PR #7262, opened 2020-04-13
Generalized shop url by creating SiteConfig field for it

jitendra1998:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Added `shop_url` field to the `SiteConfig` and replaced all occurences of "shop.dev.to" with `SiteConfig.shop_url`. 

## Related Tickets & Documents
Closes #7252 
## Added tests?

- [x] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

## [optional] Are there any post deployment tasks we need to perform?
No
## [optional] What gif best describes this PR or how it makes you feel?

alt_text

PR #7319, opened 2020-04-16
Remove leftover harcoded shop string

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Now that #7262 is in, we can remove the last hardcoded shop URL.

I also added a note explaining why we can't remove it from the routes just yet.

## Added tests?

- [x] yes
- [ ] no, because I need help


Ridhwana:
Just pinging @jitendra1998 on this PR as I think he was planning on removing the hardcoded sidebar nav element and route in a follow up PR: https://github.com/thepracticaldev/dev.to/pull/7262#issuecomment-614265100

PR #7336, opened 2020-04-16
Flaky Spec Fix: Credits Spec, Compare org updated_at's Properly 

mstruve:
- [x] Bug Fix

## Description
Looks like a little copy/pasta caused us to be comparing the org updated_at with the user updated_at which led to this spec failing intermittently. 

## Added tests?
- [x] yes

alt_text


mstruve:
LOL @rhymes yes that is intended. Its what I say when copy/paste goes awry 😂

PR #7343, opened 2020-04-16
Bug Fix: Rescue Addressable::URI::InvalidURIError and Mark Podcast Episode unreachable

mstruve:
- [x] Bug Fix

## Description
[A couple weeks ago I fixed a similar error](https://github.com/thepracticaldev/dev.to/pull/7137) but apparently there is another type of invalid error to catch. 
Fixes: https://app.honeybadger.io/fault/66984/51226b47c7edbb2ef566c681bbff2e98
 
## Added tests?
- [x] yes

alt_text


rhymes:
Ah ah so many different errors. Reminds me we should decide on using only one URI library :D

PR #7279, opened 2020-04-14
Update whitelisted params on Fastly on deployments

atsmith813:
## What type of PR is this? (check all applicable)
- [x] Feature

## Description
The [documentation](https://github.com/thepracticaldev/dev.to/pull/7279/files#diff-d626c3e4cb69aefc89b6b410c5a864d0) I added will have more details. I also annotated the code with PR comments.

**TLDR**
This PR introduces functionality to automatically update the whitelisted params in Fastly on deployments.

First, it checks if an update is necessary aka did we change any whitelisted params. If we have, it goes through an automated version of the manual process which is:
1. Clone the most recent Fastly version.
2. Update the VCL with the new params.
3. Activate the new version including the new VCL.

## Related Tickets & Documents
https://github.com/thepracticaldev/dev.to/projects/9#card-33234734

## Added tests?
- [x] yes

## Added to documentation?
- [x] docs.dev.to

## Are there any post deployment tasks we need to perform?
No more than usual, we'll want to check the logs to make sure the task ran and behaved appropriately when this deploys. I'm leaving this as a draft so I can merge, deploy, and check Fastly.

**Note:** once this is deployed, we can _**NOT**_ change the name of the VCL script in Fastly without also updating this code. We also should _**NOT**_ update the VCL directly in Fastly because our code is now the "source of truth" and it will overwrite any changes made _only_ in Fastly. I'll announce this to the team as well when I go to deploy this. We can always update the code to be "add only", but right now I don't think we need that. That would open a door to some painful bugs where the params defined in code and the params defined in Fastly (which very few people have admin access to) are different.

magic_gif

PR #7349, opened 2020-04-16
Generalize Fastly rake task

atsmith813:
- [x] Optimization

## Description
In #7279 we added the `fastly:update_whitelisted_params` task to `release-script.sh`. This will fail for other communities who don't have Fastly configured. This PR adds a guard to not run the task if Fastly isn't configured so deployments don't fail.

There's still work to be done to generalize Fastly altogether.

**P.S.**
I updated Rubocop to ignore spec names for rake tasks. It was complaining:
```
 RSpec/DescribeClass: The first argument to describe should be the class or module being tested.
```

#7279
https://github.com/thepracticaldev/dev.to/pull/7279#issuecomment-614879618
https://github.com/thepracticaldev/dev.to/projects/9#card-36459306

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

## Post deploy
@mstruve can you add the ENV variable `FASTLY_WHITELIST_PARAMS_SNIPPET_NAME`?

welcome_gif

### DIFF
--- a/Envfile
+++ b/Envfile
@@ -114,7 +114,7 @@ variable :HCTI_API_KEY, :String, default: "Optional"
 # (https://docs.fastly.com/api/)
 variable :FASTLY_API_KEY, :String, default: ""
 variable :FASTLY_SERVICE_ID, :String, default: ""
-variable :WHITELIST_PARAMS_SNIPPET_NAME, :String, default: ""
+variable :FASTLY_WHITELIST_PARAMS_SNIPPET_NAME, :String, default: ""
 
 # Honeycomb for monitoring and observability
 # (https://www.honeycomb.io/)
--- a/Envfile
+++ b/Envfile
@@ -112,8 +112,8 @@ variable :HCTI_API_KEY, :String, default: "Optional"
 
 # Fastly for edge caching
 # (https://docs.fastly.com/api/)
-variable :FASTLY_API_KEY, :String, default: "Optional"
-variable :FASTLY_SERVICE_ID, :String, default: "Optional"
+variable :FASTLY_API_KEY, :String, default: ""
+variable :FASTLY_SERVICE_ID, :String, default: ""
 
 # Honeycomb for monitoring and observability
 # (https://www.honeycomb.io/)
--- a/Envfile
+++ b/Envfile
@@ -114,6 +114,7 @@ variable :HCTI_API_KEY, :String, default: "Optional"
 # (https://docs.fastly.com/api/)
 variable :FASTLY_API_KEY, :String, default: ""
 variable :FASTLY_SERVICE_ID, :String, default: ""
+variable :WHITELIST_PARAMS_SNIPPET_NAME, :String, default: ""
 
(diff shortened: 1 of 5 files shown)


atsmith813:
cc @leewynne

lib/tasks/fastly.rake
```diff
@@ -1,6 +1,11 @@
 namespace :fastly do
   desc "Update VCL for whitelisted params on Fastly"
   task update_whitelisted_params: :environment do
+    if ApplicationConfig["FASTLY_API_KEY"].blank? || ApplicationConfig["FASTLY_SERVICE_ID"].blank?
+      puts "Fastly not configured. Please set FASTLY_API_KEY and FAASTLY_SERVICE_ID in your environment."
+      exit
```
atsmith813, line 6: IIRC using `exit` won't stop the shell script from continuing on.

citizen428: Do we maybe want to change this to a non-zero exit code so `$?` is only 0 when the task ran and completed?

rhymes: If it's a non zero exit code won't it block deployment for installations that don't have fastly?

leewynne:
Also worth noting that I have Fastly configured, just in case this for anyone that doesn't.

lib/tasks/fastly.rake
atsmith813: That's correct, non zero exit code would stop the deployment

lib/tasks/fastly.rake
```diff
@@ -1,6 +1,11 @@
 namespace :fastly do
   desc "Update VCL for whitelisted params on Fastly"
   task update_whitelisted_params: :environment do
+    if ApplicationConfig["FASTLY_API_KEY"].blank? || ApplicationConfig["FASTLY_SERVICE_ID"].blank?
```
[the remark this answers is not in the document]
atsmith813: Updated!

atsmith813:
I had to force push after rebasing master to fix the conflict.

I made some more changes. I realize it's possible to have Fastly configured, but not the custom VCL we're using. I moved the snippet name to the ENV, for now, to account for that. @mstruve if this is approved, can you help me add the new ENV variable?

Envfile
```diff
@@ -112,8 +112,9 @@ variable :HCTI_API_KEY, :String, default: "Optional"
 
 # Fastly for edge caching
 # (https://docs.fastly.com/api/)
-variable :FASTLY_API_KEY, :String, default: "Optional"
-variable :FASTLY_SERVICE_ID, :String, default: "Optional"
+variable :FASTLY_API_KEY, :String, default: ""
+variable :FASTLY_SERVICE_ID, :String, default: ""
+variable :WHITELIST_PARAMS_SNIPPET_NAME, :String, default: ""
```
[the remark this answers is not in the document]
atsmith813: Beat me to it, docs are coming!

atsmith813:
Requesting re-review after updates!

cc: @rhymes

PR #7361, opened 2020-04-17
Buttons

ludwiczakpawel:

- [x] Refactor

## Description

This PR introduces new buttons styles as well as some refactoring for buttons in general.

image

## Added tests?

## Added to documentation?

- [x] no documentation needed


app/assets/stylesheets/classified_listings.scss
```diff
@@ -255,7 +255,7 @@
       a {
         text-decoration: underline;
         display: inline-block;
-        color: var(--link-branded-color);
+        color: var(--link-brand-color);
```
ludwiczakpawel, line 258: I've changed the name of one variable, that's why there's bunch of random files affected.

app/assets/stylesheets/config/_generator.scss
```diff
@@ -413,7 +413,7 @@
   (
     'w',
     'width',
-    ('0': 0, '25': 25%, '50': 50%, '100': 100%, 'auto': auto, 'full': 100vw),
+    ('0': 0, '25': 25%, '50': 50%, '100': 100%, 'auto': auto, 'full': 100vw, 'max': max-content),
```
ludwiczakpawel, line 416: I've found it useful so i just added it shrug.gif

rhymes:
@ludwiczakpawel I regenerated the snapshots as they contained some strings in Polish. We'll eventually find a way to avoid this error in tests :D

ludwiczakpawel:
@rhymes thanks! well, I think I'll never understand and tackle tests the right way :D

lisasy:
😀Overall, this looks great 😀 I like this system you have developed and consideration for all of the use cases it addresses. 

Some of my feedback are visual design:

### Spacing gap for `icon-left + label` button seem visually too large
79571820-1f909f00-80bc-11ea-9e68-8e9668b8bc9a_png__2430×2008_

The gap between the icon and label seem unsettling, whereas the gap between the icon and label for the icon-right version seem balanced. I'm going to double-check in code to see if there's any difference, but this is something that caught my eye immediately.


### Color choices don't make these very readable
79571820-1f909f00-80bc-11ea-9e68-8e9668b8bc9a_png__2430×2008_

When looking at the button label colors for `ghost-success` and `ghost-warning` on the white background, I find it hard to pick up because of lack of contrast — I don't think there is enough saturation in these colors to have it stand out from the white background. This consideration might be addressed in another PR or other, but I wanted to bring this up here.

Also, the `ghost-danger` on the `--inverted` background is pretty hard to read.

I haven't put these values in any CSS color contrast checker, but this is what I'm noticing from glancing at the button sheet alone (and I have pretty good eyes! 🤓)

app/assets/stylesheets/components/buttons.scss
```diff
@@ -1,109 +1,241 @@
...
 
-$outlined-border-width: 2px;
-
 // Basic styling
 .crayons-btn {
+  // Setup
+  --border-width: 0px;
+  --shadow-color: rgba(0, 0, 0, 0.05);
+  --vertical-padding: #{$su-2};
+  --horizontal-padding: #{$su-4};
```
lisasy, line 9: ooooo can you explain to me what is the difference between writing `#{$su-2}` versus just `$su-2`?

app/javascript/crayons/Button/__stories__/button.html.stories.jsx
```diff
@@ -7,57 +7,97 @@ export default {
...
+  <button type="button" className="crayons-btn crayons-btn--ghost-success">
+    Ghost Success Button label
+  </button>
+);
+
+GhostSuccess.story = {
+  name: 'Ghost Success',
+};
+
+export const GhostWarning = () => (
```
lisasy, line 79: 👻<- This emoji is all I'm thinking about 😂

ludwiczakpawel:
@lisasy 

> The gap between the icon and label seem unsettling, whereas the gap between the icon and label for the icon-right version seem balanced. I'm going to double-check in code to see if there's any difference, but this is something that caught my eye immediately.

This gap will not exist in real life.. For some reason, on my demo page, there's an extra white space right there. Not sure why and not sure how to remove it... And that screenshot is coming from my demo page. Shouldn't be an issue in real life.

> When looking at the button label colors for ghost-success and ghost-warning on the white background, I find it hard to pick up because of lack of contrast — I don't think there is enough saturation in these colors to have it stand out from the white background. This consideration might be addressed in another PR or other, but I wanted to bring this up here.

Definitely another PR - buttons reuse our color palette. If we want them darker, we have to reconsider our colors palette. Also, I don't expect too many scenarios/cases where we will be using `crayons-btn-ghost-*` variants. (It's there actually for dark variant:P)

app/assets/stylesheets/components/buttons.scss
ludwiczakpawel: `$...` is a SASS variable. When SASS variable is being used in CSS function that has brackets (for example `calc()`) you need this additional syntax. Another example that uses brackets is calling CSS variables: `property: var(--vertical-padding)` and that's why you need this extra syntax here.

PR #7363, opened 2020-04-17
Just in Case: Revert "Execute Feed Searches through Elasticsearch (Round 2)"

mstruve:
Reverts thepracticaldev/dev.to#7231

Having on standby just in case!


rhymes:
LOOOL :D

PR #7366, opened 2020-04-17
Record Identity Creation Metrics

mstruve:
- [x] Feature

## Description
In order to alert us of problems with our authentication user creation flow lets record every time we create a new identity and if none are created for an extended period of time alert so we can look into it. 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/spec/services/authentication/authenticator_spec.rb
+++ b/spec/services/authentication/authenticator_spec.rb
@@ -78,6 +78,15 @@ RSpec.describe Authentication::Authenticator, type: :service do
           described_class.call(auth_payload)
         end
       end
+
+      it "records successful identity creation metric" do
+        allow(DatadogStatsClient).to receive(:increment)
+        service.call
+
+        expect(DatadogStatsClient).to have_received(:increment).with(
+          "identity.created", tags: [provider: "github"]
+        )
+      end
     end
 
     describe "existing user" do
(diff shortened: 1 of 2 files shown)


spec/services/authentication/authenticator_spec.rb
```diff
@@ -273,6 +298,13 @@
         end.not_to change(Identity, :count)
       end
 
+      it "does not record an identity creation metric" do
+        allow(DatadogStatsClient).to receive(:increment)
+        service.call
+
+        expect(DatadogStatsClient).not_to have_received(:increment)
+      end
```
vaidehijoshi, line 306: Hm, what if an existing user created a new identity? I'm thinking of if someone auths with Twitter, and then later auths with GitHub...

rhymes: cc @Zhao-Andy here as I'm not sure how that scenario works if the user has two different emails. 

If they have the same email (as I do for both twitter and github) they will end up in 

https://github.com/thepracticaldev/dev.to/blob/8340051649b38cc35ea6ec18405bc975131e0aba/app/services/authentication/authenticator.rb#L69

Zhao-Andy: If someone has two different emails, we assume it's a new user being created, so a new user (and identity) will be created. That's where we get issues with users having two accounts and then later ask for them to be merged.

I think we _have_ to assume that for security reasons.

rhymes: Great write up @Zhao-Andy, thanks!

rhymes:
Still waiting for the build to finish in the other PR - https://github.com/thepracticaldev/dev.to/pull/7365 - hopefully it won't conflict

spec/services/authentication/authenticator_spec.rb
mstruve: @vaidehijoshi this is def only tracking identity creation rather than user creation. We will then alert under a given threshold for each provider. I choose to do it this way since we don't always create new users when authenticating like this. We could add tracking for user creation and other model creations as well in the future but here I thought focusing on identity was the way to go.

PR #7367, opened 2020-04-17
Properly parse params for feed content Tag searching

mstruve:
- [x] Bug Fix

## Description
Need to properly parse arrays for feed content searches since they don't use `fetchSearch` helper

## Added tests?
- [x] no, I tested this locally, I will work on adding tests after this is out

alt_text


mstruve:
This PR makes me feel to my core we need a single search flow 🤦‍♀️

PR #7368, opened 2020-04-17
Wrap user/identity coupled logic in a transaction

vaidehijoshi:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Creating a `user` and then setting an `identity` should occur as an all-or-nothing step.
Wrapping this in an `ActiveRecord::Transaction` ensures that we don't update an identity
unless we successfully `create`/`update` the `user` first in the same database transaction.

Lol, #emergency.

N/A

## Added tests?

- [ ] yes
- [x] no, because ~I need help~ how do you...test this?

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/services/authentication/authenticator.rb
+++ b/app/services/authentication/authenticator.rb
@@ -64,8 +64,10 @@ module Authentication
       # Return the successfully-authed used from the transaction.
       authed_user
     rescue StandardError => e
-      # Notify DataDog if something goes wrong in the transaction.
+      # Notify DataDog if something goes wrong in the transaction,
+      # and then ensure that we re-raise and bubble up the error.
       DatadogStatsClient.increment("identity.errors", tags: ["error:#{e.class}"], message: e.message)
+      raise e
     end
 
     private
--- a/app/services/authentication/authenticator.rb
+++ b/app/services/authentication/authenticator.rb
@@ -66,7 +66,7 @@ module Authentication
     rescue StandardError => e
       # Notify DataDog if something goes wrong in the transaction,
       # and then ensure that we re-raise and bubble up the error.
-      DatadogStatsClient.increment("identity.errors", tags: ["error:#{e.class}"], message: e.message)
+      DatadogStatsClient.increment("identity.errors", tags: ["error:#{e.class}", "message:#{e.message}"])
       raise e
     end
 
--- a/app/services/authentication/authenticator.rb
+++ b/app/services/authentication/authenticator.rb
@@ -48,8 +48,6 @@ module Authentication
         user.save!
         user
       end
-    rescue StandardError => e
-      Rails.logger.error(e)
     end
 
     private
--- a/app/services/authentication/authenticator.rb
+++ b/app/services/authentication/authenticator.rb
@@ -26,24 +26,30 @@ module Authentication
 
     def call
       identity = Identity.build_from_omniauth(provider)
-
       return current_user if current_user_identity_exists?
 
-      user = proper_user(identity)
-      user = if user.nil?
-               build_user
-             else
-               update_user(user)
-             end
+      ActiveRecord::Base.transaction do
+        user = proper_user(identity)
+        user = if user.nil?
+                 build_user!
+               else
+                 update_user(user)
+               end
 
-      save_identity(identity, user)
+        identity.user = user if identity.user_id.blank?
+        new_record = identity.new_record?
+        identity.save!
+        record_identity_creation(identity) if new_record
 
-      user.skip_confirmation!
+        user.skip_confirmation!
 
-      flag_spam_user(user) if account_less_than_a_week_old?(user, identity)
+        flag_spam_user(user) if account_less_than_a_week_old?(user, identity)
 
-      user.save!
-      user
+        user.save!
+        user
+      end
+    rescue StandardError => e
+      Rails.logger.error(e)
     end
 
     private
@@ -70,7 +76,7 @@ module Authentication
       end
     end
 
-    def build_user
+    def build_user!
       existing_user = User.where(
         provider.user_username_field => provider.user_nickname,
       ).take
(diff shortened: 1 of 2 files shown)


app/services/authentication/authenticator.rb
```diff
@@ -70,7 +76,7 @@ def proper_user(identity)
       end
     end
 
-    def build_user
+    def build_user!
```
vaidehijoshi, line 79: Changed this method name since we call `save!` not just `save` :)

app/services/authentication/authenticator.rb
```diff
@@ -113,13 +118,6 @@ def update_profile_updated_at(user)
       user.profile_updated_at = Time.current if user.public_send(field_name)
     end
 
-    def save_identity(identity, user)
```
vaidehijoshi, line 116: Pulled this all into the transaction...I'm assuming it's okay/safe to do `record_identity_creation` from within the transaction but let me know if I've got that wrong!

app/services/authentication/authenticator.rb
```diff
@@ -26,24 +26,30 @@ def self.call(*args)
...
-      flag_spam_user(user) if account_less_than_a_week_old?(user, identity)
+        flag_spam_user(user) if account_less_than_a_week_old?(user, identity)
 
-      user.save!
-      user
+        user.save!
+        user
+      end
+    rescue StandardError => e
+      Rails.logger.error(e)
```
vaidehijoshi, line 52: I think in order to maintain parity with what this service was doing before, I have to return `user` here from direction in the transaction...but I wonder if we actually really _need_ to return it here?

Also, I just logged any errors that pop up here (this would catch any kind of validation errors from when we `save!` either the `identity` or the `user`.

mstruve: Why don't we want these errors bubbling up? Seems like the only reason to use `save!` is so we raise an error

vaidehijoshi: I guess that's a fair point, I just wasn't sure if we wanted to let it explode in that way or know about it via logging. But I think I'll just remove the logger.

app/services/authentication/authenticator.rb
rhymes: Technically no. If DataDog raises a HTTP exception the whole transaction will be rolled back, I don't think we want to block a user sign up/login if DD is down or if for any reason the connection times out. `flag_spam_user` which calls Slack asynchronously can stay there as it's correct in the flow. 

See the section "Exception handling and rolling back" in https://api.rubyonrails.org/v5.2/classes/ActiveRecord/Transactions/ClassMethods.html 

You can find an example on error handling within transaction in 

https://github.com/thepracticaldev/dev.to/blob/e0b932e569bb5bcf926f0973032761897834416d/app/services/pro_memberships/biller.rb#L32

Obviously we want to log to Datadog so I think we should have 2 DD calls: one after the transaction which will trigger if everything went well and, optionally, one in an exception handler which triggers if we get an error.

Maybe something like:

```ruby
DatadogStatsClient.increment("identity.errors", tags: ["error:#{err.class}"], message: err.message)
```

cc @mstruve for the best way to log these inline errors

app/services/authentication/authenticator.rb
rhymes: we can also call it `find_existing_user_or_create_it!` or something like that. Not blocking obviously :D

vaidehijoshi:
@rhymes Ready for re-review here, specifically the [last commit](https://github.com/thepracticaldev/dev.to/pull/7368/commits/2d1e7484fa3d0416cbea0980cf109c7e8f750b9f).

Fair warning: I could use some help with the specs (the new ones currently fail for some reason I cannot figure out!). I spent quite some time this morning trying to understand why the DataDog service doesn't `increment` in the way that I expect when I mock a `StandardError`...but I didn't want to spend too much more time on this without getting a second pair of eyes from someone.

vaidehijoshi:
Sigh, I'm going to set this aside for today because now a different spec is failing and I'm pretty tired and probably am not going to be able to figure it out. 😖 

cc'ing @rhymes in case you have any idea why `./spec/models/user_spec.rb:889` could be failing in this way (I know that the transaction work touches the authentication stuff, but why would we no longer be getting a validation error raised once putting the same logic into a transaction??).

app/services/authentication/authenticator.rb
```diff
@@ -26,24 +26,46 @@ def self.call(*args)
...
+        DatadogStatsClient.increment("identity.created", tags: [provider: id_provider])
+      end
 
-      user.save!
-      user
+      # Return the successfully-authed used from the transaction.
+      authed_user
+    rescue StandardError => e
+      # Notify DataDog if something goes wrong in the transaction.
+      DatadogStatsClient.increment("identity.errors", tags: ["error:#{e.class}"], message: e.message)
```
rhymes, line 68: I think we need to re-raise the exception after logging to DD, otherwise the user will never now what went wrong, correct? This is also why the test fails

rhymes: I would also add a test in the authenticator for that

vaidehijoshi: Done!

app/services/authentication/authenticator.rb
```diff
@@ -26,24 +26,48 @@ def self.call(*args)
...
+      end
 
-      user.save!
-      user
+      # Return the successfully-authed used from the transaction.
+      authed_user
+    rescue StandardError => e
+      # Notify DataDog if something goes wrong in the transaction,
+      # and then ensure that we re-raise and bubble up the error.
+      DatadogStatsClient.increment("identity.errors", tags: ["error:#{e.class}"], message: e.message)
```
mstruve, line 69: It looks like [the only additional data you can send with increment is tags](https://docs.datadoghq.com/developers/metrics/dogstatsd_metrics_submission/?tab=ruby#count) so we should either add the message to the tags or remove it.

vaidehijoshi: OH, great point! I'm going to just add it to the tags 👍

benhalpern:
This looks good to me. I think we should hold off merging this until prime US hours tomorrow just in case there is anything unexpected about it in prod.

PR #7369, opened 2020-04-17
Create onboarding task card

maestromac:
- [x] Feature

## Description
Add an Onboarding task card that will appear after the user finish onboarding. The rendering logic works as follows
- It will only appear for user created within 7 days ago.
- It will disappear after clicking "x" button on and never appear again.
- It uses localStorage to store the state, so this is browser and device-specific.

Resolves https://github.com/thepracticaldev/dev.to/issues/6552
Kapture 2020-04-17 at 16 08 32

## Added tests?
- [ ] WIP

## Added to documentation?
- [x] no documentation needed

n/a


vaidehijoshi:
Hm, I'm not seeing the task card when I try to QA locally...I know you said that it is browser/device specific...am I missing something?
Screen Shot 2020-04-20 at 3 33 39 PM

app/assets/images/staggered-dev.svg
```diff
@@ -0,0 +1,13 @@
...
+  </g>
+  <defs>
+    <clipPath id="clip0">
+      <rect width="106" height="103" fill="white" />
+    </clipPath>
+  </defs>
+</svg>
```
vaidehijoshi, line 13: we should optimize this svg (with `svgo`?) if it hasn't been optimized already

app/assets/javascripts/initializers/initializeOnboardingTaskCard.js.erb
```diff
@@ -0,0 +1,9 @@
+function initializeOnboardingTaskCard() {
```
vaidehijoshi, line 1: could use some documentation here :)

app/assets/javascripts/initializers/initializeOnboardingTaskCard.js.erb
```diff
@@ -0,0 +1,9 @@
...
+  if (localStorage.getItem('task-card-closed') === 'yes') { return }
+
+  var taskCard = document.getElementsByClassName("onboarding-task-card")[0];
+
+  if (taskCard !== undefined) {
```
vaidehijoshi, line 6:
```suggestion
  if (taskCard) {
```
If we find a `taskCard` by class name in the previous line, then is this should work, right? IIRC, `undefined` is falsy in JS, so even `null` or an empty string would pass this check? 🤔

app/assets/stylesheets/onboarding.scss
```diff
@@ -0,0 +1,66 @@
+@import 'variables';
+@import 'mixins';
+@import 'config/import';
+
+.onboarding-task-card {
+  background: #3b49df;
```
vaidehijoshi, line 6: are there any global colors that already work for this? if there is, we should use that instead (i suspect there is?).

if not, let's use a `$` prefixed variable at the top of this class, even if it's just in the context of this one file.

app/assets/stylesheets/onboarding.scss
```diff
@@ -0,0 +1,66 @@
...
+  border-radius: $br-default;
+  padding: $su-7 $su-8 $su-8 $su-8;
+  display: none;
+
+  .task-card-title {
+    font-weight: $fw-heavy;
+    font-size: $fs-4xl;
+    line-height: $lh-tight;
+    color: #ffffff;
```
vaidehijoshi, line 15: let's use a global variable here since this is just white and we should have white available in crayons.

if i remember correctly it should be: `var(--base-inverted)`, but you can double check in `app/assets/stylesheets/config/_colors.scss`!

app/assets/stylesheets/onboarding.scss
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  .task-card-subtitle {
+    color: #ffffff;
```
vaidehijoshi, line 19: use crayons white

app/assets/stylesheets/onboarding.scss
```diff
@@ -0,0 +1,66 @@
...
+    opacity: 0.7;
+    font-weight: $fw-medium;
+    font-size: $fs-s;
+    line-height: $lh-tight;
+    letter-spacing: -0.02em;
+  }
+
+  .task-card-action {
+    border-radius: $br-default;
+    background-color: rgba(255, 255, 255, 0.2);
```
vaidehijoshi, line 29: i feel like we should be using a crayons color here, too? i'm sure either @ludwiczakpawel or @lisasy could confirm if crayons handles opacity. if we absolutely _have_ to use `rgba` here, i think it warrants a comment. feels weird to see this in the file given the fact that we're defaulting to using hex or crayons...

app/assets/stylesheets/onboarding.scss
```diff
@@ -0,0 +1,66 @@
...
+  a.task-card-link {
+    display: block;
+    border-radius: $br-default;
+    padding: $su-5 $su-5;
+    margin-top: $su-4;
+    font-weight: $fw-bold;
+    font-size: $fs-base;
+    line-height: $lh-base;
+    letter-spacing: -0.02em;
+    color: #ffffff;
```
vaidehijoshi, line 45: 🖍️ color here

app/assets/stylesheets/onboarding.scss
```diff
@@ -0,0 +1,66 @@
...
+
+    .emoji {
+      margin-right: $su-2;
+    }
+  }
+
+  .close {
+    background: none;
+    border: none;
+    position: absolute;
```
vaidehijoshi, line 62: are these styles overriding something else? i can't try this out locally so i don't know for sure...but they seem to be "unsetting" some other style. 

if so, we should make a comment.

app/views/articles/index.html.erb
```diff
@@ -36,6 +36,11 @@
 
     <%# BEGIN Feed menu bar %>
     <div class="articles-list" id="articles-list">
+
```
vaidehijoshi, line 39: ✂️ empty new line

app/views/onboardings/_task_card.html.erb
```diff
@@ -0,0 +1,43 @@
+<script>
+  function closeTaskCard() {
```
vaidehijoshi, line 2: i think it would be helpful to include a comment that points to `initializeOnboardingTaskCard` so that it's clear where the corresponding JS lives.

app/views/onboardings/_task_card.html.erb
```diff
@@ -0,0 +1,43 @@
...
+    <svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
+      <path d="M6.99974 5.58623L11.9497 0.63623L13.3637 2.05023L8.41374 7.00023L13.3637 11.9502L11.9497 13.3642L6.99974 8.41423L2.04974 13.3642L0.635742 11.9502L5.58574 7.00023L0.635742 2.05023L2.04974 0.63623L6.99974 5.58623Z" fill="white" />
+    </svg>
+  </button>
+
+  <div>
+    <%= inline_svg_tag("staggered-dev.svg", aria: true, title: "staggered dev stickers") %>
+  </div>
+
+  <h1 class="task-card-title">You're now apart of the community</h1>
```
vaidehijoshi, line 21:
```suggestion
  <h1 class="task-card-title">You're now a part of the community!</h1>
```
the `!` is a suggestion; i'm not sure what the actual copy says but we should include some ending punctuation here

app/views/onboardings/_task_card.html.erb
```diff
@@ -0,0 +1,43 @@
...
+  <p class="task-card-subtitle">SUGGESTED THINGS YOU CAN DO</p>
+  <div class="task-card-action">
+    <a class="task-card-link" href="/welcome"><p><span class="emoji">😊</span>Join the Welcome thead</p><svg width="8" height="14" viewBox="0 0 8 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
vaidehijoshi, line 24: curious why we have some svgs inline and some that we render with the `inline_svg_tag`? i see we use that on line 18, so just wondering what the difference/discrepancy is here

app/views/onboardings/_task_card.html.erb
```diff
@@ -0,0 +1,43 @@
...
+        <path d="M5.17168 7.00023L0.22168 2.05023L1.63568 0.63623L7.99968 7.00023L1.63568 13.3642L0.22168 11.9502L5.17168 7.00023Z" fill="white" />
```
vaidehijoshi, line 25: why `fill="white"` here instead of a hex color?

app/views/onboardings/_task_card.html.erb
```diff
@@ -0,0 +1,43 @@
...
+      </svg>
+    </a>
+  </div>
+
+  <div class="task-card-action">
+    <a class="task-card-link" href="/new"><p><span class="emoji">✍🏾</span>Write your first <%= community_name %> post</p><svg width="8" height="14" viewBox="0 0 8 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
vaidehijoshi, line 31: > `<%= community_name %>`

YAY for generalization!

app/views/onboardings/_task_card.html.erb
juliannatetreault: I believe the copy includes an exclamation point if I'm not mistaken, so good catch @vaidehijoshi . 😄

juliannatetreault:
> Hm, I'm not seeing the task card when I try to QA locally...I know you said that it is browser/device specific...am I missing something?
> Screen Shot 2020-04-20 at 3 33 39 PM

I didn't see the task card initially either but after completing onboarding, I dropped into the `rails console` and updated when I had been `created_at` (since users have to be 7 days old or less) and set it to be equal to now and then saw the task card. Hopefully that helps your issue as well @vaidehijoshi 
(???)! 😊

vaidehijoshi:
> but after completing onboarding, I dropped into the rails console and updated when I had been created_at (since users have to be 7 days old or less) and set it to be equal to now and then saw the task card. 

@juliannatetreault good call! i totally forgot about that while QAing. sadly, even though i updated my user's `created_at` to 3 days ago, i'm still seeing the old closing slide and not the new task card :(

juliannatetreault:
Quick clarification: The individual tasks on the task card are supposed to remain in place even after a task has been completed, correct?

And just to make sure I'm on the same page as you 😅 : As it stands, the task card will be displayed for the first seven days that a user has been created and then only after the user has been created for 7+ days or the user clicks the "x" to close the task card, the task card will be removed, correct?

juliannatetreault:
> > but after completing onboarding, I dropped into the rails console and updated when I had been created_at (since users have to be 7 days old or less) and set it to be equal to now and then saw the task card.
> 
> @juliannatetreault good call! i totally forgot about that while QAing. sadly, even though i updated my user's `created_at` to 3 days ago, i'm still seeing the old closing slide and not the new task card :(

Hmm, I also see the old closing slide, but I was able to see and QA the new task card after completing the onboarding process (clicking the old closing slide to end the onboarding process and get to the feed to see the task card was necessary, though), but I just went through the onboarding process _again_ and this time updated my user's `created_at` time to be four days ago and did not see the task card. 🤔

app/views/articles/index.html.erb
```diff
@@ -36,6 +36,11 @@
...
+      <% if user_signed_in? && current_user.created_at > 7.days.ago %>
```
benhalpern, line 40: Reminder that we don't have access to `current_user` in this context. You'd probably need to hide this HTML and show it via `userData()` in the JS if the user qualifies for this condition.

app/views/onboardings/_task_card.html.erb
maestromac: My reasoning is I'd like to avoid adding physical file asset if it's not a big svg. I created `staggered-dev.svg` because it was taking up too many lines.

app/assets/javascripts/initializers/initializeOnboardingTaskCard.js.erb
maestromac: You are right that would work! I got bitten by a null safety before which led me to do this janky alternative.

app/assets/javascripts/initializers/initializeOnboardingTaskCard.js.erb
maestromac: Added!

app/assets/stylesheets/onboarding.scss
maestromac: I didn't use the crayon var at first because I was grepping for `#ffffff` 😆 . I didn't know that this is practically the same. Updated

app/assets/stylesheets/onboarding.scss
maestromac: I initially had concerns with this too then I notice the figma spec provided `background-color: rgba(fff,fff,fff,0.2)` as a guideline. I tweak `fff` to 255 because it wasn't valid.

This on-hover effect can't be produced with combination use of hex and opacity because the opacity will affect the text within it.

app/views/onboardings/_task_card.html.erb
maestromac: Added!

app/assets/stylesheets/onboarding.scss
maestromac: Oops left this out. Thanks!

app/views/onboardings/_task_card.html.erb
vaidehijoshi: I think we should use a hex code here; I suspect figma just doesn't default to that, for whatever reason. Same for these other svgs in this file coming from figma.

app/assets/stylesheets/onboarding.scss
vaidehijoshi: Okay, let's just leave a comment about it then so that someone doesn't accidentally change this in the future.

app/assets/images/staggered-dev.svg
maestromac: TIL! updated! 21% reduction

app/assets/images/staggered-dev.svg
```diff
@@ -0,0 +1 @@
...358 2.56-1.422 2.42zm6.14 3.606c-.807-.474-1.121-1.418-.942-2.832l3.82-13.71 2.663 1.563-3.05 10.738 7.892-7.894 2.663 1.564L80.615 52.41c-1.148.846-2.125 1.031-2.932.557z" fill="#fff"/><rect x="2.491" y="59.013" width="48" height="48" rx="3" transform="rotate(-32.667 2.491 59.013)" fill="#08090A"/><path d="M32.575 72.4c.707 1.104.932 2.24.677 3.408-.253 1.159-.957 2.108-2.11 2.847l-3.706 2.377-8.486-13.236 3.781-2.425c1.103-.707 2.236-.93 3.398-.671 1.165.25 2.101.925 2.808 2.028l3.637 5.673zm-2.364 1.516l-3.637-5.672c-.303-.473-.687-.725-1.152-.756-.468-.038-.86.045-1.175.247l-1.419.909 5.456 8.509 1.418-.91c.315-.202.556-.52.721-.956.162-.441.091-.898-.212-1.37zm12.437-2.638l-4.964 3.183c-.472.303-.967.398-1.484.285-.518-.114-.928-.406-1.23-.88l-6.365-9.926c-.304-.473-.399-.968-.285-1.485.113-.517.405-.927.878-1.23l4.963-3.183 1.516 2.364-4.254 2.728 1.97 3.072 2.6-1.667 1.515 2.364-2.6 1.667 1.97 3.072 4.255-2.727 1.515 2.363zm5.994-3.843c-.788.505-1.772.358-2.952-.442l-10.498-9.611 2.6-1.667 8.195 7.579-3.468-10.61 2.6-1.667 4.353 13.551c.235 1.406-.042 2.362-.83 2.867z" fill="#fff"/></g><defs><clipPath id="clip0"><path fill="#fff" d="M0 0h106v103H0z"/></clipPath></defs></svg>
```
vaidehijoshi, line 1: I think we should also make sure to format this correctly (svgo sadly screws up formatting which makes the svg elements hard to read!)

app/assets/stylesheets/onboarding.scss
vaidehijoshi: tried this out locally and it looks like they are necessary to style the button, so you can ignore this comment!

app/assets/images/staggered-dev.svg
vaidehijoshi: oh yeah, it's amazing! gotta save space where you can 😅

app/assets/stylesheets/onboarding.scss
maestromac: Yeah, they are overriding buttons. Added!

app/assets/images/staggered-dev.svg
maestromac: There's a neat flag for this, `svgo file.svg --pretty`

app/assets/stylesheets/onboarding.scss
maestromac: Added

app/views/onboardings/_task_card.html.erb
maestromac: I applied svgo here and it updated for me 😄

maestromac:
> Quick clarification: The individual tasks on the task card are supposed to remain in place even after a task has been completed, correct?
> 
> And just to make sure I'm on the same page as you 😅 : As it stands, the task card will be displayed for the first seven days that a user has been created and then only after the user has been created for 7+ days or the user clicks the "x" to close the task card, the task card will be removed, correct?

@juliannatetreault yes & yes. The issue specified that if any of the tasks is clicked, the task card should disappear. I find this a bit problematic as the user would not expect the task card to disappear if they didn't click on `x`.

juliannatetreault:
> > Quick clarification: The individual tasks on the task card are supposed to remain in place even after a task has been completed, correct?
> > And just to make sure I'm on the same page as you 😅 : As it stands, the task card will be displayed for the first seven days that a user has been created and then only after the user has been created for 7+ days or the user clicks the "x" to close the task card, the task card will be removed, correct?
> 
> @juliannatetreault yes & yes. The issue specified that if any of the tasks is clicked, the task card should disappear. I find this a bit problematic as the user would not expect the task card to disappear if they didn't click on `x`.

Okay, gotcha! Thank you so much for the clarification.

I see what you mean about it being a bit problematic. 🤔 I wish we could instead have the individual tasks be removed from the task card after they've been completed, e.g. after a user comments on a welcome thread, the "Join the Welcome Thread" task would disappear, but the others would remain. _But_, maybe this is something we can think about implementing for future iterations. 😄

app/views/articles/index.html.erb
maestromac: Ah, thanks for this. I was misled by the use of `user_signed_in?` in another part of the file thinking that it's okay to use this here.

app/assets/javascripts/initializers/initializeOnboardingTaskCard.js.erb
```diff
@@ -8,7 +8,13 @@ function initializeOnboardingTaskCard() {
 
   var taskCard = document.getElementsByClassName("onboarding-task-card")[0];
 
-  if (taskCard) {
+  if (taskCard == null) { return }
```
vaidehijoshi, line 12:
```suggestion
  if (taskCard === null) { return }
```

app/assets/javascripts/initializers/initializeOnboardingTaskCard.js.erb
```diff
@@ -8,7 +8,13 @@ function initializeOnboardingTaskCard() {
...
+
+  var createdAt = new Date(userData().created_at);
+  var now = new Date();
+  var aWeekAgo = now.setDate(now.getDate() - 7);
+
+  if (createdAt > aWeekAgo) {
```
vaidehijoshi, line 18: We should include something in the documentation for this function that explains the "one week ago" logic and changing the style on the card accordingly.

maestromac: Yeah I actually updated this before you reviewed https://github.com/thepracticaldev/dev.to/pull/7369/commits/236906112fc00379657932befe38b3bd6e485af2 🤔

vaidehijoshi: weird. i blame github!!

app/assets/javascripts/initializers/initializeOnboardingTaskCard.js.erb
maestromac: I should've left a note that I did this on purpose. This actually catches both [undefined and null](https://stackoverflow.com/a/2647888/5494108).

lisasy:
LGTM! :) :)

app/views/onboardings/_task_card.html.erb
lisasy: Woah! Is this going to make this extensible to when we do have more Forem communities?

maestromac:
⚠️ Please do not merge. I need to prep for 2nd portion of this first ⚠️

leewynne:
Can we generalise the images? Maybe use the mascot from siteconfig?

PR #7375, opened 2020-04-18
Fix podcast pagination client/API mismatch

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
The current podcast pagination is not working due to a change in expectations about ingestion because it was switched for the search context. This fixes it.

Also fixed a nagging css issue with podcast header.


benhalpern:
@rhymes yeah you're right. Added `type_of` back in.

PR #7378, opened 2020-04-18
Improve docker setup script

### LINKED ISSUE #7060: Investigate Elasticsearch on Gitpod
While working on #7040/#7059 I encoutered the following issue on Gitpod:

> Elasticsearch::Transport::Transport::Errors::BadRequest: [400] {"error":{"root_cause":[{"type":"invalid_alias_name_exception","reason":"Invalid alias name [users_development_alias], an index exists with the same name as the alias","index_uuid":"xDhWvsDqQnGuJtTtNcdvxA","index":"users_development_alias"}],"type":"invalid_alias_name_exception","reason":"Invalid alias name [users_development_alias], an index exists with the same name as the alias","index_uuid":"xDhWvsDqQnGuJtTtNcdvxA","index":"users_development_alias"},"status":400}

I tried to see if I could find a quick solution, but didn't, so it will have to wait until next week.

dtinth:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

I tried setting up the project according to the **[Installing DEV with Docker](https://docs.dev.to/installation/docker/)** docs page. The first issue I found that the app server failed to start. Upon closer inspection, I found that the **Setting up database** failed, but the setup script marched on anyways.

The first thing I did was to add `set -e` to the setup script, so that the script fail early, making it easier to inspect.

The error message when Setting up database is:

    Could not resolve host: elasticsearch

But the console output log did not show any relevant information. After a bit of fiddling, I figured out I had to run `docker-compose logs` to see what error occurred during Elasticsearch’s launch (it is not shown when running `docker-compose run`).

So, I added some error handling message to inform the developer to also check Docker logs.

In Elasticsearch’s logs, it said:

    Permission denied

Basically, the UID of the process inside Elasticsearch container does not match the the host’s. So I added a step to properly `chown` the data folder.

I ran the script again, but the DB setup script seems to have problems with partially-set up database. So I added an extra message to tell the developer to run `docker-compose down -v` to clean up all volumes.

I ran the script again, but the DB setup script failed with

    elasticsearch port 9200: Connection refused

I checked the Elasticsearch logs but this time it ran fine. The DB setup script runs too fast and begins setting up Elasticsearch before it is ready to serve requests.

To fixed this by using the [wait-on](https://www.npmjs.com/package/wait-on) package to wait until Elasticsearch is ready before beginning the DB setup.

I ran the script again, but I encountering the issue described in #7060.

    Invalid alias name [users_development_alias], an index exists with the same name as the alias

I tried to fix this, but it seems to keep happening. Until I figured out that `docker-compose down -v` does not delete bind-mount volumes! I had to also run `sudo rm -rf docker_data`. I updated the error message to inform the developer to do that as well.

With that out of the way, I continue inspecting until I found that the DB setup script adds some documents to the Elasticsearch index, even before `search:setup` is run. Since it is not yet run, automatic index creation is not disabled yet, causing the index to pre-exist.

The final fix is to swap the order of the `rails run` so that `search:setup` comes first.

This also fixes #7060 but only when running inside Docker.

image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed
- [x] error message

nyah

cool

### DIFF
--- a/bin/docker-setup
+++ b/bin/docker-setup
@@ -20,7 +20,7 @@ on_db_setup_failed() {
   echo "    docker-compose down -v && sudo rm -rf docker_data" >&2
   exit $1
 }
-docker-compose run web rails search:setup db:setup db:migrate data_updates:run || on_db_setup_failed $?
+docker-compose run web sh -c 'yarn global add wait-on && wait-on -t 30000 http-get://elasticsearch:9200/ && rails search:setup db:setup db:migrate data_updates:run' || on_db_setup_failed $?
 
 echo "== Starting app =="
 docker-compose up
--- a/bin/docker-setup
+++ b/bin/docker-setup
@@ -20,7 +20,7 @@ on_db_setup_failed() {
   echo "    docker-compose down -v && sudo rm -rf docker_data" >&2
   exit $1
 }
-docker-compose run web sh -c 'yarn wait-on -t 30000 http-get://elasticsearch:9200/ && rails db:setup db:migrate search:setup data_updates:run' || on_db_setup_failed $?
+docker-compose run web sh -c 'yarn wait-on -t 30000 http-get://elasticsearch:9200/ && rails search:setup db:setup db:migrate data_updates:run' || on_db_setup_failed $?
 
 echo "== Starting app =="
 docker-compose up
--- a/bin/docker-setup
+++ b/bin/docker-setup
@@ -20,7 +20,7 @@ on_db_setup_failed() {
   echo "    docker-compose down -v && sudo rm -rf docker_data" >&2
   exit $1
 }
-docker-compose run web sh -c 'yarn wait-on -t 30000 http-get://elasticsearch:9200/ && rails search:setup db:setup db:migrate data_updates:run' || on_db_setup_failed $?
+docker-compose run web rails search:setup db:setup db:migrate data_updates:run || on_db_setup_failed $?
 
 echo "== Starting app =="
 docker-compose up
--- a/bin/docker-setup
+++ b/bin/docker-setup
@@ -20,7 +20,7 @@ on_db_setup_failed() {
   echo "    docker-compose down -v" >&2
   exit $1
 }
-docker-compose run web rails db:setup db:migrate search:setup data_updates:run || on_db_setup_failed $?
+docker-compose run web sh -c 'yarn wait-on -t 30000 http-get://elasticsearch:9200/ && rails db:setup db:migrate search:setup data_updates:run' || on_db_setup_failed $?
 
 echo "== Starting app =="
 docker-compose up
--- a/bin/docker-setup
+++ b/bin/docker-setup
@@ -5,8 +5,22 @@ set -e
 echo "== Building Docker image (this may take a while) =="
 docker-compose build
 
+echo "== Ensuring Elasticsearch data directory is owned by the correct uid =="
+docker-compose run elasticsearch chown elasticsearch:elasticsearch /usr/share/elasticsearch/data
+
 echo "== Setting up database =="
-docker-compose run web rails db:setup db:migrate search:setup data_updates:run
+on_db_setup_failed() {
+  echo "Database setup failed with exit code $1." >&2
+  echo "" >&2
+  echo "Check the Docker logs using:" >&2
+  echo "    docker-compose logs" >&2
+  echo "" >&2
+  echo "Once you fixed the error, try running the script again." >&2
+  echo "To also delete all data, run this before retrying:" >&2
+  echo "    docker-compose down -v" >&2
+  exit $1
+}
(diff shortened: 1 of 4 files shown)


bin/docker-setup
```diff
@@ -1,10 +1,26 @@
...
+  echo "" >&2
+  echo "Check the Docker logs using:" >&2
+  echo "    docker-compose logs" >&2
+  echo "" >&2
+  echo "Once you fixed the error, try running the script again." >&2
+  echo "To also delete all data, run this before retrying:" >&2
+  echo "    docker-compose down -v && sudo rm -rf docker_data" >&2
+  exit $1
+}
+docker-compose run web sh -c 'yarn wait-on -t 30000 http-get://elasticsearch:9200/ && rails search:setup db:setup db:migrate data_updates:run' || on_db_setup_failed $?
```
citizen428, line 23: First off, hi Thai, fancy seeing you here 😄 and thanks for the help! 🙏  

I'm a bit reluctant to add the `wait-on` package for this, as Docker Compose supports something similar natively ([healthcheck documentation](https://docs.docker.com/compose/compose-file/#/healthcheck#healthcheck). Containers that depend on Elasticsearch running, can declare the dependency like this:

```yaml
    depends_on:
      elasticsearch:
        condition: service_healthy
```

For the Elasticsearch container you can then define a health check like this:

```yaml
    healthcheck:
        test: ["CMD", "curl", "-f", "http://localhost:3000"]
        interval: 30s
        timeout: 30s
        retries: 3
```

See [here](https://docs.docker.com/engine/reference/builder/#healthcheck) for more informationr regarding interval, timeout and retries.

dtinth: @citizen428 Hey Michael! Thanks for your review, I just learned about healthcheck support just now. Never knew it existed before (I depend on npm for most things nowadays :joy:)

@maestromac But it works on my machine™! Guess we did not get what Docker promised here… 😅 Jokings aside, I was able to replicate this when trying to run the build on GitHub Actions. I will use healthcheck as @citizen428 suggested instead. Thanks for your review.

dtinth: @citizen428 I checked, but unfortunately [`condition` is no longer supported](https://docs.docker.com/compose/compose-file/#depends_on)…

> Version 3 no longer supports the condition form of `depends_on`.

[docker-compose’s docs](https://docs.docker.com/compose/startup-order/) suggests users to “Use a tool such as wait-for-it, dockerize, or sh-compatible wait-for”. So I am bringing back `wait-on` for now. Instead of adding it as project dependency, I made the DB setup install `wait-on` and wait for it ad-hoc instead.

dtinth: @maestromac 

I get it now… This happens because the `web` container has this in `docker-compose.yml`:

```
    volumes:
      - .:/usr/src/app
```

Now, `node_module` exists _inside_ the container, but not inside the host. That is why you get the command not found error. Now, why did I not get that error? I ran `yarn add wait-on` outside, so my local machine has `node_modules`.

This is fixed by adding another entry. It will cause `node_modules` to go into a separate volume with the initial contents copied from inside the container. (Thanks [StackOverflow](https://stackoverflow.com/a/32785014/559913)!)

```diff
     volumes:
       - .:/usr/src/app
+      - /usr/src/app/node_modules
```

citizen428: @dtinth Ah, sorry, I missed that they removed `depends_on`. The new solution seems good.

> But it works on my machine™! Guess we did not get what Docker promised here… 😅 

This is a depressingly common occurence for me. 🙄

bin/docker-setup
```diff
@@ -1,10 +1,26 @@
...
+  echo "" >&2
+  echo "Check the Docker logs using:" >&2
+  echo "    docker-compose logs" >&2
+  echo "" >&2
+  echo "Once you fixed the error, try running the script again." >&2
+  echo "To also delete all data, run this before retrying:" >&2
+  echo "    docker-compose down -v && sudo rm -rf docker_data" >&2
+  exit $1
+}
+docker-compose run web sh -c 'yarn global add wait-on && wait-on -t 30000 http-get://elasticsearch:9200/ && rails search:setup db:setup db:migrate data_updates:run' || on_db_setup_failed $?
```
citizen428, line 23: If this turns out to take too long we can go back to adding it as a dev dependency.

citizen428:
@maestromac You are the single most unlucky Docker user I've encountered I think 😆 

@dtinth Note that we recently merged some changes regarding the yarn integrity check (#7478), so you may want to update this PR.

maestromac:
@citizen428 I honestly think so too lol. Anyway, don't let me be the blocker for this PR. I would like a different team member to review this.

Issue #3739, opened 2019-08-18
Refactor CodeClimate Issues


benhalpern:
We have CodeClimate hooked up to our repo. It identifies code smells, which can be low-hanging fruit for improving the application.

Issues can be found here:

https://codeclimate.com/github/thepracticaldev/dev.to/issues

You can sort by JavaScript and Ruby and other filters.

AnkitaSinghal:
I would like to work on this issue

jessleenyc:
Awesome. Just a note here that this is an evergreen issue -- there will (almost certainly) always be CodeClimate issues up for grabs so anyone should feel free to submit a PR here!

FeruzOripov:
Hi, I've just made my first PR. But I got Travis CI failure. I can't figure out the reason. Can someone review my code, #3787 . Thanks )

FeruzOripov:
Here another one #3793

FeruzOripov:
#3794

Aswathprabhu:
Hi, I've made my [first PR](https://github.com/thepracticaldev/dev.to/pull/4104). Can anyone review it?

Aswathprabhu:
Hi I've made my [first PR](https://github.com/thepracticaldev/dev.to/pull/4107), Can anyone review it?

Aswathprabhu:
Hi I've made some PRs ([4154](https://github.com/thepracticaldev/dev.to/pull/4154))([4156](https://github.com/thepracticaldev/dev.to/pull/4156))([4157](https://github.com/thepracticaldev/dev.to/pull/4157))([4158](https://github.com/thepracticaldev/dev.to/pull/4158))([4159](https://github.com/thepracticaldev/dev.to/pull/4159)) I don't know why for some of them, Travis CI fails. I can't figure it out, Can anyone check on them?

hoangvvo:
Hey, happy weekend everyone.

I need help with https://github.com/thepracticaldev/dev.to/pull/4241#issuecomment-538683355 if anyone is available.

nickytonline:
There are issues in regards to OrgSettings.jsx in Code Climate. I've started work on this. This is not the same work being done in #2470 in regards to eslint issues that @ruffle1986 is addressing.

hoangvvo:
If someone can help me review #4241, it would be great. Thanks!

redxtech:
I've just submitted two PRs in order to address some CodeClimate issues - #4442 & #4443.

netochaves:
[...]

PR #7381, opened 2020-04-19
Create Articles builder and refactor articles_controller

renanbona:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

#3739 

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

alt_text

### DIFF
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -34,34 +34,13 @@ class ArticlesController < ApplicationController
 
   def new
     base_editor_assigments
-    article_builder = Articles::Builder.new(current_user&.id, @user&.name, @tag, @prefill)
 
-    if @tag.present? && @user&.editor_version == "v2"
+    @article, needs_authorization = Articles::Builder.new(@user, @tag, @prefill).build
+
+    if needs_authorization
       authorize Article
-
-      @article = article_builder.tag_user_editor_v2
-    elsif @tag&.submission_template.present? && @user
-      authorize Article
-
-      @article = article_builder.tag_user
-    elsif @prefill.present? && @user&.editor_version == "v2"
-      authorize Article
-
-      @article = article_builder.prefill_user_editor_v2
-    elsif @prefill.present? && @user
-      authorize Article
-
-      @article = article_builder.prefill_user
-    elsif @tag.present?
-      skip_authorization
-
-      @article = article_builder.tag
     else
       skip_authorization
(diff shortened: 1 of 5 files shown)


benhalpern:
Approach looks good to me

app/controllers/articles_controller.rb
```diff
@@ -34,58 +34,35 @@ def feed
...
-                   Article.new(
-                     body_markdown: "---\ntitle: \npublished: false\ndescription: \ntags: \n---\n\n",
-                     processed_html: "",
-                     user_id: current_user&.id,
-                   )
-                 end
-               end
+    article_builder = Articles::Builder.new(current_user&.id, @user&.name, @tag, @prefill)
+
+    if @tag.present? && @user&.editor_version == "v2"
```
citizen428, line 39: I wonder if this should move into the builder too? To me it'd make more sense to call something like 

```ruby
@article = Articles::Builder.new(current_user&.id, @user&.name, @tag, @prefill).build
```

or something and be done with it, rather than having this functionality split between builder and controller. 

You can manually call the policy with something like

```ruby
ArticlePolicy.new(current_user, @article).new?
```

inside the builder or return two values, e.g. 

```ruby
@article, needs_authorization = Articles::Builder.new(current_user&.id, @user&.name, @tag, @prefill).build
```

whichever you like better.

renanbona: Good call. I'll do that.

app/services/articles/builder.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    def tag_user_editor_v2
+      submission_template = @tag.submission_template_customized(@user.name).to_s
+
+      Article.new(
+        body_markdown: submission_template.split("---").last.to_s.strip,
+        cached_tag_list: @tag.name,
+        processed_html: "",
+        user_id: @user.id,
+        title: submission_template.split("title:")[1].to_s.split("\n")[0].to_s.strip,
```
marclerodrigues, line 48: You could move this to a function:
```ruby
LINE_BREAK = "\n"

def normalized_text(source, spĺit_pattern)
  text = source.split(split_pattern)[1]
  text.to_s.split(LINE_BREAK).first.to_s.strip
end
```

Issue #4545, opened 2019-10-22
Copy-to-clipboard not working from sidebar


VarunBarad:
**Describe the bug**
The "copy post link to clipboard" from the side bar is not working.

Tested on:
- Desktop (works)
- Mobile
- Android app

**To Reproduce**
1. Go to [https://dev.to/aleccool213/how-to-make-money-from-your-software-blog-without-hosting-ads-cjn](https://dev.to/aleccool213/how-to-make-money-from-your-software-blog-without-hosting-ads-cjn) (any article should work)
2. Click on `...` on the left-hand sidebar
3. That will open a menu
4. Click on the Article link presented there
5. It should copy the article link to clipboard but doesn't

**Expected behavior**
The url for the article shown should be copied to the clipboard.

**Current behavior**
The clipboard isn't affected at all.

**Smartphone (please complete the following information):**
- Device: OnePlus 6
- OS: Android Pie (9)
- Browser:
  - Firefox - Version 68.1.1
  - Google Chrome - Version 77.0.3865.116

**Additional context**
I tested it and it works on Google Chrome (Version - 77.0.3865.120) on Microsoft Windows.
**Edit:** For some reason it started working on Firefox on Windows after a PC restart 🤷‍♂ 

Following are related issues and PRs I found across this and the android app repository.
- [https://github.com/thepracticaldev/DEV-Android/issues/58](https://github.com/thepracticaldev/DEV-Android/issues/58)
- [https://github.com/thepracticaldev/DEV-Android/pull/60](https://github.com/thepracticaldev/DEV-Android/pull/60)
- #4543

rhymes:
Additional info: Firefox 70 and Chrome 77 on macOS 10.14.6 both work

rclanan:
I just tested this on a desktop:
* OS: Windows 10 Pro, version 1809, 64-bit
* Browsers: 
	- Firefox 69.0.3 - 64-bit
	- Firefox 70.0 - 64-bit

and it appears to be functioning properly.

VarunBarad:
For some reason it started working on Firefox on Windows after a system restart 🤷‍♂

rclanan:
When in doubt, restart! :)

rhymes:
I'm closing this since the OP verified the feature is working again!

VarunBarad:
It works only on desktop, still doesn't work on the mobile browsers as I originally mentioned.

Zhao-Andy:
[...]

Issue #7251, opened 2020-04-13
Copy markdown image button not working when uploading a new photo for a post


simonxabris:
**Describe the bug**

Hi!

I tried to create my first post on DEV and when I tried to upload a picture and copy the generated markdown URL, the text was just simply not copied to my clipboard. I tried it in basically every major browser (Firefox, Safari, Chrome, Edge) and it did not work in any of them. However I tried this out on my phone too (iPhone, safari) and it worked as expected.

I found two issues that I think are related to this.

[This one](https://github.com/thepracticaldev/dev.to/issues/6169) is definitely related, because it's the same UI where I experienced the issue.

and [this one](https://github.com/thepracticaldev/dev.to/issues/7109) might be related since it's about copying as well, but at a different part of the app.

I've opened this new issue, because the first one I referenced makes it seems like it's platform specific (it has area:mobile label on it), while in my experience it is not tied to platform. So this one could be a more generic and possible solve the problem mentioned in the issues I referenced.

**To Reproduce**

- Start writing a new article.
- Upload a picture.
- Click on the copy icon next to the generated markdown url text.
- Check if it actually copied the text.

**Expected behavior**

The expected behavior is to actually copy the markdown text to the user's clipboard when the button is clicked.

**Desktop (please complete the following information):**

- OS: macOS 10.14.6
- As I stated above, I tested this in multiple browsers and it did not work in any of them

**Smartphone (please complete the following information):**

- Device: iphone 8
- OS: ios 13.4.1
- Browser: safari
- Version: 13

**Additional context**

So I went ahead and searched fo the bug in the code and I think I found it. The bug is in this function: https://github.com/thepracticaldev/dev.to/blob/80737b540453afe8775128cb37bd379b7c09c7e8/app/javascript/article-form/elements/imageManagement.jsx#L77

The command to actually copy the text is only called on iOS devices for some reason.

I'd be happy to submit a PR with a fix if this is something you would be interested in!

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

[...]

PR #7382, opened 2020-04-20
Copy to clipboard fixes including support for Android Native Bridge

### LINKED ISSUE #6169: I can't use "copy to clipboard" in add image in new post
Alt Text

Alt Text

My device: Realme C2 - details https://www.gsmarena.com/realme_c2-9670.php
Operating system: Android 9.0 (Pie); ColorOS 6 Lite
My browser: Via Browser

fdoxyz:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR intends to fix the "copy to clipboard" issue for the Android native app when writing a post by using the native bridge interface.

It uses the [Clipboard async API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) for all other cases (not Native Android app), which seems to have better support on most browsers/devices. It also has a `catch` clause that will fallback to the previous implementation when the Clipboard API isn't supported.

This approach relies on [this Android PR](https://github.com/thepracticaldev/DEV-Android/pull/70) which makes the `copyToClipboard` function available in JS.

Originally intended to fix #6169 but might fix other clipboard issues too.

*Sample image of the 'copy to clipboard' area that will be affected*
Screen Shot 2020-04-18 at 21 27 57

## Added tests?

- [ ] yes
- [x] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

clipboard thrown to the floor


fdoxyz:
This PR should fix the problem found in #7251 

Now looking into a way to test the Component (for diff-coverage), might ask for help with this if can't figure it out soon.

fdoxyz:
Glad to know it works on Firefox thanks for the confirmation @rhymes!

I'll definitely try to add the other copy to clipboard soon

fdoxyz:
@rhymes FYI just pushed the changes for your suggestion in this same PR so it should hopefully close #4545 too 🙂

If codeclimate agrees it should be ready, but I may still need to add some coverage to the JSX

PR #7385, opened 2020-04-20
🚀 Feature: Ability to search discoverable channels and send request to join

sarthology:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
1. Mod will be able to make their groups discoverable is the channel setting.
image
2. These groups then can be searched from the search option in the chat channel list.

Screen Recording 2020-04-20 at 05 46 PM

3. One can slick and send the request to join.

Screen Recording 2020-04-20 at 05 47 PM

4. Mod can accept and reject the request.
Screenshot 2020-04-20 at 5 48 24 PM

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


sarthology:
### 🚨 Still polishing it a bit.
Sent PR to take your feedback before finalizing it.

app/services/search/chat_channel_membership.rb
```diff
@@ -7,6 +7,14 @@ class ChatChannelMembership < Base
     DEFAULT_PER_PAGE = 30
 
     class << self
+      def search_documents(params:)
```
mstruve, line 10: This is in the `search/base.rb` class so you shouldn't need to define this method again here.

app/services/search/query_builders/chat_channel_membership.rb
```diff
@@ -33,8 +34,14 @@ def initialize(params:)
 
       def build_queries
         @body[:query] = {}
-        @body[:query][:bool] = { filter: filter_conditions }
-        @body[:query][:bool][:must] = query_conditions if query_keys_present?
+        if !filter_conditions.select { |x| x[:term].key?(:channel_discoverable) }.empty?
```
mstruve, line 37: Rather than doing this complicated logic, here is a solution that I think would work better and requires no additional field or logic change here. 
- Remove channel_discoverable field
- If a channel is discoverable set `viewable_by` to `all`
```ruby
params[:viewable_by] = [<user_id>, "all"]
```
By default include multiple terms will be queried using an `OR`

mstruve:
Also can we please make this a Draft PR until it is ready?

sarthology:
> Also can we please make this a Draft PR until it is ready?

Yes Sure.

app/services/search/query_builders/chat_channel_membership.rb
Parasgr-code: Hi @mstruve , i have optimized the above code to this. Please let me know what you think. 
`[      def build_queries
        @body[:query] = {}
        if @params[:user_id] == "all"
          @body[:query][:bool] = { filter: [{ terms: { status: %w[active joining_request] } }, { term: { channel_discoverable: true } }] }
          @body[:query][:bool][:must] = query_conditions
        else
          @body[:query][:bool] = { filter: filter_conditions }
          @body[:query][:bool][:must] = query_conditions if query_keys_present?
        end
      end](url)`

sarthology:
## Feature addon: Show use if already applied for the group.

Screen Recording 2020-04-25 at 10 16 PM

app/controllers/search_controller.rb
```diff
@@ -63,6 +63,19 @@ def chat_channels
     render json: { result: ccm_docs }
   end
 
+  def chat_channels_discoverable
+    ccm_docs_original = Search::ChatChannelMembership.search_documents(
+      params: chat_channel_params.merge(user_id: current_user.id).to_h,
+    )
+    discoverable_params = chat_channel_params.merge(user_id: "all").to_h
+
+    ccm_docs_discoverable = Search::ChatChannelMembership.search_documents(
```
mstruve, line 72: If you update the docs and search to use the `viewable_by` field as [suggested here](https://github.com/thepracticaldev/dev.to/pull/7385#discussion_r411386390) then you won't need to make 2 requests to Elasticsearch.

Parasgr-code: Ok @mstruve, we're looking more to optimize this piece of code as you've suggested.
Thanks

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -39,6 +39,24 @@ def create
     redirect_to edit_chat_channel_membership_path(membership)
   end
 
+  def join_channel
+    membership_params = params[:chat_channel_membership]
+    chat_channel = ChatChannel.find(membership_params[:chat_channel_id])
+    user = User.find_by(id: membership_params[:user_id])
+    existing_membership = ChatChannelMembership.find_by(user_id: user.id, chat_channel_id: chat_channel.id)
+    if existing_membership.present? && %w[active joining_request].exclude?(existing_membership.status)
+      success = existing_membership.update(status: "joining_request", role: "member") ? true : false
```
mstruve, line 48: `update` will return true or false so there is no need for the extra ternary statement.

Also, if the membership already exists then aren't we done? Why would we mark a current membership as a "joining_request"?

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -39,6 +39,24 @@ def create
     redirect_to edit_chat_channel_membership_path(membership)
   end
 
+  def join_channel
+    membership_params = params[:chat_channel_membership]
+    chat_channel = ChatChannel.find(membership_params[:chat_channel_id])
+    user = User.find_by(id: membership_params[:user_id])
```
mstruve, line 45: Can we use `current_user` here instead to avoid someone being able to send any user_id they want?

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -39,6 +39,24 @@ def create
...
+    existing_membership = ChatChannelMembership.find_by(user_id: user.id, chat_channel_id: chat_channel.id)
+    if existing_membership.present? && %w[active joining_request].exclude?(existing_membership.status)
+      success = existing_membership.update(status: "joining_request", role: "member") ? true : false
+    else
+      membership = ChatChannelMembership.new(user_id: user.id, chat_channel_id: chat_channel.id, role: "member", status: "joining_request")
+      success = membership.save ? true : false
```
mstruve, line 51: Same as above, save will return true or false already.

app/controllers/search_controller.rb
```diff
@@ -63,6 +63,17 @@ def chat_channels
     render json: { result: ccm_docs }
   end
 
+  def chat_channels_discoverable
```
mstruve, line 66: I still don't think we need this extra search method or this extra search. 

We should be able to execute the search for discoverable channels through the regular chat_channel search endpoint using the necessary params to return the discoverable docs as well as the user viewable channels. Then once the documents are sent to the view, the view can figure out which ones to display where if indeed we want them displayed in different sections. However, from the looks of the PR description, these are just displayed inline with the rest of the channels.

app/controllers/chat_channel_memberships_controller.rb
Parasgr-code: There might be few cases like if a person has left the channel or removed from the channel so in that case, if they again want to join the closed group they can send the request.
That's why we're looking for existing memberships and updating its status  as "joining_request"

app/controllers/chat_channel_memberships_controller.rb
Parasgr-code: Yes, we can do that. Modified.
Thanks

app/controllers/chat_channel_memberships_controller.rb
Parasgr-code: Thanks for pointing this out. Modified

app/controllers/search_controller.rb
Parasgr-code: We have now used a single chat_channel search endpoint for fetching discoverable channels. Please review and let us know.
Thanks

app/controllers/search_controller.rb
```diff
@@ -56,11 +56,22 @@ def tags
   end
 
   def chat_channels
-    ccm_docs = Search::ChatChannelMembership.search_documents(
-      params: chat_channel_params.merge(user_id: current_user.id).to_h,
-    )
-
-    render json: { result: ccm_docs }
+    if chat_channel_params["discoverable"].present?
```
mstruve, line 59: All the chat channel docs should be able to be returned together using a single search without the need for an `if/else` statement. This endpoint should be able to remain exactly how it is with us just tweaking the params that we are sending to it. The params we send should include `user_id: [<the_user_id>, "all"]` then we run our search and get back all relevant docs. 

We don't want to make multiple calls to the server or Elasticsearch when we could do it with a single call and then let the frontend sort out the documents. That will greatly decrease the number of searches we have to run and will help keep our application running fast.

sarthology: @mstruve Just optimized the code further. 

The complexity of repetitive channels will be too much to handle in frontend. Also if in future let say there are 100+ user in a channel then it will be a bigger mess. But I agree that multiple calls to Elasticsearch also should be avoided. So the new code addressed both the things. Please have a look and if further optimization is required we are up for it. 🙂

benhalpern:
Hmmm I can't seem to get it to work for me locally. I'll try again tomorrow.

sarthology:
> Hmmm I can't seem to get it to work for me locally. I'll try again tomorrow.

😱 Can you tell me what's exactly not working?

app/controllers/search_controller.rb
```diff
@@ -57,10 +57,16 @@ def tags
 
   def chat_channels
     ccm_docs = Search::ChatChannelMembership.search_documents(
-      params: chat_channel_params.merge(user_id: current_user.id).to_h,
+      params: chat_channel_params.merge(user_id: chat_channel_params[:user_id] || current_user.id).to_h,
```
mstruve, line 60: If we are going to allow this as a param we should validate that it is either `all` or the current_user ID otherwise someone could theoretically fetch another user's chat channels

app/services/search/query_builders/chat_channel_membership.rb
```diff
@@ -6,6 +6,7 @@ class ChatChannelMembership < QueryBase
         channel_type
         status
         viewable_by
+        channel_discoverable
```
mstruve, line 9: Since we are not going to be using this field in Elasticsearch we should remove all the Elasticsearch code that was put in place for it.

app/services/search/query_builders/chat_channel_membership.rb
```diff
@@ -33,8 +34,14 @@ def initialize(params:)
 
       def build_queries
         @body[:query] = {}
-        @body[:query][:bool] = { filter: filter_conditions }
-        @body[:query][:bool][:must] = query_conditions if query_keys_present?
+        if @params[:user_id] == "all"
+          @body[:query][:bool] = { filter: [{ terms: { status: %w[active joining_request] } }] }
```
mstruve, line 38: Rather than having this if statement here I think it would be better to set these params before we send the request to the query builder or [up in the line above that already has a comment](https://github.com/thepracticaldev/dev.to/pull/7385/files#diff-c8d5692235cc1a81cb7ed64d40688e2cR28).

The query builder's sole job is to take a set of params/filters and make an ES query out of it. I want to try as much as possible to not have it being what sets the params/filters which is why I commented on that line above where I set it myself.

app/services/search/query_builders/chat_channel_membership.rb
```diff
@@ -33,8 +34,14 @@ def initialize(params:)
...
+          @body[:query][:bool][:must] = query_conditions
+          # TODO: Optimize it more to fetch uniq results
```
mstruve, line 40: What problems are you having where the results are not unique?

sarthology:
@mstruve Implemented what we have discussed.

thomasbnt:
Youhouu :tada:

PR #7389, opened 2020-04-20
Filter Elasticsearch by Organization ID Properly

mstruve:
- [x] Bug Fix

## Description
Organizations in Elasticsearch need to be searched using their ID not by tags

## Added tests?
- [x] no, I will add them in a followup PR

alt_text


fdoxyz:
Locally I'm getting an infinite scroll that includes other posts from outside the Org. Any chance I'm missing a manual update to ES indexes/tags at all? 

I did run `bin/setup` so at least `bin/rails search:setup` was executed

PR #7390, opened 2020-04-20
Removal of morrrre hardcoded dev.to links

Ridhwana:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Removing even MORRRREE hardcoded links. We now have 90% of the hardcoded links removed. 
Just a couple of more left in some controllers that need some more testing. Will leave those for subsequent PR's 😓, but we getting there 😁.

https://github.com/thepracticaldev/dev.to/projects/15#card-35230751

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

alt_text

### DIFF
--- a/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
+++ b/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
@@ -158,7 +158,7 @@ function toggleCodeOfConduct() {
   var checkboxWrapper = document.getElementById('toggle-code-of-conduct-checkbox');
   if (checkboxWrapper && !codeOfConduct && commentCount < 1) {
     checkboxWrapper.innerHTML = '<input type="checkbox" name="checked_code_of_conduct" class="checkbox" required/>\
-                                  <label for="checked_code_of_conduct"/>I\'ve read the <a href="<%= ApplicationConfig["APP_PROTOCOL"] %><%= ApplicationConfig["APP_DOMAIN"] %>/code-of-conduct">code of conduct</a></label>'
+                                  <label for="checked_code_of_conduct">I\'ve read the <a href="<%= URL.url('code-of-conduct') %>">code of conduct</a></label>'
   }
 }
 
--- a/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
+++ b/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
@@ -155,10 +155,10 @@ function toggleCodeOfConduct() {
   }
   var codeOfConduct = user.checked_code_of_conduct
   var commentCount = user.number_of_comments
-  var checkboxWrapper = document.getElementById('toggle-code-of-conduct-checkbox')
+  var checkboxWrapper = document.getElementById('toggle-code-of-conduct-checkbox');
   if (checkboxWrapper && !codeOfConduct && commentCount < 1) {
     checkboxWrapper.innerHTML = '<input type="checkbox" name="checked_code_of_conduct" class="checkbox" required/>\
-                                  <label for="checked_code_of_conduct"/>I\'ve read the <a href="https://dev.to/code-of-conduct">code of conduct</a></label>'
+                                  <label for="checked_code_of_conduct"/>I\'ve read the <a href="<%= ApplicationConfig["APP_PROTOCOL"] %><%= ApplicationConfig["APP_DOMAIN"] %>/code-of-conduct">code of conduct</a></label>'
   }
 }
 
--- a/app/controllers/internal/events_controller.rb
+++ b/app/controllers/internal/events_controller.rb
@@ -1,11 +1,12 @@
 module Internal
   class EventsController < ApplicationController
     layout "internal"
+    include ApplicationHelper
 
     def index
       @event = Event.new(
-        location_name: "dev.to/live",
-        location_url: "https://dev.to",
+        location_name: "#{ApplicationConfig['APP_DOMAIN']}/live",
+        location_url: app_url,
         description_markdown: "*Description* *Pre-requisites:* *Bio*",
       )
       @events = Event.order("starts_at DESC")
--- a/app/controllers/internal/events_controller.rb
+++ b/app/controllers/internal/events_controller.rb
@@ -5,7 +5,7 @@ module Internal
 
     def index
       @event = Event.new(
-        location_name: "#{ApplicationConfig['APP_DOMAIN']}/live",
+        location_name: "#{URL.domain}/live",
         location_url: app_url,
         description_markdown: "*Description* *Pre-requisites:* *Bio*",
       )
--- a/app/views/internal/tools/index.html.erb
+++ b/app/views/internal/tools/index.html.erb
@@ -1,12 +1,12 @@
 <h2>General Purpose URL Purge</h2>
 <p>For a page that should 404. This will purge the page via Fastly.</p>
-<p>Use the full path (https://dev.to/xyz) or the relative path: (/xyz)</p>
+<p>Use the full path (<%= app_url('xyz') %>) or the relative path: (/xyz)</p>
 <div class="row m-3">
   <%= form_tag bust_cache_internal_tools_path do %>
     <div class="form-row">
       <div class="form-group">
         <%= label_tag("Link:") %>
-        <%= text_field_tag :dead_link, nil, placeholder: "/devteam/page" %>
+        <%= text_field_tag :dead_link, nil, placeholder: "/xyz" %>
         <%= submit_tag("Clear Cache") %>
       </div>
     </div>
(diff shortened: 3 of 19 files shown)


app/assets/javascripts/initializers/initializeCommentsPage.js.erb
```diff
@@ -155,10 +155,10 @@ function toggleCodeOfConduct() {
   }
   var codeOfConduct = user.checked_code_of_conduct
   var commentCount = user.number_of_comments
-  var checkboxWrapper = document.getElementById('toggle-code-of-conduct-checkbox')
+  var checkboxWrapper = document.getElementById('toggle-code-of-conduct-checkbox');
   if (checkboxWrapper && !codeOfConduct && commentCount < 1) {
     checkboxWrapper.innerHTML = '<input type="checkbox" name="checked_code_of_conduct" class="checkbox" required/>\
-                                  <label for="checked_code_of_conduct"/>I\'ve read the <a href="https://dev.to/code-of-conduct">code of conduct</a></label>'
+                                  <label for="checked_code_of_conduct"/>I\'ve read the <a href="<%= ApplicationConfig["APP_PROTOCOL"] %><%= ApplicationConfig["APP_DOMAIN"] %>/code-of-conduct">code of conduct</a></label>'
```
Ridhwana, line 161: I'm not sure if maybe theres a way that I'm not aware of to pull in the app helper in this file, instead of using <%= ApplicationConfig["APP_PROTOCOL"] %><%= ApplicationConfig["APP_DOMAIN"] %>. This was the most effective way.

app/views/internal/tools/index.html.erb
```diff
@@ -1,12 +1,12 @@
...
-<p>Use the full path (https://dev.to/xyz) or the relative path: (/xyz)</p>
+<p>Use the full path (<%= app_url('xyz') %>) or the relative path: (/xyz)</p>
 <div class="row m-3">
   <%= form_tag bust_cache_internal_tools_path do %>
     <div class="form-row">
       <div class="form-group">
         <%= label_tag("Link:") %>
-        <%= text_field_tag :dead_link, nil, placeholder: "/devteam/page" %>
+        <%= text_field_tag :dead_link, nil, placeholder: "/xyz" %>
         <%= submit_tag("Clear Cache") %>
```
Ridhwana, line 10: changed this from `/devteam/page` as there was no way to neat way to do this, and xyz seemes inline with the description

app/views/users/show.html.erb
```diff
@@ -30,7 +30,6 @@
     <% end %>
   </style>
   <div class="user-profile-header" style="<%= user_colors_style(@user) %>">
-    <a href="https://dev.to/<%= user_url(@user) %>"></a>
```
Ridhwana, line 33: This href was not being shown on the page. Hence I removed it. Hopefully it wasn't there for some other weird reason that I'm unaware of?

spec/support/fixtures/approvals/email_subscriptionsunsubscribe.approved.html
```diff
@@ -7,11 +7,11 @@
   <body>
     <div class="unsubscribed">
   <div class="unsubscribed__title">
-    You have been unsubscribed from DEV digest emails. 😔
+    You have been unsubscribed from DEV(local) digest emails. 😔
   </div>
   <br /><div class="unsubscribed__message">
     You will no longer receive emails like this. If you'd like to re-subscribe or manage your full email preferences, please
-    <a href="https://dev.to/settings/notifications">click here</a>.
+    <a href="http://localhost:3000/settings/notifications">click here</a>.
```
Ridhwana, line 14: I'm not 100% sure that this is going to work when we build travis from github, and I'm not sure this is the best way to do it. May need to find another solution

Ridhwana, line 14: This is pretty dependent on the EnvFile values, this the default values so this test will pass unless you have overridden the values. Not sure how to cater for this otherwise, as I cant reference the ApplicationConfig in this file..

fdoxyz: Seems like `URL.url('/settings/notifications')` may be the way to go here to get this one to work with people's defaults? Although when testing not sure why you'd override these

app/assets/javascripts/initializers/initializeCommentsPage.js.erb
juliannatetreault, line 161: I don't believe that the `/` is needed for this opening `<label>` tag since you close the tag at the end of this line of code. 😄

app/views/pages/badges.html.erb
```diff
@@ -32,7 +32,7 @@
     </p>
     <% if user_signed_in? %>
       <p>
-        <a href="https://dev.to/<%= current_user.username %>">
+        <a href="<%= user_url(current_user) %>">
```
juliannatetreault, line 35: There are two code embeds on lines 48 and 56 in this file that have the hardcoded `https://dev.to/` URL in them--is it necessary to change those URLs to `app_url` as well?

spec/support/fixtures/approvals/email_subscriptionsunsubscribe.approved.html
juliannatetreault: I agree with @fdoxyz that `URL.url(/settings/notifications)` seems like the best approach here!

app/assets/javascripts/initializers/initializeCommentsPage.js.erb
Ridhwana: Nice one thanks @fdoxyz !

app/views/pages/badges.html.erb
Ridhwana: Yes definitely, thank you!

spec/support/fixtures/approvals/email_subscriptionsunsubscribe.approved.html
Ridhwana: Hmm I don't think this is possible because it's a plain html file.

Ridhwana:
@juliannatetreault and @fdoxyz thanks for the review, I made the suggested changes :)

app/controllers/internal/events_controller.rb
```diff
@@ -1,11 +1,12 @@
 module Internal
   class EventsController < ApplicationController
     layout "internal"
+    include ApplicationHelper
 
     def index
       @event = Event.new(
-        location_name: "dev.to/live",
-        location_url: "https://dev.to",
+        location_name: "#{ApplicationConfig['APP_DOMAIN']}/live",
```
lightalloy, line 8: Could be `URL.domain` ?

app/views/mailers/notify_mailer/tag_moderator_confirmation_email.text.erb
```diff
@@ -1,14 +1,14 @@
 Hey <%= @user.name %>!
 
-Thanks for volunteering to be a <%= community_name %> tag moderator.  We’ve added you as a moderator for https://dev.to/t/<%= @tag %>. You’re currently the only non-core <%= community_name %> member that’s moderating this tag. Over time, we’ll add additional mods to join you because we view moderation as a collaborative endeavor that benefits from the support of multiple people.
+Thanks for volunteering to be a <%= community_name %> tag moderator.  We’ve added you as a moderator for <%= tag_url(@tag) %>. You’re currently the only non-core <%= community_name %> member that’s moderating this tag. Over time, we’ll add additional mods to join you because we view moderation as a collaborative endeavor that benefits from the support of multiple people.
```
lightalloy, line 3: Here `tag_url` is used, which will use domain from action mailer `smtp_settings` (`ENV["APP_DOMAIN"]`) which could be fine, but in other places `app_url` is used. I suppose we should be consistent with the urls in email templates.

Ridhwana: @lightalloy I'm not quite sure what you mean, but I'll take a stab at a response. When an application helper like tag_url, comment_url, article_url, or user_url is more apt I use that to substitute the hardcoded url. But if we cannot use one of the more specific helpers, I then use app_url. This is true for the most part but I'm sure that there is one or two places where I need to clean up.

app/controllers/internal/events_controller.rb
Ridhwana: Thanks Anna, going to update this now. In retrospect, theres a few other places in the codebase I'm still using `ApplicationConfig['APP_DOMAIN']`, I'll update those in subsequent PR's as I go along.

PR #7391, opened 2020-04-20
Allow Filtering by Approved Attribute for Articles

mstruve:
- [x] Bug Fix

## Description
Make sure we can filter by the article `approved` attribute

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,7 +1,7 @@
 class SearchController < ApplicationController
   before_action :authenticate_user!, only: %i[tags chat_channels reactions]
   before_action :format_integer_params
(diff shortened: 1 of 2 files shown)


app/controllers/search_controller.rb
```diff
@@ -1,7 +1,7 @@
 class SearchController < ApplicationController
   before_action :authenticate_user!, only: %i[tags chat_channels reactions]
   before_action :format_integer_params
-  before_action :sanitize_params, only: %i[classified_listings reactions]
+  before_action :sanitize_params, only: %i[classified_listings reactions feed_content]
```
mstruve, line 4: We need this bc approved is passed as a blank string for all tags that don't require it so we want to make sure it is removed from the params going to Elasticsearch.

PR #7393, opened 2020-04-20
Settings style updates: Crayons!

ludwiczakpawel:

- [x] Refactor

## Description

This PR brings new styling and a bit of frontend cleanup to Settings section. It adds lots of Crayons classes etc.
This PR is **not** about rethinking information architecture or making general improvements. Not everything will look ideal but it's a step forward. More improvements in further PRs.

[Little demo hero](https://d.pr/v/PTH1hz).

image

## Added tests?

- [x] yes

## Added to documentation?

- [x] no documentation needed


benhalpern:
Very excited for this!

app/views/users/_profile.html.erb
```diff
@@ -1,144 +1,187 @@
...
+    </div>
+
+    <div class="crayons-field">
+      <%= f.label :summary, class: "crayons-field__label" %>
+      <%= f.text_area :summary, maxlength: 200, class: "crayons-textfield", placeholder: "A short bio...", rows: 3 %>
+    </div>
+
+    <div class="crayons-field">
+      <%= f.label :location, class: "crayons-field__label" %>
+      <%= f.text_field :location, maxlength: 100, class: "crayons-textfield", placeholder: "San Francisco, CA" %>
```
benhalpern, line 59: Petition for the placeholder text to be "Halifax, Nova Scotia"

thomasbnt:
woaah very clean :o

app/views/users/_profile.html.erb
ludwiczakpawel: Petition accepted!

app/views/users/_additional_authentication.html.erb
```diff
@@ -1,17 +1,16 @@
-<% if SiteConfig.auth_allowed?("github") && !@user.identities.exists?(provider: "github") %>
-  <div class="field">
-    <a href="/users/auth/github" class="big-button cta" data-no-instant>
-      <img src="<%= asset_path("github-logo.svg") %>" alt="github logo"> CONNECT GITHUB ACCOUNT
-    </a>
+<% if SiteConfig.auth_allowed?("github") && SiteConfig.auth_allowed?("twitter") && (!@user.identities.exists?(provider: "github") || !@user.identities.exists?(provider: "twitter")) %>
```
rhymes, line 1: I think you changed the funtionality here, as the boolean conditions of before are different from the ones there are now. 

Before it said: "if github is enabled and the user hasn't already connected with github, show this section" (and separately from twitter, as they can be enabled separately)

Now it checks if both github and twitter are enabled, but either could be disabled

db/schema.rb
```diff
@@ -308,7 +308,6 @@
     t.string "slug"
     t.string "status", default: "active"
     t.datetime "updated_at", null: false
-    t.index ["slug"], name: "index_chat_channels_on_slug", unique: true
```
rhymes, line 311: this should stay in

ludwiczakpawel: uhh. sorry. my local env is messing with me.. :/

app/views/users/_additional_authentication.html.erb
ludwiczakpawel: hmm.. i'll double check. I think this is what I wanted:

```
<% if (SiteConfig.auth_allowed?("github") || SiteConfig.auth_allowed?("twitter")) && (!@user.identities.exists?(provider: "github") || !@user.identities.exists?(provider: "twitter")) %>
```

but i'll double check that.

ludwiczakpawel:
@rhymes as for selects - yea we don't really have styling for selects in Crayons just yet. I'll pop it in in a separate PR some day next week probably.

> The color text field in the organizations also seem to be missing a left padding

Actually I'd love your help with that! :) 

When you create an org you have these two color fields that look off... But when you edit an org, this is way different, it uses `_logo_design` partial. The same partial is used for editing user's profile. I'm not really sure how to use this partial instead of the two fields you showed on screenshot. Would you find couple mins to help?

rhymes:
> @rhymes as for selects - yea we don't really have styling for selects in Crayons just yet. I'll pop it in in a separate PR some day next week probably.

cool, thanks!

> When you create an org you have these two color fields that look off... But when you edit an org, this is way different, it uses `_logo_design` partial. The same partial is used for editing user's profile. I'm not really sure how to use this partial instead of the two fields you showed on screenshot. Would you find couple mins to help?

sure! let me look into this

lisasy:
@ludwiczakpawel 

Nice work! I'm really excited about this.

On mobile, can you add a margin around the title "Settings for ..."?

In our backlog somewhere, we can make sure we track mobile navigation for the top section to make sure it's not so overwhelming 🙃

Screenshot 2020-04-21 07 14 18

rhymes:
@ludwiczakpawel updated the PR to use the same component as the other pages. Please test it as well :)

Screenshot_2020-04-21 Settings - DEV(local) Community 👩‍💻👨‍💻(2)

benhalpern:
Biggest issues I'm seeing are the experience level buttons....

which could definitely be crayonsified I think. It could also be a different kind of UI if we wanted to do something delightful here (cc @lisasy)

Screen Shot 2020-04-21 at 2 40 15 PM

The other thing I'm seeing as weird is that the "selected" state has the background in the menu but the "hover" state does not, so if you select something you don't see the BG change until you move away...



Unless this is intended behavior should be a quick adjustment in the ordering of stuff or whatever.

ludwiczakpawel:
@benhalpern im aware of the experience selector looking off. I have ideas but it will be a new component which I dont want to introduce here. defo a separate PR.

re: hover state - its already there waiting for a moment when we update body background color. that’s the only reason it not visible yet. i think that’s fine for temporary state but I’m pretty sure we will get there soon.

PR #7394, opened 2020-04-20
Update sidebar filter menu on Notifications

lisasy:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
- Update sidebar filter menu on Notifications
- Use `crayons-card` component to build out filter menu on desktop

#### Before
Screenshot 2020-04-20 09 13 19

#### Proposed
2020-04-21_Sidebar-Filter

#### Before
Screenshot 2020-04-20 09 13 30

#### Proposed
Screenshot 2020-04-20 09 13 39

#### Interaction demo
2020-04-17_Notifications-Sidebar-Mobile

## Added tests?

- [ ] yes
- [x] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed

alt_text


app/assets/images/arrow-down-fill.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M12 16L6 10H18L12 16Z" fill="#08090A"/>
```
ludwiczakpawel, line 2: We can drop the `fill` attribute - it's being overwritten anyway by `crayons-icon` class.

app/assets/images/checkmark.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M9.99999 15.172L19.192 5.979L20.607 7.393L9.99999 18L3.63599 11.636L5.04999 10.222L9.99999 15.172Z" fill="#08090A"/>
```
ludwiczakpawel, line 2: We can drop the `fill` attribute - it's being overwritten anyway by `crayons-icon` class.

app/assets/stylesheets/components/navigation.scss
```diff
@@ -112,4 +112,26 @@
       }
     }
   }
+
+  &--item {
+    padding: $su-3;
+    text-align: center;
+    width: 100%;
+
+    &.selected {
```
ludwiczakpawel, line 121: In most cases, wherever we have this "selected/current" state we use "current" keyword. 
Also, since we are all-in with BEM naming convention, it should probably be `.crayons-tabs--item--selected`.

app/assets/stylesheets/notifications.scss
```diff
@@ -1,9 +1,12 @@
 @import 'variables';
 @import 'mixins';
 @import 'functions';
+@import 'config/import';
 
 .notifications-index {
   .home {
+    overflow: initial;
```
ludwiczakpawel, line 8: Could you help me understand what does it do?

app/assets/stylesheets/notifications.scss
```diff
@@ -428,3 +431,134 @@
...
+      content: "";
+      position: fixed;
+      left: 0;
+      right: 0;
+      top: 0;
+      right: 0;
+      height: 100%;
+      width: 100%;
+      background: rgba(0,0,0,0.5);
+      z-index: 499;
```
ludwiczakpawel, line 478: `499`? :D

app/views/notifications/index.html.erb
```diff
@@ -19,55 +19,185 @@
...
       <%= render "notifications_list", params: params %>
+
     </div>
 
     <%# new or less active users that don't have a page worth of notifications
         won't be shown the "load more" button %>
     <% if @notifications.size >= @initial_page_size %>
       <div class="load-more-wrapper">
-        <button id="load-more-button" type="button">
+        <button id="load-more-button" type="button" class="crayons-btn--secondary">
```
ludwiczakpawel, line 200: I think you need two classes here: `crayons-btn crayons-btn--secondary` (BEM naming).

app/assets/stylesheets/components/navigation.scss
```diff
@@ -112,4 +112,26 @@
...
+      border-bottom: 4px solid var(--accent-brand);
+
+      .crayons-tabs--label {
```
ludwiczakpawel, line 124: `__` instead of `--` since [it's a child element, not a modifier](http://getbem.com/naming/).

app/assets/stylesheets/components/navigation.scss
```diff
@@ -112,4 +112,26 @@
...
+        color: var(--card-color);
```
ludwiczakpawel, line 125: I'd use one of the `--link-*` variables instead of `--card-color` since card is, in most cases, specifically made for `.crayons-card`.

app/assets/javascripts/initializers/initNotifications.js
```diff
@@ -127,6 +128,27 @@ function listenForNotificationsBellClick() {
   }, 180);
 }
 
+function initFilter() {
```
vaidehijoshi, line 131: We don't have a good habit of adding documentation around new JS code, but I would definitely encourage adding docs anytime we add new code :)

I believe we're using [JSDoc](https://jsdoc.app/), and there's [a good example of how to use it](https://github.com/thepracticaldev/dev.to/blob/d461d835405e023630bf694037fccef7efaea8df/app/javascript/src/utils/debounceAction.js#L3-L16) in some of our preexisting JS functions!

app/assets/stylesheets/notifications.scss
vaidehijoshi: this feels like a magic number

app/assets/stylesheets/notifications.scss
```diff
@@ -428,3 +431,134 @@
...
+
+  &__menu {
+    display: flex;
+    background: var(--base-inverted);
+    flex-direction: column;
+    position: absolute;
+    width: 100%;
+    left: 0;
+    right: 0;
+    z-index: 500;
```
vaidehijoshi, line 498: another magic number, i suspect

app/views/notifications/index.html.erb
```diff
@@ -19,55 +19,185 @@
 
 <% if user_signed_in? %>
   <div class="home" id="notifications-container">
+
```
vaidehijoshi, line 22: ✂️ empty line

vaidehijoshi:
This might be out of scope for this PR...but I wonder if we should render a "you have no notifications!" type of message when there's nothing to display. Right now it feels like the page is just hanging and waiting to render something/potentially broken 🤔 

Screen Shot 2020-04-20 at 3 06 20 PM

app/assets/stylesheets/components/navigation.scss
lisasy: Noted! This is helpful because a lot of the existing markup/class names use "selected" to indicate status. I'll update this to `current`.

app/assets/stylesheets/notifications.scss
lisasy: I'm not sure why, but without this line, the `.home` container was hiding a portion of the menu that comes up. I noticed that in another file, `.home` has `overflow: hidden` as the default.

app/assets/stylesheets/notifications.scss
lisasy: 😂 The menu uses 500, so I chose a lower number. Do we have best practices around this?

app/views/notifications/index.html.erb
lisasy: OoOooOoo okay!

app/assets/stylesheets/components/navigation.scss
lisasy: You caught me not knowing BEM and guessing 😂

benhalpern:
> This might be out of scope for this PR...but I wonder if we should render a "you have no notifications!" type of message when there's nothing to display. Right now it feels like the page is just hanging and waiting to render something/potentially broken 🤔

Yeah, I was going to say "this situation doesn't really happen because we send onboarding notifications", but then I remembered that we periodically remove older notifications from the system, so if you have no activity for a while you might come back to empty... Either way yeah probably worth dealing with, not investing too much time in.

benhalpern:
I love the mobile nav!

thomasbnt:
Me too, the navbar to change notifications profiles is very clean :+1:

app/assets/stylesheets/components/cards.scss
```diff
@@ -20,6 +20,18 @@ $card-divider: 1px solid var(--body-bg);
     align-items: center;
   }
 
+  &__section-header {
```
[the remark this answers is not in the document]
lisasy: I like this.

app/assets/javascripts/initializers/initNotifications.js
lisasy: Gotcha! I'm copying the other ways we're using JS as a filter — the code I borrowed is from the "chronofilter" on the home feed where we let users filter their feed based on Feed / Week / Month / Year/ Infinity.

app/assets/stylesheets/components/navigation.scss
lisasy: Resolved in latest commits.

app/views/notifications/index.html.erb
lisasy: Resolved in latest commits.

app/assets/images/arrow-down-fill.svg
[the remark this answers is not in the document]
lisasy: Resolved in latest commits.

app/assets/stylesheets/components/navigation.scss
lisasy: Resolved in latest commits.

rhymes:
This is really cool!

app/assets/stylesheets/components/navigation.scss
```diff
@@ -140,4 +140,25 @@
...
+
+  &--item {
+    padding: $su-3;
+    text-align: center;
+    width: 100%;
+
+    &--current {
+      border-bottom: 4px solid var(--accent-brand);
+
+      .crayons-tabs__label {
```
ludwiczakpawel, line 152: Any reason why this element exist instead of applying styles within to the parent element (`*--current`)? There's an extra span for that in the markup as well so we could probably drop it too.

app/assets/stylesheets/notifications.scss
```diff
@@ -428,3 +431,134 @@
...
+  display: none;
+
+  &.showing {
+    display: flex;
+  }
+
+  background:  var(--base-inverted);
+}
+
+.crayons-dropdown {
```
ludwiczakpawel, line 459: `crayons-dropdown` is already existing component in Crayons (https://dev.to/page/crayons - cmd+f for "dropdown" since I can't link it :D). It's the one we use for profile dropdown in header for example (and few other places too).

But in the markup you don't actually use the `crayons-dropdown` itself - instead it's `crayons-dropdown__*` without this parent element.

Also, since dropdowns already exits in Crayons, any thoughts about reusing them instead of introducing new one? I'm not talking code-wise but also design-wise.

app/assets/stylesheets/notifications.scss
ludwiczakpawel: No, I don't think we do... I was just curious :)

app/assets/stylesheets/notifications.scss
```diff
@@ -428,3 +431,134 @@
...
+    width: 100%;
+    left: 0;
+    right: 0;
+    z-index: 500;
+  }
+
+  &__select {
+    appearance: none;
+    -webkit-appearance: none;
+    -moz-appearance: none;
```
ludwiczakpawel, line 504: No need for vendor prefixes, we add them automatically on the fly.

app/views/notifications/index.html.erb
```diff
@@ -20,54 +20,171 @@
...
+            <% end %>
+            <fieldset class="crayons-fieldset">
+              <a class="query-filter-button <%= "selected" if params[:filter].blank? %>" href="/notifications">
+                <div class="crayons-field crayons-field--radio">
+                  <input type="radio" name="all" id="r2" class="crayons-radio" <%= "checked" if params[:filter].blank? %>>
+                  <label for="all" class="crayons-field__label">
+                    All
+                  </label>
+                </div>
+              </a>
```
ludwiczakpawel, line 44: Having form element with clickable label inside `<a>` is probably a pretty hacky thing to do... What if instead of having a form here AND actual radio buttons we would go with just links and checkmarks on the right like you have here:

image

1. It would make much more sense
2. It would be consistent with actual mobile usage

app/views/notifications/index.html.erb
```diff
@@ -20,54 +20,171 @@
 <% if user_signed_in? %>
   <div class="home" id="notifications-container">
     <div class="side-bar">
-      <div class="widget fixed-widget">
-        <header>
-          <h4>notifications</h4>
+      <div class="crayons-card fixed-widget">
```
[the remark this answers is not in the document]
lisasy: Noted. It was already there, but I can remove it.

app/views/notifications/index.html.erb
```diff
@@ -20,54 +20,171 @@
...
-        <div class="widget-body">
-          <a class="query-filter-button <%= "selected" if params[:filter].blank? %>" href="/notifications">ALL</a>
-          <a class="query-filter-button <%= "selected" if params[:filter].to_s.casecmp("comments").zero? && params[:org_id].blank? %>" href="/notifications/comments">COMMENTS</a>
-          <a class="query-filter-button <%= "selected" if params[:filter].to_s.casecmp("posts").zero? && params[:org_id].blank? %>" href="/notifications/posts">POSTS</a>
+        <div class="crayons-card__body widget-body">
+          <section class="crayons-card__inner-section">
+            <% if @organizations.present? %>
+              <header class="crayons-card__subheader">
+                <h4 class="crayons-card__subheader__title">
+                  <span>@<%= @user.username %></span>
```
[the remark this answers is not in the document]
lisasy: I used it because there is a `@` that is in front `<%= @user.username %>`. What do you think?

app/assets/stylesheets/components/forms.scss
```diff
@@ -181,3 +181,12 @@
...
   }
 }
+
+///////////////////////////////////////////////////
+
+// Fieldset
+.crayons-fieldset {
+  margin: 0;
+  padding: 0;
+  border: none;
```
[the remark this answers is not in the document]
lisasy: Good call!

app/assets/stylesheets/notifications.scss
lisasy: I think this is something we can iterate on a little bit. I think that the existing `crayons-dropdown` menu which uses the box styling works really great on desktop, but as we scale down to mobile, I want to design out a version that felt mobile-first. I think this is something that would take more consideration if we're thinking about it for our crayons design system, so what I can do in this file is rename it so that it's specific to Notifications' filtering menu, and revisit this at a later time when we've worked out those details together.

app/views/notifications/index.html.erb
lisasy: Interesting! I thought to use radio options here (even though the existing markup relies on `a` to do the filtering) because from a UX perspective, the radio buttons indicate that these are selectable, mutually exclusive elements.

But you have a good point about this being not semantic. The way I'm using radio options are a bit "artificial", and it's really the `a` that is performing the action. I'm posting this image of the side-by-side of what is the two options we could have. I'll go ahead and remove crayons form elements here so that it's more semantic, and so that it's more specific to notifications filtering menu.

Screenshot 2020-04-21 14 24 41

lisasy:
@ludwiczakpawel Thanks for all of your help! TLDR I'm going to rename some of the CSS class names so that it's more specific to this notifications-filter-menu use case, and not reference/create too many additional crayons components. I think that there are further considerations and explorations we have before that can happen.

lisasy:
Hello! Here are some of the things I did in my latest updates:

### Responsiveness between 768px and 950px viewport width
Apr-21-2020 21-37-12

In the latest updates, I resolved breakpoint conflict by reverting to what it was previously.

### Updating markup + visual UI of sidebar filter
2020-04-21_Sidebar-Filter
Here, I removed form elements like `radio` and instead styled these as links.

app/assets/stylesheets/notifications.scss
ludwiczakpawel: Sounds good and I definitely agree about renaming this one and coming up with better mobile version for actual `crayons-dropdown`!

app/views/notifications/index.html.erb
ludwiczakpawel: 👍👍👍

Something to also think about re future iterations of this nav is Reading List where filtering, I believe, does **not** have mutually exclusive options. In fact you can select multiple option. For that reason I think your new design with checkmark would work better since it makes sense as replacement for radio as well as checkboxes (in terms of behaviour concept). 👍

app/assets/stylesheets/notifications.scss
nickytonline: I've removed the `overflow: hidden` in my PR for the comments subscription because I had the same issue. (PR is still in Draft mode though). Unless there is some use case that we need to hide the overflow, I think it's totally fine to remove `overflow: hidden` from _top.html.erb. Maybe @benhalpern had a reason for this at some point. I see it was added in https://github.com/thepracticaldev/dev.to/commit/630c1b7425ed38b187f15e21415708145472fd46

app/assets/stylesheets/notifications.scss
nickytonline: We should probably audit the project, maybe during cooldown to see if in fact all these `z-index`es are required and if so, replace them with more sensible ones using the utility classes, e.g. `z-10`.

app/views/notifications/index.html.erb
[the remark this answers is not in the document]
nickytonline: The `@` at the front shouldn't matter as it will render in HTML like this `@someUser`.

PR #7395, opened 2020-04-20
Add header to actions panel

Zhao-Andy:
- [x] Feature

## Description
Add the actions panel "Moderate Post" header.

Screen Shot 2020-04-20 at 1 18 30 PM

### DIFF
--- a/app/assets/stylesheets/moderators.scss
+++ b/app/assets/stylesheets/moderators.scss
@@ -155,7 +155,39 @@
 }
 
 .mod-container {
-  padding: 30px 0px;
+  padding: 16px 0px;
+
+  header.top-header {
+    padding: 0px 16px;
+    display: flex;
+    flex-direction: column;
+    text-align: left;
+    h1 {
+      font-size: 18px;
+      font-weight: bold;
+    }
+    h2 {
+      font-size: 16px;
+      font-weight: normal;
+      color: var(--base-70);
+    }
+    button.circle {
+      position: absolute;
+      width: 40px;
+      height: 40px;
+      border: none;
+      border-radius: 40px;
+      right: 16px;
+      top: 16px;
+      background: var(--base-10);
+      img.chevron-right {
+        position: absolute;
+        top: 20%;
+        right: 33%;
+        left: 21%;
+      }
+    }
+  }
   h2 {
     a {
       color: var(--body-color);
--- a/app/assets/stylesheets/moderators.scss
+++ b/app/assets/stylesheets/moderators.scss
@@ -180,7 +180,7 @@
       right: 16px;
       top: 16px;
       background: var(--base-10);
-      img.chevron-right {
+      .chevron-right {
         position: absolute;
         top: 20%;
         right: 33%;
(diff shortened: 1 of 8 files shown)


Zhao-Andy:
@msarit will probably conflict with your current work 😅

app/assets/stylesheets/moderators.scss
```diff
@@ -155,7 +155,39 @@
 }
 
 .mod-container {
-  padding: 30px 0px;
+  padding: 16px 0px;
```
lisasy, line 158: We can use some `crayons` variables here.

As long as we call `@import 'config/import';` at the top of the file, we can say this:

`padding: $su-4 0;`

In the `config/_import.scss` file, you'll see documentation for our spacing-units (that's what `su` represents).

app/assets/stylesheets/moderators.scss
```diff
@@ -155,7 +155,39 @@
...
+
+  header.top-header {
+    padding: 0px 16px;
+    display: flex;
+    flex-direction: column;
+    text-align: left;
+    h1 {
+      font-size: 18px;
```
lisasy, line 166: We also have font-size variables with `crayons`. Instead of `18px`, we can say: `$fs-l`

app/assets/stylesheets/moderators.scss
```diff
@@ -155,7 +155,39 @@
...
+      font-weight: bold;
```
lisasy, line 167: Instead of `font-weight: bold;`, we can say: `font-weight: $fw-bold;`

app/assets/stylesheets/moderators.scss
```diff
@@ -155,7 +155,39 @@
...
+      position: absolute;
+      width: 40px;
+      height: 40px;
+      border: none;
+      border-radius: 40px;
+      right: 16px;
+      top: 16px;
+      background: var(--base-10);
+      .chevron-right {
+        position: absolute;
```
lisasy, line 184: Instead of using absolute positioning with these values, you can vertically and horizontally position the `chevron-right` icon with flex-box. Here is an example of how you can do that here:

https://codepen.io/lisasy/pen/XWmjZKL

benhalpern:
Screen Shot 2020-04-20 at 5 51 46 PM

Seems kind of weird in this context eh?

Zhao-Andy:
@benhalpern ah good point. I think because `/mod` in desktop view will be rewritten into the Mod Center, it might make sense to start putting this stuff in a new view file and route, like `/:article_path/actions_panel` or something. Thoughts?

benhalpern:
@Zhao-Andy yeah, maybe a new route and use a partial so it can be in both places for now?

jacobherrington:
I like the idea of using a partial to house the shared markup. :+1: I'm curious about adding a new route, it might be weird if someone was to directly access that route, but I'm guessing that's a wart we can deal with later.

Zhao-Andy:
Closing and reopening to see if this will update properly...

Zhao-Andy:
>  it might be weird if someone was to directly access that route

Yeah, I was thinking about this too. I wonder if we can make that route visible via only by iframe?

PR #7396, opened 2020-04-20
Replace Cover Image <div> Tags With <img> Tags

### LINKED ISSUE #7289: Replace Cover Image <div> Tags with <img> Tags
We currently use `<div>`s with an `image` class and a `background-image:url` for cover images in place of `` tags, which is not semantic. The solution here is to replace all of the `<div>` tags with `` tags to make things more semantic in addition to boosting DEV's SEO. **This issue is a discrete part of issue #7228 and represents the initial work that needs to be done.**

Cover Image Files to Change:
- [x] `articles/show.html.erb`
- [x] `stories/_main_stories_feed.html.erb`

Other Non-Semantic Files:
- `comments/index.html.erb`
- `articles/_bottom_content.html.erb` (questionable)

juliannatetreault:
- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
We currently use `<div>`s with an image class and a `background-image:url` for cover images in place of `` tags, which is not very semantic. This PR not only replaces the `<div>`s with ``s, but also adjusts the styling of cover images for a particular article's show page accordingly.

**A few days post-merge, the background styling (`background` and `background-size`) will need to be removed from `article-show.scss` since it is now dead code. There is both a comment in the stylesheet, as well as an issue (seen below under "Related Tickets & Documents") for this code removal.**

Closes #7289 
Related To #7401 

**Article Show Page Before:**

Screen Shot 2020-04-17 at 3 39 18 PM
Screen Shot 2020-04-17 at 3 39 35 PM

**Article Show Page After:**

Screen Shot 2020-04-20 at 11 40 39 AM
Screen Shot 2020-04-20 at 11 39 47 AM

## Added tests?
- [ ] yes
- [ ] no, because I need help

## Added to documentation?
- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

CSS Family Guy

### DIFF
--- a/app/assets/stylesheets/article-show.scss
+++ b/app/assets/stylesheets/article-show.scss
@@ -59,16 +59,17 @@ article {
     margin: auto;
     max-width: 1024px;
     z-index: 2;
-    padding-top: 0;
+    background: transparent no-repeat center center; /* TODO[team-delightful]: Remove background styling a few days after merge to avoid conflicts */
+    background-size: cover;
     height: 42vh;
+
+    @media screen and (min-width: 880px) {
+      height: 370px;
+    }
 
     @media screen and (min-width: 950px) {
       border-top-left-radius: 2px;
       border-top-right-radius: 2px;
-    }
-
-    @media screen and (min-width: 880px) {
-      height: 370px;
     }
   }
 
--- a/app/assets/stylesheets/article-show.scss
+++ b/app/assets/stylesheets/article-show.scss
@@ -58,10 +58,8 @@ article {
     width: 100%;
     margin: auto;
     max-width: 1024px;
-    background: transparent no-repeat center center;
-    background-size: cover;
     z-index: 2;
-    padding-top: 42%;
+    padding-top: 14%;
 
     @media screen and (min-width: 950px) {
       border-top-left-radius: 2px;
--- a/app/assets/stylesheets/article-show.scss
+++ b/app/assets/stylesheets/article-show.scss
@@ -62,7 +62,7 @@ article {
     z-index: 2;
     background: transparent no-repeat center center;
     background-size: cover;
-    height: 42vh;
+    height: 42vw;
 
     @media screen and (min-width: 880px) {
       height: 370px;
--- a/app/assets/stylesheets/article-show.scss
+++ b/app/assets/stylesheets/article-show.scss
@@ -59,11 +59,16 @@ article {
     margin: auto;
     max-width: 1024px;
     z-index: 2;
-    padding-top: 14%;
+    padding-top: 0;
+    height: 42vh;
 
     @media screen and (min-width: 950px) {
       border-top-left-radius: 2px;
       border-top-right-radius: 2px;
+    }
+
+    @media screen and (min-width: 880px) {
+      height: 370px;
     }
   }
 
--- a/app/assets/stylesheets/article-show.scss
+++ b/app/assets/stylesheets/article-show.scss
@@ -53,13 +53,14 @@ article {
   padding: 0 0;
   position: relative;
 
+  /* TODO[team-delightful]: Remove background and background-size a few days post merge to avoid caching conflicts */
   .image {
     position: relative;
     width: 100%;
     margin: auto;
     max-width: 1024px;
     z-index: 2;
-    background: transparent no-repeat center center; /* TODO[team-delightful]: Remove background styling a few days after merge to avoid conflicts */
+    background: transparent no-repeat center center;
     background-size: cover;
     height: 42vh;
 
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/article-show.scss
```diff
@@ -58,10 +58,8 @@ article {
     width: 100%;
     margin: auto;
     max-width: 1024px;
-    background: transparent no-repeat center center;
-    background-size: cover;
     z-index: 2;
-    padding-top: 42%;
+    padding-top: 14%;
```
benhalpern, line 62: This element will probably work if you go `padding:0` and `height:42vh`, but for media query over 880px, go `height:420px` and that should ultimately make it work.

app/assets/stylesheets/articles.scss
```diff
@@ -680,10 +680,9 @@
     &.big-article {
       padding: 0px;
       .picture {
-        padding-top: 42%;
+        padding-top: 14%;
```
benhalpern, line 683: I feel like we can start with _just_ the article show page and remove this and the HTML adjustment. The index pages are less important for SEO, we _may_ want to do it but for the sake of making this PR simpler we might just want to focus on the show view for now.

app/assets/stylesheets/article-show.scss
benhalpern: ... Because this behavior is going to remain cached on some pages, I think we should also actually keep the `background` css here and remove it a few days after this gets merged, to make sure there are no conflicts.

So with these suggestions I'd say this can look like this....

```css
  .image {
    position: relative;
    width: 100%;
    margin: auto;
    max-width: 1024px;
    z-index: 2;
    background: transparent no-repeat center center;
    background-size: cover; 
    height: 42vw;
    @media screen and (min-width: 880px) {
      height: 370px;
    }
    @media screen and (min-width: 950px) {
      border-top-left-radius: 2px;
      border-top-right-radius: 2px;
    }
  }
```

And we'd be good to go I think.

benhalpern:
I _think_ with my suggestions this should be shippable.

app/assets/stylesheets/articles.scss
juliannatetreault: Okay, sounds reasonable to me! I'll revert these changes. 👍

app/assets/stylesheets/article-show.scss
juliannatetreault: Okay! I've made these changes already and will add the original code back as well. 😄

juliannatetreault:
> You will also need to update the following file,
> https://github.com/thepracticaldev/dev.to/blob/master/app/javascript/articles/FeaturedArticle.jsx#L48

Because I am no longer changing the `<div>` in `_main_stories_feed.html.erb` or any of that particular article's styling, is it still necessary that I change the `<div>` you highlighted?

benhalpern:
@juliannatetreault nope. I believe since we're punting on that part, @nickytonline's suggestion is no longer needed.

juliannatetreault:
> @juliannatetreault nope. I believe since we're punting on that part, @nickytonline's suggestion is no longer needed.

Okay, awesome! I just wanted to double-check. 👍

app/assets/stylesheets/article-show.scss
```diff
@@ -58,10 +58,14 @@ article {
     width: 100%;
     margin: auto;
     max-width: 1024px;
-    background: transparent no-repeat center center;
-    background-size: cover;
     z-index: 2;
-    padding-top: 42%;
+    background: transparent no-repeat center center; /* TODO[team-delightful]: Remove background styling a few days after merge to avoid conflicts */
```
vaidehijoshi, line 62: GREAT use of notes-to-self here :) Might want to just move this up to the top of the class definition and explicitly note that we can do this for both `background` and `background-cover`

vaidehijoshi: Also, I'd recommend creating an issue for this, and linking to it here. That way there will be more context for anyone who touches this, and you can track it with an issue! But this is just my personal workflow/style, so definitely not blocking.

juliannatetreault: Thank you! 😄 

In addition to making the comment slightly more explicit and moving it to the top of the class, I think opening an issue for this TODO is an excellent idea--I'm on it. 👍 👍

app/assets/stylesheets/article-show.scss
```diff
@@ -53,15 +53,20 @@ article {
...
     position: relative;
     width: 100%;
     margin: auto;
     max-width: 1024px;
+    z-index: 2;
     background: transparent no-repeat center center;
     background-size: cover;
-    z-index: 2;
-    padding-top: 42%;
+    height: 42vh;
```
benhalpern, line 65:
```suggestion
    height: 42vw;
```

Whoops, sorry typo if this is what I suggested.

PR #7402, opened 2020-04-20
Ensure Readinglist Reactions are Removed from Elasticsearch when Deleted

mstruve:
- [x] Bug Fix

## Description
When clearing out user data we use `.delete_all` on reactions which does not trigger any callbacks. This means any reading list reactions will be left in Elasticsearch. This PR updates the that delete class to remove the reactions from Elasticsearch after we have removed them from Postgres.

I also made sure to reindex reactions that were updated during a user merge since `update_all` also does not trigger any callbacks. 

https://github.com/thepracticaldev/dev.to/projects/6#card-36679085

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/services/users/delete_activity.rb
+++ b/app/services/users/delete_activity.rb
@@ -42,7 +42,7 @@ module Users
 
     def delete_profile_info(user)
       user.notifications.delete_all
-      user.reactions.delete_all
+      remove_reactions(user)
       user.follows.delete_all
       Follow.followable_user(user.id).delete_all
       user.messages.delete_all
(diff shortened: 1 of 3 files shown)


mstruve:
> there's also a user.reactions.delete_all here:

@Zhao-Andy this is only for readinglist reactions which I don't think you can give to a User unless I am mistaken.

Zhao-Andy:
@mstruve A user `has_many` reactions though so `user.reactions` would include their reading list reactions that they've left on articles.

mstruve:
OH shit I was thinking about it as reacting to a user 🤦‍♀️

PR #7405, opened 2020-04-20
Validate filenames server-side

atsmith813:
- [x] Feature

## Description
This PR validates filename lengths server-side. I'm going to circle back in another PR to remove the excessive Datadog logging and rescues.

https://github.com/thepracticaldev/dev.to/projects/9#card-36356085

create_article
create_comment
create_organization
create_podcast

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

eye_roll_gif

### DIFF
--- a/app/controllers/concerns/image_uploads.rb
+++ b/app/controllers/concerns/image_uploads.rb
@@ -1,0 +1,31 @@
+module ImageUploads
+  extend ActiveSupport::Concern
+
+  MAX_FILENAME_LENGTH = 250
+
+  def log_image_data_to_datadog
+    images = Array.wrap(
+      params.dig("user", "profile_image") ||
+      params.dig("podcast", "image") ||
+      params.dig("organization", "profile_image") ||
+      params["image"],
+    )
+
+    raise if images.empty?
+
+    images.each do |image|
+      tags = [
+        "controller:#{params['controller']}",
+        "action:#{params['action']}",
+        "content_type:#{image.content_type}",
+        "original_filename:#{image.original_filename}",
+        "tempfile:#{image.tempfile}",
+        "size:#{image.size}",
+      ]
+
+      DatadogStatsClient.increment("image_upload_error", tags: tags)
+    end
+
+    raise
+  end
+end
--- a/app/controllers/concerns/image_uploads.rb
+++ b/app/controllers/concerns/image_uploads.rb
@@ -3,6 +3,10 @@ module ImageUploads
 
   MAX_FILENAME_LENGTH = 250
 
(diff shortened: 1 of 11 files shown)


atsmith813:
Yeah, I'll move the method calls inside the controller action instead of `before_action`.

atsmith813:
I'm going to ignore codeclimate since this is just an interim solution anyways. In the (near?) future, these code chunks will be removed.

app/controllers/image_uploads_controller.rb
```diff
@@ -46,6 +53,12 @@ def cloudinary_link(uploaders)
 
   private
 
+  def valid_filename?
+    images = Array.wrap(params.dig("image"))
+    invalid_image = images.detect { |image| long_filename?(image) }
```
citizen428, line 58: Ruby has `Enumerable#none?`, which I think reads quite nicely here (you then won't need the last line, as the result of `none?` is the boolean we want):

```suggestion
    images.none? { |image| long_filename?(image) }
```

app/controllers/concerns/image_uploads.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def log_image_data_to_datadog
+    images = Array.wrap(
+      params.dig("user", "profile_image") ||
+      params.dig("podcast", "image") ||
+      params.dig("organization", "profile_image") ||
+      params["image"],
+    )
+
+    raise if images.empty?
```
rhymes, line 19: +1 on not raising a RuntimeError here

app/controllers/users_controller.rb
```diff
@@ -44,6 +44,12 @@ def edit
   # PATCH/PUT /users/:id.:format
   def update
     set_tabs(params["user"]["tab"])
+
+    unless valid_filename?
+      render :edit, status: :bad_request
```
rhymes, line 49: maybe `:unprocessable_entity` here?

atsmith813: I thought about that as well. But I wanted to keep it consistent with how the `update` action was already responding with errors:

https://github.com/thepracticaldev/dev.to/blob/e45363d63e0ed5a28db74b5dfe64c03d0a06c15b/app/controllers/users_controller.rb#L72

app/controllers/image_uploads_controller.rb
atsmith813: Love it - updated! Always game to be reminded of the nifty Ruby methods 💯. Thank you!

atsmith813:
Force pushed master just to double-check specs after the Travis CI debacle #7486

PR #7411, opened 2020-04-21
Refactor jbuilder

citizen428:

- [X] Refactor

## Description

While working on the `ClassifiedListing` refactoring I noticed that we're fairly inconsistent in our use of JBuilder (i.e. we use the more compact `extract!` in some places, but not in others). This PR makes it more consistent, introduces helpers as necessary, etc. Since I updated the relevant files for classified listings in the other PR, the are missing here.

n/a

## Added tests?

- [X] yes for new things, for the rest the old ones should suffice

## Added to documentation?

- [X] no documentation needed


citizen428:
@rhymes @maestromac What do you think about this? I know there's been some talk about eventually moving away from JBuilder, but until that day comes I'd at least like us to be a bit more consistent.

spec/helpers/articles_helper_spec.rb
```diff
@@ -53,4 +53,14 @@ def internal_navigation?
       it { is_expected.to include('height="18" width="18"') }
     end
   end
+
+  describe "#formated_date" do
+    it "correctly formats the date when present" do
+      expect(helper.formated_date(Time.now.to_i)).to match(/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z/)
```
rhymes, line 59: I think this needs to be `Time.current` and without `to_i`

app/helpers/articles_helper.rb
```diff
@@ -54,4 +54,8 @@ def get_host_without_www(url)
     host.gsub!("medium.com", "Medium")
     host.delete_prefix("www.")
   end
+
+  def formated_date(timestamp)
```
rhymes, line 58: missing `t`

I would also call this `utc_iso_timestamp` or something more explicit :D

app/views/api/v0/articles/_article.json.jbuilder
```diff
@@ -1,20 +1,25 @@
...
+  :collection_id,
+  :published_timestamp,
+)
+
+json.cover_image     cloud_cover_url(article.main_image)
+json.social_image    article_social_image_url(article)
+json.canonical_url   article.processed_canonical_url
+json.created_at      formated_date(article.created_at)
+json.edited_at       formated_date(article.edited_at)
+json.crossposted_at  formated_date(article.crosspost)
```
rhymes, line 23: `article.crossposted_at`

app/views/api/v0/articles/_article.json.jbuilder
```diff
@@ -1,20 +1,25 @@
...
+json.published_at    formated_date(article.published)
```
rhymes, line 24: `article.published_at`

citizen428: Same as above, my macro messed up 😆

spec/helpers/articles_helper_spec.rb
citizen428: It works with either, but agreed, `Time.current` is closer to how the AR timestamps look.

citizen428:
@rhymes Sorry for adding one more commit after the review, but I noticed the `content_truncated` attribute on response templates yesterday and confirmed with @Zhao-Andy  that it's no longer needed so this PR seemed like the most natural place to remove it.

app/views/api/v0/articles/_flare_tag.json.jbuilder
```diff
@@ -1,5 +1,3 @@
 json.flare_tag do
-  json.name             flare_tag.name
-  json.bg_color_hex     flare_tag.bg_color_hex
-  json.text_color_hex   flare_tag.text_color_hex
+  json.extract!(flare_tag, :name, :bg_color_hex, :text_color_hex)
```
juliannatetreault, line 2: I like the use of `.extract!` here

citizen428:
@juliannatetreault 

> Quick, non-blocking question: why is this file (app/views/api/v0/articles/me.json.jbuilder) named me.json.jbuilder

JBuilder works like views, so follows usual Rails naming conventions. This means it corresponds to the `me` action in `Api::V0::ArticlesController`, which you can find in the API docs [here](https://docs.dev.to/api/#operation/getUserArticles).

PR #7412, opened 2020-04-21
Snackbars Improvements

ludwiczakpawel:

- [x] Refactor

## Description

This PR changes the way Snackbars are stacked on top of each other.

image

## Added tests?

## Added to documentation?

- [x] no documentation needed

### DIFF
--- a/app/views/pages/crayons.html.erb
+++ b/app/views/pages/crayons.html.erb
@@ -1116,9 +1116,7 @@
           <p>You have just downvoted this article.</p>
         </div>
         <div class="crayons-snackbar__actions">
-          <div class="crayons-btn-group">
-            <button type="button" class="crayons-btn crayons-btn--secondary">Undo</button>
-          </div>
+          <button type="button" class="crayons-btn crayons-btn--ghost-success crayons-btn--inverted">Undo</button>
         </div>
       </div>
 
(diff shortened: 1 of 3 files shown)


ludwiczakpawel:
@nickytonline Just FYI...  I know we've been discussing how snackbars should be stacked on top of each other and this is against what we agreed on. But @lisasy convinced me to this style :)
Also, I left you few comments.

app/views/pages/crayons.html.erb
```diff
@@ -1116,9 +1116,7 @@
           <p>You have just downvoted this article.</p>
         </div>
         <div class="crayons-snackbar__actions">
-          <div class="crayons-btn-group">
```
ludwiczakpawel, line 1119: @nickytonline Can we require using only [INVERTED](https://github.com/thepracticaldev/dev.to/blob/master/app/javascript/crayons/Button/Button.jsx#L31-L33) buttons? If yes, then we can easily drop `crayons-btn-group` from markup even when using multiple buttons.

Ghost buttons don't really need that and we should always be using only ghost buttons in snackbar.

nickytonline: We can since it's in the internals of the <SnackbarItem /> component.

PR #7416, opened 2020-04-21
Ensure Records are Updated Properly after a User Merge

mstruve:
- [x] Bug Fix

## Description
When we merge content for users we use `update_all` which does not trigger any callbacks, This can cause things to get out of sync in Elasticsearch and with scores and caches. This PR moves all that callback syncing into a job that can run after updating the records in Postgres. 

https://github.com/thepracticaldev/dev.to/projects/6#card-36680670

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/spec/support/sidekiq_test_helpers.rb
+++ b/spec/support/sidekiq_test_helpers.rb
@@ -73,6 +73,11 @@ module SidekiqTestHelpers
     end
   end
 
+  # Perform all Sidekiq jobs until there are no longer any in the queues
+  def drain_all_sidekiq_jobs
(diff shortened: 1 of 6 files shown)


spec/support/sidekiq_test_helpers.rb
```diff
@@ -73,6 +73,11 @@ def sidekiq_perform_enqueued_jobs(only: nil, except: nil)
     end
   end
 
+  # Perform all Sidekiq jobs until there are no longer any in the queues
+  def drain_all_sidekiq_jobs
```
atsmith813, line 77: +1 for this name!

PR #7418, opened 2020-04-21
Throttle General Site Requests

mstruve:
- [x] Bug Fix
- [x] Optimization

## Description
Don't allow more than 10 requests per second from a single IP address. I am open to suggestions on the limit number. 10 felt like it was super lenient and safe but I could see us ratcheting it down to 5 or less. I just didn't want to bork any background requests Fastly might be making during a page load or things like that. 

## Added tests?
- [x] no, because the time frame on this is so low and the limit so high we can't fire off 10 requests in sequence before the key expires. Even though we use Timecop, there is an expiration set on the keys in Redis and that does not adhere to any freezing from the Timecop helper. 

alt_text

### DIFF
--- a/config/initializers/rack/attack.rb
+++ b/config/initializers/rack/attack.rb
@@ -17,7 +17,7 @@ class Rack::Attack
     end
   end
 
-  throttle("site_hits", limit: 20, period: 1) do |request|
+  throttle("site_hits", limit: 100, period: 2) do |request|
     if request.env["HTTP_FASTLY_CLIENT_IP"].present?
       request.env["HTTP_FASTLY_CLIENT_IP"].to_s
     end
--- a/config/initializers/rack/attack.rb
+++ b/config/initializers/rack/attack.rb
@@ -16,4 +16,10 @@ class Rack::Attack
       request.env["HTTP_API_KEY"]
     end
   end
+
+  throttle("site_hits", limit: 10, period: 1) do |request|
+    if request.env["HTTP_FASTLY_CLIENT_IP"].present?
+      request.env["HTTP_FASTLY_CLIENT_IP"].to_s
+    end
+  end
 end
--- a/config/initializers/rack/attack.rb
+++ b/config/initializers/rack/attack.rb
@@ -17,7 +17,7 @@ class Rack::Attack
     end
   end
 
-  throttle("site_hits", limit: 10, period: 1) do |request|
+  throttle("site_hits", limit: 20, period: 1) do |request|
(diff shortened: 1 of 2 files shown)


benhalpern:
If I load the home page I actually make async 10 requests immediately to `dev.to` as a total default (so not counting async requests for assets or third party etc.)

So we'd probably need more like 20 to be safe?

Is it possible to base this based on whether the `session_current_user_id` is present? Unauthenticated traffic typically has a lot fewer legit async requests.

I'm not sure where this fits in middleware in terms of what we could have available, but `session_current_user_id` is a pretty simple method designed for our "metal" controllers, so might be a situation where `session["warden.user.user.key"].first.first`...

```ruby
  def session_current_user_id
    return unless session["warden.user.user.key"]

    # the value is in the format [[1], "..."] where 1 is the ID
    session["warden.user.user.key"].first.first
  end
```

mstruve:
> So we'd probably need more like 20 to be safe?

sounds good!

> Is it possible to base this based on whether the session_current_user_id is present?

Yes, we can definitely do that. I think regardless of whether you are authenticated or not though the limit should be the same. The limit we set should be considered beyond what anyone with normal use should reach whether they are logged in or not.

config/initializers/rack/attack.rb
```diff
@@ -16,4 +16,10 @@ class Rack::Attack
       request.env["HTTP_API_KEY"]
     end
   end
+
+  throttle("site_hits", limit: 20, period: 1) do |request|
```
benhalpern, line 20: In thinking about this now, even 20 in one is not enough for how we handle at least one poorly-handled situation:

Open up the network tab and go here: https://dev.to/search?q=ben&filters=class_name:User

I think it's worth refactoring this situation to handle a bulk following request that handles _all_ the follow buttons on the page at once for whether there is a follow like how we do reactions on comment thread pages....

But for now it would cause problems here.

Either we merge this now as something like `limit: 75, period: 1)`

Or even `limit: 100, period: 3)` or something like that...

But we could hold off and try to adjust the way we bulk-fetch follow checks? I could probably whip up the long-overdue change.

mstruve: lol oh that's a good one, about 100 requests in 2 seconds. Though a bunch of them look like they are being made to cloudinary and not dev. I like 100 for a period of 3 and I [made a ticket to optimize that situation](https://github.com/thepracticaldev/dev.to/projects/9#card-36804382)

PR #7421, opened 2020-04-21
Style UpVote, DownVote and Vomit buttons

msarit:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR styles the upvote, downvote and vomit reactions in the new Mod Actions Panel. Only UI changes made; a separate PR will ensure that the appropriate points are added/taken away, as appropriate.

Note: in this PR, one can select BOTH upvote and downvote. The selection is indicated by the class`.reacted`. One solution would be to remove all `.reacted` classes first, before adding the class to the chosen reaction. However, the `.reacted` class is added when the method `Reaction.cached_any_reactions_for?` returns true. I will need some help with this aspect of the voting feature.

none

https://www.loom.com/share/40acabd4014447f8a3dbd467e01d7f96

## Added tests?

- [ ] yes
- [ ] no, because I need help
- [X] no, not at this stage

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [X] no documentation needed

none

Black man smiling and dusting hands off


app/assets/stylesheets/moderators.scss
```diff
@@ -213,19 +213,38 @@
...
     border-radius: 12px;
   }
+
+  // Reaction Buttons START
+
+  .thumb-reactions {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: center;
+    margin-top: 20px;
```
Zhao-Andy, line 223: Let's use the space unit variables here. :) Lisa's comments on my other PR were really helpful here: https://github.com/thepracticaldev/dev.to/pull/7395

With space unit variables, this would be `margin-top: $su-5;`. You can find them in `_import.scss`

app/assets/stylesheets/moderators.scss
```diff
@@ -213,19 +213,38 @@
...
-    margin: 30px 30px;
-    border-radius: 200px;
     border: 0px;
-    padding: 12px 12px;
+    display: flex;
+    flex-direction: column;
+    flex-wrap: wrap;
+    width: min-content;
+    height: min-content;
+    padding: 16px 24px;
```
Zhao-Andy, line 236: Also space unit variables here, and anywhere else that uses spacing.

app/assets/stylesheets/moderators.scss
```diff
@@ -213,19 +213,38 @@
...
+    border-radius: 3px;
```
Zhao-Andy, line 237: There's also a border radius variable! `$br-default`

msarit:
Thanks Andy! I will address 😅

app/assets/stylesheets/moderators.scss
```diff
@@ -213,36 +213,205 @@
...
+    border: solid 2px var(--base-20);
+    border-radius: 100%;
+    background: $white;
+  }
+
+  .reaction-button-circle img {
+    width: $su-7;
+  }
+
+  img.upvote {
```
lisasy, line 269: Rather than manipulate another image to use the "upvote" emoji, I can export one for you so we can use that.

msarit: Sounds good Lisa! If it's exported by tomorrow (Thurs Apr 23 EST) then I'll use it. Thanks!

app/assets/stylesheets/moderators.scss
```diff
@@ -213,36 +213,205 @@
...
+    color: var(--base-90);
+  }
+
+  .reaction-checkmark {
+    opacity: 0;
     position: absolute;
-    top: 5px;
-    width: 70px;
-    left: 5px;
+    top: 63px;
```
lisasy, line 288: These values are very pixel-specific and would be hard to maintain, but I'm wondering if it would be possible for you to move the `.reaction-checkmark` so that it's a child element of `.reaction-button-circle`, and then set the absolute position of `.reaction-checkmark` relative to .reaction-button-circle. It might look like this:

```
.reaction-button-circle {
  position: relative;
  .position-checkmark {
    position: absolute;
    right: 0;
    bottom: 0;
  }
}
```

This way, we always know that the `.reaction-checkmark` is pinned to the bottom-right of the `.reaction-circle-button`

app/assets/stylesheets/moderators.scss
```diff
@@ -213,36 +213,205 @@
...
     border-color: white;
-    margin: 30px 30px;
-    border-radius: 200px;
-    border: 0px;
-    padding: 12px 12px;
+    border: 0;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    width: 156px;
```
lisasy, line 232: An alternative way to setting the position of the two `reaction-button` element would be the grid layout. Doing this would mean not needing to declare a width value for the button because it would be based on the parent element's padding. Here is an illustration of how you can do it: https://codepen.io/lisasy/pen/xxwgOEO

app/assets/stylesheets/moderators.scss
```diff
@@ -213,36 +213,205 @@
...
+    height: $su-5;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    background-color: var(--accent-success);
+    border-radius: 100%;
+    border: 2px solid $white;
+    padding: 2px;
+    z-index: 3;
+    color: $white;
```
lisasy, line 300: `$white` is a variable in the color system we're eventually going to deprecate. Instead, you can use` var(--base-inverted)` here 🎨

app/assets/stylesheets/moderators.scss
```diff
@@ -213,36 +213,205 @@
...
+  }
+
+  .vomit-reaction-section,
+  .how-this-works-section {
+    display: flex;
+    justify-content: center;
+    margin: 0 $su-4 $su-4 $su-4;
```
lisasy, line 307: I believe that you can write this shorthand like this:

`margin: 0 $su-4 $su-4;`

`margin: <top> <left and right> <bottom>`

app/assets/stylesheets/moderators.scss
```diff
@@ -213,36 +213,205 @@
...
+  }
+
+  .reaction-vomit-button {
+    display: flex;
+    align-items: center;
+    border: none;
+    background: $white;
```
lisasy, line 314: same $white comment as above

app/assets/stylesheets/moderators.scss
```diff
@@ -213,36 +213,205 @@
...
+    height: $su-7;
+    padding: $su-1 $su-2;
+    border-radius: $br-default;
+
+    img {
+      position: absolute;
+      left: 107px;
```
lisasy, line 321: is it possible to use flex box here instead of absolute positioning?

app/assets/stylesheets/moderators.scss
```diff
@@ -213,36 +213,205 @@
...
+      width: $su-6;
+      height: $su-6;
+    }
+
+    span {
+      margin-left: 35px;
```
lisasy, line 327: is it possible to use flex box here instead of `margin-left: 35px` to set the position?

app/assets/stylesheets/moderators.scss
```diff
@@ -213,36 +213,205 @@
...
+    border-radius: 100%;
+    background: $white;
+  }
+
+  .reaction-button-circle img {
+    width: $su-7;
+  }
+
+  img.upvote {
+    -webkit-transform: rotate(180deg);
```
nickytonline, line 270: The browser vendor prefixes shouldn't be necessary. Using just `transform` should be fine.

app/assets/stylesheets/moderators.scss
lisasy: Here is a cleaner way to position the elements for Mark as Spam:

https://codepen.io/lisasy/pen/xxwgOEO

app/assets/stylesheets/moderators.scss
msarit: Thanks Nick! Lisa gave me the images for the upvote, so I'm going to use those instead of using this transformation.

app/assets/stylesheets/moderators.scss
```diff
@@ -213,36 +213,181 @@
...
   }
+
+  // Reaction Buttons START
+
+  .reactions-container {
+    padding: $su-4;
+  }
+
+  .thumb-reactions-container {
+    display: grid;
```
lisasy, line 224: Nice :)

app/assets/stylesheets/moderators.scss
```diff
@@ -213,36 +213,181 @@
...
   }
 
-  .reaction-button img {
-    position: absolute;
-    top: 5px;
-    width: 70px;
-    left: 5px;
+  .reaction-button-circle {
+    position: relative;
+    display: flex;
```
lisasy, line 252: Amazing!

app/assets/stylesheets/moderators.scss
msarit: 💃

app/assets/stylesheets/moderators.scss
msarit: 💃

PR #7425, opened 2020-04-21
Allow Reaction Search to Use AND Operator for Tag Searching

mstruve:
- [x] Feature
- [x] Bug Fix

## Description
Last piece of the puzzle before I PR the frontend Readinglist code! When a user filters by tags on their Reading List we want to use the AND operator between those tags. For example a search for #ruby and #rails should only return articles with BOTH tags.

https://github.com/thepracticaldev/dev.to/projects/6#card-36754138

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/services/search/query_builders/reaction.rb
+++ b/app/services/search/query_builders/reaction.rb
@@ -64,7 +64,13 @@ module Search
         TERM_KEYS.map do |term_key, search_key|
           next unless @params.key? term_key
 
-          { terms: { search_key => Array.wrap(@params[term_key]) } }
+          values = Array.wrap(@params[term_key])
+
+          if params[:tag_boolean_mode] == "all"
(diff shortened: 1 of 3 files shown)


app/services/search/query_builders/reaction.rb
```diff
@@ -64,7 +64,13 @@ def terms_keys
         TERM_KEYS.map do |term_key, search_key|
           next unless @params.key? term_key
 
-          { terms: { search_key => Array.wrap(@params[term_key]) } }
+          values = Array.wrap(@params[term_key])
+
+          if params[:tag_boolean_mode] == "all"
```
mstruve, line 69: Ideally, I would like to add this to all search query builders but I will save that for a future refactor. Then in the view we can allow users to toggle between AND and OR options for searching multiple items.

PR #7250, opened 2020-04-13
Use new classified listing categories

citizen428:
## What type of PR is this? (check all applicable)

- [X] Refactor
- [X] Generalization

## Description

This is the second PR in the process of generalizing classified listings. PR #7172 introduced a model, this PR starts the cleanup process and switches our code to use it. Since this is live already, I'm trying to do this carefully, i.e. most methods in `ClassifiedListing` that wrapped the constant are still there and access the DB instead. We can refactor this in a follow-up PR.

## Related Tickets & Documents

#6740 

## Added tests?

- [X] updated existing one

## Added to documentation?

- [X] no documentation needed

## What gif best describes this PR or how it makes you feel?

image

PR #7432, opened 2020-04-22
Filter classified listings by published status for mods

citizen428:
## What type of PR is this?

- [X] Refactor
- [X] Feature

## Description

This PR adds filtering by published status to `/internal/listings`. Unpublished listings will be filtered out by default.

Closes https://github.com/thepracticaldev/tech-private/issues/424

ezgif com-video-to-gif

## Added tests?

- [X] yes

## Added to documentation?
- [x] no documentation needed


app/controllers/internal/classified_listings_controller.rb
```diff
@@ -18,14 +19,14 @@ def update
     update_listing_details
     clear_listings_cache
     flash[:success] = "Listing updated successfully"
-    redirect_to "/internal/listings/#{@classified_listing.id}/edit"
+    redirect_to edit_internal_listing_path(id: @classified_listing.id)
```
citizen428, line 22: Since I was here already...

app/controllers/internal/classified_listings_controller.rb
```diff
@@ -18,14 +19,14 @@ def update
...
   end
 
   def destroy
     @classified_listing = ClassifiedListing.find(params[:id])
     @classified_listing.destroy
     flash[:warning] = "'#{@classified_listing.title}' was destroyed successfully"
-    redirect_to "/internal/listings"
+    redirect_to internal_listings_path
```
citizen428, line 29: Same as above.

app/views/internal/classified_listings/index.html.erb
```diff
@@ -11,7 +11,12 @@
         </div>
         <div class="form-group">
           <%= label_tag(:filter, "Category") %>
-          <%= select_tag(:filter, options_for_select([""] + ClassifiedListing.categories_available.keys, params[:filter])) %>
+          <%= select_tag(:filter, options_for_select(ClassifiedListing.categories_available.keys, params[:filter]), include_blank: true) %>
```
citizen428, line 14: Changed this to use the more standard `include_blank` option.

app/views/internal/classified_listings/index.html.erb
```diff
@@ -11,7 +11,12 @@
...
+        </div>
+        <div class="form-group">
+          <%= label_tag(:published, "Published only") %>
+          <%= hidden_field_tag(:published, 0) %>
```
citizen428, line 18: Unlike `f.check_box` `check_box_tag` does not take care of this by itself, so we have to add the unchecked option ourselves via `hidden_tag`.

app/controllers/internal/classified_listings_controller.rb
```diff
@@ -4,6 +4,7 @@ class Internal::ClassifiedListingsController < Internal::ApplicationController
 
   def index
     @classified_listings = ClassifiedListing.includes(%i[user organization]).page(params[:page]).order("bumped_at DESC").per(50)
+    @classified_listings = @classified_listings.published unless include_unpublished?
```
citizen428, line 7: @jacobherrington Do we actually use the search box of the form anywhere?

app/views/internal/classified_listings/index.html.erb
```diff
@@ -11,7 +11,12 @@
...
+          <%= check_box_tag(:published, 1, true) %>
```
lightalloy, line 19: If I uncheck the checkbox and click "filter", the checkbox will be checked after the filtering. I would expect it to remain unchecked after unchecking. How do you feel about making it checked or unchecked based on the corresponding param (when it exists)?

msarit:
I second what Anna said about the `Published only` checkbox bug. Here is a brief video with my observations:

https://www.loom.com/share/198b5444f1ab4465afbd1461e1fcb2cf

app/views/internal/classified_listings/index.html.erb
msarit, line 19: I second what Anna has flagged re: `Published only` checkbox bug; I captured my observations in the video below - hope it's helpful! 🙏 

https://www.loom.com/share/198b5444f1ab4465afbd1461e1fcb2cf

citizen428: The reason I did that was as a workflow optimization for @michael-tharrington. My rationale:

1. Most of the time he wants to see the filtered list.
2. If he really wants to see all listings he can uncheck the box and hit "filter".
3. If he wants to get back to the filtered list, he only needs to press "filter" instead of fiddling with the check box. 

I see how it's confusing though, so will change it.

citizen428: 1. What the checkbox is supposed to do is to toggle between all listings (including published "no") and only the ones with a published status of "yes". It should never only show unpublished listings. Maybe the label of the checkbox is not great, so I'm open to suggestions here  😃 

2. The checkbox behavior was somewhat intentional, as this is an internal tool with one main user (@michael-tharrington) and this is going to save him one click. Given that it confused two people though I'll change it to behave more like a traditional checkbox.

BTW, @lightalloy was only referencing point 2. above in her comment, not point 1.

citizen428:
Thanks for the feedback @lightalloy and @msarit. Based on that I made the following changes:

1. Changed the label to "Include unpublished listings".
2. Changed the params name to `include_unpublished`.
3. Flipped the logic around accordingly.
4. Make the checkbox respect the param.
5. Resolve merge conflicts caused by merging #7250.

Here's a recording of how it works now, the listing names illustrate the intended behavior:

app/views/internal/classified_listings/index.html.erb
msarit: ahhhhh ok I understand. Yes the checkbox label didnt make it very clear to me. Thanks for explaining.

PR #7438, opened 2020-04-22
Generalization: Replace static dev emails with configurable ones.

Ridhwana:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Replace static dev emails with configurable emails:
- [x] yo@dev.to => replace with SiteConfig.default_site_email
- [x]  partners@dev.to, partnerships@dev.to, biz@dev.to - use 1 email for all of these, add a configurable option, for DEV should be partners@dev.to
- [x]  members@dev.to - used only in events, probably this legacy feature will be disabled for other communities once we have feature toggles
- [x]  privacy@dev.to - additional config

I have added a specific email_addresses section to the site config, we have set fields to capture the default, business, privacy and members. It contains the following hash structure:
```
 field :email_addresses, type: :hash, default: {
    default: "yo@dev.to",
    business: "partners@dev.to",
    privacy: "privacy@dev.to",
    members: "members@dev.to"
  }
```

Note: We will write a script that will remove the defaults in subsequent PR's so that other communities can setup more easily.
 
Something to consider:
In devise we set the email to yo@dev.to https://github.com/thepracticaldev/dev.to/blob/b2e18aeb72c69ac21dd178b20dad0cb6a5710d79/config/initializers/devise.rb#L14 We need to make this general but the SiteConfig may be loaded later than the devise initializer, which poses a problem. Will need to bring this up with the team - feel free to drop any suggestions here too if you have any. 
  
https://github.com/thepracticaldev/dev.to/issues/7413

Screenshot 2020-04-22 at 14 27 28

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

1. Notify #new-communities once this is in.

alt_text


app/controllers/internal/configs_controller.rb
```diff
@@ -26,7 +26,7 @@ def create
 
   def config_params
     allowed_params = %i[
-      default_site_email mascot_user_id
```
Ridhwana, line 29: removed default_site_email and used a hash of email_addresses for the different types

app/models/site_config.rb
```diff
@@ -28,6 +27,14 @@ class SiteConfig < RailsSettings::Base
     twitch: nil
   }
 
+  # Emails
+  field :email_addresses, type: :hash, default: {
+    default: "yo@dev.to",
+    business: "partners@dev.to",
+    privacy: "privacy@dev.to",
+    members: "members@dev.to"
```
Ridhwana, line 35: used only in events, probably this legacy feature will be disabled for other communities once we have feature toggles

app/views/credits/new.html.erb
```diff
@@ -37,7 +37,7 @@
     </div>
       <br />
     </div>
-    <p class="bulk-description"><em>Contact <a href="mailto:biz@dev.to">biz@dev.to</a> for custom bulk pricing and partnerships.</em></p>
+    <p class="bulk-description"><em>Contact <a href="mailto:<%= SiteConfig.email_addresses[:business] %>?subject=Custom Bulk Pricing/Partnerships"><%= SiteConfig.email_addresses[:business] %></a> for custom bulk pricing and partnerships.</em></p>
```
Ridhwana, line 40: Added an aside subject line here just to make the UX better

app/views/internal/configs/show.html.erb
```diff
@@ -105,6 +97,32 @@
           </div>
         </div>
 
+        <div class="card mt-3">
```
Ridhwana, line 100: Split out the form sections, one for social media and one sections holds the emails.

app/views/internal/configs/show.html.erb
```diff
@@ -105,6 +97,32 @@
...
+                     locals: {
+                       header: "Emails",
+                       state: "collapse",
+                       target: "emailBodyContainer",
+                       expanded: "false"
+                     } %>
+          <div id="emailBodyContainer" class="card-body collapse hide" aria-labelledby="emailBodyContainer">
+           <div class="form-group">
+             <%= f.fields_for :email_addresses do |email_field| %>
+               <% SiteConfig.email_addresses.keys.each do |email_type| %>
```
lightalloy, line 111: I still insist on `SiteConfig.email_addresses.each do |email_type, email| ...end` instead of `keys.each` :D

Ridhwana: @lightalloy amended as per your suggestion

Ridhwana:
@citizen428 I like both those changes, but If you don't mind, since there's no breaking changes could you possibly approve this PR, and I'll create a new one this morning for those refactors. Keeping up with this PR has been a bit challenging with all the merge conflicts and I can test afresh with a new one.

citizen428:
Sorry @Ridhwana, missed this message. No problem at all, do the follow-up PR whenever you have time/feel like it.

PR #4004, opened 2019-09-10
Use Redis for session storage

Zhao-Andy:
## What type of PR is this? (check all applicable)
- [x] Feature
- [x] Bug Fix

## Description
This adds Redis to our stack, but only for session storage. We'll need to add the following ENV variables:
```
variable :REDIS_URL, :String, default: "redis://localhost:6379/0"
variable :SESSION_KEY, :String, default: "DevCommunity_Session"
variable :SESSION_EXPIRY_SECONDS, :String, default: "1209600" # two weeks
```

We'll also need to add a Redis server to our production stack.

PR #7439, opened 2020-04-22
Authentication refactoring: improve testing on OmniauthCallbacksController

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

The main goal of this PR is to start splitting out and refactoring authentication code that's specific to either GitHub or Twitter in their own modules. There are still a few things to do:

- [x] introduce `Authenticator`, extract GitHub and Twitter providers, add tests - https://github.com/thepracticaldev/dev.to/pull/7259
- [x] test OmniauthCallbacksController in depth (this PR)
- [ ] have a concept of "available providers" and "enabled providers"
- [ ] enable UI and functionality based on which providers are enabled
- [ ] detach the provider identities from being used to fetch external data (the github client and twitter client use random identities in the DB to fetch data, this can't work anymore if auth providers are pluggable, they should use one set of keys for all)

Once these are done, we can start adding additional auth providers with additional peace of mind and surety.

**This is step 2**

**While testing this I noticed how there are 2 paths in the controller that we never get to. I tried for a while to enter those logic paths with tests but I wasn't able to reproduce them, I wrote down why as comments. I would like to remove them but I want to be extra careful about them. so please review the code.**

Screenshot of the local coverage after running both Twitter and Github's tests:

Screenshot_2020-04-22 Code coverage for Devto

I'll check CodeClimate when the build is done to see if they are covered somehow.

## Added tests?

- [x] yes
- [ ] no, because I need help


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -44,38 +49,54 @@ def callback_for(provider)
       cta_variant: cta_variant,
     )
 
-    if persisted_and_valid?
+    if user_persisted_and_valid?
+      # TODO: [thepracticaldev/oss] remove this
       # delete legacy session based cookie once the user has logged in again
       request.cookie_jar.delete(LEGACY_COOKIE_NAME) if request.cookies[LEGACY_COOKIE_NAME]
```
rhymes, line 55: We switched to the new cookie in October 2019 in #4004 

I think we can remove this code, let me know if there are any objections

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -35,7 +38,9 @@ def failure
 
   private
 
-  # TODO: [thepracticaldev/oss] test all of this
+  # TODO: [thepracticaldev/oss] investigate if there's a way to simplify this
+  # by using Devise+OmniAuth builtin integration
+  # <https://github.com/heartcombo/devise/wiki/OmniAuth:-Overview>
```
rhymes, line 43: OmniAuth has builtin support for Devise, I wonder if there's a way to have less code or not, I'll investigate post refactoring as there's no need to slow down the extraction of providers for this

benhalpern:
Looks like the transaction code will need to be de-conflicted

rhymes:
> Looks like the transaction code will need to be de-conflicted

ahah thanks! love how adding an empty line threw off Git :-(

app/controllers/omniauth_callbacks_controller.rb
mstruve: DO IT!

app/controllers/application_controller.rb
```diff
@@ -38,6 +38,8 @@ def customize_params
     params[:signed_in] = user_signed_in?.to_s
   end
 
+  # This method is used by Devise to decide which is the path to redirect
+  # the user to after a successful log in
```
mstruve, line 42: I feel like this comment isn't really needed, I think the method name is pretty self-explanatory.

spec/system/authentication/user_logs_in_with_github_spec.rb
```diff
@@ -82,6 +121,35 @@
         )
       end
     end
+
+    context "when a validation failure occurrs" do
+      before do
+        OmniAuth.config.mock_auth[:github].extra.raw_info.name = "X" * 101
```
vaidehijoshi, line 127: What is this doing? Specifically the `"X" * 101`?

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -44,38 +49,54 @@ def callback_for(provider)
...
       flash[:alert] = user_errors
       redirect_to new_user_registration_url
     end
   rescue StandardError => e
-    Rails.logger.error "Log in error: #{e}"
-    Rails.logger.error "Log in error: auth data hash - #{request.env['omniauth.auth']}"
-    Rails.logger.error "Log in error: auth data hash - #{request.env['omniauth.error']&.inspect}"
+    Rails.logger.error("Log in error: #{e}")
+    Rails.logger.error("Log in error: auth data hash - #{request.env['omniauth.auth']}")
+    Rails.logger.error("Log in error: auth data hash - #{request.env['omniauth.error']&.inspect}")
```
vaidehijoshi, line 89: This is more for my own edification, but why do we log to the Rails logger here instead of sending this data to DataDog? (Maybe this is a silly question but seems like when we rescue errors we tend to send them to DataDog?)

app/controllers/application_controller.rb
rhymes: I added because I didn't know it was used TBH. I stumbled upon it casually by doing a debug session step by step. If you grep for it there's no code that calls it, it's called internally by Devise, that's why I added the comment :D

spec/system/authentication/user_logs_in_with_github_spec.rb
rhymes: Good point, let me add a comment about it. We have a name validator that renders names invalid if they are over 100 chars, so i created a string 101 chars long

app/controllers/omniauth_callbacks_controller.rb
rhymes: Asked myself the same question, @mstruve should we change that? I would but maybe not in this PR. We also need to figure out if we actually use those two logic paths we don't seem to use at all

mstruve: The next project on my list for SRE is to move our logging from Timber to Datadog so everyone can see it and then come up with a logging framework that makes things more searchable beyond message strings. Lets leave this for now, I will be overhauling the logging soon. Anything we want to give access to the whole team immediately feel free to send to Datadog as a metric.

PR #7440, opened 2020-04-22
Execute Reading List searches in Elasticsearch

mstruve:
- [x] Feature

## Description
This PR updates the frontend of the Reading List code to execute searches through Elasticsearch instead of Algolia. Couple things to note.
- Bug fix: We were filtering for only valid status reactions before, however, any reading list reaction created by an admin will have the status `confirmed` so we need to search for either. 
- Removed History item: I grep'ed the code based for `visited_at` and found no references to it in the code which led me to believe we are not ever serving "history items" to the view. With that assumption, I removed all history item code. If we do indeed use this code please let me know!
- Bug fix: Make sure `tag_boolean_mode` only applies to tags

CodeClimate Refactor Failures: I know exactly how to refactor these but I would like to keep that for a separate PR since this has enough in it already. 

https://github.com/thepracticaldev/dev.to/projects/6#card-32289399

## Added tests?
- [x] yes
I added some tests for generating the reading list items. If anyone has suggestions for other tests to add let me know! If they are frontend suggestions please include details or examples I can pull from 😉 

alt_text


atsmith813:
Did we intentionally leave off the trailing wildcard search in #7308?

mstruve:
@atsmith813 yes that was intentional. When we are searching full body analyzed text we want to search for exact matches. When we are doing name searching like tags we want to autocomplete as they type so we use the wildcard to accomplish that.

PR #7441, opened 2020-04-22
Add some placeholder pages

Ridhwana:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
As per slack [conversation](https://dev-internal.slack.com/archives/CSYHV3NJ2/p1587377079153800): 

Some of our mailers reference pages that come from the Pages db entries (i.e. pages created from the internal admin). These include 
- /community-moderation, 
- /post-a-job, 
- /tag-moderation, and 
- post-a-job. 

Since e-mails will have broken links without these pages, and it's not obvious to other communities that they need to create these backend pages, I've added placeholders for these too, like Anna did with FAQ's, about and bounty. 

We'd then let communities know (through perhaps some setup doc or banner etc.) to override these placeholders. These links become more discoverable instead of just failing silently 🤫

There is also some small refactors with pathnames.

https://github.com/thepracticaldev/dev.to/projects/15#card-36810021

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

alt_text


Ridhwana:
@maestromac can you give the code climate a peek. I tried pushing another commit in the hope that it will clear the error.

maestromac:
Hey @Ridhwana 
Those coverage reports are accurate. You wrote additional codes that were not tested.

image

While those checks are not mandatory, it might be a good idea to add an additional test in `spec/requests/pages_spec.rb`

Ridhwana:
> Hey @Ridhwana
> Those coverage reports are accurate. You wrote additional codes that were not tested. See https://codeclimate.com/github/thepracticaldev/dev.to/compare/3d985319dd53c37705c3d2e407084872c98ab523...c93a61b41b08bc7c1cbf35fc3e89ceaa6f05e588#ratings
> 
> While those checks are not mandatory, it might be a good idea to add an additional test in `spec/requests/pages_spec.rb`

Oh gosh, sorry I didn't pick that up. I'll add those tests 👍  thank you

app/controllers/pages_controller.rb
```diff
@@ -25,6 +25,30 @@ def bounty
     set_surrogate_key_header "bounty_page"
   end
 
+  def community_moderation
```
mstruve, line 28: Would it be helpful to create these pages in the seeds file for local environments?

Ridhwana: I personally think that the placeholders suffice. We don't really need these files on our local environment as its pretty much static information.

Ridhwana:
> Is the end game to eventually remove the links from the emails or do we plan on making these pages required for a new app?

I don't have a conclusive answer for this. For now they are all required but I don't think it makes sense to have all of these required for the app if they are not being used, however I believe that in order to determine which should be required, we'd need to evaluate the use cases of new communities. I think we'll get there in the coming weeks.

app/controllers/pages_controller.rb
```diff
@@ -13,10 +13,10 @@ def about
     set_surrogate_key_header "about_page"
   end
 
-  def faq
-    @page = Page.find_by(slug: "faq")
+  def about_listings
+    @page = Page.find_by(slug: "about-listings")
     render :show if @page
-    set_surrogate_key_header "faq_page"
+    set_surrogate_key_header "about_listings"
```
rhymes, line 19: is this the only key that doesn't end in `_page`  ?

app/views/pages/about_listings.html.erb
```diff
@@ -0,0 +1 @@
+<%= render partial: "placeholder", locals: { title: "About DEV Listings" } %>
```
rhymes, line 1: should this be `"about #{community_name} listings"` ?

app/views/pages/post_a_job.html.erb
```diff
@@ -0,0 +1 @@
+<%= render partial: "placeholder", locals: { title: "Posting a Job on DEV Listings" } %>
```
rhymes, line 1: should this use `community_name` ?

Ridhwana:
@rhymes these all need to be implemented. Great catches. Can we merge this one and I'll update the surrogate key (if it won't have an impact on the site, I'm not sure? have to still do some research, do you know?) and the generalisation a little later today once I'm done with my current branch work.

PR #7442, opened 2020-04-22
Bug Fix: Use strings with colons for all Datadog tags

mstruve:
- [x] Bug Fix

## Description
Using symbols and a hash for tags seems to work OK for `gauge` metrics, however, they do not work for increment and count metrics properly. I am sure the underlying gem code has some differences for each scenario. Rather than worrying about what type of metric we are sending to Datadog we should get in the habit of always using an array with strings.

Since I updated the existing specs so I am inclined to ignore the CodeClimate failure in this case. 

## Added tests?
- [x] updated

alt_text

### DIFF
--- a/app/labor/mailchimp_bot.rb
+++ b/app/labor/mailchimp_bot.rb
@@ -182,7 +182,7 @@ class MailchimpBot
 
   def report_error(exception)
     Rails.logger.error(exception)
-    DatadogStatsClient.increment("mailchimp.errors", tags: [action: "failed", user_id: user.id, source: "gibbon-gem"])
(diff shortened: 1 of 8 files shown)


rhymes:
Nice! 

Idea for the future: we could wrap `DatadogStatsClient.increment` into a key-value based utility, something like `Monitoring.increment_metric(metric_name, key: value, key2: value2)` which internally calls `.to_s` to each parameter and sends it to the `DatadogStatsClient`

mstruve:
@rhymes  you read my mind! I was definitely thinking about that as well, I love the idea and maybe raising an error if the wrong format is passed causing specs to break and people to fix the problem rather than us discovering it in Datadog.

PR #7443, opened 2020-04-22
Remove Onboarding's ClosingSlide

maestromac:
- [x] Feature

## Description
- Removing the ClosingSlide
- Upon clicking next on the `FollowUsers` slide, it will redirect the user to the homepage.

The only discrepancy is that "next" button on the final page should say "finish" instead but that's will be addressed in different PR.
Follow up to https://github.com/thepracticaldev/dev.to/pull/7443
TBD
## Added tests?
- [x] yes/updated

## Added to documentation?
- [x] no documentation needed

n/a


app/assets/javascripts/initializers/initializeOnboardingTaskCard.js.erb
```diff
@@ -8,8 +8,7 @@ function initializeOnboardingTaskCard() {
   if (localStorage.getItem('task-card-closed') === 'yes') { return }
 
   var taskCard = document.getElementsByClassName("onboarding-task-card")[0];
-
-  if (taskCard == null) { return }
+  if (taskCard == null) { return } // This guard against both null and undefined taskCard
```
vaidehijoshi, line 11:
```suggestion
  if (taskCard == null) { return } // This guards against both a null and undefined taskCard
```

app/javascript/onboarding/__tests__/Onboarding.test.jsx
```diff
@@ -311,18 +312,32 @@ describe('<Onboarding />', () => {
...
+
+      expect(window.location.href).toBe(url);
       onboardingSlides.find('.next-button').simulate('click');
       await flushPromises();
+      // No longer advance slide.
       expect(onboardingSlides.state().currentSlide).toBe(
-        emailPreferencesFormIndex + 1,
+        emailPreferencesFormIndex,
       );
+      expect(window.location.href).toBe('/');
```
vaidehijoshi, line 340: This seems odd; why does the `onboardingSlides.state().currentSlide` still point to the `emailPreferencesFormIndex`, while the `window.location.href` is also `/`? Seems like both of them cannot be true at the same time?

app/javascript/onboarding/Onboarding.jsx
```diff
@@ -47,6 +45,8 @@ export default class Onboarding extends Component {
       this.setState({
         currentSlide: nextSlide,
       });
+    } else {
+      window.location.href = '/';
```
vaidehijoshi, line 49: Might just want to include a comment here since it feels a bit random to change the window's location to `/` without any context

app/javascript/onboarding/__tests__/Onboarding.test.jsx
maestromac: I'm not too sure what context the Jest test environment has because technically there shouldn't be any more onboardingSlides due to the redirect. To check if `window.location.href` change is essential but making sure the slide number of the state didn't change is just to account for unintended behavior.

app/javascript/onboarding/Onboarding.jsx
maestromac: We can refine this in https://github.com/thepracticaldev/dev.to/issues/6544

rhymes: you can also use window.location.assign("/") which is a little bit more explicit

vaidehijoshi: I'm going to be in the onboarding code today and will be making a PR, so I can swap `href` out for `assign`!

PR #7444, opened 2020-04-22
Add Retry-After for API rate limit errors and document limits in API spec

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

- Added `Retry-After` to tell "the machine" when they can retry a request, see https://httpstatuses.com/429

- Added docs about the two API endpoints that an return `HTTP 429 Too Many Requests`

Segue to #7422 

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [x] docs.dev.to
- [ ] readme
- [ ] no documentation needed


app/controllers/api/v0/api_controller.rb
```diff
@@ -37,8 +35,9 @@ def error_not_found
     render json: { error: "not found", status: 404 }, status: :not_found
   end
 
-  def too_many_requests
-    render json: { error: "too many requests", status: 429 }, status: :too_many_requests
+  def error_too_many_requests(exc)
+    response.headers["Retry-After"] = exc.retry_after
```
rhymes, line 39: 🧙

spec/requests/image_uploads_spec.rb
```diff
@@ -99,15 +98,20 @@
...
         upload = proc do
           Rack::Test::UploadedFile.new(
-            Rails.root.join("spec/support/fixtures/images/images1.jpeg"), "image/jpeg"
+            Rails.root.join("spec/support/fixtures/images/image1.jpeg"),
+            "image/jpeg",
           )
         end
-        expect do
-          11.times { post "/image_uploads", headers: headers, params: { image: [upload.call] } }
-        end.to raise_error(RuntimeError)
```
rhymes, line 110: we were raising `RateLimitChecker::UploadRateLimitReached` and sending back a JSON response in `ImageUploadsController` but this test was never changed. The reason why we never noticed it is that the test never failed as it was looking for `images1.jpeg` (instead of `image1.jpeg`) which is a file that does not exist on disk, which eventually raised `RuntimeError` which was compatible with the expectation set. 

👀

spec/requests/image_uploads_spec.rb
```diff
@@ -54,7 +54,7 @@
         post "/image_uploads", headers: headers, params: { image: [image, image2] }
 
         expect(response.body).to match("\/i\/.+\.")
-        expect(JSON.parse(response.body)["links"].length).to eq(2)
+        expect(response.parsed_body["links"].length).to eq(2)
```
mstruve, line 57: 🔥

app/controllers/image_uploads_controller.rb
```diff
@@ -46,6 +53,15 @@ def cloudinary_link(uploaders)
 
   private
 
+  def rate_limit!
```
mstruve, line 56: I wonder if we could move all these `rate_limit!` methods into the `RateLimitChecker` since it seems like their logic is all very similar.

rhymes: hahaha @mstruve I thought about it but I didn't want to revolutionize too much. I'll make a note for another PR

app/labor/rate_limit_checker.rb
```diff
@@ -22,11 +38,13 @@ def limit_by_action(action)
   end
 
   def track_image_uploads
-    Rails.cache.increment("#{@user.id}_image_upload", 1, expires_in: 30.seconds)
+    expires_in = RETRY_AFTER[:image_upload].seconds
+    Rails.cache.increment("#{@user.id}_image_upload", 1, expires_in: expires_in)
```
tamcv, line 42: Hi @rhymes, I would like to report a problem here:
Rails version 5.2.4.2 (latest version of 5.2) does not support `expires_in` for `Rails.cache.increment` for redis_cache_store. As a result, value will be expired after 1 hour (by default), not 30 seconds as expected.
https://github.com/rails/rails/blob/v5.2.4.2/activesupport/lib/active_support/cache/redis_cache_store.rb#L262
This was fixed in master branch:
https://github.com/rails/rails/blob/master/activesupport/lib/active_support/cache/redis_cache_store.rb#L277
Please help to correct me if I'm wrong...

rhymes: @tamcv omg amazing find! Though our default expiration is 24 hours, not 1 hour:

https://github.com/thepracticaldev/dev.to/blob/1e5d831a3ed6339e9886c8300d418bd7bce13ac0/config/environments/production.rb#L115-L116


I tested with:

```ruby
[3] pry(main)> Rails.cache.increment("rhymes_counter_x", 1, expires_in: 45.seconds)
Cache increment: rhymes_counter_x ({:amount=>1})
=> 1
```

```shell
➜  ~ redis-cli get rhymes_counter_x && redis-cli ttl rhymes_counter_x
"1"
(integer) -1
```

and indeed there's no expiration date :( 

I'm going to merge this as is a it also refactors the code, would you like to open a separate bug issue about this and propose a solution? I personally would like to keep the atomic `INCRBY` separation but we can't upgrade to Rails 6 just yet, maybe there's a way to have both the cake and eat it too :D

rhymes: ah @tamcv, a possible avenue (thanks @maestromac for the suggestion) would be move these rate limits up to `rack-attack`, see https://github.com/thepracticaldev/dev.to/blob/master/config/initializers/rack/attack.rb

what do you think?

tamcv: Yes @rhymes , Rack::Attack is the correct way to throttle the requests.

tamcv: @rhymes I create new issue here: https://github.com/thepracticaldev/dev.to/issues/7503
In my opinion, we might need to clone the patch instead of moving the limit to `rack-attack` because `rack-attack` throttle the requests based on `ip_address`, there is no `user_id` at that step...

PR #7447, opened 2020-04-22
Specify SEO-Optimized Width and Height for Cover Images

### LINKED ISSUE #7228: Ensure Cover Images Use Image Tags and Adhere to the 16:9 Ratio
We currently use `<div>`s with an `image` class and a `background-image:url` for cover images on DEV. However, there is a better solution which is to replace the `<div>`s with `` tags. In addition to this, it is imperative that we ensure that cover images adhere to the 16:9 ratio, with a width of around 1200px for SEO purposes. To improve DEV's overall SEO, it is important that we optimize cover images within the existing JSON-LD data structure.

**This issue represents an overarching task that is made up of three discrete parts.** The discrete parts are represented by the following open issues:
- Part One: Issue #7289 (Done ✔️ )
- Part Two: Issue #7290 (Done ✔️ )
- Part Three: Issue #7291 

### LINKED ISSUE #7291: Add Structured Data for SEO-Optimized Images
Once we have replaced `<div>` tags with `` tags ( issue #7289 ) and have SEO-optimized cover images ( issue #7290 ), it is important that we add the SEO-optimized images to the JSON-LD within `articles/show.html.erb`. The hope is that by adding the images to the data structure, DEV's overall SEO will improve.  **This issue is a discrete part of issue #7228 and represents the final piece of the work that needs to be done.**

Changes to Make:
- [x] Enforce the 1:1 ratio for cover images via `articles/show.html.erb`'s JSON-LD
- [x] Enforce the 4:3 ratio for cover images via `articles/show.html.erb`'s JSON-LD
- [x] Enforce the 16:9 ratio for cover images via `articles/show.html.erb`'s JSON-LD

juliannatetreault:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
This PR represents the final work necessary to optimize cover images for SEO purposes. In order to ensure that cover images adhere to a `1:1`, `4:3`, and `16:9` ratio, I specified the necessary `height`s and `width`s for cover images in the JSON-LD found in `stories_controller.rb`, using the aspect ratio logic I set up in `social_image.rb` in my last PR (https://github.com/thepracticaldev/dev.to/pull/7417).

Closes #7291 
Closes #7228 

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed
- [x] inline documentation will be added in a future PR

That's All Folks

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -373,7 +373,7 @@ class StoriesController < ApplicationController
       "image": [
         ApplicationController.helpers.article_social_image_url(@article, width: 1080, height: 1080),
         ApplicationController.helpers.article_social_image_url(@article, width: 1280, height: 720),
-        ApplicationController.helpers.article_social_image_url(@article, width: 1920, height: 1080),
+        ApplicationController.helpers.article_social_image_url(@article, width: 1600, height: 900),
       ],
       "publisher": {
         "@context": "http://schema.org",
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -370,7 +370,11 @@ class StoriesController < ApplicationController
         "@id": URL.article(@article)
       },
       "url": URL.article(@article),
-      "image": ApplicationController.helpers.article_social_image_url(@article),
+      "image": [
+        ApplicationController.helpers.article_social_image_url(@article, width: 1080, height: 1080),
+        ApplicationController.helpers.article_social_image_url(@article, width: 1280, height: 720),
+        ApplicationController.helpers.article_social_image_url(@article, width: 1920, height: 1080),
+      ],
       "publisher": {
         "@context": "http://schema.org",
         "@type": "Organization",
--- a/app/helpers/social_image_helper.rb
+++ b/app/helpers/social_image_helper.rb
@@ -17,8 +17,8 @@ module SocialImageHelper
     listing_social_preview_url(listing, format: :png)
   end
 
-  def article_social_image_url(article)
-    Articles::SocialImage.new(article).url
+  def article_social_image_url(article, **options)
+    Articles::SocialImage.new(article, options).url
   end
 
   def comment_social_image_url(comment)


app/controllers/stories_controller.rb
```diff
@@ -370,7 +370,11 @@ def set_article_json_ld
         "@id": URL.article(@article)
       },
       "url": URL.article(@article),
-      "image": ApplicationController.helpers.article_social_image_url(@article),
+      "image": [
+        ApplicationController.helpers.article_social_image_url(@article, width: 1080, height: 1080),
+        ApplicationController.helpers.article_social_image_url(@article, width: 1280, height: 720),
+        ApplicationController.helpers.article_social_image_url(@article, width: 1920, height: 1080),
```
benhalpern, line 376: Is there a reason this needs to be so big?

juliannatetreault: I was just following a JSON-LD example for this, but I am happy to try making this more efficient!

benhalpern: I'm inclined to follow the example all else equal, can you link to it?

juliannatetreault: Sure thing! it can be found here: https://developers.google.com/search/docs/data-types/article

Now that I have access to the aspect ratio code, I'm going to work on making this work. 😄

benhalpern: 1920x1080 just seems a bit bigger than we'd realistically _need_ to provide based on the suggestions, why don't we do an even 1600x900? It's kind of arbitrary but that would work.

juliannatetreault: OH! I think I misunderstood you. 😅 I can make that change, sure!

app/controllers/stories_controller.rb
```diff
@@ -370,7 +370,11 @@ def set_article_json_ld
         "@id": URL.article(@article)
       },
       "url": URL.article(@article),
-      "image": ApplicationController.helpers.article_social_image_url(@article),
+      "image": [
+        ApplicationController.helpers.article_social_image_url(@article, width: 1080, height: 1080),
```
Ridhwana, line 374: @juliannatetreault  nice one, I've been learning more about SEO through your PR's :D 

On that note, some thoughts on this:
When using this helper would it not be useful to enforce the ratio through a maths calculation rather than just allowing a width and height as input. Perhaps the input parameters could be the ratio and width, and the height is calculated dynamically. The reason being is that for someone not aware that the numbers are chosen specifically for SEO benefits, it might be very easy for them to change the numbers without realising the impact.

benhalpern: @Ridhwana that's a good thought in terms of communicating intent. I still think passing in the actual h/w values provides more flexibility here.

But maybe this array could be extracted as a named method like `seo_optimized_images` so it's clear enough. And then even a comment that links to the Google guide.

In general with all of these json ld methods it might be useful to have a comment that includes the url of the guide because that's the context that one would need to really understand how and why to modify this.

@juliannatetreault I'd still vote to ship this and consider refactoring after.

juliannatetreault: Thanks so much, @Ridhwana ! I'm happy to hear that you're learning alongside me. 😄

Regarding your thought: I am happy to look into the possibility of using a calculation for this to lessen confusion for contributors and the team, but I was also thinking maybe me adding a comment in the code to help alleviate the confusion in the meantime could be a useful solution?

juliannatetreault: That's a good suggestion @benhalpern and I can open an issue and the card to the project board so that I can get this shipped and refactor later. 😄

Ridhwana: This sounds good 🎉its definitely not a blocker.

PR #7449, opened 2020-04-22
Fix task card unstyled issue for rollout

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Because we store styles in the app shell with service workers, certain features can roll out where the body is not lined up with the assets from the shell. This is kind of rare, but a problem we have not yet fully tackled.

This adds some in-html styling to ensure the `display:none` attribute is paid attention to here.

Currently some scenarios result in this...

Screen Shot 2020-04-22 at 2 03 24 PM

### DIFF
--- a/app/views/onboardings/_task_card.html.erb
+++ b/app/views/onboardings/_task_card.html.erb
@@ -10,6 +10,11 @@
     localStorage.setItem('task-card-closed', 'yes');
   }
 </script>
+<style>
+.onboarding-task-card {
+  display: none;
+}
+</style>
 
 <div class="onboarding-task-card">
   <button class="close" onclick="closeTaskCard()">


app/views/onboardings/_task_card.html.erb
```diff
@@ -10,6 +10,11 @@
     localStorage.setItem('task-card-closed', 'yes');
   }
 </script>
+<style>
```
vaidehijoshi, line 13: Should we include a comment here so we don't accidentally remove this going forward?

benhalpern: Yeah this couldn't hurt. But of note: We actually _can_ remove this after not too long. The comment should probably give all the context. I should have included that.

But a week from now we should safely feel like we can remove this.

This is the PR meant to deal with this but I sort of let it fall off... I wasn't totally sure I liked the solution but it would be good to revisit this shortly...

https://github.com/thepracticaldev/dev.to/pull/6301

vaidehijoshi: Ah, gotcha. If it's not long term, I feel like we don't need to bother  🙃

PR #7450, opened 2020-04-22
Add a progress stepper to onboarding flow

vaidehijoshi:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
✨ Adds the world's _cutest-ever_ progress stepper to the onboarding slides! ✨

The four main "interactive" steps of the onboarding flow now are represented by a little progress stepper that shows up in the navigation of the onboarding slides.

As a part of this PR, I have also cleaned up the onboarding JS tests as well. 💪 

Addresses the first half of https://github.com/thepracticaldev/dev.to/issues/6544. (I will address the skip/continue feature in a later PR! 😄)

To QA this, you'll want to step through the onboarding flow locally.

You'll want to check that the stepper does _not_ render for the intro slide:
Screen Shot 2020-04-22 at 12 11 16 PM

You'll also want to make sure that the stepper changes based on which slide you're looking at!

For example, the first slide looks like this:
Screen Shot 2020-04-22 at 12 11 26 PM

And the last slide looks like this:
Screen Shot 2020-04-22 at 12 11 38 PM

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] ~no documentation needed~ **I added inline docs!**

glittery computer face

### DIFF
--- a/app/javascript/onboarding/Onboarding.jsx
+++ b/app/javascript/onboarding/Onboarding.jsx
@@ -14,9 +14,6 @@ export default class Onboarding extends Component {
     const url = new URL(window.location);
     const previousLocation = url.searchParams.get('referrer');
 
-    this.nextSlide = this.nextSlide.bind(this);
-    this.prevSlide = this.prevSlide.bind(this);
-
     const slides = [
       IntroSlide,
       FollowTags,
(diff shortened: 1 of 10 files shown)


vaidehijoshi:
Hm, is it possible to turn off Code Climate for `propTypes = {` definitions? I don't think we can really "refactor" that.

I also could use help from someone in manually approving Code Climate because it's only other complaint is the size of a `render` function, which I feel like, personally, is good to know about...but probably shouldn't block merging?

app/javascript/onboarding/__tests__/Onboarding.test.jsx
```diff
@@ -83,7 +102,7 @@ describe('<Onboarding />', () => {
     });
 
   describe('IntroSlide', () => {
-    const introSlideIndex = 0;
+    const slideIndex = slides.indexOf('IntroSlide');
```
Zhao-Andy, line 105: Love this refactor 👍

app/javascript/onboarding/Onboarding.jsx
```diff
@@ -46,6 +49,7 @@ export default class Onboarding extends Component {
         currentSlide: nextSlide,
       });
     } else {
+      // Redirect to the main feed at the end of onboarding.
```
maestromac, line 52: Thanks for adding this!

vaidehijoshi: You got it! 😉

PR #7452, opened 2020-04-22
Generalize Onboarding's task-card image

maestromac:
- [x] Generalization?

## Description
Generalize Onboarding's task-card image
n/a
It looks the same?
image

## Added tests?
- [x] not needed

## Added to documentation?
- [x] no documentation needed

n/a


juliannatetreault:
It looks like I'm running into a caching issue (maybe?) and I'm having a hard time getting the task card to show locally. 😕

vaidehijoshi:
@juliannatetreault what is your user's `created_at`, and also, did you dismiss the card locally before? if so, you'll have a key in your local storage and you'll need to manually delete. (feel free to ping if you have questions on this!)

maestromac:
Sorry all, if you already dismiss the card the in the pass. You need to bring it back by doing
```
localStorage.setItem("task-card-closed", undefined)
```
in your browser console

juliannatetreault:
> @juliannatetreault i would double check what your user's `created_at` is. also, did you dismiss the card locally before? if so, you'll have a key in your local storage and you'll need to manually delete. (feel free to ping if you have questions on this!)

My user's `created_at` timestamp was correct, it was just a local storage issue that @maestromac reached out to me about! Thanks though!

app/models/site_config.rb
```diff
@@ -46,6 +46,7 @@ class SiteConfig < RailsSettings::Base
   field :logo_png, type: :string, default: "https://practicaldev-herokuapp-com.freetls.fastly.net/assets/devlogo-pwa-512.png"
   field :logo_svg, type: :string, default: ""
   field :primary_sticker_image_url, type: :string, default: "https://practicaldev-herokuapp-com.freetls.fastly.net/assets/rainbowdev.svg"
+  field :onboarding_taskcard_sticker, type: :string, default: "https://practicaldev-herokuapp-com.freetls.fastly.net/assets/staggered-dev.svg"
```
Ridhwana, line 49: @maestromac we should probably be adding this new field to the Site config view too so that other communities can configure this value. I'm not sure what the scope of your task is but if it's not within the scope, you can probably add it as an issue in Team OSS's generalization pipeline, or ping me to add it :) .

maestromac: Thanks for pointing that out! I went ahead and implemented it. 

Re-requesting review

app/controllers/internal/configs_controller.rb
```diff
@@ -26,19 +26,37 @@ def create
...
+      periodic_email_digest_min
+      primary_sticker_image_url
+      rate_limit_comment_creation
+      rate_limit_email_recipient
+      rate_limit_follow_count_daily
+      rate_limit_image_upload
+      rate_limit_published_article_creation
+      shop_url
+      sidebar_tags
+      suggested_tags
```
Ridhwana, line 59: @maestromac this is probably a style preference and it has unfortunately come up multiple times in PR's 🙈 but it  makes it a lot easier to read and work with when the parameters are grouped by section as it was previously with a column limit. Personally I think one per line causes the file to become unnecessarily long.

Ridhwana:
A final note that I just remembered now, as a post deployment task can you please ping Lee about this change on new-communities. It's always useful to just give him a heads up so that he can set the correct config :)

PR #7454, opened 2020-04-22
Remove Algolia from User Model

mstruve:
- [x] Refactor 🔪 

## Description
I decided to remove Algolia from all the FeedContent models in separate PRs to make them a little more digestible and so if any bugs pop up they are easier to track down and fix. 

Codeclimate can be ignored. 

https://github.com/thepracticaldev/dev.to/projects/6#card-32955055

## Added tests?

alt_text


app/models/user.rb
```diff
@@ -580,7 +528,7 @@ def check_for_username_change
   end
 
   def conditionally_resave_articles
-    Users::ResaveArticlesWorker.perform_async(id) if core_profile_details_changed? && !user.banned
```
rhymes, line 583: how did it work until now? :D

benhalpern: 🙃

Screen Shot 2020-04-24 at 6 38 22 PM

PR #7483, opened 2020-04-23
Don't Allow Travis to Return False Successes When Specs Fail

mstruve:
- [x] Bug Fix

## Description
Code snippet from [travis trouble shooting docs](https://docs.travis-ci.com/user/common-build-problems/#ruby-rspec-returns-0-even-though-the-build-failed) bc ya know its not that big of a deal if your suite tells you all is well when shit is broken 🤦‍♀️ 

alt_text

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ script:
   - bundle exec rails webpacker:compile
   - bundle exec rails db:schema:load
   - './cc-test-reporter before-build'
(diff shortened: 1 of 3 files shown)


rhymes:
@mstruve thanks! While it's in draft, after we verified it works, let's also add the link of what that comes from in the code as a comment so that we don't have some odd looking code without context in the code base.

Let's hope this works :cry

PR #7487, opened 2020-04-23
Fix Fastly rake task spec and Travis CI

atsmith813:
- [x] Bug Fix

## Description
Following up on #7486, this should fix Travis CI and the Fastly spec. RSpec in Travis CI doesn't like the use of `exit`, so I changed it to `next`.

#7486 

## Added tests?

## Added to documentation?
- [x] no documentation needed

whiteclaw_gif

### DIFF
--- a/lib/tasks/search.rake
+++ b/lib/tasks/search.rake
@@ -13,7 +13,7 @@ namespace :search do
   task destroy: :environment do
     if Rails.env.production?
       puts "Will NOT destroy indexes in production"
-      exit
+      next
     end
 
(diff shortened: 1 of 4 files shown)


lib/tasks/search.rake
```diff
@@ -13,7 +13,7 @@ namespace :search do
   task destroy: :environment do
     if Rails.env.production?
       puts "Will NOT destroy indexes in production"
-      exit
+      next
```
atsmith813, line 16: May as well since I'm here 🤷‍♂️

PR #7488, opened 2020-04-23
Add skip/continue functionality to onboarding navigation

vaidehijoshi:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Updates the "next" button in the onboarding slides in the three "interactive" components: `FollowTags`, `ProfileForm`, `FollowUsers`. 

For these three slides, the text reads "Skip for now" unless the user has taken an action on the page. If they have (for example, selected a tag, or filled out a field on the form), the button text will update to read "Continue". The button state should return back to "Skip for now" if the user reverts their action (for example, hasn't selected a tag or removes all the text from the form inputs).

Also refactors the `ProfileForm` component to nest values on the `state` within a `formValues` object in order to make it easier to determine whether any of the form inputs have a value or not.

Closes https://github.com/thepracticaldev/dev.to/issues/6544.

To QA this feature, you'll want to pull down this branch and check the `FollowTags` and `FollowUsers` components, and make sure that the button text updates correctly:
Screen_Shot_2020-04-23_at_11_28_07_AM
Screen_Shot_2020-04-23_at_11_28_19_AM

For the `ProfileForm` component, you'll want to check that the button text updates whenever _any_ input in the form is filled out. **Keep in mind that you need to click off the input, since we're not updating the text on each keystroke, just when the input's value changes.**
Screen Shot 2020-04-23 at 11 28 51 AM
Screen Shot 2020-04-23 at 11 28 42 AM

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed **(but I added inline docs!)**

### DIFF
--- a/app/assets/stylesheets/preact/onboarding-modal.scss
+++ b/app/assets/stylesheets/preact/onboarding-modal.scss
@@ -141,7 +141,7 @@
       $button-size: 40px; // Explicitly set size for accessibility.
 
       border-radius: 100%;
-      width: $button-size;
+      min-width: $button-size;
       height: $button-size;
       align-items: center;
       display: flex;
--- a/app/assets/stylesheets/preact/onboarding-modal.scss
+++ b/app/assets/stylesheets/preact/onboarding-modal.scss
@@ -124,6 +124,7 @@
 
     button {
       @extend .crayons-btn;
+      width: $su-10; // Explicitly set width to avoid buttons re-sizing with text.
     }
 
     .back-button {
@@ -135,9 +136,15 @@
       display: flex;
     }
 
+    .skip-for-now {
+      @extend .crayons-btn;
+      @extend .crayons-btn--secondary;
+    }
+
     .stepper {
       display: flex;
       align-items: center;
+      justify-content: center;
     }
 
     .dot {
--- a/app/assets/stylesheets/preact/onboarding-modal.scss
+++ b/app/assets/stylesheets/preact/onboarding-modal.scss
@@ -124,7 +124,13 @@
 
     button {
       @extend .crayons-btn;
-      width: $su-10; // Explicitly set width to avoid buttons re-sizing with text.
+    }
+
+    .back-button-container,
+    .next-button {
+      // Explicitly set width to avoid next-button re-sizing with text,
+      // and to ensure that back-button-container takes up the same amount of space.
+      width: $su-10;
     }
 
     .back-button {
@@ -132,8 +138,18 @@
       @extend .crayons-btn--ghost;
       @extend .crayons-btn--icon;
 
+      $button-size: 40px; // Explicitly set size for accessibility.
+
+      border-radius: 100%;
+      width: $button-size;
+      height: $button-size;
       align-items: center;
       display: flex;
+
+      &:hover,
+      &:active {
+        background: var(--base-10);
+      }
     }
 
     .skip-for-now {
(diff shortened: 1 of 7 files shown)


app/javascript/onboarding/components/ProfileForm.jsx
```diff
@@ -45,21 +49,31 @@ class ProfileForm extends Component {
...
 
-    this.setState({
-      [name]: value,
-    });
+    currentFormState[name] = value;
+
+    // Once we've derived the new form values, check if the form is empty
+    // and use that value to set the `canSkip` property on the state.
+    const formIsEmpty =
+      Object.values(currentFormState).filter((v) => v.length > 0).length === 0;
```
vaidehijoshi, line 61: I don't love this, but it works! I am open to other ways to write this :)

vaidehijoshi:
Included one additional feature in this PR since it was easy enough to do! The final onboarding slide now has special button text, which reads "Finish". You can see this reflected in the last commit. This was something we decided on with @lisasy's input (and isn't reflected in the design):

Screen Shot 2020-04-23 at 12 02 37 PM

vaidehijoshi:
While waiting for review on this, I noticed that the back button was growing in size/width because of changes introduced in this PR. I have pushed up a commit to fix this, and I've also addressed [an accessibility concern](https://github.com/thepracticaldev/dev.to/pull/7455#discussion_r413913391) that @lisasy brought up in a recent PR!

app/javascript/onboarding/components/ProfileForm.jsx
maestromac: I like it 😆 . I can't think of anything better

vaidehijoshi:
I noticed an annoying thing and didn't want to merge/deploy this until I fixed it. When I previously fixed my arrow styles, I caused a regression that misaligned the stepper such that it was no longer centered: 
Screen_Shot_2020-04-23_at_3_13_05_PM

I fixed this regression and amended my previous commit, so that everything is centered as per the designs :)
Screen_Shot_2020-04-23_at_3_14_17_PM

app/javascript/onboarding/__tests__/__snapshots__/Onboarding.test.jsx.snap
```diff
@@ -7,21 +7,23 @@ preact-render-spy (1 nodes)
...
+          <button
+            onClick={[Function bound prevSlide]}
+            class="back-button"
+            type="button"
           >
-            <path d="M3.828 7H16v2H3.828l5.364 5.364-1.414 1.414L0 8 7.778.222l1.414 1.414L3.828 7z"></path>
-          </svg>
-        </button>
+            <svg
+              width="24"
```
lisasy, line 17: @vaidehijoshi Yay! 😀

app/assets/stylesheets/preact/onboarding-modal.scss
```diff
@@ -126,18 +126,41 @@
...
+
     .back-button {
       @extend .crayons-btn;
       @extend .crayons-btn--ghost;
       @extend .crayons-btn--icon;
 
+      $button-size: 40px; // Explicitly set size for accessibility.
+
+      border-radius: 100%;
+      width: $button-size;
```
lisasy, line 144: I think that instead of width which is a fixed value, setting this as `min-width: $button-size` may ensure that button is always a minimum size.

lisasy:
One small comment on using `min-width` instead of `width `to set the button size, but overall, it LGTM!

app/assets/stylesheets/preact/onboarding-modal.scss
vaidehijoshi: Ah, I wasn't really thinking that the width was ever going to change/grow for this button, since it has an explicitly-sized `svg` within it, which is why I set it as `width` directly 🤷. But I've gone ahead and changed it to `min-width`.

PR #7489, opened 2020-04-23
Remove Algolia from Podcast Episodes w/ Specs 😉

mstruve:
- [x] Refactor 🔪 

## Description
Specs passing for reallz this time!!!

I decided to remove Algolia from all the FeedContent models in separate PRs to make them a little more digestible and so if any bugs pop up they are easier to track down and fix. 

https://github.com/thepracticaldev/dev.to/projects/6#card-32955089

## Added tests?

alt_text

### DIFF
--- a/app/decorators/podcast_episode_decorator.rb
+++ b/app/decorators/podcast_episode_decorator.rb
@@ -34,4 +34,8 @@ class PodcastEpisodeDecorator < ApplicationDecorator
       podcastImageUrl: image_url
     }
   end
+
+  def published_at_int
+    published_at.to_i
+  end
 end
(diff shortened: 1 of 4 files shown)


app/decorators/podcast_episode_decorator.rb
```diff
@@ -34,4 +34,8 @@ def mobile_player_metadata
       podcastImageUrl: image_url
     }
   end
+
+  def published_at_int
```
rhymes, line 38: how's this move not breaking anything? 🔬 
I guess we call `.decorate` already somewhere :D

benhalpern: Yeah, seems like we only call this in one place and it's after we use `PodcastEpisodeDecorator.decorate_collection`

PR #7494, opened 2020-04-23
Refactor SEO Optimized Images

### LINKED ISSUE #7479: Refactor SEO Optimized Images
I recently shipped some work (https://github.com/thepracticaldev/dev.to/pull/7447) to optimize images on DEV for SEO purposes. While the work in this PR is good, it could be better. In order to make the work that I've done more explicit and in order to avoid future confusion from the community and team, a refactor is necessary. To accomplish this, I need to do the following:

- [x] Pull the SEO optimized images array into a method 
- [x] Add inline documentation that links to the SEO docs for clarification

juliannatetreault:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description
I recently shipped some work (#7447) to optimize images on DEV for SEO purposes. This PR addresses the concern brought up in the PR that the code was not explicit enough--contributors and team members might not be aware that the images are sized for a specific reason, which might result in someone accidentally altering the sizes. To make the code more explicit, I extracted the images into a method, `seo_optimized_images`, and added inline documentation to give curious contributors and team members more background information on the code.

**Side note**: I did notice that there were some inconsistencies in the `stories_show_spec.rb` while I was working and took care of fixing some of those in this PR as well. 😊 

Closes #7479 

## Added tests?

- [ ] yes
- [ ] no, because I need help
- [x] I did clean up some `stories_controller.rb` specs, however

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed
- [x] inline documentation pointing to the docs

DJ Roomba

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -371,11 +371,7 @@ class StoriesController < ApplicationController
         "@id": URL.article(@article)
       },
       "url": URL.article(@article),
-      "image": [
-        ApplicationController.helpers.article_social_image_url(@article, width: 1080, height: 1080),
-        ApplicationController.helpers.article_social_image_url(@article, width: 1280, height: 720),
-        ApplicationController.helpers.article_social_image_url(@article, width: 1600, height: 900),
-      ],
+      "image": seo_optimized_images,
       "publisher": {
         "@context": "http://schema.org",
         "@type": "Organization",
@@ -398,6 +394,15 @@ class StoriesController < ApplicationController
       "datePublished": @article.published_timestamp,
       "dateModified": @article.edited_at&.iso8601 || @article.published_timestamp
     }
+  end
+
+  def seo_optimized_images
+    # This array of images exists for SEO optimization purposes.
+    # For more info on this structure, please refer to this documentation:
+    # https://developers.google.com/search/docs/data-types/article
+    [ApplicationController.helpers.article_social_image_url(@article, width: 1080, height: 1080),
+     ApplicationController.helpers.article_social_image_url(@article, width: 1280, height: 720),
+     ApplicationController.helpers.article_social_image_url(@article, width: 1600, height: 900)]
   end
 
   def set_organization_json_ld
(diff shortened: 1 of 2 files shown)


app/controllers/stories_controller.rb
```diff
@@ -400,6 +396,15 @@ def set_article_json_ld
     }
   end
 
+  def seo_optimized_images
+    # This array of images exists for SEO optimization purposes.
+    # For more info on this structure, please refer to this documentation:
+    # https://developers.google.com/search/docs/data-types/article
+    [ApplicationController.helpers.article_social_image_url(@article, width: 1080, height: 1080),
```
vaidehijoshi, line 403: Is it possible that we'll ever want to build this array for other articles, too? Right now, we're just always using `@article` here, but if we wanted to make this more reusable, we could allow someone to pass in an article, and then use that.

Not sure if we are going to want to build this kind of SEO-optimized array of images anytime soon, but if we do, this would definitely make that easier.

Something else to consider (not blocking): if we _do_ expect that we will want to build this array of variously-sized images in other places, maybe we don't want it to live in this controller, but rather, in a more reusable/shared location. Just a question in the back of my mind after looking at this code :)

juliannatetreault: These are good questions that I don't have great answers to. 😅 I could potentially see us using this array for other articles in the future, but I don't know to what extent and I don't know if refactoring this code and pulling it out into another controller is out of scope for this currently.

Does anyone else have any other thoughts or opinions on this? 😅

maestromac: It's a good idea to extract for reusability but I think we can do it when we're applying SEO optimization else too.

juliannatetreault: I just opened an issue (https://github.com/thepracticaldev/dev.to/issues/7495) and added it to my project board that addresses possible generalization of this image array in addition to refactoring the current JSON-LD code (probably during cool-down). 😊

Ridhwana: I agree with postponing extracting it until we have our next use case.

vaidehijoshi: This all sounds good to me; again, I definitely don't think any of this should block this PR!

PR #7502, opened 2020-04-24
Refactor to create a helper for the mailer links 

Ridhwana:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
We use mailer links in multiple place and so it makes sense to pull it out into a helper and just pass through the necessary parameters.

We also add a helper to set the email_from attribute for a mail.

This is based of the PR [#7438](https://github.com/thepracticaldev/dev.to/pull/7438)

There are no UI changes but here's an example of a mail test (I set the email address yodsahdgafdghsj@dev.to on site config so please don't be alarmed 😁, and I've concealed the from field):

Screenshot 2020-04-27 at 12 23 27

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

alt_text

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -304,15 +304,12 @@ class UsersController < ApplicationController
     community_name = ApplicationConfig["COMMUNITY_NAME"]
     @email_body = <<~HEREDOC
       Hello #{community_name} Team,
-      %0A
-      %0A
+      \n
       I would like to delete my account.
-      %0A%0A
+      \n
       You can keep any comments and discussion posts under the Ghost account.
-      %0A
-      %0A
+      \n
       Regards,
-      %0A
       YOUR-#{community_name}-USERNAME-HERE
     HEREDOC
   end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -303,12 +303,9 @@ class UsersController < ApplicationController
   def handle_account_tab
     community_name = ApplicationConfig["COMMUNITY_NAME"]
     @email_body = <<~HEREDOC
-      Hello #{community_name} Team,
-      \n
-      I would like to delete my account.
-      \n
-      You can keep any comments and discussion posts under the Ghost account.
-      \n
+      Hello #{community_name} Team,\n
+      I would like to delete my account.\n
+      You can keep any comments and discussion posts under the Ghost account.\n
       Regards,
       YOUR-#{community_name}-USERNAME-HERE
     HEREDOC
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -172,20 +172,9 @@ module ApplicationHelper
 
   def email_link(type = :default, text: nil, additional_info: nil)
     # The allowed types for type is :default, :business, :privacy, and members.
-    # The options can be found in field :email_addresses of models/site_config.rb
+    # These options can be found in field :email_addresses of models/site_config.rb
     email = SiteConfig.email_addresses[type] || SiteConfig.email_addresses[:default]
-    additional_info = (additional_info || {}).stringify_keys
-
-    sanitized_add_info = %w[cc bcc body subject reply_to].map! do |item|
-      option = additional_info.delete(item).presence || next
-      "#{item.dasherize}=#{ERB::Util.url_encode(option)}"
-    end.compact
-    sanitized_add_info = sanitized_add_info.empty? ? "".freeze : "?" + sanitized_add_info.join("&")
-
-    encoded_email_address = ERB::Util.url_encode(email).gsub("%40", "@")
-    additional_info["href"] = "mailto:#{encoded_email_address}#{sanitized_add_info}"
-
-    content_tag("a".freeze, text || email, additional_info)
+    mail_to email, text || email, additional_info
   end
 
   # Creates an app internal URL
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -170,7 +170,9 @@ module ApplicationHelper
     "#{start_year} - #{current_year}"
   end
 
-  def mail_link(type = :default, text: nil, additional_info: nil)
+  def email_link(type = :default, text: nil, additional_info: nil)
+    # The allowed types for type is :default, :business, :privacy, and members.
+    # The options can be found in field :email_addresses of models/site_config.rb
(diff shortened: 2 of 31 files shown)


app/views/pages/_coc_text.html.erb
```diff
@@ -31,7 +31,7 @@
 </ul>
 <h2>Enforcement</h2>
 <p>Violations of the Code of Conduct may be reported by contacting the team via the
-  <a href="<%= app_url("/report-abuse") %>">abuse report form</a> or by sending an email to <a href="mailto:<%= SiteConfig.email_addresses[:default] %>" rel="noopener noreferrer"><%= SiteConfig.email_addresses[:default] %></a>. All reports will be reviewed and investigated and will result in a response that is deemed necessary and appropriate to the circumstances. Further details of specific enforcement policies may be posted separately.
```
Ridhwana, line 34: After some research and according to some [documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) we only need an rel="noopener noreferrer" when there is a target=_blank on the `a` tag. For this reason I did not cater for rel="noopener noreferrer". If anyone knows any differently please let me know.

citizen428:
@Ridhwana Restarted the Travis run for you, all green now 😄

Ridhwana:
> @Ridhwana Restarted the Travis run for you, all green now 😄

Thanks @citizen428! and yes I think it was definitely the right choice to make! Looks way cleaner.

app/helpers/application_helper.rb
```diff
@@ -170,6 +170,14 @@ def copyright_notice
     "#{start_year} - #{current_year}"
   end
 
+  def mail_link(type = :default, text: nil, additional_info: nil)
+    email = SiteConfig.email_addresses[type] || SiteConfig.email_addresses[:default]
+    href = "mailto:#{email}#{"?#{additional_info}" if additional_info}"
```
rhymes, line 175: what if additional info is a key value param you parse? this way it can be used as:

```ruby
(e)mail_link(:default, text: "abc", additional_info: {subject: "Title"})
```

and inside you do:

```ruby
[5] pry(main)> URI.encode_www_form(subject: "This is a long subject", body: "This is a longer body with a question mark ?")
=> "subject=This+is+a+long+subject&body=This+is+a+longer+body+with+a+question+mark+%3F"
``` 

currently it's not going to escape the chars like space and question mark, which would, for the latter case, result in a double question mark:

```ruby
[8] pry(main)> additional_info = URI.encode_www_form(subject: "This is a long subject", body: "This is a longer body with a question mark ?")
=> "subject=This+is+a+long+subject&body=This+is+a+longer+body+with+a+question+mark+%3F"
[9] pry(main)> content_tag(:a, "text", {href: "mailto:a@bc.com?#{additional_info}"})
=> "<a href=\"mailto:a@bc.com?subject=This+is+a+long+subject&amp;body=This+is+a+longer+body+with+a+question+mark+%3F\">text</a>"
[11] pry(main)> content_tag(:a, "text", {href: "mailto:a@bc.com?subject=Is this it?"})
=> "<a href=\"mailto:a@bc.com?subject=Is this it?\">text</a>"
```

app/helpers/application_helper.rb
```diff
@@ -170,6 +170,14 @@ def copyright_notice
     "#{start_year} - #{current_year}"
   end
 
+  def mail_link(type = :default, text: nil, additional_info: nil)
```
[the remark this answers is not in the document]
Ridhwana: lol coz it totally did not occur to me 😁email is better thanks! and done

app/helpers/application_helper.rb
Ridhwana: @rhymes this is good, and I like the premise of this. I did find a caveat though - URI.encode_www_form will break on spaces because the `+` is used to encode spaces in URLs, not in HTML and not in SMTP.  So for example the apple mail app would render the subject spaces as `This+is+a+long+subject`.  

However I did take inspiration from [Rail's mail_to](https://apidock.com/rails/ActionView/Helpers/UrlHelper/mail_to) and used ERB::Util.url_encode along with some other nice details for code safety that it has  (like deleting sanitizing the additional parameters.

app/controllers/users_controller.rb
```diff
@@ -304,15 +304,12 @@ def handle_account_tab
     community_name = ApplicationConfig["COMMUNITY_NAME"]
     @email_body = <<~HEREDOC
       Hello #{community_name} Team,
-      %0A
-      %0A
+      \n
       I would like to delete my account.
-      %0A%0A
```
Ridhwana, line 310: needed to change this because we pass through an ERB::Util.url_encod encoded body that will work on SMTP

app/helpers/application_helper.rb
rhymes: Nice work! BTW, why can't we simply use mail_to internally instead of reimplementing it?

rhymes: Something like:

```ruby
def email_link(type = :default, text: nil, additional_info: nil)
  # extract email address(es)
  # call mail_to
end
```

Ridhwana: @rhymes I was literally just implementing that now, when I typed out that comment I thought why in the world am I reimplementing mail_to when we can use it 🙈

app/controllers/users_controller.rb
Ridhwana: Internal workings of mail_to

app/helpers/application_helper.rb
Ridhwana: Lol what a round about way to get here but I think we may have finally got it! 🎉

app/helpers/application_helper.rb
```diff
@@ -170,6 +170,13 @@ def copyright_notice
     "#{start_year} - #{current_year}"
   end
 
+  def email_link(type = :default, text: nil, additional_info: nil)
+    # The allowed types for type is :default, :business, :privacy, and members.
+    # These options can be found in field :email_addresses of models/site_config.rb
+    email = SiteConfig.email_addresses[type] || SiteConfig.email_addresses[:default]
+    mail_to email, text || email, additional_info
```
citizen428, line 177: I forgot this existed 🤣 Rails' API is vast.

PR #7505, opened 2020-04-24
Authentication refactoring: add available and enabled providers

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

The main goal of this PR is to start splitting out and refactoring authentication code that's specific to either GitHub or Twitter in their own modules. There are still a few things to do:

- [x] introduce `Authenticator`, extract GitHub and Twitter providers, add tests - https://github.com/thepracticaldev/dev.to/pull/7259
- [x] test OmniauthCallbacksController in depth - https://github.com/thepracticaldev/dev.to/pull/7439
- [x] have a concept of "available providers" and "enabled providers" (this PR)
- [ ] enable UI and functionality based on which providers are enabled (partially done in this PR)
- [ ] detach the provider identities from being used to fetch external data (the github client and twitter client use random identities in the DB to fetch data, this can't work anymore if auth providers are pluggable, they should use one set of keys for all)

**This is step 3**

**One major difference: providers are listed alphabetically, so github will appear before twitter everywhere** (let me know if this is a huge issue and I'll set an order, possibly using the order in the `SiteConfig`)

Contrary to what I said in https://github.com/thepracticaldev/dev.to/pull/7259 and https://github.com/thepracticaldev/dev.to/pull/7439 I think after this we can start implementing new providers.

The code won't be perfect but we're not going to increase technical debt adding a new provider as we can refactor all those "if twitter_username do this, if github_username do that" while adding a new provider.

I also decided to postpone a bit introducing the feature flag framework as it's not strictly necessary to start having users login with a provider, we can live with using the clunky `SiteConfig` with the array a bit longer.

I'm not cutting corners, just re-prioritizing.

Regarding the last step (which is quite big):

> detach the provider identities from being used to fetch external data (the github client and twitter client use random identities in the DB to fetch data, this can't work anymore if auth providers are pluggable, they should use one set of keys for all)

Ideally I would like to have this implemented before but I thought about it a bit and think it can also be done afterwards:

 - the twitter liquid tag defaults to the global twitter auth tokens if there are no users tokens
 - the github liquid tag and github repo functionality can be disabled if github is disabled globally (will do it in another PR, maybe in parallel to adding a new provider)

If I did my "calculations" and code explorations correctly after this I can work on a new provider.

## Screenshots

**sidebar**

Screenshot_2020-04-24 DEV(local) Community

Screenshot_2020-04-24 DEV(local) Community(1)

**sign up page**

Screenshot_2020-04-24 Screenshot

Screenshot_2020-04-24 Screenshot(1)

**drop down**

Screenshot 2020-04-24 at 7 05 47 PM

Screenshot 2020-04-24 at 7 06 17 PM

**settings** (here I cheated a bit and removed the condition to show them both, too lazy to do 2 screenshots)

Screenshot_2020-04-24 Settings - DEV(local) Community 👩‍💻👨‍💻

## Added tests?

- [x] yes
- [ ] no, because I need help


spec/views/users/edit.html.erb_spec.rb
```diff
@@ -1,30 +0,0 @@
-require "rails_helper"
```
rhymes, line 1: already tested by `spec/requests/user/user_settings_spec.rb`

citizen428:
> I also decided to postpone a bit introducing the feature flag framework as it's not strictly necessary to start having users login with a provider, we can live with using the clunky SiteConfig with the array a bit longer.

💯 I'd rather defer this decision and come up with a proper solution than introduce the wrong thing now.

> the github liquid tag and github repo functionality can be disabled if github is disabled globally 

I also just remembered that we use a random user's token for these tags (`app/liquid_tags/github_tag/github_readme_tag.rb:88`):

```ruby
Identity.where(provider: "github").last(250).sample.token
```

Maybe it's time we introduce a site-wide token owned by us?

app/services/authentication/providers.rb
```diff
@@ -1,11 +1,69 @@
...
-    # TODO: [thepracticaldev/oss] raise exception if provider is available but not enabled for this app
-    # TODO: [thepracticaldev/oss] add available providers, enabled providers
+    # Retrieves a provider that is both available and enabled
     def self.get!(provider_name)
-      "Authentication::Providers::#{provider_name.titleize}".constantize
-    rescue NameError => e
-      raise ::Authentication::Errors::ProviderNotFound, e
+      name = provider_name.to_s.titleize
+
+      "Authentication::Providers::#{name}".safe_constantize.tap do |klass|
```
citizen428, line 7: `safe_constantize` potentially raises and handles the error internally, which is relatively expensive for something that can be done without an exception:

```ruby
name = provider_name.to_s.titleize
unless Authentication::Providers.const_defined?(name)
  ...
end

unless enabled?(provider_name)
  ...
end

Authentication::Providers.const_get(name)
```

This way we can also eliminate the use of `tap`, which I'm not a big fan of in situations like this.

app/services/authentication/providers.rb
```diff
@@ -1,11 +1,69 @@
...
+            ::Authentication::Errors::ProviderNotEnabled,
+            "Provider #{name} is not enabled!",
+          )
+        end
+      end
+    end
+
+    # Returns available providers
+    def self.available
+      # as providers are lazily loaded, we need to load them all
```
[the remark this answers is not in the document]
rhymes: Thank you! I'll make the change!

app/services/authentication/providers.rb
rhymes: Perfect, i'll change this too. Thanks a lot!

rhymes:
> Maybe it's time we introduce a site-wide token owned by us?

We do have it, same for Twitter, the reason why we're not using them I think is to skirt rate limits, but I believe there's a strong case for not doing that in the first place. I'll "detach" them shortly, correlated with an analysis of rate limits for the endpoints we need (for both Twitter and GitHub) and then we can see if we need to increase rate limits or make them smarter.

We shouldn't tie authentication tokens to these features.

app/services/authentication/providers.rb
```diff
@@ -1,11 +1,70 @@
...
+    # Retrieves a provider that is both available and enabled
     def self.get!(provider_name)
-      "Authentication::Providers::#{provider_name.titleize}".constantize
-    rescue NameError => e
-      raise ::Authentication::Errors::ProviderNotFound, e
+      name = provider_name.to_s.titleize
+
+      unless Authentication::Providers.const_defined?(name)
+        raise(
+          ::Authentication::Errors::ProviderNotFound,
```
mstruve, line 16: Is the view going to handle this or will we end up seeing this in Honeybadger? 

It looks like we would see it in Honeybadger which I think could get noisy if someone comes in and tries to throw junk into the form.

app/services/authentication/providers.rb
```diff
@@ -1,11 +1,70 @@
...
+
+    # Returns available providers
+    def self.available
+      # the magic is done in <config/initializers/authentication_providers.rb>
+      Authentication::Providers::Provider.subclasses.map do |subclass|
+        subclass.name.demodulize.downcase.to_sym
+      end.sort
+    end
+
+    # Returns enabled providers
```
mstruve, line 39: Personally, I think all of these methods are named well enough that we don't need the additional comments.

app/services/authentication/providers/github.rb
```diff
@@ -2,6 +2,7 @@ module Authentication
   module Providers
     # GitHub authentication provider, uses omniauth-github as backend
     class Github < Provider
+      OFFICIAL_NAME = "GitHub".freeze
```
mstruve, line 5: 🔥

app/views/layouts/_nav_menu.html.erb
```diff
@@ -3,27 +3,29 @@
     <div id="user-profile-link-placeholder" class="border-0 border-b-1 border-solid border-base-20 p-1">...</div>
 
     <div class="border-0 border-b-1 border-solid border-base-20 p-1">
-      <a href="/dashboard" class="crayons-link crayons-link--block">Dashboard</a>
-      <a href="/new" class="crayons-link crayons-link--block">Write a Post</a>
-      <a href="/readinglist" class="crayons-link crayons-link--block">Reading List</a>
-      <a href="/settings" class="crayons-link crayons-link--block">Settings</a>
-      <a href="/p/information" class="crayons-link crayons-link--block" id="second-last-nav-link">Key Links</a>
+      <a href="<%= dashboard_path %>" class="crayons-link crayons-link--block">Dashboard</a>
```
mstruve, line 6: 🙏

app/services/authentication/providers.rb
```diff
@@ -1,11 +1,70 @@
+# We require all authentication modules to make sure providers
+# are correctly preloaded both in development and in production and
+# ready to be used when needed at runtime
```
vaidehijoshi, line 3: > to make sure providers are correctly preloaded both in development and in production and ready to be used when needed at runtime 

Shouldn't Rails do this for us? I read up a bit on `require_dependency` and it seems like it should rarely be used. I guess, even with this comment, I am still confused why we explicitly need it here. Maybe I'm just missing something?

app/services/authentication/providers.rb
rhymes: The chances for this happen are really slim, the available providers are loaded on disk, someone should basically hijack the login URL to trigger this I think, but if it happens, it'll end up in DD first:

https://github.com/thepracticaldev/dev.to/blob/2d26318cf96c0f1c5c2e827b74bbfa6d27d292d3/app/services/authentication/authenticator.rb#L69

and then in the log file as we haven't decided what to do with the `OmniauthCallbacksController` yet:

https://github.com/thepracticaldev/dev.to/blob/2d26318cf96c0f1c5c2e827b74bbfa6d27d292d3/app/controllers/omniauth_callbacks_controller.rb#L81-L86

app/services/authentication/providers.rb
rhymes: adding to the TODO as tests are too slow for only this change hahaha

app/views/layouts/_nav_menu.html.erb
```diff
@@ -3,27 +3,29 @@
...
+      <a href="<%= new_path %>" class="crayons-link crayons-link--block">Write a Post</a>
+      <a href="<%= readinglist_path %>" class="crayons-link crayons-link--block">Reading List</a>
+      <a href="<%= user_settings_path %>" class="crayons-link crayons-link--block">Settings</a>
+      <a href="<%= information_path %>" class="crayons-link crayons-link--block" id="second-last-nav-link">Key Links</a>
     </div>
     <div class="p-1">
-      <a href="/signout_confirm" class="crayons-link crayons-link--block" id="last-nav-link">Sign Out</a>
+      <a href="<%= signout_confirm_path %>" class="crayons-link crayons-link--block" id="last-nav-link">Sign Out</a>
```
vaidehijoshi, line 13: Nice work cleaning all these paths up! 🚀

app/views/shared/authentication/_providers_signup_modal.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  <% provider = Authentication::Providers.get!(provider_name) %>
+
+  <a
+    href="<%= Authentication::Providers.sign_in_path(provider_name, state: "signup-modal") %>"
+    class="sign-up-link cta"
+    data-no-instant>
+    <img
+      src="<%= asset_path("#{provider_name}-logo.svg") %>"
+      class="icon-img"
+      alt="<%= provider.official_name %> logo" /> Auth With <%= provider.official_name %>
```
vaidehijoshi, line 11: Should this be `Sign In With` rather than `Auth With` for consistency? It looks like we're using both throughout

app/views/users/_additional_authentication.html.erb
```diff
@@ -1,16 +1,17 @@
...
-  </div>
+<% Authentication::Providers.enabled.each do |provider_name| %>
+  <% unless @user.identities.exists?(provider: provider_name) %>
+    <% provider = Authentication::Providers.get!(provider_name) %>
+    <a
+      href="<%= Authentication::Providers.authentication_path(provider_name) %>"
+      class="crayons-btn crayons-btn--outlined mr-2"
+      data-no-instant>
+      <%= inline_svg_tag(
+            "#{provider_name}.svg",
```
vaidehijoshi, line 9: Do we have any checks to ensure that an svg exists for a given provider?

spec/factories/users.rb
```diff
@@ -23,7 +23,7 @@
     email_digest_periodic        { false }
 
     trait :with_identity do
-      transient { identities { %i[github twitter] } }
+      transient { identities { Authentication::Providers.available } }
```
vaidehijoshi, line 26: NICE

vaidehijoshi: I just love the way this reads 👍

spec/system/authentication/user_logs_in_with_github_spec.rb
```diff
@@ -49,6 +49,10 @@
     end
 
     context "when using invalid credentials" do
+      let(:params) do
+        '{"callback_url"=>"http://localhost:3000/users/auth/github/callback", "state"=>"navbar_basic"}'
```
vaidehijoshi, line 53: Hm, iirc, there is a way in rspec to set a host and then use it to build a path? Not blocking, but we might want to consider that instead of explicitly hardcoding `localhost:3000` into our tests 🤔

vaidehijoshi:
This overall looks _great_. I had one big question, and a few little ones :)

app/services/authentication/providers.rb
rhymes: it doesn't auto load them unfotunately as I need to know which subclasses `Provider` has to load them as plugins. This is the previous version of that method:

https://github.com/thepracticaldev/dev.to/pull/7505/commits/9f0d2c08956630e0e4dbf64840707d0710ec603c#diff-24cf9abb14a9d705c426c66026c693eaR27-R35

and this is the conversation I had with Michael about it https://github.com/thepracticaldev/dev.to/pull/7505#discussion_r415491037

app/views/shared/authentication/_providers_signup_modal.html.erb
rhymes: I didn't change the wording from the current version. I'm open to change them obviously but it's long standing copy on the website...

vaidehijoshi: Ah, okay. I don't think we should block for that, but at some point if we're in the code again maybe we can make them consistent

app/views/users/_additional_authentication.html.erb
rhymes: Not, not at present :-( 

Also, `inline_svg_tag` won't break:

```ruby
[11] pry(main)> inline_svg_tag("foobar.svg")
=> "<svg><!-- SVG file not found: 'foobar.svg' --></svg>"
```

maybe I can add a spec to check that "SVG file not found" isn't in

spec/system/authentication/user_logs_in_with_github_spec.rb
rhymes: I honestly don't know :D

rhymes: Also, `config/test.rb` is setup like this:

```ruby
Rails.application.routes.default_url_options = { host: "test.host" }
```

which apparently is being ignored :-D

app/services/authentication/providers.rb
citizen428: Like I said, in production, it should always work. In development you'll experience what you described, due to eager loading. `require_dependency` is generally the way to resolve this while still being part of Rails' usual constant tracking for reloading etc.

spec/system/authentication/user_logs_in_with_github_spec.rb
citizen428: So the `default_url_options` are correctly set to `{:host=>"test.host"}` when I inspect them in the spec. 

But this is a system test, which means it run through Capybara so you're potentially having a problem with `Capybara.default_host` or something.

app/views/shared/authentication/_providers_signup_modal.html.erb
rhymes: noted, thanks!

PR #7506, opened 2020-04-24
Move initialize actions panel button to Webpack

Zhao-Andy:
- [x] Refactor

## Description
This moves some code out of the asset pipeline and into Webpack, where it will only be loaded on the actions panel page.

## Added tests?

## Added to documentation?
- [x] no documentation needed

### DIFF
--- a/app/javascript/actionsPanel/actionsPanel.js
+++ b/app/javascript/actionsPanel/actionsPanel.js
@@ -1,0 +1,29 @@
+function toggleModActionsMenu() {
+  document.querySelector('.mod-actions-menu').classList.toggle('showing');
+}
+
+export function initializeModActionsMenu() {
+  // eslint-disable-next-line no-undef
+  const user = userData();
+  const articlePath = `${
+    document.getElementById('article-show-container').dataset.path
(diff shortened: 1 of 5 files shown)


app/javascript/actionsPanel/actionsPanel.js
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export function initializeModActionsMenu() {
+  // eslint-disable-next-line no-undef
+  const user = userData();
+  const articlePath = `${
+    document.getElementById('article-show-container').dataset.path
+  }/actions_panel`;
+  const articleAuthorId = document.getElementById('article-show-container')
+    .dataset.authorId;
```
nickytonline, line 12: You can use destructuring on the dataset.

```suggestion
  const { authorId: articleAuthorId, path } = document.getElementById('article-show-container').dataset;
  const articlePath = `${path}/actions_panel`;
```

Zhao-Andy: Wow TIL that you can destructure and rename the keys 😱

PR #7511, opened 2020-04-26
Moving babel to dev dependency

Rafi993:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR moves babel and its supporting packages from **dependencies** to **devDependencies**

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


rhymes:
@Rafi993 could you explain the reasoning behind this? Sorry but it's not clear from the PR description.

Rafi993:
> @Rafi993 could you explain the reasoning behind this? Sorry but it's not clear from the PR description.

Sorry I did not make the description clear. The reason I moved these babel and supporting libs from dependencies to devDependencies is because these packages are used during development  (and for creating production build) but are not part of user facing code.

rhymes:
Thanks @Rafi993 for the explanation.

Agreed in principle, though I think they were moved there due to issues we were encountering with Herou deploys, see https://github.com/thepracticaldev/dev.to/pull/6799 and https://github.com/thepracticaldev/dev.to/pull/6800

cc @nickytonline which might know if these should stay put where they are still or they can be moved back to `devDependencies`

nickytonline:
I’ve encountered this at my last job as well. The reason there are some devDependencies in dependencies is because when we build on Heroku, Heroku only installs dependencies, not devDependencies.

See [pruning dependencies](https://devcenter.heroku.com/articles/nodejs-support#skip-pruning) in Heroku’s bodega documentation.

nickytonline:
Closing as we won't be moving these dependencies.

PR #7515, opened 2020-04-26
Add expires_in option for ActiveSupport::Cache::RedisStore#increment method

tamcv:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Add expires_in option for ActiveSupport::Cache::RedisStore#increment method
https://github.com/thepracticaldev/dev.to/issues/7503

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

alt_text


config/initializers/active_support/cache/redis_store.rb
```diff
@@ -0,0 +1,22 @@
+module ActiveSupport
+  module Cache
+    class RedisCacheStore
+      def increment(name, amount = 1, options = nil)
```
rhymes, line 4: I would add comments to explain where this comes from and the why for this monkey patch

spec/initializers/active_support/cache/redis_store_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "rails_helper"
+
+RSpec.describe ActiveSupport::Cache::RedisStore do
+  describe "ActiveSupport::Cache::RedisCacheStore" do
```
rhymes, line 4: here you're describing one class but then putting another in the string (btw the innermost describe isn't needed). You should describe `ActiveSupport::Cache::RedisCacheStore`, not `ActiveSupport::Cache::RedisStore` which is a different class

spec/initializers/active_support/cache/redis_store_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    let(:redis_url) { "redis://localhost:6379" }
+    let(:redis_client) { Redis.new(url: redis_url) }
```
rhymes, line 6: This should read the url from the already configured one. Actually you don't need the URL at all, you can just use `Redis.current` but even better, you don't need any of it. You can just use: "let(:cache) { ActiveSupport::Cache.lookup_store(:redis_cache_store) }"

See:

```ruby
[9] pry(main)> ActiveSupport::Cache.lookup_store(:redis_cache_store).redis
=> #<Redis client v4.1.3 for redis://127.0.0.1:6379/0>
```

PR #7518, opened 2020-04-26
Fix reading list pagination

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently the pagination for a reading list is broken in production as instead of loading more items at the bottom it will load the new page and replace the existing one. 

You can see the behavior in this demo (I set the page size to 1 to make it easier to see):

listings_before

There are 3 scenarios where changing the displayed items is correct: new text in the search box, new tag selected or unselected and "clear all tags"

There is a scenario where items need to be appended instead of replaced, which is the behavior of "load more"

(we don't really have JS tests for the reading list)

**after the fix**

listings_after

**search**

listings_search

**pagination of search results**

listings_search_pagination

**pagination of tags results**

listings_tags

### DIFF
--- a/app/javascript/searchableItemList/searchableItemList.js
+++ b/app/javascript/searchableItemList/searchableItemList.js
@@ -30,7 +30,11 @@ export function onSearchBoxType(event) {
   const { selectedTags, statusView } = component.state;
 
   component.setState({ page: 0 });
-  component.search(query, { tags: selectedTags, statusView });
+  component.search(query, {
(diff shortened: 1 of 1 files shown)


mstruve:
Ah, I didnt realize that was a bug since that is how it worked before I rolled us over to Elasticsearch. Thanks for the fix!

PR #7519, opened 2020-04-26
Improve comment posting error handling

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently validation errors or other errors are treated as `HTTP 200 OK`, which makes it harder for the client to discern when the posting was successful and when it wasn't.

**when the comment text is too long**

comment_body_too_long

comment_body_too_long_2

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
+++ b/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
@@ -198,13 +198,6 @@ function handleCommentSubmit(event) {
       if (response.status === 200) {
         response.json().then(function (newComment) {
           var newNode = document.createElement('div');
-          if (newComment.status == "comment already exists") {
-            return false;
-          }
-          if (newComment.status == "errors") {
-            alert("There was a problem submitting this comment.")
-            return false;
-          }
           newNode.innerHTML = buildCommentHTML(newComment);
           var docBody = document.getElementsByTagName('body')[0]
 
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -54,7 +54,7 @@ class CommentsController < ApplicationController
   def create
     if RateLimitChecker.new(current_user).limit_by_action("comment_creation")
       skip_authorization
-      render json: {}, status: :too_many_requests
+      render json: { error: "Too many requests" }, status: :too_many_requests
       return
     end
 
@@ -75,9 +75,10 @@ class CommentsController < ApplicationController
 
       if @comment.invalid?
         @comment.destroy
-        render json: { status: "comment already exists" }
+        render json: { error: "comment already exists" }, status: :conflict
         return
       end
+
       render json: {
         status: "created",
         css: @comment.custom_css,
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/initializers/initializeCommentsPage.js.erb
```diff
@@ -198,13 +198,6 @@ function handleCommentSubmit(event) {
       if (response.status === 200) {
         response.json().then(function (newComment) {
           var newNode = document.createElement('div');
-          if (newComment.status == "comment already exists") {
```
rhymes, line 201: these just go in the else now. See line 198 above and 

https://github.com/thepracticaldev/dev.to/blob/d4c1b44a21eb4e767eb012e075edc5e72a8ca1c6/app/assets/javascripts/initializers/initializeCommentsPage.js.erb#L249-L255

app/controllers/comments_controller.rb
```diff
@@ -98,13 +99,17 @@ def create
           github_username: current_user.github_username
         }
       }
-    elsif (@comment = Comment.where(body_markdown: @comment.body_markdown,
-                                    commentable_id: @comment.commentable.id,
-                                    ancestry: @comment.ancestry)[1])
-      @comment.destroy
-      render json: { status: "comment already exists" }
+    elsif (comment = Comment.where(
```
rhymes, line 102: this assignment previously overrode `@comment` which is the comment created by the incoming params, thus hiding the validation errors we send now send back on line 111 below

app/controllers/comments_controller.rb
```diff
@@ -75,9 +75,10 @@ def create
 
       if @comment.invalid?
         @comment.destroy
-        render json: { status: "comment already exists" }
+        render json: { error: "comment already exists" }, status: :conflict
```
citizen428, line 78: Interesting choice. I usually use 422 in these cases, since 409 [says](https://httpstatuses.com/409) 

> This code is used in situations where the user might be able to resolve the conflict and resubmit the request.

app/controllers/comments_controller.rb
```diff
@@ -54,7 +54,7 @@ def edit
   def create
     if RateLimitChecker.new(current_user).limit_by_action("comment_creation")
       skip_authorization
-      render json: {}, status: :too_many_requests
+      render json: { error: "Too many requests" }, status: :too_many_requests
```
Ridhwana, line 57: This is a small comment here, I notice we seem to start the error below with lowercase, maybe we can do the same for this one 
```suggestion
      render json: { error: "too many requests" }, status: :too_many_requests
```
or capitalize the one below - whatever is the right fit

app/controllers/comments_controller.rb
rhymes: I agree with you! I was conflicted (LOL) in which one to use but I didn't check the specs

app/controllers/comments_controller.rb
rhymes: you're right! I'll change it with the other change suggested by Michael as well

PR #7525, opened 2020-04-27
Fix settings billing section and other profile improvements

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR contains the following:

1. fix for the settings/billing page (currently the buttons don't work)
1. display of server errors in credits/purchase
1. better error notice display for the settings pages
1. small fix to avoid breaking the settings profile page if the username is took long

I tried to write system tests for the billing page but I just can't make them work 🤷‍♂️ - I think it has something to do with the StripeMock library and JS timeouts. 

**truncate username above 30**

Screenshot_2020-04-27 Settings - DEV(local) Community 👩‍💻👨‍💻

**new error notice**
Screenshot_2020-04-27 Settings - DEV(local) Community 👩‍💻👨‍💻(1)

Screenshot_2020-04-27 Settings - DEV(local) Community 👩‍💻👨‍💻(5)

**billing**

(using [Stripe testing cards](https://stripe.com/docs/testing), no leak ;-))

The UI isn't much better than the current one but we can work on that later I think

Screenshot_2020-04-27 Settings - DEV(local) Community 👩‍💻👨‍💻(4)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -19,6 +19,7 @@ class User < ApplicationRecord
   STREAMING_PLATFORMS = %w[twitch].freeze
   THEMES = %w[default night_theme pink_theme minimal_light_theme ten_x_hacker_theme].freeze
   TWITCH_URL_REGEXP = /\A(http(s)?:\/\/)?(www.twitch.tv|twitch.tv)\/.*\z/.freeze
+  USERNAME_MAX_LENGTH = 30
   USERNAME_REGEXP = /\A[a-zA-Z0-9_]+\z/.freeze
   MESSAGES = {
     invalid_config_font: "%<value>s is not a valid font selection",
@@ -125,7 +126,7 @@ class User < ApplicationRecord
   validates :twitch_url, length: { maximum: 100 }, allow_blank: true, format: TWITCH_URL_REGEXP
   validates :twitter_username, uniqueness: { allow_nil: true }, if: :twitter_username_changed?
   validates :username, presence: true, exclusion: { in: ReservedWords.all, message: MESSAGES[:invalid_username] }
-  validates :username, length: { in: 2..30 }, format: USERNAME_REGEXP
+  validates :username, length: { in: 2..USERNAME_MAX_LENGTH }, format: USERNAME_REGEXP
   validates :username, uniqueness: { case_sensitive: false }, if: :username_changed?
   validates :website_url, :employer_url, url: { allow_blank: true, no_local: true }
   validates :website_url, length: { maximum: 100 }, allow_nil: true
--- a/app/views/credits/new.html.erb
+++ b/app/views/credits/new.html.erb
@@ -3,6 +3,7 @@
 <div class="credits-container">
   <div class="credits-banner">
     <h1>Purchase credits for @<%= @purchaser.username %></h1>
+
     <% if @organizations %>
         <% @organizations.each do |org| %>
           <% unless @purchaser.id == org.id %>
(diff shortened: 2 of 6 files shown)


app/views/credits/new.html.erb
```diff
@@ -14,8 +15,17 @@
         <a href="/credits/purchase" data-no-instant>Switch to personal purchase</a>
       <% end %>
   </div>
+
+  <% if flash[:error].present? %>
```
rhymes, line 19: we weren't displaying errors before

app/views/users/edit.html.erb
```diff
@@ -48,7 +48,9 @@
           </span>
         </h1>
       <% else %>
-        <h1 class="fs-xl s:fs-3xl">Settings for <a href="/<%= @user.username %>">@<%= @user.username %></a></h1>
+        <h1 class="fs-xl s:fs-3xl">
+          Settings for <a href="<%= user_url(@user) %>">@<%= @user.username.truncate(User::USERNAME_MAX_LENGTH) %></a>
```
rhymes, line 52: This is to truncate the displaying of the username in case the user manages to send more than 30 chars (which breaks the layout)

app/views/users/_billing.html.erb
```diff
@@ -1,43 +1,43 @@
...
+          <div class="crayons-card crayons-card--secondary crayons-notice crayons-notice--danger">
             <p><%= flash[:error] %></p>
           </div>
         <% end %>
       </article>
-      <div class="stripe-button-wrapper" id="stripe-button-wrapper">
-        <button id="custom-stripe-button" class="crayons-btn" type="button">Add Credit Card</button>
-      </div>
+
+      <button id="add-credit-card-button" class="crayons-btn credit-card-button" type="button">Add Credit Card</button>
```
Ridhwana, line 14: nice catch with the id and event listener

app/models/user.rb
```diff
@@ -19,6 +19,7 @@ class User < ApplicationRecord
   STREAMING_PLATFORMS = %w[twitch].freeze
   THEMES = %w[default night_theme pink_theme minimal_light_theme ten_x_hacker_theme].freeze
   TWITCH_URL_REGEXP = /\A(http(s)?:\/\/)?(www.twitch.tv|twitch.tv)\/.*\z/.freeze
+  USERNAME_MAX_LENGTH = 30
```
mstruve, line 22: Do any users in production have a username length greater than this that we need to fix to ensure they are not rendered invalid?

app/views/users/edit.html.erb
```diff
@@ -74,27 +76,11 @@
...
-      <div id="error_explanation">
-        <h2><%= pluralize(@user.errors.count, "error") %> prohibited your profile from being saved:</h2>
-
-        <ul>
-          <% @user.errors.full_messages.each do |message| %>
-            <li><%= message %></li>
-          <% end %>
-        </ul>
-      </div>
+      <%= render partial: "users/errors", locals: { errors: @user.errors, title_suffix: "prohibited your profile from being saved" } %>
```
mstruve, line 79: prohibited is such a strong word 😅 I feel like prevented should be used here but that's outside the scope of this PR.

app/models/user.rb
mstruve: Confirmed with @rhymes no users in production have a username longer than 30 characters

PR #7536, opened 2020-04-27
Fix 503 error on organisation rename

eyewritecode:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Handles Organization's articles paths update in a background job

Closes https://github.com/thepracticaldev/dev.to/issues/7355

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/workers/organizations/update_organization_articles_paths_worker.rb
+++ b/app/workers/organizations/update_organization_articles_paths_worker.rb
@@ -9,7 +9,9 @@ module Organizations
 
       return unless organization
 
-      organization.articles.find_each { |article| article.update(path: article.path.gsub(old_slug, new_slug)) }
+      organization.articles.find_each do |article|
+        article.update(path: article.path.gsub(old_slug, new_slug))
+      end
     end
   end
 end
--- a/app/workers/organizations/update_organization_articles_paths_worker.rb
+++ b/app/workers/organizations/update_organization_articles_paths_worker.rb
@@ -5,8 +5,11 @@ module Organizations
     sidekiq_options queue: :high_priority
 
     def perform(organization_id, old_slug, new_slug)
-      articles = Organization.find_by(id: organization_id).articles
-      articles.find_each { |article| article.update(path: article.path.gsub(old_slug, new_slug)) }
+      organization = Organization.find_by(id: organization_id)
+
+      return unless organization
+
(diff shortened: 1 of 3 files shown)


Zhao-Andy:
Thanks for the PR @eyewritecode! Looks like some tests are failing though.

eyewritecode:
Ah!! let me look into that @Zhao-Andy

benhalpern:
Whoops, didn't notice the tests

app/workers/organizations/update_organization_articles_paths_worker.rb
```diff
@@ -0,0 +1,15 @@
+module Organizations
+  class UpdateOrganizationArticlesPathsWorker
+    include Sidekiq::Worker
+
+    sidekiq_options queue: :high_priority
```
rhymes, line 5: Not sure if there should be a retry here

cc @mstruve

app/workers/organizations/update_organization_articles_paths_worker.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    def perform(organization_id, old_slug, new_slug)
+      organization = Organization.find_by(id: organization_id)
+
+      return unless organization
+
+      organization.articles.find_each { |article| article.update(path: article.path.gsub(old_slug, new_slug)) }
```
rhymes, line 12: this could be split in multiple lines using do 😊

eyewritecode: Definitely :)
Also, wouldn't this worker be in the `workers/articles` folder instead? @rhymes

rhymes: Mmm I think either is fine to be honest, your call :) The worker is receiveing an organization as its argument but then acting on their articles 🤷‍♂️

app/workers/organizations/update_organization_articles_paths_worker.rb
mstruve: We will retry automatically, without the option means we will do the full 25 retries rather than limiting that number.

PR #7540, opened 2020-04-27
Remove Algolia from Article Model

mstruve:
- [x] Refactor
- [x] Optimization

## Description
Remove Algolia from Article model since we are no longer using it 🙌 

CodeClimate can be ignored, not sure why it has an issue when this is just removing code. 

https://github.com/thepracticaldev/dev.to/projects/6#card-32955097

## Added tests?

alt_text

### DIFF
--- a/lib/data_update_scripts/20200217131245_re_index_existing_articles_with_approved.rb
+++ b/lib/data_update_scripts/20200217131245_re_index_existing_articles_with_approved.rb
@@ -1,7 +1,7 @@
 module DataUpdateScripts
   class ReIndexExistingArticlesWithApproved
     def run
-      Article.published.find_each { |article| Article.trigger_index(article, false) }
+      # Article.published.find_each { |article| Article.trigger_index(article, false) }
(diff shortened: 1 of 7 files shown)


lib/data_update_scripts/20200217131245_re_index_existing_articles_with_approved.rb
```diff
@@ -1,7 +1,7 @@
 module DataUpdateScripts
   class ReIndexExistingArticlesWithApproved
     def run
-      Article.published.find_each { |article| Article.trigger_index(article, false) }
+      # Article.published.find_each { |article| Article.trigger_index(article, false) }
```
rhymes, line 4: as `trigger_index` is deleted, maybe we can remove the body altogether (as uncommenting this line would break anyway)

mstruve: replaced with a brief comment!

PR #7542, opened 2020-04-27
Add SVG optimizer

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Redoing of https://github.com/thepracticaldev/dev.to/pull/7020

SVG files are markup, markup can be optimized and minimized. [svgo](https://www.npmjs.com/package/svgo) is a tool for that. It also has a UI: https://jakearchibald.github.io/svgomg/ if one wants to play with the settings. 

I kept the default ones and I've optimized all the existing SVGs.

I recommend testing this manually by clicking around. I couldn't find anything odd though.

ps. I added `--pretty` to add a semblance of readability, let me know if nobody cares about that and I'll remove it so we can save some more bytes :D 

## Added tests?

- [ ] yes
- [ ] no, because I need help


app/assets/images/arrow-down-fill.svg
```diff
@@ -1,3 +1,3 @@
-<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path d="M12 16L6 10H18L12 16Z"/>
+<svg width="24" height="24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
vaidehijoshi, line 1: I don't think we want to remove `viewBox` since we'll need it if we ever decide to scale the SVG to fit a parent/container's space

vaidehijoshi:
Also, I am really excited for this PR to be shipped!! 😻 Thanks for doing this, @rhymes !!

app/assets/images/arrow-down-fill.svg
rhymes: Great catch! I'm going to bring it back :D

nickytonline: Correct. Same with all the other viewBox attributes removed below.

rhymes:
Ok, I reverted the previous commit with the SVGs, added https://github.com/thepracticaldev/dev.to/pull/7542/commits/4dacbfff3cd8a8198d28759213ea96180f706d17 and then re-optimized the SVGs, the viewboxes are in. 

I think the build will fail on two tests I don't know how to fix as they don't fail locally. The same exact thing happened in https://github.com/thepracticaldev/dev.to/pull/7020 🤷‍♂️ 

I'm tempted to get rid of approvals for the liquid tags as they keep failing each time we change on of the SVGs :D

cc @maestromac

spec/support/fixtures/approvals/liquid_tweet_tag_spec.approved.html
```diff
@@ -18,7 +18,7 @@
         @thepracticaldev
       </div>
       <div class="ltag__twitter-tweet__twitter-logo">
-        <img src="/assets/twitter-95c212266bf9a35c02dd777b6d438dfbc5a6a4c1e82708c3ab617b069928387a.svg" alt="twitter logo" /></div>
+        <img src="/assets/twitter-99c56e7c338b4d5c17d78f658882ddf18b0bbde5b3f42f84e7964689e7e8fb15.svg" alt="twitter logo" /></div>
```
vaidehijoshi, line 21: Seems weird to me that _this_ changed? 🤔

maestromac:
@rhymes since the cooldown is near, I think it's okay for us to temporary `xit` those specs for now and deal with it post-merge.

rhymes:
> @rhymes since the cooldown is near, I think it's okay for us to temporary `xit` those specs for now and deal with it post-merge.

I agree, otherwise this PR will never go through 🗡️ 

Done!

PR #7549, opened 2020-04-27
Flaky Spec Fix: Clear Elasticsearch data for user viewing timeline specs

mstruve:
- [x] Bug Fix

## Description
These tests are some of our more flaky tests and it seems to be because they are showing leftover Elasticsearch data from other specs. I "proved" this by [creating a branch that also checked the Postgres db counts](https://github.com/thepracticaldev/dev.to/compare/mstruve/flaky-test-info) before checking the page. If the db counts were as expected then that means what we are loading on the page is not in Postgres but likely in Elasticsearch. [Here is a green spec run](https://travis-ci.com/github/thepracticaldev/dev.to/builds/161705100) that contains a single failure(retried successfully)
```
expected to find visible css ".single-article-small-pic" 4 times, found 10 matches:
```

In an attempt to prevent this from happening I added a statement to clear the feed content data before each spec. I did not want to completely reset Elasticsearch bc that adds a lot of overhead to tests while this simply adds a single command. If this works well we could attempt to use the helper for other specs as well instead of completely resetting Elasticsearch. 

https://github.com/thepracticaldev/dev.to/issues/4884

## Added tests?
- [x] updated

alt_text

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -69,7 +69,7 @@ RSpec.configure do |config|
   config.include FactoryBot::Syntax::Methods
   config.include OmniauthMacros
   config.include SidekiqTestHelpers
-  config.include ElasticsearchHelpers, elasticsearch: true
+  config.include ElasticsearchHelpers
 
   config.before(:suite) do
     Search::Cluster.recreate_indexes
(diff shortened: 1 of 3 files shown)


rhymes:
I think this also happens because we recycle data from previous tests all the time, many of our tests don't start with a clean slate and hence you have this behavior, as data from ES is removed on destroy, but if we don't destroy the record, the data is still there.

Thanks for this BTW. Hopefully we can fix the underlying issue at some point in the future!

PR #7550, opened 2020-04-27
Redirect Uppercase Routes to Lowercase Routes in the Stories::Controller

### LINKED ISSUE #7229: Redirect Uppercase Routes to Lowercase Routes
There are some routes on DEV that are uppercase rather than lowercase. To better DEV's overall SEO, it is imperative that the uppercase routes be redirected to lowercase routes and that in general, we ensure that routes are case-sensitive. To accomplish this, we must ensure that uppercased `usernames` redirect to lowercased `usernames` (for starters) before building out this functionality for all routes.

To Do:
- [x] Find uppercased routes for user profile pages (usernames)
- [x] Ensure that if a route is uppercase that it redirects to the lowercase version of the route
- [x] Ensure that redirect is a `301`

juliannatetreault:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
There are some routes on DEV that are uppercase rather than lowercase. To better DEV's overall SEO, it is imperative that the uppercase routes be redirected to lowercase routes. This PR ensures that uppercased usernames redirect to lowercase usernames and serves as a proof of concept that redirecting like this is possible.

Closes #7229 

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -16,6 +16,7 @@ class StoriesController < ApplicationController
 
   before_action :authenticate_user!, except: %i[index search show]
   before_action :set_cache_control_headers, only: %i[index search show]
+  before_action :redirect_to_lowercase_username, only: %i[show]
 
   rescue_from ArgumentError, with: :bad_request
 
@@ -333,6 +334,13 @@ class StoriesController < ApplicationController
 
   def assign_classified_listings
     @classified_listings = ClassifiedListing.where(published: true).select(:title, :category, :classified_listing_category_id, :slug, :bumped_at)
+  end
+
+  def redirect_to_lowercase_username
+    username = params[:username]
+    return unless username.match?(/\A[[:upper:]]+\z/)
+
+    redirect_to user_path(username.downcase), status: :moved_permanently
   end
 
   def set_user_json_ld
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -337,7 +337,7 @@ class StoriesController < ApplicationController
   end
 
   def redirect_to_lowercase_username
-    return unless params[:username] && params[:username]&.match?(/[A-Z]/)
+    return unless params[:username] && params[:username]&.match?(/[[:upper:]]/)
 
     redirect_to user_path(params[:username].downcase), status: :moved_permanently
   end
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -16,7 +16,7 @@ class StoriesController < ApplicationController
 
   before_action :authenticate_user!, except: %i[index search show]
   before_action :set_cache_control_headers, only: %i[index search show]
-  before_action :redirect_to_lowercase_username, only: %i[show]
+  before_action :redirect_to_lowercase_username, only: %i[index]
 
   rescue_from ArgumentError, with: :bad_request
 
@@ -337,10 +337,9 @@ class StoriesController < ApplicationController
   end
 
   def redirect_to_lowercase_username
-    username = params[:username]
-    return unless username.match?(/\A[[:upper:]]+\z/)
+    return unless params[:username] && params[:username]&.match?(/[A-Z]/)
 
-    redirect_to user_path(username.downcase), status: :moved_permanently
+    redirect_to "/#{params[:username].downcase}", status: :moved_permanently
   end
 
   def set_user_json_ld
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -339,7 +339,7 @@ class StoriesController < ApplicationController
   def redirect_to_lowercase_username
     return unless params[:username] && params[:username]&.match?(/[A-Z]/)
 
-    redirect_to "/#{params[:username].downcase}", status: :moved_permanently
+    redirect_to user_path(params[:username].downcase), status: :moved_permanently
   end
 
   def set_user_json_ld
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -339,7 +339,7 @@ class StoriesController < ApplicationController
   def redirect_to_lowercase_username
     return unless params[:username] && params[:username]&.match?(/[[:upper:]]/)
 
-    redirect_to user_path(params[:username].downcase), status: :moved_permanently
+    redirect_to "/#{params[:username].downcase}", status: :moved_permanently
   end
 
   def set_user_json_ld
--- a/spec/requests/stories_index_spec.rb
+++ b/spec/requests/stories_index_spec.rb
@@ -336,8 +336,8 @@ RSpec.describe "StoriesIndex", type: :request do
     it "redirects to the lowercase route for usernames", :aggregate_failures do
       get "/#{user.username.capitalize}"
       expect(response).to have_http_status(:moved_permanently)
-      expect(response).to redirect_to(user_path(user.username.downcase))
-      expect(response).not_to redirect_to(user_path(user.username.capitalize))
+      expect(response).to redirect_to("/#{user.username.downcase}")
+      expect(response).not_to redirect_to("/#{user.username.capitalize}")
     end
   end
 end
--- a/spec/requests/stories_index_spec.rb
+++ b/spec/requests/stories_index_spec.rb
@@ -334,10 +334,10 @@ RSpec.describe "StoriesIndex", type: :request do
     let(:user) { create(:user) }
 
     it "redirects to the lowercase route for usernames", :aggregate_failures do
-      get "/#{user.username.capitalize}"
+      get user_path(user.username.capitalize)
       expect(response).to have_http_status(:moved_permanently)
-      expect(response.body).to redirect_to("/#{user.username.downcase}")
-      expect(response.body).not_to redirect_to("/#{user.username.capitalize}")
+      expect(response).to redirect_to(user_path(user.username.downcase))
+      expect(response).not_to redirect_to(user_path(user.username.capitalize))
     end
   end
 end
--- a/spec/requests/stories_index_spec.rb
+++ b/spec/requests/stories_index_spec.rb
@@ -334,7 +334,7 @@ RSpec.describe "StoriesIndex", type: :request do
     let(:user) { create(:user) }
 
     it "redirects to the lowercase route for usernames", :aggregate_failures do
-      get user_path(user.username.capitalize)
+      get "/#{user.username.capitalize}"
       expect(response).to have_http_status(:moved_permanently)
       expect(response).to redirect_to(user_path(user.username.downcase))
       expect(response).not_to redirect_to(user_path(user.username.capitalize))
--- a/spec/requests/stories_index_spec.rb
+++ b/spec/requests/stories_index_spec.rb
@@ -329,4 +329,14 @@ RSpec.describe "StoriesIndex", type: :request do
       end
     end
   end
+
+  describe "GET user_path" do
+    let(:user) { create(:user) }
+
+    it "redirects to the lowercase route for usernames" do
+      get "/#{user.username.capitalize}"
+      expect(response.body).to include("/#{user.username.downcase}")
+      expect(response.body).not_to include("/#{user.username.capitalize}")
+    end
+  end
 end
--- a/spec/requests/stories_index_spec.rb
+++ b/spec/requests/stories_index_spec.rb
@@ -333,10 +333,11 @@ RSpec.describe "StoriesIndex", type: :request do
   describe "GET user_path" do
     let(:user) { create(:user) }
 
-    it "redirects to the lowercase route for usernames" do
+    it "redirects to the lowercase route for usernames", :aggregate_failures do
       get "/#{user.username.capitalize}"
-      expect(response.body).to include("/#{user.username.downcase}")
-      expect(response.body).not_to include("/#{user.username.capitalize}")
+      expect(response).to have_http_status(:moved_permanently)
+      expect(response.body).to redirect_to("/#{user.username.downcase}")
+      expect(response.body).not_to redirect_to("/#{user.username.capitalize}")
     end
   end
 end
--- a/spec/requests/user/user_profile_spec.rb
+++ b/spec/requests/user/user_profile_spec.rb
@@ -159,4 +159,20 @@ RSpec.describe "UserProfiles", type: :request do
       expect(response.body).to redirect_to "/internal/users/#{user.id}"
     end
   end
+
+  describe "redirect to lowercase route" do
+    let(:user) { create(:user) }
+
+    before do
+      get "/#{user.username.capitalize}"
+    end
+
+    it "redirects to the lowercase route for usernames" do
+      expect(response.body).to include("/#{user.username.downcase}")
+    end
+
+    it "does not display the uppercase route for usernames" do
+      expect(response.body).not_to include("/#{user.username.capitalize}")
+    end
+  end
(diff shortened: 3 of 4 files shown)


app/controllers/users_controller.rb
```diff
@@ -4,6 +4,7 @@ class UsersController < ApplicationController
   before_action :set_user, only: %i[update update_twitch_username update_language_settings confirm_destroy request_destroy full_delete remove_association]
   after_action :verify_authorized, except: %i[index signout_confirm add_org_admin remove_org_admin remove_from_org]
   before_action :authenticate_user!, only: %i[onboarding_update onboarding_checkbox_update]
+  before_action :redirect_to_lowercase_route, only: %i[index]
```
vaidehijoshi, line 7: I'm a bit confused here. I don't think the `index` route is the right place to be checking this. The `index` route is being used for follow and sidebar suggestions, right? I also tried to QA this locally and put a `binding.pry` and never actually hit `users#index`.

When I look at the routes, it looks like when we make a GET request to `/:username` we should be going through `stories#index`, and internally, that method calls `handle_user_index`, which is where we're actually finding the user by the passed in `params[:username]`.

vaidehijoshi: Also, for more context, the way that I was QAing this is by going to a local user's page and then changing their user name to something that was _not_ downcased (for example, `http://localhost:3000/CartwrightFrancesco`. Do let me know how you were QAing this locally and if I should be doing something different, but I'm assuming this is the behavior/request that we're trying to catch!

vaidehijoshi: Huh, it looks like we actually _do_ downcase usernames [here](https://github.com/thepracticaldev/dev.to/blob/master/app/controllers/stories_controller.rb#L197) 🤔

maestromac: Yeah, `users#index` here is actually `dev.to/users`, an API endpoint that returns json as supposed to `dev.to/ben` which is in stories.

juliannatetreault: Okay, these are all good points and this should definitely live in the `Stories::Controller`--I should have known that. 🤦‍♀️ I've since moved this logic there.

Huh, that is strange...

And to answer your question about QAing, the way that you've been QAing my work is exactly how I've been QAing it as well.

spec/requests/user/user_profile_spec.rb
```diff
@@ -159,4 +159,20 @@
...
+
+  describe "redirect to lowercase route" do
+    let(:user) { create(:user) }
+
+    before do
+      get "/#{user.username.capitalize}"
+    end
+
+    it "redirects to the lowercase route for usernames" do
+      expect(response.body).to include("/#{user.username.downcase}")
```
maestromac, line 171: What you can do instead is 
```
expect(response.body).to redirect_to some_correct_path
```

but of course, your test should be in `spec/requests/stories_index_spec.rb` instead

app/controllers/users_controller.rb
rhymes: > Yeah, `users#index` here is actually `dev.to/users`, an API endpoint that returns json as supposed to `dev.to/ben` which is in stories.

lol seems like a good candidate for removal @maestromac :D Going to bring it up for team OSS

spec/requests/stories_index_spec.rb
```diff
@@ -317,4 +317,14 @@ def create_live_sponsor(org, tag)
...
     end
   end
+
+  describe "GET user_path" do
+    let(:user) { create(:user) }
+
+    it "redirects to the lowercase route for usernames" do
+      get "/#{user.username.capitalize}"
+      expect(response.body).to include("/#{user.username.downcase}")
+      expect(response.body).not_to include("/#{user.username.capitalize}")
```
juliannatetreault, line 327: Note: I am unable to test the redirect here with `redirect_to` because the redirect happens in a `before_action`, which "halts the request cycle", as shown in the Rails docs: https://guides.rubyonrails.org/action_controller_overview.html#filters 😄

vaidehijoshi:
I noticed an interesting thing with the params while QAing this locally. Let's say I go to `http://localhost:3000/HaMillMarcellUS` locally; here's what happens initially from the server's perspective:
```bash
Processing by StoriesController#index as */*
  Parameters: {"i"=>"i", "username"=>"HaMillMarcellUS"}
{:message=>"GET /async_info/shell_version completed with 200 OK in 107.585ms", :event=>{:http_response_sent=>{:body=>nil, :content_length=>nil, :headers_json=>"{\"X-Frame-Options\":\"SAMEORIGIN\",\"X-XSS-Protection\":\"1; mode=block\",\"X-Content-Type-Options\":\"nosniff\",\"X-Download-Options\":\"noopen\",\"X-Permitted-Cross-Domain-Policies\":\"none\",\"Referrer-Policy\":\"strict-origin-when-cross-origin\",\"Cache-Control\":\"public, no-cache\",\"Surrogate-Control\":\"max-age=86400, stale-if-error=26400\",\"Surrogate-Key\":\"shell-version-endpoint\",\"Content-Type\":\"application/json; charset=utf-8\",\"Vary\":\"Accept-Encoding\",\"Content-Encoding\":\"gzip\"}", :request_id=>"68efc4b3-f78e-44a2-9dfa-6643ceecde2d", :service_name=>nil, :status=>200, :duration_ms=>107.585}}}
```

Right at this point, I hit the `before_action`, as expected. It redirects, and this is what I see next:


```bash
{:message=>"GET /HaMillMarcellUS completed with 301 Moved Permanently in 18170.207000000002ms", :event=>{:http_response_sent=>{:body=>nil, :content_length=>nil, :headers_json=>"{\"X-Frame-Options\":\"SAMEORIGIN\",\"X-XSS-Protection\":\"1; mode=block\",\"X-Content-Type-Options\":\"nosniff\",\"X-Download-Options\":\"noopen\",\"X-Permitted-Cross-Domain-Policies\":\"none\",\"Referrer-Policy\":\"strict-origin-when-cross-origin\",\"Cache-Control\":\"public, no-cache\",\"Surrogate-Control\":\"max-age=86400, stale-if-error=26400\",\"Location\":\"http://localhost:3000/hamillmarcellus\",\"Content-Type\":\"text/html; charset=utf-8\",\"Vary\":\"Accept-Encoding\",\"Content-Encoding\":\"gzip\"}", :request_id=>"2b5f77f5-0d89-4f26-96d4-d5f35ffdccb6", :service_name=>nil, :status=>301, :duration_ms=>18170.207000000002}}}
Processing by Notifications::CountsController#index as */*
Processing by StoriesController#index as */*
  Parameters: {"username"=>"hamillmarcellus"}
Processing by AsyncInfoController#shell_version as JSON
```

Note that the parameters object is actually different! Initially, it is `Parameters: {"i"=>"i", "username"=>"HaMillMarcellUS"}`, and after the redirect, it is `Parameters: {"username"=>"hamillmarcellus"}`. 

It looks like that `i` parameter comes from [here](https://github.com/thepracticaldev/dev.to/blob/b6664456c0e9db5ab03f9968904e64f23a61ecce/app/controllers/application_controller.rb#L68), and now that I see it, I wonder if we should be using a url helper rather than using the string redirect (`redirect_to "/#{params[:username].downcase}"`)? I'm not actually 100% sure if that means we'd go through the ApplicationController, or whether this even matters...but just wanted to bring it up since I noticed it.

I'll let others weigh in here.

juliannatetreault:
> I noticed an interesting thing with the params while QAing this locally. Let's say I go to `http://localhost:3000/HaMillMarcellUS` locally; here's what happens initially from the server's perspective:
> 
> ```shell
> Processing by StoriesController#index as */*
>   Parameters: {"i"=>"i", "username"=>"HaMillMarcellUS"}
> {:message=>"GET /async_info/shell_version completed with 200 OK in 107.585ms", :event=>{:http_response_sent=>{:body=>nil, :content_length=>nil, :headers_json=>"{\"X-Frame-Options\":\"SAMEORIGIN\",\"X-XSS-Protection\":\"1; mode=block\",\"X-Content-Type-Options\":\"nosniff\",\"X-Download-Options\":\"noopen\",\"X-Permitted-Cross-Domain-Policies\":\"none\",\"Referrer-Policy\":\"strict-origin-when-cross-origin\",\"Cache-Control\":\"public, no-cache\",\"Surrogate-Control\":\"max-age=86400, stale-if-error=26400\",\"Surrogate-Key\":\"shell-version-endpoint\",\"Content-Type\":\"application/json; charset=utf-8\",\"Vary\":\"Accept-Encoding\",\"Content-Encoding\":\"gzip\"}", :request_id=>"68efc4b3-f78e-44a2-9dfa-6643ceecde2d", :service_name=>nil, :status=>200, :duration_ms=>107.585}}}
> ```
> 
> Right at this point, I hit the `before_action`, as expected. It redirects, and this is what I see next:
> 
> ```shell
> {:message=>"GET /HaMillMarcellUS completed with 301 Moved Permanently in 18170.207000000002ms", :event=>{:http_response_sent=>{:body=>nil, :content_length=>nil, :headers_json=>"{\"X-Frame-Options\":\"SAMEORIGIN\",\"X-XSS-Protection\":\"1; mode=block\",\"X-Content-Type-Options\":\"nosniff\",\"X-Download-Options\":\"noopen\",\"X-Permitted-Cross-Domain-Policies\":\"none\",\"Referrer-Policy\":\"strict-origin-when-cross-origin\",\"Cache-Control\":\"public, no-cache\",\"Surrogate-Control\":\"max-age=86400, stale-if-error=26400\",\"Location\":\"http://localhost:3000/hamillmarcellus\",\"Content-Type\":\"text/html; charset=utf-8\",\"Vary\":\"Accept-Encoding\",\"Content-Encoding\":\"gzip\"}", :request_id=>"2b5f77f5-0d89-4f26-96d4-d5f35ffdccb6", :service_name=>nil, :status=>301, :duration_ms=>18170.207000000002}}}
> Processing by Notifications::CountsController#index as */*
> Processing by StoriesController#index as */*
>   Parameters: {"username"=>"hamillmarcellus"}
> Processing by AsyncInfoController#shell_version as JSON
> ```
> 
> Note that the parameters object is actually different! Initially, it is `Parameters: {"i"=>"i", "username"=>"HaMillMarcellUS"}`, and after the redirect, it is `Parameters: {"username"=>"hamillmarcellus"}`.
> 
> It looks like that `i` parameter comes from [here](https://github.com/thepracticaldev/dev.to/blob/b6664456c0e9db5ab03f9968904e64f23a61ecce/app/controllers/application_controller.rb#L68), and now that I see it, I wonder if we should be using a url helper rather than using the string redirect (`redirect_to "/#{params[:username].downcase}"`)?
> 
> I'm not actually 100% sure if that means we'd go through the ApplicationController, or whether this even matters...but just wanted to bring it up since I noticed it.
> 
> I'll let others weigh in here on whether any of this is necessary.

That is such an interesting observation - good catch! 👀  I am happy to make use a url helper instead if need be, but I would like to see what others have to say as well.  😄

vaidehijoshi:
@juliannatetreault Ok! Even if the `i` doesn't end up being significant, it might be nice to use the url helper in any case. I'm not sure why we do it differently in different files, but the more I think about it, the more I think it makes sense to use the helper if it is available to us.

Another thought: now that we're catching any `params[:username]` that isn't downcased and handling it, we should be able to remove a bunch of the `.downcase` calls we have throughout this controller I think. Probably doesn't need to be done as part of this PR, but could clean up this file significantly.

spec/requests/stories_index_spec.rb
maestromac: Hmm, I was able to get it working. I also think it's curcial here we are testing the write http status.
```suggestion
      expect(response).to have_http_status(:moved_permanently)
      expect(response).to redirect_to("/#{user.username}")
```

juliannatetreault: Hmm, so am I now... 🤔 Anyway, I think that's a great suggestion and I've made this change, but have also ensured that we're checking that the redirect goes to `/#{user.username.downcase}`. 😄

juliannatetreault:
> @juliannatetreault Ok! Even if the `i` doesn't end up being significant, it might be nice to use the url helper in any case. I'm not sure why we do it differently in different files, but the more I think about it, the more I think it makes sense to use the helper if it is available to us.
> 
> Another thought: now that we're catching any `params[:username]` that isn't downcased and handling it, we should be able to remove a bunch of the `.downcase` calls we have throughout this controller I think. Probably doesn't need to be done as part of this PR, but could clean up this file significantly.

I've made the change to use the url helper rather than the hardcoded URL--thanks for suggesting this! RE the `.downcases` throughout the file: I will open an issue for that and remove those in a separate PR as you suggested. 😄

app/controllers/stories_controller.rb
```diff
@@ -335,6 +336,12 @@ def assign_classified_listings
     @classified_listings = ClassifiedListing.where(published: true).select(:title, :category, :classified_listing_category_id, :slug, :bumped_at)
   end
 
+  def redirect_to_lowercase_username
+    return unless params[:username] && params[:username]&.match?(/[A-Z]/)
```
rhymes, line 340: even though we only allow letters in this set `[a-zA-Z0-9_]`  for the usernames and slugs, this might change at some point in the future, so to be a little bit more future proofing we might want to use the special `/[[:upper:]]/` that will match any upper case in Unicode. For example:

```ruby
[25] pry(main)> "ŸÊ".match?(/[A-Z]/)
=> false
[26] pry(main)> "ŸÊ".match?(/[[:upper:]]/)
=> true
```

Note that `.upcase` and `.downcase` work correctly:

```ruby
[27] pry(main)> "rhŸmÊs".upcase
=> "RHŸMÊS"
[28] pry(main)> "rhŸmÊs".downcase
=> "rhÿmês"
```

you can find more information about this in https://ruby-doc.org/core-2.7.1/Regexp.html#class-Regexp-label-Character+Classes

vaidehijoshi: Wow, TIL about `:upper:` and what it actually does! Surprisingly, it doesn’t show up on Rubular! I think it makes sense to use it and future-proof 👍

juliannatetreault: @rhymes I actually used `/[[:upper:]]/` here initially and then changed it! 😂 That's good to know though! I can make the change back to it using `/[[:upper:]]/` here quickly.

vaidehijoshi:
Good catch, @benhalpern. Somehow, the rest of us missed that. 

@juliannatetreault You should be able to grep for the appropriate route using `rails routes`, give it a name, and then use a url helper based on the name that you give it:
`GET      /:username(.:format)                                                  stories#index`

juliannatetreault:
> We should also test that we’re redirecting to the right url; seems like the tea should have caught that the redirect was going to the wrong path, no?

The last changes that I pushed up updated the specs to test for the correct URL.

benhalpern:
I think the original PR used the same helper in both the code and the test so it didn't really exercise things properly.

The change happens to be a little more explicit and satisfy @vaidehijoshi's concerns IMO.

vaidehijoshi:
@benhalpern @juliannatetreault I think when I reviewed this earlier this morning, the latest commit(s) hadn't been pushed up? 

In any case, what I was trying to get at was that the test should have been picking up on the fact that `user_path` was redirecting to `/users/:username` instead of just `/:username`, which you caught. My concern was just that the test for this redirect should be actually testing that the redirect goes to the right path, and the original test must have not been since none of us caught it until you pointed out the incorrect path!

PR #7551, opened 2020-04-27
Dont Try Removing Reactions Not in Elasticsearch

mstruve:
- [x] Bug Fix

## Description
Often we will never get a chance to execute an indexing job for a Reaction bc it is destroyed before the job can execute. 
```ruby
def perform(object_class, id)
  object = object_class.constantize.find(id)
  object.index_to_elasticsearch_inline
rescue ActiveRecord::RecordNotFound => e
  return if object_class == "Reaction"

  raise e
end
```
When it is destroyed however we will still attempt to remove it from Elasticsearch. This PR checks for its presence in Elasticsearch before it attempts to destroy it. Another solution would be to ignore `Search::Errors::Transport::NotFound` for reactions when removing them like we do when adding them but that worries me that we might run into missing some bugs so I want to try this more accurate approach first. If others think we should jump straight to the easy solution though let me know, this does add a bit of complexity and an ES call to a callback. 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -26,4 +26,8 @@ module Searchable
   def sync_related_elasticsearch_docs
     self.class::DATA_SYNC_CLASS.new(self).call
   end
+
+  def elasticsearch_doc_exists?
+    self.class::SEARCH_CLASS.document_exists?(search_id)
+  end
 end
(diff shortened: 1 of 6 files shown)


mstruve:
> Wouldn't it be possible to perform this check inside RemoveFromElasticsearchIndexWorker ?

Definitely, except we wouldn't even check we would do the same thing we do for indexing and rescue a not_found error. That is why I asked what others thought. I am game to do that for performance reasons!

PR #7556, opened 2020-04-28
Add automated email ownership emails

jacobherrington:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This is task Michael does manually a lot. There are a lot of cases where
we need to validate that an individual has access to the email address
associated with their DEV account. This automates a singificant portion
of that, and hopefully we can use it in a lot of places.

A user that has never verified their email:
image

A user that has verified their email:
image

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


app/controllers/email_authorizations_controller.rb
```diff
@@ -0,0 +1,12 @@
+class EmailAuthorizationsController < ApplicationController
+  def verify
+    user = User.find_by(username: params[:username])
+    raise ActionController::RoutingError, "Not Found" unless user && current_user == user
```
jacobherrington, line 4: Error handling here isn't super elegant, but I think that a 404 is best.

spec/requests/internal/users_spec.rb
```diff
@@ -51,6 +51,24 @@
...
+
+      expect(ActionMailer::Base.deliveries.count).to eq(1)
+      expect(ActionMailer::Base.deliveries.first.subject).to eq("Verify Your #{ApplicationConfig['COMMUNITY_NAME']} Account Ownership")
+      expect(ActionMailer::Base.deliveries.first.text_part.body).to include(verification_link)
+
+      sign_in(user)
+      get verification_link
+      expect(user.email_authorizations.last.verified_at).to be_within(1.minute).of Time.now.utc
+
+      ActionMailer::Base.deliveries.clear
```
jacobherrington, line 68: Not making this a before/after each because this is the only spec messing with mailers.

app/controllers/email_authorizations_controller.rb
Zhao-Andy: Hmm maybe a `before_action :authenticate_user!` and `not_found unless current_user == user`. You could also make an `email_authorization_policy.rb` here if you want to go a step further.

app/views/mailers/verification_mailer/account_ownership_verification_email.html.erb
```diff
@@ -0,0 +1,18 @@
+<% host = "#{ApplicationConfig['APP_PROTOCOL']}#{ApplicationConfig['APP_DOMAIN']}" %>
+<% verify_url = "#{host}/verify_email_ownership?confirmation_token=#{@user.email_authorizations.order('created_at DESC').first.confirmation_token}&username=#{@user.username}" %>
```
Zhao-Andy, line 2: IMO we should get the token from the controller and put it in an instance variable to render here.

app/controllers/email_authorizations_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+
+    email_authorization = user.email_authorizations.order("created_at DESC").first
+    raise ActionController::RoutingError, "Not Found" unless email_authorization.confirmation_token == params[:confirmation_token]
+
+    email_authorization.update(verified_at: Time.now.utc)
```
rhymes, line 9: `Time.current` as `Time.now` objects technically have no timezone, it's good to keep the convention

app/controllers/email_authorizations_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+    redirect_to "/"
```
rhymes, line 10: this can be `root_path`

app/controllers/email_authorizations_controller.rb
```diff
@@ -0,0 +1,12 @@
+class EmailAuthorizationsController < ApplicationController
+  def verify
+    user = User.find_by(username: params[:username])
+    raise ActionController::RoutingError, "Not Found" unless user && current_user == user
+
+    email_authorization = user.email_authorizations.order("created_at DESC").first
```
rhymes, line 6: is there a chace that this will return nothing?

app/controllers/internal/users_controller.rb
```diff
@@ -106,6 +106,15 @@ def send_email
     end
   end
 
+  def verify_email_ownership
+    if VerificationMailer.account_ownership_verification_email(params).deliver
+      flash[:success] = "Email Verification Mailer sent!"
+      redirect_back(fallback_location: "/users")
```
rhymes, line 112: `fallback_location: users_path`, though I have a question, if this is in the internal admin, why is it falling back to https://dev.to/users which is not a real page?

app/models/email_authorization.rb
```diff
@@ -1,12 +1,20 @@
...
-  validates :user_id, uniqueness: { scope: %i[type_of] }
 
   alias_attribute :sent_at, :created_at
+
+  private
+
+  def generate_confirmation_token
+    return if confirmation_token.present?
+
+    self.confirmation_token = SecureRandom.urlsafe_base64.to_s
```
rhymes, line 18: You don't need `.to_s` in this case

app/views/internal/users/_email_tools.html.erb
```diff
@@ -0,0 +1,35 @@
+<div class="row mb-3">
+  <div class="col-12">
+    <h2>Email Tools</h2>
+    <%= form_with url: verify_email_ownership_internal_user_path(@user), local: true do |f| %>
+      Email last verified at: <%= @user.email_authorizations.where.not(verified_at: nil).order("created_at DESC").first&.verified_at || "Never" %>
```
rhymes, line 5: I think we can move up this query to the controller

app/views/internal/users/_email_tools.html.erb
```diff
@@ -0,0 +1,35 @@
...
+    <% end %>
+  </div>
+</div>
+
+<div class="row mb-3">
+  <div class="col-12">
+    <h3>Recent Emails</h3>
+    <div class="list-group-flush">
+      <% @user.email_messages.order("sent_at DESC").limit(50).each do |email| %>
+        <a href="/admin/email_messages/<%= email.id %>" class="list-group-item list-group-item-action">
```
rhymes, line 28: I see it was already like this but, is it normal that it is pointing to Administrate? Is it because we don't have an email view in the internals for now?

app/views/mailers/verification_mailer/account_ownership_verification_email.html.erb
rhymes, line 2: you can use something like:

```erb
<%= app_url(verify_email_authorizations_path(confirmation_token: token, username: @user.username)) %>
```

where token comes from the email action as suggested by @Zhao-Andy

app/views/internal/users/_email_tools.html.erb
Zhao-Andy: AFAIK we don't have an email view for internal yet.

jacobherrington: I believe so, I can investigate this further, but personally I think it should be left for this PR.

app/controllers/email_authorizations_controller.rb
jacobherrington: It shouldn't because the controller action that sends the mailer also creates the email_authorization. An edge case probably exists though.

jacobherrington:
@Zhao-Andy @rhymes Made the suggested changes, would appreciate another look! Thanks!

Issue #7383, opened 2020-04-20
Running a single system test is very slow (takes over 40 seconds)


ledestin:
**Is your feature request related to a problem? Please describe.**

I worked on a PR and wrote some system tests. Running a single system test takes 40-50 seconds.

I'm willing to investigate why it's so slow, and potentially fixing it.

**Describe the solution you'd like**

A system test taking ~5 seconds would be great.

**Describe alternatives you've considered**

CI is great, but immediate feedback is often needed.

**Additional context**

```
$time ./bin/rspec spec/system/articles/user_creates_an_article_spec.rb
DEPRECATION WARNING: Default values will be removed in the next minor-release of ENVied (i.e. > v0.9). For more info see https://gitlab.com/envied/envied/tree/0-9-releases#defaults. (called from block in load at /home/ledestin/work/dev.to/Envfile:4)
[Zonebie] Setting timezone: ZONEBIE_TZ="Chatham Is."
..

Finished in 1 minute 0.74 seconds (files took 8.82 seconds to load)
2 examples, 0 failures

Coverage report generated for RSpec to /home/ledestin/work/dev.to/coverage. 2723 / 15816 LOC (17.22%) covered.

real    1m12.136s
user    1m8.128s
sys     0m7.584s
```

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

rhymes:
Hi @ledestin, they are slow indeed!

Booting Chrome headless and clicking around with Capybara is not a fast operation.

Do you have any ideas on how to improve that?

ledestin:
Hi @rhymes.

I think it's not Chrome and Capybara that are slow. I have a side project that uses headless Chrome, and running a comparable single test takes ~5 seconds (no Spring either).

It needs to be investigated, where the time is spent. One thing it does is call Webpack to compile JS, but I don't know how expensive that is. It really needs to  be profiled to get the idea where the time is spent.

ledestin:
[...]

PR #7558, opened 2020-04-28
Don't reindex Elasticsearch before suite

citizen428:

- [X] Refactor
- [X] Optimization

## Description

I looked into our Elasticsearch spec setup during an effort to make our specs run faster. 
As @ledestin mentioned that Elasticsearch I/O slows down the spec for him, I got curious, since he shouldn't need ES for that specific spec.

Looking `spec/rails_helper.rb` I noticed that we perform this operation twice, once before the whole test suite, once before specs tagged with `elasticsearch`:

```ruby
  config.before(:suite) do
    Search::Cluster.recreate_indexes
  end

  config.around(:each, elasticsearch: true) do |example|
    Search::Cluster.recreate_indexes
    example.run
  end
```

Looking at the implementation of  `Search::Cluster.delete_indexes` it seems to be able to handle non-existant indexes:

```ruby
next unless Search::Client.indices.exists(index: search_class::INDEX_NAME)
```

This makes me wonder if we need the `berfore_suite` block at all since any index we create there would be removed and recreated as soon as we hit the first spec tagged with `:elasticsearch`.

I experimented with removing `before(:suite)` locally, which shaved off a few more seconds so the particular spec we investigated now finished in ~16 seconds for me. 

I created this draft PR to see how the whole suite deals with the `before(:suit)` index creation removed.

#7383

## Added tests?

## Added to documentation?

- [X] no documentation needed

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -71,6 +71,10 @@ RSpec.configure do |config|
   config.include SidekiqTestHelpers
   config.include ElasticsearchHelpers
 
+  config.before(:suite) do
+    Search::Cluster.delete_indexes
+  end
+
   config.before do
     Sidekiq::Worker.clear_all # worker jobs shouldn't linger around between tests
   end
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -71,10 +71,6 @@ RSpec.configure do |config|
   config.include SidekiqTestHelpers
   config.include ElasticsearchHelpers
 
-  config.before(:suite) do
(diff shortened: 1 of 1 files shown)


citizen428:
As expected, this fails a lot of specs (83), though it seems that many fail  with the same error. I'll see if deleting but not recreating them in `before(:suite)` fairs any better, but the relatively low impact of the speedup does not justify much more time investment than that I'm afraid.

citizen428:
Closing this, but #7602 has other interesting Elasticsearch changes related to specs.

PR #7564, opened 2020-04-28
Configure the community_member_label (rails console update required)

Ridhwana:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Our app/views/mailers/digest_email.html.erb and some other files contains dev-specific texts like "Hiring `developers`?" or "our community of `developers`". This PR relates to [Issue 7280](https://github.com/thepracticaldev/dev.to/issues/7280) and it focuses on the first step of providing a configurable community_member_label. The next part will focus on configuring the digest email further.

https://github.com/thepracticaldev/dev.to/issues/7280

Screenshot 2020-04-28 at 15 58 44

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Yes, 
1. we need to set `SiteConfig.community_member_label = "developer"` in the console.
2. alert new-communities 

alt_text


app/views/dashboards/pro.html.erb
```diff
@@ -18,8 +18,8 @@
 
   <section class="header-card card">
     <h1 class="pro-header">Pro Dashboard for <%= @user_or_org.name %></h1>
-    <p>Welcome to the Pro Dashboard, the home of in-depth user metrics so that authors can make data-driven decisions about the developer ecosystem.</p>
-    <p>This dashboard highlights deep insights especially relevant to developer relations, authors, and serious bloggers.</p>
+    <p>Welcome to the Pro Dashboard, the home of in-depth user metrics so that authors can make data-driven decisions about the  <%= SiteConfig.community_member_label %> ecosystem.</p>
+    <p>This dashboard highlights deep insights especially relevant to  <%= SiteConfig.community_member_label %> relations, authors, and serious bloggers.</p>
```
Ridhwana, line 22: I was very two minded about whether to also add a helper called `community_member_label` like I did with `community_members_label`

Ridhwana: It seems a bit weird to be referencing SiteConfig.community_member_label for the singular version and community_members_label for the plural. Any thoughts?

vaidehijoshi: You could call `pluralize` directly in-line instead of wrapping it in a method, if that makes it feel less weird?

app/views/mailers/digest_mailer/digest_email.html.erb
```diff
@@ -15,7 +15,7 @@
 <div>
   👋 <b>Looking for work?</b> <a href="<%= app_url("/listings/jobs") %>">Browse job openings</a> or create an <a href="<%= app_url("/listings/forhire") %>">available for hire</a> listing.
   <br />
-  🔍 <b>Hiring developers?</b> Post your position(s) on <a href="<%= app_url("/page/post-a-job") %>"><%= community_name %> Listings</a>.
+  🔍 <b>Hiring  <%= community_members_label %>?</b> Post your position(s) on <a href="<%= app_url("/page/post-a-job") %>"><%= community_name %> Listings</a>.
```
vaidehijoshi, line 18: I‘m reviewing this on mobile so it’s hard to tell but it looks like there’s an extra space before the erb here.

app/views/dashboards/pro.html.erb
Ridhwana: Yeah that's also an option. I thought it was a bit repetitive to keep calling pluralize coz the most common use case is to use it in its plural form

app/views/mailers/digest_mailer/digest_email.html.erb
```diff
@@ -30,7 +30,7 @@
       You can change it any time in the future. Happy coding!
     <% elsif @user.follows.size == 0 %>
       <b>
-        <%= community_name %> Digest is a new periodic email featuring the best posts from our community of developers.
+        <%= community_name %> Digest is a new periodic email featuring the best posts from our community of  <%= community_members_label %>.
```
vaidehijoshi, line 33: Same comment re: extra space?

app/views/dashboards/pro.html.erb
vaidehijoshi: Yeah I tend to agree with you. I think it’s okay to wrap in a method. You could also wrap the singular one in a method too, if that feels better? 🤷🏽‍♀️ but what you already have doesn’t feel bad to me :)

app/helpers/application_helper.rb
```diff
@@ -177,6 +177,10 @@ def email_link(type = :default, text: nil, additional_info: nil)
     mail_to email, text || email, additional_info
   end
 
+  def community_members_label
+    SiteConfig.community_member_label.pluralize
```
rhymes, line 181: although it's fine for now, the assumption that pluralization will always work for other communities member's designations might not always be true. Maybe the next cycle we can start thinking of using i18n or something :D

app/views/mailers/digest_mailer/digest_email.html.erb
Ridhwana: Removed the extra space thanks!

Ridhwana:
> I think this looks good (I didn’t QA it but the code functionally seems okay!). I noticed in the post-deployment tasks, you mentioned that we want to set this variable for our own community...but I think that variable needs to be set before deploying (rather than after), no?

Thanks @vaidehijoshi 😊 So, that field `community_member_label` won't exist on the model as yet if we try and set it before deployment, hence we'll probably get an `NoMethodError: undefined method `community_member_label=' for #class`

app/views/dashboards/pro.html.erb
Ridhwana: Thanks for this insight, it helps to hear a different perspective. I think since it feels okay, I'm going to leave it as is for now .

rhymes: hahaha I just added a note about this before reading this conversation

app/views/mailers/digest_mailer/digest_email.html.erb
rhymes: yep, two spaces

app/helpers/application_helper.rb
benhalpern: Good call

benhalpern:
@Ridhwana this is a great touch! I believe Taylor Swift fans are called "Swifties", and Harry Potter fans are "Potterheads". So I could see this having a lot of legs in a lot of communities 😄

PR #7567, opened 2020-04-28
Ignore Search::Errors::Transport::NotFound error when Removing Reactions

mstruve:
- [x] Bug Fix

## Description
Often we will never get a chance to execute an indexing job for a Reaction bc it is destroyed before the job can execute. When it is destroyed however we will still attempt to remove it from Elasticsearch. This PR will ignore `Search::Errors::Transport::NotFound` errors for reactions when removing them like we do when adding them.

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/workers/search/remove_from_elasticsearch_index_worker.rb
+++ b/app/workers/search/remove_from_elasticsearch_index_worker.rb
@@ -7,6 +7,8 @@ module Search
     def perform(search_class, id)
       search_class.safe_constantize.delete_document(id)
     rescue Search::Errors::Transport::NotFound => e
+      # Reactions are often destroyed before the indexing job can execute
+      # so we ignore this error when trying to remove them
       return if search_class == "Search::Reaction"
 
       raise e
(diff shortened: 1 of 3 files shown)


app/workers/search/remove_from_elasticsearch_index_worker.rb
```diff
@@ -6,6 +6,10 @@ class RemoveFromElasticsearchIndexWorker
 
     def perform(search_class, id)
       search_class.safe_constantize.delete_document(id)
+    rescue Search::Errors::Transport::NotFound => e
```
atsmith813, line 9: Can we add a comment explaining why we have to do this? Maybe something like:
```ruby
# Reactions are often destroyed before the indexing job can execute so we ignore this error
```
It might help our future selves remember why we did this. Or if we end up having to rescue for other document types, we'll know why for each one.

mstruve: added to both workers actually.

Issue #7254, opened 2020-04-13
Add a twitter hashtag as a configurable option


lightalloy:
- create a `SiteConfig` option for twitter hastag, e.g. `SiteConfig.twitter_hashtag`
- make it configurable via `/internal/config`
- replace hardcoded twitter hashtag references with a configurable option (not complete)

lightalloy:
To our amazing contributors: This issue is labeled as ready for dev, so it's up for grabs.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

JuanVqz:
@lightalloy May I take it? Please

lightalloy:
@JuanVqz feel free to take it :+1:

JuanVqz:
Sorry for the time I took for making changes on this issue but I wasn't able to run the dev.to project on Docker containers, actually I'm going to open an issue for that

Ridhwana:
To tackle `replace hardcoded twitter hashtag references with a configurable option` in [a separate issue](https://github.com/thepracticaldev/dev.to/issues/7835)

lightalloy:
Closing as done with #7573

PR #7573, opened 2020-04-28
Add twitter hashtag to the site config

JuanVqz:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Issue #7254

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

**We need to set `SiteConfig.twitter_hastag = "DEVCommunity"` in the console.**


JuanVqz:
Should I add some twitter hashtag by default?
Should I change the specific hashtag to this hashtag config on all the pages I may see?

lightalloy:
Hi, @JuanVqz . Thanks for the pr!
> Should I add some twitter hashtag by default?
I think there is no need to add the default hashtag, but you'll need to add a note to the pr for a person who will merge it to update the current value on DEV:
`SiteConfig.twitter_hastag = "DEVCommunity"` (similar to the note in this pr => https://github.com/thepracticaldev/dev.to/pull/7564)

>  Should I change the specific hashtag to this hashtag config on all the pages I may see?
You're right, the hardcoded `DEVCommunity` should be replaced with the configurable one. You will need to update the related tests as well.

Another thing to add is the field in the `app/views/internal/configs/show.html.erb` so that the admins would be able to set up the hashtag.

Ridhwana:
Hi @JuanVqz 😊 This PR is coming along nicely - if you'd like we can take this PR through review and live and then replace the twitter hashtags with the Site Config variables in a subsequent one, if this would be easier and quicker for you?

JuanVqz:
Please don't forget this.
We need to set SiteConfig.twitter_hastag = "DEVCommunity" in the console.

Ridhwana:
> Please don't forget this.
> We need to set SiteConfig.twitter_hastag = "DEVCommunity" in the console.

Thanks @JuanVqz 😊 I've added this to the PR description under post deployment tasks and I've added a label too. I'll add another issue tomorrow to replace the hashtags with the Site Config.

Ridhwana:
@Zhao-Andy noticed that you on pager duty so dropping you a note on this PR: 

I'm merging this but not deploying it as yet as its still early for the US team :). 
After our next deploy (which will take this PR along) we'd need to set the record from the post deployment task. It won't break anything if it's not set immediately because we not using this config variable anywhere as yet 😊

Zhao-Andy:
Updated, thanks for the ping!

PR #7575, opened 2020-04-28
Fix 'Suggested people' small viewport issue

maestromac:
- [x] Bug Fix

Resolve https://github.com/thepracticaldev/dev.to/issues/6884
Before: 

After:
imageimage

## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a


vaidehijoshi:
@maestromac what's the best way to QA this (is there a specific phone size that causes the viewport issue?)

PR #7583, opened 2020-04-28
Support docker-selenium

maestromac:
- [x] Feature

## Description
I am using WSL2 and it does not support chromedriver. The current workaround is to rely on docker. It does work pretty well but will require this change for additional configuration.

This change should be harmless most of our user. 
The source is from https://www.neotericdesign.com/articles/2018/04/running-your-rails-test-suite-with-dockerized-selenium-on-gitlab-ci/
n/a
## Added tests?
- [x] yea, doc.dev.to

## Added to documentation?
Document to be added in the future once I work out the kinks

n/a


spec/support/initializers/capybara.rb
```diff
@@ -21,7 +21,16 @@
   end
 
   config.before(:each, type: :system, js: true) do
-    driven_by :headless_chrome
+    if ENV["SELENIUM_URL"].present?
+      # Support use of selenium-docker
+      Capybara.server_host = "0.0.0.0"
+      ip = Socket.ip_address_list.detect(&:ipv4_private?).ip_address
+      host! "http://#{ip}:#{Capybara.server_port}"
```
citizen428, line 28: I generally prefer building URI's with appropriate tools, but don't let that stop you form merging:

```suggestion
      URI::HTTP.build(host: ip, port: Capybara.server_port).to_s
```

spec/support/initializers/capybara.rb
```diff
@@ -21,7 +21,16 @@
   end
 
   config.before(:each, type: :system, js: true) do
-    driven_by :headless_chrome
+    if ENV["SELENIUM_URL"].present?
+      # Support use of selenium-docker
```
citizen428, line 25: Technically this would work for other scenarios as well, like running it on another machine. So if we introduce this, maybe we can add another option and also change the next line to this:

```ruby
Capybara.server_host = ENV.fetch("CAPYBARA_SERVER_HOST") { "0.0.0.0" }
```

citizen428:
The problem is that hostnames/IP addresses are no longer shared between Windows and Linux with WSL2. There are several workarounds for that, e.g. forwarding with `socat`.

Anyway, we'd probably want to add a little bit of documentation around this, could save other people some time when they are trying to get this to work on Windows. Maybe a small section on WSL2 [here](https://docs.dev.to/installation/windows/).

spec/support/initializers/capybara.rb
maestromac: that looks much neater than piecing together strings.

spec/support/initializers/capybara.rb
maestromac: Updated!

maestromac:
@citizen428 I hope you don't mind that I used your comment there in the documentation. I added a brief instruction on how to get started with system spec but I can't provide instruction for `socats` as I haven't gave it a chance yet.

PR #7592, opened 2020-04-29
Upgrade field_test to 0.3.2

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Upgrading `field_test` to the latest 0.3.2 version. [dependabot tried](https://github.com/thepracticaldev/dev.to/pull/7475) but fortunately one of the tests broke as there's breaking change in one of the transitive dependencies 👀 , the gem `browser`.

Basically the empty user agent, sent by request specs, is considererd a bot now, thus failing to create the new experiment.

For more details, see [browser: Bots](https://github.com/fnando/browser#bots).

**note:**, this will change the behavior in production as well:

- requests with a missing user agent will be considered a bot
- requests with these keywords - `/crawl|fetch|search|monitoring|spider|bot/` - will be considered a bot.

If we want to keep the existing behavior, we need to undo this change like suggested:

> To restore v2's bot detection, remove the following matchers:
> Browser::Bot.matchers.delete(Browser::Bot::KeywordMatcher)
> Browser::Bot.matchers.delete(Browser::Bot::EmptyUserAgentMatcher)

I also linked user to memberships as suggested in [field_test: Associations](https://github.com/ankane/field_test#associations), so that we can have `user.field_test_memberships`

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -35,6 +35,7 @@ RSpec.describe User, type: :model do
       it { is_expected.to have_many(:display_ad_events).dependent(:destroy) }
       it { is_expected.to have_many(:email_authorizations).dependent(:delete_all) }
       it { is_expected.to have_many(:email_messages).class_name("Ahoy::Message").dependent(:destroy) }
+      it { is_expected.to have_many(:field_test_memberships).class_name("FieldTest::Membership").dependent(:destroy) }
       it { is_expected.to have_many(:github_repos).dependent(:destroy) }
       it { is_expected.to have_many(:html_variants).dependent(:destroy) }
       it { is_expected.to have_many(:identities).dependent(:destroy) }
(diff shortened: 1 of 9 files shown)


spec/rails_helper.rb
```diff
@@ -56,6 +56,12 @@
 
 Rack::Attack.enabled = false
 
+# `browser`, a dependency of `field_test`, starting from version 3.0
+# considers the empty user agent a bot, with will fail tests as we
```
mstruve, line 60: `which will fail tests as we`

spec/models/user_spec.rb
```diff
@@ -139,10 +140,8 @@ def user_from_authorization_service(service_name, signed_in_resource, cta_varian
         expect(fourth_field).not_to be_valid
       end
 
-      it { is_expected.to have_many(:organization_memberships).dependent(:destroy) }
```
mstruve, line 142: hahaha my worker generalization PR #7594  just found these as well 👍

spec/rails_helper.rb
```diff
@@ -56,6 +56,12 @@
 
 Rack::Attack.enabled = false
 
+# `browser`, a dependency of `field_test`, starting from version 3.0
+# considers the empty user agent a bot, which will fail tests as we
+# explicitly configure field tests to exclude bots
+# see https://github.com/fnando/browser/blob/master/CHANGELOG.md#300
+Browser::Bot.matchers.delete(Browser::Bot::EmptyUserAgentMatcher)
```
joshpuetz, line 63: Excellent note, thanks!

PR #7594, opened 2020-04-29
Search Generalization: Remove Elasticsearch from Index Worker Names

mstruve:

- [x] Refactor

## Description
Sorry for the large dif! Now that we have removed the Algolia Index Workers we can steal their names for our Elasticsearch workers. 

YES, I choose to do it inline, I will deal with any fallout or failed jobs that occur from this switch. I don't expect there to be a ton, we don't have a super high indexing volume yet. 

https://github.com/thepracticaldev/dev.to/projects/6#card-32887789

## Added tests?
- [x] updated

alt_text

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -139,10 +139,8 @@ RSpec.describe User, type: :model do
         expect(fourth_field).not_to be_valid
       end
 
-      it { is_expected.to have_many(:organization_memberships).dependent(:destroy) }
       it { is_expected.to have_one(:counters).class_name("UserCounter").dependent(:destroy) }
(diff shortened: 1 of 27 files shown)


spec/models/user_spec.rb
```diff
@@ -139,10 +139,8 @@ def user_from_authorization_service(service_name, signed_in_resource, cta_varian
         expect(fourth_field).not_to be_valid
       end
 
-      it { is_expected.to have_many(:organization_memberships).dependent(:destroy) }
```
mstruve, line 142: Rubocop caught these as duplicate assertions so I removed the dups.

PR #7595, opened 2020-04-29
Rate Limit Organization Creation

mstruve:
- [x] Bug Fix

## Description
Limit the number of organizations a user can make to 1 per 5 minute period. I went with the same approach as article creation and just raised a rate limit reached error. 

There is A LOT of duplicated code here that could use a good refactor. I plan to tackle that in a separate PR. 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -3,6 +3,8 @@ class OrganizationsController < ApplicationController
   rescue_from Errno::ENAMETOOLONG, with: :log_image_data_to_datadog
 
   def create
+    rate_limit!
+
     @tab = "organization"
     @user = current_user
     @tab_list = @user.settings_tab_list
@@ -15,6 +17,7 @@ class OrganizationsController < ApplicationController
     @organization = Organization.new(organization_params)
     authorize @organization
     if @organization.save
+      rate_limiter.track_organization_creation
       @organization_membership = OrganizationMembership.create!(organization_id: @organization.id, user_id: current_user.id, type_of_user: "admin")
       flash[:settings_notice] = "Your organization was successfully created and you are an admin."
       redirect_to "/settings/organization/#{@organization.id}"
@@ -108,5 +111,18 @@ class OrganizationsController < ApplicationController
 
     @organization.errors.add(:profile_image, FILENAME_TOO_LONG_MESSAGE)
     false
+  end
+
+  def rate_limit!
+    rate_limiter.tap do |rate_limiter|
+      if rate_limiter.limit_by_action(:organization_creation)
+        retry_after = RateLimitChecker::RETRY_AFTER[:organization_creation]
+        raise RateLimitChecker::LimitReached, retry_after
+      end
+    end
+  end
+
+  def rate_limiter
+    RateLimitChecker.new(current_user)
   end
 end
--- a/app/labor/rate_limit_checker.rb
+++ b/app/labor/rate_limit_checker.rb
@@ -4,8 +4,17 @@ class RateLimitChecker
   RETRY_AFTER = {
     article_update: 30,
     image_upload: 30,
-    published_article_creation: 30
+    published_article_creation: 30,
+    organization_creation: 300
   }.with_indifferent_access.freeze
+
+  CONFIG_LIMIT_KEYS = %i[
+    rate_limit_follow_count_daily
+    rate_limit_comment_creation
+    rate_limit_published_article_creation
+    rate_limit_image_upload
+    rate_limit_email_recipient
+  ].freeze
 
   def initialize(user = nil)
     @user = user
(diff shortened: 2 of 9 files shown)


app/views/internal/configs/show.html.erb
```diff
@@ -374,6 +374,16 @@
               <div class="alert alert-info">Used to limit the amount of articles a user can create within 30 seconds</div>
             </div>
 
+            <div class="form-group">
+              <%= f.label :rate_limit_organization_creation %>
+              <%= f.number_field :rate_limit_organization_creation,
+                                 class: "form-control",
+                                 value: SiteConfig.rate_limit_organization_creation,
+                                 min: 0,
```
rhymes, line 382: min should be 1 otherwise no org will be created I guess, right?

mstruve: Good call!

app/controllers/organizations_controller.rb
```diff
@@ -3,6 +3,8 @@ class OrganizationsController < ApplicationController
   rescue_from Errno::ENAMETOOLONG, with: :log_image_data_to_datadog
 
   def create
+    rate_limit!
```
vaidehijoshi, line 6: Any reason to not put this in a `before_action`? Seems like we wouldn't even want to continue with this request if the rate limiter decides that we've exceeded the limit?

app/controllers/organizations_controller.rb
```diff
@@ -109,4 +112,17 @@ def valid_filename?
...
+    rate_limiter.tap do |rate_limiter|
+      if rate_limiter.limit_by_action(:organization_creation)
+        retry_after = RateLimitChecker::RETRY_AFTER[:organization_creation]
+        raise RateLimitChecker::LimitReached, retry_after
+      end
+    end
+  end
+
+  def rate_limiter
+    RateLimitChecker.new(current_user)
```
vaidehijoshi, line 126: It's probably okay since we only call this from the context of one method, but if we end up using it elsewhere in the controller we might want to consider memoizing.

app/labor/rate_limit_checker.rb
```diff
@@ -4,9 +4,18 @@ class RateLimitChecker
   RETRY_AFTER = {
     article_update: 30,
     image_upload: 30,
-    published_article_creation: 30
+    published_article_creation: 30,
+    organization_creation: 300
   }.with_indifferent_access.freeze
```
vaidehijoshi, line 9: Can we include a comment to indicate that these are all in seconds? Initially I thought that these were the number of attempts and had to look through the rest of the file to verify that I was wrong and it was actually seconds.

app/views/internal/configs/show.html.erb
```diff
@@ -374,6 +374,16 @@
...
 
+            <div class="form-group">
+              <%= f.label :rate_limit_organization_creation %>
+              <%= f.number_field :rate_limit_organization_creation,
+                                 class: "form-control",
+                                 value: SiteConfig.rate_limit_organization_creation,
+                                 min: 1,
+                                 placeholder: 1 %>
+              <div class="alert alert-info">Number of organizations a user can create within a 5 minute period</div>
+            </div>
```
vaidehijoshi, line 385: Nice thought to make this all configurable from the get-go! 👍

spec/requests/user/user_organization_spec.rb
```diff
@@ -28,14 +28,20 @@
...
-      org_params = build(:organization).attributes
       org_params["profile_image"] = Rack::Test::UploadedFile.new(Rails.root.join("app/assets/images/android-icon-36x36.png"), "image/jpeg")
-      post "/organizations", params: { organization: org_params }
+      allow(RateLimitChecker).to receive(:new).and_return(rate_limiter)
+      allow(rate_limiter).to receive(:limit_by_action).and_return(false)
     end
 
     it "creates the correct organization_membership association" do
+      create_org
+
```
vaidehijoshi, line 44: Extra line? (The other spec you modified doesn't have one 🙃)

app/controllers/organizations_controller.rb
```diff
@@ -109,4 +112,17 @@ def valid_filename?
     @organization.errors.add(:profile_image, FILENAME_TOO_LONG_MESSAGE)
     false
   end
+
+  def rate_limit!
+    rate_limiter.tap do |rate_limiter|
```
citizen428, line 117: I don't think we need `tap` here, as we're ignoring the return value of this method anyway.

app/controllers/organizations_controller.rb
mstruve: It is only being used in this method so I didn't put it in a before action. I think a refactor is due for this were this `rate_limit!` method could possibly be pulled out of individual controllers and put in application controller or the Rate limiter itself.

app/controllers/organizations_controller.rb
mstruve: I was following the convention @rhymes setup, as he stated, this will get refactored so I Im going to leave it for now.

app/controllers/organizations_controller.rb
vaidehijoshi: This still feels pretty before-action-y to me, and I'd still vote for it to be pulled out (we have plenty of before_actions for only one method). But if you plan to refactor it into the application controller, I think that works too, and if that means that we'll overhaul all of the rate limiting stuff into the application controller, I think that is definitely preferable.

PR #7597, opened 2020-04-29
Authentication refactoring: generalize account settings

rhymes:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In this PR we generalize the last (hopefully) remaining part of the user facing code, the settings/account panel.

- renamed `remove_association` to `remove_identity`
- made sure one can only remove enabled providers (we don't show disabled providers anyway, but it's good to check server side as well)
- added `settings_url` which points to the URL where the user can land to remove the oauth association
- future proof some code and tests which checked if the total number of identities was 2 (we'll soon have more)
- use `sign_in_path` everywhere

Closes https://github.com/thepracticaldev/dev.to/issues/7585

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,7 +1,9 @@
 class UsersController < ApplicationController
   before_action :set_no_cache_header
   before_action :raise_suspended, only: %i[update]
-  before_action :set_user, only: %i[update update_twitch_username update_language_settings confirm_destroy request_destroy full_delete remove_association]
+  before_action :set_user, only: %i[
+    update update_twitch_username update_language_settings confirm_destroy request_destroy full_delete remove_identity
+  ]
   after_action :verify_authorized, except: %i[index signout_confirm add_org_admin remove_org_admin remove_from_org]
   before_action :authenticate_user!, only: %i[onboarding_update onboarding_checkbox_update]
   rescue_from Errno::ENAMETOOLONG, with: :log_image_data_to_datadog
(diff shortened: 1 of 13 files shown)


app/controllers/users_controller.rb
```diff
@@ -130,22 +132,34 @@ def full_delete
...
 
-    if @user.identities.size == 2 && identity
+    error_message = "An error occurred. Please try again or send an email to: #{SiteConfig.email_addresses[:default]}"
+    unless Authentication::Providers.enabled?(params[:provider])
+      flash[:error] = error_message
+      redirect_to user_settings_path(@tab)
+      return
+    end
+
+    provider = Authentication::Providers.get!(params[:provider])
```
mstruve, line 145: could use your new helper here `authentication_provider(provider_name)`

app/controllers/users_controller.rb
```diff
@@ -130,22 +132,34 @@ def full_delete
...
+
+    identity = @user.identities.find_by(provider: provider.provider_name)
+
+    if identity && @user.identities.size > 1
       identity.destroy
 
-      identity_username = "#{provider}_username".to_sym
-      @user.update(identity_username => nil, :profile_updated_at => Time.current)
+      @user.update(
+        provider.user_username_field => nil,
```
mstruve, line 153: Should this be `provider.class.user_username_field` like it is [below in the authenticator class](https://github.com/thepracticaldev/dev.to/pull/7597/files#diff-863cabff965798c30566df1bb5e2b3a8R99)?

app/views/users/_account.html.erb
```diff
@@ -45,7 +45,7 @@
 <div class="crayons-box crayons-box--danger p-6 grid gap-6">
   <h2 class="color-accent-danger">Danger Zone</h2>
 
-  <% if @user.identities.size == 2 %>
+  <% if @user.identities.size > 1 %>
```
mstruve, line 48: 🔥

app/controllers/users_controller.rb
citizen428: It's a view helper, so you'd need to call it as`helpers.authentication_provider(params[:provider])`.

app/services/authentication/providers/github.rb
```diff
@@ -25,10 +26,22 @@ def existing_user_data
         }
       end
 
+      def self.user_created_at_field
```
citizen428, line 29: Tbh I liked them better when they were instance methods, as you always seem to be calling them from an instance context as illustrated by the `provider.class.user_created_at_field` calls. Ruby is not quite as strict as Python about making things static when you don't access `self` in the method body. As a sort of compromise you can leave them as is and add 

```ruby
delegate :user_created_at_field, to: :class
```

Which will keep them on the class but make them a bit less awkward to acess from an instance.

app/controllers/users_controller.rb
rhymes: it's a bit scary that it didn't fail any test 👀 
Going to see why

app/services/authentication/providers/github.rb
rhymes: Awesome!

app/controllers/users_controller.rb
rhymes: Oh I know why, that's the class object, not the instance object, that's why it works :D I'm going to apply @citizen428 suggestion here https://github.com/thepracticaldev/dev.to/pull/7597#discussion_r417809095 so we don't get confused ;-)

app/services/authentication/providers/provider.rb
```diff
@@ -2,12 +2,8 @@ module Authentication
   module Providers
     # Authentication provider
     class Provider
-      # name of the field to store the upstream identity's creation date in
-      CREATED_AT_FIELD = "".freeze
-      # name of the field to store the upstream identity's username in
-      USERNAME_FIELD = "".freeze
-
       delegate :email, :nickname, to: :info, prefix: :user
+      delegate :user_created_at_field, :user_username_field, to: :class
```
mstruve, line 6: 🔥

PR #7598, opened 2020-04-29
Add focus-within for tab navigation in FollowUsers

maestromac:
Both of these I guess?
- [x] Feature
- [x] Bug Fix

## Description
Add `:focus-within` for better tab navigation
Resolve https://github.com/thepracticaldev/dev.to/issues/6843
In this picture, I am using `tab` ,`shift + tab` , and `space bar` to navigate and select each tags.
test

## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a

### DIFF
--- a/app/assets/stylesheets/components/buttons.scss
+++ b/app/assets/stylesheets/components/buttons.scss
@@ -48,7 +48,8 @@
 
   &[href]:hover,
   &:hover:enabled,
-  &:active:enabled {
+  &:active:enabled,
+  &:focus:enabled {
     background-color: var(--bg-hover);
     border-color: var(--border-hover);
     box-shadow: var(--shadow-hover);
--- a/app/assets/stylesheets/preact/onboarding-modal.scss
+++ b/app/assets/stylesheets/preact/onboarding-modal.scss
@@ -319,7 +319,8 @@
       z-index: 2;
     }
 
-    &:hover {
+    &:hover,
+    &:focus-within {
       #{$item}__inner {
         background: transparent;
       }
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/preact/onboarding-modal.scss
```diff
@@ -319,15 +319,17 @@
       z-index: 2;
     }
 
-    &:hover {
+    &:hover,
+    &:focus-within {
```
maestromac, line 323: It has to be `:focus-within` instead of `:focus` because parent element that the focused button belongs to needs to be change transparency.

app/assets/stylesheets/components/buttons.scss
```diff
@@ -48,7 +48,8 @@
 
   &[href]:hover,
   &:hover:enabled,
-  &:active:enabled {
+  &:active:enabled,
+  &:focus:enabled {
```
maestromac, line 52: I've decided to also the focus effect to the `next-button` which is using `.crayons-btn`. This change should be harmless but would love to hear input from @ludwiczakpawel

app/assets/stylesheets/preact/onboarding-modal.scss
lisasy: Nothing! Awesome 😀

app/assets/stylesheets/components/buttons.scss
lisasy: Fascinating. Looks cool.

maestromac:
I had the same thought too @vaidehijoshi . I think we'll use [tabindex](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) for that. I would apply that to every page so I would keep that into a different PR

rhymes:
❤️

PR #7602, opened 2020-04-29
Clear Elasticsearch Data Instead of Resetting Entire Index for Specs

mstruve:
- [x] Optimization

## Description
By default if we are using Elasticsearch for a spec we will reset ALL of our indexes with a very brute force approach. This is fine until you have a lot of Elasticsearch dependant tests, then it gets very sloooooow. In all of our search specs I replaced the full elasticsearch reset with a simple clear data method which uses a `delete_by_query` to ensure there is no data in the index.

So how much faster does this make the specs? Here are before and afters of all the `/search/` specs.
```
# BEFORE: Complete Reset
.........................................................................................................................

Finished in 2 minutes 40.4 seconds (files took 7.88 seconds to load)
121 examples, 0 failures

# AFTER: Only Clearing Data
.........................................................................................................................

Finished in 50.01 seconds (files took 7.73 seconds to load)
121 examples, 0 failures

````

I kept the full reset helper and use it in the `base_spec` for search bc in that spec we are deleting and creating indexes all over the place so its easier to just do a full reset for those specs. 

https://github.com/thepracticaldev/dev.to/projects/6#card-37267458

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/spec/lib/data_update_scripts/index_chat_channel_memberships_to_elasticsearch_spec.rb
+++ b/spec/lib/data_update_scripts/index_chat_channel_memberships_to_elasticsearch_spec.rb
@@ -1,8 +1,9 @@
 require "rails_helper"
 require Rails.root.join("lib/data_update_scripts/20200218195023_index_chat_channel_memberships_to_elasticsearch.rb")
 
-describe DataUpdateScripts::IndexChatChannelMembershipsToElasticsearch, elasticsearch: true do
+describe DataUpdateScripts::IndexChatChannelMembershipsToElasticsearch do
   it "indexes chat channel memberships to Elasticsearch" do
+    clear_elasticsearch_data(Search::ChatChannelMembership)
     chat_channel_membership = FactoryBot.create(:chat_channel_membership)
     expect { chat_channel_membership.elasticsearch_doc }.to raise_error(Search::Errors::Transport::NotFound)
     described_class.new.run
--- a/spec/lib/data_update_scripts/index_chat_channel_memberships_to_elasticsearch_spec.rb
+++ b/spec/lib/data_update_scripts/index_chat_channel_memberships_to_elasticsearch_spec.rb
@@ -1,9 +1,8 @@
 require "rails_helper"
 require Rails.root.join("lib/data_update_scripts/20200218195023_index_chat_channel_memberships_to_elasticsearch.rb")
 
-describe DataUpdateScripts::IndexChatChannelMembershipsToElasticsearch do
+describe DataUpdateScripts::IndexChatChannelMembershipsToElasticsearch, elasticsearch: "ChatChannelMembership" do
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

File: spec/services/search/base_spec.rb [hunk 1]
```diff
@@ -1,13 +1,15 @@
...
-RSpec.describe Search::Base, type: :service, elasticsearch: true do
+RSpec.describe Search::Base, type: :service do
   let(:document_id) { 123 }
 
   before do
     # Need to use an existing index name to ensure proper data cleanup
     stub_const("#{described_class}::INDEX_NAME", "tags_#{Rails.env}")
     stub_const("#{described_class}::INDEX_ALIAS", "tags_#{Rails.env}_alias")
     stub_const("#{described_class}::MAPPINGS", Search::Tag::MAPPINGS)
+    Search::Tag.refresh_index
```
mstruve, line 11: When dealing with [bulk updating operations](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-update) and [bulk deleting operations](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html#docs-delete-by-query) you need to make sure you refresh your index before trying to do a second one otherwise you will end up with conflict errors.

spec/lib/data_update_scripts/index_chat_channel_memberships_to_elasticsearch_spec.rb
```diff
@@ -1,8 +1,9 @@
 require "rails_helper"
 require Rails.root.join("lib/data_update_scripts/20200218195023_index_chat_channel_memberships_to_elasticsearch.rb")
 
-describe DataUpdateScripts::IndexChatChannelMembershipsToElasticsearch, elasticsearch: true do
+describe DataUpdateScripts::IndexChatChannelMembershipsToElasticsearch do
```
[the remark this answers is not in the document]
mstruve: I like this! We only need to refresh after the spec in a few scenarios so I am going to skip that for most of them and only apply it for the ones that really need it.

spec/lib/data_update_scripts/resync_elasticsearch_documents_spec.rb
```diff
@@ -1,7 +1,12 @@
 require "rails_helper"
 require Rails.root.join("lib/data_update_scripts/20200410152018_resync_elasticsearch_documents.rb")
 
-describe DataUpdateScripts::ResyncElasticsearchDocuments, elasticsearch: true do
+describe DataUpdateScripts::ResyncElasticsearchDocuments, elasticsearch: %w[FeedContent User Tag] do
+  after do
+    Article::SEARCH_CLASS.refresh_index
```
mstruve, line 6: We only need to refresh after specs that have bulk updates so rather than do this for every Elasticsearch spec I choose to keep it to the ones that were bulk update specific

PR #7603, opened 2020-04-29
Modals

ludwiczakpawel:

- [x] Refactor
- [x] Optimization

## Description

This PR does three things:
- updates modals component (html&css)
- updates instances of that modal (only onboarding)
- introduces very first draft of modals in preact.

We will definitely iterate and expand it in future.

## Screenshots

image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [x] no documentation needed


rhymes:
How do i test this manually?

ludwiczakpawel:
@rhymes you have to run storybook `npm run storybook` and you can just preview it there.

Ridhwana:
@ludwiczakpawel on these PR's I think it would definitely be useful to add before and after screenshots to the description if possible 😊. It will make the manual reviews quicker because it will be easy to spot what to look for.

app/javascript/crayons/Modal/__stories__/Modal.stories.jsx
```diff
@@ -0,0 +1,42 @@
+import { h } from 'preact';
+import { withKnobs, text, boolean, select } from '@storybook/addon-knobs/react';
+import { Modal, Button } from '@crayons';
+
+import '../../storybook-utiltiies/designSystem.scss';
```
Zhao-Andy, line 5: Noticed this folder name typo; probably better to leave this for a different PR 🙃

ludwiczakpawel: cc @nickytonline

ludwiczakpawel:
@Ridhwana good point, i've added a screenshot!

@rhymes this guy https://github.com/thepracticaldev/dev.to/pull/7656 should fix the divider color issue for theme. Actually I've merged that branch into my PR here, so it should be good here as well.

app/views/classified_listings/index.html.erb
```diff
@@ -36,7 +36,7 @@
 
     <div class="classified-filters">
     <div class="classified-filters-categories">
-      <a href="<%= classified_listings_path %>" class="<%= "selected" if params[:category].blank? %> data-no-instant="true">all</a>
```
rhymes, line 39: this has come back from the side :-D the previous version was correct, the one without the spaces in the attribute

ludwiczakpawel: ugh??? i have no idea why is this included in my PR... git is weird... reverting that..

ludwiczakpawel: reverted.

app/javascript/crayons/Modal/__stories__/Modal.stories.jsx
nickytonline: Fixed it in #7671

app/assets/stylesheets/components/modals.scss
```diff
@@ -1,32 +1,80 @@
...
-  padding: $su-7;
-  max-width: 640px;
+  position: fixed;
+  left: 0;
+  right: 0;
+  top: 0;
+  bottom: 0;
+  display: flex;
+  align-items: center;
+  justify-content: center;
```
nickytonline, line 11:
```suggestion
  place-items: center;
```

app/assets/stylesheets/components/modals.scss
```diff
@@ -1,32 +1,80 @@
...
+      $bg: var(--card-bg),
+      $border: var(--card-color),
+      $color: var(--card-color)
+    );
+    display: flex;
+    flex-direction: column;
     max-width: 800px;
-    padding: $su-8;
+    width: 100%;
+    background: var(--card-bg); // TODO: ?
```
nickytonline, line 38: TODO? 👀

app/assets/stylesheets/components/modals.scss
```diff
@@ -1,32 +1,80 @@
...
+    z-index: 700;
```
nickytonline, line 39: Just making sure the z-indices merit being 500 up top and 700 here since we now have our utility classes for `z-index` which go up to `z-50`. If so, in another PR, we should audit the app for z-indices and see if they really need to be as high as they are set to.

app/assets/stylesheets/components/modals.scss
```diff
@@ -1,32 +1,80 @@
...
+    max-height: 100%;
+    pointer-events: auto;
+    @media (max-width: $breakpoint-s) {
+      border: 0 !important; // TODO: this is dirty...
```
nickytonline, line 43: TODO 👀

app/assets/stylesheets/components/modals.scss
```diff
@@ -1,32 +1,80 @@
...
+    }
+
+    @media (min-width: $breakpoint-s) {
+      max-height: calc(100% - #{$su-6});
+    }
+
+    &__header {
+      position: static;
```
nickytonline, line 51: Is this necessary? The default position of an element is `static`.

app/javascript/onboarding/components/EmailPreferencesForm.jsx
```diff
@@ -53,53 +53,55 @@ class EmailPreferencesForm extends Component {
...
-          currentSlideIndex={currentSlideIndex}
-        />
-        <div className="onboarding-content terms-and-conditions-wrapper">
-          <header className="onboarding-content-header">
-            <h1 className="title">Almost there!</h1>
-            <h2 className="subtitle">
-              Review your email preferences before we continue.
-            </h2>
-          </header>
+      <div className="onboarding-main crayons-modal">
```
nickytonline, line 56: Any reason the `<Modal />` component is not being used for onboarding? I'm guessing you just want the styling of a modal for onboarding? The same question applies to all the onboarding slides.

nickytonline: Looking at how you're using it, I now understand why you did it this way. However, there is a way to reuse the same `<Modal />` component, even in onboarding. Instead of forcing a title prop to be a string, it could be a string or component, allowing the flexibility to render whatever in the title area of the `<Modal /> component. We can go over this tomorrow if you'd like.

For more information on this, check out the [Slot Pattern](https://daveceddia.com/pluggable-slots-in-react-components/) in (P)React.

app/assets/stylesheets/components/modals.scss
ludwiczakpawel: woooo, smart! i love it! <3

app/javascript/onboarding/components/EmailPreferencesForm.jsx
ludwiczakpawel: Onboarding was created while ago and I didn't want to update it right away with this PR. This PR is only meant to introduce modal as a component, but not really implementing it anywhere in particular – I have only updated onboarding frontend just a tiny bit so it doesn't look broken.

I think modals also have some logic and JS wired that I don't want to "unwire" right now :D

But eventually - 100% agree, onboarding should be based on the Modal component.

app/assets/stylesheets/components/modals.scss
ludwiczakpawel: oh wait.. it doesn't work.. it does `justify-items` instead of `justify-content`. This would work if that was a `grid` instead of a `flex`. But grid won't behave the same way, so I'll stick to original solution.

app/assets/stylesheets/components/modals.scss
ludwiczakpawel: Yea z-indexes overall are magic numbers... Since it's a component I didn't really want to use utility-classes... But I 100% agree, we have to audit all our z-indexes at some point.

ludwiczakpawel:
@nickytonline good catches! thanks for reviewing. they were mainly leftovers from development stage.. i should have checked my CSS one more time before making a PR :)

PR #7604, opened 2020-04-29
Make DATABASE_URL_TEST optional

maestromac:
- [x] Bug Fix/DX

## Description
When relying on `DATABASE_URL` and forgetting to set `DATABASE_URL_TEST`, users would be hit with error similar to the following when setting up database.
```
Couldn't create 'PracticalDeveloper_test' database. Please check your configuration.
```

The need for `DATABASE_URL_TEST` could be pretty niche and I think this a good middle ground for  things to work as expected while providing the option for a separate database URL for those who look within `database.yml`

Resolve https://github.com/thepracticaldev/dev.to/issues/7557

## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a

### DIFF
--- a/docs/installation/postgresql.md
+++ b/docs/installation/postgresql.md
@@ -40,6 +40,9 @@ connection string.
 
 ```yml
 DATABASE_URL: postgresql://USERNAME:PASSWORD@localhost
+
+# Incase you use a separate database for test. This is optional.
+DATABASE_URL_TEST: postgresql://USERNAME:PASSWORD@localhost
 ```
 
 1. Replace `USERNAME` with your database username, `PASSWORD` with your database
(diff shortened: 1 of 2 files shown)


docs/installation/postgresql.md
```diff
@@ -40,6 +40,9 @@ connection string.
 
 ```yml
 DATABASE_URL: postgresql://USERNAME:PASSWORD@localhost
+
+# Incase you use a separate database for test. This is optional.
```
mstruve, line 44: This might be a little confusing, we do use a separate database but sometimes that database might be at a different URL. Might be more clear to reword this a little and are URL to it.

maestromac: Good point. Updated!

PR #7608, opened 2020-04-30
Fix finding existing podcast episode by website url

lightalloy:
- [x] Bug Fix

## Description
Stop trying to find an existing podcast episode by website url in case item link is empty in the rss.
I also added a comment for our controversial logic related to `unique_website_url?`.

## Added tests?
- [x] yes

### DIFF
--- a/app/models/podcast.rb
+++ b/app/models/podcast.rb
@@ -32,7 +32,9 @@ class Podcast < ApplicationRecord
     episode = PodcastEpisode.where(media_url: item.enclosure_url).
       or(PodcastEpisode.where(title: item.title)).
       or(PodcastEpisode.where(guid: item.guid.to_s)).presence
-    episode ||= PodcastEpisode.where(website_url: item.link).presence if unique_website_url?
+    # if unique_website_url? is set to true (the default value), we try to find an episode by website_url as well
+    # if unique_website_url? is set to false it usually means that website_url is the same for different episodes
+    episode ||= PodcastEpisode.where(website_url: item.link).presence if item.link.present? && unique_website_url?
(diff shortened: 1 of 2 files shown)


app/models/podcast.rb
```diff
@@ -32,7 +32,9 @@ def existing_episode(item)
     episode = PodcastEpisode.where(media_url: item.enclosure_url).
       or(PodcastEpisode.where(title: item.title)).
       or(PodcastEpisode.where(guid: item.guid.to_s)).presence
-    episode ||= PodcastEpisode.where(website_url: item.link).presence if unique_website_url?
+    # if unique_website_url? is set to true (the default value), we try to find an episode by website_url as well
+    # if unique_website_url? is set to false it usually means that website_url is the same for different episodes
+    episode ||= PodcastEpisode.where(website_url: item.link).presence if item.link.present? && unique_website_url?
```
lightalloy, line 37: Currently, `unique_website_url?` can only be set from the console, and is set to `false` only for one of the podcasts ("Base cs podcast"). I don't like the fact that this logic is hidden from the community admins but at the same time it is used, so we can't just remove it.

rhymes: How would we get rid of it?

lightalloy: The "feature" is not explicitly used, and, for now, I don't know if there are any episodes actually found by `website_url`, which are not possible to find by other fields.
So, it would be nice to explore this and either remove the feature or make it explicit.

rhymes: Let's add a card somewhere in our projects columns, we can explore this at some point. So we don't forget

PR #7609, opened 2020-04-30
Add multiple select to authentication providers configuration

rhymes:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Right now the authentication providers selection in the configuration is a bit cumbersome.

This PR switches to a select (I had to use [bootstrap-select](https://developer.snapappointments.com/bootstrap-select/) as the regular select had issues with multiple selection and default values, too much went into trying to get that right :D).

With this new select, see the `selectpicker` class in the select, it also opens the possibility of it being used in other configuration options. Out of the scope for this PR

auth_providers

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/controllers/internal/configs_controller.rb
+++ b/app/controllers/internal/configs_controller.rb
@@ -11,7 +11,9 @@ class Internal::ConfigsController < Internal::ApplicationController
     clean_up_params
 
     config_params.each do |key, value|
-      if value.respond_to?(:to_h)
+      if value.is_a?(Array)
+        SiteConfig.public_send("#{key}=", value.reject(&:blank?)) unless value.empty?
+      elsif value.respond_to?(:to_h)
         SiteConfig.public_send("#{key}=", value.to_h) unless value.empty?
       else
         SiteConfig.public_send("#{key}=", value.strip) unless value.nil?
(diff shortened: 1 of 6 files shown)


app/controllers/internal/configs_controller.rb
```diff
@@ -11,7 +11,9 @@ def create
     clean_up_params
 
     config_params.each do |key, value|
-      if value.respond_to?(:to_h)
+      if value.is_a?(Array)
```
rhymes, line 14: arrays also respond to `to_h` as they could be arrays of array to build a hash, this is not a case, so we need a special condition for them

app/controllers/internal/configs_controller.rb
```diff
@@ -59,11 +59,16 @@ def config_params
       shop_url
       sidebar_tags
       suggested_tags
+      tagline
```
rhymes, line 62: simply reordered

rhymes:
@Ridhwana it's already like this, it was implemented in https://github.com/thepracticaldev/dev.to/pull/7505 - it's just that we only have 2 providers and we use both on DEV so you can't see that :D 

It will be more evident if and when we add more

Ridhwana:
> @Ridhwana it's already like this, it was implemented in #7505 - it's just that we only have 2 providers and we use both on DEV so you can't see that :D
> 
> It will be more evident if and when we add more

Ah missed this PR, perfect thanks Rhymes. Nice work on this feature, it shaped up really well 😁

PR #7630, opened 2020-04-30
Create FastlyConfig & refactor params safe list

atsmith813:
- [x] Refactor
- [x] Feature
- [x] Optimization

## Description
After some discussion, we came to the conclusion that bringing all of our Fastly configurations into our application code would be best. It provides transparency and makes Fastly easier to manage in the sense that anyone can update application code and submit PR.

This PR sets up the framework to do that while refactoring our existing Fastly updates into the new framework.

**New behavior** - taking other communities into account, the VCL snippet logic is now find _**or create**_. This also means if we change the name of a snippet in code, we'll want to delete the old one manually in Fastly.

In the next PR I'll bring all the VCL scripts into the application code. In the one following that I'll bring conditions in. And so on. I'd rather test my process on the script I'm most familiar with before jumping into the deep end without my floaties on.

For more on my approach, check out this [document](https://www.notion.so/devto/FastlyConfig-Outline-b71bdf5aa6b94e129836bf2f99ee85d7) I wrote up.

If you're interested, check out VCL syntax highlighting for a few editors:
- [VS Code](https://marketplace.visualstudio.com/items?itemName=thomas-baumgaertner.vcl)
- [Vim](https://vimawesome.com/plugin/vcl-vim-plugin)
- [Sublime](https://github.com/brandonwamboldt/sublime-varnish)
- [Atom](https://atom.io/packages/language-vcl)

https://github.com/thepracticaldev/dev.to/projects/9#card-36754397
https://github.com/thepracticaldev/dev.to/projects/9#card-36127635
https://github.com/thepracticaldev/dev.to/projects/9#card-36455908

## Added tests?
- [x] yes
I'm not sure how to test `fastly_config/base.rb`. I'm very open to ideas!

## Added to documentation?
- [x] yes

- We need to delete the old safe params VCL in Fastly.
- Announce to the team the method of updating safe params has been updated.

men_in_black_will_smith_memory_flash_gif


app/services/fastly_config/snippets.rb
```diff
@@ -0,0 +1,62 @@
...
+          dynamic: 0,
+          name: snippet_name,
+          service_id: ApplicationConfig["FASTLY_SERVICE_ID"],
+          type: "init",
+          version: new_version.number
+        }
+
+        snippet = fastly.create(Fastly::Snippet, snippet_options)
+        log_to_datadog("create", snippet, new_version)
+      end
```
atsmith813, line 38: This is new behavior, find or create the snippet. If a community has Fastly configured, the scripts found in code will be updated on their Fastly.

app/services/fastly_config/snippets.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    def find_snippet(fastly, new_version, snippet_name)
+      fastly.get_snippet(ApplicationConfig["FASTLY_SERVICE_ID"], new_version.number, snippet_name)
+    rescue Fastly::Error => e
+      error_message = JSON.parse(e.message)
+      raise e unless error_message["msg"] == "Record not found"
+
+      nil
+    end
```
atsmith813, line 50: Fastly operates on the assumption that snippets exist so it raises when error when one isn't found 😞

app/services/fastly_config/update.rb
```diff
@@ -0,0 +1,58 @@
...
+        raise e unless unauthorized_error?(error_msg) && Rails.env.development?
+
+        nil
+      end
+
+      private
+
+      def get_active_version(service)
+        service.versions.reverse_each.detect(&:active?)
+      end
```
atsmith813, line 34: I want to make note of this logic in case it becomes an issue in the future.

This logic is to find the active version on Fastly. Unfortunately, it's not on the service object itself. The other way to find it is to parse out the details on the service object, pluck the version number of the active version, and make another API call to get the active version object. I'm trying to keep API calls to a minimum.

Versions is a slowly growing infinite Array. Anytime we make an update to Fastly, we create a new version. As of this writing we're on version 125. It's common for the active version to be the most recent, but that's not always true. This solution should be performant enough for the foreseeable future as the Array grows. And there's no way to delete a version, you'd have to create an entirely new service to start the count over.

app/services/fastly_config/update.rb
```diff
@@ -0,0 +1,58 @@
+module FastlyConfig
```
atsmith813, line 1: All of this logic was built on the thinking that the most common case is that we do _not_ update Fastly. Updates are few and far between. That's why I added logic to determine if an update is even required - to save us from several unnecessary API calls during deployment.

There are 1 or 2 spots that could be further optimized to re-use data from the initial check when we go to update. But I'm holding off on that until I see how the other configs shakeout.

app/services/fastly_config/base.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    def update(new_version)
+      updated_files.each { |filename| update_config(new_version, filename) }
+    end
+
+    private
+
+    def get_updated_files(files: self.class::FASTLY_FILES)
+      updated_files = []
+      Dir.glob(files).each { |filename| updated_files << filename if file_updated?(filename) }
```
citizen428, line 23: Ruby recently added `filter_map` which performs a mapping and removes elements that would be `nil` afterward so you don't have to create an array first and push into it.

```ruby
Dir.glob(files).filter_map { |filename| filename if file_updated?(filename) }
```

app/services/fastly_config/base.rb
```diff
@@ -0,0 +1,35 @@
...
+      updated_files
+    end
+
+    def file_updated?
+      raise FastlyConfig::Errors::Error, "Fastly configs must implement their own file_updated? method"
```
citizen428, line 28: We have a generic `SubclassResponsibility` error that would work well here too.

app/services/fastly_config/snippets.rb
citizen428: To make this a bit more obvious, how about calling the method `upsert_snippet`?

app/services/fastly_config/snippets.rb
```diff
@@ -0,0 +1,62 @@
...
+      true
+    end
+
+    def find_snippet(fastly, new_version, snippet_name)
+      fastly.get_snippet(ApplicationConfig["FASTLY_SERVICE_ID"], new_version.number, snippet_name)
+    rescue Fastly::Error => e
+      error_message = JSON.parse(e.message)
+      raise e unless error_message["msg"] == "Record not found"
+
+      nil
```
citizen428, line 49: Silently swallowing all Fastly errors that are not "Record not found" makes me a bit uneasy. What other errors are possible? Should we at least log them?

app/services/fastly_config/update.rb
```diff
@@ -0,0 +1,58 @@
...
+  # Handles updates to our Fastly configurations
+  class Update
+    FASTLY_CONFIGS = %w[Snippets].freeze
+
+    class << self
```
citizen428, line 6: This `run` method has become pretty long. Maybe we should make this follow a pattern we use in other services, i.e.

```ruby
...
    def self.call
      new.run
    end

    attr_reader :service

    def initialize
        fastly = Fastly.new(api_key: ApplicationConfig["FASTLY_API_KEY"])
        @service = fastly.get_service(ApplicationConfig["FASTLY_SERVICE_ID"])
       ...
    end

    def run(configs: FASTLY_CONFIGS)
      ...
    end
```

docs/backend/fastly.md
```diff
@@ -28,27 +41,23 @@ to see examples of this.
 
 Previously this was a manual process done by an internal team member. Now we do
 it programmatically using the Fastly
-[gem](https://github.com/fastly/fastly-ruby) as of
-[this pr](https://github.com/thepracticaldev/dev.to/pull/7279).
+[gem](https://github.com/fastly/fastly-ruby).
 
-## How it works
+### How it works
```
citizen428, line 46: :tada:

app/services/fastly_config/snippets.rb
atsmith813: I'd like to keep it generic because it's inherited from the `Base` class to be used in all configs. I changed it to `upsert_config`

app/services/fastly_config/snippets.rb
atsmith813: Unfortunately Fastly raises all errors as `Fastly::Error` and the devil is in the details (the message). What do you mean to log them? We re-raise all errors except that one. Do you mean to log them anyways in case that logic is ever faulty?

app/services/fastly_config/base.rb
atsmith813: That's awesome! TIL 🎉

app/services/fastly_config/snippets.rb
citizen428: 1. I misread the code.
2. I'm just paranoid.

app/services/fastly_config/snippets.rb
citizen428: Good point and sensible change, thanks 👍

PR #7631, opened 2020-04-30
Limit 1 message every second for  single user

mstruve:
- [x] Bug Fix

## Description
Limit the number of messages one user can send to another. It is worth it to note that when a [throttle block returns nil, we do NOT hit Redis](https://github.com/kickstarter/rack-attack/blob/master/lib/rack/attack/throttle.rb#L26). 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/config/initializers/rack/attack.rb
+++ b/config/initializers/rack/attack.rb
@@ -23,7 +23,7 @@ class Rack::Attack
     end
   end
 
-  throttle("message_throttle", limit: 1, period: 1) do |request|
+  throttle("message_throttle", limit: 2, period: 1) do |request|
     if request.path.starts_with?("/messages") && request.post? && request.env["HTTP_FASTLY_CLIENT_IP"].present?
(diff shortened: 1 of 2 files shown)


config/initializers/rack/attack.rb
```diff
@@ -22,4 +22,10 @@ class Rack::Attack
       request.env["HTTP_FASTLY_CLIENT_IP"].to_s
     end
   end
+
+  throttle("message_throttle", limit: 1, period: 1) do |request|
```
atsmith813, line 26: Maybe we bump this up to 2 messages per second? Sometimes if I'm pasting a link or doing short burst messages, I legitimately exceed 1 message/second. I imagine that doesn't happen with most people though.

mstruve: I started with 2 then thought is that really possible and lowered it to 1. I could see in a copy paste scenario maybe so I will up it

PR #7636, opened 2020-04-30
RateLimit Refactor: Consolidate Logic for Checking a Limit and Raising an Error

mstruve:
- [x] Refactor

## Description
Consolidate the controller logic used to check the rate limit for an action and then raise an error if that action has exceeded its limit. 

## Added tests?

alt_text

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -78,4 +78,12 @@ class ApplicationController < ActionController::Base
   def touch_current_user
     current_user.touch
   end
+
+  def rate_limit!(action)
+    rate_limiter.check_limit!(action)
+  end
+
+  def rate_limiter
+    RateLimitChecker.new(current_user)
+  end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,6 +9,10 @@ class ApplicationController < ActionController::Base
 
   rescue_from ActionView::MissingTemplate, with: :routing_error
 
+  rescue_from RateLimitChecker::LimitReached do |exc|
+    error_too_many_requests(exc)
+  end
+
   def not_found
     raise ActiveRecord::RecordNotFound, "Not Found"
   end
@@ -24,6 +28,11 @@ class ApplicationController < ActionController::Base
 
   def bad_request
     render json: "Error: Bad Request", status: :bad_request
+  end
+
+  def error_too_many_requests(exc)
+    response.headers["Retry-After"] = exc.retry_after
+    render json: { error: exc.message, status: 429 }, status: :too_many_requests
   end
 
   def authenticate_user!
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -109,7 +109,7 @@ class CommentsController < ApplicationController
     end
   # See https://github.com/thepracticaldev/dev.to/pull/5485#discussion_r366056925
   # for details as to why this is necessary
-  rescue Pundit::NotAuthorizedError
+  rescue Pundit::NotAuthorizedError, RateLimitChecker::LimitReached
     raise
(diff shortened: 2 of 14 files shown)


app/controllers/comments_controller.rb
```diff
@@ -52,7 +52,7 @@ def edit
   # POST /comments
   # POST /comments.json
   def create
-    if RateLimitChecker.new(current_user).limit_by_action("comment_creation")
+    if rate_limiter.limit_by_action(:comment_creation)
```
rhymes, line 55: what if we use the old and tested technique of raising the error and rescueing from it in a central way? Same as we do in the API.

https://github.com/thepracticaldev/dev.to/blob/db7aa197fa75bbe15ce73df4fa6bfea5627d6a06/app/controllers/api/v0/api_controller.rb#L20-L22

Afer all here we just render and bail

Same goes for the other "if rate limit; then bail" examples :D

app/controllers/image_uploads_controller.rb
```diff
@@ -7,7 +7,7 @@ def create
     authorize :image_upload
 
     begin
-      rate_limiter = rate_limit!
+      limit_uploads
```
rhymes, line 10: I remember @vaidehijoshi suggesting the usage of a before action. If we establish that pattern it will also be very clear for us which endpoints are rate limited (in addition to those that are rate limited inside `rack-attack`)

app/controllers/comments_controller.rb
```diff
@@ -52,11 +52,7 @@ def edit
   # POST /comments
   # POST /comments.json
   def create
-    if RateLimitChecker.new(current_user).limit_by_action("comment_creation")
-      skip_authorization
-      render json: { error: "too many requests" }, status: :too_many_requests
-      return
-    end
+    rate_limit!(:comment_creation)
```
vaidehijoshi, line 55: should this one also be in a `before_action`, similar to [what you did with](https://github.com/thepracticaldev/dev.to/pull/7636/files#diff-8d80e37c99b30c15c5a8044a782ba1a6R3) `ImageUploadsController`?

app/controllers/organizations_controller.rb
```diff
@@ -3,7 +3,7 @@ class OrganizationsController < ApplicationController
   rescue_from Errno::ENAMETOOLONG, with: :log_image_data_to_datadog
 
   def create
-    rate_limit!
+    rate_limit!(:organization_creation)
```
vaidehijoshi, line 6: same question re: `before_action` here

app/controllers/application_controller.rb
```diff
@@ -9,6 +9,10 @@ class ApplicationController < ActionController::Base
 
   rescue_from ActionView::MissingTemplate, with: :routing_error
 
+  rescue_from RateLimitChecker::LimitReached do |exc|
```
vaidehijoshi, line 12: is `exc` the actual error? (curious why we didn't go with `e` or `err` or `error` here...

asking because `exc` makes me think it's not an error object

app/controllers/comments_controller.rb
mstruve: I would need to make another method for it since you can't pass arguments to before_actions, I tried! which is why I left it this way.

app/controllers/application_controller.rb
maestromac: I believe that's `exception` ?

mstruve: Yes, exc is the error or exception. This is how we had it in the API controller and I literally just moved the code over here without messing with it 🤷@rhymes is the fan of exec, I'm good with whatever.

vaidehijoshi: Ok, good to know. Don't bother with it, I was just curious.

PR #7637, opened 2020-04-30
Allow admins to configure default users to follow during onboarding

vaidehijoshi:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
- [x] Allows the onboarding "follow users" slide to be configured by admins.
- [x] Also does some refactoring of the `user_suggestions_spec.rb` spec.

Closes https://github.com/thepracticaldev/dev.to/issues/7106.

The suggested "users to follow" can now be set by admins via the `SiteConfig`:
Screen_Shot_2020-04-30_at_2_12_00_PM

Those suggested default users will show up in onboarding as a fallback:
Screen_Shot_2020-04-30_at_11_24_01_AM

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed (although I added some comments internally)

### DIFF
--- a/app/controllers/internal/configs_controller.rb
+++ b/app/controllers/internal/configs_controller.rb
@@ -60,6 +60,7 @@ class Internal::ConfigsController < Internal::ApplicationController
       shop_url
       sidebar_tags
       suggested_tags
+      suggested_users
       tagline
     ]
 
@@ -79,11 +80,13 @@ class Internal::ConfigsController < Internal::ApplicationController
 
   def clean_up_params
     config = params[:site_config]
-    config[:suggested_tags] = config[:suggested_tags].downcase.delete(" ") if config[:suggested_tags]
     config[:sidebar_tags] = config[:sidebar_tags].downcase.delete(" ") if config[:sidebar_tags]
+    config[:suggested_tags] = config[:suggested_tags].downcase.delete(" ") if config[:suggested_tags]
+    config[:suggested_users] = config[:suggested_users].downcase.delete(" ") if config[:suggested_users]
   end
 
   def bust_relevant_caches
-    CacheBuster.bust("/tags/onboarding") # Needs to change when suggested_tags is edited
+    # Needs to change when suggested_tags is edited.
+    CacheBuster.bust("/tags/onboarding")
   end
 end
--- a/app/controllers/internal/configs_controller.rb
+++ b/app/controllers/internal/configs_controller.rb
@@ -80,9 +80,9 @@ class Internal::ConfigsController < Internal::ApplicationController
 
   def clean_up_params
     config = params[:site_config]
-    config[:sidebar_tags] = config[:sidebar_tags].downcase.delete(" ") if config[:sidebar_tags]
-    config[:suggested_tags] = config[:suggested_tags].downcase.delete(" ") if config[:suggested_tags]
-    config[:suggested_users] = config[:suggested_users].downcase.delete(" ") if config[:suggested_users]
+    %i[sidebar_tags suggested_tags suggested_users].each do |param|
+      config[param] = config[param].downcase.delete(" ") if config[param]
+    end
   end
 
   def bust_relevant_caches
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -8,29 +8,26 @@ class UsersController < ApplicationController
   before_action :authenticate_user!, only: %i[onboarding_update onboarding_checkbox_update]
   rescue_from Errno::ENAMETOOLONG, with: :log_image_data_to_datadog
 
-  DEFAULT_FOLLOW_SUGGESTIONS = %w[ben jess peter maestromac andy liana].freeze
+  DEFAULT_FOLLOW_SUGGESTIONS = SiteConfig.suggested_users
+  INDEX_ATTRIBUTES_FOR_SERIALIZATION = %i[id name username summary profile_image].freeze
+  private_constant :INDEX_ATTRIBUTES_FOR_SERIALIZATION
 
   def index
     if !user_signed_in? || less_than_one_day_old?(current_user)
-      @users = User.where(username: DEFAULT_FOLLOW_SUGGESTIONS)
+      # Theoretically, no user should get to this point since they should
+      # already be signed in, but this is here as a safeguard in case they do.
+      @users = default_suggested_users
       return
     end
-
     @users =
       if params[:state] == "follow_suggestions"
-        Suggester::Users::Recent.new(
-          current_user,
-          attributes_to_select: INDEX_ATTRIBUTES_FOR_SERIALIZATION,
-        ).suggest
+        determine_follow_suggestions(current_user)
       elsif params[:state] == "sidebar_suggestions"
         Suggester::Users::Sidebar.new(current_user, params[:tag]).suggest.sample(3)
       else
         User.none
       end
   end
-
-  INDEX_ATTRIBUTES_FOR_SERIALIZATION = %i[id name username summary profile_image].freeze
-  private_constant :INDEX_ATTRIBUTES_FOR_SERIALIZATION
 
   # GET /settings/@tab
   def edit
@@ -275,6 +272,19 @@ class UsersController < ApplicationController
 
   def sanitize_user_params
     params[:user].delete_if { |_k, v| v.blank? }
+  end
+
+  def default_suggested_users
+    @default_suggested_users ||= User.where(username: DEFAULT_FOLLOW_SUGGESTIONS)
+  end
+
+  def determine_follow_suggestions(current_user)
+    recent_suggestions = Suggester::Users::Recent.new(
+      current_user,
+      attributes_to_select: INDEX_ATTRIBUTES_FOR_SERIALIZATION,
+    ).suggest
+
+    recent_suggestions.empty? ? default_suggested_users : recent_suggestions
   end
 
   def render_update_response
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -282,7 +282,7 @@ class UsersController < ApplicationController
       attributes_to_select: INDEX_ATTRIBUTES_FOR_SERIALIZATION,
     ).suggest
 
-    recent_suggestions.empty? ? default_suggested_users : recent_suggestions
+    recent_suggestions.presence || default_suggested_users
   end
 
   def render_update_response
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -6,9 +6,9 @@ class UsersController < ApplicationController
   ]
   after_action :verify_authorized, except: %i[index signout_confirm add_org_admin remove_org_admin remove_from_org]
   before_action :authenticate_user!, only: %i[onboarding_update onboarding_checkbox_update]
+  before_action :set_suggested_users, only: %i[index]
   rescue_from Errno::ENAMETOOLONG, with: :log_image_data_to_datadog
 
-  DEFAULT_FOLLOW_SUGGESTIONS = SiteConfig.suggested_users
   INDEX_ATTRIBUTES_FOR_SERIALIZATION = %i[id name username summary profile_image].freeze
   private_constant :INDEX_ATTRIBUTES_FOR_SERIALIZATION
 
@@ -268,8 +268,12 @@ class UsersController < ApplicationController
     params[:user].delete_if { |_k, v| v.blank? }
   end
 
+  def set_suggested_users
+    @suggested_users = SiteConfig.suggested_users
+  end
+
   def default_suggested_users
-    @default_suggested_users ||= User.where(username: DEFAULT_FOLLOW_SUGGESTIONS)
+    @default_suggested_users ||= User.where(username: @suggested_users)
   end
 
   def determine_follow_suggestions(current_user)
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -13,12 +13,6 @@ class UsersController < ApplicationController
   private_constant :INDEX_ATTRIBUTES_FOR_SERIALIZATION
 
   def index
-    if !user_signed_in? || less_than_one_day_old?(current_user)
-      # Theoretically, no user should get to this point since they should
-      # already be signed in, but this is here as a safeguard in case they do.
-      @users = default_suggested_users
-      return
-    end
     @users =
       if params[:state] == "follow_suggestions"
         determine_follow_suggestions(current_user)
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -62,6 +62,7 @@ class SiteConfig < RailsSettings::Base
   # Onboarding
   field :onboarding_taskcard_image, type: :string, default: "https://practicaldev-herokuapp-com.freetls.fastly.net/assets/staggered-dev.svg"
   field :suggested_tags, type: :array, default: %w[beginners career computerscience javascript security ruby rails swift kotlin]
+  field :suggested_users, type: :array, default: %w[ben jess peter maestromac andy liana]
 
   # Images
   field :main_social_image, type: :string, default: "https://thepracticaldev.s3.amazonaws.com/i/6hqmcjaxbgbon8ydw93z.png"
--- a/app/views/internal/configs/show.html.erb
+++ b/app/views/internal/configs/show.html.erb
@@ -298,7 +298,7 @@
               <%= f.text_field :suggested_users,
                                class: "form-control",
                                value: SiteConfig.suggested_users.join(","),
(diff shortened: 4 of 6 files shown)


app/controllers/users_controller.rb
```diff
@@ -8,30 +8,27 @@ class UsersController < ApplicationController
...
-  DEFAULT_FOLLOW_SUGGESTIONS = %w[ben jess peter maestromac andy liana].freeze
+  DEFAULT_FOLLOW_SUGGESTIONS = SiteConfig.suggested_users
+  INDEX_ATTRIBUTES_FOR_SERIALIZATION = %i[id name username summary profile_image].freeze
+  private_constant :INDEX_ATTRIBUTES_FOR_SERIALIZATION
 
   def index
     if !user_signed_in? || less_than_one_day_old?(current_user)
-      @users = User.where(username: DEFAULT_FOLLOW_SUGGESTIONS)
+      # Theoretically, no user should get to this point since they should
+      # already be signed in, but this is here as a safeguard in case they do.
```
vaidehijoshi, line 18: I truly couldn't find a scenario when a user could hit this endpoint since we should be checking that a user is signed in via the `ApplicationController`. I added this note for more context, because otherwise, this feels a bit confusing (at least, it was for me!) 😕

app/models/site_config.rb
```diff
@@ -62,6 +62,7 @@ class SiteConfig < RailsSettings::Base
   # Onboarding
   field :onboarding_taskcard_image, type: :string, default: "https://practicaldev-herokuapp-com.freetls.fastly.net/assets/staggered-dev.svg"
   field :suggested_tags, type: :array, default: %w[beginners career computerscience javascript security ruby rails swift kotlin]
+  field :suggested_users, type: :array, default: %w[ben jess peter maestromac andy liana]
```
vaidehijoshi, line 65: I just set this as the default for DEV since that's what we seemed to be falling back to [here](https://github.com/thepracticaldev/dev.to/pull/7637/files#diff-4e05ad0d64e6100656b63ad1e78f32c5L11).

spec/requests/user/user_suggestions_spec.rb
```diff
@@ -1,69 +1,100 @@
...
 
-      other_user = create(:user)
-      create(:article, user: other_user, tags: [tag.name])
+    context "when the user is not signed in" do
+      it_behaves_like "a default suggested_users request", "/users"
+    end
 
-      sign_in user
+    context "when no state params are present" do
+      before { sign_in user }
```
vaidehijoshi, line 35: Cleaned up these specs quite a bit and added some shared examples (🎉!!!). Notably, I added two specific contexts for when we `sign_in user` versus when we don't (since we have a path for that scenario [here](https://github.com/thepracticaldev/dev.to/pull/7637/files#diff-4e05ad0d64e6100656b63ad1e78f32c5R16)).

spec/requests/internal/configs_spec.rb
```diff
@@ -145,6 +145,16 @@
           post "/internal/config", params: { site_config: { suggested_tags: "hey, haha,hoHo, Bobo Fofo" }, confirmation: confirmation_message }
           expect(SiteConfig.suggested_tags).to eq(%w[hey haha hoho bobofofo])
         end
+
+        it "removes space suggested_users" do
+          post "/internal/config", params: { site_config: { suggested_users: "piglet, tigger,eeyore, Christopher Robin, kanga,roo" }, confirmation: confirmation_message }
```
maestromac, line 150: These are great sets of suggested users

spec/requests/user/user_suggestions_spec.rb
```diff
@@ -2,68 +2,106 @@
 
 RSpec.describe "Users", type: :request do
   describe "GET /users" do
-    it "returns no users if state is not present" do
-      get users_path
+    let(:user) { create(:user) }
+    let(:suggested_users_list) { %w[eeyore] }
+    let!(:suggested_user) { create(:user, name: "Eeyore", username: "eeyore", summary: "I am always sad :(") }
```
maestromac, line 7: :(

spec/requests/user/user_suggestions_spec.rb
```diff
@@ -2,68 +2,106 @@
...
+        expect(response).to have_http_status(:ok)
 
-      get users_path
+        response_user = response.parsed_body.first
+        expect(response_user["id"]).to eq(suggested_user.id)
+        expect(response_user["name"]).to eq(suggested_user.name)
+        expect(response_user["username"]).to eq(suggested_user.username)
+        expect(response_user["summary"]).to eq(suggested_user.summary)
+        expect(response_user["profile_image_url"]).to eq(ProfileImage.new(suggested_user).get(width: 90))
+        expect(response_user["following"]).to be(false)
```
maestromac, line 25: Not a blocker but there's another neat way to [test json with only one expectation ](https://stackoverflow.com/a/8393268)

app/controllers/users_controller.rb
benhalpern: Hmmmmmmmmm since we don't call `authenticate_user!` on this action someone could get here but I think this is unnecessary and the dirty remnants of some past behavior and a quick fix.

benhalpern: So we could remove this IMO

benhalpern:
Is this just for scenarios when the suggestions return none or is it for general configuration by admins to come up with curated suggestions to complement the others?

vaidehijoshi:
> Is this just for scenarios when the suggestions return none or is it for general configuration by admins to come up with curated suggestions to complement the others?

This is only for when the default suggestions service returns no users. I've noticed that in new communities, the suggested list of "users to follow" is blank by default, so this is to mitigate that, and to allow new community admins to set a fallback.

app/controllers/users_controller.rb
vaidehijoshi: Cool, I'm definitely a fan of removing this and the test I added for it entirely 👍

spec/requests/user/user_suggestions_spec.rb
vaidehijoshi: gorgeousssssss, using that now ✨

benhalpern:
> That being said, I think we could definitely explore letting admins come up with curated suggestions; that just seemed out of scope for my ultimate goal of just ensuring that the "suggested users" page on onboarding never ends up being empty :)

Cool this makes sense. If it were curated, I'd say we should be able to tag suggested users by tag (which we actually used to have but got rid of it for some reason)....

If this is only as a fallback, maybe we want to make that clear in the naming/description?

vaidehijoshi:
@benhalpern Sounds good. I'm going to make a ticket for the "curated suggested users" feature and plan to tackle that during cool down. I suppose that once we implemented that feature, we'd have both `suggested_users` and `curated_users` within the `SiteConfig`?

The latest commit includes some updated copy on the `/internal/config` page that explicitly states how these "suggested users" are used in the system:
Screen Shot 2020-04-30 at 4 28 15 PM

benhalpern:
Screen Shot 2020-04-30 at 7 34 56 PM

On narrow article show views does it make sense to lift the snackbar above the reactions bar so it doesn't block what the user wants to do?

Also just a copy thing, let's change "article" to "post" in the snackbar.

spec/requests/user/user_suggestions_spec.rb
vaidehijoshi: Update: I had to remove the shared examples; for some (currently inexplicable to me) reason, shared examples caused failures on CI that I couldn't repro locally 🤷

vaidehijoshi:
Update: It seems like Travis is failing, even though the tests are actually..._passing_? I see a ton of warnings and an error at the end of the log that seems to indicate that the "database does not exist" 🤯.

If anyone has any ideas about what is happening and/or why...I'm all ears. Not really sure where to go from here at this moment, but maybe tomorrow an answer will reveal itself.

app/controllers/internal/configs_controller.rb
```diff
@@ -79,11 +80,13 @@ def extra_authorization_and_confirmation
 
   def clean_up_params
     config = params[:site_config]
-    config[:suggested_tags] = config[:suggested_tags].downcase.delete(" ") if config[:suggested_tags]
     config[:sidebar_tags] = config[:sidebar_tags].downcase.delete(" ") if config[:sidebar_tags]
+    config[:suggested_tags] = config[:suggested_tags].downcase.delete(" ") if config[:suggested_tags]
+    config[:suggested_users] = config[:suggested_users].downcase.delete(" ") if config[:suggested_users]
```
citizen428, line 85: Maybe at one point we can change this to make it less repetitive?

```ruby
%i[sidebar_tags suggested_tags suggested_users].each do |param|
  config[param] = config[param].downcase.delete(" ") if config[param]
end
```

app/controllers/users_controller.rb
```diff
@@ -277,6 +268,19 @@ def sanitize_user_params
...
+    @default_suggested_users ||= User.where(username: DEFAULT_FOLLOW_SUGGESTIONS)
+  end
+
+  def determine_follow_suggestions(current_user)
+    recent_suggestions = Suggester::Users::Recent.new(
+      current_user,
+      attributes_to_select: INDEX_ATTRIBUTES_FOR_SERIALIZATION,
+    ).suggest
+
+    recent_suggestions.empty? ? default_suggested_users : recent_suggestions
```
citizen428, line 281: This seems to be the only place in the controller where you use this, so maybe we can move it into `Suggester::Users::Recent` which performs the check itself. Then this method can follow the "tell don't ask" principle :

```ruby
def determine_follow_suggestions(current_user)
  Suggester::Users::Recent.new(
    current_user,
    attributes_to_select: INDEX_ATTRIBUTES_FOR_SERIALIZATION,
  ).suggest
end
```

I'd even wrap up the `new` and `suggest` calls into a combined `call` class method as we do for several other service objects:

```ruby
def self.call(user, attributes_to_select:)
  new(...).suggest
end
```

then this whole method essentially is just

```ruby
Suggester::Users::Recent.call(current_user,  attributes_to_select: INDEX_ATTRIBUTES_FOR_SERIALIZATION)
```

app/models/site_config.rb
citizen428: I think it's ok to leave this in the PR, but at some point we may to stop adding defaults at all, as they make no sense for other communities. Instead I think we need a sort of onboarding process for community admins that walks them through all of these settings. I'll open an issue for this.

Ridhwana: @citizen428 there is a card that outlines this at a high level, but we haven't fleshed it out as yet, maybe you could add some detail to it? https://github.com/thepracticaldev/dev.to/projects/15#card-37093809

app/views/internal/configs/show.html.erb
```diff
@@ -293,6 +293,16 @@
               </div>
             </div>
 
+            <div class="form-group">
+              <%= f.label :suggested_users %>
+              <%= f.text_field :suggested_users,
+                               class: "form-control",
+                               value: SiteConfig.suggested_users.join(","),
+                               placeholder: "List of valid usernames: comma separated, letters only e.g. beginners,javascript,ruby,swift,kotlin" %>
+              <div class="alert alert-info">Determines which users are suggested to follow to new users during onboarding (comma
```
Ridhwana, line 302:
```
Determines which users are suggested to follow to new users during onboarding (comma separated, letters only). Please note that these users will be shown as a fallback if no recently-active commenters or producers can be suggested
```

I feel like we should try to keep these descriptions to be shorter and more direct, and so I took a suggestive stab at trying to capture your points in the sentence below. I removed `comma separated, letters only` as its a repetition from the placeholder.

```
Determines suggested followers to new users during onboarding. N.B. These will be your fallback users if no recently-active content creators can be suggested
```

vaidehijoshi: @Ridhwana I’m open to changing this but personally I don’t know what “N.B.” is, so I find this more confusing.

Ridhwana: Ah, N.B is Note well, and is more commonly used as part of British English, which is what I speak 😁 let's keep it as please note instead of N.B then :)

vaidehijoshi: Ahhh makes much more sense now. I’ll try to rework this. I do worry that if we don’t mention the “comma separated” but that some admins might not know that this is how to list the usernames though. Is that not a valid concern?

Ridhwana: Hmmm so the use case for it would mostly likely be if there's only a single item in the list. I think let's just keep it 👍🏽

vaidehijoshi: Okay. I’m going to leave this sentence as-is in that case! 👍

rhymes: Trivia: we use N.B. in Italian as well, it's the latin `Nota Bene` https://en.wikipedia.org/wiki/Nota_bene

app/views/internal/configs/show.html.erb
```diff
@@ -293,6 +293,16 @@
               </div>
             </div>
 
+            <div class="form-group">
+              <%= f.label :suggested_users %>
+              <%= f.text_field :suggested_users,
+                               class: "form-control",
+                               value: SiteConfig.suggested_users.join(","),
+                               placeholder: "List of valid usernames: comma separated, letters only e.g. beginners,javascript,ruby,swift,kotlin" %>
```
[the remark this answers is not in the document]
rhymes: you can switch to using the `bootstrap-select` component I added in #7609 if you'd prefer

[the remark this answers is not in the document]
vaidehijoshi: @Ridhwana WOOPS, looks like some late evening copy-pasta 🍝. Will fix this!

app/views/internal/configs/show.html.erb
vaidehijoshi: @Ridhwana @rhymes I'm gonna start using this and trip up all my American English speaking friends 😂

app/controllers/internal/configs_controller.rb
vaidehijoshi: Ah of course. I was trying to not do too much refactor for this PR, but this seems obviously better, so I've fixed this 👍

app/controllers/users_controller.rb
vaidehijoshi: I think when I initially PRed this, I was using `default_suggested_users` in two locations (but I have since [removed it](https://github.com/thepracticaldev/dev.to/pull/7637/commits/61aa0a80c9b514228dfd489bc56b87fbda0c838e) when I discovered that it was covering a situation that probably shouldn't ever happen).

I expect to come back to this code next week during cool down, so if it's okay with you, I'm going to hold off on refactoring this and come back to this refactor in a separate PR :) I do think there is definitely room to fix this up, though, so I definitely agree with you there 👍

vaidehijoshi:
Figured out the issue with the failing tests and fixed them in the last commit. Turns out, the issue was with how/when `SiteConfig` was being referenced/loaded. It happens in a `before_action` now :)

app/controllers/users_controller.rb
```diff
@@ -277,6 +268,23 @@ def sanitize_user_params
...
+    @default_suggested_users ||= User.where(username: @suggested_users)
+  end
+
+  def determine_follow_suggestions(current_user)
+    recent_suggestions = Suggester::Users::Recent.new(
+      current_user,
+      attributes_to_select: INDEX_ATTRIBUTES_FOR_SERIALIZATION,
+    ).suggest
+
+    recent_suggestions.empty? ? default_suggested_users : recent_suggestions
```
mstruve, line 285: You could actually do presence here! I had no idea if that would work with an AR relation but it does! TIL!
```ruby
recent_suggestions.presence || default_suggested_users
```

vaidehijoshi: Nice, I should've remembered that! Gonna update with your suggestion and then merge.

rhymes:
A question: what happens if one or more of the suggested users delete their accounts over time? Or if they change their usernames. I wonder if there should be a "backup" in case the number of users fetched from the DB doesn't match the size of the array or if the array is simply too short.

vaidehijoshi:
> A question: what happens if one or more of the suggested users delete their accounts over time? Or if they change their usernames. I wonder if there should be a "backup" in case the number of users fetched from the DB doesn't match the size of the array or if the array is simply too short.

@rhymes I was thinking that admins should be responsible for knowing/checking if the featured users actually exist/match the usernames or not 🤔 If we added another backup, I feel like we'd start to have many configs, no? I'm not opposed to it, but I am a little weary of adding a "backup" config. For what it's worth, even if the array was short, the page will still render correctly.

PR #7644, opened 2020-05-01
Update bin/setup's db setup logic

maestromac:
- [x] Bug Fix/DX

## Description
Instead of trying to check if a database exists, this change will run db:migrate and it will run db:setup as a backup. This is more inline with how [Rails 6 will handle this](https://github.com/rails/rails/blob/98754de1412870d7dae9eba1c7bac944b9b90093/activerecord/lib/active_record/railties/databases.rake#L311). The downside is new contributor will see an error briefly. 😬 

I made this change because I rely on `DATABASE_URL` to access postgresql in Docker; Checking if database exist with `psql -o /dev/null...` does not work for me and I can't think of a way for `bin/setup` to access env var in `application.yml`.

Also, I removed dome dupe seeds.
n/a
n/a
## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -372,49 +372,6 @@ end
 ##############################################################################
 
 counter += 1
-Rails.logger.info "#{counter}. Creating Classified Listings"
-
-users_in_random_order.each { |user| Credit.add_to(user, rand(100)) }
-users = users_in_random_order.to_a
-
-listings_categories = ClassifiedListingCategory.pluck(:id)
-listings_categories.each.with_index(1) do |category_id, index|
-  # rotate users if they are less than the categories
-  user = users.at(index % users.length)
-  2.times do
-    ClassifiedListing.create!(
-      user: user,
-      title: Faker::Lorem.sentence,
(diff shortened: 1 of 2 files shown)


db/seeds.rb
```diff
@@ -371,49 +371,6 @@
 
 ##############################################################################
 
-counter += 1
```
maestromac, line 374: Pretty sure these dupes appear because of some merge conflict.

db/seeds.rb
```diff
@@ -371,49 +371,6 @@
...
-Rails.logger.info "#{counter}. Creating Classified Listings"
```
maestromac, line 375: Pretty sure this giant chunk of dupe appear because of a merge conflict.

rhymes:
Will this fail with Gitpod? In reference to this https://github.com/thepracticaldev/dev.to/issues/7040

maestromac:
@rhymes just tested it, it doesn't fail gitpod 🙌 !

PR #7646, opened 2020-05-01
Remove Unnecessary .downcases From the Stories::Controller

### LINKED ISSUE #7580: Remove Unnecessary .downcase From Stories::Controller
While working on handling redirects for uppercase usernames, I came across other methods in the `Stories::Controller` that use `.downcase` as well. The PR that handles redirecting from uppercase usernames to lowercase usernames (#7550 ) handles the downcasing necessary to make this happen, yet there are some places that still use `.downcase` unnecessarily. In order to remove what is now dead code, I must remove the `.downcase`s that no longer need to be used.

juliannatetreault:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
While working on handling redirects for uppercase usernames (#7550 ), I came across other methods in the `Stories::Controller` that use `.downcase`. The previous PR (#7550 ) handles redirecting from uppercase usernames to lowercase usernames, making some of the `.downcase`s in the `Stories::Controller` unnecessary. This PR removes some of the`.downcases` from the `Stories::Controller` since the downcasing should be handled by PR #7550 .

To test that this works properly navigate to one of the index pages effected (user, organization, or podcast) by the removal of the `.downcase`s and test that the route (on the index page only) redirects to the lowercase version of the route when the capitalized version is given.

Closes #7580 

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

bye bye bye

### DIFF
--- a/spec/requests/stories_index_spec.rb
+++ b/spec/requests/stories_index_spec.rb
@@ -182,7 +182,7 @@ RSpec.describe "StoriesIndex", type: :request do
 
   describe "GET podcast index" do
     include_examples "redirects to the lowercase route" do
-      let(:path) { "/#{create(:podcast).slug.upcase}" }
+      let(:path) { "/#{build(:podcast).slug.upcase}" }
     end
 
     it "renders page with proper header" do
@@ -346,13 +346,13 @@ RSpec.describe "StoriesIndex", type: :request do
 
   describe "GET user_path" do
     include_examples "redirects to the lowercase route" do
-      let(:path) { "/#{create(:user).username.upcase}" }
+      let(:path) { "/#{build(:user).username.upcase}" }
     end
   end
 
   describe "GET organization_path" do
     include_examples "redirects to the lowercase route" do
-      let(:path) { "/#{create(:organization).slug.upcase}" }
+      let(:path) { "/#{build(:organization).slug.upcase}" }
     end
   end
 end
--- a/spec/requests/stories_index_spec.rb
+++ b/spec/requests/stories_index_spec.rb
@@ -177,6 +177,14 @@ RSpec.describe "StoriesIndex", type: :request do
       get "/" + podcast.slug
       expect(response.body).to include(podcast.title)
     end
+
+    it "redirects to the lowercase route for podcast slugs" do
+      podcast = create(:podcast)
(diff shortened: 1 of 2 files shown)


spec/requests/stories_index_spec.rb
```diff
@@ -171,6 +181,10 @@
   end
 
   describe "GET podcast index" do
+    include_examples "redirects to the lowercase route" do
+      let(:path) { "/#{create(:podcast).slug.upcase}" }
```
rhymes, line 185: small suggestion, if you only need the slug, you don't need to create the object in the database, using `create`, but you can simply build it in memory, using `build()` (same parameters).

For example:

```ruby
irb(main):004:0> FactoryBot.build(:podcast).slug.upcase
=> "SLUG-3"
```

same for user and organization :-)

spec/requests/stories_index_spec.rb
```diff
@@ -1,5 +1,15 @@
...
 
+RSpec.shared_examples "redirects to the lowercase route" do
+  context "when a path contains uppercase characters" do
+    it "redirects to the lowercase route" do
+      get path
+      expect(response).to have_http_status(:moved_permanently)
+      expect(response).to redirect_to(path.downcase)
+    end
+  end
+end
```
vaidehijoshi, line 11: Nice use of shared examples. Another option is also to pass path in as a parameter `...do |path|` so that you don't need to use a block (and you can use `it_behaves_like`).

But that's more just a ProTip; I think what you have works just fine 👍

spec/requests/stories_index_spec.rb
juliannatetreault: Good suggestion; I'll make this change!

spec/requests/stories_index_spec.rb
juliannatetreault: Thanks! I actually had it that way initially but that setup didn't read as well or make as much sense for this particular spec, so I opted to set it up this way. 😄

vaidehijoshi: Yep, I think this does work better because of the `build` stuff.

PR #7651, opened 2020-05-01
Adds an extra check for Android Native app for playing podcasts

fdoxyz:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Since we merged #7554 I ran a quick test and found out we do have older native app versions that won't be able to play podcasts because **they do have** `DEV-Native-android` as user agent but the native bridge side is still not implemented. 

This simple check fixes this issue by falling back to the full on web player until the users have a a version installed that does include the necessary native functions.

#7554 

When the native side is in production (https://github.com/thepracticaldev/DEV-Android/pull/72) we can have this check removed although it wouldn't hurt to keep it for a while since not everyone updates immediately

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

double check

### DIFF
--- a/app/assets/javascripts/initializers/initializePodcastPlayback.js
+++ b/app/assets/javascripts/initializers/initializePodcastPlayback.js
@@ -114,7 +114,8 @@ function initializePodcastPlayback() {
     return (
       navigator.userAgent === 'DEV-Native-android' &&
       typeof AndroidBridge !== 'undefined' &&
-      AndroidBridge !== null
+      AndroidBridge !== null &&
+      AndroidBridge.metadataPodcast !== undefined
     );
   }
 


app/assets/javascripts/initializers/initializePodcastPlayback.js
```diff
@@ -114,7 +114,8 @@ function initializePodcastPlayback() {
     return (
       navigator.userAgent === 'DEV-Native-android' &&
       typeof AndroidBridge !== 'undefined' &&
-      AndroidBridge !== null
+      AndroidBridge !== null &&
```
fdoxyz, line 117: @maestromac I kept the double check against `undefined` & `null` because of the linter complaining about using a single `==` comparison. If you rather only do one I think we can scrap the `null` comparison instead

maestromac: Ah, it's no biggie 👍

PR #7653, opened 2020-05-01
Log RateLimit hits to Datadog for Better Reporting

mstruve:
- [x] Feature
- [x] Optimization

## Description
Rather than sending (almost) every single rate limit hit to Slack, lets send a metric to Datadog that we can track and alert on. The goal here is to send this data to Datadog. Once the data is there @michael-tharrington can then decide what thresholds he wants to set for what actions to determine when he gets alerted. Datadog then will be responsible for sending that information to Slack rather than our app. 

This will allow us to set thresholds and alert only when users are excessively hitting rate limits rather than every single time which sometimes is not concerning if they only hit it once, like a pentester trying to see if a rate limiter exists. 

https://github.com/thepracticaldev/dev.to/projects/9#card-37397577

## Added tests?
- [x] yes

alt_text


mstruve:
> will DataDog be logging this to a specific channel in Slack?

Yep! We can put it in whatever channel @michael-tharrington wants

michael-tharrington:
So excited for this! I'm going to make sure we talk about this in our next support meeting. I think this is a super 🆒 thing to monitor.

PR #7654, opened 2020-05-01
Add new VCL snippets for Fastly

atsmith813:
- [x] Feature
- [x] Optimization

## Description
This PR adds 3 new VCL snippets for Fastly
- **Stale while revalidate controls** - existing snippet. Bringing it into our application code.
- **Check for remember user token in cookie to set header** - existing snippet. Bringing it into our application code.
- **Bad bot detection** - new VCL snippet :trollface:. It will block bot traffic at the edge. Inspired by [this VCL](https://github.com/wolstena/varnish-bad-bot-detection/blob/master/bad_bot_detection.vcl).

Fastly strongly recommends using custom HTTP codes in the 600s or 700s as to not conflict with common HTTP codes. I'm thinking we'll start by using 700s for our custom Fastly errors.

https://github.com/thepracticaldev/dev.to/projects/9#card-37400263
https://github.com/thepracticaldev/dev.to/projects/9#card-37400247

## Added tests?

## Added to documentation?
- [x] no documentation needed

Hit https://dev.to with a bad user-agent and confirm it throws `403`.

block_that_shit_gif


config/fastly/snippets/bad_bot_detection.vcl
```diff
@@ -0,0 +1,140 @@
...
+    || req.http.user-agent == "EmailSiphon"
+    || req.http.user-agent == "EmailWolf 1.00"
+    || req.http.user-agent == "ESurf15a 15"
+    || req.http.user-agent == "ExtractorPro"
+    || req.http.user-agent == "Franklin Locator 1.8"
+    || req.http.user-agent == "FSurf15a 01"
+    || req.http.user-agent == "Full Web Bot 0416B"
+    || req.http.user-agent == "Full Web Bot 0516B"
+    || req.http.user-agent == "Full Web Bot 2816B"
+    || req.http.user-agent == "Fuzz Faster"
```
benhalpern, line 67: This is a direct compare, but the actual user agent is `Fuzz Faster U Fool` and then a version, so it should be a `~` for partial match eh?

https://github.com/ffuf/ffuf/blob/08ec6bad2aad1e265b20d9036a16a8f30679ed3e/pkg/runner/simple.go#L102

atsmith813: Ah, yep - good call! I didn't realize that wasn't the actual user agent. Updated.

config/fastly/snippets/bad_bot_detection.vcl
```diff
@@ -0,0 +1,140 @@
...
+    || req.http.user-agent ~ "^Jakarta"
+    || req.http.user-agent ~ "IDBot"
+    || req.http.user-agent ~ "id-search"
+    || req.http.user-agent ~ "User-Agent"
+    || req.http.user-agent ~ "compatible ;"
+    || req.http.user-agent ~ "ConveraCrawler"
+    || req.http.user-agent ~ "^Mozilla$"
+    || req.http.user-agent ~ "libwww"
+    || req.http.user-agent ~ "lwp-trivial"
+    || req.http.user-agent ~ "PHP/"
```
mstruve, line 15: Does this mean any script created written with PHP would be considered a bot?

atsmith813: Most likely - I can remove this one.

rhymes:
I wonder if we can use the same logic that we use for A/B testing to detect bots.
There's a very handy Ruby library called `browser` (which for now it's a transitive dependency in our `Gemfile.lock` but could be "promoted") that would give us two free things:

1. something in the guise of:

```ruby
> Browser.new("lwp-trivial").bot?
true
```

1. a curated list of [bots](https://github.com/fnando/browser/blob/master/bots.yml), [non-bots](https://github.com/fnando/browser/blob/master/bot_exceptions.yml) and [search engines](https://github.com/fnando/browser/blob/master/search_engines.yml). These lists are in yml files, that could be accesed (and extended with our own) simply with:

```ruby
> Browser::Bot.bots.keys.size
=> 298
```

We could add the ones that are in the script and not in that list, and use a string template in the VCL file to add the list dynamically before we upload it to Fastly.

Let me know if this makes any sense or it's too much.

Great work by the way!

config/fastly/snippets/check_for_remember_user_token_in_cookie_to_set_header.vcl
```diff
@@ -0,0 +1,19 @@
+sub vcl_recv {
```
rhymes, line 1: I don't understand why we have this script at all, we never use this header upstream...

atsmith813: I think @benhalpern might know. IIRC, it was there to address some quirk...maybe with something on how we're doing shielding?

atsmith813:
@rhymes that's a good idea! It brings up a discussion. I'm certainly not opposed to something like that, it would give us more power.

2 points of hesitation with this specifically. Maybe we have some ideas to get past them?

**1.** I'm trying to avoid dynamic insertion as much as possible - it introduces a lot of complexity.

 If we did decide we'd gain enough from dynamic insertion (very likely there will be at some point), I think what might be best is to set it up with template VCL files as follows. First, we create a template. So it would be `config/fastly/snippets/bad_bot_detection.vcl` and `config/fastly/snippets/templates/bad_bot_detection.vcl` (same file in the templates folder). The template file would hold some `SOME_GENERIC_PLACEHOLDER_TO_GSUB`. We'd use the template to generate the VCL file `bad_bot_detection.vcl`, and then continue on in our script as usual. The reason for this is it would be painful to debug if you don't output the result of the VCL into the codebase.

For example, if that dynamic bot list added a good bot (my second hesitation), contributors wouldn't be able to confirm that bot was added to the list unless they had production Fastly access to view the outputted VCL. They might find it by running that bot list command and seeing the results, but they would never know for sure what VCL was exactly in place. That would be a painful bug to track. I know me personally, I'd be pulling my hair out to find a production only bug where the server was never receiving any requests at all (because it was using an unknowingly blocked user-agent). So we'd want a way to see the results in code for debugging instead of keeping them behind an admin wall. It's always nice to be able to `grep` for answers too.

Even the old way where I had params in a yml file wasn't the best for debugging. If someone added a weird param that's valid yml but messed up the VCL syntax, that'd be difficult to debug without Fastly access. FWIW - Fastly tells you clear as day about any syntax errors and where. We could add validation, but that doesn't exist for every configuration type. Whereas, when a contributor has the actual VCL file in code, they have many tools available to debug (syntax highlighting, VCL validators, etc.).

Whatever we do, the ideal solution would fit all configuration types (conditions, snippets, etc.).

**2.** That [bot list](https://github.com/fnando/browser/blob/master/bots.yml) referenced above contains several bots we wouldn't want to block (Ruby, Google, etc.). Accidentally blocking a good bot would introduce some painful bugs mentioned above.

**TLDR/Summary**
I'm down to comb through the list and add any we don't have manually while keeping the idea of dynamic insertion in the back of our minds. If you skipped to this summary, above I mention:
> Whatever we do, the ideal solution would fit all configuration types (conditions, snippets, etc.).

So even if we decide dynamic insertion is valuable, I'd like to hold off until all the Fastly configs are brought into code so I can better understand how to handle all configuration types.

We are also planning to have a meeting to add more configurations to Fastly. I suspect we'll make some updates so powerful that even this list might become obsolete :trollface:

rhymes:
@atsmith813 thanks a lot for the insight. I wholeheartedly agree, having a static list tied to a specific version and `Browser` is intentend to identify which user agents are users and which are bots but as you said, not all bots are malicious. 

Also, there's already a lot of value to be gained by this "internalization" of VCL.

I still wish we had a VCL parser to check syntax and test but we'll get there :D

atsmith813:
@rhymes completely agree - I'll add that in soon. It felt like more of a "nice to have" than a "need to have".

When I wrote the `bad_bot_detection.vcl`, I used an online VCL validator and I tested it in the Fastly fiddle to make sure my logic worked. So, ideally, anyone submitting VCL will have gone through that process. I plan to be diligent on PR reviews adding VCL by doing that same process myself.

Under the current setup, our script (via the `fastly` gem itself) will raise an error for invalid VCL syntax. Fastly automatically validates it when you upload it. This error would stop a deployment and we'd have to address it.

I was thinking of adding some logic that can be called in a console for dev purposes (i.e. `FastlyConfig::Validate.call("some_vcl_file.vcl")`). And we could write up something (a spec?) to validate all VCL files. If someone added a bad `.vcl`, the spec fails. That way we're programmatically catching syntax issues before deployment and not relying solely on contributors and PR reviews. I added a [card](https://github.com/thepracticaldev/dev.to/projects/9#card-37427987) so it's publicly documented 😈

benhalpern:
This is great. In addition to checking for validity it would be nice to ultimately get the logic tested as well. It would help us move past really basic logic.

I think it might also be fun to try and start experimenting with some [Compute@Edge](https://www.fastly.com/products/edge-compute/beta) if we can justify it under our overarching philosophy. (That philosophy being a preference for the "majestic monolith" and avoiding total vendor lock-in);

PR #7478, opened 2020-04-23
Removed yarn integrity check

rhymes:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

With https://github.com/thepracticaldev/dev.to/pull/7469 and https://github.com/thepracticaldev/dev.to/pull/7466 the whole concept of "yarn integrity checked" was removed from Webpacker.

Also changed `yarn install` to `yarn install --frozen-lockfile` per [new README](https://github.com/rails/webpacker#deployment):

> When compiling assets for production on a remote server, such as a continuous integration environment, it's recommended to use yarn install --frozen-lockfile to install NPM packages on the remote host to ensure that the installed packages match the yarn.lock file.

PR #5329, opened 2020-01-02
Rails 5.2 features: enable cache versioning (recyclable keys)

rhymes:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Among the notable Rails 5.2 features is the ability to recycle cache keys, aka cache versioning. It is supported by all builtin cache stores and now that we've migrated to Redis and are using the builtin `:redis_cache_store` - https://github.com/thepracticaldev/dev.to/blob/ee8fd7cbc22f2edaced8163b4eabb192ebde5849/config/environments/production.rb#L106 - we can effectively enable this feature.

This will get us one step closer to Rails 5.2 defaults and help transitioning to Rails 6

**NOTE: this does NOT generate cache misses, nor it generates new keys, it just helps with a smarter invalidation**

**What does this feature do?**

Adds efficiency to cache invalidation by separating the cache key from the cache version. 

> With the new method of cache versioning, the keys stay consistent, but the cache_version is stored inside the cache entry and manually checked when pulling an entry from the cache

**Are there any issues?**

The issues are related to `:dalli_store` that doesn't support cache versioning, but we've migrated away from that

**Other useful info**

Model objects have a [.cache_key_with_version](https://api.rubyonrails.org/v5.2/classes/ActiveRecord/Integration.html#method-i-cache_key_with_version) method that could be useful when we do low level caching around the code (as an alternative to explicitly join `.updated_at`) which I just found out about, it's already available even with cache versioning disabled.

## Related Tickets & Documents

- [Cache Invalidation Complexity: Rails 5.2 and Dalli Cache Store](https://www.schneems.com/2018/10/17/cache-invalidation-complexity-rails-52-and-dalli-cache-store/) this explains what cache versioning is and how it benefits Rails builtin caching logic

- [Rails 5.2 guides: Low-Level caching](https://guides.rubyonrails.org/v5.2/caching_with_rails.html#low-level-caching) an example on how to use `cache_key_with_version`, see also this example:

```ruby
[1] pry(main)> User.last.cache_key_with_version
=> "users/11-20191231095247258085"
[2] pry(main)> x = "users/11-20191231095247258085"
[3] pry(main)> u = User.last
[4] pry(main)> u.touch
[5] pry(main)> u.reload.cache_key_with_version
=> "users/11-20200102101249941793"
[6] pry(main)> u.reload.cache_key_with_version == x
=> false
```

[...]

PR #4827, opened 2019-11-15
Rails 5.2 features: Enable default protect from forgery

rhymes:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Documentation Update

## Description

After https://github.com/thepracticaldev/dev.to/pull/4216 I'm submitting this PR to enable "protect from forgery" by default.

This won't affect us as all our controllers (internals included) inherit from our `ApplicationController` which already has a call to `protect_from_forgery`: 

https://github.com/thepracticaldev/dev.to/blob/fbbb02e241d4d91f5df4941ffe904220f3f636dc/app/controllers/application_controller.rb#L2

What Rails does when this is enabled by default is to add the following in the `ActionController::Base` controller:

https://github.com/rails/rails/blob/2b42f33fda9733d0ca91c0c1b88cc4df4b7230e8/actionpack/lib/action_controller/railtie.rb#L77-L79

Our call already overrides any possible default.

PR #4826, opened 2019-11-15
Rails 5.2 features: Enable AES-256-GCM on encrypted messages

rhymes:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Documentation Update

## Description

After https://github.com/thepracticaldev/dev.to/pull/4216 I'm submitting this PR to enable `AES-256-GCM` on encrypted messages as well.

Rails uses this algorithm to encrypt messages and to generate `rails secrets`. In our code we use `ActiveSupport::MessageEncryptor` in the Google Analytics controller to scramble the IP:

https://github.com/thepracticaldev/dev.to/blob/fbbb02e241d4d91f5df4941ffe904220f3f636dc/app/controllers/ga_events_controller.rb#L30-L33

After manual test I did determine that indeed the default encryption algorithm has been changed. For obvious reasons switching algorithms will generate different strings, but we do not decrypt those strings anywhere so this shouldn't be a problem.

PR #7658, opened 2020-05-02
Upgrade to Rails 6.0

etagwerker:
Hey @rhymes @tannakartikey,

I went ahead and started this exploratory PR. 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR adds a couple of changes to the project: 

- Adds `Gemfile.next` to make it easier to dual boot the app 
- Adds one gemfile to the Travis CI configuration

- [ ] Sorry about the Rubocop TODO changes, but it wouldn't let me commit otherwise. I'll try to remove them later. If you are curious, it was complaining about a duplicate `gem "rails"` in the Gemfile. 😺 

https://github.com/thepracticaldev/dev.to/issues/2879

N/A 

## Added tests?

- [ ] yes
- [x] no, because I'm not sure this PR needs them? 

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

I'm pretty sure you cookies generated by Rails 5.2 are not going to be compatible with Rails 6.0. So users might get signed out? (not sure) 

Both excited and freaked out! xD

giphy


rhymes:
Look great! I think we can ignore rubocop for now with `git commit --no-verify`.

I think we should eventually run `rails app:update` as indicated in the [upgrade guide](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#the-update-task) though by doing that we'd necessarily change files that are shared by both.

Anyway, I think that this multi step strategy makes sense:

- this PR activates CI on Rails 6
- we fix deprecations and errors
- another PR actually upgrades the code to use Rails 6 and switches `Gemfile.next` to Rails master

What do you think @etagwerker ?

etagwerker:
@rhymes Sounds like a good plan. 

I think this PR can be a guide for the upgrade. Any changes that are backwards compatible, should be submitted as a PR targeting `master`.

rhymes:
@etagwerker both me and @maestromac tried to push changes to this PR but we can't:

```shell
remote: Permission to fastruby/dev.to.git denied to rhymes.
fatal: unable to access 'https://github.com/fastruby/dev.to.git/': The requested URL returned error: 403
```

is it a protected fork?

etagwerker:
@rhymes @maestromac Not sure why that happened. Maybe I checked the wrong box when I submitted the PR (?) 

Just sent you invitations to the repo which should fix it: https://github.com/fastruby/dev.to/invitations

rhymes:
@etagwerker it works, thanks!

rhymes:
@etagwerker sorry I had to basically undo the `Gemfile.next` trick as it's wasn't working both locally and on Travis - https://travis-ci.com/github/thepracticaldev/dev.to/jobs/338821108.

I kept encountering various failures in tests (running them with `bundle exec --gemfile=Gemfile.next`) and after booting the app with that Gemfile I couldn't login or logout.

After spending sometime in the debugger I realized there was an easier way to figure out if the app was or wasn't working with Rails 6: actually update it to it.

So I did, and as you can see here https://travis-ci.com/github/thepracticaldev/dev.to/builds/167789826, all tests were green. Also locally everything was working.

I still think the `Gemfile.next` trick is interesting but maybe we can circle back on that after this upgrade so that we can track Rails 6.1 effectively. One thing that I foresee is going to be hard to work with is how quickly stale `Gemfile.next.lock` becomes. Yesterday dependabot sent us a bunch of updates and I had to manually update it with `bundle update --gemfile=Gemfile.next` and still I had slightly different versions of gems that weren't related to Rails (that's because we don't lock specific versions for most of the gems and the Bundle gemfile resolution algorithm).

Long story short: after fixing a bunch of visible deprecations, I think we have DEV running on Rails 6 and it was pretty easy 🔥 Still a few kinks but we're almost there.

Thanks for the help and the initial nudge. As I said I plan to circle back on `Gemfile.next` with a branch to track 6.1 once all the dust is settled.

Next steps are all about "Rails-six-fying" the code by at least migrating to Zeitwerk as the auto loader and enabling parallel testing if possible. We'll do that throughout the following weeks, let me know if you want to help :)

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -117,7 +117,9 @@ def respond_to_invitation(previous_status)
         flash[:settings_notice] = "Invitation to  #{channel_name} accepted. It may take a moment to show up in your list."
       else
         send_chat_action_message("@#{current_user.username} added @#{@chat_channel_membership.user.username}", current_user, @chat_channel_membership.chat_channel_id, "joined")
-        NotifyMailer.channel_invite_email(@chat_channel_membership, @chat_channel_membership.user).deliver_later
+
+        NotifyMailer.with(membership: @chat_channel_membership, inviter: @chat_channel_membership.user).channel_invite_email.deliver_later
```
rhymes, line 121: this is the new API in Rails 6 for parameterized mailers: https://github.com/rails/rails/pull/34591

app/labor/sticky_article_collection.rb
```diff
@@ -24,7 +24,7 @@ def tag_articles
     @tag_articles ||= Article.published.tagged_with(article_tags, any: true).
       includes(:user).
       where("positive_reactions_count > ? OR comments_count > ?", reaction_count_num, comment_count_num).
-      where.not(id: article.id, user_id: article.user_id).
+      where.not(id: article.id).where.not(user_id: article.user_id).
```
rhymes, line 27: Rails 6.1 is going to change the algorithm from NOR to NAND if you pass multiple arguments to a `where.not`, so I fixed the deprecation, see https://github.com/rails/rails/blob/6-0-stable/activerecord/CHANGELOG.md#rails-600rc1-april-24-2019

app/services/audit/event/util.rb
```diff
@@ -1,3 +1,4 @@
+# NOTE: [Rails 6] Util class is obsolete in Rails 6
```
rhymes, line 1: Going to leave it there for now, I'll slowly chip out other deprecated stuff and aim to upgrade to 6.0 defaults in the future

bin/setup
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env ruby
```
rhymes, line 1: ran `rails app:update` here

bin/update
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env ruby
```
rhymes, line 1: ran `rails app:update` here. We could probably remove this but I want to look into the new `rails db:*` tasks before, I think there are some that can simplify our setup

config/application.rb
```diff
@@ -50,24 +50,9 @@ class Application < Rails::Application
     # Therefore we disable "per_form_csrf_tokens" for the time being.
     config.action_controller.per_form_csrf_tokens = false
 
-    # Enable CORS for API v0
-    # (logging is only activated when debug is enabled)
-    debug_cors = ENV["DEBUG_CORS"].present? ? true : false
```
rhymes, line 55: moved in the default CORS initializer Rails 6 creates

config/application.rb
```diff
@@ -50,24 +50,9 @@ class Application < Rails::Application
...
-          # allow read operations, disallow custom headers (eg. api-key) and enable preflight caching
-          # NOTE: Chrome caps preflight caching at 2 hours, Firefox at 24 hours
-          # see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age#Directives
-          resource "/api/#{resource_name}/*", methods: %i[head get options], headers: [], max_age: 2.hours.to_i
-        end
-      end
-    end
+    # NOTE: [Rails 6]
+    # To improve security, Rails embeds the purpose and expiry metadata inside encrypted or signed cookies value.
+    config.action_dispatch.use_cookies_with_metadata = false
```
rhymes, line 55: this won't invalidate existing cookies but I'd rather activate it separately, see https://github.com/rails/rails/pull/32937

config/boot.rb
```diff
@@ -1,5 +1,4 @@
 ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../Gemfile", __dir__)
```
rhymes, line 1: ran `rails app:update` here

config/environments/production.rb
```diff
@@ -1,7 +1,4 @@
 Rails.application.configure do
-  # Verifies that versions and hashed value of the package contents in the project's package.json
-  config.webpacker.check_yarn_integrity = false
```
rhymes, line 3: this was long deprecated and removed. See #7478

spec/models/organization_spec.rb
```diff
@@ -240,14 +240,22 @@
 
       it "updates the paths of the organization's articles" do
         new_slug = "slug_#{rand(10_000)}"
-        organization.update(slug: new_slug)
+
+        sidekiq_perform_enqueued_jobs(only: Organizations::UpdateOrganizationArticlesPathsWorker) do
```
rhymes, line 244: no idea why this is currently working in master as it shouldn't since it didn't execute the jobs needed for the to pass 🤷

spec/requests/api/v0/articles_spec.rb
```diff
@@ -45,7 +45,7 @@
     context "without params" do
       it "returns json response" do
         get api_articles_path
-        expect(response.content_type).to eq("application/json")
+        expect(response.media_type).to eq("application/json")
```
rhymes, line 48: `content_type` will contain the full content type (with encoding and other options) in Rails 6.1, so 6.0 signals this access with a deprecation and suggests to move to `media_type` to check only that

rhymes:
I would like to merge https://github.com/thepracticaldev/dev.to/pull/8004 ahead of this and then rebase this one on top of master

rhymes:
Also, not sure why these https://travis-ci.com/github/thepracticaldev/dev.to/builds/167890168#L1327-L1330 are failing 🤷

etagwerker:
@rhymes Thanks for the update re: dual booting! Do you want me to submit an issue to keep track of this idea? It can be revisited once you get Rails 6.0 up and running in production. 😄 

Regarding the failures in `spec/system/notifications/notifications_page_spec.rb` I'm not sure what is the deal there. I tried to debug it a bit and it looks like a timing issue (?) -- I don't intend for this change to make it to `master`, but I wanted to try it in Travis CI: https://github.com/thepracticaldev/dev.to/pull/7658/commits/51630de2eb054994a785475281aa7bbbe738af4c 

One thing that is weird is that I see that spec working fine if I drop the `headless` option for the driver: 

```
Capybara.register_driver :headless_chrome do |app|
  options = Selenium::WebDriver::Chrome::Options.new(
    args: %w[no-sandbox disable-gpu window-size=1920,1080 --enable-features=NetworkService,NetworkServiceInProcess],
    log_level: :error,
  )

  Capybara::Selenium::Driver.new app, browser: :chrome, options: options
end
```

So I think it is a failure related to JS + running the scenario with a headless driver. Maybe some scenarios require to run with a non-headless driver?

rhymes:
@etagwerker 

> @rhymes Thanks for the update re: dual booting! Do you want me to submit an issue to keep track of this idea? It can be revisited once you get Rails 6.0 up and running in production. 😄

yeah, sure! We could also use it for a discussion around that as we have also to figure out what impact having double the amount of builds has

> Regarding the failures in `spec/system/notifications/notifications_page_spec.rb` I'm not sure what is the deal there. I tried to debug it a bit and it looks like a timing issue (?) -- I don't intend for this change to make it to `master`, but I wanted to try it in Travis CI: [51630de](https://github.com/thepracticaldev/dev.to/commit/51630de2eb054994a785475281aa7bbbe738af4c)

We do have a `wait_for_javascript` helper here

https://github.com/thepracticaldev/dev.to/blob/9c58ebc61e3fd581da3f8bf83c9904d162cb03e9/spec/support/initializers/capybara.rb#L38

I'll see if I can replace the `sleep` with that. Unfortunately we have several flaky tests and our code base is dependent on order, introducing a new version of the framework is bound to surface some of these problems or generate new ones.

I'll also ping other team members that are more familiar with such notifications code. Thanks for the effort!

> One thing that is weird is that I see that spec working fine if I drop the `headless` option for the driver:

That is weird indeed.

> So I think it is a failure related to JS + running the scenario with a headless driver. Maybe some scenarios require to run with a non-headless driver?

I always thought there was no difference, except the Chrome window appearing or not. I wonder also if it's a question of Chrome version. We do test against `chrome: 'stable'` in `.travis.yml` but Chrome stable changes in time...

rhymes:
@etagwerker for example, locally I have several of `./spec/system/notifications/notifications_page_spec.rb` tests fail with:

```shell
     3.2) Failure/Error: raise ActionController::RoutingError, "No route matches [#{env['REQUEST_METHOD']}] #{env['PATH_INFO'].inspect}"

          ActionController::RoutingError:
            No route matches [GET] "/stories/warm_comments/notifications"
```

and they fail both with or without `headless`

app/controllers/chat_channel_memberships_controller.rb
maestromac: Oh this is super neat!

vaidehijoshi: Whoa! Gotta teach myself to start thinking in the Rails 6 way 😅

app/queries/internal/moderators_query.rb
```diff
@@ -14,7 +14,7 @@ def self.call(relation: User.all, options: {})
                      role_id_for(:trusted),
                    ).order("users.comments_count DESC")
                  else
-                   relation.joins(:users_roles).
+                   relation.joins(:roles).
```
vaidehijoshi, line 17: Is this change due to a deprecation?

rhymes: Nope, it just doesn't work without that change. Not totally sure why, though it works in current master as well if I change it to `:roles` 🤷

.travis.yml
```diff
@@ -15,6 +15,8 @@ addons:
       - $(ls tmp/screenshots/*.png | tr "\n" ":")
       - $(ls tmp/csvs/*.csv | tr "\n" ":")
     debug: true
+gemfile:
+- Gemfile
```
etagwerker, line 19: @rhymes I guess now that we are not using dual booting, this change is unnecessary. 😄

config/application.rb
```diff
@@ -20,7 +20,7 @@
 module PracticalDeveloper
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 5.1
+    config.load_defaults 5.1 # NOTE: [Rails 6] we should at least work towards updating this to 5.2
```
etagwerker, line 23: @rhymes Does it cause a lot of problems if we try with 5.2 or 6.0?

lib/tasks/app_setup.rake
```diff
@@ -5,12 +5,7 @@ namespace :app_initializer do
     Rake::Task["search:setup"].execute
 
     puts "\n== Preparing database =="
-    begin
-      Rake::Task["db:migrate"].execute
-    rescue ActiveRecord::NoDatabaseError
-      puts "\n== Creating and Seeding database =="
-      system("bin/rails db:setup")
-    end
+    system!('bin/rails db:prepare')
```
mstruve, line 8: 🔥 🔥 🔥

app/queries/internal/moderators_query.rb
citizen428: `roles` is the actual association table and `users_roles` the join table, so I think it actually makes more sense like this anyway.

config/initializers/cors.rb
```diff
@@ -0,0 +1,19 @@
+# Be sure to restart your server when you modify this file.
+
+# Enable CORS for API v0 (logging is only activated when debug is enabled)
+debug_cors = ENV["DEBUG_CORS"].present? ? true : false
```
citizen428, line 4: Did Rails add it like this? `present?` already returns a boolean, so no need for the ternary.

.travis.yml
rhymes: yeah, I kept it as a reminder :D

config/application.rb
rhymes: Yep. if you look at https://github.com/thepracticaldev/dev.to/blob/master/config/initializers/new_framework_defaults_5_2.rb we still have at least one setting disabled. I think we can bump it later on making sure that setting stays disabled for the time being and them bump to 5.2.

I wouldn't bump to 6.0 yet, before we check each setting.

Anyway, I plan to send follow up PRs to address all the remaining deprecations and possibly "Rails-sixify" the code :D

rhymes:
@mstruve 

> This looks SUPER freaking solid! Question. We have some AR relations that are cached in Redis, do we know if those are going to break with this upgrade? For example:

Great question! How are they serialized now in the cache? Does Rails call `Marshal.dump()` on the objects?

If so, I did a super layman check (loaded the same object in Rails 5.2 and Rails 6 and checked if the strings were the same) and I can tell you they are not. So yeah, the marhsalling changes.

**BUT**

I did the following and it worked:

```ruby
# with Rails 6
> File.open('user.dump', 'wb') { |f| f.write(Marshal.dump(user)) }
# then with Rails 5.2
[1] pry(main)> user = Marshal.load(File.read('user.dump'))
=> #<User id: 11, articles_count: 0, available_for: nil, badge_achievements_count: 0, behance_url: nil, bg_color_hex: nil, blocked_by_count: 0, blocking_others_count: 0, checked_co...

# then with Rails 5.2
[4] pry(main)> File.open('user2.dump', 'wb') { |f| f.write(Marshal.dump(user2)) }

# and reloaded with Rails 6
[1] pry(main)> user2 = Marshal.load(File.read('user2.dump'))
=> #<User id: 11, articles_count: 0, available_for: nil, badge_achievements_count: 0, behance_url: nil, bg_color_hex: nil, blocked_by_count: 0, blocking_others_count: 0, checked_...
```

so the format is compatible :)

config/initializers/cors.rb
rhymes: Nope, I just copy pasted from our code from 

https://github.com/thepracticaldev/dev.to/blob/5e3f645d299f4e583b11c9322cebc14b414404c3/config/application.rb#L55

:D 

I'll fix it eventually!

Gemfile
```diff
@@ -101,7 +101,7 @@ gem "storext", "~> 3.1" # Add type-casting and other features on top of ActiveRe
 gem "stripe", "~> 5.22" # Ruby library for the Stripe API
 gem "strong_migrations", "~> 0.6" # Catch unsafe migrations
 gem "timber", "~> 3.0" # Great Ruby logging made easy
-gem "timber-rails", "~> 1.0" #  Timber integration for Rails
+gem "timber-rails", github: "timberio/timber-ruby-rails", branch: "master" # Timber integration for Rails
```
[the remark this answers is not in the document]
rhymes: @joshpuetz subcribed to the issue. Hopefully we can get rid of the gem altogether (I know it's in the SRE team's plans)

config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -0,0 +1,45 @@
+# Be sure to restart your server when you modify this file.
```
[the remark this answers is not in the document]
rhymes: yeah, this is installed by default by `rails app:update`. I'm going to flip them on (the ones we can enable at least), one at at a time, the same way we did for rails 5.2

See #4216, #4829, #4826, #4827, #5005, #5057 and #5329 :D

etagwerker:
@rhymes @joshpuetz awesome, thanks! 💯

rhymes:
@etagwerker thank you!

PR #7660, opened 2020-05-03
Decouple GitHub authentication from fetching GitHub issues

rhymes:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Currently we use users's OAuth tokens (randomly) to fetch *any* GitHub (GH from now on) issue.
The likely reason that was put in place is, I imagine, to spread out rate limits, but that's not really necessary with a few expedients.

There are a few reasons we don't want to do that anymore:

  - right now we have a **tight coupling** between the concept of "GH identity" and functionalities that use the GH client (fetching issues - this PR - and repositories)
  - users GH **identities might not exist** (other communities might opt not to have GH or disable it in the future in their lifetime if they see no one is using GH as a provider)
  - **tokens might have been revoked** (if a user disconnects the server app from their own GH settings or if a user has their app access restricted - aka banned - to the app, we shouldn't use their token in the first place)
  - GH could have **suspended an account** for any reason (there's an actual HTTP exception for that in Octokit as well :D)

There's also an argument to be made about removing the `GithubIssue` model altogether, but I kept it for the time being. The argument is that with HTTP caching plus Redis caching we might not need it anymore as it essentially serves stale content forever and ever (we check if the serialized issue is in the DB and then never fetch it ever again, which means that in the chance that any of the data of the issue or comment are ever altered, the users on DEV will never see that, same if an issue is deleted upstream)

Initially I set out only to decouple auth from the client, but I eventually expanded a bit as I found a few bugs, so this PR contains the following:

1. a new `Github::Client` object, very similar to our `Search::Client` which uses the server credentials, not the user's. This is what [GH documentation has to say about that](https://developer.github.com/v3/#oauth2-keysecret):

> Using your client_id and client_secret does not authenticate as a user, it will only identify your OAuth application to increase your rate limit. Permissions are only granted to users, not applications, and you will only get back data that an unauthenticated user would see. For this reason, you should only use the OAuth2 key/secret in server-to-server scenarios. Don't leak your OAuth application's client secret to your users.

So, that's good to know :-)

2. HTTP client side caching: Octokit [supports client side http caching](https://github.com/octokit/octokit.rb#caching) respecting the usual suspects 🤣 : `Cache-Control`, `Last-Modified` and so on
3. a refactoring of the code in `GithubIssue` and `GithubTag::GithubIssueTag` (which weirdly uses inner classes), with in addition bug fixes for the PR url generation and a bit more validation
4. UI fixes to display the avatar that had disappeared (it's hidden in production right now) and to have a title for comment embeds
5. a bunch of more tests
6. a data update script to cleanup existing `GithubIssue` rows

## Rate limiting and client side caching

GH's v3 API [documentation on rate limiting](https://developer.github.com/v3/#rate-limiting) is very clear:

> For API requests using Basic Authentication or OAuth, you can make up to 5000 requests per hour.

and it also gives a bunch of suggestions on how to extend those rate limits:

> If you exceed your rate limit using Basic Authentication or OAuth, you can likely fix the issue by caching API responses and using conditional requests.

Our client now, in addition to respecting rate limits and `retry-after` as before, [will also use conditional requests](https://developer.github.com/v3/#conditional-requests) (conditional requests are the standard HTTP caching mechanism).

Also:

> Note: Making a conditional request and receiving a 304 response does not count against your Rate Limit, so we encourage you to use it whenever possible.

Screenshot_2020-05-03 text(1)

Screen Shot 2020-05-03

*There's still this issue when a comment is too short and it's covered by the box shadow* but I'm not sure how to tackle that and we can work on it in a different PR:

Screen Shot 2020-05-03

## Added tests?

- [x] yes
- [ ] no, because I need help


app/liquid_tags/github_tag/github_issue_tag.rb
```diff
@@ -68,7 +85,19 @@ def valid_link?(link)
     end
 
     def raise_error
-      raise StandardError, "Invalid Github issue link"
+      raise StandardError, "Invalid GitHub issue, pull request or comment link"
+    end
+
+    def title
+      content_json[:title] || "Comment for"
```
rhymes, line 92: comments don't have a `title` attribute, so we add a verbatim title, see the screenshots

app/liquid_tags/github_tag/github_issue_tag.rb
```diff
@@ -1,40 +1,47 @@
...
     def render
       ActionController::Base.new.render_to_string(
         partial: PARTIAL,
         locals: {
-          title: @content_json[:title],
-          issue_number: @content_json[:number],
-          user_html_url: @content_json[:user][:html_url],
+          body: @body,
+          created_at: @created_at.rfc3339,
+          date: @created_at.utc.strftime("%b %d, %Y"),
```
rhymes, line 29: we use both for the `<time>` HTML element in the view

app/liquid_tags/github_tag/github_issue_tag.rb
```diff
@@ -46,9 +53,19 @@ def parse_link(link)
...
-      input = input.gsub(/\d{1,}#issuecomment-/, "comments/") if input.include?("#issuecomment-")
-      "https://api.github.com/repos/#{input.gsub(/.*github\.com\//, '')}"
+      path = input.gsub(/\?.*/, "")
+      path = path.gsub(/.*github\.com\//, "")
+
+      # issue fragments are ignored in the API call
+      path = path.gsub(/#issue-\d{1,}/, "") if path.include?("#issue-")
+
+      # GitHub's public PRs URLs are "/pull/{id}" but the API requires "/pulls/{id}"
+      path = path.gsub(/\/pull\//, "/pulls/") if path.include?("/pull/")
```
rhymes, line 63: this was missing, we ended up generating the wrong API url (we have a few of them in the DB)

app/services/github/client.rb
```diff
@@ -0,0 +1,98 @@
...
+
+      # adapted from https://api.rubyonrails.org/classes/Module.html#method-i-delegate_missing_to
+      def respond_to_missing?(method, _include_all = false)
+        target.respond_to?(method, false) || super
+      end
+
+      private
+
+      def request
+        Honeycomb.add_field("app.name", "github.client")
```
rhymes, line 31: @mstruve let me know if these are fine, same in `record_error`. I mirrored the same structure of `Search::Client`

app/services/github/client.rb
```diff
@@ -0,0 +1,98 @@
...
+          "github.errors",
+          tags: ["error:#{class_name}", "message:#{error_message}"],
+        )
+      end
+
+      def target
+        @target ||= Octokit::Client.new(
+          client_id: ApplicationConfig["GITHUB_KEY"],
+          client_secret: ApplicationConfig["GITHUB_SECRET"],
+          user_agent: "#{Octokit::Default::USER_AGENT} (#{URL.url})",
```
rhymes, line 66: being a good internet citizen, adding the URL to the user agent so they know it's us ;-)

app/services/github/client.rb
```diff
@@ -0,0 +1,98 @@
...
+        # <https://github.com/octokit/octokit.rb#caching>
+        # and <https://github.com/octokit/octokit.rb/blob/master/lib/octokit/default.rb>
+        Faraday::RackBuilder.new do |builder|
+          builder.use Faraday::HttpCache, store: Rails.cache, serializer: Marshal, shared_cache: false
+          builder.use Faraday::Request::Retry, exceptions: [Octokit::ServerError]
+          builder.use Octokit::Middleware::FollowRedirects
+          builder.use Octokit::Response::RaiseError
+          builder.use Octokit::Response::FeedParser
+
+          builder.adapter :patron
```
rhymes, line 84: switching from `Net::Http` to `patron`, the same we use for the `Search::Client` which is also based on Faraday

app/services/github/client.rb
```diff
@@ -0,0 +1,98 @@
...
+        end
+      end
+
+      def connection_options
+        Octokit.connection_options.merge(
+          request: {
+            open_timeout: 5,
+            timeout: 5
+          },
```
rhymes, line 93: there's no timeout by default, there are strong reasons to have default on HTTP clients as we aren't in control of the upstream server anyway (GH can go down or can take long to respond). Event Octokit [recommends adding them](https://github.com/octokit/octokit.rb#timeouts):

> By default, Octokit does not timeout network requests.
> You should set a timeout in order to avoid Ruby’s Timeout module, which can hose your server.

lib/data_update_scripts/20200503134234_remove_github_issues.rb
```diff
@@ -0,0 +1,7 @@
+module DataUpdateScripts
+  class RemoveGithubIssues
+    def run
+      GithubIssue.in_batches.delete_all
```
rhymes, line 4: this is to cleanup the status quo in the DB. 
We have a bunch of rows that have an invalid URL, which now that we fixed the URL will result in duplicates. This doesn't impact performance as we only use this model to render HTML in liquid tags, which then it's saved in the DB as processed HTML.

app/liquid_tags/github_tag/github_issue_tag.rb
```diff
@@ -46,9 +53,19 @@ def parse_link(link)
     end
 
     def generate_api_link(input)
-      input = input.gsub(/\?.*/, "")
-      input = input.gsub(/\d{1,}#issuecomment-/, "comments/") if input.include?("#issuecomment-")
-      "https://api.github.com/repos/#{input.gsub(/.*github\.com\//, '')}"
+      path = input.gsub(/\?.*/, "")
+      path = path.gsub(/.*github\.com\//, "")
```
citizen428, line 57: You can use the `%r{}` regex literals here so you don't have to escape the path slashes:

```suggestion
  path = path.gsub(%r{.*github\.com/}, "")
```

We can also leave off the `if`, since `gsub` doesn't do anything if the first argument doesn't match:

```ruby
"hello rhymes".gsub(%r{/pull/}, "pulls")
#=> "hello rhymes"
```

app/liquid_tags/github_tag/github_issue_tag.rb
```diff
@@ -46,9 +53,19 @@ def parse_link(link)
     end
 
     def generate_api_link(input)
-      input = input.gsub(/\?.*/, "")
-      input = input.gsub(/\d{1,}#issuecomment-/, "comments/") if input.include?("#issuecomment-")
-      "https://api.github.com/repos/#{input.gsub(/.*github\.com\//, '')}"
+      path = input.gsub(/\?.*/, "")
```
citizen428, line 56: Seeing this makes me think we should deal with an URI object here. Then we could interact with `query`, `fragment` etc. instead of doing this manually.

```ruby
uri = URI.parse(input)
uri.host = nil if uri.host == "github.com"
uri.path = uri.path.gsub(%r{/pull/}, "/pulls/")
uri.fragment = nil if uri.fragment.start_with?("#issue-")
# other transformation
uri.to_s
```

app/models/github_issue.rb
```diff
@@ -1,65 +1,78 @@
+# NOTE: we are using `GithubIssue` to store issues, pull requests and comments
 class GithubIssue < ApplicationRecord
+  CATEGORIES = %w[issue issue_comment].freeze
+  API_URL_REGEXP = /\Ahttps:\/\/api.github.com\/repos\/.*\z/.freeze
```
citizen428, line 4: This and most other constants here would benefit from the `%r{}` regex literals:

```suggestion
  API_URL_REGEXP = %r{\Ahttps://api.github.com/repos/.*\z}.freeze
```

app/services/github/client.rb
```diff
@@ -0,0 +1,98 @@
...
+        target.respond_to?(method, false) || super
+      end
+
+      private
+
+      def request
+        Honeycomb.add_field("app.name", "github.client")
+        yield
+      rescue Octokit::Error => e
+        class_name = e.class.name.demodulize
```
citizen428, line 34: This seems long enough to be replaced by `handle_errror(e)` or something.

lib/data_update_scripts/20200503134234_remove_github_issues.rb
citizen428: Hm, since you're using `delete_all` and not `destroy_all` I don't think we need `in_batches` here. Usually, you do that to keep memory consumption from object instantiation lower, but we don't instantiate anything here anyway.

app/liquid_tags/github_tag/github_issue_tag.rb
rhymes: make sense! I'll apply it!

app/liquid_tags/github_tag/github_issue_tag.rb
rhymes: Sure! I only thought it was faster as it doesn't have to scan the entire string, not that it matters being so short :D

app/models/github_issue.rb
rhymes: +1 I'll do it with the other ones

app/services/github/client.rb
rhymes: makes sense

lib/data_update_scripts/20200503134234_remove_github_issues.rb
rhymes: I actually copy-pasted from https://api.rubyonrails.org/v5.2/classes/ActiveRecord/Batches.html#method-i-in_batches but I think I'll keep it as it's not about loading the objects, but to issue `DELETE`s in batches instead of one large `DELETE`

app/liquid_tags/github_tag/github_issue_tag.rb
citizen428: It would scan the string anyway for the `if`, wouldn't it?

app/services/github/client.rb
mstruve: This is actually not going to work how you want it to. For example, with searching, I thought it would label the search span itself but that span is not created until after our HTTP client starts making the request much further down the stack so the only thing we end up adding a field to is the controller method or sidekiq worker span that is implementing the request. Its not hurting anything so I left it but we need to dig deep into how HTTP requests are tracked if we want to add to their spans.

app/liquid_tags/github_tag/github_issue_tag.rb
```diff
@@ -46,9 +53,29 @@ def parse_link(link)
     end
 
     def generate_api_link(input)
-      input = input.gsub(/\?.*/, "")
-      input = input.gsub(/\d{1,}#issuecomment-/, "comments/") if input.include?("#issuecomment-")
-      "https://api.github.com/repos/#{input.gsub(/.*github\.com\//, '')}"
+      uri = URI.parse(input).normalize
+      uri.host = nil if uri.host == "github.com"
+
+      # public PRs URLs are "/pull/{id}" but the API requires "/pulls/{id}"
```
mstruve, line 59: 😂 I bet there was some good facepalming when they realized this discrepancy

app/services/github/client.rb
```diff
@@ -0,0 +1,103 @@
...
+
+      def handle_error(exception)
+        class_name = exception.class.name.demodulize
+
+        # raise specific error if known, generic one if unknown
+        error_class = "::Github::Errors::#{class_name}".safe_constantize
+        raise error_class, exception.message if error_class
+
+        error_class = if exception.class < Octokit::ClientError
+                        Github::Errors::ClientError
```
mstruve, line 57: `Errors` feels redundant here, I think these could be 
```
Github::ClientError
Github::ServerError
Github::Error
```

app/views/liquids/_github_issue.html.erb
```diff
@@ -18,12 +18,12 @@
       <div class="timeline-comment-header-text">
         <strong>
           <a href="<%= user_html_url %>"><%= username %></a>
-        </strong> commented on <a href="<%= date_link %>"><%= date %></a><span class="timestamp"></span>
+        </strong> <%= tagline %> <a href="<%= html_url %>"><time datetime="<%= created_at %>"><%= date %></time></a>
```
mstruve, line 21: 🔥

app/liquid_tags/github_tag/github_issue_tag.rb
rhymes: It's even weirder as the actual API call was done on the right URL thanks to a following string replace, it was just the URL stored in the DB that is invalid. But it works now because we generate the wrong URL which matches with the wrong URL stored in the DB :D

app/services/github/client.rb
rhymes: true, but I followed the convention in `Search::`, not that it can't be changed, we just need to make a decision about this or ignore it until we make our minds up :D

mstruve: Maybe followup Refactor PR to remove redundant Error from everything 😄

rhymes: I just thought that there might be an advantage though to the namespacing: it's easier to do runtime reflection. Don't know if it's enough to justify it though ahhahaha

rhymes:
> Looks great.
> 
> (We also do the same thing with Twitter, for the same bad reasons)

hahaha yeah, next step is Twitter ;)

Issue #7495, opened 2020-04-23
Refactor JSON-LD Data Structures in the StoriesController


juliannatetreault:
Adding structured data to our site using JSON-LD was a major SEO optimization that was recently shipped and has since been iterated on. While it has proven to bolster our SEO, the code for the data structure could be refactored to be more readable for those unfamiliar with the work. In order to make the code more readable, refactoring the methods used to set specific JSON-LD values in the structure (`set_user_json_ld` and `set_user_same_as_json_ld`) to be closer to the implementation used for the `seo_optimized_images` method would be ideal. In addition to this, linking to JSON-LD documentation in the code could be useful for contributors and the team to remove any ambiguity around what the purpose of the code is.

Food for thought: generalizing the `seo_optimized_images` code even further to account for other articles potentially using the array. 🤔 

**✨ A note to our amazing contributors: this issue will be worked on by a DEV team member and is not up for grabs! ✨**

PR #7672, opened 2020-05-04
Refactor JSON-LD Within user_show_spec to Improve Testing

juliannatetreault:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
While the original implementation worked, this PR refactors the `user_show_spec.rb` to better test the JSON-LD data within the `Stories::Controller`. 

Rather than having multiple single-line expectations, the spec now tests against the JSON-LD in one test. In addition to being more concise, the spec now tests for more data and does so against the actual data structure by parsing the data after grabbing the `response.body` with Nokogiri.

These improvements should make refactoring the actual JSON-LD (see issue #7495 ) within the `Stories::Controller` easier and should overall make the `user_show_spec.rb` stronger.

Relates to #7495 - this is the initial refactor needed to ensure further refactoring is accurate

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Cats Cleaning

### DIFF
--- a/spec/requests/user/user_show_spec.rb
+++ b/spec/requests/user/user_show_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 RSpec.describe "UserShow", type: :request do
-  let_it_be(:user) { create(:user, email_public: true, education: "DEV University") }
+  let_it_be(:user) { create(:user, :with_all_info, email_public: true) }
   let(:doc) { Nokogiri::HTML(response.body) }
   let(:text) { doc.at('script[type="application/ld+json"]').text }
   let(:response_json) { JSON.parse(text) }
@@ -37,26 +37,6 @@ RSpec.describe "UserShow", type: :request do
 
   describe "GET /:slug (user)" do
     before do
-      user.update_columns(
-        employment_title: "SEO",
-        employer_name: "DEV",
-        employer_url: "www.dev.to",
-        currently_learning: "Preact",
-        mostly_work_with: "Ruby",
-        currently_hacking_on: "JSON-LD",
-        mastodon_url: "www.example.com",
-        facebook_url: "www.facebook.com/example",
-        linkedin_url: "www.linkedin.com/company/example/",
-        youtube_url: "www.youtube.com/example",
-        behance_url: "www.behance.com/example",
-        stackoverflow_url: "www.stackoverflow.com/example",
-        dribbble_url: "www.dribbble.com/example",
-        medium_url: "www.medium.com/example",
-        gitlab_url: "www.gitlab.com/example",
-        instagram_url: "www.instagram.com/example",
-        twitch_username: "Example007",
-        website_url: "www.example.com/example",
-      )
       get user.path
     end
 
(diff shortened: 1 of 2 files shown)


spec/requests/user/user_show_spec.rb
```diff
@@ -1,52 +1,71 @@
...
 
   describe "GET /:slug (user)" do
     before do
-      user.update_columns(employment_title: "SEO", employer_name: "DEV", linkedin_url: "www.linkedin.com/company/example/")
+      user.update_columns(employment_title: "SEO", employer_name: "DEV", employer_url: "www.dev.to")
+      user.update_columns(currently_learning: "Preact", mostly_work_with: "Ruby", currently_hacking_on: "JSON-LD")
+      user.update_columns(mastodon_url: "www.example.com", facebook_url: "www.facebook.com/example", linkedin_url: "www.linkedin.com/company/example/")
+      user.update_columns(youtube_url: "www.youtube.com/example", behance_url: "www.behance.com/example", stackoverflow_url: "www.stackoverflow.com/example")
+      user.update_columns(dribbble_url: "www.dribbble.com/example", medium_url: "www.medium.com/example", gitlab_url: "www.gitlab.com/example")
+      user.update_columns(instagram_url: "www.instagram.com/example", twitch_username: "Example007", website_url: "www.example.com/example")
```
Zhao-Andy, line 45: Could probably refactor this into calling it just once, and passing the arguments all at once? Like:
```
user.update_columns(
  employment_title: "SEO",
  employer_name: "DEV",
  # etc etc
)
```

juliannatetreault: Excellent suggestion! I just made this change and the spec looks so much nicer. 😂

maestromac: You could also expand user factory with a new transient.  `create(:user, :with_all_info)`  would be pretty clean

juliannatetreault: I like that idea as well--trying it out now! 😄

spec/requests/user/user_show_spec.rb
```diff
@@ -1,52 +1,71 @@
...
+
+  let(:disambiguating_description) do
+    [
+      user.mostly_work_with,
+      user.currently_hacking_on,
+      user.currently_learning,
+    ]
+  end
+
+  let(:works_for) { [{ "@type" => "Organization", "name" => user.employer_name, "url" => user.employer_url }] }
```
maestromac, line 36: Having these single-use `let` is a bad practice. You might be doing this because of rubocop rule for too many line. I suggest you disable that rule inline and keep these variable within the spec on line 53

juliannatetreault: That's good to know! I definitely do not like having it this way, but I was running into an `ExampleLength` Rubocop error, so I extracted it out to satisfy that. I'll make the change to disable it instead then. 😊

juliannatetreault:
Alright, I think that I've addressed everyone's concerns--let me know what you both think. now @Zhao-Andy and @maestromac 😄

PR #7674, opened 2020-05-04
Refactor Checking Rate Limits and Their Retry After settings

mstruve:
- [x] Refactor

## Description
This is the last rate limit refactor I am going to do for now. With this final refactor we have the ability to simply add a key to `ACTION_LIMITERS` and `CONFIGURABLE_RATES` and have all the functionality and methods we need to start tracking and limiting simple user actions. I also moved the checker from labor to services since that is what this is really morphing into. 

There is a lot more complicated refactoring that could be done here but I thought this gave us some good bang for our buck and I should be able to whip through some of those rate limit tickets we have now. 

## Added tests?
- [x] updated

alt_text

### DIFF
--- a/spec/services/rate_limit_checker_spec.rb
+++ b/spec/services/rate_limit_checker_spec.rb
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-RSpec.describe RateLimitChecker, type: :labor do
+RSpec.describe RateLimitChecker, type: :service do
   let(:user) { create(:user) }
   let(:article) { create(:article, user: user) }
   let(:rate_limit_checker) { described_class.new(user) }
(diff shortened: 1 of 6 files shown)


spec/services/rate_limit_checker_spec.rb
```diff
@@ -133,7 +133,7 @@
 
     it "raises an error if limit_by_action is true" do
       allow(rate_limit_checker).to receive(:limit_by_action).and_return(true)
-      expect { rate_limit_checker.check_limit!(:image_upload) }.to raise_error(RateLimitChecker::LimitReached)
+      expect { rate_limit_checker.check_limit!(:image_upload) }.to raise_error(described_class::LimitReached)
```
atsmith813, line 136: 🎉

PR #7675, opened 2020-05-04
Unescape user's summary in followUser slide

maestromac:
- [x] Bug Fix

## Description
I added a relatively lightweight dependency, [he](https://github.com/mathiasbynens/he), to unescape strings.
Resolves https://github.com/thepracticaldev/dev.to/issues/4650
Supersedes https://github.com/thepracticaldev/dev.to/pull/6772
n/a
## Added tests?

## Added to documentation?
- [x] no documentation needed


citizen428:
@maestromac Read my approval as "I approve of the change in principal, but you need to fix the spec it seems to be a legit error". 😉

app/javascript/onboarding/components/FollowUsers.jsx
```diff
@@ -168,7 +169,9 @@ class FollowUsers extends Component {
                 </figure>
                 <div className="user-info">
                   <h4 className="user-name">{user.name}</h4>
-                  <p className="user-summary">{user.summary}</p>
+                  <p className="user-summary">
+                    {he.unescape(user.summary || '')}
```
rhymes, line 173: is there a chance that the summary, unescaped, can be used for XSS ?
I'm thinking this because the summary can be edited by the user and the onboarding can be accessed again after a user is already registered.

maestromac: I've tried a few things and it seems to be safe. Do you have any other suggestion for me to test out?
image

PR #7676, opened 2020-05-04
Bulk Fetch Follow Button Data for Users

mstruve:
- [x] Optimization

## Description
Currently, we make individual requests to the server to figure out what to display on every single "Follow" button on a webpage. A better approach would be to make a single bulk request for all of them at once and then let the frontend loop through the response. That is exactly what this PR does. 

Suggestions for testing approaches are appreciated! And whether I can ignore the CodeClimate errors or not. I noticed other functions are used before they are defined so that to me seemed OK to let go.

https://github.com/thepracticaldev/dev.to/projects/9#card-36804382

## Added tests?
- [x] yes, I added a controller test but I did not add any JS tests. I need help/direction on what I should do for those. 

alt_text


app/assets/javascripts/initializers/initializeLocalStorageRender.js
```diff
@@ -8,6 +8,7 @@ function initializeLocalStorageRender() {
       initializeBaseUserData();
       initializeReadingListIcons();
       initializeAllFollowButts();
+      initializeUserFollowButts();
```
nickytonline, line 11: You can replace `'use strict';` with `/* global initializeUserFollowButts */`. The `'use strict';` is not required and the global comment indicates to eslint that `initializeUserFollowButts` is already defined somewhere else globally.  This should get rid of one of the code climate issues.

app/assets/javascripts/initializers/initializeAdditionalContentBoxes.js
```diff
@@ -32,6 +32,7 @@ function initializeAdditionalContentBoxes() {
             el.innerHTML = html;
             initializeReadingListIcons();
             initializeAllFollowButts();
+            initializeUserFollowButts();
```
nickytonline, line 35: You can replace `'use strict';` at the top of the file with `/* global initializeUserFollowButts */`. The `'use strict';` is not required and the global comment indicates to eslint that `initializeUserFollowButts` is already defined somewhere else globally. This should get rid of one of the code climate issues.

app/assets/javascripts/initializers/initializeAllFollowButts.js
```diff
@@ -1,10 +1,65 @@
...
   for (var i = 0; i < followButts.length; i++) {
-    initializeFollowButt(followButts[i]);
+    if (!followButts[i].className.includes("follow-user")) {
+      initializeFollowButt(followButts[i]);
+    };
   }
 }
 
+function fetchUserButtons(idButtonHash) {
+  var dataRequester;
```
nickytonline, line 11: Since this is new code, consider using `fetch`.

app/assets/javascripts/initializers/initializeAllFollowButts.js
```diff
@@ -1,10 +1,65 @@
...
   var followButts = document.getElementsByClassName('follow-action-button');
   for (var i = 0; i < followButts.length; i++) {
-    initializeFollowButt(followButts[i]);
+    if (!followButts[i].className.includes("follow-user")) {
+      initializeFollowButt(followButts[i]);
+    };
   }
 }
 
+function fetchUserButtons(idButtonHash) {
```
nickytonline, line 10: Consider a different name as this is really fetching whether or not a user is being followed.

nickytonline:
@mstruve, for testing, since these aren't Preact components or ES modules, I'd suggest a Capybara test.

benhalpern:
This looks great works great from what I can tell with manual testing...

The one thing that we'll need to change due to this implementation is the safe params vcl. I'm not _sure_ how it will handle `ids[]` from a characters perspective. We've had issues with this in the past but maybe us just being sloppy. We could add a conditional to whether the URL includes `/bulk_show` or whatever because we'll never need to cache this endpoint. Either way something like this.

mstruve:
@benhalpern good call on Fastly! 

I will get to work on that and @nickytonline thanks for the direction on testing! I will get some feature tests written up!

app/services/search/user.rb
```diff
@@ -25,7 +25,8 @@ def prepare_doc(hit)
           "comments_count" => source["comments_count"],
           "badge_achievements_count" => source["badge_achievements_count"],
           "last_comment_at" => source["last_comment_at"],
-          "roles" => source["roles"]
+          "roles" => source["roles"],
+          "user_id" => source["id"]
```
mstruve, line 29: This was a minor bug fix, I noticed when inspecting the frontend that users did not have data-content-user-id set bc it was calling `user_id` on the Elasticsearch hash which was missing for the user so I added it. 
```
data-content-user-id="'+article.user_id+'
```

app/assets/javascripts/initializers/initializeAllFollowButts.js
```diff
@@ -1,10 +1,54 @@
...
   }
 }
 
+function fetchUserFollowStatuses(idButtonHash) {
+  var idString = "";
+  Object.keys(idButtonHash).forEach(function(id) {
+    idString += ("ids[]=" + id + "&");
+  });
+
+  fetch(`/follows/bulk_show?${idString}followable_type=User`, {
```
rhymes, line 16: this could be 

```js
const url = new URL("/follows/bulk_show", document.location);
const searchParams = new URLSearchParams();
Object.keys(idButtonHash).forEach((id) => {
  searchParams.append("ids[]", id);
});
searchParams.append("followable_type", "User");
url.search = searchParams;

fetch(url, { // etc
```

it's bit more lines but I think it's cleaner and uses less string interpolation.

app/assets/javascripts/initializers/initializeAllFollowButts.js
```diff
@@ -1,10 +1,54 @@
...
+  }).then((response) => response.json())
+    .then((idStatuses) => {
+      Object.keys(idStatuses).forEach(function(id) {
+        addButtClickHandle(idStatuses[id], idButtonHash[id]);
+      })
+    });
+}
+
+function initializeUserFollowButtons(buttons) {
+  var userIds = {};
```
rhymes, line 33: maybe we can just exit this function if `buttons.length` is 0, to avoid the check at line 44

app/assets/javascripts/initializers/initializeAllFollowButts.js
```diff
@@ -1,10 +1,54 @@
...
+  for (var i = 0; i < buttons.length; i++) {
+    var userStatus = document.getElementsByTagName('body')[0].getAttribute('data-user-status');
```
rhymes, line 35: `document.body.getAttribute()` is shorter and faster

app/assets/javascripts/initializers/initializeAllFollowButts.js
```diff
@@ -1,10 +1,54 @@
...
+    .then((idStatuses) => {
+      Object.keys(idStatuses).forEach(function(id) {
+        addButtClickHandle(idStatuses[id], idButtonHash[id]);
+      })
+    });
+}
+
+function initializeUserFollowButtons(buttons) {
+  var userIds = {};
+  for (var i = 0; i < buttons.length; i++) {
```
rhymes, line 34: not necessary but a tip: `Array.from(buttons)` lets you iterate using `forEach` as well, but the for is likely faster here

app/controllers/follows_controller.rb
```diff
@@ -25,6 +25,28 @@ def show
     end
   end
 
+  def bulk_show
```
rhymes, line 28: as this is a new endpoint unauthorized, we should probably rate limit it :-D

app/controllers/follows_controller.rb
```diff
@@ -25,6 +25,28 @@ def show
...
+    skip_authorization
+
+    response = params[:ids].map(&:to_i).each_with_object({}) do |id, hsh|
```
rhymes, line 31: this results in a 500 error if the client has no ids, we should use the classic `params.require(:ids)` check, so the client gets a 400 bad request.

Also:

```ruby
> "a".to_i
0
```

which means that it does stuff with user id that don't exist.

app/controllers/follows_controller.rb
```diff
@@ -25,6 +25,28 @@ def show
...
+      hsh[id] = if current_user.id == id
```
rhymes, line 32: this breaks if I'm not logged in, I tried with:

```shell
curl -i "localhost:3000/follows/bulk_show?ids[]=1"
```

from the shell

app/controllers/follows_controller.rb
```diff
@@ -25,6 +25,28 @@ def show
...
+                  "self"
+                else
+                  following_them_check = FollowChecker.new(current_user, params[:followable_type], id).cached_follow_check
+                  following_you_check = FollowChecker.new(User.find_by(id: id), params[:followable_type], current_user.id).cached_follow_check
```
rhymes, line 36: this again will fail if `current_user` is not there

rhymes, line 36: I also think we should load users in bulk. Right now we do a "select * from users" for each user that's not the current user sent by the JS client, could we load all of them in one query?

Also, it seems to me that `FollowChecker` only requires a few columns from user to function, so we don't need to load the entire table, that could also be an optimization :)

config/fastly/snippets/safe_params_list.vcl
```diff
@@ -1,7 +1,7 @@
 import querystring;
 sub vcl_recv {
     # return this URL with only the parameters that match this regular expression
-    if (req.url !~ "/internal/" && req.url !~ "/search/") {
+    if (req.url !~ "/internal/" && req.url !~ "/search/" && req.url !~ "/bulk_show") {
```
rhymes, line 4: should this be the entire path `/follows/bulk_show` ? Just in case we add one that shouldn't be considered. 

ps. what if we whitelist `ids` instead?

app/controllers/follows_controller.rb
rhymes: Also we might want to avoid triggering behavior for user ids that don't exist as well

app/controllers/follows_controller.rb
mstruve: I think those are great optimizations to make but I would rather make them in additional PRs rather than changing a ton of behavior in this PR. The goal of this PR is to decrease the hits to the server from the view so we can tighten up our general site throttling which alone is a win. Beyond that I am inclined to use separate PRs to get further optimizations.

config/fastly/snippets/safe_params_list.vcl
mstruve: My assumption was that in any bulk_show path we are going to be using `ids[]` or some sort of array param in order to send multiple items so this will cover those situations nicely. 

I have not been able to figure out how to successfully ignore the array syntax for params. I tried a lot of different things for search and none of them worked which is why search ended up on this list.

app/controllers/follows_controller.rb
mstruve: Interesting works incognito for me.  I'll look into it, we might just not call it when incognito.

app/assets/javascripts/initializers/initializeAllFollowButts.js
mstruve: We still need the 44 check bc we might add the `addModalEventListener(buttons[i]);` to buttons if the user is logged out.

app/controllers/follows_controller.rb
mstruve: Since I added the guard clause for current_user like the action above I am less worried about this endpoint now. I also expect that we will be bringing down our overall throttle once this is merged which will help a lot.

PR #7680, opened 2020-05-05
Change string concatenation to interpolation

citizen428:

- [X] Refactor

## Description

String interpolation is often preferred over concatenation in Ruby. While the latter may be faster for two strings, it does generate more intermediate string objects when concatenating more strings. Interpolation also calls `to_s` automatically, which simplifies the resulting code.

This is what I do for fun. I'm also great at parties.

## Added tests?

## Added to documentation?

- [X] no documentation needed

### DIFF
--- a/app/controllers/concerns/valid_request.rb
+++ b/app/controllers/concerns/valid_request.rb
@@ -22,12 +22,12 @@ module ValidRequest
     case options
     # Yet another monkeypatch required to send proper protocol out.
     # In this case we make sure the redirect ends in the app protocol.
-    # This is the same as the base Rails method except ApplicationConfig["APP_PROTOCOL"]
+    # This is the same as the base Rails method except URL.protocol
     # is used instead of request.protocol.
     when /\A([a-z][a-z\d\-+\.]*:|\/\/).*/i
       options
     when String
-      (URL.protocol || request.protocol) + request.host_with_port + options
+      "#{(URL.protocol || request.protocol)}#{request.host_with_port}#{options}"
     when Proc
       _compute_redirect_to_location request, instance_eval(&options)
     else
--- a/app/controllers/concerns/valid_request.rb
+++ b/app/controllers/concerns/valid_request.rb
@@ -10,7 +10,7 @@ module ValidRequest
     return if Rails.env.test?
 
     if request.referer.present?
-      request.referer.start_with?(ApplicationConfig["APP_PROTOCOL"].to_s + ApplicationConfig["APP_DOMAIN"].to_s)
+      request.referer.start_with?(URL.url)
     else
       raise ::ActionController::InvalidAuthenticityToken, ::ApplicationController::NULL_ORIGIN_MESSAGE if request.origin == "null"
 
@@ -27,7 +27,7 @@ module ValidRequest
     when /\A([a-z][a-z\d\-+\.]*:|\/\/).*/i
       options
     when String
-      (ApplicationConfig["APP_PROTOCOL"] || request.protocol) + request.host_with_port + options
+      (URL.protocol || request.protocol) + request.host_with_port + options
     when Proc
       _compute_redirect_to_location request, instance_eval(&options)
     else
--- a/app/liquid_tags/organization_tag.rb
+++ b/app/liquid_tags/organization_tag.rb
@@ -13,7 +13,7 @@ class OrganizationTag < LiquidTagBase
     ActionController::Base.new.render_to_string(
       partial: PARTIAL,
       locals: {
-        organization: @organization,
+        organization: @organization.decorate,
         follow_button: @follow_button,
         organization_colors: @organization_colors
       },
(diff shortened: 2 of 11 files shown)


citizen428:
I wasn't done yet @rhymes, just checked in a draft PR. I'll re-request another review later.

app/controllers/concerns/valid_request.rb
```diff
@@ -10,7 +10,7 @@ def valid_request_origin?
     return if Rails.env.test?
 
     if request.referer.present?
-      request.referer.start_with?(ApplicationConfig["APP_PROTOCOL"].to_s + ApplicationConfig["APP_DOMAIN"].to_s)
+      request.referer.start_with?(URL.url)
```
citizen428, line 13: Instead of just changing from concatenation I changed this to use the helper.

app/controllers/concerns/valid_request.rb
```diff
@@ -27,7 +27,7 @@ def _compute_redirect_to_location(request, options) #:nodoc:
     when /\A([a-z][a-z\d\-+\.]*:|\/\/).*/i
       options
     when String
-      (ApplicationConfig["APP_PROTOCOL"] || request.protocol) + request.host_with_port + options
+      (URL.protocol || request.protocol) + request.host_with_port + options
```
citizen428: Switched to use of the `URL` module here too.

app/liquid_tags/organization_tag.rb
```diff
@@ -13,7 +13,7 @@ def render(_context)
     ActionController::Base.new.render_to_string(
       partial: PARTIAL,
       locals: {
-        organization: @organization,
+        organization: @organization.decorate,
```
citizen428, line 16: Added `decorate` call here to remove it from the model where it was used more than once.

app/liquid_tags/user_tag.rb
```diff
@@ -13,7 +13,7 @@ def render(_context)
     ActionController::Base.new.render_to_string(
       partial: PARTIAL,
       locals: {
-        user: @user,
+        user: @user.decorate,
```
citizen428, line 16: Added `decorate` call here to remove it from the model where it was used more than once.

app/models/user.rb
```diff
@@ -577,7 +577,7 @@ def validate_mastodon_url
   end
 
   def tag_keywords_for_search
-    employer_name.to_s + mostly_work_with.to_s + available_for.to_s
+    "#{employer_name}#{mostly_work_with}#{available_for}"
```
citizen428, line 580: Neiter @maestromac nor @benhalpern  really knew what this method is for. It's only used in `search_core`, which is then wrapped one more time in `hotness_score`.

```ruby
  def hotness_score
    search_score
  end

  def search_score
    counts_score = (articles_count + comments_count + reactions_count + badge_achievements_count) * 10
    score = (counts_score + tag_keywords_for_search.size) * reputation_modifier
    score.to_i
  end
```

Given that `tag_keywords_for_search` just concatenates 3 strings into one, I think the idea was to give new users (where `counts_score` will be 0) with filled in profiles a higher score than people with no profile information.

spec/requests/stories_show_spec.rb
```diff
@@ -115,13 +115,13 @@
     it "renders canonical url when exists" do
       article = create(:article, with_canonical_url: true)
       get article.path
-      expect(response.body).to include('"canonical" href="' + article.canonical_url.to_s + '"')
+      expect(response.body).to include(%("canonical" href="#{article.canonical_url}"))
```
citizen428, line 118: The `%()` string literals allow interpolation without needing to escape double quote inside them.

spec/requests/stories_show_spec.rb
```diff
@@ -115,13 +115,13 @@
...
     end
 
     it "does not render canonical url when not on article model" do
       article = create(:article, with_canonical_url: false)
       get article.path
-      expect(response.body).not_to include('"canonical" href="' + article.canonical_url.to_s + '"')
+      expect(response.body).not_to include(%("canonical" href="#{article.canonical_url}"))
```
citizen428, line 124: The `%()` string literals allow interpolation without needing to escape double quote inside them.

citizen428:
I think we also have quite a few unnecessary `to_s` calls sprinkled through our codebase, e.g.

```ruby
<%= field.to_s %>
```

because `<%= %>` call `to_s` anyway:

```ruby
x = nil
ERB.new("x: <%= x %>").result(binding) == ERB.new("x: <%= x.to_s %>").result(binding)
#=> true
```

But I'll leave that for another PR.

citizen428:
I'm not entirely sure why CodeClimate thinks there were changes in files I didn't touch 🤷

rhymes:
> I'm not entirely sure why CodeClimate thinks there were changes in files I didn't touch 🤷

classic CC :D

spec/requests/stories_show_spec.rb
msarit: Learned something new! 💃

PR #7681, opened 2020-05-05
Color Updates

ludwiczakpawel:

- [x] Feature
- [x] Optimization

## Description

@lisasy and myself both have been noticing that we may benefit from having additional shades of certain colors. Right now, we have several shades for base colors and two additional shades for each accent color. But sometimes we need extra shades that are not solid colors - instead they are generated by setting certain opacity.

For that reason, this PR brings more variables for `base`:

```
--base-a90: #{rgba(#08090a, 0.9)};
...
--base-a5: #{rgba(#08090a, 0.05)};
```

And one (for now) extra variable for accents:

```
--accent-brand-a10: #{rgba(#3b49df, 0.1)};
--accent-success-a10: #{rgba(#26d9ca, 0.1)};
--accent-warning-a10: #{rgba(#ffcf4c, 0.1)};
--accent-danger-a10: #{rgba(#dc1818, 0.1)};
```

## Added tests?

## Added to documentation?

- [x] no documentation needed

### DIFF
--- a/app/assets/stylesheets/config/_colors.scss
+++ b/app/assets/stylesheets/config/_colors.scss
@@ -13,7 +13,20 @@
   ////////////////////////////////////////////////////////////////////////////////////
 
   // Base
-  --base-100: #08090a;
+  --base: #08090a;
+
+  --base-a90: #{rgba(#08090a, 0.9)};
+  --base-a80: #{rgba(#08090a, 0.8)};
+  --base-a70: #{rgba(#08090a, 0.7)};
+  --base-a60: #{rgba(#08090a, 0.6)};
+  --base-a50: #{rgba(#08090a, 0.5)};
+  --base-a40: #{rgba(#08090a, 0.4)};
+  --base-a30: #{rgba(#08090a, 0.3)};
+  --base-a20: #{rgba(#08090a, 0.2)};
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/config/_colors.scss
```diff
@@ -13,7 +13,20 @@
...
+  --base-a90: #{rgba(#08090a, 0.9)};
+  --base-a80: #{rgba(#08090a, 0.8)};
+  --base-a70: #{rgba(#08090a, 0.7)};
+  --base-a60: #{rgba(#08090a, 0.6)};
+  --base-a50: #{rgba(#08090a, 0.5)};
+  --base-a40: #{rgba(#08090a, 0.4)};
+  --base-a30: #{rgba(#08090a, 0.3)};
+  --base-a20: #{rgba(#08090a, 0.2)};
+  --base-a10: #{rgba(#08090a, 0.1)};
+  --base-a5: #{rgba(#08090a, 0.05)};
```
lisasy, line 27: If we wanted for the names of these variables to be more "even", this one could be:

`--base-a05` so that they all have 2 digits. Otherwise, I wonder if there might be some mix-ups between `--base-a50` and `--base-a5` because one ends in a `0` and the other doesn't.

PR #7683, opened 2020-05-05
Limit Reactions Created in a 30 second Timeframe

mstruve:
- [x] Feature
- [x] Optimization

## Description
Limit the number of reactions created in a 30 second window, default to 10, and allow it to be configurable. 

## Added tests?
- [x] yes
- [ ] no, because I need help

alt_text

### DIFF
--- a/app/controllers/reactions_controller.rb
+++ b/app/controllers/reactions_controller.rb
@@ -1,5 +1,6 @@
 class ReactionsController < ApplicationController
   before_action :set_cache_control_headers, only: [:index], unless: -> { current_user }
+  before_action :check_limit, only: [:create]
   after_action :verify_authorized
 
   def index
@@ -73,6 +74,7 @@ class ReactionsController < ApplicationController
       reaction = build_reaction(category)
 
       if reaction.save
+        rate_limiter.track_limit_by_action(:reaction_creation)
         Moderator::SinkArticles.call(reaction.reactable_id) if reaction.vomit_on_user?
 
         Notification.send_reaction_notification(reaction, reaction.target_user)
@@ -129,5 +131,9 @@ class ReactionsController < ApplicationController
                       user_id: current_user.id,
                       context: "readinglist_reaction",
                       rating: current_user.experience_level)
+  end
+
+  def check_limit
+    rate_limit!(:reaction_creation)
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/reactions_controller.rb
```diff
@@ -1,5 +1,6 @@
 class ReactionsController < ApplicationController
   before_action :set_cache_control_headers, only: [:index], unless: -> { current_user }
+  before_action :check_limit, only: [:create]
```
citizen428, line 3: If you want to save yourself from always having to define an extra method:

```ruby
before_action(only: [:create]) { rate_limit!(:reaction_creation) }
```

I'm not entirely convinced that the `before_action`s really are necessary for most of these cases. They cancel the action if the callback renders or redirects and also cancel additional callbacks that might otherwise run. Our action here does either nothing or raises, so the result is pretty much the same.

But then I think callbacks are generally overused in Rails, especially when they are only applied to single actions.

app/services/rate_limit_checker.rb
```diff
@@ -15,7 +16,8 @@ class RateLimitChecker
     rate_limit_published_article_creation: { min: 0, placeholder: 9, description: "The number of articles a user can create within 30 seconds" },
     rate_limit_image_upload: { min: 0, placeholder: 9, description: "The number of images a user can upload within 30 seconds" },
     rate_limit_email_recipient: { min: 0, placeholder: 5, description: "The number of emails we send to a user within 2 minutes" },
-    rate_limit_organization_creation: { min: 1, placeholder: 1, description: "The number of organizations a user can create within a 5 minute period" }
+    rate_limit_organization_creation: { min: 1, placeholder: 1, description: "The number of organizations a user can create within a 5 minute period" },
```
citizen428, line 19: This constant only seems to be used in `app/views/internal/configs/show.html.erb`, should we maybe move it to a view helper?

app/controllers/reactions_controller.rb
mstruve: I had methods before and it was suggested I put them in before_actions 🤷 6/half dozen the other I guess.

app/services/rate_limit_checker.rb
mstruve: That would be a good idea going forward

PR #7685, opened 2020-05-05
Dont Allow Multiple User Destroy Requests

mstruve:
- [x] Optimization

## Description
Don't allow users to request destroy multiple times. 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -93,7 +93,11 @@ class UsersController < ApplicationController
 
   def request_destroy
     set_tabs("account")
-    if @user.email?
+
+    if destroy_request_in_progress
+      flash[:settings_notice] = "You have already requested account deletion. Please, check your email for further instructions."
+      redirect_to "/settings/#{@tab}"
+    elsif @user.email?
       Users::RequestDestroy.call(@user)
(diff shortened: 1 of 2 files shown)


app/controllers/users_controller.rb
```diff
@@ -93,7 +93,11 @@ def update_language_settings
 
   def request_destroy
     set_tabs("account")
-    if @user.email?
+
+    if destroy_request_in_progress
+      flash[:settings_notice] = "You have already requested account deletion. Please, check your email for further instructions."
+      redirect_to "/settings/#{@tab}"
```
citizen428, line 99: Lately, we're making more an effort to switch to path helpers. Here it'd be 

```ruby
redirect_to user_settings_path(@tab)
```

if you want to update it (and the rest of the file/method). It's fine if you don't, I'll eventually get to it.

app/controllers/users_controller.rb
```diff
@@ -93,7 +93,11 @@ def update_language_settings
 
   def request_destroy
     set_tabs("account")
-    if @user.email?
+
+    if destroy_request_in_progress
```
citizen428, line 97: Maybe we want to make this a predicate method too, `destroy_request_in_progress?`?

PR #7688, opened 2020-05-05
Force Unique Jobs for User Initiated Jobs and Cache Jobs

mstruve:
- [x] Optimization

## Description
This PR introduces unique constraints to all of our CacheBustWorkers and to workers that are kicked off by user actions. These will help ensure that we are not doing duplicate unnecessary work. 

I bumped the default retry for CacheBustWorkers to 15 since 10 gives us 4 hours but I could see a service possibly having issues longer than that so I increased it. 

## Added tests?

alt_text

### DIFF
--- a/app/workers/bust_cache_base_worker.rb
+++ b/app/workers/bust_cache_base_worker.rb
@@ -1,0 +1,5 @@
+class BustCacheBaseWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :high_priority, retry: 15, lock: :until_executing
+end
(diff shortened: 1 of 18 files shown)


app/workers/bust_cache_base_worker.rb
```diff
@@ -0,0 +1,5 @@
+class BustCacheBaseWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :high_priority, retry: 15, lock: :until_executing
```
mstruve, line 4: These can be overridden in the individual workers if we want.

PR #7692, opened 2020-05-06
Introduce Rack::HeadersFilter

citizen428:

- [X] Feature

## Description

This PR came out of the discussion in a security issue. The code is copied from https://github.com/pusher/rack-headers_filter (with minor updates to follow our Rubocop config), as the middleware itself is trivial and I and @rhymes both thought that it'd be best to fully control this.

Issue 160 in the private repo.

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [X] no documentation needed

### DIFF
--- a/lib/rack/headers_filter.rb
+++ b/lib/rack/headers_filter.rb
@@ -1,3 +1,4 @@
+# Taken from https://github.com/pusher/rack-headers_filter
 module Rack
   class HeadersFilter
     SENSITIVE_HEADERS = %w[
--- a/lib/rack/headers_filter.rb
+++ b/lib/rack/headers_filter.rb
@@ -1,0 +1,38 @@
+module Rack
+  class HeadersFilter
+    SENSITIVE_HEADERS = %w[
+      HTTP_X_FORWARDED_FOR
+      HTTP_X_FORWARDED_HOST
+      HTTP_X_FORWARDED_PORT
+      HTTP_X_FORWARDED_PROTO
+      HTTP_X_FORWARDED_SCHEME
+      HTTP_X_FORWARDED_SSL
+    ].freeze
+
+    # Headers sent by the Heroku router
+    HEROKU_HEADERS = %w[
+      HTTP_CONNECTION
+      HTTP_CONNECT_TIME
+      HTTP_HOST
+      HTTP_TOTAL_ROUTE_TIME
+      HTTP_UPGRADE_INSECURE_REQUESTS
+      HTTP_VIA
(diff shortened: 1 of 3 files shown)


lib/rack/headers_filter.rb
```diff
@@ -0,0 +1,38 @@
...
+      HTTP_X_FORWARDED_FOR
+      HTTP_X_FORWARDED_HOST
+      HTTP_X_FORWARDED_PORT
+      HTTP_X_FORWARDED_PROTO
+      HTTP_X_FORWARDED_SCHEME
+      HTTP_X_FORWARDED_SSL
+    ].freeze
+
+    # Headers sent by the Heroku router
+    HEROKU_HEADERS = %w[
```
citizen428, line 13: Are there any other headers we want/need to trust, e.g. coming from Fastly?

We may eventually also need to revisit this in the context of generalization where Heroku may not always be used.

spec/lib/rack/headers_filter_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    header "Host", "myhost.com"
+    header "X-Forwarded-Host", "fake.com"
+    get "/"
+
+    expect(recording_app.last_env["HTTP_HOST"]).to eq("myhost.com")
+    expect(recording_app.last_env["HTTP_X_FORWARDED_HOST"]).to be_nil
+  end
+
+  it "lets other headers through" do
+    header "X-Funions", "yum"
```
rhymes, line 28: 🤣

lib/rack/headers_filter.rb
```diff
@@ -0,0 +1,38 @@
+module Rack
```
rhymes, line 1: maybe we can nod to the gem in a comment here

citizen428: Yup, we probably should, good call!

citizen428:
@mstruve the current set of headers is what the original gem used, there are other things we may want to add here. Also, in terms of generalization, not everyone will be using Fastly.

mstruve:
> Also, in terms of generalization, not everyone will be using Fastly.

For now, unfortunately, it's a requirement so I think we can punt on doing this until we overhaul how we use our CDN.

citizen428:
@mstruve Sure, we can either close this PR for now, or we just make the setting in `production.rb` conditional by checking on some env variable.

mstruve:
I lean towards closing in favor of less complexity as we figure out our CDN

PR #7696, opened 2020-05-06
Adjust GitHub and Twitter Connect Copy for Broadcasts

### LINKED ISSUE #431: Add tracking for live events
### User Story / Description
Ever since we launched dev live, we haven't had an accurate way to tell how engaged people are with our live workshops. We can see how many people visit, but no sense of how long they stick around unless they're actively chatting. This info would help us determine which types of workshops are more interesting to the community. 

### Definition of Done
A google analytics ping on /live would be a good start.

juliannatetreault:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR adjusts the copy for both the GitHub and Twitter connect notifications.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

I will need to update the production Broadcasts to include the new copy.

Tape

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -238,7 +238,7 @@ Rails.logger.info "#{counter}. Creating Broadcasts and Welcome Thread"
 broadcast_messages = {
   set_up_profile: "Welcome to DEV! 👋 I'm Sloan, the community mascot and I'm here to help get you started. Let's begin by <a href='/settings'>setting up your profile</a>!",
   welcome_thread: "Sloan here again! 👋 DEV is a friendly community. Why not introduce yourself by leaving a comment in <a href='/welcome'>the welcome thread</a>!",
-  twitter_connect: "You’re on a roll! 🎉 Do you have a Twitter account? Consider <a href='/settings'>connecting it</a> so we can @mention you if we share your post via our Twitter account <a href='https://twitter.com/thePracticalDev'>@thePracticalDev</a>.",
+  twitter_connect: "You're on a roll! 🎉 Do you have a Twitter account? Consider <a href='/settings'>connecting it</a> so we can @mention you if we share your post via our Twitter account <a href='https://twitter.com/thePracticalDev'>@thePracticalDev</a>.",
   github_connect: "You're on a roll! 🎉  Do you have a GitHub account? Consider <a href='/settings'>connecting it</a> so you can pin any of your repos to your profile.",
   customize_feed: "Hi, it's me again! 👋 Now that you're a part of the DEV community, let's focus on personalizing your content. You can start by <a href='/tags'>following some tags</a> to help customize your feed! 🎉",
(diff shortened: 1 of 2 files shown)


db/seeds.rb
```diff
@@ -238,8 +238,8 @@
 broadcast_messages = {
   set_up_profile: "Welcome to DEV! 👋 I'm Sloan, the community mascot and I'm here to help get you started. Let's begin by <a href='/settings'>setting up your profile</a>!",
   welcome_thread: "Sloan here again! 👋 DEV is a friendly community. Why not introduce yourself by leaving a comment in <a href='/welcome'>the welcome thread</a>!",
-  twitter_connect: "You're on a roll! 🎉 Let's connect your <a href='/settings'> Twitter account</a> to complete your identity so that we don't think you're a robot. 🤖",
-  github_connect: "You're on a roll! 🎉 Let's connect your <a href='/settings'> GitHub account</a> to complete your identity so that we don't think you're a robot. 🤖",
+  twitter_connect: "You’re on a roll! 🎉 Do you have a Twitter account? Consider <a href=‘/settings’>connecting it</a> so we can @mention you if we share your post via our Twitter account <a href='https://twitter.com/thePracticalDev'>@thePracticalDev</a>.",
```
rhymes, line 241: I think we need to fix the quotes in the `<a>` tag around the settings link from `‘` to `'`

db/seeds.rb
```diff
@@ -238,8 +238,8 @@
...
+  github_connect: "You're on a roll! 🎉  Do you have a GitHub account? Consider <a href=‘/settings’>connecting it</a> so you can pin any of your repos to your profile.",
```
rhymes, line 242: same thing about the quotes, the one in the factories as well

db/seeds.rb
juliannatetreault: That's so strange... just made the change!

juliannatetreault:
> The apostraphe you are using (`’`) will break html. It has to be the standard `'`.
> 
> This is what it will generate
> image

I believe I changed all of the apostrophes in my last commit, but let me know if I'm missing something @maestromac !

db/seeds.rb
vaidehijoshi: Oh dang, thanks for catching that @rhymes! I agree with you. I missed that when reviewing on mobile.

vaidehijoshi: @juliannatetreault sometimes if you copy-paste from other sites, the quotation marks end up actually being different characters. So that might be the explanation here. Generally you probably want to actually type out the ‘ character so that you accidentally don’t get a different one!

juliannatetreault: @vaidehijoshi That's exactly what happened... copy-pasta gone wrong. 🤦‍♀️

juliannatetreault:
Would someone with merging abilities please merge this for me whenever they get a chance? 😊  I'm unable to do since it still shows that there is one change request pending. 😕

PR #7699, opened 2020-05-06
Mod Actions Panel: Prevent nonsensical reaction pairs

msarit:

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
In production, we are able to select **both** a 'thumbs-down' and a 'vomit' reaction. Now that we have a 'thumbs-up' reaction, it does not make sense to be able to select both 'thumbs-up' and 'thumbs-down' reactions (or 'thumbs-up' and 'vomit' reactions). This PR prevents the selection of these nonsensical reaction pairs.

None

## Added tests?

- [ ] yes
- [ ] no, because I need help
- [X] no, not yet

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [X] no documentation needed

None

### DIFF
--- a/app/controllers/reactions_controller.rb
+++ b/app/controllers/reactions_controller.rb
@@ -49,9 +49,17 @@ class ReactionsController < ApplicationController
 
   def create
     authorize Reaction
+
     Rails.cache.delete "count_for_reactable-#{params[:reactable_type]}-#{params[:reactable_id]}"
 
-    clear_article_reactions(params[:reactable_id], params[:reactable_type], current_user, params[:category]) if params[:reactable_type] == "Article"
+    if params[:reactable_type] == "Article"
+      clear_article_reactions(
+        params[:reactable_id],
+        params[:reactable_type],
+        current_user,
+        params[:category],
+      )
+    end
 
     category = params[:category] || "like"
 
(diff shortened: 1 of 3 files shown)


app/controllers/reactions_controller.rb
```diff
@@ -130,4 +140,13 @@ def rate_article(reaction)
                       context: "readinglist_reaction",
                       rating: current_user.experience_level)
   end
+
+  def clear_article_reactions(id, type, mod, category)
+    reactions = if category == "thumbsup"
+                  Reaction.where(reactable_id: id, reactable_type: type, user: mod).where.not(category: category)
+                elsif %w[thumbsdown vomit].include?(category)
```
citizen428, line 147: We often follow a pattern of assigning things like this array to constants. Otherwise, we have to re-create the array and two strings on each method invocation. For example:

```ruby
NONSENSICAL_PAIRS = %w[thumbsdown vomit].freeze
...
elsif category.in?(NONSENSICAL_PAIRS)
```

app/controllers/reactions_controller.rb
```diff
@@ -52,7 +52,17 @@ def create
 
     Rails.cache.delete "count_for_reactable-#{params[:reactable_type]}-#{params[:reactable_id]}"
 
+    if params[:reactable_type] == "Article"
```
citizen428, line 55: I feel like this whole action has gotten uncomfortably big. Since you're here already, do you maybe want to see if we can refactor things a bit?

msarit: I agree enthusiastically but I'm wary of doing it alone 😅

msarit: Okay I abstracted the code that handles an existing reaction - hope this works?

citizen428: @msarit just trust yourself 😃

PR #7701, opened 2020-05-06
Generalize internal logo

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Switch from rainbowdev to the sticker as an alt internal logo.

Screen Shot 2020-05-06 at 3 30 55 PM

### DIFF
--- a/app/views/layouts/internal.html.erb
+++ b/app/views/layouts/internal.html.erb
@@ -28,7 +28,7 @@
   <div class="d-flex">
     <nav class="sidebar-nav">
       <a href="/internal">
-        <img class="d-block mx-auto mt-3" src="<%= asset_path("rainbowdev.svg") %>" width="40" height="40" alt="DEV Logo">
+        <img class="d-block mx-auto mt-3" src="<%= cloudinary(SiteConfig.primary_sticker_image_url, 190) %>" width="40" height="40" alt="DEV Logo">
       </a>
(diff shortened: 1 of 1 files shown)


app/views/layouts/internal.html.erb
```diff
@@ -28,7 +28,7 @@
   <div class="d-flex">
     <nav class="sidebar-nav">
       <a href="/internal">
-        <img class="d-block mx-auto mt-3" src="<%= asset_path("rainbowdev.svg") %>" width="40" height="40" alt="DEV Logo">
+        <img class="d-block mx-auto mt-3" src="<%= cloudinary(SiteConfig.primary_sticker_image_url, 190) %>" width="40" height="40" alt="DEV Logo">
```
rhymes, line 31: if it's restricted to 40 pixel by the `width` and `height` attributes, should we simplu fetch 40 pixels  from cloudinary? Or is it changed by CSS?

benhalpern: That's a good question but I'd say we probably want to fetch images we may have fetched elsewhere, all else equal. 190 happens to be the width we fetched in the other place we fetch it so it's most likely to be cached by browser...

This is a fairly trivial thing so I'd say we just ship this and keep this in mind as for ways we can insure system-wide there are some sensible ways we do this.

PR #7707, opened 2020-05-07
Decouple GithubReadmeTag liquid tag from GitHub auth

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

On the heels of #7660 this PR decouples `GithubReadmeTag` from using a random user token and uses the newly added `Github::Client`.

Next step is revisiting the logic in `GithubRepo` and `GithubReposController` and make their usage optional and enabled only for sites that require Github authorization, including the "display github repo on your profile" feature

#7660

## Added tests?

- [x] yes
- [ ] no, because I need help


app/liquid_tags/github_tag/github_readme_tag.rb
```diff
@@ -3,71 +3,79 @@
 class GithubTag
   class GithubReadmeTag
     PARTIAL = "liquids/github_readme".freeze
+    GITHUB_DOMAIN_REGEXP = %r{.*github.com/}.freeze
+    OPTION_NO_README = "no-readme".freeze
+    VALID_OPTIONS = [OPTION_NO_README].freeze
```
mstruve, line 8: I'm confused about how the readme plays into this. Can you explain how that works? The `show_readme?` being true when `options.none?([OPTION_NO_README])` was kinda making my head spin with the negatives.

app/liquid_tags/github_tag/github_readme_tag.rb
```diff
@@ -3,71 +3,79 @@
...
       ActionController::Base.new.render_to_string(
         partial: PARTIAL,
         locals: {
           content: content,
-          show_readme: show_readme? && readme_html.present?,
+          show_readme: readme_html.present?,
           readme_html: readme_html
         },
       )
+    rescue Github::Errors::NotFound, Github::Errors::InvalidRepository
```
mstruve, line 29: Are these going to bubble all the way up to Honeybadger? If so what action do these indicate we have to take?

app/liquid_tags/github_tag/github_readme_tag.rb
```diff
@@ -3,71 +3,79 @@
...
-      options
+      [repository_path.normalize.to_s, options]
+    end
+
+    def validate_options!(*options)
+      return if options.empty?
+      return if options.all? { |o| VALID_OPTIONS.include?(o) }
+
+      message = "GitHub tag: invalid options: #{options - VALID_OPTIONS} - supported options: #{VALID_OPTIONS}"
+      raise StandardError, message
```
mstruve, line 56: Same as above, will these end up bubbling up to Honeybadger? If so, we may want to consider only showing them to the user rather than pinging us when they happen since there is likely not much we can do.

app/liquid_tags/github_tag/github_readme_tag.rb
rhymes: yeah, that's how to the tag was implemented. By default it shows the readme, the option `no-readme` doesn't render one

```liquid
{% github https://github.com/rust-lang/rust %}
```

renders

Screenshot_2020-05-07 New Post - DEV Community 👩‍💻👨‍💻

```liquid
{% github https://github.com/rust-lang/rust no-readme %}
```

renders

Screenshot_2020-05-07 New Post - DEV Community 👩‍💻👨‍💻(1)

the `show_readme?` returns true unless `no-readme` is present in the options

app/liquid_tags/github_tag/github_readme_tag.rb
rhymes: Nope,  no need, they only signal to the user that the repository they are trying to render is either not found or invalid

app/liquid_tags/github_tag/github_readme_tag.rb
rhymes: They are shown to the user, I didn't change the current behavior. The `GithubReadmeTag` is rendered in here:

https://github.com/thepracticaldev/dev.to/blob/06c8be370f5c6e6b341ef0ae51617ade7b804f32/app/liquid_tags/github_tag.rb#L18-L27

Liquid errors are shown to the user already:

**master**

Screenshot_2020-05-07 New Post - DEV Community 👩‍💻👨‍💻(2)

Screenshot_2020-05-07 New Post - DEV Community 👩‍💻👨‍💻(3)

**this branch**

Screenshot_2020-05-07 New Post - DEV(local) Community 👩‍💻👨‍💻

Screenshot_2020-05-07 New Post - DEV(local) Community 👩‍💻👨‍💻(1)

app/liquid_tags/github_tag/github_readme_tag.rb
```diff
@@ -3,71 +3,79 @@
...
-    def parse_options(link)
-      opts = sanitize_link(link)
-      _, *options = opts.split(" ")
+      validate_options!(*options)
 
-      validated_options = options.map { |option| valid_option(option) }.reject(&:nil?)
-      raise StandardError, "GitHub tag: `#{link}`: Invalid option - did you mean `no-readme`?" unless options.empty? || validated_options.any?
+      path.gsub!(%r{/\z}, "") # remove optional trailing forward slash
+      repository_path = URI.parse(path)
+      repository_path.query = repository_path.fragment = nil
```
vaidehijoshi, line 46: Hm, I'm unfamiliar this syntax; can you explain how we're deciding what to assign `repository_path.query` to here? I feel like I'm about to learn a new Ruby thing

app/liquid_tags/github_tag/github_readme_tag.rb
lightalloy: Why `StandardError`, not super cool custom error?
Though, as you haven't changed the logic, this doesn't need to be changed in this pr.

rhymes: @lightalloy I don't know, I'd like to fix that eventually. All liquid tags use `StandardError` and that's bothering me a little :D

app/liquid_tags/github_tag/github_readme_tag.rb
rhymes: If path contains query parameters or fragments, `path.query` and `path.fragment` will be populated:

```ruby
[2] pry(main)> path.fragment
=> "issue?s=query"
[3] pry(main)> path = URI.parse("/path?s=query#fragment")
=> #<URI::Generic /path?s=query#fragment>
[4] pry(main)> path.fragment
=> "fragment"
[5] pry(main)> path.query
=> "s=query"
[6] pry(main)> path.to_s
=> "/path?s=query#fragment"
```

since we don't need neither parameters not the fragment we just remove them, so that the string version of the path will be devoid of them:

```ruby
[7] pry(main)> path.query = path.fragment = nil
=> nil
[8] pry(main)> path.to_s
=> "/path"
```

PR #7708, opened 2020-05-07
Replaced ApplicationConfig with app_url in views

lightalloy:

- [x] Refactor

## Description
Replaced urls that contained`ApplicationConfig["APP_DOMAIN"]` with `app_url` in views.

### DIFF
--- a/app/views/mailers/pro_membership_mailer/expiring_membership.text.erb
+++ b/app/views/mailers/pro_membership_mailer/expiring_membership.text.erb
@@ -1,5 +1,5 @@
-<% credits_url = url_for(host: app_url, controller: :credits, action: :index) %>
-<% pro_membership_url = url_for(host: app_url, controller: :users, action: :edit, tab: "pro-membership") %>
+<% credits_url = app_url(credits_path) %>
+<% pro_membership_url = app_url(user_settings_path("pro-membership")) %>
 
 Hi <%= @user.name %>, your Pro Membership will expire in <%= @days %> days on <%= @date.to_s(:long) %>!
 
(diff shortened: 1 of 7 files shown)


app/views/mailers/pro_membership_mailer/expiring_membership.text.erb
```diff
@@ -1,6 +1,5 @@
-<% host = "#{ApplicationConfig['APP_PROTOCOL']}#{ApplicationConfig['APP_DOMAIN']}" %>
-<% credits_url = url_for(host: host, controller: :credits, action: :index) %>
-<% pro_membership_url = url_for(host: host, controller: :users, action: :edit, tab: "pro-membership") %>
+<% credits_url = url_for(host: app_url, controller: :credits, action: :index) %>
```
rhymes, line 1: maybe we can replace this with just `app_url(credits_path)`

app/views/mailers/pro_membership_mailer/expiring_membership.text.erb
```diff
@@ -1,6 +1,5 @@
...
+<% pro_membership_url = url_for(host: app_url, controller: :users, action: :edit, tab: "pro-membership") %>
```
rhymes, line 2: this page in the settings doesn't exist anymore

app/views/users/_profile.html.erb
```diff
@@ -4,7 +4,7 @@
 
 <div class="crayons-card mb-6 p-6 grid gap-6">
   <div class="flex items-center">
-    <a href="https://<%= ApplicationConfig["APP_DOMAIN"] %>/<%= current_user.username %>" aria-label="Go to your profile page" class="mr-2">
+    <a href="<%= app_url(current_user.username) %>" aria-label="Go to your profile page" class="mr-2">
```
rhymes, line 7: this can also be `user_url(current_user)` though the result is the same for now :D

app/views/mailers/pro_membership_mailer/expiring_membership.text.erb
lightalloy: I was still able to open it on localhost though. Does it mean that this mailer template doesn't make sense anymore?

rhymes: You're right! The page is still there, I thought it was deleted when we started taking out features for Pro. Anyway, let's keep it there. We can revisit what to do about pro in the future :D

PR #7730, opened 2020-05-07
Track Client IPs to Help Block Bad Actors

mstruve:
- [x] Feature

## Description
Allow us to see IPs in Honeycomb to block bad actors. Chatted with Ben about this approach and we agree it seems like a good one. 

## Added tests?
- [x] yes

alt_text


mstruve:
Admin merging since I verbally confirmed the approach with Ben and we need to get this bad actor blocked

PR #7731, opened 2020-05-07
Refactor User-Specific JSON-LD in the Stories::Controller

### LINKED ISSUE #7495: Refactor JSON-LD Data Structures in the StoriesController

juliannatetreault:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
This PR refactors the existing structured data that exists in the `Stories::Controller`. In addition to the refactoring, this PR also adds inline documentation to User-related JSON-LD and a test around nonexistent keys/values within `user_show_spec.rb`.

Closes #7495 

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed
- [x] inline documentation in the `Stories::Controller`

Erase

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -361,20 +361,13 @@ class StoriesController < ApplicationController
       "sameAs": user_same_as,
       "image": ProfileImage.new(@user).get(width: 320),
       "name": @user.name,
-      "email": "",
-      "jobTitle": "",
+      "email": @user.email.presence,
+      "jobTitle": @user.employment_title.presence,
       "description": @user.summary.presence || "404 bio not found",
       "disambiguatingDescription": user_disambiguating_description,
-      "worksFor": [
-        {
-          "@type": "Organization",
-          user_works_for
-        },
-      ],
-      "alumniOf": ""
+      "worksFor": [user_works_for],
+      "alumniOf": @user.education.presence
     }
-    # set_user_profile_json_ld
-    # set_user_same_as_json_ld
   end
 
   def set_article_json_ld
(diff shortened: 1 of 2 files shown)


app/controllers/stories_controller.rb
```diff
@@ -351,22 +353,16 @@ def set_user_json_ld
...
       },
       "url": URL.user(@user),
-      "sameAs": [],
+      "sameAs": user_same_as,
       "image": ProfileImage.new(@user).get(width: 320),
       "name": @user.name,
-      "email": "",
-      "jobTitle": "",
+      "email": @user.email.presence,
+      "jobTitle": @user.employment_title.presence,
```
juliannatetreault, line 360: How do we feel about this and the potential for `nil` values to be present here if the `email` and/or `employment_title` is not present?

vaidehijoshi: IIRC, JSON-LD doesn't allow for `nil` values? I played around with `nil` values in the structure [here](https://json-ld.org/playground/#startTab=tab-expanded&json-ld=%7B%22%40context%22%3A%22http%3A%2F%2Fschema.org%22%2C%22email%22%3A%22vaidehi%40dev.to.%22%2C%22jobTitle%22%3A%22senior%20software%20engineer%22%7D), and I don't think `nil` values would work.

juliannatetreault: That's what I was worried about @vaidehijoshi . 😕

vaidehijoshi: If a value doesn't exist, do you know if JSON-LD prefers an empty string or no key at all? If it prefers an empty string, I guess we can fallback to that (which is what it looks like you already had!), but if it doesn't even make use of empty strings, then we can just not include the key at all if its value is `nil`.

juliannatetreault: No key it is!

spec/requests/user/user_show_spec.rb
```diff
@@ -62,6 +59,16 @@
       )
     end
     # rubocop:enable RSpec/ExampleLength
+
+    it "does not render a key if no value is given" do
```
juliannatetreault, line 63: Does anyone have an opinion on adding/not adding `:aggregate_failures` here?

app/controllers/stories_controller.rb
```diff
@@ -434,31 +432,40 @@ def set_organization_json_ld
...
+      @user.linkedin_url,
+      @user.behance_url,
+      @user.stackoverflow_url,
+      @user.dribbble_url,
+      @user.medium_url,
+      @user.gitlab_url,
+      @user.instagram_url,
+      @user.twitch_username,
+      @user.website_url,
+    ].compact
```
rhymes, line 469: keep in mind that compact only removes `nil`, not empty string. Don't know which of these can be empty but it could happen that you have `["value", "", "value"]` in the array. I think :D

app/controllers/stories_controller.rb
```diff
@@ -358,22 +360,16 @@ def set_user_json_ld
         "@id": URL.user(@user)
       },
       "url": URL.user(@user),
-      "sameAs": [],
+      "sameAs": user_same_as,
       "image": ProfileImage.new(@user).get(width: 320),
       "name": @user.name,
-      "email": "",
-      "jobTitle": "",
+      "email": @user.email.presence,
```
rhymes, line 366: this can only be shown if the email is public, see https://github.com/thepracticaldev/dev.to/pull/7731/files#diff-036c389e944275ed9596b5d2c9b30027L444

spec/requests/user/user_show_spec.rb
rhymes: We're definitely not consistent in using `:aggregate_failures`. I', good either way :D

app/controllers/stories_controller.rb
```diff
@@ -358,22 +360,16 @@ def set_user_json_ld
...
-        },
-      ],
-      "alumniOf": ""
-    }
-    set_user_profile_json_ld
-    set_user_same_as_json_ld
+      "disambiguatingDescription": user_disambiguating_description,
+      "worksFor": [user_works_for].compact,
+      "alumniOf": @user.education.presence
+    }.reject { |_, v| v.blank? }
```
vaidehijoshi, line 372: This is clever. Definitely missed this in my first pass-through of this PR :)

app/controllers/stories_controller.rb
vaidehijoshi: `your_array.reject { |item| item.blank? }` might help here ;)

app/controllers/stories_controller.rb
vaidehijoshi: A follow up thought: are there any keys that absolutely _must_ exist to render JSON-LD? If so, is there a chance that they'll be `blank?` and then not show up in the JSON-LD object?

Don't know if this is something that is likely to happen, but just want to double check.

app/controllers/stories_controller.rb
juliannatetreault: Good catch! I can add a check for the public email here.

app/controllers/stories_controller.rb
juliannatetreault: That's a good point! Maybe I should replace this with `.reject` instead... 🤔

app/controllers/stories_controller.rb
juliannatetreault: That's a good question! The keys that are absolutely required have hard-coded values (e.g. `"@context": "Schema.org"`) or values that should not evaluate to `true` when `.blank?` is called (e.g. `"@type": "Person"`). That being said, I'll double-check, but I think that I've mostly accounted for the edge cases. 😅

app/controllers/stories_controller.rb
juliannatetreault: I think so @vaidehijoshi !

app/controllers/stories_controller.rb
```diff
@@ -350,6 +350,8 @@ def redirect_to_lowercase_username
   end
 
   def set_user_json_ld
+    # For more info on structering data with JSON-LD,
```
vaidehijoshi, line 353:
```suggestion
    # For more info on structuring data with JSON-LD,
```

app/controllers/stories_controller.rb
```diff
@@ -358,22 +360,16 @@ def set_user_json_ld
         "@id": URL.user(@user)
       },
       "url": URL.user(@user),
-      "sameAs": [],
+      "sameAs": user_same_as,
       "image": ProfileImage.new(@user).get(width: 320),
       "name": @user.name,
-      "email": "",
-      "jobTitle": "",
+      "email": @user.email_public && @user.email.presence,
```
rhymes, line 366: this will result in `email: false` if the email is not public:

```ruby
[1] pry(main)> email = false && "bob@acme.com"
=> false
[2] pry(main)> email
=> false
```

it should be:

```ruby
"email": @user.email_public ? @user.email : nil
```

juliannatetreault:
I've tried rerunning the Travis check since the spec that is failing is passing for me locally, but I never seem to have luck when I rerun it. Would someone else mind rerunning the Travis build for me, please? 😅

rhymes:
> hm, not sure why my review was dismissed (rebase?)
> 
> anyways, this looks 👍 pending green CI checks!

@vaidehijoshi no, that was me, i re-requested your review after mine :D

PR #7732, opened 2020-05-07
Onboarding fix for mobile

ludwiczakpawel:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR fixes scrolling issues on mobile resolutions.

## Added tests?

## Added to documentation?

- [x] no documentation needed

### DIFF
--- a/app/javascript/onboarding/__tests__/__snapshots__/Onboarding.test.jsx.snap
+++ b/app/javascript/onboarding/__tests__/__snapshots__/Onboarding.test.jsx.snap
@@ -91,7 +91,7 @@ preact-render-spy (1 nodes)
 -------
 <main class="onboarding-body">
   <div class="onboarding-main crayons-modal">
-    <div class="crayons-modal__box">
+    <div class="crayons-modal__box overflow-auto">
       <nav class="onboarding-navigation">
         <div class="navigation-content">
           <div class="back-button-container">
@@ -201,7 +201,7 @@ preact-render-spy (1 nodes)
 -------
 <main class="onboarding-body">
   <div class="onboarding-main crayons-modal">
-    <div class="crayons-modal__box">
+    <div class="crayons-modal__box overflow-auto">
       <nav class="onboarding-navigation">
         <div class="navigation-content">
           <div class="back-button-container">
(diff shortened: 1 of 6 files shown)


app/javascript/onboarding/__tests__/__snapshots__/Onboarding.test.jsx.snap
```diff
@@ -91,7 +91,7 @@ preact-render-spy (1 nodes)
 -------
 <main class="onboarding-body">
   <div class="onboarding-main crayons-modal">
-    <div class="crayons-modal__box">
+    <div class="crayons-modal__box overflow-auto">
```
vaidehijoshi, line 94: How is this different from adding a class and putting `overflow: auto;` into it?

ludwiczakpawel: No different at all.

`overflow-auto` though is one of our utility-classes - thing you may have heard about from me in the past.. :)

Utility classes are basically bunch of classes for almost every CSS property. You can read a bit more about general concept of [utility classes](https://tailwindcss.com) here. So the intention of Crayons is to structure frontend as follow:

### Components

These are ready to use pieces of code that you can basically copy'n'paste anywhere... think of buttons or form elements... you should be able to grab piece of frontend code (whether it's html or preact) and use it in your work without touching any css.

### Utility classes

Let's go through two scenarios...

1. You  are using one of the existing components but designer was moron and came up with some weird variation of the component that doesn't exist in our codebase... You can then use utility classes to customize a component... This is exactly how I used `overflow-auto` above: `crayons-modal__box` is part of existing component. But this element is not scrollable when it's used properly (so with the rest of the `crayons-modal` markup used as recommended/intended). So adding utility class in the markup to customize it is a bit cleaner way than creating 71298th SCSS file with just bunch of overwrites (even though both ways would work) :D

2. You need to build new feature. It's more complex, almost entire view... Obviously there's no component yet supporting that feature design so you have to build it from scratch. You can either create a component or quickly build something with utility classes. Building component is much more time-consuming because you have to make sure it gonna work in every context (not only for your feature), test edge cases, think it through etc... So instead you build it with utility classes.. It's technically almost like writing CSS from scratch but you don't have to worry about sizes, spacings etc since these are all built-into utility classes...

---

One of my projects for current cool down period is to document and explain our frontend architecture - Crayons. that also includes documenting components, listing all of our utility classes etc...

vaidehijoshi: @ludwiczakpawel OHHHH this makes more sense. I was missing the "utility class" part of the puzzle (I still am learning what our utility classes are!).

PR #7733, opened 2020-05-07
Refactor articles_show_spec to Better Test JSON-LD

juliannatetreault:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
This PR refactors the `articles_show_spec.rb` to better test the structured data for articles and organizations within the `Stories::Controller`.

Relates to #7495 

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Leslie Knope

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -423,7 +423,7 @@ class StoriesController < ApplicationController
       "url": URL.organization(@organization),
       "image": ProfileImage.new(@organization).get(width: 320),
       "name": @organization.name,
-      "description": @user.summary.presence || "404 bio not found"
+      "description": @organization.summary.presence || "404 bio not found"
     }
   end
 
--- a/spec/requests/articles/articles_show_spec.rb
+++ b/spec/requests/articles/articles_show_spec.rb
@@ -52,7 +52,7 @@ RSpec.describe "ArticlesShow", type: :request do
           "name" => user.name
         },
         "datePublished" => article.published_timestamp,
-        "dateModified" => article.edited_at&.iso8601 || article.published_timestamp,
+        "dateModified" => article.published_timestamp,
       )
     end
   end
--- a/spec/requests/articles/articles_show_spec.rb
+++ b/spec/requests/articles/articles_show_spec.rb
@@ -2,8 +2,6 @@ require "rails_helper"
 
 RSpec.describe "ArticlesShow", type: :request do
   let_it_be(:user) { create(:user) }
-  let_it_be(:organization) { create(:organization) }
-  let_it_be(:organization_article) { create(:article, organization_id: organization.id) }
   let_it_be(:article, reload: true) { create(:article, user: user, published: true) }
 
   describe "GET /:slug (articles)" do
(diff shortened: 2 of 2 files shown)


app/controllers/stories_controller.rb
```diff
@@ -423,7 +423,7 @@ def set_organization_json_ld
       "url": URL.organization(@organization),
       "image": ProfileImage.new(@organization).get(width: 320),
       "name": @organization.name,
-      "description": @user.summary.presence || "404 bio not found"
+      "description": @organization.summary.presence || "404 bio not found"
```
juliannatetreault, line 426: I'm pretty sure this should be `@organization` and not `@user` like it was prior to this commit... 🤔

spec/requests/articles/articles_show_spec.rb
```diff
@@ -15,31 +17,64 @@
...
-    # the date is off. See https://github.com/thepracticaldev/dev.to/issues/7086.
-    xit "renders the proper modified at date" do
-      article.update(edited_at: Time.zone.now)
-      get article.path
-      expect(response.body).to include GI.escapeHTML(article.edited_at.strftime("%b %d, %Y"))
-    end
-
-    it "renders the proper author" do
-      expect(response.body).to include CGI.escapeHTML(article.cached_user_username)
+    # rubocop:disable Rspec/ExampleLength
```
rhymes, line 20: the `S` is capital, just telling you because rubocop flagged this in another PR, see https://github.com/thepracticaldev/dev.to/pull/7743/files#diff-ba9ab82dc4668287786466b0edda73ffR18 :D

spec/requests/articles/articles_show_spec.rb
```diff
@@ -15,31 +17,64 @@
...
+          "@id" => URL.organization(organization)
+        },
+        "url" => URL.organization(organization),
+        "image" => ProfileImage.new(organization).get(width: 320),
+        "name" => organization.name,
+        "description" => organization.summary
+      },
+    )
   end
+  # rubocop:enable Rspec/ExampleLength
```
rhymes, line 77: same thing about the `S`

spec/requests/articles/articles_show_spec.rb
juliannatetreault: Thank you for pointing that out! I've made this change. 😄

spec/requests/articles/articles_show_spec.rb
```diff
@@ -15,31 +17,64 @@
...
+        },
+        "headline" => article.title,
+        "author" => {
+          "@context" => "http://schema.org",
+          "@type" => "Person",
+          "url" => URL.user(user),
+          "name" => user.name
+        },
+        "datePublished" => article.published_timestamp,
+        "dateModified" => article.edited_at&.iso8601 || article.published_timestamp,
```
maestromac, line 55: Instead of using a `||` here, I think we should be more decisive on what we expect `response_json` to have

juliannatetreault: That's fair! The reason why I left the `||` here is because we've been running into issues with testing the `edited_at` date--this spec was flaky prior to this refactor because of the way we localize date times, causing the date to be off sometimes.

That being said though, I can remove the `edited_at` date in favor of the `published_timestamp`.

PR #7736, opened 2020-05-07
Optimize FeedbackMessagesController#index

maestromac:
- [x] Optimization

## Description
Following @guivinicius 's work to optimize internal controllers. Adding `limit` and only selecting necessary column to speed up load time.

Resolves https://github.com/thepracticaldev/dev.to/issues/6674
n/a
## Added tests?
- [ ] no, because I need help

## Added to documentation?
- [x] no documentation needed

n/a

### DIFF
--- a/app/controllers/internal/feedback_messages_controller.rb
+++ b/app/controllers/internal/feedback_messages_controller.rb
@@ -5,13 +5,13 @@ class Internal::FeedbackMessagesController < Internal::ApplicationController
     @q = FeedbackMessage.includes(:reporter, :offender, :affected).
       order(created_at: :desc).
       ransack(params[:q])
-    @feedback_messages = @q.result.page(params[:page] || 1).per(5)
+    @feedback_messages = @q.result.limit(5).page(params[:page] || 1).per(5)
 
     @feedback_type = params[:state] || "abuse-reports"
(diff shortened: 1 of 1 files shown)


app/controllers/internal/feedback_messages_controller.rb
```diff
@@ -5,13 +5,13 @@ def index
     @q = FeedbackMessage.includes(:reporter, :offender, :affected).
       order(created_at: :desc).
       ransack(params[:q])
-    @feedback_messages = @q.result.page(params[:page] || 1).per(5)
+    @feedback_messages = @q.result.limit(5).page(params[:page] || 1).per(5)
```
guivinicius, line 8: I think this is already paginating this right? Unless you are changing the page behaviour to only show 5 feedback messages.

Zhao-Andy: Yeah not sure if this is redundant? Worst case scenario though is that it doesn't do anything, which is not bad at all.

maestromac: Ah right it is doing the same thing. Thanks!
```suggestion
    @feedback_messages = @q.result.page(params[:page] || 1).per(5)
```

PR #7737, opened 2020-05-07
NOT MERGE Add tabs to Crayons and to the notifications page

ludwiczakpawel:

- [x] Feature

## Description

This PR updates styles for Tabs and it adds better responsiveness behaviours. This component has already existed in the codebase in a very basic form and was used in one place: Notifications - i've also updated that.

https://d.pr/i/ND6UIK

## Added tests?

## Added to documentation?

- [x] no documentation needed


app/assets/stylesheets/components/forms.scss
```diff
@@ -113,7 +113,7 @@
 // Selects
 .crayons-select {
   line-height: var(--lh-base);
-  padding: 0.5em;
+  padding: calc(0.5em - var(--border-width)) 0.5em;
```
ludwiczakpawel, line 116: Just a little bug I found in the meantime....

app/views/notifications/index.html.erb
```diff
@@ -95,95 +95,43 @@
...
-                  <span class="notifications-filter__dropdown__label--subtitle">
-                    @<%= @user.username %> · Personal
-                  </span>
-                </div>
-                <div class="notifications-filter__icon">
-                  <%= inline_svg_tag("checkmark.svg", aria: true, class: "crayons-icon", title: "Checkmark") %>
-                </div>
-              </a>
+          <div class="p-2 s:inline-flex">
+            <select onChange="window.location.href=this.value" class="crayons-select">
```
rhymes, line 101: I'm a bit uneasy about this.

cc. @nickytonline about it

ludwiczakpawel:
> Right now in production we have all this info visible:
> Screenshot 2020-05-08 at 12 36 18 PM
> Instead now we show only the username or slug, could we at least replace it with `user.name` and `organization.name` ?

I think so... But I wanted to be consistent with what we have on desktop version where we display username/slug and not full names. In that case we should probably unify these on desktop version too. Thoughts?

> Also, there's not much padding (between checkbox and name) or margin between the entries in the select
> Screenshot 2020-05-08 at 12 37 00 PM

It's a native browser select - nothing we can control. Current version (the one with full info like name, slug etc) is a bit overkill in terms of information and implementation. After chatting with @lisasy who worked on it earlier, we came to conclusion we can simplify that to be browser-native `select`. If you preview the smallest breakpoint on your desktop browser it looks off - that's true. But you usually open the smallest breakpoint on actual mobile device in which select becomes sliding drawer from the bottom: https://d.pr/v/QQ2OC0

rhymes:
@ludwiczakpawel agreed, let's keep it like this for now. thanks for changing names. I'd like @nickytonline to take a look at this before we merge it but overall, looks awesome!

ludwiczakpawel:
@rhymes cheers! I've added _NOT MERGE_ to the title until Nick review it.

app/assets/stylesheets/components/tabs.scss
```diff
@@ -0,0 +1,136 @@
...
+
+  &__item {
+    padding: var(--tab-vertical-padding) var(--tab-horizontal-padding);
+    display: inline-block;
+    text-align: center;
+    outline: none;
+    font-size: var(--font-size);
+    line-height: var(--su-6);
+    text-decoration: none;
+    cursor: pointer;
```
nickytonline, line 29: Just curious about how you make the decision to use a CSS property vs. a utility class? In this case, i.e. `cursor-pointer` utility class vs. a CSS propert/value, `cursor: pointer`.

app/views/notifications/index.html.erb
nickytonline: This should be set up using [addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) via JS. Do you need help with that @ludwiczakpawel? If so, we can sync in the morning/your afternoon.

app/assets/stylesheets/components/tabs.scss
ludwiczakpawel: Utility classes are more for scenarios when you want to either customize existing component or for example create w feature quickly without building the component itself. Or you need to build a particular view or piece of view and not sure how to handle that in frontend infra...

I've tried to give few scenarios for utility classes usage in one [comment for Vaidehi](https://github.com/thepracticaldev/dev.to/pull/7732#discussion_r421734732) - I hope that's helpful.

app/views/notifications/index.html.erb
ludwiczakpawel: would you be able to help me out with this one?

not really sure how to implement that

PR #7738, opened 2020-05-07
Shorten page title for unauthenticated traffic

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This shortens the page title from `Productive and simple way to deploy Dockerized applications - DEV Community 👩‍💻👨‍💻` to just `Productive and simple way to deploy Dockerized applications - DEV` for non-signed-in traffic. Home page is still `DEV Community 👩‍💻👨‍💻`.

This is because search engines have struggled to display this consistently and moving just to DEV should help. I love the funkiness of the emojis but now that we're not using them for signed-out traffic we can be even funkier with them because we won't care about confusing the bots. We had a PR about randomizing the skin tones of the characters in the emoji and I was sort of unsure about doing that if the search engines are expecting consistency but this could free us up in that regard.

### DIFF
--- a/spec/requests/stories_show_spec.rb
+++ b/spec/requests/stories_show_spec.rb
@@ -19,6 +19,19 @@ RSpec.describe "StoriesShow", type: :request do
       expect(response).to have_http_status(:moved_permanently)
     end
 
+    ## Title tag
+    it "renders signed-in title tag for signed-in user" do
+      sign_in user
+      get article.path
+      p community_qualified_name
+      expect(response.body).to include "<title>#{CGI.escapeHTML(article.title)} - #{community_qualified_name} 👩‍💻👨‍💻</title>"
+    end
+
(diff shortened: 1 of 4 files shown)


spec/requests/stories_show_spec.rb
```diff
@@ -19,6 +19,19 @@
       expect(response).to have_http_status(:moved_permanently)
     end
 
+    ## Title tag
+    it "renders signed-in title tag for signed-in user" do
+      sign_in user
+      get article.path
+      p community_qualified_name
```
rhymes, line 26: this can be removed

benhalpern: Oh yeah

benhalpern, line 26:
```suggestion
```

PR #7745, opened 2020-05-08
Add Feature Search Tests for Users and Articles

mstruve:
- [x] Feature Specs

## Description
These two specs test that when we load users and articles from Elasticsearch they have all the data necessary to correctly display on the page. I plan to do this as well for podcasts and comments. We can further expand on these with many scenarios but I thought this was a good base to get us going. 

https://github.com/thepracticaldev/dev.to/projects/6#card-36667928

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/spec/system/search/display_articles_search_spec.rb
+++ b/spec/system/search/display_articles_search_spec.rb
@@ -1,0 +1,40 @@
+require "rails_helper"
+
+RSpec.describe "Display articles search spec", type: :system, js: true, elasticsearch: "FeedContent" do
+  let(:found_article_one) { create(:article) }
+  let(:found_article_two) { create(:article) }
+  let(:not_found_article) { create(:article) }
+
+  before do
+    stub_request(:post, "http://www.google-analytics.com/collect")
+  end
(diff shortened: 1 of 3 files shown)


spec/system/search/display_articles_search_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  it "returns correct results for a search" do
+    found_article_one.tags << create(:tag, name: "ruby")
+    allow(found_article_two).to receive(:body_text).and_return("Ruby Tuesday")
+    allow(not_found_article).to receive(:body_text).and_return("Python All Day Long")
+    articles = [found_article_one, found_article_two, not_found_article]
+    index_documents_for_search_class(articles, Search::FeedContent)
+    visit "/search?q=ruby&filters=class_name:Article"
+
+    expect(page).to have_content(found_article_one.title)
+    expect(page).to have_content(found_article_two.title)
```
atsmith813, line 21: Why is this article found when searching with `q=ruby`?

PR #7746, opened 2020-05-08
Limit User Email Changes

mstruve:
- [x] Optimization
- [x] Refactor

## Description
Only allow 2 email changes within a 2 minute period. If limit is exceeded show the user an error message with a retry message. As you can see below, since the email is not updated we still show that a verification email was sent to the previous updated email address. At the same time we also tell the user they will have to wait 120 seconds before retrying to update their email.

Screen Shot 2020-05-08 at 4 46 19 PM

I also took this opportunity to do a solid refactor of the RateLimitCheck spec. There is more that could be done but this eliminates the needs to add multiple specs when we add simple cache rate limit keys.

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -140,6 +140,7 @@ class User < ApplicationRecord
   validate :unique_including_orgs_and_podcasts, if: :username_changed?
   validate :validate_feed_url, if: :feed_url_changed?
   validate :validate_mastodon_url
+  validate :can_send_confirmation_email
 
   alias_attribute :positive_reactions_count, :reactions_count
   alias_attribute :subscribed_to_welcome_notifications?, :welcome_notifications
@@ -450,6 +451,10 @@ class User < ApplicationRecord
     search_score
   end
 
+  def rate_limiter
+    RateLimitChecker.new(self)
+  end
+
   private
 
   def estimate_default_language
@@ -603,5 +608,14 @@ class User < ApplicationRecord
 
   def index_roles(_role)
     index_to_elasticsearch_inline
+  end
+
+  def can_send_confirmation_email
+    return if changes[:email].blank?
+
+    rate_limiter.track_limit_by_action(:send_email_confirmation)
+    rate_limiter.check_limit!(:send_email_confirmation)
+  rescue RateLimitChecker::LimitReached => e
+    errors.add(:email, "confirmation could not be sent. #{e.message}")
   end
 end
--- a/spec/initializers/rack/attack_spec.rb
+++ b/spec/initializers/rack/attack_spec.rb
@@ -2,10 +2,9 @@ require "rails_helper"
 
 describe Rack::Attack, type: :request, throttle: true do
   before do
-    redis_url = "redis://localhost:6379"
-    cache_db = ActiveSupport::Cache::RedisStore.new(redis_url)
+    cache_db = ActiveSupport::Cache.lookup_store(:redis_cache_store)
     allow(Rails).to receive(:cache) { cache_db }
-    cache_db.data.flushdb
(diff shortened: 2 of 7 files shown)


app/models/user.rb
```diff
@@ -450,6 +451,10 @@ def hotness_score
     search_score
   end
 
+  def rate_limiter
```
mstruve, line 454: Made this public bc I think we could refactor other parts of the code like in the application controller `current_user.rate_limiter` rather than `RateLimitChecker.new(current_user)`

spec/initializers/rack/attack_spec.rb
```diff
@@ -2,10 +2,9 @@
 
 describe Rack::Attack, type: :request, throttle: true do
   before do
-    redis_url = "redis://localhost:6379"
-    cache_db = ActiveSupport::Cache::RedisStore.new(redis_url)
+    cache_db = ActiveSupport::Cache.lookup_store(:redis_cache_store)
```
mstruve, line 5: We actually want redis cache store here which is what Rails.cache uses.

spec/services/rate_limit_checker_spec.rb
```diff
@@ -20,14 +43,6 @@
         expect(rate_limit_checker.limit_by_action("comment_creation")).to be(true)
       end
 
-      it "sends a slack message if a user leaves too any messages" do
```
mstruve, line 23: We no longer send messages to slack for rate limit hits, this was a false positive spec

spec/services/rate_limit_checker_spec.rb
```diff
@@ -5,11 +5,34 @@
...
+    rate_limit_checker.send("limit_cache_key", action)
+  end
+
   describe "#limit_by_action" do
     it "returns false for invalid action" do
       expect(rate_limit_checker.limit_by_action("random-nothing")).to be(false)
     end
 
+    # published_article_creation limit we check against the databse rather than our cache
+    RateLimitChecker::ACTION_LIMITERS.except(:published_article_creation).each do |action, _options|
```
mstruve, line 18: More refactoring could be done in this spec but this was a quick win that I thought I would punch out in this PR

spec/services/rate_limit_checker_spec.rb
```diff
@@ -5,11 +5,34 @@
   let(:article) { create(:article, user: user) }
   let(:rate_limit_checker) { described_class.new(user) }
 
+  def cache_key(action)
+    rate_limit_checker.send("limit_cache_key", action)
+  end
+
   describe "#limit_by_action" do
```
rhymes, line 12: we should probably add `:aggregate_failures` here as some of the expectations are created at runtime and it'd be complicated to find them otherwise, as the same line corresponds to different test cases

spec/services/rate_limit_checker_spec.rb
```diff
@@ -5,11 +5,34 @@
...
     it "returns false for invalid action" do
       expect(rate_limit_checker.limit_by_action("random-nothing")).to be(false)
     end
 
+    # published_article_creation limit we check against the databse rather than our cache
```
rhymes, line 17: small typo

spec/services/rate_limit_checker_spec.rb
rhymes: wait, can we aggregate failures in a describe? RSpec is magical :D

spec/services/rate_limit_checker_spec.rb
mstruve: That took me reading this twice to figure it out 😂

Issue #7663, opened 2020-05-03
Issue with Rails in Docker setup


farzadbagheri:
**Describe the bug**
Rails aborting while databases are being set up with Docker (LoadError: Error relocating /usr/local/bundle/gems/mini_racer-0.2.10/lib/mini_racer_extension.so...)

**Screenshots**
<img width="1437" alt="Screen Shot 2020-05-03 at 6 31 18 PM" src="https://user-images.githubusercontent.com/10763487/80927533-6d6bff00-8d6c-11ea-8eba-fb2d3f9adf18.png">

OS: Catalina

citizen428:
Thanks for the issue @farzadbagheri! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

citizen428:
This seems to be a problem with Alpine's standard C library (musl instead of glibc), see https://github.com/sgerrand/alpine-pkg-glibc/issues/27

I have to think about how to resolve this, as to the best of my knowledge nobody on the core team uses Docker for their daily development and `mini_racer` was introduced to speed up specs in local development, see #7383.

Possible solutions:

1. Switch from Alpine to a glibc-based distribution. This would require more changes and I'm a bit reluctant to do that.
2. Don't install mini_racer when running inside Docker, which should be achievable with some `Gemfile` hackery.

joshpuetz:
Pinging @jdoss as he's been experimenting with alternate containerization for DEV's development environment

jdoss:
Hey @FarazPatankar give our revamped container setup a try. It should fix this issue. I am going to close this issue for now, but if you have any other problems with our container setup, please open a new one.

FarazPatankar:
@jdoss I think you may have mentioned the wrong person. 😛

jdoss:
@FarazPatankar Woops! My bad! Too much multi-tasking and not enough coffee on my end. My apologizes. I was moving too fast this morning.

Issue #39, opened 2017-05-22
Side-by-side editing or Edit-here


tmr232:
When editing large posts, every transition from preview to markdown makes me go and look for the same paragraph I was in.
Being able to edit the markdown side-by-side with the preview would make it a lot easier.
Another approach that will make editing easier is an "edit here" option, that will switch from preview to markdown, but keep my position in the text.

pkfrank:
Hey @tmr232 — thanks for the suggestion.  We'll definitely be giving the markdown editor a lot more love in the coming months.  I've also experienced that frustration of flipping between `Preview` and `Markdown` and having to go find my spot again.

The suggestion is definitely not falling on deaf ears, we just want to be strategic about improving the whole editing experience when we're at a point to give it a focused and major investment of our time and energy.  Hopefully that'll be soon.

Thanks again for the thought 👍 !

pbnj:
I would like to re-raise this issue 4 years later.

As a platform for developers to write technical articles and blog posts, the editor/writer experience is as important as reader experience.

Having a side-by-side markdown editor and preview window would be a great feature to have in forem.

PR #7747, opened 2020-05-08
Bugfix for #7663 and refactor of container setup

jdoss:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This is a PR that is a a fix for #7663 and a start of a refactor of our usage of Containers.

#7663

## Added tests?
- [ ] yes
- [ ] no, because I need help

### What are the chances that none of this stuff in this PR works??

It's High. It's very high...


joshpuetz:
So running locally I have an issue booting rails (which manifests in the ElasticSearch setup step):

`== Preparing Elasticsearch ==
/usr/share/ruby/irb/input-method.rb:18: warning: already initialized constant IRB::STDIN_FILE_NAME
/usr/share/gems/gems/irb-1.2.3/lib/irb/input-method.rb:18: warning: previous definition of STDIN_FILE_NAME was here
sh: /sbin/ip: No such file or directory
I, [2020-05-09T02:10:56.367671 #39]  INFO -- : Allowing  for BetterErrors and Web Console
rails aborted!
IPAddr::InvalidAddressError: invalid address: 
/usr/share/gems/gems/better_errors-2.7.0/lib/better_errors/middleware.rb:36:in `new'
/usr/share/gems/gems/better_errors-2.7.0/lib/better_errors/middleware.rb:36:in `allow_ip!'
/opt/apps/devto/config/initializers/docker.rb:9:in `<main>'`

I _think_ it's the way `https://github.com/thepracticaldev/dev.to/blob/2d26318cf96c0f1c5c2e827b74bbfa6d27d292d3/config/initializers/docker.rb` is grabbing an IP address...

jdoss:
OK pull down the latest commit and give it a try.

joshpuetz:
Still running into some issues getting the IP address:

```

== Preparing Elasticsearch ==
/usr/share/ruby/irb/input-method.rb:18: warning: already initialized constant IRB::STDIN_FILE_NAME
/usr/share/gems/gems/irb-1.2.3/lib/irb/input-method.rb:18: warning: previous definition of STDIN_FILE_NAME was here
sh: ip: command not found
I, [2020-05-09T20:52:54.015360 #40]  INFO -- : Allowing  for BetterErrors and Web Console
rails aborted!
IPAddr::InvalidAddressError: invalid address: 

```

config/initializers/docker.rb
```diff
@@ -1,7 +1,7 @@
 # Docker specific development configuration
 if Rails.env.development? && File.file?("/.dockerenv")
   # Using shell tools so we don't need to require Socket and IPAddr
-  host_ip = `/sbin/ip route|awk '/default/ { print $3 }'`.strip
+  host_ip = `ip route|awk '/default/ { print $3 }'`.strip
```
joshpuetz, line 4: This won't work on MacOS: I think `ifconfig` is the equivalent: https://unix.stackexchange.com/questions/319350/ip-command-of-iproute2-package

citizen428: @joshpuetz This doesn't need to work on  macOS, as it will only run when the app starts in a Docker container (that's the `File.file?("/.dockerenv")` part).

Containerfile
```diff
@@ -0,0 +1,30 @@
+FROM quay.io/devto/ruby:2.7.1
```
citizen428, line 1: What're the main differences between this and the official Docker images? The distribution?

Dockerfile
```diff
@@ -1,81 +1,30 @@
...
 
-# timeout extension required to ensure
-# system work properly on first time load
-ENV RACK_TIMEOUT_WAIT_TIMEOUT=10000 \
-  RACK_TIMEOUT_SERVICE_TIMEOUT=10000 \
-  STATEMENT_TIMEOUT=10000
-
-# Run mode configuration between dev / demo
-# for entrypoint script behaviour
-ENV	RUN_MODE="demo"
```
citizen428, line 58: We were talking about completely removing this from the entry point script too at one point.

docker-compose.yml
```diff
@@ -41,27 +52,37 @@ services:
...
       memlock:
         soft: -1
         hard: -1
-    volumes:
-        - ./docker_data/elasticsearch/data:/usr/share/elasticsearch/data
-    ports:
-      - "9200:9200"
+
+volumes:
+  db_data:
```
citizen428, line 85: We have a `.gitignore` for `docker_data`:

```
# ES docker volume
docker_data/
```

So I guess we either need to update that or keep both `db_data` and `es_data` in there.

Containerfile
```diff
@@ -0,0 +1,30 @@
...
+
+USER root
+
+RUN curl -sL https://dl.yarnpkg.com/rpm/yarn.repo -o /etc/yum.repos.d/yarn.repo && \
+    dnf install -y bash git ImageMagick iproute less libcurl libcurl-devel \
```
citizen428, line 6: Not sure how I feel about Fedora in this context. Personally I like the distro, and Alpine currently doesn't work for us because of the musl vs libc problem with `mini_racer`. But almost all containers I interact with seem to either inherit from scratch (e.g. multi-stage Go, with statically linked binaries), Alpine or Debian. 

I really don't want to start discussing distributions here, my only point is expectations/familiary for external contributors.

config/initializers/docker.rb
joshpuetz: Ah, good point! That makes the error `sh: ip: command not found` all that more confounding!

Dockerfile
jdoss: We should make the Entrypoints do one thing, not many. That is why I removed it and we can figure out how to make demo mode a thing if it is a needed feature.

Containerfile
jdoss: It's a tailored ruby container to fit our needs which will give us more control over our stack.

config/initializers/docker.rb
jdoss: @joshpuetz make sure you rebuilt the containers as the Containerfile changed with the iproute package being added. I think you are still using the older Container that doesn't have the `iproute` package installed. 

This whole part most likely should be refactored so it just uses Ruby to figure out the default route vs shelling out to `ip` or just removed. I am not really sure what we are whitelisting here but hard coding paths to system binaries in the app makes the code less portable.

Containerfile
jdoss: As long as we provide support and documentation for the containers we are providing our users the distribution inside the container shouldn't be a point of contention IMO.

joshpuetz:
Rebuilt my local containers and got farther along: seeing an issue with rails setup now:

```rails aborted!
LoadError: Error relocating /usr/local/bundle/gems/mini_racer-0.2.10/lib/mini_racer_extension.so: backtrace_symbols: symbol not found - /usr/local/bundle/gems/mini_racer-0.2.10/lib/mini_racer_extension.so
/usr/local/bundle/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:23:in `require'
/usr/local/bundle/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:23:in `block in require_with_bootsnap_lfi'
/usr/local/bundle/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/loaded_features_index.rb:92:in `register'
/usr/local/bundle/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:22:in `require_with_bootsnap_lfi'
/usr/local/bundle/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:31:in `require'
/usr/local/bundle/gems/activesupport-5.2.4.2/lib/active_support/dependencies.rb:291:in `block in require'
/usr/local/bundle/gems/activesupport-5.2.4.2/lib/active_support/dependencies.rb:257:in `load_dependency'
/usr/local/bundle/gems/activesupport-5.2.4.2/lib/active_support/dependencies.rb:291:in `require'
/usr/local/bundle/gems/mini_racer-0.2.10/lib/mini_racer.rb:2:in `<main>' ```

(This seems to be the same error in #7663)

File: Containerfile (hunk 2 above)
Comment (joshpuetz) [reply Containerfile:1]: Do we own it? Is it open source?

**Comment by jdoss** [comment]
Are you on master or my branch? If you are on master you will build the Container with Alpine Linux.

File: Containerfile (hunk 2 above)
Comment (jdoss) [reply Containerfile:1]: Yes, we own that and yes I plan on open sourcing the Containerfile that made it later once I get the other repos in order.

File: config/initializers/docker.rb (hunk 1 above)
Comment (citizen428) [reply config/initializers/docker.rb:4]: @jdoss We are adding these IPs to the list of allowed addresses for Better Errors and the web console, otherwise the debugging experience in Docker will be worse than locally. 

IMHO the hard-coded path was not a problem, as this only applies to the one environment we actually fully control, the distribution running inside the container. The only time it stops working is when we switch that, as right now. I have no problem with doing it in Ruby either, the comment already hints at the two libraries (`Socket` and `IPAddr`) you'll need for that.

File: Containerfile (hunk 5 above)
Comment (citizen428) [reply Containerfile:6]: Fair enough. Let's at least keep some of the documentation close to the code, i.e. 

```docker
# Custom Ruby imaged based on Fedora
FROM quay.io/devto/ruby:2.7.1
```

so less experienced users don't need to figure this out by googling what DNF is.

config/initializers/docker.rb
citizen428: @jdoss As a baseline to get you started:

```ruby
Socket.ip_address_list.detect{ |intf| intf.ipv4_private? }
```

This will return `Addrinfo` objects, calling `ip_address` on those will return the IPs as strings.

Containerfile
jdoss: :+1: I can do some better documentation before I merge this for sure.

omenlog:
Using this branch, during the build process I got the same error regarding to 

`IPAddr::InvalidAddressError: invalid address: `

image

config/initializers/docker.rb
jdoss: My latest commit should fix the issue with `ip` not being found. A total PEBKAC on my end.

```
[jdoss@sts7 devto]$ podman exec -it devto_web_1 /bin/bash
[root@3d439f7595ff devto]# ip route|awk '/default/ { print $3 }'
10.0.2.2
```

jdoss:
@joshpuetz and @omenlog Pull down my latest commits and remove all of your containers and rebuild if you get a few to test it out.

omenlog:
@jdoss using your latest changes I don't get the `InvalidAddressError` and eveything seem to be working properly 

image

the containers are up and running

image

but after this I don't have response from http://locahost:3000 on my machine

omenlog:
@jdoss  the build process is working without problem after your changes

docker-compose.yml
jdoss: Resolved with https://github.com/thepracticaldev/dev.to/pull/7747/commits/23375fb71cf8615b15bfc6ce4f446da80ae9f41e

jdoss:
I did some pretty big refactoring on this PR. Pull it down and nuke all of your containers and give it a try. I will work on some documentation later this week after gathering some feedback.

citizen428:
@jdoss You are the hero DEVham deserves and also the one it needs right now 🦇

I started from a complete `docker system prune` (I do that regularly to reclaim disk space anyway), and failed at step 20:

```
Step 20/28 : COPY ./Gemfile ./Gemfile.lock "${APP_HOME}"
ERROR: Service 'web' failed to build: When using COPY with more than one source file, the destination must be a directory and end with a /
```

To continue I made the following minor change:

```diff
- ENV APP_HOME=/opt/apps/devto
+ ENV APP_HOME=/opt/apps/devto/
```

I then managed to finish the setup, but the app never came up:

```
2020/05/21 09:54:47 Timeout after 5m0s waiting on dependencies to become available: [tcp://db:5432 http://elasticsearch:9200 tcp://redis:6379 file:///opt/apps/bundle/bundle_finished]
Database setup failed with exit code 1.
```

`docker-compose logs` shows an endless stream of verbose Java debug messages from ES. 

Trying to hit ES via `curl` leads to the following error:

```
{"error":{"root_cause":[{"type":"master_not_discovered_exception","reason":null}],"type":"master_not_discovered_exception","reason":null},"status":503}%
```

And so ends another episode of "But Docker works on my machine" 😉

jdoss:
I fixed the `APP_HOME` on the latest commit. @citizen428 that Elasticsearch container works on Linux... :wink:

citizen428:
> @citizen428 that Elasticsearch container works on Linux... 😉

What's running inside that container again? 😛

scripts/entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ -f tmp/pids/server.pid ]; then
+  rm -f tmp/pids/server.pid
+fi
+
+bundle exec rake search:setup
+bundle exec rake db:migrate 2>/dev/null || bundle exec rake db:setup
```
mstruve, line 10: hooking Elasticsearch into these rake tasks so we should be able to just do `db:whatever` and that will handle postgres and ES.

Containerfile
```diff
@@ -0,0 +1,52 @@
...
```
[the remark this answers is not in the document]
jdoss: You and me both...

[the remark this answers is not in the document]
rhymes: TLDR; there might be a light at the end of the tunnel to get rid of ImageMagick: `libvips`

While going through Rails 6 changelog I discovered the [image_processing](https://github.com/janko/image_processing) (default for ActiveStorage in Rails 6) library which is a layer on top of IM and [libvips](https://github.com/libvips/libvips) which is a faster alternative to ImageMagick: <https://github.com/libvips/libvips/wiki/Speed-and-memory-use> which I totally forgot about but knew from when I read https://images.guide back then when Addy Osmani published it.

`image_processing` in turn uses [ruby-vips](https://github.com/libvips/ruby-vips) which is also used by [carrierwave-vips](https://github.com/libvips/ruby-vips) which could be used by us instead of ImageMagick (we currently use IM only to strip EXIFs before sending the image up to S3)

jdoss:
@citizen428 How much RAM do you have allocated to your Docker on Mac? We had to bump up to 4GB of RAM for everything to run well in our tests today.

citizen428:
@jdoss I have it set to 4GB. I noticed the same limitation when I originally worked on our Docker setup and increased it back then too. See https://docs.dev.to/installation/docker/#known-problems--solutions, first point.

jdoss:
OK as promised I have updated the docs to and refactored `bin/docker-setup` to `bin/container-setup` to support both Docker and Podman container engines. It is time to get this out of draft mode and ready for review.

bin/setup
```diff
@@ -15,7 +15,8 @@ FileUtils.chdir APP_ROOT do
 
   puts "== Installing dependencies =="
   system! "gem install bundler --conservative"
-  system! "bundle config --local set cache_all true"
+  system! "bundle config --local cache_all true"
+  system! "bundle config --local cache_all_platforms true"
```
citizen428, line 19: For anyone curious, we decided to add this since some gems (e.g. `libv8`) install platform specific versions like `libv8-7.3.492.27.1-x86_64-linux.gem`.

docs/installation/containers.md
```diff
@@ -0,0 +1,117 @@
+---
```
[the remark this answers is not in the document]
jdoss: Good catch. Updated!

PR #7761, opened 2020-05-11
Add org credit management to org mod page

citizen428:

- [X] Refactor
- [X] Feature

## Description

As requested by @michael-tharrington in https://github.com/thepracticaldev/tech-private/issues/427, this PR adds credit management to the org mod page. It also improves the credit editing experience on the user mod page and cleans up/refactors some code.

https://github.com/thepracticaldev/tech-private/issues/427

Org mod page:

1. The forms for managing credits are now directly on `/internal/organizations/:id` and the remove option will only be shown when there are credits to remove. You also see how many credits are currently available:
    Screen Shot 2020-05-08 at 11 41 09
2. The inputs are now actually of type number, with the little arrows for increasing/decreasing, and all that. In the case of adding credits, values from 1 to 99999 are valid, for removing it's 1 to the number of current credits. So you can't submit the form if you try to remove more credits than an org has (indicated by the red border):
    Screen Shot 2020-05-08 at 11 38 19

User mod page:

Screen Shot 2020-05-11 at 10 37 42

1. Remove user credits form will be hidden if user has no credits.
2. All credits inputs are now number fields. For orgs there's no error checking on the upper bound of remove though, as we don't have that readily available at that point.
3. I also hide the whole org credit management block if the user isn't in any orgs.

## Added tests?

- [X] yes

Less than I would have liked tbh, but I seem to have some trouble with my local test setup and I didn't want to sink any more time into it right now.

## Added to documentation?

- [X] no documentation needed

### DIFF
--- a/app/controllers/internal/organizations_controller.rb
+++ b/app/controllers/internal/organizations_controller.rb
@@ -1,8 +1,13 @@
 class Internal::OrganizationsController < Internal::ApplicationController
   layout "internal"
 
+  CREDIT_ACTIONS = {
+    add: :add_to_org,
+    remove: :remove_from_org
+  }.with_indifferent_access.freeze
+
   def index
-    @organizations = Organization.order("name DESC").page(params[:page]).per(50)
+    @organizations = Organization.order(name: :desc).page(params[:page]).per(50)
 
     return if params[:search].blank?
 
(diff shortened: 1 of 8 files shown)


citizen428:
In the future, we generally may want to look a bit more into how credits curently work. The code often distinguishes between users and orgs (e.g. `add_credits` and `add_credits_to_org`), even though the methods behave exactly the same. There are also methods like `enough_credits?`, though we don't seem to use them in all the appropriate places (for example `ClassifiedListingsToolkit` manually performs similar checks).

michael-tharrington:
This is seriously awesome!

PR #7771, opened 2020-05-11
Decrease General Site Request Limit

mstruve:
- [x] Optimization

## Description
After merging #7676 we have decreased the number of hits we make to our backend in a short period of time. After doing some poking around Honeycomb it looks like we max out around 6 req/second and those spikes are usually not sustained so I figured 20 req/2 secs would give us enough buffer for normal functionality. Let me know your thoughts!  

alt_text

### DIFF
--- a/config/initializers/rack/attack.rb
+++ b/config/initializers/rack/attack.rb
@@ -22,7 +22,7 @@ class Rack::Attack
     end
   end
 
-  throttle("site_hits", limit: 100, period: 2) do |request|
+  throttle("site_hits", limit: 20, period: 2) do |request|
     if request.env["HTTP_FASTLY_CLIENT_IP"].present?
       Honeycomb.add_field("fastly_client_ip", request.env["HTTP_FASTLY_CLIENT_IP"])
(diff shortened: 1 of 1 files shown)


mstruve:
I think the easiest way to do that beyond looking at the page in an inspector would be to figure out what paths those endpoints call and then look at the activity in Honeycomb for those paths.

config/initializers/rack/attack.rb
```diff
@@ -22,7 +22,7 @@ class Rack::Attack
     end
   end
 
-  throttle("site_hits", limit: 100, period: 2) do |request|
+  throttle("site_hits", limit: 20, period: 2) do |request|
```
benhalpern, line 25: In going to the site I feel like it's a legit situation where a user could visit the home page and then click into an article within a second, and you'd pass 20 hits this way. So maybe cautiously bump this to 40 in two seconds, which is an improvement from 100 but still has a margin of safety for legit traffic.

mstruve: Can do!

PR #7773, opened 2020-05-11
Only Allow Direct Channels Between 2 Different People

mstruve:
- [x] Bug Fix

## Description
Somehow we are allowing people to create chat channels with themselves. When this happens we break our serialization bc the "other_user" is not present. This prevents chat channels for a single user from being created. Their ChatChannelMemberships would be invalid anyways bc we enforce uniqueness between user_id and channel_id

Fixes: https://app.honeybadger.io/fault/66984/436a2859b03e24cb43dda4036b498ab9

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/models/chat_channel.rb
+++ b/app/models/chat_channel.rb
@@ -59,7 +59,7 @@ class ChatChannel < ApplicationRecord
 
   class << self
     def create_with_users(users:, channel_type: "direct", contrived_name: "New Channel", membership_role: "member")
-      raise "Invalid direct channel" if users.size != 2 && channel_type == "direct"
+      raise "Invalid direct channel" if invalid_direct_channel?(users, channel_type)
 
       usernames = users.map(&:username).sort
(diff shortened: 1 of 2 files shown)


app/models/chat_channel.rb
```diff
@@ -91,6 +91,10 @@ def find_or_create_chat_channel(channel_type, slug, contrived_name)
       end
       channel
     end
+
+    def invalid_direct_channel?(users, channel_type)
+      (users.size != 2 || users.map(&:id).uniq.count < 2) && channel_type == "direct"
```
atsmith813, line 96: Should we split up the errors? Maybe create an error (in `./app/errors/` or in the model) like `InvalidDirectChannel`.

And then change the logic to:
```ruby
raise InvalidDirectChannel, "Users are the same"
raise InvalidDirectChannel, "Can't have more than 2 users in a direct channel"
```

mstruve: Since this happens so infrequently I think leaving it is fine for now.

PR #7777, opened 2020-05-11
New Feature: Mod Actions Panel

Zhao-Andy:
- [x] Feature

## Description
This adds the mod actions panel for moderators to use while browsing the site, or in `dev.to/mod`! It features a much cleaner UI, and lays a foundation for adding more features in the future.

## Added tests?
- [x] yes... but needs more

## Added to documentation?
- [ ] update community moderation docs

Michael Scott says, "Oh my God. Okay, it's happening!" In reference to a fire starting in the office.


app/assets/stylesheets/article-show.scss
```diff
@@ -31,6 +31,97 @@
   height: 10px;
 }
 
+.mod-actions-menu {
+  position: fixed;
+  top: var(--header-height);
+  right: 0;
+  height: calc(100vh - 56px);
+  z-index: 200;
```
nickytonline, line 39: Does this really need to be set to 200? If not, consider adding a z-index utility class on the DOM element that has the `.mod-actions-menu` on it, i.e. `z-50` (z-index: 50) or lower if possible.

Othere z-index utilty classes available.

```
z-10
z-20
z-30
z-40
z-50
```

app/javascript/actionsPanel/actionsPanel.js
```diff
@@ -0,0 +1,321 @@
+import { request } from '../utilities/http/request';
```
nickytonline, line 1: You can use the webpack alias here
```suggestion
import { request } from '@utilities/http';
```

app/javascript/actionsPanel/actionsPanel.js
```diff
@@ -0,0 +1,321 @@
...
+
+function addCloseListener() {
+  const button = document.querySelector('.close-actions-panel');
+  button.addEventListener('click', () => {
+    // getting the article show page document because this is called within an iframe
+    const articleDocument = window.parent.document;
```
nickytonline, line 7: You can use `top` instead. See https://developer.mozilla.org/en-US/docs/Web/API/Window/top
```suggestion
    const articleDocument = top.document;
```

app/javascript/actionsPanel/actionsPanel.js
```diff
@@ -0,0 +1,321 @@
...
+    .getElementById('adjustment-reason-container')
+    .classList.toggle('hidden');
+}
+
+function clearAdjustmentReason() {
+  document.getElementById('tag-adjustment-reason').value = '';
+}
+
+function renderTagOnArticle(tagName, colors) {
+  const articleTagsContainer = window.parent.document.getElementsByClassName(
```
nickytonline, line 123:
```suggestion
  const articleTagsContainer = top.document.getElementsByClassName(
```

app/javascript/actionsPanel/actionsPanel.js
```diff
@@ -0,0 +1,321 @@
...
+          el.value = '';
+        }
+
+        toggleSubmitContainer();
+        clearAdjustmentReason();
+
+        if (json.result === 'addition') {
+          renderTagOnArticle(adjustedTagName, json.colors);
+        } else {
+          const tagOnArticle = window.parent.document.querySelector(
```
nickytonline, line 173:
```suggestion
          const tagOnArticle = top.document.querySelector(
```

app/javascript/packs/loadActionsPanelButton.js
```diff
@@ -0,0 +1,7 @@
+import { initializeActionsPanel } from '../actionsPanel/initializeActionsPanelToggle';
+import { initializeFlagUserModal } from './flagUserModal';
+
+if (!window.parent.document.location.pathname.endsWith('/mod')) {
```
nickytonline, line 4:
```suggestion
if (!top.document.location.pathname.endsWith('/mod')) {
```

app/assets/stylesheets/article-show.scss
```diff
@@ -31,6 +31,97 @@
...
+    box-shadow: -4px 0px 15px rgba(0, 0, 0, 0.1);
+    width: 360px;
+  }
+}
+
+.mod-actions-menu-btn {
+  position: fixed;
+  bottom: 30px;
+  right: 30px;
+  z-index: 201;
```
nickytonline, line 58: Same question about z-index here as well.

app/assets/stylesheets/moderators.scss
```diff
@@ -155,8 +159,446 @@
...
+  }
+}
+
 .mod-container {
-  padding: 30px 0px;
+  display: flex;
+  flex-direction: column;
+  background-color: var(--base-0);
+  border: none;
+  box-shadow: none;
```
nickytonline, line 177: Is there another CSS class in a parent element that has `border` and `box-shadow` set which is why you need to set them to `none`?

app/assets/stylesheets/moderators.scss
```diff
@@ -155,8 +159,446 @@
...
+    &:hover .reaction-button-circle {
+      border-color: var(--accent-brand);
+    }
+  }
+
+  .reaction-button-circle {
+    position: relative;
+    display: flex;
+    align-items: center;
+    justify-content: center;
```
nickytonline, line 258: See https://developer.mozilla.org/en-US/docs/Web/CSS/place-items

```suggestion
    place-items: center;
```

app/assets/stylesheets/moderators.scss
```diff
@@ -155,8 +159,446 @@
...
+    .reaction-checkmark {
+      opacity: 0;
+      position: absolute;
+      bottom: 0;
+      right: 0;
+      width: var(--su-5);
+      height: var(--su-5);
+      display: flex;
+      justify-content: center;
+      align-items: center;
```
nickytonline, line 279:
```suggestion
    place-items: center;
```

app/assets/stylesheets/moderators.scss
```diff
@@ -155,8 +159,446 @@
...
+    font-size: var(--fs-base);
+    line-height: var(--lh-base);
+    margin-top: var(--su-2);
+    color: var(--base-90);
+  }
+
+  .reaction-vomit-button {
+    display: flex;
+    justify-content: center;
+    align-items: center;
```
nickytonline, line 299:
```suggestion
    place-items: center;
```

app/assets/stylesheets/moderators.scss
```diff
@@ -155,8 +159,446 @@
...
+            font-weight: var(--fw-normal);
+          }
+        }
+      }
+      .toggle-chevron-container {
+        width: 16px;
+        height: 16px;
+        display: flex;
+        justify-content: center;
+        align-items: center;
```
nickytonline, line 436:
```suggestion
    place-items: center;
```

app/assets/stylesheets/moderators.scss
```diff
@@ -155,8 +159,446 @@
...
+          background: rgba(59, 73, 223, 0.05);
+        }
+
+        .level-rating-content {
+          display: flex;
+
+          .level-rating-number {
+            display: flex;
+            justify-content: center;
+            align-items: center;
```
nickytonline, line 554:
```suggestion
    place-items: center;
```

app/assets/stylesheets/article-show.scss
Zhao-Andy: Hmm sorry how I would do this? I tried `z-index: var(--z-10)` but that didn't seem to work. Also looked at `_generator.scss` but just got confused because my front end game is weak :(

app/javascript/actionsPanel/actionsPanel.js
```diff
@@ -0,0 +1,321 @@
...
+      const {
+        reactableType: reactable_type,
+        category,
+        reactableId: reactable_id,
+      } = butt.dataset;
+
+      correctReactedClasses(category);
+      butt.classList.toggle('reacted');
+
+      request('/reactions', {
```
nickytonline, line 66: If this errors out for whatever reason, a friendly error message should be displayed to the user. The same goes for any other `request`s.

app/javascript/actionsPanel/actionsPanel.js
```diff
@@ -0,0 +1,321 @@
...
+      .json()
+      .then((json) => {
+        if (json.result === 'Success') {
+          clearExpLevels();
+          document
+            .getElementById(`js__rating__vote__${rating}`)
+            .classList.add('selected');
+        } else {
+          // eslint-disable-next-line no-alert
+          alert(json.error);
```
nickytonline, line 103: As you mentioned, you will move to the snackbar component in a future PR.

app/javascript/actionsPanel/actionsPanel.js
```diff
@@ -0,0 +1,321 @@
...
+
+  articleTagsContainer.appendChild(newTag);
+}
+
+function adjustTag(el) {
+  const reasonForAdjustment = document.getElementById('tag-adjustment-reason')
+    .value;
+  const body = {
+    tag_adjustment: {
+      // TODO: change to tag ID
```
nickytonline, line 143: Is this a TODO for this PR or future PR?

app/javascript/actionsPanel/actionsPanel.js
```diff
@@ -0,0 +1,321 @@
...
+        el.dataset.adjustmentType === 'subtract' ? 'removal' : 'addition',
+      reason_for_adjustment: reasonForAdjustment,
+    },
+  };
+
+  request('/tag_adjustments', {
+    method: 'POST',
+    body: JSON.stringify(body),
+  })
+    .then((response) => response.json())
```
nickytonline, line 156: Using `.then` is fine, but in case you weren't aware, you also can use async/await now, e.g. https://github.com/thepracticaldev/dev.to/blob/master/app/javascript/CommentSubscription/commentSubscriptionUtilities.jsx#L10

app/javascript/packs/flagUserModal.js
```diff
@@ -0,0 +1,158 @@
+import { request } from '../utilities/http';
+
+const modalSnackbarHTML = `
```
nickytonline, line 3: As discussed this afternoon, is this code being used at all, or do the existing alerts you have  cover the "ghetto" snackbar items for now. If so, feel free to remove this code.

app/javascript/packs/flagUserModal.js
```diff
@@ -0,0 +1,158 @@
...
+
+export function initializeFlagUserModal() {
+  // eslint-disable-next-line no-undef
+  const user = userData();
+  const {
+    authorId: articleAuthorId,
+    authorClassName: articleAuthorClassName,
+  } = document.getElementById('article-show-container').dataset;
+
+  const flagUserModalHTML = `
```
nickytonline, line 42: In a future PR, it would be nice to use the Preact `<Modal />` component instead, but at the moment, it is still under review.

app/javascript/packs/flagUserModal.js
```diff
@@ -0,0 +1,158 @@
...
+      } else {
+        toggleFlagUserModal();
+      }
+    });
+
+  document
+    .getElementById('confirm-flag-user-action')
+    .addEventListener('click', (e) => {
+      e.preventDefault();
+      const vomitAllOption = document.getElementById('vomit-all');
```
nickytonline, line 127: You can use the [:checked(https://developer.mozilla.org/en-US/docs/Web/CSS/:checked) pseudo class.
```suggestion
      const vomitAllOption = document.querySelector('vomit-all:checked');
```

app/javascript/packs/flagUserModal.js
```diff
@@ -0,0 +1,158 @@
...
+      const vomitAllSnackbar = document.getElementById('vomit-all-snackbar');
+
+      if (vomitAllOption.checked) {
```
nickytonline, line 130: This change is required if you go with the `:checked` pseudo-selector above.
```suggestion
      if (vomitAllOption != null) {
```

app/views/articles/show.html.erb
```diff
@@ -1,7 +1,7 @@
 <% title @article.title %>
 
 <%= render "shared/webcomponents_loader_script" %>
-<%= javascript_packs_with_chunks_tag "clipboardCopy", "webShare", "articlePage", defer: true %>
+<%= javascript_packs_with_chunks_tag "clipboardCopy", "webShare", "articlePage", "loadActionsPanelButton", defer: true %>
```
nickytonline, line 4: Consider renaming the pack to `articleModerationTools` or something more descriptive.

app/views/moderations/index.html.erb
```diff
@@ -80,7 +80,7 @@
       if (el.innerHTML.length > 10) {
```
nickytonline, line 80: Just curious what this logic means. Consider adding a comment explaining why the `innerHTML` gets cleared out if it's length is > 10.

app/controllers/reactions_controller.rb
```diff
@@ -53,7 +55,17 @@ def create
 
     Rails.cache.delete "count_for_reactable-#{params[:reactable_type]}-#{params[:reactable_id]}"
 
+    if params[:reactable_type] == "Article"
+      clear_article_reactions(
```
rhymes, line 59: I don't understand this part, why is the code removing all existing reactions?

app/controllers/reactions_controller.rb
```diff
@@ -133,6 +139,27 @@ def rate_article(reaction)
                       rating: current_user.experience_level)
   end
 
+  def clear_article_reactions(id, type, mod, category)
+    reactions = if category == "thumbsup"
+                  Reaction.where(reactable_id: id, reactable_type: type, user: mod).where.not(category: category)
+                elsif category.in?(NEG_ARTICLE_REACTIONS)
+                  Reaction.where(reactable_id: id, reactable_type: type, user: mod, category: "thumbsup")
+                end
+    reactions.each { |reaction| destroy_reaction(reaction) }
```
rhymes, line 148: Two questions (see my previous question about understanding the purpose):

1. can we use `reactions.find_each` to only load a few reactions at the time
2. can all of this be executed asynchronously?

db/schema.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
rhymes, line 1: as there are no migration files, I think this schema file should be removed.

app/assets/images/book.svg
```diff
@@ -0,0 +1 @@
+<svg width="18" height="20" xmlns="http://www.w3.org/2000/svg"><path d="M18 2H4a2 2 0 100 4h14v13a1 1 0 01-1 1H4a4 4 0 01-4-4V4a4 4 0 014-4h13a1 1 0 011 1v1zM2 16a2 2 0 002 2h12V8H4a3.982 3.982 0 01-2-.535V16zM17 5H4a1 1 0 010-2h13v2z"/></svg>
```
ludwiczakpawel, line 1: our glyphs should be 24 x 24px and the SVG tag should contain `viewBox` attribute as well.

app/assets/images/chevron-right.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M13.172 12L8.22198 7.04999L9.63598 5.63599L16 12L9.63598 18.364L8.22198 16.95L13.172 12Z" fill="#08090A"/>
```
ludwiczakpawel, line 2: I think you can drop both `fill` attributes since we overwrite them anyway.

app/assets/images/chevron-toggle.svg
```diff
@@ -0,0 +1 @@
+<svg width="12" height="6" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M6 6L0 0h12L6 6z" fill="#363D44"/></svg>
```
ludwiczakpawel, line 1: Same here in terms of sizing and fill attribute.

app/assets/images/mod-filled.svg
```diff
@@ -0,0 +1 @@
+<svg width="24" height="24" xmlns="http://www.w3.org/2000/svg"><path d="M3.783 2.826L12 1l8.217 1.826a1 1 0 01.783.976v9.987a6 6 0 01-2.672 4.992L12 23l-6.328-4.219A6 6 0 013 13.79V3.802a1 1 0 01.783-.976zM13 10V5l-5 7h3v5l5-7h-3z"/></svg>
```
ludwiczakpawel, line 1: `viewBox` attribute missing

app/assets/images/chevron-toggle.svg
ludwiczakpawel: Same for other SVGs in this PR.

app/assets/stylesheets/article-show.scss
```diff
@@ -31,6 +31,97 @@
   height: 10px;
 }
 
+.mod-actions-menu {
+  position: fixed;
+  top: var(--header-height);
+  right: 0;
+  height: calc(100vh - 56px);
```
ludwiczakpawel, line 38: If `56px` is coming from header's height you could do this:

```
height: calc(100vh - var(--header-height));
```

...or your can do:

```
top: var(--header-height);
bottom: 0;
//drop the height.
```

app/assets/stylesheets/article-show.scss
```diff
@@ -31,6 +31,97 @@
...
+    border: none;
+    box-shadow: -4px 0px 15px rgba(0, 0, 0, 0.1);
+    width: 360px;
+  }
+}
+
+.mod-actions-menu-btn {
+  position: fixed;
+  bottom: 30px;
+  right: 30px;
```
ludwiczakpawel, line 57: Can we go with `var(--su-7)` which will be equal to `32px`?

app/assets/stylesheets/article-show.scss
```diff
@@ -31,6 +31,97 @@
...
+
+  iframe {
+    height: 100%;
+    border: none;
+    box-shadow: -4px 0px 15px rgba(0, 0, 0, 0.1);
+    width: 360px;
+  }
+}
+
+.mod-actions-menu-btn {
```
ludwiczakpawel, line 54: Why not using one of our existing buttons?

```
<button class="crayons-btn crayons-btn--icon-rounded crayons-btn--l"><svg ...></button>
<!-- ^that's also available as preact component -->
```

app/assets/stylesheets/article-show.scss
```diff
@@ -31,6 +31,97 @@
...
+  position: fixed;
+  bottom: 30px;
+  right: 30px;
+  z-index: 201;
+  cursor: pointer;
+  transition: 0.3s all ease;
+  background-color: var(--base-90);
```
ludwiczakpawel, line 61: We do have quite a lot of styles to choose from for buttons. I know we do not have the color that is defined here, but I think we could use one of the existing colors here so A) you don't have to worry about styling and B) I think we this code you may have some issues on other themes, no?

cc @lisasy

app/views/articles/show.html.erb
```diff
@@ -224,6 +225,11 @@
   <% end %>
 <% end %>
 
+<div class="mod-actions-menu"></div>
+<div id="mod-actions-menu-btn-area"></div>
+<div class="flag-user-modal-container hidden"></div>
+<div class="modal-snackbar-container"></div>
+
```
ludwiczakpawel, line 232: Feels like this should be behind some IF statement for mods only? Also, is this a good practice to have these kind of empty containers/placeholders anyway?

Also, would that conflict anyhow with snackbar we already have on that view? cc @nickytonline

app/javascript/actionsPanel/actionsPanel.js
```diff
@@ -0,0 +1,321 @@
...
+  if (type === 'set-experience') {
+    document
+      .querySelector('.set-experience-options')
+      .classList.toggle('hidden');
+  } else if (type === 'adjust-tags') {
+    document.querySelector('.adjust-tags-options').classList.toggle('hidden');
+  }
+}
+
+function correctReactedClasses(category) {
```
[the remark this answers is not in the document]
msarit: I wrote this function Nick and I agree the name could be more descriptive 🙈 
This function applies the `reacted` class to the reactions based on what makes sense
Only `thumbsup` should be highlighted if it is clicked

app/javascript/actionsPanel/actionsPanel.js
msarit: I agree, thanks!

app/javascript/actionsPanel/actionsPanel.js
msarit: Yes! 😄

app/javascript/actionsPanel/actionsPanel.js
msarit: Thank you. This is an area I need to get better at 👍

app/javascript/packs/flagUserModal.js
msarit: Ok!

app/javascript/packs/flagUserModal.js
msarit: Thank you!

app/javascript/packs/flagUserModal.js
msarit: Ok!

app/assets/stylesheets/moderators.scss
Zhao-Andy: Hmm I needed this in the beginning but after testing with the latest changes maybe I don't... Will revisit.

app/javascript/actionsPanel/actionsPanel.js
```diff
@@ -0,0 +1,321 @@
...
+      .classList.toggle('showing');
+    articleDocument
+      .querySelector('.mod-actions-menu-btn')
+      .classList.toggle('hidden');
+  });
+}
+
+function initializeHeight() {
+  document.documentElement.style.height = '100%';
+  document.body.style.cssText = 'height: 100%; margin: 0; padding-top: 0;';
```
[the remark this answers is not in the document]
Zhao-Andy: Yeah, so because the actions panel is loaded via iframe we need to do some JS + CSS hacking here to configure the height of the page. Otherwise the actions panel doesn't take up the full page of the article show page.

app/javascript/actionsPanel/actionsPanel.js
Zhao-Andy: Yep, TODO for a future PR

app/assets/stylesheets/article-show.scss
Zhao-Andy: I think this was before we had the buttons in, will update 👍

app/assets/stylesheets/article-show.scss
Zhao-Andy: Hmm interesting, will test this out

app/controllers/reactions_controller.rb
msarit: @rhymes, the idea behind this was to prevent the selection of both the `upVote` reaction and either of the `downVote` or `vomit` reactions.

In the view, the class `reacted` is added to a reaction dynamically if it has any cached reactions (see `app > views > moderations > actions_panel`):
`<button class="reaction-button <%= Reaction.cached_any_reactions_for?(@moderatable, current_user, "thumbsup") ? "reacted" : "" %>"></button>`

For the UI to be accurate, we needed to clear the reactions before adding the `reacted` class to appropriate buttons in the UI

Hope this makes sense?

app/controllers/reactions_controller.rb
msarit: re: Question 1 - I will do so 👍 
re: Question 2 - probably so! 😄 I think this goal could be part of a future PR though

app/assets/stylesheets/article-show.scss
Zhao-Andy: Hmm so it would be a bit bigger if I used the smallest crayons button... here's a preview, with the original being on the left:


@lisasy what do you think, regarding the sizing and background color?

app/controllers/reactions_controller.rb
Zhao-Andy: TBH I think this should only be destroying two reactions at most, since we're looking for either a thumbs down + vomit reaction from the mod that is reacting.

app/views/moderations/index.html.erb
Zhao-Andy: Hmm I think this was written by someone else but I _think_ this is to clear the innerHTML if the iframe is loaded. Will add a comment here.

app/javascript/packs/flagUserModal.js
Zhao-Andy: This is still currently being used, so we'll keep it for now. Going to try to refactor and use the snackbar component but it's more of a stretch goal at this point.

app/javascript/packs/flagUserModal.js
nickytonline: Ahh, didn't notice that. In that case, ignore my suggestion @msarit.

app/javascript/actionsPanel/actionsPanel.js
msarit: this has been done 👍

app/views/articles/show.html.erb
msarit: done! 👍

app/controllers/reactions_controller.rb
rhymes: Got it! Thanks. What happens if we accidentally post `like` or `readinglist` as a category to this action?

app/controllers/reactions_controller.rb
rhymes: Cool! I thought we were removing all reactions :D

app/views/articles/show.html.erb
Zhao-Andy: Tested the subscribe button, doesn't seem to affect the snackbar there 👍 

I agree that we should have some if statement here, but not sure how we would hide it behind that statement since we can't check for `current_user` because of caching. 😕 I think for now it's okay to have this here and then move it later once we refactor to using the modal + snackbar from Preact.

app/controllers/reactions_controller.rb
msarit: Great question! I will investigate.

Zhao-Andy: Sorry @msarit, I updated this! I added a check for the category type just to make the code clearer. That said, posting a `like` or `readinglist` wouldn't have done anything with the other reactions so I think we're okay here :)

app/assets/stylesheets/article-show.scss
lisasy: I think using the "Crayons" button with the larger size (the one on the right) is a good approach. Thanks for looking me in @Zhao-Andy and @ludwiczakpawel

benhalpern:
The functionality is really beautiful. I think this is going to be really useful!

Given the choice to make this functionality work as a new type of positive reaction, there will need to be some additional behavior changes because this is intended to be a private reaction.

If I leave a `thumbsup` reaction it still does the normal _public_ stuff because we differentiate by positive/negative for this count and functionality....

Screen Shot 2020-05-13 at 2 17 09 PM
Screen Shot 2020-05-13 at 2 17 03 PM

I think the `counter_culture` logic as well as the notifications logic will need to be adjusted now that positive/negative is not synonymous with public/private.

It's probably too extreme in this PR to actually rename `positive_reactions_count` to `public_reactions_count` but I'd say that that is how we'd need to think about it, and perhaps make that change in another PR. Even though the naming would be off I think we need to make the functionality adjustment to reflect that.

Does that make sense?

Zhao-Andy:
@benhalpern that all makes sense! I can make a separate PR to update the `counter_culture` column.

To push this to the finish line, I think:
1. merge the additional PR
2. merge it into master
3. update this branch with master

benhalpern:
Are we thinking that `/mod` is officially deprecated because of this?

I see it kind of messes that page up.

Screen Shot 2020-05-13 at 2 31 23 PM

Zhao-Andy:
@benhalpern Yeah, in terms of moderators viewing `/article_path/mod` through any clicks, they won't really be able to. Viewing `dev.to/mod` they should see the actions panel instead now.

I think for our purposes as admins `/article_path/mod` is still useful since the admin functionality (internal article/user link, etc) hasn't been moved over to the actions panel yet.

benhalpern:
@Zhao-Andy Cool. For now we could just hack-job remove some stuff from `article-view/mod` and in the future I think it could make sense if admin actions/links were also accessible via this sidecar (maybe a tab ala how we do mobile notification views etc if the user has those permissions.)

app/views/articles/show.html.erb
```diff
@@ -1,7 +1,7 @@
 <% title @article.title %>
 
 <%= render "shared/webcomponents_loader_script" %>
-<%= javascript_packs_with_chunks_tag "clipboardCopy", "webShare", "articlePage", defer: true %>
+<%= javascript_packs_with_chunks_tag "clipboardCopy", "webShare", "articlePage", "articleModerationTools", defer: true %>
```
benhalpern, line 4: We could hide this behind a `user_signed_in?` flag because we don't need this for non-logged-in traffic.

app/javascript/packs/flagUserModal.js
```diff
@@ -0,0 +1,148 @@
...
+      '.crayons-modal__overlay, .modal-header-close-icon, #cancel-flag-user-action',
+    ),
+  );
+
+  closeModalElements.forEach((element) => {
+    element.addEventListener('click', toggleFlagUserModal);
+  });
+
+  document
+    .getElementById('confirm-flag-user-action')
```
benhalpern, line 94: I'm getting `Uncaught TypeError: Cannot read property 'addEventListener' of null` when I'm not logged in as `trusted`, so we'll need a guard for this. We could _probably_ add that guard much higher up and use [dynamic import](https://dev.to/ben/we-reduced-our-vendorjs-from-210kb-to-16kb-in-about-five-minutes-of-work-and-ten-lines-of-code-1ole) to access this file at all, eh?

msarit: Will do! 😄

app/views/articles/show.html.erb
msarit: Done! 😄

msarit:
`/article-slug/mod` has been modified to work with the other implementations 👍
@benhalpern @Zhao-Andy

benhalpern:
I think the vomit button should probably say "Mark as Abuse/Spam" or something because it's for abuse like harassment and other CoC violations as well as spam.

Also, what's the story on comment moderation. I see it currently opens to the `comment/mod` endpoint, would this eventually also trigger the sidecar?

Zhao-Andy:
> Also, what's the story on comment moderation. I see it currently opens to the comment/mod endpoint, would this eventually also trigger the sidecar?

We haven't discussed that part out yet but I think triggering the actions panel makes sense. I'll bring this up in our snackCase meeting today.

thomasbnt:
Awesome :+1: :o 
I check PR sometimes, and I think there are some good proposals. That's great!

wow_ted

Zhao-Andy:
Tests are passing but build is failing 🤔 not sure why.

app/controllers/internal/privileged_reactions_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Internal::PrivilegedReactionsController < Internal::ApplicationController
+  layout "internal"
+
+  PRIVILEGED_REACTION_CATEGORIES = %i[thumbsup thumbsdown vomit].freeze
```
rhymes, line 4: We should move tihs in the model and use it everywhere, same as `ReactionsController::MODERATION_CATEGORIES`

in another PR though :D

app/javascript/packs/articlePage.jsx
```diff
@@ -6,6 +6,9 @@ const snackZone = document.getElementById('snack-zone');
 
 render(<Snackbar lifespan="3" />, snackZone, snackZone.firstElementChild);
 
+// eslint-disable-next-line no-restricted-globals
+top.addSnackbarItem = addSnackbarItem;
```
rhymes, line 10: what is this hack for?

app/javascript/packs/flagUserModal.js
```diff
@@ -0,0 +1,137 @@
...
+      <h2>Flag User</h2>
+      <button type="button" class="crayons-btn crayons-btn--icon crayons-btn--ghost modal-header-close-icon">
+        <svg width="24" height="24" viewBox="0 0 24 24" class="crayons-icon" xmlns="http://www.w3.org/2000/svg">
+          <path d="M12 10.586l4.95-4.95 1.414 1.414-4.95 4.95 4.95 4.95-1.414 1.414-4.95-4.95-4.95 4.95-1.414-1.414 4.95-4.95-4.95-4.95L7.05 5.636l4.95 4.95z" />
+        </svg>
+      </button>
+    </header>
+    <div class="crayons-modal__box__body flag-user-modal-body">
+      <span>
+        Thanks for keeping DEV safe. Here is what you can do to flag this user:
```
rhymes, line 17: this should be generalized using `community_name` instead of `DEV`. it's a bit weird to have this much HTML in JavaScript. I would move it in a controller action and load it from JS

app/models/reaction.rb
```diff
@@ -4,7 +4,7 @@ class Reaction < ApplicationRecord
   SEARCH_SERIALIZER = Search::ReactionSerializer
   SEARCH_CLASS = Search::Reaction
 
-  CATEGORIES = %w[like readinglist unicorn thinking hands thumbsdown vomit].freeze
+  CATEGORIES = %w[like readinglist unicorn thinking hands thumbsup thumbsdown vomit].freeze
```
rhymes, line 7: this is an example of why you don't want to have enums in the DB after all @atsmith813  :D 
Our `reactions` table is probably one of the biggest ones and it would have required a complicated dance of `ALTER TABLE` with such a massive table

app/views/articles/show.html.erb
```diff
@@ -1,7 +1,11 @@
 <% title @article.title_with_query_preamble(user_signed_in?) %>
 
 <%= render "shared/webcomponents_loader_script" %>
-<%= javascript_packs_with_chunks_tag "clipboardCopy", "webShare", "articlePage", defer: true %>
+<% if user_signed_in? %>
+  <%= javascript_packs_with_chunks_tag "clipboardCopy", "webShare", "articlePage", "articleModerationTools", defer: true %>
+<% else %>
+  <%= javascript_packs_with_chunks_tag "clipboardCopy", "webShare", "articlePage", defer: true %>
+<% end %>
```
rhymes, line 8: 🔥

app/views/internal/mods/index.html.erb
```diff
@@ -49,7 +49,7 @@
       <% if params[:state] == "potential" %>
         <td>
           <%= form_with model: [:internal, mod], url: internal_mod_path(mod.id), method: :patch, local: true do |f| %>
-            <%= f.submit "Make Trusted Mod", class: "btn btn-light" %>
+            <%= f.submit "Make Trusted Mod", class: "btn btn-light js-add-to-mod-channel" %>
```
rhymes, line 52: yes, I so agree with @msarit about making classes used by JS explicit with the prefix, it's a game changer to be able to understand where JS behavior comes from. Not always easy and not always necessary but so helpful

app/views/internal/privileged_reactions/index.html.erb
```diff
@@ -2,14 +2,14 @@
...
-    <%= search_form_for @q, url: internal_negative_reactions_path, class: "form-inline justify-content-end" do |f| %>
+    <%= search_form_for @q, url: internal_privileged_reactions_path, class: "form-inline justify-content-end" do |f| %>
 
       <%= f.label :user_username_cont, "User", class: "sr-only" %>
       <%= f.search_field :user_username_cont, placeholder: "User", class: "form-control mx-3" %>
 
       <%= f.select(
             :category_eq,
-            options_for_select(["", "vomit", "thumbsdown"], @q.category_eq),
+            options_for_select(["", "thumbsup", "thumbsdown", "vomit"], @q.category_eq),
```
rhymes, line 12: another place where we could use the constant "moderator reactions" or "privileged reactions"

app/javascript/packs/articlePage.jsx
Zhao-Andy: The actions panel is loaded in an iframe, so to reference the proper `snackbar` we have to assign it to `top` so it's available globally. :(

Won't be problem once we make the actions panel via Preact.

app/javascript/actionsPanel/__tests__/actionsPanel.test.js
```diff
@@ -0,0 +1,236 @@
...
+  describe('when the user is tag moderator of #discuss', () => {
+    beforeEach(() => {
+      const tagName = 'discuss';
+      document.body.innerHTML = `
+      <a href="/t/${tagName}" class="tag">${tagName}</a>
+      <button class="adjustable-tag" data-adjustment-type="subtract" data-tag-name="${tagName}">
+        #${tagName}
+      </button>
+      <form id="adjustment-reason-container" class="adjustment-reason-container hidden">
+        <textarea class="crayons-textfield" placeholder="Reason for tag adjustment" id="tag-adjustment-reason" required></textarea>
```
maestromac, line 164:
```suggestion
        <textarea class="crayons-textfield" placeholder="Reason for tag adjustment (required)" id="tag-adjustment-reason" required></textarea>
```

I tried submitted tag changes and wasn't aware this is was required. This minor change would help.

msarit: Thanks!

benhalpern:
Hey @Zhao-Andy in a final pass I noticed an issue that needs to be changed...

Screen Shot 2020-06-08 at 10 38 41 AM

Our actual rating system is out of 10, so if we're only displaying five options we need to essentially double the scores. Currently it looks like we just submit the exact value?

I'd say changing to them to equal: `1`, `3`, `5`, `8`, `10`. (which is the same scheme we use on the ux settings)

Zhao-Andy:
Build failed b/c of flaky test 😡

PR #7782, opened 2020-05-11
Raise an Error If Trying to Rate Limit with a Missing User ID

mstruve:
- [x] Save us from ourselves Feature

## Description
To prevent generic cache keys from incorrectly being made if an unpersisted user is passed to the RateLimitChecker we will raise an error if the user ID is not present when creating the cache key.

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/services/rate_limit_checker.rb
+++ b/app/services/rate_limit_checker.rb
@@ -66,6 +66,8 @@ class RateLimitChecker
   end
 
   def limit_cache_key(action)
+    raise "Invalid Cache Key: user ID can't be blank" unless @user.id
+
     "#{@user.id}_#{action}"
   end
 
--- a/spec/services/rate_limit_checker_spec.rb
+++ b/spec/services/rate_limit_checker_spec.rb
@@ -14,8 +14,14 @@ RSpec.describe RateLimitChecker, type: :service do
       expect(rate_limit_checker.limit_by_action("random-nothing")).to be(false)
     end
 
+    it "raises an error if user does not have an ID" do
+      action = described_class::ACTION_LIMITERS.keys.first
+      limiter = described_class.new(build(:user))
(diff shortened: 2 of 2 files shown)


spec/services/rate_limit_checker_spec.rb
```diff
@@ -14,8 +14,14 @@ def cache_key(action)
       expect(rate_limit_checker.limit_by_action("random-nothing")).to be(false)
     end
 
+    it "raises an error if user does not have an ID" do
+      action = described_class::ACTION_LIMITERS.keys.first
+      limiter = described_class.new(build(:user))
```
vaidehijoshi, line 19: lol this is one of those few times when `build` is truly the best tool for the job! 💪

app/services/rate_limit_checker.rb
```diff
@@ -66,6 +66,8 @@ def limit_by_email_recipient_address(address)
   end
 
   def limit_cache_key(action)
+    raise "Invalid Cache Key: user ID can't be blank" unless @user.id
```
citizen428, line 69: Quick question, how does this happen in the actual app? I can see `@user` being `nil`, but under that circumstances would we hit the case of the user being present but not having an ID (which is the primary key)?

rhymes: Curious as well 👍

mstruve: I discovered this problem when I implemented a rate limit on email confirmations #7780 and added it to a validate which happens when a user is created which means we pass a new user object to the rate limiter.

citizen428: I see, thanks for clarifying.

The way I most commonly see this check performed is with `@user.persisted?` since that makes it more obvious that we're dealing with a new record.

PR #7783, opened 2020-05-11
Add Percy 🦔

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

**WHAT**
This PR introduces a proof-of-concept for what visual testing would look like for our app with [Percy](https://percy.io/).

Visual testing is a way of automatically detecting visual changes in our UI. Percy is a service that allows you to add "snapshots", which is a rendering of a webpage or component. The snapshot can be run against different page sizes and browsers (Chrome + Firefox). The snapshots of our `master` build are the baseline for the correct, "approved" snapshots. Every PR will (eventually) run its own snapshots against the `master` build snapshots. If a visual change is detected, the branch's snapshot appears as a "diff" that needs to be approved by someone as part of the review.

**WHY**
The idea behind adding Percy is that we will be able to:
1) detect any visual bugs introduced by a PR and,
2) quickly QA for any visual changes introduced in a PR (which we can "approve" if they are expected changes!).

**HOW**
This PR itself won't do anything to get Percy working with our CI; we'll need to add the token and [integrate it with GitHub](https://docs.percy.io/docs/github) + integrate with [Travis CI](https://docs.percy.io/docs/travis-ci). I can work on doing that if this PR is acceptable to everyone!

**FINE PRINT**
1) When you look at some of the Percy snapshots, you'll notice that some of them look a little funny (they might have content missing or might have the footer in the middle of the page). This is because either the test is missing some setup that we can now see pretty clearly in the snapshot _or_ there are some CSS-related changes that need to happen for Percy, specifically. I didn't bother doing anything to change any of the tests since this is a proof-of-concept PR. We can definitely improve our snapshots in the future, though!
2) The snapshots represented in Percy are _only_ the Rails Capybara tests, and do not represent our JavaScript/component tests. We probably _do_ want to add snapshots for those, but that will take some more learning on my part, and again, since this is proof-of-concept, I didn't want to spend time doing that without knowing that we'd actually be implementing Percy. If we are 👍 with this, I will add snapshots to our JS tests in a separate PR. 😃

I've temporarily made the (free) Percy account that I set up for DEV open source, which means that the snapshots are viewable to anyone. You'll want to look at [this build](https://percy.io/thepracticaldev/dev.to/builds/5229721) to see the relevant snapshots for this proof-of-concept PR.

N/A

## Added tests?

- [x] yes (ALL THE VISUAL TESTING! 🔥)
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed
- [x] I will create documentation for this if we go ahead with it!

See the "HOW" in the description!

hedgehog


Gemfile
```diff
@@ -67,6 +67,7 @@ gem "omniauth", "~> 1.9" # A generalized Rack framework for multiple-provider au
 gem "omniauth-github", "~> 1.3" # OmniAuth strategy for GitHub
 gem "omniauth-twitter", "~> 1.4" # OmniAuth strategy for Twitter
 gem "patron", "~> 0.13.3" # HTTP client library based on libcurl, used with Elasticsearch to support http keep-alive connections
+gem "percy-capybara", "~> 4.0.0"
```
rhymes, line 70: I think this should be moved down to the `:test` group

config/sample_application.yml
```diff
@@ -12,3 +12,5 @@ PUSHER_CLUSTER:
 TWITCH_CLIENT_ID:
 TWITCH_CLIENT_SECRET:
 TWITCH_WEBHOOK_SECRET:
+
+PERCY_TOKEN:
```
rhymes, line 16: is this needed only during CI testing? If so, we probably don't need it here

(ps. also missing a new line here)

Gemfile
rhymes: ps. don't we need the percy agent as well? https://docs.percy.io/docs/capybara#setup

joshpuetz:
I love this! A few requests (these might already be true, I haven't dug into how the Percy gem handles it yet)
- this should only work for us when the token is present. I wouldn't expect open source contributors to have to open a Percy account.
- I see Percy has a free tier: do they offer free pricing for Open Source projects?

config/sample_application.yml
vaidehijoshi: Ah, yep; it's only ever necessary for CI (unless you needed to debug Percy and run it locally, which you probably won't be doing unless you know what you're doing :)

I guess I added that a bit eagerly. I have removed it now 👍

Gemfile
vaidehijoshi: Good call on the `:test` group! I honestly don't know how this worked _without_ me adding the percy agent...? Maybe because I was running it locally 🤔 But yes, you are right that I need it and I have no idea why it worked but OH WELL, I have added it!

vaidehijoshi:
@rhymes Great questions! Here are my answers 😸 

> 1. how can we run this locally if we need a token? same thing for external contributors
You definitely need a token to run this locally; however, I would say that, for the most part, you don't want to run Percy locally. For both our team and for external contributors, we'd want CI to handle this. Eventually, we'll want to [integrate this with GitHub](https://docs.percy.io/docs/github) such that checking Percy snapshots is akin to CodeClimate. 

Another reason we _don't_ want people to run Percy locally is that we'd likely end up with a skyrocketing number of builds, which means a lot of snapshots and a lot of $$$! So, I would say that only DEV core team folks should even have access to the Percy organization and be able to "approve" or "request changes" on Percy snapshots (if there are no changed snapshots, the build is automatically approved!).

If, for some reason, they needed to run Percy locally to try to debug a very weird situation (in my past experience, this is pretty rare!), they could use the token and be cautious about how many snapshots they were creating. 

> 2. how do we keep the snapshots "clean" and ordered? I see you've named them all, but a team member or a contributor could either change the name or add a different name for essentially the same screenshot and we'd end up with multiple similar screenshots. Will this be a potential problem? Do snapshots "expire" ? Should we have a centralized catalogue of snapshots?

Provided we decide to adopt Percy as a tool, we will set the "baseline" of our approved snapshots on `master`. Eventually, instead of running against my local branch, I'll merge this onto master and then will manually click "Approve" on the snapshots that look correct, which will make them our "baseline" build:
Screen Shot 2020-05-13 at 9 17 52 AM

You can read a bit more about how Percy determines the baseline for a build [here](https://docs.percy.io/docs/baseline-picking-logic). The base build of master effectively will become our "centralized catalogue of snapshots" :)

As far as keeping snapshots clean, it's up to us to make sure we follow Percy's guidelines for how a snapshot should be named (they should always be unique per snapshot), and it's up to us as reviewers to look at the Percy build of a contributors PR and make sure that they haven't overwritten a pre-existing snapshot (aka used a name for a snapshot that already exists).

I was planning on writing up a short document on [how to approach Percy](https://docs.percy.io/docs/percy-platform-basics#what-you-should-know-about-screenshots-snapshots-and-diffs) for both ourselves, internally, and for external contributors (but I figured it didn't make sense to do this until we decide we want to actually adopt Percy!). For example, if we rename a snapshot, Percy will interpret this as "one snapshot from a previous build was removed and a new one was added to this build". I think this stuff will be easier to grasp once we all start using Percy.

> 3. is there a potential speed impact if we add too many ? Though I checked build time and it doesn't seem to be slower than our usual times

Percy runs continuously alongside our test suite on CI and it starts "receiving and rendering screenshots every time a CI build runs." So I don't think it should slow down our builds or impact them...and I haven't had this issue in the past. Their [enterprise page](https://percy.io/enterprise) claims that "Percy is optimized for speed, and because we handle all snapshot rendering, there’s no impact on your existing CI build times".

vaidehijoshi:
@joshpuetz:

> this should only work for us when the token is present. I wouldn't expect open source contributors to have to open a Percy account.

Yep, see my answer to @rhymes above! My thought is that we only invite DEV core team folks to the Percy organization, and we'll be the ones to "approve" or "request changes" on Percy diffs. No external contributor should need a token, nor should they ever run Percy locally. We'll set up CI to trigger Percy on every PR automatically, so if a Percy diff changes as a result of a contributors PR, we can request changes on it and ask them to address anything that is broken or an issue.

> I see Percy has a free tier: do they offer free pricing for Open Source projects?

Yes, they have [a free tier](https://blog.percy.io/introducing-our-free-visual-testing-plan-2cd70a34d89d) and that tier is also the equivalent of the "open source" tier. That tier includes:

* Unlimited team members
* Demo app and unlimited projects
* 5,000 screenshots / month

I'm 100% sure this tier won't be enough for us, though. So we need to decide what we want to do there... (cc @benhalpern). I suspect we could also chat with the Percy folks and see if they'd be open to working something out with us (I'm pretty friendly with them and would be happy to start that conversation).

Relatedly, Percy is also set up for public projects, which is what we'll have to use in order for our external contributors to [view snapshots](https://docs.percy.io/docs/public-projects) from outside of our organization. As they explain, "Viewers who aren’t members of your organization will not have the ability to approve builds or snapshots but will be able to see build information and statuses.".

snackattas:
Hey @vaidehijoshi, this is a pretty cool idea! I have a lot of questions in addition to the ones asked already, sorry for the brain dump of questions! But I think its worthwhile to ask these sorts of things whenever adding a new dimension to testing :)

Okay, so here we go:

1. Before merging this in, it would be good to have a wiki page that answers the questions 'what happens when my Percy screenshot is different and the test fails', 'what happens when I need to change the baseline Percy snapshot (because I changed the page under test)', etc

2. Does Percy deal with dynamic data well? We use Faker in dev.to, and this is Percy’s stance on dynamic data: https://docs.percy.io/docs/freezing-dynamic-data.  It’s interesting that Percy proposes we make Faker deterministic, as opposed to it being able to handle the dynamic data itself. I know other platforms like Applitools handle dynamic data.  Not sure how I feel about this stance. 

3. Does Percy take a long time? We already have tech debt with regards to the speed of tests, wouldn’t want to make tests more slow. It might be useful to wrap the Percy calls around a timer to see how log each call takes.  Maybe doing something like this just for testing
```ruby
RSpec.configure do |config|
  ...
  config.before(:suite) do
    $percy_timings = []
    module Percy
      class << self
        alias_method :original_snapshot, :snapshot
        def snapshot(page, *args)
          start = Time.now
          original_snapshot(page, *args)
          time = (Time.now - start).round(2)
          $percy_timings << {name: args.first[:name], time: time}
        end
      end
    end
  end
  config.after(:suite) do
    puts "============Percy timings"
    puts $percy_timings
  end
  ```


4. What does an actual Percy fail look like? It would be good to show off what a 'real' fail appears like in Percy and in the Travis fail, and determine if it is clear what has failed. 

5. Are Percy tests deterministic? When adding new test infrastructure of this nature, it may be a good idea to run like to run it 10 times to ensure that Percy is deterministic and works with the current tests. Just an idea. If it is a source of more flakiness, that would not be ideal.

6. Has dev.to been bitten by visual bugs like these before, that we hope Percy to catch us from? I would be curious to see what percentage of bugs are visual bugs vs functional bugs. I haven't taken the time to look through the dev.to issues yet, sorry.

7. Here's an idea that could help keeping Percy names consistent! Have the spec metadata passed as a parameter in the spec block, so we have access to the full spec name in the spec.
```ruby
  it "shows a thread" do |spec|
    visit "#{article.path}/comments/#{comment.id_code_generated}"
    Percy.snapshot(page, name: spec.full_description)
    expect(page).to have_selector(".single-comment-node", visible: :visible, count: 2)
    expect(page).to have_selector(".comment-deep-0#comment-node-#{comment.id}", visible: :visible, count: 1)
    expect(page).to have_selector(".comment-deep-1#comment-node-#{child_comment.id}", visible: :visible, count: 1)
  end
```

8. I created a Percy token with my own personal Percy account.  I added the credential to my local application.yml, didn’t make an effect when put there, it seems the npm agent needs the Percy token, not the rspec process? Having a Percy token ONLY in the application.yml didn't work, and then having the token present in the env, AND in application.yml, produced this msg seen in the below screenshot. I think the node Percy process is the thing that needs the token, not the ruby process.


2020-05-13 11 58 32_i908sO png

benhalpern:
This all passes muster with me. I think ultimately we'll need to get a _feel_ for Percy and how we use it. Would I be correct in thinking that this will let us initially kind of run Percy in observer mode without it having any affect and then we can adopt rules around it once we get a feel?

snackattas:
About the deterministic test thing I mentioned earlier (and here's [Percy's opinion on determinism](https://docs.percy.io/docs/freezing-dynamic-data))

We could set a specific seed on the specs that have the new tag `js: true` to set the [determinism seed](https://github.com/faker-ruby/faker#deterministic-random)

In spec_helper.rb
```ruby
RSpec.configure do |config|
  config.around(:each, js: true) do |example|
    prev_random_seed = Faker::Config.random
    Faker::Config.random = Random.new(42)
    example.run
    Faker::Config.random = prev_random_seed
  end
```

vaidehijoshi:
Hi, @snackattas! 👋 

Thanks for taking the time to write all your questions up.

> 1. Before merging this in, it would be good to have a wiki page that answers the questions 'what happens when my Percy screenshot is different and the test fails', 'what happens when I need to change the baseline Percy snapshot (because I changed the page under test)', etc

I mentioned this in [my reply to @rhymes](https://github.com/thepracticaldev/dev.to/pull/7783#issuecomment-628101677), but I definitely do think we'll need a document/wiki for this (and probably need to add it to the contributing guide!). I would vote to do this in a separate PR, and I think that it could probably happen once we add the `PERCY_TOKEN` to Travis CI>

> 2. Does Percy deal with dynamic data well?

I think we chatted about this offline on Slack so we should be all good on this one! 😁 

> 3. Does Percy take a long time? We already have tech debt with regards to the speed of tests, wouldn’t want to make tests more slow. It might be useful to wrap the Percy calls around a timer to see how log each call takes. Maybe doing something like this just for testing

I tried out your little script in my local Percy module and didn't see anything that took a significant amount of time. 🤷 

> 4. What does an actual Percy fail look like? It would be good to show off what a 'real' fail appears like in Percy and in the Travis fail, and determine if it is clear what has failed.

Failures in Percy are rescued from [here](https://github.com/percy/percy-capybara/blob/master/lib/percy.rb) but can also come from the agent. You can see errors when you run Percy locally, but I haven't investigated what this will look like when we integrate with CI yet.

An example of a "duplicate snapshot" error
```
[percy] StatusCodeError 400 - {"errors":[{"status":"bad_request"},{"source":{"pointer":"/data/attributes/base"},"detail":"The name of each snapshot must be unique, and this name already exists in the build: 'Homepage: /dashboard/following_tags infinite scroll' -- You can fix this by passing a 'name' param when creating the snapshot. See the docs for more info on identifying snapshots for your specific client: https://percy.io/docs"}]} | Wed May 13 2020 14:11:25 GMT-0700 (Pacific Daylight Time)
```

Usually though, you'd see something like this when things are working correctly:
```
[percy] snapshot taken: 'Article: confirm deletion'
```

> 5. Are Percy tests deterministic? When adding new test infrastructure of this nature, it may be a good idea to run like to run it 10 times to ensure that Percy is deterministic and works with the current tests. Just an idea. If it is a source of more flakiness, that would not be ideal.

I integrated the "deterministic seed" onto Faker for Percy-tagged tests (aka `js: true`), and have confirmed that it works well. You can see this in [this snapshot in build 15](https://percy.io/thepracticaldev/dev.to/builds/5244336/view/308413465/1280?mode=diff&browser=chrome&snapshot=308413465) which used the seed, and then the [same snapshot in build 16](https://percy.io/thepracticaldev/dev.to/builds/5244537)

One issue, however, is that I don't think I've handled date/time for these snapshots; I'm thinking that maybe I can use `TimeCop` to explicitly set the time.

> 6. Has dev.to been bitten by visual bugs like these before, that we hope Percy to catch us from? I would be curious to see what percentage of bugs are visual bugs vs functional bugs. I haven't taken the time to look through the dev.to issues yet, sorry.

We definitely have! There was one CSS change recently that changed all the link styles on the main feed to be underlined by accident 😬 

> 7. Here's an idea that could help keeping Percy names consistent! Have the spec metadata passed as a parameter in the spec block, so we have access to the full spec name in the spec.

I tried this out locally to see what the `full_description` looks like on the `spec`. I'd prefer to manually name the specs rather than use the generated description because, IMO, Percy snapshots are easier to parse through when you can easily see what a snapshot is trying to represent (what page/route and what state/condition). The test descriptions are wordy and some of the time they don't even contain all that information in and of themselves.

> 8. I created a Percy token with my own personal Percy account. I added the credential to my local application.yml, didn’t make an effect when put there, it seems the npm agent needs the Percy token, not the rspec process? Having a Percy token ONLY in the application.yml didn't work, and then having the token present in the env, AND in application.yml, produced this msg seen in the below screenshot. I think the node Percy process is the thing that needs the token, not the ruby process.

Yep, I updated this PR to include a commit with the percy (node) agent; I think it was working for me before because I was just setting the variable globally in my terminal before running Percy locally!

snackattas:
Hey @vaidehijoshi, thanks for answering my questions :).  I agree with all your answers, just had a question about #4. But this is not a blocker, just more of a question, probably worth figuring out in a subsequent PR that actually adds the PERCY_TOKEN

> What does an actual Percy fail look like? It would be good to show off what a 'real' fail appears like in Percy and in the Travis fail, and determine if it is clear what has failed.

>>Failures in Percy are rescued from here but can also come from the agent. You can see errors when you run Percy locally, but I haven't investigated what this will look like when we integrate with CI yet...

What I meant in the question was, what if the code under test is legitimately changed or perhaps not legitimately changed, aka, the diff will fail. What does that look like/where is the exception raised?

In that situation, would the `Percy.snapshot` call raise the exception right there in the spec? Or would the Percy check in Github fail,  so the Travis run might pass, but the Percy check would fail ([seems like it from this area of Percy docs](https://docs.percy.io/docs/getting-started#step-3-review-and-approve-visual-changes))?

vaidehijoshi:
@snackattas Ah, gotcha. Thanks for clarifying!

If someone made a legitimate code change that impacted one/many Percy diffs, the Percy check would show `x diffs` that would need to be approved by someone on the team. The only time that `Percy.snapshot` would fail is if something was wrong with how we were naming the snapshot or if the browser had an issue or if the snapshot literally couldn't be _captured_. But not for legitimate code changes where there is a visual diff that is caught (if that makes sense?).

So, in short, yes: the Percy check on CI would show that a diff had been detected and a visual review was required :)

spec/spec_helper.rb
```diff
@@ -121,4 +121,12 @@
...
   # as the one that triggered the failure.
   Kernel.srand config.seed
+
+  # Explicitly set a seed to ensure deterministic Percy snapshots.
+  config.around(:each, js: true) do |example|
+    prev_random_seed = Faker::Config.random
+    Faker::Config.random = Random.new(42)
+    example.run
+    Faker::Config.random = prev_random_seed
+  end
```
vaidehijoshi, line 131:
```suggestion
    config.around(:each, js: true) do |example|
    prev_random_seed = Faker::Config.random
    Faker::Config.random = Random.new(42)
    Timecop.freeze("2019-03-04T10:00:00Z")

    example.run

    Faker::Config.random = prev_random_seed
    Timecop.return
  end
```

vaidehijoshi:
> Would I be correct in thinking that this will let us initially kind of run Percy in observer mode without it having any affect and then we can adopt rules around it once we get a feel?

@benhalpern This PR on its own won't do anything; as soon as we set this up in CI, we'll approve the generate Percy diffs off of the `master` build of Percy snapshots. At that point, every other PR that is opened against master will build snapshots and compare them against that "baseline" build on `master`. 

tl;dr, once I integrate this with GitHub + Travis, we'll start seeing snapshots on every new PR!

spec/spec_helper.rb
vaidehijoshi: @rhymes Yeah, I know what you mean. The other option that I can think of is something like this:
```ruby
# rails_helper.rb

config.around(:each, freeze_time: true) do |example|
  Timecop.freeze("2019-03-04T10:00:00Z")
  example.run
  Timecop.return
end
```

and then wrapping the Percy examples that specifically need a deterministic time set like this:
```ruby
it "renders the page", js: true, freeze_time: true do
  Percy.snapshot(page, name: "User: /:user_id for logged in user's own profile")
end
```

I don't love it, but it is the more conservative course of "change only what we need to change" rather than freezing time for all the JS tests (which is breaking things anyways!).

vaidehijoshi: > Can you give more details on which examples failed with that error?

@rhymes I pushed my commit so now you can see my many, many failing specs that show up when I try to freeze time for those `js: true` specs :(

rhymes: I'll take a look, thanks!

rhymes: I honestly have no idea why they fail, the first thing i can think of is the frozen time being too much in the past, this tricks the Chrome process which thinks is 2019 and then it fails? I wonder what those local capybara snapshots contain. Another thing I could try to check if what happens if you set that datetime with `date` in the terminal and then start Chrome manually

vaidehijoshi:
OK! The latest commit now freezes time and deterministically sets data (as opposed to dynamically) for tests that take a Percy snapshot.

I've also removed some duplicate/unnecessary snapshots. ✂️ 

You can see the latest snapshots [in this build](https://percy.io/thepracticaldev/dev.to/builds/5280541).

Going forward, for pages/states that we want to snapshot, we'll need to do something like this:
```ruby
it "shows the sign-with page", js: true do
  Percy.snapshot(page, name: "Logged out user: sign-in page")
  expect(page).to have_content(/Sign In With/i, count: 2)
end
```

PR #7787, opened 2020-05-12
Fix profile page rendering when a username contains spaces

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When a username is invalid and contains spaces, the profile page that is supposed to render the errors will break due to URI parsing. 

Fixes https://github.com/thepracticaldev/dev.to/issues/7758

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -59,8 +59,7 @@ class UsersController < ApplicationController
       @user.touch(:profile_updated_at)
       redirect_to "/settings/#{@tab}"
     else
-      Honeycomb.add_field("error",
-                          @user.errors.messages.reject { |_, v| v.empty? })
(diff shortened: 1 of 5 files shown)


lightalloy:
@benhalpern the username is not actually updated in the database but when `user` is not saved we render `edit`, at that time `user.username` can be still set to invalid, we try to render `app_url` on the `edit` page which used to cause 500 when the username contains a space.

rhymes:
> @benhalpern the username is not actually updated in the database but when `user` is not saved we render `edit`, at that time `user.username` can be still set to invalid, we try to render `app_url` on the `edit` page which used to cause 500 when the username contains a space.

@lightalloy thanks for the awesome explanation! :D

PR #7794, opened 2020-05-12
Add query-friendly article additions (Initial step)

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This feature allows admins to add a label which helps clarify the content of a post, primarily for our own search engine as well as Google, and of course clarification for humans. For now this is just going to affect non-logged-in traffic but I could see this being pretty useful across all posts.

It puts a preamble in the `<title>` tag and on the post, but it is distinct from the actual article content—I think it is helpful and appropriate to modify things about _the page_, but not the title/body of the content itself, which is authored by the user.

This also adds an option to re-word the description, which again is part of site meta info and not necessarily part of the content itself.

Here is an example where this could be helpful: The title itself isn't enough to convey that this is a helpful post if you search "Rails 6" in our search, even though it is all about Rails 6.

Screen Shot 2020-05-13 at 8 48 59 AM

This PR is only the data layer, as I'd like to experiment with this a bit right now before we get into how it would fit properly into `/internal` (Or even `/mod` possibly). This is a small, but somewhat powerful tool and we want to make sure we're thinking about it the right way.

### DIFF
--- a/db/migrate/20200511224704_add_query_friendly_admin_fields_to_articles.rb
+++ b/db/migrate/20200511224704_add_query_friendly_admin_fields_to_articles.rb
@@ -1,0 +1,6 @@
+class AddQueryFriendlyAdminFieldsToArticles < ActiveRecord::Migration[5.2]
+  def change
+    add_column :articles, :query_friendly_title_preamble, :string
+    add_column :articles, :query_friendly_description_alternative, :string
+  end
+end
--- a/db/migrate/20200511224704_add_query_friendly_admin_fields_to_articles.rb
+++ b/db/migrate/20200511224704_add_query_friendly_admin_fields_to_articles.rb
@@ -1,6 +1,6 @@
 class AddQueryFriendlyAdminFieldsToArticles < ActiveRecord::Migration[5.2]
   def change
-    add_column :articles, :query_friendly_title_preamble, :string
-    add_column :articles, :query_friendly_description_alternative, :string
+    add_column :articles, :search_optimized_title_preamble, :string
+    add_column :articles, :search_optimized_description_replacement, :string
   end
 end
--- a/app/assets/stylesheets/article-show.scss
+++ b/app/assets/stylesheets/article-show.scss
@@ -153,6 +153,17 @@ article {
         font-size: 57px;
       }
 
+      .article-title-preamble {
+        display: block;
+        color: var(--base-60);
+        font-size: 16px;
+        margin-bottom: -1.25vw;
+        @media screen and (min-width: 1600px) {
+          font-size: 17px;
+          margin-bottom: -18px;
+        }
+      }
+
       .title-block {
         display: inline-block;
       }
--- a/app/assets/stylesheets/article-show.scss
+++ b/app/assets/stylesheets/article-show.scss
@@ -156,10 +156,10 @@ article {
       .article-title-preamble {
         display: block;
         color: var(--base-60);
-        font-size: 16px;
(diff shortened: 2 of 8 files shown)


joshpuetz:
So... `query_friendly_title_preamble` is a moderator created alternative tag, and `query_friendly_description_alternative` is a moderator created alternative title?

The `Rails6` example is helpful: there's something nice about being able to categorize an article's topic within a tag. I'm less clear about the problem `query_friendly_description_alternative` is trying to solve though.

benhalpern:
@joshpuetz `query_friendly_description_alternative` is an _alternative_ text for the `meta description` tag which is not part of the visible post but is what Google often displays here...


Screen Shot 2020-05-13 at 11 55 36 AM

If the text we otherwise generate to go here is not very descriptive in a search context, we can manually present an _alternative_.

The `query_friendly_title_preamble` is something that will go _before_ the title in cases where it could be needed to provide information about the topic _for the primary purpose of ensuring the post responds well to appropriate searches._

The short-term need for this is to help ensure articles that should rank for certain topics are given the best chance by us giving a better hint at the topic of the post if the title is not already organized well for this purpose. But the longterm benefits could also be a step towards generally improved platform labeling systems for structured and non-structured queries of all sorts.

mstruve:
> query_friendly_description_alternative is an alternative text for the meta description

Why not use the description field as this alternative rather than making another hidden field?

I am also hesitant if this is something mods or admins will be filling in. The more work we make for mods/admins the less easily scalable our platform is and the more work we have to ask from others who want communities like ours.

benhalpern:
@mstruve this is all 100% optional, and we intend to use it only to address the highest impact 1% of posts which could go from, say, 8th on Google rankings, etc. to 1st with a few keyword arrangements in key details.

We already allow regular users to fill in the description and title, and ideally we can offer help in instructing them to pick the most appropriate title, but this is just a backend tool to enhance where needed. Everything will otherwise work exactly as it does and no community will _need_ to have this in place.

This is a light progressive enhancement to let search-focused admins make smart choices in a curation-oriented way. I 100% agree that we wouldn't want to institute non-scalable functionality, but our search traffic is entirely power-law distribution and this lets us double down on our best horses. There's also a roadmap to where this could be fully or partially automated based on some clever programming, but since I'm confident this would have a pretty immediate impact as-is I don't think we should try and get all the way to that point too soon. 😄

db/migrate/20200511224704_add_query_friendly_admin_fields_to_articles.rb
```diff
@@ -0,0 +1,6 @@
+class AddQueryFriendlyAdminFieldsToArticles < ActiveRecord::Migration[5.2]
+  def change
+    add_column :articles, :query_friendly_title_preamble, :string
```
mstruve, line 3: I'm trying to think of a name here that would make it clear it's for search engines. 
- discovery_friendly_title_preamble
- optimized_search_title_preamble
- search_engine_title_preamble

benhalpern: Good call. I like `optimized_search_title_preamble`

joshpuetz: Wonderful suggestions @mstruve: makes me realize how much of my hesitate was related to the naming of these fields!

PR #7799, opened 2020-05-12
Adjust overflow in comments and revert inline css

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This makes an adjustment so longer comments subtly fade away... Maybe not the permanent solution but I think it works...

Screen Shot 2020-05-12 at 12 12 25 PM

This also fixes the ill-fated inlined `<style>`

### DIFF
--- a/app/assets/stylesheets/components/comments.scss
+++ b/app/assets/stylesheets/components/comments.scss
@@ -16,17 +16,32 @@
   &__body {
     margin-top: calc(var(--su-1) * -1);
     padding-left: var(--su-7);
+    padding-bottom: var(--su-2);
+    max-height:330px;
+    overflow: hidden;
 
     a:hover {
       text-decoration: underline; //todo: default for all in-content links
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/components/comments.scss
```diff
@@ -16,19 +16,34 @@
...
       -webkit-box-orient: vertical;
       overflow: hidden;
+      max-width: 100%;
+      width: auto;
+      object-fit: contain; 
     }
   }
 
+  &__footer {
+    box-shadow: 0px 0px var(--su-2) var(--su-2) var(--story-comments-bg);
```
benhalpern, line 42: This looks _a little_ funny on some of the themes because it uses a transparent value here so it overlaps weirdly but we can squash that after the fact.

PR #7800, opened 2020-05-12
Delay RemoveFromIndex Job to Fix Callback Race Conditions

mstruve:
- [x] Refactor
- [x] Bug Fix

## Description
We recently had two incidents where an Article and a Comment were updated and then right after deleted. This led to index and removal from index jobs to be enqueued close enough together that the removal job finished before the indexing jobs causing the records to still be in Elasticsearch even though they are no longer in Postgres. A big reason for this is our callbacks which ideally we should get rid of. However, that is a huge undertaking so **my "hack" to solve this problem for now is to put a small delay on the removal job of 5 seconds.**

While trying to make heads and tails of our callbacks I went ahead and organized the comment callbacks which were kinda haphazardly thrown together. 

## Added tests?

alt_text

### DIFF
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -13,7 +13,7 @@ module Searchable
 
   def remove_from_elasticsearch
     # Callbacks can cause index and removal jobs to be enqueued at the same time
-    # to avoid indexing a document after removing it we delay the removal job by 5 seconds to 
+    # to avoid indexing a document after removing it we delay the removal job by 5 seconds to
     # ensure it is run last
     Search::RemoveFromIndexWorker.perform_in(5.seconds, self.class::SEARCH_CLASS.to_s, search_id)
   end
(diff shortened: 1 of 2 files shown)


app/models/concerns/searchable.rb
```diff
@@ -12,7 +12,7 @@ def index_to_elasticsearch_inline
   end
 
   def remove_from_elasticsearch
-    Search::RemoveFromIndexWorker.perform_async(self.class::SEARCH_CLASS.to_s, search_id)
+    Search::RemoveFromIndexWorker.perform_in(5.seconds, self.class::SEARCH_CLASS.to_s, search_id)
```
maestromac, line 15: Not a block but it would be a good idea to leave a comment that this is to avoid that edge-case.

mstruve: good call!

PR #7814, opened 2020-05-12
Open in new window properly on ctrl+click etc

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

ctrl+click or middle-click should trigger `_blank` in this case as it would by default if this were an anchor...

### DIFF
--- a/app/javascript/articles/Article.jsx
+++ b/app/javascript/articles/Article.jsx
@@ -34,6 +34,7 @@ export const Article = ({
     'crayons-story__title',
     'crayons-story__tags',
     'crayons-story__bottom',
+    'crayons-story__tertiary',
   ];
 
   return (
@@ -49,10 +50,14 @@ export const Article = ({
         role="presentation"
         onClick={(event) => {
           const { classList } = event.target;
-
           if (clickableClassList.includes(...classList)) {
-            InstantClick.preload(article.path);
-            InstantClick.display(article.path);
+            if (event.which > 1 || event.metaKey || event.ctrlKey) { // Indicates should open in _blank
+              window.open(article.path, '_blank');
+            } else {
+              const fullUrl = window.location.origin + article.path // InstantClick deals with full urls
+              InstantClick.preload(fullUrl);
+              InstantClick.display(fullUrl);
+            }
           }
         }}
       >
--- a/app/javascript/articles/components/CommentListItem.jsx
+++ b/app/javascript/articles/components/CommentListItem.jsx
@@ -13,8 +13,13 @@ export const CommentListItem = ({ comment }) => (
     className="crayons-comment pl-2 cursor-pointer"
     role="presentation"
     onClick={(_event) => {
-      InstantClick.preload(comment.path);
-      InstantClick.display(comment.path);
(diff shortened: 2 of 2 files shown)


app/javascript/articles/Article.jsx
```diff
@@ -34,6 +34,7 @@ export const Article = ({
     'crayons-story__title',
     'crayons-story__tags',
     'crayons-story__bottom',
+    'crayons-story__tertiary',
```
benhalpern, line 37: This is the `x min read` area, which should be included here.

app/javascript/articles/Article.jsx
```diff
@@ -49,10 +50,14 @@ export const Article = ({
...
         onClick={(event) => {
           const { classList } = event.target;
-
           if (clickableClassList.includes(...classList)) {
-            InstantClick.preload(article.path);
-            InstantClick.display(article.path);
+            if (event.which > 1 || event.metaKey || event.ctrlKey) { // Indicates should open in _blank
+              window.open(article.path, '_blank');
+            } else {
+              const fullUrl = window.location.origin + article.path // InstantClick deals with full urls
```
benhalpern, line 57: This is just getting this to work more effectively alongside `InstantClick`

rhymes:
Which area of the code is affected to this? I tried with the current production by meta clicking and it works but maybe I didn't find the right area :D

benhalpern:
@rhymes clicking on the _title_ of a feed card works as appropriate because it's a true `a` element. Currently we also allow any other part of the card which is not also a link to act as click through on the card, but it does not respect the meta-click behavior.

app/javascript/articles/Article.jsx
```diff
@@ -49,10 +50,14 @@ export const Article = ({
         role="presentation"
         onClick={(event) => {
           const { classList } = event.target;
-
           if (clickableClassList.includes(...classList)) {
-            InstantClick.preload(article.path);
-            InstantClick.display(article.path);
+            if (event.which > 1 || event.metaKey || event.ctrlKey) { // Indicates should open in _blank
```
benhalpern, line 54: `event.which > 1` is "middle click" on mouse. `event.metaKey` and `event.ctrlKey` account for command/ctrl + click.

app/javascript/articles/Article.jsx
```diff
@@ -49,10 +50,14 @@ export const Article = ({
...
+              window.open(article.path, '_blank');
```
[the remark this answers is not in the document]
benhalpern: Given that this will only link to other content _within_ the platform it shouldn't really matter—If anything there _could_ be a legit purpose for possibly having access to the prior window for any reason.

benhalpern:
@rhymes got it working on the comment area too. (Probably could be an abstraction, but I'll take rule of three because we could wind up modifying one of these behaviors to be different down the line)

Interestingly in Chrome it _does_ open in background but doesn't in Firefox. Unless there is an instruction we can give to make it consistent I think this is closer to expected behavior than not doing this.

PR #7828, opened 2020-05-13
Validate both images and pattern_images

atsmith813:
- [x] Bug Fix

## Description
https://app.honeybadger.io/fault/66984/3bb2fc797ca1c1fb069b50bb8876a3c7

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed

alt_text


app/controllers/podcasts_controller.rb
```diff
@@ -58,30 +60,28 @@ def skip_bullet
...
-  def valid_image_file?(image)
-    return true if file?(image)
-
-    @podcast.errors.add(:image, IS_NOT_FILE_MESSAGE)
-    false
-  end
+  def valid_image_files_and_names?(images)
+    images.each do |field, image|
+      @podcast.errors.add(field, IS_NOT_FILE_MESSAGE) unless file?(image)
+      break if @podcast.errors.any?
```
atsmith813, line 80: I'm breaking here to avoid doing extra logic once the first error is found. These types of errors are _very_ rare so it's not necessary to always perform all the checks at once.

app/controllers/podcasts_controller.rb
```diff
@@ -58,30 +60,28 @@ def skip_bullet
     Bullet.enable = previous_value
   end
 
-  def valid_image?
-    image = params.dig("podcast", "image")
-    return true unless image
+  def valid_images?
+    images = podcast_params.select { |k, _v| IMAGE_KEYS.include?(k) }
```
mstruve, line 64:
```ruby
podcast_params.slice(*IMAGE_KEYS)
```

app/controllers/podcasts_controller.rb
mstruve: This could mean some relevant error messages are not given back to the user but I dont think that is a huge deal.

atsmith813:
Force pushed to merge master and fix conflicts from #7829

PR #7832, opened 2020-05-13
Rate Limit Feedback Message Creation By User ID or IP

mstruve:
- [x] Feature

## Description
Rate limit number of abuse reports a user can create in a 5 minute period. This PR applies the limit to both logged-in and not logged-in users by adding an additional IP option for making the cache keys in our RateLimitChecker.  I think this will be a huge benefit as rate-limiting is most important for those not logged in. 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -92,5 +92,11 @@ class ApplicationController < ActionController::Base
     rate_limiter.check_limit!(action)
   end
 
-  delegate :rate_limiter, to: :current_user
+  def rate_limiter
+    (current_user || anonymous_user).rate_limiter
+  end
+
+  def anonymous_user
+    User.new(ip_address: request.env["HTTP_FASTLY_CLIENT_IP"])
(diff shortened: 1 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -92,5 +92,11 @@ def rate_limit!(action)
     rate_limiter.check_limit!(action)
   end
 
-  delegate :rate_limiter, to: :current_user
+  def rate_limiter
+    (current_user || anonymous_user).rate_limiter
+  end
+
+  def anonymous_user
```
citizen428, line 99: :fire: Finally, a proper null object! 😍 I'm a big fan of this pattern and would like if we used this more widely instead of `nil` checks on `current_user` as it has the potential of removing many conditionals and making the code more "tell, don't ask".

rhymes: Also: anonymous users are users :-)

PR #7833, opened 2020-05-13
filter/sample noisy spans

lizthegrey:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
Drops noisy redis and sql spans from being sent to honeycomb

## Added tests?

- [ ] yes
- [ ] no, because I need help

https://ui.honeycomb.io/thepracticaldev/datasets/rails/result/abvTbedLExw
https://ui.honeycomb.io/thepracticaldev/datasets/rails/result/cPKPv3mVQZf

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/config/initializers/honeycomb.rb
+++ b/config/initializers/honeycomb.rb
@@ -1,3 +1,31 @@
+class CustomSampler
+  extend Honeycomb::DeterministicSampler
+
+  def self.sample(fields)
+    noisy = Set[
+      'GET rails-settings-cached/v1',
+      'TIME',
+      'BEGIN',
+      'COMMIT',
+    ]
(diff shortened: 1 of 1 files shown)


lizthegrey:
> I wonder if we could eliminate grabbing those cache settings every time we load a template 
> 
> Anyways, we could do some refactoring here but I am fine to push this out as is and then refactor later.

Yeah, it's clearly a sign you are doing something inefficient, so good eye to think about optimizing it away even as we remove the spans from your honeycomb traces. The worst load is load that is invisible to your telemetry.

Issue #6255, opened 2020-02-24
Reorganize internal and public API controllers.


citizen428:
During the discussions in #6136 it became clear that not all controllers currently residing in `app/controllers/api/v0` are meant for the public API, some are used by the DEV website itself. Since we use slightly different authorization methods and checks this can lead to subtle bugs, so we should move things that aren't mean for public consumption to a different namespace or to a `responds_to block` in their "main" controller.

Todo:

- [x] `POST     /api/reactions/onboarding           api/v0/reactions#onboarding`
- [x] `POST     /api/reactions                      api/v0/reactions#create`
- [x] `POST     /api/github_repos/update_or_create  api/v0/github_repos#update_or_create`
- [x] `GET      /api/github_repos                   api/v0/github_repos#index`
- [X] `GET      /api/tags/onboarding                api/v0/tags#onboarding`
- [x] `Api::V0::FollowingsController`
- [x] `Api::V0::ChatChannelsController`
- [x] `Api::V0::UsersController#index`

rhymes:
Adding a note here as the API docs are not the correct metric of evaluation since they are largely incomplete, see https://github.com/thepracticaldev/dev.to/issues/4474

I think we can only move those controllers that do not currently require an API key or OAuth as auth

citizen428:
Thanks @rhymes, description updated.

citizen428:
@rhymes Going through #4474 it looks like it just includes everything that's in `api/v0/controllers`, including things like Github repos that started the whole conversation in #6136. So I think step 1 would be to clean up that issue and you definitely have more context on the API than I do.

rhymes:
Updated the list there, thanks @citizen428!

These are those that should be moved:

```text
- [ ] `POST     /api/reactions/onboarding           api/v0/reactions#onboarding`
- [ ] `POST     /api/reactions                      api/v0/reactions#create`
- [ ] `GET      /api/tags/onboarding                api/v0/tags#onboarding`
- [ ] `POST     /api/github_repos/update_or_create  api/v0/github_repos#update_or_create`
- [ ] `GET      /api/github_repos                   api/v0/github_repos#index`
```

There's also this:

```text
- [ ] `POST     /api/follows                        api/v0/follows#create`
```

this shouldn't be part of the public API but it currently is, it uses `authenticate_with_api_key_or_current_user`, so we we might have to investigate into Datadog to see if anybody other than us is using it or just leave it there.

citizen428:
[...]

PR #7834, opened 2020-05-13
Create write_articles oauth scope

maestromac:
- [x] Feature

## Description
This PR adds `:write_articles` scope to expand our doorkeeper capability. I've also enabled `enforce_configured_scopes` so that application can only be created with scope we provided
n/a
image
## Added tests?
- [y] yes

## Added to documentation?
- [x] no documentation needed

n/a

### DIFF
--- a/app/controllers/api/v0/articles_controller.rb
+++ b/app/controllers/api/v0/articles_controller.rb
@@ -1,9 +1,9 @@
 module Api
   module V0
     class ArticlesController < ApiController
-      before_action :authenticate_with_api_key_or_current_user!, only: %i[create update]
-      before_action :authenticate!, only: :me
+      before_action :authenticate!, only: %i[create update me]
       before_action -> { doorkeeper_authorize! :public }, only: %w[index show], if: -> { doorkeeper_token }
+      before_action -> { doorkeeper_authorize! :write_articles }, only: %w[create update], if: -> { doorkeeper_token }
 
       before_action :validate_article_param_is_hash, only: %w[create update]
 
--- a/spec/requests/api/v0/articles_spec.rb
+++ b/spec/requests/api/v0/articles_spec.rb
@@ -462,14 +462,6 @@ RSpec.describe "Api::V0::Articles", type: :request do
         post api_articles_path, headers: { "api-key" => api_secret.secret, "content-type" => "application/json" }
         expect(response).to have_http_status(:unauthorized)
       end
-
-      it "fails when oauth's access_token" do
-        access_token = create(:doorkeeper_access_token, resource_owner_id: user.id)
-        headers = { "authorization" => "Bearer #{access_token.token}", "content-type" => "application/json" }
-
-        post api_articles_path, params: { article: { title: Faker::Book.title } }.to_json, headers: headers
-        expect(response).to have_http_status(:unauthorized)
-      end
     end
 
     describe "when authorized" do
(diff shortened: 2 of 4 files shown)


app/controllers/api/v0/articles_controller.rb
```diff
@@ -1,9 +1,9 @@
 module Api
   module V0
     class ArticlesController < ApiController
-      before_action :authenticate_with_api_key_or_current_user!, only: %i[create update]
```
mstruve, line 4: General question, since `authenticate!` checks doorkeeper plus API key plus current_user, why cant we just get rid of this method `authenticate_with_api_key_or_current_user ` and use `authenticate!` everywhere in the API?

app/controllers/api/v0/articles_controller.rb
```diff
@@ -1,9 +1,9 @@
...
-      before_action :authenticate!, only: :me
+      before_action :authenticate!, only: %i[create update me]
       before_action -> { doorkeeper_authorize! :public }, only: %w[index show], if: -> { doorkeeper_token }
+      before_action -> { doorkeeper_authorize! :write_articles }, only: %w[create update], if: -> { doorkeeper_token }
```
mstruve, line 6: Will tokens that have previously been created have to "update" themselves somehow to get this new permission?

spec/requests/api/v0/articles_spec.rb
```diff
@@ -481,6 +473,22 @@ def post_article(**params)
         post api_articles_path, params: { article: params }.to_json, headers: headers
       end
 
+      it "returns a 403 if :write_articles scope is missing (oauth)" do
```
mstruve, line 476: 🔥 specs!

app/controllers/api/v0/articles_controller.rb
maestromac: I think for a time we felt most assured using `authenticate_with_api_key_or_current_user` will distinguish between what OAuth apps can or cannot have access to. It's also because anything OAuth app have access to will need scope checks and we didn't have that figured out yet. This is just my assumption though.

app/controllers/api/v0/articles_controller.rb
maestromac: Create and update previously can only be accessed via cookie and API token. This is new permission and will not be breaking existing access.

app/controllers/api/v0/articles_controller.rb
citizen428: For context: In the past, we had things that were only meant for internal consumption in the public API, so I believe this was added to account for that. However, as of #6255, this is no longer true, so maybe you want to investigate a bit and potentially simplify this @maestromac

lightalloy: Agree with @maestromac , it's for restricting access when a user is authorized with the doorkeeper_token for the time while we only have basic scopes.
As I see, we still don't have doorkeeper scopes for all the api features, so we may want to keep limiting the access in this case.

lightalloy: Anyway, I would keep the possible changes for another pr :)

rhymes: I think too we should err on the side of caution on opening up OAuth features to registered apps.

mstruve:
Just realized this is going to break API write throttling bc we use the API token to track it. issuing a fix PR

PR #7836, opened 2020-05-13
Add max height to content images to ensure they can't over-eat vertical space

benhalpern:
…al space

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently in comments and articles if we have an image which is _super tall_ it sometimes kind of breaks things because we default to 100% max-width. This PR adds sensible max heights to complement max width and ensure images are more easily consumed if they are super tall.

Tall images like this are now much less out of place.

Screen Shot 2020-05-13 at 5 25 31 PM

Screen Shot 2020-05-13 at 5 38 02 PM

Screenshots are probably the worst offender in that they tend to be _super tall_ and kind of ugly when blown up huge. This helps account for more types of images. Clicking still lets the user see a bigger pic (And we could close the loop on lighthouse view to really make this a slick experience).

### DIFF
--- a/app/assets/stylesheets/article-show.scss
+++ b/app/assets/stylesheets/article-show.scss
@@ -612,6 +612,7 @@ article {
       margin: auto;
       left: -4%;
       max-width: 108%;
+      max-height: calc(50vh + 180px);
 
       @media screen and (min-width: 430px) {
         left: -6px;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/article-show.scss
```diff
@@ -612,6 +612,7 @@ article {
       margin: auto;
       left: -4%;
       max-width: 108%;
+      max-height: calc(50vh + 180px);
```
ludwiczakpawel, line 615: These values feel pretty random but I think I have nothing against it :D

PR #7837, opened 2020-05-13
Revert "filter/sample noisy spans"

mstruve:
Reverts thepracticaldev/dev.to#7833


mstruve:
Merging this revert to unblock deploys

PR #7838, opened 2020-05-13
Adds csv formatter only in CI and makes screenshot names unique with timestamp

snackattas:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This is a  RSpec formatter that detects if we are in the travis environment, and if we are, saves test runtimes, test pass/fail, the travisci link, the exception, etc, to a csv file.

At the end of the run, the csv is sent up to an s3 bucket that the dev.to core team configures. These csvs/screenshots are only sent for builds on master, unfortunately :( https://github.com/travis-ci/travis-ci/issues/5579

In addition this PR adds a monkeypatch to ensure screenshots get unique names. 

I can explain some of the more meaty changes in detail:
### rspec/retry
* We use the rspec/retry gem in this suite to rerun `{js: true}` specs.  Retrying specs is sort of an antipattern, but one that a lot of suites live with (and I'm fine with it for now). But over time, retrying can hide real issues, and bloat test suite runtimes.  

* So I thought it'd be good to add retry attempts to the csv reporter. However, rspec/retry goes around the `RSpec::Core::Formatter#example_failed` hook (or any hook for that matter).  If we just followed RSpecExampleReporter blogs, we'd be missing out on the data from retried specs, and that kind of hides retry attempts in an insidious way.

* Luckily, the `rspec/retry` gem has a method `retry_callback` which executes a block with the `example` argument that has all the reporting data we need! But we need to  get that data from the RSpec configure block to the CSVReporter (where it can be combined with other reporting data).

*  I decided to use a Singleton to pass data from `retry/callback` to our custom reporter.  RSpec Reporters are listeners, its hard to get data to them....they just listen for rspec events and get tossed ExampleNotification instances. Since they don't take extra args, Singleton seemed like the way to get data to the `close` event. And now we can keep an eye on those retries :)

### monkeypatch screenshot
* Screenshots are great for seeing what went wrong at the moment of a fail. For a while I was wondering, though, how screenshots are saved in this repo. Turns out, if you have a `{type: system}` tag, `rspec/rails` will automagically save a screenshot for you in `tmp/screenshots`, and will handle other Capybara state mgmt stuff. [_\~Rails magic~_](https://guides.rubyonrails.org/testing.html#screenshot-helper)

* The trouble with that is, the screenshot names aren't configurable. So, combined with rspec/retry, each screenshot taken for a spec has the same name, even if its the 1st, 2nd, or 3rd screenshot.  Again, missing out on that retry data -- later screenshots overwrite earlier ones. Also we don't want to skip rspec/rails cleanup entirely, we still want the Rails/Capybara state mgmt stuff

* From these issue notes/stackoverflow articles I learned we could monkeypatch it! Now each screenshot gets a unique timestamp
  * https://github.com/mattheworiordan/capybara-screenshot/issues/225
  * https://stackoverflow.com/questions/44336581/turn-off-screenshots-in-rails-system-tests
  * https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/system_testing/test_helpers/screenshot_helper.rb#L64

### other tiny change
* added `--format progress` to the `bundle exec rspec` call in the `.travis.yml` because travis times out/fails your run if it delivers no output to stdout in 10 minutes, and want to avoid that issue. Don't want to depend on deprecation warnings to stave us from it

### Proof
* I intentionally added some different kinds of fails in here, and screenshots/csvs dumped at end of the run https://github.com/thepracticaldev/dev.to/pull/7838/commits/b45f3cfa2d65d6d356bd343ffecd432d92721ec7
  * https://travis-ci.com/github/thepracticaldev/dev.to/builds/166825441
  * [intentional_fail.txt](https://github.com/thepracticaldev/dev.to/files/4648200/intentional_fail.txt)

2020-05-18 22 50 47_IMTbgI png

* Another run of slightly different intentional fails
  * https://travis-ci.com/github/thepracticaldev/dev.to/builds/166828393
  * [intentional_fail_again.txt](https://github.com/thepracticaldev/dev.to/files/4648203/intentional_fail_again.txt)

* I removed my intentional fails for the 3rd run, and it passed, but as it turns out, it had 2 specs that were retried but passed upon 2nd run (proving the value of logging retries)
  * https://travis-ci.com/github/thepracticaldev/dev.to/builds/166830190
  * [has_retry_fails_but_still_passed_overall.txt](https://github.com/thepracticaldev/dev.to/files/4648204/has_retry_fails_but_still_passed_overall.txt)
2020-05-18 23 28 35_59j8d7 png
2020-05-18 23 28 07_roaeWL png

Addresses https://github.com/thepracticaldev/dev.to/issues/7830 (although, not going to close that Epic out yet because, want to evaluate if this meets the need for Test reporting by using this over time)

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

typer


spec/support/initializers/rspec_csv_formatter.rb
```diff
@@ -0,0 +1,35 @@
+require "csv"
+
+class CSVFormatter < RSpec::Core::Formatters::JsonFormatter
+  RSpec::Core::Formatters.register self
+
+  def close(_notification)
+    headers = ["Spec", "File", "Status", "Run Time", "Exception"]
```
citizen428, line 7: Should we move the headers to a constant, they don't change.

spec/support/initializers/rspec_csv_formatter.rb
```diff
@@ -0,0 +1,35 @@
...
+        csv << row
+      end
+    end
+  end
+end
+
+if ENV["TRAVIS"]
+  RSpec.configure do |config|
+    tmp_dir = File.join(Dir.pwd, "tmp")
+    FileUtils.mkdir(tmp_dir) unless File.directory?(tmp_dir)
```
citizen428, line 25: Ruby also has [`FileUtils.mkdir_p`](https://ruby-doc.org/stdlib-2.7.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-mkdir_p) which behaves like `mkdir -p` in the shell, so you can simplify this a bit to

```ruby
csvs_dir = File.join("tmp", "csvs")
FileUtils.mkdir_p(csvs_dir)
```

spec/support/initializers/rspec_csv_formatter.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    csvs_dir = File.join(tmp_dir, "csvs")
+    FileUtils.mkdir(csvs_dir) unless File.directory?(csvs_dir)
+
+    timestamp = DateTime.now.utc.strftime("D%Y-%m-%dT%H-%M-%S")
```
rhymes, line 30: Not sure if the leading `D` is on purpose:

```ruby
[1] pry(main)> DateTime.now.utc.strftime("D%Y-%m-%dT%H-%M-%S")
=> "D2020-05-14T08-43-52"
```

my suggestion is to use `Time.current.utc` and add the `Z` at the end, so by looking at files we know the time is always in UTC instead of having to remember down the line at each passage the file will do:

```ruby
Time.current.utc.strftime("%Y-%m-%dT%H-%M-%SZ")
```

will result in:

```ruby
"2020-05-14T08-44-23Z"
```

we could also simply use: 

```ruby
[3] pry(main)> Time.current.utc.iso8601
=> "2020-05-14T08:46:30Z"
```

which is already pretty similar and printable

snackattas: Thanks for the suggestions! I'll do the iso8601 one, gotta stay iso compliant :)

citizen428:
Looking good now, thanks @snackattas 😃 Please mark this PR as "Ready for review" when you're done making changes.

PR #7847, opened 2020-05-14
Track and Limit API writes by IP Address + Rack Attack Refactor 

mstruve:
- [x] Refactor
- [x] Bug Fix

## Description
Since merging #7834 we may no longer have an API token when an Oauth user uses the API to update or create Articles. With that API key our current throttling will no longer work.  This updates the API write throttle to us IP address like the rest of the throttles. 

On my TODO list I want to extract the fastly env variable and make it configurable so in the future, a community doesn't need a fastly client IP to rate limit. 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/config/initializers/rack/attack.rb
+++ b/config/initializers/rack/attack.rb
@@ -2,37 +2,45 @@ Rack::Attack.throttled_response_retry_after_header = true
 
 class Rack::Attack
   throttle("search_throttle", limit: 5, period: 1) do |request|
-    if request.path.starts_with?("/search/") && request.env["HTTP_FASTLY_CLIENT_IP"].present?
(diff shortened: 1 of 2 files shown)


config/initializers/rack/attack.rb
```diff
@@ -2,37 +2,45 @@
...
+
+    if request.path.starts_with?("/messages") && request.post?
+      track_and_return_ip(request.env["HTTP_FASTLY_CLIENT_IP"])
     end
   end
+
+  def self.track_and_return_ip(ip_address)
+    Honeycomb.add_field("fastly_client_ip", ip_address)
+    ip_address.to_s
+  end
```
atsmith813, line 45: 🔥

PR #7848, opened 2020-05-14
Refactor GitHub::Client to extract Github::OauthClient for users

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR contains the following:

- `Github::OauthClient.for_user(user)` which creates a GitHub client tied to the specific user's access token
- keeps `Github::Client` for application credentials
- adds dynamic scopes for `Identity`, like `Identity.twitter`, `Identity.github`

(thanks @citizen428 for the suggestions on how to build this)

## Added tests?

- [x] yes
- [ ] no, because I need help


app/controllers/github_repos_controller.rb
```diff
@@ -10,7 +10,7 @@ def index
     # NOTE: this will invoke autopaging, by issuing multiple calls to GitHub
     # to fetch all of the user's repositories. This could eventually become slow
     @repos = fetch_repositories_from_github(known_repositories_ids)
-  rescue Octokit::Unauthorized => e
+  rescue Github::Errors::Unauthorized => e
```
citizen428, line 13: 🔥 I always prefer wrapping dependencies and not directly exposing them to most of the codebase.

app/services/github/oauth_client.rb
```diff
@@ -0,0 +1,135 @@
...
+    def respond_to_missing?(method, _include_all = false)
+      target.respond_to?(method, false) || super
+    end
+
+    def check_credentials!(credentials)
+      if credentials.present?
+        return credentials.select { |k| k == :access_token } if credentials[:access_token].present?
+
+        all_auth_credentials = proc { |key, value| APP_AUTH_CREDENTIALS.include?(key) && value.present? }
+        return credentials.select { |key| APP_AUTH_CREDENTIALS.include?(key) } if credentials.all?(all_auth_credentials)
```
citizen428, line 53: Am I misreading this, or could you  

```ruby
return credentials if credentials.all?(all_auth_credentials)
```

since the condition already verifies that all the keys are in `APP_AUTH_CREDENTIALS`?

I think actually all of this could be a bit simplified (untested):

```ruby
# assumes we move the proc into a constant
result = credentials[:access_token] || (credentials if credentials.all?(ALL_AUTH_CREDENTIALS))
return result if result
```

rhymes: Good idea! will update it shortly

app/services/github/oauth_client.rb
```diff
@@ -0,0 +1,134 @@
...
+    APP_AUTH_CREDENTIALS = %i[client_id client_secret].freeze
+    ALL_AUTH_CREDENTIALS = proc { |key, value| APP_AUTH_CREDENTIALS.include?(key) && value.present? }.freeze
+
+    # @param credentials [Hash] the OAuth credentials, {client_id:, client_secret:} or {access_token:}
+    def initialize(credentials)
+      @credentials = check_credentials!(credentials)
+    end
+
+    def self.for_user(user)
+      access_token = user.identities.github.select(:token).take!.token
```
mstruve, line 13: TIL I learned about `take`!

app/services/github/oauth_client.rb
```diff
@@ -0,0 +1,134 @@
...
+        result = credentials[:access_token].presence || (credentials if credentials.all?(ALL_AUTH_CREDENTIALS))
+        return result if result
+      end
+
+      message = "The client either needs a valid 'client_id'/'client_secret' pair or an 'access_token'!"
+      raise ArgumentError, message
+    end
+
+    def request
+      Honeycomb.add_field("app.name", "github.client")
```
mstruve, line 60: This field should just be `name` otherwise you get `app.app.name` which I found out the hard way. Also as I mentioned before this wont tag the actual Octokit request, just whatever kicks it off.

app/models/identity.rb
```diff
@@ -4,14 +4,16 @@ class Identity < ApplicationRecord
 
   scope :enabled, -> { where(provider: Authentication::Providers.enabled) }
 
+  Authentication::Providers.available.each do |provider_name|
+    scope provider_name, -> { where(provider: provider_name) }
```
mstruve, line 8: 🔥

app/models/identity.rb
```diff
@@ -4,14 +4,16 @@ class Identity < ApplicationRecord
 
   scope :enabled, -> { where(provider: Authentication::Providers.enabled) }
 
+  Authentication::Providers.available.each do |provider_name|
```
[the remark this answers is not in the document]
rhymes: mmm I thought about it but this would mean that a random script for maintenance or any other code could fail if an admin in the meantime has disabled one of the providers

PR #7849, opened 2020-05-14
Use bulk_show Endpoint for Follow Buttons on Followers Dashboard

mstruve:
- [x] Bug Fix
- [x] Optimization

## Description
This updates the user followers Dashboard buttons to use the same bulk_show endpoint we use in our search for fetching the status for the buttons. 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/views/dashboards/followers.html.erb
+++ b/app/views/dashboards/followers.html.erb
@@ -31,7 +31,7 @@
               <span class="dashboard-username">@<%= user.username %></span>
               <a href="#"
                 id="dashboard-follow"
-                class="cta follow-action-button dashboard-follow"
+                class="cta follow-action-button dashboard-follow follow-user"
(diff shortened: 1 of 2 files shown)


app/views/dashboards/followers.html.erb
```diff
@@ -31,7 +31,7 @@
               <span class="dashboard-username">@<%= user.username %></span>
               <a href="#"
                 id="dashboard-follow"
-                class="cta follow-action-button dashboard-follow"
+                class="cta follow-action-button dashboard-follow follow-user"
```
mstruve, line 34: [This indicates to us to user the bulk user endpoint](https://github.com/thepracticaldev/dev.to/pull/7676/files#diff-100ce16bf208667b69c5a3725bbe6a60R4)

rhymes: such a subtle fix :D

PR #7854, opened 2020-05-14
Return boolean value instead of undefined for listings open modal

Zhao-Andy:
- [x] Bug Fix

## Description
This fixes a small bug where individual listings can't be opened properly because the variable wasn't a boolean.

## Added tests?
- [x] no, because I need help

Alvin the chipmunk bends down like, "Oops!"


Zhao-Andy:
Yeah, maybe @nickytonline has an idea on how to write a test for this in Jest?

I was considering writing it as a system test but that seemed overreaching.

PR #7873, opened 2020-05-14
Fixing the markup in the badge embed code

JulianG:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

On the badges page (https://dev.to/p/badges) the `img` tag was missing the self-closing forward slash, both in the actual markup and in the embed code.

When using the embed code in JSX you get the following error:

```
Expected corresponding JSX closing tag for 'img'.ts(17002)
```
jsx error highlight

I'm afraid I could not bring myself to setup a local dev environment. Not even using Docker. **So I haven't even run this once.**

I understand if you can't approve this PR but someone might want to fix this anyways.

No

No

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

success

### DIFF
--- a/app/views/pages/badges.html.erb
+++ b/app/views/pages/badges.html.erb
@@ -33,7 +33,7 @@
     <% if user_signed_in? %>
       <p>
         <a href="<%= user_url(current_user) %>">
-          <img src="https://d2fltix0v2e0sb.cloudfront.net/dev-badge.svg" alt="<%= current_user.name %>'s <%= community_name %> Profile" height="30" width="30" class="dev-badge">
+          <img src="https://d2fltix0v2e0sb.cloudfront.net/dev-badge.svg" alt="<%= current_user.name %>'s <%= community_name %> Profile" height="30" width="30" class="dev-badge" />
         </a>
       </p>
       <p>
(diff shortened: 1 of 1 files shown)


rhymes:
Hi @JulianG! Thanks for the PR!

According to HTML specs the `` tag does not need self closing: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img

where did you get that error?

JulianG:
Hi @rhymes. I suppose you're right. I've must have been writing JSX for so long that I forgot.

The error happens when you paste the embed code in JSX.

I can now see this is not necessarily a problem. Users should know how to change the embed code to make it compatible with JSX. However I see no harm in making it compatible.

Perhaps we could leave the actual markup alone, and only add a forward slash to the embed code.

In any case I understand if you prefer not to merge at all.

rhymes:
Hi @JulianG, thanks for the clarification. I think we should leave it alone in this case.

JulianG:
I agree. I'll close this PR

Issue #3583, opened 2019-07-30
Duplicate notifications


shushugah:
**Describe the bug**

When someone tags me in a comment while replying to me, I get notified twice. 

**To Reproduce**
It occurs in this url, https://dev.to/notifications/comments see image below

**Expected behavior**
Should be one notification

**Screenshots**

![Screen Shot 2019-07-30 at 20 56 57](https://user-images.githubusercontent.com/7111514/62157220-f4d3a000-b30c-11e9-9684-105e11c51804.png)

**Desktop (please complete the following information):**

- OS: Mac OS
- Browser: Chrome

Zhao-Andy:
Thanks for the report! My first guess is that we can prevent mention notifications from firing if the person mentioned is receiving notifications for a post or comment thread already.

jessleenyc:
Once #3672 is complete there should be better documentation on how to set up notifications locally. There's already some info in that issue for reference.

Janglee123:
I want to work on this issue for hacktoberfest. But dont know where to start.

blunket:
Hey @Janglee123, check out these files! Maybe this will help you a bit?

- [app/services/mentions/create_all.rb](https://github.com/thepracticaldev/dev.to/blob/master/app/services/mentions/create_all.rb)
- [spec/services/mentions/create_all_spec.rb](https://github.com/thepracticaldev/dev.to/blob/master/spec/services/mentions/create_all_spec.rb)
- [app/views/notifications/_mention.html.erb](https://github.com/thepracticaldev/dev.to/blob/master/app/views/notifications/_mention.html.erb)
- [spec/requests/notifications_spec.rb](https://github.com/thepracticaldev/dev.to/blob/master/spec/requests/notifications_spec.rb#L349)

benhalpern:
Yup @blunket you're onto it correctly. @Janglee123 This is going to be a matter of adding some conditional in the logic where the `mention` is created so that it checks whether the mentioned user will not get a notification by other logic.

blunket:
@Janglee123 If you check out the Comments controller, you can find where the NotificationSubscription and Notification are both created.

https://github.com/thepracticaldev/dev.to/blob/master/app/controllers/comments_controller.rb#L59

Maybe that helps more?

ledestin:
I understand that if a notification is present for the comment and the user, mention shouldn't be created. Correct?

I'm trying to write a comment request spec, and notification isn't being created when another user replies to my article:
```
[...]

PR #7880, opened 2020-05-15
Refactor Mentions::CreateAll

ledestin:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I'm working on the issue #3583 and to understand how mentions work, I've done this refactoring. It's not a fix, it's just a preparation for me to understand how things work.

#3583

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/services/mentions/create_all.rb
+++ b/app/services/mentions/create_all.rb
@@ -10,7 +10,7 @@ module Mentions
 
     def call
       # Only works for comments right now.
-      mentioned_users = users_mentioned_in_notifiable_text_that_are_existing_and_not_its_owners
+      mentioned_users = users_mentioned_in_text_excluding_author
 
       delete_mentions_removed_from_notifiable_text(mentioned_users)
       create_mentions_for(mentioned_users)
@@ -18,23 +18,21 @@ module Mentions
 
     private
 
-    def users_mentioned_in_notifiable_text_that_are_existing_and_not_its_owners
+    def users_mentioned_in_text_excluding_author
       mentioned_usernames = extract_usernames_from_mentions_in_text
 
       collect_existing_users(mentioned_usernames).
-        then do |existing_mentioned_users|
-          reject_notifiable_owner(existing_mentioned_users)
+        yield_self do |existing_mentioned_users|
+          reject_notifiable_author(existing_mentioned_users)
         end
     end
 
     def collect_existing_users(usernames)
-      usernames.filter_map do |username|
-        User.find_by(username: username)
-      end
+      User.where(username: usernames)
     end
 
     def create_mentions_for(users)
-      users.map { |user| create_mention_for(user) }
+      users.each { |user| create_mention_for(user) }
     end
 
     def extract_usernames_from_mentions_in_text
@@ -45,11 +43,11 @@ module Mentions
       end
     end
 
-    def reject_notifiable_owner(users)
-      users.reject { |user| user_owns_notifiable?(user, @notifiable) }
+    def reject_notifiable_author(users)
+      users.reject { |user| authored_by?(user, @notifiable) }
     end
 
-    def user_owns_notifiable?(user, notifiable)
+    def authored_by?(user, notifiable)
       user.id == notifiable.user_id
     end
 
--- a/app/services/mentions/create_all.rb
+++ b/app/services/mentions/create_all.rb
@@ -10,32 +10,55 @@ module Mentions
 
     def call
       # Only works for comments right now.
-      # Paired with the process that creates the "comment-mentioned-user"
-      doc = Nokogiri::HTML(notifiable.processed_html)
-      usernames = []
-      mentions = []
-      doc.css(".comment-mentioned-user").each do |link|
-        username = link.text.delete("@").downcase
-        user = User.find_by(username: username)
+      mentioned_users = users_mentioned_in_notifiable_text_that_are_existing_and_not_its_owners
 
-        if user && user.id != notifiable.user_id
-          usernames << username
-          mentions << create_mention(user)
-        end
-      end
-      delete_removed_mentions(usernames)
-      mentions
+      delete_mentions_removed_from_notifiable_text(mentioned_users)
+      create_mentions_for(mentioned_users)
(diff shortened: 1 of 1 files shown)


app/services/mentions/create_all.rb
```diff
@@ -10,32 +10,55 @@ def self.call(*args)
...
       # Only works for comments right now.
+      mentioned_users = users_mentioned_in_notifiable_text_that_are_existing_and_not_its_owners
+
+      delete_mentions_removed_from_notifiable_text(mentioned_users)
+      create_mentions_for(mentioned_users)
+    end
+
+    private
+
+    def users_mentioned_in_notifiable_text_that_are_existing_and_not_its_owners
```
vaidehijoshi, line 21: This method name feels a bit verbose to me; how would you feel about shortening it and perhaps adding a clarifying comment?

app/services/mentions/create_all.rb
```diff
@@ -10,32 +10,55 @@ def self.call(*args)
...
+      mentioned_usernames = extract_usernames_from_mentions_in_text
+
+      collect_existing_users(mentioned_usernames).
+        then do |existing_mentioned_users|
```
vaidehijoshi, line 25: Oh wow, TIL about `then` in Ruby! I'd only ever heard of `yield_self` :)

app/services/mentions/create_all.rb
```diff
@@ -10,32 +10,55 @@ def self.call(*args)
...
+          reject_notifiable_owner existing_mentioned_users
```
vaidehijoshi, line 26:
```suggestion
          reject_notifiable_owner(existing_mentioned_users)
```
Just to follow the convention in the rest of this file/codebase :)

app/services/mentions/create_all.rb
```diff
@@ -10,32 +10,55 @@ def self.call(*args)
...
     end
 
-    private
+    def reject_notifiable_owner(users)
+      users.reject { |user| user_owns_notifiable?(user, @notifiable) }
+    end
+
+    def user_owns_notifiable?(user, notifiable)
+      user.id == notifiable.user_id
+    end
```
vaidehijoshi, line 54: It seems to me that this doesn't really need to be a method since it's only ever used in `reject_notifiable_owner`'s `reject` block; we could probably just do it inline there.

app/services/mentions/create_all.rb
```diff
@@ -10,32 +10,55 @@ def self.call(*args)
 
     def call
       # Only works for comments right now.
+      mentioned_users = users_mentioned_in_notifiable_text_that_are_existing_and_not_its_owners
```
vaidehijoshi, line 13: I wonder if `mentioned_usernames` and `mentioned_users` could just become memoized instance variables that are set the first time that they are called, and then reused in the service whenever necessary? Something like:

```ruby
def mentioned users
  @mentioned_users ||= ...
end
```

app/services/mentions/create_all.rb
citizen428: I find the `then` alias a bit confusing and less intention-revealing than `yield_self`.

app/services/mentions/create_all.rb
lightalloy: As I see, currently, `Mentions::CreateAll` is only created (and `call` called) once inside a worker, so memoization won't do much for now.

app/services/mentions/create_all.rb
```diff
@@ -10,32 +10,55 @@ def self.call(*args)
...
+      mentioned_usernames = extract_usernames_from_mentions_in_text
+
+      collect_existing_users(mentioned_usernames).
+        then do |existing_mentioned_users|
+          reject_notifiable_owner existing_mentioned_users
+        end
+    end
+
+    def collect_existing_users(usernames)
+      usernames.filter_map do |username|
```
lightalloy, line 31: We could find users with one query `User.where(username: usernames)`, am I right?

app/services/mentions/create_all.rb
```diff
@@ -10,32 +10,55 @@ def self.call(*args)
...
+        User.find_by(username: username)
+      end
+    end
+
+    def create_mentions_for(users)
+      users.map { |user| create_mention_for(user) }
```
lightalloy, line 37: I think `each` would make more sense here since we don't use the result of `map`.

app/services/mentions/create_all.rb
citizen428: Yes, we can 👍 This will also implicitly takes care of the filtering.

ledestin: Oh, such a great idea!

app/services/mentions/create_all.rb
ledestin: Hi @vaidehijoshi, thanks for help.

Regarding the long name, I was inspired by [Naming as a process](https://www.digdeeproots.com/articles/on/naming-as-a-process/) series. It's called "a honest name", and the next step would be to replace it with a name that reveals domain intent.

Fortunately, I was suggest a better shorter name.

app/services/mentions/create_all.rb
ledestin: I strongly disagree. It's easier to understand and `authored_by?(user, notifiable)` (improved on the previous name) is revealing a domain concept.

app/services/mentions/create_all.rb
ledestin: @vaidehijoshi I think line 13 is revealing where mentioned users are coming from, so I like it that way.

app/services/mentions/create_all.rb
ledestin: I don't mind `yield_self`, so I switched to that.

ledestin:
Thanks for help! Conformed to all suggestions, except one from @vaidehijoshi , and changed `owner` -> `author` everywhere.

PR #7890, opened 2020-05-15
Limit Response Template Count for Users

mstruve:
- [x] Bug Fix

## Description
Limit the number of templates a user can make to 30 unless they are a trusted user. I choose to exclude trusted users since many of those are mods who might need more templates. As it stands the user with the most templates(who is a mod actually) has 19. I figured a normal user would likely have much less and figured 30 was more than enough to cover them

Screen Shot 2020-05-15 at 1 05 53 PM

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/spec/models/response_template_spec.rb
+++ b/spec/models/response_template_spec.rb
@@ -27,4 +27,22 @@ RSpec.describe ResponseTemplate, type: :model do
       end
     end
   end
+
+  describe "user validation" do
+    it "validates the number of templates for a normal user" do
+      user = create(:user)
+      create_list(:response_template, 29, user_id: user.id)
+      invalid_template = create(:response_template, user_id: user.id)
+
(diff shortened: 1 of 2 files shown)


spec/models/response_template_spec.rb
```diff
@@ -27,4 +27,22 @@
       end
     end
   end
+
+  describe "user validation" do
+    it "validates the number of templates for a normal user" do
+      user = create(:user)
+      create_list(:response_template, 29, user_id: user.id)
```
maestromac, line 34: Shouldn't this be 30 because of "limit of 30"? and it should raise error on the 31st attempt?
```suggestion
      create_list(:response_template, 30, user_id: user.id)
```

mstruve: Good catch, had to update the validation as well

PR #7891, opened 2020-05-15
Sample Noisy Honeycomb Spans

mstruve:
- [x] Feature
- [x] Optimization

## Description
This is a second pass at #7833. We have some spans that are super noisy in Honeycomb and not very valuable so rather than capturing every single one this allows us to sample them. 
 
## Added tests?
- [x] yes

alt_text


app/services/search/client.rb
```diff
@@ -33,7 +33,6 @@ def respond_to_missing?(method, _include_all = false)
       ].freeze
 
       def request
-        Honeycomb.add_field("app.name", "elasticsearch")
```
mstruve, line 36: This was not tracking what I wanted it to so I am cleaning it up here.

PR #7892, opened 2020-05-15
Limit User Updates

mstruve:
- [x] Bug Fix

## Description
Limit the number of times a user can update themselves and make it configurable. Default to 5 times in 30 seconds

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/helpers/rate_limit_checker_helper.rb
+++ b/app/helpers/rate_limit_checker_helper.rb
@@ -2,8 +2,13 @@ module RateLimitCheckerHelper
   CONFIGURABLE_RATES = {
     rate_limit_article_update: {
       min: 1,
-      placeholder: 150,
+      placeholder: 30,
       description: "The number of article updates a user can make in 30 seconds"
+    },
+    rate_limit_user_update: {
+      min: 1,
+      placeholder: 5,
+      description: "The number of user updates a user can make in 30 seconds"
     },
     rate_limit_feedback_message_creation: {
       min: 1,
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -81,9 +81,10 @@ class SiteConfig < RailsSettings::Base
   field :rate_limit_reaction_creation, type: :integer, default: 10
   field :rate_limit_image_upload, type: :integer, default: 9
   field :rate_limit_email_recipient, type: :integer, default: 5
-  field :rate_limit_article_update, type: :integer, default: 150
+  field :rate_limit_article_update, type: :integer, default: 30
   field :rate_limit_send_email_confirmation, type: :integer, default: 2
   field :rate_limit_feedback_message_creation, type: :integer, default: 5
+  field :rate_limit_user_update, type: :integer, default: 5
 
   # Google Analytics Reporting API v4
   # <https://developers.google.com/analytics/devguides/reporting/core/v4>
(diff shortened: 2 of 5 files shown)


app/helpers/rate_limit_checker_helper.rb
```diff
@@ -2,9 +2,14 @@ module RateLimitCheckerHelper
   CONFIGURABLE_RATES = {
     rate_limit_article_update: {
       min: 1,
-      placeholder: 150,
+      placeholder: 30,
```
mstruve, line 5: I recently lowered this in our system to 30 which I think is much more practical so I changed the placeholder and default.

app/services/rate_limit_checker.rb
```diff
@@ -4,13 +4,14 @@ class RateLimitChecker
   # retry_after values are the seconds until a user can retry an action
   ACTION_LIMITERS = {
     article_update: { retry_after: 30 },
-    send_email_confirmation: { retry_after: 120 },
```
mstruve, line 7: Just a little alphabetizing 😉

spec/models/user_spec.rb
```diff
@@ -206,9 +206,25 @@ def user_from_authorization_service(service_name, signed_in_resource, cta_varian
...
+      user = create(:user)
+      limiter = RateLimitChecker.new(user)
+      allow(user).to receive(:rate_limiter).and_return(limiter)
+      allow(limiter).to receive(:limit_by_action).and_return(true)
+      allow(limiter).to receive(:track_limit_by_action)
+      user.update(articles_count: 5)
+
+      expect(user).not_to be_valid
+      expect(user.errors[:base].to_s).to include("could not be saved. Rate limit reached")
+      expect(limiter).to have_received(:track_limit_by_action).with(:user_update).twice
```
mstruve, line 227: I added this here and above bc this spec was passing but when I tested the code locally it wasn't working bc I never actually tracked the updates. This ensures that if the tracking for some reason is removed, this and the above spec will break.

app/models/site_config.rb
```diff
@@ -81,9 +81,10 @@ class SiteConfig < RailsSettings::Base
   field :rate_limit_reaction_creation, type: :integer, default: 10
   field :rate_limit_image_upload, type: :integer, default: 9
   field :rate_limit_email_recipient, type: :integer, default: 5
-  field :rate_limit_article_update, type: :integer, default: 150
+  field :rate_limit_article_update, type: :integer, default: 30
```
Ridhwana, line 84: Just a thought, are we not able to read the default from rate_limit_checker_helper.rb here,  so that they stay in sync?

mstruve: I think we would want to do it the other way around, read THIS default into the helper. I will write that down as a refactor to do in a future PR. Good call!

app/models/user.rb
```diff
@@ -624,4 +625,13 @@ def can_send_confirmation_email
   rescue RateLimitChecker::LimitReached => e
     errors.add(:email, "confirmation could not be sent. #{e.message}")
   end
+
+  def update_rate_limit
+    return unless persisted?
```
maestromac, line 630: Nice! I would've easily overlooked this

mstruve: I missed it with the confirmation email which is why I caught it here. Experience really is directly related to the number of mistakes you make 😅

PR #7894, opened 2020-05-15
Remove Unused Pro Membership Feature

mstruve:
- [x] Refactor

## Description
Remove Pro Membership Code but leave `pro?` method on user. Since we have handy analytics that are useful to users with the `pro` role I switched the `pro?` method just to look for the role. This will allow those analytics to still be used but also allows us to remove the membership code. The 5 users who had pro memberships I gave the role too so everyone should be covered. 

In the future. **I would like to rename the `pro` role `analytics`** so its more descriptive about what it allows the user to do. 

Travis can be ignored as its a CVE issue and CodeClimate seems to have problems when files are removed.

## Added tests?

alt_text


app/models/user.rb
```diff
@@ -291,8 +290,8 @@ def tech_admin?
   end
 
   def pro?
-    Rails.cache.fetch("user-#{id}/has_pro_membership", expires_in: 200.hours) do
-      pro_membership&.active? || has_role?(:pro)
+    Rails.cache.fetch("user-#{id}/has_pro_role", expires_in: 200.hours) do
```
nickytonline, line 293: If there is no longer a pro feature, why do we even have this `def`?

app/views/articles/show.html.erb
```diff
@@ -205,7 +205,7 @@
   <% end %>
 </div>
 
-<% stick_nav_cache_key = "sticky-sidebar-#{@article.id}-#{(@organization || @user).profile_updated_at}-#{(@organization || @user).last_article_at}-#{@variant_number}-#{(@organization || @user).pro?}" %>
+<% stick_nav_cache_key = "sticky-sidebar-#{@article.id}-#{(@organization || @user).profile_updated_at}-#{(@organization || @user).last_article_at}-#{@variant_number}-#{@user.pro?}" %>
```
nickytonline, line 208: Same question about why we still have a `pro` property,

app/models/user.rb
mstruve: There is no longer a "Pro Membership" feature. We still have the concept of `pro` role which allows users access to analytics. Following this PR I want to rename that role `analytics` so that is makes more sense.

rhymes:
Ah, should we remove the table as well or in a separate PR?

mstruve:
>  what about /dashboard/pro ?

Discovered this should stay as its an analytics dashboard.

> Ah, should we remove the table as well or in a separate PR?

Was going to do a separate PR for database cleanup

config/routes.rb
```diff
@@ -223,12 +223,10 @@
   resources :partnerships, only: %i[index create show], param: :option
   resources :display_ad_events, only: [:create]
   resources :badges, only: [:index]
-  resource :pro_membership, path: :pro, only: %i[show create update]
   resources :user_blocks, param: :blocked_id, only: %i[show create destroy]
   resources :podcasts, only: %i[new create]
   resources :article_approvals, only: %i[create]
   resources :video_chats, only: %i[show]
-  resolve("ProMembership") { [:pro_membership] } # see https://guides.rubyonrails.org/routing.html#using-resolve
```
vaidehijoshi, line 231: me: "what is this"
also me: "guess i don't need to know now!!"

vaidehijoshi: (i still clicked the guides link, lol)

vaidehijoshi:
Approved, pending green specs!  ✅

config/routes.rb
rhymes: I think I originally wrote that line. I have no idea what it does, I didn't click the link, I want to forget how I got there :D

Issue #7835, opened 2020-05-13
Replace all hardcoded Twitter Hashtags with the twitter_hashtag site config variable


Ridhwana:
In [this issue](https://github.com/thepracticaldev/dev.to/issues/7254) and [this PR](https://github.com/thepracticaldev/dev.to/pull/7573) the twitter hashtag site config variable was added. 

The next step is to now replace all hardcoded #DEVCommunity references with this dynamic option.

Ridhwana:
To our amazing contributors: This issue is labeled as ready for dev, so it's up for grabs.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

icncsx:
@Ridhwana If this is not taken already, I can take this!

lightalloy:
@icncsx you're welcome to work on this issue :sparkles:

icncsx:
@lightalloy please see the relevant [PR](https://github.com/thepracticaldev/dev.to/pull/7895). 
@Ridhwana thank you for linking the relevant PR and issue.

maestromac:
Resolved with https://github.com/thepracticaldev/dev.to/pull/7895

PR #7895, opened 2020-05-16
replace hardcoded twitter hashtag

icncsx:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
In [this PR](https://github.com/thepracticaldev/dev.to/pull/7573/commits) the twitter hashtag site config variable was added. The next step was to replace all hardcoded #DEVCommunity references with w/ the dynamic `SiteConfig.twitter_hashtag` variable set to "DevCommunity" in `site_config.rb`. Tests were also updated to reflect the dynamic addition.

This PR is in direct relation to [this issue](https://github.com/thepracticaldev/dev.to/issues/7835#event-3340045458)

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -31,7 +31,7 @@ class SiteConfig < RailsSettings::Base
     instagram: nil,
     twitch: nil
   }
-  field :twitter_hashtag, type: :string, default: "DEVCommunity"
+  field :twitter_hashtag, type: :string, default: "#DEVCommunity"
 
   # Emails
   field :email_addresses, type: :hash, default: {
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -31,7 +31,7 @@ class SiteConfig < RailsSettings::Base
     instagram: nil,
     twitch: nil
   }
-  field :twitter_hashtag, type: :string
+  field :twitter_hashtag, type: :string, default: "DEVCommunity"
 
   # Emails
   field :email_addresses, type: :hash, default: {
--- a/spec/labor/bufferizer_spec.rb
+++ b/spec/labor/bufferizer_spec.rb
@@ -19,10 +19,10 @@ RSpec.describe Bufferizer, type: :labor do
   end
 
   it "sends to buffer sattelite twitter" do
-    tweet = "test tweet ##{SiteConfig.twitter_hashtag}"
+    tweet = "test tweet #{SiteConfig.twitter_hashtag}"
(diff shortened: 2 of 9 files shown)


app/models/site_config.rb
```diff
@@ -31,7 +31,7 @@ class SiteConfig < RailsSettings::Base
     instagram: nil,
     twitch: nil
   }
-  field :twitter_hashtag, type: :string
+  field :twitter_hashtag, type: :string, default: "DEVCommunity"
```
rhymes, line 34: shouldn't the pound sign, `#`, be part of the string? After all the variable name is `twitter hashtag`.

icncsx: @rhymes That makes sense! By the way, I've noticed that %23 is used for URL encoding purposes. Would it be appropriate to then use "%23DEVCommunity" as the default value or "#DEVCommunity" and replace # w/ %23 where appropriate?

spec/labor/bufferizer_spec.rb
```diff
@@ -19,10 +19,10 @@
   end
 
   it "sends to buffer sattelite twitter" do
-    tweet = "test tweet #DEVCommunity"
+    tweet = "test tweet #{SiteConfig.twitter_hashtag}"
```
lightalloy, line 22: I suggest setting `SiteConfig.twitter_hashtag` explicitly in tests before performing checks that include it.  It would be nice not to rely on the default value, especially regarding that we plan to remove the default values in the future, so may accidentally keep tests that check for empty strings then.

spec/requests/buffer_updates_spec.rb
```diff
@@ -33,7 +33,7 @@
       post "/buffer_updates",
            params:
            { buffer_update: { body_text: "This is the text!!!!", tag_id: "javascript", article_id: article.id } }
-      expect(BufferUpdate.first.body_text).to include("#DEVCommunity")
+      expect(BufferUpdate.first.body_text).to include(SiteConfig.twitter_hashtag.to_s)
```
lightalloy, line 36: Same as above.

spec/labor/bufferizer_spec.rb
icncsx: @lightalloy Thanks for the tip!

icncsx:
Closes #7835

PR #7904, opened 2020-05-17
🚀 Feature: Chat Channel Request manager for mods

sarthology:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

#### Feature 
Mods will receive an alert for new requests on their discoverable channels. Which they can manage from a sidecar component.
#### Bug Fix 
Also fixes a small bug coming in mobile screen. 
image

Screen Recording 2020-05-17 at 06 30 PM

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


.vscode/settings.json
```diff
@@ -7,5 +7,6 @@
     }
   },
   "ruby.format": "rubocop",
-  "ruby.intellisense": "rubyLocate"
+  "ruby.intellisense": "rubyLocate",
+  "eslint.enable": false
```
rhymes, line 11: not sure this should be disabled for everyone, what if you just disabled it in your user settings?

app/controllers/chat_channels_controller.rb
```diff
@@ -177,9 +177,11 @@ def render_unopened_ids_response
...
 
   def render_joining_request_json_response
-    requested_memberships = current_user.chat_channel_memberships.includes(:chat_channel).
-      where(chat_channels: { discoverable: true }, role: "mod").pluck(:chat_channel_id).map { |membership_id| ChatChannel.find_by(id: membership_id).requested_memberships }
-    render json: { joining_requests: requested_memberships.flatten }
+    requested_memberships_id = current_user.chat_channel_memberships.includes(:chat_channel).
+      where(chat_channels: { discoverable: true }, role: "mod").pluck(:chat_channel_id).map { |membership_id| ChatChannel.find_by(id: membership_id).requested_memberships }.flatten.map(&:id)
+    @chat_channels_memberships = ChatChannelMembership.includes(%i[user chat_channel]).where(id: requested_memberships_id)
+
+    render "index.json"
```
rhymes, line 184: this is quite hard to read, can we split it in multiple lines? For example:

```ruby
    requested_memberships_id = current_user.chat_channel_memberships.
      includes(:chat_channel).
      where(chat_channels: { discoverable: true }, role: "mod").
      pluck(:chat_channel_id).map do |membership_id|
        ChatChannel.find_by(id: membership_id).requested_memberships
      end.flatten.map(&:id)
    @chat_channels_memberships = ChatChannelMembership.includes(%i[user chat_channel]).where(id: requested_memberships_id)

    render "index.json"
```

Looking at the query I'm also thinking if there is no way to simplify it with:

```ruby
chat_channel_ids = current_user.chat_channel_memberships.includes(:chat_channel).where(chat_channels: { discoverable: true }, role: "mod").pluck(:chat_channel_id)
@chat_channel_memberships = ChatChannelMembership.includes(%i[user chat_channel]).where(status: "joining_request", chat_channel_id: chat_channel_ids)
```

am I missing something?

app/javascript/chat/actions/requestActions.js
```diff
@@ -0,0 +1,71 @@
+export function rejectJoiningRequest(
+  channelId,
+  membershipId,
+  successCb,
+  failureCb,
+) {
+  fetch(`/chat_channel_memberships/remove_membership`, {
```
rhymes, line 7: you could use the new `http/request` util, like this:

```js
import { request } from '@utilities/http';
const response = await request('/path');
```

it'll greatly simplify this code.

see https://github.com/thepracticaldev/dev.to/blob/master/app/javascript/utilities/http/request.js

db/schema.rb
```diff
@@ -310,6 +310,7 @@
     t.string "slug"
     t.string "status", default: "active"
     t.datetime "updated_at", null: false
+    t.index ["slug"], name: "index_chat_channels_on_slug", unique: true
```
rhymes, line 313: this should be removed, my suggestion is to open `psql PracticalDeveloper_development` and `DROP INDEX index_chat_channels_on_slug` as it's a remnant of some old change we don't have in production anymore

app/controllers/chat_channels_controller.rb
sarthology: So this code will be obsolete after next PR. So we didn't optimised it.

app/javascript/chat/actions/requestActions.js
sarthology: Will do it.

.vscode/settings.json
sarthology: Sure. Will disable it.

sarthology:
> Hi @sarthology, how do we manually test this?

@rhymes Search a discoverable channel and send joining request to it. Then login as mod to take action on that request.

app/controllers/chat_channels_controller.rb
rhymes: ah ok!

benhalpern:
I'm going to merge this as it ultimately is pretty straightforward logic and we can get it up.

sarthology:
Thanks ben. 🙂

PR #7910, opened 2020-05-18
Rename classified listings

citizen428:

- [X] Refactor
- [X] Masochism

## Description

During all my recent work on classified listings I noticed that we consistently use the term "listing" externally:

* The link
    Screen Shot 2020-05-18 at 15 00 44
* The actual URL: https://dev.to/listings
* The buttons on the listings page
    Screen Shot 2020-05-18 at 15 01 37
* The API endpoint: https://dev.to/api/listings
* etc

However, in our codebase, they are generally referred to as "classified listings", except for the occasional case where they aren't. As a strong believer in consistency and clarity, I decided to unify it all under the single name listing (and listing category) and this PR makes the necessary changes. 

Things left alone:

* Table names: changing them could be potentially disruptive, so I'm using an explicit `self.table_name=` in the 2 affected models.
* The ES search name, since that can't easily be changed.

n/a

## Added tests?

- [x] yes

## Added to documentation?

- [X] no documentation needed
- [X] added inline comments for clarity where it seemed appropriate


app/views/listings/delete_confirm.html.erb
```diff
@@ -1,18 +1,18 @@
 <%= javascript_include_tag "application" %>
 
 <div class="container delete-confirm">
-  <h4><%= @classified_listing.title %></h4>
+  <h4><%= @lassified_listing.title %></h4>
```
rhymes, line 4: this looks like a typo

docs/api_v0.yml
```diff
@@ -691,7 +691,7 @@ components:
     ListingCreate:
       type: object
       properties:
-        classified_listing:
```
rhymes, line 694: this and the rest of changes are going to break the API though, I think we should either accept both versions of leave it there as is

spec/requests/classified_listings_spec.rb
```diff
@@ -1,27 +1,27 @@
 require "rails_helper"
```
rhymes, line 1: shouldn't this file be renamed?

citizen428:
Thanks for the feedback @rhymes. I expected some things to still be broken, I'm surprised there aren't more of them. It probably means we just haven't found them yet.

> The UI is a bit borked

Expected, I renamed all the CSS classes, something may have slipped through the cracks.

> when I tried to create a listing I got the following error

Caused by an over-eager replace operation, inside the method body it should set `classified_listing_category_id` 😅 Thanks!

app/views/listings/delete_confirm.html.erb
citizen428: Thanks!

spec/requests/classified_listings_spec.rb
citizen428: The problem is we always had two specs here, `classified_listings_spec.rb` and `listings_spec.rb`. So simple renaming won't work. I wanted to figure out if it would be sensible to just merge them (probably is), but decided to tackle that a bit later.

docs/api_v0.yml
citizen428: I'm always a bit unsure about this: we have now on numerous occasions broken the API and justified this by citing the v0 version number. Since we separated internal and external API, this would only be used by tools built by others, do we know if they exist/how many requests they make?

citizen428: @rhymes and I discussed this on Slack and decided to leave this as-is for now and wait for TheGreatAPIRefactor™ before we release the v1 API.

citizen428:
So, this is how writers must feel when they finish a 1000 page novel: happy, exhausted, a litte afraid of what they unleashed on the world and reasonably sure that most people won't read it. In other words, this PR is now ready for review.

citizen428:
Hm, it seems I can't approve the Percy changes myself. Also, it seems quite a few of them are related to things like different avatars or Faker content.

app/controllers/listings_controller.rb
```diff
@@ -11,39 +11,40 @@ class ClassifiedListingsController < ApplicationController
...
-        ClassifiedListing.none
+        Listing.none
       end
 
-    @listings_json = @classified_listings.to_json(JSON_OPTIONS)
-    @displayed_listing_json = @displayed_classified_listing.to_json(JSON_OPTIONS)
+    @listings_json = @listings.to_json(JSON_OPTIONS)
+    @displayed_listing_json = @displayed_listing.to_json(JSON_OPTIONS)
 
+    # TODO: [mkohl] Can we change this to listings-#{params[:category]}?
```
mstruve, line 42: Yes, it will just force the cache to refresh with the new code

app/controllers/api/v0/listings_controller.rb
```diff
@@ -1,48 +1,46 @@
...
+        @listings = @listings.order(bumped_at: :desc)
 
         per_page = (params[:per_page] || 30).to_i
         num = [per_page, 100].min
         page = params[:page] || 1
-        @classified_listings = @classified_listings.page(page).per(num)
+        @listings = @listings.page(page).per(num)
 
-        set_surrogate_key_header ClassifiedListing.table_key, @classified_listings.map(&:record_key)
+        set_surrogate_key_header Listing.table_key, @listings.map(&:record_key)
```
mstruve, line 31: These keys will actually stay the same since the table name is the same. I think that's fine.

citizen428:
@mstruve Agreed regarding the deployment, we have a team OSS planning call tomorrow evening my time, so maybe do it around then, since it's more likely that people in the US will be online too.

lib/data_update_scripts/20200217215802_index_listings_to_elasticsearch.rb
```diff
@@ -1,9 +1,9 @@
 module DataUpdateScripts
```
rhymes, line 1: I don't think we can change the name of this, as the data update script machinery stores the name of the file. This will either result in a situation of "there are missing files to be run" or re-run this script. Not sure which of the two

PR #7911, opened 2020-05-18
Remove User#destroy_empty_dm_channels callback

rhymes:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I noticed this accidentally as users are rightly deleted with `Users::DeleteWorker` both in the frontend and the backend. If you try to delete manually a user in the console with `user.destroy`, `empty_dm_channels.destroy_all` will fail as it's an array, not an ActiveRecord relation.

That got me thinking and I realized this is not needed nor activated normally, as direct message channels are deleted already in 

https://github.com/thepracticaldev/dev.to/blob/decae056f10fe9740769d3154fde08c8794347c8/app/services/users/delete_activity.rb#L49

which calls

https://github.com/thepracticaldev/dev.to/blob/decae056f10fe9740769d3154fde08c8794347c8/app/services/users/cleanup_chat_channels.rb#L4-L6

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -166,7 +166,6 @@ class User < ApplicationRecord
   before_validation :set_config_input
   before_validation :downcase_email
   before_validation :check_for_username_change
-  before_destroy :destroy_empty_dm_channels, prepend: true
   before_destroy :destroy_follows, prepend: true
   before_destroy :unsubscribe_from_newsletters, prepend: true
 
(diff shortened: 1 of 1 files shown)


rhymes:
> Looks good.
> I wonder how we haven't noticed this before :D

@lightalloy

Because it never gets triggered from the admin or self delete, the channels are deleted before we call `user.destroy`

lightalloy:
@rhymes that was more like a philosophical question :D

PR #7912, opened 2020-05-18
Enable comment liquid tag and alias devcomment to it

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In the spirit of generalization, it would be weird to have `{% devcomment id %}` to show comments. Even though the [original liquid comment tag](https://shopify.github.io/liquid/tags/comment/) is intended to embed commented out lines AFAIK we don't need that and so I propose we add `{% comment id %}` to show user's comments.

I also left `devcomment` as an alias, so that existing embeds can keep working forever.

## Added tests?

- [x] yes
- [ ] no, because I need help


app/liquid_tags/null_tag.rb
```diff
@@ -4,7 +4,7 @@ def initialize(tag_name, _markup, _options)
   end
 end
 
-disabled_tags = %w[assign break capture case comment cycle decrement for if ifchanged include increment unless tablerow]
+disabled_tags = %w[assign break capture case cycle decrement for if ifchanged include increment unless tablerow]
```
maestromac, line 7: Just a side note, this is here to disable Liquid's build-in [comment feature](https://shopify.github.io/liquid/tags/comment/) for security reasons. Technically it would now be overwritten by our "dev comment" so It shouldn't be a problem.

PR #7914, opened 2020-05-18
Update safe params VCL for Array params

atsmith813:
- [x] Refactor
- [x] Optimization

## Description
In the past, we've had issues trying to add Array parameters so the safe list in our VCL snippets in Fastly. This PR demonstrates an updated `safe_params_list.vcl` to work with Array params.

~**Next steps:** if we want to refactor this VCL, just let me know! We can go about it a few ways:~
~- Comment the params that need to be added if we remove the entire guard of `/search` and `/bulk_show` endpoints and I'll add them in and mark this ready for review.~
~- A contributor pulls this PR down and updates the list themself.~
~- Do nothing, I'll close this out and we'll have it for future reference.~

~_I'm leaving this as draft for now as it is a demonstration and not ready to be merged._~

You can see how the updates snippet works on this [Fastly fiddle](https://fiddle.fastlydemo.net/fiddle/1957a5cf) and [here](https://docs.fastly.com/vcl/functions/urldecode/) is the documentation for `urldecode()`.

This PR addresses the comments from https://github.com/thepracticaldev/dev.to/pull/7676#issuecomment-624814011 and https://github.com/thepracticaldev/dev.to/pull/7676#discussion_r421628638 by adding `ids[]` to the safe list for `/bulk_show`. By decoding the URL first, we can just add `ids\[\]` to our safe list.

https://github.com/thepracticaldev/dev.to/pull/7676#issuecomment-624814011
https://github.com/thepracticaldev/dev.to/pull/7676#discussion_r421628638

## Added tests?
See the [fiddle](https://fiddle.fastlydemo.net/fiddle/1957a5cf) for testing.

## Added to documentation?
- [x] no documentation needed

decoded_gif

### DIFF
--- a/config/fastly/snippets/safe_params_list.vcl
+++ b/config/fastly/snippets/safe_params_list.vcl
@@ -1,7 +1,9 @@
 import querystring;
 sub vcl_recv {
-    # return this URL with only the parameters that match this regular expression
-    if (req.url !~ "/internal/" && req.url !~ "/search/" && req.url !~ "/bulk_show") {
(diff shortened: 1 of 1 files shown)


mstruve:
@rhymes neither did we until I contacted support and they revealed it to us. They really need to document it more.

atsmith813:
Posting this here for reference, Fastly support came up with another solution where you don't decode the URL and you add `ids%255B%255D` to the regex instead. I prefer the decoding because it's easier to read so I'm going to leave it as is.

https://fiddle.fastlydemo.net/fiddle/1957a5cf

PR #7919, opened 2020-05-18
Add public reactions count columns

Zhao-Andy:
- [x] Feature

## Description
This adds the `public_reactions_count` columns to the articles and comments tables. This column will be replacing the `positive_reactions_count` columns soon.

https://github.com/thepracticaldev/dev.to/pull/7777#issuecomment-628165795

## Added tests?

## Added to documentation?
- [x] no documentation needed

Make sure to `[deploy]` this and to ensure that the migration is successful.

### DIFF
--- a/db/migrate/20200514212601_add_public_reactions_count_columns.rb
+++ b/db/migrate/20200514212601_add_public_reactions_count_columns.rb
@@ -1,0 +1,8 @@
+class AddPublicReactionsCountColumns < ActiveRecord::Migration[5.2]
+  def change
+    add_column :articles, :public_reactions_count, :integer, default: 0, null: false
+    add_column :articles, :previous_public_reactions_count, :integer, default: 0, null: false
+
+    add_column :comments, :public_reactions_count, :integer, default: 0, null: false
+  end
+end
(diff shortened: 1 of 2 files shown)


Zhao-Andy:
My current plan is to do this in four steps:
1. make a PR that creates the public count columns
2. Merge and backfill data
3. Make another PR that moves over functionality to public count columns
4. Make a last PR to remove positive reactions count columns

I think this is safer than simply renaming the columns and renaming the columns in the right places of the code... right?

db/migrate/20200514212601_add_public_reactions_count_columns.rb
```diff
@@ -0,0 +1,8 @@
+class AddPublicReactionsCountColumns < ActiveRecord::Migration[5.2]
+  def change
+    add_column :articles, :public_reactions_count, :integer, default: 0, null: false
+    add_column :articles, :previous_public_reactions_count, :integer, default: 0, null: false
```
msarit, line 4: Why are we adding this column?

Zhao-Andy: We have a `previous_positive_reactions_count` column that we'll want to replace too :)

PR #7920, opened 2020-05-18
Decouple Twitter authentication from fetching tweets

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Similar to #7660, here we decouple the concept of Twitter identity from fetching tweets for the Twitter liquid tag.

Rate limit wise we have [300 each 15 minutes](https://developer.twitter.com/en/docs/tweets/post-and-engage/api-reference/get-statuses-show-id) which should be okay. If we start seeing rate limit issues we can make the client smarter, as unfortunately the Twitter API doesn't seem to support HTTP caching. Which shouldn't be an issue anyway because Tweets are saved on disk and never updated anyway.

ps. the reason why the name of the namespace is `TwitterClient::` is because the [twitter gem](https://rubygems.org/gems/twitter) already sits in `Twitter`

#7660 

## Added tests?

- [x] yes
- [ ] no, because I need help


app/services/twitter_client/client.rb
```diff
@@ -0,0 +1,72 @@
+module TwitterClient
+  # Twitter client (users twitter gem as a backend)
+  class Client
```
rhymes, line 3: I know the various HTTP clients have code in common by now but it's still a bit early to group the variations in a common base class. There's still a bit of refactoring to be done

spec/fixtures/approvals/rssreader/fetch_user/gets_articles_for_user.approved.txt
```diff
@@ -1 +1 @@
-1
\ No newline at end of file
+10
```
rhymes, line 1: these change because I re-fetched all live RSS feeds, so the articles have increased by now :D

spec/support/initializers/vcr.rb
```diff
@@ -17,21 +17,17 @@
 
   # Removes all private data (Basic Auth, Set-Cookie headers...)
   config.before_record do |i|
+    # Twitter embeds the Bearer access token in the JSON HTTP response
```
rhymes, line 20: damn you Twitter API :D

mstruve: Nice catch!

app/services/twitter_client/client.rb
mstruve: 100% agree, let's finish this refactor first then see what we can consolidate. That method worked great for Search.

PR #7924, opened 2020-05-18
Remove Unnecessary User Touch From Stripe Controller

mstruve:
- [x] Optimization

## Description
Saving a user here seems to be completely unnecessary so I am removing it. When we save a user we kick off a bunch of background jobs that waste resources so we should avoid saving them when we don't need to. 

I thought maybe the reason for this was cache-busting but we do that all async so it wouldn't even matter. 

## Added tests?

alt_text

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -87,10 +87,6 @@ class ApplicationController < ActionController::Base
     response.headers["Expires"] = "Fri, 01 Jan 1990 00:00:00 GMT"
   end
 
-  def touch_current_user
-    current_user.touch
-  end
-
   def rate_limit!(action)
     rate_limiter.check_limit!(action)
   end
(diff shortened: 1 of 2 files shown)


rhymes:
> > Saving a user here seems to be completely unnecessary so I am removing it.
> 
> 👍 We should consider adding a sort of log similar to the mod actions, where important account events (credit card added/removed etc) get stored. Updating the timestamp is not super helpful, as we don't know what updated it anyway.

We do have the log of purchases but not the log of actions, great idea!

rhymes:
> > Saving a user here seems to be completely unnecessary so I am removing it.
> 
> 👍 We should consider adding a sort of log similar to the mod actions, where important account events (credit card added/removed etc) get stored. Updating the timestamp is not super helpful, as we don't know what updated it anyway.

We do have the log of purchases but not the log of actions, great idea!

PR #7925, opened 2020-05-18
Fix removal of tags/tag_list validation

atsmith813:
- [x] Bug Fix

## Description
Without going down the rabbit hole of adding case sensitivity to tags in general, this PR is a quick fix to allow a moderator to remove a tag.

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

lowercase_gif

### DIFF
--- a/app/models/tag_adjustment.rb
+++ b/app/models/tag_adjustment.rb
@@ -29,7 +29,7 @@ class TagAdjustment < ApplicationRecord
   end
 
   def article_tag_list
-    errors.add(:tag_id, "selected for removal is not a current live tag.") if adjustment_type == "removal" && article.tag_list.exclude?(tag_name)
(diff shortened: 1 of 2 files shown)


app/models/tag_adjustment.rb
```diff
@@ -29,7 +29,7 @@ def has_privilege_to_adjust?
   end
 
   def article_tag_list
-    errors.add(:tag_id, "selected for removal is not a current live tag.") if adjustment_type == "removal" && article.tag_list.exclude?(tag_name)
+    errors.add(:tag_id, "selected for removal is not a current live tag.") if adjustment_type == "removal" && article.tag_list.none? { |tag| tag.casecmp(tag_name).zero? }
```
atsmith813, line 32: Shoutout to Rubocop for this sorcery...I had it as `downcase` 🤣

PR #7926, opened 2020-05-18
Rename positive_reactions_count to public_reactions_count in logic

Zhao-Andy:
- [x] Refactor

## Description
This moves the functionality from `positive_reactions_count` to `public_reactions_count`. All changes are simply renaming `positive` to `public`.

#7919
#7932 

## Added tests?

## Added to documentation?
- [x] no documentation needed

Spongebob and Patrick are thinking very, very hard.

### DIFF
--- a/app/controllers/reactions_controller.rb
+++ b/app/controllers/reactions_controller.rb
@@ -24,10 +24,10 @@ class ReactionsController < ApplicationController
     else
       comments = Comment.
         where(commentable_id: params[:commentable_id], commentable_type: params[:commentable_type]).
-        select(%i[id positive_reactions_count])
+        select(%i[id public_reactions_count])
 
       reaction_counts = comments.map do |comment|
-        { id: comment.id, count: comment.positive_reactions_count }
+        { id: comment.id, count: comment.public_reactions_count }
       end
 
       reactions = if session_current_user_id
--- a/app/controllers/reactions_controller.rb
+++ b/app/controllers/reactions_controller.rb
@@ -10,7 +10,7 @@ class ReactionsController < ApplicationController
       id = params[:article_id]
 
       reactions = if session_current_user_id
-                    Reaction.positive.
+                    Reaction.public_category.
                       where(
                         reactable_id: id,
                         reactable_type: "Article",
@@ -32,12 +32,12 @@ class ReactionsController < ApplicationController
 
       reactions = if session_current_user_id
                     comment_ids = reaction_counts.map { |rc| rc[:id] }
-                    cached_user_positive_reactions(current_user).where(reactable_id: comment_ids)
+                    cached_user_public_reactions(current_user).where(reactable_id: comment_ids)
                   else
                     Reaction.none
                   end
 
-      result = { positive_reaction_counts: reaction_counts }
+      result = { public_reaction_counts: reaction_counts }
     end
 
     render json: {
@@ -95,9 +95,9 @@ class ReactionsController < ApplicationController
     render json: { result: result, category: category }
   end
 
-  def cached_user_positive_reactions(user)
-    Rails.cache.fetch("cached_user_reactions-#{user.id}-#{user.positive_reactions_count}", expires_in: 24.hours) do
-      user.reactions.positive
+  def cached_user_public_reactions(user)
+    Rails.cache.fetch("cached_user_reactions-#{user.id}-#{user.public_reactions_count}", expires_in: 24.hours) do
+      user.reactions.public_category
     end
   end
 
--- a/app/models/reaction.rb
+++ b/app/models/reaction.rb
@@ -18,6 +18,7 @@ class Reaction < ApplicationRecord
                   }
   counter_culture :user
 
+  scope :positive, -> { where(category: PUBLIC_CATEGORIES) }
   scope :public_category, -> { where(category: PUBLIC_CATEGORIES) }
   scope :readinglist, -> { where(category: "readinglist") }
   scope :for_articles, ->(ids) { where(reactable_type: "Article", reactable_id: ids) }
--- a/app/models/reaction.rb
+++ b/app/models/reaction.rb
@@ -18,7 +18,6 @@ class Reaction < ApplicationRecord
                   }
   counter_culture :user
 
-  scope :positive, -> { where(category: PUBLIC_CATEGORIES) }
   scope :public_category, -> { where(category: PUBLIC_CATEGORIES) }
   scope :readinglist, -> { where(category: "readinglist") }
   scope :for_articles, ->(ids) { where(reactable_type: "Article", reactable_id: ids) }
--- a/app/models/reaction.rb
+++ b/app/models/reaction.rb
@@ -5,6 +5,7 @@ class Reaction < ApplicationRecord
   SEARCH_CLASS = Search::Reaction
 
   CATEGORIES = %w[like readinglist unicorn thinking hands thumbsdown vomit].freeze
+  PUBLIC_CATEGORIES = %w[like readinglist unicorn thinking hands].freeze
   REACTABLE_TYPES = %w[Comment Article User].freeze
   STATUSES = %w[valid invalid confirmed archived].freeze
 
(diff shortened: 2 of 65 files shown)


Zhao-Andy:
Leaving a note to let me merge this so I can monitor the status :)

config/elasticsearch/mappings/feed_content.json
```diff
@@ -84,7 +84,7 @@
     "path": {
       "type": "keyword"
     },
-    "positive_reactions_count": {
+    "public_reactions_count": {
```
mstruve, line 87: I think this was an accidental replacement since you have the public_reactions_count below. Can you switch this back and put a comment here saying this field is deprecated? That way when we have a chance we can remove it.

config/elasticsearch/mappings/users.json
```diff
@@ -38,7 +38,7 @@
     "path": {
       "type": "keyword"
     },
-    "positive_reactions_count": {
```
mstruve, line 41: same as above, can you add a comment here as well.

app/models/user.rb
```diff
@@ -143,8 +143,8 @@ class User < ApplicationRecord
   validate :can_send_confirmation_email
   validate :update_rate_limit
 
-  alias_attribute :positive_reactions_count, :reactions_count
   alias_attribute :public_reactions_count, :reactions_count
+  alias_attribute :positive_reactions_count, :reactions_count
```
Zhao-Andy, line 147: I added this back in, in case any old caches are rely on it, like the cache key in `Reaction.cached_any_reactions_for?`. I don't think it needs it since I'm updating it in the same PR but maybe it does...?

app/models/reaction.rb
```diff
@@ -53,7 +55,7 @@ def count_for_article(id)
 
     def cached_any_reactions_for?(reactable, user, category)
       class_name = reactable.class.name == "ArticleDecorator" ? "Article" : reactable.class.name
-      cache_name = "any_reactions_for-#{class_name}-#{reactable.id}-#{user.reactions_count}-#{user.positive_reactions_count}-#{category}"
+      cache_name = "any_reactions_for-#{class_name}-#{reactable.id}-#{user.reactions_count}-#{user.public_reactions_count}-#{category}"
```
Zhao-Andy, line 58: Think this should be safe, re: comment on `user.rb`

app/models/reaction.rb
```diff
@@ -13,11 +14,12 @@ class Reaction < ApplicationRecord
...
   counter_culture :reactable,
                   column_name: proc { |model|
-                    model.points.positive? ? "positive_reactions_count" : "reactions_count"
+                    PUBLIC_CATEGORIES.include?(model.category) ? "public_reactions_count" : "reactions_count"
                   }
   counter_culture :user
 
-  scope :positive, -> { where("points > ?", 0) }
+  scope :positive, -> { where(category: PUBLIC_CATEGORIES) }
+  scope :public_category, -> { where(category: PUBLIC_CATEGORIES) }
```
Zhao-Andy, line 22: Not my favorite method name: `.public_category` but I can't do `.public` because Ruby 😅

app/models/reaction.rb
```diff
@@ -13,11 +14,12 @@ class Reaction < ApplicationRecord
 
   counter_culture :reactable,
                   column_name: proc { |model|
-                    model.points.positive? ? "positive_reactions_count" : "reactions_count"
+                    PUBLIC_CATEGORIES.include?(model.category) ? "public_reactions_count" : "reactions_count"
                   }
   counter_culture :user
 
-  scope :positive, -> { where("points > ?", 0) }
+  scope :positive, -> { where(category: PUBLIC_CATEGORIES) }
```
Zhao-Andy, line 21: Leaving this in here because we have one cache relying on this.

app/controllers/reactions_controller.rb
```diff
@@ -95,9 +95,9 @@ def create
     render json: { result: result, category: category }
   end
 
-  def cached_user_positive_reactions(user)
-    Rails.cache.fetch("cached_user_reactions-#{user.id}-#{user.positive_reactions_count}", expires_in: 24.hours) do
-      user.reactions.positive
+  def cached_user_public_reactions(user)
+    Rails.cache.fetch("cached_user_reactions-#{user.id}-#{user.public_reactions_count}", expires_in: 24.hours) do
+      user.reactions.public_category
```
Zhao-Andy, line 100: Pointing this section out, since this cache was previously using on `.positive` and therefore the reason I'm keeping the `.positive` scope.

config/elasticsearch/mappings/feed_content.json
Zhao-Andy: Fixed 👍

Zhao-Andy:
I'll also be updating the count columns again in console before this goes live, to make sure the count columns match closely.

app/models/user.rb
mstruve: You can kill this, its only being used in the cache key, there is nothing in the cache that relies on it.

app/services/slack/messengers/moderator_create_comment.rb
```diff
@@ -0,0 +1,42 @@
+module Slack
```
mstruve, line 1: Is this meant to be here?

Zhao-Andy: ...I have no idea. Must be a git issue... Will update 😓

app/models/reaction.rb
mstruve: You also don't need this since what is in the cache is the AR relation results rather than `.positive` scope

```
[30] pry(main)> Rails.cache.fetch("test-key") do
[30] pry(main)*   User.last.reactions.positive
[30] pry(main)* end
Cache read: test-key ({:namespace=>nil, :compress=>true, :compress_threshold=>1024, :expires_in=>3600, :race_condition_ttl=>nil})
Cache generate: test-key ({:namespace=>nil, :compress=>true, :compress_threshold=>1024, :expires_in=>3600, :race_condition_ttl=>nil})
  User Load (27.9ms)  SELECT  "users".* FROM "users" ORDER BY "users"."id" DESC LIMIT $1  [["LIMIT", 1]]
Cache write: test-key ({:namespace=>nil, :compress=>true, :compress_threshold=>1024, :expires_in=>3600, :race_condition_ttl=>nil})
  Reaction Load (15.1ms)  SELECT "reactions".* FROM "reactions" WHERE "reactions"."user_id" = $1 AND (points > 0)  [["user_id", 11]]
=> [#<Reaction:0x00007fe2c21ab448 id: 35, category: "unicorn", created_at: Mon, 18 May 2020 20:09:40 UTC +00:00, points: 1.0, reactable_id: 8, reactable_type: "Article", status: "valid", updated_at: Mon, 18 May 2020 20:09:40 UTC +00:00, user_id: 11>]
[31] pry(main)> Rails.cache.fetch("test-key").count
Cache read: test-key ({:namespace=>nil, :compress=>true, :compress_threshold=>1024, :expires_in=>3600, :race_condition_ttl=>nil})
   (5.4ms)  SELECT COUNT(*) FROM "reactions" WHERE "reactions"."user_id" = $1 AND (points > 0)  [["user_id", 11]]
=> 1
[32] pry(main)> Rails.cache.fetch("test-key")
Cache read: test-key ({:namespace=>nil, :compress=>true, :compress_threshold=>1024, :expires_in=>3600, :race_condition_ttl=>nil})
  Reaction Load (0.5ms)  SELECT "reactions".* FROM "reactions" WHERE "reactions"."user_id" = $1 AND (points > 0)  [["user_id", 11]]
=> [#<Reaction:0x00007fe2c1755598 id: 35, category: "unicorn", created_at: Mon, 18 May 2020 20:09:40 UTC +00:00, points: 1.0, reactable_id: 8, reactable_type: "Article", status: "valid", updated_at: Mon, 18 May 2020 20:09:40 UTC +00:00, user_id: 11>]
```

Zhao-Andy: Ohh interesting. Think I was trying to account for that last AR object cache, but I get it now. 👍

app/services/slack/messengers/moderator_create_comment.rb
Zhao-Andy: Force pushed and resolved

PR #7927, opened 2020-05-18
Drop Elasticsearch Indexes When Database is Dropped in Development

mstruve:
- [x] Bug Fix

## Description
When the database is dropped in development also drop Elasticsearch

## Added tests?

alt_text

### DIFF
--- a/lib/tasks/search.rake
+++ b/lib/tasks/search.rake
@@ -18,4 +18,8 @@ namespace :search do
 
     Search::Cluster.delete_indexes
   end
+
+  if %(development).include?(Rails.env)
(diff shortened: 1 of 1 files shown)


lib/tasks/search.rake
```diff
@@ -18,4 +18,8 @@ namespace :search do
 
     Search::Cluster.delete_indexes
   end
+
+  if %(development).include?(Rails.env)
+    Rake::Task["db:drop"].enhance(["search:destroy"])
```
joshpuetz, line 23: Whoa! 🤯

maestromac: 🤯

PR #7931, opened 2020-05-18
One Liner: Only Count Hot Stories Once Rather Than Every Iteration

mstruve:
- [x] Optimization

## Description
The `count` will actually issue a database query for every iteration. Instead lets count in the database once then store the count in a variable and use that. 

## Added tests?

alt_text

### DIFF
--- a/app/services/articles/feed.rb
+++ b/app/services/articles/feed.rb
@@ -209,7 +209,8 @@ module Articles
         order("hotness_score DESC")
       featured_story = hot_stories.where.not(main_image: nil).first
       if user_signed_in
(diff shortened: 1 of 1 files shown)


lizthegrey:
#foundwithhoneycomb

lizthegrey:
How we found it:

https://share.getcloudapp.com/P8uEKOOk

https://ui.honeycomb.io/thepracticaldev/datasets/rails/result/8pSU2b7BHMr/trace/fBsp7EqruSR

https://ui.honeycomb.io/thepracticaldev/datasets/rails/result/dKJLas6YDu2

https://share.getcloudapp.com/E0ubNOBl

rhymes:
This is awesome! So many queries saved! Thanks @mstruve @lizthegrey !

PR #7932, opened 2020-05-18
Add data update script to backfill public reaction count

Zhao-Andy:
- [x] Data update

## Description
This adds the data update script to backfill the public reaction count.

https://github.com/thepracticaldev/dev.to/pull/7919


lib/data_update_scripts/20200518173504_update_public_reactions_count_from_positive_reactions_count.rb
```diff
@@ -0,0 +1,9 @@
+module DataUpdateScripts
+  class UpdatePublicReactionsCountFromPositiveReactionsCount
+    def run
+      ActiveRecord::Base.connection.execute("UPDATE articles SET public_reactions_count = positive_reactions_count")
+      ActiveRecord::Base.connection.execute("UPDATE articles SET previous_public_reactions_count = previous_positive_reactions_count")
+      ActiveRecord::Base.connection.execute("UPDATE comments SET public_reactions_count = positive_reactions_count")
```
Zhao-Andy, line 6: So this is only 3 transactions!!! I don't think this will fail or timeout even though we have a lot of rows to update here. This should all happen pretty quickly. Not an expert here though so feel free to comment if you feel otherwise!

lib/data_update_scripts/20200518173504_update_public_reactions_count_from_positive_reactions_count.rb
```diff
@@ -0,0 +1,8 @@
+module DataUpdateScripts
+  class UpdatePublicReactionsCountFromPositiveReactionsCount
+    def run
+      ActiveRecord::Base.connection.execute("UPDATE articles SET public_reactions_count = positive_reactions_count, previous_public_reactions_count = previous_positive_reactions_count")
+      ActiveRecord::Base.connection.execute("UPDATE comments SET public_reactions_count = positive_reactions_count")
```
Zhao-Andy, line 5: So this is only 3 transactions!!! I don't think this will fail or timeout even though we have a lot of rows to update here. This should all happen pretty quickly. Not an expert here though so feel free to comment if you feel otherwise!

PR #7935, opened 2020-05-18
Track Deploys In Honeycomb

mstruve:
- [x] Feature

## Description
Track deploys in Honeycomb

alt_text

### DIFF
--- a/config/initializers/honeycomb.rb
+++ b/config/initializers/honeycomb.rb
@@ -25,6 +25,8 @@ else
 
       # Scrub unused data to save space in Honeycomb
       config.presend_hook do |fields|
+        fields["global.build_id"] = ApplicationConfig["HEROKU_SLUG_COMMIT"]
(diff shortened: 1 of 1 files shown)


mstruve:
@benhalpern the reason I want to use the commit slug is that anyone can correlate it with a merged commit in the repo. When we get a public dashboard going people can see when their code is deployed and how it affects the app. Furthermore, commit slugs will always remain the same no matter what platform we are on even if they are not coming from Heroku. The Heroku releases are very specific to Heroku and most people do not have access to them.

lizthegrey:
Why not both?

mstruve:
Even if we add a new field I still think tying us to Heroku releases which are very Heroku specific is not a great idea. I plan on generalizing the slug itself so it's not generated from Heroku so as people spin up communities of their own they can use whatever other platforms they want.

Issue #6499, opened 2020-03-06
Gracefully Handle Missing Linked Users From Comments


mstruve:
We recently received a Honeybadger where we went to save comments that both raised errors due to no being able to find linked users.
Honeybadgers:
https://app.honeybadger.io/fault/66984/ac0e5bc2313665cb976dee458505349b
```
StandardError: Invalid username
 user_tag.rb  25 parse_username_to_user(...)
[PROJECT_ROOT]/app/liquid_tags/user_tag.rb:25:in `parse_username_to_user'
23   def parse_username_to_user(user)
24     user = User.find_by(username: user)
25     raise StandardError, "Invalid username" if user.nil?
26 
27     user
```

In these situations, we should gracefully handle the missing user or comments and NOT raise an error. There will be times when users leave the platform and any internal markdown links should not break their related models because of it.

Math3v:
@mstruve I would like to take a look. Have you got any particular solution in mind? When users or articles "leave" the platform, are they gone for good, or are they only marked as deleted? Thanks!

danascheider:
@Math3v are you still working on this? If not I can take a stab at it.

Math3v:
@danascheider I'm not currently, would be awesome if you had a look at it. Maybe it will be really beneficial if you take a look at my solution and the comments I received from the team. :)

danascheider:
Cool, thanks, I'll take a look at what you worked on and see what I can figure out!

rhymes:
@danascheider I'm going to assign you to the issue :)

danascheider:
I have a couple questions about this issue:
1. If a user is missing, presumably we will want to use some kind of default colours. What should those be?
2. What should we do with the parts of the partial that display data that won't be available for a deleted user, such as the follow button, user Twitter handle, etc.?

One idea I had, going off the comments in #6638, was to use an object like this for the user:
```ruby
OpenStruct.new(username: "username", object: <User object or nil>)
```
Then if the user does exist, we can still use the decorator on the user object in `render`, which wouldn't be as easy if we just used the `:persisted?` key as was suggested in the other PR.  If the `:object` is `nil`, we will know the user doesn't exist and can handle it differently.

Does this sound like a reasonable path forward?

rhymes:
Tagging @mstruve and @citizen428 who have been talking about handling `NullUser`s

mstruve:
@danascheider I am not sure how the url would work but if we want to go the empty user route I would try this. 
[...]

PR #7941, opened 2020-05-18
Gracefully handle missing linked users from comments

danascheider:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

There were issues rendering the liquid `UserTag` when the user in question had been deleted, as described in #6499. This PR fixes the bug.

There is now a `DELETED_USER` constant in the `ApplicationHelper` class, which is an `OpenStruct` object that can be passed into the template in case the user has been deleted. This can be referenced when there is a need to reference a deleted user.

I've also rendered links in the tag conditionally on whether the user is deleted or not.

#6499 

Tag for existing user:
Screen Shot 2020-05-19 at 8 21 41 am

Tag for deleted user:
Screen Shot 2020-05-19 at 8 21 28 am

Note: The "profile image" for the deleted user kept returning a 404 in my local environment. This wasn't a result of the changes I made.

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -2,10 +2,7 @@ module ApplicationHelper
   # rubocop:disable Performance/OpenStruct
   DELETED_USER = OpenStruct.new(
     id: nil,
-    darker_color: HexComparer.new({
-                                    bg: "#19063A",
-                                    text: "#dce9f3"
-                                  }).brightness,
+    darker_color: HexComparer.new(bg: "#19063A", text: "#dce9f3").brightness,
     username: "[deleted user]",
     name: "[Deleted User]",
     summary: nil,
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,6 @@
 module ApplicationHelper
+  DELETED_USER = User.new(username: "[deleted user]", name: "[Deleted User]")
+
   def user_logged_in_status
     user_signed_in? ? "logged-in" : "logged-out"
   end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,5 +1,18 @@
 module ApplicationHelper
-  DELETED_USER = User.new(username: "[deleted user]", name: "[Deleted User]")
+  # rubocop:disable Performance/OpenStruct
+  DELETED_USER = OpenStruct.new(
+    id: nil,
+    darker_color: HexComparer.new({
+                                    bg: "#19063A",
+                                    text: "#dce9f3"
+                                  }).brightness,
+    username: "[deleted user]",
+    name: "[Deleted User]",
+    summary: nil,
+    twitter_username: nil,
+    github_username: nil,
+  )
+  # rubocop:enable Performance/OpenStruct
 
   def user_logged_in_status
     user_signed_in? ? "logged-in" : "logged-out"
@@ -115,6 +128,8 @@ module ApplicationHelper
   end
 
   def follow_button(followable, style = "full")
+    return if followable == DELETED_USER
+
     tag :button, # Yikes
         class: "cta follow-action-button",
         data: {
@@ -129,6 +144,8 @@ module ApplicationHelper
   end
 
   def user_colors(user)
+    return { bg: "#19063A", text: "#dce9f3" } if user == DELETED_USER
+
     user.decorate.enriched_colors
   end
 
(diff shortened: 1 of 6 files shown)


app/models/user.rb
```diff
@@ -40,6 +40,8 @@ class User < ApplicationRecord
   SEARCH_CLASS = Search::User
   DATA_SYNC_CLASS = DataSync::Elasticsearch::User
 
+  DELETED_USER = User.new(username: "[deleted user]", name: "[Deleted User]")
```
mstruve, line 43: Not sure how I feel about this as a constant here. It's nice bc it is very accessible but at the same time we currently only using this in a view which makes it feel like it should be in a helper. Maybe @vaidehijoshi has some thoughts.

spec/liquid_tags/user_tag_spec.rb
```diff
@@ -21,9 +21,11 @@ def generate_user_tag(id_code)
     end
   end
 
-  it "rejects invalid id_code" do
-    expect do
-      generate_user_tag("this should fail")
-    end.to raise_error(StandardError)
+  context "when given an invalid id_code" do
```
mstruve, line 24: I think this is leftover copy/paste from the comment specs. Can you change this to `username` rather than `id_code`?

app/models/user.rb
vaidehijoshi: It makes me a little nervous to have an actual `User` instance here that we're effectively using as a fallback for deciding how to render a null `User` object within `UserTag` (somehow it feels like it has the potential to be buggy down the road?). I definitely agree that we could be using helpers here, instead!

danascheider:
Thanks @vaidehijoshi and @mstruve! I've moved the `DELETED_USER` constant into the `ApplicationHelper`. Let me know if that works or if you'd rather I create a more specific helper module.

app/helpers/application_helper.rb
```diff
@@ -1,4 +1,6 @@
 module ApplicationHelper
+  DELETED_USER = User.new(username: "[deleted user]", name: "[Deleted User]")
```
citizen428, line 2: Instead of creating an actual user instance, we could also create an `OpenStruct`. This could have the extra advantage that we can also override things that are defined as methods for `User`, so we don't need the conditional check for `path` in the Liquid tag:

```ruby
OpenStruct.new(username: "[deleted user]", name: "[Deleted User]", path: nil)
```

Pros:
* Removes conditional logic
* Addresses @vaidehijoshi's concerns about using an actual `User` object.

Con:
* Needs to be kept up to date with how we use this null object, though to a certain degree that's always true.

citizen428:
This is more of a general comment for us internally and nothing that needs to be fixed in this PR: @mstruve recently also introduced a null object, so maybe it's time we talk as a team how we handle this going forward. I quite like this approach and we can build on that:

https://medium.com/@vonchristian/how-to-use-null-object-pattern-with-ruby-on-rails-a4168400ad56

app/helpers/application_helper.rb
danascheider: There's definitely a part of me that likes this idea and considered it, but another part of me wasn't sure it was desirable to create an `OpenStruct` with as many keys as there are methods called on the user (`:username`, `:name`, `:path`, `:summary`, `:twitter_username`, `:github_username`). Still, I don't love having the unpersisted `User` object either so perhaps this makes sense.

danascheider:
@mstruve @citizen428 I've made the change to using an OpenStruct instead of the unpersisted User object. I had to do a couple funny things with the decorator methods, as well as I had to disable a Rubocop cop that requires the use of `Struct` instead of `OpenStruct` - I'm assuming you suggested `OpenStruct` for a reason so I went with that. Let me know what you think.

app/helpers/application_helper.rb
citizen428: I don't think we need keys for all methods of `User`. If we have a  generic `NullObject` to inherit from that sets a couple of sensible defaults for things like `persisted?` I'd be happy to only have things we actually used defined on the specific objects.

citizen428:
The reason I had suggested `OpenStruct` was simply that it doesn't need to be predefined. I'm open to all approaches, including a plain Ruby object, a singleton, or anything else that would make sense.

danascheider:
Thanks for clarifying @citizen428 - if it looks good the way it is I'm happy to stick with that but if you don't want the cops disabled I can do it another way. :)

PR #7946, opened 2020-05-19
Refactored creating sponsorships

lightalloy:
- [x] Refactor

## Description
- removed unused functionality related to self-serving media and devrel sponsorships, added a redirect when a user tries to create such sponsorships (e.g. via editing html)
- renamed misleading `LEVELS_WITH_EXPIRATION` (not only those are with expiration date)
- refactored `PartnershipsController#create`
- replaced an exception with a redirect when there are not enough credits 

One of the further improvements could be moving creating sponsorship to a PORO.

TODO:
- make `flash[:notice]` visible

## Added tests?

- [x] yes


app/models/sponsorship.rb
```diff
@@ -1,6 +1,6 @@
 class Sponsorship < ApplicationRecord
   LEVELS = %w[gold silver bronze tag media devrel].freeze
-  LEVELS_WITH_EXPIRATION = %w[gold silver bronze].freeze
+  METAL_LEVELS = %w[gold silver bronze].freeze
```
rhymes, line 3: ahahha I laughed at `METAL_LEVELS` :D love it!

lightalloy: Naming problems :D

PR #7952, opened 2020-05-19
Add "Sign in with Apple" authentication provider

### LINKED ISSUE #7445: Authentication option: Sign in with Apple
As we look to expand authentication options, I think [
Sign in with Apple ](https://developer.apple.com/sign-in-with-apple/) would be a good one to look at.

**Is your feature request related to a problem? Please describe.*

We are limited in the usecases we can provide with future communities by only having Twitter and GitHub right now, and in expanding our auth we need to take baby steps to get this right. I think Apple as another auth method could be a really great step.

I think its _raison d'etre_ fits well with what we're trying to do in terms of privacy and "antifraud" features to ensure people are real users. Of course Apple is just one more giant platform and we'll want to just make this one piece of a broader evolution of our authentication menu.

I think Sign in with Apple is an option we could roll out for DEV itself while also making it immediately available for other communities.

Here is a gem we might want to use...

https://github.com/nhosoya/omniauth-apple

I haven't given this a deep dive technically so I'm not aware of any technical gotchas, but I feel like this wouldn't be an overly difficult feature to get implemented.

I'm nervous about any code we have in the codebase that _assumes_ the current state of just Twitter and GitHub but I think we'd find that pretty quickly in developing this feature and it wouldn't cause critical bugs.

***

This one has a special case that we'd only want to display it in the case of the user being on an Apple device so there's some extra considerations there. I think it might make sense to limit the areas we display the actual log in options to the modal and registration landing pages so we can reduce display complexity, especially with regards to how we are limited by how we do caching across parts of the site.

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR adds support for users to register and log in through [Sign in with Apple](https://developer.apple.com/sign-in-with-apple/)

### TODO

quoting the internal document:

* [ ] add SPF to relay private emails: https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/communicating_using_the_private_email_relay_service
* [x] `username` doesn't exist for Apple, here we're using `first_name.downcase` - is that ok?
* [x] same for name, here we're using `#{first_name} #{last_name}`, is that ok?
* [x] Apple users will rely on `Users::ProfileImageGenerator` for avatar ~do we need a default profile image or the mascot URL is fine?~ 
* [x] we need to make sure that we are in conformity with Apple UX guidelines: https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple/overview/buttons/
* [x] we need SVG icons for Apple logos

Closes #7445
Closes https://github.com/forem/InternalProjectPlanning/issues/4

## Added tests?

- [x] yes
- [ ] no, because I need help


citizen428:
> username doesn't exist for Apple, here we're using `first_name.downcase` - is that ok?

As suggested on Slack, I'd probably use the username part of the relay email address Apple generates. `first_name.downcase` sounds like it has a lot of potential for duplicates (said by a Michael, I know what I'm talking about).

app/services/authentication/providers.rb
```diff
@@ -48,5 +48,11 @@ def self.enabled
     def self.enabled?(provider_name)
       enabled.include?(provider_name.to_sym)
     end
+
+    def self.username_fields
+      Authentication::Providers::Provider.subclasses.map do |subclass|
+        subclass.const_get("USERNAME_FIELD").to_sym
```
citizen428, line 54: I think we can provide reasonably defaults in the base class for `user_username_field`, `user_created_at_field` and `official_name`:

```ruby
def self.official_name
  name
end

def self.user_username_field
  "#{name.underscore}_username".to_sym
end 

def self.user_created_at_field
  "#{name.underscore}_created_at".to_sym
end

def self.settings_url
  raise SubclassResponsibility
end
```

Then we can get rid of the constants and stop defining these methods in every subclass, instead only override them where necessary (e.g. "GitHub" as `official_name`). If you're worried about performance you can cache them in instance variables, but I think most of these will be accessed mostly during initial app load for metaprogramming purposes. 

If we do this, this block would become:

```ruby
Authentication::Providers::Provider.subclasses.map(&:user_username_field).sort
```

app/helpers/authentication_helper.rb
```diff
@@ -14,4 +14,11 @@ def authentication_enabled_providers
       Authentication::Providers.get!(provider_name)
     end
   end
+
+  def current_user_authentication_enabled_providers
+    providers = Authentication::Providers.enabled.to_set & current_user.identities.pluck(:provider).map(&:to_sym).to_set
```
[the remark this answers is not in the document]
rhymes: ah! TIL!

app/models/user.rb
```diff
@@ -126,14 +125,21 @@ class User < ApplicationRecord
...
   validates :username, length: { in: 2..USERNAME_MAX_LENGTH }, format: USERNAME_REGEXP
   validates :username, uniqueness: { case_sensitive: false }, if: :username_changed?
   validates :website_url, :employer_url, url: { allow_blank: true, no_local: true }
   validates :website_url, length: { maximum: 100 }, allow_nil: true
   validates :youtube_url, length: { maximum: 1000 }, format: YOUTUBE_URL_REGEXP, allow_blank: true
 
+  # add validators for provider related usernames
+  Authentication::Providers.username_fields.each do |username_field|
+    # make sure usernames are not empty, to be able to use the database unique index
+    before_validation(proc { |record| record.assign_attributes(username_field => nil) if record.attributes[username_field.to_s] == "" })
```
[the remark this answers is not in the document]
rhymes: this should only be set if the username is nil, so I'm going to to a remix of your suggestion keeping the if! thanks!

app/services/authentication/providers.rb
rhymes: done! :)

app/models/user.rb
```diff
@@ -511,10 +509,11 @@ def temp_name_exists?
   end
 
   def temp_username
-    if twitter_username
-      twitter_username.downcase.gsub(/[^0-9a-z_]/i, "").delete(" ")
-    elsif github_username
-      github_username.downcase.gsub(/[^0-9a-z_]/i, "").delete(" ")
+    Authentication::Providers.username_fields.each do |username_field|
+      value = public_send(username_field)
+      if value.present?
```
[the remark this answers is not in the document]
rhymes: done!

rhymes:
Sorry @citizen428 I force pushed to bring master in and fix a conflict

rhymes:
@citizen428 

> As suggested on Slack, I'd probably use the username part of the relay email address Apple generates. first_name.downcase sounds like it has a lot of potential for duplicates (said by a Michael, I know what I'm talking about).

Technically the person can send their real email along, the relay is a choice they can make. 

There's still going to be a probability of collision: "bob@acme.com" and "bob@google.com" will result in the same `apple_username`. 

As there can't be two identical emails in the Apple DB, I'm wondering if we should just change it to a hash of the email. We don't really care as we never display the Apple username, as, unlike Twitter and Github, there's no social clout or social features attached to it, it's just for our internal purposes.

We'd still have an issue if a person [changes their email on Apple ID](https://support.apple.com/en-us/HT202667) but we already have that same issue if for Twitter and Github so it won't be worse than the status quo.

I'm going to try to see what happens with an hash in the meantime

citizen428:
@rhymes Yeah, that makes sense. Thanks for all the changes so far, this PR is shaping up nicely!

benhalpern:
I'm not sure I _entirely_ follow the username logic/discussion...

But based on how we currently do it, we take, for example, `twitter_username` and if it already exists, we append a random number.

So could we do that? `first_name.downcase`... if `bob` exists, that's the username, otherwise `bob_488` and then they can change it later?

...

I got the SPF stuff working for Sendgrid and Apple. It was simple enough. Will look into Mailchimp tomorrow.

UX Guidelines mean we need the logo and black color eh? I'd say that means we should add the Twitter and GitHub logos and brand colors as well, eh? I'd say we go ahead and do that.

This all make sense?

benhalpern:
If the design guidelines end up being a lot of yak shaving, we could merge this functionality behind a feature flag and that part as a follow-up.

rhymes:
Rebased on top of master to make sure we included all the recent changes

rhymes:
> But based on how we currently do it, we take, for example, twitter_username and if it already exists, we append a random number.

@benhalpern yeah, I think we gost lost in the conversation about using the first name as I initially did. I'm going to revert it to that as the logic already takes care of it.

> I got the SPF stuff working for Sendgrid and Apple. It was simple enough. Will look into Mailchimp tomorrow.

cool thanks! 

> UX Guidelines mean we need the logo and black color eh? I'd say that means we should add the Twitter and GitHub logos and brand colors as well, eh? I'd say we go ahead and do that.

cc @ludwiczakpawel on this

> If the design guidelines end up being a lot of yak shaving, we could merge this functionality behind a feature flag and that part as a follow-up.

that makes sense, thanks!

ludwiczakpawel:
> > UX Guidelines mean we need the logo and black color eh? I'd say that means we should add the Twitter and GitHub logos and brand colors as well, eh? I'd say we go ahead and do that.
> 
> cc @ludwiczakpawel on this
> 
> > If the design guidelines end up being a lot of yak shaving, we could merge this functionality behind a feature flag and that part as a follow-up.

https://github.com/thepracticaldev/dev.to/pull/8679 it should also cover apple's one.

benhalpern:
How's this look for default profile image?

#8695

benhalpern:
Alright, I'm not sure what the final steps for Mailchimp are to close the loop on that. The process was straightforward for sendgrid.

This wouldn't be a full-on blocker right?

fdoxyz:
> This wouldn't be a full-on blocker right?

Shouldn't be a blocker, that's right. If Sendgrid works with the Private Email Relay we'll be able to send the verification email.

@rhymes are we still going with releasing behind a feature flag to try this out first? With a few registrations we can make sure Mailchimp is also setup correctly. But most importantly we can make sure all configs are working well before making this public since these configs are notoriously tricky to get right.

I gave this a try (using ngrok with https) configuring all four ENV Variables (`APPLE_CLIENT_ID`, `APPLE_TEAM_ID`, `APPLE_KEY_ID` & `APPLE_PEM`) but ran into an `ActionController::InvalidAuthenticityToken` error. Maybe something I'm missing?

fdoxyz:
Whoops, just double checked and it's under `ApplicationConfig`. I'll give it another try in just a minute

rhymes:
@fdoxyz 

> I gave this a try (using ngrok with https) configuring all four ENV Variables (APPLE_CLIENT_ID, APPLE_TEAM_ID, APPLE_KEY_ID & APPLE_PEM) but ran into an ActionController::InvalidAuthenticityToken error. Maybe something I'm missing?

Oh gosh, forgot about that. I had the same issue, that's why I would like to try with a real server.

I'm not entirely sure how to feature flag this though. I think we need to "patch" `Authentication::Providers.available` and `Authentication::Providers.enabled` to add `:apple` only if it's enabled. 

I'll give it a try

rhymes:
I'm not sure how we can use feature flags for this to enable/disable it for both authenticated and unauthenticated users without passing down the `current_user` everywhere we use `.available/.available?` and `.enabled/.enabled?` in `Authentication::Providers`

For example:

```ruby
    def self.available
      providers = Authentication::Providers::Provider.subclasses.map do |subclass|
        subclass.name.demodulize.downcase.to_sym
      end.sort

      # hide :apple behind a feature flag
      return providers - [:apple] unless FeatureFlag.exist?(:auth_apple)
      providers
    end
```

and 

```ruby
    def self.enabled
      providers = SiteConfig.authentication_providers.map(&:to_sym).sort
      return providers - [:apple] unless FeatureFlag.enabled?(:auth_apple, CURRENT_USER)
    end
```

any ideas?

A possible alternative would be to hide it everywhere except as a connect option for already authenticated users and enable it with the Actor flag so that a few of us can test it by going to https://dev.to/settings/account

cc @joshpuetz

benhalpern:
> A possible alternative would be to hide it everywhere except as a connect option for already authenticated users and enable it with the Actor flag so that a few of us can test it by going to https://dev.to/settings/account

This sounds good to me....So we wouldn't even add it to the `SiteConfig`, but we would display it custom in settings as a feature flag to start?

rhymes:
> This sounds good to me....So we wouldn't even add it to the SiteConfig, but we would display it custom in settings as a feature flag to start?

I think so, we modify https://github.com/thepracticaldev/dev.to/blob/master/app/views/users/_additional_authentication.html.erb to display the "connect apple account" button only for users that have the feature enabled and that's it.

I'll give it a try

joshpuetz:
> A possible alternative would be to hide it everywhere except as a connect option for already authenticated users and enable it with the Actor flag so that a few of us can test it by going to https://dev.to/settings/account
> 
> cc @joshpuetz

It really depends what you're after in terms of testing/flagging. Just giving access to a few us of with the Actor gate sounds like a good first step. We could roll it out to either a percentage of users or a percentage of the time for all user with the Percentage of Actors or Percentage of Time gates (https://github.com/jnunemaker/flipper/blob/master/docs/Gates.md#3-percentage-of-actors).

However: we'll run into issues with user *creation*: without a user object Flipper is going to have issues tracking if said user is in the 'they get it bucket': I think only the 'percentage of time' gate would work in this case.

rhymes:
Thanks!

@joshpuetz I think once we have enough confidence of it working, we can go back to all or nothing using `SiteConfig` only. I don't think we want to keep it as a flag for long

joshpuetz:
Absolutely agreed @rhymes: flags should be short lived!

rhymes:
@joshpuetz @benhalpern I added this flag https://github.com/thepracticaldev/dev.to/pull/7952/commits/105bde0373389a4eb9b6e948f60734c7e0e99cba

rhymes:
@joshpuetz @benhalpern I'm mistaken this is not the right approach. An authentication provider to be activated in the code has to be enabled and right now enabling it means enabling it for everybody as we consider this a configuration setting which is per site, not per user.

By removing it from the list of enabled providers we effectively disable it so it won't work globally, irregardless of what the feature flag says. 

Basically the issue we have here is that the whole authentication functionality has been developed as a site plugin, not a feature for users.

I'm going to revert the last commit.

I think we should still figure out a way to spin up a new Forem with this branch and test it there, then we can roll it out. Maybe I can ask @jdoss ?

If that's an avenue we don't want to pursue we need to go back to the drawing table and find a way to merge the concept of "enabled authentication method" to "but keep it disabled for everyone but me" but as I mentioned here - https://github.com/thepracticaldev/dev.to/pull/7952#issuecomment-645311046 -, there's no concept of authenticated user down at the plugin layer as it was built as a site wide feature

> Does Apple allow us to only present the 'sign in with Apple' to already authenticated users?

Good question, not sure.


Let me know!

jdoss:
@rhymes Soon. Very very soon...

leewynne:
Hows about I push the branch to TMML and test it? Would that help?

mstruve:
@rhymes, you want to update this branch with master and have @leewynne  give it a test?!

rhymes:
@mstruve sure! let me update it. we're also likely going to work on this in the next cycle expanding the bandwidth on this project

app/services/authentication/providers/apple.rb
```diff
@@ -0,0 +1,63 @@
...
+      # For Apple we override this method because the `info` payload doesn't
+      # include `nickname`. On top of not having a username, Apple allows users
+      # to 'choose' the first_name & last_name sent our way so they are
+      # definitely not assured to be unique. We still need `user_nickname` to
+      # always be the same on each login so we use the email hash as suffix to
+      # avoid collisions with other registrations with the same first_name
+      def user_nickname
+        email_digest = Digest::SHA512.hexdigest(info.email)
+        "#{info.first_name&.downcase}_#{email_digest}"[0...25]
+      end
```
fdoxyz, line 43: I left the note explaining why the need for the hash but it comes down to multiple users coming in with the same name. This is overriding `user_nickname` from `Provider.rb` so it remains clean/simple in the Base class and customizable for Providers that require some extra logic. (Anyone) feel free to ask/discuss

In this case the username is based on the user's "real name" which makes it seem a bit more natural than just a hash and looks like this:

Screen Shot 2020-08-06 at 16 58 36

app/services/authentication/providers/apple.rb
```diff
@@ -0,0 +1,63 @@
+module Authentication
+  module Providers
+    # Apple authentication provider, uses omniauth-apple as backend
+    class Apple < Provider
+      SETTINGS_URL = "https://appleid.apple.com/account/manage".freeze
+
+      def new_user_data
+        # Apple sends `first_name` and `last_name` as separate fields
+        name = "#{info.first_name} #{info.last_name}"
+        timestamp = raw_info.auth_time || raw_info.id_info.auth_time
```
fdoxyz, line 10: To be honest, this is something that has made me scratch my head quite a bit. I've had times where `auth_time` comes nested directly in `raw_info` and other times within `id_info`.

Because of this mismatch every now and then ran into an error with `Time.zone.at(nil)`, this will catch both cases. If anyone is absolutely certain we'll get one or the other let me know, but from my experience I rather leave it this way 😅

Same a below for `existing_user_data `

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -1,14 +1,14 @@
 class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   include Devise::Controllers::Rememberable
 
-  # Called upon successful redirect from Twitter
-  def twitter
-    callback_for(:twitter)
-  end
+  skip_before_action :verify_authenticity_token, only: %i[apple] # rubocop:disable Rails/LexicallyScopedActionFilter
```
fdoxyz, line 4: This may be the main caveat of the PR. I don't think disabling this for Apple auth should be a problem but there may be a more elegant solution (I believe Apple sends in the data within a signed JWT). Anyways this is the reason behind having to open this

__Problem:__ Apple uses POST for the OAuth callback and because of this Rails "protects against forgery" from a different Origin and raises a CSRF exception. If you enable CORS debug you'll see this in the logs

```
HTTP Origin header (https://appleid.apple.com) didn't match request.base_url (https://rpi-forem.ngrok.io)
```

__Solution 1:__ Keep this `skip_before_action` for Apple's callback to avoid the CSRF exception

__Solution 2:__ Edit `config/initializers/cors.rb` in order to add an exception for this endpoint, only allowing `https://appleid.apple.com` instead of bypassing all checks. This is the equivalent to "_giving Apple the keys to enter through this very specific door_" vs "_leave the door open for anyone to come through_"

rhymes: Love the CORS solution, I'm still baffled as to their choice but everything about implementing this provider has been "slightly different" than the two previous ones

app/services/authentication/providers/apple.rb
rhymes: It might be the difference between first time vs second time. I noticed the payload Apple sends the first time (so you have to disconnect yourself on Apple's dashboard) is slightly different. I probably didn't catch this one :(

app/services/authentication/providers/apple.rb
rhymes: I did implement something like this initially and then reverted it as we have an automatic mechanism who handle's collision. See https://github.com/forem/forem/pull/7952/commits/b82e5c71c380f4ccf79436035633ce4ee2f47516 which is a revert due to this comment https://github.com/forem/forem/pull/7952#issuecomment-643022179

app/services/authentication/providers/provider.rb
```diff
@@ -22,8 +22,7 @@ def existing_user_data
       end
 
       def user_nickname
-        # Apple doesn't have the `nickname` field in the `info` payload
-        (info.first_name&.downcase || info.nickname).to_s
+        info.nickname.to_s
```
rhymes, line 25: isn't `info.nickname` empty for Apple?

app/services/authentication/providers/apple.rb
fdoxyz: Absolutely! I actually saw what you did there and modified it a bit. The difference was that before their username would only be a hash (very cryptic username as in not personalized at all) and here it's their first name with the appended hash.

The automatic mechanism is made up by [these functions in User](https://github.com/fdoxyz/dev.to/blob/rhymes/add-sign-in-with-apple/app/models/user.rb#L510-L534) right? There is a problem with it though, the registration process will encounter an error ([in this save!](https://github.com/fdoxyz/dev.to/blob/rhymes/add-sign-in-with-apple/app/services/authentication/authenticator.rb#L47)) before making it to where the User is created (`before_validation :set_username` within User). The reason is that multiple Identities can be created that render the same "apple_username" and that's why I added this again. 

This is what I found happens:

1. User `johndoe@dev.to` signs up with `first_name="John" last_name="Doe"` and is assigned username `johndoe`
2. User `martin@othercompany.com` signs up but modifies their first & last name and also comes in as "johndoe".  [Authentication::Authenticator#find_or_create_user](https://github.com/fdoxyz/dev.to/blob/master/app/services/authentication/authenticator.rb#L98) checks and finds that a user with the username already exists, and instead of creating a new user it returns it and then an exception is raised because of this (before the automatic mechanism is processed)

The problem is not only that Apple OAuth process (when registering) users can change the name, but some common names will almost certainly collide. In the end `"#{first_name}_#{last_name}` is not guaranteed to be a unique way to know if a user is new or existing specifically within the Apple Provider "_namespace_".

The alternative I considered was not perfect either, and it consists of modifying `Authentication::Authenticator#find_or_create_user` so instead of this:

```
def find_or_create_user!
  existing_user = User.where(
    provider.user_username_field => provider.user_nickname,
  ).take
  return existing_user if existing_user
...
```

change it to look something like this:

```
def find_or_create_user!
  existing_user = User.where(
    provider.user_username_field => provider.user_nickname,
    'email' => provider.info.email,
  ).take
  return existing_user if existing_user
...
```

This way at least we know Apple won't send us users with the same username (regardless of how we define Apple users' username). __The problem I see with this__ is that common names will almost certainly repeat (a lot) and the existing mechanism uses a `rand(100)` call, so if we end up having 100 users with the same `first_name` & `last_name` (keep in mind people can "edit their name") it's likely for users to try to register and having to do lots of failed attempts until they get assigned a unique random username (once we have 100 it will become impossible for them to register).

It's all very complicated, if you sign up for Medium using Apple they use a worse username actually (doesn't mean we can't come up with a better solution!)

Screen Shot 2020-08-07 at 09 43 21

I think an improvement (I can push this change soon) is to make this method look this way:

```
def user_nickname
  [
    info.first_name&.downcase,
    info.last_name&.downcase,
    Digest::SHA512.hexdigest(info.email),
  ].join("_")[0...25]
end
```

With this change the username would probably be something like `fernando_valverde_3a5f38a` instead of `fernando_3a5f38ad4e722de8`

Sorry for the super lengthy response! 😅

app/services/authentication/providers/provider.rb
fdoxyz: Yes, but we can use `info.nickname` for all providers (keeping it in the base class `Provider`) and whenever we implement a new provider that is different (Apple 😆) we can override it in their own class `Authentication::Providers::Apple::user_nickname`.

Hypothetical example: If we add Linkedin sign in and their hash __does have__ `info.nickname` then no need to override. Later on we implement Facebook auth and they __don't have__ `info.nickname` but instead they have something like `info.profile_name` we can just override it in `Authentication::Providers::Facebook::user_nickname`.

This way no matter which provider we're referring to they will have `user_nickname` respond with a valid username and the base class implementation (most common scenario) remains a bit cleaner/simpler.

An alternative I thought about was making it a "virtual method" (using `raise SubclassResponsibility`) so each provider can implement their own. Maybe this sounds better?

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -1,14 +1,21 @@
...
-  def github
-    callback_for(:github)
+  # Rails actionpack only allows POST requests that come with an ORIGIN header
+  # that matches `request.base_url`, it raises CSRF exception otherwise.
+  # There is no way to allow specific ORIGIN values in order to securely bypass
+  # trusted origins (i.e. Apple OAuth) so `protect_from_forgery` is skipped
+  # ONLY when it's safe to do so (i.e. ORIGIN == 'https://appleid.apple.com').
+  # The hardcoded CSRF check can be found in the method `valid_request_origin?`:
+  # https://github.com/rails/rails/blob/master/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L453-L463
+  protect_from_forgery unless: -> { safe_apple_callback_request? }
```
fdoxyz, line 11: Okay so I gave up on `rack-cors`, I now understand how this works 😅 Here's the gist:

1. `rack-cors` sets the CORS headers and these headers are used by the browser to allow/disallow requests. Our problem was not coming from here
2. The CSRF check is done in Rail's `actionpack` [here](https://github.com/rails/rails/blob/master/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L453-L463), notice how __there is no way to "safely override" this__. So I added this check to bypass `protect_from_forgery` only when it's safe to do so.

Basically, we only want to "open our door" when Apple (trusted origin) is knocking, instead of bypassing this altogether ([see this comment](https://github.com/forem/forem/pull/7952#discussion_r466851860) for the backstory of the analogy)

Also, check this out as it's now deployed and should be working in https://rpi-forem.ngrok.io/

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -98,4 +105,11 @@ def user_persisted_and_valid?
   def user_persisted_but_username_taken?
     @user.persisted? && @user.errors_as_sentence.include?("username has already been taken")
   end
+
+  # We only bypass CSRF checks on Apple callback path & Apple trusted ORIGIN
+  def safe_apple_callback_request?
+    trusted_origin = Authentication::Providers::Apple::TRUSTED_CALLBACK_ORIGIN
+    request.fullpath == Authentication::Providers::Apple::CALLBACK_PATH &&
+      request.headers["ORIGIN"] == trusted_origin
+  end
```
rhymes, line 114: cc @justinlarson here for a quick sanity check

app/services/authentication/providers/apple.rb
rhymes: Have you gotten around to test this?

app/services/authentication/providers/provider.rb
rhymes: Makes sense, thank you!

app/controllers/omniauth_callbacks_controller.rb
fdoxyz: This is a bit crazy, it worked for me yesterday. Now that I'm trying it again today I got the same error as you (CSRF), but after I clear my caches and service worker (basically delete all Application data) it works again :(

I've seen this happen a few times now and I think something in the service worker may be interfering with the flow. Can you maybe try it again after clearing the caches?

I think Apple's redirect may be cached by the service worker, I worry we'll see the same result in a real deployment but it may be good to test it out. I'll try to see if there's something we can avoid to cache over in the service worker JS

rhymes: Booted up Safari from scratch.
Tried the first time with the app already connected to the rpi-forem and it failed. Went to appleid.apple.com, disconnected the app, didn't work. 

Tried again from Chrome, it worked!

db/migrate/20200514162708_add_apple_authentication_fields.rb
```diff
@@ -0,0 +1,6 @@
+class AddAppleAuthenticationFields < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :apple_created_at, :datetime
```
joshpuetz, line 3: Not for this PR, but in the future we might consider breaking out auth columns to a separate table as we add additional auth providers!

Envfile
```diff
@@ -42,20 +42,24 @@ variable :ELASTICSEARCH_URL, :String, default: "http://localhost:9200"
...
-# for more development experience
-
-# GitHub
+###############################################################
+############## 3rd Party Authentication Services ##############
+###############################################################
+
+# Apple for authentication
+# https://developer.apple.com/sign-in-with-apple/
+variable :APPLE_CLIENT_ID, :String, default: ""
```
mstruve, line 51: Need to move these to the new `.env_sample` file

app/models/user.rb
```diff
@@ -524,10 +526,11 @@ def temp_name_exists?
   end
 
   def temp_username
-    if twitter_username
-      twitter_username.downcase.gsub(/[^0-9a-z_]/i, "").delete(" ")
-    elsif github_username
-      github_username.downcase.gsub(/[^0-9a-z_]/i, "").delete(" ")
+    Authentication::Providers.username_fields.each do |username_field|
```
mstruve, line 529: This is a very large PR, it might be nice to pull out a bunch of these refactors into their own PR so this is not as big and we can push them out and ensure everything works as refactored before trying to add an additional auth provider.

fdoxyz: After a quick conversation with Josh we both agree that this seems like a very good idea. We still haven't been able to fix this CSRF issue despite lots of debugging and I agree having a smaller PR could help making it easier to receive feedback/help.

I'm hoping to split this one today into a new PR with the generalization refactor and this PR will remain as "Sign in with Apple" PR :)

fdoxyz:
Marked the PR as Draft again in order to clearly share when it's ready for review again :)

I'll be running more tests and maybe clean things up a bit. Hoping to open for review again when all comments/suggestions have been addressed 🙂

spec/factories/broadcasts.rb
```diff
@@ -22,13 +22,11 @@
...
-    factory :twitter_connect_broadcast do
-      title          { "Welcome Notification: twitter_connect" }
+    factory :apple_connect_broadcast do
+      title          { "Welcome Notification: apple_connect" }
       type_of        { "Welcome" }
       processed_html do
-        "You're on a roll! 🎉 Do you have a Twitter account? Consider " \
-        "<a href='/settings'>connecting it</a> so we can @mention you if we share " \
-        "your post via our Twitter account <a href='https://twitter.com/thePracticalDev'>@thePracticalDev</a>."
+        "You're on a roll! 🎉  Do you have an Apple account? Consider <a href='/settings'>connecting it</a>."
```
joshpuetz, line 29: I think this broadcast may need to be added to a) seeds.rb, and b) a data migration to add it to the database!

joshpuetz: Here's what I came up with for the Facebook auth PR: https://github.com/forem/forem/commit/65de0e7b408f55726da0dc271c80a686b628d342

Envfile
mstruve: We actually need to completely remove this file and transfer these to `.env_sample` now

app/services/authentication/providers/apple.rb
```diff
@@ -0,0 +1,78 @@
+module Authentication
+  module Providers
+    # Apple authentication provider, uses omniauth-apple as backend
+    class Apple < Provider
+      SETTINGS_URL = "https://appleid.apple.com/account/manage".freeze
+      TRUSTED_CALLBACK_ORIGIN = "https://appleid.apple.com".freeze
+      CALLBACK_PATH = "/users/auth/apple/callback".freeze
```
joshpuetz, line 7: I don't this this is used anywhere, was there a reason for adding it?

app/services/authentication/providers/apple.rb
```diff
@@ -0,0 +1,78 @@
...
+
+      # Currently `beta` state until thoroughly tested in production level env
+      def self.beta_access?
+        true
```
joshpuetz, line 11: If we wanted to make this more dynamic we could hook this value up to a feature flag (totally optional though)

joshpuetz (approved):
In favor of getting this into production for testing!

app/services/authentication/providers/apple.rb
fdoxyz: Yes, this was added to bypass `protect_from_forgery` only if the request comes from the trusted ORIGIN **and** only on Apple's callback path.

Bypass declared [here](https://github.com/forem/forem/pull/7952/files#diff-aed97266b0fe72994225a9cac3e4230bR4-R11) and the check is [here](https://github.com/forem/forem/pull/7952/files#diff-aed97266b0fe72994225a9cac3e4230bR117-R123).

This allows us to "give Apple the keys to enter this door vs leaving the door open for any origin".

app/services/authentication/providers/apple.rb
fdoxyz: Sounds like a good idea. I'm trying it out locally, it does give more flexibility to this beta restriction.

The feature flag if enabled would make the auth provider publicly available (act as if not beta) but otherwise would be hidden behind the param `state=beta_providers_enabled`. Forem instances would always be able decide to enable or disable the providers altogether no matter if they're beta or not.

Envfile
```diff
@@ -0,0 +1,181 @@
+# Most of the things are optional. If you need to use a
```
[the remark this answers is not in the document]
mstruve: yes please!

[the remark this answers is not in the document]
fdoxyz: Thanks spotting this again! Done

fdoxyz:
Haven't forgotten about this but will require some time to get back in shape given recent changes in Authentication. I'll mark as ready for review then

leewynne:
Might be worth some live testing with me @fdoxyz 🤩

rhymes:
Superseded by https://github.com/forem/forem/pull/11934

PR #7961, opened 2020-05-19
Add config for whether campaign posts require approval

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This adds a configuration for whether a "campaign" requires approval on which posts show up in sidebar.

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -69,7 +69,7 @@ class StoriesController < ApplicationController
 
   def get_latest_campaign_articles
     campaign_articles_scope = Article.tagged_with(SiteConfig.campaign_featured_tags, any: true).
-      where("published_at > ?", 4.weeks.ago).where(approved: true).
+      where("published_at > ? AND score > ?", 4.weeks.ago, 0).where(approved: SiteConfig.campaign_articles_require_approval?).
       order("hotness_score DESC")
 
     @campaign_articles_count = campaign_articles_scope.count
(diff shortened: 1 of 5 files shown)


app/controllers/stories_controller.rb
```diff
@@ -69,7 +69,7 @@ def assign_hero_html
 
   def get_latest_campaign_articles
     campaign_articles_scope = Article.tagged_with(SiteConfig.campaign_featured_tags, any: true).
-      where("published_at > ?", 4.weeks.ago).where(approved: true).
+      where("published_at > ? AND score > ?", 4.weeks.ago, 0).where(approved: SiteConfig.campaign_articles_require_approval?).
```
fdoxyz, line 72: Is it a problem that this check will filter out all posts that are explicitly approved when approval is not required?

The query will evaluate to `.where(approved: false)` which will filter out all explicitly approved posts. 

See other inline comment below

spec/requests/stories_index_spec.rb
```diff
@@ -159,12 +159,27 @@
...
+        SiteConfig.campaign_articles_require_approval = true
+        Article.last.update_column(:score, -2)
+        get "/"
+        expect(response.body).not_to include(CGI.escapeHTML("Unapproved-post"))
+      end
+
+      it "displays unapproved post if approval is not required" do
+        SiteConfig.campaign_sidebar_enabled = true
+        get "/"
+        expect(response.body).to include(CGI.escapeHTML("Unapproved-post"))
```
fdoxyz, line 180: What I mean by that other inline comment is that the spec will fail if we add the following line here:

`expect(response.body).to include(CGI.escapeHTML("Super-sheep"))`

Which I think makes sense: "If we don't need approval with the `SiteConfig` I should be able to see Posts whether they're approved or not"

PR #7524, opened 2020-04-27
"Write a Post" view redesign

ludwiczakpawel:
## What type of PR is this? (check all applicable)

- [x] Refactor

## Description

This PR introduces new styling for "Write a Post" view.

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)

image

## Added tests?

- [ ] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [x] no documentation needed

PR #7967, opened 2020-05-19
Remove /new from serviceworkers

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
`/new` foregoes the top bar via #7524. So we don't want to load the shell.

There are maybe other ways we could do this ideally but this is quick and easy enough I think.


benhalpern:
Going to merge this so that the editor redesign is maximally compatible when it goes live.

PR #7970, opened 2020-05-19
Add Percy to Travis yml

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
This PR adds Percy to Travis CI and enables the base build for our app, which you can get to by clicking through the GitHub integration link below:

Screen Shot 2020-05-19 at 4 13 48 PM

### A couple things to note:
1) For security purposes, we're only enabling Percy on against this repo (PRs made from branches). This means that if you fork off of our repo, branch, and PR back up, Percy **will not run**. This is not ideal, but it's the safest way to actually try our Percy without exposing secure information. We need to decide as a team whether we're all going to shift to making PRs from the main repo (which triggers Percy), or continue forking (which means we don't get to use Percy).
2) OSS contributors can _see_ our builds because [our Percy project is public](https://docs.percy.io/docs/public-projects). This means that anyone outside of the dev.to organization on Percy will _not_ have the ability to approve builds or snapshots but will be able to see build information and statuses.
3) We have a limited number of free snapshots; I suspect we're going to run out of them very quickly, and we'll run out of them quickly. If we want to truly try out Percy, we likely should pay for at least a month of snapshots and see how we like it.
4) You'll see that this PR [adds 66 new snapshots](https://percy.io/thepracticaldev/dev.to/builds/5321617). I'm leaving the visual diffs unapproved so that everyone can see them. Before merging this PR, I will "approve all", which will set our base build. From that point on, anyone who PRs in this repo will kick off a Percy build, and the snapshots generated from _their_ PR's code will be compared against this `master` build.

This [massive PR](https://github.com/thepracticaldev/dev.to/pull/7783).

Check out the [related Percy build](https://percy.io/thepracticaldev/dev.to/builds/5336425).

## Added tests?

- [x] yes **(Percy basically now _is_ our way of visual testing!)**
- [ ] no, because I need help

## Added to documentation?

_I plan to add documentation in a separate PR if we decide to go ahead with Percy._

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed

N/A

super cute running hedgehog

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ script:
   - bundle exec rails webpacker:compile
   - bundle exec rails db:schema:load
   - './cc-test-reporter before-build'
-  - 'bundle exec rspec spec --color --tty --format progress'
+  - 'npx percy exec -- bundle exec rspec spec --color --tty --format progress'
   - '[ ! -f .approvals ] || bundle exec approvals verify --ask false'
   - './cc-test-reporter format-coverage -t simplecov -o coverage/codeclimate.simplecov.json'
   - 'yarn test --colors'


vaidehijoshi:
> I think this is the best option. We could also make PRs from the main repo only when we need to use Percy. For me that's probably most of the time since I work on more user-facing stuff 🙃 That said I don't think it's a terrible shift.

I think to a certain extent, this works. But, notably...this means that one of us could make a PR that we don't _think_ or _realize_ will make a visual change or break things...and definitely could! And if we don't make a directly PR into the repo (as opposed to forking) then we'll miss this and potentially deploy code that causes a regression. 😞 IMO, I think that we should probably all make PRs into the main repo rather than forking for some given amount of time while we "try out" Percy. And then, should we all find it useful, we should prioritize figuring out how to make this work for forks/OSS contributors.

mstruve:
@vaidehijoshi does this mean we have to always manually go into Percy and approve a build before it can be merged? I was under the assumption that Percy was comparing images itself and then failing or passing for us. Adding another manual component in makes me a little leary that it might burden our process a bit.

vaidehijoshi:
@mstruve By default, Percy approvals aren’t _required_ before merging, but we can enable this feature once we're ready for it. You can read a bunch more about this in the description of [my initial Percy PR](https://github.com/thepracticaldev/dev.to/pull/7783#issue-416390948).

If you read that description, I explain that the only time that you need to go into Percy and do anything is if a visual change is detected; for example, if you make a bunch of spec changes, Percy shouldn't be triggered and you shouldn't have to do anything! Ultimately, I _do_ think we should require Percy diffs to be manually approved by someone on the team. We already take on the burden of QAing and checking that things work locally. Percy should be much easier than that, and will actually make our process for quickly catching visual regressions even better.

So, IMO, it does the opposite of burdening our code review process: it actually speeds it up and is a pixel-catching net for us to detect regressions.

PR #7977, opened 2020-05-20
Add missing unique indexes - part 1

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Our PostgreSQL is missing quite a few unique indexes that are only checked via Rails, with the side-effect of potentially generating slow SQL queries and leaving it open to duplicate rows if the data comes in via other sources.

I tried to group some tables that have relatively little rows. There are still quite a few, I'll send them in "parts"

> This helper validates that the attribute's value is unique right before the object gets saved. It does not create a uniqueness constraint in the database, so it may happen that two different database connections create two records with the same value for a column that you intend to be unique. To avoid that, you must create a unique index on that column in your database.

https://guides.rubyonrails.org/active_record_validations.html#uniqueness

https://github.com/rubocop-hq/rubocop-rails/pull/197

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/db/migrate/20200520091835_add_unique_index_to_webhook_endpoints_target_url.rb
+++ b/db/migrate/20200520091835_add_unique_index_to_webhook_endpoints_target_url.rb
@@ -1,0 +1,7 @@
+class AddUniqueIndexToWebhookEndpointsTargetUrl < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def change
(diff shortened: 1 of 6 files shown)


rhymes:
@mstruve yeah, I couldn't find duplicated adjustments, but please double check as it's best to be extra sure

PR #7979, opened 2020-05-20
Refactor OrganizationsController

JuannFerrari:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR adds a refactoring to `OrganizationsController` by applying Form Object Pattern

I did this refactor because I thought that it would be good to move the validations and some logic from the controller to a [FormObject](https://medium.com/selleo/essential-rubyonrails-patterns-form-objects-b199aada6ec9) using the [yaaf](https://github.com/rootstrap/yaaf) gem. I believe this pattern will reduce the complexity of the controller and can be used in others as well.

I also added `organizations#create` request tests.

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


Gemfile
```diff
@@ -105,6 +105,7 @@ gem "uglifier", "~> 4.2" # Uglifier minifies JavaScript files
 gem "ulid", "~> 1.2" # Universally Unique Lexicographically Sortable Identifier implementation for Ruby
 gem "validate_url", "~> 1.0" # Library for validating urls in Rails
 gem "webpacker", "~> 5.1.1" # Use webpack to manage app-like JavaScript modules in Rails
+gem "yaaf", "~> 0.1.1" # To create form objects in a standardised, railsy way
```
citizen428, line 108: I'm of the opinion that we already have too many gems, some of which are only used in specific contexts (e.g. `dry-struct`). I'm very reluctant to add another gem for a single controller, especially since it's reasonably straightforward to do this using a PORO with `ActiveModel` included.

citizen428:
Thanks for your PR @JuannFerrari,  it's much appreciated. I left a general comment that's something we should find a consensus on as team, so it's a bit of a meta discussion triggered by this PR.

Gemfile
rhymes: If it can be accomplished without the gem I'm +1 on that approach

mstruve: I 100% agree with @citizen428 I think we should avoid using gems for refactors and stick to using them when we really need their features.

juanmanuelramallo: I might be biased on this comment, but the very main purpose of yaaf is to standardise the way we write form objects. @citizen428 mentioned that we could accomplish this with a PORO object with ActiveModel and that's right, in fact that's what most of us do in our projects. But when several form objects are around and each one of them is written differently (with/without db transactions, with/without a save with bang method, ...) it's hard to keep consistency in the codebase. 

What I see from this PR is an introduction to the form object pattern (which indeed cleaned up a bit our organizations controller) and can be later adopted by other sections of the app.

I agree that adding a new gem (not so popular, and rather new) to a very mature app as dev.to can be daunting, but you might agree with [Soveran's point on popularity](https://soveran.com/popularity.html), it's easy to measure popularity, but not so much for quality. So that's why maybe it'd be nice for you to take a look at yaaf's source code https://github.com/rootstrap/yaaf/blob/master/lib/yaaf/form.rb which is only 64 lines long leveraging directly from ActiveModel. 

If you reached this point, thanks for reading, and thanks for the amazing community you have built in dev.to!

mstruve: > What I see from this PR is an introduction to the form object pattern

While I think it's GREAT to standardize the way we do things, right now we are in the process of getting ready for a Rails upgrade #7658  and have a lot of other balls in the air that adding a new dependency just for the purpose of standardization is not something additional we want to take on. 

Our rails controllers might not be the prettiest things right now but they don't require knowledge of additional gems to understand and work with and I think that makes them more approachable to the rest of our community.

juanmanuelramallo: Fair enough! Looking forward to the upgrade! 🎉 

We can revisit this approach later if needed. Thank you for the explanation, you really are an SRE aren't you? 😆

JuannFerrari: > > What I see from this PR is an introduction to the form object pattern
> 
> While I think it's GREAT to standardize the way we do things, right now we are in the process of getting ready for a Rails upgrade #7658 and have a lot of other balls in the air that adding a new dependency just for the purpose of standardization is not something additional we want to take on.
> 
> Our rails controllers might not be the prettiest things right now but they don't require knowledge of additional gems to understand and work with and I think that makes them more approachable to the rest of our community.

Thanks for the explanation! What do you think about removing the gem and apply the refactoring with a PORO like @citizen428 suggest? (I know that it may be weird having a single controller following a FormObject pattern but it could be used as an example)

citizen428: @juanmanuelramallo @JuannFerrari I looked at the source, it's nice 👍 But if anything, the fact that it's only 64 lines makes me even more wary to add it. I've seen many small gems like this abandoned in the past. Alas they had their own dependencies locked which made upgrading the overall app more painful. This is why I generally just add an `ApplicationForm` class, which is IMHO the best of both worlds, as it standardizes the way forms are handled but without introducing another dependency. If you want to reuse (parts of) YAAF for such a base class that's fine too. 

All that said, happy with whatever the majority of the team feels is the right decision here.

mstruve:
Since it seems the team here agrees this is not something we want to move forward with at the moment I am going to close this PR. Feel free to keep the discussion going though!!! Thanks again @JuannFerrari for the PR!

citizen428:
@juanmanuelramallo @JuannFerrari Thanks for the PR, the contributions and also for making YAAF in the first place 😃 Would you be interested in taking up a PR which introduces an `ApplicationForm` class and changes 1 or 2 controllers to use it?

JuannFerrari:
@citizen428 I would love to! that'll be my next contribution then.

citizen428:
@JuannFerrari I made an issue for this, please assign yourself, thanks! 

https://github.com/thepracticaldev/dev.to/issues/8207

PR #7985, opened 2020-05-20
Apply HTML encoder to FeedContent's query

maestromac:
- [x] Feature

## Description
This encoder option will escape the search result prior to applying the highlight tags (`<em>`).

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#request-body-search-highlighting
n/a
## Added tests?
- [x] ~no, because they aren't needed~ actually I should've added some

## Added to documentation?
- [x] no documentation needed

n/a


rhymes:
@maestromac is there a way to test for this? I have the feeling this can easily disappear in a refactoring or regression and we won't notice until the next time something breaks.

spec/system/search/display_comments_search_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "rails_helper"
+
+RSpec.describe "Display articles search spec", type: :system, js: true, elasticsearch: "FeedContent" do
+  before do
+    stub_request(:post, "http://www.google-analytics.com/collect")
```
rhymes, line 5: I started to see this around the testing failures sometimes. I thought we didn't have GA anymore :-|

mstruve: This always pops up for me locally, I am not sure why, maybe something is enabled still/"cached" locally despite us getting rid of it in the main app... its one of those easy fixes so no one has really investigated much I think

rhymes: It could be that the Google JS script has changed and now posts to that URL. I can't wait to get rid of GA :D

PR #7990, opened 2020-05-20
Cache Article IDs Rather than ActiveRecord Objects

mstruve:
- [x] Bug Fix
- [x] Optimization

## Description
In general, we should not be catching AR objects in Redis. Redis is meant to store simple data structures, not AR relationships. The reason we dont want to do this is that then that cache key working relies on specific code and if we change that code suddenly the cache key can no longer be loaded. This changes the cache key to store ids rather than an AR relationship. I also updated the actual cache key to ensure we dont load any data that will break with the new code.

The downside to this approach is that when the cache is not present we will make 2 db requests instead of 1. I think this is a small price to pay for having a reliable simple cache. I also updated the code to use `limit` instead of `sample` which will limit the records pulled from the database rather than pulling them all and then sampling them. 

#7988 
#7986 
#7987 
Yes it took me 3 tries to fix this cache issue. 

## Added tests?

alt_text

### DIFF
--- a/app/services/suggester/articles/classic.rb
+++ b/app/services/suggester/articles/classic.rb
@@ -7,14 +7,16 @@ module Suggester
 
       def initialize(input = nil, options = {})
         @input = input
-        @not_ids = options[:not_ids]
+        @not_ids = options[:not_ids] || []
       end
 
       def get(num = 1)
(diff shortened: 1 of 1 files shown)


rhymes:
OMG

PR #7995, opened 2020-05-20
Use Faker to generate user + article factory data

vaidehijoshi:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
We need to deterministically generate data for tests tagged with `percy: true` so that our snapshots are reliably the same, each time we try to build Percy. Currently, however, we have a couple factories that do _not_ use `Faker` to do this, which means we can't deterministically generate data.

This PR ensures that we use `Faker` instead of `generate` to make sure that we don't get intermittent Percy diffs where the colors, ~usernames, or emails~ and article titles change on each build. 😢 

## Added tests?

- [x] yes **(check snapshots)**
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/spec/factories/articles.rb
+++ b/spec/factories/articles.rb
@@ -3,7 +3,7 @@ FactoryBot.define do
 
   factory :article do
     transient do
-      title { generate :title }
+      title { Faker::Food.dish }
       published { true }
       date { "01/01/2015" }
       tags { Faker::Hipster.words(number: 4).join(", ") }
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -21,6 +21,8 @@ FactoryBot.define do
     checked_terms_and_conditions { true }
     signup_cta_variant           { "navbar_basic" }
     email_digest_periodic        { false }
+    bg_color_hex                 { Faker::Color.hex_color }
+    text_color_hex               { Faker::Color.hex_color }
 
     trait :with_identity do
       transient { identities { Authentication::Providers.available } }


vaidehijoshi:
Hm, need to investigate why the username validations are failing here. Going to pick back up on this tomorrow.

spec/factories/users.rb
```diff
@@ -8,11 +8,11 @@
 
   factory :user do
     name                         { Faker::Name.name }
-    email                        { generate :email }
-    username                     { generate :username }
+    email                        { Faker::Internet.email }
+    username                     { Faker::Internet.username }
```
rhymes, line 12: this won't work, as username has a few rules:

1. can't be one of `ReservedWords.all` (which also includes all the routes)
1. can only be `/\A[a-zA-Z0-9_]+\z/` but `Faker::Internet.username` generates sepators like dash or dot, maybe with `Faker::Internet.username(separators: [])` you can go on a bit, see https://github.com/faker-ruby/faker/blob/master/doc/default/internet.md

same thing for `.email`

There's still the fact that these names will be random, so Percy would fail anyway? Maybe I didn't understand fully the scope of the problem.

One last thing: as we generate thousands of user objects, there' still the chance that a collision can occurr, not likely but also not impossible. I guess that's the reason they decided to use [factory bot sequences](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#sequences).

See the implementation of `Faker::Internet.username`:

https://github.com/faker-ruby/faker/blob/a7aa80877f7bcad435b9a19a62cf5a82b5f449b0/lib/faker/default/internet.rb#L35-L74

vaidehijoshi: @rhymes yeah, i clearly got too excited late last night when I pushed this up and was hopeful that I could quickly fix this (LOL 😭)

FWIW, we _should_ be able to assert that the same value will be deterministically generated by Faker (at least, in the same _order_) because of [this](https://github.com/thepracticaldev/dev.to/blob/master/spec/rails_helper.rb#L158-L168) bit in our `rails_helper.rb`.

vaidehijoshi:
I'm still running into issues while trying to get the user `emails` and `usernames` to be rendered deterministically (I also think I tried to fix too much in this one PR 😓).

Instead, I've updated this PR to only use Faker to generate the colors on a user (which were previously just random!) and to generate the `title` on an article (which was also being done randomly).

FYI: for the curious, Faker [allows you to deterministically](https://github.com/faker-ruby/faker#deterministic-random) render data (tl;dr Faker's data is always random, except for when you [do this](https://github.com/thepracticaldev/dev.to/blob/master/spec/rails_helper.rb#L158-L168), in which case it generates data in the same order each time.

spec/factories/articles.rb
```diff
@@ -3,7 +3,7 @@
 
   factory :article do
     transient do
-      title { generate :title }
+      title { Faker::Food.dish }
```
maestromac, line 6: I want to point out that this attribute is not used unless `with_title` is used with it`

PR #7996, opened 2020-05-21
Add Feature Flags support to DEV

joshpuetz:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds feature flagging to Dev.to via the [Flipper gem](https://github.com/jnunemaker/flipper) as well as a management UI. Management UI is mounted at `/internal/flipper`, and only available to users with the `tech_admin` roles. Flipper is wrapped with a `FeatureFlag` helper class: due to the way Flipper's DSL is exposed we'll have to add wrapped methods here one by one. I've only exposed the `enabled?` flag check right now: I envision most creation and manipulation of flags will be done via the UI console.

### Why?
The use case for adding feature flags is guiding access to pages/features before they publicly launch. @fdoxyz and I plan to do this for CodeLandConf pages and features.

### Feature flags are so easy, why don't we just write our own?

It's a just a flag, right? Then you need groups. Then you need users. Then you need percentages. And timers. And analytics.

We could spend our time writing a feature flag gem, or we could work on DEV. Trust me: I worked for a [company making millions of dollars](https://rollout.io) rescuing startups from their home grown feature flags. 

The tricky bit with feature flags is not implementing them, but **using** them. Flags are meant to be short lived: once a feature gets rolled out to everyone, there should be a PR to clean up and remove said flag.

### Why Flipper?
It's relatively popular, well supported, open source, and free.

### How can I test it?
To test locally: grant yourself the `tech_admin` role, navigate to `/internal/flipper`.

### Some open questions:
- What level should this be tested at?
- Does documentation of this feature belong in docs.dev.to, or somewhere else?

@Ridhwana and @rhymes has previously made some excellent notes about feature flipping at https://www.notion.so/devto/Feature-flags-9f755970fb384385a5d1775c4cd1b351

Enabling feature - logged in user
https://share.getcloudapp.com/d5uvd6WA

Enabling feature - logged out user
https://share.getcloudapp.com/12uyZPN0

Enabling feature by user id
https://share.getcloudapp.com/RBuvpEqY

## Added tests?

- [ ] yes
- [x] no, open for discussion on what level to test at!

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed
- [X] open for discussion...needed? Appropriate for docs.dev.to?

Flipper!


citizen428:
IRRC we used some other gem for this purpose before and then removed it again. Feature flag libraries are a bit like tagging libraries in my experience: great to get a quick start, but in the long run, you always end up hacking around the limitations for your specific use case.

That said, if everyone is ok with the Flipper gem, there's nothing wrong with this PR.

app/controllers/codeland_controller.rb
```diff
@@ -0,0 +1,9 @@
+class CodelandController < ApplicationController
```
rhymes, line 1: This is probably unrelated to this PR as I know this is a POC to test them.

If FF are meant to be shortlived and Codeland is obviously a feature that does not register with other communities, should this be a site config instead of a FF?

Just a thought

app/controllers/codeland_controller.rb
```diff
@@ -0,0 +1,9 @@
...
+  # No authorization required for viewing Codeland
+
+  def show
+    return if Flipper[:codeland].enabled?(current_user)
```
rhymes, line 5: Mental note to add a little wrapper on these so that we don't sprinkle "Flipper" everywhere when we start using it

config/initializers/inflections.rb
```diff
@@ -14,3 +14,7 @@
 # ActiveSupport::Inflector.inflections(:en) do |inflect|
 #   inflect.acronym 'RESTful'
 # end
+
+ActiveSupport::Inflector.inflections do |inflect|
+  inflect.irregular "codeland", "codeland"
```
rhymes, line 19: ❤️

Gemfile
```diff
@@ -44,6 +44,9 @@ gem "fastly", "~> 2.5" # Client library for the Fastly acceleration system
 gem "feedjira", "~> 3.1" # A feed fetching and parsing library
 gem "field_test", "~> 0.3" # A/B testing
 gem "figaro", "~> 1.2" # Simple, Heroku-friendly Rails app configuration using ENV and a single YAML file
+gem "flipper", "~> 0.17.2" # Feature flipping / flags for Ruby
+gem "flipper-active_record", "~> 0.17.2" # Store Flipper flags in ActiveRecord
```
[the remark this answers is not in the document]
mstruve: I think FeatureFlags should be in the DB and then we can optimize them with Redis. The reason I think this is that I always try to design a system so that if Redis goes down we dont lose anything crucial. Redis in my opinion should be there to improve things but when possible not be the source of truth for anything. At least that is my opinion when Redis is used in our context. Obviously many people do use it as their main datastore and that works too!

app/controllers/codeland_controller.rb
mstruve: I definitely agree with Rhymes here I think we should try and avoid putting a Codeland Controller into our codebase since it does not apply to any other communities and will just be one more thing we have to cleanup

db/schema.rb
```diff
@@ -312,6 +312,7 @@
     t.string "slug"
     t.string "status", default: "active"
     t.datetime "updated_at", null: false
+    t.index ["slug"], name: "index_chat_channels_on_slug", unique: true
```
mstruve, line 315: accidental addition?

Gemfile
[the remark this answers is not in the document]
rhymes: Thanks, makes perfect sense to me! Going back to redundancy conversations :)

db/schema.rb
joshpuetz: I'm not sure: this index keep re-appearing ever time I run migrations. Hypershield doesn't appear to let me see the underlying structure of a table: can you tell me if this index is in production @mstruve?

app/controllers/codeland_controller.rb
joshpuetz: Is your objection to the name `Flipper` in code? We definitely wrap that in a more neutral name.

vaidehijoshi: maybe we could wrap it with `FeatureFlags`?

app/controllers/codeland_controller.rb
joshpuetz: I need some guidance on an alternative from both of you then: Codeland is going to require a few pages of interactivity. Should I try to get by with something like records in `pages.rb` like the SheCoded campaign used? Or might this be a good chance to add an 'online event' feature to DEV/Forem?

Pinging @benhalpern for some input here as well.

rhymes: cc @Ridhwana @lightalloy as the resident experts in generalization on team OSS :D

Ridhwana: @joshpuetz I agree with the above discussion here. We can chat in more detail if you'd like and I can give you an overview of what we did for SheCoded, but it sounds like since Codeland is not applicable to other communities, perhaps it should be modelled after Shecoded (maybe with a few tweaks in the process).

Ridhwana:
Nice work on this @joshpuetz! I'm super excited about adding a feature flag gem 🎉. 

Let me know if you want to chat more about Codeland vs Shecoded 😊.

citizen428:
> We rolled our own Feature Flags at Kenna, stored them in the db, and optimized them with Redis and it worked fantastically.

This would be my preferred approach, but since everyone else seems on board with using a gem for this purpose I'll go with the majority. Either way, thanks for getting the ball rolling on this @joshpuetz  😃

config/routes.rb
```diff
@@ -42,6 +42,7 @@
 
     authenticate :user, ->(user) { user.has_role?(:tech_admin) } do
       mount Blazer::Engine, at: "blazer"
+      mount Flipper::UI.app(Flipper), at: "flipper"
```
citizen428, line 45: Should we change the route to something like `/feature_flags`?

db/schema.rb
citizen428: @joshpuetz If you consistently have this problem, drop the index from your local DB and it will stop reappearing in the diffs.

joshpuetz: @citizen428 This index appears even after I drop my entire database!

.nvmrc
```diff
@@ -1 +1 @@
-lts/erbium
+12.16
```
mstruve, line 1: surprised this is in the dif since its also in master.... maybe a github issue bc its not listing a conflict

joshpuetz: I cherry-picked @maestromac's commit so I could get my branch to build on Travis!

PR #8000, opened 2020-05-21
Fix db:reset by adding search:setup after search:destroy

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Ran `rails db:reset` locally, ran into "ES index not found", figured the problem was that we drop the indexes but not re-create them after the drop. 

This works.

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/lib/tasks/search.rake
+++ b/lib/tasks/search.rake
@@ -18,8 +18,8 @@ namespace :search do
 
     Search::Cluster.delete_indexes
   end
+end
 
-  if %(development).include?(Rails.env)
-    Rake::Task["db:drop"].enhance(["search:destroy"])
-  end
+if %(development).include?(Rails.env)
+  Rake::Task["db:drop"].enhance(["search:destroy", "search:setup"])
 end
--- a/lib/tasks/search.rake
+++ b/lib/tasks/search.rake
@@ -20,6 +20,6 @@ namespace :search do
   end
 end
 
-if %(development).include?(Rails.env)
+if Rails.env.development?
   Rake::Task["db:drop"].enhance(["search:destroy", "search:setup"])
 end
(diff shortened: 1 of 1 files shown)


lib/tasks/search.rake
```diff
@@ -18,8 +18,8 @@ namespace :search do
 
     Search::Cluster.delete_indexes
   end
+end
 
-  if %(development).include?(Rails.env)
-    Rake::Task["db:drop"].enhance(["search:destroy"])
-  end
+if %(development).include?(Rails.env)
```
citizen428, line 23: Should we fix this condition too? `%(development)` is a string literal and I don't think we wanted `String#include?` here. Probably best to add the `w`:

```ruby
%(development)
#=> "development"
%w(development)
#=> ["development"]
```

rhymes: Totally missed that in the refactoring. Fixed!

rhymes:
@mstruve `db:reset` calls `db:drop` though:

```ruby
    # desc 'Resets your database using your migrations for the current environment'
    task reset: ["db:drop", "db:create", "db:migrate"]
```

https://github.com/rails/rails/blob/abcbfe87e5c4be16f32f49983382cdf799efc533/activerecord/lib/active_record/railties/databases.rake#L156-L157

mstruve:
Right, ok what about hooking setup to create? I just dont want setup hooked to drop since that seems counter-intuitive.

rhymes:
@mstruve makes sense! works well as well. Sent the change :)

rhymes:
@mstruve can't, that flaky test is back 😢 

https://travis-ci.com/github/thepracticaldev/dev.to/builds/167622109#L1476

Issue #7998, opened 2020-05-21
After enabling Ten X hacker theme username/email is hidden 


krishnadevz:
**Describe the bug**
hello
I set my theme to Ten X hacker theme my username is hidden because in default theme my username is everything is white except contents ex:-name and articles but after switching to dark theme my is get mixed with that and then that thing is hidden if you guys let me in I can change CSS color properties for showing text it is just a color property thanks 😊

**To Reproduce**

Go the Profile icon then UX but first your theme is set to the default mode and then switch to Ten X hacker theme and then go and see the difference that's all about how to reproduce.

**Expected behavior**

**Screenshots**
![Annotation 2020-05-21 144847](https://user-images.githubusercontent.com/42638797/82544099-366e5980-9b72-11ea-8cb1-e4f52e3b03c0.png)
![Annotation 2020-05-21 144920](https://user-images.githubusercontent.com/42638797/82544141-4b4aed00-9b72-11ea-84da-f75e917bed68.png)
![Annotation 2020-05-21 144948](https://user-images.githubusercontent.com/42638797/82544171-5867dc00-9b72-11ea-81d3-11e11233755b.png)

**Desktop (please complete the following information):**

- OS:Windows 10 X64
- Browser:firefox 
- Version:76.0.1 (64-bit)

citizen428:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

citizen428:
@krishnadevz Thanks for the report? Were you planning on fixing this? If yes, please assign yourself the issue and open a PR. If not, this is up for grabs for anyone interested.

simskij:
I can have a stab at this. 👍

joshpuetz:
@simskij This issue was fixed in https://github.com/thepracticaldev/dev.to/pull/8001, closing now!

simskij:
Was it? It looked like they didn’t touch the part this issue is refering to but peripheral issues.

joshpuetz:
[...]

PR #8001, opened 2020-05-21
--base-50: changed to #fff unreadble text in ten X hacker theme

krishnadevz:
--base-50: changed to #fff unreadable text in ten X hacker theme  light-dark text making eye strain their that's why I changed to white it looks good I tested in firefox and another bug in this theme is in case of mail 
  and usernames CSS properties are overriding  and that text is getting black out there Ex:-Hidden

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
--base-50: changed to #fff unreadable text in ten X hacker theme  light-dark text making eye strain their that's why i changed to white it looks i tested in firefox and another bug in this theme is in case of mail 
  and usernames CSS properties are overriding  and that text is getting black out there Ex:-Hidden

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [X] no documentation needed

Annotation 2020-05-21 144920
Annotation 2020-05-21 144948
# After changing base 50 value to white text appeared clear🔽
devto

 but name and email property is overridden by something I didn't  get I tried but I am not able to find that property there I think that is overridden by some other properties 
Issue assigned from #7998

### DIFF
--- a/app/assets/stylesheets/themes/hacker.scss
+++ b/app/assets/stylesheets/themes/hacker.scss
@@ -27,13 +27,14 @@
   --base-80: #c8c8c8;
   --base-70: #adadad;
   --base-60: #939393;
-  --base-50: #7a7a7a;
+  --base-50: #fff;// unreadable in Ten X hacker theme to light to view and in case of mail 
+  ///and usernames CSS properties are overriding please and that text is getting black out there Ex:-Hidden 
   --base-40: #626262;
   --base-30: #4b4b4b;
   --base-20: #353535;
   --base-10: #1c1c1c;
   --base-0: #0a0a0a;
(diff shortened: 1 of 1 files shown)


joshpuetz:
Here's what this PR looks like for me locally in Chrome and Safari:

Josh_Puetz_-_DEV_local__Community_👩💻👨💻

The email and visited links look ok-ish to me: thoughts @ludwiczakpawel ?

krishnadevz:
Thanks @joshpuetz and @benhalpern 😊 looking forward for more contributions.

PR #8002, opened 2020-05-21
Add missing unique indexes - part 2

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Follow up to #7977

Our PostgreSQL is missing quite a few unique indexes that are only checked via Rails, with the side-effect of potentially generating slow SQL queries and leaving it open to duplicate rows if the data comes in via other sources.

I tried to group some tables that have relatively little rows. There are still quite a few, I'll send them in "parts"

> This helper validates that the attribute's value is unique right before the object gets saved. It does not create a uniqueness constraint in the database, so it may happen that two different database connections create two records with the same value for a column that you intend to be unique. To avoid that, you must create a unique index on that column in your database.

https://guides.rubyonrails.org/active_record_validations.html#uniqueness

- #7977
- https://github.com/rubocop-hq/rubocop-rails/pull/197

## Added tests?

- [ ] yes
- [ ] no, because I need help

**pre-deployment**, we need to remove duplicated results in the followin tables:

**mentions**

```sql
select user_id, mentionable_id, mentionable_type, count(*)
from mentions
group by user_id, mentionable_id, mentionable_type
HAVING count(*) > 1
```

**notifications**

```sql
select user_id, organization_id, notifiable_id, notifiable_type, action, count(*)
from notifications
group by user_id, organization_id, notifiable_id, notifiable_type, action
having count(*) > 1
```

**notification subscriptions**

```sql
select user_id, notifiable_type, notifiable_id, count(*)
from notification_subscriptions
group by user_id, notifiable_type, notifiable_id
having count(*) > 1
```

**reactions**

```sql
select user_id, reactable_id, reactable_type, category, count(*)
from reactions
group by user_id, reactable_id, reactable_type, category
having count(*) > 1
```


mstruve:
- [x] mentions
- [ ] notifications
- [x] notification subscriptions
- [x] reactions

Notifications is going to take a bit to figure out the best approach and it will likely be slow. Want to remove it and we can send this out first and do notifications separate?

rhymes:
@mstruve sure! I'll remove that in the meantime!

PR #8004, opened 2020-05-21
Update webpacker configuration

rhymes:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As we're slowly approaching upgrading to Rails 6, I'm starting to set foundation stones in place.
This was (hopefully) an easy win.

Ran `rails webpacker:install` and merged the new configuration with the existing one.

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/config/webpack/test.js
+++ b/config/webpack/test.js
@@ -1,3 +1,7 @@
+/* global process, module, require */
+
+process.env.NODE_ENV = process.env.NODE_ENV || 'development';
+
 const environment = require('./environment');
 
 const config = environment.toWebpackConfig();
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,8 @@ latest.dump
 /public/packs
 /public/packs-test
 /node_modules
+yarn-debug.log*
+.yarn-integrity
 
 # Ignore storybook static site generation
 storybook-static/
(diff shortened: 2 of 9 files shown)


config/webpack/test.js
```diff
@@ -1,3 +1,7 @@
+/* global process, module, require */
+
+process.env.NODE_ENV = process.env.NODE_ENV || 'development';
```
nickytonline, line 3: Shouldn't this be `'test'`?

```suggestion
process.env.NODE_ENV = process.env.NODE_ENV || 'test';
```

rhymes: @nickytonline maybe? https://github.com/rails/webpacker/blob/master/lib/install/config/webpack/test.js

🤷

nickytonline: I asked a question, https://github.com/rails/webpacker/issues/2602, in their repo about this.

rhymes: I have the feeling the reason why it's working is that when you run webpacker it sets `NODE_ENV` correctly, so it never gets to the `OR` second clause

PR #8018, opened 2020-05-21
Revert "Update safe params for Array params (#7914)"

atsmith813:
This reverts commit 4eba14f15008564d53470ca7dd7ca2b2c8ad0b11.

- [x ] Bug Fix

## Description
Ahoy encodes URL params and our new Fastly way of decoding messes up the URL - therefore breaking Ahoy links.

#7914
https://github.com/thepracticaldev/dev.to/issues/7982

## Added tests?
Already tested live in Fastly

## Added to documentation?
- [x] no documentation needed

### DIFF
--- a/config/fastly/snippets/safe_params_list.vcl
+++ b/config/fastly/snippets/safe_params_list.vcl
@@ -1,9 +1,7 @@
 import querystring;
 sub vcl_recv {
-  # return this URL with only the parameters that match this regular expression
-  if (req.url !~ "/internal/" && req.url !~ "/search/") {
-    set req.http.decodeurl = urldecode(req.url);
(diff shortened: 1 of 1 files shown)


atsmith813:
I think there's a way to handle it better in Fastly by escaping instead of decoding...I'm sure this won't be the last time a gem encodes requests in an unexpected way :trollface:

rhymes:
@atsmith813 @Zhao-Andy what if you open an issue to https://github.com/ankane/ahoy_email ? We are friendly with the developer and maybe he can suggest a solution

atsmith813:
@rhymes that's a good idea. I believe the encoding is technically valid, it just starts to get weird when we do our Fastly trickery of decoding on it. I suspect it's because Ahoy ends up including the encoded URL as a query string param value. Off the top of my head, I'm actually not sure how Fastly handles decoding in that case.

I think by going away from decoding on Fastly to proper escaping, it'll all work just fine. I'll circle back on it on the next cool down to test that theory and if it fails, I'll open an issue 💯

PR #8020, opened 2020-05-21
Remove ProMembership Table and Factory

mstruve:
- [x] Refactor

## Description
Final piece of #7894 removing the table in the database (which is tiny) and a leftover factory file. 

## Added tests?

alt_text


db/migrate/20200521153435_remove_pro_memberships.rb
```diff
@@ -0,0 +1,21 @@
+class RemoveProMemberships < ActiveRecord::Migration[5.2]
+  def up
```
rhymes, line 2: FYI: `drop_table` in `change` accepts a block that is used for the creation step: https://github.com/thepracticaldev/dev.to/blob/master/db/migrate/20191227113154_drop_push_notification_subscriptions.rb

mstruve: oooh that is great to know! Thank you!

PR #8024, opened 2020-05-21
Fix tag removal case sensitivity for mods

atsmith813:
- [x] Bug Fix

## Description
Our Tags, in general, are messed up. This is a quick band-aid to allow mods to remove tags regardless of case.

_Note: when adding a tag a mod must specify the `name` of the tag and **it is case sensitive**. We can't properly address that issue until we fix the core underlying issue._

https://github.com/thepracticaldev/dev.to/issues/5325

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

tag_youre_it_gif

### DIFF
--- a/app/services/tag_adjustment_creation_service.rb
+++ b/app/services/tag_adjustment_creation_service.rb
@@ -20,7 +20,7 @@ class TagAdjustmentCreationService
   private
 
   def update_article
-    article.update!(tag_list: article.tag_list.remove(@tag_adjustment.tag_name)) if @tag_adjustment.adjustment_type == "removal"
+    article.update!(tag_list: article.tag_list.remove(article.tag_list.select { |tag| tag.casecmp(tag_name).zero? })) if @tag_adjustment.adjustment_type == "removal"
     article.update!(tag_list: article.tag_list.add(@tag_adjustment.tag_name)) if @tag_adjustment.adjustment_type == "addition"
   end
 
--- a/app/services/tag_adjustment_creation_service.rb
+++ b/app/services/tag_adjustment_creation_service.rb
@@ -20,7 +20,11 @@ class TagAdjustmentCreationService
   private
 
   def update_article
-    article.update!(tag_list: article.tag_list.remove(article.tag_list.select { |tag| tag.casecmp(tag_name).zero? })) if @tag_adjustment.adjustment_type == "removal"
+    if @tag_adjustment.adjustment_type == "removal"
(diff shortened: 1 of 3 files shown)


app/services/tag_adjustment_creation_service.rb
```diff
@@ -20,14 +20,30 @@ def update_tags_and_notify
...
   def creation_args
     args = @tag_adjustment_params
+
+    tag =
+      case args[:adjustment_type]
+      when "removal"
+        article.tags.detect { |article_tag| article_tag.name.casecmp(args[:tag_name]).zero? }
+      when "addition"
+        Tag.find_by(name: args[:tag_name])
+      end
```
atsmith813, line 42: This is done to do an efficient case insensitive search - we just grab the article's tag records (max 4) and loop through them, ignoring case.

This was done instead of adding an index to the `tags` table to do a `lower(name)` query. This is all going to change once we _actually_ fix tags.

app/services/tag_adjustment_creation_service.rb
```diff
@@ -20,14 +20,30 @@ def update_tags_and_notify
...
     article.update!(tag_list: article.tag_list.add(@tag_adjustment.tag_name)) if @tag_adjustment.adjustment_type == "addition"
   end
 
   def creation_args
     args = @tag_adjustment_params
+
+    tag =
+      case args[:adjustment_type]
+      when "removal"
+        article.tags.detect { |article_tag| article_tag.name.casecmp(args[:tag_name]).zero? }
```
atsmith813, line 39: This is done to do an efficient case insensitive search - we just grab the article's tag records (max 4) and loop through them, ignoring case.

This was done instead of adding an index to the `tags` table to do a `lower(name)` query. This is all going to change once we _actually_ fix tags.

app/services/tag_adjustment_creation_service.rb
```diff
@@ -20,14 +20,30 @@ def update_tags_and_notify
   private
 
   def update_article
-    article.update!(tag_list: article.tag_list.remove(@tag_adjustment.tag_name)) if @tag_adjustment.adjustment_type == "removal"
+    if @tag_adjustment.adjustment_type == "removal"
+      removed_tags = article.tag_list.select { |tag| tag.casecmp(@tag_adjustment.tag_name).zero? }
+      return if removed_tags.empty?
```
atsmith813, line 25: Don't even bother touching the polymorphic sorcery of the `acts-as-taggable-on-gem` if we don't have any removed tags

PR #8046, opened 2020-05-25
RSS feed: add article tags as categories

rhymes:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We list tags in the API but not in the RSS feeds. According to the [RSS 2.0 spec](https://validator.w3.org/feed/docs/rss2.html#ltcategorygtSubelementOfLtitemgt) a `<item>` can have optional `<category>` elements, one for each category.

Closes https://github.com/thepracticaldev/dev.to/issues/1128

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

**We need to purge the "feed" surrogate key in Fastly**


app/controllers/articles_controller.rb
```diff
@@ -7,6 +7,17 @@ class ArticlesController < ApplicationController
...
+  FEED_ALLOWED_TAGS = %w[
+    strong em a table tbody thead tfoot th tr td col colgroup del p h1 h2 h3 h4
+    h5 h6 blockquote iframe time div span i em u b ul ol li dd dl dt q code pre
+    img sup cite center br small
+  ].freeze
+
+  FEED_ALLOWED_ATTRIBUTES = %w[
+    href strong em class ref rel src title alt colspan height width size rowspan
+    span value start data-conversation data-lang id
+  ].freeze
```
rhymes: these were buried in the RSS builder

app/controllers/articles_controller.rb
```diff
@@ -7,6 +7,17 @@ class ArticlesController < ApplicationController
...
+  FEED_ALLOWED_TAGS = %w[
+    a b blockquote br center cite code col colgroup dd del div dl dt em em h1 h2
+    h3 h4 h5 h6 i iframe img li ol p pre q small span strong sup table tbody td
+    tfoot th thead time tr u ul
+  ].freeze
+
+  FEED_ALLOWED_ATTRIBUTES = %w[
+    alt class colspan data-conversation data-lang em height href id ref rel
+    rowspan size span src start strong title value width
+  ].freeze
```
rhymes, line 19: These were buried in the RSS builder, I brought them up here and sorted them

app/controllers/articles_controller.rb
```diff
@@ -24,10 +35,10 @@ def feed
     end
 
     set_surrogate_key_header "feed"
+    set_cache_control_headers(10.minutes.to_i, stale_while_revalidate: 30, stale_if_error: 1.day.to_i)
```
rhymes, line 38: a little bit clearer to the human eye.

app/controllers/articles_controller.rb
```diff
@@ -24,10 +35,10 @@ def feed
...
 
-    set_cache_control_headers(600,
-                              stale_while_revalidate: 30,
-                              stale_if_error: 86_400)
+    @allowed_tags = FEED_ALLOWED_TAGS
```
citizen428, line 40: Since we already have a `render` right below, maybe pass them in as `locals`?

rhymes: done :) https://github.com/thepracticaldev/dev.to/pull/8046/commits/f775af85e1ac8422e11198be0aef1f019ad2635f

PR #8048, opened 2020-05-25
Update the tag moderation email

Ridhwana:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Update the tag moderation email body, as per https://github.com/thepracticaldev/tech-private/issues/428

https://github.com/thepracticaldev/tech-private/issues/428

screencapture-localhost-3000-rails-mailers-notify-mailer-tag-moderator-confirmation-email-2020-05-27-13_35_14

## Added tests?
- updated tests
- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

alt_text


app/views/mailers/notify_mailer/tag_moderator_confirmation_email.text.erb
```diff
@@ -1,18 +1,25 @@
...
 
-To aid in communication, we’ve added you to a tag moderator <%= community_name %> Connect channel: <%= app_url("/connect/tag-moderators") %>.  You can use this channel to meet other tag moderators across site, and to coordinate with the <%= community_name %> team. When in doubt about how to exercise your new privileges, please ask!
+For details on all available features and how to access them, visit our tag moderation page.
 
-Tag moderation is still very much in beta, so we’ll be making adjustments and expanding features over time as we receive feedback.
+To aid in communication, we’ve added you to the following <%= community_name %> Connect channels:
 
-Thanks for volunteering your time to benefit the <%= community_qualified_name %>.
+- Tag Moderators <%= app_url("/connect/tag-moderators") %>
+- Name of tag mod specific channel
```
Ridhwana, line 16: Still going to make this var dynamic

app/labor/assign_tag_moderator.rb
```diff
@@ -17,7 +17,7 @@ def self.add_tag_moderators(user_ids, tag_ids)
       add_tag_mod_role(user, tag)
       add_trusted_role(user)
       add_to_chat_channels(user, tag)
-      NotifyMailer.tag_moderator_confirmation_email(user, tag).deliver unless tag.name == "go"
```
Ridhwana, line 20: Go was the first tag that we added multiple tag mods too. We initially stopped the welcome email from going out for Go Mods b/c we thought the welcome mail might be confusing. Hence, this can be removed now.
https://dev-internal.slack.com/archives/C7GE84DEJ/p1587112951346400

app/labor/assign_tag_moderator.rb
```diff
@@ -48,4 +48,8 @@ def self.remove_tag_moderator(user, tag)
     Rails.cache.delete("user-#{user.id}/tag_moderators_list")
     MailchimpBot.new(user).manage_tag_moderator_list
   end
+
+  def self.chat_channel_slug(tag)
+    tag.mod_chat_channel&.slug
```
Ridhwana, line 53: We need to pass the chat channel slug through the email now if there is one to point the user to the new channel that they were added to

Ridhwana:
Will look into why the tests on Travis are failing, they seem to be unrelated to my PR.

app/mailers/notify_mailer.rb
```diff
@@ -112,9 +112,10 @@ def export_email(user, attachment)
     mail(to: @user.email, subject: "The export of your content is ready")
   end
 
-  def tag_moderator_confirmation_email(user, tag)
+  def tag_moderator_confirmation_email(user, tag, channel_slug)
     @tag = tag
     @user = user
+    @channel_slug = channel_slug
```
nickytonline, line 118: I see that `@channel_slug is set to be used in the email templates, but how is it passed in? Ruby/Rails newb question probably. 😉

spec/mailers/previews/notify_mailer_preview.rb
```diff
@@ -39,7 +39,7 @@ def channel_invite_email
   end
 
   def tag_moderator_confirmation_email
-    NotifyMailer.tag_moderator_confirmation_email(User.first, Tag.find(1))
+    NotifyMailer.tag_moderator_confirmation_email(User.last, Tag.find(1), nil)
```
nickytonline, line 42: Just curious what the reasoning is for changing this to `User.last`.

Ridhwana: I changed it to User.last just coz I was testing with my user and it happened to be the last user. It really wouldn't make a difference, but I changed it back to User.first coz I shouldn't have committed that. Thanks!

app/mailers/notify_mailer.rb
Ridhwana: You'll see in assign_tag_moderator in the file above, the channel slug gets passed in from NotifyMailer.tag_moderator_confirmation_email(user, tag, chat_channel_slug(tag)).deliver. 

Then, when we set channel_slug to an instance variable i.e. `@channel_slug`. Rails is able to access instance variables in the view.

Hope that answers your question?

citizen428: @nickytonline Rails has a concept called "view contexts". When calling `render`, this gets instantiated like this (code from `ActionView::Rendering`):

```ruby
    def view_context
      view_context_class.new(lookup_context, view_assigns, self)
    end
```

`view_assingns` is a helper defined in `AbstractController::Rendering`):

```ruby
    def view_assigns
      variables = instance_variables - _protected_ivars

      variables.each_with_object({}) do |name, hash|
        hash[name.slice(1, name.length)] = instance_variable_get(name)
      end
    end
```

So the instance variables essentially get copied from the controller to the view context, which is why you can access them in the views.

PR #8054, opened 2020-05-25
Podcast Player native bridge calls refactor to use message based communication

fdoxyz:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR refactors the interaction with the Native Bridge for mobile apps. It reduces some `if else` logic and now both Android & iOS rely on "message based" (Pub/Sub style) which also makes it easier to read/understand. 

Meaning this type of code:
```
if (isNativeIOS()) {
  // iOS code
} else if (isNativeAndroid()) {
  // Android code
} else {
  // Web code
}
```

now becomes this:
```
if (isNativePlayer()) {
  // message to Native Player (both iOS & Android)
} else {
  // Web code
}
```

Also supports some small bug fixes (see PRs linked below for details).

- [Android repo PR](https://github.com/thepracticaldev/DEV-Android/pull/75)
- [iOS repo PR](https://github.com/thepracticaldev/DEV-ios/pull/227)

I tested all three PR's and **there's no problem with merging them in any order** 🎉

The Podcast experience in both iOS & Android remains stable until all three PRs are released, which is when we start to enjoy the best experience.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

makeover

### DIFF
--- a/app/assets/javascripts/initializers/initializePodcastPlayback.js
+++ b/app/assets/javascripts/initializers/initializePodcastPlayback.js
@@ -521,13 +521,14 @@ function initializePodcastPlayback() {
     if (message.action === 'tick') {
       currentState.currentTime = message.currentTime;
       currentState.duration = message.duration;
-      saveMediaState(currentState);
       updateProgress(currentState.currentTime, currentState.duration, 100);
     } else if (message.action === 'init') {
       getById('time').innerHTML = 'initializing...';
+      currentState.currentTime = 0;
     } else {
       console.log('Unrecognized podcast message: ', message); // eslint-disable-line no-console
     }
+    saveMediaState(currentState);
   }
 
   function addMutationObserver() {
--- a/app/assets/javascripts/initializers/initializePodcastPlayback.js
+++ b/app/assets/javascripts/initializers/initializePodcastPlayback.js
@@ -1,5 +1,3 @@
-
-
 /**
  * This script hunts for podcast's "Record" for both the podcast_episde's
  * show page and an article page containing podcast liquid tag. It handles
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/initializers/initializePodcastPlayback.js
```diff
@@ -188,13 +189,11 @@ function initializePodcastPlayback() {
   }
 
   function loadAudio(audio) {
-    if (isNativeIOS()) {
-      sendPodcastMessage({
+    if (isNativePlayer()) {
```
citizen428, line 192: A potential idea for the future: introduce a strategy pattern ([example](https://dev.to/carlillo/design-patterns---strategy-pattern-in-javascript-2hg3)) or something similar to only perform this check once, then call the correct strategy (native/non-native) throughout the rest of the code, removing a lot of conditionals in the process.

fdoxyz: That's a very interesting approach! There definitely are many ways we can optimize and refactor to streamline the logic here. I bookmarked that pattern design series to circle back to it eventually, not just for this specific area 😃

citizen428: @fdoxyz Since our choices seem to be mutually exclusive (web, iOS, or browser) in certain cases, I think we could use this to greatly simplify quite a bit of code. I'd be happy to discuss this in more detail if you ever want to tackle this, maybe during the next downtime or something.

PR #8055, opened 2020-05-26
Fix broken tag filters in listings

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently in production tag filters can't be removed:

before

This PR fixes it and adds tabbing index to the tags so they can be selected with the keyboard.

While checking for a11y issues I also noticed how `categories` were mis-rendered by using a non existent ID. I fixed that as well.

after

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/assets/stylesheets/classified_listings.scss
+++ b/app/assets/stylesheets/classified_listings.scss
@@ -49,6 +49,7 @@
       }
     }
     .classified-filters-tags {
+      display: flex;
       padding: 20px 0px;
       position: relative;
       input {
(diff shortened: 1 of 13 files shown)


citizen428:
@rhymes If we merge this before #7910 I'll cry. That monster of a PR finally has green specs and I already merged master back twice.

rhymes:
@citizen428 ok, let's hold off for that, this one is quite smaller

citizen428:
Thank you @rhymes, that's much appreciated 🙏

rhymes:
Going to re-do this as it's easier than to fix the conflicts :D

PR #8059, opened 2020-05-26
Add unique indexes - part 4

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Follow up to https://github.com/thepracticaldev/dev.to/pull/8050

- `organizations.secret`
- `chat_channel_memberships.chat_channel_id` + `chat_channel_memberships.user_id`
- `chat_channels.slug`
- `badge_achievements.badge_id` + `badge_achievements.user_id`

## Added tests?

- [x] yes
- [ ] no, because I need help

**pre-deployment**, we need to remove duplicates in the following tables:

**chat channels**

```sql
select slug, count(*)
from chat_channels
group by slug
having count(*) > 1
```

**badge achievements**

```sql
select badge_id, user_id, count(*)
from badge_achievements
group by badge_id, user_id
having count(*) > 1
```

### DIFF
--- a/app/models/chat_channel_membership.rb
+++ b/app/models/chat_channel_membership.rb
@@ -5,13 +5,17 @@ class ChatChannelMembership < ApplicationRecord
   SEARCH_SERIALIZER = Search::ChatChannelMembershipSerializer
   SEARCH_CLASS = Search::ChatChannelMembership
 
+  ROLES = %w[member mod].freeze
+  STATUSES = %w[active inactive pending rejected left_channel removed_from_channel joining_request].freeze
+
   belongs_to :chat_channel
   belongs_to :user
 
-  validates :user_id, presence: true, uniqueness: { scope: :chat_channel_id }
   validates :chat_channel_id, presence: true, uniqueness: { scope: :user_id }
-  validates :status, inclusion: { in: %w[active inactive pending rejected left_channel removed_from_channel joining_request] }
-  validates :role, inclusion: { in: %w[member mod] }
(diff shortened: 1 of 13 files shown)


app/models/response_template.rb
```diff
@@ -14,7 +13,7 @@ class ResponseTemplate < ApplicationRecord
   USER_NIL_TYPE_OF_MSG = "cannot have a user ID associated.".freeze
 
   validates :type_of, :content_type, :content, :title, presence: true
-  validates :content, uniqueness: { scope: UNIQUENESS_SCOPE }
+  validates :content, uniqueness: { scope: %i[user_id type_of content_type] }
```
rhymes, line 16: this is because Rubocops unique index static checker needs the columns to be defined inline to be able to validate composite indexes

app/models/chat_channel_membership.rb
```diff
@@ -89,7 +93,11 @@ def other_user
   end
 
   def permission
-    errors.add(:user_id, "is not allowed in chat") if chat_channel.direct? && chat_channel.slug.split("/").exclude?(user.username)
+    return unless chat_channel
```
mstruve, line 96: Wouldn't it be invalid without a chat_channel? Is it necessary to even check this?

rhymes: I added it because in one of the tests it ended up triggering an error as chat_channel wasn't nil. I think it's because ordering could trigger one before the other?

PR #8063, opened 2020-05-26
Preload InstantClick Linked Pages on Mousedown instead of Hover

mstruve:
- [x] Optimization

## Description
We use [InstantClick](http://instantclick.io/) to help preload certain pages for links so when someone clicks on a link the transition seems instantaneous. Given our heavy use of Fastly many of our pages are already lightning quick. For this reason I think it would be worth it to add a couple of ms to a request in order to save what I think will be thousands of useless server requests.  

Possbily related: https://github.com/thepracticaldev/dev.to/issues/1771

## Added tests?

alt_text

### DIFF
--- a/app/assets/javascripts/base.js.erb
+++ b/app/assets/javascripts/base.js.erb
@@ -672,7 +672,7 @@ var instantClick
     InstantClick.on('change', function() {
       initializePage();
     });
-    InstantClick.init();
+    InstantClick.init('mousedown');
   }
 
 // INITIALIZE/ERROR HANDLING
--- a/app/assets/javascripts/base.js.erb
+++ b/app/assets/javascripts/base.js.erb
@@ -672,7 +672,7 @@ var instantClick
     InstantClick.on('change', function() {
       initializePage();
     });
-    InstantClick.init('mousedown');
+    InstantClick.init(50);
   }
 
 // INITIALIZE/ERROR HANDLING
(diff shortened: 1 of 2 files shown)


benhalpern:
I think we should keep IntantClick as is. A huge amount of the "lightning quick" feel is because we use Instantclick in the first place. The Fastly thing is just a part of that.

mstruve:
Right, but how much "instant" is expected? I think by preloading on mousedown we will save ourselves a lot of useless server requests and I don't really think people will notice much. Not to mention, those who are short on bandwidth will have a worse experience bc it will be tied up with requests they probably dont even want made.

mstruve:
Could we at least try it and see how many requests we save? Maybe it's negligible, and we go back, or maybe its huge and now we have a lot more bandwidth to scale

benhalpern:
Since ~88% of these types of requests never hit our servers and are instead handled at the edge, we might save _some_ but I don't think it's worth such a big change. Of course, we still pay for the Fastly data, but it's not much.

I think we could modify the code to account for device setting APIs like [`navigator.connection.saveData`](https://dev.to/addyosmani/adaptive-loading-improving-web-performance-on-low-end-devices-1m69), as the bandwidth used on these requests is in the realm of 6-12kb per request which can be dwarfed by a single 100kb+ image, so for users not connecting in these ways I think the math favors loading on hover.

InstantClick does have the concept of delaying by a few milliseconds the preload to lessen the # of requests for incidental hovers and stuff. I'd probably be more in favor of that.

mstruve:
Changed it to 50ms which was a suggested value http://instantclick.io/preloading

mstruve:
Changing this timing is not going to work bc it messes with our spec timing and I don't feel like wrestling with our specs for a day trying to fix it.

PR #8064, opened 2020-05-26
Sync public reactions count with positive reactions count

Zhao-Andy:
- [x] Bug Fix

## Description
This adds a data update script to sync `article.public_reactions_count`, `article.previous_public_reactions_count`, and `comment.public_reactions_count` with positive reactions count again. This is necessary since #7926 was merged today.

"Fantastic! One more time!"

### DIFF
--- a/lib/data_update_scripts/20200526181850_update_public_reaction_counts_again.rb
+++ b/lib/data_update_scripts/20200526181850_update_public_reaction_counts_again.rb
@@ -1,0 +1,10 @@
+module DataUpdateScripts
+  class UpdatePublicReactionCountsAgain
+    def run
+      ActiveRecord::Base.connection.execute("UPDATE articles SET public_reactions_count = positive_reactions_count, previous_public_reactions_count = previous_positive_reactions_count WHERE id < #{Article.count / 2}")
+      ActiveRecord::Base.connection.execute("UPDATE articles SET public_reactions_count = positive_reactions_count, previous_public_reactions_count = previous_positive_reactions_count WHERE id > #{Article.count / 2}")
+      ActiveRecord::Base.connection.execute("UPDATE comments SET public_reactions_count = positive_reactions_count WHERE id < #{Comment.count / 2}")
+      ActiveRecord::Base.connection.execute("UPDATE comments SET public_reactions_count = positive_reactions_count WHERE id > #{Comment.count / 2}")
+    end
+  end
+end
--- a/lib/data_update_scripts/20200526181850_update_public_reaction_counts_again.rb
+++ b/lib/data_update_scripts/20200526181850_update_public_reaction_counts_again.rb
@@ -1,10 +1,12 @@
 module DataUpdateScripts
   class UpdatePublicReactionCountsAgain
     def run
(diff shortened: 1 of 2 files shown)


lib/data_update_scripts/20200526181850_update_public_reaction_counts_again.rb
```diff
@@ -0,0 +1,10 @@
+module DataUpdateScripts
+  class UpdatePublicReactionCountsAgain
+    def run
+      ActiveRecord::Base.connection.execute("UPDATE articles SET public_reactions_count = positive_reactions_count, previous_public_reactions_count = previous_positive_reactions_count WHERE id < #{Article.count / 2}")
+      ActiveRecord::Base.connection.execute("UPDATE articles SET public_reactions_count = positive_reactions_count, previous_public_reactions_count = previous_positive_reactions_count WHERE id > #{Article.count / 2}")
+      ActiveRecord::Base.connection.execute("UPDATE comments SET public_reactions_count = positive_reactions_count WHERE id < #{Comment.count / 2}")
+      ActiveRecord::Base.connection.execute("UPDATE comments SET public_reactions_count = positive_reactions_count WHERE id > #{Comment.count / 2}")
```
Zhao-Andy, line 7: I ran this in production console earlier, and don't think I need to adjust our database timeout for this.

lib/data_update_scripts/20200526181850_update_public_reaction_counts_again.rb
```diff
@@ -0,0 +1,10 @@
+module DataUpdateScripts
+  class UpdatePublicReactionCountsAgain
+    def run
+      ActiveRecord::Base.connection.execute("UPDATE articles SET public_reactions_count = positive_reactions_count, previous_public_reactions_count = previous_positive_reactions_count WHERE id < #{Article.count / 2}")
+      ActiveRecord::Base.connection.execute("UPDATE articles SET public_reactions_count = positive_reactions_count, previous_public_reactions_count = previous_positive_reactions_count WHERE id > #{Article.count / 2}")
```
rhymes, line 5: I think one of these two should have the `<=` or `>=` but I think it's fine anyway. Nice trick to split the table in two :)

lib/data_update_scripts/20200526181850_update_public_reaction_counts_again.rb
mstruve: Can you comment out the first data update script so we don't end up running this twice?

Also, since this will be the final script we should have a `>=` for each one bc an ID could be equal to the count divided by 2.

lib/data_update_scripts/20200526181850_update_public_reaction_counts_again.rb
mstruve: Wow we are definitely on the same page

Zhao-Andy: Ahh right haha

mstruve: Oh in addition to this, we should be storing the counts in a variable bc technically the counts could change between running each statement. I know its a small chance but its still a chance

Zhao-Andy: 👍 updated

PR #8065, opened 2020-05-26
Consolidate Setup Tasks to a Single Rake Task

mstruve:
- [x] Optimization

## Description
This PR consolidates our application setup database logic to a single rake task. The reason we want to do this is so we don't have to worry about which steps to execute and in what order every time we want to setup the app somewhere. Eventually this might have things like health checks etc but for now I stuck with consolidating what we have.

- Code climate can be ignored
- Spec that failed is a flaky one

https://github.com/thepracticaldev/dev.to/projects/9#card-38620911

## Added tests?

alt_text


lib/tasks/app_setup.rake
```diff
@@ -0,0 +1,20 @@
...
+  task setup: :environment do
+    puts "\n== Preparing Elasticsearch =="
+    Rake::Task["search:setup"].execute
+
+    puts "\n== Preparing database =="
+    begin
+      Rake::Task["db:migrate"].execute
+    rescue ActiveRecord::NoDatabaseError
+      puts "\n== Creating and Seeding database =="
+      Rake::Task["db:create"].execute
```
mstruve, line 12: `Rake::Task["db:setup"].execute` for some reason does not work here. I had to break it down and explicitly call each task that it executes.

lib/tasks/app_setup.rake
```diff
@@ -0,0 +1,20 @@
+namespace :app do
+  desc "Prepare Application on Boot Up"
+  task setup: :environment do
+    puts "\n== Preparing Elasticsearch =="
+    Rake::Task["search:setup"].execute
+
+    puts "\n== Preparing database =="
+    begin
+      Rake::Task["db:migrate"].execute
+    rescue ActiveRecord::NoDatabaseError
```
mstruve, line 10: This should also work better than the way we were doing it before. I have been caught off guard a couple of times when my database was reset due to a random migration that had nothing to do with a database missing.

mstruve:
@jacobherrington I don't believe so at this point! The docs talk about running `bin/setup` but never really break down what it does. We have a lot of docs about setting up databases, etc and I think once we have a "DEV in a box" solution or are closer to it then we will need to update setup docs with the one path that will be used to stand up the app.

lib/tasks/app_setup.rake
```diff
@@ -0,0 +1,20 @@
+namespace :app do
```
rhymes, line 1: FYI, `app:` is used by Rails. As of Rails 6: `app:template` and `app:update` are defined. This won't conflict with `app:setup` but we might need an update script in the future.

lib/tasks/app_setup.rake
rhymes: Cool! Fortunately Rails 6 adds `db:prepare` which does just that. I'll update the branch when we merge this

lib/tasks/app_setup.rake
rhymes: Don't know exactly what the difference is, it might be because `db:setup` is just a reference to other 3 tasks run inline. You can use `system!()` here I guess.

Also to be forward looking we might want to just clone the tasks `db:setup` runs:

https://github.com/rails/rails/blob/7b5cc5a5dfcf38522be0a4b5daa97c5b2ba26c20/activerecord/lib/active_record/railties/databases.rake#L188-L189

My preference goes to `system()` though, as it allows us to pick up Rails changes. If that doesn't work with Docker, we can stick with the list of tasks in the Rails code

lib/tasks/app_setup.rake
mstruve: oh snap, thanks for the heads up! I will change it so we don't have issues.

lib/tasks/app_setup.rake
mstruve: I'll play around with this

lib/tasks/app_setup.rake
mstruve: in the future `app_initializer` could be a project name but this keeps it general for now

PR #8071, opened 2020-05-26
Add App, Database, and Search Health Check Endpoints

mstruve:
- [x] Feature

## Description
Going forward if we want to support an "automatic setup" flow we need health check endpoints to help our processes determine when portions of our app are up and ready to receive traffic. We can also monitor these using tools like Datadog to alert if something goes down. 

This PR covers Postgres, Elasticsearch, and the App in general. We will likely want one for all our cache stores as well but I will handle that in a separate PR.

https://github.com/thepracticaldev/dev.to/projects/9#card-38620995

## Added tests?
- [x] yes

alt_text


config/routes.rb
```diff
@@ -1,6 +1,11 @@
 # rubocop:disable Metrics/BlockLength
 
 Rails.application.routes.draw do
+  # Health Check Endpoints
+  get "/health_check" => "health_check#ping"
+  get "/search_health_check" => "health_check#search_ping"
+  get "/database_health_check" => "health_check#database_ping"
```
jdoss, line 7: What about a /redis_health_check too?

jdoss: Nevermind! I cannot read...

>This PR covers Postgres, Elasticsearch, and the App in general. We will likely want one for all our cache stores as well but I will handle that in a separate PR.

app/controllers/health_check_controller.rb
```diff
@@ -0,0 +1,21 @@
+class HealthCheckController < ApplicationController
```
rhymes, line 1: should we rate limit these? :D 

Aside from rate limiting, whose default could be fine, should we use some sort of authentication method? Basic authentication with a server side key is fine. I'm pretty sure any "pinger" out there can use basic auth, even `curl` can. 

I'm a bit wary of the idea that third parties can check this endpoint at random, especially the one about search. Why should they know if Elasticsearch is up or not?

Something like:

```ruby
class HealthCheckController < ApplicationController
   http_basic_authenticate_with name: AppConfig["HEALTH_CHECK_AUTH_USERNAME"], password: AppConfig["HEALTH_CHECK_AUTH_PASSWORD"], except: :index
end
```

and then:

```shell
$ curl -u username:password https://dev.to/health-checks/app
```

Last but not least, what if we use the "canary healtheck" strategy ? Instead of 3 endpoints, we build one, which returns a structured JSON with all the results, something like:

```json
{
  "app": {
    "success": true
  },
  "db": {
    "success": true
  },
  "search": {
    "success": true
  }
}
```

this way metrics can be added in the future without re-doing the client (which, if not updated will just keep checking true or false) and we'd have one endpoint

ps. have you considered making this an actual API controller?

config/routes.rb
rhymes, line 7: what about `/health-checks/app`, `/health-checks/search`, ``/health-checks/database` ?

app/controllers/health_check_controller.rb
mstruve: > should we rate limit these? :D

Nah I think our defaults are sufficient

> have you considered making this an actual API controller?

Yeah I considered making an API controller with authentication but then did some testing and figured out that we did not authenticate these endpoints at Kenna. I couldn't remember if there was a reason for that or not so I stuck with that approach in case there was a reason. After chatting with @jdoss he would rather have them authenticated so I will update that. 

> what if we use the "canary healtheck" strategy

I think having separate ones is more explicit and allows us more flexibility on our checks so once the db comes back as up we can stop checking that, etc. This could totally change down the road but I like separate endpoints to start.

rhymes: >  I think having separate ones is more explicit and allows us more flexibility on our checks so once the db comes back as up we can stop checking that, etc. This could totally change down the road but I like separate endpoints to start.

I was thinking that the main limitation with this is that if the app is down, both ES and the DB will be flagged as down as these checks both depend on Rails to be up to work.

I guess that's why I had the idea of consolidating them in one single endpoint

app/controllers/api/v0/health_checks_controller.rb
```diff
@@ -0,0 +1,37 @@
+module Api
+  module V0
+    class HealthChecksController < ApiController
+      before_action :authenticate_with_token
```
rhymes, line 4: I'd favor a standard HTTP way to do this, using Rails's [ActionController::HttpAuthentication::Token](https://api.rubyonrails.org/classes/ActionController/HttpAuthentication/Token.html) method:

```ruby
def authenticate
  authenticate_or_request_with_http_token do |token, options|
    # Compare the tokens in a time-constant manner, to mitigate
    # timing attacks.
    ActiveSupport::SecurityUtils.secure_compare(token, TOKEN)
  end
end
```

`ActiveSupport::SecurityUtils.secure_compare` is important, we use it for the `api-key` authentication as well.

This way the client could just be:

`curl -H "Authorization: Token TOKEN_VALUE" https://dev.to/api/healt-checks/app`

mstruve: For sensitive endpoints, I think that would be more important but since these are simple health check endpoints I think the simple comparison will suffice.

PR #8072, opened 2020-05-27
Add unique indexes to articles - part 5

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Follow up to https://github.com/thepracticaldev/dev.to/pull/8059

This only adds missing unique indexes to the `articles` table:

- `slug`+ `user_id` - https://dev.to/admin/blazer/queries/78-duplicate-articles-by-slug-and-user_id
- `feed_source_url` - https://dev.to/admin/blazer/queries/79-duplicate-articles-feed_source_url
- `canonical_url` - https://dev.to/admin/blazer/queries/80-duplicate-articles-canonical_url
- `body_markdown` + `user_id` + `title` - https://dev.to/admin/blazer/queries/81-duplicate-articles-by-body_markdown-user_id-title

## Added tests?

- [x] yes
- [ ] no, because I need help

**pre deployment**: unfortunately we have duplicates in each of the indexes. It's a bit odd we have duplicates on canonical URL and duplicated articles for the same user 😞 

- `slug`+ `user_id`:

```sql
select slug, user_id, count(*)
from articles
group by slug, user_id
having count(*) > 1
```

- `feed_source_url`:

```sql
select feed_source_url, count(*)
from articles
group by feed_source_url
having count(*) > 1
```

- `canonical_url`:

```sql
select canonical_url, count(*)
from articles
group by canonical_url
having count(*) > 1
```

- `body_markdown` + `user_id` + `title` (this times out at time but when it doesn't, it yields results)

```sql
select body_markdown, user_id, title, count(*)
from articles
group by body_markdown, user_id, title
having count(*) > 1
```

### DIFF
--- a/db/migrate/20200527101013_add_unique_indexes_to_articles_slug_feed_source_url_canonical_url_body_markdown.rb
+++ b/db/migrate/20200527101013_add_unique_indexes_to_articles_slug_feed_source_url_canonical_url_body_markdown.rb
@@ -6,16 +6,35 @@ class AddUniqueIndexesToArticlesSlugFeedSourceUrlCanonicalUrlBodyMarkdown < Acti
       remove_index :articles, column: :feed_source_url, algorithm: :concurrently
     end
 
+    # needed for the digest() function, checks if the extension already exists
+    ActiveRecord::Base.connection.enable_extension("pgcrypto")
+
     add_index :articles, %i[slug user_id], unique: true, algorithm: :concurrently
     add_index :articles, :feed_source_url, unique: true, algorithm: :concurrently
     add_index :articles, :canonical_url, unique: true, algorithm: :concurrently
-    add_index :articles, %i[body_markdown user_id title], unique: true, algorithm: :concurrently
+
+    # to avoid the error "Values larger than 1/3 of a buffer page cannot be indexed."
+    # due to the fact that large text column cannot be indexed by btree indexes
+    # we are going to need to build an index on the hash of the `body_markdown` column.
+    # We also cannot use a `HASH` index as it's not supported by unique columns.
+    # I don't recommend using GiN or GiST as well
+    # See <https://www.postgresql.org/message-id/AANLkTin3p6VS1Z=TtqUV-5cG4TZpTjUMfuPNzWJFgnr5@mail.gmail.com>,
+    # <https://www.postgresql.org/docs/11/pgcrypto.html#id-1.11.7.34.5> and
+    # <https://www.postgresql.org/docs/11/indexes-types.html>
+    # NOTE: using SQL as I couldn't find a way to have Rails generate it correctly
+    ActiveRecord::Base.connection.execute(
+      <<~SQL
+        CREATE UNIQUE INDEX CONCURRENTLY "index_articles_on_digest_body_markdown_and_user_id_and_title"
+        ON "articles"
+        USING btree (digest("body_markdown", 'sha512'::text), "user_id", "title");
+      SQL
+    )
   end
 
   def down
     remove_index :articles, column: %i[slug user_id], algorithm: :concurrently
     remove_index :articles, column: :feed_source_url, algorithm: :concurrently
     remove_index :articles, column: :canonical_url, algorithm: :concurrently
-    remove_index :articles, column: %i[body_markdown user_id title], algorithm: :concurrently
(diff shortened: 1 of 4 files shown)


rhymes:
The failing test is the usual flaky one :-(

db/migrate/20200527101013_add_unique_indexes_to_articles_slug_feed_source_url_canonical_url_body_markdown.rb
```diff
@@ -0,0 +1,40 @@
+class AddUniqueIndexesToArticlesSlugFeedSourceUrlCanonicalUrlBodyMarkdown < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def up
+    if index_exists?(:articles, :feed_source_url)
+      remove_index :articles, column: :feed_source_url, algorithm: :concurrently
+    end
+
+    # needed for the digest() function, checks if the extension already exists
+    ActiveRecord::Base.connection.enable_extension("pgcrypto")
```
mstruve, line 10: Should we add an `unless` statement here like they do in the edge docs? https://edgeguides.rubyonrails.org/active_record_postgresql.html#uuid-primary-keys
```ruby 
enable_extension 'pgcrypto' unless extension_enabled?('pgcrypto')
```

Looks like we can remove the `ActiveRecord::Base.connection` as well.

rhymes: as the comment says, that check is builtin in Rails :-D 

> is there any chance that this extension can't be enabled for some reason on a Postgres db

nope, it's builtin - https://www.postgresql.org/docs/11/pgcrypto.html.

If we ever encounter a custom version of PostgreSQL that has that module disabled somehow I think it should explode as pgcrypto contains all the cryptographic functions anyway

rhymes:
Marking this as a draft as we have quite a few duplicated rows in production which seem to be created by the RSS importer and we need to figure out why before we clean them up.

db/schema.rb
```diff
@@ -122,14 +123,17 @@
     t.string "video_source_url"
     t.string "video_state"
     t.string "video_thumbnail_url"
+    t.index "digest(body_markdown, 'sha512'::text), user_id, title", name: "index_articles_on_digest_body_markdown_and_user_id_and_title", unique: true
```
benhalpern, line 126: Even though I loved the exploration in your post, I think we may want to question whether `body_markdown` should actually be "unique" from a usability perspective.

We've had this in, but I'm not sure it's the ideal behavior in all cases. I think getting the logic into the database is the point of no return vs the rails validators which are more flexible.

I think the biggest reason this was originally added to the validators was to deal with the mishap of situations where folks could accidentally submit the same new post twice depending on how they hit refresh or whatever in the browser at the wrong time... Which could still be avoided.

But I could see a legit case for a "repeating post" or something along those lines where you don't want the `body_markdown` to be entirely unique.

rhymes: The main advantage of having this in the DB is that there's still not a 100% guarantee that doing so in Rails won't lead to duplicate posts. Mainly because of concurrency. The chances are very slim, but double posting is not completely unheard of, which is why we have the uniqueness check anyway. I don't think it's completely a point of no-return as we can easily remove it with an `alter table`.

Rails validators would stay in, the whole idea is to have both. But a Rails validator is going to be less performing if it's not backed by a unique index. While I was exploring this I ended up in Rails source code and what Rails basically does is to do a `SELECT` with all the scoped fields (`body_markdown`, `user_id`, `title` in this instance), see if it returns something and if it does, they trigger the validator, if it doesn't they let the operation go through.

This is where the unique index inserts itself:

- it quickens that select by using the index instead of walking through all rows
- it avoid the improbable but still impossible case of an insert happening at the same time (the double posting)

We likely found thousands of duplicates (most likely related to the RSS importer, not probably due to doubleposting) in the database which is probably why a validator alone is not sufficient.

I still have to investigate why we have so many duplicates (how they got in).

rhymes:
Now that all of this has been tackled in different PRs, of which https://github.com/forem/forem/pull/10273 and https://github.com/forem/forem/pull/10284 should be the latest, we can close this and bid it farewell.

PR #8077, opened 2020-05-27
Comment out Percy Code Failing Specs to Unblock Master

mstruve:
- [x] Bug Fix

## Description
Specs are currently failing due to having to freeze time for Percy but that timestamp is too far in the past so users are constantly logged out after a single action. This comments out that Percy code until we can come up with a fix for it. 

alt_text

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -156,14 +156,14 @@ RSpec.configure do |config|
   # config.filter_gems_from_backtrace("gem name")
 
   # Explicitly set a seed and time to ensure deterministic Percy snapshots.
-  config.around(:each, percy: true) do |example|
-    Timecop.freeze("2020-05-13T10:00:00Z")
-    prev_random_seed = Faker::Config.random
-    Faker::Config.random = Random.new(42)
+  # config.around(:each, percy: true) do |example|
+  #   Timecop.freeze("2020-05-13T10:00:00Z")
+  #   prev_random_seed = Faker::Config.random
+  #   Faker::Config.random = Random.new(42)
 
-    example.run
+  #   example.run
 
-    Faker::Config.random = prev_random_seed
-    Timecop.return
-  end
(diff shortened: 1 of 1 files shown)


spec/rails_helper.rb
```diff
@@ -156,14 +156,14 @@
...
 
   # Explicitly set a seed and time to ensure deterministic Percy snapshots.
-  config.around(:each, percy: true) do |example|
-    Timecop.freeze("2020-05-13T10:00:00Z")
-    prev_random_seed = Faker::Config.random
-    Faker::Config.random = Random.new(42)
+  # config.around(:each, percy: true) do |example|
+  #   Timecop.freeze("2020-05-13T10:00:00Z")
+  #   prev_random_seed = Faker::Config.random
+  #   Faker::Config.random = Random.new(42)
```
rhymes, line 162: this can probably remain but let's unblock master first and figure that out later. thanks Molly!

spec/rails_helper.rb
```diff
@@ -156,14 +156,14 @@
...
 
-    example.run
+  #   example.run
 
-    Faker::Config.random = prev_random_seed
-    Timecop.return
-  end
+  #   Faker::Config.random = prev_random_seed
+  #   Timecop.return
+  # end
```
vaidehijoshi, line 168: If we're going to comment this out, we should probably also just remove the PERCY_TOKEN from the travis env vars.

mstruve: I left it bc I figured we would want to try to fix it at some point.

vaidehijoshi: (Otherwise Percy is going to be getting a bunch of wonky snapshots without this setting)

vaidehijoshi: @mstruve Sorry, to clarify: I mean that we should temporarily remove the PERCY_TOKEN from Travis settings until we can figure out how to fixed, which will effectively disable Percy until we can fix this.

PR #7528, opened 2020-04-27
Remove index_chat_channels_on_slug from schema

maestromac:
## What type of PR is this? (check all applicable)
- [x] clean up

## Description
This has to be okay to remove because it still lingers even after `db:reset`

## Related Tickets & Documents
n/a
## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
n/a
## Added tests?
- [x] no, because they aren't needed

## Added to documentation?
- [x] no documentation needed

PR #8103, opened 2020-05-28
Maybe remove index

citizen428:

- [X] Clean up

## Description

Welcome to another episode of "Walking Dead: `index_chat_channels_on_slug` Edition". 

History (`git grep index_chat_channels_on_slug $(git rev-list --all -- db/schema.rb)`:

1. Originally added in #7105
2. Removed, but I can't find where
3. Re-added in #7495 
4. Removed again in #7528
5. Added again in #7777, 514516b18b67b77a5c614f6ab181cf04becc212e and #8059

So, should this index be present or not? If yes, and I'm starting to suspect that might actually be the case, we should leave it in and close this PR. If not, we should remove it and make sure it stays removed, so everyone should drop it manually from their system just in case:

```
$ rails dbconsole
psql (12.3)
Type "help" for help.

PracticalDeveloper_development=# DROP INDEX IF EXISTS index_chat_channels_on_slug;
DROP INDEX
```

n/a

## Added tests?

## Added to documentation?

- [x] no documentation needed

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -314,7 +314,6 @@ ActiveRecord::Schema.define(version: 2020_05_26_151807) do
     t.string "slug"
     t.string "status", default: "active"
     t.datetime "updated_at", null: false
-    t.index ["slug"], name: "index_chat_channels_on_slug", unique: true
   end
 
   create_table "classified_listing_categories", force: :cascade do |t|


rhymes:
I vote for remain, as there is a `uniqueness: true` check in the `ChatChannel` model and all of these Ruby level checks should be backed by a DB index, otherwise we end up in the current situation in which we have duplicate rows in the DB.

See

https://github.com/thepracticaldev/dev.to/blob/29269cd58970a8f8fe2449c41816cc0f5948612c/app/models/chat_channel.rb#L23

and https://github.com/thepracticaldev/dev.to/pull/8059

citizen428:
Finally, some clarity!

image

joshpuetz:
My hypothesis is that somehow this index doesn't appear until chat functionally is tested/visited..which is normally isn't right out of the box with our seed data. I'm also voting leaving it in. All of the times I've seen it appear, I've never had it _removed_ by the system.

citizen428:
Thanks, @rhymes & @joshpuetz, let's keep this index.

image

PR #8107, opened 2020-05-28
Layout updates

ludwiczakpawel:

- [x] Refactor
- [x] Feature

## Description

This PR does few things:

1. It introduces CSS for different types of layout with its responsive options... 1 column layout, 2 columns layout, 3 columns layout + special snowflakes.
2. It updates layout on the home page to use new crayons-layout. Visually there shouldn't be too many differences with what we have currently though.
3. It updates settings layout to use new crayons-layout.
4. It's introduces couple styling changes: 
    - updated body background (slightly darker comparing to what we have currently), 
    - widgets in the right sidebar are now contained in boxes, with new style
    - some labels in buttons were updated to not be capitalized anymore since it doesn't play well with Crayons styling.. **so you may see lots of files in this PR were actually just text changes**

image

## Added tests?

- [x] yes
- [x] no, because I need help

## Added to documentation?

- [x] no documentation needed

1. Go through each display ad in `/admin/display_ads` and check whether images have defined `width` and `height` attributes.
2. Update Campaign image to preserve 100:42 ratio. New artwork below:

image


app/assets/stylesheets/top-bar.scss
```diff
@@ -25,11 +25,17 @@
...
+    padding-left: var(--su-2);
+    padding-right: var(--layout-padding);
     margin: auto;
     display: flex;
     align-items: center;
     position: relative;
+
+    @media (min-width: $breakpoint-m) {
+      padding-left: var(--layout-padding);
+    }
```
ludwiczakpawel, line 38: Little changes here required to make it better responsive.

app/javascript/sidebar-widget/sidebarUser.jsx
```diff
@@ -34,7 +34,7 @@ class SidebarUser extends Component {
             onClick={this.onClick}
             id={`widget-list-item__follow-button-${user.username}`}
           >
-            {user.following ? '✓ FOLLOWING' : '+ FOLLOW'}
+            {user.following ? 'Following' : 'Follow'}
```
nickytonline, line 37: Do we still want to keep the checkmark as well for `Following`?

spec/system/notifications/notifications_page_spec.rb
```diff
@@ -42,8 +42,8 @@ def validate_reply(id)
     Notification.send_new_follower_notification_without_delay(follow, "Published")
     visit "/notifications"
     expect(page).to have_css("div.single-notification")
-    click_button("+ FOLLOW BACK")
-    expect(page).to have_text("FOLLOWING")
+    click_button("Follow back")
```
nickytonline, line 45: Do we want to keep the `+` or is it being omitted on purpose for the new design changes

app/javascript/sidebar-widget/sidebarUser.jsx
ludwiczakpawel: Skipping checkmark and plus was actually intentional. ✅

spec/system/notifications/notifications_page_spec.rb
ludwiczakpawel: Same here

ludwiczakpawel:
Oh good catch on social buttons - i'll fix it!

nickytonline:
One other thing I noticed @ludwiczakpawel is there is no longer an X to close Suggested things you can do section.

Your branch

image

Master

image

ludwiczakpawel:
oh wow... where the heck is that change coming from?! well, i'll investigate :) another good catch @nickytonline !

app/assets/stylesheets/articles.scss
```diff
@@ -85,27 +85,13 @@
     }
 
     .sidebar-nav {
-      margin-bottom: 16px;
       .sidebar-nav-block {
-        @media screen and (min-width: 950px) {
```
lisasy, line 90: It must have felt so good deleting these lines.

app/javascript/leftSidebar/__tests__/__snapshots__/TagsFollowed.test.jsx.snap
```diff
@@ -15,164 +15,120 @@ exports[`<TagsFollowed /> should render the tags followed when tags are passed i
...
+      #webdev
     </a>
   </div>
   <div
     class="sidebar-nav-element"
     id="sidebar-element-react"
   >
     <a
-      class="sidebar-nav-link sidebar-nav-link-tag"
+      class="crayons-link crayons-link--block"
```
lisasy, line 40: 😍

ludwiczakpawel:
Merging this with new "Write a post" view required few additional updates but it should now be all good and ready for merge.

---

@nickytonline all the issues you've mentioned should now be fixed. Really good catch with that onboarding `x` icon :)

benhalpern:
Looks like there's an issue with the editor...

Screen Shot 2020-06-02 at 5 37 25 PM

There's also an issue with the campaign image "jumping" on load. We had previously relied on the assumption that this would be constant width, so we set the height as constant as well. It's not good if the whole sidebar jumps because the image may be slow to load.

ezgif-4-0189c0a3c514

Same with this sidebar... We had been hardcoding the height because width was constant. This is what it would look like with current data (e.g. inline styles)

We may need to create some limitations about the aspect ratio of these things and/or come up with some clever solutions in these places...

ezgif-4-a166608b3c19

I'm also a little concerned that there's no longer as much distinction between "follow" and "following" across the site, harder to tell at a glance your relationship with a user or tag. Twitter has the "follow" state as outlined and the "following" as filled, we may want to consider something like that?

ludwiczakpawel:
@benhalpern  can you double check the editor issue? It works fine on my local env on several browsers.. Caching maybe?

I'll look into sidebar artworks height - definitely shouldn't jump like that, so I'll try to fix it with aspect ratio and we may indeed need to set some guidelines probably. For classic ads (your example with stickers)  I'd probably go for square (1:1 ratio). As for campaign image, maybe we could stick to the same ratio as we have for featured image (100:42) so if needed we could use the same artwork for post cover..?

> I'm also a little concerned that there's no longer as much distinction between "follow" and "following" across the site, harder to tell at a glance your relationship with a user or tag. Twitter has the "follow" state as outlined and the "following" as filled, we may want to consider something like that?

I don't think I have changed the button styling (well, except label - so no more `+` and `✓`) - it has always been _filled_. BUT I do think we should have alternative style whether or not user follows a tag (or user), although I think that should come in a separate PR - there will be lots of JS work involved.

ludwiczakpawel:
@benhalpern I've added a bit of CSS so we can preserve the aspect ratio. I left the appropriate comments in the code. Overall the rule was as mentioned:

**For DisplayAds**, any image within this component should have defined width and height attribute. That combined with CSS will preserve the space even if the image is not loaded. There's no need to have aspect ratio defined as long as we have width & height attributes in the IMG. So some ads can be square, some can be rectangle. **🚨So we may need to update some of our existing ads code to reflect that, after or before deploy.**

```
<img ... width="300" height="100">
```

---

**For Campaign images**, we have to define aspect ratio because we don't have control over code (therefor we can't define width & height) when uploading an image. I've decided to go with 100:42 aspect ratio (the same as featured image in articles).

ludwiczakpawel:
something it going off with the PR.. im marking it again as draft until i figure it out.

benhalpern:
Hmmmmm I am seeing general layout issues, but not sure if it's a problem with the PR or locally. I'll have to come back to this tomorrow. I'll also propose a solution to ensure we can merge this a little more cleanly tomorrow. I have an idea.

Nice stuff all around by the looks of it though.

ludwiczakpawel:
@benhalpern sweet, let me know what's the issue and im looking forward for your proposal.

PR #4221, opened 2019-10-03
Implement Reddit liquid tag

cesc1989:
## What type of PR is this? (check all applicable)
- [x] Feature

## Description
Implement Liquid tag for reddit content.

It takes the post URL and renders a block of content based on the JSON response of requesting that URL and appending `.json`:

```
url =  https://www.reddit.com/r/IAmA/comments/afvl2w/im_scott_from_scotts_cheap_flights_my_profession

json_url =  https://www.reddit.com/r/IAmA/comments/afvl2w/im_scott_from_scotts_cheap_flights_my_profession.json
```

I tried to emulate the same way the Medium liquid tag was coded and that's why the request is made and handled in a different class(`RedditJsonFromUrlService`)

I'm still pending to add tests and improve the styling of the reddit content.

## Related Tickets & Documents
This PR resolves #1536 

## Added to documentation?

- [x] docs.dev.to

## What gif best describes this PR or how it makes you feel?

grrrreat

Issue #1536, opened 2019-01-14
Reddit Liquid Tag


tiagomsmagalhaes:
**Describe the solution you'd like**
It would be nice to have a liquid tag to integrate reddit submissions

**Describe alternatives you've considered**
I haven't researched intensively but found that Reddit has an embedding mechanism (oEmbed ?)

**Additional context**
[A maybe useful GH wiki page about Reddit embedding.](https://github.com/reddit-archive/reddit/wiki/oEmbed)

benhalpern:
This makes sense. I think we'd do it via the API and embed via plain old backend HTML. This is our preference when it's not rich content like CodePen, etc.

Link2Twenty:
If you take any post and put `.json` at the end of it you see the post details in json form.


https://www.reddit.com/r/aww/comments/ag3s4b/ive_waited_28_years_to_finally_havr_my_first_pet/

becomes

https://www.reddit.com/r/aww/comments/ag3s4b/ive_waited_28_years_to_finally_havr_my_first_pet.json

the first item in the array is the original post the second is replies.
If we dig down like this `json[0].data.children[0].data` we get to the data we want

```json
[...]

PR #8111, opened 2020-05-28
Add Reddit Liquid Tag

cesc1989:
## What type of PR is this?

- [x] Feature

## Description
Implement Liquid tag for Reddit content.

It takes the post URL and renders a block of content based on the JSON response of requesting that URL and appending `.json`:

```
url =  https://www.reddit.com/r/IAmA/comments/afvl2w/im_scott_from_scotts_cheap_flights_my_profession

json_url =  https://www.reddit.com/r/IAmA/comments/afvl2w/im_scott_from_scotts_cheap_flights_my_profession.json
```

I followed example from the Medium liquid tag. This is why the request is made and handled in a different class(`RedditJsonFromUrlService`).

In #4221 I had used VCR for testing but it created a huge file. Now, I'm mocking many aspects of the request to Reddit.

This PR solves #1536 and is a continuation of #4221 

## Added tests?

- [x] yes

## Added to documentation?

- [x] docs.dev.to

energized


app/services/reddit_json_from_url_service.rb
```diff
@@ -0,0 +1,58 @@
...
+  def validate_url
+    cleaned_url = @url.delete(" ")
+
+    return true if valid_url?(cleaned_url) && link_exists?
+
+    raise StandardError, "Invalid Reddit link: #{@url}"
+  end
+
+  def valid_url?(url)
+    url = URI.parse(url)
```
rhymes, line 50: I think we should also employ a regexp for reddit URLs, as we do in other liquid tags, for example 

https://github.com/thepracticaldev/dev.to/blob/1a40e10df8d82ed06d4c37cb5a05f4ea71efed0d/app/liquid_tags/codepen_tag.rb#L3

maybe not as complex but enought to not just validate anything thrown at it. 

Otherwise you get the following:

Screenshot_2020-05-28 New Post - DEV(local) Community 👩‍💻👨‍💻

app/services/reddit_json_from_url_service.rb
```diff
@@ -0,0 +1,58 @@
...
+    url.is_a?(URI::HTTP)
+  end
+
+  def link_exists?
```
rhymes, line 54: this double HTTP call is not necessary, the URL is going to be fetched anyway in the render

db/schema.rb
```diff
@@ -454,6 +454,7 @@
     t.datetime "updated_at", null: false
     t.bigint "user_id"
     t.datetime "verified_at"
+    t.index ["user_id"], name: "index_email_authorizations_on_user_id"
```
rhymes, line 457: this looks like it's not supposed to sneak in :D

benhalpern:
This looks great! Yeah, implement @rhymes changes and we should be pretty close with this!

db/schema.rb
cesc1989: oops. Probably after running migrations. What would be a good way to remote this? Do I just remove the line?

app/services/reddit_json_from_url_service.rb
cesc1989: I put the second fetch because some "good looking URLs" might pass the Regexp check but would return 404.

app/services/reddit_json_from_url_service.rb
cesc1989: I did give it a try but a hard time figuring out a good regex 😅  Could you, please give me an idea for this?

A valid Reddit URL usually looks like this:

```
https://www.reddit.com/r/IAmA/comments/afvl2w
```

db/schema.rb
rhymes: yeah, just remove the line :)

app/services/reddit_json_from_url_service.rb
rhymes: the user is still going to notice if the URL leads to a 404, right?

app/services/reddit_json_from_url_service.rb
```diff
@@ -0,0 +1,58 @@
+class RedditJsonFromUrlService
```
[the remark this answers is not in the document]
cesc1989: In the first implementation, I didn't want to split it in two classes but now I see `RedditJsonFromUrlService` doing so much stuff (fetching, digging from hash, parsing, validating).

Would it be fine if I move the `RedditJsonFromUrlService` under `RedditTag::RedditJsonFromUrlService` in the same file?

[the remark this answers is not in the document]
rhymes: I know it's counter intuitive but I think you can just keep the entire logic in the `RedditTag` class, 99% of the tags are written like that. 

I'm not sure having an inner class is much better :)

app/services/reddit_json_from_url_service.rb
rhymes: Yeah, this is though, also because it could also be https://reddit.com/r/IAmA/comments/afvl2w or https://www.reddit.com/r/IAmA/comments/afvl2w/im_scott_from_scotts_cheap_flights_my_profession/ or I guess any combination of the subreddits?

What if we just check for `\Ahttps://(www.)?reddit.com` ? or something like that?

rhymes: Basically: starts with https://reddit.com or https://www.reddit.com

app/services/reddit_json_from_url_service.rb
cesc1989: Good point 😄

cesc1989:
This is so cool 😄 

Thank you @rhymes for guiding me through this PR and everyone from the team who gave me feedback from the very beginning.

PR #8135, opened 2020-05-28
Update vendor/cache After Dependabot

jdoss:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

We need to update vendor/cache after Dependabot does it's bot duties. This GitHub Action should do the business. 

It was adopted from https://gist.github.com/filipkis/fab89a09f54ffdd8e70c063755e7ce7f which was posted to this Dependabot feedback issue: https://github.com/dependabot/feedback/issues/474#issuecomment-583741305

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

This close

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -25,6 +25,8 @@ jobs:
       - name: Bundle install
         run: |
           gem install bundler --no-rdoc --no-ri
+          bundle config set cache_all true
+          bundle config cache_all_platforms true
           bundle install
       - name: Commit and push
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.7.1
+      - name: Installing additional dependencies
+        run: |
+          sudo apt-get install postgresql-client libpq-dev
+      - name: Bundle install
+        run: |
+          gem install bundler --no-rdoc --no-ri
+          bundle install
```
mstruve, line 28: do we need to run `bundle config set cache_all true` before this or is it the default when bundler is installed?

jdoss: Good catch, added `cache_all` and `cache_all_platforms`.

PR #8149, opened 2020-05-29
Add feature flag support to Pages

joshpuetz:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds support to gating access to pages created via `/internal/pages` via feature flags! Adding a feature flag at `/internal/feature_flags` with a name in the format of `page_#{page.slug}` will guard access to that page.

Changes in this PR include:
- Feature flag indicator added to page form
- Flag check in both `PagesController` and `StoriesController`
- Refactored the `FeatureFlag` helper

#7996 - Add Feature Flag support to DEV

Feature flag not found matching slug name:

Pages

Feature flag present matching slug name:
Pages

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

same page

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -4,9 +4,8 @@ class PagesController < ApplicationController
 
   def show
     @page = Page.find_by!(slug: params[:slug])
-    if @page.feature_flag && !FeatureFlag.enabled?(@page.feature_flag, current_user)
-      not_found
-    end
+    not_found unless FeatureFlag.accessible?(@page.feature_flag, current_user)
+
     set_surrogate_key_header "show-page-#{params[:slug]}"
   end
 
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -4,7 +4,7 @@ class PagesController < ApplicationController
 
   def show
     @page = Page.find_by!(slug: params[:slug])
-    not_found unless FeatureFlag.accessible?(@page.feature_flag, current_user)
+    not_found unless FeatureFlag.accessible?(@page.feature_flag_name, current_user)
 
     set_surrogate_key_header "show-page-#{params[:slug]}"
   end
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -4,7 +4,7 @@ class PagesController < ApplicationController
 
   def show
     @page = Page.find_by!(slug: params[:slug])
-    not_found if FeatureFlag.exist?(@page.feature_flag) && !FeatureFlag.enabled?(@page.feature_flag, current_user)
+    not_found unless FeatureFlag.enabled?(@page.feature_flag, current_user)
     set_surrogate_key_header "show-page-#{params[:slug]}"
   end
 
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -4,7 +4,9 @@ class PagesController < ApplicationController
 
   def show
     @page = Page.find_by!(slug: params[:slug])
-    not_found unless FeatureFlag.enabled?(@page.feature_flag, current_user)
+    if @page.feature_flag && !FeatureFlag.enabled?(@page.feature_flag, current_user)
+      not_found
+    end
     set_surrogate_key_header "show-page-#{params[:slug]}"
   end
 
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -4,6 +4,7 @@ class PagesController < ApplicationController
 
   def show
     @page = Page.find_by!(slug: params[:slug])
+    not_found if FeatureFlag.exist?(@page.feature_flag) && !FeatureFlag.enabled?(@page.feature_flag, current_user)
     set_surrogate_key_header "show-page-#{params[:slug]}"
   end
 
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -111,7 +111,7 @@ class StoriesController < ApplicationController
       Honeycomb.add_field("stories_route", "org")
       handle_organization_index
     elsif @page
-      if !@page.feature_flag || FeatureFlag.enabled?(@page.feature_flag, current_user)
+      if FeatureFlag.accessible?(@page.feature_flag, current_user)
         Honeycomb.add_field("stories_route", "page")
         handle_page_display
       else
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -111,11 +111,11 @@ class StoriesController < ApplicationController
       Honeycomb.add_field("stories_route", "org")
       handle_organization_index
     elsif @page
-      if FeatureFlag.exist?(@page.feature_flag) && !FeatureFlag.enabled?(@page.feature_flag, current_user)
-        not_found
-      else
+      if FeatureFlag.enabled?(@page.feature_flag, current_user)
         Honeycomb.add_field("stories_route", "page")
         handle_page_display
+      else
+        not_found
       end
     else
       Honeycomb.add_field("stories_route", "user")
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -111,7 +111,7 @@ class StoriesController < ApplicationController
       Honeycomb.add_field("stories_route", "org")
       handle_organization_index
     elsif @page
-      if FeatureFlag.enabled?(@page.feature_flag, current_user)
+      if !@page.feature_flag || FeatureFlag.enabled?(@page.feature_flag, current_user)
         Honeycomb.add_field("stories_route", "page")
         handle_page_display
       else
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -111,8 +111,12 @@ class StoriesController < ApplicationController
       Honeycomb.add_field("stories_route", "org")
       handle_organization_index
     elsif @page
-      Honeycomb.add_field("stories_route", "page")
-      handle_page_display
+      if FeatureFlag.exist?(@page.feature_flag) && !FeatureFlag.enabled?(@page.feature_flag, current_user)
+        not_found
+      else
+        Honeycomb.add_field("stories_route", "page")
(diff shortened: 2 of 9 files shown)


app/controllers/pages_controller.rb
```diff
@@ -4,6 +4,7 @@ class PagesController < ApplicationController
 
   def show
     @page = Page.find_by!(slug: params[:slug])
+    not_found unless FeatureFlag.enabled?(@page.feature_flag, current_user)
```
citizen428, line 7: This is just an idea, but what if we remove the feature flag name from the form and replace it with a check box that says "Protected by feature flag" or similar? We can then default the flag name to something like `page_#{slug}` which means we could perform this change without having to load a page record from the DB. Potentially we wouldn't even need to add a column to the `pages` table, just a method that returns the name from the slug. 

This would be easier for site admins (only one action instead of two), would save us an extra DB access and they only minor drawback would be that the flag name isn't fully configurable.

app/helpers/feature_flag.rb
```diff
@@ -1,7 +1,15 @@
 module FeatureFlag
   extend self # rubocop:disable Style/ModuleFunction
```
citizen428, line 2: I forgot to comment on this in the original PR: we currently do this in several different ways across the code-base:

1. `extend self`
2. Prefixing every method name with `self.`
3. `class << self`
4. `module_function`

For the sake of consistency, we should probably settle on one and given that the one used here needs a temporary Rubocop override I'd vote for 2. or 3. (unless you need the extra feature of `module_function`, in which case 4. is ok).

app/helpers/feature_flag.rb
```diff
@@ -1,7 +1,15 @@
...
+    Flipper.enabled?(feature_name, *args)
+  end
+
+  def exist?(feature_name)
+    Flipper.exist?(feature_name)
+  end
+
   def enabled?(feature_name, *args)
-    Flipper[feature_name].enabled?(*args)
+    !exist?(feature_name) || strict_enabled?(feature_name, *args)
```
citizen428, line 13: Maybe we can do something like instead:

```ruby
def enabled?(feature_name, *args)
  Flipper.enabled?(feature_name, *args)
end

def exist?(feature_name)
  Flipper.exist?(feature_name)
end

def active?
  # what the current enabled? is trying to do
end
```

If we do this we can also use delegates to wrap the `Flipper` class methods:

```ruby
module FeatureFlag
  class << self
    delegate :enabled?, :exists?, to: Flipper

    def active?
      ...
    end
  end
end 
```

BTW, not sure about returning true for missing features, that seems a bit counterintuitive to me. The only way I'd expect something to not show up is when it's (1) gated AND (2) enabled.

spec/helpers/feature_flag_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      expect(Flipper).to have_received(:exist?).with("foo")
+    end
+  end
+
+  describe ".enabled?" do
+    let(:user) { UserStruct.new(flipper_id: 1) }
+
+    context "when flag doesn't exist" do
+      it "returns true" do
+        expect(described_class.enabled?("missing_flag")).to be_truthy # rubocop:disable Rspec/PredicateMatcher
```
citizen428, line 31: If we return a truthy value for pages where we don't have feature flags defined, wouldn't that disable all existing pages unless we retroactively add flags for them?

app/helpers/feature_flag.rb
joshpuetz, line 13: The default behavior of Flipper is to return false is a flag isn't defined. I've added this helper method to return true if:

feature flag is globally enabled
feature flag is disabled, but enabled for the specified user id
feature flag has not been defined
I need input on this last point: if you were to call FeatureFlag.enabled? for a flag that didn't exist...what would you expect to get?

app/controllers/pages_controller.rb
joshpuetz: I **love** this idea @citizen428! The only downsides are a) can't customize the flag name like you mentioned, and b) it wouldn't work if you had multiple pages you wanted guarded by a single feature flag.

Since I don't have those requirements for Codeland (which is driving this feature), I'm going to go for it!

app/helpers/feature_flag.rb
fdoxyz: Oh I see, that check totally went over my head. 

From that perspective I'd maybe say it could be good to make `FeatureFlag.enabled?` return `false` for a flag that doesn't exist.

My reasoning is that if someone sets a check for a feature_flag with a typo then the page will leak as generally available (regardless if @citizen428 idea of using the slug is implemented or not).

app/helpers/feature_flag.rb
joshpuetz: Good point: I pretty much took this from the definition of `Flipper::DSL`. I'll opt for option 2 :)

app/helpers/feature_flag.rb
joshpuetz: How could I have forgotten about `delegate` 🤦 Thanks, this reads much cleaner.

I'm pretty firm that the name of the class should remain FeatureFlag: a Feature is a much more generic term. I like how specific FeatureFlag is.

joshpuetz:
Oof, good catch @fdoxyz! (https://github.com/thepracticaldev/dev.to/pull/8149#pullrequestreview-421041265)

I think I'll move the null check inside `FeatureFlag.accessible?`: I hate how having null checks pepper throughout the code looks!

app/helpers/feature_flag.rb
joshpuetz: Hmm, that's a fair point. I think I'm covered with this logic for `FeatureFlag.accessible?`

feature flag is globally enabled: **returns true**
feature flag is disabled, but enabled for the specified user id: **returns true**
feature flag specified doesn't exist: **returns false**
nil or empty string passed in: **returns true**

app/controllers/stories_controller.rb
```diff
@@ -111,8 +111,12 @@ def handle_user_or_organization_or_podcast_or_page_index
       Honeycomb.add_field("stories_route", "org")
       handle_organization_index
     elsif @page
-      Honeycomb.add_field("stories_route", "page")
-      handle_page_display
+      if FeatureFlag.enabled?(@page.feature_flag, current_user)
+        Honeycomb.add_field("stories_route", "page")
```
mstruve, line 115: 🙏  Thanks for adding this

db/schema.rb
```diff
@@ -794,6 +794,7 @@
     t.text "body_markdown"
     t.datetime "created_at", null: false
     t.string "description"
+    t.string "feature_flag"
```
mstruve, line 797: Rather than add a column here, it would be nice to keep the feature flags contained to just Flipper, why not use the slug as the flag instead? The slug will ensure uniqueness and is very unlikely to change for the life of the page.

joshpuetz: Feature flagging a page is going to be optional: we don't want to ever _require_ a feature flag: they are meant to be short lived. I'm going to use the slug name as the flag name as @citizen428 suggested  (https://github.com/thepracticaldev/dev.to/pull/8149#discussion_r432276233), but will still need to record the fact this page is guarded SOMEWHERE.

joshpuetz:
@mstruve @citizen428 I hear your concerns about flag data leaking into the database. I've updated the PR to a) not store _anything_ in the database, and b) alert the admin if a feature flag matching `page_#{slug_name}` is present. I've added text and links at the bottom of the page forma to alert the user if the flag is present or not!

app/controllers/pages_controller.rb
citizen428: Sorry, I meant additional reads in the case the page is disabled. If we could check the flag independent of the page object, this read `@page = Page.find_by!(slug: params[:slug]` becomes unncessary in case the site is blocked by the flag. 

Anyway, you're probably right, like I said, it was just an idea 😃

app/helpers/feature_flag.rb
citizen428: > I'm pretty firm that the name of the class should remain FeatureFlag

Sure, let's stick to the name `FeatureFlag`. I sometimes just throw ideas out during reviews, unless I use the "request changes" feature (almost never) it's always meant as a suggestion that can be ignored.

spec/helpers/feature_flag_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      expect(Flipper).to have_received(:exist?).with("foo")
+    end
+  end
+
+  describe ".accessible?" do
+    let(:user) { UserStruct.new(flipper_id: 1) }
+
+    it "returns false when flag doesn't exist" do
+      expect(described_class.accessible?("missing_flag")).to be_truthy # rubocop:disable Rspec/PredicateMatcher
```
citizen428, line 30: No blocker for this PR, but for a future refactor it may be easier to move the `rubocop:disable` to the top and add a `rubocop:enable Rspec/PredicateMatcher` at the bottom of the file as we do in some other files.

https://rubocop.readthedocs.io/en/latest/configuration/#disabling-cops-within-source-code

app/helpers/feature_flag.rb
```diff
@@ -1,7 +1,9 @@
 module FeatureFlag
-  extend self # rubocop:disable Style/ModuleFunction
+  class << self
+    delegate :enabled?, :exist?, to: Flipper
 
-  def enabled?(feature_name, *args)
-    Flipper[feature_name].enabled?(*args)
+    def accessible?(feature_flag_name, *args)
+      feature_flag_name.blank? || !exist?(feature_flag_name) || enabled?(feature_flag_name, *args)
```
citizen428, line 6: Right now we call `accessible?` in `StoriesController`, where we pass in `@page.feature_flag_name`. In `Page` we define this method to return `"page_#{slug}"`. So am I missing something or can this never be `blank?`?

joshpuetz: You're right: the way I'm calling it now it won't be blank, but I want this helper class to be resilient for any kind of use case: I don't want to always assume feature flag name will be passed in!

PR #8150, opened 2020-05-29
Leverage announcement broadcasts

### LINKED ISSUE #8089: Leverage Announcement broadcasts
Differentiate between `Welcome` and `Announcement` broadcasts in Admin UI (`/internal/broadcasts`).

Note: we only want to allow one `active` announcement broadcast at a time (will need to validate this upon create/update).

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
In preparation for a new "site-wide messaging" feature (see https://github.com/thepracticaldev/dev.to/issues/8090), I've started cleaning up some of our admin UI around the `Broadcast` model, and adding some constraints around the Broadcasts with a `type_of: "Announcement"`, which is what we'll use for our site-wide messaging service.

This PR does the following:

- [x] Differentiates between welcome + announcement broadcasts in the admin UI.
- [x] Allow admins to filter between these two types in the `/internal/broadcasts` UI by passing a `type_of` param into the route.
- [x] Only allows one `Announcement` broadcast to be active at a time (AND adds a bunch of tests related to this validation).
- [x] Adds some useful `announcement` and `welcome` scopes on the `Broadcast` class.
- [x] Adds some `flash[:success]` messages in the `BroadcastsController`.

Closes #8089! 🚀 

**To QA this, you will want to do the following:**

1. Check out the `/internal/broadcasts` page! You'll see an improved UI that lets you filter based on the `type_of` on the broadcast:
Screen Shot 2020-05-28 at 6 00 21 PM

2. Check that you can create some inactive announcement broadcasts:
Screen Shot 2020-05-28 at 3 52 19 PM

3. Check that you can set one of those announcement broadcasts to `active`, and that you see a super cute flash message:
Screen Shot 2020-05-28 at 3 28 16 PM
Screen Shot 2020-05-28 at 3 53 06 PM
Screen Shot 2020-05-28 at 3 28 22 PM

4. Check that you CANNOT set another announcement broadcast to `active`, thus ensuring that only _one_ of these types of broadcasts can ever be active at once!
Screen Shot 2020-05-28 at 3 52 42 PM

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/controllers/internal/broadcasts_controller.rb
+++ b/app/controllers/internal/broadcasts_controller.rb
@@ -21,7 +21,11 @@ class Internal::BroadcastsController < Internal::ApplicationController
   end
 
   def index
-    @broadcasts = Broadcast.all
+    @broadcasts = if params[:type_of]
+                    Broadcast.where(type_of: params[:type_of].capitalize)
+                  else
+                    Broadcast.all
+                  end
   end
 
   private
--- a/app/controllers/internal/broadcasts_controller.rb
+++ b/app/controllers/internal/broadcasts_controller.rb
@@ -45,8 +45,4 @@ class Internal::BroadcastsController < Internal::ApplicationController
   def authorize_admin
     authorize Broadcast, :access?, policy_class: InternalPolicy
   end
-
-  def active_announcement_params
-    params[:type_of].capitalize == "Announcement" && params[:active] == true
-  end
 end
--- a/app/controllers/internal/broadcasts_controller.rb
+++ b/app/controllers/internal/broadcasts_controller.rb
@@ -3,13 +3,21 @@ class Internal::BroadcastsController < Internal::ApplicationController
 
   def create
     @broadcast = Broadcast.create!(broadcast_params)
+    flash[:success] = "Broadcast has been created!"
+    redirect_to "/internal/broadcasts"
+  rescue ActiveRecord::RecordInvalid => e
+    flash[:danger] = e.message
     redirect_to "/internal/broadcasts"
   end
 
   def update
     @broadcast = Broadcast.find_by!(id: params[:id])
     @broadcast.update!(broadcast_params)
+    flash[:success] = "Broadcast has been updated!"
     redirect_to "/internal/broadcasts"
+  rescue ActiveRecord::RecordInvalid => e
+    flash[:danger] = e.message
+    redirect_to "/internal/broadcasts/#{params[:id]}/edit"
   end
 
   def new
@@ -36,5 +44,9 @@ class Internal::BroadcastsController < Internal::ApplicationController
 
   def authorize_admin
     authorize Broadcast, :access?, policy_class: InternalPolicy
+  end
+
+  def active_announcement_params
+    params[:type_of].capitalize == "Announcement" && params[:active] == true
   end
 end
--- a/app/controllers/internal/broadcasts_controller.rb
+++ b/app/controllers/internal/broadcasts_controller.rb
@@ -33,7 +33,7 @@ class Internal::BroadcastsController < Internal::ApplicationController
                     Broadcast.where(type_of: params[:type_of].capitalize)
                   else
                     Broadcast.all
-                  end
+                  end.order(title: :asc)
   end
 
   private
--- a/app/models/broadcast.rb
+++ b/app/models/broadcast.rb
@@ -5,10 +5,28 @@ class Broadcast < ApplicationRecord
 
   validates :title, :type_of, :processed_html, presence: true
   validates :type_of, inclusion: { in: %w[Announcement Welcome] }
+  validate  :single_active_announcement_broadcast
 
   scope :active, -> { where(active: true) }
+  scope :announcement, -> { where(type_of: "Announcement") }
+  scope :welcome, -> { where(type_of: "Welcome") }
 
   def get_inner_body(content)
     Nokogiri::HTML(content).at("body").inner_html
+  end
+
+  private
+
+  def single_active_announcement_broadcast
+    # Only add errors if we are trying to modify an announcement
+    # broadcast while another announcement broadcast is already
+    # active to ensure that only one can be active at a time.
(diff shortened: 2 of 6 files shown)


app/models/broadcast.rb
```diff
@@ -5,10 +5,28 @@ class Broadcast < ApplicationRecord
...
+    # broadcast while another announcement broadcast is already
+    # active to ensure that only one can be active at a time.
+    active_broadcasts = Broadcast.announcement.active
+    return unless active_broadcasts.count.positive? &&
+      active &&
+      type_of == "Announcement" &&
+      active_broadcasts.first.id != id
+
+    errors.add(:base, "You can only have one active announcement broadcast")
+  end
```
vaidehijoshi, line 31: @atsmith813 here are the fun timez that became 30 minutes of my afternoon 😛

app/models/broadcast.rb
```diff
@@ -5,10 +5,28 @@ class Broadcast < ApplicationRecord
...
+
+  def single_active_announcement_broadcast
+    # Only add errors if we are trying to modify an announcement
+    # broadcast while another announcement broadcast is already
+    # active to ensure that only one can be active at a time.
+    active_broadcasts = Broadcast.announcement.active
+    return unless active_broadcasts.count.positive? &&
+      active &&
+      type_of == "Announcement" &&
+      active_broadcasts.first.id != id
```
rhymes, line 28: super small optimization since they are in `AND`: if you put the count as the third condition, it's going to be executed only if the previous ones are all true. This way we don't hit the DB with a SQL count if we don't need to

You might also not need the count at all actually, though it's going to make the code less readable, since the last condition loads the first item just to check the `ID`, so instead of going to the DB twice, you can do (untested): 

```ruby
active && type_of == "Announcement" && ![nil, id].include?(active_broadcasts.order(id: :asc).limit(1).pluck(:id).first)
```

a cleaner version could be:

```ruby
first_active_broadcast = active_broadcasts.order(id: :asc).limit(1) # this doesn't execute a query
return unless active && type_of == "Announcement" &&  ![nil, id].include?(first_active_broadcast.pluck(:id).first)
```


which in Rails 6 will become:

```ruby
first_active_broadcast = active_broadcasts.order(id: :asc).limit(1) # this doesn't execute a query
return unless active && type_of == "Announcement" &&  ![nil, id].include?(first_active_broadcast.pick(:id))
```

:-D

app/controllers/internal/broadcasts_controller.rb
```diff
@@ -21,7 +29,11 @@ def edit
   end
 
   def index
-    @broadcasts = Broadcast.all
+    @broadcasts = if params[:type_of]
+                    Broadcast.where(type_of: params[:type_of].capitalize)
```
rhymes, line 33: should these be ordered alphabetically or by descending creation date?

app/models/broadcast.rb
vaidehijoshi: @rhymes Ooof yes I should have known to put the query last to avoid unnecessary optimizations. This is what happens when I try to code in the evenings 🙈

app/controllers/internal/broadcasts_controller.rb
vaidehijoshi: Good call! We don't have `created_at` on these (which is odd, and which I didn't notice til just now!). But for the time being, I am going with ordering alphabetically 👍

vaidehijoshi:
@rhymes Thanks for the suggestions, I've added both in the latest commit! This is why I love working with you: you always teach me new things and make my code better ❤️

spec/requests/internal/broadcasts_spec.rb
```diff
@@ -3,7 +3,7 @@
 
 RSpec.describe "/internal/broadcasts", type: :request do
   let(:get_resource) { get "/internal/broadcasts" }
-  let(:params) { { title: "Hello!", processed_html: "<pHello!</p>", type_of: "Welcome", sent: true } }
+  let(:params) { { title: "Hello!", processed_html: "<pHello!</p>", type_of: "Welcome", active: true } }
```
juliannatetreault, line 6: I spy a missing closing `>` here! 👀

spec/requests/internal/broadcasts_spec.rb
```diff
@@ -87,4 +87,29 @@
       end
     end
   end
+
+  context "with type_of Announcement" do
+    let(:super_admin) { create(:user, :super_admin) }
+    let(:params) { { title: "Hello!", processed_html: "<pHello!</p>", type_of: "Announcement", active: true } }
```
juliannatetreault, line 93: Same here: there is a missing `>` on the `p` tag.

spec/requests/internal/broadcasts_spec.rb
vaidehijoshi: Ah, this was pre-existing! But I have pushed up the fix.

spec/requests/internal/broadcasts_spec.rb
vaidehijoshi: This is what happens when you copypasta pre-existing code and don't read it closely enough 🙊 Fixed now.

vaidehijoshi:
@jacobherrington Womp womp. 😞 Restarted them! 🤞

app/controllers/internal/broadcasts_controller.rb
rhymes: Ooh! that's weird. We should definitely send a PR to add those fields ahhaha. I'm suprised Rails works without the two timestamps at all.

TIL!

rhymes: https://github.com/thepracticaldev/dev.to/pull/8181

PR #8152, opened 2020-05-29
Improve development experience of custom HTML Pages

fdoxyz:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Not entirely sure if this is something we want, but I thought it may be useful for the team going forward 🙂

**Context:** Local development of a `Page` (usually ahead of campaigns) has the potential to be a little more developer friendly. The two options that I see at the moment are:
- Serve the HTML file separate from Rails and iterate on it. Later on include it in a `Page` to check how it looks inside the site's shell
- With each change to the HTML you can update the page in `/internal/pages` dashboard and refresh the Page's URL to check how it looks. The iteration here feels a bit slow (lots of copy & paste)

This PR introduces the ability to use the designated partial `app/views/pages/_wip_page.html.erb` to make the process of working on a custom HTML page a bit quicker/easier.
- This feature will be invisible in production (only available when working locally)
- The process now becomes:
  1. Create the Page in `/internal/pages` with the "Use partial" option enabled
  1. Start working on the page by editing `app/views/pages/_wip_page.html.erb` and iterate by refreshing the page locally
  1. When finished working on the page: Take the contents of the partial and save them as the HTML body of the Page

**New section**
(it won't be visible in production)
Screen Shot 2020-05-29 at 11 36 55

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

- DB Migration

Is this useful?
maybe meme


fdoxyz:
Brakeman is not happy with the "Unescaped model attribute" being rendered, which makes sense. 

This is always a problem when rendering plain HTML even from `@page.processed_html` in the DB. The local file workaround is restricted to not work in production, I wonder if there is a way to whitelist this.

Anyways, open to hearing opinions about the PR in general first

app/models/page.rb
```diff
@@ -18,6 +18,16 @@ def path
     is_top_level_path ? "/#{slug}" : "/page/#{slug}"
   end
 
+  def local_html?
+    !Rails.env.production? && File.file?(local_path.to_s)
+  end
+
+  def html
+    return File.read(local_path) if local_html?
```
joshpuetz, line 26: Instead of saving paths to the database, since this is just for local development could we just have reading a local file default to a directory in the project? For example: if the page is named `codeland`, just have the local option look for a file at `<project>/tmp/local/codeland.html` (or a different folder added to `.gitignore`)? That way NOTHING leaks into production!

app/views/internal/pages/_form.html.erb
```diff
@@ -17,6 +17,13 @@
         <%= form.label :body_html %> (Only if not using markdown. HTML is dangerous ⚠️)
         <%= form.text_area :body_html, class: "form-control" %>
       </div>
+      <% unless Rails.env.production? %>
+        <div class="form-group">
+          <%= form.label :local_path %> (Use for local development only: <strong>This will not be available in production</strong>. Be careful! ⚠️)
+          <%= form.text_field :local_path, class: "form-control" %>
```
joshpuetz, line 23: If we just defaulted paths, we wouldn't have to store anything in the database!

app/models/page.rb
fdoxyz: I very much see the value of this! I hadn't thought of a good default and also at some point also considered having an ENV variable that can be set for this purpose.

You just sparked an idea in me for this though 🙏🏼  I'll push some changes soon!

fdoxyz:
@joshpuetz I just pushed out the changes. The PR now has a partial designated for this local development, which can only be enabled in development. No paths can be leaked to the production DB and this is also a way less sketchy way to render compared to `File.read('file_path').html_safe`

I updated the PR original comment to explains the entire process too so it's clear. Let me know what you think of this approach 🙂

app/views/pages/_wip_page.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="content">
```
joshpuetz, line 1: Do we perhaps want to add this page to `.gitignore` so changes to it don't get written to Github? I like the idea of having a sample page in source control, but I'm concerned about constant updates to it. Perhaps something like `sample_application.yml` (we tell users to copy that to `application.yml`, and that file is in *.gitignore)

db/schema.rb
```diff
@@ -801,6 +802,7 @@
     t.string "template"
     t.string "title"
     t.datetime "updated_at", null: false
+    t.boolean "use_partial", default: false
```
joshpuetz, line 805: I bet everyone's gonna hate this 😆 

Since this is just for development, perhaps we could just use a partial if one if found at a location like `app/views/pages/#{page.slug}.html.erb`?

app/views/internal/pages/_form.html.erb
```diff
@@ -38,6 +38,15 @@
         <%= form.check_box :is_top_level_path %>
```
joshpuetz, line 38: This feature doesn't work with the `top level path` option: when I turn on `WIP` and `Top level path`, visiting `/<page_name` gives me an error

db/schema.rb
fdoxyz: I can see the point and its a very valid one, to make sure nothing gets into the DB that doesn't really need to be there. Also that column name is just looks terrible 🙃

Automatically recognizing the presence of a partial that matches the slug sounds interesting (the type of "magic" that comes out of the box in Rails). I can give this approach a try to see how it feels.

joshpuetz: I did something similar here, just having a method construct the name instead of storing it in the db: https://github.com/thepracticaldev/dev.to/pull/8149/files#diff-8135377fbb811479614b50a9eca1d721R21-R23

fdoxyz: Noted! Thanks for the reference, definitely useful 😄

fdoxyz:
@mstruve I see the value of pulling the pages out of the codebase. This was an attempt to improve the productivity of the development environment when coding new custom HTML pages. The intention at all times was to not make this something that could be used in production (i.e. add partials for Pages).

If the approach is not on point and aligned with the bigger picture of Pages I absolutely understand that too. If I can't figure out an enhancement that fits this criteria I'm totally okay with stashing the idea for the moment. Thank you for the insight 👍🏼

mstruve:
> The intention at all times was to not make this something that could be used in production 

@fdoxyz its not a production issue, its a generalization issue. This codebase is soon to become Forem which means anything community specific like pages should not be in it. 

I am all for making the development experience better but in this case I think we should try something different. Maybe write a seed script that someone could use to make a page in their local console and we could add it to the docs for people to use.

fdoxyz:
I have nothing to refute, even the GIF I chose says it was a "maybe" idea 😅 I'm closing this now.

Thank you also for the seeds/script suggestion @mstruve! I'll keep it in mind in case I/we try something else in the future.

PR #8153, opened 2020-05-29
Make db:seed idempotent

citizen428:

- [X] Refactor

## Description

Yesterday @jdoss mentioned to me that from a systems engineering perspective it would be good for the `db:seed` task to be idempotent. I added a little helper for that purpose, which should also come in handy during development, since it will selectively seed new and/or missing entities.

This is what it looks like in action:

```
$ rails db:seed
Seeding with multiplication factor: 1

  1. Creating Organizations.
  2. Users already exist. Skipping.
  3. Tags already exist. Skipping.
  4. Creating 25 Articles.
  5. Comments already exist. Skipping.
  6. Creating Podcasts.
  7. Broadcasts already exist. Skipping.
  8. ChatChannels already exist. Skipping.
  9. Creating HtmlVariants.
  10. Badges already exist. Skipping.
  11. FeedbackMessages already exist. Skipping.
  12. ListingCategories already exist. Skipping.
  13. ListingCategories already exist. Skipping.
  14. Creating Pages.

   [DEV ASCII art snipped]
```

n/a

## Added tests?

- [X] no, because they aren't needed. At least I don't think so, we never had tests around this.

## Added to documentation?

- [X] no documentation needed

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,128 +1,146 @@
+# rubocop:disable Rails/Output
+
+# NOTE: when adding new data, please use this class to ensure the seed tasks
+# stays idempotent.
+class Seeder
+  def initialize
+    @counter = 0
+  end
+
+  def create_if_none(klass, count = nil)
+    @counter += 1
+    plural = klass.name.pluralize
+
+    if klass.none?
+      message = ["Creating", count, plural].compact.join(" ")
+      puts "  #{@counter}. #{message}."
+      yield
+    else
+      puts "  #{@counter}. #{plural} already exist. Skipping."
+    end
+  end
+end
+
 # we use this to be able to increase the size of the seeded DB at will
 # eg.: `SEEDS_MULTIPLIER=2 rails db:seed` would double the amount of data
+seeder = Seeder.new
 SEEDS_MULTIPLIER = [1, ENV["SEEDS_MULTIPLIER"].to_i].max
-counter = 0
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -1,128 +1,146 @@
+# rubocop:disable Rails/Output
```
citizen428, line 1: We already had this for the ASCII art output, I just moved it to the top.

db/seeds.rb
```diff
@@ -1,128 +1,146 @@
...
+    @counter = 0
+  end
+
+  def create_if_none(klass, count = nil)
+    @counter += 1
+    plural = klass.name.pluralize
+
+    if klass.none?
+      message = ["Creating", count, plural].compact.join(" ")
+      puts "  #{@counter}. #{message}."
```
citizen428, line 16: I thought displaying the output could be nicer than using the Rails logger, so people know exactly what changed (they may miss the logs). I can change it back though if people prefer that.

db/seeds.rb
```diff
@@ -1,128 +1,146 @@
+# rubocop:disable Rails/Output
+
+# NOTE: when adding new data, please use this class to ensure the seed tasks
+# stays idempotent.
+class Seeder
```
citizen428, line 5: Just a minimal helper class to wrap up the creation logic.

PR #8160, opened 2020-05-29
Enqueue feed_customization_notification

### LINKED ISSUE #8069: Properly Enqueue send_feed_customization_notification
Currently, we check whether an individual notification is enqueued or not in `generator.rb`. All notifications should set `@notification_enqueued` equal to `true` within their individual "send X notification" method, but `send_feed_customization_notification` does not - this needs to be fixed. 

juliannatetreault:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR properly enqueues the `customize_feed_notification` within the `WelcomeNotification::Generator` in addition to removing an extra word from the `generator_spec.rb`. 

After a _ton_ of debugging yesterday, it is safe to say that no additional specs are needed for this addition and that the current spec will fail properly if `@notification_enqueued = true` is removed specifically from line 34 of the generator - the only place in the generator where there is a potential enqueuing-related bug. 

The potential bug is caused by the removal of `@notification_enqueued = true` for the _first_ Welcome Notification--if that line is removed, a User could potentially receive the first _and_ second Welcome Notification if the Notification is not properly enqueued, but this would only happen if someone intentionally removed this line from the generator. 

TL;DR: Please refrain from removing `@notification_enqueued = true` from line 34. 😂 

Closes #8069 

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

rubber_duck

### DIFF
--- a/app/services/broadcasts/welcome_notification/generator.rb
+++ b/app/services/broadcasts/welcome_notification/generator.rb
@@ -31,7 +31,7 @@ module Broadcasts
         return if received_notification?(welcome_broadcast) || commented_on_welcome_thread? || user.created_at > 3.hours.ago
 
         Notification.send_welcome_notification(user.id, welcome_broadcast.id)
-        # Please do not remove the @notification_enqueued = true here - it is necessary for the notifications to run in the proper order
+        # Setting @notification_enqueued here prevents us from sending a user two welcome notifications in one day.
         @notification_enqueued = true
       end
 
--- a/app/services/broadcasts/welcome_notification/generator.rb
+++ b/app/services/broadcasts/welcome_notification/generator.rb
@@ -31,6 +31,7 @@ module Broadcasts
         return if received_notification?(welcome_broadcast) || commented_on_welcome_thread? || user.created_at > 3.hours.ago
 
         Notification.send_welcome_notification(user.id, welcome_broadcast.id)
+        # Please do not remove the @notification_enqueued = true here - it is necessary for the notifications to run in the proper order
         @notification_enqueued = true
       end
 
--- a/app/services/broadcasts/welcome_notification/generator.rb
+++ b/app/services/broadcasts/welcome_notification/generator.rb
@@ -45,6 +45,7 @@ module Broadcasts
         return if user_is_following_tags? || received_notification?(customize_feed_broadcast) || user.created_at > 3.days.ago
 
         Notification.send_welcome_notification(user.id, customize_feed_broadcast.id)
+        @notification_enqueued = true
       end
 
       def send_ux_customization_notification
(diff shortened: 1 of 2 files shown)


vaidehijoshi:
I think we might want to include a short comment about which notification actually is the one that needs the enqueued flag; chances are high that someone will remove that line and not look at this PR and its description, so it will probably be more future-proof if we provide context in the code itself!

Also...based on your PR description, seems like we might not need all the other `notification_enqueued` flags in the _other_ notifications; am I understanding that right? Should we just remove the ones that we don't need?

cc-ing @maestromac here in case I am missing something about why we added those in the first place.

juliannatetreault:
> I think we might want to include a short comment about which notification actually is the one that needs the enqueued flag; chances are high that someone will remove that line and not look at this PR and its description, so it will probably be more future-proof if we provide context in the code itself!
> 
> Also...based on your PR description, seems like we might not need all the other `notification_enqueued` flags in the _other_ notifications; am I understanding that right? Should we just remove the ones that we don't need?
> 
> cc-ing @maestromac here in case I am missing something about why we added those in the first place.

Done! 

From what @atsmith813 and I could tell, the `@notification_enqueued = true` is not necessary for all of the methods, as there don't seem to be situations in which a User could receive multiple notifications. That being said though, I think that the removal of `@notification_enqueued = true` should be part of a larger refactor if we decide to remove them.

app/services/broadcasts/welcome_notification/generator.rb
```diff
@@ -31,6 +31,7 @@ def send_welcome_notification
         return if received_notification?(welcome_broadcast) || commented_on_welcome_thread? || user.created_at > 3.hours.ago
 
         Notification.send_welcome_notification(user.id, welcome_broadcast.id)
+        # Please do not remove the @notification_enqueued = true here - it is necessary for the notifications to run in the proper order
```
vaidehijoshi, line 34:
```suggestion
        # Setting @notification_enqueued here prevents us from sending a user two welcome notifications in one day.
```

It's not so much that the flag here is running the notifications in a specific order, it's preventing it from sending two in one day.

juliannatetreault:
> Looks like `app/services/broadcasts/welcome_notification/generator.rb` could be refactored a bit if we decide it is worth the time!
> 
> Otherwise this looks good to me. 👍

I agree @jacobherrington ! I'm likely going to add a cool-down card for it. 😄

vaidehijoshi:
@jacobherrington @juliannatetreault FWIW, the generator will probably be refactored when [this card](https://github.com/thepracticaldev/dev.to/projects/7#card-38686701) is tackled (hopefully) later on in the cycle.

PR #8161, opened 2020-05-29
Add path_redirects table and internal view

atsmith813:
- [x] Feature

## Description
**PR Summary**
This PR creates the backend for `path_redirects` (migration, model, and updated seed file) and creates a new area on internal for admins to access it appropriately.

I included a `version` column so we can keep an eye on many times we're updating a re-direct. Constantly updating a URL with a new `301` destination hurts SEO, so we want to be mindful of that.

The `source` column is so we know the origin of the `new_path` field. It can be one of two things:
1. `"service"` - the `new_path` was determined programmatically.
2. `"admin"` - the `new_path` was set by an admin in the internal panel.

**Steps to test**
1. Re-seed your DB or create `PathRedirect` records manually.
2. [Make sure your user has admin access.](https://docs.dev.to/internal/#internal-guide)
3. Fire-up the app and visit [http://localhost:3000/internal/path_redirects](http://localhost:3000/internal/path_redirects).
4. View, update, create, and destroy `PathRedirects`. Don't forget to try and break it (i.e. submit blank data in the form to make sure it throws an error 😎 .

**Admin usage**
Admins can:
- Create `PathRedirects`.
- Edit only the `new_path` of an existing `PathRedirect`.
- Destroy `PathRedirects`.
- View `PathRedirects`.

The version number is automatically incremented anytime the `new_path` field is updated. The `source` field is automatically set to `"admin"` for any action made on a record via the internal/admin panel. 

And lastly, the version number is color-coded:
- The color is green when the version is <= 3.
- The color is yellow when the version is <= 5.
- The color is red when the version is > 5.

**Project Summary**
_I'll be copy and pasting this summary on each PR related to this project_

We are working on adding better handling for 404 pages to improve SEO. To start, we're focusing on Articles. When an Article is unpublished, we'll add logic to store a new place to redirect to. Then, if a user tries to visit an Article that is going to result in a `404` response, we'll check if we have a defined redirect and do that accordingly.
- [x] [Create a redirects table to hold redirects and expose it to admins](https://github.com/thepracticaldev/dev.to/issues/8079)
- [ ] [Create an `ArticleRedirectService`](https://github.com/thepracticaldev/dev.to/issues/8080)
- [ ] [Create a Sidekiq worker to call `ArticleRedirectService` when an Article is unpublished](https://github.com/thepracticaldev/dev.to/issues/8081)
- [ ] [Create a controller concern and handle Article redirects](https://github.com/thepracticaldev/dev.to/issues/8083)
- [ ] [Log redirects to Datadog](https://github.com/thepracticaldev/dev.to/issues/8082)
- [ ] [Create a JS redirect notice](https://github.com/thepracticaldev/dev.to/issues/8084)

https://github.com/thepracticaldev/dev.to/issues/8079

## Mobile & Desktop Screenshots/Recordings
**The index/show view**
- An admin can search by `old_path` or `new_path`.
- Clicking on the `ID` takes the admin to the edit view. Clicking on `old_path` or `new_path` will take you to the URL (`old_path` should always redirect to `new_path` once this feature is complete - a quick way for an admin to test).
- Clicking on the `Date` column header will sort the rows by `Date`.
index_show_view

**The create view**
create_view

**The edit view**
edit_view

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed
**I'll add documentation on the final PR for this feature**

dancing_air_traffic_controller


db/seeds.rb
```diff
@@ -326,6 +326,8 @@ def create_if_none(klass, count = nil)
 ##############################################################################
 
 seeder.create_if_none(Badge) do
+  users_in_random_order = User.order(Arel.sql("RANDOM()"))
```
atsmith813, line 329: I added this line to fix the seed file. I kept getting an error at:
https://github.com/thepracticaldev/dev.to/blob/e13b3de5865ff9f3557f819a66f7ea7bcb70c3ed/db/seeds.rb#L339
With:
```
undefined method limit for NilClass
```

db/seeds.rb
```diff
@@ -470,6 +472,21 @@ def create_if_none(klass, count = nil)
...
+
+  articles_for_old_paths.each_with_index do |old_article, i|
+    new_article = articles_for_new_paths[i]
+    PageRedirect.create!(old_path: old_article.path, new_path: new_article.path)
+  end
+end
+
+##############################################################################
+
+# rubocop:disable Rails/Output
```
atsmith813, line 489: I'm having second thoughts about including this in the seed file. I could see it being confusing in development when a couple of random articles always redirect. Any thoughts?

app/views/internal/shared/_navbar.html.erb
```diff
@@ -1,9 +1,34 @@
...
+     tools
+     users
+     welcome
+   ].each_with_object({}) { |v, h| h[v] = v } %>
+
+<% menu_items["webhooks"] = "webhook_endpoints" %>
 
 <div class="mt-4">
-  <% menu_items.each do |name, controller| %>
+  <% menu_items.sort_by { |name, _controller| name }.each do |name, controller| %>
```
atsmith813, line 31: This sorts the menu items so they render alphabetically - more user friendly.

mstruve:
What is the plan for implementing this? Will we catch 404s and then check for a redirect path in an attempt to serve some sort of content?

atsmith813:
@mstruve yep - spot on! To start it will be only on articles. If it's going to end up resulting in a `404`, we'll check if we have an assigned redirect to serve up instead.

app/models/page_redirect.rb
```diff
@@ -0,0 +1,25 @@
+class PageRedirect < ApplicationRecord
+  SOURCES = %w[admin service].freeze
+
+  validates :old_path, presence: true, uniqueness: true
```
rhymes, line 4: the uniqueness index here doesn't correspond to the uniqueness index built in the migration (well, depends on which of the unique indexes you decide to keep)

app/controllers/internal/page_redirects_controller.rb
```diff
@@ -0,0 +1,62 @@
+class Internal::PageRedirectsController < Internal::ApplicationController
```
rhymes, line 1: should we add auditing support here? 

See https://github.com/thepracticaldev/dev.to/pull/6797/files and https://github.com/thepracticaldev/dev.to/blob/master/app/controllers/internal/articles_controller.rb

app/controllers/internal/page_redirects_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+    if @page_redirect.update(edit_page_redirect_params)
+      flash[:success] = "Page Redirect updated successfully!"
+      redirect_to edit_internal_page_redirect_path(@page_redirect)
+    else
+      flash[:danger] = @page_redirect.errors.full_messages.to_sentence
+      render :edit
+    end
+  end
+
+  def destroy
```
rhymes, line 39: Won't removing page redirects have an impact on SEO?

I wonder if this model should have soft delete, so instead of destroying redirects, we should `enable`/`disable` them

app/views/internal/page_redirects/edit.html.erb
```diff
@@ -0,0 +1,33 @@
...
+
+<h2>Page Redirect - ID:  <%= @page_redirect.id %></h2>
+<% badge_type =
+     if @page_redirect.version <= 3
+       "success"
+     elsif @page_redirect.version <= 5
+       "warning"
+     else
+       "danger"
+     end %>
```
rhymes, line 11: ? What does this do?

app/views/internal/shared/_navbar.html.erb
```diff
@@ -1,9 +1,34 @@
-<% menu_items = %w[comments articles users tags welcome broadcasts reports response_templates pages tools chat_channels permissions growth mods config events badges organizations sponsorships podcasts negative_reactions moderator_actions].sort.each_with_object({}) { |v, h| h[v] = v } %>
-<% menu_items[:listings] = "listings" %>
-<% menu_items[:webhooks] = "webhook_endpoints" %>
+<% menu_items = %w[
```
rhymes, line 1: maybe we should freeze this list in the `Internal::ApplicationController` once and for all :D

db/migrate/20200527161505_create_page_redirect.rb
```diff
@@ -0,0 +1,17 @@
...
+      t.string :old_path, null: false
+      t.string :new_path, null: false
+      t.string :source, null: false, default: "service"
+      t.integer :version, null: false, default: 0
+      t.timestamps null: false
+    end
+
+    add_index :page_redirects, :old_path, unique: true
+    add_index :page_redirects, :new_path
+    add_index :page_redirects, %i[old_path new_path], unique: true
```
rhymes, line 13: I don't think we need this. `old_path` is either unique on the entire table, or unique in the context of `new_path`. Don't know which should be, depends on your intentions.

db/seeds.rb
rhymes: this was fixed in https://github.com/thepracticaldev/dev.to/pull/8192/ I believe

spec/models/page_redirect_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+RSpec.describe PageRedirect, type: :model do
+  subject { build(:page_redirect) }
+
+  describe "validations" do
+    it { is_expected.to validate_uniqueness_of(:old_path) }
+    it { is_expected.to validate_presence_of(:old_path) }
+    it { is_expected.to validate_presence_of(:new_path) }
+    it { is_expected.to validate_presence_of(:source) }
+    it { is_expected.to validate_inclusion_of(:source).in_array(described_class::SOURCES) }
```
rhymes, line 11: the trick with this is that it will always be true no matter how many times we change `SOURCES`. I tend to keep the names hardcoded here, so that these checks will intentionally fail if I didn't mean to change it or forgot to update them

atsmith813:
> Nice work!
> 
> I have an observation about naming. Is this model directly related to the `Page` model? If not, as it doesn't seem to be, should we just call this `Redirect` and its table `redirects`?
> 
> ps. are you going to use a gem like https://github.com/jtrupiano/rack-rewrite for the actual redirects?

@rhymes It's not related to pages. I was hesitant to call it just `Redirect` because I fear that would conflict at some point in the future being such a common namespace (i.e. with another gem).

For example, I could see someone doing
```ruby
redirect = Redirect.find_by(old_path: '/some-old-path')`
redirect_to = redirect.new_path
```
 in a controller which could start leading to issues. I felt like by naming the table `redirects` it forces our future selves to have to be more careful about how we name the variables. I'm happy to change it if we feel that's not a big deal!

And no, I don't plan using a gem because where/when we redirect will vary, and sometimes it will require further conditions so we wouldn't want to redirect as high up as the middleware.

For example, looking at Articles, the redirect logic will be called if we are about to 404 [after all checks](https://github.com/thepracticaldev/dev.to/blob/38634e5316140d82c6872673ac3c317034afce82/app/controllers/stories_controller.rb#L45):
1) Check if there's a valid `PageRedirect` record.
2) If not, redirect to the user's dashboard if they're signed in.
3) Otherwise, redirect to the home page.

This could start to vary by object type. So for now, I wasn't planning on moving this to middleware.

app/models/page_redirect.rb
atsmith813: I believe this corresponds to https://github.com/thepracticaldev/dev.to/pull/8161/files#diff-1acd2e7e27a227829d5d14a91c863bb6R784 which I plan to keep.

app/controllers/internal/page_redirects_controller.rb
atsmith813: I didn't even know this was a thing 🙈 . At a glance, it seems like that makes sense. I'll have to look at the documentation and see what that's all about 💯.

app/controllers/internal/page_redirects_controller.rb
atsmith813: That's a good question, it would. But it should be pretty rare that we would need to do this. Whether it's destroyed or disabled - the effect would be the same in terms of SEO. So it's more about the UX.

My original thinking was an `enabled/disabled` flag adds one more layer of complexity (a redirect can exist, but it's inactive). It could open up the mistake of an admin not realizing something was set incorrectly.

I could honestly go either way here. If we think it's better, I can add the flag in there.

app/views/internal/page_redirects/edit.html.erb
atsmith813:
```erb
<% badge_type =
   if @page_redirect.version <= 3
     "success"
   elsif @page_redirect.version <= 5
     "warning"
   else
     "danger"
   end %>
```
```erb
<h4 class="d-inline-block mb-5">
  Version:
  <span class="align-top badge badge-<%= badge_type %>"><%= @page_redirect.version %></span>
</h4>
```

It sets the `badge_type` so the version number displays with the appropriate bootstrap class. For example, a badge with a class of `badge-success` displays as the green/success badge.

edit_view

app/views/internal/shared/_navbar.html.erb
atsmith813: Good idea - done!

db/migrate/20200527161505_create_page_redirect.rb
atsmith813: My thinking here was for when we run a query such as:
```ruby
Redirect.find_by(old_path: "/some-old-path", new_path: "/some-new_path")
```
Which we aren't doing yet. And maybe that index wouldn't even help with that?

db/seeds.rb
atsmith813: Forced push with master to get that update, thanks!

spec/models/page_redirect_spec.rb
atsmith813: That's fair. This spec is really just making sure it's something we validate, so it would fail if someone removed the validation from the model. So we could add another spec to test the options or take your suggestion. I took your suggestion 💯

rhymes:
@atsmith813 thanks for the explanation, makes sense, maybe `PathRedirect` then?

atsmith813:
@rhymes I like that! Will update 👍

app/controllers/internal/page_redirects_controller.rb
atsmith813: Added 👍

app/models/page_redirect.rb
rhymes: Sure, but the migration file has 2 unique indexes, I think the one that's scoped to two columns is redundant

app/controllers/internal/page_redirects_controller.rb
rhymes: I agree with you, better not to have them in the DB, let's destroy! :D

db/migrate/20200527161505_create_page_redirect.rb
rhymes: Ah got it! Ok, so the third one is definitely redundant then. Because PostgreSQL is smart enough to use the two indexes you already have on line 11 and 12. 

Two options:

- one composite unique index on both old_path and new_path (but this would not make old_path unique per se, just in relation of new path)
- the two indexes without the third

atsmith813: Makes sense - I forgot that PostgreSQL knows to use the first 2! Deleted the third index 💯

db/schema.rb
```diff
@@ -268,10 +268,10 @@
   create_table "broadcasts", id: :serial, force: :cascade do |t|
     t.boolean "active", default: false
     t.text "body_markdown"
-    t.datetime "created_at"
     t.text "processed_html"
     t.string "title"
     t.string "type_of"
+    t.datetime "created_at"
```
atsmith813, line 274: I'm not exactly sure why this moved for me 🤔

app/controllers/internal/path_redirects_controller.rb
```diff
@@ -0,0 +1,66 @@
+class Internal::PathRedirectsController < Internal::ApplicationController
+  layout "internal"
+
+  before_action :set_path_redirect, only: %i[edit update destroy]
+
+  after_action only: %i[update destroy create] do
+    Audit::Logger.log(:moderator, current_user, params.dup)
```
juliannatetreault, line 7: I've never seen this before! What does `Audit::Logger` do here? 👀

app/controllers/internal/path_redirects_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def new
+    @path_redirect = PathRedirect.new
+  end
+
+  def create
+    @path_redirect = PathRedirect.new(new_path_redirect_params)
+
+    if @path_redirect.save
+      flash[:success] = "Path Redirect created successfully!"
```
juliannatetreault, line 18: 🎉

app/controllers/internal/path_redirects_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      render :edit
+    end
+  end
+
+  def destroy
+    if @path_redirect.destroy
+      flash[:success] = "Path Redirect destroyed successfully!"
+      redirect_to internal_path_redirects_path
+    else # This should never be the case
+      flash[:danger] = @path_redirect.errors.full_messages.to_sentence
```
juliannatetreault, line 48: TIL `.to_sentence` turns the error message array into a readable sentence! Why am I just learning this now!? 😂

app/controllers/internal/path_redirects_controller.rb
atsmith813: I hadn't either - it's also my first time! We are recording moderator/admin actions to the DB so we can keep track of changes - https://github.com/thepracticaldev/dev.to/pull/6797.

So here it's recording anytime an admin creates, updates, or destroys a `PageRedirect`. When debugging, we'll be able to see the admin action for a given redirect and which admin made the changes.

Screen Shot 2020-06-01 at 4 46 19 PM

app/views/internal/path_redirects/index.html.erb
```diff
@@ -0,0 +1,56 @@
...
+    <% @path_redirects.each do |path_redirect| %>
+      <% badge_type =
+           if path_redirect.version <= 3
+             "success"
+           elsif path_redirect.version <= 5
+             "warning"
+           else
+             "danger"
+           end %>
+
```
juliannatetreault, line 41: Do we need this whitespace here? 👀

app/controllers/internal/path_redirects_controller.rb
atsmith813: On second thought...thanks for drawing attention to this! I should change this from `:moderator` to `:internal` - this not an action that can be performed by moderators. Updated!

app/views/internal/path_redirects/index.html.erb
atsmith813: Good catch! Technically, we don't. It doesn't show up in the markup and my thinking was it's a little more readable since the lines above it are using indentation heavily. It's harder to catch the closing `%>` on line 40 when it's up against the next block of code - especially since the opening `<%` is several lines above that.

app/controllers/internal/path_redirects_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      flash[:success] = "Path Redirect created successfully!"
+      redirect_to internal_path_redirects_path
+    else
+      flash[:danger] = @path_redirect.errors.full_messages.to_sentence
+      render new_internal_path_redirect_path
+    end
+  end
+
+  def index
+    @q = PathRedirect.order(created_at: :desc).ransack(params[:q])
```
vaidehijoshi, line 27: Is `@q` an elasticsearch convention?

app/models/path_redirect.rb
```diff
@@ -0,0 +1,25 @@
+class PathRedirect < ApplicationRecord
+  SOURCES = %w[admin service].freeze
+
+  validates :old_path, presence: true, uniqueness: true
```
vaidehijoshi, line 4: good call on the `uniqueness` here!

app/models/path_redirect.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def new_path_url
+    URL.url(new_path)
+  end
+
+  private
+
+  def increment_version
+    self.version += 1
```
vaidehijoshi, line 23: The `self` is redundant here, no?
```suggestion
    version += 1
```

app/views/internal/path_redirects/edit.html.erb
```diff
@@ -0,0 +1,33 @@
...
+       "success"
+     elsif @path_redirect.version <= 5
+       "warning"
+     else
+       "danger"
+     end %>
+<h4 class="d-inline-block mb-5">
+  Version:
+  <span class="align-top badge badge-<%= badge_type %>"><%= @path_redirect.version %></span>
+</h4>
```
vaidehijoshi, line 15: May want to include a note (warning?) here about not updating paths once you hit a certain version?

app/controllers/internal/application_controller.rb
```diff
@@ -2,6 +2,34 @@ class Internal::ApplicationController < ApplicationController
   before_action :authorize_admin
   after_action :verify_authorized
 
+  MENU_ITEMS = [
```
vaidehijoshi, line 5: i think it would be nice to either add a comment or tweak the const name here; at first glance, i was confused why this was here and it wasn't until i got to [the navbar](https://github.com/thepracticaldev/dev.to/pull/8161/files#diff-0b025b48805ca7b6093e4b356b4a723aR2) that i figured out what the deal was. especially would be useful i think because you don't usually see this kind of large frozen object in a controller

vaidehijoshi: another thought: can we put this in an instance variable that is then accessible by the view and can be at the bottom of the controller? thinking that this could be an option because it looks like this object is used solely for display purposes in the navbar...

db/schema.rb
vaidehijoshi: My guess is it's related to https://github.com/thepracticaldev/dev.to/pull/8181 ?

db/seeds.rb
```diff
@@ -470,6 +470,21 @@ def create_if_none(klass, count = nil)
 
 ##############################################################################
 
+num_path_redirects = 2 * SEEDS_MULTIPLIER
+
+seeder.create_if_none(PathRedirect, num_path_redirects) do
+  articles_for_old_paths = Article.where(published: true).order(Arel.sql("RANDOM()")).limit(num_path_redirects)
```
vaidehijoshi, line 476: wow, arel! #throwback

db/migrate/20200527161505_create_path_redirect.rb
```diff
@@ -0,0 +1,16 @@
+class CreatePathRedirect < ActiveRecord::Migration[5.2]
+  def change
+    create_table :path_redirects do |t|
+      t.string :old_path, null: false
+      t.string :new_path, null: false
+      t.string :source, null: false, default: "service"
```
vaidehijoshi, line 6: > The source column is so we know the origin of the new_path field. It can be one of two things:
> "service" - the new_path was determined programmatically.
> "admin" - the new_path was set by an admin in the internal panel.

i have a question about this: are we doing to be programmatically creating path_redirects? how is that going to work?

the reason i ask is because it seems to me that we actually want `"service`" as the default here, but i am guessing i am missing something about how this will work eventually since you made it the default...

app/views/internal/path_redirects/index.html.erb
```diff
@@ -0,0 +1,56 @@
+<div class="row m-3">
+  <div class="col">
+    <a href="<%= new_internal_path_redirect_path %>" class="btn btn-primary" role="button">Create a New Path Redirect</a>
+  </div>
+  <div class="col">
+    <%= search_form_for @q, url: internal_path_redirects_path, class: "form-inline justify-content-end" do |f| %>
+
+      <%= f.label :old_path_or_new_path_cont, "Old or new path", class: "sr-only" %>
+      <%= f.search_field :old_path_or_new_path_cont, placeholder: "Old or new path", class: "form-control mx-3" %>
```
juliannatetreault, line 9: This is not a blocker by any means and is just me being nit-picky, but I would prefer the placeholder to read "Old or New Path" rather than what it currently is--I think it looks slightly weird as is locally. 😅

app/controllers/internal/path_redirects_controller.rb
atsmith813: It's a [Ransack](https://docs.dev.to/internal/internal-search/) convention

juliannatetreault: I believe that it is specific to ransack, but I'm not sure. Good q @vaidehijoshi -  I'm curious too. 😄

app/views/internal/path_redirects/edit.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    <%= form.label :new_path %>
+    <%= form.text_field :new_path, class: "form-control" %>
+  </div>
+  <%= form.submit "Update Path Redirect", class: "btn btn-primary" %>
+<% end %>
+
+<hr class="mt-5">
+
+<div class="d-flex justify-content-start">
+  <%= link_to "Destroy Path Redirect", url_for(action: :destroy, id: @path_redirect.id), method: :delete, data: { confirm: "Are you sure?" }, class: "btn btn-danger ml-2 float-right" %>
```
juliannatetreault, line 32: I'm not sure of the best way to show this, so I've attached two screenshots to help clarify (see below), but I don't believe that these two buttons are aligned.
Buttons:
Screen Shot 2020-06-01 at 3 56 49 PM

Better view of them not looking aligned:
Screen Shot 2020-06-01 at 3 57 30 PM

app/models/path_redirect.rb
atsmith813: No, it has to be `self` when setting the value to refer to the attribute on the record itself. Without `self` Ruby will interpret as trying to set/modify a local variable of `version` which isn't defined and results in:
```ruby
NoMethodError: undefined method `+' for nil:NilClas
```

db/migrate/20200527161505_create_path_redirect.rb
atsmith813: Yes, we'll be programmatically creating `path_redirects` - and that's why it defaults to `"service"`.

When an `Article` is unpublished, we'll kick off a Sidekiq worker to call a service (`"service"`) that determines what the best `path_redirect` to create is. That's the default case.

An admin would only override this in rare circumstances where we want to control a situation more directly for some 1-off reason. For example, maybe a super popular React post is unpublished and the service calculated an appropriate redirect as some `Article` of something that's far less popular and not quite in the same vein. At that point, an admin could override what was determined programmatically and make the redirect point to the second most popular and more relevant React article.

The only thing we have to go off of when determining an appropriate redirect is the path itself. So when we go to search for an appropriate redirect, there will be some situations where the calculated result isn't quite the best since we're going off of so little. Or maybe we find no article at all and fallback on a tag instead. In these cases, an admin can override it. In theory, this is really only necessary in the event of a _super_ hot Article.

app/views/internal/path_redirects/edit.html.erb
atsmith813: That's somewhat intentional. The `destroy` button is supposed to be in a pseudo "footer". It would look better if we had more buttons and actions in the footer. Since this is Internal, I didn't bother with it.

juliannatetreault: Ahh, I see! Thanks for the explanation. 👍 👍

app/views/internal/path_redirects/index.html.erb
atsmith813: Updated!

app/controllers/internal/application_controller.rb
atsmith813: Good call - both make sense to me and it is confusing as is 😂 

I added a comment to clarify the use. I see what you mean about using an instance variable since it gets used in a view.

I'm thinking of sticking with a constant since the value is constant - we only need it initialized once and we can save ourselves the memory.

It also feels a little weird to put instance variables for a shared view in the application controller...maybe that part is just me 🙈 If it were a siloed view attached to a controller and an action, it would feel less weird. And since this is all in internal which is kinda, sorta a free for all until it's completely re-done, I think it still fits.

What do you think?

app/views/internal/path_redirects/edit.html.erb
atsmith813: Love that - updated!

Screen Shot 2020-06-01 at 6 29 28 PM

app/controllers/internal/application_controller.rb
vaidehijoshi: Yeah, I think I agree on all accounts, and the comment does the job 👍

db/migrate/20200527161505_create_path_redirect.rb
vaidehijoshi: Hm, based on all of this, I’m curious why we don’t have the service that is creating the redirect object be the one to actually set the source as service? The reason I think that this makes more sense is that it aligns with what you described, and it avoids potential bugs where perhaps someone creates a service in...the console (?) and the service defaults to a source of “service” when that is actually not the true source. I can imagine down the road that we’re trying to debug a situation (something something race condition) and we actually have a redirect that is not created by the service but its default value will lead us to believe that it is...

atsmith813: Ahh, I see - that makes sense! Updated to allow `null` instead 👍

vaidehijoshi:
> A question: is there a chance for creating an infinite loop here?

Ahh this is a super great point, I didn't think of this. Shall we put a unique constraint on the `new_path` to avoid this?

vaidehijoshi:
I don't think that this necessarily should block this PR, but when it comes time to actually do the "redirect" logic, we may want to add some validation around the path itself:
Screen Shot 2020-06-02 at 7 05 03 AM

Relatedly, we also will want to either validate _or_ handle the scenario when someone does not include a prepended `/` in the relative path, since right now I don't actually need to put it in to create the path :)

atsmith813:
@rhymes @vaidehijoshi

That's a great point!

In code...no. When we redirect at the controller level, we'll be adding something like
```?context=not-found```
to the URL so the frontend knows to show a banner. We'll use that as well to skip the redirect if it's present because that means it was already done.

In the DB...yes. You can have a situation where you have `/a` --> `/b` and then `/b` -->`/a`. Both `old_path` values are unique, but they point back to each other. We also want to leave open the idea of having `/a` --> `/b` and `/c` --> `/b` being able to exist side-by-side. It's actually better for SEO to allow that because we're concentrating SEO juice to fewer locations.

I like the idea of adding a custom validation at the model level for more protection though...but it will cost us a DB query. I'll whip up a validation real quick and add it in.

**Comment by atsmith813** [comment]
> Relatedly, we also will want to either validate _or_ handle the scenario when someone does not include a prepended `/` in the relative path, since right now I don't actually need to put it in to create the path :)

@vaidehijoshi Agreed! My plan is to address this later on in this feature lifecycle, but still before the production release of v1. Right now, I'm making some assumptions about these URLs in my head. I won't know for sure until I go to implement the service and controller concern in detail. In those PRs, I plan to add a quick URL validation once I confirm some thoughts.

At a minimum, making sure baseline things such as `/` being present at the beginning makes sense. I have some unclarified thoughts about handling URLs when it comes to new domains, subdomains, etc. when handling URLs, especially with Forem.

It's mostly a concern for admin overrides since the services will be picking up _only_ valid and existing URLs when creating redirects.

**Comment by atsmith813** [comment]
@vaidehijoshi @rhymes re-requesting review to confirm you're both okay with the validations I added:
- Validate that `old_path` isn't the same as `new_path` (pointless redirect).
- Validate that `new_path` isn't an existing redirect. This will prevent `/a` --> `/b` and `/b` --> `c`. Instead, that should be a single record of `/a` --> `/c` resulting in 1 redirect instead of 2. It also prevents the aforementioned infinite redirect loop possibility of `/a` --> `/b` and `/b` --> `/a`.

File: app/models/path_redirect.rb [hunk 25]
```diff
@@ -0,0 +1,40 @@
...
+  validates :old_path, presence: true, uniqueness: true
+  validates :new_path, presence: true
+
+  # Validate old_path != new_path
+  validates :old_path, exclusion: { in: ->(path_redirect) { [path_redirect.new_path] }, message: "the old_path cannot be the same as the new_path" }
+
+  validates :source, inclusion: { in: SOURCES }, allow_blank: true
+
+  # This issues a DB query so best to keep this validation as far down as possible
+  validate :new_path_wont_redirect
```
rhymes, line 13: If you want to eventually replace this with a DB unique index so you can avoid the DB query, you can do the following:

```sql
rhymes@/tmp:test> create extension if not exists hstore;
rhymes@/tmp:test> create unique index uniqut_ab on uniqt using btree (hstore(ARRAY[a,b], ARRAY[a,b]));
CREATE INDEX
Time: 0.009s
rhymes@/tmp:test> insert into uniqt values ('a', 'b');
INSERT 0 1
Time: 0.004s
rhymes@/tmp:test> insert into uniqt values ('b', 'a');
duplicate key value violates unique constraint "uniqut_ab"
DETAIL:  Key (hstore(ARRAY[a, b], ARRAY[a, b]))=("a"=>"a", "b"=>"b") already exists.
```

taken from https://stackoverflow.com/a/18361118/4186181

The reason why this works is because is because PostgreSQL is using its builtin key value store (hstore) to create an index where the permutation of `a, b` and `b, a` has to be unique.

In Rails terms it would translate into something like (untested):

```ruby
ActiveRecord::Base.connection.enable_extension("hstore")
ActiveRecord::Base.connection.execute(
  <<~SQL
    CREATE UNIQUE INDEX CONCURRENTLY "index_name"
    ON "page_redirects"
    USING btree (hstore(ARRAY[a,b], ARRAY[a,b]));
  SQL
)
```

in the migration file.

cc @vaidehijoshi

atsmith813: TIL - this is _sweet_! Thanks for sharing 👍

PR #8163, opened 2020-05-29
Spec Fix: Update Button to Match New Editor Changes

mstruve:
Spec Fix: We changed the buttons for publishing articles with our recent editor changes so the word is no longer all caps. This updates the button in the spec to match


mstruve:
Solved by #8162

PR #8175, opened 2020-05-29
Streamline Tag parsing for both editors

maestromac:
- [x] Feature

## Description
We handle tags differently between our two editors and this PR serves as a means to streamline that. v1 editor will not correct user as they input but will do the following 3 with `TagParser`
1. Downcase input.
2. Remove all non-alphanumric input, ie `web-assembly` to `webassembly` but also `c++` to `c` ☹️
3. Automatically replace it with the alias that we prefer, ie replace `webassembly` with `wasm`.

Because v2 editor does not currently utilize `TagParser`, none of this is applied to those tags and there's a mismatch, more importantly it does not downcase. v1 editor also has the problem of fixing user's input which can cause unexpected behavior.

What I did in this PR is the following
- I removed the Article model's dependency on `TagParser`; this is because I believe `ActAsTaggable`'s parser class's sole responsibility is to decide how to split input instead of modifying user's entry. More importantly, we want to inform the user that their `c++` tag is invalid instead of automatically changing it to `c` . This is something the default parser, ie `parser: true`, handles out of the box.
- `Article#set_tag_list` will be the source of truth for dealing with tags for both v1 and v2 editor's content.
- Use `ActsAsTaggableOn.force_lowercase` to make sure all tags will be lowercase.

The biggest change here is we will not automatically remove non-alphanumeric from the user's tag input and will instead, raise an error to the user. This will be the same on both editors and API. 

I can't fully remove `TagParser` just yet as ClassifiedListing depends on it but it is on my roadmap.

n/a
image

## Added tests?
- [x] yup yup
## Added to documentation?
- [x] no documentation needed

n/a


app/models/article.rb
```diff
@@ -446,12 +458,7 @@ def before_destroy_actions
 
   def evaluate_front_matter(front_matter)
     self.title = front_matter["title"] if front_matter["title"].present?
-    if front_matter["tags"].present?
-      self.tag_list = [] # overwrite any existing tag with those from the front matter
-      tag_list.add(front_matter["tags"], parser: ActsAsTaggableOn::TagParser)
-      remove_tag_adjustments_from_tag_list
-      add_tag_adjustments_to_tag_list
```
maestromac, line 453: Removed these because they are redundant. This will be handled with `validate_tag`

spec/liquid_tags/link_tag_spec.rb
```diff
@@ -3,7 +3,7 @@
 RSpec.describe LinkTag, type: :liquid_tag do
   let(:user) { create(:user, username: "username45", name: "Chase Danger", profile_image: nil) }
   let(:article) do
-    create(:article, user_id: user.id, title: "test this please", tags: "tag1 tag2 tag3")
```
maestromac, line 6: these were invalid the whole time 😬

app/models/listing.rb
```diff
@@ -73,7 +73,9 @@ def evaluate_markdown
   end
 
   def modify_inputs
-    ActsAsTaggableOn.default_parser = ActsAsTaggableOn::TagParser
```
maestromac, line 76: Setting the default_parser caused conflict in testing so I have to change how Tag is parsed in Listing.

app/services/rss_reader/assembler.rb
```diff
@@ -32,7 +32,9 @@ def assemble
     private
 
     def get_tags
-      @categories.first(4).map { |tag| tag.lstrip[0..19] }.join(",")
+      @categories.first(4).map do |tag|
+        tag.delete(" ").gsub(/[^[:alnum:]]/i, "")[0..19]
+      end.join(",")
```
maestromac, line 37: I have to make an exception in RSS parsing because feed may contain tags with hyphens. Rather than flat out rejecting those feed import because of malformed tag, this will follow what v1 editor used to do.

app/models/tag.rb
```diff
@@ -76,6 +76,20 @@ def self.aliased_name(word)
     tag.alias_for.presence || tag.name
   end
 
+  def self.find_alias_for(candidate)
```
mstruve, line 79: Why do we have multiple levels of aliases? This is not going to be very performant and arbitrarily going 4 deep feels a little weird. I think we should be eliminating the in-between aliases if the goal is always to arrive at the lowest common denominator.

app/models/listing.rb
mstruve: > caused conflict in testing 

What do you mean by conflict?

app/models/article.rb
```diff
@@ -383,12 +383,24 @@ def evaluate_markdown
...
+    end
+
     self.description = processed_description if description.blank?
   rescue StandardError => e
     errors[:base] << ErrorMessageCleaner.new(e.message).clean
   end
 
+  def set_tag_list(tags)
+    self.tag_list = [] # overwrite any existing tag with those from the front matter
+    tag_list.add(tags, parse: true)
```
mstruve, line 400: 🔥

app/models/listing.rb
maestromac: I should've expanded on that. When running listing (and related) specs and my new spec together, ie 

```
bundle exec rspec spec/lib/data_update_scripts/index_listings_to_elasticsearch_spec.rb spec/models/article_spec.rb:245
```

My new spec will fail. This is because if the code path went through Listing Model which will go through the code `ActsAsTaggableOn.default_parser = ActsAsTaggableOn::TagParser` , setting the default parser I'm trying to avoid, hence failing my spec (automatically removing non-alphanumerics).

app/models/tag.rb
maestromac: That actually is an arbitrary number I thought of as I extracted the code from [TagParser ](https://github.com/thepracticaldev/dev.to/blob/master/app/lib/acts_as_taggable_on/tag_parser.rb#L20) in case we ever accidentally looped an alias onto itself. 

The original idea was we would swap in a more preferred tag and it would be more straight forward to apply new alias onto the existing preferred tag instead of changing the alias of the tail. We were definitely wrong there and it should've been the other way around but this isn't a cause for concern as there is not a lot of Tags that currently have an alias. Would it be better if I lower that number down to 2?

mstruve: If we have a low number of "alias chains" then I think we should jump in a console and remove the alias chains in favor of a single alias and then we don't need this looping logic. What do you think?

maestromac: Let's do that

app/models/tag.rb
```diff
@@ -76,6 +76,10 @@ def self.aliased_name(word)
     tag.alias_for.presence || tag.name
   end
 
+  def self.find_preferred_alias_for(word)
```
mstruve, line 79: We should dedup this at some point with the method above but its fine for now.

maestromac: I also gave that some thought but ended up keeping it like so. They are different though; the method above makes sure the given tag exists, this one does not.

app/models/article.rb
```diff
@@ -497,12 +504,15 @@ def validate_tag
 
   def remove_tag_adjustments_from_tag_list
     tags_to_remove = TagAdjustment.where(article_id: id, adjustment_type: "removal", status: "committed").pluck(:tag_name)
-    tag_list.remove(tags_to_remove, parser: ActsAsTaggableOn::TagParser) if tags_to_remove.present?
+    tag_list.remove(tags_to_remove, parse: true) if tags_to_remove.present?
```
rhymes, line 507: @maestromac what is "parse: true" ?

config/initializers/acts_as_taggable_on.rb
```diff
@@ -0,0 +1 @@
+ActsAsTaggableOn.force_lowercase = true
```
rhymes, line 1: 🔥 do we have upper cased tags in the DB?

app/models/article.rb
maestromac: previously with `parser: ActsAsTaggableOn::TagParser` we ask the tags string (or array) to be parsed with a particular parser. With `parse: true`, we ask it to use the default parser.

config/initializers/acts_as_taggable_on.rb
maestromac: yes we currently do 😬

app/models/tag.rb
maestromac: @atsmith813 yep that's correct but I just got fooled by my own code (again) so I'm going to verbosely downcase it anyway as this method on its own is not tied to ActAsTaggableOn

PR #8189, opened 2020-05-30
Reload on comment error instead of alert

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

There is currently a Datadog SSL issue

https://status.datadoghq.com/incidents/6bqpd511nj4h

Since the only real _user-facing_ problem right now is that they get this message due to our catch-all for errors is to toss up this alert, I feel like we may be better off just refreshing the page on error for now.

Then they will see that their comment was submitted successfully.

Screen Shot 2020-05-30 at 1 49 48 PM


mstruve:
Hotfix for an issue that ended up resolving itself. Closing

PR #8190, opened 2020-05-30
pages:sync rake task for easier local coding of HTML Page

fdoxyz:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Rake task that listens for changes to a local HTML file. It will keep in sync a Page's HTML body with the content of this file. 

**Usage:**
1. Start your local server
1. Bind a Page by passing the `slug` and the local `filepath` of the HTML file (absolute path)
1. Edit the local HTML file on your Text Editor of choice and refresh the Page on your browser

Screen Shot 2020-06-01 at 07 54 52

[The first attempt](https://github.com/thepracticaldev/dev.to/pull/8152) at improving this process was overly complex and not aligned with the generalization behind the project.

This is an unobtrusive approach that speeds up coding and relies on the `Listen` gem, which was already available in the development environment

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

maybe?


lib/tasks/pages.rake
```diff
@@ -0,0 +1,19 @@
+namespace :pages do
+  desc "Automated sync of Page HTML body by listening to changes of local file"
+  task :bind, [:slug, :filepath] => [:environment] do |task, args|
+    listener = Listen.to(Pathname.new(args.filepath).dirname) do
```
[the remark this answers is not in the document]
rhymes: also, this is going to watch not just the file, but the entire folder which we should probably be very explicit about with messages display on the console, like "Watching FOLDER_NAME for changes..."

lib/tasks/pages.rake
```diff
@@ -0,0 +1,19 @@
...
+      puts "Reloading page..."
+      page = Page.find_by(slug: args.slug)
+      page.body_html = File.read(args.filepath)
+      puts "Error updating page: #{page.errors.messages}" unless page.save
```
rhymes, line 8: I think we should let the user know when the page is updated, the happy path, not just when the save fails.

fdoxyz:
I implemented the suggestions @rhymes, I thought were all great. Also added a test for this, which was a bit difficult but it seems to cover the use case 🙂

spec/tasks/pages_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "rails_helper"
```
rhymes, line 1: LOL I shouldn't say this but I feel like we should get rid of this test if it's too costly or fix its present state.

Hear me out:

- this test is testing a task that should only be enabled in development (forgot to flag that we should make sure that the task doesn't do anything in production mode at least).
- this test spawns threads which are going to change data on disk, so if the user hits Ctrl-C between before and after but the thread finishes, they'll have changes on disk

If we want to keep this test in, we should either:

- use Tempfile as the destination (you could copy the fixture file to a temp file and then unleash the threads on that)
- mock `File.read`/`File.write` to write in a memory mapped file but this might be trickier

fdoxyz: Makes a lot of sense! I agree that we should avoid the possible side effects the test can have.

Just sent in the fix for this.

mstruve: I'm with @rhymes, I really think we should skip the test for this.  There are lots of sleeps in there which are going to be slow and its not code we rely on in production so if it breaks its not the end of the world.

fdoxyz:
> We should probably also add some documentation on this so that other contributors can make use of it as well.

I agree @mstruve, would it make sense to add it over in [Architecture](https://docs.dev.to/technical-overview/architecture/)? 

I'd have to add a section at the bottom to explain the concept of Pages first and then make reference to this utility task. I couldn't really find any other section that would fit this purpose.

I think I'll need someone with privileges to merge this for me because of the missing test coverage

mstruve:
>  would it make sense to add it over in Architecture?

Someone from @thepracticaldev/oss should chime in but I think that would be perfect. Explain what pages are, how we create them, then link the script for testing.

PR #8194, opened 2020-05-31
Embed github code support for liquid tags

### LINKED ISSUE #2046: Embed GitHub Code
I tried to embed a few lines of code in a post from github. It did not work, but would be cool if it did.

This would be the expected format, I think

{% github https://github.com/johnpapa/vscode-peacock/blob/master/.vscodeignore#L12-L17 %}

colorfusion:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description
This feature will add support to display GitHub code permalink through liquid tags.

Fixes #2046.

image

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [x] docs.dev.to
- [x] readme
- [ ] no documentation needed

awesome


app/lib/redcarpet/render/html_rouge.rb
```diff
@@ -5,11 +5,31 @@ module Render
     class HTMLRouge < HTML
       include Rouge::Plugins::Redcarpet
 
-      # Rouge requires the hint language to be lower case, by overriding this
-      # method we can allow the hint language to be specified with other casings
-      # eg. `Ada` instead of `ada`
-      def block_code(code, language)
-        super(code, language.to_s.downcase)
+      # overrided method to add line number support
+      def block_code(code, language, opts = {})
```
rhymes, line 9: is this going to impact code blocks in the markdown around the website. I'm asking this as `HTMLRouge` is used both in comments and articles.

Should we add specific tests for this?

app/liquid_tags/github_tag/github_code_tag.rb
```diff
@@ -0,0 +1,160 @@
...
+          file_path: @content[:file_path],
+          body: @content[:body],
+          line_info: @content[:line_info],
+          original_link: @original_link,
+          ref_name: @content[:ref_name]
+        },
+      )
+    end
+
+    def parse_link(link)
```
rhymes, line 24: I think from here down everything can be `private`

app/liquid_tags/github_tag/github_code_tag.rb
```diff
@@ -0,0 +1,160 @@
...
+        file_path: file_path,
+        file_type: file_type,
+        start_line: start_line,
+        end_line: end_line
+      }
+    end
+
+    def get_file_contents(repo_info, file_info)
+      client = Octokit::Client.new(access_token: token)
+      file = client.contents(repo_info[:repo_path],
```
rhymes, line 75: you should use `Github::Client.contents` here. See https://github.com/thepracticaldev/dev.to/blob/3b188d349ad018d41c88ee2a9e25ca1dedac06a3/app/liquid_tags/github_tag/github_readme_tag.rb#L15

app/liquid_tags/github_tag/github_code_tag.rb
```diff
@@ -0,0 +1,160 @@
...
+      raise_error unless /.*github\.com\//.match?(link)
+      link.gsub(/.*github\.com\//, "")
+    end
+
+    def raise_line_number_error
+      raise StandardError, "Line number is invalid"
+    end
+
+    def raise_error
+      raise StandardError, "Invalid Github link"
```
rhymes, line 149: `GitHub`

app/liquid_tags/github_tag/github_code_tag.rb
```diff
@@ -0,0 +1,160 @@
...
+    end
+
+    def token
```
rhymes, line 152: this is not needed, see my comment about using the global `Github::Client`

app/views/liquids/_github_code.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="ltag-github-code-tag">
```
rhymes, line 1: I wonder if the UI could have two improvements:

1. a header similar to this:

Screenshot_2020-06-01 New Post - DEV(local) Community 👩‍💻👨‍💻

(you can get the style from `GithubReadmeTag.scss`)

2. (totally optional and can be done in another PR): syntax highlighting. 

Unfortunately GitHub's contents API doesn't return the detected language it seems: https://developer.github.com/v3/repos/contents/#get-contents

spec/liquid_tags/github_tag/github_code_tag_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    let(:file_name) { "johnpapa/vscode-peacock/.vscodeignore" }
+    let(:path) { "https://github.com/johnpapa/vscode-peacock/blob/master/.vscodeignore" }
+
+    setup { Liquid::Template.register_tag("github", GithubTag) }
+
+    def generate_github_code(path, line_number, _options = "")
+      Liquid::Template.parse("{% github #{path}#{line_number} %}")
+    end
+
+    # need to figure out how to generate a token that works
```
rhymes, line 19: you can configure your local dev installation with github, see https://docs.dev.to/backend/auth-github/ and then when you'll have `GITHUB_KEY` and `GITHUB_SECRET` configured in your `application.yml` they will be picked up by the `Github::Client` class.

Then you can use VCR. Don't worry, VCR won't store your secrets in the yml file (double check :)).

Example of you to use `VCR`: 

https://github.com/thepracticaldev/dev.to/blob/3b188d349ad018d41c88ee2a9e25ca1dedac06a3/spec/liquid_tags/github_tag/github_readme_tag_spec.rb#L36-L41

don't forget to enable VCR for the file:

https://github.com/thepracticaldev/dev.to/blob/3b188d349ad018d41c88ee2a9e25ca1dedac06a3/spec/liquid_tags/github_tag/github_readme_tag_spec.rb#L3

colorfusion:
Thanks for the feedback, I'll work on those adjustments.

colorfusion:
Hey @benhalpern, I'm still looking into the changes and working on it. I got a little busy on my end, so I'll try to iron things out by this weekend hopefully. 😎

app/lib/redcarpet/render/html_rouge.rb
colorfusion: Based on the changes I've made, for default markdown cases, it should call the function correct. I've extracted the original code from [redcarpet.rb](https://github.com/rouge-ruby/rouge/blob/master/lib/rouge/plugins/redcarpet.rb) in the original repo, and made a few adjustments.

This allows `Formatters::HTMLLegacy` to take in additional parameters for line number rendering.

If we need to add tests cases for this, where should I add them in?

rhymes: I'm cceing @maestromac which probably has insight about this

app/views/liquids/_github_code.html.erb
colorfusion: > 1. a header similar to this:

When you're saying a header similar to the GitHubReadme you're referring the path having different links for each part? The way I wrote the second line was based on Github's way of rendering code snippets.

> 2. syntax highlighting

If I remember correctly, passing the file type through the `HTMLRogue` will provide the syntax highlighting, the screenshot was using `.vscodeignore` that might not have highlight support. I  will need to fix my current dev environment (WSL + Windows), to get the page preview working again to confirm the behavior.

app/lib/redcarpet/render/html_rouge.rb
maestromac: This looks like a safe change as long as it passes all of our `spec/labor/markdown_parser_spec.rb` specs. @colbymelvin could you please add new tests to verify the functionality of this code? it can be added to `markdown_parser_spec.rb`

colorfusion: Sure, will do that.

app/views/liquids/_github_code.html.erb
colorfusion: Hey @rhymes, may I know what you're referring to when you said a header similar to the screenshot above?

colorfusion:
Yeah, I'll work on that.

app/lib/redcarpet/render/html_rouge.rb
colorfusion: @maestromac may I know what kind of test I should be adding into the `markdown_parser.spec.rb`? The `MarkdownParser` class uses `HTMLRouge` as it's renderer, but it passes it into `Redcarpet::Markdown` to render the markdown. But I'm using the HTMLRogue class directly to render the file itself.

Am I supposed to use `HTMLRogue` directly in the test to check it renders properly, or use the MarkdownParser class? I'm not exactly sure whether the class will call the `block_code` function.

rhymes:
Hi @colorfusion, we've let this PR slip a bit but we're going to review it ASAP and hopefully finalize it. I've merged master and fixed one line to make it work again after so many months

rhymes (commented):
One thing I've noticed while testing it is that `https://github.com/forem/forem/blob/9af120561ca8edfdacded615231f984fcad8e8c1/app/models/article.rb#L1-L7` works but `https://github.com/forem/forem/blob/9af120561ca8edfdacded615231f984fcad8e8c1/app/models/article.rb` (so without the line numbers) doesn't. It'd be nice if it worked nonetheless by embedding the first few lines or the entire file with a max height and scrollbars

ludwiczakpawel (changes requested):
I think there're couple things we should handle before shipping it. There're some issues with how code and line numbers are being displayed:

1. Double border-radius on both pane: for line numbers and code box itself - it looks a bit off.
2. De-emphasize visually line numbers - maybe smaller font, maybe less opacity - it's just something that isn't crucial for end use experience and should be somehow distinguishable from rest of the code - so people don't think it's part of code.
3. Right-aligned line numbers - that's pretty common practice and makes scanning numbers a bit easier.
4. More bullet-proof frontend for line numbers part - it's currently fixed with and can cause glitches when line number is 3+ digits - for clipping and/or overflow.
5. Limit height - it's a bit more drastic change request and I'm keen to hear others feedback, but shouldn't we limit the height of this liquid tag and if it gets too big show the "View on github" button like we do elsewhere?

@colorfusion - let me know if these are sensible requests. I'm also very happy to help with 1-4 if you need it.

rhymes:
> Limit height - it's a bit more drastic change request and I'm keen to hear others feedback, but shouldn't we limit the height of this liquid tag and if it gets too big show the "View on github" button like we do elsewhere?

I agree, that would be nice! :-)

cmgorton:
Hi @colorfusion !

PR #8205, opened 2020-05-31
Fix gists sometimes rendered in wrong order (attempt)

### LINKED ISSUE #1679: Post with multiple embedded gists have order mixed up
**Describe the bug**
On my post https://dev.to/rasmusvhansen/rxjs-transducer---harness-the-power-of-rxjs-operators-1ai8
I have 5 gists embedded in the page. I noticed that the order of the gists is not fixed. On some reloads, the first and e.g third gists are swapped. 

Other users have also noticed this (see comments on post)

**To Reproduce**

Note: It is kind of hard to reproduce. I saw it twice but cannot consistently reproduce.

1. Go to https://dev.to/rasmusvhansen/rxjs-transducer---harness-the-power-of-rxjs-operators-1ai8
2. Check order of gists
3. Reload and check if order is the same
4. Repeat step 3 until error occurs

**Expected behavior**
Order of gists should be as specified in markdown, not random.

**Screenshots**

- OS: Windows 10
- Browser: Chrome
- Version: 71

- Device: iPhone 6.1
- OS: iOS 12.1
- Browser Safari

cardoso:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Debugging finds

When debugging, I found that the `beforeWrite` function is sometimes called with the wrong text, causing the bug. The `context` variable seems to be correct, which I think rules out my solution and seems to point to a bug in postscribe itself. Or maybe it should only be called once at a time?

## Description

Sending this as an attempt to fix the gist ordering bug (think of it as a place for discussion). I haven't confirmed the fix works (probably not!), since I don't have time to set up the dev environment this weekend, but, looking at the code, this seems like a mistake that could cause it.

I made a small script to reproduce something similar:

```javascript
function randomInt(min, max) {
	return min + Math.floor((max - min) * Math.random());
}

function makeRequest() {
    return new Promise((resolve) => {
       setTimeout(resolve, randomInt(0, 5000);
    });
}

async function process() {
    var els = ["one", "two", "three", "four", "five"];
    for(var i = 0; i < els.length; i++) {
      let a = i + 1;
      let current = els[i]; // change to `var` to reproduce bug
      makeRequest().then(e => console.log(`${a}: ${current}`));
    }
}

process();
```

I normally wouldn't send a PR without fully testing it in practice, but I really wanted to kick start this, since it's affecting my tutorials and leaving readers confused. I'll try and make time next weekend to set everything up if no one can confirm/debunk this fix.

## Reproducing the bug

Link to my tutorial: https://dev.to/cardoso/how-to-build-a-twitch-clone-game-live-streaming-app-for-ios-5f0p
Image of bug happening:
IMG_E20C3831DE80-1
Correct snippet is: https://gist.github.com/cardoso/889bb3933d0f135d8889b028f4153708#file-podfile

I reproduced in Safari and Chrome. macOS and iOS. Normally the first snippet is affected, but also throughout the tutorial other snippets may be wrong.

**IMPORTANT:** It seems to happen very frequently when navigating from another page (such as dashboard) to the tutorial. On reload, it's always correct for me.

Attempts to fix #1679 

## Added tests?

- [ ] yes
- [ ] no, because I need help

Might be needed, if so, point me in the right direction :)

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/views/articles/show.html.erb
+++ b/app/views/articles/show.html.erb
@@ -14,7 +14,7 @@
           clearInterval(waitingOnPostscribe);
           var els = document.getElementsByClassName("ltag_gist-liquid-tag");
           for (var i = 0; i < els.length; i++) {
-            var current = els[i];
+            let current = els[i];
             postscribe(current, current.firstElementChild.outerHTML, {
               beforeWrite: function (context) {
                 return function (text) {


app/views/articles/show.html.erb
```diff
@@ -14,7 +14,7 @@
           clearInterval(waitingOnPostscribe);
           var els = document.getElementsByClassName("ltag_gist-liquid-tag");
           for (var i = 0; i < els.length; i++) {
-            var current = els[i];
+            let current = els[i];
```
nickytonline, line 17: This looks like it should fix the issue. If you don't have time to get all set up on your local machine, you can use GitPod. See our [documentation on Gitpod](https://docs.dev.to/installation/gitpod/).

PR #8210, opened 2020-06-01
Fix badge count update after merge

citizen428:

- [X] Bug Fix

## Description

We recently merged two accounts on a user's request, but badges did not show up on the new account. Trying to reassign them also did not work. I did some digging and found the following:

1. The badges had been moved over and assigned to the new user correctly. This is why they couldn't be reassigned. 
2. In `app/views/articles/_badges_area.html.erb` we perform the following conditional check:
    ```ruby
    if @user.badge_achievements_count > 0 
    ````
3. I checked, and the `badge_achievements_count` for the new user was indeed 0. 

I then had a look at `Moderator::MergeUser` and fixed the count update logic during merges. The newly added spec fails with the old code and succeeds with this change.

https://github.com/thepracticaldev/tech-private/issues/445#issuecomment-636615311

## Added tests?

- [X] yes

## Added to documentation?

- [X] no documentation needed

image

### DIFF
--- a/app/services/moderator/merge_user.rb
+++ b/app/services/moderator/merge_user.rb
@@ -1,15 +1,15 @@
 module Moderator
   class MergeUser < ManageActivityAndRoles
+    def self.call(admin:, keep_user:, delete_user_id:)
+      new(keep_user: keep_user, admin: admin, delete_user_id: delete_user_id).merge
+    end
+
     attr_reader :keep_user, :admin, :delete_user_id
 
     def initialize(admin:, keep_user:, delete_user_id:)
       @keep_user = keep_user
       @admin = admin
       @delete_user = User.find(delete_user_id.to_i)
-    end
-
-    def self.call_merge(admin:, keep_user:, delete_user_id:)
-      new(keep_user: keep_user, admin: admin, delete_user_id: delete_user_id).merge
     end
 
     def merge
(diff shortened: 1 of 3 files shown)


app/services/moderator/merge_user.rb
```diff
@@ -2,16 +2,16 @@ module Moderator
   class MergeUser < ManageActivityAndRoles
     attr_reader :keep_user, :admin, :delete_user_id
 
+    def self.call(admin:, keep_user:, delete_user_id:)
```
citizen428, line 3: Since I was here already I made this conform more to our other service objects:

1. Main API method called `call`.
2. Class methods before instance methods (a convention often found in Ruby projects).

app/services/moderator/merge_user.rb
```diff
@@ -56,7 +56,7 @@ def merge_profile
       end
       if @delete_user.badge_achievements.any?
         @delete_user.badge_achievements.update_all(user_id: @keep_user.id)
-        @keep_user.badge_achievements_count = @keep_user.badge_achievements.size
+        BadgeAchievement.counter_culture_fix_counts(where: { users: { id: @keep_user.id } })
```
citizen428, line 59: See: https://github.com/magnusvk/counter_culture#manually-populating-counter-cache-values

PR #8211, opened 2020-06-01
Clean up BadgeRewarder

citizen428:

- [X] Refactor

## Description

I was looking at our `BadgeRewarder` in the context of generalization and while doing so noticed that there are some low hanging fruits for making the code a bit nicer, i.e. using our new `URL` helper module. etc.

n/a

## Added tests?

## Added to documentation?

- [X] no documentation needed

### DIFF
--- a/app/labor/badge_rewarder.rb
+++ b/app/labor/badge_rewarder.rb
@@ -13,7 +13,7 @@ module BadgeRewarder
     (1..3).each do |i|
       message = "Happy #{ApplicationConfig['COMMUNITY_NAME']} birthday! Can you believe it's been #{i} #{'year'.pluralize(i)} already?!"
       badge = Badge.find_by!(slug: "#{YEARS[i]}-year-club")
-      User.where(created_at: i.year.ago..i.year.ago - 2.days).find_each do |user|
+      User.where("created_at < ? AND created_at > ?", i.year.ago, i.year.ago - 2.days).find_each do |user|
         achievement = BadgeAchievement.create(
           user_id: user.id,
           badge_id: badge.id,
@@ -97,7 +97,7 @@ module BadgeRewarder
       count = 0
       num_weeks.times do |i|
         num = i + 1
-        count += 1 if user.articles.published.where(published_at: num.weeks.ago..(num - 1).weeks.ago).any?
+        count += 1 if user.articles.published.where("published_at > ? AND published_at < ?", num.weeks.ago, (num - 1).weeks.ago).any?
       end
       usernames << user.username if count >= num_weeks
     end
--- a/app/labor/badge_rewarder.rb
+++ b/app/labor/badge_rewarder.rb
@@ -1,12 +1,19 @@
 module BadgeRewarder
-  YEARS = { 1 => "one", 2 => "two", 3 => "three", 4 => "four", 5 => "five", 6 => "six", 7 => "seven", 8 => "eight", 9 => "nine", 10 => "ten" }.freeze
+  YEARS = { 1 => "one", 2 => "two", 3 => "three" }.freeze
   REPOSITORIES = ["thepracticaldev/dev.to", "thepracticaldev/DEV-ios", "thepracticaldev/DEV-Android"].freeze
+
+  LONGEST_STREAK_WEEKS = 16
+  LONGEST_STREAK_MESSAGE = "16 weeks! You've achieved the longest #{ApplicationConfig['COMMUNITY_NAME']} writing " \
+    "streak possible. This makes you eligible for special quests in the future. Keep up the amazing contributions to" \
+    " our community!".freeze
+
+  MINIMUM_QUALITY = -25
 
   def self.award_yearly_club_badges
     (1..3).each do |i|
       message = "Happy #{ApplicationConfig['COMMUNITY_NAME']} birthday! Can you believe it's been #{i} #{'year'.pluralize(i)} already?!"
       badge = Badge.find_by!(slug: "#{YEARS[i]}-year-club")
-      User.where("created_at < ? AND created_at > ?", i.year.ago, i.year.ago - 2.days).find_each do |user|
+      User.where(created_at: i.year.ago..i.year.ago - 2.days).find_each do |user|
         achievement = BadgeAchievement.create(
           user_id: user.id,
           badge_id: badge.id,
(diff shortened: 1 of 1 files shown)


app/labor/badge_rewarder.rb
```diff
@@ -1,12 +1,19 @@
...
+    " our community!".freeze
+
+  MINIMUM_QUALITY = -25
+
   def self.award_yearly_club_badges
     (1..3).each do |i|
       message = "Happy #{ApplicationConfig['COMMUNITY_NAME']} birthday! Can you believe it's been #{i} #{'year'.pluralize(i)} already?!"
       badge = Badge.find_by!(slug: "#{YEARS[i]}-year-club")
-      User.where("created_at < ? AND created_at > ?", i.year.ago, i.year.ago - 2.days).find_each do |user|
+      User.where(created_at: i.year.ago..i.year.ago - 2.days).find_each do |user|
```
citizen428, line 16: I switched this to a range, which turns this into a `BETWEEN` query.  There's one difference though: `BETWEEN` is inclusive while the existing query was exclusive. I felt that this wsa ok in this case, but if people prefer that I can change it back. /cc @benhalpern

app/labor/badge_rewarder.rb
```diff
@@ -17,11 +24,10 @@ def self.award_yearly_club_badges
     end
   end
 
-  def self.award_beloved_comment_badges
-    # ID 3 is the proper ID in prod. We should change in future to ENV var.
-    badge_id = Badge.find_by(slug: "beloved-comment")&.id || 3
-    Comment.where("public_reactions_count > ?", 24).find_each do |comment|
-      message = "You're famous! [This is the comment](https://#{ApplicationConfig['APP_DOMAIN']}#{comment.path}) for which you are being recognized. 😄"
+  def self.award_beloved_comment_badges(comment_count = 24)
+    badge_id = Badge.find_by!(slug: "beloved-comment").id
```
citizen428, line 28: I removed the hard-coded badge id and made this follow behave like `award_badges`. I used Blazer to confirm that a badge with the correct slug exists in production.

app/labor/badge_rewarder.rb
```diff
@@ -17,11 +24,10 @@ def self.award_yearly_club_badges
...
+    Comment.where("public_reactions_count > ?", comment_count).find_each do |comment|
+      message = "You're famous! [This is the comment](#{URL.comment(comment)}) for which you are being recognized. 😄"
```
citizen428, line 30: Use URL helper.

app/labor/badge_rewarder.rb
```diff
@@ -49,14 +55,14 @@ def self.award_tag_badges
...
-        order("score DESC").
+        order(score: :desc).
         where("published_at > ?", 7.5.days.ago). # More than seven days, to have some wiggle room.
         cached_tagged_with(tag).first
       if winning_article
         award_badges(
           [winning_article.user.username],
           tag.badge.slug,
-          "Congratulations on posting the most beloved [##{tag.name}](#{ApplicationConfig['APP_PROTOCOL'] + ApplicationConfig['APP_DOMAIN']}/t/#{tag.name}) post from the past seven days! Your winning post was [#{winning_article.title}](#{ApplicationConfig['APP_PROTOCOL'] + ApplicationConfig['APP_DOMAIN'] + winning_article.path}). (You can only win once per badge-eligible tag)",
+          "Congratulations on posting the most beloved [##{tag.name}](#{URL.tag(tag)}) post from the past seven days! Your winning post was [#{winning_article.title}](#{URL.article(winning_article)}). (You can only win once per badge-eligible tag)",
```
citizen428, line 65: Use URL helper.

app/labor/badge_rewarder.rb
```diff
@@ -78,9 +84,10 @@ def self.award_contributor_badges_from_github(since = 1.day.ago, message_markdow
   end
 
   def self.award_streak_badge(num_weeks)
-    article_user_ids = Article.published.where("published_at > ? AND score > ?", 1.week.ago, -25).pluck(:user_id) # No cred for super low quality
-    message = if num_weeks == 16
-                "16 weeks! You've achieved the longest #{ApplicationConfig['COMMUNITY_NAME']} writing streak possible. This makes you eligible for special quests in the future. Keep up the amazing contributions to our community!"
+    # No credit for super low quality
+    article_user_ids = Article.published.where("published_at > ? AND score > ?", 1.week.ago, MINIMUM_QUALITY).pluck(:user_id)
+    message = if num_weeks == LONGEST_STREAK_WEEKS
```
citizen428, line 89: Replaced magic number with a constant.

app/labor/badge_rewarder.rb
```diff
@@ -90,7 +97,7 @@ def self.award_streak_badge(num_weeks)
       count = 0
       num_weeks.times do |i|
         num = i + 1
-        count += 1 if user.articles.published.where("published_at > ? AND published_at < ?", num.weeks.ago, (num - 1).weeks.ago).any?
+        count += 1 if user.articles.published.where(published_a: num.weeks.ago..(num - 1).weeks.ago).any?
```
citizen428, line 100: Same as above, switched to a `BETWEEN` query. Same concerns apply /cc @benhalpern

citizen428:
There are more things one could have done in this file, but this seemed enough for now. Better than what we had, not too many changes. 

What's more intersting is the long term strategoy for badges in terms of different communities, I'll try to write something up over the next few days.

app/labor/badge_rewarder.rb
```diff
@@ -1,12 +1,13 @@
 module BadgeRewarder
-  YEARS = { 1 => "one", 2 => "two", 3 => "three", 4 => "four", 5 => "five", 6 => "six", 7 => "seven", 8 => "eight", 9 => "nine", 10 => "ten" }.freeze
+  YEARS = { 1 => "one", 2 => "two", 3 => "three" }.freeze
```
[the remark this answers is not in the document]
Ridhwana: I think @benhalpern mentioned yesterday that the 4 year badge will be coming up soon (https://dev-internal.slack.com/archives/C69MABC03/p1591122156079600). Not sure if we want to amend it at this point in preparation for 4 years or just open up a new PR later but thought I'd mention it anyways :)

app/labor/badge_rewarder.rb
Ridhwana: @citizen428 whats the likelihood of other communities like mma not having this badge with the correct slug in production. Should we cater for that?

app/labor/badge_rewarder.rb
Ridhwana: Should we not break this long single line string into multiple lines?

app/labor/badge_rewarder.rb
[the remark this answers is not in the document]
citizen428: Whoever adds the 4 year badge will already have to make the PR, since the awaring is currently hard-coded for 1 - 3 years (see above), so I'd say we can add it then.

app/labor/badge_rewarder.rb
citizen428: Making badges completely optional was left for later. In the meantime I at least made the comment count configurable through an argument.

app/labor/badge_rewarder.rb
citizen428: We can, but it's in the middle of an argument list and I decided to rather leave it alone for now. No particular reason, I just figured we may soon need to find a better way to deal with copy anyway when dealing with I18n.

citizen428:
Thanks for the feedback @Ridhwana 😃

PR #8214, opened 2020-06-01
Add unique index to comments - part 6

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Follow up to https://github.com/thepracticaldev/dev.to/pull/8072

This adds a missing index to the `comments` table:

- `body_markdown user_id ancestry commentable_id commentable_type`

## Added tests?

- [ ] yes
- [ ] no, because I need help

**pre deployment**: we have a few duplicate comments in the DB:

```sql
select body_markdown, user_id, ancestry, commentable_id, commentable_type, count(*)
from comments
group by body_markdown, user_id, ancestry, commentable_id, commentable_type
having count(*) > 1
```


app/models/comment.rb
```diff
@@ -25,7 +24,7 @@ class Comment < ApplicationRecord
   before_validation :evaluate_markdown, if: -> { body_markdown }
   validate :permissions, if: :commentable
   validates :body_markdown, presence: true, length: { in: BODY_MARKDOWN_SIZE_RANGE }
-  validates :body_markdown, uniqueness: { scope: BODY_MARKDOWN_UNIQUENESS_SCOPES }
+  validates :body_markdown, uniqueness: { scope: %i[user_id ancestry commentable_id commentable_type] }
```
Ridhwana, line 27: @rhymes Why did we remove the variable?

rhymes: Because rubocop uses a static checker and it's not going to detect the scoped variables and complain forever that we don't have a unique index on this row :D 

Rubocop woes

Ridhwana: Ah I see, thanks for the explanation!

citizen428: I'm not sure how I feel about this. Rubocop is a tool that is supposed to help us if we need to change reasonable code to please it, something feels off.

PR #8215, opened 2020-06-01
Add unique indexes to follows - part 7

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Follow up to https://github.com/thepracticaldev/dev.to/pull/8214

This adds a unique index to the `follows` table:

- `followable_id followable_type follower_id`

## Added tests?

- [x] yes
- [ ] no, because I need help

**pre deployment**: we need to check if there are any duplicates in the DB. I'm unsure about that:

```sql
select followable_id, followable_type, follower_id, count(*)
from follows
group by followable_id, followable_type, follower_id
having count(*) > 1
```


mstruve:
I was not able to check how many of these we have in total but I did pull up how many we have that were created in the last 2 days and it was 3 of them so clearly we are still making duplicates and need to get that cleaned up before we can add this index.

rhymes:
@mstruve thanks!

rhymes:
@mstruve @jacobherrington 

**I started investigating and thought the following was a clue but I now think it's unrelated. I'm going to leave it here for posterity:**

We create follows with 

https://github.com/thepracticaldev/dev.to/blob/db17e98804058fa4ff172904b97ff7797e51deff/app/controllers/follows_controller.rb#L92

and we remove follows with 

https://github.com/thepracticaldev/dev.to/blob/db17e98804058fa4ff172904b97ff7797e51deff/app/controllers/follows_controller.rb#L101

`user.follow` is implemented by the [acts_as_follower gem](https://github.com/tcocca/acts_as_follower) as:

https://github.com/tcocca/acts_as_follower/blob/c5ac7b9601c4af01eb4d9112330b27be4d694ecc/lib/acts_as_follower/follower.rb#L30-L35

`first_or_create` is subject to race condition (between the `SELECT` and the `CREATE` another thread can do the same and create a duplicate)

We could override that method using [create_or_find_by](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-create_or_find_by) which is a new Rails 6 method which decreases the possibility for the race condition between `SELECT` and `INSERT`.

But this doesn't take into account `follower_id` so I'm not sure I'm any closer to where we're generating duplicates

mstruve:
I'm game to try this!!!

awasum:
> /rebase

@rhymes  for all the events caught by our Rebase action: https://github.com/thepracticaldev/dev.to/actions?query=workflow%3A%22Automatic+Rebase%22


Just one has been run: https://github.com/thepracticaldev/dev.to/runs/830925719?check_suite_focus=true, which says the PR is rebaseable.

Are you part of the Org containing this repo as @nickytonline  was able to launch the command and it worked (even if it was not successful. Does it mean Draft PRs cannot be used with this action (/rebase) ?

rhymes:
@awasum are you sure it worked? Because I can't see neither the log nor the commit for the rebase

rhymes:
@awasum that's weird, because by the log it seems the rebase failed and all those events are attributed to the wrong branch. If you see https://github.com/thepracticaldev/dev.to/runs/830925719?check_suite_focus=true points to https://github.com/thepracticaldev/dev.to/pull/9073 which has no rebase comment

awasum:
> @awasum that's weird, because by the log it seems the rebase failed and all those events are attributed to the wrong branch. If you see https://github.com/thepracticaldev/dev.to/runs/830925719?check_suite_focus=true points to #9073 which has no rebase comment

The Github Action is reading or listening to all PR comments then filtering that via the if statement in the Github Action...i.e if the event is a PR and the comment body begins with /rebase and the author of the comment is part of the https://github.com/thepracticaldev Organization. 

That rebase failed because the PR has merge conflicts with master.

rhymes:
**Update**: we are using our own fork of `acts_as_follower` so I applied the patch directly on the gem, see https://github.com/thepracticaldev/acts_as_follower/pull/1

Unfortunately, by not being an actual published gem, we vendor the whole verbatim source so the PR size skyrocketed

rhymes:
@mstruve @jacobherrington @Ridhwana  ready for a re-review

rhymes:
I set it again to draft to avoid an accidental merge, as this needs data to be cleaned up in prod first

rhymes:
@mstruve the problem is that `create_or_find_by!` won't work without the index: 

> The underlying table must have the relevant columns defined with unique constraints.

see https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-create_or_find_by

rhymes:
@mstruve let's improve the strategy here:

1. we push the index change
1. we wait a bit for the index to be created (we can't be 100% sure as we can't monitor concurrent index creation [until we update to PostgreSQL 12](https://www.depesz.com/2019/04/18/waiting-for-postgresql-12-report-progress-of-create-index-operations/))
1. we push the new gem with the `create_or_find_by`

I realized that pushing it all together means that `create_or_find_by` could be triggered before the index has been completed.

What do you think?

mstruve:
That works for me.

rhymes:
Ok @mstruve reverted the gem update and moved it to its own PR https://github.com/thepracticaldev/dev.to/pull/9145

mstruve:
Working on cleaning up existing data then will deploy this

PR #8234, opened 2020-06-02
Article Design update

ludwiczakpawel:

- [x] Refactor

## Description

This PR is another step towards implementing Crayons. This time I took Article view for a ride. It introduces new layout as well as some updated styling here and there on Article view. It's only first step of Article view updates since I wanted to split the work into smaller PRs.

image

## Added tests?

- [x] no, because I need help

## Added to documentation?

- [x] no documentation needed

## Todos

- [x] Actions bar: button styling
- [x] Actions bar: dropdown
- [x] Actions bar: mobile breakpoint
- [x] Author section at the bottom
- [x] Organization badge in content
- [x] Organization widget in sidebar
- [x] Check responsiveness
- [x] Sidebar widget: add IF statements if widget is empty
- [x] Videos
- [ ] 
- [ ] 
- [ ] 
- [ ]


app/views/articles/_user_metadata.html.erb
```diff
@@ -1,122 +1,104 @@
-<style>
-  .user-metadata-details-inner a {
```
rhymes, line 2: what about this?

ludwiczakpawel: What do you mean?

I basically got rid of profile colors for metadata - it was usually causing issues anyway (remember all the github PRs about text readability in different themes?) and instead I'll be introducing profile colors in other places (see that stripe at the top of the profile widget in sidebar?)

ludwiczakpawel:
@rhymes good catches about circles and copy to clipboard style.. I'll fix it!

@msarit 

> On first load, everything seemed out of whack; may be just my dev env though

it's just caching, don't worry about it :)

> footer looks wonky to me

Good catch.. I actually have another PR with footer redesign but until it's done, I'll fix this glitch.

> the color of the author's block top bar doesn't show up as nicely on Night and Hacker themes

Yea, it's because it's custom per profile. User can set own colors and they don't always look nicely depending on the theme. I think it's still better than previous solution where we had custom colors applied to text making it less readable.

app/views/articles/_user_metadata.html.erb
rhymes: oh cool! :D

lisasy:
Screen Shot 2020-06-08 at 11 19 12 AM

Regarding future iterations on the flow, I think it'll be curious to see how we approach comments & replies here and how it impacts this whole section.

lisasy:
Screen Shot 2020-06-08 at 11 21 24 AM

I think it's great to remove the unnecessary author information that we currently include between the end of the article and the start of the discussion. We already have the author-card in the right sidebar, which remains fixed.

lisasy:
Screen Shot 2020-06-08 at 11 22 41 AM

Re: colored tags. An eventual consideration is: would we want to keep the colored tags or move to something more generic (e.g. all tags are just in grey text without any color distinction)? Not anything to decide at this moment, but seeing this made me think about it.

benhalpern:
This looks really great. I love the reaction animation and all the cleaning up of the reading experience! Layout looks really great too.

A few issues



When the number changes width it shoves the other icons over slightly.

The one change I'd ask for is that the reactions on the left sidebar be bumped down a bit. They just seem a little too _uptight_ or something when pinned right to the top of the page.

The equivalent on Medium is about 160px from the top of the page, and I think that gives a bit more balance to things.

Another thing that could be an issue....

The sticky right sidebar will show "cut off" articles that will eventually be visible once the user gets to the bottom of the page (the sticky element is beautiful btw). The current design cuts off articles in their entirety so there isn't an unreachable half article. Currently it's done with Flexbox.



Also, I see that this changes the cover image back to a background image from an `img`. The element was originally changed to an image tag for improved SEO. I don't think this is a _huge_ thing, but when in doubt I think we should keep the status quo in that regard.

At around 430px width with an image in the post, the page has a subtle side-to-side scroll. I think that's because the image is _slightly_ larger than 100%. I think this was a thing that had been fixed with `overflow:hidden` in the past. As one possible way to deal that might be a nicer aesthetic is to frame images a bit. Right now they're edge-to-edge, but maybe if there were a bit of space (and maybe a few px of border radius), it could be a nice look if you want to modify this.

Also there's a quick _jump_ in the height of the `follow` button before the `follow/following` shows up. Maybe modify the default text to be `&nbsp;`?

ludwiczakpawel:
> One small suggestion is in eliminating the grid-gap on small screens

@lisasy I'll see what I can do about it.. Surprisingly it's not so trivial  :) 

> see how we approach comments & replies here and how it impacts this whole section.

Definitely! We should collaborate in future on this.. I have some ideas but would love another pair of eyes for this.

> I think it's great to remove the unnecessary author information that we currently include between the end of the article and the start of the discussion. We already have the author-card in the right sidebar, which remains fixed.

Surprise, surprise - I did not remove it :D There's a logic in our code that hides it IF post content is short (markdown below 900 characters). But I do have similar feelings that we should do something about it but similarly to Comments section - that will come in future.

ludwiczakpawel:
> When the number changes width it shoves the other icons over slightly.

@benhalpern I've noticed that only on smaller breakpoints (and I've just fixed it).. Have you noticed that in desktop resolutions too?

> The one change I'd ask for is that the reactions on the left sidebar be bumped down a bit. They just seem a little too uptight or something when pinned right to the top of the page.

Hmm, I've added `6vh` to the calculation of its position but honestly I'm not sure if I like it... It is now: 

```
var(--header-height) + var(--layout-padding) + 6vh;
```

> The sticky right sidebar will show "cut off" articles that will eventually be visible once the user gets to the bottom of the page (the sticky element is beautiful btw). The current design cuts off articles in their entirety so there isn't an unreachable half article. Currently it's done with Flexbox.

Is that wrong? With this solution users at least can see these cut off articles when they scroll to the bottom.. When hiding them with flexbox, user won't be able to see them at all.. I think this behavior feels more natural but that's just me. Btw. hiding these articles with flexbox will be definitely much harder with current design (when each article is part of widget).

Also, maybe we could **additionally** limit number of articles in the widget so there are up to 3 for each widget? Not sure if having more of them actually drives more traffic on other articles?

> I see that this changes the cover image back to a background image from an img. The element was originally changed to an image tag for improved SEO. I don't think this is a huge thing, but when in doubt I think we should keep the status quo in that regard.

Ok, I brought back `` tag 👍.

---

The scroll with image and `&nbsp;` for buttons should be fixed now as well.

benhalpern:
> My only feedback is I thought the reading column is a bit narrow but I can get used to it.

Yeah I also agree this could be a bit wider, but I wouldn't argue too strongly about this. I could see benefits either way.

...What if we just dropped the "Trending on ..." section altogether to deal with this sticky issue. I think the profile card + more from user boxes suffice and we could revisit this section in general later.

In terms of getting this deployed, I think we'll need to add some styles at the bottom of the file like...

```
   .stories-show .home {
      ....temporary styles to show when old styles are still around. that we can tear out later.
     probably a bunch of !important stuff at the bottom of the file.
     basically the old HTML will still be present for about a day. We can clear *a bunch* of caches, but it's impractical to bust all the caches site-wide.
   }
```

thomasbnt:
Hello, I created a PR on your fork @ludwiczakpawel but I didn't know how to get it notified here. See https://github.com/ludwiczakpawel/dev.to/pull/5

> ## Description
> Just added `:hover` to `.mod-actions-menu-btn`
> 
> ```diff
> .mod-actions-menu-btn {
>   position: fixed;
>   bottom: var(--su-7);
>   right: var(--su-7);
>   z-index: 201;
>   cursor: pointer;
>   transition: 0.3s all ease;
>   border-radius: 50px;
>   
> +  &:hover {
> +    z-index: 201;
> +  }
> 
> [...]
> ```
> ## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
> **Old**	**New**
> image	image

ludwiczakpawel:
> I thought the reading column is a bit narrow but I can get used to it.

Yea, I think it's a matter of getting used to it. Technically ideal line length for readability is around 60 characters per line.. This new design has about 74 characters per line so "technically" still too much but I think it's a good balance altogether. Although I'm more than happy to adjust it a little in future with the cost of right sidebar. And totally agree - it's definitely matter of getting used to it.

> What if we just dropped the "Trending on ..." section altogether to deal with this sticky issue. 

@benhalpern Maybe we drop it IF there's no other posts from that user?  Also, limit "Trending posts" number to max 5. 

Re styles: Im not sure if I understand this..?

nickytonline:
@ludwiczakpawel, I know that the footer is lining up with the width of the article and reaction icons, but I wonder if the footer would look better if it was only the width of the article.

image

ludwiczakpawel:
@nickytonline I'm aware of the footer issues and I intentionally skipped it since I have [this PR](https://github.com/thepracticaldev/dev.to/pull/8684) in queue so I thought it doesn't make much sense to adjust the alignment here.

app/assets/stylesheets/article-show.scss
```diff
@@ -1406,20 +898,18 @@ article {
...
 .show-comments-header {
-  width: 800px;
-  max-width: 90%;
+  // width: 800px;
+  // max-width: 90%;
   margin: auto;
   margin-bottom: -50px;
   text-align: center;
   font-family: var(--ff-default);
 }
```
benhalpern, line 908:
```suggestion
.primary-sticky-nav {
  display: none;
}
.primary-sticky-nav {
  display: none;
}
.article-actions {
  display: none;
}
```

Basically some of the old pages will look weird with the old style removed but old HTML still present. This will hide those elements for the edge cases. There will be some light confusion, but better than styles being off.

app/views/articles/show.html.erb
```diff
@@ -81,143 +81,165 @@
...
-    <% @classic_article = Suggester::Articles::Classic.new(@article, not_ids: [@article.id]).get.first %>
-    <% if @classic_article %>
-      <%= render "additional_content_boxes/article_box",
-                 article: @classic_article,
-                 classification: "classic",
-                 classification_text: "Classic #{community_name} Post from #{@classic_article.readable_publish_date}",
-                 follow_cue: "Follow <a href='#{@classic_article.user.path}'>@#{@classic_article.user.username}</a> to see more of their posts in your feed." %>
+
+  <div class="crayons-layout__sidebar-right">
+    <% stick_nav_cache_key = "sticky-sidebar-#{@article.id}-#{(@organization || @user).profile_updated_at}-#{(@organization || @user).last_article_at}-0-#{@user.pro?}" %>
```
benhalpern, line 236:
```suggestion
    <% stick_nav_cache_key = "sticky-sidebar-#{@article.id}-#{(@organization || @user).profile_updated_at}-#{(@organization || @user).last_article_at}-1-#{@user.pro?}" %>
```

app/views/articles/_sticky_nav.html.erb
```diff
@@ -34,91 +34,99 @@
...
+        <div>
+          <a href="<%= @actor.cta_button_url || "Learn more" %>" class="crayons-btn crayons-btn--outlined w-100">
+            <%= @actor.cta_button_text %>
+          </a>
+        </div>
+      <% end %>
     <% end %>
   </div>
+
   <% cache("article-sticky-nav-articles-#{@article.id}-#{@actor.last_article_at}", expires_in: 48.hours) do %>
```
benhalpern, line 79:
```suggestion
  <% cache("article-sticky-nav-articles-0-#{@article.id}-#{@actor.last_article_at}", expires_in: 48.hours) do %>
```

benhalpern:
@ludwiczakpawel cool. This either/or logic works for me, but we should limit it to 3 IMO.

I made a few comment suggestions which contain some code to bump a couple fragment caches and hide elements in scenarios where the old HTML is still being rendered.

So with all of this I think it's pretty much good to go.

app/assets/stylesheets/comments.scss
```diff
@@ -966,27 +933,26 @@ a.header-link {
...
+        @include generate-box(
+          $level: 2,
+          $bg: var(--card-bg),
+          $border: var(--card-color),
+          $color: var(--card-color)
+        );
         z-index: 20;
-        width: 130px;
-        box-shadow: $shadow;
+        width: 160px;
```
nickytonline, line 947: Just a general question here in regards to width. Why aren't we using `10rem`?

app/assets/stylesheets/components/tags.scss
```diff
@@ -18,7 +18,7 @@
     opacity: 0.5;
   }
 
-  &:hover {
+  &[href]:hover {
```
nickytonline, line 21: Is the addition of `[href]` because not all `.crayons-tag`s are anchor tags?

app/assets/stylesheets/scaffolds.scss
```diff
@@ -37,15 +37,15 @@ body {
 
 .zen-mode {
   .top-bar,
-  .primary-sticky-nav,
-  .article-actions {
+  .crayons-article-sticky,
+  .crayons-article-actions {
     display: none !important;
```
nickytonline, line 42: Do we still need the `important!`?

app/assets/stylesheets/syntax.scss
```diff
@@ -13,7 +13,7 @@ div.inner-comment div.body div.highlight pre.highlight {
   }
 }
 
-$code-highlight-background-color: #29292e;
+$code-highlight-background-color: #08090a;
```
nickytonline, line 16: For a future PR perhaps, but does it make sense to convert these to CSS properties like we've done with many other SASS vars?

app/assets/stylesheets/views/article.scss
```diff
@@ -4,55 +4,83 @@
...
   &__cover {
-    margin: calc(var(--article-padding-y) * -1)
-      calc(var(--article-padding-x) * -1) var(--article-padding-x)
-      calc(var(--article-padding-x) * -1);
-    width: calc(100% + (var(--article-padding-x) * 2));
+    position: relative;
+    width: 100%;
+    padding-top: 42%;
+
+    // &__image {
```
nickytonline, line 32: Remove this if it's not being used.

app/javascript/article-form/components/__tests__/Preview.test.jsx
```diff
@@ -97,7 +97,7 @@ describe('<Preview />', () => {
       />,
     );
 
-    expect(container.find('.crayons-article__tags').text()).toEqual(
+    expect(container.find('.spec-article__tags').text()).toEqual(
```
nickytonline, line 100: cc: @Ridhwana . We'll need to convert this test if this gets merged before our Preact upgrade PRs.

app/views/articles/show.html.erb
nickytonline: Just out of curiousity @benhalpern, what's the reason for changing part of the cache key from `0` to `1`? Is it simply a way to invalidate the cache for the sticky nav by creating a new cache key?

app/assets/stylesheets/syntax.scss
ludwiczakpawel: Probably but not necessarily per theme because syntax highlighter likely won't change no matter what theme. So not global `:root` variables but yea - we can convert them to be CSS variables indeed.

app/assets/stylesheets/scaffolds.scss
ludwiczakpawel: Good question, I don't think we do. I'll check and remove it if it's not needed anymore.

app/assets/stylesheets/components/tags.scss
ludwiczakpawel: Correct.

app/assets/stylesheets/comments.scss
ludwiczakpawel: I haven't thought about it and I have no idea... But thinking about it right now... If we ever change the font size a little bit, do we also want to adjust the layout of some elements (since rem units are based on font size). But I don't have any strong opinions there yet, I'd have to google it a bit more.

Another thing is that it's easier to visualize what "160px" mean instead of "10 rem" mostly because you don't have to do the math.

Lemme know if you have any thoughts or opinions there.

nickytonline: I was thinking more about if people zoom. `rem` and `em` help with that.

nickytonline: As I mentioned it's not specific to this PR, just more of a general question.

ludwiczakpawel: Definitely worth talking about - I'm not very experienced with accessibility and I don't think I know exactly how zooming works.

But that may seem indeed as bigger discussion issue to be solved in separate PR..

ludwiczakpawel:
@benhalpern Ok, I think I've made all the suggested changes and fixes.

After your suggested changes re caching, we should not have an issue like on the screenshot below, is that correct?

image

ludwiczakpawel:
Ouch, looks like the build is failing... Seems like an issue with [limit thingy I've added](https://github.com/thepracticaldev/dev.to/pull/8234/files#diff-dce65e53568bc516c9529e2c53b62de5R92-R113)... Anyone know how to fix/solve this?

app/javascript/article-form/components/__tests__/Preview.test.jsx
Ridhwana: Will keep an 👀 out for it 👍🏽

benhalpern:
@ludwiczakpawel yeah. Instead of what you have, you'll have to go into  `StickyArticleCollection` and change these three values to (`limit` and `sample` to `3`) 

```ruby
  def user_stickies
    author.articles.published.
      limited_column_select.
      tagged_with(article_tags, any: true).
      where.not(id: article.id).order("published_at DESC").
      limit(2)
  end

  def suggested_stickies
    (tag_articles.load + more_articles).sample(8)
  end

  def tag_articles
    @tag_articles ||= Article.published.tagged_with(article_tags, any: true).
      includes(:user).
      where("public_reactions_count > ? OR comments_count > ?", reaction_count_num, comment_count_num).
      where.not(id: article.id).where.not(user_id: article.user_id).
      where("featured_number > ?", 5.days.ago.to_i).
      order(Arel.sql("RANDOM()")).
      limit(8)
  end
```

The suggested changes actually handle the _rerverse_ scenario to what you have displayed there, where the new shell is present, but the old HTML is still around. The above situation is old shell, new HTML.

A one-off blip for some users isn't _so_ terrible. But the situation where people are stick with _totally broken HTML_ until that page is un-cached is worse. So where we stand now is a reasonble deploy. As we solidify crayons and can reach for utility classes, these issues will become much less prevalent over time.

I'd say this is basically ready to go. I definitely want to be online and ready for this deployment though 😄

ludwiczakpawel:
@benhalpern sweet, then consider it done :)

nickytonline:
Looks like there is a conflict @ludwiczakpawel. This should be good to go once that is sorted.

image

benhalpern:
Yeah, my schedule is open tomorrow. @ludwiczakpawel why don't you get this conflict resolve and I'll ping you first thing in the morning for us to get this deployed.

ludwiczakpawel:
@benhalpern cool. I'm ready whenever you are!

PR #8239, opened 2020-06-02
Enforce Uniqueness for Broadcast Titles

### LINKED ISSUE #8070: Ensure that broadcast names are unique
Currently, for the welcome notifications feature, we look for a broadcast based on its name (and whether or not it is `active` or not), for example: `Broadcast.active.find_by!(title: "bananas")`.

However, as we expand the `Broadcast` model to be used for _other_ types of notifications (such as as site-wide notification feature), there is a possibility for multiple broadcasts to be named similarly.

We should enforce uniqueness on a Broadcasts name, and render an appropriate error message to an admin user who tries to create a broadcast with the same name as another. 

We should also make sure that this is well-tested :)

juliannatetreault:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR enforces uniqueness on a Broadcast titles, scoped specifically to `type_of` (announcement or welcome), and renders an appropriate error message to an Admin that tries to create a Broadcast with the same title as another Broadcast of the same `type_of`.

Closes #8070 

When attempting to add a Broadcast of the same `type`, with the same `title`:

Screen Shot 2020-06-02 at 12 01 35 PM
Also note the successful scoping of Broadcasts with the same `title`, but with different `type_of`s.

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Heads-Up Team SRE (@mstruve and @maestromac): I have double-checked that there are no duplicate Broadcast titles, so this should be safe to add and should not cause any uniqueness-related errors when running the migration. :crossed_fingers:

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -83,6 +83,7 @@ Rails/UniqueValidationWithoutIndex:
     - 'app/models/comment.rb'
     - 'app/models/follow.rb'
     - 'app/models/notification.rb'
+    - 'app/models/broadcast.rb'
 
 # Offense count: 33
 # Cop supports --auto-correct.
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
+<<<<<<< HEAD
 ActiveRecord::Schema.define(version: 2020_06_01_121243) do
+=======
+ActiveRecord::Schema.define(version: 2020_06_02_174329) do
+>>>>>>> Add unique index for title and type_of on Broadcasts
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -273,6 +277,7 @@ ActiveRecord::Schema.define(version: 2020_06_01_121243) do
     t.string "title"
     t.string "type_of"
     t.datetime "updated_at"
+    t.index ["title", "type_of"], name: "index_broadcasts_on_title_and_type_of", unique: true
   end
 
   create_table "buffer_updates", force: :cascade do |t|
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,11 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-<<<<<<< HEAD
-ActiveRecord::Schema.define(version: 2020_06_01_121243) do
-=======
 ActiveRecord::Schema.define(version: 2020_06_02_174329) do
->>>>>>> Add unique index for title and type_of on Broadcasts
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
--- a/spec/models/broadcast_spec.rb
+++ b/spec/models/broadcast_spec.rb
@@ -5,6 +5,7 @@ RSpec.describe Broadcast, type: :model do
   it { is_expected.to validate_presence_of(:type_of) }
   it { is_expected.to validate_presence_of(:processed_html) }
   it { is_expected.to validate_inclusion_of(:type_of).in_array(%w[Announcement Welcome]) }
+  it { is_expected.to validate_uniqueness_of(:title).scoped_to(:type_of) }
 
   it { is_expected.to have_many(:notifications) }
 
--- a/spec/requests/internal/broadcasts_spec.rb
+++ b/spec/requests/internal/broadcasts_spec.rb
@@ -112,4 +112,19 @@ RSpec.describe "/internal/broadcasts", type: :request do
       end
     end
   end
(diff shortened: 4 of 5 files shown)


.rubocop_todo.yml
```diff
@@ -83,6 +83,7 @@ Rails/UniqueValidationWithoutIndex:
     - 'app/models/comment.rb'
     - 'app/models/follow.rb'
     - 'app/models/notification.rb'
+    - 'app/models/broadcast.rb'
```
vaidehijoshi, line 86: Not impacting @juliannatetreault's work here at all, but should we consider removing this `Rails/UniqueValidationWithoutIndex` rule if we're ignoring it in so many files? Is it actually enforcing anything?

db/schema.rb
```diff
@@ -268,9 +268,12 @@
   create_table "broadcasts", id: :serial, force: :cascade do |t|
     t.boolean "active", default: false
     t.text "body_markdown"
+    t.datetime "created_at"
     t.text "processed_html"
     t.string "title"
     t.string "type_of"
+    t.datetime "updated_at"
```
vaidehijoshi, line 275: I think this PR needs to be rebased on master; looks like [#8181](https://github.com/thepracticaldev/dev.to/pull/8181/files) added the `created_at` and `updated_at` timestamps recently.

Once it's rebased, those two lines shouldn't show up in the `schema.db`, and it should just be the index, as expected :)

spec/models/broadcast_spec.rb
```diff
@@ -5,6 +5,7 @@
   it { is_expected.to validate_presence_of(:type_of) }
   it { is_expected.to validate_presence_of(:processed_html) }
   it { is_expected.to validate_inclusion_of(:type_of).in_array(%w[Announcement Welcome]) }
+  it { is_expected.to validate_uniqueness_of(:title).scoped_to(:type_of) }
```
vaidehijoshi, line 8: We might want to add a quick test or two to `spec/requests/internal/broadcasts_spec.rb` to test than an admin can't create duplicate Broadcasts, similar to how we test that you can't create [multiple active broadcasts](https://github.com/thepracticaldev/dev.to/blob/a4be185720a1e8067b54668541ca1fcb974413a4/spec/requests/internal/broadcasts_spec.rb#L91-L113).

juliannatetreault: @vaidehijoshi: I've added a test per your request (see below file). 😄

spec/requests/internal/broadcasts_spec.rb
```diff
@@ -112,4 +112,19 @@
       end
     end
   end
+
+  context "with the same title and the same type_of" do
+    let(:super_admin) { create(:user, :super_admin) }
+    let(:params) { { title: "Hello!", processed_html: "<p>Hello!</p>", type_of: "Announcement", active: true } }
```
vaidehijoshi, line 118: One small thing: I think this is actually testing that we can't create two active `Announcement` broadcasts (because we have a validation on that specific case, too!) In fact, you could try this out: if you tried this test out `master`, I suspect it would pass, because of the validation that ensures that only one `Announcement` broadcast is `active` at a time :)

But!!! There are two ways around this, which will make the test more explicit:
1. remove the active: 
```suggestion
    let(:params) { { title: "Hello!", processed_html: "<p>Hello!</p>", type_of: "Announcement" } }
```
2. or, test this against `Welcome` type broadcasts.
```suggestion
    let(:params) { { title: "Hello!", processed_html: "<p>Hello!</p>", type_of: "Welcome", active: true} }
```

I only thought of this because I wrote the logic for the Announcement validation! 😸

spec/models/broadcast_spec.rb
vaidehijoshi: I LUV 2 SEE IT 👀

spec/requests/internal/broadcasts_spec.rb
juliannatetreault: Whoops! Good catch. 

I've made the change!

.rubocop_todo.yml
rhymes: @vaidehijoshi I'm slowly removing all of those in https://github.com/thepracticaldev/dev.to/pulls?q=is%3Apr+author%3Arhymes+%22unique+indexes%22

I'll remove the rule when we're done, it's a pretty recent one :)

rhymes:
@juliannatetreault @vaidehijoshi isn't this PR missing the migration file which modified the schema?

PR #8242, opened 2020-06-02
CI Speed Up: Use Gems in Vendor Cache for Travis

mstruve:
- [x] Optimization

## Description
Rather than wasting up to 3+ minutes installing gems during a Travis run, lets use the ones we have cached in our vendor/cache file by setting the `--local` flag. I also added [`--jobs 2`](https://bundler.io/bundle_install.html) which allows us to use 2 workers in parallel to load the gems.

## Added tests?

alt_text

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,8 @@ before_install:
   - sudo chown -R elasticsearch:elasticsearch /etc/default/elasticsearch
   - sudo systemctl start elasticsearch
 install:
-  - bundle install --path vendor/bundle
+  - bundle config set path 'vendor/bundle'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -48,7 +48,8 @@ before_install:
   - sudo chown -R elasticsearch:elasticsearch /etc/default/elasticsearch
   - sudo systemctl start elasticsearch
 install:
-  - bundle install --path vendor/bundle
+  - bundle config set path 'vendor/bundle'
```
mstruve, line 51: This was moved bc of a deprecation warning that stated this is the new expected way to set the path.

PR #8244, opened 2020-06-03
Add vault-rails gem

citizen428:

- [X] Feature

## Description

We're currently leaning towards introducing Vault for secret management, this draft PR gives @jdoss something to play with.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed

alt_text


jdoss:
@mstruve is checking this out too. Let's try and sync efforts here soon.

citizen428:
@jdoss For sure. This PR only addds the gem and a basic initializer so far.

mstruve:
After doing some work with this yesterday I am pretty sure we don't need vault-rails, we only need the vault Ruby Gem. This gem is what we would want if we were using vault to encrypt fields on our ActiveRecord models. Instead, we want to use it to store auth and third party tokens.

citizen428:
@mstruve Good point, changed this PR to use `vault-ruby`.

mstruve:
Closing in favor of https://github.com/thepracticaldev/dev.to/pull/8891

PR #8246, opened 2020-06-03
Make Devise.mailer_sender configurable

### LINKED ISSUE #8245: Generalize Devise mailer config
Replace hardcoded `config.mailer_sender` in devise initializer with a configurable one.

citizen428:

- [X] Refactor

## Description

Our `Devise.mailer_sender` was hardcoded to `dev.to <yo@dev.to>`. This PR changes it to `#{ENV["COMMUNITY_NAME"]} <#{ENV["DEFAULT_EMAIL"]}>` (so `DEV <yo@dev.to>`) as a first attempt at generalizing this. There's more we could do here, but unless we have a comprehensive strategy around `SiteConfig`, default values, etc. this seems sufficient.

Closes #8245 

## Added tests?

## Added to documentation?

- [X] no documentation needed

* Set the `DEFAULT_EMAIL` env variable to `yo@dev.to`. Other communities need to do this as well.

### DIFF
--- a/Envfile
+++ b/Envfile
@@ -15,6 +15,9 @@ variable :APP_PROTOCOL, :String, default: "http://"
 variable :COMMUNITY_COPYRIGHT_START_YEAR, :String, default: "2016"
 variable :COMMUNITY_NAME, :String, default: "DEV(local)"
 
+# Email related variables
+variable :DEFAULT_EMAIL, :String, default: "yo@dev.to"
+
 # Service timeout length
 variable :RACK_TIMEOUT_WAIT_TIMEOUT, :Integer, default: 100_000
 variable :RACK_TIMEOUT_SERVICE_TIMEOUT, :Integer, default: 100_000
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -11,7 +11,7 @@ Devise.setup do |config|
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class
   # with default "from" parameter.
-  config.mailer_sender = "dev.to <yo@dev.to>"
+  config.mailer_sender = "#{ENV['COMMUNITY_NAME']} <#{ENV['DEFAULT_EMAIL']}>"
 
   # Configure the class responsible to send e-mails.
   # config.mailer = 'Devise::Mailer'
--- a/spec/requests/internal/configs_spec.rb
+++ b/spec/requests/internal/configs_spec.rb
@@ -41,7 +41,7 @@ RSpec.describe "/internal/config", type: :request do
 
       describe "API tokens" do
         it "updates the health_check_token" do
-          token = "#{rand(20)}"
+          token = rand(20).to_s
           post "/internal/config", params: { site_config: { health_check_token: token }, confirmation: confirmation_message }
           expect(SiteConfig.health_check_token).to eq token
         end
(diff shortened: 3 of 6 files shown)


Envfile
```diff
@@ -15,6 +15,9 @@ variable :APP_PROTOCOL, :String, default: "http://"
 variable :COMMUNITY_COPYRIGHT_START_YEAR, :String, default: "2016"
 variable :COMMUNITY_NAME, :String, default: "DEV(local)"
 
+# Email related variables
+variable :DEFAULT_EMAIL, :String, default: "yo@dev.to"
```
Ridhwana, line 19: Hmm It feels super weird that we're setting a default email in the environment and also in the site config. Is there no way we can use it from the Site config, or should we take it out of the site config and use the default email from the ENV (in that case we'll still be keeping alternate emails in the site config)? Just throwing some thoughts going through my head out.

citizen428: You are correct. I confused myself. Let me updates this quickly.

citizen428: You cannot use `SiteConfig` in initializers. If you try to do that you'll get the following error:

```ruby
You cannot use settings before Rails initialize. (RuntimeError)
```

How about updating `SiteConfig` like this:

```ruby
  field :email_addresses, type: :hash, default: {
    default: ENV["DEFAULT_EMAIL"],
    business: "partners@dev.to",
    privacy: "privacy@dev.to",
    members: "members@dev.to"
  }
```

Not sure if that makes more sense though...

mstruve: Yeah SiteConfig can be finicky, only took down the app a couple of times for us to figure that one out 🙃 

@citizen428 I love the idea of making the ENV variable the default!

Ridhwana: Making the ENV variable the default is a neat idea, however, I can easily see it getting out of sync and confusing the users. 
For example, a user updates their default email in the site config variable and thinks that the default email is changed throughout the site, hence they expect the emails to be sending from the new default email address that they've set in the site config.

I'm leaning towards less ambiguity by in keeping it set in one place with the drawback that its isolated from the other additional emails. i.e. setting the default email in the ENV variable and changing the Site config section to read Additional emails with a note that states that the default email is set as an ENV variable.

Ridhwana: In addition, we can make the default email in the site config a readonly field so that they can see what the value is set to.

Ridhwana: ^^ _updated my comment_

citizen428: @Ridhwana Good call!

I now did the following: 

1. Change the internal mod page:
    Screen Shot 2020-06-04 at 14 02 07
2. Updated the params for the controller to not accept the default email:
    ```ruby
    SiteConfig.email_addresses.except(:default).keys
    ```
3. Default `SiteConfig.email_addresses[:default]` to the ENV var:
    ```ruby
    field :email_addresses, type: :hash, default: {
      default: ApplicationConfig["DEFAULT_EMAIL"],
      business: "partners@dev.to",
      privacy: "privacy@dev.to",
      members: "members@dev.to"
    }
```

Theoretically, you could still set the value in a Rails console with `SiteConfig.email_addresses[:default] = ...`, but preventing that would be too hacky. 

Better?

Comment (Ridhwana) [reply Envfile:19]: @citizen428 I like this much better 🎉. I have a question follow up question - why are we keeping the default email in the Site Config Hash, could we not remove it completely and just use the ENV variable wherever we referencing the default email address like we doing now in show.html.erb ?

File: spec/requests/internal/configs_spec.rb [hunk 2]
```diff
@@ -97,17 +97,15 @@
       describe "Emails" do
         it "updates email_addresses" do
           expected_email_addresses = {
-            default: "foo@bar.to",
-            business: "partners@dev.to",
-            privacy: "privacy@bar.to",
-            members: "members@bar.to"
+            business: "partners@example.com",
```
citizen428, line 100: Since I updated the spec I changed the domain to `example.com` which is a [IANA reserved domain](https://www.iana.org/domains/reserved) unlike `bar.to`.

config/initializers/devise.rb
```diff
@@ -11,7 +11,7 @@
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class
   # with default "from" parameter.
-  config.mailer_sender = "dev.to <yo@dev.to>"
+  config.mailer_sender = "#{ENV['COMMUNITY_NAME']} <#{ENV['DEFAULT_EMAIL']}>"
```
mstruve, line 14: 🔥

Envfile
citizen428: @Ridhwana I could pretend that I really like the consistency of having all the emails in one place, but to be honest I did a quick grep and was too lazy to make all the changes right now.

I think what we currently have is safe enough™ and with the ongoing generalization work, we may completely remove all default values anyway.

Ridhwana: Lol kudos for your honestly 😂 This sounds fine to me though, we can make it an iterative approach :)

citizen428:
Had to re-run the specs since this had hit the flaky Pundit one.

rhymes:
@citizen428 does this require any pre or post deploy manual work for us or other communities?

Ridhwana:
Other communities will most definitely have to set this environment variable to ensure their devise mail settings don't bork

citizen428:
I may have accidentally re-requested reviews because it's Monday morning the :coffee: hasn't licked in yet, sorry for that.

PR #8250, opened 2020-06-03
Generalize search.xml

### LINKED ISSUE #8247: Make OpenSearch description configurable
We haven an [OpenSearch description file](https://developer.mozilla.org/en-US/docs/Web/OpenSearch) in `public/search.xml`. This is currently hard-coded to DEV specific values and several attributes should be generalized for other communities.

```xml
<?xml version="1.0" encoding="UTF-8"?>
 <OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
   <ShortName>DEV Search</ShortName>
   <Description>Find posts from the DEV community.</Description>
   <Tags>software development</Tags>
   <Contact>yo@dev.to</Contact>
   <Url type="text/html"
        template="https://dev.to/search?q={searchTerms}"/>
 </OpenSearchDescription>
```

citizen428:

- [X] Generalization

## Description

This PR generalizes our [OpenSearch specification](https://developer.mozilla.org/en-US/docs/Web/OpenSearch). The file moved from `dev.to/search.xml` to `dev.to/assets/search.xml` but this should matter since I updated the `link` accordingly. Firefox still recognizes the search engine:

Screen Shot 2020-06-03 at 14 52 55

Closes #8247

## Added tests?

## Added to documentation?

- [X] no documentation needed

### DIFF
--- a/app/assets/xml/search.xml.erb
+++ b/app/assets/xml/search.xml.erb
@@ -1,0 +1,9 @@
+<% helper = Module.new { extend ApplicationHelper } %>
+<?xml version="1.0" encoding="UTF-8"?>
+ <OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+   <ShortName><%= helper.community_name %> Search</ShortName>
+   <Description>Find posts from the <%= helper.community_qualified_name %>.</Description>
+   <Contact><%= SiteConfig.email_addresses[:default] %></Contact>
+   <Url type="text/html" 
+     template="<%= URL.url("search") %>?q={searchTerms}"/>
+ </OpenSearchDescription>
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -7,3 +7,4 @@
 // = link administrate/application.css
 // = link administrate/application.js
 // = link katex.css
+// = link search.xml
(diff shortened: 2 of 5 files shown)


app/assets/xml/search.xml.erb
```diff
@@ -0,0 +1,9 @@
+<% helper = Module.new { extend ApplicationHelper } %>
```
citizen428, line 1: A little hack to make the helpers available here so we can reuse `community_name` and `community_qualified_name` to make sure they stay in sync with the rest of the app.

app/assets/xml/search.xml.erb
```diff
@@ -0,0 +1,9 @@
...
+<?xml version="1.0" encoding="UTF-8"?>
+ <OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+   <ShortName><%= helper.community_name %> Search</ShortName>
+   <Description>Find posts from the <%= helper.community_qualified_name %>.</Description>
```
citizen428, line 5: We previously had a `<Tags>` attribute with the value "software development" here. I removed this for now, since it's optional and the current one wouldn't work across communities. I was thinking about using `SiteConfig.suggested_tags`, but not sure if that's really what we want.

Issue #8158, opened 2020-05-29
PWA error on icons


lomby92:
**Describe the bug**
The PWA app is not installable since the `manifest.json` has some errors on links used for icons.

**To Reproduce**
Go with Chrome (tested on Android) into `https://dev.to` and see that the prompt for install the PWA does not appear

**Expected behavior**
The prompt should be visible

**Additional context**

I have tried to directly access the second part of the URL of the 512px icon: https://practicaldev-herokuapp-com.freetls.fastly.net/assets/devlogo-pwa-512.png

It works so i think that the error could be of: https://res.cloudinary.com

jacobherrington:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

fdoxyz:
Thank you again for the report lomby92. I'm checking this out now and it does seem like you're right about the problem in the manifest. I'll follow up on this issue soon.

fdoxyz:
Hi @lomby92, we merged a fix for the image links thanks again for reporting this!

After giving it some time (because of caching) I'm getting the correct links in the manifest, but I still can't see the PWA prompt with my Galaxy S8 (Android 9). If you're able to check again I'd appreciate it, to confirm the issue is closed.

Also a bit out of topic but, I'd also recommend giving [the native app](https://dev.to/downloads) a try if you haven't already too 😄

lomby92:
Thanks @fdoxyz, I can confirm that now it works. (Tested both on Chrome Desktop on Linux and Chrome on Android 9)

PR #8253, opened 2020-06-03
Fixes Cloudinary helper calls that specify custom format

fdoxyz:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The [cloudinary helper](https://github.com/fdoxyz/dev.to/blob/master/app/helpers/application_helper.rb#L77) method has 1 required parameter (url) and 3 optional parameters (width, quality & format). 

Since these are positional params (not [keyword arguments](https://thoughtbot.com/blog/ruby-2-keyword-arguments)) if we want to specify the **format** (4th param) we must also pass in all 3 other params, otherwise there will be a mismatch in them which was causing this bug.

To see this in action check out the `manifest.json` from production and all 3 images have wrong arguments passed in to Cloudinary, which causes this error:

Screen Shot 2020-06-03 at 08 18 20

With this fix the URLs generated now have the following valid params: `c_limit,f_png,fl_progressive,q_80,w_192`

This error surfaced because of #8158 which mentioned PWA support was failing due to bad image links

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

aha!

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -396,7 +396,7 @@ class StoriesController < ApplicationController
         "logo": {
           "@context": "http://schema.org",
           "@type": "ImageObject",
-          "url": ApplicationController.helpers.cloudinary(SiteConfig.logo_png, 192, "png"),
+          "url": ApplicationController.helpers.cloudinary(SiteConfig.logo_png, 192, 80, "png"),
           "width": "192",
           "height": "192"
         }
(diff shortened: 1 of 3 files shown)


rhymes:
@fdoxyz great find! Let me know if I can close https://github.com/thepracticaldev/dev.to/issues/8158 if the issue is fixed

fdoxyz:
Thanks @rhymes! I'll write a follow up on the issue, because the images are working now but I can't seem to find the PWA bannner on my device.

Issue #8238, opened 2020-06-02
Text out of bounds in Search results feed 


madarsbiss:
**Describe the bug**

Today I was using DEV search and noticed text out of bounds in one of the results.

**To Reproduce**

Search: https://dev.to/search?q=project%20ideas
Scroll down till article below shows up (should be around 20th-25th result).

**Screenshots**
![Screenshot](https://user-images.githubusercontent.com/42511917/83550981-5efe3800-a510-11ea-98c2-b33f0fd6c4bb.png)

**Desktop (please complete the following information):**

- OS: Win10 x64
- Browser: Chrome
- Version: 83.0.4103.61

I did check on Firefox too (v 77.0), the problem still exists.

**Smartphone (please complete the following information):**

- Device:
- OS:
- Browser:
- Version:

**Additional context**

Ridhwana:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

icncsx:
Looks like the issue is the non-breaking space: `&nbsp`. 

![Screenshot from 2020-06-03 12-53-40](https://user-images.githubusercontent.com/60417145/83678999-a3d7ac80-a5ac-11ea-97d4-4d9b55b81324.png)

@Ridhwana, it looks like the overwhelming majority of articles don't have that issue. Do we want to handle this server side or client side?

If we want to change it client side, it's a pretty trivial fix on line 221 [here](https://github.com/thepracticaldev/dev.to/blob/master/app/assets/javascripts/utilities/buildArticleHTML.js#L221). 

```js
bodyTextSnippet =
        startingEllipsis + article.highlight.body_text.map(x=>x..replace(/&nbsp;/g, " ")).join('...') + '…';
```

If we want to change it server side, it really depends on how downstream or upstream we want to 
[...]

Issue #7351, opened 2020-04-17
Liquid comment in notifications list bad


qm3ster:
**Describe the bug**

![Annotation 2020-04-17 031307](https://user-images.githubusercontent.com/10551234/79518498-e9223800-8059-11ea-8390-f0fff13d1451.png)

**To Reproduce**

1. Have a notification that shows a comment, for example a reply notification, wherein the text contains a liquid tag for another comment on dev.to
2. Have a viewport width of over `1120px`, so that `#articles-list` is no longer full-width, but fixed width instead.
3. At viewport widths of `760px` and over, `.liquid-comment` becomes `width: 620px` which kills the crab.
4. Removing that fixes notifications view but breaks the normal views.

**Expected behavior**

![Annotation 2020-04-17 031526](https://user-images.githubusercontent.com/10551234/79518496-e7f10b00-8059-11ea-8e64-ff2267a644f7.png)

**Desktop (please complete the following information):**

- OS: Classified
- Browser: Chromium
- Version: Classified

**How to actually fix**

Use flexbox for everything. Even post body. *Especially* post body.

qm3ster:
I swear I saw a similar issue, but I can't find it in search again :(
If anyone knows how to find it, please link this there.

citizen428:
Thanks for the issue @qm3ster! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

rhymes:
There might be some helpful clues in this other issue https://github.com/thepracticaldev/dev.to/issues/7170

qm3ster:
The issue I believe I saw had the words "liquid" "comment" and "overflow" in the title, but I can't find anything with those words or their permutations. I am shooketh.

ludwiczakpawel:
I think this should be now resolved (or at least outdated since we have updated Notifications center).

Issue #7170, opened 2020-04-08
Long web URL goes out of text box


bernardbaker:
**Describe the bug**
I was viewing my notifications. And then I scrolled down to a reply. There was a URL in the box and the text bled out of the box and out of the viewport. I was using a mobile phone.

**To Reproduce**

Click the bell on mobile.
Find a notification which has a long URL in it.
See if it bleeds out of the text box.

**Expected behavior**

The word should wrap around to the next line.

**Screenshots**

![Screenshot_20200408-235842](https://user-images.githubusercontent.com/13556172/78841724-06ba2680-79f6-11ea-82f0-40f8c71d136f.png)

**Desktop (please complete the following information):**

- OS: 
- Browser: 
- Version: 

**Smartphone (please complete the following information):**

- Device: Pixel 3 XL
- OS: ANDROID
- Browser: CHROME
- Version: ??

**Additional context**

citizen428:
Thanks for the issue @bernardbaker! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

pedrowindisch:
This problem also happens on the desktop. Working around the website, I found two solutions to this issue. 

We can fix it by applying a `word-break: break-word` property on the text. It's a quick and easy solution, but if the URL is too big, the notification can get really large. Take a look at the screenshot below.

![dev to link](https://user-images.githubusercontent.com/30203228/78855398-2eab8900-79fa-11ea-8bc0-3b2b4840c720.png)

The second solution I found requires two properties, `overflow: hidden` and `text-overflow: ellipsis`. This is how it looks like:

![dev to link-2](https://user-images.githubusercontent.com/30203228/78855840-6404a680-79fb-11ea-8183-943e59018fc6.png)

Both solutions fix the problem on all platforms and are [widely supported](https://caniuse.com/#search=text-overflow). I find the second fix better and I'd be more than happy to implement it to the codebase.

rhymes:
I think this might be related, I just noticed it this morning

[...]

PR #8263, opened 2020-06-04
Fix text out of bounds in Search results feed

icncsx:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Issue #8238 is caused by the non-breaking space (`&nbsp`). See below.

nbsp

To reproduce the issue on production server, follow the directions:
- Search: https://dev.to/search?q=project%20ideas
- Scroll down till article below shows up (should be around 20th-25th result).

Screenshot from 2020-06-03 21-50-24

I originally thought about 3 ways to solve this issue.

### Client side option

If we want to change it client side, we need to do something on line 221 [here](https://github.com/thepracticaldev/dev.to/blob/master/app/assets/javascripts/utilities/buildArticleHTML.js#L221) where we just replace `&nbsp` with `" "`. 

```js
// something like this
bodyTextSnippet =
        startingEllipsis + article.highlight.body_text.map(x=>x.replace(/&nbsp;/g, " ")).join('...') + '…';
```

### Server side option

If we want to change it server side, it really depends on how downstream or upstream we want to 
gsub out the `&nbsp`. Upon cursory inspection, a good candidate seemed to be the `prepare_doc` method in [feed_content.rb](https://github.com/thepracticaldev/dev.to/blob/master/app/services/search/feed_content.rb#L22).

```rb
def prepare_doc(hit)
        source = hit.dig("_source")
        hit.dig("highlight", "body_text").map! { |p| p.gsub(/\&nbsp;/, " ") }
        source["highlight"] = hit["highlight"]
```

But really the most simple solution is to change the CSS.

### CSS option

```css
  &__snippet {
    overflow-wrap: break-word;
  }
```

This solution is what this PR implements. Note that this solution is consistent with how we treat `.container.body` which holds the `body_text`.

container-body

After break-word is applied:

break-word

### What caused this?

I wrote a test article entitled 

```md
"THIS IS A TEST &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; TEST &nbsp;&nbsp;&nbsp;"
```

and I got back this:

test

The root issue, then, is that some people will choose to write their titles and articles with `&nbsp` instead of `" "`. I don't have the production DB in front of me to procure the statistic, but I would say that it's an extremely low sample size - those people that choose to encode their white space explicitly. 

## Added tests?

- [ ] yes
- [x] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/assets/stylesheets/components/stories.scss
+++ b/app/assets/stylesheets/components/stories.scss
@@ -187,6 +187,7 @@
       color: var(--card-color);
       background: var(--accent-warning-lighter);
     }
+    overflow-wrap: break-word;
   }
 
   &__headline {
--- a/app/services/search/feed_content.rb
+++ b/app/services/search/feed_content.rb
@@ -26,7 +26,6 @@ module Search
         source["tag_list"] = source["tags"]&.map { |t| t["name"] } || []
         source["flare_tag"] = source["flare_tag_hash"]
         source["user_id"] = source.dig("user", "id")
-        hit.dig("highlight", "body_text").map! { |p| p.gsub(/\&nbsp;/, " ") }
         source["highlight"] = hit["highlight"]
         source["readable_publish_date"] = source["readable_publish_date_string"]
         source["podcast"] = {
--- a/app/services/search/feed_content.rb
+++ b/app/services/search/feed_content.rb
@@ -26,6 +26,7 @@ module Search
         source["tag_list"] = source["tags"]&.map { |t| t["name"] } || []
         source["flare_tag"] = source["flare_tag_hash"]
         source["user_id"] = source.dig("user", "id")
+        hit.dig("highlight", "body_text").map! { |p| p.gsub(/\&nbsp;/, " ") }
         source["highlight"] = hit["highlight"]
         source["readable_publish_date"] = source["readable_publish_date_string"]
         source["podcast"] = {


ludwiczakpawel:
Oh that's a good catch.. I can't give feedback on backend solutions since I'm not an expert. I'm not sure how we ended up having a post with `&nbsp;` instead of actual white spaces.. I guess it was weirdly imported from other source? Not sure..

But your CSS suggestion should work I think... But it would work only on excerpt (which is on search results only) and not on the other example you showed (with long title). So this may need a separate fix. Do you think you could add that in your PR?

Ridhwana:
Thank you for the verbose write-up @icncsx, it really helps to weigh all the options.

This seems like a good solution to me.  Since this is not a common occurrence, choosing the simplest solution seems like a good idea to me. I'd love other opinions from the team on this though.

icncsx:
@ludwiczakpawel @Ridhwana I agree that we need some cleaning server side so that we don't run into more of these kinds of articles. That can either be a separate linked issue or the scope of this issue and accompanying PR can expand to address both needs: the need to format already existing articles that might have this issue and the need to process future article titles and their body text upon posting. And the CSS fix seems to address the former.

Ridhwana:
Thanks @icncsx. I'm just discussing this internally with the team. I'll feedback to you on this PR asap.

icncsx:
@Ridhwana @ludwiczakpawel I just realized that this issue (#8238) could be related to: #7351 and #7170.

ludwiczakpawel:
#7170 is definitely a different issue and it [may have been fixed already](https://github.com/thepracticaldev/dev.to/issues/7170#issuecomment-614710083).

#7351 seems like an unrelated issue at first glance, but I haven't been able to investigate that yet.

Generally I think none of these issue are related (it would be too easy! :))

PR #8280, opened 2020-06-04
Bug Fix: Change Cache Key to Bust Cache

mstruve:
- [x] Bug Fix

Fixes: https://app.honeybadger.io/projects/66984/faults/64426845 which was caused by caching ActiveRecord objects that are not compatible with Rails 6. This changes the cache key so we start fresh. YES it is on my todo list to remove all of the AR object caching, I just haven't gotten to it yet. https://github.com/thepracticaldev/dev.to/projects/9#card-39519016

alt_text

### DIFF
--- a/app/controllers/reactions_controller.rb
+++ b/app/controllers/reactions_controller.rb
@@ -96,7 +96,7 @@ class ReactionsController < ApplicationController
   end
 
   def cached_user_public_reactions(user)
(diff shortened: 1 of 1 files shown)


mstruve:
Flaky test strikes again 😝

rhymes:
@mstruve omg! sorry, I really thought they were compatible with my manual marshal test, I should have created a cache of objects with Rails 5 and then booted on Rails 6 :(

PR #8282, opened 2020-06-04
Feed Layout Fix

### LINKED ISSUE #8278: Central articles overflow to right
**Describe the bug**
The central articles column overflows completely to the right, without an overflow scroll bar.  
This hides the right sidebar completely, and most of the article text is also obscured.

The page seems to load initially fine, with the skeleton loaders in the correct central position and not overflowing. However, when articles load in, the central columns seems to overflow, with article sizes increasing as well.

Error usually persists, but sometimes the central column also loads properly.

**To Reproduce**
1. Load Dev.to page
2. See error

**Expected behavior**
Central article columns remains in the center and does not overflow.

**Screenshots**

Screenshot (5)

- OS: Windows 10
- Browser: Chrome
- Version: v83

Seems to work fine on my phone.

- Device: - 
- OS: - 
- Browser: -
- Version: -

**Additional context**
Bug is reproducible, but not 100% of the time.

ludwiczakpawel:

- [x] Bug Fix

## Description

This PR fixes an issue with layout on feed which would cause expanding middle column exceeding its designated area...

- Closes #8277 
- Closes #8278 

## Added tests?

## Added to documentation?

- [x] no documentation needed

- [ ] @benhalpern to remove _hack_ from internal


vaidehijoshi:
@ludwiczakpawel i restarted your build since it was failing! i'm not sure what is the best way to QA this (how did you repro the bug/QA it locally?). none of the css seems overtly problematic to me, but i can't be sure if it fixes the bug :) what would you recommend to test?

ludwiczakpawel:
@vaidehijoshi I tried to manually test CSS changes on prod with web inspector... Very suboptimal way of testing but we don't have better for now.

If you wanna test it locally, you'd have to:

1. Make comments show up in the feed locally.
2. Make sure one of the visible comments contain code block with longer piece of code.

PR #8259, opened 2020-06-03
Display Jobs Banner if User searches jobs or job

mstruve:
## What type of PR is this? (check all applicable)
- [x] Feature

## Description
Earlier, I went to go look up one of our posted jobs at DEV. I typed "dev.to jobs" into Chrome and it landed me on dev.to's search page with the search query "jobs" which wasn't quite what I was going for. I wondered how many others might be looking for jobs AT DEV and end up where I did so I brought it up in #hiring and we can up with this little fix. Thanks @ludwiczakpawel for the design and Gracie for the copy. 

We have the ability to turn this on or off using SiteConfig. 

Screen Shot 2020-06-03 at 1 57 53 PM

## Added tests?
- [x] yes

alt_text

PR #8283, opened 2020-06-04
Add q Back to Whitelist Params for Fastly

mstruve:
- [x] Bug Fix

## Description
Earlier I merged [a PR](#8259) that tried to make use of `params[:q]` from the query. This however is not present bc this `/search` is not exempt from params whitelisting and q is not on the params list. This changes that so we can use that `params[:q]` to decide whether or not to show the jobs banner. 
 
## Added tests?

alt_text

### DIFF
--- a/config/fastly/snippets/safe_params_list.vcl
+++ b/config/fastly/snippets/safe_params_list.vcl
@@ -1,7 +1,7 @@
 import querystring;
 sub vcl_recv {
     # return this URL with only the parameters that match this regular expression
-    if (req.url !~ "/internal/" && req.url !~ "/search/" && req.url !~ "/bulk_show") {
+    if (req.url !~ "/internal/" && req.url !~ "/search" && req.url !~ "/bulk_show") {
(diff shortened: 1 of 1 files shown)


benhalpern:
I'm going to close this because the intended behavior is that the first request is always cached and `q` is ignored.

Here is the rationale which I dropped in a Slack convo...


> The part where we don't use that param is intentional since we do the search via a JS fetch and therefore we don't have to go to the origin twice for the info.
Otherwise we could probably allow the q to be passed as a parameter.
> This is the intended user experience and I don't think we'd want to change it unless we made the choice to overhaul everything and render all search pages from the server. (But I think that would be going backwards in terms of ideal UX and where the web is going)

I'm happy to open the convo, but this would be a bigger change than this, so just going to close this for now.

mstruve:
Ah shit, I'm slow at this stuff but I see what you are saying now. What if we whitelisted, `q=job`, (not even sure if we can) then have that cached partial have the banner and anything else wont. Otherwise we will have to toggle it with js I guess when we load the results, actually that is probably easiest.

PR #8284, opened 2020-06-04
Document the usage of Audit:Logger

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

One of the DEV maintainers pointed out that our action logging mechanism
isn't documented anywhere. This add some context for that code!

(Thanks @atsmith813)

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [x] docs.dev.to
- [ ] readme
- [ ] no documentation needed

### DIFF
--- a/docs/backend/audit-log.md
+++ b/docs/backend/audit-log.md
@@ -1,0 +1,45 @@
+---
+title: Audit Log
+---
+
+# Audit Log
+
+To help maintain accountability for users with elevated permissions the DEV
+application has a special model that records certain actions.
+
+For example, when a user with the `trusted` role creates a negative reaction on
+an article a record is created with certain information about that action.
+
+That record (which we call an `AuditLog`) looks something like this:
+
+```ruby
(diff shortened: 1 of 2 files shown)


atsmith813:
I'm not sure where this would belong, but 2 things I noticed when I was implementing it:
- I didn't realize that `:moderator` and `:internal` were the 2 options when calling `Audit::Logger.log()`. Maybe we add a blurb about that? And/or when it's appropriate to create a new one? It looks like we aren't quite sure yet on how we want to use that so maybe that would be preemptive.
- As it should be, I eventually realized only ones tagged as `:moderator` will show up in /internal under `Moderator Actions`. Maybe this is worth just a quick footnote elsewhere? (Do we even document `/internal`? 😅 )

jacobherrington:
@atsmith813 I'll make a PR to start documenting how the internal tools work. That's a great idea. :+1:

PR #8286, opened 2020-06-04
Stub Percy.snapshot During Specs

mstruve:
- [x] Optimization

## Description
I did some digging into Percy and found that even when we are not running the agent, [`Percy.snapshot` will still attempt to make a request to the agent every time it is called.](https://github.com/percy/percy-capybara/blob/ffe355ef3e2c5544c9a1215eb03d463ada5cc450/lib/percy.rb#L109) which can waste time. It's not a ton but its something. This stubs that request and eliminates the HTTP call. 

What do you all think? Honestly, I want to remove it to try and simplify our spec suite as I try and make the leap to parallel builds but I also didn't want to shut this all down completely if we want to use it in the future. 

Requests being made during a build
Screen Shot 2020-06-04 at 5 11 55 PM

## Added tests?

alt_text

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -82,7 +82,8 @@ RSpec.configure do |config|
   end
 
   config.before do
-    Sidekiq::Worker.clear_all # worker jobs shouldn't linger around between tests
+    # Worker jobs shouldn't linger around between tests
+    Sidekiq::Worker.clear_all
   end
 
   config.around(:each, elasticsearch_reset: true) do |example|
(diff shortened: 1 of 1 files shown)


rhymes:
@vaidehijoshi I have the same feeling TBH. Our specs are slow and have ordering issues at present. Until we make them faster and truly randomizable and parallelizable, it's possible that Percy is going to be an effort of putting out fires left and right. I still can't run most system tests locally for example. Have no idea why but at the same time, pushing a broken test, waiting 20 minutes to find out it's broken, is an issue.

Trivia: with `rspec --bisect` I once ended up with a failing tests rendering data that was NOT in the test nor in the Ruby code that was generating the HTML. Like ghost data. 👀 

I'm actually surprised that Percy has been more or less working so far :D

PR #8287, opened 2020-06-04
Remove Explict Gemfile from travis.yml to Allow vendor/cache to be used

mstruve:
- [x] Bug Fix

## Description
The addition of this line seems to be preventing bundler from using `vendor/cache` and causing it to install all of the gems for new builds so I removed it. These lines were added in our Rails 6 PR #7658 but based [on the comment](https://github.com/thepracticaldev/dev.to/pull/7658/files#r432139357) and the fact that the specs run green I dont think we need it.

alt_text

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,6 @@ addons:
       - $(ls tmp/screenshots/*.png | tr "\n" ":")
       - $(ls tmp/csvs/*.csv | tr "\n" ":")
     debug: true
-gemfile:
-- Gemfile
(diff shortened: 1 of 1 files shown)


rhymes:
I totally agree with both, that was a leftover mostly to remind myself to setup "Gemfile.next" again in the midterm, but I made a note in my todo list and I agree that until we have faster specs, having two sets of slow specs isn't going to help anyone :D 

cc @citizen428 @mstruve

Issue #8185, opened 2020-05-30
Adapt ratio for asciinema tag or support custom ratio parameters


davinkevin:
**Is your feature request related to a problem? Please describe.**

When using an asciinema tag, the heigh/width is hardcoded and doesn't match the terminal size used during recording.
For example, this https://asciinema.org/a/335281 one, when embedded has a strangely rendered when showing `vim`

![image](https://user-images.githubusercontent.com/1970922/83330016-d9b72100-a28c-11ea-8ad7-801d9d9c998f.png)

Everything in the red square is added because the ratio of the terminal is not kept after inclusion.

**Describe the solution you'd like**

LiquidTag could fetch the ratio of the terminal and include it to fit the content. Or, as a user, I can include values in liquidtag to be used to populate https://github.com/thepracticaldev/dev.to/pull/4152/files#diff-992801ce0976336128e2e2268d8a18a6R6 values.

**Describe alternatives you've considered**

We can record "only for dev.to", by using the ratio used by the dev.to website. But sometimes this is not possible (cross posting on other blog with other layout) and this could lead to the same problem if dev.to change its design.

**Additional context**

I think all is in this issue. Thanks for your help and for your great product ! 

/cc @ruizalexandre

davinkevin:
In some context, this leads to totally wrong result: 

![image](https://user-images.githubusercontent.com/1970922/83330653-dfaf0100-a290-11ea-9c31-b0859eeb28d4.png)

When embedding this one https://asciinema.org/a/335296

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

citizen428:
I'll be closing this issue now as #8339 has now been merged. @davinkevin please try it out and feel free to reopen this issue if you still have problems with Asciinema embeds.

PR #8288, opened 2020-06-05
Also allow Asciinema URLs, not only IDs

citizen428:

- [X] Feature

## Description

While working on #8185 I noticed a few things about the Asciinema tag that I wanted to change. Since these changes make sense by themselves, I didn't want to hold them back cause of my personal struggly with CSS.

Somewhat related to #8185, but doesn't close it.

## Added tests?

- [X] yes

## Added to documentation?

- [X] editor guide

### DIFF
--- a/app/liquid_tags/asciinema_tag.rb
+++ b/app/liquid_tags/asciinema_tag.rb
@@ -1,5 +1,6 @@
 class AsciinemaTag < LiquidTagBase
   PARTIAL = "liquids/asciinema".freeze
+  ASCIINEMA_URL_REGEX = %r{https://asciinema.org/a/(?<id>\d+)}.freeze
 
   def initialize(tag_name, id, tokens)
     super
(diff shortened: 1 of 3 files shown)


app/liquid_tags/asciinema_tag.rb
```diff
@@ -1,5 +1,6 @@
 class AsciinemaTag < LiquidTagBase
   PARTIAL = "liquids/asciinema".freeze
+  ASCIINEMA_URL_REGEX = %r{https://asciinema.org/a/(?<id>\d+)}.freeze
```
citizen428, line 3: We now also accept full URLs, e.g.`https://asciinema.org/a/239367`.

app/liquid_tags/asciinema_tag.rb
```diff
@@ -18,14 +19,15 @@ def render(_context)
...
-    input_no_space
+    sanitized_input = input.strip
+    match_data = sanitized_input.match(ASCIINEMA_URL_REGEX)
+    match_data ? match_data["id"] : validate(sanitized_input)
   end
 
-  def valid_id?(id)
-    id.match?(/\A\d+\Z/)
+  def validate(id)
+    raise "Invalid Asciinema ID: #{id}" unless id.match?(/\A\d+\z/)
```
citizen428, line 28: Changed this from `\Z` to `\z`, since the former would accept a trailing newline, which we don't want.

PR #8293, opened 2020-06-05
Routine Rubocop fixes

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Ran `rubocop -a && rubocop --auto-gen-config` that now picks up on Rails 6 new methods like [index_with](https://api.rubyonrails.org/classes/Enumerable.html#method-i-index_with)


app/services/podcasts/episode_rss_item.rb
```diff
@@ -25,8 +25,8 @@ def self.from_item(item)
     end
 
     def to_h
-      ATTRIBUTES.each_with_object({}) do |key, hash|
-        hash[key] = instance_variable_get("@#{key}")
+      ATTRIBUTES.index_with do |key|
```
mstruve, line 28: I never knew about `index_with` neat!

rhymes: @mstruve it was added in Rails 6, I didn't either :D

PR #8296, opened 2020-06-05
Replace defunct Algorithmia with Miro for color detection

benhalpern:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We have a longstanding feature in place that detects the proper color to use for main images while they load, for a smooth transition for the user.

However, this basically is fully broken and just getting rescued.

This replaces the longstanding useless code and removes Algorithmia gem from our codebase, replaced here with Miro.


spec/rails_helper.rb
```diff
@@ -148,7 +148,15 @@
     stub_request(:post, "http://www.google-analytics.com/collect").
       to_return(status: 200, body: "", headers: {})
 
+    stub_request(:any, /robohash.org/).
```
mstruve, line 151: This seems a little strange, was this needed to run the specs locally?

app/labor/color_from_image.rb
```diff
@@ -4,21 +4,13 @@ def initialize(url)
...
+    get_hex
+  rescue StandardError
+    "#dddddd"
   end
 
-  def get_palette
-    input = {
-      url: @url
-    }
-    client = Algorithmia.client(ApplicationConfig["ALGORITHMIA_KEY"])
```
mstruve, line 20: Did we just remove an external API call with this switch?! 💃

mstruve:
FWIW the gem you chose does not look well maintained, any concerns about that? It seems simple enough so I am not too concerned but wanted to point it out

benhalpern:
Yeah @mstruve this new gem happens to replace some functionality that was _fully not maintained_ and happens not to crash anything due to the `rescue`.

And this seems to work pretty well, and is simple, as you point out.

Issue #7760, opened 2020-05-10
Footnote and content on return from footnote hidden by top bar


wayofthepie:
**Describe the bug**
When following links to footnotes and back from footnotes the line with the anchor gets hidden behind the top bar.

**To Reproduce**
[This post](https://dev.to/wayofthepie/footnote-issue-43m6-temp-slug-580403?preview=3efd6c798b2c7cae5e4947f5fab920b339b18ba1684ee6e6cf23ea667d3a94860ae6a8d8867b7a496110656c5644d2da5c2a1ee83e20f0ad919af8f2) has a simple example and is the one used in the screenshots below.

1. Add a footnote in a post.
2. Save the post.

![image](https://user-images.githubusercontent.com/1102174/81507870-23e55c00-92f8-11ea-8774-84a7f3b93e6f.png)

3. Click the footnote link in the actual content, it will go to the matching footnote but that will be hidden - e.g. the image below should show the foot note `1. A test ↩` but it's behind the top bar.

![image](https://user-images.githubusercontent.com/1102174/81507896-53946400-92f8-11ea-898a-425101a05bbe.png)

4. Similarly try clicking the link back to the content (`↩`) the content will now be hidden behind the top bar.

![image](https://user-images.githubusercontent.com/1102174/81507929-88082000-92f8-11ea-826e-f5208fc7cda7.png)

**Expected behavior**
I would expect clicking the footnote link in the content to scroll to the footnote like this:

![image](https://user-images.githubusercontent.com/1102174/81507955-b980eb80-92f8-11ea-9397-3b7323c55ddf.png)

And I would expect clicking the back reference (`↩`) in the footnote to look like this:

![image](https://user-images.githubusercontent.com/1102174/81507970-e6350300-92f8-11ea-9ffe-90e0b55a457d.png)

**Desktop:**

- OS: Ubuntu 20.04 and Windows 10
- Browser: Latest Firefox and Google Chrome

Thanks!

citizen428:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

[...]

PR #8298, opened 2020-06-05
fix footnote anchor tag misposition

icncsx:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When following links to footnotes and back from footnotes the line with the anchor gets hidden behind the top bar. This is due to the fact that we have to offset an html anchor to adjust for fixed header.

The solution is to add something like this to the CSS:

```css
:target {
  padding-top: var(--header-height);
  margin-top: calc(-1 * var(--header-height));
}
```

Once we make that change, we no longer have the problem referenced at issue #7760. The header no longer covers the footnote.

Screenshot from 2020-06-05 11-14-45

The JavaScript solution was originally considered, where we listen to the `onhashchange` event. On change we want to select the dom element and use `window.scrollTo()` to scroll to the correct Y position using position data from `getboundingclientrect()`. Ultimately, the CSS solution was preferred for its simplicity.

## To Reproduce

Author an article and view/preview it. The article should have anchor tags that refer to in-document ids (in-page anchors).

```html
<p>This is a test <sup id="fnref1"><a href="#fn1">1</a></sup>.</p>

<p>This is just some random content to fill the space.</p>

<ol>
  <li id="fn1">
    <p>A test.&nbsp;<a href="#fnref1">↩</a></p>
  </li>
</ol>
```

Then click on one of any in-page anchor and you should navigated to the correct position in the document.
 
This PR attempts to resolve issue #7760.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/assets/stylesheets/top-bar.scss
+++ b/app/assets/stylesheets/top-bar.scss
@@ -2,9 +2,10 @@
 @import 'variables';
 @import 'mixins';
 
-:target {
-  padding-top: var(--header-height);
-  margin-top: calc(-1 * var(--header-height));
+body.default-navbar-config {
+  .article .body [id] {
+    scroll-margin-top: var(--header-height);
+  }
 }
 
 .top-bar {
--- a/app/assets/stylesheets/top-bar.scss
+++ b/app/assets/stylesheets/top-bar.scss
@@ -2,6 +2,11 @@
 @import 'variables';
 @import 'mixins';
 
+:target {
+  padding-top: var(--header-height) !important;
+  margin-top: calc(-1 * var(--header-height)) !important;
+}
+
 .top-bar {
   --indicator-outline: var(--header-bg);
   --dropdown-display: none;
--- a/app/assets/stylesheets/top-bar.scss
+++ b/app/assets/stylesheets/top-bar.scss
@@ -3,8 +3,8 @@
 @import 'mixins';
 
 :target {
-  padding-top: var(--header-height) !important;
-  margin-top: calc(-1 * var(--header-height)) !important;
+  padding-top: var(--header-height);
+  margin-top: calc(-1 * var(--header-height));
 }
 
 .top-bar {
--- a/app/assets/stylesheets/top-bar.scss
+++ b/app/assets/stylesheets/top-bar.scss
@@ -2,12 +2,6 @@
 @import 'variables';
 @import 'mixins';
 
-body.default-navbar-config {
-  .article .body [id] {
-    scroll-margin-top: var(--header-height);
-  }
-}
-
 .top-bar {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/top-bar.scss
```diff
@@ -2,6 +2,11 @@
 @import 'variables';
 @import 'mixins';
 
+:target {
+  padding-top: var(--header-height) !important;
```
jacobherrington, line 6: Is `!important` strictly necessary here? I'm not a fan of adding it when there are alternatives.

icncsx: `:target` will have lower specificity, so I thought this would be problematic but it turns out that the visual effect is more or less the same. 

lower-spec

Here is what happens when I remove `!important`. 

margin-override

Yay still working 🙂

app/assets/stylesheets/top-bar.scss
```diff
@@ -2,6 +2,11 @@
 @import 'variables';
 @import 'mixins';
 
+:target {
+  padding-top: var(--header-height);
+  margin-top: calc(-1 * var(--header-height));
+}
+
```
ludwiczakpawel, line 9: I think this is not optimal solution because of two reasons:

1. It adds additional spacing above the footnote list.
2. User can make header not being sticky in /settings/ux and in that case that fix isn't needed at all. So we should better target it (I believe we have specific body class applied when header is sticky/notsticky).

---

Generally there's much better solution for that issue: 

```
body.default-navbar-config { // targeting sticky header only
  .article .body [id] { // targeting any element with [id] attribute within article content
    scroll-margin-top: var(--header-height); // magic :)
  }
}
```

`scroll-magin-top` is much better for handling issues like that :)

app/assets/stylesheets/top-bar.scss
```diff
@@ -2,6 +2,12 @@
 @import 'variables';
 @import 'mixins';
 
+body.default-navbar-config {
+  .article .body [id] {
+    scroll-margin-top: var(--header-height);
+  }
+}
+
```
ludwiczakpawel, line 10: Lovely! Can I have one more request to actually put it in [this file](https://github.com/thepracticaldev/dev.to/blob/master/app/assets/stylesheets/article-show.scss) instead of topbar? It is more article-related issue rather than top bar :). But overall, it's great!

icncsx: Done! See most recent commit~

benhalpern:
This is really great!

PR #8301, opened 2020-06-05
Display site-wide announcements

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This feature allows active announcement broadcasts to be displayed, site-wide.

Specifically, this PR does the following: 
- [x] Adds a `broadcast:` key to our `AsyncInfoController`'s `base_data`, regardless of whether a user is logged in or not. If a broadcast is not present, passes along a `nil` value.
- [x] Initializes the active broadcast (if present) when rendering _any_ page.
- [x] Allows admins to preview a broadcast's `processed_html` when editing the broadcast on `/internal/broadcasts/:id`.
- [x] Adds specs around serializing and rendering an active broadcast (request + system-level specs).
- [x] Add some temporary styling to the broadcast banner such that it is `fixed` to the top of the page, below the nav. I actually don't really like how this looks, but it ensures that the broadcast renders with the highest `z-index` on an article page. **HOWEVER, once https://github.com/thepracticaldev/dev.to/pull/8234 is merged, the default style on a broadcast will Just Work™️, and we can remove this styling (you'll notice I added a TODO for this).**

The next step in this feature will be https://github.com/thepracticaldev/dev.to/issues/8091, which will allow the user to dismiss this active announcement. But, since this feature isn't being used yet in production and it doesn't affect any users just yet, I'll be handling that in a subsequent PR.

Closes https://github.com/thepracticaldev/dev.to/issues/8090.

**TO QA THIS PR:**
1. Go to `/internal/broadcasts`, and create/update a broadcast with some html/inline css. Once you save it, you can preview the broadcast and update it to be active:
Screen_Shot_2020-06-04_at_4_39_31_PM
2. Go to any page on the site. You should see the broadcast displayed:
Screen Shot 2020-06-04 at 4 39 48 PM
3. Try setting that same broadcast to "inactive" from the internal page. You should no longer see the broadcast displayed anywhere on the site!

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Nope!

Seth Meyers has an announcement gif

### DIFF
--- a/app/assets/javascripts/initializers/initializeBroadcast.js
+++ b/app/assets/javascripts/initializers/initializeBroadcast.js
@@ -1,0 +1,25 @@
+/**
+ * Parses the broadcast object on the document into JSON.
+ *
+ * @function broadcastData
+ */
+function broadcastData() {
+  const { broadcast = null } = document.body.dataset;
+  return JSON.parse(broadcast);
+}
+
+/**
+ * Inserts the broadcast's HTML into `active-broadcast` element
+ * as the first child within the document's body, and only inserts the HTML once.
+ *
+ * @function initializeBroadcast
+ */
+function initializeBroadcast() {
+  const { html } = broadcastData();
+  var el = document.getElementById('active-broadcast');
+
+  if (el.firstElementChild) {
+    return;
+  } // Only append HTML once, on first render.
+  el.insertAdjacentHTML('afterbegin', html);
+}
--- a/app/assets/javascripts/initializers/initializeBroadcast.js
+++ b/app/assets/javascripts/initializers/initializeBroadcast.js
@@ -5,6 +5,7 @@
  */
 function broadcastData() {
   const { broadcast = null } = document.body.dataset;
+
   return JSON.parse(broadcast);
 }
 
@@ -15,7 +16,11 @@ function broadcastData() {
  * @function initializeBroadcast
  */
 function initializeBroadcast() {
-  const { html } = broadcastData();
+  if (!broadcastData()) {
+    return;
+  }
+  const { html = null } = broadcastData();
+
   var el = document.getElementById('active-broadcast');
 
   if (el.firstElementChild) {
--- a/app/assets/javascripts/initializers/initializeBroadcast.js
+++ b/app/assets/javascripts/initializers/initializeBroadcast.js
@@ -16,10 +16,11 @@ function broadcastData() {
  * @function initializeBroadcast
  */
 function initializeBroadcast() {
-  if (!broadcastData()) {
+  const data = broadcastData();
+  if (!data) {
     return;
   }
-  const { html = null } = broadcastData();
+  const { html } = data;
 
   var el = document.getElementById('active-broadcast');
 
--- a/app/assets/stylesheets/scaffolds.scss
+++ b/app/assets/stylesheets/scaffolds.scss
@@ -242,6 +242,14 @@ body.hidden-shell {
   z-index: -1;
 }
 
+.broadcast-wrapper {
+  // TODO: [@thepracticaldev/delightful]: Remove these styles once
+  // https://github.com/thepracticaldev/dev.to/pull/8234 is merged.
+  z-index: 101; // Must be higher than the z-index of the sticky-nav.
+  position: fixed;
+  width: 100%;
+}
+
 @keyframes loading-fadein {
   from {
     opacity: 0;
--- a/app/assets/stylesheets/scaffolds.scss
+++ b/app/assets/stylesheets/scaffolds.scss
@@ -164,10 +164,24 @@ body.ten-x-hacker-theme {
   }
 }
 
+.broadcast-wrapper {
+  // TODO: [@thepracticaldev/delightful]: Remove these styles once
+  // https://github.com/thepracticaldev/dev.to/pull/8234 is merged.
+  z-index: 101; // Must be higher than the z-index of the sticky-nav.
+  position: fixed;
+  width: 100%;
+  font-size: 20px;
+  text-align: center;
+}
+
 body.static-navbar-config {
   padding-top: 0;
   .top-bar {
     position: relative !important;
+  }
+
+  .broadcast-wrapper {
+    position: relative; // Unset fixed position for static navbar config.
   }
 }
 
@@ -193,7 +207,7 @@ body.hidden-shell {
   }
 }
 
-// Remove post confirmaiton
+// Remove post confirmation
 .delete-confirm.container {
   padding: 150px 10px;
   min-height: 300px;
@@ -240,14 +254,6 @@ body.hidden-shell {
   animation-delay: 0.6s;
   font-size: 1.2em;
   z-index: -1;
-}
-
-.broadcast-wrapper {
-  // TODO: [@thepracticaldev/delightful]: Remove these styles once
-  // https://github.com/thepracticaldev/dev.to/pull/8234 is merged.
-  z-index: 101; // Must be higher than the z-index of the sticky-nav.
-  position: fixed;
-  width: 100%;
 }
 
 @keyframes loading-fadein {
--- a/app/controllers/async_info_controller.rb
+++ b/app/controllers/async_info_controller.rb
@@ -7,6 +7,7 @@ class AsyncInfoController < ApplicationController
     flash.discard(:notice)
     unless user_signed_in?
       render json: {
+        broadcast: broadcast_data,
         param: request_forgery_protection_token,
         token: form_authenticity_token
       }
@@ -22,9 +23,10 @@ class AsyncInfoController < ApplicationController
     respond_to do |format|
       format.json do
         render json: {
+          broadcast: broadcast_data,
           param: request_forgery_protection_token,
           token: form_authenticity_token,
-          user: user_data.to_json
+          user: user_data
         }
       end
     end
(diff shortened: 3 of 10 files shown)


app/assets/stylesheets/scaffolds.scss
```diff
@@ -242,6 +242,14 @@ body.hidden-shell {
   z-index: -1;
 }
 
+.broadcast-wrapper {
+  // TODO: [@thepracticaldev/delightful]: Remove these styles once
+  // https://github.com/thepracticaldev/dev.to/pull/8234 is merged.
+  z-index: 101; // Must be higher than the z-index of the sticky-nav.
+  position: fixed;
+  width: 100%;
+}
```
vaidehijoshi, line 251: As mentioned in the PR description, these styles can be removed once @ludwiczakpawel's PR, https://github.com/thepracticaldev/dev.to/pull/8234, is merged.

I've tried this out against his changes and we shouldn't need these styles since the default `<div>` styling will Just Work™️. Without these styles, the broadcast looks like this on an article page: 
Screen Shot 2020-06-04 at 4 56 07 PM

So that's why I'm adding this fixed positioning. But with @ludwiczakpawel's PR changes, it will look like this, so I think we'll be able to remove most of this styling:
Screen Shot 2020-06-05 at 8 43 17 AM

I'm

spec/system/homepage/user_visits_homepage_spec.rb
```diff
@@ -29,8 +30,28 @@
       expect(page).to have_text("DESIGN YOUR EXPERIENCE")
     end
 
+    context "when rendering broadcasts" do
+      let!(:broadcast) { create(:announcement_broadcast) }
+
+      it "renders the broadcast if active", js: true do
+        get "/async_info/base_data" # Explicitly ensure broadcast data is loaded before doing any checks
```
vaidehijoshi, line 37: I don't love that I have to do this, but it seems necessary since, afaik, the request to `async_info/base_data` doesn't happen by default. If someone has better ideas, open to hearing them!

app/views/internal/broadcasts/_form.html.erb
```diff
@@ -14,3 +14,12 @@
   <%= select_tag :active, options_for_select([false, true]) %>
 </div>
 <br>
+
+<% if @broadcast.processed_html %>
+  <div>
+    </p><strong>Preview</strong></p>
+    <%= @broadcast.processed_html.html_safe %>
```
vaidehijoshi, line 21: Hm, looks like brakeman is complaining that this is an `Unescaped model attribute`? This is a new one for me, but from the source code it looks like brake man [does indeed think](https://github.com/presidentbeef/brakeman/blob/c9fc366db717f759c064371a0321cccc88c2affb/test/apps/rails4/app/views/another/html_safe_is_not.html.erb) that this is unsafe.

Not sure what is the best way to get around this, since we _do_ want to be able to let admins preview here, and this html is a value that comes _directly from the database_ and not a cookie/param. Any thoughts? 🤔

juliannatetreault: Huh... 🤔 Maybe something like this could be useful?: https://stackoverflow.com/questions/39099489/how-to-fix-cross-site-scripting-security-warning-in-rails-when-using-html-safe

vaidehijoshi: @juliannatetreault Great find!! Trying it out with `sanitize`.

Sadly, it seems I need to explicitly pass in the sanitized attributes (according to [the docs](https://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html#method-i-sanitize_css) `style` and css should be sanitized by default), but I think that might be because of the way sanitize works with nested elements. In any case, I think passing in explicitly is safer.

Thanks again sending that method my way! 👍

app/views/shell/_top.html.erb
```diff
@@ -58,6 +58,7 @@
           <%= yield(:audio) %>
         </div>
         <%= render "layouts/top_bar" %>
+        <div id="active-broadcast" class="broadcast-wrapper"></div>
```
ludwiczakpawel, line 61: Do we want this empty DIV to always be there even if there's no announcement visible? Is there any way to avoid that?

ludwiczakpawel: Also, we do have something like `crayons-banner` (not documented so it's absolutely fair you weren't aware of that :D) that basically supposed to be this kind of a banner... We do have several variations:

```
<div class="crayons-banner">...</div>
```
image


```
<div class="crayons-banner crayons-banner--brand">...</div>
```
image

```
<div class="crayons-banner crayons-banner--success">...</div>
```
image

```
<div class="crayons-banner crayons-banner--warning">...</div>
```
image

```
<div class="crayons-banner crayons-banner--error">...</div>
```
image



We could obviously add additional ones and/or expend that component if needed.

app/controllers/async_info_controller.rb
```diff
@@ -37,6 +39,16 @@ def shell_version
...
   end
 
+  def broadcast_data
+    broadcast = Broadcast.announcement.active.first.presence
+    return unless broadcast
+
+    {
+      title: broadcast&.title,
+      html: broadcast&.processed_html
+    }.to_json
```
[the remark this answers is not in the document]
vaidehijoshi: @rhymes This is a good point, and I have no idea, honestly. tbqh, I was following [what was in the file](https://github.com/thepracticaldev/dev.to/blob/38634e5316140d82c6872673ac3c317034afce82/app/controllers/async_info_controller.rb#L27), but according to [the rails docs](https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-json):
> You don't need to call `to_json` on the object that you want to render. If you use the `:json` option, render will automatically call `to_json` for you.

It looks like we also use it [here](https://github.com/thepracticaldev/dev.to/blob/38634e5316140d82c6872673ac3c317034afce82/app/controllers/async_info_controller.rb#L37) and [here](https://github.com/thepracticaldev/dev.to/blob/38634e5316140d82c6872673ac3c317034afce82/app/controllers/async_info_controller.rb#L47) for some reason.

I kinda don't want to touch this in the context of this PR but it seems like we should fix it in a separate PR and test it to ensure that there actually _isn't_ a reason we're explicitly doing this.

app/views/shell/_top.html.erb
vaidehijoshi: @ludwiczakpawel:
> Do we want this empty DIV to always be there even if there's no announcement visible? Is there any way to avoid that?

The only point where we know whether an announcement exists or not is when we fetch it [and assign it from the backend](https://github.com/thepracticaldev/dev.to/pull/8301/files#diff-c9d5cb67f974ae68d44442f67dea97f2R41). When we initialize the page, we check to see if any data [was assigned on fetch](https://github.com/thepracticaldev/dev.to/pull/8301/files#diff-a99b7a23c274fa6b57e161e372ad2e28R24), and that's when we do something with the `<div id="active-broadcast" class="broadcast-wrapper"></div>`. The only other option here is to actually insert the div _into_ the body using `append()`, but I personally find that more confusing and a vector for potential UI bugs: we'd have to know _where_ in the body to append it, and we'd be relying on no other elements in the body being shifted around or accidentally added/removed, otherwise the `<div>` could render in a completely incorrect location. Also, it's worth mentioning that the `<div id="message-notice"></div>` element already exists (and is empty except for when used for chat channels), and it would be odd to do something entirely different to that, I think. 


> Also, we do have something like crayons-banner (not documented ...

Yeah, I was chatting with @lisasy about this! Since our `/internal` pages don't have access to crayons, the "preview" on `/internal/broadcasts/:id/edit` does not render any crayons variables. I can definitely add it, but I don't think that this blocks the PR, and would prefer to do it in a separate PR to avoid this one from getting big :)

app/assets/javascripts/initializers/initializeBroadcast.js
```diff
@@ -0,0 +1,30 @@
...
+ * Inserts the broadcast's HTML into `active-broadcast` element
+ * as the first child within the document's body, and only inserts the HTML once.
+ *
+ * @function initializeBroadcast
+ */
+function initializeBroadcast() {
+  if (!broadcastData()) {
+    return;
+  }
+  const { html = null } = broadcastData();
```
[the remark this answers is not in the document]
vaidehijoshi: LOL whoops, I was so focused on fixing my borked tests that I totally didn't even see myself doing this, haha. 🙈

app/assets/stylesheets/scaffolds.scss
vaidehijoshi: Addressed! Added styles to handle the two different header configs (totally forgot that we had that, good catch on pointing that out, @ludwiczakpawel!):

_Sticky:_
Screen Shot 2020-06-08 at 9 53 18 AM

_Static:_
Screen Shot 2020-06-08 at 9 53 09 AM

vaidehijoshi:
@rhymes I have addressed your comments, and included some default styling as you requested. As I mentioned to @ludwiczakpawel, I'm going to add [more default crayon style options](https://github.com/thepracticaldev/dev.to/pull/8301#discussion_r436826417) in a separate PR :)

@ludwiczakpawel I have addressed your questions and fixed the styling with the default navbar config view!

@joshpuetz The bug you noticed was actually pre-existing, and was just because of the way we were _not_ specifying a default value in the select options in the form. I've gone ahead and fixed that 👍

app/views/shell/_top.html.erb
ludwiczakpawel: Thanks, that was very helpful explanation :) Even though I don't like empty containers 🙈

PR #8303, opened 2020-06-05
Use Javascript to Display Jobs Banner

mstruve:
- [x] Feature try 2

## Description
Earlier in the week I tried to add a jobs banner to one of our cached partials (#8259) that did not work out so well bc of cacheing. This moves the display logic for the banner from the partial to the dynamic javascript that also loads our search results
 
## Added tests?
- [x] yes

alt_text


app/views/articles/_search.html.erb
```diff
@@ -59,6 +59,16 @@
     }
   }
 
+  function addOptionalBanner(query, displayBanner, jobsUrl) {
+    if (query) {
+      var lowerCaseQuery = query.toLowerCase();
+
+      if (displayBanner && (lowerCaseQuery == "job" || lowerCaseQuery == "jobs")) {
+        document.getElementById("articles-list").innerHTML == '<div class="crayons-notice crayons-notice--info mb-3 fs-base">Interested in joining our team? Explore our <a href=' + jobsUrl+ '>open roles</a>.</div>';
```
fdoxyz, line 67: This line throws me off a bit. Double equal is assigning the notice to the `innerHTML`? 👀

mstruve: oh typo! let me fix that, thanks!

app/views/articles/_search.html.erb
```diff
@@ -59,6 +59,16 @@
     }
   }
 
+  function addOptionalBanner(query, displayBanner, jobsUrl) {
+    if (query) {
+      var lowerCaseQuery = query.toLowerCase();
+
+      if (displayBanner && (lowerCaseQuery == "job" || lowerCaseQuery == "jobs")) {
```
vaidehijoshi, line 66: a cleaner alternative:
```suggestion
      if (displayBanner && lowerCaseQuery == ("job" || "jobs") {
```

do we want to handle the case when a user could search for "developer job" or "new job" or "tech jobs"? should we just do: `lowerCaseQuery.includes("job")`?

app/views/articles/_search.html.erb
fdoxyz: Settled for quirk instead of magic then! I was surprised too 😂

app/views/articles/_search.html.erb
mstruve: I didnt want to include all instances of "job" initially bc I wasn't sure if we wanted to be showing the banner in the event someone searches "How to get a dev job" "How to do job interviews" in those cases, they are clearly seeking advice and I wasn't sure if throwing the banner up for them was too forward. What do you think?

vaidehijoshi: I don't feel like that banner is _too_ intrusive that a user would be annoyed if we tried to guess that they were looking for job-related content (maybe we'd get people looking at the jobs page who wouldn't ordinarily looked for it!)

...But I also see your point! I was mostly just asking to know if we specifically _didn't_ want to do that, which you have answered :) I'm ok with just basing the conditional off the single word, too.

mstruve: > looking at the jobs page who wouldn't ordinarily looked for it!)

Ya know, I kinda like this idea. People who love DEV but never considered they could work there I think it would be awesome to show to. I also think, especially when we have jr positions open we could get some people applying who might not have otherwise even thoughts about it. I'll let this sit for the weekend see if anyone else wants to chime in otherwise I will make your suggested change.

vaidehijoshi: woohoo 🎉 sounds great!

spec/system/search/display_jobs_banner_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
+
+RSpec.describe "Display Jobs Banner spec", type: :system, js: true do
+  before do
+    stub_request(:post, "http://www.google-analytics.com/collect")
+    SiteConfig.jobs_url = "www.very_cool_jobs_website.com"
```
rhymes, line 6: we should mock this, otherwise it will have this value for any other test at the end

spec/system/search/display_jobs_banner_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  context "when SiteConfig.display_jobs_banner is false" do
+    it "does not show jobs banner" do
+      SiteConfig.display_jobs_banner = false
```
rhymes, line 11: some thing about mocking

spec/system/search/display_jobs_banner_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
+
+RSpec.describe "Display Jobs Banner spec", type: :system, js: true do
+  before do
+    stub_request(:post, "http://www.google-analytics.com/collect")
```
rhymes, line 5: Can't wait to get rid of this :D

app/views/articles/_search.html.erb
rhymes: +1 on adding it for all "jobs" related searches. maybe not "steve jobs" :D 
It'd be great if we could use a/b testing to see which strategy people prefer

spec/system/search/display_jobs_banner_spec.rb
vaidehijoshi: +1 on mocking, good catch @rhymes.

mstruve: Good call!!!

app/views/articles/_search.html.erb
```diff
@@ -59,6 +59,16 @@
     }
   }
 
+  function addOptionalBanner(query, displayBanner, jobsUrl) {
```
nickytonline, line 62: Just curious about this function. Is this for our own internal job postings? Also, I'd remove the `if(query)` from this function and put it down below in the Promise's then since the job of this function is to display a banner, not check if it should before it does.

app/views/articles/_search.html.erb
```diff
@@ -109,6 +119,8 @@
     })
       .then(response => response.json())
     .then((content) => {
+        addOptionalBanner(searchHash["search_fields"], content.display_jobs_banner, content.jobs_url);
```
nickytonline, line 122:
```suggestion
	if (searchHash["search_fields"]) {
        addOptionalBanner(searchHash["search_fields"], content.display_jobs_banner, content.jobs_url);
	}        
```

app/views/articles/_search.html.erb
mstruve: Oh snap I didnt include the original screenshot in this PR like the first one, it will show a little banner pointing to our jobs page like this. 

Screen Shot 2020-06-03 at 1 57 53 PM

I brought it up in Slack bc I would constantly type into Chome "dev.to jobs" when I wanted to get a job link to send to someone and I always ended up on dev.to with jobs as my search query which I didn't want. This ensures that anyone who does that can easily get to our job page.

PR #8304, opened 2020-06-05
Allow users to set a payment pointer for their content

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR allows individual users to add a payment pointer to their own profiles so that they can receive web monetization payments directly for traffic to their own posts, their own profiles and comment sections of posts they created.

This will help users directly monetize their content without our intervention in any way. Any earnings will be fairly minimal early on while the web monetization ecosystem is still early technology, but it's likely that our technical community will adopt some of this earlier than others and we are proud to be able to let users benefit early on.

The app itself has a payment pointer for traffic to pages not monetized by individual authors, which will help all Forem creators monetize in the future as well. In a lot of cases the platform is the only one monetizing in this fashion, rather than individual creators, but we felt like the functionality that allows both parties to take part across the Forem ecosystem is the only way we would want this to work.

In this PR I just linked to an explainer post about web monetization, but I think we should follow up with a more straightforward guide about how precisely it works on the platform.

Here is the Grant for the Web Hackathon page where you can find more information on web monetization and payment pointers:

https://dev.to/devteam/announcing-the-grant-for-the-web-hackathon-on-dev-3kd1


app/views/shared/_payment_pointer.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if user.payment_pointer.present? %>
+  <div id="author-payment-pointer" data-payment-pointer="<%= user.payment_pointer %>" ></div>
+<% end %>
```
joshpuetz, line 3: ⚠️

benhalpern, line 3:
```suggestion
<% end %>

```

PR #8305, opened 2020-06-05
Spec Speed Up: Memoize Feed objects and Combine some Tests

mstruve:
- [x] Optimization

## Description
Going through our slowest spec list and this one was at the top. Ran through and made a couple quick optimizations by memorizing some feed object, combining some tests, and reducing the number of cycles for testing for empty feed. This doesnt speed it up like crazy but makes it a few seconds faster. 

but its gotta get done! 
alt_text

### DIFF
--- a/spec/services/articles/feed_spec.rb
+++ b/spec/services/articles/feed_spec.rb
@@ -30,8 +30,9 @@ RSpec.describe Articles::Feed, type: :service do
     let(:tagged_article) { create(:article, tags: tag) }
 
     it "returns published articles" do
-      expect(feed.published_articles_by_tag).to include article
-      expect(feed.published_articles_by_tag).not_to include unpublished_article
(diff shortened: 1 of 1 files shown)


spec/services/articles/feed_spec.rb
```diff
@@ -456,7 +447,7 @@
       # We manually called `feed.globally_hot_articles` here because `let` caches it!
       it "still returns articles" do
         empty_feed = false
-        20.times do
+        5.times do
```
mstruve, line 450: @joshpuetz you set this to 20 and I am not sure if there was a reason you picked that number but since it is only returning 1 article I decreased it to 5. If there was a reason you choose 20 let me know!

joshpuetz: I can't recall offhand, but it smells like a hedge against test flakiness. I say ship it!

PR #8314, opened 2020-06-07
Flaky Spec Fixes: User Block and Sitemap Specs

mstruve:
- [x] Bug Fix

## Description
* User block spec: Ensure the id we are using is going to be missing
* Sitemap: Don't rely on a randomly picked datetime (essentially what we get with random Time zone setting) and ensure dates will always be within the month we want. 

Running builds in parallel is totally awesome and fast but it has really exposed some of our flaky specs. Rather than throwing them all in the parallel build PR I want to spread them out in small PRs. This should also help our current build! 

alt_text

### DIFF
--- a/spec/requests/user_blocks_spec.rb
+++ b/spec/requests/user_blocks_spec.rb
@@ -83,8 +83,10 @@ RSpec.describe "UserBlock", type: :request do
     end
 
     it "raises ActiveRecord::RecordNotFound error if UserBlock not found" do
+      missing_id = blocked.id
+      blocked.destroy
       expect do
-        delete "/user_blocks/#{blocked.id}", params: { user_block: { blocked_id: blocked.id + 10 } }
+        delete "/user_blocks/#{missing_id}", params: { user_block: { blocked_id: missing_id } }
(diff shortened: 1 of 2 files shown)


spec/requests/user_blocks_spec.rb
```diff
@@ -83,8 +83,10 @@
     end
```
rhymes, line 83: I think blocker and blocked should be reloaded with `let_it_be_changeable()` at the top, as by destroying the object we are making sure that if we ever switch on the random order, another test won't have the objects we destroyed in another one.

rhymes: An alternative strategy, if recreating the object anew is too costly, is to create an object inside the test and destroy it right after. This way the other tests can keep using the shared object and the test which call destroy is truly independent

mstruve: > This way the other tests can keep using the shared object

The objects right now(blocked and blocker) are being created with `let` so they only exist for the duration of the test and don't persist in between which means none of these tests should have shared objects so we should be good here.

rhymes: oh, right! :D

PR #8316, opened 2020-06-07
expand hover radius with before element

### LINKED ISSUE #7606: Dropdown under profile picture disappears when cursor is diagonally
**Describe the bug**

Dropdown under user profile disappears when the cursor is moved diagonally 

**To Reproduce**

1. Go to dev.to
2. Move cursor diagonally from the profile picture

**Expected behavior**

The dropdown should stay there when the cursor is moved diagonally upto some distance.

**Screenshots**

dropdown

- OS: Ubuntu 18.04
- Browser: Brave
- Version: 1.7.92 Chromium: 80.0.3987.163 (Official Build) (64-bit)

icncsx:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Closes #7606.

In brief, the drop-down under user profile disappears when the cursor is moved diagonally because the hover area of the top bar menu is insufficient. The solution is to use a pseudo element such as `:before` to keep the original element size but make the hover area bigger. Now the user can move the cursor diagonally and access the user profile without issue.

Screenshot from 2020-06-07 00-16-08

Issue #7606.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/assets/stylesheets/top-bar.scss
+++ b/app/assets/stylesheets/top-bar.scss
@@ -113,6 +113,7 @@
   display: flex;
   align-items: center;
   z-index: 20;
+  position: relative;
   &__trigger {
     all: unset;
     padding: var(--su-1);
@@ -125,6 +126,17 @@
   &__dropdown {
     display: var(--dropdown-display);
     top: var(--header-height);
+  }
+
+  &:before {
+    content: '';
+    position: absolute;
+    --menu-width: 150%;
+    --menu-height: 110%;
+    width: var(--menu-width);
+    height: var(--menu-height);
+    top: calc((100% - var(--menu-height)) / 2);
+    left: calc((100% - var(--menu-width)) / 2);
   }
 
   &.showing,


icncsx:
@ludwiczakpawel I agree. It's not ideal but it's consistent with the current implementation of toggling `display` on hover. 

How about a JS solution? `onmouseover`, we display the dropdown and on `mouseout` we hide the dropdown but only after a small delay. This gives a person 100-200 ms longer to move the cursor diagonally and hover over the dropdown. This feels gimmicky, so I want to ask your opinion about it.

ludwiczakpawel:
@icncsx yea i think this will make the most sense. Before we dive into JS, we should explore if it's doable with CSS only - here's interesting proof of concept: https://stackoverflow.com/questions/9393125/delay-mouseout-hover-with-css3-transitions but I'm not sure if it will work in our case.

icncsx:
@ludwiczakpawel If CSS can do it, I'll always take it over a JS one. Thanks, I'll check it out soon :)

PR #8332, opened 2020-06-07
Flaky Spec Fix: Bulk Delete SQL Spec

mstruve:
- [x] Bug Fix

## Description
In my Knapsack branch this spec would fail about 50% of the time bc the count of Notifications did not change after the delete was wrong. I think it might be something to so with the way the expect block executes and the SQL statement possibly getting rolledback due to database cleaning before the assertion. Changing it to be explicit like this seems to have fixed the problem. I would get this spec to fail almost every run before and with this change it has yet to fail after 4 runs. 

Failed specs are our internal flaky ones, unrelated to this change. 

alt_text

### DIFF
--- a/spec/services/bulk_sql_delete_spec.rb
+++ b/spec/services/bulk_sql_delete_spec.rb
@@ -7,7 +7,7 @@ describe BulkSqlDelete, type: :service do
       WHERE notifications.id IN (
         SELECT notifications.id
         FROM notifications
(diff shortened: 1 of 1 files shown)


mstruve:
@snackattas I tried assigning you some of these spec PRs so you could take a look/keep track of the changes but you dont come up as an option so I will try and get in the habit of tagging you in the description

PR #8333, opened 2020-06-07
Make Video Permission a Valid Role

mstruve:
- [x] Bug Fix

## Description
When we validate a role we check to make sure it is included in our list of `Role::ROLES` which lists the names of all the roles we have. One of the things we are doing to fix our spec suite is ensure roles are correctly added during tests. This uncovered the fact that this was a role we could not successfully add bc it was not valid. This fixes that. 
 
## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -12,6 +12,7 @@ class Role < ApplicationRecord
     mod_relations_admin
     tech_admin
     trusted
+    video_permission
     warned
     workshop_pass
   ].freeze
--- a/spec/models/role_spec.rb
+++ b/spec/models/role_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe Role, type: :model do
         admin banned chatroom_beta_tester comment_banned
         podcast_admin pro single_resource_admin super_admin
         tag_moderator mod_relations_admin tech_admin
-        trusted warned workshop_pass
+        trusted video_permission warned workshop_pass
       ]
       expect(described_class::ROLES).to eq(expected_roles)
     end


snackattas:
Oh, I realize I approved this already, but I just thought of a question. Is `:video_permission` actually a valid role? I don't _think_ I see it used anywhere in the codebase, although I could be missing it

rhymes:
@snackattas you do have a point! it's added in specs, but I don't see it added in code:

```ruby
 ➜  devto git:(master) rg video_permission
spec/factories/users.rb
76:    trait :video_permission do

spec/requests/videos_spec.rb
5:  let(:authorized_user)   { create(:user, :video_permission) }

spec/requests/api/v0/videos_spec.rb
4:  let_it_be_readonly(:user) { create(:user, :video_permission) }

spec/system/internal/admin_bans_or_warns_user_spec.rb
64:    user.add_role :video_permission
71:    expect(user.has_role?(:video_permission)).to eq(false)

app/services/moderator/manage_activity_and_roles.rb
28:      @user.remove_role :video_permission
```

mstruve:
@rhymes @snackattas I wondered the same thing so I removed it and all the specs in `spec/requests/videos_spec.rb` failed so I added it back and didn't question it further and assumed it was some Rolify magic going on

snackattas:
Looks like `spec/requests/videos_spec.rb` is using a fake `:video_permission` https://github.com/thepracticaldev/dev.to/blob/master/spec/factories/users.rb#L76

I think we can probably gut it then, right?

rhymes:
Oooh, lol! Great find @snackattas!

mstruve:
WTF?! Amazing find, I didnt even look at the block. Ok updated PR coming shortly

PR #8335, opened 2020-06-08
Fix Ruby 2.7.1 download in Gitpod with non cached images

snackattas:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

https://github.com/thepracticaldev/dev.to/issues/8334

## Description
I was seeing the errors documented in the issue with my branches on Gitpod.  

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed

Gitpod makes me think of tide pods which led me here on giphy
guy fieri tide pod


.gitpod.dockerfile
```diff
@@ -2,8 +2,13 @@ FROM gitpod/workspace-postgres
 
 # Install Ruby
 ENV RUBY_VERSION=2.7.1
+RUN rm /home/gitpod/.rvmrc && touch /home/gitpod/.rvmrc && echo "rvm_gems_path=/home/gitpod/.rvm" > /home/gitpod/.rvmrc
 RUN bash -lc "rvm install ruby-$RUBY_VERSION && rvm use ruby-$RUBY_VERSION --default"
 
+# Install Node
```
citizen428, line 8: Hm, I'm not sure if this is necessary. The image we use (`gitpod/workspace-postgres`) inherits from `gitpod/workspace-full` which includes Node. On my still working cached pod of master the following Node version is available:

```
gitpod /workspace/dev.to $ node
Welcome to Node.js v12.16.1.
```

snackattas: weird,  I was getting this error.  Yea I should probably check which node versions are installed with nvm prior and see if I can just reassign to v12.16 instead of reinstalling

2020-06-09 00 35 51_ALrRAp png

snackattas: https://hub.docker.com/r/gitpod/workspace-full/dockerfile this says it uses `v8.14.0`? so confused

snackattas: then, clearing up the node version error, get this last error :(
2020-06-09 08 53 26_94wvM5 png

snackattas: feel free to commit to this branch if you have ideas as well

snackattas: oh wait 🤦 it was still using 12.18.3 I didn't change the path or .nvmrc to ensure it uses 12.16.3. that fixed the above

citizen428:
@snackattas This seems to work! I've rebased #8337 against this so I can continue there.

.gitpod.dockerfile
akhil-naidu: > weird, I was getting this error. Yea I should probably check which node versions are installed with nvm prior and see if I can just reassign to v12.16 instead of reinstalling
> 
> 2020-06-09 00 35 51_ALrRAp png
> 
> I think its a caching thing again, that base img must have changed their node version?

You can fix this, by changing the version number of the node to 12.18.x in the package.json

PR #8339, opened 2020-06-08
Use official Asciinema player embed

citizen428:

- [X] Bug Fix

## Description

1. While doing the [original PR](4152) we noticed that the Asciinema IFrames are quite small, so added a `vh50` attribute to them (`style="width:100%;height:50vh;"`).
2. The problem with hardcoding this value: every Asciinema recording has a different width and height, and the site doesn't scale them in any way to avoid legibility issues. Instead, they emit an event containing the proper size after the video is loaded.
3. The event is then handled by [this JS function](https://github.com/asciinema/asciinema-server/blob/7250aaca1cc739bf53de9c3e804dbe6a431b4ef2/app/assets/javascripts/widget.js#L71-L82) for resizing the IFrame based on an event. 

All of this is handled by their official player embed and this PR replaces the direct IFrame embed with a script tag. 

#8185

Screenshot of a properly resized player with one of the examples linked in the issue:

Screen Shot 2020-06-08 at 15 16 19

The other video linked in the issue:

Screen Shot 2020-06-08 at 15 17 30

## Added tests?

- [X] Existing tests were recently amended in the refactor PR (#8288).

## Added to documentation?

- [X] no documentation needed

### DIFF
--- a/app/javascript/article-form/articleForm.jsx
+++ b/app/javascript/article-form/articleForm.jsx
@@ -32,6 +32,17 @@ export default class ArticleForm extends Component {
     activateRunkitTags();
   }
 
+  // Scripts inserted via innerHTML won't execute, so we use this handler to
+  // make the Asciinema player work in previews.
+  static handleAsciinemaPreview() {
+    const els = document.getElementsByClassName('ltag_asciinema');
+    for (let i = 0; i < els.length; i += 1) {
+      const el = els[i];
+      const script = el.removeChild(el.firstElementChild);
+      postscribe(el, script.outerHTML);
+    }
+  }
+
   static propTypes = {
     version: PropTypes.string.isRequired,
     article: PropTypes.string.isRequired,
@@ -103,6 +114,7 @@ export default class ArticleForm extends Component {
     if (previewResponse) {
       this.constructor.handleGistPreview();
       this.constructor.handleRunkitPreview();
+      this.constructor.handleAsciinemaPreview();
     }
   }
 
--- a/app/views/liquids/_asciinema.html.erb
+++ b/app/views/liquids/_asciinema.html.erb
@@ -1,9 +1,3 @@
 <div class="ltag_asciinema">
-  <iframe frameborder="0"
-    scrolling="no"
-    id="asciinema_<%= id %>"
-    src="https://asciinema.org/a/<%= id %>/iframe"
-    loading="lazy"
-    style="width:100%;height:50vh;">
-  </iframe>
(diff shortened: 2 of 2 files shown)


citizen428:
Yeah, I noticed that too, but I thought correct looking embeds in the final article trump the previews. But I'll look into the solution @maestromac linked, thanks for that!

citizen428:
@maestromac I had a look at that now. According to the documentation, it's for making synchronous scripts load asynchronously, but that's not the problem we have, as the Asciinema player uses an async tag by default.

citizen428:
@maestromac @rhymes The reason the Asciinema tags don't show in previews is that we're using `dangerouslySetInnerHTML` in `app/javascript/article-form/components/Preview.jsx`:

```js
dangerouslySetInnerHTML={{ __html: previewResponse.processed_html }}
```

Alas the [HTML 5 spec](https://www.w3.org/TR/2008/WD-html5-20080610/dom.html#innerhtml0) is quite clear on that matter:

> Note: script elements inserted using innerHTML do not execute when they are inserted.

I thought I'll be clever and use an `IFrame` as workaround, but alas that gets blocked:

Screen Shot 2020-06-09 at 14 27 02

This leaves us in an awkward spot: the JS player seems to be by far the most straightforward way to get correctly resized embeds, but it breaks previews. :-( So I can think of a few potential solutions, but none of them excite me:

1. Leave as is. I don't think it makes for a great publishing experience and we may generate extra support requests.
2. Don't use the official player and write our own JS to resize the IFrame. I'd rather not duplicate the work that the Asciinema developers are already doing.
3. Use some sort of placeholder (e.g. the Asciinema preview image) in the preview **only**. Decent user experience, but requires a special case in our component for a single liquid tag.

Other ideas?

rhymes:
I like #3, it's not perfect but it's better than 1 I think. We definitely don't want to do 2

citizen428:
@rhymes I'm leaning towards 3 too. I'm just not super happy about embeding a special case for one Liquid tag in the Preact component, but I may just have had an idea.

maestromac:
@citizen428 I'd go with 3 too but I have a working solution. I should've expanded on my pervious on commment; I was suggested you use postscribe to force that script to load only when previewing an article. I was able to (sort of) get it working by messing around with the name of the element to seek, see [my commit here](https://github.com/maestromac/dev.to/commit/44099eaa20971a9a0d73f77ad78070928b5c998c)

citizen428:
@maestromac Thanks, that extra context helped. I added a new handler for this purpose, as the approach used for gists ended up duplicating the player tags.

PR #8341, opened 2020-06-08
Replace whitelisted_ips with permissions in Docker init

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Rails [web-console](https://github.com/rails/web-console) officially uses the `.permissions` settings instead of the deprecated `.whitelisted_ips`. I've updated the code to follow suit.

See https://github.com/rails/web-console#configweb_consolepermissions

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/config/initializers/docker.rb
+++ b/config/initializers/docker.rb
@@ -9,5 +9,5 @@ if Rails.env.development? && File.file?("/.dockerenv")
     BetterErrors::Middleware.allow_ip!(host_ip)
   end
 
-  Rails.application.config.web_console.whitelisted_ips << host_ip
+  Rails.application.config.web_console.permissions = host_ip
 end


config/initializers/docker.rb
```diff
@@ -9,5 +9,5 @@
     BetterErrors::Middleware.allow_ip!(host_ip)
   end
 
-  Rails.application.config.web_console.whitelisted_ips << host_ip
+  Rails.application.config.web_console.permissions = host_ip
```
mstruve, line 12: Why did you choose to replace the shovel with the equals here?

rhymes: Because of the documentation: 

```ruby
Rails.application.configure do
  config.web_console.permissions = '192.168.0.0/16'
end
```

in https://github.com/rails/web-console#configweb_consolepermissions

technically it can receive an array, so we could use `<<` but by default is nil and as it's only used there, my laziness won :D

Issue #8091, opened 2020-05-27
Allow users to dismiss active broadcasts


vaidehijoshi:
Once we implement the feature to render an "active" broadcast throughout the site (https://github.com/thepracticaldev/dev.to/issues/8090), we should also ensure that users can dismiss the currently-active broadcast by `x`-ing out of it.

We probably want to save this in localStorage, similar to how we implement the "close" functionality on the hero banner. This should be broadcast-specific; in other words, if an admin activates _another_ broadcast, the user should still see it, even if they have dismissed a different active broadcast previously.

vaidehijoshi:
To our awesome contributors, please note that this issue will be worked on by a DEV team member and is not up for grabs.

benhalpern:
> We probably want to save this in localStorage, similar to how we implement the "close" functionality on the hero banner. This should be broadcast-specific; in other words, if an admin activates another broadcast, the user should still see it, even if they have dismissed a different active broadcast previously.

Yup, I'd say you can follow the same exact format as the hero banner (and probably generalize that whole function).

Basically store _which_ broadcast has been dismissed and don't show the broadcast if that is stored as such.

Issue #8090, opened 2020-05-27
Display active announcement broadcast across site


vaidehijoshi:
If an admin toggles on an `Announcement` broadcast, we need to display the currently-active broadcast across the site (for both logged-in users and non-logged-in users).

We may also want to allow admins to preview what this will look like before setting a broadcast live (should be able to do this by re-using the same code on the admin panel that we will use to render on the homepage itself).

Current working design from @lisasy:
![image](https://user-images.githubusercontent.com/6921610/83047124-cdb53000-9ffc-11ea-9678-313c7b5fc4ba.png)

vaidehijoshi:
To our awesome contributors, please note that this issue will be worked on by a DEV team member and is not up for grabs.

PR #8343, opened 2020-06-08
Allow Users to Opt-out of Announcements

### LINKED ISSUE #8092: Allow users to opt-out of announcements
On the `/settings` page, we should provide an option for users with accounts to unsubscribe/turn off "announcement" broadcasts completely (similar to our checkbox for "display ads", which corresponds to `display_sponsors` on a `User`).

We could probably follow the convention and call the column on the table `display_announcements`.

juliannatetreault:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR gives Users the option to display or not display Announcements while browsing. This is accomplished by adding a `boolean` to the Users table, `display_announcements`, that defaults to `true` (similar to `display_sponsors`). On the backend, it also adds `display_announcements` to the permitted attributes in `user_policy.rb` and to `#user_data` in the `AsyncInfoController`. On the frontend, it adds an "Announcements" section to `/settings/misc`, where a User can toggle between displaying or not displaying Announcements.

Closes #8092 
Relates to #8090 and #8091 

**Default (`display_announcements` is true):**
Screen Shot 2020-06-08 at 9 33 54 AM
**When a User does not want to display announcements:**
Screen Shot 2020-06-08 at 9 34 05 AM

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Announcement


juliannatetreault:
> > Added tests? no, because they aren't needed until the Announcement work is merged in, making this feature testable!
> 
> I think we should definitely add some tests to make sure that a user can update these settings successfully, since the Rails updating aspect of this is in the scope of this PR!

I've added a spec to the `user_settings_spec.rb` that ensures that Announcements properly render under `/settings/misc` and I've added a new system spec (similar to the `user_selects_looking_for_work_spec.rb`),`user_updates_announcements_spec.rb`, that tests that Announcement settings can be updated. Let me know what you think!

spec/system/user/user_updates_announcements_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "rails_helper"
+
+RSpec.describe "Announcements", type: :system do
```
vaidehijoshi, line 3:
```suggestion
RSpec.describe "User updates announcements", type: :system do
```

spec/system/user/user_updates_announcements_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  let(:user) { create(:user) }
+
+  before do
+    sign_in(user)
+    visit "/settings/misc"
+  end
+
+  it "renders the announcements settings" do
+    expect(page).to have_text("Announcements")
```
vaidehijoshi, line 12: What is this test actually checking for? It seems to do the same thing as the [one over here](https://github.com/thepracticaldev/dev.to/pull/8343/files#diff-dabdd41261904eae23b477c8ae3a61adR34-R37).

spec/system/user/user_updates_announcements_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  it "properly updates the announcements settings", js: true do
+    page.check "Display Announcements (When browsing)"
+    sidekiq_perform_enqueued_jobs do
+      click_button("Save Announcements Settings")
+    end
+    expect(page).to have_text("Your profile was successfully updated.")
```
vaidehijoshi, line 20: Hm, this test checks that the flash message shows up, but if you think about it, it's not really testing that the broadcast was updated. For example, someone could go into the broadcasts controller, make a change to the `save!` method, and this flash message might still show up (meaning the test would pass even though the code didn't actually work!).

I think you want to take a look at `spec/requests/user/user_settings_spec.rb`... 👀

vaidehijoshi: I think we eventually _do_ want to add a system spec for announcements in order to test if the announcement banner does not appear if you uncheck the "display announcements" box (but that feature isn't hooked up yet 😆)

Currently though, I don't think that currently these tests are really testing _your_ changes to the backend.

PR #8344, opened 2020-06-08
Remove video_permissions Role Reference From Specs and Code

mstruve:
- [x] Refactor

## Description
We are not using the video_permissions role that is referenced in our codebase so this removes it and replaces it with the actual video authorization method which is ensuring the user was created more than 2 weeks ago. 

Spec Failure is a flaky one 

@snackattas 

alt_text


mstruve:
Yeah separate flaky spec 🤦‍♀️

PR #8348, opened 2020-06-08
Spec Speedup: Comment Out More Percy Only Specs

mstruve:
- [x] Optimization

## Description
Find all fail! [When I did the first round of these](https://github.com/thepracticaldev/dev.to/pull/8347) I searched with tab indentation in my search string which led to me missing the vast majority of these Percy only specs. Here are the rest of them commented out.  I was also only looking for specs with a single Percy.snapshot line, however, as you can see in this PR, some specs have lines before the Percy snapshot to setup the environment. The key is that NONE of them have any additional expectations. 

Thats ok, its another excuse to put a gif of a hedgehog in a PR 😄 
alt_text

### DIFF
--- a/spec/system/articles/user_creates_an_article_spec.rb
+++ b/spec/system/articles/user_creates_an_article_spec.rb
@@ -13,7 +13,8 @@ RSpec.describe "Creating an article with the editor", type: :system do
     sign_in user
   end
 
-  it "renders the page", js: true, percy: true do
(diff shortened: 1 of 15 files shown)


mstruve:
> I'd recommend just removing them and documenting it in git  

We probably should document something about Percy and its state right now. I do kinda like the comments bc it will make those specs easier to find if/when we decide to enable this again though I do get having a bunch of them is a bit annoying.

PR #8350, opened 2020-06-08
Flaky Spec Fix: Set Profile Image to Nil to Avoid Server Errors in js Specs

mstruve:
- [x] Bug Fix

## Description
When running tests in parallel about 50% of the time we would see these failures. ~Given we dont care about loading the pictures I think we can safely ignore these.~ I set profile_image to nil for the users in these specs to avoid these errors trying to load the images
```
  2) User visits a homepage when no options specified when main featured article shows the main article readable date
     Failure/Error: raise ActionController::RoutingError, "No route matches [#{env['REQUEST_METHOD']}] #{env['PATH_INFO'].inspect}"
     ActionController::RoutingError:
       No route matches [GET] "/uploads/user/profile_image/648/57285fc5-ca55-4721-be7e-d09b931fe325.jpeg"
     # /home/travis/build/thepracticaldev/dev.to/vendor/bundle/ruby/2.7.0/gems/actionpack-6.0.3.1/lib/action_dispatch/middleware/debug_exceptions.rb:36:in `call'
     # ------------------
     # --- Caused by: ---
     # Capybara::CapybaraError:
     #   Your application server raised an error - It has been raised in your test code because Capybara.raise_server_errors == true
     #   ./spec/rails_helper.rb:122:in `block (3 levels) in <top (required)>'
  3) User visits a homepage when no options specified when all other articles shows all articles dates
     Failure/Error: raise ActionController::RoutingError, "No route matches [#{env['REQUEST_METHOD']}] #{env['PATH_INFO'].inspect}"
     ActionController::RoutingError:
       No route matches [GET] "/uploads/user/profile_image/648/57285fc5-ca55-4721-be7e-d09b931fe325.jpeg"
     # /home/travis/build/thepracticaldev/dev.to/vendor/bundle/ruby/2.7.0/gems/actionpack-6.0.3.1/lib/action_dispatch/middleware/debug_exceptions.rb:36:in `call'
     # ------------------
     # --- Caused by: ---
     # Capybara::CapybaraError:
     #   Your application server raised an error - It has been raised in your test code because Capybara.raise_server_errors == true
     #   ./spec/rails_helper.rb:122:in `block (3 levels) in <top (required)>'
```

@snackattas 

alt_text


spec/support/initializers/capybara.rb
```diff
@@ -20,6 +20,12 @@
     driven_by :rack_test
   end
 
+  config.around(:each, raise_server_errors: false) do |example|
```
nickytonline, line 23: Do we know why the images fail to load though? I'm only asking as we'll need them once Percy starts running again.

mstruve: I am not sure why this doesnt happen all the time but we load the image path into the HTML/javascript and **then my best guess** would be something like InstantClick then tries to follow it but ends up trying to run the path against our server instead of through a helper like cloudinary which would resolve it correctly. There is probably some host env variable we can set to ensure the pictures always get loaded but I have a lot of specs that need fixing right now and this seemed like the easiest and quickest fix. At this point I dont think Percy will be coming back anytime soon but when it does, if this fails those snapshots will report the problem and we can address it then more "correctly"

snackattas:
Looks like the image in that error is from the [user factory's profile_image urls](https://github.com/thepracticaldev/dev.to/blob/master/spec/factories/users.rb#L13), which are invoked because of their factory association with the [article factory](https://github.com/thepracticaldev/dev.to/blob/master/spec/factories/articles.rb#L20)

I'm fine with this workaround since we want to get parallelization, stat.

One idea I had while investigating though was to pass in the user factory yourself, setting the img to nil. Just an idea. Might be worth a test, before trying this strategy.

```
  let!(:article) { create(:article, reactions_count: 12, featured: true, user: create(:user, profile_image: nil)) }
  let!(:article2) { create(:article, reactions_count: 20, featured: true, user: create(:user, profile_image: nil)) }
```

In future would also be good to look into what `Rack::Test::UploadedFile` is about (never seen it before)

also, just wondering, how are you running things in parallel?

mstruve:
@snackattas I was considering that option as well. I do think having the ignore server block could be useful for the future. But another spec started failing with this issue so I will try making that nil bc I think that would be easier an even cleaner fix.

.travis.yml
```diff
@@ -49,7 +49,7 @@ before_install:
   - sudo systemctl start elasticsearch
 install:
   - bundle config set path 'vendor/bundle'
-  - bundle install --local --jobs 2
+  - bundle install --local --jobs=2
```
mstruve, line 52: Snuck in this little typo fix.

PR #8357, opened 2020-06-08
Add json template to Pages

joshpuetz:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds a `json` template type to `/internal/pages`. The concept is Pages could be used to host lightweight data structures that change infrequently and serve as an alternative to a heavier weight data store like a database or API.

The use case we have for this is dynamic data on the CodeLand page: as videos are aired, we want to update information about the talk's speaker, title, and summary. Instead of adding this information to our database or redis and serving it over an API (all of which would require additions to the codebase), we can serve up this information AND change it easily by exporting it via a `Pages` page. We'll change this data as different videos air in order to update clients that are currently watching the content without requiring them to refresh their pages.

The only problem is that currently `pages` have one of two layouts (`contained` or `full_within_layout`) both of are HTML layouts with a full header, styles, etc. The `json` template options serves up the contents of a page as content-type `application/json`.

Originally I added this feature as a template option to `html_variants`, but I think it makes more sense as a `pages` addition:
- `html_variants` are designed around testing different pieces of content in the same spot. For example: you could have 3 different banners that are rotated. It's unlikely data would be treated in this manner
- `pages` already contains code for setting cache headers
- `pages` allow admins to change content in place, whereas `html_variants` requires deactivation and reactivation
- `html_variants` are only exposed by id right now, `pages` have friendly slug names  

Let's make some data...

Pages

..and then let's view it!

localhost_3000_page_codeland_data

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Jason?

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -7,6 +7,7 @@ class PagesController < ApplicationController
     not_found unless FeatureFlag.accessible?(@page.feature_flag_name, current_user)
 
     set_surrogate_key_header "show-page-#{params[:slug]}"
+    render json: @page.processed_html if @page.template == "json"
   end
 
   def about
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -157,7 +157,12 @@ class StoriesController < ApplicationController
   def handle_page_display
     @story_show = true
     set_surrogate_key_header "show-page-#{params[:username]}"
-    render template: "pages/show"
+
+    if @page.template == "json"
+      render json: @page.processed_html
+    else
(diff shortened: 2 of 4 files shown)


joshpuetz:
Great points @Ridhwana and @rhymes! I'll try to improve the UI, and I'm down to add a `body_json` field on the `pages` table.

joshpuetz:
Repining @Ridhwana and @rhymes for input on my UX additions!

PR #8361, opened 2020-06-08
Provide default crayon styles when creating banners

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

- [x] Imports crayons styles into `/internal` pages.
- [x] Allows admins to use crayons banner styles when creating broadcasts.
- [x] Previews banner styles with crayons.
- [x] Fixes a UI bug with buttons on the `/internal/broadcast/:id/edit` page.

Closes https://github.com/thepracticaldev/dev.to/issues/8355.

### HOW TO QA THIS PR:

**1. Go to the `/internal/broadcasts` page locally, and pick a `type_of: Announcement` broadcast that you want to activate, or is already active.**

**2. Choose a crayons banner style, or leave it blank. If blank, you should see plain old HTML represented:**
Screen Shot 2020-06-08 at 3 18 22 PM

**3. If you select a banner style, you should be able to save the broadcast, and see the banner style reflected. For example, with the style of `warning`:**
Screen Shot 2020-06-08 at 3 18 02 PM

**And this is with the style of `brand`:**
Screen Shot 2020-06-08 at 3 17 22 PM

**4. Finally, you should see that same banner style from the crayons design system reflected when you view the active announcement throughout the site:**
Screen Shot 2020-06-08 at 4 00 55 PM

This PR also fixes a UI regression on the `/internal/broadcasts/:id/edit` page:
_Before:_
Screen Shot 2020-06-08 at 3 15 50 PM

_After:_
Screen Shot 2020-06-08 at 3 16 09 PM

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

N/A

alt_text

### DIFF
--- a/app/assets/stylesheets/internal/layout.scss
+++ b/app/assets/stylesheets/internal/layout.scss
@@ -1,4 +1,5 @@
 @import '../variables';
+@import '../crayons';
 
 #siteConfigBodyContainer {
   padding: 20px;
@@ -26,9 +27,9 @@
 }
 
 .buffer-cell {
-    /* Max-width derived from width of a Bootstrap container */
-    max-width: 1140px;
-    word-wrap: break-word;
+  /* Max-width derived from width of a Bootstrap container */
+  max-width: 1140px;
+  word-wrap: break-word;
 }
 
 .bg-featured {
--- a/app/models/broadcast.rb
+++ b/app/models/broadcast.rb
@@ -6,6 +6,7 @@ class Broadcast < ApplicationRecord
   validates :title, uniqueness: { scope: :type_of }, presence: true
   validates :type_of, :processed_html, presence: true
   validates :type_of, inclusion: { in: %w[Announcement Welcome] }
+  validates :banner_style, inclusion: { in: %w[default brand success warning error] }, allow_blank: true
   validate  :single_active_announcement_broadcast
 
   scope :active, -> { where(active: true) }
@@ -14,6 +15,16 @@ class Broadcast < ApplicationRecord
 
   def get_inner_body(content)
     Nokogiri::HTML(content).at("body").inner_html
+  end
+
+  def banner_class
+    return if banner_style.blank?
+
+    if banner_style == "default"
+      "crayons-banner"
+    else
+      "crayons-banner crayons-banner--#{banner_style}"
+    end
   end
 
   private
--- a/app/models/broadcast.rb
+++ b/app/models/broadcast.rb
@@ -1,4 +1,5 @@
 class Broadcast < ApplicationRecord
+  VALID_BANNER_STYLES = %w[default brand success warning error].freeze
   resourcify
 
   has_many :notifications, as: :notifiable, inverse_of: :notifiable
@@ -6,7 +7,7 @@ class Broadcast < ApplicationRecord
   validates :title, uniqueness: { scope: :type_of }, presence: true
   validates :type_of, :processed_html, presence: true
   validates :type_of, inclusion: { in: %w[Announcement Welcome] }
-  validates :banner_style, inclusion: { in: %w[default brand success warning error] }, allow_blank: true
+  validates :banner_style, inclusion: { in: VALID_BANNER_STYLES }, allow_blank: true
   validate  :single_active_announcement_broadcast
 
   scope :active, -> { where(active: true) }
@@ -15,16 +16,6 @@ class Broadcast < ApplicationRecord
 
   def get_inner_body(content)
     Nokogiri::HTML(content).at("body").inner_html
-  end
-
-  def banner_class
-    return if banner_style.blank?
-
-    if banner_style == "default"
-      "crayons-banner"
-    else
-      "crayons-banner crayons-banner--#{banner_style}"
-    end
   end
 
   private
--- a/app/views/internal/broadcasts/_form.html.erb
+++ b/app/views/internal/broadcasts/_form.html.erb
@@ -22,7 +22,7 @@
 <% if @broadcast.processed_html %>
   <div>
     </p><strong>Preview</strong></p>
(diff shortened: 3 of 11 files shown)


app/assets/stylesheets/internal/layout.scss
```diff
@@ -1,4 +1,5 @@
 @import '../variables';
+@import '../crayons';
```
vaidehijoshi, line 2: I was toying with the ideal of just importing _certain_ crayons styles into our `/internal` pages, like colors + banners...but then realized that someone might want to use _other_ crayons variables inside the html, too. This _might_ be too heavyweight to import, but did it here and figured that someone would tell me if this was a bad idea or not :)

app/models/broadcast.rb
```diff
@@ -16,6 +17,16 @@ def get_inner_body(content)
     Nokogiri::HTML(content).at("body").inner_html
   end
 
+  def banner_class
```
rhymes, line 20: can we move this in a `BroadcastHelper`? Seems a better place instead of the model object

app/views/internal/broadcasts/_form.html.erb
```diff
@@ -9,6 +9,10 @@
   <%= label_tag :type, "Type:" %>
   <%= select_tag "type_of", options_for_select(%w[Welcome Announcement], selected: @broadcast.type_of) %>
 </div>
+<div class="form-group">
+  <%= label_tag :banner_style, "Banner Style:" %>
+  <%= select_tag "banner_style", options_for_select(%w[default brand success warning error], selected: @broadcast.banner_style), include_blank: true %>
```
rhymes, line 14: the repetition of the styles reminds me we should put it in a contant in the model

db/migrate/20200608175130_add_banner_style_to_broadcasts.rb
```diff
@@ -0,0 +1,5 @@
+class AddBannerStyleToBroadcasts < ActiveRecord::Migration[6.0]
+  def change
+    add_column :broadcasts, :banner_style, :string
```
rhymes, line 3: should this have a `null: false` plus default value or it doesn't matter if existing broadcasts don't have a style set?

mstruve:
Tested, works great! 2 questions
1) is the preview supposed to update as you change the banner options? I had to save then click back to the page to get the preview to update. Totally fine if that is intended. Might also be nice to redirect back to the individual broadcast after the update instead of the index.
2) Any idea why this happens? I assume its supposed to push everything down. 
Screen Shot 2020-06-09 at 9 36 43 AM

vaidehijoshi:
@mstruve:
> 1. is the preview supposed to update as you change the banner options? I had to save then click back to the page to get the preview to update. Totally fine if that is intended. Might also be nice to redirect back to the individual broadcast after the update instead of the index.

nope, this is just plain rails so you need to persist the record in order to see the changes in the preview. FWIW, we don't actually have a "show" view for the broadcast record. i can add that later on (i'm sure we'll want it) and fix the redirect issue in a separate PR, since i'll be working in this area of the codebase throughout the cycle.

> 2. Any idea why this happens? I assume it's supposed to push everything down.

This is intentional (see [this comment in GitHub](https://github.com/thepracticaldev/dev.to/pull/8301#discussion_r436102261) and [this one in Slack](https://dev-internal.slack.com/archives/C7GE84DEJ/p1591315623108200). Basically we want the announcement to render directly below the nav, but I haven't shipped the `x` close button yet (which is fine because we aren't using this feature even though it's in production!). It's coming in the next PR (I don't like overloading PRs and prefer to split them up 😛)

db/migrate/20200608175130_add_banner_style_to_broadcasts.rb
vaidehijoshi: No, I don't think it needs this; welcome broadcasts don't require a style at all (we don't use them), and you don't even _need_ a style set for announcement broadcasts, either (though we ought to recommend it). So, there's no reason to enforce this on a broadcast-wide basis.

app/views/internal/broadcasts/_form.html.erb
vaidehijoshi: Ah, good call 👍

app/models/broadcast.rb
vaidehijoshi: @rhymes Hm, so the only reason I didn't do that was because I needed access to it in the `async_info_controller` so that it can be [serialized and used](https://github.com/thepracticaldev/dev.to/pull/8361/files#diff-02a7d87efea39bec38ecc2737b8eba6cR49) in the frontend. I was under the impression that the helper can only be used in views? ...But maybe I am wrong, and it actually would be accessible from that controller?

rhymes: yeah, you can access helpers from controllers with https://api.rubyonrails.org/classes/ActionController/Helpers.html#method-i-helpers

db/migrate/20200608175130_add_banner_style_to_broadcasts.rb
rhymes: ah, make sense! :)

app/models/broadcast.rb
vaidehijoshi: Addressed in last commit!

app/views/internal/broadcasts/_form.html.erb
vaidehijoshi: Addressed in last commit!

vaidehijoshi:
Since I've addressed @rhymes's comments + have enough approvals, I'm going to go ahead and deploy this! 🚀

Issue #8370, opened 2020-06-09
Errors when running `bundle exec erblint --lint-all`


AlbertoPdRF:
**Describe the bug**

Running `bundle exec erblint --lint-all` outputs `130 error(s) were found in ERB files`.

**To Reproduce**

1. Run `bundle exec erblint --lint-all`
2. See 130 errors

**Expected behavior**

There shouldn't be any errors.

**Screenshots**

**Desktop (please complete the following information):**

- OS: Ubuntu 20.04 (WSL2)
- Browser:
- Version:

**Smartphone (please complete the following information):**

- Device:
- OS:
- Browser:
- Version:

**Additional context**

I found this while working on #8369, as one error (fixed on that PR) was preventing me from creating a commit with changes on the file with such error. I had to fix that error manually as the pre-commit hook, which runs erblint with the `--autocorrect` flag, wouldn't work.

I'm happy to work on fixing this issue!

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

rhymes:
Time to take a second stab at this I guess :D

First part was https://github.com/thepracticaldev/dev.to/issues/1842

AlbertoPdRF:
As that's how it was done in #1842, I'll submit a separate pull request for each of the `app/views/` folder (e.g. `app/views/listings`) to keep them small and easier to review.

rhymes:
Thank you @AlbertoPdRF! Small PRs are the key :D

AlbertoPdRF:
There are PRs to fix all the errors already, but there's still something I haven't fixed. If I run `bundle exec erblint --lint-all` I have the following output:

<details>
<summary>Output</summary>

```ruby
Linting 438 files with 12 linters...

Exception occured when processing: app/views/internal/listings/index.html.erb
If this file cannot be processed by erb-lint, you can exclude it in your configuration file.
Parser::Source::Range: end_pos must not be less than begin_pos
[...]

PR #8374, opened 2020-06-09
Fix erblint errors on the `app/views/listings` folder

AlbertoPdRF:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR fixes the errors resulting from running `bundle exec erblint app/views/listings`.

This PR is part of #8370.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Fix errors gif


app/views/listings/dashboard.html.erb
```diff
@@ -1,18 +1,9 @@
...
-)%>" data-usercredits="<%= @user_credits %>"
-      data-orglistings="<%= @org_listings.to_json(
-        only: %i[title tag_list created_at expires_at bumped_at updated_at category id user_id slug organization_id location published],
-        include: {
-          author: { only: %i[username name], methods: %i[username profile_image_90] },
-        },
-      ) %>" data-orgs="<%= @orgs.to_json(
-        only: %i[id name slug unspent_credits_count]
-      ) %>" >
+<div id="listings-dashboard"
```
rhymes, line 3: I was thinking, could we move these `.to_json()` methods in controller variables? what do you think?

app/views/listings/index.html.erb
```diff
@@ -38,7 +38,7 @@
     <div class="listing-filters-categories">
       <a href="<%= listings_path %>" class="<%= "selected" if params[:category].blank? %> data-no-instant=" true ">all</a>
       <% categories_for_display.each do |cat| %>
-        <a href="<%= listing_category_path(category: cat[:slug]) %>" class="<%= "selected" if params[:category] == cat[:slug] %> data-no-instant="true"><%= cat[:name] %></a>
+        <a href="<%= listing_category_path(category: cat[:slug]) %> " class="<%= "selected" if params[:category] == cat[:slug] %> data-no-instant="true"><%= cat[:name] %></a>
```
rhymes, line 41: weird that it created a space between the tag and the quote

app/views/listings/dashboard.html.erb
AlbertoPdRF: That sounds good to me, it will definitely look cleaner.

app/views/listings/index.html.erb
AlbertoPdRF: I still don't fully-get why this space is necessary to be honest, as erblint isn't complaining for other similar lines 😕

AlbertoPdRF:
How does this look @rhymes?

(I had to force-push because I had rebased my branch with `master`, should I not rebase in the future unless it's necessary?)

rhymes:
@AlbertoPdRF we encourage people not to force push after the first review has been given, unless there's a need to incorporate something from master, but no worries, you didn't know and it's also not a big deal :)

app/views/listings/index.html.erb
citizen428: I'm not sure why it's needed but I'm definitely guilty of using `--no-verify` to avoid adding it. 😉

Issue #8170, opened 2020-05-29
Create a new database table to hold subscriptions


atsmith813:
_We are building a liquid tag where permitted authors can embed a button that users/readers can click to share the email address associated with their DEV account with the author - almost like a subscribe button._

**We'll likely want to explore a polymorphic set up to have this fit with comments too**
- Model
- Migration
- Specs
- Associations
- Validations

atsmith813:
To our awesome contributors, please note that this issue will be worked on by a DEV team member and is not up for grabs.

atsmith813:
Completed with #8375.

PR #8375, opened 2020-06-09
Email signups liquid tag backend

atsmith813:
- [x] Feature

## Description
In this PR I'm setting up the backend for email signups which includes:
- Migration
- Model
- Associations
- Validations
- Spec

I chose to go polymorphic so we can easily extend the liquid tag to other contexts (i.e. `Comments`) in the future. Users (subscribers) can subscribe to other users (authors) via some content (articles, comments, etc.).

I want to draw attention to the naming. My plan is to call it the `emailsignup` liquid tag and use that wording on the frontend. This current setup is also great for if we decide to create more liquid tags to share more than just email in the future 🙃 .

We are building a liquid tag where permitted authors can embed a button that users/readers can click to share the email address associated with their DEV account with the author - almost like a subscribe button.

_Details are still hazy, but we'll be restricting this liquid tag only for use in `Articles`, not `Comments`, or `Pages`. We may also restrict it to a new role to start._
- [ ] [Create a new endpoint to handle email signups from the new liquid tag](https://github.com/thepracticaldev/dev.to/issues/8171)
- [ ] [Create an email signup button liquid tag](https://github.com/thepracticaldev/dev.to/issues/8168)
- [ ] [Add new tab to User dashboard to display email addresses from the email signup button liquid tag](https://github.com/thepracticaldev/dev.to/issues/8169)

#8170 

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

youtube_subscribe_gif


citizen428:
I'm just throwing a totally random thought out there, feel free to ignore it completely: we are very concerned with privacy and keeping the amount of data we store reasonably small. How much of these relationships do we really need to store permanently/know about? Could it be enough to just send an email to the author's registered email address saying "X subscribed to Y" and they process this information in whichever way suits them (probably not if we want to offer list management like features like downloading all subscribers as CSV)? There's probably a minimum amount of data we need to store to avoid spamming, but other than that how much is necessary and how much is nice to have?

atsmith813:
That's a great point @citizen428 and something to keep top of mind 🤔 .

FWIW, the use of this liquid tag will likely start out limited to a new role and certain users as it's for CodeLand.

app/models/article.rb
```diff
@@ -28,6 +28,8 @@ class Article < ApplicationRecord
   counter_culture :organization
 
   has_many :comments, as: :commentable, inverse_of: :commentable
+  has_many :user_subscriptions, as: :user_subscription_sourceable
+  has_many :subscribers, class_name: "User", through: :user_subscriptions, foreign_key: :user_id
```
mstruve, line 32: I think we should take this out completely or rename it to something along the lines of `sourced_author_subscribers` or `sourced_subscribers` . The reason I suggest this is bc the users are not subscribing to the article they are subscribing to the author and I think this relationship could make that confusing.

atsmith813: I'm game, how about `source_author_subscribers` (`source` > `sourced`)?

app/models/user_subscription.rb
```diff
@@ -0,0 +1,12 @@
+class UserSubscription < ApplicationRecord
+  ALLOWED_TYPES = %w[Article].freeze
+
+  belongs_to :author, class_name: "User", foreign_key: :author_id, inverse_of: :user_subscriptions
+  belongs_to :subscriber, class_name: "User", foreign_key: :subscriber_id, inverse_of: :user_subscriptions
+  belongs_to :user_subscription_sourceable, polymorphic: true
```
rhymes, line 6: can you explain this?

atsmith813: @rhymes sure! Do you mean how polymorphism works? Or the context of the relationships?

atsmith813:
@rhymes A user can subscribe to another user. For now, subscribing simply means the user is sharing the email address associated with their DEV account, through explicit consent, with the author of an article or comment.

So an approved author will be able to see all users that elected to subscribe (share their email address) with them. And from where they subscribed (specific article or comment).

app/models/article.rb
atsmith813: Updated to `sourced_subscribers` 🚀

app/models/user_subscription.rb
atsmith813: This is part of the complexity of polymorphic and how you "auto-magically" setup polymorphic to work by telling this `UserSubscription` model that it's a part of that relationship. It's like saying, "the `UserSubscription` belongs to the `user_subscription_sourceable` entity". It's usually more straight forward when we keep the names the same (i.e. `:user_subscribable`). I don't know the technical term as far as grammar goes...but making the word "-able" is a convention for polymorphic associations. And when you look at the migration, you'll see:

```ruby
t.references :user_subscription_sourceable, polymorphic: true
```

From that, Rails will do its magic and create the `user_subscription_source_type` and `user_subscription_source_id` columns for you. `type` and `id` are what we need to create the object we're referencing. So that would look like: 
```
user_subscription_source_type: "Article"
user_subscription_source_id: 1
```
which is saying that the `source` of the `UserSubscription` is the `Article` object with `id` of `1`. It's convenient because Rails/ActiveRecord will return the actual `Article` record by simply calling `UserSubscription.first.user_subscription_sourceable`.

And the name `user_subscription_sourceable` is something we choose. In other words, we're saying this object is `sourceable` (can be a source) for a `UserSubscription`.

I think it's more clear with a generic example. Say you were to build some review website (like a Yelp clone). And say you had different types of things that could be reviewed (`Restaurant` and `Concert`). Those are 2 different object types that we want to associate with `Review`. So, in this example, we'd likely just call it `reviewable` which would make the columns `reviewable_type` and `reviewable_id`. We're saying both `Restaurants` and `Concerts` are `reviewable`.

In our example, we're saying a `UserSubscription`can be `sourced` from an `Article` or `Comment`.

maestromac:
The potential problem I can think of down the line is the distinction between `follow` and `subscribe`. So you can be a subscriber but not a follower?

atsmith813:
Agreed, the naming here is especially difficult. We also have a concept of subscriptions with regards to notifications as well 😬 .

And yes, a user can be a subscriber, but not a follower. It's similar to other content-driven websites. Using Medium an example, I can "subscribe"/share my email/signup for emails on a super sweet article an author wrote. Maybe they did something like, "sign up for my mailing list for this AWESOME pdf". But I don't want to follow the author, because maybe they mostly post about a different topic.

atsmith813:
One option, if people find it more clear, is to change `subscription` to `sign_up`.

citizen428:
@atsmith813 Is the word "disclose" to specific? A user sharing his email is a "discloser", the action is a "disclosure" etc. Maybe slightly awkward verbiage, but at least we'd have zero domain language overlap.

app/models/user_subscription.rb
rhymes: ahahahha sorry, I should have been more clear so to avoid having you spend so much time replying to what was essentially the following question: why do we differentiate sources? 

But you also replied to that just now, thank you and sorry again for having you explain the entire context! I just didn't understand why a person would be subscribed to either an article or a comment. But I was wrong, it's not subscribed to an article or a comment, it's tracking where the subscription came from!

citizen428:
I think what @vaidehijoshi said is pretty much necessary. 

I'm still worried about the naming overlap with the unrelated concept of subscribing to articles/comments, especially given how much work it was to retroactively fix the domain language for listings, but don't let that stop you from merging @atsmith813.

PR #8379, opened 2020-06-10
Flaky Spec Fixes: Override Warden Test Helpers to Prevent Auth Race Conditions

mstruve:
- [x] Bug Fix

## Description

alt_text


spec/requests/dashboard_spec.rb
```diff
@@ -221,7 +221,7 @@
         org = create :organization
         create(:organization_membership, user: user, organization: org, type_of_user: "admin")
         user.add_role(:pro)
-        login_as user
+        sign_in user
```
mstruve, line 224: let's keep things consistent and always use `sign_in` for our sepcs

spec/rails_helper.rb
```diff
@@ -77,6 +77,14 @@
   config.include SidekiqTestHelpers
   config.include ElasticsearchHelpers
 
+  config.after(:each, type: :system) do
```
citizen428, line 80: This is 🔥 

The only thing that worries me a bit is that `_on_request` is considered private (the doc has a `:api: private` annotation), so it may change at any time. I wonder if [`Warden.test_mode!`](https://www.rubydoc.info/github/hassox/warden/Warden#test_mode!-class_method) in combination with [`Warden.test_reset!`](https://www.rubydoc.info/github/hassox/warden/Warden/Test/WardenHelpers#test_reset!-instance_method) would also solve our problem, I looked through some of my old Rails projects and that's the approach used there.

PR #8380, opened 2020-06-10
Fix a11y issues flagged by axe on homepage

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This commit brings the number of violations found on homepage (local,
not prod) down from ~100 to ~10. It should have no serious impact on the
visual layout or design of the page.

Before:
Screenshot from 2020-06-09 20-57-25

After:
Screenshot from 2020-06-09 20-59-56

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/views/articles/index.html.erb
+++ b/app/views/articles/index.html.erb
@@ -36,7 +36,8 @@
     <%= render "articles/sidebar" %>
 
     <%# BEGIN Feed menu bar %>
-    <div class="articles-list crayons-layout__content" id="articles-list">
+    <main class="articles-list crayons-layout__content" id="articles-list" role="main">
+      <h1 style="position: absolute; left: -10000px; width: 1px; height: 1px; overflow: hidden;">Articles</h1>
       <%= render(partial: "onboardings/task_card") if user_signed_in? %>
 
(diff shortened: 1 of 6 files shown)


app/views/articles/index.html.erb
```diff
@@ -36,7 +36,8 @@
     <%= render "articles/sidebar" %>
 
     <%# BEGIN Feed menu bar %>
-    <div class="articles-list crayons-layout__content" id="articles-list">
+    <main class="articles-list crayons-layout__content" id="articles-list" role="main">
+      <h1 style="position: absolute; left: -10000px; width: 1px; height: 1px; overflow: hidden;">Articles</h1>
```
jacobherrington, line 40: This is something worth discussing.

We should think about doing this in more places to improve the performance of screen readers. I added this here to satisfy axe, but it could be used all over the app.

There are a few methods for adding hidden headings that will still be useful for screen readers, this is just the one I picked.

Issue #8377, opened 2020-06-09
Repositories still appears on profile sidebar after deleted from Github, but doesnt appear on configurations (impossible to remove them from the profile sidebar)


felippe-regazio:
Deleted repositories still appear on integrations list. If you go to my profile https://dev.to/felipperegazio you will see two repositories on right sidebar. there were deleted, so they doesnt appear on configurations > integrations anymore (so i cant remove them from that list), but still appears on the sidebar with a 404 target link.

Try:

1. Integrate with github.
2. Pick a repository to show on your profile.
3. Delete this repository
4. Try to remove the deleted repo from your profile.

**Expected behavior**

The repositories on the sidebar must disappear when deleted from github, or the (remove) action must be linked to the repos already listed on the profile sidebar also.

**Screenshots**

Not needed since the link on the description tells the history.

**Desktop (please complete the following information):**

- OS: Ubuntu 18
- Browser: Chrome
- Version: 80.0.3987.132 (Official Build) (64-bit)

**Smartphone (please complete the following information):**

- Device: N/A
- OS: N/A
- Browser: N/A
- Version: N/A

**Additional context**

citizen428:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

rhymes:
@felippe-regazio could you try again? We shipped #8382 which should help!

felippe-regazio:
Hello @rhymes !

If you visit my profile https://dev.to/felipperegazio and click on any of the repositories on my left side menu, you will see they doenst exist anymore. anyway they still there, and they still are not on the integrations list, so i still cant remove them.

rhymes:
@felippe-regazio if you visit https://dev.to/settings/integrations, wait for the list to load and then visit https://dev.to/felipperegazio - do they disappear?

felippe-regazio:
@rhymes  yes they disappeared, from the `/integrations` and from my profile side menu also now. i think the issue has been solved : )

rhymes:
[...]

PR #8382, opened 2020-06-10
Clean up inaccessible repositories

citizen428:

- [X] Bug Fix

#8377 

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed

alt_text

### DIFF
--- a/app/controllers/github_repos_controller.rb
+++ b/app/controllers/github_repos_controller.rb
@@ -5,11 +5,11 @@ class GithubReposController < ApplicationController
   def index
     authorize GithubRepo
 
-    known_repositories_ids = current_user.github_repos.featured.distinct.pluck(:github_id_code)
+    known_repositories = current_user.github_repos.featured.distinct.to_a
 
     # NOTE: this will invoke autopaging, by issuing multiple calls to GitHub
     # to fetch all of the user's repositories. This could eventually become slow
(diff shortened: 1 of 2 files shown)


app/controllers/github_repos_controller.rb
```diff
@@ -38,13 +38,18 @@ def update_or_create
...
-      repo.featured = known_repositories_ids.include?(repo.id)
+    repos = client.repositories(visibility: :public).map do |repo|
+      repo.featured = known_repositories.delete_if { |known| known.github_id_code == repo.id }.present?
       repo
-    end.sort_by(&:name)
+    end
+
+    # Remove pinned repositories that were removed from GH or are now private,
+    # since the user will not be able to remove them by themselves.
+    known_repositories.each(&:destroy)
```
citizen428, line 50: Since we're using `delete_if` above, this will only contain the repositories we still consider featured but that are no longer part of the user's repos with public visibility according to GH.

citizen428:
Ha, good point @rhymes, probably worth investigating.

citizen428:
@rhymes I now added a spec and marked this PR as ready for review. I have no insight yet as to when/how often this task runs and this fix is not particularly intrusive so should be ok to run additionally.

PR #8391, opened 2020-06-10
Conditionally install foreman during setup

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We should avoid install foreman during `bin/setup` if it is already
installed.

I'm not sure if we should dictate a version here - anyone have thoughts?

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -20,7 +20,7 @@ chdir APP_ROOT do
   system! "bundle config --local cache_all_platforms true"
   system("bundle check") || system!("bundle install")
 
-  system! "gem install foreman"
+  system! "gem list \"^foreman$\" -v 0.87.1 -i --silent || gem install foreman"
 
   # Install JavaScript dependencies if using Yarn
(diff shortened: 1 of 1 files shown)


benhalpern:
@jacobherrington @rhymes yeah could we check what version of foreman exists and include that in the logic here?

jacobherrington:
@benhalpern At what version do we force an update? It's worth noting that [foreman seems really stable](https://rubygems.org/gems/foreman/versions).

rhymes:
@jacobherrington i think we can just simply check the latest available, if we ever break foreman in development we'll know soon enough as we all use it

jacobherrington:
@rhymes @benhalpern Updated!

PR #8394, opened 2020-06-10
Update Users::Sidebar not to cache AR object

maestromac:
- [x] Optimization

## Description
Prefer caching ActiveRecord ids instead of ActiveRecord objects. Also updated cache name for clarity and invalidation.
https://github.com/thepracticaldev/dev.to/projects/9#card-39519016
n/a
## Added tests?
- [x] improved existing spec

## Added to documentation?
- [x] no documentation needed

n/a


app/services/suggester/users/sidebar.rb
```diff
@@ -4,33 +4,39 @@ class Sidebar
...
-            where(id: user_ids).
-            order("reputation_modifier DESC").limit(20).to_a
-          group_two = User.select(:id, :name, :username, :profile_image, :summary).
-            where(id: user_ids).
-            order(Arel.sql("RANDOM()")).limit(20).to_a
-          (group_one + group_two).uniq
+        suggested_user_ids = Rails.cache.fetch(generate_cache_name, expires_in: 120.hours) do
+          (reputable_user_ids + random_user_ids).uniq
         end
+        User.select(:id, :name, :username, :profile_image, :summary).where(id: suggested_user_ids).to_a
```
rhymes, line 14: is there a reason to make this array instead of keeping it an active record relation?

app/services/suggester/users/sidebar.rb
```diff
@@ -4,33 +4,39 @@ class Sidebar
       def initialize(user, given_tag)
         @user = user
         @given_tag = given_tag
+        @minimum_reaction_count = Rails.env.production? ? 25 : 0
```
rhymes, line 7: what if we just keep it to 25? is there a reason why in development it should be zero?

maestromac: With a minimum of 25 reactions, I don't think any local dev user will meet that threshold from the get-go. With 0, it will ensure we will always see suggested user when browsing to `localhost:3000/t/anytagpage`. 

Yeah I agree this code doesn't look very peachy to me either and I'm okay with keeping it locked at 25 but I also don't want us dev to run into confusion when developing. What do you think?

app/services/suggester/users/sidebar.rb
maestromac: I was keeping to identical to the original spec but you are right, it doesn't have to be an array. Updated

app/services/suggester/users/sidebar.rb
rhymes: let's keep it!

app/services/suggester/users/sidebar.rb
```diff
@@ -4,33 +4,39 @@ class Sidebar
...
       end
 
       private
 
-      attr_reader :user, :given_tag
+      attr_reader :user, :given_tag, :minimum_reaction_count
 
       def generate_cache_name
-        "tag-#{given_tag}_user-#{user.id}-#{user.last_followed_at}/tag-follow-sugggestions"
+        "tag-#{given_tag}-user-#{user.id}-#{user.last_followed_at}/tag-follow-sugggestions"
```
mstruve, line 22: May as well pull the extra g out of suggestions here 😅

app/services/suggester/users/sidebar.rb
```diff
@@ -4,33 +4,39 @@ class Sidebar
...
+        @active_authors_for_given_tags ||= Article.published.tagged_with([given_tag], any: true).
+          where("public_reactions_count >= ?", minimum_reaction_count).
+          where("published_at > ?", 4.months.ago).
+          where("user_id != ?", user.id).
+          where.not(user_id: user.following_by_type("User")).
+          pluck(:user_id)
+      end
+
+      def reputable_user_ids
+        User.where(id: active_authors_for_given_tags).order("reputation_modifier DESC").limit(20).pluck(:id).to_a
```
mstruve, line 35: No need for `to_a` here or the line below since [Pluck returns an Array of attribute values](https://apidock.com/rails/v6.0.0/ActiveRecord/Calculations/pluck)

app/services/suggester/users/sidebar.rb
maestromac: Nice catch 😆

Issue #7600, opened 2020-04-29
Contrasts of colours makes it hard to read email address


bernardbaker:
**Describe the bug**

I was viewing somebody who placed the heart on my article. When I went to their profile I couldn't read the email address. Because of the contrast of colours.

**To Reproduce**

1. Go to profile in the picture.

![Screenshot_20200429-205512](https://user-images.githubusercontent.com/13556172/80641144-5374a880-8a5c-11ea-85bd-76a05c39c049.png)

2. Read the email address.

**Expected behavior**

I should be able to read the email address.

**Screenshots**

**Desktop (please complete the following information):**

- OS:
- Browser:
- Version:

**Smartphone (please complete the following information):**

- Device: pixel 3 xl
- OS: android
- Browser: dev.to app
- Version: n/a

**Additional context**

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

rhymes:
Il looks like this depends on which colors the user has set:

https://github.com/thepracticaldev/dev.to/blob/c734457a4447520c5cc30a9d41496b4ef22ce5c7/app/views/articles/_user_metadata.html.erb#L2-L4

but as we know from the discussion here https://github.com/thepracticaldev/dev.to/issues/7434 that logic fails to take into account theming.

rhymes:
I believe this was fixed by https://github.com/forem/forem/pull/8406

@bernardbaker let us know otherwise

PR #8406, opened 2020-06-11
Fix Colour Contrast on the User Profile Header

Oghenebrume50:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR attempts to fix the issue #7600 and #7998 basically both issues complained about the username and/or the user email being hidden when a certain theme is used, mainly the X hacker theme or the Night theme.
One thing I did was to set the text to the body text color because this seems to be well contrasted against the background-color, I do not know if this is a good approach but it gets the job done properly. 
I am kind of skeptical because I am not really sure what the HexComparer does but the color it sets in these cases, causes the bad contrast.  But I noticed that in the case of the user email the <a> tag was set to a different color hence the problem same with h1 tag in the case of the username,  the body color though sets a better contrast.

This is how it looks on the Night theme
devprofile1

This is how it looks on the X Hacker theme
devprofile2

All other themes are great too.
## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

## [optional] Are there any post-deployment
No

## tasks we need to perform?
No

About to go down


nickytonline:
cc: @ludwiczakpawel

app/views/articles/_user_metadata.html.erb
```diff
@@ -1,6 +1,6 @@
 <style>
   .user-metadata-details-inner a {
-    color: <%= HexComparer.new([user_colors(@user)[:bg], user_colors(@user)[:text]]).brightness(0.95) %>
+    color: var(--body-color);
```
lisasy, line 3: very nice!

Oghenebrume50:
@rhymes any input on this?

rhymes:
@Oghenebrume50 I think it's sound but I know there was a reason for the `HexComparer`, though I don't know exactly which, thus I requested a review from team members who might have a wider view on the subject

app/views/articles/_user_metadata.html.erb
```diff
@@ -1,6 +1,6 @@
...
   }
 </style>
```
ludwiczakpawel, line 5: Readability issue here happens (and is being reported) often enough to consider it harm to customize text colors with profile/brand colors. So my thinking is we should NOT use profile/brand colors for any texts. Instead profile/brand colors should only be used as decorations so they never ever affect readability of important information.

That being said I suggest we remove this entire custom styling here instead of replacing link color with `var(--body-color)` - it's not ideal because we actually DO want to differentiate links within text.

app/views/users/show.html.erb
```diff
@@ -20,7 +20,7 @@
       color: #cedae2;
     }
     .user-profile-header .user-profile-header-container .profile-details h1{
-      color: <%= HexComparer.new([user_colors(@user)[:bg], user_colors(@user)[:text]]).brightness(0.78) %> !important;
```
ludwiczakpawel, line 23: Same here, I think we should just drop this style entirely as well as customization below for night theme: https://github.com/thepracticaldev/dev.to/pull/8406/files#diff-c7c9a522f39f5d8cd9b512cd928b2d14L25-L27

Oghenebrume50:
@rhymes @ludwiczakpawel  I had followed @ludwiczakpawel review and I think the output looks amazing

How it looks on the night theme 
nigthem

How it looks on the Xhacker theme
xhackthem

nickytonline:
Thanks for the PR @Oghenebrume50! 👏🏻

PR #8407, opened 2020-06-11
Replace activerecord-import with Rails 6 insert_all and upsert_all

rhymes:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We're currently using the gem `activerecord-import` to import records in bulk (like with credits and podcast episodes) and take advantage of PostgreSQL builtin `UPSERT` capabilities see [INSERT ON CONFLICT DO UPDATE](https://www.postgresql.org/docs/11/sql-insert.html#SQL-ON-CONFLICT) and its support for partial indexes.

Starting in Rails 6, see [Bulk insert support in Rails 6](https://blog.bigbinary.com/2019/04/15/bulk-insert-support-in-rails-6.html), the gem is not necessary anymore as they've added (with a clearer syntax in my opinion) support for bulk insert via [insert_all](https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-insert_all) and [insert_all!](https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-insert_all-21) and support for upsert (which means insert or update) via [upsert_all](https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-upsert_all) and its shortcut [upsert](https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-upsert).

The big advantage here is eliminating race conditions of two jobs concurrently running the same insertions, in the upsert case, by letting the DB.

Currently we still have an open issue related to this, https://github.com/thepracticaldev/dev.to/issues/3583, which probably hasn't popped up out of luck but technically is still present.

Funnily enough, the code to solve this issue was already added by @lightalloy in https://github.com/thepracticaldev/dev.to/pull/3012 but never activated as it didn't uncomment the patch in #3010. 

While replacing the gem, I also re-activated the fix, so that we should be safe from race conditions.

Closes https://github.com/thepracticaldev/dev.to/issues/3583

## Added tests?

- [x] yes
- [ ] no, because I need help


spec/models/notification_spec.rb
```diff
@@ -21,79 +21,55 @@
   end
 
   describe "when trying to create duplicate notifications" do
-    # Duplicate notifications are not allowed even when validations are skipped
-    it "doesn't allow to create a duplicate notification via import" do
```
rhymes, line 25: here we're testing activerecord-import, we're not testing the actual code, which at this moment in time wasn't even called at all :D 

replaced it with some safety tests

app/models/credit.rb
```diff
@@ -19,9 +19,20 @@ class Credit < ApplicationRecord
   end
 
   def self.add_to(user_or_org, amount)
+    return unless amount.positive?
```
mstruve, line 22: 😍

app/services/notifications/notifiable_action/send.rb
```diff
@@ -21,28 +21,32 @@ def call
...
-        Notification.import! notifications,
-                             on_duplicate_key_update: {
-                               conflict_target: conflict_target,
-                               index_predicate: index_predicate,
-                               columns: %i[json_data notified_at read]
-                             }
+
+        return if notifications_attributes.blank?
+
+        upsert_index = choose_upsert_index(action)
```
mstruve, line 44: 🔥

Gemfile
```diff
@@ -11,7 +11,6 @@ group :production do
 end
 
 gem "active_record_union", "~> 1.3" # Adds proper union and union_all methods to ActiveRecord::Relation
-gem "activerecord-import", "~> 1.0" # Adds ability to bulk create activerecord objects
```
citizen428, line 14: Another one bites the dust!

PR #8409, opened 2020-06-11
Hide site-wide messaging for opted-out users

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Takes the `display_announcements` column on the `user` into account when displaying an announcement. If a logge-in user has opted-out of announcements, the banner will not display. If the user has opted-in to announcements (the default!), then the banner will display.

This should not affect logged-out users, since they are not persisted anywhere (although they can dismiss the announcement still via clicking the `X`, which will persist a key in `localStorage`).

Closes https://github.com/thepracticaldev/dev.to/issues/8262.

HOW TO QA THIS PR:
**1. Create an active broadcast via `/internal/broadcasts`:**
Screen Shot 2020-06-11 at 8 54 32 AM

**2. You should see the broadcast render throughout the site:**
Screen Shot 2020-06-11 at 8 55 20 AM

**3. Go to your `/settings/misc` page, and opt-out of the announcements in your settings:** 
Screen_Shot_2020-06-11_at_8_54_56_AM

**4. Navigate throughout the site; you shouldn't see the banner displayed anywhere!**
Screen Shot 2020-06-11 at 8 55 10 AM

**5. If you change your setting back and opt-in again, the banner should render as expected.**

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [x] ~readme~ internal documentation
- [ ] no documentation needed

N/A

### DIFF
--- a/app/assets/javascripts/initializers/initializeBroadcast.js
+++ b/app/assets/javascripts/initializers/initializeBroadcast.js
@@ -80,14 +80,20 @@ function renderBroadcast(broadcastElement, data) {
 }
 
 /**
- * A function to determine if a broadcast should render
- * Does not render broadcast it has already been inserted,
- * or if the key for the broadcast's title exists in localStorage.
+ * A function to determine if a broadcast should render.
+ * Does not render a broadcast if the current user has opted-out.
+ * Does not render a broadcast it has already been inserted, or
+ * if the key for the broadcast's title exists in localStorage.
  *
  * @function initializeBroadcast
  */
 function initializeBroadcast() {
+  const user = userData();
   const data = broadcastData();
+
+  if (user && !user.display_announcements) {
+    return;
+  }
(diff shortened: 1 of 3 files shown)


vaidehijoshi:
@atsmith813 

> the broadcast blocks the settings menu

I intentionally set the css so that the announcement would render fixed with a higher z-index, but I do agree it looks a bit...odd. I have a fix for this, but I'll PR it separately since adding tiny one-line CSS changes scares the living daylights out of me 😨 

> When creating a new broadcast with errors, I get redirected back to the index page and lose everything I typed.

I'll fix this too, separately. I never noticed this but good catch on that!

PR #8413, opened 2020-06-11
Allow Knapsack to work for Forks

mstruve:
In order to get KnapsackPro to work on forks we need [to use this workaround from the Knapsack docs](https://knapsackpro.com/faq/question/how-to-make-knapsack_pro-works-for-forked-repositories-of-my-project)


.travis.yml
```diff
@@ -65,7 +65,7 @@ script:
   - bundle exec rails webpacker:compile
   - bundle exec rails db:schema:load
   - './cc-test-reporter before-build'
-  - 'bundle exec rake knapsack_pro:rspec'
+  - 'bundle exec bin/knapsack_pro_rspec'
```
rhymes, line 68: I think it's just `bin/...` as it calls `bundle exec` inside

PR #8426, opened 2020-06-11
Flaky Spec Fix: Stub Requests to Elasticsearch For some js: true Specs

mstruve:
- [x] Bug Fix
- [x] Optimization
- [x] Documentation Update

## Description
Prior to this change, in order to fix a flaky js spec that had unwanted data showing up on the page we would have to clear all the data out of Elasticsearch. Clearing out that data is costly. I think a better way to solve this problem is to stub elasticsearch for these specs. I also think stubbing elasticsearch with the `stub_elasticsearch` flag is more clear about what is happening than having the flag `elasticsearch: <class name>`  on your spec.

Stubbing elasticsearch is also a lot faster than clearing it out and hitting it during the spec so +1 for speed up.

Credit to @vaidehijoshi for making rethink this approach to fixing these specs.

@snackattas 

#8388 

This is me putting Elasticsearch in a timeout
alt_text

### DIFF
--- a/docs/tests/test-flags.md
+++ b/docs/tests/test-flags.md
@@ -54,10 +54,10 @@ There are two ways to do this:
 ### `js: true` Flag
 
 `js: true` indicates to our specs that we want the javascript on the page to be
-execute when the page is rendered. One side effect of running our javascript is
+executed when the page is rendered. One side effect of running our javascript in our specs is
 that a lot of pages will hit Elasticsearch. Since we don't clean out
-Elasticsearch between every single spec(because it is very costly) this can lead
-to unexpected data being loaded for a spec. To prevent this use the
+Elasticsearch between every single spec (because it is very costly) this can lead
+to unexpected data being loaded for a spec. To prevent this from happening, we can use the
 `:stub_elasticsearch` flag. The `:stub_elasticsearch` flag will stub all index
 and search requests made to Elasticsearch and return an empty response. This
-will ensure that no unwanted data shows up on your page.
+will ensure that no unwanted data shows up on your spec's page.
--- a/docs/tests/test-flags.md
+++ b/docs/tests/test-flags.md
@@ -11,8 +11,9 @@ test example.
 Some flags that we use are:
 
 - `js: true`
-- `elasticsearch: reset`
+- `elasticsearch_reset: true`
 - `elasticsearch: <search class name>`
+- `stub_elasticsearch: true`
 - `throttle: true`
 - `type: <test type>`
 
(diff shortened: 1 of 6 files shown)


docs/tests/test-flags.md
```diff
@@ -50,7 +51,13 @@ There are two ways to do this:
...
 ```
 
-**NOTE** Any specs that use the `js: true` flag might be hitting Elasticsearch.
-You may want to consider clearing out Elasticsearch data for those specs even if
-you don't intend on working with Elasticsearch directly to ensure that you have
-a clean page load with no unexpected data.
+### `js: true` Flag
+
+`js: true` indicates to our specs that we want the javascript on the page to be
+execute when the page is rendered. One side effect of running our javascript is
```
vaidehijoshi, line 57:
```suggestion
executed when the page is rendered. One side effect of running our javascript in our specs is
```

docs/tests/test-flags.md
```diff
@@ -50,7 +51,13 @@ There are two ways to do this:
...
+that a lot of pages will hit Elasticsearch. Since we don't clean out
+Elasticsearch between every single spec(because it is very costly) this can lead
```
vaidehijoshi, line 59:
```suggestion
Elasticsearch between every single spec (because it is very costly) this can lead
```

docs/tests/test-flags.md
```diff
@@ -50,7 +51,13 @@ There are two ways to do this:
...
+to unexpected data being loaded for a spec. To prevent this use the
```
vaidehijoshi, line 60:
```suggestion
to unexpected data being loaded for a spec. To prevent this from happening, we can use the
```

docs/tests/test-flags.md
```diff
@@ -50,7 +51,13 @@ There are two ways to do this:
...
+`:stub_elasticsearch` flag. The `:stub_elasticsearch` flag will stub all index
+and search requests made to Elasticsearch and return an empty response. This
+will ensure that no unwanted data shows up on your page.
```
vaidehijoshi, line 63:
```suggestion
will ensure that no unwanted data shows up on your spec's page.
```

i think it's worth clarifying this because this is referencing the capybara `page` object when running `js: true` tests, right?

spec/system/user/view_user_index_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-RSpec.describe "User index", type: :system do
+RSpec.describe "User index", type: :system, stub_elasticsearch: true do
```
vaidehijoshi, line 3: I am REALLY happy with this change! I can promise you that future me in 6 months will know exactly what this flag is and when to use it :)

docs/tests/test-flags.md
mstruve: yep!

rhymes:
Love this!

snackattas:
This is a cool idea!

Issue #8434, opened 2020-06-12
Show banner for not configured sites


citizen428:
Remove defaults related to DEV from the site config and add a banner that the site has not been configured yet.

PR #8435, opened 2020-06-12
Display banner if site hasn't been configured (PRE DEPLOYMENT TASKS)

citizen428:
Note: this now needs https://github.com/thepracticaldev/dev.to/pull/9068 to be merged first.

- [X] Feature

## Description

A lot of the defaults in `SiteConfig` are DEV specific. In a Slack discussion, we (@thepracticaldev/oss) agreed on eventually removing all these defaults and displaying a banner to the admin, directing them to the configuration page. I started this PR as a PoC and since we agreed on the implementation @Ridhwana kindly joined and helped me push this over the finish line.

Note that some default values were removed but not added as mandatory config values. Where appropriate we checked usages and added the necessary conditional logic.

**Some more notes:**

1. Changed `primary_sticker_image_url` to `secondary_logo_url` 
2. If there is no `secondary_logo`, `onboarding_logo_image` or `onboarding_taskcard_image` we show the `logo_png` 
3. Rename `logo_png` to `primary_logo_url`
4. Gated some parameters like `payment_pointer`, `tagline`
5. Removed `community_member_description` as @Ridhwana thoughts it was repetitive. We used  `community_member_description` only on the sign-in invitation banner and it was configured to use "DEV is a community of x amazing humans who code", where `community_member_description` is `amazing humans who code`. Instead, we now just use the `community_member_labe`l and changed it to "DEV(local) is a community of 11 amazing developers". Reusing config for these kinds of things seems to make more sense than adding additional options.
6. Show *'s next to the mandatory fields on `/internal/config`

In a subsequent PR, we will rename some variables to be consistent, like adding `_url` suffix for all URL related options. We also want to display what values are missing so the user doesn't have to look for them through the form. In addition, we'll do validation before hitting save.

#8434

### Banner on normal pages

Screen Shot 2020-06-12 at 11 27 59

### Banner on `/internal` except `/internal/config`

Screen Shot 2020-06-12 at 11 28 25

### Banner on `/admin`

Screen Shot 2020-06-12 at 11 28 53

To show/hide the banner remove or add values listed in `VerifySetupCompleted::MANDATORY_CONFIGS` from your `SiteConfig`.

We did a smoke test for the site but if you're able to do one too that would be awesome!

## Added tests?

- [X] added and updated as needed

## Added to documentation?

- [X] no documentation needed

## Pre-deployment tasks

Ensure the following `SiteConfig` values are actually set in the DB and don't rely on defaults:

* [x] `community_description`: `"A constructive and inclusive social network. Open source and radically transparent."`
* [x] `community_member_label`: `"developer"`
* [x] `community_action`: `"coding"`
* [x] `tagline`: `"We're a place where coders share, stay up-to-date and grow their careers."`

* [x] `email_addresses`:  
    ```
    {
      default: "yo@dev.to",
      business: "partners@dev.to",
      privacy: "privacy@dev.to",
      members: "members@dev.to"
    }
    ```

* [x] `jobs_url`: `"https://jobs.dev.to/"`

* [x] `main_social_image`: `"https://thepracticaldev.s3.amazonaws.com/i/6hqmcjaxbgbon8ydw93z.png"`
* [x] `favicon_url`: `https://dev-to.s3.us-east-2.amazonaws.com/favicon.ico`
* [x] `logo_png`: `"https://practicaldev-herokuapp-com.freetls.fastly.net/assets/devlogo-pwa-512.png"`

* [x] `mascot_image_url`: `https://dev-to-uploads.s3.amazonaws.com/i/y5767q6brm62skiyywvc.png"`
* [x] `mascot_image_description`: `"Sloan, the sloth mascot"`
* [x] `mascot_footer_image_url`: `https://dev-to-uploads.s3.amazonaws.com/i/wmv3mtusjwb3r13d5h2f.png`

* [x] `meta_keywords`
    ```
    {
      default: "software development, engineering, rails, javascript, ruby",
      article: "software, coding, development, engineering, inclusive, community",
      tag: "software development, engineering",
    }
   ```

* [x] `payment_pointer`: `"$ilp.uphold.com/24HhrUGG7ekn"`
* [x] `shop_url`: `"https://shop.dev.to`

* [x] `onboarding_logo_image`: `https://dev.to/assets/purple-dev-logo.png"`
* [x] `onboarding_background_image`: `"https://dev.to/assets/onboarding-background-white.png"`
* [x] `onboarding_taskcard_image`: `"https://practicaldev-herokuapp-com.freetls.fastly.net/assets/staggered-dev.svg"`
* [x] `sidebar_tags`: `%w[help challenge discuss explainlikeimfive meta watercooler]`
* [x] `suggested_users`: `%w[ben jess peter maestromac andy liana]`

* [x] THIS IS A NEW VALUE - `secondary_logo_url`:`"https://practicaldev-herokuapp-com.freetls.fastly.net/assets/rainbowdev.svg"`

* [x] `secondary_logo_url`: `https://practicaldev-herokuapp-com.freetls.fastly.net/assets/rainbowdev.svg `
* [x] `twitter_hashtag`: `"#DEVCommunity"`


Ridhwana:
I like this approach too @citizen428.

app/controllers/concerns/verify_setup_completed.rb
```diff
@@ -0,0 +1,36 @@
+module VerifySetupCompleted
+  extend ActiveSupport::Concern
+
+  MANDATORY_CONFIGS = %i[
```
citizen428, line 4: Removed defaults that are not added here:

* `shop_config`: checked, it seems this is generally gated via `if SiteConfig.shop_url.present?`

citizen428:
@thepracticaldev/oss In case any of you want to help with this (just push into this PR):

1. Remove defaults
2. Decide whether or not they should be added to `VerifySetupCompleted::MANDATORY_CONFIGS`. For example, I didn’t do it for the shop URL, as that can conceivably be missing. In that case, make sure we are not using the value in a way that could break.
3. List them at the end of the PR description as pre-deployment tasks.

Ridhwana:
@citizen428 just a note that I'll do some work on this today

citizen428:
Thanks @Ridhwana! I'll probably do some too, I'll ping you on Slack so we avoid conflicts.

app/controllers/concerns/verify_setup_completed.rb
```diff
@@ -0,0 +1,36 @@
...
+    community_description
+    community_member_description
+    community_action
+    tag_url
```
Ridhwana, line 8: @citizen428 I'm getting an error on tag_url, was this meant to be tagline. I don't see a site config variable called tag_url?

app/controllers/concerns/verify_setup_completed.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    main_social_image
+    favicon_url
+    logo_png
```
Ridhwana, line 12: We can technically gate `favicon_url`, `logo_png` or even just use a [placeholder](https://placeholder.com/) image but I feel like a logo and favicon are mandatory items that should be dealt with on setup.

app/controllers/concerns/verify_setup_completed.rb
```diff
@@ -0,0 +1,41 @@
...
+    primary_sticker_image_url
```
Ridhwana, line 13: The `primary_sticker_image_url` config is strange to me, because it seems that we have a different logo for our internal admin interface (i.e. the primary_sticker_image_url which is the rainbowdev logo) and a different one for our external facing UI (i.e. the logo pg/svg). We also use  primary_sticker_image_url on the sign in invitation. 

I don't feel that every forem will have the luxury of having two logos and so my suggestion is as follows: 
- Let's make primary_sticker_image_url optional and without a default (so I'll take it out of this mandatory configs), and then we can use `logo_png` as the fallback when we do not have a primary_sticker_image_url. 
- In addition, perhaps we rename primary_sticker_image_url to `secondary_logo` since that is actually what our rainbowdev image is.

Any thoughts on this proposal?  (cc @ludwiczakpawel - would be good to have your thoughts here as well since it has a design/ui element)

app/controllers/concerns/verify_setup_completed.rb
citizen428: 🤦 Yes, that was most certainly mean to be `tagline`, sorry for that.

app/controllers/concerns/verify_setup_completed.rb
citizen428: I agree on `logo_png`, not so sure about `favicon_url`, as 404s for those are common.

app/controllers/concerns/verify_setup_completed.rb
citizen428: This sounds like a very sensible approach to me, the name `primary_sticker_image_url` is pretty confusing.

ludwiczakpawel: @Ridhwana I definitely agree.

I believe most communities will have only one logo by default AND MAYBE another one for when they think it's for dark theme (which isn't valid assumption of course)... I don't think too many brands have "alternative" logo. So for whenever we reference any kind of logo in the UI - it should always be the same thing... TBH I can't really think of good usage for `secondary_logo` in the UI - there's no good reason to replace "normal" logo in certain places (like sign in invitation etc).

I'd personally nuke "secondary" logo altogether.

---

Another topic is that we need to provide a way for admins to setup other assets, like app icon, PWA icon (in multiple sizes too), favicon etc. etc.. Right now I think all of them are pulled from `/assets` folder in codebase but maybe this could be configurable from admin panel as well?

app/controllers/concerns/verify_setup_completed.rb
ludwiczakpawel: If logos, favicons and app icons are configurable through admin panel, we should ask admin to upload at least two assets:

1. Logo used everywhere in UI
2. Icon being a square image that would be also used for any other icons assets (favicon, app icon, etc.) unless admin decides to overwrite that with custom ones if that makes sense.

ludwiczakpawel:
Should we make other things configurable too? For example Listings, Podcasts, Videos..?

citizen428:
@ludwiczakpawel Many more things will get generalized for as we continue working on Forem, this specif PR's purpose is to remove DEV specific defaults from configuration to make DEV just another Forem instance.

app/assets/stylesheets/preact/onboarding-modal.scss
```diff
@@ -81,7 +81,6 @@
     justify-content: center;
     align-items: center;
     overflow: hidden;
-    background-image: url('/assets/onboarding-background-white.png');
```
Ridhwana, line 84: This image is specific to DEV since it contains the DEV badges and should not be hardcoded in our codebase for Forem. However, I just want to make a note that when I remove this there is a flash before the asset from the site config shows up on the page. I'm not sure whats the alternate solution to this?

app/controllers/internal/configs_controller.rb
```diff
@@ -120,7 +120,6 @@ def mascot_params
     %i[
       mascot_image_description
       mascot_image_url
-      mascot_footer_image_url
```
Ridhwana, line 123: I couldn't find any reference to this in the codebase and so I've removed it

app/controllers/internal/configs_controller.rb
```diff
@@ -131,7 +130,7 @@ def image_params
       logo_png
       logo_svg
       main_social_image
-      primary_sticker_image_url
+      secondary_logo_url
```
Ridhwana, line 133: If you look through the conversations on this PR (sorry there's a lot 😅) you'll see that we had a discussion about `primary_sticker_image_url`. We concluded that it was not an intuitive name, and that `secondary_logo` made more sense.

app/helpers/application_helper.rb
```diff
@@ -159,6 +159,10 @@ def logo_svg
     end
   end
 
+  def show_logo logo
```
Ridhwana, line 162: We should start using `show_logo` helper to display what logo to use. We're going to find that many Forems will most likely have only one logo, so in the case where additional logos/images are not available (like a secondary_logo, onboarding_logo_image and onboarding_taskcard_image), we will use the primary logo image in its place 😊 Hence, this allows a user to get started with Forem more easily and allows us to reduce the number of madatory configs

citizen428: Total nitpick, but we don't generally use the no parenthesis form for method defintions.

citizen428: Ah, sorry for that, sometimes it gets confusing between the conversation s and the actual diffs, especially when you don't notice the "Outdated" tag.

app/assets/stylesheets/preact/onboarding-modal.scss
citizen428: Maybe @ludwiczakpawel wants to weigh in on this?

app/views/internal/configs/show.html.erb
```diff
@@ -144,23 +144,14 @@
                      } %>
           <div id="contentBodyContainer" class="card-body collapse hide" aria-labelledby="contentBodyContainer">
             <div class="form-group">
-              <%= f.label :community_description, "Community description" %>
+              <%= f.label :community_description, "Community description*" %>
```
citizen428, line 147: I'm afraid that if we add the stars for mandatory fields manually they might get out of sync. Maybe we can make a custom label helper that checks against `VerifySetupCompleted::MANDATORY_CONFIGS`, I'll take a stab at that after finishing the review.

citizen428:
@Ridhwana I now added a helper `internal_config_labels` that is used throughout `app/views/internal/configs/show.html.erb`. It uses `VerifySetupCompleted::MANDATORY_CONFIGS` to always keep the list of required labels in sync with the concern, so the constant is now no longer private. I kept the `*` as a suffix for now, but thought maybe `[required]` would be move explicit. Anyway, it's just a simple change in a single location now.

citizen428:
The CodeClimate warning is for an unescaped model attribute: `SiteConfig.logo_svg.html_safe` Since `logo_svg` gets set in the internal config page by admins, I think we can ignore this.

citizen428:
Even though this is still in becasue of spec fixes, @thepracticaldev/sre might want to get a headstart on making sure all the `SiteConfig` variables in the PR description are correctly set, as we will no longer be able to rely on defaults.

app/assets/javascripts/initializers/initializePaymentPointers.js
```diff
@@ -2,10 +2,10 @@ function initializePaymentPointers() {
   var userPointer = document.getElementById('author-payment-pointer');
   var basePointer = document.getElementById('base-payment-pointer');
   var meta = document.querySelector("meta[name='monetization']");
-  
+
   if (userPointer) {
     meta.content = userPointer.dataset.paymentPointer;
-  } else {
+  } else if (basePointer) {
```
Ridhwana, line 8: basePointer will not exist if the site config payment pointer is not configured, hence we put it in an else if

spec/requests/stories_index_spec.rb
```diff
@@ -16,11 +16,11 @@
       article = create(:article, featured: true)
 
       get "/"
-
       expect(response.body).to include(CGI.escapeHTML(article.title))
     end
 
     it "renders proper description" do
+      allow(SiteConfig).to receive(:community_description).and_return("Some description")
```
citizen428, line 23: I wonder if we should move some of the more common ones into a global spec config, because I'm afraid that otherwise too many people will have errors that are hard to figure out when adding new tests. WDYT @Ridhwana?

citizen428:
@benhalpern Pinging you for a final ok on the approach too.

spec/requests/stories_index_spec.rb
citizen428: `community_description` is stubbed in more than one place and maybe it would make sense to stub `logo_png` too. But I guess we can do this at a later point too.

app/helpers/application_helper.rb
```diff
@@ -159,6 +159,10 @@ def logo_svg
     end
   end
 
+  def show_logo(logo)
```
citizen428, line 162: I was wondering if we want to rename this to `logo_url`, since that's what the params represent and what this method returns. Maybe we could even make it a bit more explicit that there's a sensible default, i.e. `safe_logo_url` or something like that.

Ridhwana: I like this suggestion @citizen428. I've updated 👌🏽

app/models/site_config.rb
```diff
@@ -26,26 +26,25 @@ class SiteConfig < RailsSettings::Base
   field :campaign_articles_require_approval, type: :boolean, default: 0
 
   # Community Content
-  field :community_description, type: :string, default: "A constructive and inclusive social network. Open source and radically transparent."
-  field :community_member_description, type: :string, default: "amazing humans who code."
```
Ridhwana, line 30: Removed community_member_description as I think its repetitive. We used  community_member_description only on the sign in invitation banner and it was configured to read `DEV is a community of x amazing humans who code`, community_member_description had a value of `amazing humans who code`. 

Instead I've amended the sentence to use the community_member_label and changed it to `DEV(local) is a community of 11 amazing developers`. 

I feel like reusing config for these kind of things makes more sense than to add some more.

spec/requests/stories_index_spec.rb
Ridhwana: 💯 I moved community_description into the rails helper, but I kept the suggested tags and the on boarding stuff relevant to their files. I'm going to keep an eye on logo_png before I add it to the global config

app/views/articles/_search.html.erb
```diff
@@ -167,7 +167,7 @@
     })
       .then(response => response.json())
     .then((content) => {
-        if (searchHash["search_fields"]) {
+        if (searchHash["search_fields"] && content.jobs_url) {
```
mstruve, line 170: 🎉

app/views/layouts/application.html.erb
```diff
@@ -1,5 +1,7 @@
 <%= render "shell/top" %>
-<div id="base-payment-pointer" data-payment-pointer="<%= SiteConfig.payment_pointer %>"></div>
+<% if SiteConfig.payment_pointer.present? %>
```
mstruve, line 2: We have a lot of conditional checks for these SiteConfigs scattered all over the code. I'm wondering if it might be cleaner to hook up a `method_missing` method to handle and return something when a SiteConfig is not present.

app/controllers/concerns/verify_setup_completed.rb
```diff
@@ -0,0 +1,45 @@
...
+  included do
+    # rubocop:disable Rails/LexicallyScopedActionFilter
+    before_action :verify_setup_completed, only: %i[index new edit show]
+    # rubocop:enable Rails/LexicallyScopedActionFilter
+  end
+
+  private
+
+  def verify_setup_completed
+    return if setup_completed? || config_path?
```
mstruve, line 30: Can we flip these? `setup_completed?` will make a call to Redis while `config_path?` doesn't make any external calls and if`config_path?` returns true we can skip the Redis call.

I really wish there was a way we could verify setup completed without having to hit Redis on every call 🤔 I'm not sure of another way though off the top of my head. Maybe something to keep in mind.

mstruve:
We can't set `secondary_logo_url` yet since it is not included in the SiteConfig list, is that something we want to deploy ahead of this so we can get that set before this goes out since there will not be a default?

app/views/layouts/application.html.erb
citizen428: Many of these conditionals are in views and I'm not sure `method_missing` can help us with that, as returning view snippets from there does not seem particularly clean either.

I thought along the lines of doing a refactor later where we extract the conditional code into partials, so the logic on whether or not to render something is self-contained and the including view is conditional free. For this specific example:

```ruby
<%= render partial: "payment_pointer", locals: { payment_pointer: SiteConfig.payment_pointer } %>
```

The partial then can be something like

```ruby
<% if payment_pointer.present? %>
  ...
<% end %>
```

That's a little bit like frontend components, i.e. the enclosing "component" just renders and something and passes in data (here locals instead of props) and the "component" itself decides if/what to render.

How does that sound?

app/controllers/concerns/verify_setup_completed.rb
citizen428: Good point, I recently suggested a similar change elsewhere, so 👍 on catching that! 

I'm a bit confused about the mention of Redis though. `SiteConfig` is stored in the DB. It also caches its values for a while, so they aren't retrieved every time:

```ruby
[1] dev.to(main)> SiteConfig.public_send(:logo_png).present?
  SiteConfig Load (0.6ms)  SELECT var, value FROM "site_configs"
true
[2] dev.to(main)> SiteConfig.public_send(:secondary_logo_url).present?
false
```

I just woke up and haven't had enough coffee yet, so am probably missing something, can you elaborate which Redis calls you mean?

citizen428: @mstruve I switched to `return if config_path? || setup_completed?` now. It's probably not going to make a lot of difference since only `/internal/config` will trigger this early return, but it's a good habit to sort them like this anyway.

citizen428:
@mstruve Thanks for the feedback! I opened https://github.com/thepracticaldev/dev.to/pull/9068 which only adds `secondary_logo_url` to `SiteConfig` so we can deploy this first and set the default.

app/controllers/concerns/verify_setup_completed.rb
rhymes: @citizen428 SiteConfig caches things in memory and in Redis, it's a two layer cache system:

https://github.com/huacnlee/rails-settings-cached/blob/7ae49b0f46a14d7b0a065f0719650c168d4b1505/lib/rails-settings/base.rb#L130-L138

```ruby
          RequestStore.store[:rails_settings_all_settings] ||= begin
            Rails.cache.fetch(self.cache_key, expires_in: 1.week) do
              vars = unscoped.select("var, value")
              result = {}
              vars.each { |record| result[record.var] = record.value }
              result.with_indifferent_access
            end
          end
        end
```

Ridhwana:
I'll merge and deploy this today. @Zhao-Andy, I seen that you're on call and so I'll ping you when the US day starts to work with you to monitor the PR to ensure that nothing goes wrong. 

@mstruve has confirmed that all the values are set and I double checked them on Blazer too. 🤞🏽

Veggier:
In the pre-deployment tasks, I think we missing the `suggested_tags` value. In the `MANDATORY_CONFIGS`, `suggested_tags` is default request value.

Ridhwana:
@Veggier yeah I noticed that suggested_tags isn't in the pre deployment tasks, but then I checked in the database and it is currently a set value, so nothing should have broken as a result.

Veggier:
@Ridhwana  I am not sure, because I don't have that value. Therefore, I added it by myself.

Ridhwana:
@Veggier I'm not sure what you mean? Is this on your local setup, there are no default values so you would not see any values on your local env (I was talking about production when i said I checked the database). Did you get some sort of an error? Could you perhaps possibly open up an issue and help me with replication steps and expected behaviour so I can debug the issue?

Veggier:
@Ridhwana yea, I was talking about my local setup. There are no default values which cause the empty values of my local env. Therefore, I had to set up all the requests by myself.

Veggier:
@Ridhwana when I have the empty value of the  `suggested_tags`, It shows that Setup not completed
image
image
and when I set the value, the error will be gone
image
image

Ridhwana:
@Veggier I'll look into this UX today and get back to you on it.

app/controllers/concerns/verify_setup_completed.rb
citizen428: Thank you @rhymes 😃

PR #8436, opened 2020-06-12
Small ActiveRecord optimizations: begone race conditions and hello bulk insert

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR contains three things:

- usage of [create_or_find_by](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-create_or_find_by) in `PollSkip`: which is a new Rails 6 methods with works only in certain cases, the goal here is to avoid the possible race condition of "find or create" where something can be created by a different thread between the find and the create (given the model this is used and the fact polls are not used much, it's not strictly necessary but it's a good addition to our tools ;)) *[this only works with a corresponding unique index]*

- bulk inserts for both memberships in a chat channel (going from N queries, one per each user, to 1)
- bulk inserts for data update scripts (I also moved stuff that was only used in tests as `private`, no point keeping it public if we never use it)

Yay Rails 6 :D

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/controllers/poll_skips_controller.rb
+++ b/app/controllers/poll_skips_controller.rb
@@ -2,12 +2,15 @@ class PollSkipsController < ApplicationController
   before_action :authenticate_user!, only: %i[create]
 
   def create
-    @poll_skip = PollSkip.find_or_create_by(poll_id: poll_skips_params[:poll_id], user_id: current_user.id)
-    @poll = Poll.find(poll_skips_params[:poll_id])
-    render json: { voting_data: @poll.voting_data,
-                   poll_id: poll_skips_params[:poll_id].to_i,
-                   user_vote_poll_option_id: nil,
-                   voted: false }
+    poll = Poll.find(poll_skips_params[:poll_id])
+    poll.poll_skips.create_or_find_by(user_id: current_user.id)
+
+    render json: {
+      voting_data: poll.voting_data,
+      poll_id: poll.id,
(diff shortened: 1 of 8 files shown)


app/controllers/poll_skips_controller.rb
```diff
@@ -2,12 +2,15 @@ class PollSkipsController < ApplicationController
...
 
   def create
-    @poll_skip = PollSkip.find_or_create_by(poll_id: poll_skips_params[:poll_id], user_id: current_user.id)
-    @poll = Poll.find(poll_skips_params[:poll_id])
-    render json: { voting_data: @poll.voting_data,
-                   poll_id: poll_skips_params[:poll_id].to_i,
-                   user_vote_poll_option_id: nil,
-                   voted: false }
+    poll = Poll.find(poll_skips_params[:poll_id])
+    poll.poll_skips.create_or_find_by(user_id: current_user.id)
```
mstruve, line 6: > The primary key may auto-increment on each create, even if it fails.

Thats an interesting tidbit to keep in mind if we decide to use this more.

rhymes: @mstruve ahaha we're back to this issue https://github.com/thepracticaldev/tech-private/issues/385 and this PR https://github.com/thepracticaldev/dev.to/pull/1113

PR #8451, opened 2020-06-12
Add active_status_updated_at to the Broadcasts Table

### LINKED ISSUE #8276: Track when a broadcast was activated
Since announcement broadcasts will be use to send site-wide messages with their `active` attribute, we should make it easy to track when an admin last set a broadcast to "active". We should add a `last_active_at` to Broadcasts (specifically to be used for `type_of: "Announcement"`, but this will be useful for all broadcasts) and make sure that is rendered on the `/internal/broadcasts` page for each broadcast.

juliannatetreault:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR adds an `active_status_updated_at` column to the Broadcasts table. In addition to this, it adds a place to display the time at which a particular Broadcast was last set to `active` in `/internal/broadcasts`, an `active_status_updated_a` attribute to the Broadcasts factory, and a model spec that tests these changes.

Closes #8276 

### QA Instructions:
1. Navigate to `/internal/broadcasts` and ensure that there are Broadcasts, if not, create a new one (Broadcast `type` does not matter in this case).
Screen Shot 2020-06-12 at 1 53 52 PM
2. Set the Broadcast to `Active: true` if it is not already. Observe that the Broadcast's "Last Active On:" timestamp updates.
Screen Shot 2020-06-12 at 1 54 18 PM
3. Set the Broadcast to `Active: false` so that it is no longer active and observe that the "Last Active On:" timestamp displays the time for when you first set the Broadcast to `Active: true`.
Screen Shot 2020-06-12 at 1 54 33 PM
4. Set the Broadcast to `Active: true` so that it is active again and observe that the "Last Active On:" timestamp displays the timestamp that correlates to the time that you set it to `Active: true` again (the current time).
Screen Shot 2020-06-12 at 1 55 08 PM
5. Try to break things! 

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Clocks


app/views/internal/broadcasts/index.html.erb
```diff
@@ -31,6 +31,7 @@
           <p>
             <%= broadcast.processed_html %>
           </p>
+          <h3>Last Active: <%= broadcast.last_active_at %></h3>
```
juliannatetreault, line 34: Should I format this timestamp at all? 🤔

lib/data_update_scripts/20200612173137_backfill_last_active_at_for_broadcasts.rb
```diff
@@ -0,0 +1,15 @@
...
+  class BackfillLastActiveAtForBroadcasts
+    def run
+      # Broadcasts didn't have a last_active_at column until
+      # 20200609195523_add_last_active_at_to_broadcasts was created, thus we can
+      # safely assume that only those with `last_active_at IS NULL` need their
+      # timestamps overridden.
+      # This updates the Broadcast's last_active_at to be the date and time
+      # in which the Broadcast was last updated.
+      Broadcast.where(last_active_at: nil).order(title: :asc).each do |cast|
+        cast.update(last_active_at: cast.updated_at)
```
juliannatetreault, line 11: Please see my "post-deployment tasks" comment in the PR description regarding this script!

app/views/internal/broadcasts/index.html.erb
vaidehijoshi: Yep, we definitely should; the current format isn't really human-readable. Check out Ruby's `strftime` [method](https://apidock.com/ruby/DateTime/strftimehttps://github.com/juliannatetreault/dev.to/tree/juliannatetreault/last-active-broadcast).

lib/data_update_scripts/20200612173137_backfill_last_active_at_for_broadcasts.rb
vaidehijoshi: Hm, I have some doubts around this. Why do we need to backfill the data for pre-existing broadcasts?

The `last_active_at` column isn't _required_ and we allow that column to be `null` in [the migration](https://github.com/thepracticaldev/dev.to/pull/8451/files#diff-59c9ac156b94b4411f337406d7e5da9aR3). If we run a script to backfill previous broadcast data to whenever they were _updated_, we're setting an attribute manually which in the future will lead someone to believe that the broadcast was _at some point_, set to `active`. This is will not necessarily be true. For example, if there is a broadcast on production that has never been set to active, but has been updated/saved at some point, when we ran this script, we'd now actually be mutating the data attached to it and effectively saying that "this broadcast was once active", even though it never was. The truth is that the broadcast was _updated_ at some point, but that may not have anything to do with whether or not it was active :)

TL;DR: we really shouldn't ever set `last_active_at` manually since this will cause potential confusion (and maybe bugs) down the road; it should only be set when the broadcast was _actually_ activated.

app/models/broadcast.rb
```diff
@@ -10,6 +10,8 @@ class Broadcast < ApplicationRecord
   validates :banner_style, inclusion: { in: VALID_BANNER_STYLES }, allow_blank: true
   validate  :single_active_announcement_broadcast
 
+  before_save :update_last_active_at, if: :will_save_change_to_active?
```
vaidehijoshi, line 13: Doesn't look like `will_save_change_to_active` is being used at all here

app/views/internal/broadcasts/index.html.erb
juliannatetreault: Agreed! I have since changed this and followed the convention of some other formatted times (podcasts and events), but I could also remove the year from the timestamp if we think that it isn't necessary.

lib/data_update_scripts/20200612173137_backfill_last_active_at_for_broadcasts.rb
juliannatetreault: OH! I thought (after looking at Broadcasts) that this would be necessary since all of the Broadcasts in production will not have a `last_active_at` timestamp by default, but I see what you mean - I overoptimized here and since removed this file.

app/models/broadcast.rb
juliannatetreault: Huh, it looks like you're right! I just removed the method and it didn't result in any errors - strange. 🤔  I have updated the model to not include this dead code!

app/models/broadcast.rb
```diff
@@ -32,4 +34,12 @@ def single_active_announcement_broadcast
 
     errors.add(:base, "You can only have one active announcement broadcast")
   end
+
+  def update_last_active_at
```
mstruve, line 38: I think this might not be quite what you want. When I think `last_active_at` I think the last time the broadcast was active. Here is the scenario I am thinking that might be confusing. 

1. Set broadcast to active on March 1, last_active_at = March 1
2. Set broadcast to inactive on April 1, last_active_at does not change
3. View broadcast on May 1 and see `last_active_at` equals March 1 but you have no idea how long it was active for after it was marked active. In this case, `last_active_at` is actually `last_activated_at` If indeed this is what we want then I would suggest renaming the column to be more clear. 

If we actually want `last_active_at` then I would think you would want the value to be April 1 since that was the last time it was active. In order to set it this way you need to check if `active` has been changed from true to false and then set it when that condition occurs.

juliannatetreault: You raise a good point @mstruve! The goal here is to display the time at which the Broadcast was last set to `active`: https://github.com/thepracticaldev/dev.to/issues/8276. That being said, I might want to change the column name to `last_actived_at` per your suggestion to clear the code up, unless I'm missing something here. 😊

vaidehijoshi: I think `last_activated_at` makes sense in the context of the issue, but I also see @mstruve's point about actually wanting to know _when_ it was last active. We don't use this feature yet, so it's hard to know whether we'll ever want to know when it was last _active_ versus when it was last _activated_.

It might be nice to actually track `last_active_at` and update it whenever someone toggles the `active` property. We don't technically need to know this, but since @mstruve pointed it out, it might be worth the extra few minutes of refactoring to make sure we're updating that column whenever we update `active` on the broadcast. 

Curious if others have strong opinions here (cc @mstruve + @atsmith813).

app/models/broadcast.rb
```diff
@@ -32,4 +34,12 @@ def single_active_announcement_broadcast
...
+    # Updates the last_active_at timestamp to show
+    # when the Broadcast was last set to "active"
```
vaidehijoshi, line 40: IMHO, I don't think this comment is adding anything; the code is pretty self-documenting here :)

app/models/broadcast.rb
juliannatetreault: Okay, I have modified this to actually track when the Broadcast was `last_active_at` and will push up the changes momentarily. Prior to doing that, I just wanted to make sure that the timestamps are granular enough as is and that I shouldn't change the timestamp format along with these changes -thoughts @mstruve @vaidehijoshi cc @atsmith813?

vaidehijoshi: @juliannatetreault are you asking if the column should no longer be a `datetime` format? if that's your question, then the answer is that you can keep it in the same format, since that is the convention (if you do a quick search through `schema.rb` you'll see that `created_at` and `updated_at` use that same format).

juliannatetreault: No, what I'm asking is if the time displayed in `/internal/broadcasts` should be more granular (display it with the hour it was changed in addition to the date, for example) or if it is fine as is or if I should change the formatting of it in the view--sorry for the confusion! 😬  @vaidehijoshi

vaidehijoshi: @juliannatetreault sure, displaying the hour/minute seems specific enough 👍

juliannatetreault:
**Update**: I have adjusted the logic in `broadcast.rb` so that whenever a Broadcast is toggled to active `true`, the `last_active_at` timestamp is updated, and likewise, whenever the Broadcast is toggled to active `false`, the `last_active_at` timestamp is updated again. In addition to this, I have also adjusted the timestamp that displays when a Broadcast was "last active" in `/internal/broadcasts` to include hours and minutes in UTC so that if on the off chance, a Broadcast is updated from active to inactive (or vice versa) in the same day, it is clearer to the user. @mstruve @vaidehijoshi @atsmith813 😄

app/models/broadcast.rb
```diff
@@ -32,4 +34,8 @@ def single_active_announcement_broadcast
 
     errors.add(:base, "You can only have one active announcement broadcast")
   end
+
+  def update_last_active_at
+    self.last_active_at = Time.zone.now
```
vaidehijoshi, line 39: Should use [`Time.current`](https://apidock.com/rails/v6.0.0/Time/current/class) here

app/models/broadcast.rb
vaidehijoshi, line 13: I was about to ask why this [came back](https://github.com/thepracticaldev/dev.to/pull/8451/files#r439868062), but now I see that `will_save_change_to` is [a Rails magic thing](https://apidock.com/rails/v6.0.0/ActiveRecord/AttributeMethods/Dirty/will_save_change_to_attribute%3F)!

app/views/internal/broadcasts/index.html.erb
```diff
@@ -31,6 +31,7 @@
           <p>
             <%= broadcast.processed_html %>
           </p>
+          <h3>Last Active: <%= broadcast.last_active_at.strftime("%b %d, %Y %H:%M UTC") %></h3>
```
vaidehijoshi, line 34: What happens if a broadcast doesn't have a `last_active_at` set? 

Screen Shot 2020-06-15 at 10 51 47 AM

spec/models/broadcast_spec.rb
```diff
@@ -18,4 +18,14 @@
...
+
+  it "updates the Broadcasts last_active_at timestamp" do
+    Timecop.freeze(Time.current) do
+      current_time = Time.zone.now
+      broadcast = create(:welcome_broadcast, active: false)
+      expect(broadcast.last_active_at).to eq(2.days.ago)
+      broadcast.update(active: true)
+      expect(broadcast.last_active_at).to eq current_time
+    end
+  end
```
vaidehijoshi, line 30: I think we could also use some tests in `spec/requests/internal/broadcasts_spec.rb` that test that this attribute does/does not update correctly from the controller side

juliannatetreault: I had a spec for this in `/internal/broadcasts_spec.rb` initially, but after speaking with @atsmith813 on Friday, he suggested that I move my spec to the Broadcast model since the method that updates the timestamp is on the model itself. That being said, is it necessary and does it make sense to add a spec to `/internal/broadcasts_spec.rb` in addition to the one in the model? cc @atsmith813

app/controllers/internal/broadcasts_controller.rb
```diff
@@ -56,7 +56,7 @@ def destroy
   private
 
   def broadcast_params
-    params.permit(:title, :processed_html, :type_of, :banner_style, :active)
+    params.permit(:title, :processed_html, :type_of, :banner_style, :active, :last_active_at)
```
vaidehijoshi, line 59: Hm, looking at this again, do we actually need `last_active_at` here? We don't allow the user to update this data, we update it internally, so it shouldn't actually be necessary, right?

juliannatetreault: This is a holdover from my previous work. 🤦‍♀️

spec/models/broadcast_spec.rb
vaidehijoshi: The test that you have _does_ test that the model behavior works; but what happens if you remove `:active` from [here](https://github.com/thepracticaldev/dev.to/pull/8451/files#diff-08d4f0af1a44965d340da617d97718dbR59)?

In fact, I'd try it out; if you removed it from the params, your [test](https://github.com/thepracticaldev/dev.to/pull/8451/files#diff-37ba91f757769b082f98c5b68d981c8fR22) on line 22 would still pass, since that test is manually updating the attribute. You're testing that the model behavior works, but not testing that the attribute can be updated from the controller/via a request.

That's why I'd advocate for adding a test to make sure that you can make a request to change this attribute from the controller side. That way, if something about the controller breaks, you'll catch the regression (even if the model side changes aren't affected).

Does that make sense?

app/models/broadcast.rb
juliannatetreault: Correct! Although you were correct in that it was working properly initially.

spec/models/broadcast_spec.rb
juliannatetreault: Yes, that makes sense. I have added a spec for this per your request. 😄

app/models/broadcast.rb
mstruve, line 13: This still feels a little confusing to me, sometimes this time is when the broadcast was last active(when going from active to inactive) and sometimes this time is when the broadcast was last activated(when going from inactive to active).  What if instead, we called this timestamp `active_status_updated_at` ?

spec/requests/internal/broadcasts_spec.rb
```diff
@@ -48,6 +48,17 @@
...
     end
 
+    describe "PUT /internal/broadcasts" do
+      let!(:broadcast) { create(:welcome_broadcast, active: false) }
+
+      it "updates the Broadcasts last_active_at timestamp" do
+        expect do
+          put "/internal/broadcasts/#{broadcast.id}", params: params
+        end.to change { broadcast.reload.active }.from(false).to(true)
+        expect(broadcast.last_active_at).not_to eq(2.days.ago)
```
mstruve, line 58: I think you want Timecop here to ensure timing issues don't occur. Also, instead of this being 2 days ago, you should store the old time and assert it is not equal to it. This way if someone changes that factory default this spec's assertion will be updated with it rather than break. 
```
old_time = broadcast.last_active_at
# update
expect(broadcast.last_active_at).not_to eq(old_time)
```

app/models/broadcast.rb
juliannatetreault: I'm happy to change the name! For clarification purposes on my end though, when you refer to "timestamp" here, is it the timestamp on the Broadcast table (`last_active_at`) or the act of updating the timestamp in `#update_last_active_at` (asking because we're in the model now and I want to ensure I understand!) @mstruve.

spec/requests/internal/broadcasts_spec.rb
juliannatetreault: Great call @mstruve! I'll make this change now.

app/models/broadcast.rb
mstruve: Good question, I'm talking about the column name in the database.

juliannatetreault: Perfect; I'll make that change! 😄

spec/requests/internal/broadcasts_spec.rb
```diff
@@ -48,6 +48,20 @@
       end
     end
 
+    describe "PUT /internal/broadcasts" do
+      let!(:broadcast) { create(:welcome_broadcast, active: false) }
+
+      it "updates the Broadcasts last_active_at timestamp" do
```
vaidehijoshi, line 54:
```suggestion
      it "updates the broadcast's last_active_at timestamp" do
```

spec/requests/internal/broadcasts_spec.rb
```diff
@@ -48,6 +48,20 @@
...
+        old_time = broadcast.last_active_at
+        Timecop.freeze(Time.current) do
+          expect do
+            put "/internal/broadcasts/#{broadcast.id}", params: params
+          end.to change { broadcast.reload.active }.from(false).to(true)
+          expect(broadcast.last_active_at).not_to eq(old_time)
```
vaidehijoshi, line 60: Can we assert the the `last_active_at` is the same as `Time.current`? Rather than asserting that it's just _not_ the same as what it was?

app/views/internal/broadcasts/index.html.erb
```diff
@@ -31,6 +31,7 @@
           <p>
             <%= broadcast.processed_html %>
           </p>
+          <h3>Last Active On: <%= broadcast.active_status_updated_at&.strftime("%b %d, %Y %H:%M UTC") %></h3>
```
mstruve, line 34: Might want to reword this description for clarity

spec/requests/internal/broadcasts_spec.rb
juliannatetreault: Great question @vaidehijoshi!

I attempted to do this multiple times and ways yesterday, and while the specs always passed locally, they always failed on their Travis builds due to the _slightest_ (I'm talking milliseconds) discrepancy in timestamps. At this point in time, I don't think it's worth the time it takes for me to run this through Travis again because it seems like the odds of it passing are relatively low and I don't want to hog the build pipeline again. 😬 😅

app/views/internal/broadcasts/index.html.erb
juliannatetreault: My concern with rewording this is that it is too lengthy. 😬  "Active Status Updated On:" is slightly clearer but is pretty long with the timestamp, especially compared to the rest of the text in the `/internal/broadcasts` view and "Active Status Updated:" doesn't read as well as "Last Active On:" or "Active Status Updated On:". That being said, I'm happy to change it if you have a better suggestion @mstruve! 😄

mstruve: gotcha, this works, if people get confused in the future we can always change it. As this starts to be used by others I think we will be documenting how all this works and we can use docs to clarify if we need to.

juliannatetreault: Sounds like a plan! 👍 👍

PR #8452, opened 2020-06-12
Update ReactionsController not to cache AR Object (2nd attempt)

maestromac:
- [x] Optimization

## Description
Instead of caching whole AR object, this change makes it cache the hash result instead and then use `.select` to see the right comment.

```
Warming up --------------------------------------
        caching AR O   689.000  i/100ms
          caching Id   560.000  i/100ms
        caching hash    68.983k i/100ms
Calculating -------------------------------------
        caching AR O      7.597k (± 4.0%) i/s -     38.584k in   5.086865s
          caching Id      6.515k (± 4.4%) i/s -     33.040k in   5.081493s
        caching hash    980.685k (± 2.6%) i/s -      4.967M in   5.068166s

Comparison:
        caching hash:   980684.5 i/s
        caching AR O:     7597.4 i/s - 129.08x  slower
          caching Id:     6515.0 i/s - 150.53x  slower

```
The difference is huge and it makes me wonder if maybe I made a mistake on my benchmark
## Added tests?
- [ ] yes
- [ ] no, because I need help

## Added to documentation?
- [x] no documentation needed

n/a


app/controllers/reactions_controller.rb
```diff
@@ -102,10 +102,11 @@ def create
     render json: { result: result, category: category }
   end
 
-  def cached_user_public_reactions(user)
-    Rails.cache.fetch("cached-user-reactions-#{user.id}-#{user.public_reactions_count}", expires_in: 24.hours) do
-      user.reactions.public_category
+  def cached_user_public_reactions(user, comment_ids)
+    cache = Rails.cache.fetch("cached-user-reactions-#{user.id}-#{user.public_reactions_count}", expires_in: 24.hours) do
+      JSON.parse(user.reactions.public_category.to_json)
```
mstruve, line 107: I would throw the `.where(reactable_id: comment_ids)` on the end of your AR relation here then `.to_json` it so you can skip the select below.

maestromac: Unfortunately `comment_ids` is the request's param and doesn't fall under `cached-user-reactions-#{user.id}-#{user.public_reactions_count` cache.

mstruve: Got it! Rather than doing `.to_json` here then `JSON.parse` you could do `user.reactions.public_category.map(&:attributes)` to return a hash. And actually, I am not sure how many comment_ids we can expect here but below looping through an array and comparing to another array is going to result in a lot of iterations. I think you could speed that up if you create a hash here instead of an array of hashes. So something like this
```ruby
user.reactions.public_category.each_with_object({}){|r, h| h[r.reactable_id] = r.attributes}

# below
cache.slice(*comment_ids)
```

app/controllers/reactions_controller.rb
```diff
@@ -102,10 +102,11 @@ def create
...
     end
+    cache.select { |a| comment_ids.include?(a["reactable_id"]) }
```
mstruve, line 109: Your above comment makes sense. I think there is actually a bug here now that I think about it. We select the reactions where reactable_id is included in the comment_ids but we dont ensure that the reactions are for comments. I think we want to filter by reactable_type here as well otherwise we might get an article reaction in this group.

app/controllers/reactions_controller.rb
maestromac: Thanks for this optimization tip! Updated.

app/controllers/reactions_controller.rb
maestromac: Great point. I've added an additional `where` clause and updated this method for clarity.

app/controllers/reactions_controller.rb
```diff
@@ -102,11 +102,13 @@ def create
...
-    cache = Rails.cache.fetch("cached-user-reactions-#{user.id}-#{user.public_reactions_count}", expires_in: 24.hours) do
-      JSON.parse(user.reactions.public_category.to_json)
+  def cached_user_public_comment_reactions(user, comment_ids)
+    cache = Rails.cache.fetch("cached-user-#{user.id}-reaction-ids-#{user.public_reactions_count}", expires_in: 24.hours) do
+      user.reactions.public_category.each_with_object({}) do |r, h|
+        h[r.reactable_id] = r.attributes
+      end
     end
-    cache.select { |a| comment_ids.include?(a["reactable_id"]) }
+    cache.slice(*comment_ids).values.select { |r| r["reactable_type"] == "Comment" }
```
mstruve: I would add this to your where statement. 
```
user.reactions.public_category.where(reactable_type: "Comment")
```

maestromac: Yep I did. I think you were GitHub was showing you my older commit 😄

PR #8453, opened 2020-06-12
Spec Cleanup: Stub Elasticsearch for js: true Specs that Dont Need it

mstruve:
- [x] Optimization

## Description
In #8453 I stubbed Elasticsearch for `js: true` specs that had recently failed due to Elasticsearch data displaying on the page when not expected. With the flag now an option I put a puts statement in the Search Controller to see what system specs hit Elasticsearch. Once I got that list I went down and found the ones that don't need it and stubbed it for all of those.

alt_text

### DIFF
--- a/spec/system/user_uses_the_editor_spec.rb
+++ b/spec/system/user_uses_the_editor_spec.rb
@@ -88,7 +88,7 @@ RSpec.describe "Using the editor", type: :system do
     end
   end
 
-  describe "using v2 editor", js: true do
+  describe "using v2 editor", js: true, stub_elasticsearch: true do
     before { user.update(editor_version: "v2") }
 
     it "fill out form with rich content and click publish" do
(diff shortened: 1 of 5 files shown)


spec/system/user_uses_the_editor_spec.rb
```diff
@@ -88,7 +88,7 @@ def fill_markdown_with(content)
     end
   end
 
-  describe "using v2 editor", js: true do
+  describe "using v2 editor", js: true, stub_elasticsearch: true do
```
mstruve, line 91: This actually hit Elasticsearch once for every letter entered in the tag input 😅

mstruve:
and 2 min wasted on the deploy job which is getting 🔪  here #8450 ZOOM ZOOM!

spec/system/user_uses_the_editor_spec.rb
maestromac: 😱

Issue #8414, opened 2020-06-11
Label Experience Level Scale in User Settings


michael-tharrington:
**Is your feature request related to a problem? Please describe.**

Yes, as a user it's unclear how the 1-5 experience scale (Content Customization) is meant to be interpreted in https://dev.to/settings/ux. It's not clearly stated the 1 is for novices and 5 for experts; folks could misinterpret this and think that is 1 is expert and 5 novice.

**Describe the solution you'd like**

I think we should label each number in the scale to correspond to the labeled scale in our [moderation panel](https://dev.to/devteam/changelog-updated-mod-tooling-1ill) (see figure 1).

**Additional context**

Figure 1:

![mod panel experience level scale](https://user-images.githubusercontent.com/16007075/84413563-c9813900-abde-11ea-8bc2-68c9c8095e90.png)

It might look vaguely like this:

![experience scale in settings](https://user-images.githubusercontent.com/16007075/84413936-4dd3bc00-abdf-11ea-9300-24b70943e836.png)

I think it'd be fine to put the text inside each button or underneath, whichever looks better.

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

nickytonline:
I think labelling these is a good idea to avoid confusion of the scale.

Oghenebrume50:
@rhymes I will like to work on this

Ridhwana:
@Oghenebrume50 I've assigned the issue to you 😊

rahuldkjain:
I would love to work on this issue

rhymes:
@rahuldkjain I've assigned you to the issue :) Thank you!

rhymes:
Wasn't this solved by https://github.com/thepracticaldev/dev.to/pull/8454 ?

cc @Oghenebrume50 @jacobherrington

Oghenebrume50:
Yes it was @rhymes the PR was merged it shows up on the website too :+1:

rhymes:
[...]

PR #8454, opened 2020-06-12
Label Experience Level Scale in User Settings

Oghenebrume50:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
According to this issue, #8414 wish I think is a pretty good suggestion, I added labels to the user coding experience in the settings to make it more descriptive for users to select their appropriate experience, without any confusion.

How it looks on a desktop screen
shotdev

How it looks on smaller screens
devshotmobile

## Added tests?
- [ ] yes
- [ ] no, because I need help

## Added to documentation?
- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


Oghenebrume50:
@rhymes

Oghenebrume50:
Thank you @rhymes I look forward to contributing more

app/helpers/application_helper.rb
```diff
@@ -15,6 +15,10 @@ def user_logged_in_status
     user_signed_in? ? "logged-in" : "logged-out"
   end
 
+  def user_experience_labels
+    %w[Novice Beginner Mid-level Advanced Expert]
+  end
```
vaidehijoshi, line 20: Since this is only used in the `users/_ux.html.erb` view, it might make more sense for this to live in its own helper file? Usually when I see a helper in `ApplicationHelper`, it signals to me that it is used throughout the entire app, but since this one is much more specific and is only used in one place, it feels a bit lightweight for this file :) If you did put it in a separate file, you could also extract out the `%w[1 3 5 8 10]` array from [here](https://github.com/thepracticaldev/dev.to/pull/8454/files#diff-46a748ce93f3baf80ce27713e044542cR50), if you wanted! You could also just put this array inline, too.

Not a blocker for merging this PR, though!

Oghenebrume50: So I tried putting the array inline but that did not work, erblint kept throwing errors I would try putting it in a separate helper file though

Oghenebrume50: I have done the update to separate modules

Oghenebrume50: > Since this is only used in the `users/_ux.html.erb` view, it might make more sense for this to live in its own helper file? Usually when I see a helper in `ApplicationHelper`, it signals to me that it is used throughout the entire app, but since this one is much more specific and is only used in one place, it feels a bit lightweight for this file :) If you did put it in a separate file, you could also extract out the `%w[1 3 5 8 10]` array from [here](https://github.com/thepracticaldev/dev.to/pull/8454/files#diff-46a748ce93f3baf80ce27713e044542cR50), if you wanted! You could also just put this array inline, too.
> 
> Not a blocker for merging this PR, though!

@rhymes @vaidehijoshi I have done the update to a separate module, but now I am not sure why Travis is failing.

rhymes: unfortunately it's a flaky spec! Restarted!

app/helpers/ux_helper.rb
```diff
@@ -0,0 +1,9 @@
+module UxHelper
```
rhymes, line 1: Can we rename this as `SettingsHelper` or `UsersHelper`, something a bit more generic? I'm saying this because `UX` also means user experience in the context of design of the user experience and it can be a bit confusing. As it's a helper that's used in what it's technically the settings maybe we can name it that.

Let me know what you think

benhalpern:
Yeah this looks great, but agree with @rhymes idea to rename that module.

app/helpers/ux_helper.rb
Oghenebrume50: I think you are right, and I feel ```SettingsHelper``` is better so I renamed it already to that @rhymes

PR #8295, opened 2020-06-05
Adds config.hosts option in development environment

fdoxyz:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Rails 6 introduced a security configuration to whitelist requests coming with a custom host. When working in development we sometimes want to use services like [ngrok](https://ngrok.io) or [xip.io](http://xip.io) to access our local dev server from other devices.

This only affects the development environment.
 
Screen Shot 2020-06-05 at 08 27 18
Screen Shot 2020-06-05 at 08 20 23

## Added tests?

- [ ] yes
- [x] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

## [optional] Are there any post deployment tasks we need to perform?

None

## [optional] What gif best describes this PR or how it makes you feel?

you dont recognize me, do you

PR #8677, opened 2020-06-15
Update config.hosts for GitPod

### LINKED ISSUE #8458: Add config.hosts instructions to GitPod Installation docs
**Describe the bug**
Opening DEV in Gitpod from with https://gitpod.io/#https://github.com/thepracticaldev/dev.to leads to a `Blocked Host` error after a successful install of everything and `bin/startup` (Screenshot below). This is due to `Host Authorization` in Rails 6 where `config.hosts` needs to be configured. 

It would be nice to have a helpful indication in the GitPod docs about this process and where to place the line. A new entry needs to be created in `application.yml` with `APP_DOMAN: <URL>` where the URL is the one mentioned in the error message within `" "`. Not sure if the GitPod file can be configured to do this automatically or not.

**To Reproduce**

1. Open https://gitpod.io/#https://github.com/thepracticaldev/dev.to and wait for installation and `bin/startup` to complete.
2. Open the Browser Preview tab if it doesn't open automatically.
3. A similar error with `Blocked Host` will be printed.

**Expected behavior**
Either the `APP_DOMAN` env var is automatically configured or there are instructions printed in docs and in GitPod console to do the task before `bin/startup` runs.

**Screenshots**
image

**Additional context**
Should perform a checkup if the process can be automated.

citizen428:

- [X] Developer Experience
- [X] Bug Fix

## Description

Rails 6 only allows requests from localhost, for accepting requests from other hosts config.hosts needs to be set. #8295 addresses this for our local development workflow and this PR fixes the issue on GitPod as well. This also includes the changes made by @snackattas in #8335 as I couldn't continue without them.

Closes #8458

## QA Instructions

* Start workspace on GitPod and make sure it works.

## Added tests?

## Added to documentation?

- [X] no documentation needed


config/environments/development.rb
```diff
@@ -75,6 +75,9 @@
   config.action_mailer.perform_caching = false
 
   config.hosts << ENV["APP_DOMAIN"] unless ENV["APP_DOMAIN"].nil?
+  if (gitpod_workspace_url = ENV["GITPOD_WORKSPACE_URL"])
```
citizen428, line 78: I'm not super happy about having this in` development.rb`, but I didn't feel like doing anything too clever given that Github Codespaces could conceivably replace Gitpod in the not too distant future.

.gitpod.dockerfile
```diff
@@ -2,8 +2,13 @@ FROM gitpod/workspace-postgres
 
 # Install Ruby
 ENV RUBY_VERSION=2.7.1
+RUN rm /home/gitpod/.rvmrc && touch /home/gitpod/.rvmrc && echo "rvm_gems_path=/home/gitpod/.rvm" > /home/gitpod/.rvmrc
 RUN bash -lc "rvm install ruby-$RUBY_VERSION && rvm use ruby-$RUBY_VERSION --default"
 
+# Install Node
+ENV NODE_VERSION=12.16.3
```
rhymes, line 9: can we read from `.nvmrc` here?

citizen428: Probably could, like I said I just pulled all of this in from @snackattas branch.

snackattas:
Thanks for getting my changes out with yours, appreciate it!

citizen428:
@snackattas Thanks for making them in the first place, very helpful! 😃

PR #8684, opened 2020-06-15
Footer redesign

ludwiczakpawel:

- [x] Refactor
- [x] Feature

## Description

This PR redesigns the footer.

🚨 This PR should be merged **AFTER** #8234 

image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


app/views/layouts/_footer.html.erb
```diff
@@ -1,10 +1,49 @@
...
+
+      <div>
+        <a href="/sponsors" class="crayons-link crayons-link--block">Sponsors</a>
+        <% if SiteConfig.shop_url.present? %>
+          <a href="<%= SiteConfig.shop_url %>" class="crayons-link crayons-link--block"><%= community_name %> Shop</a>
+        <% end %>
+      </div>
+    </nav>
+
+    <img src="<%= asset_path("footer-sloan.png") %>" alt="" width="52" height="120" class="crayons-footer__mascot">
```
rhymes, line 46: this should probably be generalized somehow, as not all communities will have Sloan as a mascot

nickytonline: We can do that in another PR as part of the Forem initiative for generalizing things.

thomasbnt:
Okay this is cute :octopus: 
image

nickytonline:
@ludwiczakpawel, you could change the z-index on the footer, but not sure if that's what we'd want to go for.

Jun-15-2020 14-01-33

The other option might be to not make the reactions and right sidebar non-sticky but I'm guessing that is off the table?

ludwiczakpawel:
> I noticed one issue in the article page where the footer gets hidden by the sidebar article boxes

> change the z-index on the footer, but not sure if that's what we'd want to go for.

@rhymes @nickytonline that's why I mentioned in the PR description that this should be merged AFTER we have new [article design merged](https://github.com/thepracticaldev/dev.to/pull/8234) :D. this overlapping bug will not be a problem then :) 

> I do have one general question though: shouldn't we remove infinite scrolling before we release the updated footer?

@rhymes I don't think it's needed for this PR, or am I missing something?

app/views/layouts/_footer.html.erb
ludwiczakpawel: Agree but I have no idea how to "generalize" it... I mean it's hard to expect every community will have this kind of "hanging mascot" for footer... On the other hand I think it's a nice touch and wouldn't like to drop it from DEV for sake of generalization... Any thoughts? Ideas? Should we maybe have some sort of "community.scss" file which could potentially include all these kind of community-specific style overwrites? Not sure if I like it, but I also can't think of an alternative...

ludwiczakpawel: cc @benhalpern thoughts on this?

nickytonline:
> @rhymes @nickytonline that's why I mentioned in the PR description that this should be merged AFTER we have new [article design merged](https://github.com/thepracticaldev/dev.to/pull/8234) :D. this overlapping bug will not be a problem then :)

Ahh sorry, missed that comment @ludwiczakpawel.

app/views/layouts/_footer.html.erb
benhalpern: This wouldn't be a style override.

Would end up being like `<%= cloudinary(SiteConfig.footer_mascot_image_url) %>`... We already do that with other mascot instances via `mascot_image_url` so this is a specialized config.

It will require a separate PR. It's straightforward but definitely outside of scope for a designer so I'll whip it up....right now and will be ready when you come back to this tomorrow.

benhalpern: Okay cool, #8692 solves this problem so we can get that merged asap... Note that I named it `mascot_footer_image_url` instead of what I wrote above.

rhymes:
> @rhymes I don't think it's needed for this PR, or am I missing something?

That's why I asked. I think this footer won't be reachable in the homepage leading to some people questioning why we have a footer at all as it already happened numerous times. What i mean is: we should probably merge a fix for the infinite scrolling BEFORE merging the footer redesign.

But let's not that be a blocker 👍

ludwiczakpawel:
@rhymes I think footer redesign is one thing and infinite scroll is completely separate thing. I'd say it's more of a feed issue rather than footer itself. Footer is indeed impacted by this but only on limited number of views. I think I wouldn't make those two dependent.

benhalpern:
Footer links are also available in two other accessible places on the home page besides the footer itself.... So I don't think we _need_ to rush that. It should be a more involved product decision.

Btw, mascot PR is merged, and we'll also want to do a final pass in terms of temporary CSS which could help this cleanly deploy.

ludwiczakpawel:
Sweet! I think we could ship it right after #8234 .

nickytonline:
Just a couple of conflicts to sort out @ludwiczakpawel, but aside from that, this is good to go.

benhalpern:
Okay I think we can ship on this as soon as the hardcoded sloan is changed to `mascot_footer_image_url` and the conflicts are changed.

Since this is below the fold old not layout-oriented. We can handle some edge case CSS issues, so I'd say we're good to go here once that stuff gets taken care of.

ludwiczakpawel:
@benhalpern @nickytonline conflicts resolved and mascot is now using `SiteConfig`. ready for merge I guess..

PR #8685, opened 2020-06-15
Routine rubocop fixes

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Routine Rubocop fixes, this time as we merged https://github.com/thepracticaldev/dev.to/pull/8638 and https://github.com/thepracticaldev/dev.to/pull/8643


app/models/feedback_message.rb
```diff
@@ -1,9 +1,9 @@
 class FeedbackMessage < ApplicationRecord
   resourcify
 
-  belongs_to :offender, foreign_key: "offender_id", class_name: "User", optional: true, inverse_of: :offender_feedback_messages
-  belongs_to :reporter, foreign_key: "reporter_id", class_name: "User", optional: true, inverse_of: :reporter_feedback_messages
-  belongs_to :affected, foreign_key: "affected_id", class_name: "User", optional: true, inverse_of: :affected_feedback_messages
+  belongs_to :offender, class_name: "User", optional: true, inverse_of: :offender_feedback_messages
```
mstruve, line 4: Since we are on Rails 6, does it just assume these values now based off the object name?

rhymes: It's not related to Rails 6, it was already the default in Rails 5.2. Basically Rails already assumes the default name of the foreign key is `#{name}_id`

PR #8687, opened 2020-06-15
Always set the default email address to ENV variable

Zhao-Andy:
- [x] Bug Fix

## Description
Before, whenever a site config was updated via the controller (`/internal/config`), it would remove the `default` email address, but still show it in the UI. This fixes it so the default email address is always set to the ENV variable. It also prevents it from being edited via a duplicate form in the view.

#8246 

1. Go to `/internal/config`
2. Add yourself as site config admin with:
```
me = User.find(your_id) # replace your_id with your actual ID
me.add_role :single_resource_admin, Config
```
2. Submit the form via the update button in the view
3. Check in console that the email address is still present:
```
SiteConfig.email_addresses
#=> :default should point to your ENV variable
```

### Before:

Screen Shot 2020-06-15 at 11 55 21 AM

### After:

Screen Shot 2020-06-15 at 11 54 47 AM

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

Don't update any site config variables until this is merged!

A suspicious cat looks around, with text that says, "Pretty sure... I smell a Monday here!"

### DIFF
--- a/app/controllers/internal/configs_controller.rb
+++ b/app/controllers/internal/configs_controller.rb
@@ -48,11 +48,12 @@ class Internal::ConfigsController < Internal::ApplicationController
       onboarding_params |
       job_params
 
+    params[:site_config][:email_addresses][:default] = ApplicationConfig["DEFAULT_EMAIL"]
     params.require(:site_config).permit(
       allowed_params,
       authentication_providers: [],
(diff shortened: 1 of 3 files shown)


mstruve:
Looks like a small spec issue to cleanup :)

PR #8693, opened 2020-06-15
Optimization: Remove unused indexes from Postgres

mstruve:
- [x] Optimization

## Description
These 3 indexes according to Postgres are unused and 2 of them take up more than a gig of space so I have chosen to remove them. 
Screen Shot 2020-06-15 at 4 34 26 PM

alt_text


rhymes:
@maestromac that table is huuuuge :D

PR #8695, opened 2020-06-15
Create profile image generator

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This is a quick and dirty way to create a profile image generator to help push `Sign in with Apple` over the line.

#7952 

The idea is that this helps us have default profile images on sign ups that don't come with avatars.

Usage...

In the GitHub auth we have this...

```ruby
  remote_profile_image_url: info.image.to_s
```

So in Apple and elsewhere we can have this...

```ruby
  remote_profile_image_url: Users::AvatarGenerator.call
```

And this will create a random profile image from Twemoji and colors from our pastel rainbow.

I don't know if this is the final solution, but I think it's a decent quick way to get this working. I chose Twemoji because it's liberally open source.

I'd love to replace this with something delightful and unique, but I think this is a good starter place for that in the future, and at least it's one less blocker on Apple.

### DIFF
--- a/app/services/users/profile_image_generator.rb
+++ b/app/services/users/profile_image_generator.rb
@@ -1,0 +1,33 @@
+# rubocop:disable MixinUsage
+include CloudinaryHelper
+
+module Users
+  module ProfileImageGenerator
+    EMOJI_IMAGES =
+      %w[dog-face_1f436.png
+         monkey-face_1f435.png
+         unicorn_1f984.png
+         mouse-face_1f42d.png
+         hamster_1f439.png
+         koala_1f428.png
+         bear_1f43b.png
+         panda_1f43c.png
+         penguin_1f427.png
+         spouting-whale_1f433.png
+         honeybee_1f41d.png
+         lion_1f981.png
+         tiger-face_1f42f.png
+         fox_1f98a.png
+         wolf_1f43a.png].freeze
+    BACKGROUND_HEXES = %w[#f68d8e #fce289 #f3f096 #55c1ae #88aedc #f8b4d0].freeze
+    def self.call
+      source = "https://emojipedia-us.s3.dualstack.us-west-1.amazonaws.com/thumbs/240/twitter/248/" + EMOJI_IMAGES.sample
+      cl_image_path(
+        source,
+        type: "fetch",
+        background: BACKGROUND_HEXES.sample,
+        sign_url: true,
+      )
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


vaidehijoshi:
> The idea is that this helps us have default profile images on sign ups that don't come with avatars.

This definitely seems good to do, and feels in line with forem work. I wonder if design could come up with some more delightful/unique "default" avatars, which we could cycle through and assign initially for new signups :)

benhalpern:
> This definitely seems good to do, and feels in line with forem work. I wonder if design could come up with some more delightful/unique "default" avatars, which we could cycle through and assign initially for new signups :)

Yup 10000%. I wanted to just get this up to unblock @rhymes' Apple PR (which doesn't come with a default profile image) but we should definitely put some delightful/onbrand profile generation on the list....Along with ability to modify or upload upon registration.

ludwiczakpawel:
I love the twemojis!

Should we be hosting these from external (looks like Emojipedia's) servers instead of our own? I mean, I don't care - I just hate external dependencies :D

app/services/users/profile_image_generator.rb
```diff
@@ -0,0 +1,33 @@
+# rubocop:disable MixinUsage
+include CloudinaryHelper
```
benhalpern, line 2: I included this here because locally it was appearing to fail when included within the module, but I might have been confused.

maestromac: Globally `include Cloudinary` does break test (and I assume it's breaking ActionView). This should be removed and custom setup needs to be added to the spec itself.

I can't find a way to get it to be included in the module but there's (generally) no need to include them here because wherever this module is being evoked from will either already include CloudinaryHelper or need to include it.  The catch is you can't call `Users::ProfileImageGenerator.call` on it's own, this module would have to be included and use the method `call` on its own. 

The other solution is to covert `ProfileImageGenerator` into a class and includes `CloudinaryHelper` within it

spec/services/users/profile_image_generator_spec.rb
```diff
@@ -0,0 +1,7 @@
+require "rails_helper"
+
```
maestromac, line 2: If `Users::ProfileImageGenerator` is kept as a module, then this spec will need this setup to mimic when other class might utilize it. It works but I'm uncertain if this will break _other_ test setups.
```suggestion
require "rails_helper"

RSpec.configure do |c|
  include CloudinaryHelper
end

```

app/services/users/profile_image_generator.rb
rhymes: I also have the feeling that globally including CloudinaryHelper there means that when Rails autoloads it will be included everywhere.

PR #8696, opened 2020-06-15
Update Rack to Fix Bundle Audit Failure

mstruve:
Bundle Audit Fix
```
ruby-advisory-db: 454 advisories
Name: rack
Version: 2.2.2
Advisory: CVE-2020-8184
Criticality: Unknown
URL: https://groups.google.com/g/rubyonrails-security/c/OWtmozPH9Ak
Title: Percent-encoded cookies can be used to overwrite existing prefixed cookie names
Solution: upgrade to ~> 2.1.4, >= 2.2.3

Vulnerabilities found!
```

alt_text

PR #8698, opened 2020-06-15
Flaky Spec Fix: Replace let_it_be_changeable with let for stripe spec

mstruve:
- [x] Bug Fix

## Description
While updating rack to fix our bundle audit failure #8696 I had a bunch of specs in `spec/requests/stripe_active_cards_spec.rb` fail with the following error
```
1) StripeActiveCards POST /stripe_active_cards increments sidekiq.errors.new_subscription in Datadog on failure
     Failure/Error: sign_in user
     ActiveRecord::RecordNotFound:
       Couldn't find User with 'id'=1097
     # ./spec/requests/stripe_active_cards_spec.rb:11:in `block (2 levels) in <top (required)>'
     # ./spec/rails_helper.rb:136:in `block (3 levels) in <top (required)>'
     # ./spec/rails_helper.rb:136:in `block (2 levels) in <top (required)>'
```

Upon looking in the file I noticed we were persisting the user through the spec so if anything removed it the rest of the specs would fail. I updated the spec so the user is created for each spec rather than persisted. 

@snackattas 

alt_text

### DIFF
--- a/spec/requests/stripe_active_cards_spec.rb
+++ b/spec/requests/stripe_active_cards_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 RSpec.describe "StripeActiveCards", type: :request do
-  let_it_be_changeable(:user) { create(:user) }
+  let(:user) { create(:user) }
   let(:stripe_helper) { StripeMock.create_test_helper }
   let(:card_token) { stripe_helper.generate_card_token }
 


rhymes:
@citizen428 yeah :(

citizen428:
@rhymes That makes me wonder if we're using them wrong tbh, because knowing Vladimir it's hard to imagine that they are broken on a fundamental level.

rhymes:
We might be

PR #8701, opened 2020-06-16
Fix node version on GitPod

citizen428:

- [X] Bug Fix

## Description

Trying to start the app on GitPod leads to the following:

```
== Restarting application server ==
== STARTING UP ==
04:13:02 web.1       | started with pid 23441
04:13:02 webpacker.1 | started with pid 23442
04:13:02 sidekiq.1   | started with pid 23443
04:13:04 webpacker.1 | yarn run v1.22.4
04:13:04 webpacker.1 | error Command "webpack-dev-server" not found.
04:13:04 webpacker.1 | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
04:13:05 webpacker.1 | exited with code 1
04:13:05 system      | sending SIGTERM to all processes
04:13:05 web.1       | terminated by SIGTERM
04:13:05 sidekiq.1   | terminated by SIGTERM
```

Trying to manually restart `webpack-dev-server`:

```
gitpod /workspace/dev.to $ webpack-dev-server
bash: webpack-dev-server: command not found
```

Trying to bundle:

```
gitpod /workspace/dev.to $ yarn
yarn install v1.22.4
[1/5] Validating package.json...
error dev.to@1.0.0: The engine "node" is incompatible with this module. Expected version "12.16.x". Got "12.18.0"
```
But it seems we're not using it, as the 12.18 version is still in the path until issuing an `nvm use command`

n/a

## Added tests?

## Added to documentation?

- [X] no documentation needed

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -19,8 +19,7 @@ tasks:
   - init: cp config/sample_application.yml config/application.yml
     command: >
       gem install solargraph;
-      bin/setup &&
-      bin/startup
+      nvm use default && bin/setup && bin/startup
 github:
   prebuilds:
     # enable for the master/default branch (defaults to true)
--- a/.gitpod.dockerfile
+++ b/.gitpod.dockerfile
@@ -7,7 +7,7 @@ RUN bash -lc "rvm install ruby-$RUBY_VERSION && rvm use ruby-$RUBY_VERSION --def
 
 # Install Node
 ENV NODE_VERSION=12.16.3
-RUN bash -lc ". .nvm/nvm.sh && nvm install $NODE_VERSION"
+RUN bash -lc ". ~/.nvm/nvm.sh && nvm install $NODE_VERSION && nvm alias default $NODE_VERSION && nvm use $NODE_VERSION"
 
 # Install Redis.
 RUN sudo apt-get update \


citizen428:
Sorry for the force pushes here, made some recent email changes (commits to this repo now are from my @dev.to address) and GH/CLAassistant needed some convincing.

.gitpod.yml
```diff
@@ -14,6 +14,9 @@ ports:
   - port: 9300
     onOpen: ignore
 tasks:
+  - before: >
```
citizen428, line 17: `nvm` is already available in the base image, so we don't need to do so much manual setup. Note the `--reinstall-packages-from=node` flag, which makes sure `yarn` will be available for the newly installed version too.

Since RVM is available too we should be able to manage Ruby the same way, but when I tried that I always ended up with 2.6.6 and I didn't want to invest more time in this.

citizen428:
@snackattas FYI

.gitpod.yml
```diff
@@ -14,10 +14,15 @@ ports:
...
+      nvm install 12.16.3 --reinstall-packages-from=node;
+      rvm install ruby-2.7.1
   - command: redis-server
   - command: /home/gitpod/elasticsearch-7.5.2/bin/elasticsearch
   - init: cp config/sample_application.yml config/application.yml
     command: >
+      nvm use 12.16.3;
```
citizen428, line 24: Without this the correct versions were only available in the shell session the `before` command ran in.

joshpuetz:
I tried this on Gitpod (using their URL prefix trick to open a workspace: https://gitpod.io/#https://github.com/thepracticaldev/dev.to/pull/8721/files#r441585424), and I still had to manually switch to node 12.16.3 via the build in terminal

joshpuetz:
Finally getting back to this, and it's working for me on Gitpod!

citizen428:
Thanks for the feedback @joshpuetz, marking this as ready for review now.

donofriov:
This is still failing in the gitpod container:

```
gitpod /workspace/dev.to $ bin/yarn
yarn install v1.22.4
[1/5] Validating package.json...
error dev.to@1.0.0: The engine "node" is incompatible with this module. Expected version "12.16.x". Got "12.18.2"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
gitpod /workspace/dev.to $ nvm list
       v12.16.3
->     v12.18.2
default -> 12.16.3 (-> v12.16.3)
```

image

citizen428:
Closed in favor of #9271.

citizen428:
@donofriov Agreed, there's definitely something funky going on. It can be relatively hard to debug Gitpod because of this, because it seems containers only get rebuilt when you start a completely new session. Anyway, thanks for all the effort you put into this, I don't use Gitpod myself, which is why this was never particularly prioritized.

PR #8704, opened 2020-06-16
Remove Audit::Event::Util

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Rails 6 added support to serialize Time records in ActiveSupport events:

> We currently support String, Integer, Float, NilClass, TrueClass, FalseClass, BigDecimal, Symbol, Date, Time, DateTime, ActiveSupport::TimeWithZone, ActiveSupport::Duration, Hash, ActiveSupport::HashWithIndifferentAccess, Array or GlobalID::Identification instances

thus this class, added in https://github.com/thepracticaldev/dev.to/pull/3449, as noted in the comments, is not needed anymore.

see also https://github.com/rails/rails/tree/6-0-stable/activejob/lib/active_job/serializers

## Added tests?

- [x] yes
- [ ] no, because I need help


spec/services/audit/notification_spec.rb
```diff
@@ -16,6 +16,7 @@ def notify
     described_class.notify(listener) do |payload|
       payload.user_id = user.id
       payload.roles = user.roles.pluck(:name)
+      payload.data = build(:activesupport_event)
```
rhymes, line 19: making sure it actually works by serializing a custom even with time ;-)

spec/system/articles/user_visits_articles_by_tag_spec.rb
```diff
@@ -38,19 +38,19 @@
```
rhymes, line 38: fixed flaky spec

mstruve: What flaky spec did you fix with this? Bc it is failing a lot with this addition. any title with a special character breaks https://github.com/thepracticaldev/dev.to/pull/8714

PR #8707, opened 2020-06-16
Rails 6 autoloader: prepare liquid tags code

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Rails 6 has a new autoloader called [zeitwerk](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html) which solves a bunch of issues but it's a bit stricter than the classic. 

In preparation for the switch I'm going to tackle a few issues at a time. 

This PR contains renames and code changes for liquid tags.

## Added tests?

- [x] yes
- [ ] no, because I need help


app/liquid_tags/standard_filters.rb
```diff
@@ -1,75 +0,0 @@
-module StandardFilters
```
rhymes, line 1: variables are disabled, see test. There's no possibility to use filters anyway:

```ruby
[1] pry(main)> Liquid::Template.parse("{{ a | capitalize }}").render({"a" => "a"})
Liquid::StandardError: Liquid error: Liquid variables are disabled
from devto/lib/liquid/variable.rb:6:in `initialize'
```

also we weren't monkey patching them anyway, as by re-enabling variables and keeping this code filters are still working:

```ruby
[1] pry(main)> Liquid::Template.parse("{{ a | capitalize }}").render({"a" => "a"})
=> "A"
```

app/liquid_tags/url_decode_filter.rb
```diff
@@ -1,7 +0,0 @@
-module UrlDecodeFilter
```
rhymes, line 1: another filter that was never actually called

PR #8708, opened 2020-06-16
Bug Fix: Dont Freeze Timeframer Timestamps bc They Will be Inaccurate

mstruve:
- [x] Bug Fix

## Description
We cannot be freezing Timestamps in our Timeframer bc that will cause it to become inaccurate since the timestamps will be frozen to the time the application boots up. The longer the application has been booted, the more inaccurate these will be. This became very apparent when a spec ordering issued caused this class to be instantiated while Timecop had time frozen in March 2019 which caused later specs to fail. 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/labor/timeframer.rb
+++ b/app/labor/timeframer.rb
@@ -2,21 +2,25 @@ class Timeframer
   attr_accessor :timeframe
 
   LATEST_TIMEFRAME = "latest".freeze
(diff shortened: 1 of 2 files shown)


rhymes:
thanks! I just bumped into this :D

PR #8709, opened 2020-06-16
Mod center quick mod

Zhao-Andy:
- [x] Feature

## Description
This uses the quick mod functionality that was in the previous `dev.to/mod`

1. Click an article row in the mod center
2. open quick mod
3. attempt this for multiple articles

Screenshot_2020-06-16 Moderate - DEV(local) Community 👩‍💻👨‍💻

## Added tests?
- [x] no, not yet 😬 

## Added to documentation?
- [x] no documentation needed


app/javascript/modCenter/singleArticle/index.jsx
```diff
@@ -1,88 +1,76 @@
 import PropTypes from 'prop-types';
-import { h } from 'preact';
+import { h, Component } from 'preact';
+import { formatDate } from './util';
```
msarit, line 3: Nice abstraction! 😄

app/javascript/modCenter/singleArticle/index.jsx
```diff
@@ -1,88 +1,76 @@
...
-      </span>
-      <span className="article-author">
-        {newAuthorNotification()}
-        {user.name}
-      </span>
-      <span className="article-published-at">{formatDate(publishedAt)}</span>
-    </div>
-  );
-};
+    const newAuthorNotification = user.articles_count <= 3 ? '👋 ' : '';
```
msarit, line 41: Ugh! Ternary operator!! Why do I always forget? lol

app/javascript/modCenter/singleArticle/index.jsx
```diff
@@ -1,88 +1,76 @@
...
+          </header>
+          {tags}
+        </span>
+        <span className="article-author">
+          {newAuthorNotification}
+          {user.name}
+        </span>
+        <span className="article-published-at">{formatDate(publishedAt)}</span>
+        <div
+          className="article-iframes-container"
```
msarit, line 61: A few comments:

1. I think the `iframe` border looks clunky 😟 We could set `frameBorder="0"` on the `iframe` element, which would remove the border completely, but then the iframes look like they're floating (🚫). We could then add `border: 1px solid var(--bg-color)` which gives the iframes a subtle border, but then we'd need to hide this border when the article is collapsed. So perhaps we leave it as-is for now. Just looks clunky...

2. This is minor, but the snackbar confirmations appear way down the article, and will probably not be seen by the mod (unless the article is super-short). Do we care? I really don't

PR #8718, opened 2020-06-16
Fix axe warnings on the user dashboard

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This commit fixes most of the accessibility complaints from axe on the user dashboard. There are no visual UI changes.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/assets/stylesheets/scaffolds.scss
+++ b/app/assets/stylesheets/scaffolds.scss
@@ -65,6 +65,15 @@ body {
   font-size: 18px;
 }
 
+/* Hide headers for screen readers */
+.visually-hidden-header {
+  position: absolute;
+  left: -10000px;
+  width: 1px;
+  height: 1px;
+  overflow: hidden;
+}
+
 a {
   color: var(--link-brand-color);
   text-decoration: none;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/scaffolds.scss
```diff
@@ -65,6 +65,15 @@ body {
   font-size: 18px;
 }
 
+/* Hide headers for screen readers */
```
nickytonline, line 68: You can use `aria-hidden` on the headers and that should remove them from the accessibility tree.

jacobherrington: I think the intention is actually the opposite. Rather than hiding these headers from the Accessbility API, I want them visually gone, but available to people using assistive technologies.

Here is an example: https://www.accessibility-developer-guide.com/examples/headings/visually-hidden-headings/

We could use these headers in many more places on the site, but I'm trying to use them at minimum in the main landmarks.

jacobherrington: Although the comment is misleading now that I read it without context. I will change that 😬

nickytonline: Ahh, gotcha. All good then.

PR #8719, opened 2020-06-16
Optimization: Eager Load and Limit Columns When Fetching Articles Not IDs

mstruve:
- [x] Optimization

## Description
Recently we have had a couple alerts for requests taking longer than expected in the application. I have been able to trace these requests back to loading `additional_content_boxes` so I went digging in to see if I could find what queries were causing problems. One issue I found was that we were no longer eager loading the user or limiting our columns selected when loading the articles for the view. This fixes that. 

## Added tests?

alt_text


mstruve:
I was wondering the same thing when I found this!

PR #8721, opened 2020-06-16
Add SiteConfig for feed style

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR allows admins to set `feed_style` which is the default way the feed looks.

Right now toggling between `rich` and `basic` determines whether to always show the cover image or not. I envision a `compact` feed which would cut out a bunch of the details present in either basic or rich.

This is purposefully very high level so that we can change what "rich" or "basic" mean for new designs in future.

I see this as something users could also adjust to override defaults in future, but wanted to start with this basic functionality to help forems bring rich content to the feeds.

Likewise, we can also allow forems to set a different default font, which the user could override (we can also let users carry their configs between forems by default somehow down the road).

Closes https://github.com/thepracticaldev/dev.to/issues/7766

How it looks in development....

Screen Shot 2020-06-16 at 7 25 10 PM


app/controllers/async_info_controller.rb
```diff
@@ -71,6 +71,7 @@ def user_data
         moderator_for_tags: @user.moderator_for_tags,
         config_body_class: @user.config_body_class,
         pro: @user.pro?,
+        feed_style: SiteConfig.feed_style,
```
mstruve, line 74: Can we find a better place to serve this from? This hash is all comprised of `user_data` and this is a Site configuration. It feels like this should be its own standalone endpoint if we want a way to serve SiteConfigs to our views.

app/javascript/packs/homePageFeed.jsx
```diff
@@ -81,6 +81,7 @@ export const renderFeed = (timeFrame) => {
         };
 
         const [featuredStory, ...subStories] = feedItems;
+        const feedStyle = JSON.parse(document.body.dataset.user).feed_style;
```
joshpuetz, line 84: Similar to Molly's comment at https://github.com/thepracticaldev/dev.to/pull/8721/files#r441585424, it feels odd to be pulling a site configuration option from user information

app/controllers/async_info_controller.rb
benhalpern: The idea here is that this value will be overrideable by users in their `/settings/ux` so while that's not available yet, this will essentially be user information like anything else and this is just the default.

app/views/internal/configs/show.html.erb
```diff
@@ -689,6 +688,26 @@
...
+                     locals: {
+                       header: "User Experience",
+                       state: "collapse",
+                       target: "uxBodyContainer",
+                       expanded: "false"
+                     } %>
+          <div id="uxBodyContainer" class="card-body collapse hide" aria-labelledby="uxBodyContainer">
+            <div class="form-group">
+              <%= f.label :feed_style %>
+              <%= f.text_field :feed_style,
```
Ridhwana, line 702: Since there is a limited set of values for this field, it would make more sense for this to be a dropdown or radio buttons instead of a text field.

app/views/articles/_single_story.html.erb
```diff
@@ -1,9 +1,11 @@
 <div class="crayons-story <% if featured == true %>crayons-story--featured<% end %>" data-content-user-id="<%= story.user_id %>">
   <% if featured == true %>
-  <div id="featured-story-marker" data-featured-article="articles-<%= story.id %>"></div>
-  <a href="<%= story.path %>" class="crayons-story__cover" aria-label="featured article" data-featured-article="articles-<%= story.id %>">
-    <div style="background-color: <%= story.main_image_background_hex_color %>; background-image: url(<%= cloud_cover_url(story.main_image) %>);" class="crayons-story__cover__image"></div>
-  </a>
+    <div id="featured-story-marker" data-featured-article="articles-<%= story.id %>"></div>
+  <% end %>
+  <% if featured == true || (SiteConfig.feed_style == "rich" && story.main_image.present?) %>
```
mstruve, line 5: If we intend to make this user-specific we should probably reference the value from the user view data here so when it becomes part of the user UX there is only one spot we have to update.

app/controllers/async_info_controller.rb
mstruve: _sigh_ fine, this still makes me cringe 😉

app/views/articles/_single_story.html.erb
benhalpern: Good call. This specific area is server-rendered and is only served to logged-out users, but I added a new method in `ApplicationController` where which we could call from within async controller or here and will soon serve different content if user exists and has different preferences.

benhalpern:
Alright. Tests pass, I promise I'll follow up with the other part where users can configure this so that it makes sense in the async_info method, and in the meantime wrote the method so we don't have to change two places.

So I think it's good to go @mstruve unless there's a bug 😄

app/controllers/application_controller.rb
```diff
@@ -82,6 +82,13 @@ def internal_navigation?
   end
   helper_method :internal_navigation?
 
+  def feed_style_preference
```
mstruve, line 85: 🔥

PR #8723, opened 2020-06-17
Add subscriber_email to user_subscriptions & more backend updates

atsmith813:
- [x] Optimization

## Description
I had another PR ready to go, but it became quite large so this is me breaking it out into smaller chunks. This PR adds a `subscriber_email` field to `user_subscriptions` so we can record the subscriber's email address at the time of subscription to display to an author.

For the scope of this feature and CodeLand, a user can subscribe to an author via an article one time. If they change their email address, they cannot re-subscribe to the same article with the new email address. That functionality will be in a future iteration post-CodeLand.

We are also intentionally allowing an author to subscribe to themself. That way they can test and experiment on their own.

I've included some name changes for some relationships and updated some DB indexes.

```ruby
subscriber = User.last
article = Article.first

UserSubscription.make(source: article, subscriber: subscriber)
UserSubscription.build(source: article, subscriber: subscriber)

article.create_user_subscription(subscriber)
article.build_user_subscription(subscriber)
```

I added the latter two to a model concern and moved the associations in there as well. Adding this to comments in the future will just be:
```ruby
include UserSubscriptionSourceable
```

_We are restricting this liquid tag only for use in `Articles` (not `Comments`, or `Pages`) and for use only by approved organizations for CodeLand (using a new role)._
- [ ] [Create a new endpoint to handle email signups from the new liquid tag](https://github.com/thepracticaldev/dev.to/issues/8171)
- [ ] [Create an email signup button liquid tag](https://github.com/thepracticaldev/dev.to/issues/8168)
- [ ] [Add new tab to User dashboard to display email addresses from the email signup button liquid tag](https://github.com/thepracticaldev/dev.to/issues/8169)

https://github.com/thepracticaldev/dev.to/issues/8722

You can create some objects locally if you'd like to test the associations.

```ruby
# If you're copying and pasting, make sure you execute this setup in order.
subscriber = User.last
article = Article.first
author = article.user

user_subscription = UserSubscription.create(user_subscription_sourceable: article, subscriber_id: subscriber.id, subscriber_email: subscriber.email, author_id: author.id)

##########################

# Test source_authored_user_subscriptions

# Should return an ActiveRecordCollection with 1 UserSubscription from above
author.source_authored_user_subscriptions
=> [UserSubscription]

##########################

# Test subscribed_to_user_subscriptions

# Should return an ActiveRecordCollection with 1 UserSubscription from above
subscriber.subscribed_to_user_subscriptions
=> [UserSubscription]

##########################

# Test subscriber

# Should return 1 User object - the subscriber from above
user_subscription.subscriber
=> User

##########################

# Test author

# Should return 1 User object - the author from above
user_subscription.author
=> User

##########################

# Test user_subscription_sourceable

# Should return 1 Article object - the article from above
user_subscription.user_subscription_sourceable
=> Article

##########################

# Test subscriber_email
user_subscription.subscriber_email == subscriber.email
=> true

##########################

# Bonus points, try to create invalid user_subscriptions (i.e. missing attributes,
# duplicate records, etc.) and you should see it fail.
```

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

will_smith_getting_ready_gif

### DIFF
--- a/app/models/concerns/user_subscription_sourceable.rb
+++ b/app/models/concerns/user_subscription_sourceable.rb
@@ -12,7 +12,7 @@ module UserSubscriptionSourceable
              foreign_key: :user_id
   end
 
-  def new_user_subscription(subscriber)
+  def build_user_subscription(subscriber)
     UserSubscription.new(user_subscription_attributes(subscriber))
   end
 
--- a/app/models/concerns/user_subscription_sourceable.rb
+++ b/app/models/concerns/user_subscription_sourceable.rb
@@ -25,7 +25,7 @@ module UserSubscriptionSourceable
   def user_subscription_attributes(subscriber)
     {
       user_subscription_sourceable: self,
-      author_id: self&.user_id,
+      author_id: user_id,
       subscriber_id: subscriber&.id,
       subscriber_email: subscriber&.email
     }
--- a/app/models/concerns/user_subscription_sourceable.rb
+++ b/app/models/concerns/user_subscription_sourceable.rb
@@ -1,0 +1,33 @@
+module UserSubscriptionSourceable
+  extend ActiveSupport::Concern
+
+  # This all assumes there's an association with User under the column user_id.
+
+  included do
+    has_many :user_subscriptions, as: :user_subscription_sourceable
+    has_many :sourced_subscribers,
+             class_name: "User",
+             through: :user_subscriptions,
+             source: :subscriber,
+             foreign_key: :user_id
+  end
+
+  def new_user_subscription(subscriber)
+    UserSubscription.new(user_subscription_attributes(subscriber))
+  end
+
+  def create_user_subscription(subscriber)
+    UserSubscription.create(user_subscription_attributes(subscriber))
+  end
+
+  private
+
+  def user_subscription_attributes(subscriber)
+    {
+      user_subscription_sourceable: self,
+      author_id: self&.user_id,
+      subscriber_id: subscriber&.id,
+      subscriber_email: subscriber&.email
+    }
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -45,6 +45,10 @@ class User < ApplicationRecord
   acts_as_followable
   acts_as_follower
 
+  has_many :authored_user_subscriptions, class_name: "UserSubscription", foreign_key: :author_id, inverse_of: :author, dependent: :destroy
+  has_many :subscribers, through: :authored_user_subscriptions, dependent: :destroy
+  has_many :subscribed_to_user_subscriptions, class_name: "UserSubscription", foreign_key: :subscriber_id, inverse_of: :subscriber, dependent: :destroy
+
   has_many :access_grants, class_name: "Doorkeeper::AccessGrant", foreign_key: :resource_owner_id, inverse_of: :resource_owner, dependent: :delete_all
   has_many :access_tokens, class_name: "Doorkeeper::AccessToken", foreign_key: :resource_owner_id, inverse_of: :resource_owner, dependent: :delete_all
   has_many :affected_feedback_messages, class_name: "FeedbackMessage", inverse_of: :affected, foreign_key: :affected_id, dependent: :nullify
@@ -67,8 +71,6 @@ class User < ApplicationRecord
   has_many :display_ad_events, dependent: :destroy
   has_many :email_authorizations, dependent: :delete_all
   has_many :email_messages, class_name: "Ahoy::Message", dependent: :destroy
-  has_many :user_subscriptions, foreign_key: :author_id, inverse_of: :author, dependent: :destroy
-  has_many :subscribers, through: :user_subscriptions, dependent: :destroy
   has_many :field_test_memberships, class_name: "FieldTest::Membership", as: :participant, dependent: :destroy
   has_many :github_repos, dependent: :destroy
   has_many :html_variants, dependent: :destroy
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -45,8 +45,8 @@ class User < ApplicationRecord
   acts_as_followable
   acts_as_follower
 
-  has_many :authored_user_subscriptions, class_name: "UserSubscription", foreign_key: :author_id, inverse_of: :author, dependent: :destroy
-  has_many :subscribers, through: :authored_user_subscriptions, dependent: :destroy
+  has_many :source_authored_user_subscriptions, class_name: "UserSubscription", foreign_key: :author_id, inverse_of: :author, dependent: :destroy
+  has_many :subscribers, through: :source_authored_user_subscriptions, dependent: :destroy
   has_many :subscribed_to_user_subscriptions, class_name: "UserSubscription", foreign_key: :subscriber_id, inverse_of: :subscriber, dependent: :destroy
 
   has_many :access_grants, class_name: "Doorkeeper::AccessGrant", foreign_key: :resource_owner_id, inverse_of: :resource_owner, dependent: :delete_all
--- a/app/models/user_subscription.rb
+++ b/app/models/user_subscription.rb
@@ -13,4 +13,21 @@ class UserSubscription < ApplicationRecord
   validates :subscriber_id, presence: true, uniqueness: { scope: %i[subscriber_email user_subscription_sourceable_type user_subscription_sourceable_id] }
   validates :user_subscription_sourceable_id, presence: true
   validates :user_subscription_sourceable_type, presence: true, inclusion: { in: ALLOWED_TYPES }
+
+  def self.make_new(source:, subscriber:)
+    new(build_attributes(source, subscriber))
+  end
+
+  def self.make(source:, subscriber:)
+    create(build_attributes(source, subscriber))
+  end
+
+  def self.build_attributes(source, subscriber)
+    {
+      user_subscription_sourceable: source,
+      author_id: source&.user_id,
+      subscriber_id: subscriber&.id,
+      subscriber_email: subscriber&.email
+    }
+  end
 end
--- a/app/models/user_subscription.rb
+++ b/app/models/user_subscription.rb
@@ -14,7 +14,7 @@ class UserSubscription < ApplicationRecord
   validates :user_subscription_sourceable_id, presence: true
   validates :user_subscription_sourceable_type, presence: true, inclusion: { in: ALLOWED_TYPES }
 
-  def self.make_new(source:, subscriber:)
+  def self.build(source:, subscriber:)
     new(build_attributes(source, subscriber))
   end
 
--- a/app/models/user_subscription.rb
+++ b/app/models/user_subscription.rb
@@ -4,12 +4,13 @@
 class UserSubscription < ApplicationRecord
   ALLOWED_TYPES = %w[Article].freeze
 
-  belongs_to :author, class_name: "User", inverse_of: :user_subscriptions
-  belongs_to :subscriber, class_name: "User", inverse_of: :user_subscriptions
+  belongs_to :author, class_name: "User", inverse_of: :authored_user_subscriptions
+  belongs_to :subscriber, class_name: "User", inverse_of: :subscribed_to_user_subscriptions
   belongs_to :user_subscription_sourceable, polymorphic: true
 
   validates :author_id, presence: true
-  validates :subscriber_id, presence: true, uniqueness: { scope: %i[user_subscription_sourceable_type user_subscription_sourceable_id] }
+  validates :subscriber_email, presence: true
+  validates :subscriber_id, presence: true, uniqueness: { scope: %i[subscriber_email user_subscription_sourceable_type user_subscription_sourceable_id] }
   validates :user_subscription_sourceable_id, presence: true
   validates :user_subscription_sourceable_type, presence: true, inclusion: { in: ALLOWED_TYPES }
 end
--- a/app/models/user_subscription.rb
+++ b/app/models/user_subscription.rb
@@ -4,7 +4,7 @@
 class UserSubscription < ApplicationRecord
   ALLOWED_TYPES = %w[Article].freeze
 
-  belongs_to :author, class_name: "User", inverse_of: :authored_user_subscriptions
+  belongs_to :author, class_name: "User", inverse_of: :source_authored_user_subscriptions
   belongs_to :subscriber, class_name: "User", inverse_of: :subscribed_to_user_subscriptions
   belongs_to :user_subscription_sourceable, polymorphic: true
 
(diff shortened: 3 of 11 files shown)


app/models/user.rb
```diff
@@ -45,6 +45,10 @@ class User < ApplicationRecord
   acts_as_followable
   acts_as_follower
 
+  has_many :authored_user_subscriptions, class_name: "UserSubscription", foreign_key: :author_id, inverse_of: :author, dependent: :destroy
```
atsmith813, line 48: The `UserSubscriptions` the `User` is an `Author` of.

app/models/user.rb
```diff
@@ -45,6 +45,10 @@ class User < ApplicationRecord
...
+  has_many :subscribers, through: :authored_user_subscriptions, dependent: :destroy
+  has_many :subscribed_to_user_subscriptions, class_name: "UserSubscription", foreign_key: :subscriber_id, inverse_of: :subscriber, dependent: :destroy
```
atsmith813, line 50: The `UserSubscriptions` the `User` is a `Subscriber` of.

db/migrate/20200617014509_add_subscriber_email_index_to_user_subscriptions.rb
```diff
@@ -0,0 +1,45 @@
+class AddSubscriberEmailIndexToUserSubscriptions < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def up
+    if index_exists?(:user_subscriptions, %i[subscriber_id user_subscription_sourceable_id user_subscription_sourceable_type])
```
atsmith813, line 5: Can't wait for when we update to Rails 6.1 and get access to `if_exists` and `if_not_exists` options on indexes 🎉

db/migrate/20200617014320_add_subscriber_email_to_user_subscriptions.rb
```diff
@@ -0,0 +1,5 @@
+class AddSubscriberEmailToUserSubscriptions < ActiveRecord::Migration[6.0]
+  def change
+    add_column :user_subscriptions, :subscriber_email, :string, null: false
```
atsmith813, line 3: The `null: false` without a `default` should be fine here since there are no records in Production.

db/migrate/20200617014509_add_subscriber_email_index_to_user_subscriptions.rb
```diff
@@ -0,0 +1,45 @@
...
+      remove_index :user_subscriptions,
+                   column: %i[subscriber_id user_subscription_sourceable_id user_subscription_sourceable_type],
+                   algorithm: :concurrently
+    end
+
+    if !index_exists?(:user_subscriptions, :subscriber_email)
```
citizen428, line 11: I'm surprised Rubocop didn't try to change this to `unless index_exists?`

app/models/user.rb
citizen428: This took me a moment to untangle. You mean the user is the author of the article that another user subscribed to, right?

atsmith813: That's correct! In the future, it could be the author of the `Comment` or something else. It's the author of the source.

db/migrate/20200617014509_add_subscriber_email_index_to_user_subscriptions.rb
atsmith813: I prefer `unless index_exists?` as well. It seems we have used `!index_exists?` as well 😅 . It's a dead tie with this included. I switched it 💯

app/models/user.rb
atsmith813: Do you think it would be better if I did `:source_authored_user_subscriptions`?

citizen428, line 48: Yes, I think the passive voice does actually clear it up a bit more. 

BTW: recent discussion on `foreign_key`: https://github.com/thepracticaldev/dev.to/pull/8685#discussion_r440304148

app/models/user.rb
```diff
@@ -45,6 +45,10 @@ class User < ApplicationRecord
   acts_as_followable
   acts_as_follower
 
+  has_many :source_authored_user_subscriptions, class_name: "UserSubscription", foreign_key: :author_id, inverse_of: :author, dependent: :destroy
```
atsmith813, line 48: The `UserSubscriptions` where the `User` is the `Author` of the source.

app/models/user.rb
```diff
@@ -45,6 +45,10 @@ class User < ApplicationRecord
...
+  has_many :subscribers, through: :source_authored_user_subscriptions, dependent: :destroy
+  has_many :subscribed_to_user_subscriptions, class_name: "UserSubscription", foreign_key: :subscriber_id, inverse_of: :subscriber, dependent: :destroy
```
atsmith813, line 50: The `UserSubscriptions` where the `User` is a `Subscriber`.

atsmith813:
@rhymes 

It almost works that way. You can do:
```ruby
UserSubscription.create(user_subscription_sourceable: article, subscriber: subscriber, author: author, subscriber_email: subscriber.email)
```
- `source` breaks polymorphism/Rails convention. I can change it to `sourceable` and/or write an alias on the model. More on this below.
- The reason `author` and `subscriber` work is because the association is set up based on their respective IDs. There's no association for `subscriber_email` so that doesn't get picked up by `subscriber`.



> also: `author`, is this the author of the article? can't we get that from `article.user`? I'm confused :D 

That's correct! From my code example above:
```ruby
article = Article.first
author = article.user
```
That's where we do get it from. We need to set/save `author_id` explicitly on the table. If we took it out, the queries would be expensive to join various object types to find `User`. That and in the future the Object type for Source (`Article`, `Comment`, etc.) could have the association named differently. Maybe it's an object that has `author_id` and a relationship called `Author` instead of `user_id`/`User`. So we don't want to assume/always set it to be `User`. Instead, it's set when you create the object.



> `author.source_authored_user_subscriptions` can be `author.email_subscriptions` ?

In this case, it's not clear, are those `email_subscriptions` ones the `author` (which is just a variable name holding a `User` object, could be called `user` or anything else) is the `Author` on? Or the `Subscriber`? `.email_subscriptions` has 2 relationships to a `User` - `Author` and `Subscriber`.



>`user_subscription.user_subscription_sourceable` can be `user_subscription.source` ?

This breaks the polymorphism/Rails convention. It would make it more clear what it is, but it would further obfuscate what's going on. When you see `user_subscription_sourceable` or `sourceable`, you automatically have an idea of how that's expected to behave. You know it's polymorphic and it should have two columns - `type` and `id`.

Changing it to `source` reads better. It may be a little confusing as to how that's actually working.  Seeing source, you're more likely to expect a `source_id` column to exist on the table - in this case, it doesn't. I thought about moving sources out to another table to achieve that effect fully. It makes the queries I need more expensive by causing another join.

The other thing about shortening it to `sourceable` is it's likely we can refactor out the associations into a model concern (common for this polymorphic setup) in the future and it becomes `include UserSubscriptionSourceable` vs `include Sourceable`. Here it's more clear what the model in question is a `source` of with the way it is now.

-------

Are there some more areas of the naming that are confusing? Maybe that will help me get some ideas on how to make it more clear.

rhymes:
@atsmith813 thanks for the explanation!

> source breaks polymorphism/Rails convention. I can change it to sourceable and/or write an alias on the model. More on this below.

yeah I don't recommend changing the actual param name, just to add a factory method that makes the API easier:

```ruby
UserSubscription.make!(source: article, subscriber: subscriber)
```

and/or 

```ruby
article.create_email_subscription!(subscriber)
```

and inside you can assign the various params to the correct fields.

> In this case, it's not clear, are those email_subscriptions ones the author (which is just a variable name holding a User object, could be called user or anything else) is the Author on? Or the Subscriber? .email_subscriptions has 2 relationships to a User - Author and Subscriber.

I would imagine `x.email_subscriptions` to return the list of things `x` is subscribed to. 

> Are there some more areas of the naming that are confusing? Maybe that will help me get some ideas on how to make it more clear.

hahaha I guess I can't still wrap my head some of these concepts but I'm sure they'll be clearer with usage!

atsmith813:
@rhymes ahhhh, I see. I like that idea a lot - I'll explore that!

> I would imagine x.email_subscriptions to return the list of things x is subscribed to.

I go back and forth 🙈 . That makes sense to me in every context until you're talking about a `User` who's an `Author`. I could see changing or aliasing `subscribed_to_user_subscriptions` to `user_subscriptions`.

I'll think more on that while I explore the factory method and gather more feedback 💯 .

atsmith813:
@rhymes I added some features, thanks again for the idea! I love how this looks now.

```ruby
subscriber = User.last
article = Article.first

UserSubscription.make(source: article, subscriber: subscriber)
UserSubscription.make_new(source: article, subscriber: subscriber)

article.create_user_subscription(subscriber)
article.new_user_subscription(subscriber)
```

I added the latter two to a model concern and moved the associations in there as well. Adding this to comments in the future will just be:
```ruby
include UserSubscriptionSourceable
```

cc: @citizen428 since this is a notable change, I'm re-requesting review 🙃

app/models/concerns/user_subscription_sourceable.rb
```diff
@@ -0,0 +1,33 @@
...
+  def create_user_subscription(subscriber)
+    UserSubscription.create(user_subscription_attributes(subscriber))
+  end
+
+  private
+
+  def user_subscription_attributes(subscriber)
+    {
+      user_subscription_sourceable: self,
+      author_id: self&.user_id,
```
citizen428, line 28: You shouldn't need the `self` here.

app/models/user_subscription.rb
```diff
@@ -4,12 +4,30 @@
...
   belongs_to :user_subscription_sourceable, polymorphic: true
 
   validates :author_id, presence: true
-  validates :subscriber_id, presence: true, uniqueness: { scope: %i[user_subscription_sourceable_type user_subscription_sourceable_id] }
+  validates :subscriber_email, presence: true
+  validates :subscriber_id, presence: true, uniqueness: { scope: %i[subscriber_email user_subscription_sourceable_type user_subscription_sourceable_id] }
   validates :user_subscription_sourceable_id, presence: true
   validates :user_subscription_sourceable_type, presence: true, inclusion: { in: ALLOWED_TYPES }
+
+  def self.make_new(source:, subscriber:)
```
citizen428, line 17: Could it make sense to follow e.g. `FactoryBot` naming and use `build` here? `make_new` is less intent revealing, especially when next to `make`.

app/models/user_subscription.rb
```diff
@@ -4,12 +4,30 @@
...
-  belongs_to :author, class_name: "User", inverse_of: :user_subscriptions
-  belongs_to :subscriber, class_name: "User", inverse_of: :user_subscriptions
+  belongs_to :author, class_name: "User", inverse_of: :source_authored_user_subscriptions
+  belongs_to :subscriber, class_name: "User", inverse_of: :subscribed_to_user_subscriptions
   belongs_to :user_subscription_sourceable, polymorphic: true
 
   validates :author_id, presence: true
-  validates :subscriber_id, presence: true, uniqueness: { scope: %i[user_subscription_sourceable_type user_subscription_sourceable_id] }
+  validates :subscriber_email, presence: true
+  validates :subscriber_id, presence: true, uniqueness: { scope: %i[subscriber_email user_subscription_sourceable_type user_subscription_sourceable_id] }
```
vaidehijoshi, line 13: Shall we just pull out the this array into a constant and freeze it (similar to `ALLOWED_TYPES`)?

db/migrate/20200617014509_add_subscriber_email_index_to_user_subscriptions.rb
```diff
@@ -0,0 +1,45 @@
...
+      remove_index :user_subscriptions, column: :subscriber_email, algorithm: :concurrently
+    end
+
+    if index_exists?(:user_subscriptions, %i[subscriber_id subscriber_email user_subscription_sourceable_type user_subscription_sourceable_id])
+      remove_index :user_subscriptions,
+                   column: %i[subscriber_id subscriber_email user_subscription_sourceable_type user_subscription_sourceable_id],
+                   algorithm: :concurrently
+    end
+  end
+end
```
vaidehijoshi, line 45: I might be missing something completely obvious here, but why do we need to explicitly check whether or not each index exists before removing it (or even adding it)? I would imagine that we would know when they would exist based on whether the `up` or `down` method are being run; is this for idempotency's sake?

spec/models/article_spec.rb
```diff
@@ -11,6 +11,7 @@ def build_and_validate_article(*args)
   let!(:article) { create(:article, user: user) }
 
   include_examples "#sync_reactions_count", :article
+  it_behaves_like "UserSubscriptionSourceable"
```
vaidehijoshi, line 14: just lovely! shared examples are 🔥

db/migrate/20200617014509_add_subscriber_email_index_to_user_subscriptions.rb
atsmith813: Just for idempotency 💯 

We've had a couple of issues in production in the past. A new PR with a migration goes out, the migration run, but the deploy fails for some reason. The PR is rolled back, a fix is pushed, but then it tries to run the same migration again and the index exists already, throwing an error, causing another problem.

We should probably document this somewhere now that I think about it...but the "new rule of thumb" has become to make the migrations idempotent to avoid that headache.

vaidehijoshi: Ah okay, makes total sense :)

I basically asked the question and then answered it for myself, haha

app/models/user_subscription.rb
atsmith813: This is a little weird...I tried it with:
```ruby
SUBSCRIBER_ID_UNIQUE_SCOPE = %i[
  subscriber_email
  user_subscription_sourceable_type
  user_subscription_sourceable_id
].freeze

validates :subscriber_id, presence: true, uniqueness: { scope: SUBSCRIBER_ID_UNIQUE_SCOPE }
```

And Rubocop complains with:
```
Rails/UniqueValidationWithoutIndex: Uniqueness validation should be with a unique index.
```

I ensured my spelling, changed indentations, etc.

🤔   the index is there, but it only complains when I use a constant or variable. I didn't find anything as to _why_ this is. And I don't want to disable the rule around this line because it's a performance check. I'm going to leave it as is and I've made a note to circle back some time to learn more about why.

rhymes: Rubocop uses static analysis for that cop, so the names need to be inlined, thus I would leave it as it is

PR #8730, opened 2020-06-17
Only render div if the user is able to connect another auth

Zhao-Andy:
- [x] Bug Fix

## Description
Fixes a small bug where a blank `<div>` is rendered for users with more than 1 authentication identity.

# Before:
Screen Shot 2020-06-17 at 11 11 24 AM

# After:

Screen Shot 2020-06-17 at 11 08 52 AM


app/views/users/_profile.html.erb
```diff
@@ -1,8 +1,10 @@
 <%= javascript_packs_with_chunks_tag "colorPreview", "validateFileInputs", defer: true %>
 
-<div class="crayons-card mb-6 p-5">
-  <%= render "users/additional_authentication" %>
-</div>
+<% if @user.identities.size < 2 %>
```
rhymes, line 3: didn't know we added this too!

can you use `if user.identities.count < SiteConfig.authentication_providers.count` ? I shall add a custom method soon

Zhao-Andy: Ohh yeah that's great 👍

Zhao-Andy: I'm going to stick with `@user.identities.size` or `.count` unless you think otherwise?

rhymes: no that's fine!

rhymes: BTW in `/account` there's the same problem so it's probably easier to add it to the partial itself

Zhao-Andy:
Fixed for `/account` too, thanks @rhymes!

# Before:

Screen Shot 2020-06-17 at 11 31 56 AM

# After:

Screen Shot 2020-06-17 at 11 32 02 AM

PR #8744, opened 2020-06-17
Mod Center MVP 🎉

msarit:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This is the first iteration of our redesigned Mod Center!

Fire your local env up, then visit `/mod` as a trusted user.

Demo Video: https://streamable.com/fvl6g4

## Added tests?

- [x] yes _(almost done)_
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [X] no documentation needed

Anime girl swooning


app/assets/stylesheets/moderators.scss
```diff
@@ -2,163 +2,104 @@
...
-        border-radius: 100px;
-        margin-right: 5px;
-        display: inline-block;
-        vertical-align: -0.3em;
-      }
-      a {
-        font-weight: 500;
-        color: var(--card-color-tertiary);
+  }
+  main.mod-index-list {
```
nickytonline, line 27: The `main` element selector shouldn't be necessary.

app/assets/stylesheets/moderators.scss
```diff
@@ -2,163 +2,104 @@
...
+    border: 1px solid var(--base-10);
+    margin: 0;
+    width: 100%;
+    .mod-index-list-header {
+      display: grid;
+      grid-template-columns: 4fr 2fr 1fr;
```
nickytonline, line 33: Is this good for smaller mobile as well, or is the mod center currently only being considered for Desktop, laptops, tablets?

app/javascript/modCenter/modCenter.js
```diff
@@ -0,0 +1,18 @@
+export function highlightCurrentTag() {
```
nickytonline, line 1: Add JSDocumentation to functions that are exported from ES Modules.

app/javascript/modCenter/modCenter.js
```diff
@@ -0,0 +1,18 @@
...
+  const tagView = document
+    .querySelector('.tag-view')
+    .textContent.replace('#', '')
+    .trim();
+  const inboxTags = Array.from(document.querySelectorAll('.inbox-tags'));
```
nickytonline, line 6: No need to change, but just letting you know you can also do this.

```javascript
  const inboxTags = [...document.querySelectorAll('.inbox-tags')];
```

app/javascript/modCenter/moderationArticles.jsx
```diff
@@ -0,0 +1,46 @@
+import { h, Component } from 'preact';
+// import { request } from "../utilities/http"
```
nickytonline, line 2: Remove commented code.

app/javascript/modCenter/moderationArticles.jsx
```diff
@@ -0,0 +1,46 @@
...
+import SingleArticle from './singleArticle';
+
+export class ModerationArticles extends Component {
+  state = {
+    articles: [],
+  };
+
+  componentWillMount() {
+    const container = document.getElementById('mod-index-list');
```
nickytonline, line 11: Does the dataset of articles on this element exist before this component is rendered? If so the `componentWillMount()` method can be removed and all the articles can be part of the initia state in the constructor.

app/javascript/modCenter/moderationArticles.jsx
```diff
@@ -0,0 +1,46 @@
...
+          const {
+            id,
+            title,
+            path,
+            cached_tag_list: cachedTagList,
+            published_at: publishedAt,
+            user,
+          } = article;
+          return (
+            <SingleArticle
```
nickytonline, line 33: Mapped items need a `key` prop. See [Lists and Keys](https://reactjs.org/docs/lists-and-keys.html) from the official React documentation.

app/javascript/modCenter/singleArticle/index.jsx
```diff
@@ -0,0 +1,76 @@
...
+      document.getElementById(
+        `article-iframe-${id}`,
+      ).innerHTML = `<iframe class="article-iframe" src="${path}"></iframe><iframe class="actions-panel-iframe" src="${path}/actions_panel"></iframe>`;
+    }
+  };
+
+  render() {
+    const { id, title, publishedAt, cachedTagList, user } = this.props;
+    const tags = cachedTagList.split(', ').map((tag) => {
+      return (
```
nickytonline, line 33: Same here. Provide a `key` prop on mapped over items.

benhalpern:
Coming together really nicely

app/views/moderations/_mod_sidebar_left.html.erb
```diff
@@ -0,0 +1,51 @@
+<div class="crayons-layout__sidebar-left">
+  <p class="fw-heavy fs-3xl mod-center-heading">Mod Center</p>
+  <div class="fw-bold fs-s mod-center-section-title">
+    INBOX
```
nickytonline, line 4: Very minor, but I've noticed a lot of our text labeling is steering away from all caps to just the first letter being capitalized.

app/assets/stylesheets/moderators.scss
Zhao-Andy: Pretty much only for desktop. Based on user research they strictly use it for desktop.

For this cycle we decided that mobile view will be a nice-to-have.

app/javascript/modCenter/moderationArticles.jsx
Zhao-Andy: Ohh interesting, yes it exists before the component renders. Will update.

nickytonline: That's good. So load it up in the initial state in the constructor and you'll avoid an unnecessary rerender.

nickytonline: To be clear rerenders happen, but in this case, it would definitely change the DOM, so might as well do it all at once.

app/javascript/modCenter/singleArticle/singleArticle.test.jsx
```diff
@@ -0,0 +1,73 @@
...
+    expect(text).toContain(article.user.name);
+  });
+  it('renders the hand wave emoji if the author has less than 3 articles ', () => {
+    const { container } = renderSingleArticle();
+    const text = getNodeText(container.querySelector('.article-author'));
+    expect(text).toContain('👋');
+  });
+  it('renders the a formatted published date', () => {
+    const { getByText } = renderSingleArticle();
+    getByText('Jun 28');
```
msarit, line 61: Is it possible to add a test for when the article was published that same day? To test the `HH:MM AM/PM` format?

Zhao-Andy: Ooh good point 👍

msarit: 😄

app/javascript/packs/modCenter.jsx
```diff
@@ -5,9 +5,16 @@ let elementLoaded = false;
...
   const root = document.getElementById('mod-index-list');
+  const isMobileDevice = typeof window.orientation !== 'undefined';
+
   if (root) {
     render(<ModerationArticles />, root);
   }
+
+  if (isMobileDevice) {
+    // eslint-disable-next-line no-alert
+    alert('The Mod Center is best viewed on desktop');
```
nickytonline, line 16: Consider using the snackbar or something that is not an `alert`. cc: @ludwiczakpawel

msarit: great point Nick, thanks. Is there a way to have the snackbar NOT fade out until `Dismiss` is clicked/tapped?

app/javascript/modCenter/singleArticle/index.jsx
```diff
@@ -0,0 +1,82 @@
...
+        <span className="mod-article-tag fs-s ff-accent lh-base" key={key}>
+          <span className="article-hash-tag">#</span>
+          {tag}
+        </span>
+      );
+    });
+
+    const newAuthorNotification = user.articles_count <= 3 ? '👋 ' : '';
```
rhymes, line 42: aahha top

app/javascript/modCenter/singleArticle/util.js
```diff
@@ -0,0 +1,35 @@
+const months = [
```
rhymes, line 1: I don't think this should be a blocker but I don't think we need any of these functions. 

JavaScript has a bunch of methods to format time. We should also use the `<time>` element to make time "semantic".

You can check https://github.com/thepracticaldev/dev.to/blob/master/app/assets/javascripts/utilities/localDateTime.js for inspiration (and add to those if you want) and this is how the `<time>` element works: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time

This is also an example of a live `<time>` element in the site: https://github.com/thepracticaldev/dev.to/blob/a9c18b4359c5aa242cadb59d0b5ffa301f71f53a/app/views/comments/_comment_date.erb#L3-L5

app/javascript/packs/modCenter.jsx
rhymes: +1 on the snackbar

app/views/moderations/_mod_sidebar_right.html.erb
```diff
@@ -0,0 +1,10 @@
+<div id="sidebar-wrapper-right" class="sidebar-wrapper sidebar-wrapper-right crayons-layout__sidebar-right">
+  <div id="quick-stat-card" class="crayons-card">
+    <div class="crayons-card__body mod-saying-hello">
+      <p class="fw-medium fs-2xl">Hello! 👋</p>
+      <p>
+        Thank you for helping to keep DEV safe! ❤️
```
rhymes, line 6: we should replace `DEV` with `ApplicationConfig["COMMUNITY_NAME"]`

app/javascript/modCenter/singleArticle/util.js
msarit: Thanks Rhymes! I think we can incorporate your suggestions in this PR 👍

app/views/moderations/_mod_sidebar_right.html.erb
msarit: Will do - thanks!

app/javascript/packs/modCenter.jsx
rhymes: @msarit that would be great, by reading the code it seems like the only way to hack that currently is to have a super long lifespan

nickytonline: If it's something you want to persist for longer than 3 seconds, maybe something else would make more sense like a [notice](https://storybook.dev.to/?path=/story/3-components-notices-html--warning).

Zhao-Andy: Added snackbar, with a `lifespan` set to 0. That seemed to make it so it doesn't disappear on its own.

app/javascript/modCenter/singleArticle/util.js
msarit: This has been address @rhymes thanks!

app/views/moderations/_mod_sidebar_right.html.erb
msarit: done! @rhymes

PR #8756, opened 2020-06-17
Fix article preamble style

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
The article _preamble_ is not actually part of the title but the latest change made it lose its styling.

This helps place it distinctively separate from the title, like such:

Screen Shot 2020-06-17 at 6 19 48 PM

Done with utility classes rather than a hack. Definitely a win 😄


ludwiczakpawel:
<3

Also, how do you add preamble to the post? I did not know this even exist 👀

PR #8758, opened 2020-06-18
Add ahoy + welcome notification tracking

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

We deployed welcome notifications last cycle, but haven't been able to measure or quantify how effective they are, and which ones are performing better than others.

This PR adds tracking around welcome notifications by making use of the `ahoy` [gem](https://github.com/ankane/ahoy) and [javascript library](https://github.com/ankane/ahoy.js).

This PR does a few things:
- [x] Adds two dependencies for `ahoy` (rails + JS), and their corresponding models, migrations, initializers, and database tables.
- [x] Imports the JS dependency into `base.js.erb` so that it can be used anywhere (although currently, we're only using it in the welcome notifications view).
- [x] Unique `id`s for each `broadcast` notification that is rendered on `/notifications`
- [x] A `trackNotification()` function evaluated in a `<script>` tag, which will track clicks for welcome notifications.

Closes https://github.com/thepracticaldev/dev.to/issues/8099.

1. First, add or update a `Welcome` broadcast to include a tracking function:
Screen_Shot_2020-06-17_at_5_51_26_PM

Your broadcast should look something like this:
Screen Shot 2020-06-17 at 4 09 19 PM

2. In your terminal, open up `rails console`. You'll want to find your user, and delete any notifications you already have for the broadcast that you just modified:
```bash
$ rails console
> user = User.where(email: 'put-your-email@here.com')
> user.notifications
> # Delete the preexisting notification here, for example something like this:
> user.notifications.last.destroy
> # Check that you don't have any events already tracked:
> Ahoy::Event.count
> 0
```
3. Manually send yourself a notification from the console:
```bash
# inside of the rails console
> Broadcasts::WelcomeNotification::Generator.call(user.id)
> nil
```
You should see something like this happen in your server logs:
```bash
# bin/startup
> 18:10:41 sidekiq.1   | 2020-06-18T01:10:41.189Z pid=10759 tid=21af class=Notifications::WelcomeNotificationWorker jid=366f055672bee68e2a677af4 INFO: start
> 18:10:41 sidekiq.1   | 2020-06-18T01:10:41.638Z pid=10759 tid=21af class=Notifications::WelcomeNotificationWorker jid=366f055672bee68e2a677af4 elapsed=0.451 INFO: done
```
4. Go to your `/notifications` tab. You should see the click handler appear, as well as an unique `id` on the broadcast element. Click on the link that you jut added tracking to!
Screen_Shot_2020-06-17_at_6_15_27_PM

5. The click should take you to the page that you expect, and when you go back into your `rails console`, you should see a new `Ahoy::Event` that records the action you just took:

```bash
# inside of the rails console
> Ahoy::Event.count
> 1
> Ahoy::Event.last
> #<Ahoy::Event:0x00007fdadde4bc30
 id: 15,
 name: "Clicked Welcome Notification",
 properties: {"title"=>"welcome_notification_welcome_thread"},
 time: Thu, 18 Jun 2020 01:20:10 UTC +00:00,
 user_id: 11,
 visit_id: 6>
```

_Hooray, you just tracked the click on a welcome notification! 🎉_

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [x] ~readme~ **added inline docs**
- [ ] no documentation needed

Once we have shipped this PR, we'll need to add the `trackNotification(e)` to the broadcasts in the welcome notification flow that we want to track; without that, we won't actually be tracking them :)


vaidehijoshi:
I want to add an additional note here that this is a v1 pass on this feature. I realize it's not perfect because it's not ideal for _admins_ to be adding something manually like `trackNotification(event)`! But, at the same time, this will get the ball rolling on adding tracking to something that badly needs it, and IMHO, this is a good first step.

I also want to add that we have https://github.com/thepracticaldev/dev.to/issues/6990 as the next ticket on our plate in terms of the broadcast/notification work we've been doing. Something that @atsmith813 and I have been discussing on Slack is the fact that we're toeing the line between being fully customizable (aka our fully-customizable HTML in our broadcasts feature), and something that is more structured, but with some configuration.

I think the next step (v2) of this feature should actually move away from admins adding an "event handler" like we're doing here. Instead, IMHO, v2 should do something like:
* Continue to allow an admin to add customizable HTML.
* Allow an admin to add/remove tracking from any notifications on a broadcast in the UI of `/internal/broadcasts`, probably by a checkbox (?).
* Parse through a notification and add the `trackNotification(event)` click handler programmatically before save/update. 
* Specify on the admin internal UI that checking the checkbox in the UI to add tracking will only work if they include an `<a>` or `<button>` tag in their HTML.

Definitely open to other ideas here, but just wanted to share what I've been thinking about while building this out. The TL;DR here is that we should think more deeply about how we'll make this extensible to forem users, but for the specific use case of adding tracking to THIS feature for _us_, I think this is the right/best first step.

rhymes:
> Parse through a notification and add the trackNotification(event) click handler programmatically before save/update.

The first thing that popped up in my mind for future iterations is to add a special `js-` class to notifications that need tracking and then use JS to listen to click events with a global handler and deliver the goods to the server

Not sure if it's feasible, just an idea

db/migrate/20200617183058_create_ahoy_visits_and_events.rb
```diff
@@ -0,0 +1,61 @@
...
+      t.string :visitor_token
+
+      # the rest are recommended but optional
+      # simply remove any you don't want
+
+      # user
+      t.references :user
+
+      # standard
+      t.string :ip
```
rhymes, line 14: Privacy related question: should we track IPs? We do scramble IPs when we send them to Google Analytics. 

https://github.com/thepracticaldev/dev.to/blob/478b048fc5aa0c5607e8f3491e68811a46f044bf/app/controllers/ga_events_controller.rb#L31-L34

Can we do without them altogether for in house tracking? Or at least encrypt them?

A visit is tied to the user model, which means we're going to have multiple DBs (in each community) that associate user data with their IP.

db/migrate/20200617183058_create_ahoy_visits_and_events.rb
```diff
@@ -0,0 +1,61 @@
...
+      t.string :browser
+      t.string :os
+      t.string :device_type
+
+      # location
+      t.string :country
+      t.string :region
+      t.string :city
+      t.float :latitude
+      t.float :longitude
```
rhymes, line 30: Me again on privacy related data. Do we need to track the location down to the coordinates? Right now we tell the users "fill the location field with whatever you want" in https://dev.to/settings/ and I think vagueness is by design. 
Even if we come up with some super clear messaging, users might not know that we're storing location data and might not want to or we should at least give them the option to opt out (I do block GA scripts on purpose for this very reason :D).

Doign a step back: do we need location data? If it's useful for statistics and reports, is country enough? if not, if region enough? if not, is city enough?

db/migrate/20200617183058_create_ahoy_visits_and_events.rb
citizen428: Also, the GDPR considers IP addresses personally identifiable information, so we'd have to remove them for European users on their request.

db/migrate/20200617183058_create_ahoy_visits_and_events.rb
```diff
@@ -0,0 +1,61 @@
+class CreateAhoyVisitsAndEvents < ActiveRecord::Migration[6.0]
+  def change
+    create_table :ahoy_visits do |t|
+      t.string :visit_token
+      t.string :visitor_token
+
+      # the rest are recommended but optional
```
citizen428, line 7: I think we should remove every column we can't make an extremely compelling case for. Many of them go way past tracking how successful a campaign is and would prefer us to always take the "as little as possible" approach for storing user data.

db/migrate/20200617183058_create_ahoy_visits_and_events.rb
benhalpern: Yeah. We do not currently track or persist IP for any users for max simplicity and compatibility with GDPR, so let's keep it that way.

db/migrate/20200617183058_create_ahoy_visits_and_events.rb
benhalpern: I think country and/or region is probably fine as far as what is considered PII.

But what if we let signed-in opt in or out of this granularity of tracking? I could see a forem in the future having location-based features which _could_ benefit from location awareness.

Maybe:

Country/region -> Opt in by default (not PII, generally could be helpful for top-level metrics)
Everything more granular -> Opt out by default (maybe just not built in at all until needed/requested, but with this in mind when needed)

City is probably also technically fine but I don't think we have a real need for that and should probably not store it.

Also of note: We pass all our traffic through a Fastly shield so we actually need to reach for `request.env["HTTP_FASTLY_CLIENT_IP"]` instead of the default which might through all of this.

Of course future Forems will not have the same Fastly config so we'll probably need an abstraction for getting this right.

db/migrate/20200617183058_create_ahoy_visits_and_events.rb
vaidehijoshi: @rhymes @citizen428 @benhalpern Agree with you on all accounts re: privacy concerns. I specifically left all the columns in so we could decide on which we wanted to keep/remove 😉 (I should have left a note about that, sorry!).

I think it sounds like we should just probably remove _all_ of the columns that are not strictly required (basically what @citizen428 mentioned), and then incrementally and intentionally add back whatever we feel is necessary.

vaidehijoshi:
This is ready for re-review based on everyone's comments re: privacy. I went with the "least possible amount of information" approach, even given what @benhalpern [mentioned here](https://github.com/thepracticaldev/dev.to/pull/8758#discussion_r442274043), since it sounds like we need to think a bit more about how we'll expose some of these location tracking items, if we do decide to add them.

For now, I've trimmed this down to the most essential pieces for visits + events.

vaidehijoshi:
> > Parse through a notification and add the trackNotification(event) click handler programmatically before save/update.
> 
> The first thing that popped up in my mind for future iterations is to add a special `js-` class to notifications that need tracking and then use JS to listen to click events with a global handler and deliver the goods to the server
> 
> Not sure if it's feasible, just an idea

@rhymes I'll certainly look into this for v2! @atsmith813 had a similar idea, which might be a nice middle ground if we think it will work with forem extensibility.

db/migrate/20200617183058_create_ahoy_visits_and_events.rb
mstruve: fwiw  neither of those values in our case, are going to be the actual user IP, those will be Fastly IPs

app/models/ahoy/event.rb
```diff
@@ -0,0 +1,8 @@
+class Ahoy::Event < ApplicationRecord
```
mstruve, line 1: Since we do a lot of tracking and reporting in Datadog, down the road you might want to hook into this model on creation and send the stats to Datadog as well.

app/views/notifications/index.html.erb
```diff
@@ -17,6 +17,20 @@
...
+  /**
+   * A script that tracks clicks for welcome notifications.
+   * The notification must have `trackNotification` as its onclick event
+   * in order for this to be triggered.
+   *
+   * @param {Event} The click event on the notification.
+   */
+  function trackNotification(e) {
+    let title = e.target.parentElement.id
+    ahoy.track("Clicked Welcome Notification", { title });
```
mstruve, line 30: 🔥

db/schema.rb
```diff
@@ -37,6 +49,15 @@
     t.index ["user_id", "user_type"], name: "index_ahoy_messages_on_user_id_and_user_type"
   end
 
+  create_table "ahoy_visits", force: :cascade do |t|
+    t.datetime "started_at"
+    t.bigint "user_id"
+    t.string "visit_token"
```
mstruve, line 55: Are these tokens just something at Ahoy uses to track users for its own purposes?

spec/system/notifications/notifications_page_spec.rb
```diff
@@ -85,4 +85,58 @@ def interact_with_each_emojis
...
     end
   end
+
+  context "with welcome notifications" do
+    let(:mascot_account) { create(:user) }
+
+    before do
+      allow(Notification).to receive(:send_welcome_notification).and_call_original
+      allow(User).to receive(:mascot_account).and_return(mascot_account)
+      SiteConfig.staff_user_id = mascot_account.id
```
mstruve, line 95: You could stub this value so you dont have to worry about state before and after.

db/schema.rb
vaidehijoshi: @mstruve Good question! The ahoy docs don't really explain it at all (_le sigh 😞_) but as far as I can tell from looking at the source code, `visit_token` is a unique uuid identifier to create/identify/track visits. It is a required column, and the gem internally relies on its uniqueness.

spec/system/notifications/notifications_page_spec.rb
vaidehijoshi: Nice idea, done! ✅

app/models/ahoy/event.rb
vaidehijoshi: @mstruve What sorts of things do you think would be useful for Datadog reporting as opposed to just tracking? Trying to think through what/when we'd want to report (if there's something obvious, I can add it to the project board to tackle)

mstruve: I was thinking just basic counts to start
```ruby
DatadogStatsClient.count("welcome_notification_clicks", <number>, tags: ["title:<the title>"])
```
That would allow you to graph over time the number of clicks for each different type of welcome notification and compare them all to each other

PR #8761, opened 2020-06-18
Display 'Originally published' for canonical URLs

### LINKED ISSUE #8726: Display canonicals consistently on articles
@michael-tharrington commented on [Fri Nov 15 2019](https://github.com/thepracticaldev/tech-private/issues/314)

Currently, we allow folks two main ways of designating canonicals:
1. Publishing via RSS (importing an external blog in your dashboard)
2. Manually writing in the canonical on individual posts

When publishing via RSS, a note is generated beneath the title of the post that says:
"_Originally published at [insert-blog-name-and-link-to-canonical-URL]._"

However, when manually writing in the canonical on individual posts, there is no extra note generated beneath the title of the post.

Both methods should produce the same outcome, providing the noted attribution beneath the title. 

---

@michael-tharrington commented on [Wed Jun 17 2020](https://github.com/thepracticaldev/tech-private/issues/314#issuecomment-565489561)

Just making a note that there's a user, Dvir (https://dev.to/dejavo), who wants to be updated when this goes into place. Some of their articles wouldn't import via RSS, so they manually imported them, but they'd prefer for it to be uniform.

They wrote an issue on their RSS here - https://github.com/thepracticaldev/dev.to/issues/5086

citizen428:

- [X] Refactor

## Description

So far we only added "Originally published at" for articles imported from feeds, but not for posts that had a canonical URL set. This has confused several users in the past, including your truly.

Closes #8726

Screen Shot 2020-06-18 at 13 42 26

Note the missing "on <timestamp>" if we only have a canonical URL.

## Added tests?

- [X] yes

## Added to documentation?

- [X] no documentation needed

image

### DIFF
--- a/app/helpers/articles_helper.rb
+++ b/app/helpers/articles_helper.rb
@@ -41,11 +41,12 @@ module ArticlesHelper
   end
 
   def should_show_crossposted_on?(article)
-    article.crossposted_at &&
+    article.canonical_url ||
+      (article.crossposted_at &&
       article.published_from_feed &&
       article.published &&
       article.published_at &&
-      article.feed_source_url.present?
+      article.feed_source_url.present?)
   end
 
   def get_host_without_www(url)
--- a/app/views/articles/show.html.erb
+++ b/app/views/articles/show.html.erb
@@ -166,8 +166,8 @@
                 <em>
                   Originally published at
                   <a href="<%= @article.canonical_url || @article.feed_source_url %>" style="color:#1395b8"><%= get_host_without_www(@article.canonical_url || @article.feed_source_url) %></a>
-                  on
                   <% if @article.crossposted_at %>
+                    on
                     <time datetime="<%= (@article.originally_published_at || @article.published_at)&.utc&.iso8601 %>"><%= (@article.originally_published_at || @article.published_at)&.strftime("%b %d, %Y") %></time>
                   <% end %>
                 </em>
--- a/spec/views/articles_spec.rb
+++ b/spec/views/articles_spec.rb
@@ -47,4 +47,11 @@ RSpec.describe "articles/show", type: :view do
     expect(rendered).to have_css("form#new_comment")
     expect(rendered).to have_css("input#submit-button")
   end
+
+  it "shows a note about the canonical URL" do
+    allow(article1).to receive(:canonical_url).and_return("https://example.com/lamas")
+    render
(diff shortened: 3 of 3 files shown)


app/helpers/articles_helper.rb
```diff
@@ -41,11 +41,12 @@ def should_show_updated_on?(article)
   end
 
   def should_show_crossposted_on?(article)
-    article.crossposted_at &&
+    article.canonical_url ||
```
citizen428, line 44: I didn't see a reason about making this more complicated than a simple "if we have a canonical URL, show it". The logic for RSS feeds can probably be simplified too (e.g. if the article isn't published, the user may still want to see the "Originally published at" in their draft), but that's for another day.

app/helpers/articles_helper.rb
```diff
@@ -41,11 +41,12 @@ def should_show_updated_on?(article)
...
+      (article.crossposted_at &&
```
citizen428, line 45: The parens aren't necessary for boolean logic, but without them Rubocop wants me to indent all the following lines by 2 more spaces. You can't always let the machines win.

app/views/articles/show.html.erb
```diff
@@ -166,8 +166,8 @@
                 <em>
                   Originally published at
                   <a href="<%= @article.canonical_url || @article.feed_source_url %>" style="color:#1395b8"><%= get_host_without_www(@article.canonical_url || @article.feed_source_url) %></a>
-                  on
                   <% if @article.crossposted_at %>
+                    on
```
citizen428, line 170: Just moved the "on" into the conditional, since otherwise we'd end up "Originally published at example.com on" without a time stamp.

spec/views/articles_spec.rb
```diff
@@ -47,4 +47,11 @@ def create_comment(parent_id = nil)
     expect(rendered).to have_css("form#new_comment")
     expect(rendered).to have_css("input#submit-button")
   end
+
+  it "shows a note about the canonical URL" do
+    allow(article1).to receive(:canonical_url).and_return("https://example.com/lamas")
+    render
+    expect(rendered).to have_text("Originally published at")
```
citizen428, line 54: `rendered` is full of newlines and whitespace, so I just match twice. Alternatively we can match on `/\AOriginally published at\s+example.com\z/`, LMK which one you prefer.

Issue #8248, opened 2020-06-03
Improve developer experience by allowing code errors surface


ledestin:
**Is your feature request related to a problem? Please describe.**

While working on request specs of comments controller, I found out that exceptions wouldn't surface, when running request specs. The culprit turned out to be `resque StandardError`:
https://github.com/thepracticaldev/dev.to/blob/master/app/controllers/comments_controller.rb#L114

The problem here is two-fold:
1. To get to the error, I need to analyse response payload, and I don't get a backtrace. Given that the project is set up to surface exceptions via `config.action_dispatch.show_exceptions = false`, this practice breaks the default behaviour, which worsens developer experience.
2. `NameError` and `NoMethodError` are subclasses of `StandardError`:
```
ruby -e 'wtf'
Traceback (most recent call last):
-e:1:in `<main>': undefined local variable or method `wtf' for main:Object (NameError)
```
Thus, I can write nonsense, and I will not know, unless I take care to examine response payload. I hit this problem while fixing a bug, and I lost time on it.

`rescue StandardError` is all over the place.

**Describe the solution you'd like**

I'd like programming errors surface as exceptions with a proper backtrace, in test environment. I suggest rescuing `RuntimeError`, instead of `StandardError`.

**Describe alternatives you've considered**

For request specs, it's examining payload, but for models, there's no solution like that.

Ridhwana:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

rhymes:
I think this request also sits in a larger conversation about using custom exceptions as much as possible instead of `StandardError` which can have the byproduct of silencing meaningful ones.

I'm not sure what the perfect approach is but it's worth a discussion

ledestin:
What was the original intent of rescuing `StandardError` and not `RuntimeError`?

rhymes:
@ledestin hard to say, If I had to guess, it's because all liquid tags (that can be rendered in comments) raise StandardError (which we plan to fix eventually) and not a custom exception.

[...]

PR #8765, opened 2020-06-18
StandardError to ApplicationError

ledestin:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Try replacing `raise StandardError` with `raise ApplicationError` in a single class and rescue `ApplicationError` in controllers alongside with `StandardError`.

#8248

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


ledestin:
@rhymes take a look at a possible gradual approach.

joshpuetz:
Tagging @rhymes and the rest of @thepracticaldev/oss

ledestin:
My plan was to introduce `ApplicationError`, so that it can be used later on, in liquid tags and elsewhere. At this point, it should be safe to replace `raise StandardError` in liquid tags with `raise ApplicationError`, since controllers catch it.

As @rhymes suggested, the issues, when liquid tags still raise `StandardError` due to code errors, should be resolved one by one.

joshpuetz:
Re-pinging @forem/oss

rhymes:
@joshpuetz thanks! we're having a sort of double conversation between the PR and the issue https://github.com/forem/forem/issues/8248#issuecomment-673404569

It's a bit confusing!

Basically I still think we should go the custom route of having `LiquidTagError < ApplicationError < StandardError` so the controller can distinguish from errors thrown by liquid tags and everything else

rhymes:
Hi @ledestin, I'm sorry we dragged our decision on this but we're not sure we'd like to proceed as is and this PR has sat here for months due to such indecision. I still think you have a point there's valuable feedback in #8248 but we haven't had the chance of moving forward with this and we'd like to table this PR until we can make a team decision on the issue that this refers to.

So, in the meantime, I'm going to close this.

Thank you again!

PR #8766, opened 2020-06-18
Rails 6: enable zeitwerk autoloader

rhymes:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

[zeitwerk](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html) is the new Rails 6 thread safe autoloader. It changes a few things:

- it uses `autoload` instead of patching `const_missing` to load things automatically
- it autoloads all descendant folders in `/app`, thus there's no need to list them explicitly
- it's a bit stricter on the naming of things

- [Understanding Zeitwerk in Rails 6](https://medium.com/cedarcode/understanding-zeitwerk-in-rails-6-f168a9f09a1f)
- [Rails 6 guides: Autoloading and Reloading Constants (Zeitwerk Mode)](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html)
- [zeitwerk gem](https://github.com/fxn/zeitwerk)

- boot the app
- click around :D

## Added tests?

- [ ] yes
- [ ] no, because I need help


app/services/authentication/providers.rb
```diff
@@ -1,10 +1,3 @@
-# We require all authentication modules to make sure providers
```
rhymes, line 1: moved to an an initializer

app/services/aws/fake_client.rb
```diff
@@ -1,8 +0,0 @@
-# Fake Aws client is used in non-production environments to prevent actual calls to AWS lambda
```
rhymes, line 1: mvoed to the initializer, this class shouldn't be here anyway

config/application.rb
```diff
@@ -22,22 +22,25 @@ class Application < Rails::Application
...
-    config.autoload_paths += Dir["#{config.root}/app/labor/"]
-    config.autoload_paths += Dir["#{config.root}/app/decorators/"]
-    config.autoload_paths += Dir["#{config.root}/app/services/"]
-    config.autoload_paths += Dir["#{config.root}/app/presenters/"]
-    config.autoload_paths += Dir["#{config.root}/app/liquid_tags/"]
-    config.autoload_paths += Dir["#{config.root}/app/black_box/"]
-    config.autoload_paths += Dir["#{config.root}/app/sanitizers"]
-    config.autoload_paths += Dir["#{config.root}/app/facades"]
-    config.autoload_paths += Dir["#{config.root}/app/errors"]
-    config.autoload_paths += Dir["#{config.root}/app/view_objects"]
```
rhymes, line 39: every folder under `/app` is autoloaded by default in zeitwerk mode

config/initializers/pusher.rb
```diff
@@ -1,4 +1,3 @@
-require "pusher"
```
rhymes, line 1: pusher is automatically loaded

config/initializers/reverse_markdown.rb
```diff
@@ -1,13 +0,0 @@
-# This eagerload our custom ReverseMarkdown::Converters and allows it to be autoreloaded
```
rhymes, line 1: all of this stuff is preloaded in zeitwerk mode

config/initializers/sidekiq.rb
```diff
@@ -1,22 +1,9 @@
...
-        def stringify_keys(other)
-          other.to_hash.transform_keys(&:to_s)
-        end
-      end
-    end
+Rails.application.reloader.to_prepare do
+  Dir.glob(Rails.root.join("lib/sidekiq/*.rb")).sort.each do |filename|
+    require_dependency filename
   end
 end
```
rhymes, line 5: this patch was included in rack, see https://github.com/rack/rack/pull/1428 - we don't need it anymore

config/application.rb
maestromac: it's so clean now ✨

config/initializers/pusher.rb
```diff
@@ -1,4 +1,3 @@
...
 require "pusher/push_notifications"
```
maestromac, line 1: What does this line? should it be removed?

config/initializers/zeitwerk.rb
```diff
@@ -0,0 +1,15 @@
...
+# see https://guides.rubyonrails.org/autoloading_and_reloading_constants.html
+
+# Configures default inflections for the zeitwerk autoloader
+# see <https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#customizing-inflections>
+Rails.autoloaders.each do |autoloader|
+  autoloader.inflector.inflect(
+    "html_rouge" => "HTMLRouge",
+    "url" => "URL",
+  )
+end
```
maestromac, line 11: this is super neat!

config/initializers/pusher.rb
rhymes: no, we can't because Pusher::Notifications comes from the pusher-notifications gem which is not autoloaded. I did try, stuff exploded, put it back :D

app/services/authentication/providers.rb
citizen428: It's great that we could remove this from here 🚀

config/application.rb
citizen428: Rails 5 already autoloaded all direct descendants of `/app`, so this configuration shouldn't have been necessary.

PR #8772, opened 2020-06-18
Enable webmentions

### LINKED ISSUE #6807: Enable Webmentions

My website can receive webmentions, and I keep syndicating my blog post to dev.to. If the reactions that I received on dev.to can be reflected on my website that would be great.

dev.to enable webmentions and pingback to my original blog post on my website whenever my post on dev.to received a reaction.

Not sure if there are any.

**Additional context**

eyewritecode:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR is based on a few assumptions that i thought would be enough to test webmentions.

Since there exists different ways to [discover webmentions](https://www.w3.org/TR/webmention/#sender-discovers-receiver-webmention-endpoint), this PR assumes that the `canonical_url` points to a URL that advertises its webmention endpoint with an HTML <link> tag in the document and that it ~~should be an absolute URL~~ (i added relative URL support)

Also, Since DEV will be the source of webmentions, these have to be in a certain format a.k.a `Microformats` that allows the receiving website to parse various information of your posts like the comment's content, the author of the comment's information(name, picture, etc..)

More info can be found here:
- [Sending your first webmention](https://aaronparecki.com/2018/06/30/11/your-first-webmention)
- [Microformats Wiki](http://microformats.org/wiki/h-entry)

Closes #6807 

- Create or edit a post and add `canonical_url` that supports Webmentions(i.e https://webmention.rocks/test/4)
- Write a comment on that post and it should be handled by the Webmention server(i.e shown on https://webmention.rocks/test/4)

FYI: For this to work locally, and be able to send post requests to the Webmention endpoint, I had to:
- Use`ngrok` and change `ApplicationConfig["APP_DOMAIN"]` to `ngrok` URL
- Configure the `development` environment to allow requests from the `ngrok` URL

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


app/models/article.rb
```diff
@@ -75,6 +75,7 @@ class Article < ApplicationRecord
   before_save :fetch_video_duration
   before_save :clean_data
   before_save :update_cached_user
+  before_save :check_webmentions_support
```
mstruve, line 78: Since this is being run in a Sidekiq job and we expect the new `@canonical_url` to be present in that job this needs to be done in an `after_commit` callback to ensure the database is updated before the worker is kicked off.

app/services/web_mentions/web_mention_handler.rb
```diff
@@ -0,0 +1,37 @@
...
+      @article_url = article_url
+      @webmention_url = ""
+    end
+
+    def self.call(*args)
+      new(*args).call
+    end
+
+    def call
+      accepts_webmention? ? send_webmention(@webmention_url) : Rails.logger.info("#{@canonical_url} doesn't support Webmentions")
```
mstruve, line 14: What is the goal fo all of the error logging? Are these failures we want someone to address?

app/workers/web_mentions/send_web_mention.rb
```diff
@@ -0,0 +1,17 @@
+module WebMentions
+  class SendWebMention
+    include Sidekiq::Worker
+    sidekiq_options queue: :low_priority, retry: 15, lock: :until_executing
+
+    def perform(comment_id)
+      comment = Comment.find_by(id: comment_id)
+      return unless comment
+
+      return unless comment.commentable.support_webmentions
```
mstruve, line 10: We do keep comments even when commentable's have been deleted so we need a safety operator here. 
```ruby
comment.commentable&.support_webmentions
```

app/services/web_mentions/web_mention_handler.rb
```diff
@@ -0,0 +1,37 @@
...
+    def accepts_webmention?
+      document = Nokogiri::HTML.parse(URI.open(@canonical_url)).at_css("link[rel='webmention']")
+
+      if document
+        @webmention_url = document["href"]
+        return true unless Addressable::URI.parse(@webmention_url).relative?
+      end
+      false
+    rescue StandardError => e
+      Rails.logger.error("WebmentionsException: #{e}")
```
mstruve, line 26: In the event an error is raised, I would expect this method to return false. The way it is written, if we receive an error the message will return the boolean of whether or not the log was created successfully, which in most cases will be true.

app/services/web_mentions/web_mention_handler.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def self.call(*args)
+      new(*args).call
+    end
+
+    def call
+      accepts_webmention? ? send_webmention(@webmention_url) : Rails.logger.info("#{@canonical_url} doesn't support Webmentions")
+    end
+
+    def accepts_webmention?
```
citizen428, line 17: I find this a bit misleading. The name suggests a boolean check, not something that actually sets `@webmention_url` as a side effect.

app/services/web_mentions/web_mention_handler.rb
```diff
@@ -0,0 +1,37 @@
...
+      end
+      false
+    rescue StandardError => e
+      Rails.logger.error("WebmentionsException: #{e}")
+    end
+
+    private
+
+    def send_webmention(webmention_url)
+      RestClient.post(webmention_url, "source": @article_url, "target": @canonical_url)
```
citizen428, line 32: Why are we passing in `webmention_url` but are accessing the other two as instance variables?

app/workers/web_mentions/send_web_mention.rb
citizen428: Yes, and since we are using it several times, it may be easier to do something like

```ruby
commentable = Comment.find_by(id: comment_id)&.commentable
return unless commentable
return uless commentable.support_webmentions
```

which also protects the other uses of `commentable` in this method.

app/workers/web_mentions/send_web_mention.rb
```diff
@@ -0,0 +1,17 @@
...
+
+      canonical_url = comment.commentable.canonical_url
+      article_url = ApplicationConfig["APP_DOMAIN"] + comment.commentable.path
```
citizen428, line 13: We have a URL helper module for this:

```ruby
URL.url(commentable.path)
```

db/migrate/20200614123306_add_support_webmentions_to_articles.rb
```diff
@@ -0,0 +1,5 @@
+class AddSupportWebmentionsToArticles < ActiveRecord::Migration[6.0]
+  def change
+    add_column :articles, :support_webmentions, :boolean, default: false
```
citizen428, line 3: I'm not saying this can go here, but right now the articles table has 76 columns (according to `article.attributes.keys.size`) and I think we should weigh carefully wether or not we add another one especially given that (a) it will conceivably be false for a majority of the articles (at least until webmentions gain wider support) and (b) is only needed in relatively specific circumstances (article creation and updates) where maybe we can live with an extra query here or there, especially if it's made by an async worker.

citizen428:
Thank you @eyewritecode, this is a great start for sure! :rocket:

> this PR assumes that the canonical_url points to a URL that advertises its webmention endpoint with an HTML tag in the document and that it should be an absolute URL

I think this is a reasonable assumption and how I handle web mentions on my blog.

app/views/articles/show.html.erb
```diff
@@ -165,7 +165,7 @@
               <% elsif should_show_crossposted_on?(@article) %>
                 <em>
                   Originally published at
-                  <a href="<%= @article.canonical_url || @article.feed_source_url %>" style="color:#1395b8"><%= get_host_without_www(@article.canonical_url || @article.feed_source_url) %></a>
+                  <a href="<%= @article.canonical_url || @article.feed_source_url %>" style="color:#1395b8" class="u-in-reply-to"><%= get_host_without_www(@article.canonical_url || @article.feed_source_url) %></a>
```
nickytonline, line 168: The `u-in-reply-to` does not appear to exist anywhere in the codebase or this PR except for here. I didn't realize this was required for webmentions, https://indieweb.org/reply . 

Same goes for the other `u-*`, `h-entry` etc. CSS class names. Consider mentioning this in the PR description for those unfamiliar with webmentions.

app/services/web_mentions/web_mention_handler.rb
eyewritecode: Thanks :),  i didn't know this

app/services/web_mentions/web_mention_handler.rb
eyewritecode: Good catch @citizen428, i'm looking at it right now and can't think of a reason why i did it like this 😬

app/services/web_mentions/web_mention_handler.rb
eyewritecode: @mstruve I added these for debugging purposes but i don't think there's more one can do when a URL doesn't support webmentions! 

Should i remove it?

app/services/web_mentions/web_mention_handler.rb
eyewritecode: Good point @citizen428, wouldn't this be better? 
```
def accepts_webmention?
  webmention_url.blank? ? false : true
end

def webmention_url
  # add webmention_url setting logic here
end
```

app/views/articles/show.html.erb
eyewritecode: Thanks @nickytonline, I'll update the PR description

db/migrate/20200614123306_add_support_webmentions_to_articles.rb
eyewritecode: @citizen428 I understand your concern and would like to explore a different approach if you could point me in a different direction! 

Is it possible to persist a virtual attribute until a worker has finished running 🤔

app/services/web_mentions/web_mention_handler.rb
citizen428: Instead of

```ruby
def accepts_webmention?
  webmention_url.blank? ? false : true
end
```

we can just do 

```ruby
def accepts_webmention?
  webmention_url.present?
end
```

The predicate method already returns a boolean and `present? == !blank?`

db/migrate/20200614123306_add_support_webmentions_to_articles.rb
citizen428: If it's reasonably shortlived, Redis seems like a good choice.

app/workers/web_mentions/send_web_mention.rb
```diff
@@ -0,0 +1,17 @@
+module WebMentions
+  class SendWebMention
+    include Sidekiq::Worker
+    sidekiq_options queue: :low_priority, retry: 15, lock: :until_executing
+
+    def perform(comment_id)
+      commentable = Comment.find_by(id: comment_id)&.commentable
+      return unless commentable
+
+      return unless commentable.support_webmentions
```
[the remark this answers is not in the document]
citizen428: That sounds good. Maybe then we could also remove ` WebMentions::CheckWebMentionSupport`, since that calls the same handler to set the DB column. If we do that, we probably also don't need the validation on `Article`, so the simplified flow would: in `after_commit` of a commit, enqueue this job (which we already do unconditionally anyway), then just rely on this and the handler to do the right thing:

```ruby
    def perform(comment_id)
      commentable = Comment.find_by(id: comment_id)&.commentable
      return unless commentable
      WebMentions::WebMentionHandler.call(commentable)
    end
```

This way `Article` itself doesn't need to know about web mentions at all and the full logic (early return when there's no canonical URL, building a full article URL with the `URL` module, etc) is in one place.

app/services/web_mentions/web_mention_handler.rb
```diff
@@ -0,0 +1,44 @@
+module WebMentions
+  class WebMentionHandler
+    def initialize(canonical_url, article_url = nil, webmention_url = nil)
```
rhymes, line 3: can we use keyword parameters? switching the two optional URLs is bound to happen :D

app/services/web_mentions/web_mention_handler.rb
```diff
@@ -0,0 +1,44 @@
...
+    def call
+      accepts_webmention? ? send_webmention : Rails.logger.info("#{@canonical_url} doesn't support Webmentions")
+    end
+
+    def accepts_webmention?
+      webmention_url.blank? ? false : true
+    end
+
+    def webmention_url
+      document = Nokogiri::HTML.parse(URI.open(@canonical_url)).at_css("link[rel='webmention']")
```
rhymes, line 22: Sorry, I pressed enter too early in my previous review. Can we use `HTTParty.get` instead of `URI.open`? That's our standard interface to the external world, mostly

app/services/web_mentions/web_mention_handler.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+    rescue StandardError => e
+      Rails.logger.error("WebmentionsException: #{e}")
+      nil
+    end
+
+    private
+
+    def send_webmention
+      RestClient.post(@webmention_url, "source": @article_url, "target": @canonical_url)
```
rhymes, line 39: can we switch to `HTTParty.post()` ?

I know we have many HTTP clients but we're trying to converge on actively use only one

app/workers/web_mentions/send_web_mention.rb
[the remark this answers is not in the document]
rhymes: Small suggestion if we decide to keep it: `supports_webmentions` instead of `support_webmentions`

db/migrate/20200614123306_add_support_webmentions_to_articles.rb
rhymes: You can use Redis. Add the attribute, then remove it at the end. It might not be needed as mentioned in the other reviews

rhymes:
Hi @eyewritecode, any chance you might want to take this to the finish line? Let us know if we can provide support :)

eyewritecode:
@rhymes Oh man 🙈 
Been busy with work and completely forgot about this one. Will definitely find time to finish it soon

citizen428:
Thanks @eyewritecode! Let us know if you need any help in resolving the conflicts, looking forward to this feature 😃

eyewritecode:
>I think it was already mentioned but I would mention these new u- and p- and h- CSS classes somewhere or at least write a regression test for them. Otherwise they are bound to disappear as there's no automatic way to know they are needed by webmentions for people unfamiliar with them

FYI @rhymes, I opted to create tests for those classes and added them in `spec/system/web_mentions`

eyewritecode:
@citizen428 Can you help resolve the codeclimate check?!

app/services/web_mentions/web_mention_handler.rb
mstruve: Question, do we want to silently log these errors or should we be letting them get raised so we can address them?

In my opinion, given this is being executed in a Sidekiq worker in the background I would lean towards raising the error. Since it is in the background it wont disrupt any user flow and will still notify us if anything is broken. Logs are what we use to fallback on for investigations so anything put in them is not going to be seen unless something else triggers us to look. Rather than having a user report mentions not working, I would rather be notified immediately by an error.

db/migrate/20201010084621_remove_support_webmentions_from_articles.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveSupportWebmentionsFromArticles < ActiveRecord::Migration[6.0]
+  def change
+    safety_assured { remove_column :articles, :support_webmentions, :boolean }
```
mstruve, line 3: Rather than adding a migration for removing this column, lets just remove all the migrations from the PR if we are not going to be doing anything with the database. That will help keep the codebase nice and clean.

db/schema.rb
```diff
@@ -141,6 +140,7 @@
     t.text "slug"
     t.string "social_image"
     t.integer "spaminess_rating", default: 0
+    t.integer "third_user_id"
```
mstruve, line 143: Looks like a rogue change made it into your schema 😉

mstruve (commented):
Cannot wait for this feature. Made a few suggestions!

app/services/web_mentions/web_mention_handler.rb
```diff
@@ -0,0 +1,41 @@
...
+    rescue StandardError => e
+      raise e, "Couldn't find the webmention URL of: #{@canonical_url}"
+    end
+
+    private
+
+    def send_webmention
+      HTTParty.post(@webmention_url, "source": @article_url, "target": @canonical_url)
+    rescue HTTParty::Error => e
+      raise e, "Couldn't send webmentions to: #{@webmention_url}"
```
mstruve, line 38: I am not very familiar with sending webmentions. What kinds of errors might we see here? Are they likely going to be errors that are worth retrying or not?

app/models/comment.rb
```diff
@@ -314,4 +315,8 @@ def notify_slack_channel_about_warned_users
   def parent_exists?
     parent_id && Comment.exists?(id: parent_id)
   end
+
+  def send_web_mentions
+    WebMentions::SendWebMention.perform_async(id)
```
mstruve, line 320: In DEV's database out of 196k published articles, only 47k have canonical URLs which makes me think enqueueing this job for every single comment might be a waste of resources. What do you think about checking for a commentable canonical URL before enqueueing this job to determine if we even need it?

spec/system/web_mentions/web_mention_microformats_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "rails_helper"
+
+# For webmentions to work, we basically need two things:
+# A source of webmentions(i.e: article on FOREM)
+# A target to which webmentions will be sent (i.e article's canonical URL)
+# Since FOREM will be the source of webmentions, every article should contain Microformats
+# that allows the receiving website to parse various information from the source article
+# such as the comment's content, the author of the comment and their information(name, picture, etc..)
+# Ref: http://microformats.org/wiki/h-entry
```
mstruve, line 9: I am not a huge code comment fan but this was very insightful and informative! Thank you!

app/models/article.rb
```diff
@@ -1,4 +1,8 @@
 class Article < ApplicationRecord
+  self.ignored_columns = %w[
+    support_webmentions
```
mstruve, line 3: I'm assuming the plan is to remove this so we don't have to keep it in sync and instead check the canonical URL on the fly, is that correct?

mstruve (commented):
I learned SO much from this PR! Thank you @eyewritecode! Left a couple comments and questions :)

app/services/web_mentions/web_mention_handler.rb
citizen428: The main error we could see is the endpoint getting removed or temporarily being unavailable.

rhymes (commented):
Looks cool! I left a note about the ignored field but overall it looks cool!

I agree with @mstruve that we should only enqueue meb mentioned to be sent for articles that have a canonical URL when the callbakc executes

app/services/web_mentions/web_mention_handler.rb
mstruve: > the endpoint getting removed

We should probably handle a 404(I'm assuming that's the error we would get) more intelligently so we are not constantly retrying if an endpoint is removed.

app/models/article.rb
```diff
@@ -1,4 +1,8 @@
...
+  ]
+
```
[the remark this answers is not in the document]
citizen428: It was actually part of this PR but then removed again, so I don't think we need to ignore the column as it wasn't part of production code yet and therefore isn't in the AR cache.

cmgorton:
Hi @eyewritecode 

PR #8773, opened 2020-06-18
Fixing listing message modal handler

Rafi993:

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

It fixes send button in listing message modal

Closes https://github.com/thepracticaldev/dev.to/issues/8762

1. Click on a listing
2. Type in message and click it send to send to send it to the message
3. Click on listing again
4. Type white space and click send the message won't be sent

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/app/javascript/listings/__tests__/Modal.test.jsx
+++ b/app/javascript/listings/__tests__/Modal.test.jsx
@@ -3,9 +3,7 @@ import { render } from '@testing-library/preact';
 import { axe } from 'jest-axe';
 import Modal from '../components/Modal';
 
-
 describe('<Modal />', () => {
-  const idFromMessageModal = 'listings-message-form';
   const getDefaultListing = () => ({
     id: 22,
     category: 'misc',
@@ -29,8 +27,8 @@ describe('<Modal />', () => {
     onAddTag: () => {
       return 'onAddTag';
     },
-    onChange: () => {
-      return 'onChange';
+    onChangeDraftingMessage: () => {
+      return 'onChangeDraftingMessage';
     },
     onClick: () => {
       return 'OnClick';
@@ -51,7 +49,7 @@ describe('<Modal />', () => {
     render(<Modal {...getProps()} listing={listing} />);
 
   it('should have no a11y violations', async () => {
-    const { container } = renderModal({...getDefaultListing()});
+    const { container } = renderModal({ ...getDefaultListing() });
     const results = await axe(container);
     expect(results).toHaveNoViolations();
   });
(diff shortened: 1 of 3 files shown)


app/javascript/listings/components/MessageModal.jsx
```diff
@@ -54,9 +57,12 @@ const MessageModal = ({
...
           </em>
         ) : (
-          <em data-testid="generic-message-about-interactions" id="generic-message-about-interactions">
+          <em
+            data-testid="generic-message-about-interactions"
+            id="generic-message-about-interactions"
+          >
             Message must be relevant and on-topic with the listing. All private
-            interactions
+            interactions
```
citizen428, line 65: Random observation, you seem to have some extra whitespace here 😃

Rafi993: It is some git hook that is adding these spaces. Should I skip it?

app/javascript/listings/components/MessageModal.jsx
```diff
@@ -80,7 +86,7 @@ MessageModal.propTypes = {
     user_id: PropTypes.number.isRequired,
   }).isRequired,
   onSubmit: PropTypes.func.isRequired,
-  onChangeDraftingMessage: PropTypes.func.isRequired,
+  onChange: PropTypes.func.isRequired,
```
Rafi993, line 89: Rather than changing it here I could have renamed the prop in the parent too.

app/javascript/listings/components/MessageModal.jsx
nickytonline: The git hook should be removing it. It runs `prettier` on the files you committed.

app/javascript/listings/components/MessageModal.jsx
nickytonline: I think it may make more sense to keep the more explicit prop name `onChangeDraftingMessage` as `onChange` is pretty generic and it's not clear what it might be for.

app/javascript/listings/components/MessageModal.jsx
Rafi993: I found that I had global prettier config that was causing this issue after I removed the global config everything was fine.

app/javascript/listings/__tests__/Modal.test.jsx
```diff
@@ -3,9 +3,7 @@ import { render } from '@testing-library/preact';
 import { axe } from 'jest-axe';
 import Modal from '../components/Modal';
 
-
 describe('<Modal />', () => {
-  const idFromMessageModal = 'listings-message-form';
```
Rafi993, line 8: This was not being used so I had to remove this (eslint was throwing error)

PR #8778, opened 2020-06-18
Table component & Dashboard update

ludwiczakpawel:

- [x] Refactor
- [x] Feature

## Description

This PR introduces two things: 
1. Super basic crayons styling for tables. It's definitely far away from being _great_ but for sure it's a good start.
2. Some changes for dashboard: updating layout a little and adding crayons table (^) to display posts. Similarly to what I mentioned re tables: this one is also just a start and I'll be working iteratively to improve other parts of the dashboard and its layout itself as well.

Just go to dashboard and click around.

image

It doesn't look ideal but it's not supposed to :) It's just a kickoff and I want to keep this PR minimal.

## Added tests?

- [ ] yes
- [ ] no, because I need help
- [x] not yet.

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


ludwiczakpawel:
@maestromac cheers, i've swapped the order of the buttons a little so they should be more aligned now. eventually we will hide them and show only on hover I think.
also, i've fixed specs and waiting to see if build fails or not this time :)

@rhymes per what we chatted on slack: looks like this is due to your local env state.

PR #8779, opened 2020-06-18
User subscriptions API and backend updates

atsmith813:
- [x] Feature

## Description
This PR is setting up the necessary API updates for the new liquid tag to allow readers to share their email address with the author of articles from approved organizations for CodeLand. The liquid tag will render a `Subscribe` button. Once clicked, the client will make an API call to this endpoint to handle creating a `UserSubscription`. I restricted the use of the tag to `super_admin` and `admin` roles, for now, while I'm still iterating. Feature flags aren't quite up and running yet and this PR is already large enough.

This includes:
- A new `/user_subscriptions` endpoint.
- Rate limiting* (which causes the number of files changed to feel daunting).
- Specs.
- A new `/user_subscriptions/subscribed` endpoint for the liquid tag to know if a `current_user` is subscribed via a specified source already or not.
- Temporarily restrict the liquid tag to `admin` and `super_admin` roles.

_*Please check my rate limiting approach. The process for rate limiting is currently undocumented (AFAIK) and I just want to double-check I got it right 😅 . I'm open to adding or replacing this with throttling if that makes more sense to folks._

Quick context-setting - we're trying to walk the line of _"getting this done for CodeLand"_ and doing this _"perfectly"_. It'll become more apparent in the next PR, but the way we handle liquid tags (especially one that's going to be so dynamic), starts uncovering many areas that we can improve as a whole (caching, rendering, permissions, duplicate code, etc.). I've got a backlog of topics for a larger team discussion in the future🙃 .

- [ ] [Create a user subscription button liquid tag](https://github.com/thepracticaldev/dev.to/issues/8168)
- [ ] [Create a new role for users approved to use the new user subscription liquid tag](https://github.com/thepracticaldev/dev.to/issues/8165)

https://github.com/thepracticaldev/dev.to/issues/8171

You can fire up the server locally and make a `POST` request to `/user_subscriptions`. If doing it from a console, here's an example of a valid request body:
```ruby
article = Article.first
user = User.last

{
  user_subscription: {
    source_type: article.class_name,
    source_id: article.id,
    subscriber_email: user.email
  }
}
```

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

getting_ready_psych_gif


app/controllers/async_info_controller.rb
```diff
@@ -53,25 +53,27 @@ def broadcast_data
...
+        id: @user.id,
         moderator_for_tags: @user.moderator_for_tags,
-        config_body_class: @user.config_body_class,
+        name: @user.name,
         pro: @user.pro?,
-        created_at: @user.created_at
+        profile_image_90: ProfileImage.new(@user).get(width: 90),
+        reading_list_ids: ReadingList.new(@user).cached_ids_of_articles,
+        saw_onboarding: @user.saw_onboarding,
+        subscription_source_article_ids: @user.cached_subscription_source_article_ids,
```
atsmith813, line 74: I alphabetized this list and added:
```ruby
email: @user.email
subscription_source_article_ids: @user.cached_subscription_source_article_ids
```
We want the `email` to populate text in the liquid tag. Something like:
> Hey, user! Are you sure you want to share your email address (test@test.com) with the author_username?

The exact wording is TBD, but we'll want the email address so it's as clear as possible to the user what they're agreeing to.

And we want `subscription_source_article_ids` so we know which Authors the current user is subscribed to and through which articles. Then we know to display the "already subscribed" UX.

app/controllers/user_subscriptions_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+      error: "user email mismatch",
+      status: 422
+    }, status: :unprocessable_entity
+  end
+
+  def user_subscription_tag_enabled?(user_subscription_source)
+    liquid_tags =
+      case user_subscription_params[:source_type]
+      when "Article"
+        user_subscription_source.liquid_tags_used(:body)
```
atsmith813, line 63: We only care if the liquid tag is used in the `body` of the Article.

app/liquid_tags/email_subscription_tag.rb
```diff
@@ -0,0 +1,9 @@
+class UserSubscriptionTag < LiquidTagBase
+  def initialize(_tag_name, cta_text, _tokens)
+    @cta_text = cta_text.strip
+  end
+
+  def render(context); end
+end
+
+Liquid::Template.register_tag("user_subscription", UserSubscriptionTag)
```
atsmith813, line 9: This entire liquid tag will be built in the next PR. For now, it's "empty" so that
```ruby
liquid_tags.include?(UserSubscriptionTag)
```
doesn't blow up in the controller with `undefined`.

app/models/article.rb
```diff
@@ -351,6 +351,22 @@ def update_score
...
+      when :comments
+        comments_blob
+      else
+        "#{body_markdown}#{comments_blob}"
+      end
+
+    MarkdownParser.new(content).tags_used
+  rescue StandardError
+    []
+  end
```
atsmith813, line 368: This is a refactor. For the context of the liquid tag, we don't want to check the comments of an article, only the body of the article itself.

app/models/user.rb
```diff
@@ -227,6 +227,16 @@ def cached_following_podcasts_ids
     end
   end
 
+  # TODO: (Alex Smith) - Adjust TTL and limit query size post Codeland
+  def cached_subscription_source_article_ids
+    cache_key = "user-#{id}-#{subscribed_to_user_subscriptions_count}/subscription_source_article_ids"
+    Rails.cache.fetch(cache_key, expires_in: 12.hours) do
```
atsmith813, line 233: Very open to changing the TTL. Just stuck with the `12.hours` trend. It'll be ignored anyway with `subscribed_to_user_subscriptions_count` changing.

config/routes.rb
```diff
@@ -236,6 +236,7 @@
   resources :podcasts, only: %i[new create]
   resources :article_approvals, only: %i[create]
   resources :video_chats, only: %i[show]
+  resources :user_subscriptions, only: %i[create]
```
atsmith813, line 239: I will likely exclude this route and maybe more from the service worker in a later PR once we sort out exactly how we're going to do the author's view.

app/controllers/user_subscriptions_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+  before_action :authenticate_user!
+
+  def create
+    rate_limit!(:user_subscription_creation)
+
+    source_type = user_subscription_params[:source_type]
+    return invalid_type_error unless UserSubscription::ALLOWED_TYPES.include?(source_type)
+
+    source_id = user_subscription_params[:source_id]
+    user_subscription_source = source_type.safe_constantize.find_by(id: source_id)
```
atsmith813, line 11: Could technically use `constantize` here too since we validate its above.

app/controllers/user_subscriptions_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+  def stale_subscriber_email
+    render json: {
+      error: "subscriber email mismatch",
+      status: 422
+    }, status: :unprocessable_entity
+  end
+
+  def user_subscription_tag_enabled?(user_subscription_source)
+    liquid_tags =
+      case user_subscription_params[:source_type]
```
atsmith813, line 61: I made this a `case` to leave room for the future 🙈

app/models/article.rb
```diff
@@ -535,9 +546,12 @@ def canonical_url_must_not_have_spaces
     errors.add(:canonical_url, "must not have spaces") if canonical_url.to_s.match?(/[[:space:]]/)
   end
 
+  # TODO: (Alex Smith) remove UserSubscriptionTag from subscription when ready to
+  # release or when permission-based liquid tags are implemented.
+  #
   # Admin only beta tags etc.
   def validate_liquid_tag_permissions
-    errors.add(:body_markdown, "must only use permitted tags") if liquid_tags_used.include?(PollTag) && !(user.has_role?(:super_admin) || user.has_role?(:admin))
+    errors.add(:body_markdown, "must only use permitted tags") if (liquid_tags_used & RESTRICTED_TAGS).any? && !(user.has_role?(:super_admin) || user.has_role?(:admin))
```
atsmith813, line 554: Handling permissions, roles, and liquid tags are going to change entirely in a separate PR. But, because this PR technically introduces the liquid tag now by including `app/liquid_tags/email_subscription_tag.rb`, I need to somehow restrict it so users don't start messing with it.

I thought of just restricting the endpoint to `admin` roles as well. But I've coordinated with the design team and we decided what's easiest would be for me to push a very bare-bones UI with all the logic in place (future PR) that they can then put all their design polish on. Having it restricted this way will allow our internal team to access the tag in dev and prod 🙃 .

spec/requests/user_subscriptions_spec.rb
```diff
@@ -0,0 +1,131 @@
+require "rails_helper"
+
+RSpec.describe "UserSubscriptions", type: :request do
+  # TODO: (Alex Smith) remove super_admin restriction before final release
```
atsmith813, line 4: See my [other comment](https://github.com/thepracticaldev/dev.to/pull/8779/files#r442573640) for more details. I realize in the model it technically allows `admin` _OR_ `super_admin`...since this is going to be removed anyways in the next 2 or 3 PRs and before we officially release it, I didn't go through the trouble of testing both roles. I simply added it to the spec so all my specs are still relevant and work - I just have to remove the `super_admin` in the PR where I setup roles and permissions.

app/models/article.rb
```diff
@@ -535,9 +546,11 @@ def canonical_url_must_not_have_spaces
     errors.add(:canonical_url, "must not have spaces") if canonical_url.to_s.match?(/[[:space:]]/)
   end
 
+  # TODO: (Alex Smith) refactor liquid tag permissions
+  #
   # Admin only beta tags etc.
   def validate_liquid_tag_permissions
-    errors.add(:body_markdown, "must only use permitted tags") if liquid_tags_used.include?(PollTag) && !(user.has_role?(:super_admin) || user.has_role?(:admin))
+    errors.add(:body_markdown, "must only use permitted tags") if (liquid_tags_used & RESTRICTED_LIQUID_TAGS).any? && !(user.has_role?(:super_admin) || user.has_role?(:admin))
```
atsmith813, line 553: Handling permissions, roles, and liquid tags are going to change entirely in a separate PR. But, because this PR introduces the liquid tag now by including `app/liquid_tags/email_subscription_tag.rb`, I need to somehow restrict it so users don't start messing with it, though technically it wouldn't render anything since there's no view included.

I thought of just restricting the endpoint to `admin` roles as well. But I've coordinated with the design team and we felt it would be easier for me to push a very bare-bones UI with all the logic in place (future PR) that they can then put all their design polish on. Having it restricted this way will allow our internal team to access the tag in dev (assigning their user `super_admin` locally) while still keeping it "hidden" in prod (very few people have that role)🙃 .

app/controllers/async_info_controller.rb
rhymes: I see a privacy issue here as this means that _all_ users will have their emai suddenly exposed in the async data info blob

Can we share this in the async blob ONLY if the requesting user is an admin? AFAIK we said only privileged users would be using this feature.

cc @mstruve @vaidehijoshi

atsmith813:
@rhymes I agree, this is a tough one 😬 . Adding the email here makes me uneasy as well. I spoke with Ben about it before and we thought adding it here would be okay for now, as is.

I'm hesitant about restricting that to a role. As of right now, roles outside of a few moderator ones can only be managed via a production console. As is, managing the liquid tag with roles is already going to be somewhat of a headache since very few people have production console access. We'll have to build it out an admin UI and re-think roles a little bit in the future to make this easier.

rhymes:
@atsmith813 what about managing it with feature flags?

atsmith813:
@rhymes I tried to, but feature flags aren't fully up and running yet 😞  A feature flag would be perfect for this case.

rhymes:
@atsmith813 I think they are working, I found out for the "sign in with Apple" PR. You just need to ping someone to enable them manually as the UI dashboard doesn't work. cc @joshpuetz

atsmith813:
@rhymes when I spoke with @joshpuetz he mentioned they could be turned on manually.

I was trying to avoid that route as well because I don't have access or confidence to do it. Feature flag usage is not documented yet so it would be hard for me to direct someone with access manually when I can't see it myself. I also don't want that to slow down these iterations and introduce more complexity that I can't access to quickly resolve (production console).

Is it just the _creation_ of feature flags that are a little wonky? And it can still be managed via the UI?

atsmith813:
Following up from Slack...restricting the email by role would only apply for the next PR or two. Once this feature is ready for a production release, we don't want to restrict which readers/users/subscribers are able to subscribe via the liquid tag by a role. We are only restricting authors who are allowed to use the tag.

app/controllers/async_info_controller.rb
mstruve: Couple things I am thinking of looking at this. 
1) `subscription_source_article_ids`: We are going to make a request to Redis/Postgres for every single user page load so we can populate in very few cases an "already subscribed" UX. That feels like we are doing a lot of processing for very little in return. I think the more performant solution would be to throw up an "already subscribed" message upon submission. That will save us millions of these external db hits a day. 
2) Should we allow the user to enter an email address rather than force them to use their DEV one? This would eliminate the need to send the user email in the async info which just saves some data across the wire. Much less concerned about that than the extra db hit.

atsmith813: 1. That's totally fair. It's a "cheap" query and I see what you mean by it needs to be populated in very few cases. We explored this avenue as well. It's a far less than ideal UX to have to re-submit to get that confirmation and we felt the cost of doing this wasn't worth that compromise in UX. An alternative route we considered was moving the logic into its own API call from the liquid tag. In that case, it would fire anytime an article with this tag is viewed (only when "needed"). We figured it sticking it in `base_data` was a quick win since 1) `base_data` needs an overall update anyways so we're only slightly if at all, adding to the "tech debt" here 2) it's more consistent with how we're currently doing things, even if sub-optimally and 3) we are going to re-visit this entire thing post-Codeland to open it up more, so we could take what we learn from that to make a more informed decision that's longer-lasting.
2. We reviewed that point extensively when we scoped this project - that's not a road we want to go down right now because it opens up the door for a lot of spam.

mstruve: > We figured it sticking it in base_data was a quick win 

We did this a bunch in the past and it led to me cleaning up a lot of code to help us performance-wise which is why I want to avoid doing it. I think making an API call from the liquid tag sounds like the way to go. I would imagine it would act similarly to how we populate the text on follow buttons.

atsmith813: 😢  will do.

app/models/user_subscription.rb
```diff
@@ -4,6 +4,8 @@
 class UserSubscription < ApplicationRecord
   ALLOWED_TYPES = %w[Article].freeze
 
+  counter_culture :subscriber, column_name: "subscribed_to_user_subscriptions_count"
```
juliannatetreault, line 7: TIL `counter_culture` and `culture_cache` 😄

joshpuetz:
> Is it just the _creation_ of feature flags that are a little wonky? And it can still be managed via the UI?

No, management of them is wonky via UI as well. Everything with feature flags works just fine programmatically. The Web UI is bugged: any request to the backend that it's a simple GET fails.

app/controllers/async_info_controller.rb
atsmith813: @mstruve I updated it to follow the "follow buttons" approach, thanks for the suggestion! It feels much better now. Let me know if it can be improved further.

@rhymes I restricted the email to `admin` and `super_admin` roles for now. We'll just need to remove my quick and dirty patch before the final release.

app/controllers/async_info_controller.rb
```diff
@@ -52,27 +52,33 @@ def broadcast_data
...
+          pro: @user.pro?,
+          profile_image_90: ProfileImage.new(@user).get(width: 90),
+          reading_list_ids: ReadingList.new(@user).cached_ids_of_articles,
+          saw_onboarding: @user.saw_onboarding,
+          trusted: @user.trusted,
+          username: @user.username
+        }
+
+      # TODO: (Alex Smith) remove this when ready
+      data[:email] = @user.email if @user.has_role?(:super_admin) || @user.has_role?(:admin)
```
atsmith813, line 79: This is temporary, we'll need to remove it before the final release. We don't want to restrict which users can or cannot subscribe to an author in the final version.

mstruve, line 79: Let's just add `:email` to the hash right now with the above comment next to it rather than doing a role lookup here. That's going to add 2 more db hits which I would rather avoid in exchange for a couple more bits in the hash.

atsmith813: That's a great point, good call. I'll have to scrap `email` altogether 😢  - totally doable, just causes me much more pain in the next few PRs.

It's not ready to be applied to all users yet and I don't think it's worth the trouble of me caching the roles to make this more performant when the check is going to be removed anyway.

app/controllers/user_subscriptions_controller.rb
```diff
@@ -0,0 +1,87 @@
+class UserSubscriptionsController < ApplicationController
+  before_action :authenticate_user!
+
+  def subscribed
+    params.require(%i[source_type source_id])
+    source_type = params[:source_type]
+    source_id = params[:source_id]
+
+    is_subscribed = UserSubscriptions::CacheChecker.new(current_user, source_type, source_id).cached_subscription_check
```
citizen428, line 9: Do we want to follow the convention we use elsewhere here and add a class level `.call` method?

```suggestion
    is_subscribed = UserSubscriptions::CacheChecker.call(current_user, source_type, source_id)
```

The method can be as simple as 

```ruby
def self.call(user, source_type, source_id)
  new(user, source_type, source_id).cached_subscription_check
end
```

app/controllers/user_subscriptions_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    render json: {
+      is_subscribed: is_subscribed,
+      status: 200
```
citizen428, line 13: Isn't this a bit redundant with the HTTP status? If we want to communicate success, how about a boolean flag `success: true` which potentially also simplifies the JS consumer a bit.

app/controllers/user_subscriptions_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+    }, status: :ok
+  end
+
+  def create
```
citizen428, line 17: This method is quite long. Feel free to leave as is, but a pattern I commonly follow for this is a service object like `UserSubscriptions::Create`, which returns an object (an `OpenStruct` most likely) that responds to `success?`, `error` (for when `success?` is `false`) and `data` (when `success?` is true). Then this whole action would look something like:

```ruby
def create
  user_subscription = UserSubscriptions::Create.call(...)

  if user_subsription.success?
    rate_limiter.track_limit_by_action(:user_subscription_creation)
    render json: { message: "success", status: 200 }, status: :ok
  else
    error_response(user_subscription.error)
  end
end
```

app/controllers/user_subscriptions_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+    case source_type
+    when "Article"
+      user_subscription_source.published?
+    else
+      false
+    end
+  end
+
+  def error_response(msg)
+    render json: { error: msg, status: 422 }, status: :unprocessable_entity
```
citizen428, line 71: Same comment as above, instead of duplicating the HTTP status code, maybe a `success: false` would suffice.

app/controllers/user_subscriptions_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+  # same as their current email address. A mismatch occurs if a user updates
+  # their email address in a new/separate tab and then tries to subscribe on
+  # the old/stale tab without refreshing. In that case, the user would have
+  # consented to share their old email address instead of the current one.
+  def subscriber_email_stale?
+    current_user&.email != user_subscription_params[:subscriber_email]
+  end
+
+  def user_subscription_params
+    accessible = %i[source_type source_id subscriber_email]
```
citizen428, line 84: Maybe make this a constant?

app/models/article.rb
```diff
@@ -351,6 +352,22 @@ def update_score
                    spaminess_rating: BlackBox.calculate_spaminess(self))
   end
 
+  def liquid_tags_used(section = nil)
```
citizen428, line 355: Does this really need to live on the model? A caller of this method already has access to an article object, so unless you call this in many different places, it can live closer to where it's needed? `Article` and `User` area already very heavy on responsibilities, we should always consider if we want to add more to that.

app/controllers/user_subscriptions_controller.rb
atsmith813: I went to do that originally and decided against it to leave room for more methods in `CacheChecker`. But on second thought, I can rename it to `SubscribtionCacheChecker` and implement this suggestion 👍

app/models/article.rb
atsmith813: Nope - it doesn't! It can be removed. Right now it's currently being used for our "hacky" liquid tag permission setup. Once I refactor how that works in 2 more PRs, I'll come back and take this out.

app/controllers/user_subscriptions_controller.rb
atsmith813: I love this idea! I [marked it down](https://github.com/thepracticaldev/dev.to/projects/19#card-40564468) to circle back in another PR. I suspect this controller will be tweaked further in the next PR so it will be even easier to set up a new service once those final details are known 💯

app/controllers/user_subscriptions_controller.rb
atsmith813: That's fair, it is. I'll switch it to a boolean.

My best guess for why we do this in other places is because once you resolve the promise in JS and parse the response body, the function you pass that parsed response to no longer is aware of what the status was. By having the status code duplicated, it avoided having to pass it separately. I don't plan on differentiating logic by specific codes so boolean seems good to me 👍.

app/controllers/async_info_controller.rb
rhymes: Thanks for the update! I'm a fan of @mstruve suggestion:

> Should we allow the user to enter an email address rather than force them to use their DEV one? This would eliminate the need to send the user email in the async info which just saves some data across the wire. Much less concerned about that than the extra db hit.

We could actually do this by letting people enter it and, optionally, then validating it server side to make sure it's the correct email. This way we never send back the actual address. Just a thought

spec/requests/user_subscriptions_spec.rb
```diff
@@ -0,0 +1,174 @@
...
+             subscriber_id: user.id,
+             subscriber_email: user.email,
+             author_id: article.user_id,
+             user_subscription_sourceable: article)
+
+      valid_params = { source_type: article.class_name, source_id: article.id }
+      get subscribed_user_subscriptions_path, params: valid_params
+
+      expect(response).to have_http_status(:ok)
+      expect(JSON.parse(response.body)["is_subscribed"]).to eq true
```
mstruve, line 28: could also use `response.parsed_body` here :)

spec/requests/user_subscriptions_spec.rb
```diff
@@ -0,0 +1,174 @@
...
+    end
+
+    it "returns false if a user is not already subscribed" do
+      valid_params = { source_type: "Article", source_id: 999 }
+      get subscribed_user_subscriptions_path, params: valid_params
+
+      expect(response).to have_http_status(:ok)
+      expect(JSON.parse(response.body)["is_subscribed"]).to eq false
```
mstruve, line 36: Same as above, `response.parsed_body`

app/services/user_subscriptions/subscription_cache_checker.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    def initialize(user, source_type, source_id)
+      @user = user
+      @source_type = source_type
+      @source_id = source_id
+    end
+
+    def call
+      return false unless user
```
vaidehijoshi, line 16: I wonder if `return`ing `false` here is the right thing to do; if there is no `user` passed, then this would return `false`, and then `is_subscribed` would [be `false`](https://github.com/thepracticaldev/dev.to/pull/8779/files#diff-2ee6f8c98b7e8296071384bd2ee06ac9R11). But in reality, it's not that the user is not subscribed, it's that we didn't pass the `SubscriptionCacheChecker` a `user`... 🤔

Should we be handling the case where no `user` is passed to the `SubscriptionCacheChecker` class differently? Raise an error and then notify that we tried to call this service without a user?

db/migrate/20200618212422_add_subscribed_to_user_subscriptions_count_to_users.rb
```diff
@@ -0,0 +1,9 @@
+class AddSubscribedToUserSubscriptionsCountToUsers < ActiveRecord::Migration[6.0]
+  def self.up
+    add_column :users, :subscribed_to_user_subscriptions_count, :integer, null: false, default: 0
+  end
+
+  def self.down
+    remove_column :users, :subscribed_to_user_subscriptions_count
+  end
+end
```
vaidehijoshi, line 9: Any reason not to make this a `def change` method? IIRC, using `add_column` in a `change` should be reversible and run `remove_column` by default, right?

spec/requests/user_subscriptions_spec.rb
```diff
@@ -0,0 +1,174 @@
...
+             headers: { "Content-Type" => "application/json" },
+             params: { user_subscription: invalid_source_attributes }.to_json
+      end.to change(UserSubscription, :count).by(0)
+
+      expect(response).to have_http_status(:unprocessable_entity)
+      expect(response.parsed_body["error"]).to eq("source not found")
+    end
+
+    it "returns an error for an inactive source" do
+      unpublished_article = create(:article, user: super_admin_user, body_markdown: "---\ntitle: User Subscription#{rand(1000)}\npublished: false\n---\n\n{% user_subscription 'CTA text' %}")
```
vaidehijoshi, line 82: Not blocking at all, but at some point we might want to clean these specs up a bit for readability and extract this into a factory or trait

spec/requests/user_subscriptions_spec.rb
```diff
@@ -0,0 +1,174 @@
...
+      expect(response).to have_http_status(:unprocessable_entity)
+      expect(response.parsed_body["error"]).to eq("user subscriptions are not enabled for the requested source")
+    end
+
+    it "returns an error for an invalid UserSubscription" do
+      article = create(:article, user: super_admin_user, body_markdown: "---\ntitle: User Subscription#{rand(1000)}\npublished: true\n---\n\n{% user_subscription 'CTA text' %}")
+
+      # Create the UserSubscription directly so it results in a
+      # duplicate/invalid record and returns an error. This mimics a user
+      # trying to subscribe to the same user via the same source, twice.
```
vaidehijoshi, line 112: Thank you for adding this context! 👍

spec/services/user_subscriptions/subscription_cache_checker_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+           subscriber_id: user.id,
+           subscriber_email: user.email,
+           author_id: article.user_id,
+           user_subscription_sourceable: article)
+
+    expect(described_class.call(user, article.class_name, article.id)).to eq(true)
+  end
+
+  it "checks if subscribed to a thing and returns false if they are not" do
+    expect(described_class.call(user, article.class_name, article.id)).to eq(false)
```
vaidehijoshi, line 18: This test also makes me think we might want to account for not passing a user to the `SubscriptionCacheChecker` differently than returning `false`...if you did `described_class.call(nil, article.class_name, article.id)` here, I suspect it would be `false` as well?

app/liquid_tags/user_subscription_tag.rb
```diff
@@ -0,0 +1,9 @@
+class UserSubscriptionTag < LiquidTagBase
+  def initialize(_tag_name, cta_text, _tokens)
+    @cta_text = cta_text.strip
+  end
+
+  def render(context); end
```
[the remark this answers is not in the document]
atsmith813: Correct - I'm intentionally not rendering anything. I only included this file so the reference to `UserSubscriptionTag` doesn't blow up with `undefined`. The actual building of this liquid tag, the view, the rendering, etc. is the next PR.

I think the comment of me addressing this got lost in the shuffle 😅
https://github.com/thepracticaldev/dev.to/pull/8779#discussion_r442527660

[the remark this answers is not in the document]
vaidehijoshi: Gotcha, thanks for pointing me back to it :)

app/models/article.rb
```diff
@@ -535,9 +546,11 @@ def canonical_url_must_not_have_spaces
     errors.add(:canonical_url, "must not have spaces") if canonical_url.to_s.match?(/[[:space:]]/)
   end
 
+  # TODO: (Alex Smith) refactor liquid tag permissions
```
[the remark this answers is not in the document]
atsmith813: Agreed! I thought about doing that at first and I left it to my name since all of this is just for the sake of iterating with smaller PR sizes. Once it's "done", any remaining TODOs I'll change to `@thepracticaldev/delightful`. Under the current format, it's easy for me to `grep` for my name and catch what I need quickly as I'm iterating.

app/services/user_subscriptions/subscription_cache_checker.rb
atsmith813: I'm torn on this. This case shouldn't be hit since we authenticate first. So this would happen if we authenticated and `current_user` was `nil`. By returning `false` it's just informing the client to treat the UX as if the user is not already subscribed. The client will already have the logic of its own to detect if a user is signed in or not as well which would supersede anything related to subscribed or not 🤔 

If I returned an error, instead, I'm not sure what that would look like on the client. That would be a "you're signed in, but not really" scenario 😅 . Maybe what I'd have to do is another confirmation that `current_user` exists before calling this and remove that guard? Is that overkill? What do you think?

db/migrate/20200618212422_add_subscribed_to_user_subscriptions_count_to_users.rb
atsmith813: Yep, `change` would work! TBH I just used the built-in generator provided by the [counterculture gem](https://github.com/magnusvk/counter_culture) to create this migration. And since it's more explicit, I didn't bother changing it 🙈

app/services/user_subscriptions/subscription_cache_checker.rb
vaidehijoshi: > Maybe what I'd have to do is another confirmation that current_user exists before calling this and remove that guard? Is that overkill? What do you think?

This feels right to me; seems like the controller can just handle this and the cache checker class shouldn't need to every worry about it at all. And actually, if `UserSubscriptionsController` inherits from `ApplicationController`, then you should be able to use `authenticate_user!` to check that the `current_user` exists and then remove this guard completely!

atsmith813: Great point - I'll just drop the guard 🎉

PR #8781, opened 2020-06-18
Created the spinner for the image uploader

nickytonline:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Creates the spinner component that can be used anywhere, but for now it's for the image uploaded in the editor v2.

*The skip is simply the animated gif ending*

Jun-18-2020 18-41-47

image

image

image

image

To see the spinner in action:

* Navigate to https://deploy-preview-8781--storybookdevto.netlify.app/?path=/story/4-app-components-shared-spinner--default
* Use the theme drop down to see it rendered for different themes.

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

A dolphin spinning in the water


nickytonline:
I'm calling it a day @ludwiczakpawel, but I'm forgetting something silly about SVG scaling. It's hard coded to width/height `18px` and a `viewBox="0 0 18 18"` currently.

app/javascript/shared/components/Spinner.jsx
```diff
@@ -0,0 +1,18 @@
+import { h } from 'preact';
+
+export const Spinner = () => (
+  <svg
+    className="crayons-spinner"
+    width="18px"
+    height="18px"
+    viewBox="0 0 18 18"
```
ludwiczakpawel, line 8: Maybe we could redo the SVG a little to be within 24px artwork like rest of our icons? Im happy to do so, let me know if you need help with that.

app/javascript/shared/components/Spinner.jsx
```diff
@@ -0,0 +1,18 @@
+import { h } from 'preact';
+
+export const Spinner = () => (
+  <svg
+    className="crayons-spinner"
```
ludwiczakpawel, line 5: Should we add `crayons-icon` class there as well? Since it could be potentially used elsewhere as replacement for other icons so maybe it could use our default `crayons-icon` styling. Orrrrr, apply that styling to `crayons-spinner` instead?

app/javascript/shared/components/Spinner.jsx
nickytonline: I can put it to 24px. I was basing it off the Figma page that you showed me where it had it as 18px for the image uploaded.

app/javascript/shared/components/Spinner.jsx
nickytonline: Done

app/assets/stylesheets/spinner.scss
```diff
@@ -0,0 +1,13 @@
...
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
+
+.crayons-spinner {
+  color: var(--accent-brand);
+  animation: 0.75s linear infinite crayons-spinner;
+}
```
ludwiczakpawel, line 13: One last thing - should we have this file in `/stylesheets/components` folder?

rhymes:
I have an odd question, as this is essentially a SVG with a CSS animation, do we absolutely need it to be wrapped in a JS component?

ludwiczakpawel:
@rhymes are you ever going to be able to use loading spinner (in a meaningful way) without actually JS? :)

nickytonline:
> Love it!
> 
> I have an odd question, as this is essentially a SVG with a CSS animation, do we absolutely need to have it wrapped by a JS component?

In the context of Preact, it does need to be wrapped in JS as it has to be a component. If we were to use the spinner outside of a Preact component, we could promote this to just an SVG file and then it could be used in non-Preact stuff as well. The implications of doing that though mean you'd have to load the SVG contents and then dangerously set the innerHTML of the component. Obviously we could just opt to not make it a component, but that's another discusssion. 😅

app/assets/stylesheets/spinner.scss
nickytonline: I was thinking about this too. I currently have it as an app component in Storybook, but should this icon be considered as part of the design system instead of an app component? I'm thinking yes.

rhymes:
> Obviously we could just opt to not make it a component, but that's another discusssion

that was my question though :-)

nickytonline:
I'd say if we need it outside of the editor at some point @rhymes, we can probably make it an svg file then.

PR #8818, opened 2020-06-22
Add overflow-x to KaTeX liquid tag

### LINKED ISSUE #8812: If katex overflows horizontally there is no scroll option
**Describe the bug**
If katex overflows horizontally, there is no way to scroll. Here is an example on mobile:
image

On the web version of `dev.to` this looks as follows:
image

If there is overflow on the web version there is also no way to scroll horizontally see [this test post](https://dev.to/wayofthepie/test-katex-overflow-56j6-temp-slug-7812927?preview=ce2562f79e095273b5ae803f95fb2eabf6c6bbee4d89f3193ae246251a3b39e5b810bdf8470ed3480c5f8489803e6aa024620fd9b1f9891fbda5d093). 

image

**To Reproduce**
Add some katex which will overflow the page horizontally, for example:

```
{% katex %}
\begin{aligned}
(0 * 16 ^5) + (0 * 16 ^4) + (0 * 16 ^3) + (5 * 16 ^2) + (0 * 16 ^1) + (0 * 16 ^0) + (0 * 16 ^0) + (0 * 16 ^0) + (0 * 16 ^0) + (0 * 16 ^0) + (0 * 16 ^0) + (0 * 16 ^0) + (0 * 16 ^0) + (0 * 16 ^0) 
\end{aligned} 
{% endkatex %}
```

**Expected behavior**
When code blocks are longer than the width of the article pane they allow horizontal scrolling. Here is an example code block:
image
I would expect the same horizontal scroll bar to appear for katex also.

- OS: Ubuntu 20.04
- Browser: Firefox
- Version: 77.01

- OS: Android 9
- Browser: Firefox
- Version: 68.9.0

Thanks!

citizen428:

- [X] Bug Fix

## Description

So far when a KaTeX tag overflowed the article area there was no way to scroll to the cut off content. I added `overflow-x: auto;` to the non-inline version of the tag now. We are using style attributes on other Liquid tags so I figured it's ok.

Closes #8812

There's now a scrollbar when content in KaTeX overflows:

Screen Shot 2020-06-22 at 11 17 30

## Added tests?

- [X] no, but updated approval HTML.

## Added to documentation?

- [X] no documentation needed


app/views/liquids/_katex.html.erb
```diff
@@ -7,7 +7,7 @@
   <%= parsed_content %>
 </span>
 <% else %>
-<div class="katex-element">
+<div class="katex-element" style="overflow-x: auto;">
```
Ridhwana, line 10: Just a question on this: Is this inline for a specific purposes instead of it going into a stylesheet?

citizen428: Not really. We are doing it for several other Liquid tags too, and I sometimes find it a bit easier to have it right there when looking at the tag, but I can move it if there's a strong preference.

citizen428:
I may need help with the failing spec, already updated the approval but it still seems to fail /cc @maestromac

rhymes:
@citizen428 I ran `approvals verify` and uploaded the change

app/views/liquids/_katex.html.erb
maestromac: I don't think we properly established it yet but we should avoid adding inline style to liquid tags because the content is hardcoded as `body_html` which limits future changes.  It's best to apply this to `KatexTag.scss`, please see my temp commit here https://github.com/maestromac/dev.to/commit/c14300bed5ac15c8fffc72301a6c39572a788371

citizen428: Thanks @maestromac!

PR #8829, opened 2020-06-22
Update Connect responsiveness on smaller viewports

lisasy:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
- Update responsive experience for Connect to use single-focus views for small screens
- Remove `expanded` and `contracted` logic in order to rely on responsiveness to define experience
- Update a few related tests

https://github.com/thepracticaldev/dev.to/issues/8442

📹 [Video demo](https://share.getcloudapp.com/7Kumpjow)

Screen Shot 2020-06-22 at 9 18 33 AM

Screen Shot 2020-06-22 at 9 18 41 AM

Screen Shot 2020-06-22 at 9 18 46 AM

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

alt_text


app/javascript/chat/chat.jsx
```diff
@@ -1233,18 +1237,10 @@ export default class Chat extends Component {
           </div>
         );
       }
-      if (state.expanded) {
+      if (state.expanded || !state.expanded) {
```
lisasy, line 1240: @benhalpern @rhymes 

So.... I know that there is definitely a better way to do this 😅

But essentially, what I'm suggesting is that we always display the search filters, regardless of whether this `state` is `expanded` or not. I think we can rely more on CSS & viewport breakpoints to determine when things are exposed on not, versus relying on doing that in the JS here.

app/javascript/chat/chat.jsx
```diff
@@ -1829,9 +1846,8 @@ export default class Chat extends Component {
     return (
       <div
         data-testid="chat"
-        className={`chat chat--${
-          state.expanded ? 'expanded' : 'contracted'
-        }${detectIOSSafariClass} chat--${
+        className={`chat chat--expanded
```
lisasy, line 1849: it's possible to even remove the `chat--expanded` class from this entirely.

app/javascript/chat/chat.jsx
joshpuetz: So...could we just remove the if conditional?

lisasy: Haha yes I would love to! As a JS noobie, I wanted to share this first before I removed anything that could be potentially really important.

benhalpern:
This looks pretty great overall. One thing: The open/close animation is a sort of janky when opening a channel.

app/javascript/chat/chat.jsx
fdoxyz: +1 on removing the if altogether 😃

app/javascript/chat/chat.jsx
```diff
@@ -1309,15 +1305,6 @@ export default class Chat extends Component {
...
-            onClick={this.toggleExpand}
-            style={{ width: '100%' }}
-            type="button"
-            title="Expand channels"
-          >
-            {'>'}
-          </button>
           <Channels
             activeChannelId={state.activeChannelId}
             chatChannels={state.chatChannels}
```
fdoxyz, line 1310: By removing the `if (state.expanded || !state.expanded)` statement above it will mean that this "fallback return" can be removed too

app/javascript/chat/chat.jsx
lisasy: Removed it.

app/assets/stylesheets/chat.scss
```diff
@@ -60,100 +80,32 @@
...
-}
-
-.chat__channels--expanded {
-  width: 219px;
-  min-width: 219px;
 
-  @media screen and (min-width: 1000px) {
-    width: 200px;
-    min-width: 200px;
+  @media screen and (max-width: 599px) {
```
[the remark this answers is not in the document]
lisasy: This was already in here and I didn't make any changes to it. Because of the limitations of local development, I'm still not totally sure what this class covers, so this is something I'm not making updates on in this branch.

fdoxyz:
@lisasy I think I found out where the problem with the Channel Settings overlay is coming from: [This negative margin](https://github.com/thepracticaldev/dev.to/blob/master/app/assets/stylesheets/chat.scss#L322)

Adjusting `margin-left: 7px;` & `margin-top: 2px;` worked for me (maybe even better to use `--su-2` or something like that)

ludwiczakpawel:
@lisasy I've fixed the 50/50 split on small breakpoint for you plus I fixed the issue that @fdoxyz mentioned above as well. I think there's one or two more tests failing that should be fixed before merge.

lisasy:
Hey @ludwiczakpawel and @fdoxyz , I've managed to make the requested changes. Can you approve of this PR? :)

rhymes:
@nickytonline I think this failed because the threshold for coverage was not met, what is the actionable here? https://travis-ci.com/github/thepracticaldev/dev.to/jobs/353760749#L2099

app/javascript/chat/chat.jsx
```diff
@@ -75,7 +75,7 @@ export default class Chat extends Component {
       activeContent: {},
       fullscreenContent: null,
       videoPath: null,
-      expanded: window.innerWidth > 600,
+      expanded: window.innerWidth > 767,
```
nickytonline, line 78: `767` seems to be used in a few places. consider making it a constant. It will also make it clearer what that number means which I'm assuming is for low resolution tablets, phones in landscape mode?

app/javascript/chat/chat.jsx
nickytonline: Marie Kondo that if you need to.

nickytonline:
@rhymes, it went down by 0.01%, so maybe I was a little too premature in boosting the lines coverage threshold to 42%. @lisasy, to fix your build error change this line, https://github.com/thepracticaldev/dev.to/blob/master/jest.config.js#L24, to `41`.

rhymes:
@nickytonline I added the two changes: constants and the threshold fix

PR #8836, opened 2020-06-22
Remove trailing slashes from URLs to improve SEO

vaidehijoshi:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Across our app, we have a two different variations of how we link to URLs. In the majority of cases, we link to `/foo`, but in some instances, we link to `/foo/`. Google gets pretty confused when we do this; it interprets these as _two different URLs_. When they both resolves successfully (with a status of `200`), Google is forced to pick between the two. The issue with this is that it splits up our "link equity" for those pages, since Google sees them as two links. Instead, we'd much prefer to consolidate all of our "SEO juice" and link equity around _one, standardized link_.

Thus, in an effort to boost SEO, we really need to standardize around _one_ way of representing links, since, at the end of the day, we only want one representation of each URL.

Ideally, we will want to programmatically parse through links going forward (using a linter or something in our `routes.rb` file to handle redirects with trailing slashes), but I wanted to tackle this as a quick, low-hanging fruit that hopefully result in a big win! 🚀 

N/A

Please make sure I didn't break any links! I did a massive pass through of our app, and while there weren't too many things to change, it's important to check that none of the trailing slashes that I removed will cause broken links! 😉 

## Added tests?

- [x] yes **(updated pre-existing tests)**
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

N/A

### DIFF
--- a/app/controllers/page_views_controller.rb
+++ b/app/controllers/page_views_controller.rb
@@ -46,7 +46,7 @@ class PageViewsController < ApplicationMetalController
   def update_organic_page_views
     return if Rails.env.production? && rand(100) != 1 # We need to do this operation only once in a while.
 
-    page_views_from_google_com = @article.page_views.where(referrer: "https://www.google.com/")
+    page_views_from_google_com = @article.page_views.where(referrer: "https://www.google.com")
 
     organic_count = page_views_from_google_com.sum(:counts_for_number_of_views)
     @article.update_column(:organic_page_views_count, organic_count) if organic_count > @article.organic_page_views_count
--- a/config/initializers/s3_direct_upload.rb
+++ b/config/initializers/s3_direct_upload.rb
@@ -3,5 +3,5 @@ S3DirectUpload.config do |c|
   c.secret_access_key = ApplicationConfig["AWS_S3_VIDEO_KEY"] # your secret access key
   c.bucket = ApplicationConfig["AWS_S3_INPUT_BUCKET"] # your bucket name
   c.region = nil # region prefix. _Required_ for non-default AWS region, eg. "eu-west-1"
-  c.url = nil # S3 API endpoint (optional), eg. "https://#{c.bucket}.s3.amazonaws.com/"
+  c.url = nil # S3 API endpoint (optional), eg. "https://#{c.bucket}.s3.amazonaws.com"
 end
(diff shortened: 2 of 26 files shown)


config/initializers/s3_direct_upload.rb
```diff
@@ -3,5 +3,5 @@
   c.secret_access_key = ApplicationConfig["AWS_S3_VIDEO_KEY"] # your secret access key
   c.bucket = ApplicationConfig["AWS_S3_INPUT_BUCKET"] # your bucket name
   c.region = nil # region prefix. _Required_ for non-default AWS region, eg. "eu-west-1"
-  c.url = nil # S3 API endpoint (optional), eg. "https://#{c.bucket}.s3.amazonaws.com/"
+  c.url = nil # S3 API endpoint (optional), eg. "https://#{c.bucket}.s3.amazonaws.com"
```
maestromac, line 6: 😆

app/controllers/page_views_controller.rb
```diff
@@ -46,7 +46,7 @@ def page_view_params
   def update_organic_page_views
     return if Rails.env.production? && rand(100) != 1 # We need to do this operation only once in a while.
 
-    page_views_from_google_com = @article.page_views.where(referrer: "https://www.google.com/")
+    page_views_from_google_com = @article.page_views.where(referrer: "https://www.google.com")
```
benhalpern, line 49: @vaidehijoshi this is going to cause a problem because we are querying for existing articles that match this exact referrer so we can't change this.

benhalpern:
Just want to give some clarity here: The only thing that matters are _internal links_, so `a` elements that link to another page within the site. So in the changed files, the `FEED` links will matter, and it doesn't hurt to randomly remove them elsewhere, but removing them in every context is not necessarily what we want.

vaidehijoshi:
Thanks for clarifying that, @benhalpern. I probably should have left some comments on this PR to explain that the external site links weren't actually _required_ to boost SEO. I just removed them while I was in the codebase to help keep things consistent since I figured someone down the road might see a trailing slash in a link and follow the pattern.

PR #8838, opened 2020-06-22
Add trailing slash back in

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

PR https://github.com/thepracticaldev/dev.to/pull/8836 accidentally removes the trailing slash here, but this is a count for existing page views by referrer and can't change.

### DIFF
--- a/app/controllers/page_views_controller.rb
+++ b/app/controllers/page_views_controller.rb
@@ -46,7 +46,7 @@ class PageViewsController < ApplicationMetalController
   def update_organic_page_views
     return if Rails.env.production? && rand(100) != 1 # We need to do this operation only once in a while.
 
-    page_views_from_google_com = @article.page_views.where(referrer: "https://www.google.com")
+    page_views_from_google_com = @article.page_views.where(referrer: "https://www.google.com/")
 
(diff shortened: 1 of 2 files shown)


app/controllers/page_views_controller.rb
```diff
@@ -46,7 +46,7 @@ def page_view_params
   def update_organic_page_views
     return if Rails.env.production? && rand(100) != 1 # We need to do this operation only once in a while.
 
-    page_views_from_google_com = @article.page_views.where(referrer: "https://www.google.com")
+    page_views_from_google_com = @article.page_views.where(referrer: "https://www.google.com/")
```
vaidehijoshi, line 49: You beat me to opening this PR :) thanks for fixing my mistake!

Could we add a quick comment here that we are querying for existing articles that match this exact referrer?

benhalpern: Yeah, I think wrapping it in a named method as well is probably an additional step that could help communicate the intent here.

Basically we're querying for the referrer exactly how it's given to us in the `create` context. But if this passes I'll merge it before getting to that.

PR #8841, opened 2020-06-22
Fix accessiblity warnings on Listings page

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This should fix ~60 axe warnings on this page. This commit also removes
some markup that is being overwritten by Preact.

There are no UI changes.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/views/listings/index.html.erb
+++ b/app/views/listings/index.html.erb
@@ -26,26 +26,15 @@
   <meta name="twitter:card" content="summary_large_image">
 <% end %>
 
-<div class="home">
-  <div class="listings-container" id="listings-index-container"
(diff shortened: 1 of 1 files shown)


rhymes:
@jacobherrington  Should we keep the markup? In case we have no-js bots indexing the listing page. It might totally not needed. Just a hunch

jacobherrington:
@rhymes I would agree, but the page doesn't function at all without JS. So unless we want to write a no-JS friendly version of this that'd be great :grin:

PR #8845, opened 2020-06-23
Unflake system user_settings_response_templates_spec.rb

snackattas:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
screenshot_2020-07-20 23 40 50_sDkNkd
This is one of the most common fails in our Travis runs. This Pivot table ^ comes from analyzing the csv reports of our Travis runs

In looking at the travis fail for this spec, this is the fail/stacktrace and the screenshot of the fail
```shell
#<Capybara::ModalNotFound: Unable to find modal dialog>
``` 
[stacktrace.log](https://github.com/forem/forem/files/4953858/stacktrace.log)

failures_r_spec_example_groups_user_uses_response_templates_settings_when_user_is_signed_in_when_user_has_a_response_template_already_shows_the_proper_message_when_deleting_a_reponse_template_53

This spec 
1. Clicks the remove button on a response template, which triggers a browser modal
2. Then uses `accept_confirm` to accept the browser modal
3. If successful there's a toast message saying the template was deleted

In the failed test screenshot, you can see that there is no modal present, BUT, it seems, the modal _was present_ and did its job, since we see the toast message.  So, I think the Capybara method is simply flaky here, sometimes the action of accepting the modal happens too fast it doesn't register. That is the theory I'm going with here.

In in running this spec a bunch on my local, I was able to see the modal appear SUPER fast, and see this test fail in the same way. So I think this workaround of wrapping the Capybara `accept_confirm` in a begin/rescue is ok, since the action it does still gets done.  

This is kind of an awkward solution, but its the only use of `accept_confirm` in the repo as well, so I think its ok. Also, selenium doesn't usually interact with browser modals, its kind of a different thing, so I imagine, it's subject to its own race conditions, different from the regular Capybara methods.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed

sanic


spec/system/user/user_settings_response_templates_spec.rb
```diff
@@ -20,8 +20,15 @@
...
       it "shows the proper message when deleting a reponse template", js: true do
         visit "/settings/response-templates"
-        accept_confirm { click_button "Remove" }
+        expect(page).to have_text(response_template.title)
 
+        expect(page).to have_css(".github-repo-row-name", text: response_template.title)
+        begin
+          accept_confirm { click_button("Remove") }
+        rescue Capybara::ModalNotFound => e
+          puts "This spec is known to his this error because " \
```
mstruve, line 29:
```suggestion
          puts "This spec is known to hit this error because " \
```

mstruve: This PR description is really great, I would drop the PR number here or a link so people can find all the information and details.

spec/system/user/user_settings_response_templates_spec.rb
```diff
@@ -20,8 +20,15 @@
...
+               "accept_conform has issues. Hits this error #{e}"
```
vaidehijoshi, line 30:
```suggestion
               "accept_confirm has issues. Hits this error: #{e}"
```

spec/system/user/user_settings_response_templates_spec.rb
snackattas: thanks for committing the change 👍

snackattas:
> This is a great workaround! TBH I think we could eventually skip or delete this system spec since the request spec works. I wrote this test to test that the whole action worked, but thinking now I probably didn't need to test a basic browser confirmation.

my 2 cents, I think its still important to test with the rendering of the javascript and everything. But if you think its not providing value, can you put in a PR to remove the test then?

PR #8847, opened 2020-06-23
Progress bar fix

### LINKED ISSUE #8830: Loading bar gets stuck at times
**Describe the bug**

Loading bar gets stuck for like 1 of 3 times I want to load DEV main page multiple times in a row.

**To Reproduce**
Try to reload Home by pressing DEV icon on top left multiple times.

**Screenshots**

aaaa

**Expected behavior**
Home page gets loaded and loading bar goes away.

**Desktop:**

- OS: Win10
- Browser: Chrome
- Version: 83.0.4103.106 (64-bit)

**Additional context**

The problem only exists once I'm signed in.
Tested on Firefox 77.0 (64-bit), the same result.

ludwiczakpawel:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR hopefully fixes weird behavior of progress bar reported in #8830. I'm saying "hopefully" because it's more of a gut feeling rather than something I've actually figured out :D 

- Closes #8830 (hopefully)

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


benhalpern:
So the old way the code works was to ensure that the progress bar didn't _poke out_ just a little bit on every load even if it was super fast.

IMO the ideal practical functionality I think would be...

1. progress bar doesn't show up at all for xxx miisseconds (any pseudo-instant load doesn't need it)
2. When it does show up it's already full width and shows the gradiant flowing along, but not necessarily towards "completion" because that's kind of a false idea anyway. Maybe also fade in.

This is just personal taste as I find that the creep-in-from-the-left style has always been kind of awkward in a lot of cases in terms of throwing off the balance of the page.

ludwiczakpawel:
> progress bar doesn't show up at all for xxx miisseconds (any pseudo-instant load doesn't need it)

Now the `140%` value makes sense. I'll revert the change and make some adjustments to make it work again 👍

> When it does show up it's already full width and shows the gradiant flowing along, but not necessarily towards "completion" because that's kind of a false idea anyway. Maybe also fade in.

I'm not sold on this one TBH. Showing full width bar and making the gradient flow only is pretty uncommon pattern on other sites.. youtube, medium, figma... but if you have strong feelings about it i guess i can be bribed :)

But it will require more JS work. Current solution (with what you mentioned in point 1) will make it easy to not show the loader for xxx ms (because we can hide it with CSS by making it wider). But the other solution you proposed would be visible right away so it will need more robust JS...

rhymes:
Closing and reopening to trigger Travis

PR #8865, opened 2020-06-23
Spec Fix: Use Truncation For Tests Where Transactions Get Stuck

mstruve:
- [x] Bug Fix

## Description
As I mentioned in the original KnapsackPro PR #8390 I had to exclude `rss_reader_spec.rb` from the list of specs that were running with Knapsack bc it would constantly hang and fail. After digging in with a lot of logging I found that it would get stuck on a db create statement which led me to start looking at how we were handling the database. Upon further investigation on my own machine, I found that it was just sitting in a transaction. 
```
Chromecast-Ultra:dev.to mollystruve$ ps aux | grep postgres
mollystruve      11032   0.0  0.0  4560088   3480   ??  Ss   11:33AM   0:00.21 postgres: mollystruve PracticalDeveloper_test [local] idle in transaction
```
After some Googling, I figured out how to disable transactions for individual specs by putting 
```ruby
self.use_transactional_tests = false
``` 
at the top of the spec file. I tried putting this in `before`, `around`, and `after` blocks and none of it worked. I also tried `config.use_trasactional_fixtures=false` in all of those blocks and none of it worked. The only way I got it to work was by setting it at the top of the spec file. This will flip the value locally for all the examples in that spec file. However, once the spec is done, it reverts to the global value which is set in `rails_helper.rb`
```ruby
config.use_transactional_fixtures = true
```
Here Rspec has [`use_transactional_fixtures` hooked up to set the underlying `use_transactional_tests` value in Rails](https://github.com/rspec/rspec-rails/blob/148f7ce473e0f9e36018e9a7ca79aad4af0a6d14/lib/rspec/rails/fixture_support.rb#L18).
Since we are not using transactions in these couple examples we have to make sure we clean up the database manually when we are done. To accomplish this I have added the [DatabaseCleaner gem](https://github.com/DatabaseCleaner/database_cleaner) which will clean the database using truncation at the end of each spec example. 

The sorting spec is another one which @rhymes can attest to I have seen hang before so I decided to get ahead of the ball and use truncation for that spec as well. This may be an approach we use to solve a few other flaky system spec issues. 

With this in place, we should be able [to get queue mode going](https://github.com/thepracticaldev/dev.to/compare/mstruve/queue-mode) barring more flaky spec fixes per usual. 

@snackattas @ArturT

Apparently there are gifs for tree trunks 😂 
alt_text


mstruve:
Leave it to @rhymes to come in with the Rails 6 solution! ❤️

mstruve:
Reverted trying to remove the database cleaner, I want to do it but its going to take some debugging in Travis, I'll work on that in a separate branch and ship this one as is.

PR #8868, opened 2020-06-23
Paginate /dashboard posts

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The dashboard on _my_ account has gotten _super_ slow just due to the raw number of posts that we are rendering all at once if you have _many_ posts. So this wouldn't be an issue for most, but some users (power users) have a really really slow dashboard.

This reduces that worst case scenario by rendering 50 posts at a time.

We use the Kaminari gem which comes with this helper which generates its own classes. Not exactly the most crayons-friendly situation, but we could switch to custom implementation later. I think the plain text links are solid here.

Screen Shot 2020-06-23 at 4 33 07 PM

### DIFF
--- a/app/views/dashboards/_dashboard_article_row.html.erb
+++ b/app/views/dashboards/_dashboard_article_row.html.erb
@@ -100,10 +100,6 @@
     <% elsif @user == article.user %>
       <a href="<%= article.path %>/delete_confirm" data-no-instant class="crayons-btn crayons-btn--ghost-danger crayons-btn--s">Delete</a>
     <% end %>
-    <% if @current_user_pro %>
-      <a href="<%= article.path %>/stats" class="crayons-btn crayons-btn--ghost crayons-btn--s">Stats</a>
-    <% end %>
-
     <a href="<%= article.path %>/edit" class="crayons-btn crayons-btn--ghost crayons-btn--s">Edit</a>
 
     <div class="js-dashboard-row-more relative inline-block">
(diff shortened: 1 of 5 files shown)


app/views/dashboards/show.html.erb
```diff
@@ -42,7 +42,7 @@
         <a class="video-upload-cta cta" href="/listings/dashboard" data-no-instant>
           Create/Manage Listings
         </a>
-      <%= select_tag "dashhboard_sort", options_for_select(sort_options, params[:sort]), 'aria-label': 'Sort By' %>
+      <%= select_tag "dashhboard_sort", options_for_select(sort_options, params[:sort]), 'aria-label': "Sort By" %>
```
joshpuetz, line 45: Super nitpick, but is this a type? 
```suggestion
      <%= select_tag "dashboard_sort", options_for_select(sort_options, params[:sort]), 'aria-label': "Sort By" %>
```

app/views/dashboards/_dashboard_article_row.html.erb
```diff
@@ -112,6 +108,9 @@
       </button>
 
       <div class="crayons-dropdown top-100 right-0 z-10 align-left js-dashboard-row-more-dropdown">
+        <% if @current_user_pro %>
```
joshpuetz, line 111: Is pro coming back?!

rhymes: nope, Ben just moved it around, we use pro only for stats now

PR #8871, opened 2020-06-23
User subscriptions liquid tag MVP

atsmith813:
- [x] Feature

## Description
This PR is setting up the bare-bones/MVP of the UX. This will _**not**_ be how it looks. The stylesheet and the markup will be changed. The goal is to just get the functionality hooked up enough so that the design team can jump in and clean it all up (I've already coordinated with them). So I did it quick and dirty.

I'm still open to feedback on the stylesheet and the view if something really jumps out! My response just might be "this will change" or "I'll revisit after design does their magic".

A few things to note:
- There will always be a _slight_ jump in the UX because we load the scripts for liquid tags asynchronously and we can't initially render the markup dynamically with the "initial state" because we cache the articles. In other words, we are essentially rendering a "default state" (signed out), and once the JavaScript loads up it updates everything as needed. Any jump you see here should improve after design puts their polish on it. Currently, it's very in line and often faster than other areas on the page where we use the same approach.
- This whole tag is still restricted to admins only so it's still not "released" yet.
- This liquid tag (like others that use custom JS) won't "work" in the preview mode because we don't load the scripts there. The preview will always just display the liquid tag's default state. Updating that behavior for liquid tags is outside the scope of this PR.
- I added logic in the controller and the liquid tag to account for users who sign up to DEV using Apple Auth. Apple Auth is not released yet, but it's expected to be around the time this liquid tag is released so I'm accounting for it now. Apple Auth allows users to use what's called a [private relay email address](https://support.apple.com/en-us/HT210425). It's essentially an "alias" that DEV is allowed to email, but we are _not_ allowed to share that email address in any other way. So if a user signs up with Apple and they have this setting turned on, they will need to update their email address on their DEV account before they're allowed to subscribe.

The JavaScript here is going to be, for lack of a better word, pretty brutal. I considered moving it into its own file, but I decided against it. The JS here is treated as a raw string. I considered moving the code into a file and do something like:
```ruby
SCRIPT = File.read('app/javascripts/liquid_tags/user_subscription_tag.js').freeze

def self.script
  SCRIPT
end
```
This introduces a couple more items for discussion. It's been a long outstanding TODO for us to reorganize the JS. Also, while I would obviously add a comment to the top of this file (if this approach were taken), it wouldn't be clear that this file sort of falls outside of the usual asset pipeline. You can't include anything in this file, for example. I felt that got a little confusing. Instead, I chose to stick with the "current setup" of how we're handling liquid tags and re-visit the entire architecture of liquid tags in the future. More on that below.

- [ ] [Create a new role for users approved to use the new user subscription liquid tag](https://github.com/thepracticaldev/dev.to/issues/8165)

https://github.com/thepracticaldev/dev.to/issues/8168

**Setup**
Pull down this code locally and go to `app/controllers/async_info_controller.rb` and add:
```ruby
email: @user.email,
```
anywhere in the [`user_data` hash](https://github.com/thepracticaldev/dev.to/blob/cfa5e882ce26dd984d7a60fbbc26f636c8113d07/app/controllers/async_info_controller.rb#L55-L75). Fire up the app locally.

**Steps**
4. View your Article.
6. Refresh the Article (you should still see confirmation that you're already subscribed).
7. Sign out.
8. Revisit the Article (you should see the "sign-in" UX).
9. Sign back in via the `Sign in` button (you should see confirmation that you're already subscribed).
10. Destroy your `UserSubscription` in the DB (`UserSubscription.destroy_all`).
11. Change the email in `async_info_controller` from the setup setup step to `"test@privaterelay.appleid.com"`.
12. Refresh your article (you should see the note about being Auth'd with Apple).

That's it 🎉 

A reminder the copy and the design are going to change. This is my quick and dirty MVP version. You can check out our team Slack for the mockups and discussions 🌻 .

**Signed in and not subscribed UX**
signed_in_subscribe

**Signed out UX**
signed_out

**Subscription confirmation UX**
subscribe_confirmation

**Subscription success UX**
subscription_confirmation

**Already subscribed UX**
already_subscribed

**Apple Auth UX**
apple_auth_message

## Added tests?
- [x] yes - for the controller updates.
- [x] no - for the liquid tag. I will add tests once the design team is finished modifying the markup.

## Added to documentation?
- [x] no documentation needed

anxious_gif

### DIFF
--- a/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
+++ b/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
@@ -1,0 +1,62 @@
+.ltag__user-subscription {
+  .user-subscription-container {
+    display: flex;
+    justify-content: center;
+    border: 2px solid black;
+    box-shadow: 4px 4px 0 black;
+  }
+
+  .user-subscription-content {
+    margin: 1rem auto;
+  }
+
+  .user-subscription-cta-text {
+    text-align: center;
+  }
+
+  .user-subscription-signed-out {
+    text-align: center;
+  }
+
+  .user-subscription-response {
+    display: none;
+  }
+
+  #sign-in-btn {
+    width: max-content;
+  }
+
+  .user-subscription-signed-in {
+    text-align: center;
+    display: none;
+  }
+
+  #logged-in-text {
+    text-align: center;
+  }
+
+  #user-subscription-email {
+    font-weight: bold;
+  }
+
+  .user-subscription-confirmation-modal {
+    display: none;
+  }
+
+  .crayons-modal__overlay {
+    position: fixed;
+    z-index: 400;
+  }
+
+  .confirmation-buttons {
+    text-align: center;
+  }
+
+  #confirmation-button,
+  #cancel-button {
+    width: max-content;
+    display: inline-block;
+    margin-right: 8px;
+    margin-bottom: 1rem;
+  }
+}
--- a/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
+++ b/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
@@ -35,7 +35,7 @@
     text-align: center;
   }
 
-  #user-subscription-email {
+  .subscriber-email {
     font-weight: bold;
   }
 
--- a/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
+++ b/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
@@ -1,5 +1,5 @@
-.ltag__user-subscription {
-  .user-subscription-container {
+.ltag__user-subscription-tag {
+  .ltag__user-subscription-tag__container {
     display: flex;
     justify-content: center;
     border: 2px solid black;
@@ -14,23 +14,23 @@
     }
   }
 
-  .subscriber-profile-image {
+  .ltag__user-subscription-tag__subscriber-profile-image {
     display: none;
   }
 
-  .user-subscription-content {
+  .ltag__user-subscription-tag__content {
     margin: 1rem auto;
   }
 
-  .user-subscription-cta-text {
+  .ltag__user-subscription-tag__cta-text {
     text-align: center;
   }
 
-  .user-subscription-signed-out {
+  .ltag__user-subscription-tag__signed-out {
     text-align: center;
   }
 
-  .user-subscription-response {
+  .ltag__user-subscription-tag__response {
     display: none;
   }
 
@@ -38,7 +38,7 @@
     width: max-content;
   }
 
-  .user-subscription-signed-in {
+  .ltag__user-subscription-tag__signed-in {
     text-align: center;
     display: none;
   }
@@ -48,7 +48,7 @@
     text-align: center;
   }
 
-  .subscriber-email {
+  .ltag__user-subscription-tag__subscriber-email {
     font-weight: bold;
   }
 
@@ -65,7 +65,7 @@
     z-index: 400;
   }
 
-  .confirmation-buttons {
+  .ltag__user-subscriptino-tag__confirmation-buttons {
     text-align: center;
   }
 
--- a/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
+++ b/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
@@ -56,6 +56,10 @@
     display: none;
   }
 
+  #subscriber-apple-auth {
+    display: none;
+  }
+
   .crayons-modal__overlay {
     position: fixed;
     z-index: 400;
--- a/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
+++ b/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
@@ -43,7 +43,8 @@
     display: none;
   }
 
-  #logged-in-text {
+  #logged-in-text,
+  #apple-auth-message {
     text-align: center;
   }
 
@@ -62,6 +63,12 @@
 
   .confirmation-buttons {
     text-align: center;
+  }
+
+  #sign-in-btn,
+  #subscribe-btn {
+    margin: 0 auto;
+    display: block;
   }
 
   #confirmation-btn,
--- a/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
+++ b/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
@@ -51,7 +51,7 @@
     font-weight: bold;
   }
 
-  .user-subscription-confirmation-modal {
+  #user-subscription-confirmation-modal {
     display: none;
   }
 
@@ -64,8 +64,8 @@
     text-align: center;
   }
 
-  #confirmation-button,
-  #cancel-button {
+  #confirmation-btn,
+  #cancel-btn {
     width: max-content;
     display: inline-block;
     margin-right: 8px;
--- a/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
+++ b/app/assets/stylesheets/ltags/UserSubscriptionTag.scss
@@ -6,6 +6,18 @@
     box-shadow: 4px 4px 0 black;
   }
 
+  #profile-images {
+    text-align: center;
+
+    img {
+      border-radius: 50%;
+    }
+  }
+
+  .subscriber-profile-image {
+    display: none;
+  }
+
   .user-subscription-content {
     margin: 1rem auto;
   }
--- a/app/controllers/user_subscriptions_controller.rb
+++ b/app/controllers/user_subscriptions_controller.rb
@@ -28,6 +28,7 @@ class UserSubscriptionsController < ApplicationController
     end
 
     return error_response("subscriber email mismatch") if subscriber_email_stale?
+    return error_response("cannot subscribe with Apple private relay email") if subscriber_authed_with_apple?
 
     @user_subscription = user_subscription_source.build_user_subscription(current_user)
 
@@ -40,6 +41,10 @@ class UserSubscriptionsController < ApplicationController
   end
 
   private
+
+  def subscriber_authed_with_apple?
+    current_user.email.end_with?("@privaterelay.appleid.com")
+  end
 
   def user_subscription_tag_enabled?(source_type, user_subscription_source)
     liquid_tags =
--- a/app/controllers/user_subscriptions_controller.rb
+++ b/app/controllers/user_subscriptions_controller.rb
@@ -20,11 +20,14 @@ class UserSubscriptionsController < ApplicationController
     is_subscribed = UserSubscriptions::SubscriptionCacheChecker.call(current_user, params[:source_type], params[:source_id])
     author = source.user
 
+    # TODO: (Alex Smith) remove this when ready
+    subscriber_email = current_user.email if current_user.any_admin?
+
     render json: {
       success: true,
       subscriber: {
         is_subscribed: is_subscribed,
-        email: current_user.email
+        email: subscriber_email # TODO: (Alex Smith) change to current_user.email
       },
       author: {
         username: author.username,
--- a/app/controllers/user_subscriptions_controller.rb
+++ b/app/controllers/user_subscriptions_controller.rb
@@ -2,41 +2,6 @@ class UserSubscriptionsController < ApplicationController
   before_action :authenticate_user!
 
   USER_SUBSCRIPTION_PARAMS = %i[source_type source_id subscriber_email].freeze
-  BASE_DATA_PARAMS = %i[source_type source_id].freeze
-
-  def base_data
-    params.require(BASE_DATA_PARAMS)
-    source_type = params[:source_type]
-    return error_response("invalid source_type") unless UserSubscription::ALLOWED_TYPES.include?(source_type)
-
-    source_id = params[:source_id]
-    source = source_type.constantize.includes(:user).find_by(id: source_id)
-    return error_response("source not found") unless active_source?(source_type, source)
-
-    unless user_subscription_tag_enabled?(source_type, source)
-      return error_response("user subscriptions are not enabled for the requested source")
-    end
-
-    is_subscribed = UserSubscriptions::SubscriptionCacheChecker.call(current_user, params[:source_type], params[:source_id])
-    author = source.user
-
-    # TODO: (Alex Smith) remove this when ready
-    subscriber_email = current_user.email if current_user.any_admin?
-
-    render json: {
-      success: true,
-      subscriber: {
-        username: current_user.username,
-        is_subscribed: is_subscribed,
-        email: subscriber_email, # TODO: (Alex Smith) change to current_user.email
-        profile_image_90: ProfileImage.new(current_user).get(width: 90)
-      },
-      author: {
-        username: author.username,
-        profile_image_90: ProfileImage.new(author).get(width: 90)
-      }
-    }, status: :ok
-  end
 
   def subscribed
     params.require(%i[source_type source_id])
--- a/app/controllers/user_subscriptions_controller.rb
+++ b/app/controllers/user_subscriptions_controller.rb
@@ -2,6 +2,36 @@ class UserSubscriptionsController < ApplicationController
   before_action :authenticate_user!
 
   USER_SUBSCRIPTION_PARAMS = %i[source_type source_id subscriber_email].freeze
+  BASE_DATA_PARAMS = %i[source_type source_id author_id].freeze
+
+  def base_data
+    params.require(BASE_DATA_PARAMS)
+    source_type = params[:source_type]
+    return error_response("invalid source_type") unless UserSubscription::ALLOWED_TYPES.include?(source_type)
+
+    source_id = params[:source_id]
+    source = source_type.constantize.includes(:user).find_by(id: source_id)
+    return error_response("source not found") unless active_source?(source_type, source)
+
+    unless user_subscription_tag_enabled?(source_type, source)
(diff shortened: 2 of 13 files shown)


app/assets/stylesheets/ltags/UserSubscriptionTag.scss
```diff
@@ -0,0 +1,85 @@
+.ltag__user-subscription-tag {
```
atsmith813, line 1: You can skip right over this file if you'd like :trollface:. It's all hacky to get the MVP UX running for design to polish up 🎉 .

app/controllers/user_subscriptions_controller.rb
```diff
@@ -28,6 +28,7 @@ def create
     end
 
     return error_response("subscriber email mismatch") if subscriber_email_stale?
+    return error_response("cannot subscribe with Apple private relay email") if subscriber_authed_with_apple?
```
atsmith813, line 31: Codeclimate is complaining because I've added too many returns in here. Let's ignore for it now since I already have another issue slated to re-factor this entire controller before release. I had a feeling it would need a few more tweaks over the next couple of PRs so I postponed the refactor. That, and I need to get this all through so the design team can have at it 🎉 .

app/liquid_tags/user_subscription_tag.rb
```diff
@@ -1,9 +1,242 @@
 class UserSubscriptionTag < LiquidTagBase
+  PARTIAL = "liquids/user_subscription".freeze
+
+  SCRIPT = <<~JAVASCRIPT.freeze
```
atsmith813, line 4: In case you missed it, check out my PR description where I explain why this is set up this way 🙃 .

app/views/liquids/_user_subscription.html.erb
```diff
@@ -0,0 +1,70 @@
+<div class="ltag__user-subscription-tag" id="user-subscription-tag" data-author-username=<%= author_username %>>
```
atsmith813, line 1: Same as above, you can skip right over this file if you'd like :trollface:. It's all hacky to get the MVP UX running for design to polish up 🎉 .

app/labor/markdown_parser.rb
```diff
@@ -18,7 +19,7 @@ def finalize(link_attributes: {})
     sanitized_content = sanitize_rendered_markdown(html)
     begin
       parsed_liquid = Liquid::Template.parse(sanitized_content)
-      html = markdown.render(parsed_liquid.render)
+      html = markdown.render(parsed_liquid.render(source: @source))
```
atsmith813, line 22: In the liquid world, this is known as an [Assignment](https://github.com/Shopify/liquid/wiki/Liquid-for-Programmers#using-assignments-in-templates). It allows us to access other data inside the tag. It is accessible in `context.environments` which I made a helper to quickly access. I'm thinking of this as the "environment the liquid tag is being rendered in" or "the source (`Article`, `Comment`, etc.) where the liquid tag is being rendered".

This enables us to do context (`Article`, `Comment`, etc.) specific rendering of liquid tags. The same liquid tag could render differently in different contexts/environments.

No other liquid tag uses this yet because this the first "context-specific liquid tag". Very likely it won't be the last 🙃 . I may use this same concept here to validate liquid tags and which contexts they're permitted in. TBD in the next PR when I add role restrictions and validations to liquid tags.

If you're like me and happen to peruse this `MarkdownParser` or other files around rending liquid tags, you'll find other areas where we could improve this entire process. I'm open to other avenues! I chose this as the least invasive and most impactful 🙈 .

app/liquid_tags/liquid_tag_base.rb
```diff
@@ -3,6 +3,12 @@ def self.script
     ""
   end
 
+  def detect_in_context(key, context)
+    # Pulls the value of the specified key out of context.environments which is
+    # an Array of Hashes. Otherwise, it returns nil.
+    context.environments.detect { |c| c.key? key }.try(:[], key)
+  end
+
```
atsmith813, line 11: Said helper [mentioned above](https://github.com/thepracticaldev/dev.to/pull/8871/files#r445053974).

`context` will look something like:
```ruby
[{:source => Article}, {}]
```
So my approach is trying to "future proof" it a little bit.

app/liquid_tags/user_subscription_tag.rb
```diff
@@ -1,9 +1,242 @@
...
   def initialize(_tag_name, cta_text, _tokens)
     @cta_text = cta_text.strip
   end
 
-  def render(context); end
+  def render(context)
+    source = detect_in_context(:source, context)
+    author = source&.user
+    author_profile_image = author&.profile_image_90
+    author_username = author&.username
```
atsmith813, line 225: `source` will be `nil` in some circumstances (i.e. Article editor preview). And that's okay, we'll just pass `nil` values to the view and those items just show up "blank" 👍

app/liquid_tags/user_subscription_tag.rb
```diff
@@ -1,9 +1,242 @@
...
+
+  SCRIPT = <<~JAVASCRIPT.freeze
+    function isUserSignedIn() {
+      return document.head.querySelector('meta[name="user-signed-in"][content="true"]') !== null;
+    }
+
+    // Hiding/showing elements
+    // ***************************************
+    function clearSubscriptionArea() {
+      document.getElementById('subscription-signed-in').style.display = 'none';
```
lisasy, line 12: When I start working on this, I'll refactor this so that instead of manipulating styles in this Javascript, we add CSS classes that control its styling.

atsmith813: Love that! 💯

app/liquid_tags/liquid_tag_base.rb
mstruve: Rather than looping through the environments why not just set the key in the context hash? It seems like that would be a lot more accessible according to the docs which show an example of fetching a key from the context.
```ruby
def render(context)
  (@factor * context["multiply_by"].to_f).to_s
end
```

mstruve:
Started playing with this locally and I noticed the confirm dialog pops up when the article page is loaded
https://youtu.be/RCkZoBjNuGY

app/liquid_tags/liquid_tag_base.rb
atsmith813: Agreed - for the life of me, I couldn't get that to work. I tried both ways of setting an assignment:
```ruby
@template = Liquid::Template.parse("{% multiply 5 %}")
@template.assigns["multiply_by"] = 10
@template.render    # => "50"
```
and the current way
```ruby
@template = Liquid::Template.parse("{% multiply 5 %}")
@template.render('multiply_by' => 3)    # => 15
```

Both ways and playing with it, the argument still always shows up in the `context.environments` Array 🤔 . I eventually chalked it up as stale docs since those were last updated in 2016 😬 .

atsmith813: This has me thinking though...I've been playing around in `pry` and the source code for the gem. I'll update this. At a minimum, I shouldn't be passing it an Object as that's not a "permitted type". And `context` has its own built-in `find_variable` method, but won't work with an Object 🤔 .

atsmith813:
@vaidehijoshi yep, you're correct!
> This liquid tag (like others that use custom JS) won't "work" in the preview mode because we don't load the scripts there. The preview will always just display the liquid tag's default state. Updating that behavior for liquid tags is outside the scope of this PR.

In this specific case, we don't have access to the avatar yet. What happens with preview is it just runs the body markdown all by itself through the parser. The parser and liquid tag don't know who the "current user" is and there is no `Article` object yet.

But now that I'm thinking about it again, I might have an idea to fix it now 🤔 .

vaidehijoshi:
> The preview will always just display the liquid tag's default state

Ah, you explained this in the description! I just didn't register that the "default state" was the logged out version :) thanks for clarifying 👍

atsmith813:
@vaidehijoshi I don't think it's going to be best to fix it right now. I was thinking of updating the `MarkdownParser` parser to accept any options and pass them down to the rendering of liquid tags. That could look something like:
```ruby
class MarkdownParser
  def initialize(content, liquid_tag_options: {})
    @content = content
    @liquid_tag_options = liquid_tag_options
  end

  def finalize
    ...
    parsed_liquid = Liquid::Template.parse(sanitized_content)
    html = markdown.render(parsed_liquid.render(liquid_tag_options))
    ...
  end
end
```

Then, in the `Article` model it would be:
```ruby
parsed_markdown = MarkdownParser.new(parsed.content, liquid_tag_options: { author_id: user_id })
```

And in the `ArticlesController#preview` it would be:
```ruby
parsed_markdown = MarkdownParser.new(parsed.content, liquid_tag_options: { author_id: current_user.id })
```
From here, the liquid tag would know in both `preview` and normal rendering who the author is to populate that image.

The problem with this approach is anytime you have a new liquid tag that needs data like this, you'll have to update many references to the parser to pass in data for any and every liquid tag that accepts custom data, just in case the content includes it.

----

Fixing this is tough. There are many ways to go. Loading the JS scripts on the editor would fix it, also not ideal because of caching and performance.

I feel the `MarkdownParser` could use a revamp as a whole and/or breaking out liquid tag handling (keeping in mind the next PR will introduce liquid tag specific validations and permissions) is a possible avenue. And since liquid tags are used anywhere we are accepting and rendering markdown, the solution will have to be super flexible.

In other words, I don't know that we have a "good" way to tell this liquid tag specifically how to behave inside preview that wouldn't brush up against some much bigger changes 🙈 . I really hate that it's _always_ a broken image in preview 😞 . I'm very open to ideas to fix this ASAP since I don't have a plan yet to do that before the final release.

app/liquid_tags/user_subscription_tag.rb
```diff
@@ -1,9 +1,242 @@
...
+      addConfirmationModalClickHandlers();
+
+      // We need access to some DOM elements (i.e. csrf token, article id, userData, etc.)
+      document.addEventListener('DOMContentLoaded', function() {
+        checkIfSubscribed();
+      });
+    } else {
+      addSignInClickHandler();
+    }
+  JAVASCRIPT
```
vaidehijoshi, line 215: feel free to push back on this, but i wonder if it would be worth it to extract this into a separate file and the pull it in and assign it to `SCRIPT`. not majorly important, but because there is _so much_ JS for this tag, i was thinking that it might help with the readability of this model.

app/views/liquids/_user_subscription.html.erb
```diff
@@ -0,0 +1,70 @@
...
+          <div class="fs-s" id="logged-in-text">
+            You'll subscribe with
+            <span class="ltag__user-subscription-tag__subscriber-email"></span>
+            the email address associated with your DEV account. To use a
+            different email address, you can
+            <a href="/settings">update your email address in Settings</a>.
+          </div>
+        </div>
+        <div class="ltag__user-subscription-tag__apple-auth fs-s" id="subscriber-apple-auth">
+          <button disabled class="crayons-btn" id="subscribe-btn">Subscribe</button>
```
vaidehijoshi, line 29: should this button always be disabled? i didn't see any JS to toggle the `disabled` property here so i'm curious about this

app/views/liquids/_user_subscription.html.erb
```diff
@@ -0,0 +1,70 @@
...
+      </div>
+    </div>
+    <div id="user-subscription-confirmation-modal">
+      <div class="crayons-modal crayons-modal--l">
+        <div class="crayons-modal__box">
+          <header class="crayons-modal__box__header">
+            <h2>Confirm subscribe</h2>
+            <button type="button" class="crayons-btn crayons-btn--icon crayons-btn--ghost" id="close-confirmation-modal">
+              <svg width="24" height="24" viewBox="0 0 24 24" class="crayons-icon" xmlns="http://www.w3.org/2000/svg">
+                <path d="M12 10.586l4.95-4.95 1.414 1.414-4.95 4.95 4.95 4.95-1.414 1.414-4.95-4.95-4.95 4.95-1.414-1.414 4.95-4.95-4.95-4.95L7.05 5.636l4.95 4.95z" />
```
vaidehijoshi, line 48: we should run this [through svgo](https://jakearchibald.github.io/svgomg) if we haven't already!

app/assets/stylesheets/ltags/UserSubscriptionTag.scss
vaidehijoshi: _averts eyes knowing that we will refactor this 😄_

app/assets/stylesheets/ltags/UserSubscriptionTag.scss
```diff
@@ -0,0 +1,85 @@
...
+    display: none;
+  }
+
+  #subscriber-apple-auth {
+    display: none;
+  }
+
+  .crayons-modal__overlay {
+    position: fixed;
+    z-index: 400;
```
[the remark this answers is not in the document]
atsmith813: Yeah, this was weird, the default `z-index` of the overlay (the darkened background when the modal pops up) was showing up on top of the modal box itself 🤔 . So this is me lowering it. I've got a mental note to try and get this cut out with design because I also hate `z-index` 🙈 .

app/views/liquids/_user_subscription.html.erb
atsmith813: Yep! I had 2 choices, toggle the existing `subscribe` button in JS or toss in a second one that's always disabled. It's for this UX:
image
Where the button will never be enabled under the "Apple auth UX". Once a user changes their email, they end up in the "Signed in but not subscribed UX" which has the button enabled.

I tossed in a second button because the "Apple Auth UX" was never mocked up so I wanted to leave the Apple chunk untangled from everything else. Putting in JS to toggle the button for the "Signed in but not subscribed UX" and for "Apple Auth UX" would "join/tangle" them so I didn't go that route.

I'll likely move to the JS button toggle depending on where we land with handling "Apple Auth UX".

app/views/liquids/_user_subscription.html.erb
atsmith813: 👀  I copied and pasted it directly from [Crayons](https://dev.to/page/crayons). So if/when we do this post-design we should update that page as well!

app/liquid_tags/user_subscription_tag.rb
atsmith813: I considered moving it into its own file, but I decided against it. The JS here is treated as a raw string. I considered moving the code into a file and do something like:
```ruby
SCRIPT = File.read('app/javascripts/liquid_tags/user_subscription_tag.js').freeze

def self.script
  SCRIPT
end
```
This introduces a couple more items for discussion. It's been a long outstanding TODO for us to reorganize the JS. Also, while I would obviously add a comment to the top of this file (if this approach were taken), it wouldn't be clear that this file would sort of fall outside of the usual asset pipeline. You can't include anything in this file, for example. I felt that got a little confusing. Instead, I chose to stick with the "current setup" of how we're handling liquid tags and re-visit the entire architecture of liquid tags in the future 🙃 .

vaidehijoshi: @atsmith813 That's fair; fwiw, when I left this comment I was thinking of a `File.read` type of thing as well. I think re-visiting it potentially later is fine, I think it's 👍 to stick with convention for now.

app/liquid_tags/liquid_tag_base.rb
atsmith813: @mstruve thanks for keeping me honest with this inquiry 💯 .

**TLDR** - I [switched to register](https://github.com/Shopify/liquid/wiki/Liquid-for-Programmers#difference-between-assigns-and-registers) and removed the looping through environments and those docs/that wiki are definitely stale.

**Details** _for future reference_
`render` [will create a new `Context` object](https://github.com/Shopify/liquid/blob/1feaa6381300d56e2c71b49ad8fee0d4b625147b/lib/liquid/template.rb#L189) which will [`squash_instance_assigns_with_environments`](https://github.com/Shopify/liquid/blob/1feaa6381300d56e2c71b49ad8fee0d4b625147b/lib/liquid/context.rb#L37). So even if you create your own `Context` object to pass to `render`, inside `Context` if you try to use `environments`, `scopes`, or `instance_assigns` they will end up in the `environments` or `scopes` Array. Neither of those is really appropriate here. Instead, I switched to register which sounds more appropriate.

Registers [are set](https://github.com/Shopify/liquid/blob/1feaa6381300d56e2c71b49ad8fee0d4b625147b/lib/liquid/context.rb#L31) instead of the whole Array riff-raff. Also, a [code comment](https://github.com/Shopify/liquid/blob/1feaa6381300d56e2c71b49ad8fee0d4b625147b/lib/liquid/template.rb#L170-L171) points out that registers are
> hash with register variables. Those can be accessed from filters and tags and might be useful to integrate liquid more with its host application.

🎉

atsmith813:
@vaidehijoshi I spoke too soon! I got a nice quick way to fix preview now 🎉  Thanks for the inquiry 🕵️

atsmith813:
@rhymes good question - that is confusing right now 🙈. We don't pass the email. In the setup details in the description I have:
**Setup**
Pull down this code locally and go to `app/controllers/async_info_controller.rb` and add:
```ruby
email: @user.email,
```
anywhere in the [`user_data` hash](https://github.com/thepracticaldev/dev.to/blob/cfa5e882ce26dd984d7a60fbbc26f636c8113d07/app/controllers/async_info_controller.rb#L55-L75). Fire up the app locally.

----

So all of this code assumes it will be there. This is some of the "pain" I was alluding to on the last PR that's caused by removing `email` from `async_info` for now.

The very last PR for this feature for final release will be adding `email` back into `async_info`.

PR #6823, opened 2020-03-24
Add response templates to user settings

Zhao-Andy:
## What type of PR is this? (check all applicable)
- [x] Feature

## Description
This adds response templates to the settings page. Note that it does not make a link visible, but it is technically available via the direct route `/settings/response-templates`.

Users who access this page will be able to complete basic CRUD for personal response templates.

Also removes some dead CSS.

## Related Tickets & Documents
#5419 

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
Screen Shot 2020-03-24 at 5 24 35 PM

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

## [optional] Are there any post deployment tasks we need to perform?
No

## [optional] What gif best describes this PR or how it makes you feel?

Kermit plays the banjo with Fozzie, the two in a car ride together with the text "We're close".

PR #8876, opened 2020-06-24
Fix incorrect styling of fork in repo list

### LINKED ISSUE #8834: Incorrect styling of "fork" in the repo list on the "Integrations" tab
**Describe the bug**

The label "fork" in the GitHub repositories list in the "Integrations" tab of your settings is styled incorrectly.

**To Reproduce**

1. Make sure you have your GitHub account connected
2. Make sure you have  at least one forked repository
3. Go to the "Integrations" tab in your settings
4. Look at the list of repos

**Expected behavior**

The label "fork" is styled correctly. Seems like it originally looked like this:

Correctly styled "fork" label

**Screenshots**

Screen Shot 2020-06-22 at 11 45 08 AM

N/A

N/A

**Additional context**

This bug seems to have been introduced by https://github.com/thepracticaldev/dev.to/pull/6823#discussion_r397473497, which broke the following:

https://github.com/thepracticaldev/dev.to/blob/81fff248dd953f873161ed276ddb7f0ab576bcf8/app/javascript/githubRepos/singleRepo.jsx#L13-L15

anyruizd:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
- Fixes #8834 
- Added back in the piece of code removed in #6823 corresponding to the styles of the `Fork` label in the repo list on the "Integrations" tab. 

Before:
imagen

After:
imagen

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

alt_text

### DIFF
--- a/app/assets/stylesheets/settings.scss
+++ b/app/assets/stylesheets/settings.scss
@@ -35,6 +35,15 @@ $input-width: 650px;
       background: $light-green;
       color: #0f0f0f;
     }
+    .github-repo-fork {
+      margin-left: 5px;
+      font-weight: 400;
+      background: $purple;
+      color: $bold-blue;
+      border-radius: 3px;
+      padding: 1px 6px;
+      font-size: 0.9em;
+    }
     .github-repo-row-name {
       font-weight: bold;
       form {
--- a/app/assets/stylesheets/settings.scss
+++ b/app/assets/stylesheets/settings.scss
@@ -35,20 +35,14 @@ $input-width: 650px;
       background: $light-green;
       color: #0f0f0f;
     }
-    .github-repo-fork {
-      margin-left: 5px;
-      font-weight: 400;
-      background: $purple;
-      color: $bold-blue;
-      border-radius: 3px;
-      padding: 1px 6px;
-      font-size: 0.9em;
-    }
     .github-repo-row-name {
       font-weight: bold;
       form {
         display: block;
         float: right;
+      }
+      .crayons-indicator {
+        margin-left: 5px;
       }
       button,
       a[role='button'] {
--- a/app/javascript/githubRepos/singleRepo.jsx
+++ b/app/javascript/githubRepos/singleRepo.jsx
@@ -11,7 +11,11 @@ export class SingleRepo extends Component {
   forkLabel = () => {
     const { fork } = this.props;
     if (fork) {
-      return <span className="github-repo-fork">fork</span>;
+      return (
+        <span className="crayons-indicator crayons-indicator--accent">
+          fork
+        </span>
+      );
     }
     return null;
   };


app/assets/stylesheets/settings.scss
```diff
@@ -35,6 +35,15 @@ $input-width: 650px;
       background: $light-green;
       color: #0f0f0f;
     }
+    .github-repo-fork {
```
ludwiczakpawel, line 38: Maybe instead of introducing new style we could use one of the existing components called `crayons-indicator`? It's the badge we currently use, for example, as number indicator in header (notifications, connect). I imagine this could be

```
<span class="crayons-indicator crayons-indicator--accent">Fork</span>
```

It would render as something like this:

image

anyruizd: Hey, @ludwiczakpawel thanks for your feedback! 🤗 I can do that! But first I want to make sure we're on the same page: This PR is reverting the changes introduced in this one (#6823) in which the piece of code corresponding to the fork label styling was deleted: so no new styles are being introduced. However, it seems like those styles are a little bit outdated? So, by implementing your feedback I'll be updating them.
Having said that, should I proceed?

ludwiczakpawel: correct! 👌

nickytonline: Definitely will look awesome as an indicator. 🔥

anyruizd: Done! 🚀

app/assets/stylesheets/settings.scss
```diff
@@ -41,6 +41,9 @@ $input-width: 650px;
         display: block;
         float: right;
       }
+      .crayons-indicator {
```
nickytonline, line 44: We want to avoid modifying the CSS classes of our design system even if they are within another element. Instead you can use our margin left utility class `ml-1` which is 4px. We tend to use even numbers for spacing.

So this CSS can be removed, and simply add `ml-1` in the JSX below

`<span className="crayons-indicator crayons-indicator--accent ml-1">`

Or do you think `ml-2` (8px) would be better @ludwiczakpawel? I feel like `ml-1` should be good enough.

We're still building out our design system @anyruizd, so there's some documentation we still need to work on to surface all these utility classes.

anyruizd: @nickytonline Thaanks a lot! I was wondering if that was the right approach 😅 
This is how it looks with `ml-1`
imagen
And this with `ml-2`
imagen
LMK! 😄

nickytonline: I think that looks great, but I'll let @ludwiczakpawel give the final thumbs.

ludwiczakpawel: Both look great :) I'd probably lean towards ml-2 but that's just personal opinion... You decide :)

Thank you so much @anyruizd for looking into this!

nickytonline:
Thanks so much for your PR @anyruizd! 🔥

PR #8880, opened 2020-06-24
Enabled ESLint canned import rules as well as the import order rule (fixable via --fix)

nickytonline:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This is just another change to our ESLint rules. We're still keeping with recommended rules, this time with the [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import) recommended rules.

Recommended import warnings and errors have been put in place as well as an import order rule. The import order rule enforces an order to imports, but it gets automatically fixed via EsLint's CLI arg `--fix`. This means no friction on a contributor to the code base.

So for example, the snippet below would result in an error because an npm package is imported after a local file.

```javascript
import { h, Component } from 'preact';
import { MyComponent } from './MyComponent';
```

Once committed to the code base though, it gets corrected to the snippet below, assuming you ran the precommit hooks of the repository or explicitly ran `eslint --fix` on your file or your IDE's ESLint plugin handled it.

```javascript
import { h, Component } from 'preact';
import { MyComponent } from './MyComponent';
```

image

To test:

* Change the order of some imports in a file like in the example above.
* Commit the file to the branch
* The imports in the JavaScript or JSX file will be reordered automatically.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Actor in a film nodding yes in the woods

### DIFF
--- a/app/javascript/.eslintrc.js
+++ b/app/javascript/.eslintrc.js
@@ -4,6 +4,8 @@ module.exports = {
   parser: 'babel-eslint',
   extends: [
     'eslint:recommended',
+    'plugin:import/errors',
+    'plugin:import/warnings',
     'preact',
     'plugin:jsx-a11y/recommended',
     'prettier',
@@ -38,6 +40,7 @@ module.exports = {
   },
   plugins: ['import', 'react', 'jsx-a11y'],
   rules: {
+    'import/order': ['error'],
     'import/prefer-default-export': 'off',
     'no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
     'jsx-a11y/label-has-associated-control': [


app/javascript/.eslintrc.js
```diff
@@ -4,6 +4,8 @@ module.exports = {
   parser: 'babel-eslint',
   extends: [
     'eslint:recommended',
+    'plugin:import/errors',
```
nickytonline, line 7: The recommended canned sets of rules for the eslint-plugin-import plugin.

app/javascript/.eslintrc.js
```diff
@@ -38,6 +40,7 @@ module.exports = {
   },
   plugins: ['import', 'react', 'jsx-a11y'],
   rules: {
+    'import/order': ['error'],
```
nickytonline, line 43: Enforces import order but is fixable via `eslint --fix` so there is no friction on a contributor.

nickytonline: Not a fan? This isn't a deal breaker for me. I'm curious, if it gets fixed automatically on commit, what's the 👎 for? Too rigid a rule? Also, it could be made a warning.

citizen428: I have no idea where that thumbs-down came from 🤷‍♂️ There may just be too much blood in my coffee stream.

PR #8881, opened 2020-06-24
Hide notification when comment is hidden

Rafi993:

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Hide notification when comment is hidden

closes https://github.com/thepracticaldev/dev.to/issues/8788

removing-hidden-comments

## Added tests?

- [ ] yes
- [X] no, because I need help

I could not find test for this hide action or even the controller itself so I might need little help on that.

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [X] no documentation needed

Happy

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -221,6 +221,11 @@ class CommentsController < ApplicationController
     authorize @comment
     @comment.hidden_by_commentable_user = true
     @comment&.commentable&.update_column(:any_comments_hidden, true)
+
+    Notification.destroy_by(user_id: current_user.id,
+                            notifiable_type: "Comment",
+                            notifiable_id: params[:comment_id])
+
     if @comment.save
       render json: { hidden: "true" }, status: :ok
     else
--- a/spec/requests/comments_spec.rb
+++ b/spec/requests/comments_spec.rb
@@ -332,6 +332,23 @@ RSpec.describe "Comments", type: :request do
 
   describe "PATCH /comments/:comment_id/hide" do
     include_examples "PATCH /comments/:comment_id/hide or unhide", path: "hide", hidden: "true"
+
+    context "with notifications" do
+      let(:user2) { create(:user) }
+      let(:article)  { create(:article, :with_notification_subscription, user: user) }
+      let(:comment)  { create(:comment, commentable: article, user: user2) }
+
+      before do
+        sign_in user
+        Notification.send_new_comment_notifications_without_delay(comment)
+      end
+
+      it "Delete notification when comment is hidden" do
+        notification = user.notifications.last
+        patch "/comments/#{comment.id}/hide", headers: { HTTP_ACCEPT: "application/json" }
+        expect(Notification.exists?(id: notification.id)).to eq(false)
+      end
+    end
   end
 
   describe "PATCH /comments/:comment_id/unhide" do


db/schema.rb
```diff
@@ -688,6 +688,7 @@
   create_table "notifications", id: :serial, force: :cascade do |t|
     t.string "action"
     t.datetime "created_at", null: false
+    t.boolean "hidden", default: false
```
Rafi993, line 691: I named this column **hidden** while other columns are named **any_comments_hidden**  since there are other notifications that are not comments too.

db/migrate/20200623174413_hide_notifications.rb
```diff
@@ -0,0 +1,5 @@
+class HideNotifications < ActiveRecord::Migration[6.0]
+  def change
+    add_column :notifications, :hidden, :boolean, default: false
```
Rafi993, line 3: I don't know If I should also write data migration too?

app/controllers/comments_controller.rb
```diff
@@ -232,7 +237,12 @@ def unhide
     @comment = Comment.find(params[:comment_id])
     authorize @comment
     @comment.hidden_by_commentable_user = false
-    if @comment.save
+
+    # unhide notification related to that comment
+    @comment_notification = Notification.find_by(notifiable_type: "Comment", user_id: current_user.id, notifiable_id: params[:comment_id])
+    @comment_notification.hidden = false
+
+    if @comment.save && @comment_notification.save
```
Rafi993, line 245: Should I perform this inside transaction or is this ok?

app/controllers/comments_controller.rb
```diff
@@ -221,7 +221,12 @@ def hide
     authorize @comment
     @comment.hidden_by_commentable_user = true
     @comment&.commentable&.update_column(:any_comments_hidden, true)
-    if @comment.save
+
+    # Hide notification related to that comment
+    @comment_notification = Notification.find_by(notifiable_type: "Comment", user_id: current_user.id, notifiable_id: params[:comment_id])
```
Rafi993, line 226: I could not find existing test for this hide action or even the controller itself so I might need little help on that.

db/migrate/20200623174413_hide_notifications.rb
Rafi993: Adding this column will also provide ability to hide any notifications in the future.

Rafi993:
> @Rafi993 thank you so much for the PR!!! A little background on our notifications, we have a lot of them and because of that we periodically will remove all notifications that are older than 4 months since at that point we don't feel like they are very useful. With that said, I think a more performant and simpler approach here would be to just delete the notification all together rather than hiding it. That eliminates the need for another column on this extremely large table and simplifies the code quite a bit since once the notification is gone we dont have to worry about filtering it out anywhere else.
> 
> What do you think?

You are right it will simplify the code. I will do that change.

app/controllers/comments_controller.rb
```diff
@@ -221,7 +221,12 @@ def hide
     authorize @comment
     @comment.hidden_by_commentable_user = true
     @comment&.commentable&.update_column(:any_comments_hidden, true)
-    if @comment.save
+
+    @comment_notification = Notification.find_by(notifiable_type: "Comment",
+                                                 user_id: current_user.id,
+                                                 notifiable_id: params[:comment_id])
+
+    if @comment.save && @comment_notification.destroy
```
Rafi993, line 229: I just realized that I need to handle the case where the notification does not exist anymore.

app/controllers/comments_controller.rb
```diff
@@ -221,6 +221,13 @@ def hide
     authorize @comment
     @comment.hidden_by_commentable_user = true
     @comment&.commentable&.update_column(:any_comments_hidden, true)
+
+    @comment_notification = Notification.find_by(notifiable_type: "Comment",
```
citizen428, line 225: This probably doesn't need to be an instance variable, right? We can make it a local variable or just get rid of it completely:


```suggestion
Notification.find_by(
  notifiable_type: "Comment",
  Notification.find_by(notifiable_type: "Comment",
  notifiable_id: params[:comment_id]
)&.destroy
```

citizen428:
Thank you @Rafi993!

app/controllers/comments_controller.rb
```diff
@@ -221,6 +221,11 @@ def hide
     authorize @comment
     @comment.hidden_by_commentable_user = true
     @comment&.commentable&.update_column(:any_comments_hidden, true)
+
+    Notification.find_by(notifiable_type: "Comment",
+                         user_id: current_user.id,
+                         notifiable_id: params[:comment_id])&.destroy
```
rhymes, line 227: What about switching to the new Rails 6 method `destroy_by` ?

```ruby
Notification.destroy_by(user_id: current_user.id, notifiable_type: "Comment", notifiable_id: params[:comment_id])
```

Rafi993:
> Thanks @Rafi993!
> 
> Is there a way you can write a test for this?

I could not find test for this hide action or even the controller itself so I might need little help on that.

rhymes:
@Rafi993 sure! 

It looks like there's a a test here

https://github.com/thepracticaldev/dev.to/blob/ab4753079c3849730729256c22c774f25172d604/spec/requests/comments_spec.rb#L333-L335

which includes a shared example

https://github.com/thepracticaldev/dev.to/blob/master/spec/requests/shared_examples/comment_hide_or_unhide_request.rb

I would probably add a case either to comments_spec or there to test that the notification gets destroyed

Notifications are sent with `Notification.send_new_comment_notifications_without_delay(code)` in the code

Let me know if you need further help!

Rafi993:
> @Rafi993 sure! 
> 
> It looks like there's a a test here
> 
> https://github.com/thepracticaldev/dev.to/blob/ab4753079c3849730729256c22c774f25172d604/spec/requests/comments_spec.rb#L333-L335
> 
> which includes a shared example
> 
> https://github.com/thepracticaldev/dev.to/blob/master/spec/requests/shared_examples/comment_hide_or_unhide_request.rb
> 
> I would probably add a case either to comments_spec or there to test that the notification gets destroyed
> 
> Notifications are sent with `Notification.send_new_comment_notifications_without_delay(code)` in the code
> 
> Let me know if you need further help!

Thanks. I'll write the test.

spec/requests/comments_spec.rb
```diff
@@ -332,6 +332,24 @@
...
   describe "PATCH /comments/:comment_id/hide" do
     include_examples "PATCH /comments/:comment_id/hide or unhide", path: "hide", hidden: "true"
+
+    context "with notifications" do
+      let(:commentable_author) { create(:user) }
+      let(:article) { create(:article, user: commentable_author) }
+      let(:parent_comment) { create(:comment, commentable: article, user: commentable_author) }
+      let(:notification) do
+        Notification.send_new_comment_notifications_without_delay(parent_comment)
+      end
```
Rafi993, line 342:
```ruby
     Notification.send_new_comment_notifications_without_delay(parent_comment)
```
 Always seems to return nil hence this test passes always. I think I might need help to understand this.

rhymes: Hi @Rafi993, that method doesn't return the notification object, you can get it by fetching it from the user's notification:

1. create comment
1. create notification
1. fetch the notification from the user's `.notifications` relation
1. hide comment
1. check the notification doesn't exist in anymore (eg. `Notification.exists?(id: notification.id)` should fail)

Rafi993: > 5\. Notification.exists?(id: notification.id)

Thanks @rhymes I have fixed the test now. I'll squash my commits too.

app/controllers/comments_controller.rb
```diff
@@ -221,6 +221,11 @@ def hide
     authorize @comment
     @comment.hidden_by_commentable_user = true
     @comment&.commentable&.update_column(:any_comments_hidden, true)
+
+    Notification.destroy_by(user_id: current_user.id,
```
mstruve, line 225: TIL about `destroy_by` 🎉

spec/requests/comments_spec.rb
rhymes: @Rafi993 thanks! ps. no need to squash them in the future as we squash when we merge ;)

app/controllers/comments_controller.rb
```diff
@@ -221,6 +221,11 @@ def hide
...
+                            notifiable_type: "Comment",
```
Rafi993, line 226: Hello there,
  I just saw this code segment in **app/models/comment.rb** thanks to this comment https://forem.dev/msarit/comment/f8 thanks to @msarit 

```ruby
  after_update :update_descendant_notifications, if: :deleted
  after_update :remove_notifications, if: :deleted
```

Is it possible that I can write another after_update to remove notifications if it is hidden making the code more clean?

Rafi993: I just remembered this and commented here. I'll create separate Issue if it is needed.

Rafi993: I'll create a separate issue for this :)

PR #8883, opened 2020-06-24
[ModCenter v1] Fix bugs post-launch

msarit:

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
After launching ModCenter v1, we encountered the following bugs:
- Article titles were blacked-out in dark themes
- Articles without tags showed a solitary `#` mark
- Articles with long titles wrapped around (per [ModCenter specs](https://www.figma.com/file/Jgq2Qg3MqmjQ4uVdJtyAsS/Moderation?node-id=419%3A6974), long titles should be truncated)

This PR fixes these bugs.

Fire up your local env, navigate to `/mod` and ensure you have some articles with (1) long titles (2) no tags

Screen Shot 2020-06-24 at 9 27 54 AM

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [X] no documentation needed

Cartoon fixing car


app/javascript/modCenter/singleArticle/singleArticle.test.jsx
```diff
@@ -16,11 +17,28 @@ const testArticle = {
...
+  publishedAt: '2019-06-24T09:32:10.590Z',
+  cachedTagList: '',
+  user: {
+    articles_count: 1,
+    name: 'howdy',
+  },
+};
+
+const truncatedTitle =
+  'An article title that is quite very actually rather extre...';
```
Zhao-Andy, line 35: I think it makes sense to put this in the test itself and not outside of the scope.

app/javascript/modCenter/singleArticle/singleArticle.test.jsx
```diff
@@ -29,15 +47,32 @@ describe('<SingleArticle />', () => {
...
+    render(
+      <SingleArticle
+        id={article.id}
+        title={article.title}
+        path={article.path}
+        publishedAt={article.publishedAt}
+        cachedTagList={article.cachedTagList}
+        user={article.user}
+      />,
+    );
```
Zhao-Andy, line 60: You can reuse the `renderSingleArticle` here and pass in `testArticle2` :) The `(article = testArticle2)` here is a default argument.

app/javascript/modCenter/singleArticle/singleArticle.test.jsx
```diff
@@ -29,15 +47,32 @@ describe('<SingleArticle />', () => {
...
   });
 
   it('renders the article title', () => {
     const { getByText } = renderSingleArticle();
-    getByText(testArticle.title);
+    getByText(testArticle1.title);
+  });
+
+  it('truncates the article title if longer than 60 characters', () => {
+    const { getByText } = renderSingleArticle2();
```
Zhao-Andy, line 74: So for example here you can do `renderSingleArticle(testArticle2)`

app/javascript/modCenter/singleArticle/index.jsx
```diff
@@ -31,16 +31,22 @@ export default class SingleArticle extends Component {
     const { articleOpened } = this.state;
     const { id, title, publishedAt, cachedTagList, user, key } = this.props;
     const tags = cachedTagList.split(', ').map((tag) => {
-      return (
-        <span className="mod-article-tag fs-s ff-accent lh-base" key={key}>
-          <span className="article-hash-tag">#</span>
-          {tag}
-        </span>
-      );
+      if (tag) {
```
Zhao-Andy, line 34: Simple, nice!

app/javascript/modCenter/singleArticle/singleArticle.test.jsx
msarit: ok!

app/javascript/modCenter/singleArticle/singleArticle.test.jsx
msarit: ahhh yes! I don't my brain was firing on all cylinders at this point 😅

app/javascript/modCenter/singleArticle/singleArticle.test.jsx
msarit: 👍

app/javascript/modCenter/singleArticle/index.jsx
msarit: thanks!

nickytonline, line 34: The `if (tag) {` shouldn't be necessary. If we're mapping over an items, it will exist.

app/javascript/modCenter/singleArticle/index.jsx
```diff
@@ -31,16 +31,22 @@ export default class SingleArticle extends Component {
...
+            <span className="article-hash-tag">#</span>
+            {tag}
+          </span>
+        );
+      }
     });
 
     const newAuthorNotification = user.articles_count <= 3 ? '👋 ' : '';
 
+    const truncateTitle = (title, limit) => {
```
nickytonline, line 46: You can use CSS to do this. See https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

app/javascript/modCenter/singleArticle/singleArticle.test.jsx
```diff
@@ -29,15 +47,32 @@ describe('<SingleArticle />', () => {
       />,
     );
 
-  it('should have no a11y  violations', async () => {
+  const renderSingleArticle2 = (article = testArticle2) =>
```
nickytonline: Instead of making the `testArticle2` a default parameter, consider just doing all this within the test. Also, `testArticle2` can be mutated as well as `testArticle1`. Consider creating a `getArticle1()` (needs a better name) etc. so that each test gets it's own test article.

app/javascript/modCenter/singleArticle/index.jsx
msarit: Right Nick, I agree. The issue was, with an empty `tag` list, the function still returned a solitary `#` on account of this span in the `return` block: `<span className="article-hash-tag">#</span>`

If there is a more elegant solution, I would like to implement it 😄

app/javascript/modCenter/singleArticle/singleArticle.test.jsx
msarit: 👍

app/javascript/modCenter/singleArticle/index.jsx
```diff
@@ -31,16 +31,22 @@ export default class SingleArticle extends Component {
...
+      return title.length <= limit ? title : `${title.slice(0, limit - 3)}...`;
+    };
```
Zhao-Andy:
```suggestion
```

Adding a suggestion because I can't push to the branch.

app/javascript/modCenter/singleArticle/index.jsx
```diff
@@ -49,7 +55,9 @@ export default class SingleArticle extends Component {
       >
         <span className="article-title">
           <header>
-            <h3 className="fs-base fw-bold lh-tight">{title}</h3>
+            <h3 className="fs-base fw-bold lh-tight">
+              {truncateTitle(title, 60)}
+            </h3>
```
Zhao-Andy:
```suggestion
            <h3 className="fs-base fw-bold lh-tight">{title}</h3>
```

app/javascript/modCenter/singleArticle/index.jsx
Zhao-Andy: Yeah, not sure how to work around this either without having the `if (tag)` 🤔

Zhao-Andy: Yeah, not sure how to work around this either without having the `if (tag)` 🤔

app/javascript/modCenter/singleArticle/index.jsx
Zhao-Andy: Done 👍 was a bit tricky with grid, TLDR I did this by adding `min-width: 0` the `article-title` grid element.

Reference:
https://stackoverflow.com/questions/44430740/how-do-i-set-text-overflow-to-ellipsis-in-css-grid-layout
https://css-tricks.com/preventing-a-grid-blowout/

app/javascript/modCenter/singleArticle/index.jsx
Zhao-Andy: Ah okay, I figured out why we should stick with `if (tag)`.  If an article has no tags, `article.cachedTagList` will equal `['']`. Therefore, checking `if (tag)` seems like the most performant way to check the validity of that tag.

app/javascript/modCenter/singleArticle/singleArticle.test.jsx
```diff
@@ -16,11 +17,25 @@ const testArticle = {
...
+  },
+};
+
 describe('<SingleArticle />', () => {
-  const renderSingleArticle = (article = testArticle) =>
+  const renderSingleArticle = (article = testArticle1) =>
     render(
       <SingleArticle
-        id={article.title}
+        id={article.id}
```
mstruve, line 38: whoops! 😅

app/assets/stylesheets/moderators.scss
```diff
@@ -41,6 +41,15 @@
...
         grid-template-columns: 4fr 2fr 1fr;
+        .article-title {
+          color: var(--base-90);
+          min-width: 0;
+          header > h3 {
+            text-overflow: ellipsis;
+            overflow: hidden;
+            white-space: nowrap;
+          }
+        }
```
ludwiczakpawel, line 52: > Articles with long titles wrapped around (per ModCenter specs, long titles should be truncated)

Hmmm, is this a good idea? I mean, why would we want to truncate titles in the first place?

Also, since this mod center table has some responsiveness issues, the title area can be very short so the truncated title would be very short too...

app/javascript/modCenter/singleArticle/index.jsx
```diff
@@ -31,12 +31,14 @@ export default class SingleArticle extends Component {
...
-          <span className="article-hash-tag">#</span>
-          {tag}
-        </span>
-      );
+      if (tag) {
+        return (
+          <span className="mod-article-tag fs-s ff-accent lh-base" key={key}>
+            <span className="article-hash-tag">#</span>
+            {tag}
+          </span>
```
ludwiczakpawel, line 39: We have this guy that could be reused:

```suggestion
          <span className="crayons-tag" key={key}>
            <span className="crayons-tag__prefix">#</span>
            {tag}
          </span>
```

app/javascript/modCenter/singleArticle/index.jsx
nickytonline: `article.cachedTagList` is a string though, so if you check if it's an empty string you can skip the string split/mapping altogether. I made it a ternary but it doesn't have to be.

```jsx
    const tags = !cachedTagList ? null : cachedTagList.split(', ').map((tag) => {
        return (
          <span className="mod-article-tag fs-s ff-accent lh-base" key={key}>
            <span className="article-hash-tag">#</span>
            {tag}
          </span>
        );
    });
```

app/assets/stylesheets/moderators.scss
nickytonline: The truncation was being done in JS before. I suggested if you're going to truncate to do it in CSS.

app/javascript/modCenter/singleArticle/index.jsx
Zhao-Andy: Ohhh smart. Will update!

app/assets/stylesheets/moderators.scss
ludwiczakpawel: Fair. Still I have doubts if truncation is needed at all. cc @lisasy

app/javascript/modCenter/singleArticle/index.jsx
Zhao-Andy: Ooh good catch, thanks! Btw we'll update this to use the Crayons table component in the future, maybe cooldown? 😄

app/assets/stylesheets/moderators.scss
Zhao-Andy: I see if your point about not truncating. I think it looks okay not truncated:

Screen Shot 2020-06-25 at 11 49 20 AM

nickytonline: Yeah wrapping seems fine. I thought it was being cut off originally.

app/javascript/modCenter/singleArticle/singleArticle.test.jsx
```diff
@@ -1,10 +1,11 @@
+/* eslint-disable jest/expect-expect */
```
nickytonline, line 1: This won't be necessary once this #8919 is merged.

benhalpern:
Fix the conflicts and looks good to me

Zhao-Andy:
@ludwiczakpawel mind approving so I can merge once build passes? 😁

PR #8887, opened 2020-06-24
Refactor: Remove Database Cleaner gem in favor of Rails 6 Database tasks

mstruve:
- [x] Refactor

## Description
In #8865 I added the DatabaseCleaner gem to help us truncate tables after some specs. @rhymes then [suggested that we make use of Rails 6 database tasks](https://github.com/thepracticaldev/dev.to/pull/8865#pullrequestreview-436373877) to accomplish this. In the initial run we ran into a connection error on Travis so I choose to merge the original change and debug the failure. I could not recreate the failure locally or on Travis so I think it was something intermittent. 

alt_text


rhymes:
@mstruve this never worked :-( ?

mstruve:
I never could figure it out and other things came up so for now I closed this. I would like to get it to work at some point or maybe @snackattas can take a whack at it,

PR #8888, opened 2020-06-24
Blackboxed node_modules when debugging jest tests in VS Code

nickytonline:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [x] DX

## Description

Introduces a small change when debugging jest tests in VS Code. Now all npm packages are blackboxed so they do not get stepped into while debugging.

N/A

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Animation of a black box boxing another black box and goes on and on

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -75,7 +75,8 @@
       "internalConsoleOptions": "neverOpen",
       "windows": {
         "program": "${workspaceFolder}/node_modules/jest/bin/jest"
-      }
+      },
+      "skipFiles": ["node_modules/*"]
     }
   ]
 }


nickytonline:
> > all npm packages are blackboxed so they do not get stepped into while debugging.
> 
> This sounds _so great and helpful_! I'm excited to never step into a node module ever again. I guess the only way that this could cause issues is if there was an actual bug or error coming _from_ a npm package though, right? Since that would be obfuscated and hidden while debugging...

That's correct. In that case, I would probably comment it out at that point, but I would say skipping would be the general scenario. The stack trace would also show if an error occured in a node_module.

PR #8891, opened 2020-06-24
Feature: Setup Vault and AppSecrets Wrapper Class

mstruve:
- [x] Feature

## Description
This PR introduces Vault into our application. Vault will allow us to store and manage secrets in a single secure location no matter how many applications we need to support. I added a wrapper class for it, `AppSecrets`. This wrapper not only talks to Vault, but will also fall back on `ApplicationConfig` if Vault is not setup or a key is not found in Vault. This PR simply sets up the framework for using vault to fetch secrets. We still need to figure out our process for setting Vault up on a per-app basis. 

To ensure an app can only access it's own secrets we will be using [Vault policies](https://learn.hashicorp.com/vault/getting-started/policies#overview). Each app will get its own policy and each policy comes with its own token to give the app access to ONLY its secrets. That token is what will be set as `VAULT_TOKEN`

https://github.com/thepracticaldev/dev.to/projects/9#card-38621281

Without any setup ensure the application console and UI loads. This feature is going to be optional so we don't want it hindering anyone who doesn't want to use it. 

## Added tests?
- [x] yes

alt_text


app/lib/app_secrets.rb
```diff
@@ -0,0 +1,19 @@
...
+    result = Vault.kv(namespace).read(key)&.data&.fetch(:value) if ApplicationConfig["VAULT_TOKEN"].present?
+    result ||= ApplicationConfig[key]
+
+    result
+  rescue Vault::VaultError
+    ApplicationConfig[key]
+  end
+
+  def self.[]=(key, value)
+    Vault.kv(namespace).write(key, value: value)
```
rhymes, line 12: should this bail as well if VAULT_TOKEN is not set?

app/lib/app_secrets.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def self.namespace
+    ENV["VAULT_SECRET_NAMESPACE"]
```
rhymes, line 16: this is not defined in the `Envfile`, correct?

mstruve: Nothing is in the Envfile anymore bc I dont what any of them required in production. I issued a quick follow up PR to fix that #8895 when I forgot that Envied requires everything regardless of default in Prod

app/lib/app_secrets.rb
mstruve: I wasn't sure how I wanted to handle this honestly bc I am not sure how/if I will use this in production. I thought about raising an error or returning nil without vault. I think this is going to evolve as our Vault usage evolves.

PR #8895, opened 2020-06-24
Remove Vault ENV variables from Envfile so they arent Required in Prod

mstruve:
Envied is setup so that anything we add to the Envfile is REQUIRED to be present in Production which is not what I want for option Vault variables. I always forget this detail 🤦‍♀️  This removes the from the file so the app can load without them present. 

alt_text


mstruve:
@rhymes Totally!!!! I wanted to sort out more how it was going to work in terms of setting vault values before writing up doc on it but its on my list! https://github.com/thepracticaldev/dev.to/projects/9#card-40734895

PR #8896, opened 2020-06-24
Add traffic source columns to ahoy visits

vaidehijoshi:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

When we initially added ahoy in https://github.com/thepracticaldev/dev.to/pull/8758, we decided to limit what data we would track. This was fine (arguably really good, actually!) to start. However, I'm still seeing an inexplicably high number of `Ahoy::Visit`s being recorded [here](https://dev.to/internal/blazer/queries/101-ahoy-visits-per-day).

I've tried [all](https://github.com/thepracticaldev/dev.to/pull/8782) [of](https://github.com/thepracticaldev/dev.to/pull/8872) [the](https://github.com/thepracticaldev/dev.to/pull/8839) [recommended](https://github.com/thepracticaldev/dev.to/pull/8866) configurations to turn off visit tracking, so now I am taking a different approach. I'm adding some columns to track a visit's `landing_page`, `referrer`, and `referring_domain`.

These are [optional ahoy columns](https://github.com/ankane/ahoy/blob/b8aa52946bb533df68b2b83d17975631095263b7/test/internal/db/schema.rb#L15-L17), which I am adding back in now very intentionally to try to figure out where on earth these "unexplained" visits are coming from. 🤔 

See description above for linked PRs.

This just adds some columns, so the only thing to really QA here is to make sure that my migration can "migrate" successfully! :) 

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

N/A

### DIFF
--- a/db/migrate/20200624193436_add_traffic_source_columns_to_ahoy_visits.rb
+++ b/db/migrate/20200624193436_add_traffic_source_columns_to_ahoy_visits.rb
@@ -1,0 +1,11 @@
+class AddTrafficSourceColumnsToAhoyVisits < ActiveRecord::Migration[6.0]
+  def change
+    # strong_migrations cannot guarantee safety ofwhat happens inside a change_table
+    # block, so we need to explicitly disable the BulkChangeTable rule here.
+    # rubocop:disable Rails/BulkChangeTable
+    add_column :ahoy_visits, :referrer, :text
+    add_column :ahoy_visits, :referring_domain, :string
+    add_column :ahoy_visits, :landing_page, :text
+    # rubocop:enable Rails/BulkChangeTable
+  end
+end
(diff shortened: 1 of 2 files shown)


mstruve:
How much of a problem are these rogue visits? Are we getting tons of them or just a few? I would really rather not add columns to a database table for debugging purposes. I would much rather use logs which are a bit easier to remove when we are done.

vaidehijoshi:
@mstruve I agree with you re: not wanting to add columns just for debugging, but I apparently forgot/did not have access to logs at all so went this route. Ended up doing some log debugging with @atsmith813, and I don't think there is anything that we want to save with these columns that _isn't already available_ in the logs. It was just a matter of me having access to them :) 

I'm going to go ahead and close this now.

PR #8899, opened 2020-06-24
Add @honeybadger-io/webpack for source-map upload

maestromac:
- [x] DX

## Description
This will only upload our Preact portion source map to help debug errors we see on Honeybadger. The code within `app/assets/javascripts` will require different solution

https://github.com/thepracticaldev/dev.to/projects/9#card-38542511

Followed the instruction on 
1. https://github.com/honeybadger-io/honeybadger-webpack#rails-webpacker-configwebpackenvironmentjs
2. https://docs.honeybadger.io/lib/javascript/guides/using-source-maps.html#uploading-your-source-map

1. QA on production ☠️ 

## Added tests?

## Added to documentation?

- [x] no documentation needed

There shouldn't be. Let's hope it works on the first try 🤞

### DIFF
--- a/config/webpack/environment.js
+++ b/config/webpack/environment.js
@@ -1,6 +1,7 @@
 /* global require module */
 const path = require('path');
 const { environment } = require('@rails/webpacker');
+const HoneybadgerSourceMapPlugin = require('@honeybadger-io/webpack')
 const erb = require('./loaders/erb');
 
 /*
@@ -51,5 +52,17 @@ environment.splitChunks((config) => {
 environment.loaders.delete('nodeModules');
 
 environment.loaders.append('erb', erb);
+
+const revision = process.env.SOURCE_VERSION || 'master'
+
+environment.plugins.append(
+  'HoneybadgerSourceMap',
+  new HoneybadgerSourceMapPlugin({
+    apiKey: process.env.HONEYBADGER_API_KEY,
+    assetsUrl: process.env.ASSETS_URL,
+    silent: false,
+    ignoreErrors: false,
+    revision: revision
+  }))
 
 module.exports = environment;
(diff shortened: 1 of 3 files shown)


config/webpack/environment.js
```diff
@@ -52,4 +53,16 @@ environment.loaders.delete('nodeModules');
...
 
+if (process.env.HONEYBADGER_API_KEY) {
+  environment.plugins.append(
+    'HoneybadgerSourceMap',
+    new HoneybadgerSourceMapPlugin({
+      apiKey: process.env.HONEYBADGER_API_KEY,
+      assetsUrl: `${process.env.APP_PROTOCOL}${process.env.APP_DOMAIN}/packs/js`,
+      silent: false,
+      ignoreErrors: false,
+      revision: process.env.SOURCE_VERSION || 'master'
```
mstruve, line 64: I think using a git commit SHA here would be better since that will be easier for folks to link back to Github where as only a few of us have access to Heroku source versions. To get the git commit you want to access `HEROKU_SLUG_COMMIT`

maestromac: Oh thanks! I honestly thought `SOURCE_VERSION` was going to yield me the commit hash
```suggestion
      revision: process.env.HEROKU_SLUG_COMMIT || 'master'
```

maestromac:
@nickytonline yep, they also provide a section to show uploaded source maps.

config/webpack/environment.js
```diff
@@ -52,4 +53,16 @@ environment.loaders.delete('nodeModules');
 
 environment.loaders.append('erb', erb);
 
+if (process.env.HONEYBADGER_API_KEY) {
```
rhymes, line 56: should this be moved to `production.js` or do we want both dev and test to potentially upload this as well?

PR #8917, opened 2020-06-25
Add contexts and permissions to liquid tags

atsmith813:
- [x] Feature
- [x] Optimization

## Description
**I'm looking for some feedback on this approach before I add tests, docs, etc**

This PR adds the ability to restrict individual liquid tags by user roles and contexts.
- User roles: liquid tags can be restricted for use by users only with the defined roles.
- Contexts: I'm thinking of the context of a liquid tag as being **where** the liquid tag is being used (i.e. `Article`, `Comment`, etc.).

This is needed to restrict the new `UserSubscription` liquid tag for use by approved organizations and only in articles for CodeLand. It's also applicable for future "custom" liquid tags.

- [x] [Make liquid tags role and context specific](https://github.com/thepracticaldev/dev.to/issues/8165)
- [ ] Clean up tickets (will be a separate checklist)

https://github.com/thepracticaldev/dev.to/issues/8165

**QA Steps**

_Setup_
Grab a user record that you can sign in with. For me, that's usually `User.last`.
```ruby
user = User.last
# or
user = User.find_by(email: 'emailaddressofyourlocaluser@hey.com')
```

_Role restriction steps_
1. Destroy all existing roles - `user.roles.destroy_all`.
2. Go to the editor [http://localhost:3000/new](http://localhost:3000/new) and add the new `UserSubscriptionTag` by typing `{% user_subscription Some sweet CTA text %}` in the body of the Article.
3. Click **Preview**. You should see an error saying `"User is not permitted to use this liquid tag"`.
4. In a console add a role - `user.add_role(:admin)`.
5. Refresh your article draft and click **Preview** again, you should now see the preview 🎉 .

_Context restriction steps_
1. Make sure your user has an appropriate role - `user.add_role(:admin)`.
2. Navigate to any article locally.
3. Write up a comment that includes the restricted liquid tag by typing `{% user_subscription Some sweet CTA text %}` in the comment.
4. Click **Preview**. You should see an error saying `"Invalid context. This liquid tag can only be used in Articles."`.

## Added tests?
- [x] yes
_I will add UserSubscriptionTag specs in a future PR once the markup is completed by design. I'll need to reference elements in the view to test all the logic of this tag._

## Added to documentation?
- [x] docs.dev.to

liquid_gif

### DIFF
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -87,7 +87,7 @@ class ArticlesController < ApplicationController
     begin
       fixed_body_markdown = MarkdownFixer.fix_for_preview(params[:article_body])
       parsed = FrontMatterParser::Parser.new(:md).call(fixed_body_markdown)
-      parsed_markdown = MarkdownParser.new(parsed.content, source: Article.new(user: current_user), user: current_user)
+      parsed_markdown = MarkdownParser.new(parsed.content, source: Article.new, user: current_user)
       processed_html = parsed_markdown.finalize
     rescue StandardError => e
       @article = Article.new(body_markdown: params[:article_body])
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -87,7 +87,7 @@ class ArticlesController < ApplicationController
     begin
       fixed_body_markdown = MarkdownFixer.fix_for_preview(params[:article_body])
       parsed = FrontMatterParser::Parser.new(:md).call(fixed_body_markdown)
-      parsed_markdown = MarkdownParser.new(parsed.content, source: Article.new(user: current_user))
+      parsed_markdown = MarkdownParser.new(parsed.content, source: Article.new(user: current_user), user: current_user)
       processed_html = parsed_markdown.finalize
     rescue StandardError => e
       @article = Article.new(body_markdown: params[:article_body])
--- a/app/errors/liquid_tags.rb
+++ b/app/errors/liquid_tags.rb
@@ -1,0 +1,9 @@
+module LiquidTags
+  module Errors
+    class Error < StandardError
+    end
+
+    class InvalidParsedContext < Error
+    end
+  end
+end
--- a/app/errors/liquid_tags.rb
+++ b/app/errors/liquid_tags.rb
@@ -5,7 +5,7 @@ module LiquidTags
 
     # ParseContexts are options passed to initialize on a LiquidTag.
     # An error is raised if any of those options are invalid.
-    class InvalidParsedContext < Error
+    class InvalidParseContext < Error
     end
   end
 end
--- a/app/errors/liquid_tags.rb
+++ b/app/errors/liquid_tags.rb
@@ -3,6 +3,8 @@ module LiquidTags
     class Error < StandardError
     end
 
+    # ParseContexts are options passed to initialize on a LiquidTag.
+    # An error is raised if any of those options are invalid.
     class InvalidParsedContext < Error
     end
   end
--- a/app/labor/markdown_parser.rb
+++ b/app/labor/markdown_parser.rb
@@ -4,7 +4,7 @@ class MarkdownParser
 
   WORDS_READ_PER_MINUTE = 275.0
 
-  def initialize(content, source: {}, user: nil)
+  def initialize(content, source: nil, user: nil)
     @content = content
     @source = source
     @user = user
--- a/app/labor/markdown_parser.rb
+++ b/app/labor/markdown_parser.rb
@@ -96,7 +96,8 @@ class MarkdownParser
 
     cleaned_parsed = escape_liquid_tags_in_codeblock(@content)
     tags = []
-    Liquid::Template.parse(cleaned_parsed).root.nodelist.each do |node|
+    liquid_tag_options = { source: @source, user: @user }
+    Liquid::Template.parse(cleaned_parsed, liquid_tag_options).root.nodelist.each do |node|
       tags << node.class if node.class.superclass.to_s == LiquidTagBase.to_s
     end
     tags.uniq
--- a/app/labor/markdown_parser.rb
+++ b/app/labor/markdown_parser.rb
@@ -4,9 +4,10 @@ class MarkdownParser
 
   WORDS_READ_PER_MINUTE = 275.0
 
-  def initialize(content, source: {})
+  def initialize(content, source: {}, user: nil)
     @content = content
     @source = source
+    @user = user
   end
 
   def finalize(link_attributes: {})
@@ -18,8 +19,9 @@ class MarkdownParser
     html = markdown.render(escaped_content)
     sanitized_content = sanitize_rendered_markdown(html)
     begin
-      parsed_liquid = Liquid::Template.parse(sanitized_content)
-      html = markdown.render(parsed_liquid.render(nil, registers: { source: @source }))
+      liquid_tag_options = { source: @source, user: @user }
+      parsed_liquid = Liquid::Template.parse(sanitized_content, liquid_tag_options)
+      html = markdown.render(parsed_liquid.render)
     rescue Liquid::SyntaxError => e
       html = e.message
     end
--- a/app/liquid_tags/liquid_tag_base.rb
+++ b/app/liquid_tags/liquid_tag_base.rb
@@ -3,11 +3,53 @@ class LiquidTagBase < Liquid::Tag
     ""
   end
 
+  def validate_data(parsed_context)
+    source = parsed_context.partial_options[:source]
+    validate_contexts(source)
+
+    user = parsed_context.partial_options[:user]
+    validate_user_permissions(user)
+  end
+
   def finalize_html(input)
     input.gsub(/ {2,}/, "").
       gsub(/\n/m, " ").
       gsub(/>\n{1,}</m, "><").
       strip.
       html_safe
+  end
+
+  class LiquidTagError < StandardError; end
+
+  private
+
+  def validate_contexts(source)
+    raise LiquidTagError, "No source found" unless source
+
+    is_valid_source = self.class::VALID_CONTEXTS.include? source.class_name
+    invalid_source_error_msg = "Invalid context. This liquid tag can only be used in #{self.class::VALID_CONTEXTS.join(', ')}."
+
+    raise LiquidTagError, invalid_source_error_msg unless is_valid_source
+  end
+
+  def validate_user_permissions(user)
+    raise LiquidTagError, "No user found" unless user
+
+    raise LiquidTagError, "User is not permitted to use this liquid tag" unless user_permitted_to_use_liquid_tag?(user)
+  end
+
+  def user_permitted_to_use_liquid_tag?(user)
+    self.class::VALID_ROLES.any? { |valid_role| user_has_valid_role?(user, valid_role) }
+  end
+
+  def user_has_valid_role?(user, valid_role)
+    # Single resource roles require 2 arguments - the role and the resource
+    if valid_role.is_a? Array
+      role, resource = valid_role
+
+      user.has_role? role, resource
+    else
+      user.has_role? valid_role
+    end
   end
 end
--- a/app/liquid_tags/liquid_tag_base.rb
+++ b/app/liquid_tags/liquid_tag_base.rb
@@ -22,18 +22,18 @@ class LiquidTagBase < Liquid::Tag
   private
 
   def validate_contexts(source)
-    raise LiquidTagError, "No source found" unless source
+    raise LiquidTags::Errors::InvalidParseContext, "No source found" unless source
 
     is_valid_source = self.class::VALID_CONTEXTS.include? source.class_name
     invalid_source_error_msg = "Invalid context. This liquid tag can only be used in #{self.class::VALID_CONTEXTS.join(', ')}."
 
-    raise LiquidTagError, invalid_source_error_msg unless is_valid_source
+    raise LiquidTags::Errors::InvalidParseContext, invalid_source_error_msg unless is_valid_source
   end
 
   def validate_user_permissions(user)
-    raise LiquidTagError, "No user found" unless user
+    raise LiquidTags::Errors::InvalidParseContext, "No user found" unless user
 
-    raise LiquidTagError, "User is not permitted to use this liquid tag" unless user_permitted_to_use_liquid_tag?(user)
+    raise LiquidTags::Errors::InvalidParseContext, "User is not permitted to use this liquid tag" unless user_permitted_to_use_liquid_tag?(user)
   end
 
   def user_permitted_to_use_liquid_tag?(user)
@@ -50,7 +50,4 @@ class LiquidTagBase < Liquid::Tag
       user.has_role? valid_role
     end
   end
-end
-
-class LiquidTagError < StandardError
 end
--- a/app/liquid_tags/liquid_tag_base.rb
+++ b/app/liquid_tags/liquid_tag_base.rb
@@ -3,9 +3,15 @@ class LiquidTagBase < Liquid::Tag
     ""
   end
 
-  def validate_data(parse_context)
-    validate_contexts(parse_context) if context_restricted?
-    validate_user_roles(parse_context) if role_restricted?
+  def initialize(_tag_name, _content, _parse_context)
+    super
+    validate_contexts
+    Pundit.authorize(
+      parse_context.partial_options[:user],
+      self,
+      :initialize?,
+      policy_class: LiquidTagPolicy,
+    )
   end
 
   def finalize_html(input)
@@ -18,15 +24,9 @@ class LiquidTagBase < Liquid::Tag
 
   private
 
-  def context_restricted?
-    self.class.const_defined? "VALID_CONTEXTS"
-  end
+  def validate_contexts
+    return unless self.class.const_defined? "VALID_CONTEXTS"
 
-  def role_restricted?
-    self.class.const_defined? "VALID_ROLES"
-  end
-
-  def validate_contexts(parse_context)
     source = parse_context.partial_options[:source]
     raise LiquidTags::Errors::InvalidParseContext, "No source found" unless source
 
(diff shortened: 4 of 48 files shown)


app/controllers/articles_controller.rb
```diff
@@ -87,7 +87,7 @@ def preview
     begin
       fixed_body_markdown = MarkdownFixer.fix_for_preview(params[:article_body])
       parsed = FrontMatterParser::Parser.new(:md).call(fixed_body_markdown)
-      parsed_markdown = MarkdownParser.new(parsed.content, source: Article.new(user: current_user))
+      parsed_markdown = MarkdownParser.new(parsed.content, source: Article.new(user: current_user), user: current_user)
```
atsmith813, line 90: Here's context in action! The `source` is **where** the liquid tag will be used and `user` is **who** is using it.

app/liquid_tags/liquid_tag_base.rb
```diff
@@ -3,11 +3,54 @@ def self.script
...
+  def user_has_valid_role?(user, valid_role)
+    # Single resource roles require 2 arguments - the role and the resource
+    if valid_role.is_a? Array
+      role, resource = valid_role
+
+      user.has_role? role, resource
+    else
+      user.has_role? valid_role
+    end
+  end
```
atsmith813, line 52: When a role is a [single resource role](https://docs.dev.to/internal/), we need to define the role _and_ the resource. This method handles both. So, for example, `VALID_ROLES` might be something like:
```ruby
VALID_ROLES = [
  :super_admin,
  [:codeland_organizer, UserSubscriptionTag],
].freeze
```

So we handle both. And a liquid tag can have multiple valid roles.

app/liquid_tags/user_subscription_tag.rb
```diff
@@ -214,22 +219,22 @@ class UserSubscriptionTag < LiquidTagBase
     }
   JAVASCRIPT
 
-  def initialize(_tag_name, cta_text, _tokens)
+  def initialize(_tag_name, cta_text, parsed_context)
```
atsmith813, line 222: Either the [`liquid` gem](https://github.com/Shopify/liquid) has made some updates or we just made a slight mistake in development. You might notice that in every other liquid tag, the third argument for `initialize` is `tokens`. I dug through the source code and found that's incorrect - [it's actually `pasred_context`](https://github.com/Shopify/liquid/blob/1feaa6381300d56e2c71b49ad8fee0d4b625147b/lib/liquid/tag.rb#L27).

I will update all the other liquid tags in a separate PR since that will result in like 30+ file changes.

app/labor/markdown_parser.rb
```diff
@@ -18,8 +19,9 @@ def finalize(link_attributes: {})
     html = markdown.render(escaped_content)
     sanitized_content = sanitize_rendered_markdown(html)
     begin
-      parsed_liquid = Liquid::Template.parse(sanitized_content)
-      html = markdown.render(parsed_liquid.render(nil, registers: { source: @source }))
+      liquid_tag_options = { source: @source, user: @user }
+      parsed_liquid = Liquid::Template.parse(sanitized_content, liquid_tag_options)
```
atsmith813, line 23: Digging through the [`liquid` gem](https://github.com/Shopify/liquid) source code, I found that [the second argument here is actually `options`](https://github.com/Shopify/liquid/blob/1feaa6381300d56e2c71b49ad8fee0d4b625147b/lib/liquid/template.rb#L133). When you continue to pull on that thread, you'll find these options become a `ParsedContext` object which is then [passed to the `initialize` method by default](https://github.com/thepracticaldev/dev.to/pull/8917/files#r445812102).

app/labor/markdown_parser.rb
```diff
@@ -4,9 +4,10 @@ class MarkdownParser
 
   WORDS_READ_PER_MINUTE = 275.0
 
-  def initialize(content, source: {})
+  def initialize(content, source: nil, user: nil)
```
atsmith813, line 7: I'm thinking with this setup the `MarkdownParser` as a whole can have come "contextual awareness". It will be able to understand the context that it's parsing markdown for. Context isn't always necessary, so we default to `nil`.

app/liquid_tags/liquid_tag_base.rb
mstruve: Should we have a LiquidTag Policy class for this? Might help contain the logic a bit in a single place rather than having the liquid tag worry about it.

app/controllers/articles_controller.rb
rhymes: Is there a case where `source.user` and `user` are different?

app/errors/liquid_tags.rb
```diff
@@ -0,0 +1,9 @@
+module LiquidTags
+  module Errors
+    class Error < StandardError
+    end
+
+    class InvalidParsedContext < Error
+    end
```
rhymes, line 7: Can you add a comment to explain withat `invalid parsed context` applies to? I can see it by reading the PR but maybe an explanation is useful as well

app/labor/markdown_parser.rb
rhymes: so the `registers:` options is not needed anymore?

app/liquid_tags/liquid_tag_base.rb
rhymes: I agree with mstruve, it's a bit odd to have the role logic here in the liquid tag but maybe it can't moved up because of reasons related to parsing?

atsmith813: Hmmm...I think I'm misunderstanding. Can you both clarify what you mean?

It's in a single location because all `LiquidTags` inherit from this base class. Or do you mean contain the piece where we are specifying roles in a single location? The roles will change for each liquid tag. So maybe put that "mapping" for each tag in a single location?

app/labor/markdown_parser.rb
atsmith813: Correct! We are passing the same information in an improved way (in the initialize step and using options).

app/liquid_tags/liquid_tag_base.rb
```diff
@@ -3,11 +3,56 @@ def self.script
...
+  def role_restricted?
+    self.class.const_defined? "VALID_ROLES"
+  end
+
+  def validate_contexts(parsed_context)
+    source = parsed_context.partial_options[:source]
+    raise LiquidTags::Errors::InvalidParsedContext, "No source found" unless source
+
+    is_valid_source = self.class::VALID_CONTEXTS.include? source.class.name
+    invalid_source_error_msg = "Invalid context. This liquid tag can only be used in #{self.class::VALID_CONTEXTS.join(', ')}."
```
juliannatetreault, line 34: Quick, non-blocking question: what do you think about changing this error message to be plural? 

The reason I ask is that when an error occurs, the message reads "Invalid context. This liquid tag can only be used in Article". By pluralizing it, when an error occurs, it will read a bit nicer: "Invalid context. This liquid tag can only be used in Articles" and it will make the error message reusable if and when we decide to add more valid contexts (as opposed to how it would less readable and reusable if we added an "an" to the error message and did not pluralize it).

app/controllers/articles_controller.rb
atsmith813: Currently, no. In the future that could happen. Liquid tags are "used" wherever we render the markdown editor. In the future, it's possible an Admin creates something on behalf of another user.

Also, the relationship isn't always `user` on the `source`. In future source types, it could be `.author`, `.admin`, etc. 

And since these are used in a markdown editor, it's also possible for liquid tags to be used within an object that has no association to a user. `Page` is a current example. In the case of `Page`, we'd just want to check if the currently signed-in user is permitted to use it since `Page` has no `User` association itself.

rhymes:
I have a question: what happens to the liquid tag and its containing comment or article if a user is demoted from the previous role? Say you're an admin, you publish an article with a subcription, then you're not an admin anymore, the article is already published and cached I guess so we can't change anything.

But let's say we decide to refresh all articles when the role is removed. Will the new markdown stop rendering as it's suddenly invalid? Should we replace the block with "here lies something you cannot see" 🤣  or else?

app/liquid_tags/user_subscription_tag.rb
juliannatetreault: After looking at the source code you linked to, it looks like it actually might be `parse_context` rather than `parsed_context` - am I seeing that correctly? 😬

atsmith813: Good catch, thanks! I'll update it.

app/liquid_tags/liquid_tag_base.rb
mstruve: Store all of the logic for determining if a liquid tag can be used or not in a Policy rather than here. https://github.com/thepracticaldev/dev.to/tree/master/app/policies since that is our convention for controlling access for most things

atsmith813:
@rhymes - the refreshing of articles would fail because the article would no longer pass validations. So in that case, you couldn't refresh all the articles and it would remain the same.

Replacing the block is one way. That breaks the "liquid tag convention". We made a push a while back to raise errors vs saving and rendering errors wrt Liquid Tags. Is that an issue?

A complicated route is to use[ rolify callbacks of `after_remove`](https://github.com/RolifyCommunity/rolify#31-configure-your-user-model) to clean up any articles that were using that role. That sounds like overkill 🤔 .

What do you think?

rhymes:
@atsmith813 they all sound overkill to me :D I just had that question in mind. I'd add this to the list of things to consider post codeland if we decide to enlarge the pool of users that can use this feature from not just admins

app/liquid_tags/liquid_tag_base.rb
atsmith813: Updated!

atsmith813:
@rhymes LOL well then...

Very good question! I've been keeping track and I added this scenario as [a card](https://github.com/thepracticaldev/dev.to/projects/19#card-40811306). Feel free to take a look or add any to the column!

app/liquid_tags/liquid_tag_base.rb
vaidehijoshi: If it's possible to put these permissions-related things into a specific `Policy` class, I agree that this would be preferable here.

app/liquid_tags/liquid_tag_base.rb
```diff
@@ -3,11 +3,57 @@ def self.script
     ""
   end
 
+  def validate_data(parse_context)
+    validate_contexts(parse_context) if context_restricted?
```
citizen428, line 7: I'd probably remove the conditionals here and move them into the validation methods ("tell, don't ask"):

```ruby
def validate_contexts(parse_context)
   return unless context_restricted?
   ....
end
```

Since these methods are private you may not even be using them in other places, so it may be easier to just inline them and get rid of the predicate methods:

```ruby
def validate_contexts(parse_context)
  return unless self.class.const_defined?("VALID_CONTEXTS")
  ...
end
```

app/liquid_tags/liquid_tag_base.rb
citizen428: If you really can't use a policy class here for whatever reason, maybe we can at least remove the conditional logic:

```ruby
def user_has_valid_role?(user, valid_role)
  # Splat array for single resource roles
  user.has_role?(*Array(valid_role))
end
```

atsmith813:
@mstruve @rhymes @vaidehijoshi @citizen428 @juliannatetreault 

Marking this ready for review. Changes include:
- Moving permission validation to a new policy. It might look a little...wonky. It wasn't "straight forward" because Pundit is built on the use of Controller/Model relationships. It supports authorization outside of controllers, but it's very much an after-thought. Also, liquid tags themselves behave quite differently than the assumed MVC setup Pundit is built on. I'm open to suggestions to improve it!
- Added tests and docs.
- This PR introduces changes that affect _all_ liquid tags (i.e. changing `tokens` to `parse_context`). I'll make the updates to other liquid tags in separate PRs because they will result in many files changed, making this PR very noisy.

app/liquid_tags/liquid_tag_base.rb
```diff
@@ -3,11 +3,39 @@ def self.script
     ""
   end
 
+  def initialize(_tag_name, _content, _parse_context)
+    super
+    validate_contexts
+    # This check issues DB queries so we keep it as the last one
+    Pundit.authorize(
+      parse_context.partial_options[:user],
```
citizen428, line 11: I assume this has to do with inheriting from `Liquid::Tag` and calling `super` (assumption: the super class uses the argument and makes it available via an accessor), but it's confusing to see `parse_context` used when we have a param that suggests it isn't (`_parse_context`).

app/liquid_tags/liquid_tag_base.rb
```diff
@@ -3,11 +3,39 @@ def self.script
...
       gsub(/\n/m, " ").
       gsub(/>\n{1,}</m, "><").
       strip.
       html_safe
   end
+
+  private
+
+  def validate_contexts
+    return unless self.class.const_defined? "VALID_CONTEXTS"
```
citizen428, line 29: If you don't want to go through the class, you can use `defined?(VALID_CONTEXTS)`. It has a somewhat weird output (`nil` or `"constant"`), but works for truthiness checks.

app/policies/liquid_tag_policy.rb
```diff
@@ -0,0 +1,30 @@
...
+  attr_reader :user, :record
+
+  def initialize(user, record)
+    @user = user
+    @record = record
+  end
+
+  def initialize?
+    return true unless record.class.const_defined?("VALID_ROLES")
+    raise Pundit::NotAuthorizedError, "No user found" if record.class.const_defined?("VALID_ROLES") && user.nil?
```
citizen428, line 13: If we get to this line `VALID_ROLES` should be defined (since the previous line didn't return), so we don't need to check again for that, do we?

atsmith813: Correct, good catch! This was my mistake from moving this check around and messing with Pundit. Updated 👍.

app/liquid_tags/liquid_tag_base.rb
atsmith813: I couldn't get this to work 🤔. `defined?(VALID_CONTEXTS)` returned `nil` in the cases when I expected it not to.

app/liquid_tags/liquid_tag_base.rb
atsmith813: @citizen428 Yep - your assumption is spot on! Updated it to `parse_context` 👍.

app/liquid_tags/liquid_tag_base.rb
citizen428: Nevermind, it's not important. It may not work as I thought for inherited constants for example 😄

app/policies/liquid_tag_policy.rb
```diff
@@ -0,0 +1,30 @@
+# Intentionally not inheriting from ApplicationPolicy because liquid tags behave
```
mstruve, line 1: Thanks for the explanation!

spec/policies/liquid_tag_policy_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      end.not_to raise_error
+    end
+
+    it "handles single resource roles" do
+      # TODO: (Alex Smith) - update roles to new liquid tag role for more relevant example/use
+      user = create(:user, :single_resource_admin, resource: Article)
+      parse_context = { source: article, user: user }
+
+      allow(liquid_tag).to receive(:parse_context).and_return(parse_context)
+      stub_const("#{liquid_tag.class}::VALID_ROLES", [[:single_resource_admin, Article]])
```
vaidehijoshi, line 48: Hm, I am confused about this test. Looking [at the docs](https://github.com/thepracticaldev/dev.to/pull/8917/files#diff-050073bb34d810dba4e9692f8aca0209R128), shouldn't the resource here be `NewLiquidTag` rather than `Article`? Or am I misunderstanding?

app/policies/liquid_tag_policy.rb
```diff
@@ -0,0 +1,30 @@
...
+    @record = record
+  end
+
+  def initialize?
+    return true unless record.class.const_defined?("VALID_ROLES")
+    raise Pundit::NotAuthorizedError, "No user found" unless user
+    # Manually raise error to use a custom error message
+    raise Pundit::NotAuthorizedError, "User is not permitted to use this liquid tag" unless user_permitted_to_use_liquid_tag?
+
+    true
```
vaidehijoshi, line 17: Why do we need to return `true` here? Wouldn't `user_permitted_to_use_liquid_tag?` return `true` as the last thing that is invoked here? (Unless of course it was `false` and then `raise`d the `Pundit::NotAuthorizedError`?

spec/policies/liquid_tag_policy_spec.rb
atsmith813: `NewLiquidTag` isn't defined, so we can't use that in code. That was included in the docs to demonstrate how it would be used. Whatever new tag you're creating, you could include that as a role.

app/policies/liquid_tag_policy.rb
atsmith813: It returns `nil` because it's a conditional statement.

vaidehijoshi: Ah, of course. 🤦‍♀️

docs/frontend/liquid-tags.md
```diff
@@ -106,3 +107,46 @@ etc.
...
+It needs to be an `Array` of class names that are valid. For example, to
+restrict a liquid tag to only be usable in articles you would do:
+
+```ruby
+class NewLiquidTag < LiquidTagBase
+  VALID_CONTEXTS = %w[Article].freeze
+end
+```
+
+_If you do not define this constant the liquid tag will be usable in all
```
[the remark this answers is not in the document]
atsmith813: Bold doesn't work in our docs at the moment 😬 .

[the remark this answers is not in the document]
atsmith813: And adding `code` into italics doesn't work either 😬 .

Issue #8313, opened 2020-06-06
Be able to fetch article from API with slug instead of ID


emilpriver:
**Is your feature request related to a problem? Please describe.**

Be able to fetch article from API with slug instead of ID.

**Describe the solution you'd like**

Be able to fetch article via slug from api.

Muhimen123:
Why do you want to use slug instead of ID? Anyway, you can use web scraping. Like this.
```
URL = "https://www.dev.to/" + slug_path
```

The website is pretty well-formatted. You shouldn't have any problem scraping the web.

emilpriver:
> Why do you want to use slug instead of ID?

Mainly as it a more prettier link that using an ID. Instead of /post/123123 would I be able to use /post/hello-world which is a more prettier link and also more SEO friendly :D

nolleto:
Use slug still a desirable feature? If so, I can try to implement it.

rhymes:
Hi @nolleto, I think so in theory, have you thought about the API interface?

Right now we have `/articles/:id` to get a published article. How would "get by slug" fit in the API?

Oghenebrume50:
@rhymes so I was looking at this, just curious,  this is  (https://github.com/thepracticaldev/dev.to/blob/master/app/controllers/api/v0/articles_controller.rb) the file where the implementation is made I guess. 
Also is there an extra setup to make the API work in my local environment, because I tried running the get all with curl `curl localhost:3000/articles` but it threw an error. There is nothing in the docs with regards to this

rhymes:
Hi @Oghenebrume50, good questions! the API docs is https://docs.dev.to/api/

The API locally is reachable at `curl localhost:3000/api/articles` and similar URLs but refered to the documentation.

The documentation is auto generated from this file: https://github.com/thepracticaldev/dev.to/blob/master/docs/api_v0.yml

Here's how you can view it and edit it locally: https://docs.dev.to/contributing_api/

rhymes:
Solved by https://github.com/forem/forem/pull/8929

PR #8929, opened 2020-06-26
Fetch article by slug

Oghenebrume50:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
According to this issue #8313 the user suggested that articles should be searchable by slug and not only by id. My implement included converting the param input to an integer, so if a string that is not a valid number is passed it returns 0. So if the value is not zero then I find with the id else I find by the slug,  in both cases, the error is well handled by rails, this is why I used `find_by!` instead of `find_by` because the former return an ` ActiveRecord::ActiveRecordError` when nothing is found

This is an example of a working slug
workingslug

Example of a wrong slug that is not found 
wrongslug

Also, the id works fine
id

I tested everything, all possible endpoints and they work fine. 

I am opening this as a draft PR to check if this is a good approach, if this is okay I can update the documentation.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


Oghenebrume50:
@rhymes this is an attempt to solve issue #8313 please review if it is acceptable it works fine, as can be seen in the screenshots.

rhymes:
Hi @Oghenebrume50, clever idea! There are a few things to consider: 

Slug are not unique per article but are unique per user:

https://github.com/thepracticaldev/dev.to/blob/a7938ed287899840238beb55b65ad481bb92c4a7/app/models/article.rb#L45-L46

this means that the path should be the same as the public article: `dev.to/:username/:slug`

That said, there's also another thing to consider: we'd be bending a bit the RESTful contract. I wonder if we shouldn't simply add a separate route:

`/api/articles/:username/:slug`

what do you think?

rhymes:
ps. we do have a column named `path` which you can use to load the article: `Article.where(path: "/#{username}/#{slug}")`

Oghenebrume50:
@rhymes there is a column named path and finding the article that way works fine. 

Also, I think to bend the RESTful contract wouldn't be so bad since slugs are unique to users that will be the safest way to go about it.

rhymes:
@Oghenebrume50 as Rails will do the routing work for you if you define a `/api/articles/:username/:slug` route I don't think we need to bend `:id` to be both a number or a path. This way you have two clean paths. 

The implementation would be very similar to `API::V0::ArticlesController#show`. We can work on the documentation on a separate PR I think

Oghenebrume50:
@rhymes this should fix the issue properly, I decided to use `find_by!` instead of `where` because where returns an array but find_by! returns the actual active record object and since the path is always unique this should work fine. 
slug works

Oghenebrume50:
Hi @rhymes @mstruve I have written the test for this feature and the seem to work fine except one I wasn't so sure about, the same for the api/article/id can be found here
https://github.com/thepracticaldev/dev.to/blob/09e70896fac773ee5efb29b2311e880f2bd88220/spec/requests/api/v0/articles_spec.rb#L19
This did not pass and after debugging I wasn't sure why, however, I felt that this wasn't necessary since, a unique path to the user profile would mean that the article does not belong to the organization, going by this then I think it is safe to leave this test out :thinking: 

Also, I am not sure why Travis fails here, the output isn't so clear

spec/requests/api/v0/articles_spec.rb
```diff
@@ -342,6 +342,71 @@
...
+      expect(response.headers["Access-Control-Allow-Origin"]).to eq(origin)
+      expect(response.headers["Access-Control-Allow-Methods"]).to eq("HEAD, GET, OPTIONS")
+      expect(response.headers["Access-Control-Expose-Headers"]).to be_empty
+      expect(response.headers["Access-Control-Max-Age"]).to eq(2.hours.to_i.to_s)
+    end
+
+    it "returns correct tag list" do
+      get slug_api_articles_path(username: article.username, slug: article.slug)
+
+      expect(response.parsed_body["tag_list"]).to be_a_kind_of String
```
mstruve, line 359: It would probably a good idea to check that the returned tag list actually matches the article tag list rather than just checking if its a string. I also think to save some spec runtime you could merge the spec below into this one and just have a single spec that checks the tags. 

```ruby
    it "returns correct tags" do
      get slug_api_articles_path(username: article.username, slug: article.slug)

      expect(response.parsed_body["tags"]).to match_array(article.tags)
      expect(response.parsed_body["tag_list"]).to eq(article.tag_list) # or whatever method you need to call to get the correct match
    end
```

Oghenebrume50:
Hi @rhymes @mstruve I have fixed the test

PR #8935, opened 2020-06-26
Feature: Setup Admin Area to Update Vault Secrets

mstruve:
- [x] Feature

## Description
This PR gives admins the ability to update secrets in Vault from our internal namespace. If Vault is not enabled for an application then we will simply list the secrets with a warning. NOTE: These three keys are probably much better suited for SiteConfig but I wanted something low risk to test the first run of Vault with and I figured these were a good choice. With that said **Vault is still DISABLED** on dev. This PR is simply putting in place the infrastructure to update it once its activated. 

Vault Enabled
Screen Shot 2020-06-26 at 2 24 19 PM

Vault Disabled
Screen Shot 2020-06-26 at 2 24 33 PM

Vault Secret Updated
Screen Shot 2020-06-26 at 3 37 30 PM

https://github.com/thepracticaldev/dev.to/projects/9#card-40734891

## QA Instructions
If you would like to test this PR you will first need to setup Vault as explained in #8936. Once you have it all setup visit `http://localhost:3000/internal/secrets` and try updating secrets. 

To test what happens when Vault is disabled you can either visit that URL without setting up Vault OR comment out the Vault token in your `application.yml` file. 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/controllers/internal/secrets_controller.rb
+++ b/app/controllers/internal/secrets_controller.rb
@@ -1,0 +1,33 @@
+class Internal::SecretsController < Internal::ApplicationController
+  layout "internal"
+
+  before_action :validate_settable_secret, only: [:update]
+  after_action only: [:update] do
+    Audit::Logger.log(:internal, current_user, params.dup)
+  end
+
+  def index
+    @vault_enabled = AppSecrets.vault_enabled?
+    @secrets = AppSecrets::SETTABLE_SECRETS.map do |key|
+      secret_value = AppSecrets[key]
+      secret_value = secret_value.present? ? "#{secret_value.first(8)}******" : "Not In Vault"
+      [key, secret_value]
+    end
+  end
+
+  def update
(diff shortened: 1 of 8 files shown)


app/controllers/internal/secrets_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+  after_action only: [:update] do
+    Audit::Logger.log(:internal, current_user, params.dup)
+  end
+
+  def index
+    @vault_enabled = AppSecrets.vault_enabled?
+    @secrets = AppSecrets::SETTABLE_SECRETS.map do |key|
+      secret_value = AppSecrets[key]
+      secret_value = secret_value.present? ? "#{secret_value.first(8)}******" : "Not In Vault"
+      [key, secret_value]
```
mstruve, line 14: First 8 was pretty arbitrary, seemed like most the secret was hidden but you could see enough to know it it was right or wrong.

PR #8938, opened 2020-06-26
Feature: Introduce KnapsackPro Queue Mode

mstruve:
- [x] Feature

## Description
This PR enables [Queue mode for KnapsackPro](https://youtu.be/hUEB1XDKEFY). In Regular Mode KnapsackPro will make a single call on each node at the beginning of a CI run to request a set of specs based on historical timing data. In our case since we have 3 nodes, we get 3 batches that are as even as possible. However, as the tests run, one batch may hit some retries or failed specs and end up being much slower than the others. Queue Mode helps mitigate this problem. 

In Queue Mode, each node will start by pulling a small set of a few specs. Once those specs are run then the node will pull another few specs. This will occur until all the specs in our suite have been run. By doing it this way, if one node is cranking through specs quickly then we will end up sending more specs to it rather than it finishing early and having to wait on the other nodes. 

The downside here, in terms of our test suite, is that likely more flaky specs are going to pop up. **I feel** like our suite has come a long way thanks to KnapsackPro already lightly introducing some randomness to the spec ordering and that its ready to handle this next step.  Totally open to any other thoughts people have though if you disagree!

@snackattas @ArturT
 
https://github.com/thepracticaldev/dev.to/projects/9#card-40829203

alt_text

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: required
 language: ruby
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,13 +63,13 @@ script:
   - bundle exec rails webpacker:compile
   - bundle exec rails db:schema:load
   - './cc-test-reporter before-build'
-  - 'bin/knapsack_pro_rspec'
-  - '[ ! -f .approvals ] || bundle exec approvals verify --ask false'
   - if [ "$KNAPSACK_PRO_CI_NODE_INDEX" == "0" ]; then yarn test --colors; fi
   - if [ "$KNAPSACK_PRO_CI_NODE_INDEX" == "0" ]; then ./cc-test-reporter format-coverage -t lcov -o coverage/codeclimate.lcov.json; fi
   - if [ "$KNAPSACK_PRO_CI_NODE_INDEX" == "1" ]; then bundle exec bundle-audit check --update --ignore CVE-2015-9284; fi
   - if [ "$KNAPSACK_PRO_CI_NODE_INDEX" == "1" ]; then yarn build-storybook; fi
   - if [ "$KNAPSACK_PRO_CI_NODE_INDEX" == "2" ]; then bundle exec rails runner -e production 'puts "App booted successfully"'; fi
+  - 'bin/knapsack_pro_rspec'
+  - '[ ! -f .approvals ] || bundle exec approvals verify --ask false'
 after_script:
   - ./cc-test-reporter format-coverage -t simplecov -o ./coverage/codeclimate.$KNAPSACK_PRO_CI_NODE_INDEX.json ./coverage/spec/.resultset.json
   - ./cc-test-reporter sum-coverage --output - --parts $COVERAGE_REPORTS_TOTAL coverage/codeclimate.*.json | ./cc-test-reporter upload-coverage --input -
--- a/bin/knapsack_pro_rspec
+++ b/bin/knapsack_pro_rspec
@@ -4,5 +4,5 @@ if [ "$KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC" = "" ]; then
     KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC=disabled-for-fork \
     bundle exec rake knapsack_pro:rspec # use Regular Mode here always
 else
-  bundle exec rake knapsack_pro:rspec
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -1,4 +1,3 @@
-sudo: required
```
mstruve, line 1: Snuck in this cleanup, this is deprecated according to the warning in Travis.

Screen Shot 2020-06-26 at 4 24 25 PM

ArturT:
@mstruve Looks good. The only thing I would consider if it's possible to run knapsack pro queue mode after other commands. This way you would auto-balance parallel CI jobs execution time to be even (as a result of that you get as fast CI build as possible).

```json
  "script": [
    "date --rfc-3339=seconds",
    "nvm install",
    "pkill -9 -f elasticsearch || true",
    "curl -s -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-7.5.2-amd64.deb",
    "sudo dpkg -i --force-confnew elasticsearch-7.5.2-amd64.deb",
    "sudo sed -i.old 's/-Xms1g/-Xms128m/' /etc/elasticsearch/jvm.options",
    "sudo sed -i.old 's/-Xmx1g/-Xmx128m/' /etc/elasticsearch/jvm.options",
    "echo -e '-XX:+DisableExplicitGC\\n-Djdk.io.permissionsUseCanonicalPath=true\\n-Dlog4j.skipJansi=true\\n-server\\n' | sudo tee -a /etc/elasticsearch/jvm.options",
    "sudo chown -R elasticsearch:elasticsearch /etc/default/elasticsearch",
    "sudo systemctl start elasticsearch",
    "curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter",
    "chmod +x ./cc-test-reporter",
    "bundle config set path 'vendor/bundle'",
    "bundle install --local --jobs=2",
    "yarn install --frozen-lockfile",
    "bundle exec rails db:create",
    "bundle exec rails webpacker:compile",
    "bundle exec rails db:schema:load",
    "./cc-test-reporter before-build",
    
    "[ ! -f .approvals ] || bundle exec approvals verify --ask false",
    "if [ \"$KNAPSACK_PRO_CI_NODE_INDEX\" == \"0\" ]; then yarn test --colors; fi",
    "if [ \"$KNAPSACK_PRO_CI_NODE_INDEX\" == \"0\" ]; then ./cc-test-reporter format-coverage -t lcov -o coverage/codeclimate.lcov.json; fi",
    "if [ \"$KNAPSACK_PRO_CI_NODE_INDEX\" == \"1\" ]; then bundle exec bundle-audit check --update --ignore CVE-2015-9284; fi",
    "if [ \"$KNAPSACK_PRO_CI_NODE_INDEX\" == \"1\" ]; then yarn build-storybook; fi",
    "if [ \"$KNAPSACK_PRO_CI_NODE_INDEX\" == \"2\" ]; then bundle exec rails runner -e production 'puts \"App booted successfully\"'; fi"

    "bin/knapsack_pro_rspec", <--- keep it after other command if possible
  ],
```

Something to take into account:

* If for example ` "if [ \"$KNAPSACK_PRO_CI_NODE_INDEX\" == \"1\" ]; then bundle exec bundle-audit check --update --ignore CVE-2015-9284; fi",` step fails and next step won't start it may result in knapsack pro never run on node index 1. That's bad because other parallel jobs would run all tests and make CI build slower. The CI node index 1 would not run tests so knapsack pro won't collect timing data properly.

I'm not sure how your current Travis CI config works exactly if RSpec tests fail then yarn test is not started at all? If so then let's keep it as it is.
But if RSpec and yarn test are always started even when RSpec tests failed then it would make sense to run yarn tests first and knapsack pro in queue mode after yarn test. This way we would auto balance parallel jobs time to keep faster CI build.
https://knapsackpro.com/faq/question/what-is-optimal-order-of-test-commands

mstruve:
It does not "fast fail", when any part fails it records an exit code of 1 and keeps processing the rest of the bulld so it sounds like the right move to switch them. Thanks for the tip!

mstruve:
> Does this have any impact on other communities that are deploying autonomously?

This doesn't change anything really from KnapsackPro regular mode and given we have the fallback mode in place for missing tokens any community can still use this with Travis. With that said we do need to figure out a nice way to include CI configs so that other communities can use whatever method/CI they choose for deploying. @jdoss might have some ideas around that.

Issue #8749, opened 2020-06-17
Connect Video no longer opens


benhalpern:
This is not a _core_ feature so it's not the end of the world, but worth noting that video functionality is not currently working as expected.

Reproduce:

Go to Connect -> type `/call` -> Check console.

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

narender2031:
@benhalpern & @rhymes. This issue is fixed in another PR. We can close this issue.

PR #8945, opened 2020-06-27
🚀 Feature: Chat channel membership manager component

sarthology:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
In this PR we have completed the Chat Channel Membership Manager. Which gives you more control to perform following actions as a mod on the channel.
- Ability to send invitations using an URL.
- Make other member Mod of the channel.
- Search member from a long list. 
We also fixed few bugs like 
- [#8749](https://github.com/thepracticaldev/dev.to/issues/8749) 
Screen Recording 2020-06-27 at 05 40 PM
Screen Recording 2020-06-27 at 05 38 PM

To test:
- Set up your user to become a mod of the channel

```ruby
u = User.last # Me

cc = ChatChannel.create!(channel_type: "open", channel_name: "ACME", slug: "acme")
cc.add_users(u) # Add myself which I do see in the site, but only as a member
cc = ChatChannel.create!(channel_type: "open", channel_name: "ACME", slug: "acme")

u.chat_channel_memberships.last.update(role: "mod")
```

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

alt_text


app/presenters/chat_channel_detail_presenter.rb
```diff
@@ -39,4 +40,9 @@ def membership_users(memberships)
       }
     end
   end
+
+  def inviation_link(chat_channel)
+    host = Rails.application.config.app_domain || "localhost:3000"
```
mstruve, line 45: For me `Rails.application.config.app_domain` returns localhost locally.
```
[4] pry(main)> Rails.application.config.app_domain
=> "localhost:3000"
```
and since you define this in `test.rb` I dont think you need this `||` here.

app/services/create_chat_channel_membership_service.rb
```diff
@@ -0,0 +1,14 @@
+class CreateChatChannelMembershipService < ApplicationService
+  def initialize(chat_channel, current_user)
+    @chat_channel = chat_channel
+    @user = current_user
+  end
+
+  attr_accessor :chat_channel, :user
+
+  def perform
+    membership = ChatChannelMembership.new(user_id: user.id, chat_channel_id: chat_channel.id)
```
mstruve, line 10: Since we are simply creating a `ChatChannelMembership` this doesn't feel like it needs a whole service. If we had more logic then I would be all for this but to simply use what ActiveRecord has already given us feels a bit redundant and not worth the added complexity.

app/services/update_chat_channel_membership_service.rb
```diff
@@ -0,0 +1,15 @@
...
+    @chat_channel = chat_channel
+    @membership = membership
+    @params = params
+  end
+
+  attr_accessor :chat_channel, :membership, :params
+
+  # @TODO: Add dynamic content
+  def perform
+    membership.update(params)
```
mstruve, line 12: Same comment as above, since we are simply doing an update statement here it doesnt feel like it warrants the added complexity of an entire service.

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -115,14 +115,50 @@ def leave_membership
...
+    chat_channel = ChatChannel.find_by(slug: params[:channel_slug])
+    authorize chat_channel
+    membership = ChatChannelMembership.find_by(user_id: current_user.id, chat_channel_id: chat_channel.id)
+
+    unless membership
+      result = CreateChatChannelMembershipService.perform(chat_channel, current_user)
+      send_chat_action_message("@#{result.user.username} join the channel", current_user, chat_channel.id, "joined") unless result&.errors&.any?
+    end
+
+    if membership && membership.status != "active"
```
mstruve, line 141: I got a little lost with this `if/unless` set of statements. I think it might make it a little more clear to have a single statement like
```ruby
if !membership 
  # Create membership
elsif membership.status != "active"
  # Update membership
end
```

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -115,14 +115,50 @@ def leave_membership
...
+      result = UpdateChatChannelMembershipService.perform(chat_channel, membership, { role: "member", status: "active" })
+      send_chat_action_message("@#{membership.user.username} join the channel", current_user, membership.chat_channel_id, "joined")
+    end
+
+    if result&.errors&.any?
+      flash[:settings_notice] = "There is some error"
```
mstruve, line 147: We should be able to grab the errors off of the membership model to print here to make it more clear what the problem might be.

app/javascript/chat/ChatChannelSettings/ChatChannelSettings.jsx
```diff
@@ -305,6 +314,54 @@ export default class ChatChannelSettings extends Component {
     }
   };
 
+  toggelScreens = () => {
```
jacobherrington, line 317: I think this should be `toggleScreens` seems to be a consistent typo throughout the PR.

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -115,14 +115,49 @@ def leave_membership
...
+    authorize chat_channel
+    membership = ChatChannelMembership.find_by(user_id: current_user.id, chat_channel_id: chat_channel.id)
+
+    if !membership
+      membership = ChatChannelMembership.new(user_id: current_user.id, chat_channel_id: chat_channel.id)
+      membership.save
+      send_chat_action_message("@#{membership.user.username} join the channel", current_user, chat_channel.id, "joined") unless membership&.errors&.any?
+    elsif membership.status != "active"
+      membership.update(role: "member", status: "active")
+      send_chat_action_message("@#{membership.user.username} join the channel", current_user, membership.chat_channel_id, "joined")
```
mstruve, line 142: I think you could refractor this one step further and pull this `send_chat_action_message` out of both sides of this boolean and put it below.

```ruby
if membership&.errors&.any?
  flash[:settings_notice] = membership.errors.full_messages
  redirect_to root_path
else
    send_chat_action_message("@#{membership.user.username} join the channel", current_user, membership.chat_channel_id, "joined")
end

File: spec/requests/chat_channel_memberships_spec.rb [hunk 8]
```diff
@@ -20,8 +20,8 @@
       end
 
       it "return all details of chat channel" do
-        expect(response).to have_http_status(:ok)
-        expect(response.parsed_body["result"].keys).to eq(%w[chat_channel memberships current_membership])
+        expect(response.status).to eq(200)
+        expect(JSON.parse(response.body)["result"].keys).to eq(%w[chat_channel memberships current_membership invitation_link])
```
mstruve, line 24: Why did you remove `response.parsed_body` here? I think we should keep it since its a little cleaner and we dont have to do the parsing.

app/presenters/chat_channel_detail_presenter.rb
```diff
@@ -39,4 +40,9 @@ def membership_users(memberships)
       }
     end
   end
+
+  def inviation_link(chat_channel)
```
mstruve, line 44: It also looks like everywhere we are missing a "t" in "invitation", we should probably update that.

app/controllers/chat_channel_memberships_controller.rb
narender2031: @mstruve In that case, If a user is already a member and he used this link. then the joining message will appear for him in chat.

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -115,14 +115,49 @@ def leave_membership
...
+
+    membership.update(role: channel_membership_params[:role])
+    if membership.errors.any?
+      render json: { success: false, message: "Failed to update membership", errors: chat_channel_membership.errors.full_messages }, status: :bad_request
+    else
+      render json: { success: true, message: "User Membership is updated" }, status: :ok
+    end
+  end
+
+  def join_channel_invitation
```
benhalpern, line 131: As I understand this, since it relies on a channel slug there's no real way to "expire" the link.

Could we instead have something that could expire?

Also I feel like just _automatically joining a channel_ based on a link you could click on accidentally is not ideal. What if you needed to land on a page first and then proactively click a button that joined via a `POST` request?

narender2031: @benhalpern 
- Yes, we can expire the link.
- Also, I will add the page that will ask user to `join the channel` or `cancel`, when user use the joining link.

quick question: 
- What is the expiry time for the link?  => (1day, 6hours)

app/controllers/chat_channel_memberships_controller.rb
mstruve: Could you solve that by doing this?
```
if membership&.errors&.any?
...ruby
elsif membership.status != "active"
  send_chat_action_message(..)
end
```

app/presenters/chat_channel_detail_presenter.rb
```diff
@@ -39,4 +40,9 @@ def membership_users(memberships)
       }
     end
   end
+
+  def invitation_link(chat_channel)
+    host = Rails.application.config.app_domain
+    "#{host}/chat_channel_memberships/join_channel_invitation/#{chat_channel.slug}"
```
fdoxyz, line 46: Instead of `Rails.application.config.app_domain` it may be better if we use [URL.url](https://github.com/thepracticaldev/dev.to/blob/master/app/lib/url.rb#L13) here to maintain consistency and reliability on urls. It can then look something like this:

```
slug = chat_channel.slug
URL.url("/chat_channel_memberships/join_channel_invitation/#{slug}")
```

app/javascript/chat/ChatChannelSettings/MembershipManager/Membership.jsx
```diff
@@ -0,0 +1,128 @@
...
+  const addAsModButton =
+    membership.role === 'member' ? (
+      <button
+        className="crayons-btn crayons-btn--s"
+        type="button"
+        onClick={handleUpdateMembershipRole}
+        data-membership-id={membership.membership_id}
+        data-role="mod"
+      >
+        Permote as Mod
```
fdoxyz, line 21: Small typo, suggestion: "Promote to Mod" or "Add as Mod"

app/javascript/chat/ChatChannelSettings/MembershipManager/Membership.jsx
```diff
@@ -0,0 +1,128 @@
...
+  const addAsMemberButton =
+    membership.role === 'mod' ? (
+      <button
+        className="crayons-btn crayons-btn--s"
+        type="button"
+        onClick={handleUpdateMembershipRole}
+        data-membership-id={membership.membership_id}
+        data-role="member"
+      >
+        Add as Member
```
fdoxyz, line 34: Since this means the member is a Mod and the action would __make them normal users__ (no longer Mods) I think this could be renamed to "Demote to Member" or "Remove Mod"

app/controllers/chat_channel_memberships_controller.rb
benhalpern: I think a one day expiration makes sense

nickytonline:
@sarthology, I'll be taking a look at this PR this week.

narender2031:
@nickytonline, Today I will wrap up this PR.

app/controllers/chat_channel_memberships_controller.rb
narender2031: @benhalpern added expiry for an invitation link

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -24,15 +24,25 @@ def chat_channel_info
...
-
-    render json: { success: true, result: data, message: "" }, success: :ok
+    invitation_link = @channel.chat_channel_invitation_links.last
+    if !invitation_link
+      invitation_link = CreateChatChannelInvitationLink.perform(@channel)
+    elsif invitation_link && (invitation_link.expiry_time < DateTime.now || invitation_link.expired?)
+      invitation_link.update(status: "expired") if invitation_link.active?
+      invitation_link = CreateChatChannelInvitationLink.perform(@channel)
+    end
+    if invitation_link&.errors&.any?
```
rhymes, line 34: Is there a chance invitation link is still nil here? It doesn't seem so, but maybe I'm wrong

app/javascript/chat/actions/chat_channel_setting_actions.js
```diff
@@ -142,7 +142,31 @@ export async function leaveChatChannelMembership(membershipId) {
...
       method: 'PATCH',
+      credentials: 'same-origin',
+    },
+  );
+
+  return response.json();
+}
 
+/**
+ * This function is used to update the membnership role
```
rhymes, line 153: small typo in `membership`

app/javascript/chat/content.jsx
```diff
@@ -45,8 +45,6 @@ export default class Content extends Component {
         className="activechatchannel__activecontent activechatchannel__activecontent--sidecar"
         id="chat_activecontent"
         onClick={onTriggerContent}
-        role="button"
-        tabIndex="0"
```
rhymes, line 49: is there a reason why these were removed? They were explicitly added in #8513

app/models/chat_channel.rb
```diff
@@ -9,6 +9,7 @@ class ChatChannel < ApplicationRecord
   has_many :messages, dependent: :destroy
   has_many :chat_channel_memberships, dependent: :destroy
   has_many :users, through: :chat_channel_memberships
+  has_many :chat_channel_invitation_links, dependent: :destroy
```
rhymes, line 12: I think we should calld this `invitation_links`, the prefix is redundant 😊

app/models/chat_channel_invitation_link.rb
```diff
@@ -0,0 +1,12 @@
+class ChatChannelInvitationLink < ApplicationRecord
+  extend Enumerize
+
+  enumerize :status, in: %w[expired active].freeze, predicates: true
+
+  belongs_to :chat_channel
+
+  validates :url, presence: true, uniqueness: true
```
rhymes, line 8: isn't this a path and not a URL?

app/models/chat_channel_invitation_link.rb
```diff
@@ -0,0 +1,12 @@
...
+  validates :expiry_time, presence: true
```
rhymes, line 9: I would follow the convention and call this `expiry_at`

app/services/create_chat_channel_invitation_link.rb
```diff
@@ -0,0 +1,16 @@
+class CreateChatChannelInvitationLink < ApplicationService
```
rhymes, line 1: see my other comment on moving this in `ChatChannels::CreateInvitationLink`

app/services/create_chat_channel_invitation_link.rb
```diff
@@ -0,0 +1,16 @@
...
+  def initialize(chat_channel)
+    @chat_channel = chat_channel
+  end
+
+  attr_accessor :chat_channel, :invitation_link
+
+  def perform
+    slug = "invitation-link-" + rand(100_000).to_s(26)
```
rhymes, line 9: `"invitation-link-#{SecureRandom.hex(3)}` is probably clearer

app/services/create_chat_channel_invitation_link.rb
```diff
@@ -0,0 +1,16 @@
...
+    url = "/chat_channel_memberships/join_channel_invitation/#{chat_channel.slug}?invitation_slug=#{slug}"
+    expiry_time = DateTime.now + 1.day
```
rhymes, line 11: `1.day.from_now`, we still need to respect time zones

config/environments/test.rb
```diff
@@ -63,6 +63,9 @@
   # Debug is the default log_level, but can be changed per environment.
   config.log_level = :debug
 
+  # App domain
+  config.app_domain = "localhost:3000"
```
rhymes, line 67: mmm I'm not sure we should keep this in

package.json
```diff
@@ -146,6 +146,8 @@
     "lodash.debounce": "4.0.8",
     "postscribe": "^2.0.8",
     "preact": "^10.4.4",
+    "preact-render-spy": "^1.3.0",
+    "preact-render-to-json": "^3.6.6",
```
rhymes, line 150: cc @nickytonline

spec/models/chat_channel_invitation_link_spec.rb
```diff
@@ -0,0 +1,5 @@
+require "rails_helper"
```
rhymes, line 1: this can be removed if it doesn't contain tests

package.json
nickytonline: @sarthology and @narender2031, we no longer use `preact-render-spy` and p`react-render-to-json`. Merge latest of master to ensure everything is up to date.

There is a [changelog post on DEV](https://dev.to/devteam/changelog-frontend-edition-30l7) detailing all the changes.

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -24,15 +24,25 @@ def chat_channel_info
     @membership = ChatChannelMembership.find(params[:id])
     authorize @membership
     @channel = @membership.chat_channel
-    data = ChatChannelDetailPresenter.new(@channel, @membership).as_json
-
-    render json: { success: true, result: data, message: "" }, success: :ok
+    invitation_link = @channel.chat_channel_invitation_links.last
```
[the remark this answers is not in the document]
narender2031: @rhymes  I think we only need the last record. The reason behind it is always the last record that was `active` or `expired`. If we call all records then we need to find the active or expire record.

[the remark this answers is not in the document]
rhymes: Sorry, I meant the name of the relation, to be called `invitation_links` instead of `chat_channel_invitation_links` as `@chat_channel.` already qualifies them

Gemfile
```diff
@@ -36,6 +36,7 @@ gem "dry-struct", "~> 1.2" # Typed structs and value objects
 gem "elasticsearch", "~> 7.8" # Powers DEVs core search functionality
 gem "email_validator", "~> 2.0" # Email validator for Rails and ActiveModel
 gem "emoji_regex", "~> 3.0" # A pair of Ruby regular expressions for matching Unicode Emoji symbols
+gem "enumerize", "~> 2.3", ">= 2.3.1" # Enumerated attributes
```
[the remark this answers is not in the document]
mstruve: Oh, yes please! If it is possible this would be great, each gem increases our complexity so the less the better!

narender2031:
@rhymes PR suggestions are done.

app/presenters/chat_channel_detail_presenter.rb
```diff
@@ -1,10 +1,11 @@
 class ChatChannelDetailPresenter
```
rhymes, line 1: I didn't notice this pattern in the previous review but I wonder if we shouldn't remove presenters and just use jbuilder for rendering. They seem to be just classes used to serialize to JSON.

See https://github.com/thepracticaldev/dev.to/blob/master/app/views/articles/index.json.jbuilder for example

spec/support/fixtures/approvals/liquid_tweet_tag_spec.received.html
```diff
@@ -0,0 +1,44 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
```
rhymes, line 1: this probably slipped in

app/javascript/chat/content.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/click-events-have-key-events */
```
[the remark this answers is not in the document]
narender2031: I think we can remove this line now eslint issue is fixed not

app/presenters/chat_channel_detail_presenter.rb
narender2031: To build the response for the API this will help to build response according to our use case.

rhymes: Sure, but I think it can be done directly in jbuilder as it's essentially a class that serializes to JSON. I'm just trying not to introduce too many different patterns in a code base that already has too many :)

narender2031: done

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -157,7 +209,15 @@ def respond_to_invitation(previous_status)
       flash[:settings_notice] = "Invitation rejected."
     end
 
-    membership_user = MembershipUserPresenter.new(@chat_channel_membership).as_json
+    membership_user = {
```
mstruve, line 212: Could you use the ChatChannelMembershipHelper below here as well?

app/models/invitation_link.rb
```diff
@@ -0,0 +1,12 @@
+class InvitationLink < ApplicationRecord
```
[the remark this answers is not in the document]
mstruve: I definitely think we should be using Redis for this and not creating a whole new table in Postgres. 

I think we want something similar to how we destroy users. When a user requests to be destroyed we set a token in Redis with a 12 hour expiration. If that token is present when we attempt to destroy the user, we are good to go. If that token is missing then they have to request again. 
Setting the token:
```ruby
module Users
  module RequestDestroy
    module_function

    def call(user)
      token = SecureRandom.hex(10)

      Rails.cache.write("user-destroy-token-#{user.id}", token, expires_in: 12.hours)

      NotifyMailer.with(user: user, token: token).account_deletion_requested_email.deliver_now
    end
  end
end
```
Then in the controller we validate that the token in Redis matches the token in the link. 
```ruby
def confirm_destroy
  destroy_token = Rails.cache.read("user-destroy-token-#{@user.id}")
  raise ActionController::RoutingError, "Not Found" unless destroy_token.present? && destroy_token == params[:token]

  set_tabs("account")
end
```

narender2031:
@mstruve In our case Redis key should be unique & it should be changed every time you write on Redis, To prevent the user to use expiry link. In that case, we need to save the key in our DB. I think it should be best if we add a new field in the chat channel that updates every time when a new key is generated.

app/services/chat_channels/create_invitation_link.rb
```diff
@@ -0,0 +1,15 @@
+module ChatChannels
+  module CreateInvitationLink
+    module_function
+
+    def call(chat_channel)
+      invitation_slug = "invitation-link-#{SecureRandom.hex(3)}"
+      chat_channel.update(invitation_slug: invitation_slug)
+      unless chat_channel.errors.any?
+        path = "/chat_channel_memberships/join_channel_invitation/#{chat_channel.slug}?invitation_slug=#{invitation_slug}"
```
mstruve, line 9: since the `invitation_slug` is in the path that we are going to use as an invite we dont need to store it in the database. When the link is clicked we can take the `invitation_slug` param, check if it is in Redis and if it is, complete the invite and if it is not, then we know it was expired.

db/schema.rb
```diff
@@ -643,6 +644,18 @@
     t.index ["provider", "user_id"], name: "index_identities_on_provider_and_user_id", unique: true
   end
 
+  create_table "invitation_links", force: :cascade do |t|
```
mstruve, line 647: This should be removed

app/services/chat_channels/create_invitation_link.rb
narender2031: @mstruve 
`app/controllers/chat_channel_memberships_controller.rb`
`line no. 27`
We need to find the path to send as a response to the frontend. In that case, we don't have a `path`. So `invitation_slug` is required to save in DB. If I don't have the `invitation_slug`, It's not possible to find the path from Redis.

In this method, 1st we try to find the path using the invitation_slug. If invitation slug is not present in DB. That means there no path saved in Redis. In that case, we create the path and save it in Redis.

**Case 1** If the path is not present in the Redis `or` no Invitation slug, then we create a path and save on Redis, Also save the `invitation_slug` in DB. 
**Case 2** If we have the `invitation_slug` in DB. We read the REDIS with the `invitation_slug`. If path present in Redis. then pass to response else create a path in Redis with new `invitation_slug`

mstruve: > 1st we try to find the path using the invitation_slug. If invitation slug is not present in DB. That means there no path saved in Redis. In that case, we create the path and save it in Redis.

Rather than checking the DB, simply check Redis, if there is no path in Redis then that means the previous invite has expired and we need to generate a new one. If there is a path in Redis then we simply read that.

narender2031: @mstruve, We need the path with the unique `invitation_slug`. So if we don't know the `lnvitation_sug`. we cannot search on Redis.

**Reason**
  - If our path doesn't have a unique invitation_slug. It will be the same when you generate a new record in Redis. In that case if someone has the old URL/path after 12 hours he uses that link. He can use it to join the chat Channel. Because after 12 hours a new record will have an old path. 

That's why it is required to use the invitation_slug. It will change every 12 hours when the old expire.

mstruve: Here is what I am thinking
```ruby
def chat_channel_info
    @membership = ChatChannelMembership.find(params[:id])
    authorize @membership
    @channel = @membership.chat_channel
    
    invite_cache_key = "chat-channel-invite-#{chat_channel.id}" 
    invitation_slug = Rails.cache.fetch(invite_cache_key, expires_in: 12.hours) do 
        "invitation-link-#{SecureRandom.hex(3)}"
      end
    @invitation_link = "/join_channel_invitation/#{chat_channel.slug}?invitation_slug=#{invitation_slug}"

    if result&.errors&.any?
      render json: { success: false, errors: result.errors.full_messages }, status: :bad_request
    end
  end
```

The slug is stored in Redis and we fetch it using the chat channel ID. I do this using a handy little helper called [`fetch`](https://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html#method-i-fetch). This will read the cache key if it is present. If it is not present it will generate a new cache key. 

Once this is in place the next step is reading it when someone tries to use the link. 
```ruby
def join_channel_invitation
    @chat_channel = ChatChannel.find_by(slug: params[:channel_slug])
    authorize @chat_channel
    
    invite_cache_key = "chat-channel-invite-#{chat_channel.id}"  # could pull this out into a method
    invitation_slug = Rails.cache.read(invite_cache_key)
    
    if invitation_slug == params[:invitation_slug]
      # add them to the channel
    else
      # Show an error message that the invitation link has expired.
    end
  end
```

Let me know what you think!!!

app/javascript/chat/chat.jsx
```diff
@@ -1794,6 +1794,11 @@ export default class Chat extends Component {
         ? 'sidecar-user'
         : `chat_channel_setting`;
 
+    const contentLink =
+      activeChannel.channel_type === 'direct'
```
benhalpern, line 1798: I'm getting a runtime error now here...

```javascript
Uncaught TypeError: activeChannel is null
```

Unless it's just something wrong locally for me.

app/services/chat_channels/create_invitation_link.rb
narender2031: I got your point. Looks fine to me. I will update my code according to that.
 I was trying to save the full path on Radis.

app/javascript/chat/chat.jsx
narender2031: @benhalpern I am not getting this error locally. I will try to reproduce it.

fdoxyz:
I'm having trouble testing the invitation link, for some reason the `Rails.cache.fetch` is not persisting the element in Redis in my local environment. Every time after sharing the link with a different user (incognito window) the record is not there and the server responds that the invitation has expired (because the `Rails.cache.read` returns `nil`). I added these puts statements to try and figure out the problem:

```
# Inside ChatChannelMembershipsController#chat_channel_info

res = Rails.cache.fetch('test1', expires_in: 12.hours) do
  'this should be available later'
end
puts "WRITING"
puts invite_cache_key
puts invitation_slug
puts Rails.cache.read(invite_cache_key)
puts res
puts Rails.cache.read('test1')
```

Screen Shot 2020-07-10 at 08 30 17

```
# Inside ChatChannelMembershipsController#join_channel_invitation

puts "READING"
puts invite_cache_key
puts Rails.cache.read(invite_cache_key)
puts Rails.cache.read('test1')
```

Screen Shot 2020-07-10 at 08 30 40

Maybe this is a Redis problem in my local environment 🤔 Not too sure yet, but I'll try re-installing/updating

----------

Anyways this definitely feels closer to ship now, but I think splitting the PR similar to what @mstruve suggested before would help finish off easier and faster. Instead of frontend vs backend PRs my suggestion would be:

- First PR for the 'Chat Channel Membership Manager' without the invitation link (which already seems ready to me)
- Second PR adding the invitation link (would include the backend logic + adding the link to the component)

I'm suggesting this because the membership manager feature is very useful on its own (to moderate larger channels) and seems solid from what I was able to test, users would still be able to request join with what we currently have without the invitation link. The second PR would build on top of that existing component and we can focus on that without having that big of a big PR. Let me know what you think @narender2031 @sarthology

rhymes:
@fdoxyz 

> Maybe this is a Redis problem in my local environment 🤔 Not too sure yet, but I'll try re-installing/updating

nope, all is fine, Redis is not enableby default locally. Activate the Redis cache with `rails app:cache` see also

https://github.com/thepracticaldev/dev.to/blob/879e0ae1c893db33b6a31dd4ad92921072e9f78b/config/environments/development.rb#L19-L20

fdoxyz:
And that is an immediate fix... Thanks @rhymes! 🙏🏼 The invitation worked as expected now

I'll take another look around to try things out now

narender2031:
@fdoxyz If everything works fine. Then can we continue with this PR? I am thinking that once this PR merged. Then we can focus on find bugs and fix them, before CodeLand. If you still facing any issue. Then we can divide the PR. 
Please add your suggestions.

fdoxyz:
I ran a test with a user `fdoxyz` as mod, and then followed the invite link with user `unavailable1991`. That flow worked well, but for some reason a third user got added to the channel without having interacted with it. Here's a video of the test:

https://share.getcloudapp.com/DOuA1eGn

The only other thing I have found is that the invitation link only works on discoverable channels, which may be done that way by design (or by how the discoverable vs non-discoverable channels work). Not necessarily a problem but asking in case this is considered a bug.

About whether or not splitting the PR, I think @nickytonline was going to review the PR too so it maybe better to wait for his input and then make a decision on this

narender2031:
> https://share.getcloudapp.com/DOuA1eGn

The reason behind it:- There is a hard-code setting in local development. When you set your setting discoverable it adds the user has an `ID 1`

Please check the 
Controller: `/app/controllers/chat_channels_controller.rb` 
Line no: 45

It is an old functionality. It set the mascot_user_id, which is default 1.

narender2031:
> The only other thing I have found is that the invitation link only works on discoverable channels, which may be done that way by design (or by how the discoverable vs non-discoverable channels work). Not necessarily a problem but asking in case this is considered a bug.

Reason: - You cannot search the group. If it is not discoverable. That's why I think If the group is not discoverable it should not allow the user to join the group. 

@fdoxyz Please add your suggestions

app/javascript/chat/chat.jsx
fdoxyz: FWIW I didn't run into this problem locally either

narender2031:
> My only suggestion in this case would be to hide the invitation link from the component if the channel is not discoverable. This way users won't share the link and find it doesn't work, since there's no way for them to know this is causing the problem. How does this sound?

@fdoxyz  It sounds good to me. I can implement this. Do we need confirmation before the start or should I start implementing it?

fdoxyz:
@benhalpern do you think we will want to have the invitation links work for non-discoverable channels?

If yes: It's probably best to make the invitation links work for non-discoverable channels and leave the component as-is.

If no: We should go ahead and hide the invitation link for non-discoverable channels like suggested above

narender2031:
> @benhalpern do you think we will want to have the invitation links work for non-discoverable channels?
> 
> If yes: It's probably best to make the invitation links work for non-discoverable channels and leave the component as-is.
> 
> If no: We should go ahead and hide the invitation link for non-discoverable channels like suggested above

@benhalpern any updates on this?

fdoxyz:
Hi again @narender2031, I gave this another thought and I'm confident we can move forward making the link work for non-discoverable chats, it would indeed be very valuable. Custom private group chats would benefit from this feature, for example the one from #9234 .

So I think this is the best route to take in order to get this PR finished:

> If yes: It's probably best to make the invitation links work for non-discoverable channels and leave the component as-is.

With this change the PR would look good to approve from me 😃. Also this may be the easiest from the two options I had originally suggested since it would only take a small change in the policy, right?

app/policies/chat_channel_policy.rb
```diff
@@ -39,6 +39,18 @@ def update_channel?
     user_can_edit_channel
   end
 
+  def join_channel_invitation?
+    record.present? && user.id && record.discoverable
```
fdoxyz, line 43: I haven't tested it locally yet, but just by removing the discoverable check could make that last suggestion work 🙂

narender2031:
> With this change the PR would look good to approve from me 😃. Also, this may be the easiest from the two options I had originally suggested since it would only take a small change in the policy, right?
Yes. There is a small change in policy. I will update this update the PR with this change.

app/policies/chat_channel_policy.rb
narender2031: yes

narender2031:
@fdoxyz PR is updated with the suggested changes

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -131,14 +133,73 @@ def leave_membership
...
+    if params[:user_action] == "accept"
+      membership = ChatChannelMembership.find_by(user_id: current_user.id, chat_channel_id: chat_channel.id)
+      if !membership
+        membership = ChatChannelMembership.new(user_id: current_user.id, chat_channel_id: chat_channel.id)
+        membership.save
+        unless membership&.errors&.any?
+          send_chat_action_message("@#{membership.user.username} join the channel", current_user, chat_channel.id,
+                                   "joined")
+        end
+      elsif membership.status != "active"
```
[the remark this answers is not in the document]
fdoxyz: I think this is taking into account people that have requested access but have not been accepted yet (status = "pending" or "joining_request" not entirely sure), or people that have been invited but haven't accepted yet.

I tried it out locally, where I had requested access first (not accepted) and then joined using the link. The request was not left lingering (worked out as expected), but anyways would be great to hear if that's not the case.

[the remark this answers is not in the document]
narender2031: @nickytonline  & @fdoxyz there is an `inactive` option for `chatChannelMembershi` . But this also check checks if the user already has the chatChannelMembership with the status `pending` or `joining_request`. Then update it to as `active`.

narender2031:
@nickytonline There is one spec is failing. These specs work fine locally. Also, this is not relevant to the PR.

rhymes:
I restarted the build

narender2031:
> I restarted the build

@rhymes There is something that must be wrong in master. Because I am facing a similar issue in other PR's.

app/assets/stylesheets/chat.scss
```diff
@@ -1164,3 +1165,70 @@
 .channel_details {
   margin-top: 30px;
 }
+
+/* .view-membership-btn {
```
nickytonline, line 1169: This can be removed.

app/controllers/chat_channel_memberships_controller.rb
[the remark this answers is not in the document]
nickytonline: Consider adding a comment explaining this in the code.

app/javascript/chat/ChatChannelSettings/ActiveMembershipsSection.jsx
```diff
@@ -1,48 +1,53 @@
...
     <div
       data-testid="active-memberships"
       className="p-4 grid gap-2 crayons-card mb-4"
       data-active-count={activeMemberships ? activeMemberships.length : 0}
     >
       <h3 className="mb-2 active_members">Members</h3>
-      {activeMemberships && activeMemberships.length > 0
-        ? activeMemberships.map((pendingMembership) => (
+      {activeMembershipList && activeMembershipList.length > 0
+        ? activeMembershipList.map((activeMembership) => (
```
nickytonline, line 23: `activeMembershipList` will always be an array based on this line, https://github.com/forem/forem/pull/8945/files#diff-9c84ece864f7060e4ad87d13a14365fdR13

So the existence and length checks are unnecessary. If the array length is zero, then nothing gets mapped.

```suggestion
      {activeMembershipList.map((activeMembership) => (
```

app/javascript/chat/ChatChannelSettings/ActiveMembershipsSection.jsx
```diff
@@ -1,48 +1,53 @@
...
+              membership={activeMembership}
               removeMembership={removeMembership}
               membershipType="active"
               currentMembershipRole={currentMembershipRole}
               className="active-member"
             />
           ))
         : null}
+      <div className="row align-center">
+        <button
```
nickytonline, line 34: Use the [design system buttons](https://storybook.dev.to/?path=/story/3-components-buttons--default).

`import { Button } from '@crayons'`

app/javascript/chat/ChatChannelSettings/ChatChannelSettings.jsx
```diff
@@ -149,6 +157,7 @@ export default class ChatChannelSettings extends Component {
 
   updateMemberships = (membershipId, response, membershipStatus) => {
     if (response.success) {
+      this.componentDidMount();
```
nickytonline, line 160: Avoid calling component lifecycle methods explicitly. If there is common functionality that can be reused, wrap it in a function/method that can be used here as well as in `componentDidMount()`.

app/javascript/chat/ChatChannelSettings/ChatChannelSettings.jsx
```diff
@@ -305,6 +314,54 @@ export default class ChatChannelSettings extends Component {
...
+
+  handleUpdateMembershipRole = async (e) => {
+    const { membershipId, role } = e.target.dataset;
+    const { chatChannel } = this.state;
+    const response = await updateMembershipRole(
+      membershipId,
+      chatChannel.id,
+      role,
+    );
+    const { message } = response;
```
nickytonline, line 334: Since `message` is being destructured, might as well destructure `success` as well.

app/javascript/chat/ChatChannelSettings/ChatChannelSettings.jsx
```diff
@@ -305,6 +314,54 @@ export default class ChatChannelSettings extends Component {
...
+    if (response.success) {
+      this.componentDidMount();
```
nickytonline, line 336: Same comment as above.

app/javascript/chat/ChatChannelSettings/ChatChannelSettings.jsx
```diff
@@ -305,6 +314,54 @@ export default class ChatChannelSettings extends Component {
...
+      this.setState((prevState) => {
+        const { activeMemberships } = prevState;
+        const updatedActiveMemberships = activeMemberships.map(
+          (activeMembership) => {
+            if (activeMembership.membership_id === Number(membershipId)) {
+              // eslint-disable-next-line no-param-reassign
+              activeMembership.role = role;
```
nickytonline, line 343: To avoid mutating the existing memberships, you can do the following:

```suggestion
              return { ...activeMembership, role };
```

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -25,15 +25,17 @@ def chat_channel_info
...
     authorize @membership
     @channel = @membership.chat_channel
-    data = ChatChannelDetailPresenter.new(@channel, @membership).as_json
-
-    render json: { success: true, result: data, message: "" }, success: :ok
+    invite_cache_key = "chat-channel-invite-#{@channel.id}"
+    invitation_slug = Rails.cache.fetch(invite_cache_key, expires_in: 12.hours) do
+      "invitation-link-#{SecureRandom.hex(3)}"
+    end
+    @invitation_link = "/join_channel_invitation/#{@channel.slug}?invitation_slug=#{invitation_slug}"
```
mstruve, line 32: We need to add `invitation_slug` to our [approved list of Fastly params](https://github.com/forem/forem/blob/master/config/fastly/snippets/safe_params_list.vcl#L5) otherwise it will get stripped and this wont work.

app/javascript/chat/ChatChannelSettings/ChatChannelSettingsSection.jsx
```diff
@@ -0,0 +1,106 @@
...
+  removeMembership,
+  chatChannelAcceptMembership,
+  channelDescription,
+  chatChannel,
+  currentMembership,
+  activeMemberships,
+  pendingMemberships,
+  requestedMemberships,
+  invitationUsernames,
+  showGlobalBadgeNotification,
```
nickytonline, line 31: Just an observation, but there is quite a lot of props for this component. Membership and Channel props are being split up in the parent component. It would make sense to keep them grouped like they were in the original entities returned in the JSON payload.

Also, prefixing or using the word chat in props in here is not necessary. The name of the component they are being passed into `<ChatChannelSettingsSection ...   />` gives enough context that using `channel` is good enough for readability. e.g. `handleChatChannelInvitations` -> `handleChannelInvitations`, `handleInvitations` or even `processInvitations`

```suggestion
  updateCurrentMembershipNotificationSettings,
  handleleaveChatChannelMembership,
  handlePersonChatChennelSetting,
  handleChannelDescriptionChanges,
  handleChannelDiscoverableStatus,
  handleDescriptionChange,
  handleChatChannelInvitations,
  handleInvitationUsernames,
  toggleScreens,
  removeMembership,
  channel,
  memberships,
  invitationUsernames,
  showGlobalBadgeNotification,
```

app/javascript/chat/ChatChannelSettings/MembershipManager/InvitationLinkManager.jsx
```diff
@@ -0,0 +1,127 @@
+import { h, Component } from 'preact';
+import PropTypes from 'prop-types';
+import { Button } from '@crayons';
+
+function isNativeAndroid() {
```
nickytonline, line 5: I've seen these in a bunch of places in the code base. Consider importing them from a module so that we avoid copy pasta. For the time being import it only for stuff relevant to this PR and for other parts of the code base, it can be a follow up PR.

app/javascript/chat/ChatChannelSettings/MembershipManager/Membership.jsx
```diff
@@ -0,0 +1,128 @@
...
+
+const Membership = ({
+  membership,
+  currentMembership,
+  removeMembership,
+  handleUpdateMembershipRole,
+}) => {
+  const addAsModButton =
+    membership.role === 'member' ? (
+      <button
```
nickytonline, line 14: These can also be design system buttons as mentioned previously.

nickytonline:
@sarthology @narender203. I've finished reviewing. I'm just going to try it out locally now.

app/javascript/chat/ChatChannelSettings/ActiveMembershipsSection.jsx
```diff
@@ -19,25 +20,23 @@ const ActiveMembershipSection = ({
...
+          membership={activeMembership}
+          removeMembership={removeMembership}
+          membershipType="active"
+          currentMembershipRole={currentMembershipRole}
+          className="active-member"
+        />
+      ))}
       <div className="row align-center">
-        <button
+        <Button
```
nickytonline, line 33: 🥳 Thanks for changing to the design system button. Just FYI, you don't need to include the following CSS classes, `crayons-btn` and `crayons-btn-s`. Internally, the `<Button />` component adds the `crayons-btn` class and for the `crayons-btn-s`, just add the prop/value `size='s'`.

app/javascript/chat/ChatChannelSettings/InviteForm.jsx
```diff
@@ -26,21 +27,21 @@ const InviteForm = ({
         />
       </div>
       <div>
-        <button
+        <Button
           className="crayons-btn"
```
nickytonline, line 31: Same comment as above. The CSS class is not necessary when using the design system button.

app/javascript/chat/ChatChannelSettings/LeaveMembershipSection.jsx
```diff
@@ -13,20 +14,20 @@ const LeaveMembershipSection = ({
     <div className="crayons-card p-4 grid gap-2 mb-4 leave_membership_section">
       <h3>Danger Zone</h3>
       <div>
-        <button
+        <Button
           className="crayons-btn crayons-btn--danger leave_button"
```
nickytonline, line 18: Same as mentioned above and for a danger button, use the `variant` prop.

```suggestion
          className="leave_button"
          variant="danger"
```

app/javascript/chat/ChatChannelSettings/MembershipManager/Membership.jsx
```diff
@@ -11,33 +11,33 @@ const Membership = ({
 }) => {
   const addAsModButton =
     membership.role === 'member' ? (
-      <button
+      <Button
```
nickytonline, line 14: Same as previously mentioned about CSS classes.

app/javascript/chat/ChatChannelSettings/MembershipManager/Membership.jsx
```diff
@@ -11,33 +11,33 @@ const Membership = ({
...
       >
         Promote to Mod
-      </button>
+      </Button>
     ) : null;
 
   const addAsMemberButton =
     membership.role === 'mod' ? (
-      <button
+      <Button
```
nickytonline, line 27: Same as previously mentioned about CSS classes.

app/javascript/chat/ChatChannelSettings/MembershipManager/Membership.jsx
```diff
@@ -11,33 +11,33 @@ const Membership = ({
...
       >
         Remove Mod
-      </button>
+      </Button>
     ) : null;
 
   const removeMembershipButton =
     membership.role === 'member' ? (
-      <button
+      <Button
```
nickytonline, line 40: Same as previously mentioned about CSS classes.

app/javascript/chat/ChatChannelSettings/PersonalSetting.jsx
```diff
@@ -16,20 +17,20 @@ const PersonalSettings = ({
...
-          onChange={handlePersonChatChennelSetting}
+          onChange={handlePersonChannelSetting}
         />
         <label htmlFor="c3" className="crayons-field__label">
           Receive Notifications for New Messages
         </label>
       </div>
       <div>
-        <button
+        <Button
```
nickytonline, line 27: Same as previously mentioned about CSS classes.

app/javascript/chat/ChatChannelSettings/SettingsForm.jsx
```diff
@@ -42,13 +43,13 @@ const SettingsFrom = ({
         </label>
       </div>
       <div>
-        <button
+        <Button
```
nickytonline, line 46: Same as previously mentioned about CSS classes.

app/javascript/utilities/validateAndroidNative.js
```diff
@@ -0,0 +1,7 @@
+export function isNativeAndroid() {
```
nickytonline, line 1: Thanks for exporting this! Can you add a JS Doc comment to the function?

narender2031:
@nickytonline `Button` unnecessary classes are removed.

app/javascript/utilities/validateAndroidNative.js
narender2031: added.

narender2031:
@nickytonline any updates on this PR?

nickytonline:
I have latest of the branch and when I log on as another user with the invitation link, it says it's expired. Based on the UI, the invitation link looks generic and can be sent to anyone. Is that the case?

Here's what I get when using the invitation link right away.

image

nickytonline:
Also, if there is only one member in the channel (the mod), they shouldn't be able to remove themselves as the channel would become abandoned then.

image

nickytonline:
Also, it's not clear to me if the user is sent a notification when they are promoted or demoted as a mod.

narender2031:
> I have latest of the branch and when I log on as another user with the invitation link, it says it's expired. Based on the UI, the invitation link looks generic and can be sent to anyone. Is that the case?
> 
> Here's what I get when using the invitation link right away.

@nickytonline Did you start the cache in development mode?

narender2031:
> Also, if there is only one member in the channel (the mod), they shouldn't be able to remove themselves as the channel would become abandoned then.

Ok. I will change according to this. Also, there is one more case that we already discussed a MOD user can demote to another user. 

I will update this change, If there is one user then he cannot remove or demote himself.

narender2031:
> Also, it's not clear to me if the user is sent a notification when they are promoted or demoted as a mod.

Did you mean? If the user is promoted/demoted it should receive notification or text message in a chat?

fdoxyz:
@nickytonline like @narender2031 said, I had the same issue because I hadn't enabled the cache in development. If you haven't yet try `rails dev:cache`

See https://github.com/forem/forem/pull/8945#issuecomment-656714758

fdoxyz:
> Did you mean? If the user is promoted/demoted it should receive notification or text message in a chat?

My suggestion would be to add as message in chat (similar to when a user joins the channel), because of consistency and easier to implement.

My reasoning is that we want this feature to moderate during codeland, and if this will delay the PR from being merged for too long it may be worth tackling in a separate PR as it only affects when promoting/demoting mods.

nickytonline:
I'll try the `rails dev:cache`.

> > Did you mean? If the user is promoted/demoted it should receive notification or text message in a chat?
> 
> My suggestion would be to add as message in chat (similar to when a user joins the channel), because of consistency and easier to implement (not feeling too strongly about either option though).
> 
> My reasoning is that we want this feature to moderate during codeland, and if this will delay the PR from being merged for too long it may be worth tackling in a separate PR as it only affects when promoting/demoting mods.

Sounds good @fdoxyz 

I'm off in a couple of hours and all of next week, so I will approve the change requests. I'll let you continue the discussions here if there are any other loose ends @fdoxyz.

narender2031:
@fdoxyz , @nickytonline  I will update these suggestion within one hour.

sarthology:
> My reasoning is that we want this feature to moderate during codeland, and if this will delay the PR from being merged for too long it may be worth tackling in a separate PR as it only affects when promoting/demoting mods.


I believe this should be separate PR considering Codeland is almost here.

narender2031:
@fdoxyz @nickytonline  PR is updated as we discussed for the suggestions.

benhalpern:
Not sure what the failing test is about. @fdoxyz @nickytonline any idea?

@sarthology @narender2031 what is the order we should be looking to merge stuff?

fdoxyz:
I ran it locally and it did pass, it does looks like a problem not related to the PR. I just restarted the build.

narender2031:
@benhalpern i think we can merge this PR, once @nickytonline  or @fdoxyz review the changes as they suggested.

narender2031:
@fdoxyz can you please restart one more time. If it fails again then I will test again locally. But I am pretty much sure it’s not related to the PR.

fdoxyz:
@narender2031 can you instead sync with master? That will also restart the build and I think some recent changes might be causing this

nickytonline:
Looks like flaky tests? Anyways I already approved. Thanks for all the work on this @sarthology and @narender2031!

fdoxyz:
I'm seeing this in the CI logs

Screen Shot 2020-07-17 at 15 44 31

Not sure what's causing this right now because `master` was just merged into the branch, but I'll continue to investigate.

fdoxyz:
@narender2031 and I have been debugging this issue and seems like knapsack is causing the build to fail. This hasn't been a problem with other PRs which makes it odd, but the error says `Failed to open TCP connection to api-disabled-for-fork.knapsackpro.com:443`. 

So it seems like the API is disabled on forks? Maybe a config needs to be tweaked for knapsack to work here? Whenever you have a moment to take a looks at this @mstruve we would really appreciate it.

Screen Shot 2020-07-17 at 21 10 51

fdoxyz:
Now after some time away from the screen and a second look seems like knapsack is not the problem, they're only warnings and the failure comes way before or after the knapsack warnings (also found & read [Molly's post](https://dev.to/molly_struve/using-knapsackpro-to-parallelize-your-tests-1lbe) explaining the warning 🙂). 

It may be indeed a flaky spec:

When I checkout this PR locally and run `bundle exec rspec spec/system/comments/user_delete_a_comment_spec.rb` it's now failing __but with a different error than the one logged in Travis__.

This is what I see locally on that failing spec:

Screen Shot 2020-07-18 at 09 47 09

Which happens when a comment is deleted (`bust_cache_worker` seems to request an index). This doesn't explain why Travis is logging a different error though, and that's the most confusing part for me at the moment.

narender2031:
@fdoxyz The reason for failing the spec locally might be your elastic search is not running. This spec work fine locally for me. 

It might be a flaky spec. 
I am not sure why it is failing. Need to reproduce this issue locally for the fix.

narender2031:
> Got the build to pass, code looks good. 😅

Thanks @benhalpern 
Finally, It is ready to ship  😅

fdoxyz:
> Got the build to pass, code looks good. 😅

A restart build did the trick? You got the magic touch 😂🚀

Issue #8969, opened 2020-06-28
shadow of crayons-notice class feels out-of-place in flat design


thorstenhirsch:
The design of dev.to has made huge progress, I really like it. One thing I just stumbled upon is the shadow of the crayons (like the red "Unpublished Post. This URL is public but secret, ..." box).

```
.crayons-notice--danger {
  background: var(--accent-danger-a10);
  border: 2px solid var(--accent-danger));
  box-shadow: 2px 2px 0 var(--accent-danger));
  color: var(--card-color);
  border-radius: var(--radius);
}
```

It seems out of place now. Please remove the 2px 2px shadow of the crayons-notice class. In the code I've seen that there's a mixin that takes care of setting the shadow: generate-box() with $level as the 1st argument. I'll provide a PR with the easiest solution: calling generate-box() with $level=0 when generating all crayons-notice boxes. However you might think of removing $level completely in generate-box(), as there seems to be no place in the new design anymore for boxes with shadow, I guess.

citizen428:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

PR #8970, opened 2020-06-28
set level=0 when generating boxes for .crayons-notice

thorstenhirsch:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
shadow of crayons-notice class feels out-of-place in flat design

#8969

- check out the shadow of the red box that says "Unpublished Post. This URL is public but secret, ..." when looking at an unpublished post.

## Added tests?

- [ ] yes
- [x] no, because they aren't needed (only design is affected)
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

alt_text

### DIFF
--- a/app/assets/stylesheets/components/notices.scss
+++ b/app/assets/stylesheets/components/notices.scss
@@ -3,7 +3,7 @@
 .crayons-notice {
   padding: var(--su-4);
   @include generate-box(
-    $level: 1,
+    $level: 0,
     $bg: var(--card-bg),
     $border: var(--card-color),
     $color: var(--card-color)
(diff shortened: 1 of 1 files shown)


thorstenhirsch:
Aw, crap, I used the wrong email address in my commit. I just added it to my github account, so that the CLA check runs OK, but is there any way to update this PR with my primary email address?

ludwiczakpawel:
I'll let myself close this PR. Even though the shadow may look a little bit out of place right now, it's still kinda _on brand_. We do have primarily flat and minimalistic style but we try to keep some UI nuances referring to the _old-school-cool_ retro UI in the games and software. We will probably rethink this direction and design choices when/if we start working on rebranding. But as of today, that _3d-retro_ style is tightly connected with our core values and roots so lets keep it as it is for now.

I'll will not lock the conversation so we can keep discussing here if there are any questions or concerns.

Issue #3777, opened 2019-08-20
Lint issues with JavaScript Files


nickytonline:
Describe the bug
This compliments work being done in #2470.

The *.js files inside app/javascript and app/assets/javascripts still have plenty of lint issues

To Reproduce

From the command line in the app/javascript folder, run `eslint . --ext .js`
From the command line in the app/assets/javascripts folder, run `eslint . --ext .js`

Expected behavior
There shouldn't be any lint issues.

Time to do some cleanup 😆

![Cleanup](https://media.giphy.com/media/3oKIPCSX4UHmuS41TG/giphy-downsized.gif)

Amorpheuz:
I would love to help out with this! I assume it will be much cleaner/smoother to fix a file, create a PR and then move on to the next, or should each fix be a commit in a PR?

nickytonline:
Just create a PR and link it to this issue. Thanks for helping! 👏

palledorous:
Do you still need help with this? I'd love to take on some refactoring.

rhymes:
Hi @palledorous, please go ahead, I'm going to copy my recommendation from the other PR:

**to make reviews easier and also curb the possibility of introducing bugs, is not to tackle too many files at the same time.**

I went ahead and grouped the results of `yarn run eslint app/assets/javascripts --ext ".js"` and `yarn run eslint app/javascript --ext ".js"`:

- [ ] /app/assets/javascripts/initializers/initializeAllFollowButts.js
- [ ] /app/assets/javascripts/initializers/initializeAllTagEditButtons.js
- [ ] /app/assets/javascripts/initializers/initializeArchivedPostFilter.js
- [ ] /app/assets/javascripts/initializers/initializeArticleDate.js
- [ ] /app/assets/javascripts/initializers/initializeArticleReactions.js
- [ ] /app/assets/javascripts/initializers/initializeBaseUserData.js
- [ ] /app/assets/javascripts/initializers/initializeBodyData.js
- [ ] /app/assets/javascripts/initializers/initializeCommentDate.js
- [ ] /app/assets/javascripts/initializers/initializeCommentDropdown.js
- [ ] /app/assets/javascripts/initializers/initializeCommentPreview.js
- [ ] /app/assets/javascripts/initializers/initializeCreditsPage.js
- [ ] /app/assets/javascripts/initializers/initializeDashboardSort.js
- [ ] /app/assets/javascripts/initializers/initializeDrawerSliders.js
[...]

PR #8972, opened 2020-06-28
fix lint errors in initializeSwipeGestures

Amorpheuz:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Fixes eslint errors in `app/assets/javascripts/initializers/initializeSwipeGestures.js`.

*P.S. not sure if a hook reformatted the file or it was on my end! Will rollback if needed.*

#3777 

Simulate a touch device of Tablet and Mobile sizes and perform the following actions on home page:
- `left swipe` to trigger the left panel (with login, etc) and `right swipe` to close it.
- `right swipe` to trigger the right panel (with tags, more links, etc) and `left swipe` to close it.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

colorful text animated with swipe left, swipe right

### DIFF
--- a/app/assets/javascripts/initializers/initializeSwipeGestures.js
+++ b/app/assets/javascripts/initializers/initializeSwipeGestures.js
@@ -1,4 +1,4 @@
-/* global slideSidebar */
+/* global slideSidebar, swipeState: true */
 'use strict';
 
 function slideContent(d) {
--- a/app/assets/javascripts/initializers/initializeSwipeGestures.js
+++ b/app/assets/javascripts/initializers/initializeSwipeGestures.js
@@ -1,61 +1,116 @@
-function initializeSwipeGestures(){
+/* global slideSidebar */
+'use strict';
+
+function initializeSwipeGestures() {
   initializeSwipeGestures.called = true;
-  swipeState = "middle";
-  setTimeout(function(){
-
+  let swipeState = 'middle';
+  setTimeout(function () {
     // window.onload=function(){
-    (function(d){
-    var
-    ce=function(e,n){var a=document.createEvent("CustomEvent");a.initCustomEvent(n,true,true,e.target);e.target.dispatchEvent(a);a=null;return false},
-    nm=true,sp={x:0,y:0},ep={x:0,y:0},
-    touch={
-      touchstart:function(e){sp={x:e.touches[0].pageX,y:e.touches[0].pageY,scrollY:window.scrollY}},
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/initializers/initializeSwipeGestures.js
```diff
@@ -1,61 +1,116 @@
...
-      touchmove:function(e){nm=false;ep={x:e.touches[0].pageX,y:e.touches[0].pageY,scrollY:window.scrollY}},
-      touchend:function(e){
-        if(nm){ce(e,'fc')}
-        else {
-          var x=ep.x-sp.x,xr=Math.abs(x),y=ep.y-sp.y,yr=Math.abs(y),absScroll=Math.abs(sp.scrollY-ep.scrollY);
-          if(Math.max(xr,yr)>15){
-            var shouldScroll = (xr/2)>yr && absScroll < 5;
-            ce(e,(shouldScroll?(x<0?'swl':'swr'):(y<0?'swu':'swd')))
-          }
+    (function (d) {
```
rhymes, line 9: do you think we could move this function in a named function and then call it on line 68? It should probably fix Codeclimate error as well.

cc @nickytonline

Amorpheuz: I feel that the `function (d)` as a whole forms an integral part the `initializeSwipeGestures()`... Rather moving the function as a whole, moving the eventHandlers for `touchstart`, `touchend`, `touchmove` and `touchcancel` should make more sense as independent units.

Also, the independent `function (d)` will exceed the codeclimate 50 line limit as well.

app/assets/javascripts/initializers/initializeSwipeGestures.js
```diff
@@ -1,61 +1,112 @@
-function initializeSwipeGestures(){
-  initializeSwipeGestures.called = true;
-  swipeState = "middle";
-  setTimeout(function(){
+/* global slideSidebar */
```
nickytonline, line 1:
```suggestion
/* global slideSidebar, swipeState: true */
```

app/assets/javascripts/initializers/initializeSwipeGestures.js
nickytonline: @rhymes, this code looks like it was pulled from somewhere that had it minified already. Since this PR is just about fixing the lint issues, I would address this in another PR if we are to pursue it.

PR #8975, opened 2020-06-29
Refactor code blocks styling

### LINKED ISSUE #8946: How to highlight text (without word break) in a dev.to article as an editor?
Hello,

I don't really know where to ask this question.

I am just trying to write an article on Dev.to.   Sometimes I like to highlight a word by using backticks.

So I might highlight a word like this:  `myword`.   This works but the problem I run into is that sometimes line breaks happen right in the middle of words that I highlight like this.    So that 'mywor' is at the end of one line and 'd' starts at the beginning of the next line.

How can prevent line breaks happening right in the middle of words I highlight?  Or is there a different way to highlight words in dev.to articles?

ludwiczakpawel:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR does some frontend refactoring for all our code blocks elements. We used to overwrite codeblocks styling in several places causing some weird edge cases. AND we also used to fix those edge cases with more overwriting. All of these made code blocks being pretty hard to maintain and causing other, layout-related, issues.

This PR unifies styling for code blocks and fixes the issues I've mentioned above.

Closes #8946

Just play around with code blocks in your posts and comment.

There shouldn't be too many visual differences except I've decided to not let code blocks exceed content area (so it should be as wide as rest of the blog post content). 

Before:

image

After:

image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


app/assets/stylesheets/item-list.scss
```diff
@@ -186,7 +186,7 @@
         text-align: center;
         padding: 100px 0px;
 
-        span.highlight {
+        span.btnhighlight {
```
ludwiczakpawel, line 189: Renamed just in case, to avoid potential conflicts with code blocks.

nickytonline:
@ludwiczakpawel, can you provide before and after screenshots in the PR? Thanks!

app/assets/stylesheets/components/comments.scss
```diff
@@ -46,6 +36,12 @@
       width: auto;
       object-fit: contain;
     }
+
+    pre:not(.highlight),
+    div.highlight {
```
nickytonline, line 41: Is the `div` as part of the selector necessary?

app/assets/stylesheets/item-list.scss
nickytonline: Maybe rename to `btn--highlight`? Also, I don't think the `span` is necessary for the selector.

app/assets/stylesheets/base/reset.scss
```diff
@@ -67,7 +68,7 @@ pre,
 code,
 kbd,
 samp {
-  font-family: monospace, monospace;
+  font-family: var(--ff-accent);
```
nickytonline, line 71: Is this `nonospace`?

ludwiczakpawel:
@nickytonline I've added screenshot for code block. This PR should not change the visuals drastically.. It was more about changes under the hood. Although, yea - there are few differences:

- font-family and font-size
- code block not being edge-to-edge

app/assets/stylesheets/components/comments.scss
ludwiczakpawel: Yeap, since `.highlight` is a class name not only for DIVs but also for PREs.. that's why i have this non-standard selector a line above with `:not...`

app/assets/stylesheets/base/reset.scss
ludwiczakpawel: Yes, it's a monospace font. I think this variable name is a bit misleading - I'll rename it to `--ff-monospace` in a separate PR.

ludwiczakpawel:
@benhalpern I'm happy to bring back the "full width" blocks - it's actually very simple and it didn't cause any issues. I just thought it looks cleaner when it's limited width. But I really don't have strong opinions about it. If we decide to bring this to other block elements (images?), then I'm even more happy to bring it back.

benhalpern:
@ludwiczakpawel I just deployed this because I think that this does indeed look cleaner. At some point maybe we can follow up with my "expand" button 😄

artis3n:
FYI I am now seeing https://github.com/thepracticaldev/dev.to/issues/9028 which is maybe more of an inconvenience than a visual bug, but there it is

PR #8981, opened 2020-06-29
Do not render active announcements in iframes

vaidehijoshi:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Because of the way that the mod tools sidebar relies on iframes, any active broadcasts were being rendered twice: once on the main site, and once within the mod tools iframe, which re-invoked all the initializers and their JavaScript when the iframe itself re-rendered. Unfortunately, iframes also create their _own_ window context, which means that this issue can't simply be solved by setting a flag on the window (in other words, something like `window.broadcastVisible` wouldn't work!).

The best solution here is to not re-invoke the `initializeBroadcast` function if its caller is the mod tools sidebar; the best way to determine that is by using the path, which will include `actions_panel` within it. 😓 

Closes https://github.com/thepracticaldev/dev.to/issues/8774.

Before this change:
Screen Shot 2020-06-29 at 7 59 01 AM

After this change:
Screen Shot 2020-06-29 at 7 58 45 AM

## Added tests?

- [ ] yes
- [x] no, because they aren't needed _(I opted to not write tests for this specific functionality since it sounds like the mod actions panel will be re-written in preact soon, which means that this check won't be necessary at that point.)_
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [x] ~readme~ inline documentation
- [ ] no documentation needed

N/A

le sigh

### DIFF
--- a/app/assets/javascripts/initializers/initializeBroadcast.js
+++ b/app/assets/javascripts/initializers/initializeBroadcast.js
@@ -83,9 +83,10 @@ function renderBroadcast(broadcastElement, data) {
 
 /**
  * A function to determine if a broadcast should render.
- * Does not render a broadcast on the `/new` route, if the current user
- * has opted-out, if the broadcast has already been inserted, or
- * if the key for the broadcast's title exists in localStorage.
+ * Does not render a broadcast on the `/new` route or the mod tools panel.
+ * Does not render a broadcast if the current user has opted-out,
+ * if the broadcast has already been inserted, or if the key for
+ * the broadcast's title exists in localStorage.
  *
  * If the broadcast exists in the DOM but was hidden by the articleForm,
  * the function will re-display it again by adding a class.
@@ -93,7 +94,10 @@ function renderBroadcast(broadcastElement, data) {
  * @function initializeBroadcast
  */
 function initializeBroadcast() {
-  if (window.location.pathname === '/new') {
+  const pathname = window.location.pathname;
+  // The mod actions panel uses an iframe, which attempts to re-render the broadcast.
+  // By checking for `actions_panel` in the path, we can avoid this double render.
+  if (pathname === '/new' || pathname.includes('actions_panel')) {
     return;
   }
 
--- a/app/assets/javascripts/initializers/initializeBroadcast.js
+++ b/app/assets/javascripts/initializers/initializeBroadcast.js
@@ -83,7 +83,7 @@ function renderBroadcast(broadcastElement, data) {
 
 /**
  * A function to determine if a broadcast should render.
- * Does not render a broadcast on the `/new` route or the mod tools panel.
+ * Does not render a broadcast on the `/new` route or in an iframe.
  * Does not render a broadcast if the current user has opted-out,
  * if the broadcast has already been inserted, or if the key for
  * the broadcast's title exists in localStorage.
@@ -94,12 +94,9 @@ function renderBroadcast(broadcastElement, data) {
  * @function initializeBroadcast
  */
 function initializeBroadcast() {
-  const pathname = window.location.pathname;
-  // The mod actions panel uses an iframe, which attempts to re-render the broadcast.
-  // By checking for `actions_panel` in the path, we can avoid this double render.
-  // TODO: [@thepracticaldev/delightful]: Refactor how we decide when to render the broadcast
-  // across the site given that this check will need to happen every time we use iframes.
-  if (pathname === '/new' || pathname.includes('actions_panel')) {
+  // Iframes will attempt to re-render a broadcast, so we want to explicitly
+  // avoid initializing one if we are within `window.frameElement`.
+  if (window.frameElement || window.location.pathname === '/new') {
     return;
   }
 
--- a/app/assets/javascripts/initializers/initializeBroadcast.js
+++ b/app/assets/javascripts/initializers/initializeBroadcast.js
@@ -97,6 +97,8 @@ function initializeBroadcast() {
   const pathname = window.location.pathname;
   // The mod actions panel uses an iframe, which attempts to re-render the broadcast.
   // By checking for `actions_panel` in the path, we can avoid this double render.
+  // TODO: [@thepracticaldev/delightful]: Refactor how we decide when to render the broadcast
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/initializers/initializeBroadcast.js
```diff
@@ -83,17 +83,21 @@ function renderBroadcast(broadcastElement, data) {
...
  * the function will re-display it again by adding a class.
  *
  * @function initializeBroadcast
  */
 function initializeBroadcast() {
-  if (window.location.pathname === '/new') {
+  const pathname = window.location.pathname;
+  // The mod actions panel uses an iframe, which attempts to re-render the broadcast.
+  // By checking for `actions_panel` in the path, we can avoid this double render.
+  if (pathname === '/new' || pathname.includes('actions_panel')) {
```
ludwiczakpawel, line 100: Hmmm... Instead of targeting actions_panel, is there a way to target an element? It seems like this issue happens on all `iframe`s we have and we do have several of them in the _sidecar_ in the Connect as well. Otherwise we will have to end up with 10s of URLs in that if statement. 

Connect issue: https://d.pr/v/boafUi

jacobherrington: Cartman saying, "Oh crap, there's more of them!"

jacobherrington: I'm not totally sure if it is the right idea, but does it make more sense to render the broadcast on fewer pages (for example, the home page)? Rather than running the JS on every page, we could circumvent this issue by only using the broadcast on a small number of specific pages.

vaidehijoshi: @ludwiczakpawel The only way to really "target an element" here would be to check to see where the `<div id="active-broadcast" class="broadcast-wrapper"></div>` element is rendered. However, in both instances where the broadcast is being rendered—in the example of this bug, it is rendered once on the "article" page, and once on the "mod tools" page—the broadcast's `parentElement` is the `<body>` tag. The only thing that differentiates the two versions of the `<body>` tag is the `<div id="page-content"></div>` element, which is a child of the `<body>` tag.

You can see the difference between the two in the JS console. On the "article" page, the `data-current-page` attribute is `"stories-show"`, but on the "mod tools" page, that attribute is `"moderations-actions_panel"`. And that's about the only really "difference" in these two `<body>` tags.

The issue here now is that if we start targeting by an element (in this case, the only thing we could target by is the `data-current-page` on the `<div id="page-content"></div>` element), we run into the exact same issue that you were describing above. Every single place where we have an `iframe`, we now need to look at how that div's attribute changes, because that is the _only_ difference that really exists on the `<body>` tag. In other words, I understand your concern (and I am not a fan of this, either!) but targeting an element (the only element that we _can_ target) doesn't seem to really be much different in the result here.

@jacobherrington The announcement broadcasts are meant to be rendered site-wide (on every single page except for the `/new` route, where users write new articles). The issue with rendering the broadcast on fewer pages is that we'll then run into sort of the opposite problem as what @ludwiczakpawel described: we'll have to start figuring out the routes for where we _do_ want to render the broadcast, and then add them back in. So, it's sort of a question of do we want to be additive or subtractive here. If we take the additive approach, we're listing out all the places where the announcement _should _ render. If we're being subtractive, we're listing out all the places where the announcement _shouldn't_ render. 

IMHO, I think the real answer here is to move away from iframes since they:
1. have their own `window` context,
2. require more computational memory, and
3. can break browser's back/forward buttons + contribute to accessibility issues. 😬

jacobherrington: image

Iframes are definitely hard to work around, but we've sort of adopted them... We will need to have a larger discussion if we choose to remove them long-term, but I agree we should have the conversation.

I personally prefer the additive approach you descrived because it's more explicit -- we are more likely to discover bugs because we know exactly where we are letting this code run. I was thinking of breaking this out of the initializePage file and running that JavaScript more intentionally in specific erb files, and I'm not clever enough to think of a good hack using selectors.

I totally get your point though, and I don't think it's the ideal solution...

If you feel good about it, in the short-term we can merge this, then re-visit a better solution?

vaidehijoshi: @jacobherrington HAHAHA that tweet! 😂  Thanks for the thoughtful reply. I do agree, we should consider having the conversation here as a larger topic. To be honest, part of the reason I thought that this fix was okay was that, while chatting with @Zhao-Andy, he mentioned that the mod tools will likely be re-written in Preact, which would eliminate this issue (but definitely won't solve it for other iframes 😬).

I'm open to the idea of breaking out the initialization logic into separate files and being more intentional with it if that seems easier (we'll just need to be cautious of making sure we cover all the many possible routes in our app where the broadcast would need to render). For the time being, I've added a `TODO` comment in the latest commit for @thepracticaldev/delightful just to make sure we come back to it (but if you end up doing this, feel free to scrap the comment).

TL;DR I appreciate your reply and agree that we want some kind of better solution here 😃

benhalpern:
As mentioned here...

https://github.com/thepracticaldev/dev.to/issues/8983

I think the best way to do this is to add this div into the list of elements hidden in any iframe context. This is the same way that the header and footer magically don't show up when included in an iframe.

jacobherrington:
This now closes #8983 :tada:

vaidehijoshi:
@benhalpern I actually think the solution you described [here](https://github.com/thepracticaldev/dev.to/issues/8983#issuecomment-651243066) that leans on JS is the right approach! I think we'll ultimately prefer that over CSS since we're using the `display` property on the broadcast to toggle its visibility. I suspect if we went the CSS approach, we'd need to use an `!important`, which I'd rather avoid 😛.

It would still be nice to figure out a way to just avoid calling some of these initializers within iframes _altogether_, but that's definitely out of scope for this. On the plus side, after all this discussion, I know know about `window.frameElement`!!
the more you know

Zhao-Andy:
Oops I approved a cached version. Also in favor of Ben's proposed solution, but IMO this still works :)

PR #8987, opened 2020-06-29
Optimization: Dont Use Redis For Flare Tag IDs, Load Them On Boot

mstruve:
- [x] Optimization

## Description
Rather than having to look up Flare Tag IDs every time we load an article(which happens **over 350k times** in 2 hours), lets instead load the IDs when the application boots up and then fetch them from that hash. I think loading the IDs at boot here is fine bc I would not expect the IDs to change under normal circumstances and because if we include a new FLARE_TAG we will have to update this code and redeploy anyways. Let me know what you all think! 

https://github.com/thepracticaldev/dev.to/projects/9#card-39033876

Reboot the app and ensure your flare tags still display as expected!

## Added tests?
- [x] updated

alt_text


app/services/flare_tag.rb
```diff
@@ -0,0 +1,47 @@
...
+                       todayilearned
+                       help
+                       news
+                       healthydebate
+                       showdev
+                       challenge
+                       anonymous
+                       discuss].freeze
+
+  FLARE_TAG_IDS_HASH = Tag.where(name: FLARE_TAG_NAMES).each_with_object({}) do |tag, h|
```
citizen428, line 16: This was merged to fast, but if you ever refactor it: 

```ruby
FLARE_TAG_IDS_HASH = Tag.where(name: FLARE_TAG_NAMES).pluck(:name, :id).to_h
```

app/services/flare_tag.rb
```diff
@@ -0,0 +1,47 @@
...
+  end.freeze
+
+  def initialize(article, except_tag = nil)
+    @article = article.decorate
+    @except_tag = except_tag
+  end
+
+  def tag
+    @tag ||= if tag_id
+               Tag.select(%i[name bg_color_hex text_color_hex]).find_by(id: tag_id)
```
citizen428, line 27: Maybe this array could be moved out to a constant.

app/services/flare_tag.rb
```diff
@@ -0,0 +1,47 @@
...
+             end
+  end
+
+  def tag_hash
+    return unless tag
+
+    { name: tag.name,
```
citizen428, line 34: This can be simplified a bit

```ruby
def tag_hash
  tag&.slice(:name, :bg_color_hex, :text_color_hex)
end
```

PR #9003, opened 2020-06-29
Update styling of user subscription liquid tag

lisasy:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Update visual styling of user subscription liquid tag with CodeLand in mind.

- Refactor SCSS file to use for utility-css classes
- Add overrides to default `crayons-modal` stylings
- Refactor visibility using `CSS` classes instead of js-based styling
- Update styles for `signed-in`, `signed-out`, and `modal` views

#### Signed Out
Screen Shot 2020-06-29 at 3 57 51 PM

#### Signed In
Screen Shot 2020-06-29 at 3 58 07 PM

#### Confirmation Modal
Screen Shot 2020-06-29 at 3 58 16 PM

#### Message Notice
Screen Shot 2020-06-29 at 3 58 21 PM

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed

alt_text


app/liquid_tags/user_subscription_tag.rb
```diff
@@ -75,14 +89,16 @@ class UserSubscriptionTag < LiquidTagBase
...
-      profileImageWrappers.forEach(function(profileImageWrapper) {
-        profileImageWrapper.style.display = 'inline-block';
-      });
+    if (isUserSignedIn()) {
+      profileImageContainer.classList.remove("signed-out");
+      profileImageContainer.classList.add("signed-in");
+    } else {
+      profileImageContainer.classList.remove("signed-in");
+      profileImageContainer.classList.add("signed-out");
+      subscriberImageContainer.classList.add("hidden");
```
atsmith813, line 101: This logic is handled further down with https://github.com/thepracticaldev/dev.to/blob/c4cd91f8102a076004eb0365b0ba41c853b73a68/app/liquid_tags/user_subscription_tag.rb#L219-L231

It looks like you already updated the methods referenced there as well (i.e. `showSignedOut()`) 🎉

app/assets/stylesheets/ltags/UserSubscriptionTag.scss
```diff
@@ -1,85 +1,74 @@
...
-  #subscriber-apple-auth {
-    display: none;
+  .crayons-modal__box__body {
+    padding: var(--su-6);
   }
 
   .crayons-modal__overlay {
     position: fixed;
     z-index: 400;
   }
```
atsmith813, line 73: This "hack" can be removed with these sweet updates 🎉

atsmith813:
Heads up to other reviewers - this is still not the "final" version, especially for the JS. I'll be updating it once more after this is merged to address several things including:
- Conditionally referencing DOM elements (https://github.com/thepracticaldev/dev.to/pull/9005).
- Making the `email` conditional (it's no longer going to be guaranteed included by CodeLand).
- Updating the AppleAuth validation server-side since the client-side validation depended on `email`, which is being made conditional.
- Etc.

In other words, the last PR was passing it to design, this PR is passing it back, the next PR will be wrapping this area up 🎉

app/assets/stylesheets/ltags/UserSubscriptionTag.scss
```diff
@@ -1,85 +1,74 @@
+@import 'config/import';
+
 .ltag__user-subscription-tag {
-  .ltag__user-subscription-tag__container {
+  &__container {
+    border-radius: var(--radius);
+    border: 2px solid var(--body-color);
+    box-shadow: 4px 4px 0 var(--body-color);
```
ludwiczakpawel, line 7: You should be able to use SCSS generator we have for these boxes:

- this is a function: https://github.com/thepracticaldev/dev.to/blob/master/app/assets/stylesheets/config/_import.scss#L206-L231 - it’s documented there as well
- and this is an example of how to use it: https://github.com/thepracticaldev/dev.to/blob/master/app/assets/stylesheets/components/notices.scss#L5-L10

app/assets/stylesheets/ltags/UserSubscriptionTag.scss
```diff
@@ -1,85 +1,74 @@
...
+        margin: 0 auto var(--su-4);
+      }
+    }
 
-  .ltag__user-subscription-tag__signed-in {
-    text-align: center;
-    display: none;
+    img {
+      border-radius: 100%;
+    }
```
ludwiczakpawel, line 44: If you use `.crayons-avatar` in the markup, it will automatically make avatars rounded. So you can drop this line.

lisasy:
> 🔥 🔥 🔥 - this is _awesome!_
> 
> It looks like the preview (write a draft Article and click **Preivew**) needs tweaking and then it's set!
> Screen Shot 2020-06-29 at 8 38 06 PM
> 
> On **Preview**, the JavaScript isn't executed so we need to render the markup in the "default state" which should be "signed-out". Quickly looking at the changes here, I think we need to do something along the lines of removing the `hidden` class from [here](https://github.com/thepracticaldev/dev.to/pull/9003/files#diff-e42e13d644d4faa2ea4ccb207b8eee9aR23)
> 
> I'm not entirely sure why the images are showing up this way 🤔 . Are we adding a class somewhere in JS to get them centered up? If so, we'll want to update the markup to include the needed classes to center the images and hide the subscriber image by default.

I'm adding a `signed-in` or `signed-out` class to the parent container to set the positioning of the profile image elements. I'll address this in the upcoming commits...

app/assets/stylesheets/ltags/UserSubscriptionTag.scss
lisasy: I added this because `.crayons-article__body img {border-radius: var(--radius)}` is over-writing the style of the rounded radius of `.crayons-avatar`. Have any thoughts on this hierarchy here?

app/assets/stylesheets/ltags/UserSubscriptionTag.scss
lisasy: Thank you!

app/assets/stylesheets/ltags/UserSubscriptionTag.scss
lisasy: Removed :)

app/liquid_tags/user_subscription_tag.rb
lisasy: I refactored this in the latest commit: https://github.com/thepracticaldev/dev.to/pull/9003/commits/44d85d2a773e2064cdf9cd2ab8cd1d7634c21e90#diff-dbaa157d5eab1612b5396d10b80fe577

lisasy:
@atsmith813 Here is the screenshot of what this looks like in `Preview` mode. This PR is ready to be re-reviewed.

Screen Shot 2020-06-30 at 9 23 58 AM

atsmith813:
@lisasy AWESOME!

I think we to remove the hidden class from [here](https://github.com/thepracticaldev/dev.to/pull/9003/files#diff-e42e13d644d4faa2ea4ccb207b8eee9aR23) so that the **Sign in** button shows up in the preview. I'm thinking an author might be confused as to how that tag is going to work without seeing a button in the **preview**. And we're defaulting to the "Signed out UX" which would be with that button.

What do you think?

lisasy:
@atsmith813 

Per your earlier comment about the `Preview` state of the tag lacking any indication of how the end-user would see it, I've updated it so that it shows the `signed-out` state. If this looks good, I'm ready to merge this and let you handle additional work here.

Screen Shot 2020-06-30 at 5 48 21 PM

app/assets/stylesheets/ltags/UserSubscriptionTag.scss
ludwiczakpawel: hmmm that's weird.. keep it as it is, i'll try to investigate that later 👍

PR #9006, opened 2020-06-30
Stop allowing accented characters in tags

### LINKED ISSUE #8884: Tag limitation
**Describe the bug**

I was able to create a tag with only one special character `â` when adding a new post. But the generated page with tag did not work.

**To Reproduce**

1. Write a new post
2. Add a tag with some special character like `â` or `ą` or whatever else.. Technically you can't enter that character by hand (usually with use of `opt` key on keyboard) but you can easily paste it or long press `a` key to see diacritical marks and then you can enter any you one.
3. Post is created. Click on that tag in the post view.
4. Error.

**Video**
[Video](https://d.pr/v/1xSmcP)

**Expected behavior**

Not sure... 🤷 

- OS: macOS X
- Browser: Chrome
- Version: 83.x

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This commit will add strict frontend and backend validation for tags
disallowing accented characters.

Closes #8884

Before:
https://d.pr/v/1xSmcP

After:
Peek 2020-06-29 20-54

*Note: HTML5 validation won't let you type accented characters, but you can paste them*

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -82,7 +82,10 @@ class Tag < ActsAsTaggableOn::Tag
 
   def validate_name
     errors.add(:name, "is too long (maximum is 30 characters)") if name.length > 30
-    errors.add(:name, "contains non-alphanumeric characters") unless name.match?(/\A[[:alnum:]]+\z/)
+    # [:alnum] is not used here because it supports diacritical characters.
+    # If we decide to allow diacritics in the future, we should replace the
+    # following regex with [:alnum].
+    errors.add(:name, "contains non-alphanumeric characters") unless name.match?(/\A[[a-z0-9]]+\z/i)
   end
 
   def mod_chat_channel
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -82,10 +82,10 @@ class Tag < ActsAsTaggableOn::Tag
 
   def validate_name
     errors.add(:name, "is too long (maximum is 30 characters)") if name.length > 30
-    # [:alnum] is not used here because it supports diacritical characters.
+    # [:alnum:] is not used here because it supports diacritical characters.
     # If we decide to allow diacritics in the future, we should replace the
-    # following regex with [:alnum].
-    errors.add(:name, "contains non-alphanumeric characters") unless name.match?(/\A[[a-z0-9]]+\z/i)
+    # following regex with [:alnum:].
+    errors.add(:name, "contains non-ASCII characters") unless name.match?(/\A[[a-z0-9]]+\z/i)
   end
 
   def mod_chat_channel
--- a/spec/models/tag_spec.rb
+++ b/spec/models/tag_spec.rb
@@ -46,6 +46,16 @@ RSpec.describe Tag, type: :model do
         tag.name = nil
         expect(tag).not_to be_valid
       end
+
+      it "fails validations if name uses diacritics" do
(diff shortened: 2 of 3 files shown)


spec/models/tag_spec.rb
```diff
@@ -46,6 +46,16 @@
...
         expect(tag).not_to be_valid
       end
+
+      it "fails validations if name uses diacritics" do
+        tag.name = "łookatmé"
+        expect(tag).not_to be_valid
+      end
+
+      it "fails validations if name uses non-latin characters" do
+        tag.name = "火"
```
jacobherrington, line 56: We could certainly expand this to include more character sets, but I think that might be overdoing it. :stuck_out_tongue:

citizen428: I don't think so tbh. I left a comment on the issue, but given our coming I18n efforts it seems restricting tags to ASCII is a bit counterproductive.

app/models/tag.rb
```diff
@@ -82,7 +82,10 @@ def self.find_preferred_alias_for(word)
 
   def validate_name
     errors.add(:name, "is too long (maximum is 30 characters)") if name.length > 30
-    errors.add(:name, "contains non-alphanumeric characters") unless name.match?(/\A[[:alnum:]]+\z/)
+    # [:alnum] is not used here because it supports diacritical characters.
+    # If we decide to allow diacritics in the future, we should replace the
+    # following regex with [:alnum].
+    errors.add(:name, "contains non-alphanumeric characters") unless name.match?(/\A[[a-z0-9]]+\z/i)
```
citizen428, line 88: We should consider changing the wording to "contains non-ASCII characters". Accented characters are alphanumeric too, which is why the `:alnum:` class matches them.

atsmith813:
WRT [`acts-as-taggable-on` gem](https://github.com/mbleigh/acts-as-taggable-on), this looks solid, thanks for jumping on this 🎉 . Since tags have been cleaned up, this will be much easier to manage now and in the future 💯 .

The gem offers [minor support for special characters for MySql users](https://github.com/mbleigh/acts-as-taggable-on#for-mysql-users) (which doesn't apply since we use PostgreSQL).

spec/models/tag_spec.rb
jacobherrington: @citizen428 I agree in the long run.

jacobherrington: Lol re-reading this comment, I meant adding more tests was overdoing it... :sweat_smile: 

I think we should work toward removing this limitation on tags, especially considering that in the future there might be non-English communities on Forem.

citizen428:
> The gem offers minor support for special characters for MySql users (which doesn't apply since we use PostgreSQL).

I've maintained for a long time that it's almost always better to implement tagging from scratch. It's not particularly complicated and over the years I found that existing solutions almost always fall short and get replaced anyway, same as admin scaffolding tools (see recent discussion about removing Administrate).

djuber:
https://github.com/forem/forem/issues/14422 adding this issue link for visibility into the context (and #13720 as well)

PR #9007, opened 2020-06-30
Initial Proof of Concept: Code block "lightbox"

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Sometimes code is hard to read when it is constrained to a narrow box within a post. This PR gives an expansion button. We can re-use most of this functionality for images and other "expandable" content. But since Forem's main app is DEV, I think focusing on the code reading functionality makes sense.

This is not yet fully functional. Getting it working is easy, but I definitely want to get the crayons part right want to wait on input there. For now I just tossed in some dummy classes.

Here it is in action...

### DIFF
--- a/app/assets/stylesheets/syntax.scss
+++ b/app/assets/stylesheets/syntax.scss
@@ -13,6 +13,7 @@
 
 pre:not(.highlight),
 div.highlight {
+  position: relative;
   background: var(--syntax-background-color);
   color: var(--syntax-text-color);
   font-size: 80%;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/syntax.scss
```diff
@@ -31,6 +40,30 @@ div.highlight {
...
+  }
+  &:hover {
+    button {
+      display: block;
+    }
+  }
+}
+
+// *Definitely* doesn't belong here, just tossed it here and probably should be the fully-reusable modal bg.
+.crayons-fullscreenbg {
```
Link2Twenty, line 59: I know this is closed but I thought I'd share a little trick.
Rather than spawning in a whole new div to the dom to could do something like

```scss
&.crayons-fullscreen {
  position: fixed;
  top: calc(50px + 3vh);
  left: calc(5px + 3vw);
  bottom: 3vh;
  right: calc(5px + 3vh);
  z-index: 150;

  &--backdrop {
    box-shadow: 0 0 0 200vmax rgba(0,0,0,0.8);
  }
}
```

Now you can have  a div like this.

```html
<div class="crayons-fullscreen crayons-fullscreen--backdrop" >CONTENT</div>
```

result

PR #9027, opened 2020-06-30
Add review item model and mark as reviewed for mod center

Zhao-Andy:
- [x] Feature

## Description
This is a proof of concept of the "mark as reviewed" functionality for mods to review articles in a more inbox-style way through the Mod Center. The general idea is this:

1. an article is created, and is put into a review queue, via a `ReviewItem` creation
2. the Mod Center shows articles that need to be reviewed
3. a moderator visits the mod center, and takes an action, which marks an article as reviewed

We're still finalizing the details of what exactly a reviewed article means. This current implementation is based on these ideas:
- the Mod Center is showing **only articles you have not reviewed**
- an article that has been reviewed by someone will still be unreviewed to other moderators

I was having trouble figuring out the best way to implement this. Currently, each article will create a `ReviewItem`. This allows us to easily query for which articles to display:
```ruby
article_ids = ReviewItem.where(reviewed: false, reviewable_type: "Article").pluck(:reviewable_id)

articles = Article.where(id: article_ids)
```
However, this means that we'll have **at least** one new row per article. Additional rows per article would be linked to a `reviewer`. Like so:
```ruby
ReviewItem.find_by(reviewable: some_article, reviewer_id: nil, reviewed: false)
ReviewItem.find_by(reviewable: some_article, reviewer_id: 11, reviewed: true)
ReviewItem.find_by(reviewable: some_article, reviewer_id: 12, reviewed: true)
# ... and so on
```
I feel like this could be problematic, such as always having a `ReviewItem` with `reviewer_id: nil` and will never become `reviewed: true`.

Another implementation I thought of was making the `reviewer_id` column an array column:
```ruby
# on article creation
ReviewItem.find_by(reviewable: some_article, reviewer_ids: [], reviewed: false)
# on article being reviewed by someone
ReviewItem.update(reviewable: some_article, reviewer_ids: [1], reviewed: true)
# querying for whether current_user has reviewed it
current_user_id = 1
ReviewItem.find_by(reviewable: some_article, reviewer_ids: [1])
# actually not sure how ActiveRecord works with Postgres array column yet, but it can be solved!
```

**_Anyway, additional thoughts will definitely be appreciated!!!_**

Screen Shot 2020-06-30 at 11 09 00 AM

Steps for QA:
1. Create review items in console:
```
articles = Article.last(3)
articles.each do |article|
  ReviewItem.create!(reviewable: article, reviewer_id: nil)
end
2. Go to the Mod Center `/mod`
3. Open an article, and take any action on it via the mod panel
4. Refresh the page to see that it's marked as reviewed (yes, this is kind of janky right now but it's just a proof of concept)

## Added tests?
- [ ] no, not yet

## Added to documentation?
- [x] no documentation needed

![alt_text](gif_link)

### CODE REVIEW

**Comment by msarit** [comment]
> Another implementation I thought of was making the reviewer_id column an array column

This implementation seems more elegant to me 😄 but I am very curious to see what solutions others come up with.
Great job so far @Zhao-Andy! 👍

File: app/models/review_item.rb [hunk 1]
```diff
@@ -0,0 +1,21 @@
...
+  scope :reviewable_articles, ->(reviewer_id) { where("reviewed = false AND reviewable_type = 'Article' AND (reviewer_id IS NULL OR reviewer_id = ?)", reviewer_id) }
+
+  class << self
+    def mark_as_reviewed(reviewable, reviewer)
+      review_item = find_or_initialize_by(reviewable: reviewable, reviewer: reviewer)
+      return if review_item.reviewed
+
+      review_item.reviewed = true
+      review_item.read = true
+      review_item.save!
```
vaidehijoshi, line 18: `review_item.update!(reviewed: true, read: true)`

db/migrate/20200624185243_create_review_items.rb
```diff
@@ -0,0 +1,20 @@
+class CreateReviewItems < ActiveRecord::Migration[6.0]
+  def change
+    create_table :review_items do |t|
+      t.references :reviewable, polymorphic: true, null: false, index: { name: :index_on_reviewable_type_and_id }
+      t.references :reviewer, references: :users, foreign_key: { to_table: :users }
+      t.boolean :reviewed, null: false, default: false
+      t.boolean :read, null: false, default: false
+      t.string :action_taken
```
vaidehijoshi, line 8: Doesn't appear to be being used yet, but we probably want to validate inclusion of whatever `action_taken` is :)

rhymes:
@vaidehijoshi 

>  (in my experience, the Postgres json column is hard to work with, but can be super useful as long as you don't want to query into it. Not sure if the array column is similar in how you interact with it).

They've recently added indexes to paths in JSON columns so you can even index a single field inside a JSON blob and query it.

rhymes:
If I understood correctly we essentially need to tie a list of articles with a list of moderators and put it into a list of queues, one per each moderator.

When an item is reviewer a boolean gets flipped and that's it.

I have a few questions:

Do we need to keep the list of reviewed articles? 
Is undo a thing?
What if instead of having the concept of "review item" we create a table called something like `article_moderation_queue` which contains a single row with three columns? `moderator_id`, `array of reviewed article ids`, `array of articles to review`
This way it's just one row per moderator, when the item goes from unmoderated to moderated we just move the id from one column to the other and if undo is a thing, we do the reverse.
If then we want to add the concept of "comment moderation queue" we do the the same with a new table.

Would that make sense?

PostgreSQL has native array types that can be indexed - https://www.postgresql.org/docs/11/arrays.html - and operations you can perform in arrays - https://www.postgresql.org/docs/11/functions-array.html

This way each row of this new table is the moderator queue and when a moderator gets demoted, we just remove that row or even better, move the IDs of stuff to moderate to someone else :)

citizen428:
Not knowing the exact requirements/specification, the idea of having a table that has a direct correspondence to articles bothers me a little bit, especially since it seems we wouldn't need most of this information after an article is reviewed.

So I'm wondering: what's the minimum of amount of information we need to store to make this work? A boolean column (`reviewed`) and reviewer info (potentially an array or JSON column)? In that case, we might be able to keep the queue in more ephemeral storage and rebuild it on-demand/from a task instead of persisting/purging. The `articles` table already has quite a few columns, yet I think I'd feel better about adding 2 more, than an entirely new table while we're still not completely sure how this feature will work out in practice.

mstruve:
This feels like an architecture hiring question to me! lol 

I think we should stay away from arrays or special columns in our database. Those to me always seem to indicate there is a better, more basic solution. Here is what I was thinking.

Based on the PR description(I may totally be missing something!) It sounds like a `ReviewItem`(maybe rename just `Review`) should be created when an article is reviewed and it contains all the information about the review. The missing piece which it feels like you are trying to shove into that model with a blank reviewer_id is how do you account for articles that we want to be reviewed and still need to be reviewed. Here I think you want a second model such as `ReviewStatus`. This model would be connected solely to the article and would hold the status of the article, has it been reviewed and approved? Has it been reviewed and rejected? Does it need to be reviewed? I think keeping track of that would work well using an enum. 

In model terms....
```ruby
class Article
  has_many :reviews
  belongs_to :review_status
end

class Review
  belongs_to :reviewable # could expand this to comments or whatever
  belongs_to :reviewer
  belongs_to :review_status # I think this is what we want here
end

class ReviewStatus
  # connects the article to the reviews and holds all the logic to know what actions need to be taken. 
  belongs_to :reviewable # Article
  has_many :reviews
  enum status: [:approved, :rejected, :unreviewed]
end
``` 

I also share @citizen428 concerns about trying to keep all things synced but at least with the above approach we keep that "sync" logic contained in it's own model.

mstruve:
Hey @Zhao-Andy, what's the status on this? There are a lot of conflicts and it's been inactive for a while so I was thinking we might want to close it.

Zhao-Andy:
Hey @mstruve, good point. I think it needs a revisit at some point, but definitely deprioritized now. We're not doing anything with the Mod Center any time soon as far as I know, so I'll close this.

Appreciate everyone's comments on this! Will definitely take everything into consideration when we do work on the Mod Center again.

PR #9031, opened 2020-06-30
Refactor: Remove Unused Article Suggester Services and Specs

mstruve:
- [x] Refactor

## Description
These two article suggesters are no longer used. It looks like they might have been replaced with the class `ArticleSuggester` so I am removing them and their related specs 

alt_text


rhymes:
YES!

PR #9032, opened 2020-06-30
Hide secret code from input history

benlesh:
Currently, the secret code will show up in the input history. That, I think, means that anyone that can open that page in a shared browser instance could join the org.

NOTE: I'm not a Ruby developer, so I'm only guessing that this change will resolve the issue. The intent is to add `autocomplete="off"` to the text input. [More information about the autocomplete attribute can be found here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)


nickytonline:
Thanks for the PR @benlesh!

PR #9043, opened 2020-06-30
Add slash back for root link

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When we removed some end-of-url slashes we also removed them from root here, resulting in a "" href. This is a quick fix.

### DIFF
--- a/app/views/articles/index.html.erb
+++ b/app/views/articles/index.html.erb
@@ -47,7 +47,7 @@
         <div class="flex-1 flex items-center justify-between">
           <h2 class="fs-l fw-heavy">Posts</h2>
           <nav class="crayons-tabs hidden s:flex">
-            <a href="<%= list_path %>" class="crayons-tabs__item <%= "crayons-tabs__item--current" if %w[week month year infinity latest].exclude?(params[:timeframe]) %>">Feed</a>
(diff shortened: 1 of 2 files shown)


app/views/articles/index.html.erb
```diff
@@ -47,7 +47,7 @@
         <div class="flex-1 flex items-center justify-between">
           <h2 class="fs-l fw-heavy">Posts</h2>
           <nav class="crayons-tabs hidden s:flex">
-            <a href="<%= list_path %>" class="crayons-tabs__item <%= "crayons-tabs__item--current" if %w[week month year infinity latest].exclude?(params[:timeframe]) %>">Feed</a>
+            <a href="<%= list_path %>/" class="crayons-tabs__item <%= "crayons-tabs__item--current" if %w[week month year infinity latest].exclude?(params[:timeframe]) %>">Feed</a>
```
vaidehijoshi, line 50: @benhalpern Thinking about this more...maybe `list_path` can just return `"/"` if there's no tag and then we don't need to worry about an empty path since the helper can handle it internally...

PR #9044, opened 2020-07-01
Fix archived readinglist reactions affecting count

### LINKED ISSUE #8164: Reading list count in the homepage should reflect only non archived items
**Describe the bug**

Reading list count is not updating!

The reading list count is showing 7.

image

However I have only 2 items in the list!

image

Not sure if it a cache issue though! Yet have tried empty cache and hard reload, still the issue persists. 

**To Reproduce**

1. You can add n number of posts to the reading lists.
2. Delete some of them.
3. Check if the count matches on homepage (/) and reading list route (/readinglist)

**Expected behavior**

Count should matche on homepage (/) and reading list route (/readinglist)

- OS: Ubuntu 20.04
- Browser: Chrome Version 81.0.4044.138 (Official Build) (64-bit)

**Additional context**

The posts were there in my reading list for several days (approx 10-15). Also posts weren't removed at once, last update in reading list was couple hours ago.

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I've seen at least two issues pertaining to this bug, so I went ahead and fixed it. 

~However, I found another bug while doing so... Admin reactions have probably been created with a status of `confirmed` in production instead of the expected `valid`. Which means that this proposed change will cause admin users who have saved articles before this point to have an incorrect readinglist count badge going forward, _but_ it will not affect admins from this point forward~

~If that's considered unacceptable, we can come up with a workaround or change those reaction records in production :sweat:~

~The reading list count badge on the home page currently takes all
readinglist reactions into account, but we shouldn't be including those
that are marked "archived."~

~Secondarily this fixes a bug that affects only admin users: all
reactions created by admins were given the status "confirmed."
Essentially this means that all admin users who saved articles to their
reading list might see an odd behavior here...~

The above concern is no longer an issue thanks to @mstruve's suggestion. I will modify the commit message on merge.

Closes #8164

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

If we are concerned about a slight cosmetic bug (incorrect reading list badge count) that affects only admins, we will need to go update all of the `readinglist` reactions that admins have made while this bug existed. Their status is most likely currently `confirmed` rather than the default `valid`.

### DIFF
--- a/app/controllers/reactions_controller.rb
+++ b/app/controllers/reactions_controller.rb
@@ -124,7 +124,7 @@ class ReactionsController < ApplicationController
       reactable_type: params[:reactable_type],
       category: category
     }
-    if NEGATIVE_CATEGORIES.include?(category) && current_user&.any_admin?
+    if current_user&.any_admin? && NEGATIVE_CATEGORIES.include?(category)
       create_params[:status] = "confirmed"
     end
     Reaction.new(create_params)
--- a/app/controllers/reactions_controller.rb
+++ b/app/controllers/reactions_controller.rb
@@ -124,7 +124,9 @@ class ReactionsController < ApplicationController
       reactable_type: params[:reactable_type],
       category: category
     }
-    create_params[:status] = "confirmed" if current_user&.any_admin?
+    if NEGATIVE_CATEGORIES.include?(category) && current_user&.any_admin?
+      create_params[:status] = "confirmed"
+    end
     Reaction.new(create_params)
   end
 
--- a/app/labor/reading_list.rb
+++ b/app/labor/reading_list.rb
@@ -20,7 +20,7 @@ class ReadingList
   end
 
   def ids_of_articles
-    Reaction.where(reaction_criteria).order(created_at: :desc).pluck(:reactable_id)
+    Reaction.where(reaction_criteria).where.not(status: "archived").order(created_at: :desc).pluck(:reactable_id)
   end
 
   def count
@@ -28,6 +28,6 @@ class ReadingList
   end
 
   def reaction_criteria
-    { user_id: user.id, reactable_type: "Article", category: "readinglist", status: "valid" }
+    { user_id: user.id, reactable_type: "Article", category: "readinglist" }
   end
 end
--- a/app/labor/reading_list.rb
+++ b/app/labor/reading_list.rb
@@ -28,6 +28,6 @@ class ReadingList
   end
 
   def reaction_criteria
-    { user_id: user.id, reactable_type: "Article", category: "readinglist" }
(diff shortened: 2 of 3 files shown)


app/controllers/reactions_controller.rb
```diff
@@ -120,7 +120,9 @@ def build_reaction(category)
       reactable_type: params[:reactable_type],
       category: category
     }
-    create_params[:status] = "confirmed" if current_user&.any_admin?
+    if NEGATIVE_CATEGORIES.include?(category) && current_user&.any_admin?
```
citizen428, line 123: Should we maybe switch the two conditions around so we exit as early as possible (no current user or they are not an admin) before doing the array inclusion check? It's not particularly important in this case as the latter check is cheap, but maybe a good habit to get into.

rhymes:
@jacobherrington could you come up with a script for the post deployment? This way if the SRE team wants to run it they can easily do it. I think a SQL query is also fine.

awasum:
> /rebase

This should fail as am not part of the org containing this repo...  Can any of you try?

rhymes:
I tried on one of my PRs but with no result https://github.com/thepracticaldev/dev.to/pull/8215#issuecomment-653028504 - I wonder if it needs a new one @awasum

awasum:
I think we should update the GitHub action to allow collaborators to rebase PRs automatically. The members condition we have in the GitHub action seems to ensure only public members of the Dev.to GitHub org can rebase PRs as it worked for Nick

app/labor/reading_list.rb
```diff
@@ -28,6 +28,6 @@ def count
   end
 
   def reaction_criteria
-    { user_id: user.id, reactable_type: "Article", category: "readinglist" }
+    { user_id: user.id, reactable_type: "Article", category: "readinglist", status: "valid" }
```
mstruve, line 31: Rather than showing the incorrect counts for admins and doing only valid here why don't we just exclude the `archived` ones? So above 
```ruby
def ids_of_articles
  Reaction.where(reaction_criteria).where.not(status: "archived")...
end
```

Another option is rather than grabbing this count from the async info hash, get it from the Elasticsearch query.

app/controllers/reactions_controller.rb
jacobherrington: Good call, done!

app/labor/reading_list.rb
jacobherrington: Ah, yeah, that's a much better solution. I'd rather do that than run a script on prod data. :sweat:

PR #9047, opened 2020-07-01
Fixing scrollbar for small screens in the profile page

OmarElKhatibCS:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Go to profile -> resize the window -> you should see annoying scroll bar in the bottom ,
also same for mobile app.

Screenshot after fixing

# before
fix_respo_before

Screenshot_20200630-184951

this black shadow keep appearing because of this scrollbar

# after

fix_respo

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/assets/stylesheets/articles.scss
+++ b/app/assets/stylesheets/articles.scss
@@ -9,6 +9,7 @@
   position: relative;
   margin: 0 auto;
   max-width: $max-width;
+  width: 99%;
   text-align: left;
 
   @media screen and (max-width: 949px) {
--- a/app/assets/stylesheets/articles.scss
+++ b/app/assets/stylesheets/articles.scss
@@ -9,7 +9,6 @@
   position: relative;
   margin: 0 auto;
   max-width: $max-width;
-  width: 99%;
   text-align: left;
 
   @media screen and (max-width: 949px) {
--- a/app/assets/stylesheets/index-comments.scss
+++ b/app/assets/stylesheets/index-comments.scss
@@ -3,10 +3,6 @@
 .index-comments {
   text-align: left;
   padding-bottom: calc(0.6vw + 5px);
-  background: var(--card-bg);
-  border-radius: 3px;
-  @include themeable(border, theme-container-border, none);
-  @include themeable(box-shadow, theme-container-box-shadow, $bold-shadow);
   margin: auto;
   width: 100%;
   @media screen and (min-width: 950px) {
--- a/app/views/users/_comments_section.html.erb
+++ b/app/views/users/_comments_section.html.erb
@@ -7,7 +7,7 @@
       color: #cedae2;
     }
   </style>
-  <div class="index-comments">
+  <div class="index-comments crayons-card">
     <% if @comments.present? %>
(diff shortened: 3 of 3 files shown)


OmarElKhatibCS:
> `.home` is being used on maaaaaany views in DEV (even though most of them are not _home_). That's why I think this change would affect too many places.
> 
> Instead we should find the actual cause of seeing the scrollbar.
> 
> Also, I can't repro this bug, can you help me out with this?

Hello after more investigation I see that the border in class index-comments cause this problem.
"repro this bug" this term new to me can you please explain it?

ludwiczakpawel:
> "repro this bug" this term new to me can you please explain it?

Whoops, I apologize, should have more understandable term... So I'm trying to spot / reproduce the same issue locally, on my computer and phone and I'm having troubles. Meaning - i don't see this bug...

OmarElKhatibCS:
> > "repro this bug" this term new to me can you please explain it?
> 
> Whoops, I apologize, should have more understandable term... So I'm trying to spot / reproduce the same issue locally, on my computer and phone and I'm having troubles. Meaning - i don't see this bug...

 I try it on firefox both linux/windows , Also on APK mobile app .
But in chrome it's work fine.
it can be fixed by removing the border or width:99% not sure what is better.

ludwiczakpawel:
> Hello after more investigation I see that the border in class index-comments cause this problem.

If this really is an issue, it's because `index-comments` element has legacy (old) design. It should not have the border applied the way it is. Here's my suggestion:

1. Apply `crayons-card` class to [`index-comments` element](https://github.com/thepracticaldev/dev.to/blob/a95877e8ff655371cfd06383bd9dabc26e7f6653/app/views/users/_comments_section.html.erb#L10).
2. Remove legacy styling from [`index-comments` stylesheet](https://github.com/thepracticaldev/dev.to/blob/3a17fd2aa9312ff04ed000c115ceebfa44ce0fe7/app/assets/stylesheets/index-comments.scss#L3-L11). Remove following lines

```
background: var(--card-bg);
border-radius: 3px;
@include themeable(border, theme-container-border, none);
@include themeable(box-shadow, theme-container-box-shadow, $bold-shadow);
```

^ all of that style should be handled by `crayons-card` class and you should no longer see the extra border potentially causing  the scrollbar :)

OmarElKhatibCS:
> Thanks for the updates, let me know if it fixed your issue. LGTM.

I am building the containers right now , When it's done , I will try it locally if it fixes the problem.
Connection is slow so it gonna take some time.

OmarElKhatibCS:
I tried to run it locally , I got lot of errors , web:3000 connection refused for dev_seed and dev_sidekiq .
Also when I create a post it's not published (nothing happen when I press publish) so I can not reply to check if comments section is working.
I will try to fix it later , I think I can create user and posts from ruby (I am not a ruby developer) .
I am busy rn with finals are soon so it's need sometime to do it.
I am building using docker.

Issue #8169, opened 2020-05-29
Add subscriptions to the dashboard


atsmith813:
_We are building a liquid tag where permitted authors can embed a button that users/readers can click to share the email address associated with their DEV account with the author - almost like a subscribe button._

### Expose submitter emails to article authors
In the user's dashboard add a new tab for email submissions that shows the articles where the author included the button and then shows all the email addresses for submissions.

Could possibly provide a way to export/download the data as well - CSV, JSON, etc.

Maybe a "see all" view that isn't filtered by the article?

*Don't forget to update the service worker to exclude any new routes*

atsmith813:
To our awesome contributors, please note that this issue will be worked on by a DEV team member and is not up for grabs.

PR #9052, opened 2020-07-01
Dashboard subscriptions

ludwiczakpawel:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR introduces a subpage for articles having a liquid tag for collecting email with subscription liquid tag.
It also adds several trigger to get to this page. These should obviously be contained in appropriate IF statement.

🚨 In order to create this page I've done some things in articles_controller and other places that I should probably never touch :D Just pointing it out so someone with better backend knowledge can actually do it properly.

- #8169 
- #8779 

image

## Added tests?

- [ ] yes
- [x] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


lisasy:
No CSS file?! Amazing 🤯

app/views/articles/subscriptions.html.erb
```diff
@@ -0,0 +1,65 @@
+<% title "Subscriptions #{@article.title}" %>
+
+<% data = [
```
joshpuetz, line 3: Should this PR be a draft containing sample data like this?

app/controllers/articles_controller.rb
```diff
@@ -81,6 +81,13 @@ def manage
     @org_members = @organization.users.pluck(:name, :id) if @organization
   end
 
+  def subscriptions
```
joshpuetz, line 84: These subscriptions are `user_subscriptions`, right? Is this controller the right place for them?

ludwiczakpawel:
> No CSS file?! Amazing 🤯

hah! that's a benefit of Crayons!

app/controllers/articles_controller.rb
rhymes: I think this should be in the `DashboardsController`

cc @atsmith813

app/policies/application_policy.rb
```diff
@@ -36,6 +36,10 @@ def manage?
     update? && record.published
   end
 
+  def subscriptions?
+    true
```
rhymes, line 40: this should probably use the same logic as the auth logic that enables subscriptions, not sure how it's done right now

app/views/articles/subscriptions.html.erb
ludwiczakpawel: Very good point! Thanks :)

ludwiczakpawel:
> Hi @ludwiczakpawel, looks cool but I think this should be part of the dashboard...

@rhymes yeap, definitely. as I mentioned in PR description - i'm not the best person to figure it out, i just wanted to put something together to have this table view :P I bet someone from @thepracticaldev/delightful will have better idea of how it should be done correctly.

app/views/articles/subscriptions.html.erb
```diff
@@ -0,0 +1,65 @@
...
+     },
+     {
+       name: "Lisa Sy",
+       email: "lisa@sy.com",
+       timestamp: "Last month"
+     },
+     {
+       name: "Bill Clinton",
+       email: "billy.c@gov.com",
+       timestamp: "3 years ago"
```
vaidehijoshi, line 32: We probably should use actual timestamps here; we'll need to include additional logic to convert them and format into a string (but I think @thepracticaldev/delightful can implement this change if necessary!).

app/controllers/articles_controller.rb
vaidehijoshi: ➕ to moving this into `DashboardsController`...maybe a `subscribed_users` method?

app/views/dashboards/_dashboard_article.html.erb
```diff
@@ -64,6 +64,7 @@
     <% if @current_user_pro %>
       <a href="<%= article.path %>/stats" data-no-instant class="cta pill blue">STATS</a>
     <% end %>
+    <a href="<%= article.path %>/subscriptions" data-no-instant class="cta pill">SUBSCRIPTIONS</a>
```
vaidehijoshi, line 67: Out of scope of this PR probably, but I think we should consider having a css class that is uppercased (if we don't already) and applying it here. We have lots of variations of different text formatting (lowercase, capitalized, all uppercase), and using classes would make this a bit easier/more robust.

ludwiczakpawel: We do have one: `text-uppercase`. I think you can easily ignore this one - this view is still being considered as "legacy, pre-crayons era" type of view. Once I update its design, we will no longer have this type of buttons with uppercase labels. 

In this case I just wanted to keep it consistent with what's already there :)

app/views/articles/subscriptions.html.erb
ludwiczakpawel: Cool. Yea, i think having relative dates with a tooltip showing full timestamp is the ideal state. But if we can't have that, showing full timestamp will be enough.

atsmith813: @vaidehijoshi @ludwiczakpawel I'm slightly biased towards relative time because our timestamps aren't quite where we want them to be WRT timezones 😬 . We've got a couple of open issues regarding that. For this "v1", I'm thinking maybe what's most important is an author has an easy way to identify "recent" subscriptions, either way.

vaidehijoshi: @atsmith813 I should clarify: I do agree that we should format the timestamps to be relative to the user's timezone; I was more just trying to point out here that the sample data here has a string for the time, when more likely, from the backend, it will be a `created_at` timestamp...right?

atsmith813: @vaidehijoshi ohhhhhh yes, I see now. Correct!

atsmith813:
@ludwiczakpawel what's the easiest way to move forward on this from your perspective?

For me, this PR has what I need to finish it off - the markup. I can take on the routes, controller actions, etc. A few options - I can push to this PR, we can close this and I'll pluck the markup as I need it in another PR, we can remove everything except the markup from this PR and merge it (with no route it will be fine), or maybe something else?

I think we can remove the `remove` button column. I don't think we'll need to give an author a way to remove someone subscribed to them for "v1". Worst case, we can delete them manually for authors and learn more about that use case if/when that happens. These subscriptions are tied to users and articles so deleting them and handling the deletion of associations is something I want to think through more first.

Something else to get on the radar, not necessarily for this PR, is a button for "Export all subscriptions" on the main dashboard page so an author doesn't have to export it for each article. It's unlikely I'll get to this soon, just something for us to think about. I imagine that'll be one of the first pieces of feedback we'll get.

ludwiczakpawel:
@atsmith813 I think that’s the right way to proceed - I was indeed hoping you will take it from there and handle all backend related stuff. As I mentioned earlier: my intention was to help with frontend and design part so if you think this is in good spot then feel free to continue with other changes.

Also please do let me know if there’s anything also you’d like my input on.

atsmith813:
Closing this out in favor of #9161. Thanks again, @ludwiczakpawel 🎉

PR #9059, opened 2020-07-01
Remove report-abuse link from base HTML document into JavaScript

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Every comment has a "report abuse" link, which Google sees as useless and a waste of crawler resources since this isn't a link to new "content".

Screen Shot 2020-07-01 at 2 52 25 PM

So this PR moves that functionality such that the link does not exist until the user clicks on the dropdown menu.

In this PR, I also fixed some styling for the "hide" button, which is currently not correct.

Screen Shot 2020-07-01 at 2 50 29 PM

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -161,7 +161,7 @@
         <% if user_signed_in? %>
           <%= javascript_packs_with_chunks_tag "profileDropdown", defer: true %>
         <% end %>
-        <a href="/report-abuse?url=<%= user_url(@user) %>">Report Abuse</a>
+        <span class="report-abuse-link-wrapper" data-path="/report-abuse?url=<%= user_url(@user) %>"></span>
       </div>
     </div>
   </div>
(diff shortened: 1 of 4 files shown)


app/views/users/show.html.erb
```diff
@@ -161,7 +161,7 @@
         <% if user_signed_in? %>
           <%= javascript_packs_with_chunks_tag "profileDropdown", defer: true %>
         <% end %>
-        <a href="/report-abuse?url=<%= user_url(@user) %>">Report Abuse</a>
+        <span class="report-abuse-link-wrapper" data-path="/report-abuse?url=<%= user_url(@user) %>"></span>
```
vaidehijoshi, line 164: Ah, are we switching from `<a>` to `<span data-path=...>` because the custom data attribute won't be indexed by Google?

benhalpern: Yeah. Basically we're taking a span and including the necessary data (the path) and then only creating the link for user-triggered action.

Google sees this as a useless span instead of a div. The dumb spider isn't smart enough to just ignore these kinds of links 🤪

PR #9060, opened 2020-07-01
UserSubscription backend refactor

### LINKED ISSUE #9022: Refactor UserSubscriptions controller
The logic for creating a `UserSubscription` in the `UserSubscriptionsController` is a little dense. Let's refactor that logic.

One approach is to create a service to handle this logic: https://github.com/thepracticaldev/dev.to/pull/8779#discussion_r443952447

atsmith813:
- [x] Refactor

## Description
Now that we have a better vision of what's needed and what's not, this PR is refactoring the backend logic for `UserSubscriptions`.
- Created `UserSubscriptions::Create` service to handle the logic for creating a UserSubscription to help clean up the controller.
- Moved some checks to model validations.
- Removed the reliance on email confirmation since that will no longer be included (for now).
- Update factories to more easily create a valid `UserSubscription`.

I'm sure the factory setup could be improved 🤔 . I found it a little challenging to get transient attributes to play nicely with associations, even with the [suggested workaround](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#with-associations). I also tried using callbacks.
Closes #9022

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

over_it_gif

### DIFF
--- a/app/controllers/user_subscriptions_controller.rb
+++ b/app/controllers/user_subscriptions_controller.rb
@@ -17,11 +17,14 @@ class UserSubscriptionsController < ApplicationController
     rate_limit!(:user_subscription_creation)
 
     source_type = user_subscription_params[:source_type]
-    return error_response("invalid source_type") unless UserSubscription::ALLOWED_TYPES.include?(source_type)
+    return error_response("Invalid source_type.") unless UserSubscription::ALLOWED_TYPES.include?(source_type)
 
     source_id = user_subscription_params[:source_id]
     source = source_type.constantize.find_by(id: source_id)
-    return error_response("source not found") unless source
+    return error_response("Source not found.") unless source
+
+    # TODO: [@thepracticaldev/delightful]: uncomment this once email confirmation is re-enabled
+    # return error_response("Subscriber email mismatch.") unless subscriber_email_is_current
 
     user_subscription = source.build_user_subscription(current_user)
 
@@ -34,6 +37,15 @@ class UserSubscriptionsController < ApplicationController
   end
 
   private
+
+  # This checks if the email address the user saw/consented to share is the
+  # same as their current email address. A mismatch occurs if a user updates
+  # their email address in a new/separate tab and then tries to subscribe on
+  # the old/stale tab without refreshing. In that case, the user would have
+  # consented to share their old email address instead of the current one.
+  def subscriber_email_is_current
+    current_user.email == user_subscription_params[:subscriber_email]
+  end
 
   def error_response(msg)
     render json: { error: msg, success: false }, status: :unprocessable_entity
--- a/app/controllers/user_subscriptions_controller.rb
+++ b/app/controllers/user_subscriptions_controller.rb
@@ -20,68 +20,23 @@ class UserSubscriptionsController < ApplicationController
     return error_response("invalid source_type") unless UserSubscription::ALLOWED_TYPES.include?(source_type)
 
     source_id = user_subscription_params[:source_id]
-    user_subscription_source = source_type.constantize.find_by(id: source_id)
-    return error_response("source not found") unless active_source?(source_type, user_subscription_source)
+    source = source_type.constantize.find_by(id: source_id)
+    return error_response("source not found") unless source
 
-    unless user_subscription_tag_enabled?(source_type, user_subscription_source)
-      return error_response("user subscriptions are not enabled for the requested source")
-    end
+    user_subscription = source.build_user_subscription(current_user)
 
-    return error_response("subscriber email mismatch") if subscriber_email_stale?
-    return error_response("cannot subscribe with Apple private relay email") if subscriber_authed_with_apple?
-
-    @user_subscription = user_subscription_source.build_user_subscription(current_user)
-
-    if @user_subscription.save
+    if user_subscription.save
       rate_limiter.track_limit_by_action(:user_subscription_creation)
       render json: { message: "success", success: true }, status: :ok
     else
-      error_response(@user_subscription.errors.full_messages.to_sentence)
+      error_response(user_subscription.errors.full_messages.to_sentence)
     end
   end
 
   private
 
-  def subscriber_authed_with_apple?
-    current_user.email.end_with?("@privaterelay.appleid.com")
-  end
-
-  def user_subscription_tag_enabled?(source_type, user_subscription_source)
-    liquid_tags =
-      case source_type
-      when "Article"
-        user_subscription_source.liquid_tags_used(:body)
-      else
-        user_subscription_source.liquid_tags_used
-      end
-
-    liquid_tags.include?(UserSubscriptionTag)
-  end
-
-  def active_source?(source_type, user_subscription_source)
-    return false unless user_subscription_source
-
-    # Don't create new user subscriptions for inactive sources
-    # (i.e. unpublished Articles, deleted Comments, etc.)
-    case source_type
-    when "Article"
-      user_subscription_source.published?
-    else
-      false
-    end
-  end
-
   def error_response(msg)
     render json: { error: msg, success: false }, status: :unprocessable_entity
-  end
-
-  # This checks if the email address the user saw/consented to share is the
-  # same as their current email address. A mismatch occurs if a user updates
-  # their email address in a new/separate tab and then tries to subscribe on
-  # the old/stale tab without refreshing. In that case, the user would have
-  # consented to share their old email address instead of the current one.
-  def subscriber_email_stale?
-    current_user&.email != user_subscription_params[:subscriber_email]
   end
 
   def user_subscription_params
--- a/app/controllers/user_subscriptions_controller.rb
+++ b/app/controllers/user_subscriptions_controller.rb
@@ -5,10 +5,7 @@ class UserSubscriptionsController < ApplicationController
 
   def subscribed
     params.require(%i[source_type source_id])
-    source_type = params[:source_type]
-    source_id = params[:source_id]
-
-    is_subscribed = UserSubscriptions::SubscriptionCacheChecker.call(current_user, source_type, source_id)
+    is_subscribed = UserSubscriptions::SubscriptionCacheChecker.call(current_user, params)
 
     render json: { is_subscribed: is_subscribed, success: true }, status: :ok
   end
--- a/app/controllers/user_subscriptions_controller.rb
+++ b/app/controllers/user_subscriptions_controller.rb
@@ -13,7 +13,7 @@ class UserSubscriptionsController < ApplicationController
   def create
     rate_limit!(:user_subscription_creation)
 
-    user_subscription = UserSubscriptions::CreateFromController.call(current_user, user_subscription_params)
+    user_subscription = UserSubscriptions::CreateFromControllerParams.call(current_user, user_subscription_params)
 
     if user_subscription.success
       rate_limiter.track_limit_by_action(:user_subscription_creation)
--- a/app/controllers/user_subscriptions_controller.rb
+++ b/app/controllers/user_subscriptions_controller.rb
@@ -13,7 +13,7 @@ class UserSubscriptionsController < ApplicationController
   def create
     rate_limit!(:user_subscription_creation)
 
-    user_subscription = UserSubscriptions::Create.call(current_user, user_subscription_params)
+    user_subscription = UserSubscriptions::CreateFromController.call(current_user, user_subscription_params)
 
     if user_subscription.success
       rate_limiter.track_limit_by_action(:user_subscription_creation)
--- a/app/controllers/user_subscriptions_controller.rb
+++ b/app/controllers/user_subscriptions_controller.rb
@@ -13,39 +13,14 @@ class UserSubscriptionsController < ApplicationController
   def create
     rate_limit!(:user_subscription_creation)
 
-    source_type = user_subscription_params[:source_type]
-    return error_response("Invalid source_type.") unless UserSubscription::ALLOWED_TYPES.include?(source_type)
+    user_subscription = UserSubscriptions::Create.call(current_user, user_subscription_params)
 
-    source_id = user_subscription_params[:source_id]
-    source = source_type.constantize.find_by(id: source_id)
-    return error_response("Source not found.") unless source
-
-    # TODO: [@thepracticaldev/delightful]: uncomment this once email confirmation is re-enabled
-    # return error_response("Subscriber email mismatch.") unless subscriber_email_is_current
-
-    user_subscription = source.build_user_subscription(current_user)
-
-    if user_subscription.save
+    if user_subscription.success
       rate_limiter.track_limit_by_action(:user_subscription_creation)
       render json: { message: "success", success: true }, status: :ok
     else
-      error_response(user_subscription.errors.full_messages.to_sentence)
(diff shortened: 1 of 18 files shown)


app/models/user_subscription.rb
```diff
@@ -16,20 +16,68 @@ class UserSubscription < ApplicationRecord
...
+      when "Article"
+        user_subscription_sourceable.liquid_tags_used(:body)
+      else
+        user_subscription_sourceable.liquid_tags_used
+      end
+
+    return if liquid_tags.include?(UserSubscriptionTag)
+
+    errors.add(:base, "User subscriptions are not enabled for the source.")
+  end
```
atsmith813, line 56: This process of "figuring out `liquid_tags_used`" will be refactored in another PR. Right now, `PollTag` also leverages it.

app/models/user_subscription.rb
```diff
@@ -16,20 +16,68 @@ class UserSubscription < ApplicationRecord
...
+
+    liquid_tags =
+      case user_subscription_sourceable_type
+      when "Article"
+        user_subscription_sourceable.liquid_tags_used(:body)
+      else
+        user_subscription_sourceable.liquid_tags_used
+      end
+
+    return if liquid_tags.include?(UserSubscriptionTag)
```
atsmith813, line 53: This process of "figuring out `liquid_tags_used`" will be refactored in another PR. Right now, `PollTag` also leverages it.

app/models/user_subscription.rb
```diff
@@ -16,20 +16,68 @@ class UserSubscription < ApplicationRecord
...
+
+    source_active =
+      # Don't create new user subscriptions for inactive sources
+      # (i.e. unpublished Articles, deleted Comments, etc.)
+      case user_subscription_sourceable_type
+      when "Article"
+        user_subscription_sourceable.published?
+      when "Comment"
+        !user_subscription_sourceable.deleted?
+      else
```
atsmith813, line 75: Not needed yet, but helps demonstrate why it's set up this way so I figured I may as well toss it in now. Won't hurt 🤷‍♂️ .

spec/services/user_subscriptions/create_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    user_subscription = described_class.call(subscriber, user_subscription_params)
+
+    expect(user_subscription.data).to be_nil
+    expect(user_subscription.success).to be false
+    expect(user_subscription.error).to eq "Invalid source_type."
+  end
+
+  it "returns an error for an invalid source" do
+    source = create(:article, :with_user_subscription_tag_role_user, with_user_subscription_tag: true)
+    user_subscription_params = { source_type: source.class.name, source_id: source.id + 999, subscriber_email: subscriber.email }
```
vaidehijoshi, line 18: `source_id: source.id + 999`

Is this just a way to create an invalid source to force the validation to fail?

app/controllers/user_subscriptions_controller.rb
```diff
@@ -5,85 +5,24 @@ class UserSubscriptionsController < ApplicationController
 
   def subscribed
     params.require(%i[source_type source_id])
-    source_type = params[:source_type]
-    source_id = params[:source_id]
-
-    is_subscribed = UserSubscriptions::SubscriptionCacheChecker.call(current_user, source_type, source_id)
+    is_subscribed = UserSubscriptions::SubscriptionCacheChecker.call(current_user, params)
```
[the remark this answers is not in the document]
atsmith813: Maybe we re-name it? `Subscription` is in the name because that's the cache we're checking - the subscriptions.

spec/services/user_subscriptions/create_spec.rb
atsmith813: Correct!

app/models/user_subscription.rb
```diff
@@ -16,20 +16,68 @@ class UserSubscription < ApplicationRecord
...
   end
 
-  def self.build_attributes(source, subscriber)
+  def self.build_attributes(source, subscriber, subscriber_email)
     {
       user_subscription_sourceable: source,
       author_id: source&.user_id,
       subscriber_id: subscriber&.id,
-      subscriber_email: subscriber&.email
+      subscriber_email: subscriber_email || subscriber&.email
```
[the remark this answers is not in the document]
atsmith813: This is part of us removing the email confirmation 😢 . Originally, and in the future, the client will send to the server a param of `subscriber_email`. That gets passed into the controller and eventually down to here.

Which actually reminds me, I should update https://github.com/thepracticaldev/dev.to/pull/9060/files#diff-590c19c8df4567bb0d7c239737d8581dR28 with a comment to pass it in once it's re-enabled 🎉

app/controllers/user_subscriptions_controller.rb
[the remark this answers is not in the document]
vaidehijoshi: Yeah, I won't push for it super hard, but I think the fact that it's namespaced makes it clear enough that the cache checker is for the (user) subscription. I just found the class name a bit verbose/repetitive here, so was trying to propose an alternative.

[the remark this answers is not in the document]
atsmith813: Sorry! To clarify what I meant...`UserSubscriptions` namespace is for the `UserSubscription` concept/model/object. `SubscriptionCacheChecker` is to check a user's subscriptions in cache.

If we cached more parts of `UserSubscription`, it would start to look something like this (in total, pseudo-names):
- `UserSubscriptions::SubscriptionCacheChecker` _what the user we are checking is subscribed to_
- `UserSubscriptions::SubscribersCacheChecker` _the users subscribed to the user we're checking_
- `UserSubscriptions::SourceCacheCacheker` _the sources the user we're checking is an author of that have subscriptions on them_

app/models/user_subscription.rb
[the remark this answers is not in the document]
atsmith813: Updated with comments and code!

app/controllers/user_subscriptions_controller.rb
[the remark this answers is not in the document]
vaidehijoshi: @atsmith813 Ah, thanks for clarifying this further. Definitely didn't have the whole context here since only the one class exists right now. (Clearly, I'm finding it hard to re-load all of the moving parts of this feature into my brain cache every time I review a liquid tag PR! 😅)

I think we can leave it as it is, and worry about renaming it if we add more "cache checkers". Might be worth adding what you just explained to me (that _Subscription_ `CacheChecker` is checking  _what the user we are checking is subscribed to_) as a comment in the class or where the class is used.

app/models/user_subscription.rb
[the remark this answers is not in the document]
atsmith813: On second thought...I can get rid of all of this 🤦 . Great call!

app/controllers/user_subscriptions_controller.rb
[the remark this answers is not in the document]
atsmith813: @vaidehijoshi I went in to add the comment and went ahead and renamed it to `IsSubscribedCacheChecker` since it returns a boolean. When I originally wrote it, I planned to return the subscriptions. But we ended up going a different direction with how that works. Since we settled on this, I updated it! What do you think? Does this feel more clear now?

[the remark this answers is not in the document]
vaidehijoshi: I like it this better! And it is much more in line with [this check](https://github.com/thepracticaldev/dev.to/pull/9060/commits/62b4684584706137ab975213c8f39f676d173398#diff-2ee6f8c98b7e8296071384bd2ee06ac9R8) now:
```ruby
is_subscribed = UserSubscriptions::IsSubscribedCacheChecker.call(current_user, params)
```

spec/requests/user_subscriptions_spec.rb
```diff
@@ -140,13 +137,12 @@
       end.to change(UserSubscription, :count).by(0)
 
       expect(response).to have_http_status(:unprocessable_entity)
-      expect(response.parsed_body["error"]).to eq("subscriber email mismatch")
+      expect(response.parsed_body["error"]).to eq("Subscriber email mismatch.")
```
mstruve, line 140: For these error checks it might be easier to do `include("email mismatch")` so if you tweak the wording you dont have to always update these specs.

app/services/user_subscriptions/create.rb
```diff
@@ -0,0 +1,72 @@
+module UserSubscriptions
+  class Create
```
mstruve, line 2: This class feels a little weird to me. We have validations in the model and then we have this creation class which has more validations in it that are manually adding errors to a general Struct. It seems to be mimicking exactly what ActiveRecord is already doing for us and feels like we might not need this abstraction.

app/models/user_subscription.rb
```diff
@@ -16,6 +16,10 @@ class UserSubscription < ApplicationRecord
   validates :user_subscription_sourceable_id, presence: true
   validates :user_subscription_sourceable_type, presence: true, inclusion: { in: ALLOWED_TYPES }
 
+  validate :tag_enabled
```
mstruve, line 19: NICE!

app/services/user_subscriptions/create.rb
atsmith813: Agreed - it's to take out some weird logic that's only needed in the controller. For example, when creating a `UserSubscription` we only need to check for the email mismatch because the request is coming from the front end. If we created one strictly on the backend (no use case for that right now 🤷‍♂️ ), that check wouldn't be necessary.

mstruve: Yeah, I totally get that. My concern here is that I would see this class and assume that every create needs to go through this class in order to include all the necessary logic rather than this class is only needed for the controller. Maybe a rename or controller helper module 🤔 I dont really have a great solution off the top of my head.

spec/requests/user_subscriptions_spec.rb
atsmith813: Good idea - updated!

app/services/user_subscriptions/create.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      # TODO: [@thepracticaldev/delightful]: uncomment this once email confirmation is re-enabled
+      # return response if subscriber_email_mismatch?
+
+      user_subscription = source.build_user_subscription(user)
+      if user_subscription.save
+        response.success = true
+        response.data = user_subscription
+      else
+        response.error = user_subscription.errors.full_messages.to_sentence
```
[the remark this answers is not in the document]
citizen428: This is a pattern I have used for a long time and suggested on the previous PR/the follow-up discussion with @atsmith813 on Slack. It's the main interface I like to use for my service classes, especially if the controller is for an API, as it now basically just needs to (a) pass on input params and (b) wrap return values accordingly.

app/services/user_subscriptions/create.rb
citizen428: The context was a discussion @atsmith813 and I had about avoiding conditional validations/callbacks etc. on models, as they can be hard to reason about and add cruft to a model despite only being used in one specific circumstance.

I do agree though that `Create` by itself is a bit misleading, as it wrongly suggests that this is the canonical way of creating `UserSubscription` instances. I tend to favor relatively explicit and verbose names for services, e.g. `UserSubscriptions::CreateFromXyzForm` or `UserSubscriptions::CreateFromUserInteraction`, etc.

app/services/user_subscriptions/create.rb
```diff
@@ -0,0 +1,72 @@
+module UserSubscriptions
+  class Create
+    attr_accessor :user, :source_type, :source_id, :subscriber_email, :response
+
+    def self.call(*args)
```
citizen428, line 5: Even though it's slightly more verbose, I always spell out the arguments and explicitly forward them. This method is the main interface to this service and `*args` doesn't tell a reader anything, so they have to check the initializer.

app/services/user_subscriptions/create.rb
```diff
@@ -0,0 +1,72 @@
...
+      new(*args).call
+    end
+
+    def initialize(user, user_subscription_params)
+      @user = user
+      @source_type = user_subscription_params[:source_type]
+      @source_id = user_subscription_params[:source_id]
+      @response = new_response
```
citizen428, line 13: Suggestion:

1. Assign the response struct to a constant (like Ruby does for all classes):
    ```ruby
    Response = Struct.new(:success, :data, :error, keyword_init: true)
    ```
3. Change this to:
    ```ruby
    @response = Response.new(success: false)
    ```
    You can leave off the arguments that are `nil`, structs are smart about this:
    ```ruby
    Response = Struct.new(:success, :data, :error, keyword_init: true)
    Response.new(success: false)
    #=> #<struct Response success=false, data=nil, error=nil>
    ```

If we start using this pattern in more places, we can actually extract a `BaseService` for others to inherit that already includes things like the response struct (and I usually add a method `success?` to wrap the instance variable in a more idiomatic way, etc).

app/services/user_subscriptions/create.rb
```diff
@@ -0,0 +1,72 @@
+module UserSubscriptions
+  class Create
+    attr_accessor :user, :source_type, :source_id, :subscriber_email, :response
```
citizen428, line 3: Do these really need to be accessors, i.e. should the be writable by consumers of this service?

app/services/user_subscriptions/create.rb
citizen428, line 33: A while back I added a helper on `ApplicationRecord`, so this can be 

```ruby
response.error = user_subscription.errors_as_sentence
```

app/services/user_subscriptions/create.rb
atsmith813: I agree! My thinking here though is we do it this way (`call(*args)` when using `call` in every other service we use. At this point, it's like an unofficial convention 🤣

app/services/user_subscriptions/create.rb
atsmith813: 1. I tried that originally, you get `SyntaxError:  dynamic constant assignment`.

app/services/user_subscriptions/create.rb
atsmith813: Nope - updated!

app/services/user_subscriptions/create.rb
atsmith813: Sweet - updated!

app/services/user_subscriptions/create.rb
atsmith813: I renamed it to `UserSubscriptions::CreateFromControllerParams`

app/services/user_subscriptions/create.rb
atsmith813: JK, looks like I need to just do a little more refactoring to get it to work with the constant. Updated.

app/services/user_subscriptions/create.rb
citizen428: I may make undoing this convention one of my cooldown projects. Our service objects generally have exactly one public interface method and we know exactly what arguments they need so we can initialize the objects. The lack of clarity for the caller IMHO weighs heavier than the slight redudancy.

PR #9072, opened 2020-07-02
Tags index design update

ludwiczakpawel:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This is a little layout and design update to our `/tags` page. It should look slightly bit better on production comparing with my screenshot since I have very little data locally and no logos :C. Definitely not a final state, but I think it's good-enough first step.

Before

image

After

image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


app/views/tags/index.html.erb
```diff
@@ -19,25 +19,52 @@
...
+
+  .tag-card:not(:hover) .crayons-btn--secondary.follow-action-button {
+    background: var(--button-secondary-bg) !important;
+    color: var(--button-secodnary-color) !important;
+  }
+</style>
+
+<div class="crayons-layout crayons-layout--1-col">
+  <header class="px-2 m:px-0">
     <h1>Top 100 Tags</h1>
```
rhymes, line 36: Just thought that there's no guarantee that a website has 100 tags :D 
https://letsbuild.gg/tags has 81
https://www.thismmalife.com/tags has 78.

I think we should drop the `100` from the title

app/views/tags/index.html.erb
```diff
@@ -19,25 +19,52 @@
...
+        <h3 class="crayons-tag crayons-tag--l mb-2">
+          <a href="/t/<%= tag.name %>" class="crayons-link">
+            <span class="crayons-tag__prefix">#</span><%= tag.name %>
+          </a>
+        </h3>
+        <% if tag.short_summary.present? %>
+          <p class="mb-2 truncate-at-3"><%= sanitize tag.short_summary %></p>
+        <% end %>
+        <p class="mb-3 fs-s color-base-50">
+          <%= pluralize @num_published_articles, "post" %> published
```
rhymes, line 52: I don't think this is going to work, it comes from a different controller.
You can use `tag.taggings_count`

app/views/tags/index.html.erb
ludwiczakpawel: Very good point. I'll fix it. Just gonna have "Top Tags" for now and eventually maybe "Tags"

app/views/tags/index.html.erb
```diff
@@ -19,25 +19,52 @@
   <meta name="twitter:image:src" content="<%= SiteConfig.main_social_image %>">
 <% end %>
 
-<div class="user-profile-header tag-header bg-transparent">
-  <div class="tag-or-query-header-container">
+<style>
```
nickytonline, line 22: Any reason for the `<style />` tag addition? Is this inline critical CSS?

ludwiczakpawel: Not critical, but useful :D I've done little "redesign" since yesterday, so we should not need it anymore.

ludwiczakpawel:
@rhymes I have indeed changed "branding" strategy a little. go ahead and take it for another ride.

ludwiczakpawel:
@rhymes you're right, i've just bumped the contrast ratio - should be good now.

PR #9082, opened 2020-07-02
UserSubscription liquid tag update & re-enable JS

### LINKED ISSUE #9023: Make email completely conditional in the UserSubscription liquid tag
For the time being, we do _not_ want to display the reader's email address when confirming a user subscription. We'll need to update the code (including Apple Auth logic) to be "modular" around email being present or not. That way, once we add it in later, "it just works".

atsmith813:
- [x] Refactor

## Description
This PR cleans up the UserSubscription liquid tag now that we have designs/markup completed. The updates include:
- Making the availability of the subscriber's email completely conditional now that we've postponed that functionality.
- Make the JS null safe.
- Specs 🎉 . Given the way we handle JS for liquid tags with "raw strings", I chose to use Capybara to test this tag.

Closes #9023 

**Open up your browser console and navigate around (i.e. click articles that don't have this liquid tag) and make sure you don't see any JS errors related to the liquid tag.** You'll likely see a few errors related to 3rd party services like Pusher or Cloudinary, but you shouldn't see any from this tag, especially about `null` references.

**Steps**
4. View your Article.
6. Refresh the Article (you should still see confirmation that you're already subscribed).
7. Sign out.
8. Revisit the Article (you should see the "sign-in" UX).
9. Sign back in via the `Sign in` button (you should see confirmation that you're already subscribed).

That's it 🎉 .

_Heads up: if you are testing and using an article you previously created locally with this liquid tag, you need to edit the article and re-publish/re-save it so the new markup is rendered._

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed - _I'll include documentation for this liquid tag in the final PR_

exhausted_gif

### DIFF
--- a/app/liquid_tags/user_subscription_tag.rb
+++ b/app/liquid_tags/user_subscription_tag.rb
@@ -14,27 +14,61 @@ class UserSubscriptionTag < LiquidTagBase
     // Hiding/showing elements
     // ***************************************
     function clearSubscriptionArea() {
-      document.getElementById('subscription-signed-in')?.classList.add("hidden");
-      document.getElementById('subscription-signed-out')?.classList.add("hidden");
-      document.getElementById('response-message')?.classList.add("hidden");
-      document.getElementById('subscriber-apple-auth')?.classList.add("hidden");
+      const subscriptionSignedIn = document.getElementById('subscription-signed-in');
+      if (subscriptionSignedIn) {
+        subscriptionSignedIn.classList.add("hidden");
+      }
+
+      const subscriptionSignedOut = document.getElementById('subscription-signed-out');
+      if (subscriptionSignedOut) {
+        subscriptionSignedOut.classList.add("hidden");
+      }
+
+      const responseMessage = document.getElementById('response-message');
+      if (responseMessage) {
+        responseMessage.classList.add("hidden");
+      }
+
+      const subscriberAppleAuth = document.getElementById('subscriber-apple-auth');
+      if (subscriberAppleAuth) {
+        subscriberAppleAuth.classList.add("hidden");
+      }
 
       hideConfirmationModal();
     }
 
     function showSignedIn() {
       clearSubscriptionArea();
-      document.getElementById('subscription-signed-in')?.classList.remove("hidden");
-      document.getElementById('profile-images')?.classList.remove("signed-out");
-      document.getElementById('profile-images')?.classList.add("signed-in");
+      const subscriptionSignedIn = document.getElementById('subscription-signed-in');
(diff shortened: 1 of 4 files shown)


app/views/liquids/_user_subscription.html.erb
```diff
@@ -71,9 +71,8 @@
           </header>
           <div class="crayons-modal__box__body">
             <p id="confirmation-text" class="fs-base mb-4 mt-0">
-              You'll share your email address
-              <span class="ltag__user-subscription-tag__subscriber-email"></span>
-              with
+              You'll share your email address<span class="ltag__user-subscription-tag__subscriber-email"></span>,
+              username, name, and DEV profile URL with
```
atsmith813, line 75: Being especially clear about what the user/subscriber will be sharing. Check out the screenshot in #9052 to see how this info is used.

app/liquid_tags/user_subscription_tag.rb
```diff
@@ -13,73 +13,70 @@ class UserSubscriptionTag < LiquidTagBase
...
-
     function clearSubscriptionArea() {
-      subscriptionSignedIn.classList.add("hidden");
-      subscriptionSignedOut.classList.add("hidden");
-      responseMessage.classList.add("hidden");
-      subscriberAppleAuth.classList.add("hidden");
+      document.getElementById('subscription-signed-in')?.classList.add("hidden");
+      document.getElementById('subscription-signed-out')?.classList.add("hidden");
+      document.getElementById('response-message')?.classList.add("hidden");
+      document.getElementById('subscriber-apple-auth')?.classList.add("hidden");
```
vaidehijoshi, line 20: out of curiosity, how does this work if we do `style.display = 'none';` here instead of adding a class? i noticed the slight flicker from the `Subscribe` button to the `You are already subscribed.` text, and I wonder if adding a class is contributing to that 🤔

atsmith813: It still flickers 😞  - it was originally that way with `display = 'none'`. The flicker is definitely something we should explore and improve. This is the first liquid tag with all these different "states" using async JS 🙈 .

vaidehijoshi: Hm, weird. Maybe I'll investigate it if I'm feeling adventurous during cool down! It's not major though, and if it was further down the page with actual _text_ in the article (as opposed to my 3 word article 😂), then it wouldn't even be noticeable!

atsmith813:
That's a _major_ bummer. Especially since it would only be an issue where this tag is used (rarely). I'll update it.

PR #9085, opened 2020-07-02
Cleanup existing liquid tags

### LINKED ISSUE #9024: Cleanup existing liquid tags with new updates
#8917 introduced some changes that will affect all liquid tags. We'll want to update all the existing liquid tags for things like:
- Changing the `tokens` argument to `parse_context`. 
- Making sure `super` is called in `initialize`.
- Refactor any quick wins with the use of context.

atsmith813:
- [x] Refactor

## Description
In #8917 and digging through the [`liquid` gem](https://github.com/Shopify/liquid) source code, we found that [the third argument passed to `initialize` in liquid tags is `parse_context`, not `tokens`](https://github.com/Shopify/liquid/blob/1feaa6381300d56e2c71b49ad8fee0d4b625147b/lib/liquid/tag.rb#L27).

While updating these references, I also:
- Ensured `super` is called in every liquid tag.
- Denoted unused arguments in initialize methods.

I intentionally skipped `PollTag`, I need to update more logic on that. That'll be a separate PR.

Closes #9024

Specs should be appropriate here since I'm not changing any logic or behavior. If you'd like, you can include a bunch of different liquid tags in an article locally to ensure they work 🎉 .

## Added tests?

## Added to documentation?
- [x] no documentation needed

cleanup_gif

### DIFF
--- a/app/liquid_tags/github_tag.rb
+++ b/app/liquid_tags/github_tag.rb
@@ -1,9 +1,7 @@
 class GithubTag < LiquidTagBase
-  def initialize(tag_name, link, parse_context)
+  def initialize(_tag_name, link, _parse_context)
     super
-    @tag_name = tag_name
     @link = link
-    @parse_context = parse_context
     @rendered = pre_render
   end
 
--- a/app/liquid_tags/github_tag.rb
+++ b/app/liquid_tags/github_tag.rb
@@ -1,9 +1,9 @@
 class GithubTag < LiquidTagBase
-  def initialize(tag_name, link, tokens)
+  def initialize(tag_name, link, parse_context)
     super
     @tag_name = tag_name
     @link = link
-    @tokens = tokens
+    @parse_context = parse_context
     @rendered = pre_render
   end
 
--- a/app/liquid_tags/katex_tag.rb
+++ b/app/liquid_tags/katex_tag.rb
@@ -2,7 +2,7 @@ class KatexTag < Liquid::Block
   PARTIAL = "liquids/katex".freeze
   KATEX_EXISTED = "katex_existed".freeze
 
-  def initialize(tag_name, markup, parse_context)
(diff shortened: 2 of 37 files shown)


app/liquid_tags/katex_tag.rb
```diff
@@ -2,7 +2,7 @@ class KatexTag < Liquid::Block
   PARTIAL = "liquids/katex".freeze
   KATEX_EXISTED = "katex_existed".freeze
 
-  def initialize(tag_name, markup, tokens)
+  def initialize(_tag_name, markup, _parse_context)
```
juliannatetreault, line 5: I think that `markup` could benefit from being written as `_markup` here since it seems unused - thoughts? 😄

vaidehijoshi: @juliannatetreault Ah but it is being used [here](https://github.com/thepracticaldev/dev.to/blob/7930a1a784d975a014d2160d436fae21af3ac780/app/liquid_tags/katex_tag.rb#L38); it's just a little obfuscated since it is [initialized by the gem itself](https://github.com/Shopify/liquid/blob/1feaa6381300d56e2c71b49ad8fee0d4b625147b/lib/liquid/tag.rb#L29)

juliannatetreault: Ohh, I see! Thanks for pointing that out.

app/liquid_tags/github_tag.rb
```diff
@@ -1,9 +1,9 @@
 class GithubTag < LiquidTagBase
-  def initialize(tag_name, link, tokens)
+  def initialize(tag_name, link, parse_context)
     super
     @tag_name = tag_name
     @link = link
-    @tokens = tokens
+    @parse_context = parse_context
```
rhymes, line 6: i think this can be removed as if changing the name didn't break anything it means it's not used by the view or the tag

PR #9087, opened 2020-07-02
Update PollTag

### LINKED ISSUE #9026: Cleanup PollTag
The PollTag is a "prototyped" liquid tag that isn't available for broad use but it is live. With the changes introduced by #8917, we'll want to update this tag to keep things consistent:
- Change `tokens` to `parse_context`.
- Add context restriction.
- Add role restrictions.
- Cleanup references and validations in the `Article` model.

atsmith813:
- [x] Refactor

## Description
The `PollTag` was a beta/prototype liquid tag that is currently "live", but unused. In #8917, we refactored how permissions and contexts work with liquid tags. I updated the `PollTag` to use that new logic.

Closes #9026 

Specs or you can create a `Poll` locally and then use the tag in an article and very it works as `admin` user and only in Articles 🎉 . This liquid tag isn't documented anywhere and it's quite possible it doesn't work anymore anyway 😬 .

## Added tests?

## Added to documentation?
- [x] no documentation needed

dancing_parrot_gif

### DIFF
--- a/app/liquid_tags/poll_tag.rb
+++ b/app/liquid_tags/poll_tag.rb
@@ -1,5 +1,10 @@
 class PollTag < LiquidTagBase
   PARTIAL = "liquids/poll".freeze
+  VALID_CONTEXTS = %w[Article].freeze
+  VALID_ROLES = %i[
+    admin
+    super_admin
+  ].freeze
 
   SCRIPT = <<~JAVASCRIPT.freeze
     if (document.head.querySelector('meta[name="user-signed-in"][content="true"]')) {
@@ -95,7 +100,8 @@ class PollTag < LiquidTagBase
     }
   JAVASCRIPT
 
-  def initialize(_tag_name, id_code, _tokens)
+  def initialize(_tag_name, id_code, _parse_context)
+    super
     @poll = Poll.find(id_code)
   end
 
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -9,7 +9,6 @@ class Article < ApplicationRecord
   SEARCH_SERIALIZER = Search::ArticleSerializer
   SEARCH_CLASS = Search::FeedContent
   DATA_SYNC_CLASS = DataSync::Elasticsearch::Article
-  RESTRICTED_LIQUID_TAGS = [PollTag, UserSubscriptionTag].freeze
 
   acts_as_taggable_on :tags
   resourcify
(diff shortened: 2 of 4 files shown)


app/models/article.rb
```diff
@@ -352,22 +350,6 @@ def update_score
...
-        comments_blob
-      else
-        "#{body_markdown}#{comments_blob}"
-      end
-
-    MarkdownParser.new(content, source: self, user: user).tags_used
-  rescue StandardError
-    []
-  end
-
```
atsmith813, line 370: This was originally implemented _just_ for the validation of the `PollTag`. Now that we handle the validation differently, we don't need this on the model anymore.

app/models/user_subscription.rb
```diff
@@ -41,20 +41,21 @@ def self.build_attributes(source, subscriber)
...
     errors.add(:base, "User subscriptions are not enabled for the source.")
   end
 
+  def liquid_tags_used
+    MarkdownParser.new(
+      user_subscription_sourceable.body_markdown,
+      source: user_subscription_sourceable,
+      user: user_subscription_sourceable.user,
+    ).tags_used
+  rescue StandardError # Can occur during parsing for improper tags or bad args
```
atsmith813, line 55: If a liquid tag happens to raise an error in `initialize`, we want to ignore it. We're checking only for the use of tags. The _proper_ use of liquid tags is handled elsewhere.

app/models/article.rb
mstruve: I LOVE that this is not longer on the article!!!

app/liquid_tags/poll_tag.rb
```diff
@@ -1,5 +1,10 @@
 class PollTag < LiquidTagBase
```
mstruve, line 1: Where do we test these tags? I went looking for specs but couldn't find them for these individual classes.

atsmith813: This tag, in particular, doesn't have specs. It was a [quick proof of concept Ben did awhile back](https://dev.to/devteam/feedback-on-this-new-dev-feature-5cai). It's only used in like 2 articles by Ben. Worst case scenario, if this tag broke, it would still only be an issue if Ben re-saved those articles causing it to re-render.

The logic for liquid tag permissions, which was the change, is/will be tested in another PR.

FWIW, I tried to figure out the `PollTag` thing, but it uses `Polls` and I couldn't figure out how some of that logic was working and I suspect some of it may have been removed. It started taking more time than it was worth because that tag isn't used.

I'm willing to circle back and unpack it all. I'd have to put this on hold and circle back sometime.

atsmith813: I added specs for context and role validations (logic that was also added to this tag) in #9107 👍 .

PR #9091, opened 2020-07-03
Check for report abuse link existence before calling it

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

There is an issue where this does not always exist here. This should fix things.


benhalpern:
Nevermind, this was actually fixed in #9088 which I did not see at first. Carry on.

PR #9144, opened 2020-07-05
Fix pro link for organization users

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The "pro analytics for org" link should only appear if the user has the pro role.

Closes https://github.com/thepracticaldev/dev.to/issues/9106

1. create an org
1. add and remove the `:pro` role to see the link appearing or disappearing

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/views/dashboards/_actions_mobile.html.erb
+++ b/app/views/dashboards/_actions_mobile.html.erb
@@ -8,18 +8,19 @@
     <option value="/dashboard/following_podcasts" <%= "selected" if params[:action] == "following_podcasts" %>>Following podcasts (<%= @user.following_podcasts_count %>)</option>
 
     <option value="/listings/dashboard">Listings</option>
-    <% if @user.has_role? :pro %>
-      <option value="/dashboard/pro">Pro Analytics</option>
-    <% end %>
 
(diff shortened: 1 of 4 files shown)


app/views/dashboards/_actions_mobile.html.erb
```diff
@@ -8,18 +8,19 @@
...
+      <% if @organizations && (params[:which].blank? || params[:which] == "organization") %>
+        <% @organizations.each do |org| %>
+          <option value="<%= dashboard_pro_org_path(org.id) %>">Pro Analytics for <%= org.name %></option>
+        <% end %>
       <% end %>
     <% end %>
 
-    <% if current_user.created_at < 2.weeks.ago %>
-      <option value="/videos/new">Upload a video</option>
+    <% if policy(:video).new? %>
```
rhymes, line 22: this way we have the condition only in one place, `VideoPolicy`

mstruve: Good catch!

PR #9161, opened 2020-07-06
Add subscriptions to the dashboard

### LINKED ISSUE #8169: Add subscriptions to the dashboard

atsmith813:
- [x] Feature

## Description
This PR add subscriptions to the dashboard. Big shoutout to @ludwiczakpawel from #9052 💯 .

- [x] [Add subscriptions to the dashbaord](https://github.com/thepracticaldev/dev.to/issues/8169)
- [ ] [Create a new role for users approved to use the new user subscription liquid tag and add docs](https://github.com/thepracticaldev/dev.to/issues/8165)

Closes #8169

2. Write a post and include the liquid tag {% user_subscription Some sweet CTA text %}.
4. View your Article.
6. Go to your dashboard.
7. You should see subscriptions on the article you just created and subscribed to - click that.
8. See your subscriptions 🎉 .

**The data icon shows the number of subscriptions**
Screen Shot 2020-07-06 at 10 03 39 AM

**You can also see subscriptions in the dropdown menu**
Screen Shot 2020-07-06 at 10 03 46 AM

**Subscriptions page**
Screen Shot 2020-07-07 at 10 12 23 AM

**Accessing a route directly for an article without any subscriptions**
Screen Shot 2020-07-06 at 10 11 31 AM

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

stay_focused_gif

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -23,8 +23,6 @@ class DashboardsController < ApplicationController
       @articles = target.articles.includes(:organization)
     end
 
-    @articles = @articles.includes(:user_subscriptions)
-
     @reactions_count = @articles.sum(&:public_reactions_count)
     @page_views_count = @articles.sum(&:page_views_count)
 
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -2,6 +2,7 @@ class DashboardsController < ApplicationController
   before_action :set_no_cache_header
   before_action :authenticate_user!
   before_action :fetch_and_authorize_user, except: :pro
+  before_action :set_source, only: %i[subscriptions]
   before_action -> { limit_per_page(default: 80, max: 1000) }, except: %i[show pro]
   after_action :verify_authorized
 
@@ -21,6 +22,8 @@ class DashboardsController < ApplicationController
       # if the target is a user, we need to eager load the organization
       @articles = target.articles.includes(:organization)
     end
+
+    @articles = @articles.includes(:user_subscriptions)
 
     @reactions_count = @articles.sum(&:public_reactions_count)
     @page_views_count = @articles.sum(&:page_views_count)
@@ -70,7 +73,21 @@ class DashboardsController < ApplicationController
     @organizations = current_user.member_organizations
   end
 
+  def subscriptions
+    authorize @source
+    @subscriptions = @source.user_subscriptions.
+      includes(:subscriber).order(created_at: :desc).page(params[:page]).per(100)
+  end
+
   private
+
+  def set_source
+    source_type = params[:source_type]
+    not_found unless UserSubscription::ALLOWED_TYPES.include? source_type
+
+    source = source_type.constantize.find_by(id: params[:source_id])
+    @source = source || not_found
+  end
 
   def fetch_and_authorize_user
     @user = if params[:username] && current_user.any_admin?
--- a/app/policies/article_policy.rb
+++ b/app/policies/article_policy.rb
@@ -34,6 +34,10 @@ class ArticlePolicy < ApplicationPolicy
        archived]
   end
 
+  def subscriptions?
+    user_is_author? || user_admin?
+  end
+
   private
 
   def user_is_author?
--- a/app/views/dashboards/_dashboard_article_row.html.erb
+++ b/app/views/dashboards/_dashboard_article_row.html.erb
@@ -58,11 +58,11 @@
           <% end %>
         </span>
 
-        <% if false %>
-          <span class="flex items-center p-1 ml-1" title="Data">
+        <% if article.user_subscriptions.any? %>
+          <a href="<%= dashboard_subscriptions_path(source_type: "Article", source_id: article.id) %>" class="flex items-center crayons-btn crayons-btn--ghost crayons-btn--s fw-normal color-base-70" title="Subscriptions">
             <%= inline_svg_tag("small-data.svg", aria: true, class: "crayons-icon mr-1", title: "Data") %>
-            4
-          </span>
+            <%= article.user_subscriptions.size %>
+          </a>
         <% end %>
       </div>
     <% end %>
@@ -94,9 +94,12 @@
         <%= inline_svg_tag("overflow-horizontal.svg", aria: true, class: "crayons-icon", title: "More...") %>
       </button>
 
-      <div class="crayons-dropdown top-100 right-0 z-10 align-left js-dashboard-row-more-dropdown">
+      <div class="crayons-dropdown top-100 right-0 z-10 align-left js-dashboard-row-more-dropdown p-1">
+        <% if article.user_subscriptions.any? %>
+          <a href="<%= dashboard_subscriptions_path(source_type: "Article", source_id: article.id) %>" class="crayons-link crayons-link--block w-100"><%= "Subscriptions (#{article.user_subscriptions.size})" %></a>
+        <% end %>
         <% if @current_user_pro %>
-          <a href="<%= article.path %>/stats" class="crayons-link crayons-link--block w-100 border-0 bg-transparent">Stats</a>
+          <a href="<%= article.path %>/stats" class="crayons-link crayons-link--block w-100">Stats</a>
         <% end %>
         <%= form_for(article, method: :patch, remote: true, authenticity_token: true, html: { "data-type": "json", class: "js-archive-toggle" }) do |f| %>
           <%= f.hidden_field :archived, value: !article.archived, id: "article_archived_#{article.id}" %>
--- a/app/views/dashboards/_dashboard_article_row.html.erb
+++ b/app/views/dashboards/_dashboard_article_row.html.erb
@@ -58,10 +58,10 @@
           <% end %>
         </span>
 
-        <% if article.user_subscriptions.any? %>
+        <% if article.user_subscriptions_count > 0 %>
           <a href="<%= dashboard_subscriptions_path(source_type: "Article", source_id: article.id) %>" class="flex items-center crayons-btn crayons-btn--ghost crayons-btn--s fw-normal color-base-70" title="Subscriptions">
             <%= inline_svg_tag("small-data.svg", aria: true, class: "crayons-icon mr-1", title: "Data") %>
-            <%= article.user_subscriptions.size %>
+            <%= article.user_subscriptions_count %>
           </a>
         <% end %>
       </div>
@@ -95,8 +95,8 @@
       </button>
 
       <div class="crayons-dropdown top-100 right-0 z-10 align-left js-dashboard-row-more-dropdown p-1">
-        <% if article.user_subscriptions.any? %>
-          <a href="<%= dashboard_subscriptions_path(source_type: "Article", source_id: article.id) %>" class="crayons-link crayons-link--block w-100"><%= "Subscriptions (#{article.user_subscriptions.size})" %></a>
+        <% if article.user_subscriptions_count > 0 %>
+          <a href="<%= dashboard_subscriptions_path(source_type: "Article", source_id: article.id) %>" class="crayons-link crayons-link--block w-100"><%= "Subscriptions (#{article.user_subscriptions_count})" %></a>
         <% end %>
         <% if @current_user_pro %>
           <a href="<%= article.path %>/stats" class="crayons-link crayons-link--block w-100">Stats</a>
--- a/app/views/dashboards/subscriptions.html.erb
+++ b/app/views/dashboards/subscriptions.html.erb
@@ -1,0 +1,39 @@
+<% title "Subscriptions #{@source.title}" %>
+
+<div class="crayons-layout crayons-layout--1-col">
+  <header class="flex items-center justify-between p-2 m:p-0">
+    <div>
+      <h1 class="fs-2xl s:fs-3xl">Subscriptions</h1>
+      <h2 class="fs-base s:fs-xl fw-medium">
+        <em class="fs-italic fw-normal color-base-60">for</em>
+        <a href=<%= @source.path %>><%= @source.title %></a>
+      </h2>
+    </div>
+    <%#TODO: [@thepracticaldev/delightful]: uncomment this when ready to implement CSV exports%>
+    <%#<button type="button" class="crayons-btn crayons-btn--secondary">Export CSV</button>%>
+  </header>
+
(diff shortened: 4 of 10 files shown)


atsmith813:
I'll need an admin override for Codeclimate.

app/controllers/dashboards_controller.rb
```diff
@@ -70,8 +73,22 @@ def pro
     @organizations = current_user.member_organizations
   end
 
+  def subscriptions
+    authorize @source
+    @subscriptions = @source.user_subscriptions.
+      includes(:subscriber).order(created_at: :desc).page(params[:page]).per(100)
```
atsmith813, line 79: Paging isn't technically needed yet, but I figured I may as well 🤷‍♂️

app/policies/article_policy.rb
```diff
@@ -34,6 +34,10 @@ def permitted_attributes
        archived]
   end
 
+  def subscriptions?
+    user_is_author? || user_admin?
```
atsmith813, line 38: This can easily be extended to allow organization admins to see subscriptions for articles they didn't write themselves in the future.

app/views/dashboards/subscriptions.html.erb
```diff
@@ -0,0 +1,39 @@
+<% title "Subscriptions #{@source.title}" %>
```
atsmith813, line 1: This won't quite work if/when we extend subscriptions to other objects, like comments, in the future (`Comment` doesn't have a `title`). We can cross that bridge when we get there. It's a quick update. We also don't have the concept of a "comment dashboard" so it's possible that view would live somewhere else anyways.

app/views/dashboards/_dashboard_article_row.html.erb
```diff
@@ -58,11 +58,11 @@
           <% end %>
         </span>
 
-        <% if false %>
-          <span class="flex items-center p-1 ml-1" title="Data">
+        <% if article.user_subscriptions.any? %>
```
mstruve, line 61: This is going to issue a db call every time we call it which is twice, here and below on line 98. We should cache the value and then apply it in each place. OR if we wanted to take our db calls from 3 to 1, we could just do `article.user_subscriptions.size` and then for the booleans check if size is greater than 0

spec/requests/dashboard_spec.rb
```diff
@@ -39,6 +39,19 @@
         expect(response.body).to include "Delete"
       end
 
+      it "renders subscriptions for articles with subscriptions" do
+        user.add_role(:admin) # TODO: (Alex Smith) - update roles before release
```
mstruve, line 43: One suggestion, we have had some issues with flakiness around roles in specs, one way to avoid this would be to stub the role so `user.admin?` returns true rather than adding the role itself. This would also help speed up the specs. Just a thought.

atsmith813: The roles are going to change in the next PR, so I'll address it there 👍 .

app/views/dashboards/_dashboard_article_row.html.erb
atsmith813: Darn, I thought `.any?` was smart enough to check it in memory since we eagerly loaded them already 🤔 . I'll cache it 😢 .

atsmith813: @mstruve Updated to use `counter_culture` 🎉  Thanks for catching this!

db/migrate/20200706184804_add_user_subscriptions_count_to_articles.rb
```diff
@@ -0,0 +1,9 @@
+class AddUserSubscriptionsCountToArticles < ActiveRecord::Migration[6.0]
+  def self.up
+    add_column :articles, :user_subscriptions_count, :integer, null: false, default: 0
```
mstruve, line 3: I don't think we need to add a whole new column for this, I just meant cacheing it in an instance variable in the controller that we could then use it in the views.

atsmith813: We need the count on each article for this view. What does that look like with instance variables?

atsmith813: Bullet doesn't like us using `.size` and throws a message about using a counter cache instead.

mstruve: I think that makes sense if you are calling `.size` all over the place but in this instance we all need it in one place so in the controller I would do 
```
@article_sub_count = article.subscriptions.count
```
Then in the views use that instance variable.

mstruve: @atsmith813 and I chatted a bit about the pros and cons of adding a column vs locally caching this on the article. I have a hard time with adding a column that will be minimally used(not many articles will have it) and is only to show a nice to have stat. In the end, users are subscribing to users, not articles so this number is less a must have and more a nice to have. Adding a whole column to an already bloated model for a nice to have doesnt sit well with me. On the flip side it does eliminate the need for a query. 

Interested to hear other's opinions!

atsmith813: Going to chime in here with some additional context @mstruve.

**TLDR** - I'm thinking what we have here is the move and the challenge (issue?) is a symptom of something else that we can address separately in another PR after a team discussion.

> so this number is less a must have and more a nice to have.

In this same view, we use the same setup as this PR is now trying to do using the DB to cache counts. It's what we do in other controllers. It's how the `counter_culture` gem and ActiveRecord's `counter_cache` (which we aren't using right now) work. In all areas, the counts are stored on the records so an additional query isn't needed. The counts are there for other items in this view (page views, comments, reactions, etc.) _and_ they use the same approach. It is, in some ways, a nice to have the count of subscriptions. It's also consistent with other areas of this view.

> In the end, users are subscribing to users, not articles

I don't think that's the case anymore. That's one way of looking at it. If we look back at what we have now that's not as true as it was when we first discussed this. If it were purely about a user to user, we would have a view where an author just saw a list of emails/subscribers, regardless of where the subscription came from. It would also mean that the uniqueness of a subscription would be enforced strictly by user IDs. Neither of those is how it works. We show the subscriptions by Article (source) and a user can subscribe to the same author via multiple articles. So for all intents and purposes, it's actually setup more to be that users are subscribing to articles. When it comes to delivering the UX to both readers and authors, the context/source of the subscription always matters.

> Adding a whole column to an already bloated model

In other words, if we were taking this approach on a different model that wasn't so bloated, would it be okay?

Either way, it feels like the issue is more about the Article model and that counting subscriptions is a symptom of this issue. It doesn't feel great that we would be considering breaking convention, ignore the warnings from `Bullet`, and code up a custom exception here. This issue will inevitably arise again when we want to efficiently count something else related to Articles, which is at the crux of what we do.

It would feel better if the issue of a bloated model, especially as it relates to the counting of relationship records, were addressed separately. In the case of this PR, we are very slightly adding to an existing problem by adding another column for counts. When we figure out a gameplan to address this separately, this implementation would get swept up with the counts of comments, page views, reactions, and many other columns.

I understand there's a cost and a tradeoff that's growing with each column we keep adding to `Article` and that cost doesn't feel worth it for something used that will currently be used so little. Since this is for Codeland, this will be used by major organizations, and it's following convention, I'm hoping we can stick to it here.

ludwiczakpawel:
Yeap, looks good design-wise. Not going to accept since @lisasy has already done it for design and I can see there's an ongoing debate around some backend-related issue so I'd rather prefer someone else to give another accept so it doesn't get merged by mistake before all issues are resolved.

app/views/dashboards/_dashboard_article_row.html.erb
rhymes: for the rationale behind any and the other methods https://www.speedshop.co/2019/01/10/three-activerecord-mistakes.html

unfortunately the whole deal is quite complicated :D

db/migrate/20200706184804_add_user_subscriptions_count_to_articles.rb
mstruve: > Neither of those is how it works. We show the subscriptions by Article (source) and a user can subscribe to the same author via multiple articles.

When did that change? And what is the purpose of subscribing to a user multiple times? Would people make different email lists based on articles? If that is the case we might want to make sure its explicit that if you sign up via one article you may end up on a different email list than another.

atsmith813: I should clarify - it's less of a change and I think more of a misunderstanding and miscommunication.

> Would people make different email lists based on articles?

Correct! An author could write an article about Rails, gather subscriptions, and email those subscribers relevant content. The same author could write another article about React, gather subscriptions, and email those subscribers relevant and different content. The author may have a "one time ask" related to an event, a survey, etc.

That's where the call-to-action text by the author comes in handy. It tells the user why they're clicking "subscribe". Something like, _"want to learn more about Rails? Subscribe below to be included in our weekly updates about the top Rails tips!"_.

The purpose is for a user to express intent/interest to a specific call-to-action and for an author to be able to "segment" their subscribers enabling both sides more power to be connected to more relevant content.

In the event an author doesn't include any call-to-action text, we make it clear to the user what they're doing and what they're sharing.

Technically, going back to the example above of an author writing about Rails and React, there's nothing stopping that author from gathering all subscribers from both lists and emailing them the same content, regardless of relevancy. Once the subscription is made, the rest of the interaction between the author and the subscriber moves outside of our platform. The author would be subject to the "standards of email marketing" like giving the user a way to unsubscribe from them.

app/views/dashboards/subscriptions.html.erb
```diff
@@ -0,0 +1,39 @@
...
+  <header class="flex items-center justify-between p-2 m:p-0">
+    <div>
+      <h1 class="fs-2xl s:fs-3xl">Subscriptions</h1>
+      <h2 class="fs-base s:fs-xl fw-medium">
+        <em class="fs-italic fw-normal color-base-60">for</em>
+        <a href=<%= @source.path %>><%= @source.title %></a>
+      </h2>
+    </div>
+    <%#TODO: [@thepracticaldev/delightful]: uncomment this when ready to implement CSV exports%>
+    <%#<button type="button" class="crayons-btn crayons-btn--secondary">Export CSV</button>%>
```
vaidehijoshi, line 13: lol i like the optimism behind adding this button 😂

app/views/dashboards/subscriptions.html.erb
```diff
@@ -0,0 +1,39 @@
...
+  </header>
+
+  <main>
+    <% if @subscriptions.any? %>
+      <table class="crayons-table w-100">
+        <tbody class="crayons-card">
+          <% @subscriptions.each do |subscription| %>
+            <% subscriber = subscription.subscriber %>
+            <tr>
+              <td class="fw-medium"><a href=<%= "/#{subscriber.username}" %> <%= "#{subscriber.name} DEV profile" %>><%= subscriber.name %></a></td>
```
vaidehijoshi, line 23: is it possible to user a path helper here instead of a direct `"/#{subscriber.username}"` href?

spec/requests/dashboard_spec.rb
```diff
@@ -300,4 +313,77 @@
...
+      user.add_role(:admin)
+      article_with_user_subscription_tag = create(:article, :with_user_subscription_tag_role_user, with_user_subscription_tag: true)
+      create(:user_subscription,
+             subscriber_id: second_user.id,
+             subscriber_email: second_user.email,
+             author_id: article_with_user_subscription_tag.user_id,
+             user_subscription_sourceable: article_with_user_subscription_tag)
+
+      expect do
+        get "/dashboard/subscriptions", params: { source_type: article_with_user_subscription_tag.class.name, source_id: article_with_user_subscription_tag.id }
```
vaidehijoshi, line 351: I see this being used a couple times throughout the test; can we pull it out into a variable? This specific `get` is really hard to read/wrap your head around :)

atsmith813: It is used a couple of times and there are variations of this in the file where it's slightly different.

What if I did:

```ruby
expect do
  params = {
    source_type: article_with_user_subscription_tag.class.name,
    source_id: article_with_user_subscription_tag.id
  }

  get "/dashboard/subscriptions", params: params
end.to raise_error(Pundit::NotAuthorizedError)
```

Does this help with readability?

vaidehijoshi: @atsmith813 Hm, I thought it was the same line that is used 4 times on 332, 351, 374, and 385, which is why I suggested extracting it 🤔

atsmith813: @vaidehijoshi that's correct! And 363 and 337 are different. I usually find it more confusing to override an extraction (`params`) only in specific specs (363 and 337) so I usually only extract it out if it's going to be consistent (like putting `user` at the top level).

I'll extract it though and re-name them.

atsmith813: I'm also hesitant to extract when the params are defined by something else in the spec (`article_with_user_subscription_tag`). That has a low chance of changing in the future here so it should be fine to take it out as well.

atsmith813: @vaidehijoshi On second thought, I'll revisit this in the next PR when I change the specs again to update them for the new roles. The challenge here is sometimes I make the `article_with_user_subscription_tag` differently, depending on the spec. And I do that because of roles.

app/views/dashboards/subscriptions.html.erb
atsmith813: It is! I created a path helper for this path and used that instead.

spec/requests/dashboard_spec.rb
vaidehijoshi: @atsmith813 Ok, that's fair! Seems like it will be easier to change it in the next PR pass through.

PR #9169, opened 2020-07-06
Bug Fix: Prevent Duplicate Follow Attempts by Preventing Multiple Workers

mstruve:
- [x] Bug Fix
- [x] Optimization

## Description
We are still getting duplicate follow attempts and it looks like its bc we are kicking off multiple workers at once. This will prevent that from happening so at least part of the problem will be fixed, I am not sure if it will fix all of it. 

Fixes Sidekiq Error:
```
ActiveRecord::RecordInvalid: Validation failed: Followable has already been taken
```

Make sure you can still follow someone

## Added tests?

alt_text

### DIFF
--- a/app/workers/users/follow_worker.rb
+++ b/app/workers/users/follow_worker.rb
@@ -1,7 +1,7 @@
 module Users
   class FollowWorker
     include Sidekiq::Worker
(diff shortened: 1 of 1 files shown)


rhymes:
😢

PR #9170, opened 2020-07-06
Add role and docs for user subscription liquid tag

### LINKED ISSUE #9086: Finalize UserSubscription liquid tag - create new role and add docs
To put the bow on the `UserSubscription` liquid tag, we need a few things:
- Document the new liquid tag.
- Create a new role to apply to authorized users of this tag.
- Update the role in code.
- Clear up any remaining "TODOs".

atsmith813:
- [x] Feature

## Description
This PR adds the new role that's going to be used for the liquid tag and adds some documentation to liquid tags for using it.

A few things to note:
- The way we currently handle documenting liquid tags needs a re-factor, but that's outside the scope of this PR. Currently, we render directions for liquid tags in 3 different pages ([`/app/views/_editor_liquid_help.html.erb`](https://github.com/thepracticaldev/dev.to/blob/master/app/views/pages/_editor_liquid_help.html.erb), [`/app/views/pages/markdown_basics.html.erb`](https://github.com/thepracticaldev/dev.to/blob/master/app/views/pages/markdown_basics.html.erb), [`/app/views/pages/_editor_guide_text.html.erb`](https://github.com/thepracticaldev/dev.to/blob/master/app/views/pages/markdown_basics.html.erb)). Ideally, it would be one location that the different areas render as needed.
- Furthermore, we could probably come up with a better way to handle the conditional rendering of liquid tag documentation. For example, show only liquid tags that can be used in articles in the editor.
- The design of liquid tag documentation will likely be updated in the near future as well. You'll notice we aren't consistent with how we do it right now.
- For now, since the `user_subscription` liquid tag is the first and only liquid tag that's context and role-specific, I just took the shortcut. I render the documentation for it if the user has any of the tag's permitted roles and it's hard-coded to say it's only for use in articles.

- [x] [Add subscriptions to the dashbaord](https://github.com/thepracticaldev/dev.to/issues/8169)
- [x] [Create a new role for users approved to use the new user subscription liquid tag and add docs](https://github.com/thepracticaldev/dev.to/issues/8165)

Closes #9086 

_Setup_
Grab a user record that you can sign in with. For me, that's usually `User.last`.
```ruby
user = User.last
# or
user = User.find_by(email: 'emailaddressofyourlocaluser@hey.com')
```
and add the new role to this user:
```ruby
user.add_role(:restricted_liquid_tag, LiquidTags::UserSubscriptionTag)
```
1. Go to the editor - **Write a post** / https://localhost:3000/new.
2. Click into the body of the article editor to trigger the helper to pop up on the right.
3. Click the **Liquid Tags** link in the helper.
4. Make sure you see the documentation for the user subscription tag 🎉 .

Screen Shot 2020-07-06 at 2 01 48 PM

## Added tests?

## Added to documentation?
- [x] no documentation needed

exhausted_gif

### DIFF
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -63,6 +63,7 @@ class ArticlesController < ApplicationController
     @version = @article.has_frontmatter? ? "v1" : "v2"
     @user = @article.user
     @organizations = @user&.organizations
+    set_user_approved_liquid_tags
   end
 
   def manage
@@ -189,6 +190,10 @@ class ArticlesController < ApplicationController
     @organizations = @user&.organizations
     @tag = Tag.find_by(name: params[:template])
     @prefill = params[:prefill].to_s.gsub("\\n ", "\n").gsub("\\n", "\n")
+    set_user_approved_liquid_tags
+  end
+
+  def set_user_approved_liquid_tags
     @user_approved_liquid_tags = @user ? @user.roles.where(name: "restricted_liquid_tag").pluck(:resource_type) : []
   end
 
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -18,6 +18,8 @@ class ArticlesController < ApplicationController
     rowspan size span src start strong title value width
   ].freeze
 
+  RESTRICTED_LIQUID_TAGS = [UserSubscriptionTag].freeze
+
   def feed
     skip_authorization
 
@@ -194,7 +196,14 @@ class ArticlesController < ApplicationController
   end
 
   def set_user_approved_liquid_tags
-    @user_approved_liquid_tags = @user ? @user.roles.where(name: "restricted_liquid_tag").pluck(:resource_type) : []
+    @user_approved_liquid_tags =
+      if @user
+        RESTRICTED_LIQUID_TAGS.filter_map do |liquid_tag|
+          liquid_tag if liquid_tag::VALID_ROLES.any? { |role| @user.has_role?(*Array(role)) }
+        end
+      else
+        []
+      end
   end
 
   def handle_user_or_organization_feed
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -189,6 +189,7 @@ class ArticlesController < ApplicationController
     @organizations = @user&.organizations
     @tag = Tag.find_by(name: params[:template])
     @prefill = params[:prefill].to_s.gsub("\\n ", "\n").gsub("\\n", "\n")
+    @user_approved_liquid_tags = @user ? @user.roles.where(name: "restricted_liquid_tag").pluck(:resource_type) : []
   end
 
   def handle_user_or_organization_feed
--- a/app/models/liquid_tags/user_subscription_tag.rb
+++ b/app/models/liquid_tags/user_subscription_tag.rb
@@ -1,0 +1,9 @@
+module LiquidTags
+  class UserSubscriptionTag < ApplicationRecord
+    resourcify
+    # This class exists to take advantage of Rolify for limiting authorization
+    # on internal reports.
+    # NOTE: It is not backed by a database table and should not be expected to
+    # function like a traditional Rails model
+  end
+end
--- a/docs/frontend/liquid-tags.md
+++ b/docs/frontend/liquid-tags.md
@@ -129,8 +129,8 @@ Here we are saying that the `UserSubscriptionTag` is only usable by users with
 the `admin` role or with a role of `:restricted_liquid_tag` and a specified
 resource of `LiquidTags::UserSubscriptionTag`.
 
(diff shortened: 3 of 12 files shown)


app/models/liquid_tags/user_subscription_tag.rb
```diff
@@ -0,0 +1,9 @@
+module LiquidTags
+  class UserSubscriptionTag < ApplicationRecord
```
atsmith813, line 2: `Rolify` only works with model resources 😢 . I chose this setup to avoid naming conflicts and organize these into one directory.

app/controllers/articles_controller.rb
```diff
@@ -189,6 +192,18 @@ def base_editor_assigments
...
+
+  def set_user_approved_liquid_tags
+    @user_approved_liquid_tags =
+      if @user
+        RESTRICTED_LIQUID_TAGS.filter_map do |liquid_tag|
+          liquid_tag if liquid_tag::VALID_ROLES.any? { |role| @user.has_role?(*Array(role)) }
+        end
+      else
+        []
+      end
```
atsmith813, line 206: This is a sort of quick patch so that, in the view, we can check if the user is authorized to use a restricted liquid tag before showing the documentation for it. It's flexible enough that all we have to do for a future restricted tag is add it to the `RESTRICTED_LIQUID_TAGS` constant.

Ideally, this logic probably wouldn't live here. To _really_ get this right I think is out of scope for this PR as it would brush up against other larger refactors. I could see a world where the logic of figuring out which liquid tags are restricted and if the user has authorization for each one, living somewhere else. I could also see us refactoring liquid tag documentation logic out differently as well. Maybe the liquid tag has a method to generate the documentation for itself, given a user, and it would have the power to generate the documentation based on the context it's being used in and the roles of the user.

Either way, this approach seemed like a happy middle ground. I'm open to other ideas! I also considered removing this logic altogether, always showing for the documentation for this liquid tag, and adding a footnote along the lines of:

> This liquid tag can only be used by authorized users.

spec/requests/dashboard_spec.rb
```diff
@@ -314,75 +316,81 @@
...
+    let(:article_with_user_subscription_tag) { create(:article, user: author, with_user_subscription_tag: true) }
+    let(:params) do
+      { source_type: article_with_user_subscription_tag.class.name, source_id: article_with_user_subscription_tag.id }
+    end
+
     before do
-      sign_in user
+      # Stub roles because adding them normally can cause flaky specs
+      allow(author).to receive(:has_role?).and_call_original
+      allow(author).to receive(:has_role?).with(:restricted_liquid_tag,
```
mstruve, line 329: 😍

app/models/liquid_tags/user_subscription_tag.rb
mstruve: I kinda want to set aside these stubs in their own folder. Work for another day though.

docs/frontend/liquid-tags.md
```diff
@@ -113,21 +113,24 @@ https://github.com/thepracticaldev/dev.to/pull/3801
...
 ```
 
-Here we are saying that the `NewLiquidTag` is only usable by users with the
-`admin` role or with a role of `:single_resource_admin` and a specified resource
-of `NewLiquidTag`.
+Here we are saying that the `UserSubscriptionTag` is only usable by users with
+the `admin` role or with a role of `:restricted_liquid_tag` and a specified
+resource of `LiquidTags::UserSubscriptionTag`.
+
+`LiquidTags::UserSubscriptionTag` is a resource model to play nicely with the
```
mstruve, line 132:
```suggestion
`LiquidTags::UserSubscriptionTag` is a resource model so that we can play nicely with the
```

vaidehijoshi:
> I render the documentation for it if the user has the permitted role

Just to make sure I understand this correctly, if a user does not have the `restricted_liquid_tag` role, the documentation should _not_ appear, right?

I just tried this out locally and the docs _do_ appear when my user has the role, but when I destroy the role, it doesn't disappear the way that I would expect. I did a quick test in the UI to check what the values are, but let me know if I'm misunderstanding the expected behavior here :)

Screen Shot 2020-07-08 at 10 17 39 AM

atsmith813:
@vaidehijoshi my mistake! I should've updated the wording on the PR description. It's actually any of the _approved roles for that liquid tag_. That includes `admin`, `super_admin`, and the new role. So you would need to destroy all 3 of those roles 🙈

vaidehijoshi:
@atsmith813 No worries, thanks for clarifying! I went to QA it first, so I hadn't looked at the code yet...and now I see [this](https://github.com/thepracticaldev/dev.to/pull/9170/files#diff-dbaa157d5eab1612b5396d10b80fe577R4-R7), which makes more sense/aligns with your explanation :)

docs/frontend/liquid-tags.md
atsmith813: Good idea - done!

Issue #9116, opened 2020-07-04
Can't create a listing / tags validation fails


kbariotis:
**Describe the bug**

It seems like I can't create a listing due to the tag validation failing. If I leave tags blank and then try to edit, it then allows me to.

**To Reproduce**

1. Go to "Home page"
2. Click on "Create new listing"
3. Try to add a tag
4. See validation error 

**Expected behavior**

To be able to create a listing with tags.

**Screenshots**

![image](https://user-images.githubusercontent.com/605742/86520014-2a93d980-be38-11ea-971f-65e2b7097938.png)

**Desktop (please complete the following information):**

- OS: MacOS
- Browser: Chrome
- Version: 83.0.4103.116

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

Rafi993:
Hi there, If this issue is not taken I would like to help.

rhymes:
@Rafi993 go ahead! thanks!

Rafi993:
> @Rafi993 go ahead! thanks!

Thank you @rhymes

nickytonline:
Just to give a little more info on this. The `pattern` attribute on the input is wrong which is why it fails. It should accept a comma (maybe even whitespace that gets removed on blur or on keyup it just doesn't let a space) in between tags. To make it friendlier, consider taking off the trailing comma if there is one.

So valid use cases:

- someAlphanumericTag1, someAlphanumericTag2, someAlphanumericTag3
- someAlphanumericTag1, someAlphanumericTag2, someAlphanumericTag3,
- someAlphanumericTag1,
- someAlphanumericTag1

The pattern should also allow for n number of tags. Currently it's 8 tags (see placeholder text). This is coming from `maxTags` in tags.jsx.

![image](https://user-images.githubusercontent.com/833231/86551625-95661300-bf13-11ea-8ab7-8cc29653a0c5.png)

jacobherrington:
[...]

PR #9172, opened 2020-07-06
Restrict tags input regex

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This commit updates the regex looking for non-ASCII characters and
also makes it possible to pass a regex as a prop if necessary.

I saw a second issue come in for this bug, so I went ahead and updated the regex to be more explicitly restrictive. It should solve the issue from https://github.com/thepracticaldev/dev.to/issues/8884 for listings and articles without breaking listings.

Fixes https://github.com/thepracticaldev/dev.to/issues/9116
Duplicate https://github.com/thepracticaldev/dev.to/issues/9168

Bug was created in https://github.com/thepracticaldev/dev.to/pull/9006/files

Gif I made was way too big :(

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed

### DIFF
--- a/app/javascript/article-form/components/TagsField.jsx
+++ b/app/javascript/article-form/components/TagsField.jsx
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types';
 import Tags from '../../shared/components/tags';
 
 export const TagsField = ({ defaultValue, onInput, switchHelpContext }) => {
+  const TAG_FORMAT = '[0-9A-Za-z, ]+';
+
   return (
     <div className="crayons-article-form__tagsfield">
       <Tags
@@ -12,6 +14,7 @@ export const TagsField = ({ defaultValue, onInput, switchHelpContext }) => {
         onFocus={switchHelpContext}
         classPrefix="crayons-article-form"
         fieldClassName="crayons-textfield crayons-textfield--ghost ff-accent"
+        pattern={TAG_FORMAT}
       />
     </div>
   );
--- a/app/javascript/article-form/components/TagsField.jsx
+++ b/app/javascript/article-form/components/TagsField.jsx
@@ -2,7 +2,9 @@ import { h } from 'preact';
 import PropTypes from 'prop-types';
 import Tags from '../../shared/components/tags';
 
-export const TagsField = ({ defaultValue, onInput, switchHelpContext }) => {
+export const DEFAULT_TAG_FORMAT = '[0-9A-Za-z, ]+';
+
+export const TagsField = ({ defaultValue, onInput, switchHelpContext, tagFormat = DEFAULT_TAG_FORMAT }) => {
   const TAG_FORMAT = '[0-9A-Za-z, ]+';
 
   return (
--- a/app/javascript/listings/listingForm.jsx
+++ b/app/javascript/listings/listingForm.jsx
@@ -61,6 +61,8 @@ export default class ListingForm extends Component {
       expireDate,
     } = this.state;
 
+    const TAG_FORMAT = '[0-9A-Za-z, ]+';
+
     const selectOrg =
       organizations && organizations.length > 0 ? (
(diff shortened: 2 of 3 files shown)


app/javascript/shared/components/tags.jsx
```diff
@@ -374,6 +374,7 @@ class Tags extends Component {
       listing,
       fieldClassName,
       onFocus,
+      pattern,
```
jacobherrington, line 377: This might be unnecessary duplication, but I think having finer-grain control might be good (based on the last attempt working for articles but not listings). Happy to take this out and include the pattern only in the component if this seems like too much.

app/javascript/article-form/components/TagsField.jsx
```diff
@@ -12,6 +12,7 @@ export const TagsField = ({ defaultValue, onInput, switchHelpContext }) => {
         onFocus={switchHelpContext}
         classPrefix="crayons-article-form"
         fieldClassName="crayons-textfield crayons-textfield--ghost ff-accent"
+        pattern="[\x30-\x39\x41-\x5A\x61-\x7A\x2C\x20]+"
```
rhymes, line 15: do you think we could save this pattern in a self explanatory constant? Or at least list in a comment which chars are restricted by these. Same for the other change

jacobherrington: @rhymes I updated it to a more readable format, let me know what you think! I'm not sure why I used the hex values in retrospect :sweat_smile:

juliannatetreault:
On the issue you linked to in your PR description, #9116, Nick gave some context around the issue and what would need to be fixed moving forward. In his comment, he mentioned that the pattern should allow for "n number of tags", but it is limited to 8 tags currently: "'The pattern should also allow for n number of tags. Currently it's 8 tags (see placeholder text). This is coming from maxTags in tags.jsx.'" I just wanted to see whether or not allowing for more than 8 tags per listing was in the scope of this PR and if you were planning on addressing it here or not, as I tested these changes locally and everything worked as expected, but I was still hit with the 8-or-less-tag-limit validation. 😄

app/javascript/article-form/components/TagsField.jsx
rhymes: ahhaha that's fine too!

app/javascript/article-form/components/TagsField.jsx
```diff
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types';
 import Tags from '../../shared/components/tags';
 
 export const TagsField = ({ defaultValue, onInput, switchHelpContext }) => {
```
nickytonline, line 5:
```suggestion
export const DEFAULT_TAG_FORMAT = '[0-9A-Za-z, ]+';

export const TagsField = ({ defaultValue, onInput, switchHelpContext, tagFormat = DEFAULT_TAG_FORMAT }) => {
```

app/javascript/listings/listingForm.jsx
```diff
@@ -61,6 +61,8 @@ export default class ListingForm extends Component {
       expireDate,
     } = this.state;
 
+    const TAG_FORMAT = '[0-9A-Za-z, ]+';
```
nickytonline, line 64: Import this from the suggestion above so it's not duplicated.

nickytonline:
We'll do the requested changes later as this is blocking listings.

jacobherrington:
This is not working, I committed Nick's suggestions via the GitHub UI about 20 minutes before this was merged and some changes needed to be made on the branch to support that suggestion.

I'll push a branch up in a minute that will contain that code.

PR #9182, opened 2020-07-07
Fix sign in flow for user subscription liquid tag

atsmith813:
- [x] Bug Fix

## Description
When I re-did the JS in #9082 I accidentally messed up the sign in flow. This PR fixes that bug and adds a spec to check for this case.

3. Publish your Article.
4. Sign out.
5. View your Article.
6. Click sign in on the liquid tag - you should be directed to sign in 🎉 .

#9082 

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

blinds_a_mess_gif

### DIFF
--- a/app/liquid_tags/user_subscription_tag.rb
+++ b/app/liquid_tags/user_subscription_tag.rb
@@ -147,9 +147,11 @@ class UserSubscriptionTag < LiquidTagBase
     // ***************************************
     function addSignInClickHandler() {
       const signInBtn = document.getElementById('sign-in-btn');
-      if (signInBtn && typeof showModal !== 'undefined') {
+      if (signInBtn) {
         signInBtn.addEventListener('click', function(e) {
-          showModal('email_signup');
+          if (typeof showModal !== 'undefined') {
+            showModal('user_subscription');
+          }
         });
       }
     }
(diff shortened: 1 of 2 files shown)


app/liquid_tags/user_subscription_tag.rb
```diff
@@ -147,9 +147,11 @@ class UserSubscriptionTag < LiquidTagBase
     // ***************************************
     function addSignInClickHandler() {
       const signInBtn = document.getElementById('sign-in-btn');
-      if (signInBtn && typeof showModal !== 'undefined') {
+      if (signInBtn) {
         signInBtn.addEventListener('click', function(e) {
-          showModal('email_signup');
+          if (typeof showModal !== 'undefined') {
+            showModal('user_subscription');
+          }
```
Zhao-Andy, line 154: Huh, I guess this is because of some async stuff going on? I would think the original would work but I guess not...

atsmith813: Right!? I thought the exact same thing 🙈 . The original kept failing because `showModal` was undefined.

I think changing it back to this way works because it's not checking for `showModal` until "click" and by that point, everything is loaded.

Zhao-Andy: Ahh that makes a lot of sense!

PR #9054, opened 2020-07-01
Overview of my own series #6938

Oghenebrume50:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR attempts to address the issue #6938 this user proposed a series overview on the dashboard, I had come up with a workable solution, so I opened this draft PR so I can get more suggestion on what exactly is needed for this to be perfect

Closes #6938

## QA Instructions, Screenshots, Recordings
This is how it shows when the dashboard is opened
series1

This is how it shows when the series is clicked on, basically just list all series if any, clicking on any one links to the first article in the series
series2

since all other articles in the articles are shown here I think this works.
series3

_Please replace this line with instructions on how to test your changes, as well
as any relevant images for UI changes._

## Added tests?

- [ ] yes
- [ ] no, because they aren't needed
- [x] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

PR #9201, opened 2020-07-08
Add index and show pages for series

### LINKED ISSUE #2424: Landing pages for series
It would be cool if series posts could have a landing page where all posts in the series displayed. It could look like the main feed, with just those series posts, and the [individual posts themselves](https://dev.to/aspittel/moving-past-tutorials-receiving-a-problem-to-solve-2hf8) could have the series name clickable - in that example, the **Part of "moving past tutorials" series** text would link to the feed of series posts.

AlbertoPdRF:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I'm opening this as a draft PR as there's probably much work left to do. I think it would be really nice to get some feedback on what I've been doing to see if I'm going in the right direction!

Just one question for now: on the index page, I'm getting the following warning:

```
See 'Uniform Notifier' in JS Console for Stacktrace
user: apdrf Need Counter Cache Collection => [:articles]
```

Any recommendation on how to deal with it? I guess it's worth noting that I haven't taken care of any caching (nor pagination or other stuff for that matter), I've just went with the very basics.

Fixes #2424.

Index page

Show page

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Fill the `path` for all the existing collections!

Excited GIF

CC: @rhymes @ludwiczakpawel @jacobherrington @Oghenebrume50


app/controllers/articles_controller.rb
```diff
@@ -109,7 +109,8 @@ def preview
   def create
     authorize Article
 
-    article = Articles::Creator.call(current_user, article_params_json)
+    @user = current_user
+    article = Articles::Creator.call(@user, article_params_json)
```
rhymes, line 113: I think this change slipped in, there's no need to store in `@user`

app/controllers/collections_controller.rb
```diff
@@ -0,0 +1,13 @@
+class CollectionsController < ApplicationController
+  def index
+    if (@user = User.find_by(username: params[:username]))
+      @collections = Collection.where(user_id: @user.id)
```
rhymes, line 4: this can be: `@user.collections`

app/controllers/collections_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+    elsif (@user = Organization.find_by(slug: params[:username]))
+      @collections = Collection.where(organization_id: @user.id)
+    end
```
rhymes, line 7: Although we do have the concept of "series optionally belonging to an organization" there's currently no code path in which an organization can create a series, so I would probably drop this altogether and just load: `user.collections` for now

app/controllers/collections_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  def show
+    @collection = Collection.includes(%i[user articles]).find_by(path: "/#{params[:username].downcase}/series/#{params[:slug]}")
```
rhymes, line 11: I don't think we need to store the path, as storing the path would mean we'd need to keep it updated in case the user changes username or if the slug changes somehow and so on, in addition to having to populate the field for existing collections.

Though it would result in one more query I don't think that's a huge deal:

- load the user through the username: `user = User.find_by!()`
- load the collection through `user.collections.find_by!()`

app/views/collections/index.html.erb
```diff
@@ -0,0 +1,53 @@
...
+  <meta property="og:image" content="<%= SiteConfig.main_social_image %>" />
+  <meta property="og:description" content="<%= community_qualified_name %> is great!" />
+  <meta property="og:site_name" content="<%= community_qualified_name %>" />
+
+  <meta name="twitter:card" content="summary_large_image">
+  <meta name="twitter:site" content="@<%= SiteConfig.social_media_handles["twitter"] %>">
+  <meta name="twitter:title" content="About <%= community_name %>">
+  <meta name="twitter:description" content="<%= community_qualified_name %> is great!">
+  <meta name="twitter:image:src" content="<%= SiteConfig.main_social_image %>">
+<% end %>
```
rhymes, line 21: could this be put in a partial for both index and show pages?

app/views/collections/show.html.erb
```diff
@@ -0,0 +1,43 @@
...
+
+  <meta name="twitter:card" content="summary_large_image">
+  <meta name="twitter:site" content="@<%= SiteConfig.social_media_handles["twitter"] %>">
+  <meta name="twitter:title" content="About <%= community_name %>">
+  <meta name="twitter:description" content="<%= community_qualified_name %> is great!">
+  <meta name="twitter:image:src" content="<%= SiteConfig.main_social_image %>">
+<% end %>
+
+<% user = @collection.user %>
+<% articles = @collection.articles %>
```
rhymes, line 24: they should probably be ordered by `published_at`

app/views/collections/show.html.erb
```diff
@@ -0,0 +1,43 @@
...
+  </header>
+
+  <div class="articles-list" id="articles-list">
+    <div class="substories" id="substories">
+      <% if articles.present? %>
+        <% articles.each do |article| %>
+          <%= render "articles/single_story", story: article.decorate, featured: article.main_image.present? %>
+        <% end %>
+      <% else %>
+        <p>This series doesn't have any article</p>
```
rhymes, line 39: typo `articles`

app/controllers/collections_controller.rb
rhymes: I would also order them by descending creation order, or alphabetical title, you choose :D

app/controllers/articles_controller.rb
AlbertoPdRF: I do think this is necessary, as `article_params_json` uses `@user`, but I could be wrong.

app/controllers/collections_controller.rb
AlbertoPdRF: Sounds good! And descending creation order it will be 🙂

app/controllers/collections_controller.rb
AlbertoPdRF: Ok, will do!

app/views/collections/index.html.erb
AlbertoPdRF: Definitely!

app/controllers/collections_controller.rb
AlbertoPdRF: This in general sounds like a good way to go, but we have the problem of the slug not being a real slug again. Any clever idea on how to proceed? I see there's a `titleize` method but it doesn't quite work for our case.

app/views/collections/show.html.erb
AlbertoPdRF: Indeed!

app/controllers/collections_controller.rb
AlbertoPdRF: For now I'm using the slug as is, although this produces ugly URLs, so it should be changed.

rhymes:
Marking it as ready for review as it might not register with other reviewers being in draft

app/views/articles/_collection.html.erb
```diff
@@ -2,12 +2,18 @@
...
 <% if collection && collection_size > 1 %>
+  <% size_string = "#{collection_size} Part Series" %>
   <div class="article-collection-wrapper article-collection-wrapper-<%= position %>">
-    <% if collection.slug.present? %>
-      <p><%= collection.slug %> (<%= collection_size %> Part Series)</p>
-    <% else %>
-      <p><%= collection_size %> Part Series</p>
-    <% end %>
+    <p>
+      <a href="<%= collection.path %>">
```
citizen428, line 8: **nitpick (non-blocking):** We're using this in two places, maybe we can wrap it up in a helper, which would also remove some local variables (`size_string`) from views. Something like:

```ruby
def collection_link(collection, options = {})
  collection_size = ...
  size_string = ...
  body = if collection.slug.present? ...

  link_to body, collection.path, options
end
```

AlbertoPdRF:
> I wonder if we should switch to IDs for the series for the time being, any ideas @jacobherrington @citizen428 @Ridhwana ?
> 
> To go from `http://localhost:3000/rhymes/series/A%20series` to `http://localhost:3000/rhymes/series/2`

I've just pushed a commit that does that. One good thing about the change is that we are back to one query to get the collection! I'll get back to the rest of your comment later.

app/views/articles/_collection.html.erb
AlbertoPdRF: The way you suggest looks cleaner, I'm pushing a commit with the change!

AlbertoPdRF:
> I noticed something odd:
> 
> 1. write a draft post with a series
> 2. go to http://localhost:3000/:username/series
> 3. it says "1 post series"
> 4. if you click no posts are shown as I think it only shows published posts

This is because I was counting the size of the collection without specifying if the articles had been published or not. I thought this was the way to go because of [this method](https://github.com/forem/forem/blob/master/app/helpers/articles_helper.rb#L20-L26), but as the articles passed to this method are always published ([this is were they come from](https://github.com/forem/forem/blob/949b092dbd614ffa6a68026b72c4e89fbe1e2190/app/controllers/stories_controller.rb#L291-L293)) I've decided to clean up the code a bit and also fix the odd behaviour you pointed out. I'm pushing now two commits with the changes, although maybe the cleaning up one should go in a separate PR -- let me know what you think of this!

AlbertoPdRF:
> [CLA assistant check](https://cla-assistant.io/forem/forem?pullRequest=9201)
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/forem/forem?pullRequest=9201) before we can accept your contribution.
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/forem/forem?pullRequest=9201) it.

Argh, the commits got signed with a different email address :confounded:

AlbertoPdRF:
> Looks nice, one question, how does a user get to the page?

Right now they can get to the show page clicking on the title of a series (from an article), and to the index page by clicking on the back button of the show page. Where would you add more links (apart from #9054)?

rhymes:
@AlbertoPdRF that's fine, I temporarily forgot about https://github.com/forem/forem/pull/9054 :D

AlbertoPdRF:
> > [CLA assistant check](https://cla-assistant.io/forem/forem?pullRequest=9201)
> > Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/forem/forem?pullRequest=9201) before we can accept your contribution.
> > You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/forem/forem?pullRequest=9201) it.
> 
> Argh, the commits got signed with a different email address

By the way, I'll need to force-push to amend this, should I do it now or after the PR is considered finished?

rhymes:
@AlbertoPdRF do it now, no worries :)

Oghenebrume50:
> @AlbertoPdRF that's fine, I temporarily forgot about #9054 :D

I guess once this is merged I can polish #9054 and open it as an actual PR

AlbertoPdRF:
Are the build fails I'm getting due to some flaky tests?

ludwiczakpawel:
From design perspective this obviously needs touch of love but overall it works so I'm happy to ship it as it is for sake of delivering a feature and we can then circle back with another PR to make some design improvements.

Really good work @AlbertoPdRF !

rhymes:
> Are the build fails I'm getting due to some flaky tests?

unfortunately yes :(

AlbertoPdRF:
I've added some tests, please let me know if any should be modified or more should be added! I'll leave an inline comment regarding one of the tests now, because I can't quite get what's wrong with it.

spec/system/collections/user_views_collections_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  let(:collection2) { create(:collection, user: user) }
+
+  before do
+    sign_in user
+    visit user_series_path(user.username)
+  end
+
+  it "shows all collections" do
+    [collection1, collection2].each do |collection|
+      expect(page.body).to have_link("#{collection.slug} (#{collection.articles.published.size} Part Series)")
```
AlbertoPdRF, line 15: This expectation fails sometimes and I can't really get why 😖 I tried to debug this using `byebug`, but when I check for `user.collections.count` it returns `2` and the test never fails in that case, so I don't know how to proceed.

citizen428: Do you have an example message from one of the failed runs?

AlbertoPdRF: One of the builds actually failed for that, so you can [see the message here](https://travis-ci.com/github/forem/forem/jobs/364361396#L2148-L2167).

citizen428: Does `collection2` also need articles? You have `:with_articles` on `collection1` but not `collection2`, so its size is probably 0. That's just a guess though, I didn't look into anything in detail.

AlbertoPdRF: Hmmm, having it without articles was intentional, as the link should be present (and correct) for a collection without articles too.

citizen428: @AlbertoPdRF I had some time to look into it this morning. Here's how I debugged it in case that's useful to you in the future:

1. Add `:aggregate_failures` to the spec so we can see the results for each run.
    ```diff
    - it "shows all collections" do
    + it "shows all collections", :aggregate_failures do
    ```
    This revealed that both assertions are failing, not only the one for `collection2`.
2. Look at the page by adding `save_and_open_page` at the end of the spec. This showed the following:
    Screen Shot 2020-07-27 at 10 25 30
3. Grepped for the message, found it in `app/views/collections/index.html.erb`. It will be rendered when `@collections.present?` is `false`.
4. Set a `binding.pry` in `app/controllers/collections_controller.rb`. `user.collections` was an empty array.
5. Changed the spec to use `let!` over `let`, green now. See e.g. https://mixandgo.com/learn/let-vs-instance-variables-in-rspec

If you want to try it out, I made a patch (with `git format-patch`) but for whatever reason GH won't let me attach it:

```
From d3d169d9e38e28066802fdc21a1faf1ae95e98d8 Mon Sep 17 00:00:00 2001
From: Michael Kohl <citizen428@dev.to>
Date: Mon, 27 Jul 2020 10:38:55 +0700
Subject: [PATCH 8/8] Fix spec

---
 spec/system/collections/user_views_collections_spec.rb | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

diff --git a/spec/system/collections/user_views_collections_spec.rb b/spec/system/collections/user_views_collections_spec.rb
index c664a0892..e078e5409 100644
--- a/spec/system/collections/user_views_collections_spec.rb
+++ b/spec/system/collections/user_views_collections_spec.rb
@@ -2,15 +2,15 @@ require "rails_helper"

 RSpec.describe "Visiting collections", type: :system do
   let(:user) { create(:user) }
-  let(:collection1) { create(:collection, :with_articles, user: user) }
-  let(:collection2) { create(:collection, user: user) }
+  let!(:collection1) { create(:collection, :with_articles, user: user) }
+  let!(:collection2) { create(:collection, user: user) }

   before do
     sign_in user
     visit user_series_path(user.username)
   end

-  it "shows all collections" do
+  it "shows all collections", :aggregate_failures do
     [collection1, collection2].each do |collection|
       expect(page.body).to have_link("#{collection.slug} (#{collection.articles.published.size} Part Series)")
     end
--
2.27.0
```

You can save this to your working directory and apply it with `git apply <filename>`.

AlbertoPdRF: That's indeed very useful @citizen428! Thank you for looking into this and for the very clear explanation 😊

rhymes:
Closing and reopening it to see if Travis gets unstucked

app/helpers/application_helper.rb
```diff
@@ -193,6 +193,13 @@ def copyright_notice
     "#{start_year} - #{current_year}"
   end
 
+  def collection_link(collection, options = {})
```
rhymes, line 196: you should be able to fix the Codeclimate violation by using `**kwargs` here and at the last line:

`link_to(body, collection.path, **kwags)`

AlbertoPdRF: Ok, changing this! :slightly_smiling_face:

PR #9204, opened 2020-07-08
Fix Style/SingleLineBlockParams, RSpec/MultipleExpectations and remaining Layout/LineLength

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Instead of ignoring `RSpec/MultipleExpectations` which by default is set to 1 `expect()` per test case (with 930 violations) I set it to a more than reasonable 8 (mostly by process of elimination :D) and refactored a couple of tests that had 12 `expect()`.

In general I think we should work towards getting rid of the `.rubocop_todo.yml` over inline `rubocop:disable` statements as they are visible during the process of creating code. The todo file is likely to be overlooked most of the times.

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -446,3 +446,9 @@ RSpec/ExampleLength:
   Enabled: true
   Max: 15
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ExampleLength
+
+RSpec/MultipleExpectations:
+  Description: Checks if examples contain too many `expect` calls.
+  Enabled: true
+  Max: 8
+  StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/MultipleExpectations
--- a/spec/lib/data_update_scripts/re_index_feed_content_and_users_to_elasticsearch_spec.rb
+++ b/spec/lib/data_update_scripts/re_index_feed_content_and_users_to_elasticsearch_spec.rb
@@ -1,4 +1,3 @@
-# rubocop:disable RSpec/ExampleLength
 require "rails_helper"
 require Rails.root.join("lib/data_update_scripts/20200519142908_re_index_feed_content_and_users_to_elasticsearch.rb")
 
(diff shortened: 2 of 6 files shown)


spec/lib/data_update_scripts/re_index_feed_content_and_users_to_elasticsearch_spec.rb
```diff
@@ -8,27 +7,49 @@
...
+  it "indexes users to Elasticsearch" do
+    user = create(:user)
+
+    expect { user.elasticsearch_doc }.to raise_error(Search::Errors::Transport::NotFound)
+
+    sidekiq_perform_enqueued_jobs { described_class.new.run }
+
     expect(user.elasticsearch_doc).not_to be_nil
 
-    expect(article.elasticsearch_doc["_source"].keys).to include "public_reactions_count"
```
mstruve, line 28: Since this is going to increase the run time of this spec, could we instead use `aggregate_failures` here to keep all the Elasticsearch refreshing etc to a single time?

rhymes: Got it. I'm going to put it back as it were then and add `aggregate_failures`

.rubocop.yml
```diff
@@ -446,3 +446,9 @@ RSpec/ExampleLength:
   Enabled: true
   Max: 15
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ExampleLength
+
+RSpec/MultipleExpectations:
+  Description: Checks if examples contain too many `expect` calls.
```
mstruve, line 451: Seeing as we have the `aggregate_failures` ability, I am worried that this might cause people to end up writing very long-running specs bc they limit even heavy lifting specs to a single assertion. I'm wondering if we might be better off without this cop. Thoughts?

rhymes: The cop is going to activate after 8 `expect()` statements which I think it'd be a reasonable number to have in a single spec. If not we can raise it? I understsand the point about speed and I agree but there's also to consider the reason why this cop even exists: too many responsibilities tested in the same test.

What do you think?

mstruve: OH the threshold is 8, well that is much better, I assumed it was 1 🤦‍♀️

PR #9205, opened 2020-07-08
Refactor: Remove Omniauth Error Logs, Send to Honeybadger

mstruve:
- [x] Refactor

## Description
In an effort to remove PII information from our logs so that we can allow the entire core team to access that data for debugging purposes I have moved these error logs to Honeybadger.

# Related Ticket
https://github.com/thepracticaldev/dev.to/projects/9#card-33250889

alt_text

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -72,18 +72,20 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
       user_errors = @user.errors.full_messages
 
-      Rails.logger.error("Log in error: sign in failed. username: #{@user.username} - email: #{@user.email}")
-      Rails.logger.error("Log in error: auth data hash - #{request.env['omniauth.auth']}")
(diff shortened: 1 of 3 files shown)


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -72,18 +72,20 @@ def callback_for(provider)
...
+      Honeybadger.notify("Omniauth log in error")
 
       flash[:alert] = user_errors
       redirect_to new_user_registration_url
     end
   rescue StandardError => e
-    Rails.logger.error("Log in error: #{e}")
-    Rails.logger.error("Log in error: auth data hash - #{request.env['omniauth.auth']}")
-    Rails.logger.error("Log in error: auth data hash - #{request.env['omniauth.error']&.inspect}")
+    Honeybadger.notify(e)
```
mstruve, line 88: The error will have all of the request and env info when we send it. I am not sure if it would above when we send a  string as an error so I added it to the context to be safe.

PR #9221, opened 2020-07-09
Rubocop: fix Metrics/CyclomaticComplexity and disable Metrics/AbcSize

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR:

- fixes `Metrics/CyclomaticComplexity`
- refactors a bit `Articles::Builder` to use `.call` and to fix `Metrics/CyclomaticComplexity`
- disables `Metrics/AbcSize` as we currently have 278 violations and I believe there's no point into keeping this particular violation activated or to find the threshold to hide some of them "under the rug"

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/app/services/articles/builder.rb
+++ b/app/services/articles/builder.rb
@@ -8,27 +8,34 @@ module Articles
       @prefill = prefill
     end
 
-    def build
-      if @tag.present? && @user&.editor_version == "v2"
+    def self.call(*args)
+      new(*args).call
+    end
 
+    # rubocop:disable Metrics/CyclomaticComplexity
+    def call
+      if @tag.present? && @user&.editor_version == "v2"
         [tag_user_editor_v2, needs_authorization]
       elsif @tag&.submission_template.present? && @user
-
         [tag_user, needs_authorization]
       elsif @prefill.present? && @user&.editor_version == "v2"
-
         [prefill_user_editor_v2, needs_authorization]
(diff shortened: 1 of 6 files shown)


app/services/articles/builder.rb
```diff
@@ -6,87 +6,79 @@ def initialize(user, tag, prefill)
...
-      true
+    def self.call(*args)
+      new(*args).call
     end
 
-    def does_not_need_authorization
-      false
+    # the Builder returns a pair of [article, needs_authorization?]
+    # => `needs_authorization? can be either true or false
+    def call
```
mstruve, line 19: Whew this is a trip!

app/services/articles/builder.rb
```diff
@@ -6,87 +6,79 @@ def initialize(user, tag, prefill)
...
+      @editor_version2 = @user&.editor_version == "v2"
     end
 
-    def needs_authorization
-      true
+    def self.call(*args)
+      new(*args).call
     end
 
-    def does_not_need_authorization
```
mstruve, line 38: 🔪

app/services/articles/builder.rb
rhymes: yeah, I still don't fully understand it :D

PR #9222, opened 2020-07-09
Bug Fix: Remove Rails Validation for Follow Uniqueness, Guard against dup Follow Jobs

mstruve:
- [x] Bug Fix

## Description
We have been seeing a lot of ActiveRecord validation errors when a record tries to follow a record they are already following. We attempted to prevent this using `create_or_find_by!`, however we were still seeing the error. The reason, and you all can thank @rhymes for figuring this out is
> create_or_find_by! rescues ActiveRecord::RecordNotUnique which is the exception the database sends up through its constraints, but we have “mounted” on top of it a second layer which is Rails’s, through ActiveRecord::RecordInvalid which is not the exception create_or_find_by! handles. 
```ruby
    def create_or_find_by!(attributes, &block)
      transaction(requires_new: true) { create!(attributes, &block) }
    rescue ActiveRecord::RecordNotUnique
      find_by!(attributes)
    end
```

## Added tests?

alt_text

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -243,7 +243,7 @@ class UsersController < ApplicationController
     return unless user.looking_for_work?
 
     hiring_tag = Tag.find_by(name: "hiring")
-    return if user.following?(hiring_tag)
+    return if !hiring_tag || user.following?(hiring_tag)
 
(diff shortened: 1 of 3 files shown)


rhymes:
@Zhao-Andy the unique index is the glue holding all of this together :D

PR #9224, opened 2020-07-09
Refactor:Use if/else When Selecting Production Log Strategy

mstruve:
- [x] Refactor

## Description
**I am close to attempting to switch us over to Datadog for logging.** In order to ensure that we can do it smoothly I rewrote the production logging set up so we only configure it once for `RAILS_LOG_TO_STDOUT` and once if we are using Timber, `SEND_LOGS_TO_TIMBER`.

Next Steps
1) Set `RAILS_LOG_TO_STDOUT` to true 
2) [Add a log drain to Heroku to send logs to Datadog](https://docs.datadoghq.com/logs/guide/collect-heroku-logs/). NOTE: This feature is still in beta so if we do run into problems then I will go the fallback route.
- Fallback Plan: [Follow these directions to use the buildpack to send logs to Datadog](https://www.ironin.it/blog/sending-logs-from-your-heroku-app-to-datadog.html)

https://github.com/thepracticaldev/dev.to/projects/9#card-33250889

alt_text

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -94,25 +94,23 @@ Rails.application.configure do
   ]
 
   # Use default logging formatter so that PID and timestamp are not suppressed.
-  # config.log_formatter = ::Logger::Formatter.new
   config.log_formatter = ::Logger::Formatter.new
 
-  # Use a different logger for distributed setups.
-  # require 'syslog/logger'
-  # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new 'app-name')
-
   if ENV["RAILS_LOG_TO_STDOUT"].present?
+    # Use a different logger for distributed setups.
+    # require 'syslog/logger'
+    # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new 'app-name')
+
     logger           = ActiveSupport::Logger.new(STDOUT)
     logger.formatter = config.log_formatter
     config.logger    = ActiveSupport::TaggedLogging.new(logger)
+  elsif (ENV["SEND_LOGS_TO_TIMBER"] || "true") == "true"
+    # Timber.io logger
+    log_device = Timber::LogDevices::HTTP.new(ENV["TIMBER"])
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -94,26 +94,24 @@
...
-
   if ENV["RAILS_LOG_TO_STDOUT"].present?
+    # Use a different logger for distributed setups.
+    # require 'syslog/logger'
+    # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new 'app-name')
+
     logger           = ActiveSupport::Logger.new(STDOUT)
     logger.formatter = config.log_formatter
     config.logger    = ActiveSupport::TaggedLogging.new(logger)
+  elsif (ENV["SEND_LOGS_TO_TIMBER"] || "true") == "true"
```
atsmith813, line 107: **question:** Won't this always end up equating to true with `|| "true" == "true"`?

atsmith813: JK, is it Friday yet? 😅 

That would only be the case if `ENV["SEND_LOGS_TO_TIMBER"]` were `nil`.

mstruve: LOL This comment just made my day....btw I am right there with you https://github.com/thepracticaldev/dev.to/pull/9204#discussion_r452302134 😅

atsmith813: LOL.

Do this mean the scenario where `RAILS_LOG_TO_STDOUT` isn't set (skips first conditional of `if`) and `SEND_LOGS_TO_TIMBER` is `"false"` (skips second conditional of `elsif`) should never happen? In that case, neither of these conditionals would be hit and we don't have an `else` here.

mstruve: That should never happen but I think Rails might default to STDOUT in that case. This little case statement will be removed as soon as we make the switch and confirm everything is working

atsmith813: Ahhhh I see - thank you!

PR #9225, opened 2020-07-09
Add number_with_delimiter to post count on tags page

atsmith813:
- [x] Optimization

## Description
Format large numbers with commas on the tags page https://dev.to/tags

**Before**
Screen Shot 2020-07-09 at 1 03 02 PM

**After**
Screen Shot 2020-07-09 at 1 02 36 PM

## Added tests?

## Added to documentation?
- [x] no documentation needed

no_biggy_gif

### DIFF
--- a/app/views/tags/index.html.erb
+++ b/app/views/tags/index.html.erb
@@ -37,7 +37,7 @@
           <p class="mb-2 color-base-70 truncate-at-3"><%= sanitize tag.short_summary %></p>
         <% end %>
         <p class="mb-3 fs-s color-base-60">
-          <%= pluralize tag.taggings_count, "post" %> published
+          <%= pluralize(number_with_delimiter(tag.taggings_count), "post") %> published
           <% if tag.sponsorship && tag.sponsorship.status == "live" && tag.sponsorship.expires_at > Time.current %>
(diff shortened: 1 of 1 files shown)


atsmith813:
@Zhao-Andy I believe so because it defaults to [defaults to current locale](https://api.rubyonrails.org/classes/ActionView/Helpers/NumberHelper.html#method-i-number_with_delimiter). Do you know if that is sufficient? Or should we be passing an the option to override it instead?

Zhao-Andy:
Ahh yeah I think that's fine then! Missed that part in the description.

rhymes:
Awesome @atsmith813 It's a great idea.

Though as it's done server side it's going to use the server's locale which means that in my case, even though I know we're taking about thousands, I'm instinctively `23,861 post published` as "23 posts point 861" as in in my locale that would be written as `23.861` :-D 

If we want to do it according to the browser locale we should use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

atsmith813:
Ahhhh I see - makes sense. Thanks, @rhymes! TIL. I'll revert this PR and we can circle back in the future and use JS instead.

Issue #8861, opened 2020-06-23
Add title text to broadcasts


vaidehijoshi:
We need some way of differentiating the main title text of a broadcast from the `processed_html` in order to render the title differently from the basic text in the new welcome notification design.

See [this related issue](https://github.com/thepracticaldev/dev.to/issues/7169) for new welcome notification designs.

vaidehijoshi:
I'm not sure when we're going to tackle this redesign, but in an effort to clean up outstanding issues before Hacktoberfest, I'm going to close this.

Issue #7169, opened 2020-04-08
Implement new styles for welcome notifications


vaidehijoshi:
Now that welcome notifications are live, we can focus on making them prettier! Here is the working design by @lisasy:

![image](https://user-images.githubusercontent.com/6921610/78841602-a5bd2f00-79b2-11ea-8704-f0d652b3956a.png)

Some additional links:
* [Crayons documentation](https://dev.to/page/crayons)
* [Figma design](https://www.figma.com/file/TtB4hJxqTE26T3aUCSvjhk/Notifications?node-id=294%3A1)

vaidehijoshi:
To our awesome contributors, please note that this issue will be worked on by a DEV team member and is not up for grabs.

vaidehijoshi:
I've been thinking about this a bit more, and I think it really ought to be closely tied to #6990.

I imagine that we want to admins to be able to configure some of this; for example, we probably want to have a `CTA text` associated with a broadcast that allows admins to configure this on a per-broadcast basis.

![78841602-a5bd2f00-79b2-11ea-8704-f0d652b3956a](https://user-images.githubusercontent.com/6921610/80159981-41d56180-8581-11ea-9529-cb21739ee3ae.png)

vaidehijoshi:
I'm not sure when we're going to tackle this redesign, but in an effort to clean up outstanding issues before Hacktoberfest, I'm going to close this.

PR #9231, opened 2020-07-09
Add Polymorphism to Broadcasts (Part 1)

juliannatetreault:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR is part one of a two-part refactor that creates a polymorphic association between Broadcasts, Announcements, and Welcome Notifications. This PR adds polymorphism to the Broadcasts table and a `data_update` script to properly backfill the newly added `broadcastable_type` column, as shown below. Additionally, this PR adds an Announcement and Welcome Notification table to the database with `created_at` and `updated_at` timestamps on both tables and an additional `banner_style` column on the Announcements table. It also stubs out the Announcement and Welcome Notification models in preparation for the next PR, Part 2 of the move to polymorphic Broadcasts.

Related to #7169 and #8861 

To QA these changes, you can manually run the `data_update` script (as shown below) and make sure that the build passes and that these changes do not break anything else.

To manually run the `data_update` script, do the following:
- Create a Broadcast without a `broadcastable_type`:
```
broadcast = Broadcast.create(title: "Hello, world!", processed_html: "Hello, world!", type_of: "Welcome", active: false, broadcastable_type: nil)
```
```
=> #<Broadcast:0x00007fa982de3c98
 id: 17,
 active: false,
 active_status_updated_at: nil,
 banner_style: nil,
 body_markdown: nil,
 created_at: Thu, 09 Jul 2020 20:22:31 UTC +00:00,
 processed_html: "Hello, world!",
 title: "Hello, world!",
 type_of: "Welcome",
 updated_at: Thu, 09 Jul 2020 20:22:31 UTC +00:00,
 broadcastable_id: nil,
 broadcastable_type: nil>
```
- Run the code within the `data_update` script in your console:
```
  Broadcast.where(broadcastable_type: nil).each do |cast|
     cast.update(broadcastable_type: broadcast.type_of)
   end
```
- Check to see that the Broadcast's `broadcastable_type` column is no longer `nil` and contains the correct `broadcastable_type`:
```
=> [#<Broadcast:0x00007fa982d09340
  id: 17,
  active: false,
  active_status_updated_at: nil,
  banner_style: nil,
  body_markdown: nil,
  created_at: Thu, 09 Jul 2020 20:22:31 UTC +00:00,
  processed_html: "Hello, world!",
  title: "Hello, world!",
  type_of: "Welcome",
  updated_at: Thu, 09 Jul 2020 20:23:45 UTC +00:00,
  broadcastable_id: nil,
  broadcastable_type: "Welcome">]
```

## Added tests?

- [ ] yes
- [ ] no, because I need help
- [x] no, because the necessary tests will be included in the following PR, Polymorphism Part 2.

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

There is a `data_update` script included in this PR to port over the data from the `type_of` column to the `broadcastable_type` column on Broadcasts.

Homer Simpson

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_07_06_184804) do
+ActiveRecord::Schema.define(version: 2020_07_07_173524) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -293,6 +293,8 @@ ActiveRecord::Schema.define(version: 2020_07_06_184804) do
     t.datetime "active_status_updated_at"
     t.string "banner_style"
     t.text "body_markdown"
+    t.integer "broadcastable_id"
+    t.string "broadcastable_type"
     t.datetime "created_at"
     t.text "processed_html"
     t.string "title"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -58,6 +58,12 @@ ActiveRecord::Schema.define(version: 2020_07_07_173524) do
     t.index ["visit_token"], name: "index_ahoy_visits_on_visit_token", unique: true
   end
 
+  create_table "announcements", force: :cascade do |t|
+    t.string "banner_style"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+  end
+
   create_table "api_secrets", force: :cascade do |t|
     t.datetime "created_at", null: false
     t.string "description", null: false
@@ -1326,6 +1332,11 @@ ActiveRecord::Schema.define(version: 2020_07_07_173524) do
     t.index ["oauth_application_id"], name: "index_webhook_endpoints_on_oauth_application_id"
     t.index ["target_url"], name: "index_webhook_endpoints_on_target_url", unique: true
     t.index ["user_id"], name: "index_webhook_endpoints_on_user_id"
+  end
+
+  create_table "welcome_notifications", force: :cascade do |t|
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
   end
 
   add_foreign_key "api_secrets", "users", on_delete: :cascade
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_07_07_173524) do
+ActiveRecord::Schema.define(version: 2020_07_10_174257) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -306,6 +306,7 @@ ActiveRecord::Schema.define(version: 2020_07_07_173524) do
     t.string "title"
     t.string "type_of"
     t.datetime "updated_at"
+    t.index ["broadcastable_type", "broadcastable_id"], name: "index_broadcasts_on_broadcastable_type_and_broadcastable_id", unique: true
     t.index ["title", "type_of"], name: "index_broadcasts_on_title_and_type_of", unique: true
   end
 
--- a/lib/data_update_scripts/20200708163323_backfill_broadcastable_type_for_broadcasts.rb
+++ b/lib/data_update_scripts/20200708163323_backfill_broadcastable_type_for_broadcasts.rb
@@ -2,7 +2,7 @@ module DataUpdateScripts
   class BackfillBroadcastableTypeForBroadcasts
     def run
       Broadcast.where(broadcastable_type: nil).each do |cast|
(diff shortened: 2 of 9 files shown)


db/schema.rb
```diff
@@ -293,6 +299,8 @@
     t.datetime "active_status_updated_at"
     t.string "banner_style"
     t.text "body_markdown"
+    t.integer "broadcastable_id"
+    t.string "broadcastable_type"
```
vaidehijoshi, line 303: I think we probably want to add an index on broadcast for the `broadcastable_id` and `broadcastable_type`. Since polymorphic associations rely on querying both of these together to figure out _which_ type of broadcast we're dealing with, we probably want a combined index here, something like this?
```ruby
add_index :broadcasts, [ : broadcastable_type, : broadcastable_id]
```

To learn more about indexes + polymorphic associations, check out [this](https://mauricio.github.io/2008/09/27/handling-database-indexes-for-rails-polymorphic-associations.html) blog post!

vaidehijoshi: Also just want to note that we have a `index_broadcasts_on_title_and_type_of` currently on the `broadcasts` table; we'll want to explicitly remove that whenever we remove the `type_of` column :)

lib/data_update_scripts/20200708163323_backfill_broadcastable_type_for_broadcasts.rb
```diff
@@ -0,0 +1,9 @@
+module DataUpdateScripts
+  class BackfillBroadcastableTypeForBroadcasts
+    def run
+      Broadcast.where(broadcastable_type: nil).each do |cast|
+        cast.update(broadcastable_type: broadcast.type_of)
+      end
```
vaidehijoshi, line 6: So I ran into something interesting while QAing this locally. Check out this output:

```ruby
[2] pry(main)> Broadcast.where(broadcastable_type: nil).each do |cast|
[2] pry(main)*   cast.update(broadcastable_type: broadcast.type_of)
[2] pry(main)* end
  Broadcast Load (1.2ms)  SELECT "broadcasts".* FROM "broadcasts" WHERE "broadcasts"."broadcastable_type" IS NULL
   (0.2ms)  BEGIN
  Broadcast Exists? (1.1ms)  SELECT 1 AS one FROM "broadcasts" WHERE "broadcasts"."title" = $1 AND "broadcasts"."id" != $2 AND "broadcasts"."type_of" = $3 LIMIT $4  [["title", "Welcome Notification"], ["id", 1], ["type_of", "Onboarding"], ["LIMIT", 1]]
   (2.1ms)  ROLLBACK
```

If you read the `SELECT` statement carefully, then you'll see that I had a broadcast locally with a `type_of: "Onboarding"`, which isn't a valid `broadcastable_type`. The reason that this didn't blow up for me is that you have an `update` rather than an `update!` here, so when ActiveRecord couldn't update the broadcast, it just rolled back and moved on silently.

Although I don't think this is an issue on production (from what I can tell, there aren't any Broadcasts with a `type_of: "Onboarding"`), we probably want to decide how/if we want to handle this scenario if it happens to someone else locally.

db/schema.rb
rhymes: Question: can the combo "type" and "id" allow more than one item? Otherwise it should be a unique index as well, by adding `unique: true` to the parameters

lib/data_update_scripts/20200708163323_backfill_broadcastable_type_for_broadcasts.rb
juliannatetreault: Huh, that's strange... 🤔 

I didn't see any Broadcasts in production with `type_of: "Onboarding"` either, which is good. I'm curious about other people's thoughts on this, though. In the meantime, I've changed `update` to `update!` in the script. 

Thanks for drawing attention to this @vaidehijoshi!

db/schema.rb
juliannatetreault: Making these changes now! Thank you both for pointing all of this out and linking to the post @vaidehijoshi @rhymes. 😄

juliannatetreault:
> I think adding the `update!` here is probably sufficient 👍 Just be sure to merge and deploy this with enough buffer time during working hours so that you can rollback or handle things in case something goes wrong in the deploy or data update script step (I doubt that'll happen, though).

Awesome! Once I get another approval, I'll merge/deploy right away so that I can watch everything.

lib/data_update_scripts/20200708163323_backfill_broadcastable_type_for_broadcasts.rb
```diff
@@ -0,0 +1,9 @@
+module DataUpdateScripts
+  class BackfillBroadcastableTypeForBroadcasts
+    def run
+      Broadcast.where(broadcastable_type: nil).each do |cast|
```
rhymes, line 4: this should probably use `find_each` but as I don't think we have so much data it shouldn't be an issue :D

PR #9232, opened 2020-07-09
Update Docs, Travis,  and Containers to Use Postgres 11

mstruve:
## Description
We recently discovered that our migrations are not compatible with Postgres versions <11. Rather than continuing to try and support these versions, we have decided to update our docs to require Postgres version 11 or higher to run DEV.

This PR...
1) Updates all container files to use Postgres 11 images
2) Updates the Docs to say we require Postgres 11
3) Runs migrations in a single Travis job (I confirmed if you drop the Travis version to 10 and the StrongMigrations supported version to 10 you will end up with the migration error that @jdoss saw)
4) Bumps Travis to 10....Wait, what?! Travis does not play nice with Postgres-11. If you install it through [their services](https://travis-ci.community/t/postgresql-11-2-cannot-connect/2680) it will fire up on port 5433 bc Postgres 9 which is fired up first takes port 5432. If you get that resolved with some fancy `sed` work, then you end up with user and permission errors for the database. TL;DR Unless you want to set up Postgres 11 manually you can't use it. I don't want to go through all of that setup bc it will add more time to our builds. Also, I struggled with it for a couple of hours with no luck and decided since I think we will be moving off Travis soon to cut our losses for now.

https://github.com/thepracticaldev/dev.to/projects/9#card-41592053
 

alt_text

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ cache:
 rvm:
   - 2.7.1
 addons:
-  postgresql: '9.6'
+  postgresql: '10' # Travis does not work out of the box with Postgres 11 yet
   chrome: 'stable'
   artifacts:
     paths:
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ A more complete overview of our stack is available in
   the badge.
 - [Yarn](https://yarnpkg.com/) 1.x: please refer to their
   [installation guide](https://classic.yarnpkg.com/en/docs/install).
-- [PostgreSQL](https://www.postgresql.org/) 9.5 or higher.
+- [PostgreSQL](https://www.postgresql.org/) 11 or higher.
 - [ImageMagick](https://imagemagick.org/): please refer to ImageMagick's
   [installation instructions](https://imagemagick.org/script/download.php).
 - [Redis](https://redis.io/) 4 or higher.
--- a/container-compose.yml
+++ b/container-compose.yml
@@ -112,7 +112,7 @@ services:
     command: ["bundle", "exec", "sidekiq","-c","2"]
 
   db:
-    image: postgres:9.6.15-alpine
+    image: postgres:11-alpine
     container_name: dev_postgresql
     environment:
       POSTGRES_USER: devto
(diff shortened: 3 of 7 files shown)


rhymes:
Why is Travis such a disappointment sometimes :D PostgreSQL has been released on October 18, 2018. It's not like it's brand new :D 

Before we merge this, should we also address the "run all migrations on Travis" step? 
Do we know what was wrong with that migration?

I fear that without that part we're going to just postpone the underlying issue of mismatching of SQL statements with the correct PostgreSQL version

mstruve:
The problem migration is the [add_user_subscriptions_count_to_articles.rb](https://github.com/thepracticaldev/dev.to/blob/master/db/migrate/20200706184804_add_user_subscriptions_count_to_articles.rb) migration. The reason is because [prior to postgres 11 this requires an entire table rewrite](https://github.com/ankane/strong_migrations#adding-a-column-with-a-default-value) so Strong migrations will raise an error if the version is <11. 

> I fear that without that part we're going to just postpone the underlying issue of mismatching of SQL statements with the correct PostgreSQL version

We are "validating" migrations for Postgres 11, requiring Postgres 11 in our docs and using it ourselves so if someone writes a.migration that is only valid on Postgres 12, Travis should catch it still. I think having Travis on 10 is not a huge deal and will be resolved soon hopefully.

rhymes:
@mstruve thanks a lot! I forgot about `strong_migrations` checks. Let's be fearless and raise the minimum version. I can't wait to upgrade to 12 as well :P

rhymes:
Now I fully understand, our mistake was to tell `strong_migrations` to target PostgreSQL 11 and not 9.6 :-(

https://github.com/thepracticaldev/dev.to/blob/5094f6be9c1d6065b233654def953dbf3dad650e/config/initializers/strong_migrations.rb#L8

rhymes:
I took the liberty to update the 3 plaform installation docs as well that I noticed after approving, to avoid another PR

mstruve:
Thanks @rhymes! Totally missed those

PR #9241, opened 2020-07-10
Rubocop: fix Performance/OpenStruct

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

I replaced `OpenStruct` with `Struct` wherever possible (the [latter is faster](http://www.chrisrolle.com/en/blog/struct-vs-openstruct-benchmark)) and disabled it locally where it made sense.

I think we should eventually stop storing serialized objects in the DB and use standard formats like JSON but for now I just replaced a serialized open struct vs a struct. No need to backfill those, they don't have to interact and they behave the same way at runtime so articles will "backfill" themselves on save.

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -122,14 +122,16 @@ class Organization < ApplicationRecord
   def update_articles
     return unless saved_change_to_slug || saved_change_to_name || saved_change_to_profile_image
 
-    cached_org_object = {
-      name: name,
-      username: username,
-      slug: slug,
-      profile_image_90: profile_image_90,
-      profile_image_url: profile_image_url
-    }
-    articles.update(cached_organization: OpenStruct.new(cached_org_object))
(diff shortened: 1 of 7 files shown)


app/models/organization.rb
```diff
@@ -122,14 +122,16 @@ def downcase_slug
...
-    cached_org_object = {
-      name: name,
-      username: username,
-      slug: slug,
-      profile_image_90: profile_image_90,
-      profile_image_url: profile_image_url
-    }
-    articles.update(cached_organization: OpenStruct.new(cached_org_object))
+    # TODO: [thepracticaldev/oss] this should eventually be moved to JSON,
+    # to avoid storing a native Ruby object in the DB
```
vaidehijoshi, line 126: 👍 nice reminder!

PR #9245, opened 2020-07-10
Update instantClick's script clone logic

maestromac:
- [x] Bug Fix
- [x] DX

## Description
The instantClick that we adopted into our codebase is an older code of what's on [the official repo](https://github.com/dieulot/instantclick/blob/master). Even though the project is unmaintained, it looks like we are commits behind. This PR isn't going to address that issue but apply a tiny chunk of code change that will most likely resolve a very noisy bug. 

I can not reproduce this bug at all but I believe the cause of it is some script tags are removed as part of page change. That's probably the reason [this commit exists](https://github.com/dieulot/instantclick/commit/9fb9634c1d400e8b7ce7209c14f870cc0b8ee131) on the official project. This code can be improved further but I decided not to tamper too much since I don't have a test to back this up.
1. Current latest copy chunk of code is from here https://github.com/dieulot/instantclick/blob/master/src/instantclick.js#L296
2. The initial commit for the change https://github.com/dieulot/instantclick/commit/9fb9634c1d400e8b7ce7209c14f870cc0b8ee131
3. [Basically 90% of all](https://app.honeybadger.io/projects/67192/faults?sort=last_seen_desc&q=-is%3Aresolved%E2%80%83-is%3Aignored%E2%80%83Unexpected+token) the `Unexpected token ':'` we've been getting on HoneyBadger.

1. start up the server.
2. Browser around with the console open and there everything should be working normally.

## Added tests?
- [x] no, I actually have no idea how to write an adequate test for this but the test suite is passing.

## Added to documentation?
- [x] no documentation needed

n/a


app/assets/javascripts/base.js.erb
```diff
@@ -684,7 +688,7 @@ var instantClick
     apiKey: "<%= ApplicationConfig['HONEYBADGER_JS_API_KEY'] %>",
     environment: "<%= Rails.env %>",
     revision: "<%= ApplicationConfig['HEROKU_SLUG_COMMIT'] %>",
-    ignorePatterns: [/ResizeObserver/i, /MetaMask detected another web3/i],
```
Zhao-Andy, line 687: Huh, what's this?

app/assets/javascripts/base.js.erb
```diff
@@ -684,7 +688,7 @@ var instantClick
...
+    ignorePatterns: [/ResizeObserver/i, /MetaMask/i],
```
maestromac, line 691: @Zhao-Andy I think quoting this change is more appropriate. I'm loosening up the regex constraint so the ignore-pattern can be applied to more errors.

nickytonline:
Thanks Mac! One question I have is, perhaps we could look into the library that the InstantClick author does maintain now, [instant.page](https://instant.page/). It could make for an awesome cooldown spike.

nickytonline:
There's also a [short post](https://css-tricks.com/instant-page/) about it on CSS Tricks.

PR #9251, opened 2020-07-10
Make sidebar sponsor headline configurable

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This makes the "headline" of the gold sponsor area configurable, in case a community wants to list things differently there, such as describing "partners" or "supporters".

Screen Shot 2020-07-10 at 12 33 12 PM

### DIFF
--- a/app/views/internal/configs/show.html.erb
+++ b/app/views/internal/configs/show.html.erb
@@ -682,6 +682,26 @@
         <div class="card mt-3">
           <%= render partial: "card_header",
                      locals: {
+                       header: "Sponsors",
+                       state: "collapse",
+                       target: "sponsorsContainer",
+                       expanded: "false"
+                     } %>
+          <div id="sponsorsContainer" class="card-body collapse hide" aria-labelledby="sponsorsContainer">
+            <div class="form-group">
+              <%= f.label :sponsor_headline %>
+              <%= f.text_field :sponsor_headline,
+                               class: "form-control",
+                               value: SiteConfig.sponsor_headline,
+                               placeholder: "Community Sponsors" %>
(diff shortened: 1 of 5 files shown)


app/views/internal/configs/show.html.erb
```diff
@@ -679,6 +679,26 @@
...
+                       target: "sponsorsContainer",
+                       expanded: "false"
+                     } %>
+          <div id="sponsorsContainer" class="card-body collapse hide" aria-labelledby="sponsorsContainer">
+            <div class="form-group">
+              <%= f.label :sponsor_headline %>
+              <%= f.text_field :sponsor_headline,
+                               class: "form-control",
+                               value: SiteConfig.sponsor_headline,
+                               placeholder: "Community Sponsors" %>
```
nickytonline, line 696: The default headline is Community Sponsors but is configurable. should we not do the same for the placeholder?

app/views/internal/configs/show.html.erb
```diff
@@ -679,6 +679,26 @@
           </div>
         </div>
 
+        <div class="card mt-3">
+          <%= render partial: "card_header",
+                     locals: {
+                       header: "Sponsors",
+                       state: "collapse",
+                       target: "sponsorsContainer",
+                       expanded: "false"
```
nickytonline, line 688: Any reason we need a `state` and `expanded` property in locals? They seem like they would so the same thing.

benhalpern: Yeah, I dunno. This is the pattern we have established across all the configs. You're probably right, but refactoring this would be out-of-scope of the PR.

PR #9253, opened 2020-07-10
refactor chat_channel model - adjust for 20 methods bellow

brunnohenrique:
Refactoring model chat_channel, adjusting for 20 methods bellow

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
The issue about refactoring and code climate

Helps https://github.com/thepracticaldev/dev.to/issues/3739

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [X] no documentation needed


brunnohenrique:
hi @rhymes, I've tried to refactor as you suggested, but I got some logical errors in other classes.
For example: In this case `app/labor/assign_tag_moderator.rb` I changed only the call for class, but we had a changed of logic with `slug` variable.

Could you help me to understand which are the problems?

app/controllers/chat_channels_controller.rb
```diff
@@ -32,12 +32,12 @@ def show
 
   def create
     authorize ChatChannel
-    @chat_channel = ChatChannelCreationService.new(current_user, params[:chat_channel]).create
+    @chat_channel = ChatChannels::Creation.new(current_user, params[:chat_channel]).create
```
rhymes, line 35: could these new classes become verbs? `ChatChannels::Create` instead of `::Creation` ? Also, to keep a bit of consistency, can these new services use `.call` as their interface? See https://github.com/forem/forem/blob/master/app/services/users/delete.rb

unfortunately our code base isn't consistent in that so I understand the confusion

app/controllers/chat_channels_controller.rb
```diff
@@ -240,7 +240,7 @@ def render_chat_channel
   end
 
   def send_open_notification
-    adjusted_slug = if @chat_channel.group?
+    adjusted_slug = if @chat_channel.type?("group")
```
rhymes, line 243: I still think `.group?` is clearer than `.type?(group)`. Is there are any naming conflict with keeping it `.group?`?

app/controllers/internal/chat_channels_controller.rb
```diff
@@ -8,7 +8,7 @@ def index
     end
 
     def create
-      ChatChannel.create_with_users(
+      ChatChannels::CreateWithUser.perform(
```
rhymes, line 11: should this be `CreateWithUsers`, plural?

app/services/chat_channels/create_with_user.rb
```diff
@@ -0,0 +1,42 @@
+module ChatChannels
+  class CreateWithUser
+    def initialize(options = {})
+      @users = options.fetch(:users)
+      @channel_type = options.fetch(:channel_type, "direct")
+      @contrived_name = options.fetch(:contrived_name, "New Channel")
+      @membership_role = options.fetch(:membership_role, "member")
+    end
```
rhymes, line 8: I would replace this with `def initialize(users: [], channel_type: "direct", contrived_name: "New Channel", "membership_role: "member"`).

It's much clearer what the method requires

app/services/chat_channels/create_with_user.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    def self.perform(*args)
+      new(*args).perform
+    end
+
+    def perform
+      raise "Invalid direct channel" if invalid_direct_channel?(users, channel_type)
```
rhymes, line 15: this should probably also bail if users is an empty array

app/services/chat_channels/creation.rb
```diff
@@ -0,0 +1,15 @@
+module ChatChannels
+  class Creation
+    attr_accessor :user, :params
```
rhymes, line 3: is there a need to access this from the outside or can they just be private `attr_reader`s ?

app/services/chat_channels/creation.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    def initialize(user, params)
+      @user = user
+      @params = params
```
rhymes, line 7: same as the other. if you need `channel_name` and `slug` it's better to declare params explicitly:

`def initialize(user, channel_name:, slug:)`

this way you also get Ruby's error if the two keywords params aren't given

app/services/chat_channels/find_or_create.rb
```diff
@@ -0,0 +1,34 @@
+module ChatChannels
+  class FindOrCreate
+    def initialize(channel_type, slug, contrived_name)
+      @channel_type = channel_type
+      @slug = slug
+      @contrived_name = contrived_name
+    end
+
+    attr_accessor :channel_type, :slug, :contrived_name
```
rhymes, line 9: are this needed outside the class?

app/services/chat_channels/update.rb
```diff
@@ -0,0 +1,15 @@
+module ChatChannels
```
rhymes, line 1: I don't think we need this at all as it's simply replaceable with `channel.update(params)`

app/controllers/chat_channels_controller.rb
brunnohenrique: Ok, I've used `perform` cause is a default name for services, but I change to `call` to follow the project.
And the name of the classes I'll change too.

app/controllers/chat_channels_controller.rb
brunnohenrique: I'll remove the commit about these methods changes.

app/services/chat_channels/create_with_users.rb
```diff
@@ -0,0 +1,43 @@
...
+      raise "Invalid direct channel" if invalid_direct_channel?(users, channel_type)
+
+      usernames = users.map(&:username).sort
+      slug = if channel_type == "direct"
+               usernames.join("/")
+             else
+               "#{@contrived_name.to_s.parameterize}-#{rand(100_000).to_s(26)}"
+             end
+
+      contrived_name = channel_type == "direct" ? "Direct chat between " + usernames.join(" and ") : @contrived_name
```
brunnohenrique, line 27: We have a problem here with variable `contrived_name` without @ this is `null`. The most thing strange is in the class `initialize` the value is normal, but after that the value is null

brunnohenrique:
@rhymes I had problems with COVID, and stop with development to rest myself. But now is everything ok.
I did the changes asked and the checks have passed.

rhymes:
Hi @brunnohenrique, sorry to hear that! I hope you're doing fine. I'll take a look at the code and leave a review ASAP

app/services/chat_channels/create.rb
```diff
@@ -0,0 +1,16 @@
+module ChatChannels
```
rhymes, line 1: Sorry I didn't catch this in the initial reviews but this service doesn't seem to be useful (it wasn't in the previous incarnation as well). We should directly call `user.chat_channels.create()` and get rid of this as it's an entire service to replace one line of code

app/services/chat_channels/create_with_users.rb
```diff
@@ -0,0 +1,43 @@
...
+  class CreateWithUsers
+    def initialize(users: [], channel_type: "direct", contrived_name: "New Channel", membership_role: "member")
+      @users = users
+      @channel_type = channel_type
+      @contrived_name = contrived_name
+      @membership_role = membership_role
+    end
+
+    attr_reader :users, :channel_type, :membership_role
+    attr_accessor :contrived_name
```
rhymes, line 11: are these used outside? otherwise they can be set to `private`. also I don't think you need `contrived_name` to be an accessor. 

I would use `attr_reader :users, :channel_type, :contrived_name, :membership_role`

app/services/chat_channels/create_with_users.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    def call
+      raise "Invalid direct channel" if invalid_direct_channel?(users, channel_type)
+
+      usernames = users.map(&:username).sort
+      slug = if channel_type == "direct"
+               usernames.join("/")
+             else
+               "#{@contrived_name.to_s.parameterize}-#{rand(100_000).to_s(26)}"
```
rhymes, line 24: this can be `contrived_name.to_s.parameterize` without the `@`

app/services/chat_channels/create_with_users.rb
rhymes: I think the problem here is that the class defines an `attr_accessor` and then here there's a local variable with the same name of the accessor. You need the following:

- `attr_reader :contrived_name`
- this local variable to have a different name

app/services/chat_channels/find_or_create.rb
```diff
@@ -0,0 +1,34 @@
+module ChatChannels
+  class FindOrCreate
+    def initialize(channel_type, slug, contrived_name)
+      @channel_type = channel_type
+      @slug = slug
+      @contrived_name = contrived_name
+    end
+
+    attr_reader :channel_type, :slug, :contrived_name
```
rhymes, line 9: if this aren't accessed outside, they should be set private

app/services/chat_channels/find_or_create.rb
```diff
@@ -0,0 +1,34 @@
...
+      new(*args).call
+    end
+
+    def call
+      channel = ChatChannel.find_by(slug: slug)
+      if channel
+        raise "Blocked channel" if channel.status == "blocked"
+
+        channel.status = "active"
+        channel.save
```
rhymes, line 21: this can be `channel.update(status: "active")` I guess

brunnohenrique:
@rhymes I think is everything ok now. I concluded the suggestions and the tests are ok.

rhymes:
@brunnohenrique thanks. I'll take a look! please refrain from force pushing if possible, it's harder to follow PR changes otherwise, see our contributing guide:

> After you submit your pull request, one of the members of the core team will review your code.

> Please avoid force pushing unless you need to rebase with the master branch.

> If feedback is provided, any changes should be contained in new commits. Please don't force push or worry about squashing your commits.

https://docs.forem.com/contributing/forem/#contribution-guidelines

PR #9255, opened 2020-07-10
Update RssReader::Assembler#get_content hash syntax

maestromac:
- [x] Bug Fix

## Description
This change interaction with RSS's `@item` as hash instead of method call so it's nullable.

Resolves https://app.honeybadger.io/fault/66984/90d3be76e672b09bcf0fa82d0f7f1be7
warning: this will create a lot of local articles.
1. spin up rails console.
2. Add the feed to your personal account. ie `me.update(feed_url: "http://www.blameless.com/blog/rss.xml")`
3. Run `RssReader.get_all_articles`
4. You should be able to fetch all feed successfully without error.

## Added tests?

## Added to documentation?
- [x] no documentation needed

mark honeybadger item as resolved

### DIFF
--- a/app/services/rss_reader/assembler.rb
+++ b/app/services/rss_reader/assembler.rb
@@ -54,7 +54,7 @@ class RssReader
     end
 
     def get_content
-      @item.content || @item.summary || @item.description
+      @item[:content] || @item[:summary] || @item[:description]
     end
 
     def thorough_parsing(content, feed_url)


app/services/rss_reader/assembler.rb
```diff
@@ -54,7 +54,7 @@ def assemble_body_markdown
     end
 
     def get_content
-      @item.content || @item.summary || @item.description
+      @item[:content] || @item[:summary] || @item[:description]
```
mstruve, line 57: Is there going to be any case where `@item` is not present? If so we might want to use `@item&.dig(..)` with the safety operator. Not necessary but thinking ahead if that's a possible case.

maestromac: That's unlikely as `@item` is individual RSS entries generated by Feedjira gem.

PR #9256, opened 2020-07-10
Bug Fix: Setup STDOUT Logger bc we need it

mstruve:
- [x] Bug Fix

## Description
Apparently we need the STDOUT logging part bc, [surprise `RAILS_LOG_TO_STDOUT` is present on Heroku.](https://blog.heroku.com/container_ready_rails_5)
> This value can be set by the container or the platform on which your Rails app runs. In our case, the Ruby buildpack detects your Rails version, and if it's Rails 5 or greater will set the RAILS_LOG_TO_STDOUT environment variable.
```
irb(main):002:0> ENV["RAILS_LOG_TO_STDOUT"].present?
=> true
```

alt_text


rhymes:
Magic

PR #9262, opened 2020-07-11
Rubocop: move dot in multi-line calls to leading position

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR moves the dot for the multiline calls in the leading position. A few reasons:

1. when one "greps" for things it's immediately clear that 

```ruby
           .method()
```

is part of a multi line call, while:

```ruby
           method()
```

is not. 

2. it's easier to visually scan, especially for long lines, see the difference between:

```ruby
          .includes(:user)
          .select(SHOW_ATTRIBUTES_FOR_SERIALIZATION)
          .find(params[:id])
          .decorate
```

and 

```ruby
          includes(:user).
          select(SHOW_ATTRIBUTES_FOR_SERIALIZATION).
          find(params[:id]).
          decorate
```

3. the generated diff only includes the new mehod call when added at the bottom, without having to needlessly change the previous line as well. See the difference between:

```diff
diff --git a/app/controllers/api/v0/articles_controller.rb b/app/controllers/api/v0/articles_controller.rb
index 47348cd74..04f275e1e 100644
--- a/app/controllers/api/v0/articles_controller.rb
+++ b/app/controllers/api/v0/articles_controller.rb
@@ -23,7 +23,8 @@ module Api
           includes(:user).
           select(SHOW_ATTRIBUTES_FOR_SERIALIZATION).
           find(params[:id]).
-          decorate
+          decorate.
+          do_something_else()

         set_surrogate_key_header @article.record_key
       end
```

vs

```diff
diff --git a/app/controllers/api/v0/articles_controller.rb b/app/controllers/api/v0/articles_controller.rb
index 1889f79aa..43c26f023 100644
--- a/app/controllers/api/v0/articles_controller.rb
+++ b/app/controllers/api/v0/articles_controller.rb
@@ -24,6 +24,7 @@ module Api
           .select(SHOW_ATTRIBUTES_FOR_SERIALIZATION)
           .find(params[:id])
           .decorate
+          .do_something_else()

         set_surrogate_key_header @article.record_key
       end
```

## Added tests?

- [ ] yes
- [ ] no, because I need help


rhymes:
@mstruve it's totally fair and it make sense. I've never actually grepped for that and always worked on code bases with the leading dot, so to me it's the opposite, I found it odd when approaching DEV's codebase initially. 

The point of this and likely some of the next style PRs in the pipeline is to discuss styling among us but as the goal is to work towards not having these discussions at all in the future (so we can concentrate on content and not form) and implementing [standard](https://github.com/testdouble/standard) you should know that if we proceed to adopt this tool, it's going to put leading dots anyway (which is also the reason why I submit this PR, to "prepare the field"). 

See https://github.com/testdouble/standard/issues/75 for an expansion on the reasoning behind the leading dot.

This is our card https://github.com/thepracticaldev/dev.to/projects/2#card-35265769

mstruve:
I'm all for coming up with a single style guide so we don't discuss this stuff in the future but this is one styling approach that I think we should keep the way it is and use the trailing dots. 

In addition to grep'ing, I am a big copy/paste'r into terminals so the issues with the leading dots not supporting that are super frustrating. Coming here was a breath of fresh air bc I cam copy and paste chunks of code into terminals for debugging without having to update it all.

Plus, since I do a lot of debugging, I am often commenting out pieces of scopes [like what is mentioned here](https://github.com/testdouble/standard/issues/75#issuecomment-460661597) and with the leading dots, that causes errors. 

```ruby
foo.
  bar.
  # baz.
  boo
```
No problems. 😎
```ruby
foo
  .bar
  # .baz
  .boo
```
syntax error, unexpected '.', expecting end-of-input. 😢

> In our case it happens on things that are not automatically enforced:

This is automatically enforced as trailing and I think that is something we should keep.

rhymes:
@mstruve 

I don't disagree with your reasoning to prefer trailing dots, though I still prefer leading dots. But I think we should have a different conversation. If we start investigating the code we can find linting styles that some of us will prefer and others that won't and viceversa.

I think the actual conversation here is (again, my bad for not having brought it up since the start): are we ok to set aside personal preference (which in this case aligns with mine but it could have not as well) to adopt a tool that shall reduce the amount of time we spend debating style?

After all, what we're doing between the two of us right now it's exactly that, debating style :D And there are many other team members that could chime in on each change. And none of us would be wrong in absolute terms. 

That's precisely why we want to move towards [standard](https://github.com/testdouble/standard), and then let Rubocop worry about more complex issues that are not alignment.

ps. Ruby 2.7 allows you to do:

```ruby
class Foo
  def bar
    puts "bar"
    self
  end

  def boo
    puts "boo"
    self
  end
end

foo = Foo.new

foo
  .bar
  # .baz
  .boo
```

If you put this in a `example.rb` and execute it you'll notice there's no syntax error. Unfortunately it's our choice of editor, Sublime Text 3 that complains, Visual Studio Code doesn't.

rhymes:
I also found the fix for Sublime Text 3, open the "SublimeLinter-ruby" settings and paste this in the user one:

```json
    "paths": {
        "osx": ["/Users/rhymes/.rubies/ruby-2.7.1/bin"],
    },
```

where the path is the path to your Ruby 2.7 bin folder, otherwise by default it uses 2.6 which is the system Ruby in macOS. The syntax error disappeared after restart

mstruve:
Ok, I think the part I am confused about is why is [this the standard](https://github.com/testdouble/standard) we are looking to duplicate?

citizen428:
> Ok, I think the part I am confused about is why is this the standard we are looking to duplicate?

IIRC the original reason was to move to an extremely opinionated linter to avoid style discussions. 

I looked into Standard a bit during the last cooldown, and from what I gathered (from an issue I think) it is primarily targetting plain Ruby, not Rails, and doesn't allow for adding additional Rubocops like we do. So I think what @rhymes is doing here is moving us closer to Standard while keeping Rubocop.

> are we ok to set aside personal preference 

Personally I am. I generally favor trailing dots, mostly because leading dots didn't work until Ruby 1.9 or so (a bit shaky on the exact timeline). The main reason, however, is that copy-pasting into a Rails console doesn't work. Using your example from above:

```ruby
irb(main):013:0> foo = Foo.new
irb(main):014:0> foo
=> #<Foo:0x000000010d6b3330>
irb(main):015:0>   .bar
Traceback (most recent call last):
   [...]
SyntaxError ((irb):15: syntax error, unexpected '.')
  .bar
  ^
irb(main):016:0*   # .baz
irb(main):017:0>     .boo^C
```

That said, I'd still be ok to switch, team decision > personal preference. BUT I'm not entirely convinced by Standard anymore, it's pretty idiosyncratic in places (e.g. semantic blocks) and given that we run this project as open-source we should also consider that this will affect all our contributors.

rhymes:
@mstruve @citizen428 

I personally think the advantages trump copy-pasting in the REPL, which you can circumvent using `edit` inside pry, but even if we disagree on this I'm obviously okay either way. The point of these PRs is to spark a discussion.

I would like other team members to chime in as well:

1 - do we ultimately end up with our custom Rubocop configuration
1 - do we work towards [standard](https://github.com/testdouble/standard) + our custom Rubocop configuration for things that are not linting?

If there's no interest I'll eventually close the PR and move on :)

citizen428:
@rhymes I agree on choosing an opinionated tool and being done with it. I'm just no longer sure that Standard should be this tool. Prettier's [Ruby plugin](https://github.com/prettier/plugin-ruby) could be an alternative, as we are already using prettier for JS. Then we could use Rubocop only for things that aren't style-related (e.g. performance and security cops) and prettier for formatting. 

That said, I believe in [do-ocracy](https://communitywiki.org/wiki/DoOcracy) and since you're putting in the work right now, that counts more in my book than my random comments from the sidelines.

rhymes:
@citizen428 I'll wait a few days for the results of the poll :D I'll also take a look at prettier plugin! :D Thanks!

Issue #8849, opened 2020-06-23
Incorrect node version used in GitPod


Amorpheuz:
**Describe the bug**
Creating a new GitPod instance leads to a bug where under `gem install solargraph; bin/setup && bin/startup` the `bin/setup` script fails at `yarn install` due to incorrect node version being used (which is 12.18.x). 

The [GitPod Dockerfile](https://github.com/thepracticaldev/dev.to/blob/master/.gitpod.dockerfile) does handle the installation of nvm and correct node version, but forgets to switch to the correct node version (12.16.3).

**To Reproduce**
1. Create a GitPod Workspace with https://gitpod.io/#https://github.com/thepracticaldev/dev.to
2. Wait till `bundle install` finishes in the third `workspace/dev.to` tab completes.
3. Yarn throws the error as indicated in screenshot below.

**Expected behavior**
Correct node version is selected in nvm, which allows the `yarn install` command to complete successfully.

**Screenshots**
![image](https://user-images.githubusercontent.com/22113778/85383294-6b990d80-b55d-11ea-92ff-a74d807c917b.png)

**Desktop (please complete the following information):**

- OS: Windows 10
- Browser: Edge
- Version: 83

**Additional context**
 The `nvm use` command can be of help.

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

jacobherrington:
It looks like we might need to add `&& nvm use $NODE_VERSION` to https://github.com/thepracticaldev/dev.to/blob/2a31f7b9de7e9090ec908dd2a2d28508d807ab6f/.gitpod.dockerfile#L10

citizen428:
I have a draft PR open for this but never quite got it working. Let me try again.

akhil-naidu:
For now, you can manually change the version in package.json file. 

I'm not an expert to know the crux of what it actually does, but it gave me an active instance I'm my case.

PR #9271, opened 2020-07-12
GitPod: Use node version 12.16.3

donofriov:

- [x] Bug Fix

## Description

After installing a new node version with nvm it has to be explicitly chosen to use. This adds the nvm use command to fix the `bin/setup` script

#8849 

View the [GitPod workflow for this PR](https://gitpod.io/#https://github.com/thepracticaldev/dev.to/pull/9271) and verify an error is no longer thrown on [yarn install](https://user-images.githubusercontent.com/22113778/85383294-6b990d80-b55d-11ea-92ff-a74d807c917b.png)

## Added tests?

## Added to documentation?

- [x] no documentation needed

### DIFF
--- a/.gitpod.dockerfile
+++ b/.gitpod.dockerfile
@@ -7,7 +7,7 @@ RUN bash -lc "rvm install ruby-$RUBY_VERSION && rvm use ruby-$RUBY_VERSION --def
 
 # Install Node
 COPY ./.nvmrc "${APP_HOME}"/.nvmrc
-RUN bash -lc ". .nvm/nvm.sh && nvm install && nvm alias default $(cat .nvmrc) && nvm use default && nvm uninstall $NODE_VERSION && nvm use default"
+RUN bash -lc ". .nvm/nvm.sh && nvm install && nvm unalias default && nvm alias default $(cat /.nvmrc) && nvm use default && nvm uninstall $NODE_VERSION"
 
 # Install Redis.
 RUN sudo apt-get update \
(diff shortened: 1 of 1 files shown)


donofriov:
Yeah sorry this isn't ready for review. The initial fix of just adding `nvm use node_version` isn't working even though it _should_ be. I tried removing the initial command but that fails the workspace from starting. Continuing to debug.

donofriov:
I've opened a topic in the GitPod Community as this solution should be working: https://community.gitpod.io/t/nvm-version-not-being-used/1542

donofriov:
This gets the right version of node in the container and nvm is now reading from .nvmrc which is a nice enhancement, but now Yarn is missing 😢 

```
gitpod /workspace/dev.to $ nvm list
->     v12.16.3
default -> 12.16 (-> v12.16.3)
node -> stable (-> v12.16.3) (default)
stable -> 12.16 (-> v12.16.3) (default)
```

```
Yarn executable was not detected in the system.
Download Yarn at https://yarnpkg.com/en/docs/install
```

donofriov:
I have to take a break now, but the current status is that `yarnpkg` is working just fine from the command line in the Gitpod workspace, but is failing as part of the `bin/yarn` script.

citizen428:
@donofriov I went through a lot of exactly the same in https://github.com/thepracticaldev/dev.to/pull/8701. The history is lost, as I force-pushed a lot on the WIP branch, but according to last reports by @joshpuetz it's working for him now. Do you mind taking that branch for a spin and reporting back?

donofriov:
@citizen428 I have everything working here now all within the docker file which is where nvm commands should be run. Adding it to the beginning of the path allows us to have the defined version of node and yarn always available in every session within the gitpod container.

citizen428:
@donofriov Agreed, better approach. Funnily enough it's almost what I had started from, I just fixed the path issue in a hackier way 😆 Thanks for the PR!

PR #9285, opened 2020-07-12
Add local_date helper to handle dates with respect of user timezone

### LINKED ISSUE #9269: New posts/users have their published/joined date a day ahead
**Describe the bug**
I was looking through some user-profiles and noticed that some have a date that is in the future. Found users that have a joined date of July 12, 2020. And investigating further looks like this applies to posts and comments.

**To Reproduce**
Create a post and see if the published date is 1 day ahead

**Expected behavior**
Users should have posts and join dates of the current day

**Screenshots**

Screen Shot 2020-07-11 at 6 11 22 PM

Screen Shot 2020-07-11 at 6 05 45 PM

Screen Shot 2020-07-11 at 6 24 02 PM

- OS: macOS 10.15.5
- Browser: Safari, Firefox, Chrome
- Version:

- Device:
- OS:
- Browser:
- Version:

**Additional context**

I believe the issue is coming from the additional day we get this year.

### Users join dates
current   ->  actual
Feb 28, 2020 -> Feb 28, 2020
Mar 1, 2020 -> Feb 29, 2020
Mar 2, 2020 -> Mar 1, 2020
...
Jul 12, 2020 -> Jul 11, 2020

pashagray:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
I've added `local_date` helper with specs. And changed `_user_metadata.html.erb` for demonstration. If it's OK, I change other views. 
I want to use stimulus with date-fns (found some info that Dev.to is moving to Stimulus stack), but all stimulus controllers are available only in internal layout.

`local_date` helper shows data even without JS, but JS is used to convert it to proper timezone.

Closes #9269 

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed

alt_text


app/helpers/date_time_helper.rb
```diff
@@ -0,0 +1,13 @@
+module DateTimeHelper
+  # Show date in local timezone. Expected to be used
+  # by frontend JS library like date-fns or moment.js
+  def local_date(datetime)
+    tag.span(
```
rhymes, line 5: instead of being a span it should be a `<time>` element. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time and https://github.com/thepracticaldev/dev.to/blob/5094f6be9c1d6065b233654def953dbf3dad650e/app/views/articles/show.html.erb#L155

app/helpers/date_time_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+      datetime.strftime("%b %e, %Y"),
+      data: {
+        helper: "local-date",
+        utc: datetime.iso8601
+      },
```
rhymes, line 10: these aren't needed if you use the proper `<time>` element

app/helpers/date_time_helper.rb
```diff
@@ -0,0 +1,13 @@
+module DateTimeHelper
+  # Show date in local timezone. Expected to be used
+  # by frontend JS library like date-fns or moment.js
+  def local_date(datetime)
```
rhymes, line 4: I would call it `local_datetime` as it contains both.

db/schema.rb
```diff
@@ -483,6 +483,7 @@
     t.datetime "updated_at", null: false
```
rhymes, line 483: this should be removed from the PR

PavelTkachenko:
Sorry if I make dumb decisions, it's a first pull-request here =)

I've change everything according to remarks, but:
l've add `date` class to time tag (as saw in some files) and left helper name unchanged to justify that format should be a date. I query like this `document.querySelectorAll('time.date')` to avoid conflict with `initializeCreditsPage` which is also queries time tags: `const datetimes = document.querySelectorAll('.ledger time')`.

rhymes:
> Sorry if I make dumb decisions, it's a first pull-request here =)

no dumb decisions and thanks for the PR! :-)

> l've add date class to time tag (as saw in some files) and left helper name unchanged to justify that format should be a date. I query like this document.querySelectorAll('time.date') to avoid conflict with initializeCreditsPage which is also queries time tags: const datetimes = document.querySelectorAll('.ledger time').

Cool, so if it's just for dates I think we should call `initializeDateTimeHelpers` `initializeDateHelpers`?

Another thing: shouldn't this PR contain the fix for https://github.com/thepracticaldev/dev.to/issues/9269 as well?

One last thing: you don't need to force push each change, as we're going to squash merge in the end anyway. Only force push if you need to rebase from master, see https://docs.dev.to/contributing/forem/#contribution-guidelines

PavelTkachenko:
Thanks for patience!

1. Changed it to date_helper
2. PR contains fixes for #9269 
3. Got it! No more --force!

Article date doesn't have year in design, so I modified helper to handle this. Also `Article#published_timestamp` returns `String`, not `DateTime`, so I changed helper to work with both strings and time/datetime/date. Don't know whether it is good or bad.

rhymes:
@PavelTkachenko I updated the PR to get master in as we've moved over to `/forem/forem` and Travis still pointed to the old version

PavelTkachenko:
@rhymes Is there anything I can do here?

app/assets/javascripts/initializers/initializeDateTimeHelpers.js
```diff
@@ -0,0 +1,18 @@
+/* global localizeTimeElements */
+
+'use strict';
+
+function initializeDateHelpers() {
+  // Date without year: Jul 12
+  localizeTimeElements(document.querySelectorAll('time.date-no-year'), {
+    month: 'short',
+    day: '2-digit',
```
fdoxyz, line 9: I think this can sort out the zero-padding issue

```suggestion
    day: 'numeric',
```

app/assets/javascripts/initializers/initializeDateTimeHelpers.js
```diff
@@ -0,0 +1,18 @@
...
+  });
+
+  // Full date: Jul 12, 2020
+  localizeTimeElements(document.querySelectorAll('time.date'), {
+    year: 'numeric',
+    month: 'short',
+    day: '2-digit',
```
fdoxyz, line 16: Same as above

```suggestion
    day: 'numeric',
```

PavelTkachenko:
@fdoxyz Thanks for review! I've fixed helper to meet spec requirements.

PR #9289, opened 2020-07-13
Site Config: Get Started Section and Show Required Fields

Ridhwana:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In [PR 8435](https://github.com/thepracticaldev/dev.to/pull/8435), we had introduced required vs optional fields to the site configuration layer. When something is required we show a banner. However, currently as the interface stands, it is not easy to see which are the required fields on the site configuration page. Hence **I have introduced the 'Required' label next to fields that are required.** (See screenshots)

However, with only the label, it's still difficult to find the required fields because they are hidden within the sections that are collapsed by default, so a user would need to open up each section to find the required fields. 
Hence, the next natural step for me was to pull out the required fields in **a 'Get Started' section**, which at some point in the future will probably be its' own onboarding workflow (something like Lisa's designs that she walked us through). 

The Get started section is expanded by default when all the mandatory configs are not completed however when they have been, it collapses. Thanks @citizen428 for this idea 😁. In addition, I've pulled out the descriptions and the placeholders in a hash to make it a bit more dynamic, because when we have a Get Started section we don't want to keep hardcoding the values but we want to simply use a loop to go through the mandatory configs, and hence we need the associated description and placeholder for the config.

The one thing to note for _this feature is that it's not the highest priority at the moment and so we just need something that is going to take us to that next step, and make it less confusing for admins and developers setting up._ 

In the future I will clean up some of the template into a partial as well as clean up descriptions and placeholders.

Go to http://localhost:3000/internal/config where you will see the following changes:

screencapture-localhost-3000-internal-config-2020-07-13-16_15_02

In addition, you should be able to update fields from either the Site Config or the Get Started section and the changes should reflect on both. You will see "Required" tags on the mandatory fields, and finally the Get Started section will be expanded or contracted by default  based on whether you have filled out all the mandatory configs. 🙃

## Added tests?

- [x] yes, I've amended them
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

None

alt_text


app/helpers/application_helper.rb
```diff
@@ -246,8 +246,84 @@ def sanitize_and_decode(str)
...
   def internal_config_label(method, content = nil)
-    content ||= method.to_s.humanize
-    content << "*" if method.in?(VerifySetupCompleted::MANDATORY_CONFIGS)
-    label_tag("site_config_#{method}", content)
+    content ||= "<span>#{method.to_s.humanize}</span>"
+    content << "<span class='site-config__required'>Required</span>" if method.to_sym.in?(VerifySetupCompleted::MANDATORY_CONFIGS)
+    label_tag("site_config_#{method}", raw(content),  class: 'site-config__label')
+  end
+
+  def site_config_description
```
Ridhwana, line 254: So currently we've hardcoded the description and placeholders in the view. I'm not sure if these can be stored with (rails-settings-cached) in the database or should we just have a hash that maps them. And if we do keep the hash, where does this function belong?

The reason we need it to be dynamic is because when we have a `Get Started` section we don't want to keep hardcoding the values but we want to simply use a loop to go through the mandatory configs.

app/helpers/application_helper.rb
```diff
@@ -246,8 +246,84 @@ def sanitize_and_decode(str)
   end
 
   def internal_config_label(method, content = nil)
-    content ||= method.to_s.humanize
-    content << "*" if method.in?(VerifySetupCompleted::MANDATORY_CONFIGS)
-    label_tag("site_config_#{method}", content)
+    content ||= "<span>#{method.to_s.humanize}</span>"
+    content << "<span class='site-config__required'>Required</span>" if method.to_sym.in?(VerifySetupCompleted::MANDATORY_CONFIGS)
```
Ridhwana, line 250: I've added a required label that's more visible instead of just the asterick

app/views/internal/configs/show.html.erb
```diff
@@ -11,16 +11,74 @@
...
+  <div class="card w-100">
+    <div class="card-header" id="getStartedHeader" data-toggle="collapse"
+      data-target="#getStartedBodyContainer" aria-expanded="true" aria-controls="getStartedBodyContainer">
+      <h2 class="d-inline">Get Started</h2>
+      <button class="btn btn-secondary float-right" type="button">
+        Toggle Get Started
+      </button>
+    </div>
+
+    <div id="getStartedBodyContainer" class="collapse show hide p-3" aria-labelledby="siteConfigHeader">
```
Ridhwana, line 24: This section is a `Get Started` section that will be expanded on the Site Config page. We will list the required fields in this section so a new forem user will easily and very quickly be able to configure their site without expanding and collapsing all those tabs. This will translate into an onboarding section for the forem admin at some point.

app/views/internal/configs/show.html.erb
```diff
@@ -11,16 +11,74 @@
...
+      <%= form_for(SiteConfig.new, url: internal_config_path) do |f| %>
+
+        <% VerifySetupCompleted::MANDATORY_CONFIGS.each do |config_key| %>
+          <!-- we need to list the config as seperate fields if the data structure is a Hash -->
+          <% if SiteConfig.send(config_key).is_a?(Hash) %>
```
Ridhwana, line 29: We have this if because we need to list the config as separate fields if the data structure is a Hash. We dont need to do this for an array or string. It feels weird to do this but without hardcoding the entire get started section I'm not sure what else we can do.

app/views/internal/configs/show.html.erb
Ridhwana, line 24: We can probably move this into a partial. Once I get feedback and I'm polishing I can do this.

Ridhwana: Also we can probably move this into a partial. Once I get feedback and I'm polishing I can do this.

app/helpers/application_helper.rb
citizen428: > I'm not sure if these can be stored with (rails-settings-cached) in the database or should we just have a hash that maps them.

Let's keep things simple for now, I think a lookup hash is fine for the time being.

> And if we do keep the hash that I've outlined below, does this function belong in the application helper or should it be moved somewhere else?

Maybe add it as `SiteConfigDescriptions` to `app/lib/constants.rb` to avoid recreating this fairly big hash on every method invocation.

app/views/internal/configs/show.html.erb
citizen428: Maybe with a small change we can also reuse `setup_completed?` from `VerifySetupCompleted` and hide this section if everything's been configured already?

app/views/internal/configs/show.html.erb
citizen428: You'd be surprised at how many `is_a?` checks can be found inside Rails itself, so don't feel too bad about this.

citizen428:
> Hence, the next natural step for me was to pull out the required fields in a 'Get Started' section, which at some point in the future will probably be its' own onboarding workflow (something like Lisa's designs that she walked us through).

Thanks @Ridhwana, I think this will do nicely until we have a better onboarding flow 😃

app/views/internal/configs/show.html.erb
citizen428: @Ridhwana This is one of the few cases where I think [`module_function`](http://ruby-doc.org/core-2.7.1/Module.html#module_function-method) is the best solution. It makes a method's mixin also available on the class level, so if you add `module_function :setup_completed?` to `VerifySetupCompleted` we're able to do this:

```ruby
VerifySetupCompleted.setup_completed?
```

rhymes:
FYI: the reason why https://codeclimate.com/github/forem/forem/pull/9289 is complaining is because of `raw()`, I think with `content_tag` you can make those flagged issues disappear: https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-content_tag

Ridhwana:
> FYI: the reason why https://codeclimate.com/github/forem/forem/pull/9289 is complaining is because of `raw()`, I think with `content_tag` you can make those flagged issues disappear: https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-content_tag

I tried content_tag and it was breaking in other ways + I was getting a different code climate error about content_tag. Now that I polished up, I'll check it out again.

Ridhwana:
@rhymes getting something similar with content_tag. Any suggestions?

Screenshot 2020-07-16 at 12 39 46

app/views/internal/configs/show.html.erb
Ridhwana: @citizen428 I was really scratching my head for this one so thank you, this was helpful!

PR #9294, opened 2020-07-13
Invite users to join and create password

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As an authentication flow which could be a pretty great fundamental to forems is _invite-based_, which is, for example, how Slack works.

This integrates the Devise Invitable gem. It's an interesting implementation in that it creates users at the time of invitation, and then the users are sent an email where they can set a password.

I think it also makes sense as a feature that any existing account _can_ set a password.

This pull request adds some new fields related to the devise invitable gem and the associated functionality. But the password functionality already existed because it came with default devise and we never changed it. So finally closing the loop on this 4+ years later. 😄

This does not make use of the user-to-user invitation flow. This is admin only for now, but that functionality exists in the gem;s functionality and I think could ultimately become a pretty core feature of how a typical forem grows— that is to invite friends or whatever. 

It's a pretty typical concept: Almost all of our team tools work like this and it could really open some things up for how forems can work.

I made this PR to be somewhat brutalist, but if people are cool with this gem I can close the loop and get it all working. It's like 90% ready to go, just needs more testing, polish and refactoring before we put the stamp on this.

Here are some new visuals...

If a forem offers self-serve social auth, passwords are an option if you have already set one. If a forem is invite-only (no social auth at all, then a password is required but registration is not available.

Screen Shot 2020-07-14 at 9 06 04 PM
Screen Shot 2020-07-14 at 9 06 33 PM
Screen Shot 2020-07-14 at 9 06 59 PM
Screen Shot 2020-07-14 at 9 07 24 PM
Screen Shot 2020-07-14 at 9 52 46 PM

The above screenshots were taken before _forgot password_ functionality was added...

Latest:

Screen Shot 2020-07-15 at 9 18 22 AM
Screen Shot 2020-07-15 at 9 18 28 AM

Since authentication will be a major part of this next cycle, I feel like this kickstarts us with a fairly _technically_ fleshed out flow which we could build on in terms of improved design, UX, and polish as we introduce more flows.


app/controllers/internal/invitations_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+    layout "internal"
+
+    def index
+      @invitations = User.where(registered: false).page(params[:page]).per(50)
+    end
+
+    def new; end
+
+    def create
+      email = params[:user][:email]
```
citizen428, line 12: **suggestion (non-blocking):** I prefer using

```ruby
params.dig(:user, :email)
```

as it reads better and is `nil` safe if a key is missing.

app/controllers/internal/invitations_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+      name = params[:user][:name]
+      username = name.downcase.tr(" ", "_") + rand(1000).to_s
```
citizen428, line 14: **nitpick (non-blocking):** During one of the last cooldowns I changed most code like this, would be great if we can keep it consistent:


```suggestion
      username = "#{name.downcase.tr(" ", "_")}#{rand(1000)}"
```

app/helpers/application_helper.rb
```diff
@@ -101,6 +101,10 @@ def tag_colors(tag)
     end
   end
 
+  def any_selfserve_auth?
```
citizen428, line 104: **suggestion (non-blocking):** I find the name a bit confusing, how about the more verbose `selfserve_auth_available?`?

app/controllers/internal/invitations_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+      User.invite!(email: email,
+                   name: name,
+                   username: username,
+                   remote_profile_image_url: Users::ProfileImageGenerator.call,
+                   registered: false)
+      redirect_to "/internal/invitations"
```
citizen428, line 20: **nitpick:** Let's please use path helpers as much as possible:

```suggestion
      redirect_to internal_invitations_path
```

db/migrate/20200712150048_devise_invitable_add_to_users.rb
```diff
@@ -0,0 +1,32 @@
+class DeviseInvitableAddToUsers < ActiveRecord::Migration[6.0]
+  def up
+    safety_assured do
```
benhalpern, line 3: I added this to get this to stop yelling at me after the generated migration had issues.... but this isn't the clean way to deal with this. I'll come back to this.

app/views/internal/invitations/index.html.erb
```diff
@@ -0,0 +1,35 @@
+<div class="row my-3">
+  <div class="col">
+    <ul class="nav nav-pills">
+      <li class="nav-item">
+        <%= link_to "Registered Users", "/internal/users", class: "nav-link" %>
+      </li>
+      <li class="nav-item">
+        <%= link_to "New", "/internal/invitations/new", class: "nav-link" %>
```
citizen428, line 8:
```suggestion
        <%= link_to "New", new_internal_invitation_path, class: "nav-link" %>
```

app/views/internal/invitations/index.html.erb
```diff
@@ -0,0 +1,35 @@
+<div class="row my-3">
+  <div class="col">
+    <ul class="nav nav-pills">
+      <li class="nav-item">
+        <%= link_to "Registered Users", "/internal/users", class: "nav-link" %>
```
citizen428, line 5:
```suggestion
        <%= link_to "Registered Users", internal_users_path, class: "nav-link" %>
```

db/migrate/20200712150048_devise_invitable_add_to_users.rb
```diff
@@ -0,0 +1,32 @@
...
+      change_table :users do |t|
+        t.string     :invitation_token
+        t.boolean    :registered, default: true
```
citizen428, line 6: **question:** Adding default values on big tables can be fairly disruptive, have we considered this?

benhalpern: The `users` table isn't _that_ big at ~425k rows or so. And from my understanding this is much more problematic with the modification of existing columns vs adding new ones. I _think_ we're good here.

db/migrate/20200712150048_devise_invitable_add_to_users.rb
benhalpern: Actually in looking at this I think it's fine to use here. This seems pretty safe and straightfoward I think.

app/helpers/application_helper.rb
benhalpern: ✅

db/schema.rb
```diff
@@ -347,6 +347,7 @@
     t.datetime "created_at", null: false
     t.string "description"
     t.boolean "discoverable", default: false
+    t.string "invitation_slug"
```
mstruve, line 350: This should not be here, its from the chat channels changes PR

app/workers/metrics/record_daily_usage_worker.rb
```diff
@@ -27,7 +27,7 @@ def perform
       DatadogStatsClient.count("articles.first_past_24h", first_articles_past_24h, tags: ["resource:articles"])
 
       # Users who signed up in the past 24 hours who have made at least 1 comment so far
-      new_users_min_1_comment_past_24h = User.where("comments_count >= ? AND created_at > ?", 1, 24.hours.ago).size
+      new_users_min_1_comment_past_24h = User.where("comments_count >= ? AND registered_at > ?", 1, 24.hours.ago).size
```
mstruve, line 30: Since we are going to start using this `registered_at` column right away we need to make sure we update it to match, I would think, `created_at` for all existing users.

db/schema.rb
benhalpern: Guess I got my wires crossed a bit 😄

benhalpern, line 350:
```suggestion
```

app/workers/metrics/record_daily_usage_worker.rb
benhalpern: Yeah, my rationale for not including the data file was that it was just one more thing in a big PR and this functionality only gets triggered once per day. One less moving piece within this pull request.

How about this: After this gets merged, I whip up the data script as a separate PR?

app/views/layouts/_signup_modal.html.erb
```diff
@@ -9,19 +9,22 @@
                   class: "mascot-image",
                   alt: SiteConfig.mascot_image_description,
                   loading: "lazy") %>
+    <% if any_selfserve_auth? %>
```
rhymes, line 12: what does "self serve" mean?

app/workers/metrics/record_daily_usage_worker.rb
benhalpern: @mstruve I created #9331 to address this.

Let me know if I need to do anything differently... I actually haven't made a data script yet so I'm not sure I need to do anything else. (Docs are super great though 😄)

db/migrate/20200712150048_devise_invitable_add_to_users.rb
rhymes: should we add an index on `registered` ? The code is doing a lot of `.where(registered: true|false)` queries

app/workers/metrics/record_daily_usage_worker.rb
mstruve: Works for me!

db/migrate/20200712150048_devise_invitable_add_to_users.rb
benhalpern: @rhymes My understanding is that indexes on booleans wouldn't add a lot of value.

> PostgreSQL will use an index only if it thinks it will be cheaper that way. An index on a boolean column, which can only take two possible values, will almost never be used, because it is cheaper to sequentially read the whole table than to use random I/O on the index and the table if a high percantage of the table has to be retrieved.

[source](https://stackoverflow.com/questions/42972726/postgresql-create-index-for-boolean-column)

rhymes: TIL! :-)

citizen428: > And from my understanding this is much more problematic with the modification of existing columns vs adding new ones.

I agree with this statement, but I think it never hurts to bring it up in PR reviews.

PR #9297, opened 2020-07-13
Fix TypeError: null is not an object in initializeArticleReactions.js

vaidehijoshi:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
As [this Honeybadger error](https://app.honeybadger.io/fault/67192/6f9b4fdf475e2229c968fa26d785b90f) shows, we sometimes try to call `textContent` on an element that is `null`. My guess is that this error happens at some point when `initializeArticleReactions.js` is called, and when the DOM attempts to `getElementById("reaction-number-"+reactionName)`, but can't find that element on the page/`document` yet.

Since we can't continue with that function without the element existing, I'm adding as guard clause here, which should fix this particular error.

Should fix https://app.honeybadger.io/fault/67192/6f9b4fdf475e2229c968fa26d785b90f

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

new bugs

### DIFF
--- a/app/assets/javascripts/initializers/initializeArticleReactions.js
+++ b/app/assets/javascripts/initializers/initializeArticleReactions.js
@@ -41,13 +41,10 @@ function hasUserReacted(reactionName) {
 
 function getNumReactions(reactionName) {
   const reactionEl = document.getElementById('reaction-number-' + reactionName);
-
-  if (!reactionEl) {
-    return;
-  }
-  if (reactionEl.textContent === '') {
+  if (!reactionEl || reactionEl.textContent === '') {
     return 0;
   }
+
   return parseInt(reactionEl.textContent, 10);
 }
 
--- a/app/assets/javascripts/initializers/initializeArticleReactions.js
+++ b/app/assets/javascripts/initializers/initializeArticleReactions.js
@@ -40,12 +40,15 @@ function hasUserReacted(reactionName) {
 }
 
 function getNumReactions(reactionName) {
-  var num = document.getElementById('reaction-number-' + reactionName)
-    .textContent;
-  if (num === '') {
+  const reactionEl = document.getElementById('reaction-number-' + reactionName);
+
+  if (!reactionEl) {
+    return;
+  }
+  if (reactionEl.textContent === '') {
     return 0;
   }
-  return parseInt(num, 10);
+  return parseInt(reactionEl.textContent, 10);
 }
 
 function reactToArticle(articleId, reaction) {


app/assets/javascripts/initializers/initializeArticleReactions.js
```diff
@@ -40,12 +40,15 @@ function hasUserReacted(reactionName) {
...
 
 function getNumReactions(reactionName) {
-  var num = document.getElementById('reaction-number-' + reactionName)
-    .textContent;
-  if (num === '') {
+  const reactionEl = document.getElementById('reaction-number-' + reactionName);
+
+  if (!reactionEl) {
+    return;
+  }
```
rhymes, line 47: ❤️ defensive programming is a must in DOM based JavaScript :D

app/assets/javascripts/initializers/initializeArticleReactions.js
```diff
@@ -40,12 +40,15 @@ function hasUserReacted(reactionName) {
 }
 
 function getNumReactions(reactionName) {
-  var num = document.getElementById('reaction-number-' + reactionName)
-    .textContent;
-  if (num === '') {
+  const reactionEl = document.getElementById('reaction-number-' + reactionName);
+
+  if (!reactionEl) {
```
Ridhwana, line 45: Looking at these functions, if we have this code:
```
if (!reactionEl) {
    return;
 }
```

and we're using `getNumReactions` in the following way from `reactToArticle`:
```
function reactToArticle(articleId, reaction) {
  // Visually toggle the reaction
  function toggleReaction() {
    var currentNum = getNumReactions(reaction);
    if (hasUserReacted(reaction)) {
      hideUserReaction(reaction);
      setReactionCount(reaction, currentNum - 1);
    } else {
      showUserReaction(reaction, 'user-animated');
      setReactionCount(reaction, currentNum + 1);
    }
  }
```
currentNum seems like it would be undefined because we return early if !reactionEl, hence won't it cause the following evaluation `undefined - 1`?

vaidehijoshi: Yeah, I was wondering to myself whether we want to allow the error to bubble up to `reactToArticle`, or whether we just want `getNumReactions` to return `0` if there isn't a `reactionEl ` (to keep the math working). I think probably the latter makes sense here, now that I look at the code again.

vaidehijoshi: @Ridhwana Updated this to just return `0` in the latest commit, which is what we do if the `reactionEl`'s `textContent` is an empty string in any case 🙃

PR #9298, opened 2020-07-14
Fix TypeError: userData is null in initializeOnboardingTaskCard.js

vaidehijoshi:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
As [this Honeybadger error](https://app.honeybadger.io/fault/67192/65045742fda3390a067960f53a894e87) shows, we sometimes try to call `created_at` on `userData`, which can occasionally be `null`. My guess is that this error happens sometimes when we call `initializeOnboardingTaskCard` but for some reason (?) the `userData()` isn't there.

This particular error started occurring about 5 days ago, and seems to only be occurring on Firefox.

I suspect that there are times when the `document.body.dataset` doesn't have the `user` within it (for some reason?), and so we are unable to `JSON.parse(user)`, and then this function causes the subsequent `TypeError`. 

Probably, we should have some kind of better, global catchall system for handling the situations where `userData` doesn't exist on the page. But at a bare minimum, I don't think we want to continue execution of this function.

Should fix https://app.honeybadger.io/fault/67192/65045742fda3390a067960f53a894e87

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

oh dear dog

### DIFF
--- a/app/assets/javascripts/initializers/initializeOnboardingTaskCard.js
+++ b/app/assets/javascripts/initializers/initializeOnboardingTaskCard.js
@@ -10,13 +10,9 @@ function initializeOnboardingTaskCard() {
   }
 
   var taskCard = document.getElementsByClassName('onboarding-task-card')[0];
-  if (taskCard == null) {
-    return; // This guards against both a null and undefined taskCard.
-  }
-
   const user = userData();
-  if (!user) {
-    return;
+  if (taskCard == null || !user) {
+    return; // Guard against a null/undefined taskCard, and no user data.
   }
 
   var createdAt = new Date(user.created_at);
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/initializers/initializeOnboardingTaskCard.js
```diff
@@ -11,10 +11,15 @@ function initializeOnboardingTaskCard() {
 
   var taskCard = document.getElementsByClassName('onboarding-task-card')[0];
   if (taskCard == null) {
+    return; // This guards against both a null and undefined taskCard.
+  }
+
+  const user = userData();
+  if (!user) {
```
Ridhwana, line 18: Can we not just move this into an or statement, i.e. :
```
var taskCard = document.getElementsByClassName('onboarding-task-card')[0];
const user = userData();
if (taskCard == null ||  !user) {
  return;
}
```

vaidehijoshi: We can -- I clearly had a fried brain when I wrote this last night 😓 Updated in the latest commit!

vaidehijoshi:
Marked [this](https://app.honeybadger.io/projects/67192/faults/65501154) error as resolved!

PR #9299, opened 2020-07-14
Fix TypeError: userData is null in blockButton.js

vaidehijoshi:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
As [this Honeybadger error](https://app.honeybadger.io/fault/67192/ed6f6b392b5bd791d6dc04732f125123) shows, we sometimes try to call `id` on `userData`, which can occasionally be `null`.

This particular error seems to really only happen to one user, who is on Firefox. I'm not sure why `userData` doesn't seem to exist for this user on this browser specifically (and that too from `initBlock.js`?), but it seems like we shouldn't be throwing a TypeError here if the `userData` object doesn't even exist. 🤷 

As I mentioned in https://github.com/thepracticaldev/dev.to/pull/9298, we probably need to investigate the non-existant user data set on the page, but for now, this should bring down some of the high error counts for _this_ error that we've been seeing on the JS Honeybadger dashboard.

Should fix https://app.honeybadger.io/fault/67192/ed6f6b392b5bd791d6dc04732f125123

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

oh no panda

### DIFF
--- a/app/javascript/profileDropdown/blockButton.js
+++ b/app/javascript/profileDropdown/blockButton.js
@@ -24,8 +24,8 @@ export default function initBlock() {
         },
       }),
     })
-      .then(response => response.json())
-      .then(response => {
+      .then((response) => response.json())
+      .then((response) => {
         if (response.result === 'unblocked') {
           blockButton.innerText = 'Block';
           /* eslint-disable-next-line no-use-before-define */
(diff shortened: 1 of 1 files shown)


vaidehijoshi:
Marked [this](https://app.honeybadger.io/projects/67192/faults/65290784) error as resolved!

PR #9300, opened 2020-07-14
Introduce domain model for campaigns

citizen428:
_I tagged entirely too many people in this PR, please don't feel like you have to comment. My intention was just to make people aware of the idea and to maybe spark a general discussion about ensuring that important domain concepts don't fall through the cracks but get proper representation._

- [X] Refactor

## Description

Campaigns are clearly a concept in our app's domain, as we quite frequently talk about them. However, so far this domain concept had no actual representation in code but instead existed "between the lines" in `SiteConfig` and code in the view layer. This PR introduces a plain Ruby object (PORO) to remedy that. This not only allows us to treat campaigns as first-class citizens in code but also to use niceties like decorators, etc.

**Notes**

1. This was a 1h spike, so I did not change all instances of `SiteConfig.campaign_*` yet. Instead, I changed just a little bit to demonstrate the approach and will change the rest if we agree to move forward with this. The same for tests.
2. I personally always treated `app/models` as the home for all domain models, whether they are DB-backed or not (which IMHO is an implementation detail). Some people seem to be uncomfortable with having non-AR models there. I can move it if people feel better about it, but welcome suggestions regarding the new location.
3. I won't insist on the singleton pattern, but from what I can tell we can only have one active campaign at a time, so it made sense to me.

n/a

Configure a campaign via `/internal/config`:

Screen Shot 2020-07-14 at 10 01 35

Make sure to tick the checkbox next to `Campaign sidebar enabled` and have a look at the sidebar, both with and without a campaign URL set.

## Added tests?

- [ ] yes
- [ ] no, because I need help
- [x] not yet because spike

## Added to documentation?

- [X] Code comments


app/decorators/campaign_decorator.rb
```diff
@@ -0,0 +1,17 @@
+class CampaignDecorator < ApplicationDecorator
+  include ActionView::Helpers::AssetTagHelper
```
citizen428, line 2: I included these as the most convenient way to get access to `image_tag` and `link_to`. We can probably be smarter about this if it bothers people.

app/models/campaign.rb
```diff
@@ -0,0 +1,27 @@
...
+    hero_html_variant_name
+    featured_tags
+    sidebar_enabled?
+    sidebar_image
+    url
+    articles_require_approval?
+  ].freeze
+
+  # Define delegate methods for SiteConfig
+  METHODS.each { |m| define_method(m) { SiteConfig.public_send("campaign_#{m}") } }
```
citizen428, line 22: Note that we're only defining reader methods, no writers. Campaigns should be configured through `/internal/config` exclusively.

app/models/campaign.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def active?
```
citizen428, line 24: As I'm not super familiar with campaigns someone who is should comment on this. Is being enabled in the sidebar the same as being active? If not, do we do anything campaign-related when `SiteConfig.campaign_sidebar_enabled` is `false`?

app/views/articles/_sidebar_campaign.html.erb
```diff
@@ -1,21 +1,16 @@
+<% campaign = CampaignDecorator.new(Campaign.current) %>
```
citizen428, line 1: Used for simplifying the handling of the sidebar image and the 3 uses of the main tag (the first of the featured tags).

app/decorators/campaign_decorator.rb
```diff
@@ -0,0 +1,17 @@
...
+  include ActionView::Helpers::UrlHelper
+
+  def sidebar_image
+    return unless active?
+
+    img = image_tag(object.sidebar_image, class: "block w-100 h-auto radius-default", width: 1000, height: 420)
```
atsmith813, line 8: **question/nitpick:** how do we feel about including this markup in the decorator? Will it be confusing for future debugging to find where these classes and styles are being applied since they're not in the view?

app/models/campaign.rb
```diff
@@ -0,0 +1,27 @@
+# This "model" is not backed by the database. Its main purpose is giving one of
+# our domain concepts an actual representation in code.
+class Campaign
+  include Singleton
```
atsmith813, line 4: **thought:** I've always been under the impression of "don't use Singletons", but this setup is _sweet_ 🎉 . Since `SiteConfig` is, more or less DB driven, I don't know of a better way to get use of this "global state". We can't really use an initializer since that would require an app re-boot. It doesn't seem like re-writing a global constant would be better either. This just might be my new go-to reference for a solid use of Singletons 💯

atsmith813:
**praise:** I'm a fan 💯 . I left a couple of comments above and I wasn't sure if you were looking for a "formal approval" yet since this is still a draft.

app/decorators/campaign_decorator.rb
citizen428: Excellent point @atsmith813, I didn't think about this during my spike. The decorator now takes an options hash as argument and passes it through to `image_tag`:

```ruby
<%= campaign.sidebar_image(class: "block w-100 h-auto radius-default", width: 1000, height: 420) %>
```

app/models/campaign.rb
citizen428: > I've always been under the impression of "don't use Singletons",

1. Know the rules.
2. Know when to disregard them.
3. Profit.

As I hinted at in the PR description, I'm not super tied to the singleton pattern here, but this object is essentially a read-only "view" into a DB table with some behavior attached. I could have just made it a module and accessed everything through `Campaign.method`, but then we'd essentially have an implicit singleton and then I'd rather be explicit about it.

citizen428:
>  I wasn't sure if you were looking for a "formal approval" yet since this is still a draft

Not really. It's a somewhat unusual approach (though one that served me well in the past), so the idea was to spike out something quick and get feedback. I'm not tied to a specific implementation (e.g. the singleton), but I _really_ believe domain concepts should have corresponding code entities and not only exist as emergent behaviors.

app/models/campaign.rb
mstruve: > Know the rules.
Know when to disregard them.
Profit.

Can I Tweet this? It's gold! 😂

app/models/campaign.rb
```diff
@@ -0,0 +1,27 @@
+# This "model" is not backed by the database. Its main purpose is giving one of
+# our domain concepts an actual representation in code.
+class Campaign
```
mstruve, line 3: Given this is not an AR model, should we maybe put it in services? I'm fine with it here but wasn't sure if we wanted to keep these to AR models or not.

joshpuetz:
> I tagged entirely too many people in this PR, please don't feel like you have to comment. My intention was just to make people aware of the idea and tp maybe spark a general discussion about ensuring that important domain concepts don't fall through the cracks but get proper representation.

**Very** much appreciated @citizen428 !

app/models/campaign.rb
joshpuetz: I much prefer this class being in model as it's clearly solving the problem of missing domain object. There's also prior art for non-AR backed classes in `/models` in `language.rb`

app/models/campaign.rb
citizen428: Sure, go ahead @mstruve 😃

app/models/campaign.rb
citizen428: Here's how I generally think about this:

* `app/model`: Nouns, e.g. `User`, `Article`. Whether or not a model is DB-backed doesn't matter for this.
* `app/services`: Verbs, e.g. `User::Delete`, `Article::BustCache` (made-up examples). Actions performed on those nouns.

We're currently not quite there, as our services often use noun-names too, but for me, this separation has worked well in the past. I also tend to namespace services by the object the operate on as in the examples given above.

But I'm fine with everything the team agrees on.

citizen428:
Based on feedbach here and on Slack it seems that overall people like this, so I will go ahead and make more changes now.

citizen428:
Note: I did not change `app/views/internal/configs/show.html.erb`, as this deals with updating the `SiteConfig` itself, so it seemed more consistent to reflect that in the code.

citizen428:
Go home CodeClimate, you're drunk. 

image

The issue it complains about is in a file this PR didn't touch.

citizen428:
Fixed the spec and changed `.active?` to `.show_in_sidebar?`, which is more explicit about the intent.

PR #9303, opened 2020-07-14
Native bridge video support

### LINKED ISSUE #8908: Video not showing on mobile. Tested on iOS only.
**Describe the bug**

When I navigate to Dev video and I choose to view any videos displayed, the player is rendered but the screen is black. Audio is coming through fine. When I fiddle with it, like go full screen and back again, sometimes the video renders and sometimes it doesn't.

**To Reproduce**

Open Dev
Go to Videos
Choose any video
Video doesn't render.

**Expected behavior**

Full video on launch of article

fdoxyz:

- [x] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The native video player integration to work with the mobile apps. The PR includes the following:
- Native calls are executed based on platform (native video player will stream on Android + iOS apps)
- Article decorator added for video metadata (similar to how it's handled for Podcasts)
- JS in `app/views/articles/_video_player.html.erb` got too long once added logic to support the native players, so this was extracted into a separate JS initializer

Fixes #8908 

Triggers the feature implemented in [this iOS repo PR](https://github.com/thepracticaldev/DEV-ios/pull/229)

In order to test locally you'll need an article with valid video data. In order to do this I took a sample video article and manually populated an existing article from my local server. 

You can do this in `/admin` dashboard by adding these fields to an existing article with this sample data:

| Video  | Video code  | Video source url  | Video thumbnail url  |
|---|---|---|---|
| https://s3.amazonaws.com/dev-to-input-v0/video-upload__2f321f7af545f8682dc2623af19962a4  | video-upload__2f321f7af545f8682dc2623af19962a4  | https://dw71fyauz7yz9.cloudfront.net/video-upload__2f321f7af545f8682dc2623af19962a4/video-upload__2f321f7af545f8682dc2623af19962a4.m3u8  |  https://dw71fyauz7yz9.cloudfront.net/video-upload__2f321f7af545f8682dc2623af19962a4/thumbs-video-upload__2f321f7af545f8682dc2623af19962a4-00001.png |

There should be no difference when browsing from a web browser, but when using either the iOS/Android app the `jwplayer`
should be bypassed in favor of sending messages with the Native Bridge functionalities of each platform.

This is what the player looks like in iOS:

ios-video-integration-3

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

tricky


app/assets/javascripts/initializers/initializeVideoPlayback.js
```diff
@@ -0,0 +1,193 @@
...
+        }
+      }
+    } else if(isNativeAndroid()) {
+      nativeBridgeMessage = function(message) {
+        try {
+          AndroidBridge.videoMessage(JSON.stringify(message));
+        } catch (err) {
+          console.log(err.message); // eslint-disable-line no-console
+        }
+      }
```
fdoxyz, line 160: @citizen428 [I remember your suggestion](https://github.com/thepracticaldev/dev.to/pull/8054#discussion_r430210129) from last time I worked on native bridge JS. These checks are performed once and from then on the function assigned to `nativeBridgeMessage` is called directly.

I know in this case the effect is not as impactful as in the Podcast code (much more complex), but let me know what you think about this implementation of the strategy pattern. I think I can squeeze in a refactor to the podcast code in a separate PR to use this and avoid all those _if checks_ before our cooldown is over 🙂

app/assets/javascripts/initializers/initializeVideoPlayback.js
```diff
@@ -0,0 +1,193 @@
+/**
+ * This script hunts for video tags and initializes the correct player
+ * depending on the platform:
+ * - web: jwplayer
+ * - iOS/Android: Native player
+ *
+ * Once jwplayer is initialized there's no follow up actions to be taken.
+ * Mobile Native players send back information into the DOM in order to
+ * interact and update the UI, therefore a MutationObserver is registered.
+ */
```
joshpuetz, line 10: 👏 Fantastic inline commenting!

app/assets/javascripts/initializers/initializeVideoPlayback.js
```diff
@@ -0,0 +1,193 @@
...
+      navigator.userAgent === 'DEV-Native-android' &&
+      typeof AndroidBridge !== 'undefined' &&
+      AndroidBridge !== null &&
+      AndroidBridge.videoMessage !== undefined
+    );
+  }
+
+  function getParameterByName(name, url) {
+    if (!url) url = window.location.href;
+    name = name.replace(/[\[\]]/g, "\\$&");
```
fdoxyz, line 46: CodeQL is raising a warning here. This code was not modified from the current implementation (maybe wasn't caught before because it lived inline `_video_player.html.erb`).

Instead of fiddling with the regexp (and possibly breaking it) left it unchanged. If this is something we want to change I'd be happy to hear suggestions/comments on this

maestromac: Let's not touch it then. We also just removed CodeQL for the time being.

app/decorators/article_decorator.rb
```diff
@@ -70,4 +70,14 @@ def description_and_tags
 
     modified_description + " Tagged with #{cached_tag_list}."
   end
+
+  def video_metadata
```
maestromac, line 74: this is neat 👍

spec/decorators/article_decorator_spec.rb
```diff
@@ -183,4 +183,39 @@ def create_article(*args)
...
+      caption_track_url = video_data[:video_closed_caption_track_url]
+
+      allow(article_with_video).to receive(:video_code).
+        and_return(video_data[:video_code])
+      allow(article_with_video).to receive(:video_source_url).
+        and_return(video_data[:video_source_url])
+      allow(article_with_video).to receive(:cloudinary_video_url).
+        and_return(video_data[:video_thumbnail_url])
+      allow(article_with_video).to receive(:video_closed_caption_track_url).
+        and_return(caption_track_url)
```
maestromac, line 209: I don't think stubbing method ActiveRecord functions create a dependable test. You can simplify this whole process by generated articles with video links.

```
article_with_video = create(:article, video_code: "ABC", video_source_url: "https://cdn.com/ABC.m3u8", video_thumbnail_url: "https://cdn.com/ABC.png", video_closed_caption_track_url: "https://cdn.com/ABC_captions")
expect(article_with_video.decorate.video_metadata).to eq({
  id: article_with_video.id,
  ...
})
```

fdoxyz: This is so true! Just updated the spec to implement this

app/assets/javascripts/initializers/initializeVideoPlayback.js
citizen428: @fdoxyz I like it! 😃 As you said yourself, it's probably more impactful for podcasts, but this is a nice trial run for the idea.

app/assets/javascripts/initializers/initializeVideoPlayback.js
```diff
@@ -0,0 +1,193 @@
...
+    if (!results) return null;
+    if (!results[2]) return '';
+    return decodeURIComponent(results[2].replace(/\+/g, " "));
+  }
+
+  function timeToSeconds(hms) {
+    var a;
+    if (hms.length < 3) {
+      return hms
+    } else if (hms.length < 6) {
```
citizen428, line 58: **thought:** At one point we may want to look into making this function a bit more robust. `1:2:3` for example has a lenght of less than 6 but `+a[1]` would result in `NaN`. Disregard this comment if we have full control over the formatting of `hms` and this can never happen.

app/decorators/article_decorator.rb
citizen428, line 74: **suggestion (non-blocking):** A potential alternative with less repetition:

```ruby
def video_metadata
  as_json(only: [:id, :video_code, :video_source_url, :video_thumbnail_url, :video_closed_caption_track_url])
end
```

In action (I didn't have an article with the right attributes in my local DB):

```ruby
d = ArticleDecorator.new(a)
d.video_metadata
#=>
{
                              "id" => 52,
  "video_closed_caption_track_url" => nil,
                      "video_code" => nil,
                "video_source_url" => nil,
             "video_thumbnail_url" => nil
}
```

app/assets/javascripts/initializers/initializeVideoPlayback.js
```diff
@@ -0,0 +1,193 @@
...
+
+function initializeVideoPlayback() {
+  var nativeBridgeMessage;
+  var currentTime = "0";
+
+  function getById(name) {
+    return document.getElementById(name);
+  }
+
+  function isNativeIOS() {
```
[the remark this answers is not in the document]
fdoxyz: I love this suggested structure! I will definitely send in a PR with this approach to support both Podcast & refactor this Video feature soon. Rather do it in a separate PR so we can test and ship the associated mobile PRs, in the end the functionality will remain unchanged.

Also "Fernandroid" > "iFernando" 😆

app/decorators/article_decorator.rb
fdoxyz: I like the cleaner look, but is there a way to associate `video_thumbnail_url` to `cloudinary_video_url` method using this approach? Because we're not rendering the exact value of `video_thumbnail_url`, we are using the `cloudinary_video_url` method of the Article.

Maybe another enhancement to be added in the follow up PR

app/assets/javascripts/initializers/initializeVideoPlayback.js
fdoxyz: I think we do have full control over this, mainly because I couldn't find anywhere we share a "timestamped URL of the article".

What I mean by this is that I believe this is an open ended "hidden feature", as in I can share a link [like this one starting at 1:40](https://dev.to/terminusdb/world-of-woqlcraft-ep-16-knowledge-graph-for-nlp-1jhc?t=1:40) and it will work, but we don't have any built-in share feature that will include the current time of the player.

I just sent in a tweak for better supporting this timestamp parameter but will definitely put some more safety nets in this method in the follow up 👍🏼

app/decorators/article_decorator.rb
citizen428: @fdoxyz Oh, did I miss that one attribute gets renamed?  

In that case:

```ruby
hash = as_json(..., :cloudinary_video_url)
hash[:video_thumbnail_url] = hash.delete(:cloudinary_video_url)
hash

# or shorter
as_json(..., :cloudinary_video_url).tap |hash|
  hash[:video_thumbnail_url] = hash.delete(:cloudinary_video_url)
end
```

app/assets/javascripts/initializers/initializeVideoPlayback.js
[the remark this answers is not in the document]
maestromac: That's a great suggestion @citizen428 !

Issue #8976, opened 2020-06-29
Allow posts listed in Mod Center to open in a new tab


Turnerj:
**Is your feature request related to a problem? Please describe.**

Currently in the Mod center, when you click a post it opens inline underneath the post you clicked. This is really nice BUT it can be useful opening a bunch of posts into multiple tabs and handle it that way.

Because the post items are handled by JS magic, you can't actually open them into a new tab.

**Describe the solution you'd like**

So my thought is keep the inline JS stuff but have the post items be anchors so when we Ctrl+Click etc, it can open in a new tab. I'd probably go a bit further and make the link include a query string parameter or fragment which also indicates that the mod tools should open up by default on that link.

jacobherrington:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

jacobherrington:
I played around with this a little bit. I think the best solution is to add a link icon next to the title:
![image](https://user-images.githubusercontent.com/11466782/86133062-697b0580-baad-11ea-92fb-982e35b81774.png)

I did the above pretty quickly by changing the following lines:
https://github.com/thepracticaldev/dev.to/blob/56736aab1527c301b73fffce50b085c044145de4/app/javascript/modCenter/singleArticle/index.jsx#L53-L55

To something like this:

```javascript
import ExternalLinkIcon from 'images/external-link.svg';
...
<h3 className="fs-base fw-bold lh-tight">
  {title}
  <a href={path}>
    <span className="crayons-icon">
      <img src={ExternalLinkIcon} aria-label="External Link" />
    </span>
  </a>
</h3>
```

Turnerj:
Interesting - it does give the functionality though looking at the `index.jsx` markup, I'm not sure how well putting an anchor in a button would turn out (I didn't realise it was a button tag for the whole post before).

If it does work and doesn't seem too weird, I think it might be a little confusing as my first thought is that the whole title would be a link then. This would follow the similar design choice as shown on the "Listing" link in the dashboard: 
[...]

PR #9314, opened 2020-07-14
Mod Center: clicking on article title opens a in new tab

PrimeNoodles:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Was able to create a clickable title on the modCenter page where clicking on the article title would open a new tab to the the article. Clicking on the label will bring up the original inline article mod.
Closes https://github.com/forem/forem/issues/8976

I was also able to get a clickable link on the author to work, which seemed to have no errors on my local host end. However, when trying to commit and run the test cases on singleArticle.text.jsx, it wasn't able to pass one of the test on line 73 to 77 where it renders the author's name. I have commented the code out if anyone can look and see what the problem is.

I also wasn't able to implement @jacobherrington suggestion of having a external link icon simply because I wasn't able to import the externalicon.svg, maybe it's connect to the crayon system?

ModWindow

ModWindowDark

ModWindowMinLight

ModWindowPink

ModWindowTenXHacker

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed


app/javascript/modCenter/singleArticle/index.jsx
```diff
@@ -51,13 +59,16 @@ export default class SingleArticle extends Component {
...
+            <h3 className="fs-base fw-bold lh-tight">
+              <a href={path}>{title}</a>
+            </h3>
           </header>
           {tags}
         </span>
         <span className="article-author fs-s lw-medium lh-tight">
           {newAuthorNotification}
           {user.name}
+          {/* <a href={user.username}>{user.name}</a> */}
```
jacobherrington, line 71:
```suggestion

```
I think this should be removed? Or should this still be a link to the author's profile?

jacobherrington:
Thanks for making a PR! This looks good! :tada: 

I noticed a commented out line and had a question about it.

app/javascript/modCenter/singleArticle/index.jsx
PrimeNoodles: So while in testing on my own local host 3000, I was able to click on the author's name which would then take me to the author's profile. Problem is, while committing the file it didn't pass the test cases in singleArticle.text.jsx, specifically the one on lines 73 to 77 where it renders the author's name. I just commented it out and put the {user.name} back on line 70. Otherwise you would replace line 70 with 71 if it works.

PrimeNoodles: I've committed a newer version removing the comment.

PrimeNoodles:
I made a test case to see if the article title matches the title given in the href tag, but I'm not sure if this would be valid enough for a test case.

Zhao-Andy:
Thanks for the PR! Looks like you have merge conflicts now unfortunately.

@lisasy, do you think the link color should be there? Or should it be the color of regular text? I think it probably makes sense to have the link color now that the title has become a link.

Also @PrimeNoodles do you mind checking if the titles look okay with different themes?

PrimeNoodles:
@Zhao-Andy I've updated the original post with examples of every theme, seems to be working fine, also some themes seem to have their own link color. I'm trying to fix this problem with codeclimate and the merge conflicts.

thomasbnt:
Oh very useful thank you! 🙌🏼🌱

sachudj (approved):

lisasy:
This looks good to me!

app/javascript/modCenter/singleArticle/__tests__/singleArticle.test.jsx
```diff
@@ -46,6 +46,20 @@ describe('<SingleArticle />', () => {
...
+    const { container } = render(
+      <>
+        <SingleArticle {...getTestArticle()} toggleArticle={jest.fn()} />
+        <div
+          data-testid="flag-user-modal"
+          class="flag-user-modal-container hidden"
+        />
+      </>,
+    );
+    const text = getNodeText(container.querySelector('.article-title-link'));
```
nickytonline, line 59: refactor (non-blocking): You can use preact-testing-library's [getByText](https://testing-library.com/docs/dom-testing-library/api-queries#bytext) instead of using DOM APIs directly.

question: Is the `<div />` in the render necessary for the test?

```jsx
<div
    data-testid="flag-user-modal"
    class="flag-user-modal-container hidden"
/>
```

nickytonline (approved):
Thanks for adding some tests @PrimeNoodles. Left a suggestion.

app/javascript/modCenter/singleArticle/__tests__/singleArticle.test.jsx
rhymes: @nickytonline as the OP hasn't responded in a week and as this PR is months old, do you want to finalize this change or shall we merge as is?

nickytonline: We can merge it as it's non-blocking.

PR #9319, opened 2020-07-14
Bug Fix: Validate GIFs Frames and Add MiniMagick Timeout

mstruve:
- [x] Bug Fix

## Description
This will prevent uploading of large GIFs from timing out and set a limit to how many frames they can have.

NOTE: `FRAME_MAX` is actually 1000 but I lowered it for testing locally. 
Screen Shot 2020-07-14 at 4 57 09 PM

## Added tests?
- [x] yes

alt_text


config/initializers/carrierwave.rb
```diff
@@ -1,3 +1,11 @@
+require "mini_magick"
+
+# Carrierwave uses MiniMagick for image processing. To prevent server tmeouts
+# we are setting the MiniMagick timeout lower.
+MiniMagick.configure do |config|
+  config.timeout = 10
```
mstruve, line 6: Totally open to other timeout values if anyone has a suggestion!

app/uploaders/base_uploader.rb
```diff
@@ -2,6 +2,10 @@ class BaseUploader < CarrierWave::Uploader::Base
   include CarrierWave::BombShelter # limits size to 4096x4096
   include CarrierWave::MiniMagick # adds processing operations
 
+  FRAME_MAX = 1000
```
mstruve, line 5: I had a 768 frame gif that I used to test all of this and that GIF is really long. I don't see anyone needing more than 1000 frames, and worse case, if they do, they can upload it elsewhere and link to it.

atsmith813:
**question:** Is the expected behavior that it displays the friendly error in your screenshot?

I triggered the newly lowered timeout limit when I tested this locally:
Screen Shot 2020-07-14 at 11 01 25 PM
If we want the "pretty error", maybe we rescue there and return it?

config/initializers/carrierwave.rb
rhymes: ahahha if it can't open an image in 10 seconds we should definitely bail :D

config/initializers/carrierwave.rb
```diff
@@ -1,3 +1,11 @@
+require "mini_magick"
+
+# Carrierwave uses MiniMagick for image processing. To prevent server tmeouts
```
rhymes, line 3: typo in `timeouts`

mstruve:
Great suggestion @atsmith813!!!
 
Screen Shot 2020-07-15 at 8 55 53 AM

PR #9320, opened 2020-07-14
Attempt to fix Flipper-UI

joshpuetz:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Another try at fixing ongoing issue with the Flipper/Feature Flag UI (https://github.com/thepracticaldev/tech-private/issues/450). After some investigation: I found that omitting the `Origin` http header when using curl allowed me to successfully POST to production. Digging through both Flipper's forums and Molly's similar Sidekiq fix from last year, I suspect something is happening with the CSRF protection FlipperUI is getting from Sinatra.

(Related: I had to pin the `redis-store` gem at 1.9.0 to work around an issue noted here:  https://github.com/redis-store/redis-store/pull/333

https://github.com/thepracticaldev/tech-private/issues/450

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

alt_text


config/routes.rb
```diff
@@ -42,7 +42,13 @@
 
     authenticate :user, ->(user) { user.has_role?(:tech_admin) } do
       mount Blazer::Engine, at: "blazer"
-      mount Flipper::UI.app(Flipper, { rack_protection: {} }), at: "feature_flags"
+
+      flipper_ui = Flipper::UI.app(Flipper) do |builder|
+        builder.use Rack::Protection, origin_whitelist: ["https://dev.to"]
```
citizen428, line 47: **suggestion:** We should be able to leverage our `URL` helper module here, so we don't add anything new that later needs to be generalized again.

```suggestion
        builder.use Rack::Protection, origin_whitelist: URL.url
```

Gemfile
```diff
@@ -85,6 +85,7 @@ gem "recaptcha", "~> 5.5", require: "recaptcha/rails" # Helpers for the reCAPTCH
 gem "redcarpet", "~> 3.5" # A fast, safe and extensible Markdown to (X)HTML parser
 gem "redis", "~> 4.2.1" # Redis ruby client
 gem "redis-rails", "~> 5.0.2" # Redis for Ruby on Rails
+gem "redis-store", "~> 1.9.0" # Redis storage options for Rails (listed here to lock version above 1.9.0 - see https://github.com/redis-store/redis-store/pull/333)
```
rhymes, line 88: I don't think we need to add this to the Gemfile as it's not an explicit dependency, I ran `bundle update redis-store` on master and it did update it nonetheless.

Maybe we can move `see https://github.com/redis-store/redis-store/pull/333` and a short explanation about what's going on in the routes file

config/routes.rb
joshpuetz: This is a great suggestion!

Gemfile
joshpuetz: I swear version Redis-store 1.9 wasn't updating for me, but now that 1.9 is in `Gemfile.lock` I agree this isn't needed in the Gemfile anymore! I'll move this PR link to routes per your suggestion.

joshpuetz:
(Note that Github is having some issues showing newer commits to PR, you can see my latest here: https://github.com/forem/forem/compare/master...joshpuetz/flipper-ui-session-fix)

joshpuetz:
Update: it didn't work 😧 Info added to issue https://github.com/thepracticaldev/tech-private/issues/450

PR #9321, opened 2020-07-14
Re-enable Codeclimate's coverage support on Travis

maestromac:
- [X] DX

## Description
The general solution for test coverage to work with a parallel test is to upload each coverage to an s3 instance then combine them together. I can not go that route because it requires us to supply AWS keys which our contributors won't have access to. Instead I'm using Travis's beta feature, "workspace", to persist some files between each build.

https://docs.travis-ci.com/user/using-workspaces
n/a

## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a


maestromac:
holding this off for the time being because it will slow down our build time.

PR #9326, opened 2020-07-15
Make credit pricing configurable

citizen428:

- [X] Feature

## Description

Our credit prices so far were hardcoded (in cents). This PR makes the values configurable. I went with T-shirt sizing for the names of the limits, i.e. small, medium, large, xlarge (these is the number of credits purchased in a transaction). Note that I set the default in `SiteConfig` to the values currently used by DEV. They seem sensible and I don't want new Forem admins to feel forced to configure this.

Not in scope: making the limits themselves configurable, i.e. setting the limits for small, medium, etc. I'm not sure if we even want to do this, but if we do I'd rather make that follow-up PR.

n/a

Screen Shot 2020-07-17 at 11 34 15

## Added tests?

- [X] yes

## Added to documentation?

- [X] no documentation needed

### DIFF
--- a/app/controllers/credits_controller.rb
+++ b/app/controllers/credits_controller.rb
@@ -99,14 +99,17 @@ class CreditsController < ApplicationController
   end
 
   def cost_per_credit
-    if @number_to_purchase < 10
-      500
-    elsif @number_to_purchase < 100
-      400
-    elsif @number_to_purchase < 1000
-      300
+    prices = SiteConfig.credit_prices_in_cents
+
+    case @number_to_purchase
+    when ..9
+      prices[:small]
+    when 10..99
+      prices[:medium]
+    when 100..999
+      prices[:large]
     else
-      250
+      prices[:xlarge]
     end
   end
 end
--- a/app/controllers/internal/configs_controller.rb
+++ b/app/controllers/internal/configs_controller.rb
@@ -55,6 +55,7 @@ module Internal
         social_media_handles: SiteConfig.social_media_handles.keys,
         email_addresses: SiteConfig.email_addresses.keys,
         meta_keywords: SiteConfig.meta_keywords.keys,
+        credit_prices_in_cents: SiteConfig.credit_prices_in_cents.keys,
       )
     end
 
--- a/app/controllers/internal/configs_controller.rb
+++ b/app/controllers/internal/configs_controller.rb
@@ -70,6 +70,7 @@ module Internal
       %i[sidebar_tags suggested_tags suggested_users].each do |param|
         config[param] = config[param].downcase.delete(" ") if config[param]
       end
+      config[:credit_prices_in_cents]&.transform_values!(&:to_i)
     end
 
     def bust_relevant_caches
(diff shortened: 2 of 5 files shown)


benhalpern:
Good feature

app/controllers/credits_controller.rb
```diff
@@ -99,14 +99,11 @@ def generate_cost
...
   def cost_per_credit
-    if @number_to_purchase < 10
-      500
-    elsif @number_to_purchase < 100
-      400
-    elsif @number_to_purchase < 1000
-      300
-    else
-      250
+    case @number_to_purchase
```
msarit, line 102: Correct me if I am wrong, but wouldn't the overlap in these ranges cause issues?
I was thinking more:
```
case @number_to_purchase
  in 0...10 then 500
  in 11...100 then 400
  in 101...1000 then 300
  else 250
```

citizen428: @msarit Ranges with 3 dots don't include the upper bound (half-open intervals in math, i.e. `[lower, upper)`)

```ruby
(1..3).to_a
#=> [1, 2, 3]
(1...3).to_a
#=> [1, 2]
```

I could change it to use inclusive ranges (closed intervals, i.e. `[lower, upper]`)

```ruby
  in 0..9 then 500
  in 10..99 then 400
  in 100...999 then 300
  else 250
```

but I usually prefer the more "symmetric" version.

app/views/internal/configs/show.html.erb
```diff
@@ -181,6 +181,49 @@
...
+        <div class="card mt-3">
+          <%= render partial: "card_header",
+                     locals: {
+                       header: "Credits",
+                       state: "collapse",
+                       target: "creditsBodyContainer",
+                       expanded: "false"
+                     } %>
+
+          <div id="creditsBodyContainer" class="card-body collapse hide" aria-labelledby="creditsBodyContainer">
```
citizen428, line 193: I think this is good enough for now. The reason I explicitly list all 4 options is that I wanted to give some more context on each. Once we make the limits themselves configurable we can revisit this.

citizen428:
This still needs some work, will fix on Monday.

app/controllers/internal/configs_controller.rb
```diff
@@ -69,6 +70,7 @@ def clean_up_params
       %i[sidebar_tags suggested_tags suggested_users].each do |param|
         config[param] = config[param].downcase.delete(" ") if config[param]
       end
+      config[:credit_prices_in_cents]&.transform_values!(&:to_i)
```
citizen428, line 73: `ActiveSupport` adds some useful helpers, this is one of them.

app/controllers/credits_controller.rb
```diff
@@ -99,14 +99,17 @@ def generate_cost
...
-    if @number_to_purchase < 10
-      500
-    elsif @number_to_purchase < 100
-      400
-    elsif @number_to_purchase < 1000
-      300
+    prices = SiteConfig.credit_prices_in_cents
+
+    case @number_to_purchase
+    when ..9
```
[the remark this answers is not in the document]
Zhao-Andy: Interesting, does this include `nil`? I'd test this but I'm on my phone, and repl.it isn't on Ruby 2.7 yet.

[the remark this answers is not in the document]
citizen428: No, it doesn't:

```ruby
(..9).cover?(nil)
#=> false
```

PR #9331, opened 2020-07-15
[MERGE AFTER INVITATION PR] Add registered_at data script

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This backfills the `registered_at` value which we now want to use since we have the capacity to create non-registered users.

Merge after #9294

### DIFF
--- a/lib/data_update_scripts/20200715140848_backfill_user_registered_at.rb
+++ b/lib/data_update_scripts/20200715140848_backfill_user_registered_at.rb
@@ -1,0 +1,9 @@
+module DataUpdateScripts
+  class BackfillUserRegisteredAt
+    def run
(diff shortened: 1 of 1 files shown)


lib/data_update_scripts/20200715140848_backfill_user_registered_at.rb
```diff
@@ -0,0 +1,9 @@
+module DataUpdateScripts
+  class BackfillUserRegisteredAt
+    def run
+      User.where(registered_at: nil).find_each do |user|
```
benhalpern, line 4:
```suggestion
      User.where(registered_at: nil, registered: true).find_each do |user|
```

Issue #9324, opened 2020-07-15
Settings: Elements in content customization box leaking outside the container


icncsx:
**Describe the bug**

The text elements area leaking outside of the container.

![Screenshot from 2020-07-15 00-24-56](https://user-images.githubusercontent.com/60417145/87503287-cfc96180-c631-11ea-93ce-ce7c63e1d4af.png)

**To Reproduce**

Visit the [settings/ux](https://dev.to/settings/ux) endpoint and look at the coding experience boxes.

**Expected behavior**

Text should be contained in the box.

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

PavelTkachenko:
I can take this issue and provide png images to approve design before coding.

rhymes:
@PavelTkachenko go ahead, thank you!

benhalpern:
@PavelTkachenko We can think of the text as optional, so we could hide it for smaller breakpoints.

PR #9335, opened 2020-07-15
Enhance user experience layout (settings/ux) for small screens

pashagray:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
I removed labels on small screens as @benhalpern suggested, but I think user will lost confidence in selecting his level based only on number (1, 2, 3, 4, 5). I decided to be a little designer and offer such a solution for small screens. Tried to use pastel colors as I can see dev.to promotes. If this is unnecessary, let me know, I'll remove it.

#9324 

mobile look
## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed

alt_text

### DIFF
--- a/app/assets/stylesheets/theme-selector.scss
+++ b/app/assets/stylesheets/theme-selector.scss
@@ -153,6 +153,36 @@
       font-size: 0.5em;
     }
   }
+
+  @media screen and (max-width: 810px) {
+    .level_tag {
+      display: none;
+    }
+  }
+}
+
+.level-bar__labels {
+  display: flex;
+  justify-content: space-between;
+  flex-direction: row;
+  padding: 0.3rem var(--su-5);
+  color: white;
+  border-radius: 20px;
+
+  background: black;
+  background: -moz-linear-gradient(left, #6ebc25 0%, #e0c22a 50%, #c16868 100%);
+  background: -webkit-linear-gradient(
+    left,
+    #6ebc25 0%,
+    #e0c22a 50%,
+    #c16868 100%
+  );
+  background: linear-gradient(to right, #6ebc25 0%, #e0c22a 50%, #c16868 100%);
+  filter: progid:DXImageTransform.Microsoft.gradient( startColorstr='#6ebc25', endColorstr='#c16868',GradientType=1 );
+
+  @media (min-width: 810px) {
+    display: none;
+  }
 }
 
 .rich-selector-radio {
--- a/app/assets/stylesheets/theme-selector.scss
+++ b/app/assets/stylesheets/theme-selector.scss
@@ -166,19 +166,18 @@
   justify-content: space-between;
   flex-direction: row;
   padding: 0.3rem var(--su-5);
-  color: white;
+  color: black;
   border-radius: 20px;
 
   background: black;
-  background: -moz-linear-gradient(left, #6ebc25 0%, #e0c22a 50%, #c16868 100%);
+  background: -moz-linear-gradient(left, #66e2d5 0%, #fefa87 50%, #ff0000 100%);
   background: -webkit-linear-gradient(
     left,
-    #6ebc25 0%,
-    #e0c22a 50%,
-    #c16868 100%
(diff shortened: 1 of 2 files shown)


PavelTkachenko:
@rhymes Something weird with TravisCI. One test fails, but this PR is not a reason I guess...

rhymes:
@PavelTkachenko I restarted the build, it's likely a flaky test

benhalpern:
I think the design here is fairly brilliant!

I'd say that my only issue is that the colors are a little funny against the rest of the site. What if we did a spectrum of light to dark using the main crayons brand color for the theme?

I cede to @ludwiczakpawel and @lisasy for ultimate choice here. I'd also say that what we have here is a clear upgrade over the current funky broken look in prod, so I'd err on the side of approving this, even if we want to clean it up later.

And also—not sure what the crayonsy way of implementing such a custom thing would be, would we componentize this kind of thing?

pashagray:
Thanks for review! The same gradient with $green, $yellow and $red colors from variables.scss.

gradient

Zhao-Andy:
I prefer the SCSS variable gradient personally 🙂

pashagray:
I've changed colors according to $variables file. I saw that almost all files except `theme-selector.scss` importing these variables. I did't brake this pattern and set all colors manually. Is it ok?

app/assets/stylesheets/theme-selector.scss
```diff
@@ -153,6 +153,35 @@
...
+  }
+}
+
+.level-bar__labels {
+  display: flex;
+  justify-content: space-between;
+  flex-direction: row;
+  padding: 0.3rem var(--su-5);
+  color: black;
+  border-radius: 20px;
```
ludwiczakpawel, line 170: Just trying to avoid actual units in our CSS:

```suggestion
  border-radius:  var(--su-5); // that stands for 20px
```

app/assets/stylesheets/theme-selector.scss
```diff
@@ -153,6 +153,35 @@
...
+    }
+  }
+}
+
+.level-bar__labels {
+  display: flex;
+  justify-content: space-between;
+  flex-direction: row;
+  padding: 0.3rem var(--su-5);
+  color: black;
```
ludwiczakpawel, line 169:
```suggestion
  color: var(--body-color); // this should set the color to our _version_ of black
```

app/assets/stylesheets/theme-selector.scss
```diff
@@ -153,6 +153,35 @@
...
+
+  background: black;
+  background: -moz-linear-gradient(left, #66e2d5 0%, #fefa87 50%, #ff0000 100%);
+  background: -webkit-linear-gradient(
+    left,
+    #66e2d5 0%,
+    #fefa87 50%,
+    #ff0000 100%
+  );
+  background: linear-gradient(to right, #66e2d5 0%, #fefa87 50%, #ff0000 100%);
```
ludwiczakpawel, line 180: We automatically add vendor prefixes so for development we don't need to add them.

Also, in terms of colors, we're currently turning towards using our native CSS variables instead of hex values OR SCSS variables. [There's a file with all of them](https://github.com/forem/forem/blob/master/app/assets/stylesheets/config/_colors.scss).

So in this case I'd suggest little change:

```suggestion
  background: linear-gradient(to right, var(--accent-success) 0%, var(--accent-warning) 50%, var(--accent-danger) 100%);
```

mstruve:
Hey @pashagray!
 
Thank you so much for the PR and for contributing to Forem! We are super happy to have your help 😄

If you would still like to get this merged please feel free to update your branch with master and the suggested feedback and get the code ready to merge and then open the PR back up. We would be more than happy to help you push this through if you would like! 

Thanks again for contributing! Ping us anytime if you have questions!

PR #9354, opened 2020-07-16
 Remove codeql-analysis.yml

maestromac:
- [x] DX

## Description
Better to remove this file completely because at the moment it is giving out error notifications to forked repositories upon new branch commits
n/a
n/a
## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a


jhutchings1:
👋🏻  Hey @maestromac I'm a product manager at GitHub working on code scanning/CodeQL. Can you point me at some of the errors you were seeing in the forked PR workflows? Would love to validate we've got that fixed as part of the beta.

maestromac:
Hello @jhutchings1  I don't remember the initial error that started all of this but it was working just fine for a time before we move to a different org. Here's were the order of how things went down.

1. We migrated the repository to a different org which caused codeql to fail. 
2. We commented-out the file as a temporary measure.
3. Turns out commenting is not enough because the workflow still fails, so we had to remove it.

We added it originally because it's a suggested action from the repository's security tab but since that's didn't come back, we kind of just moved on. We can re-add it back in to test it out though.

jhutchings1:
Thanks for confirming, @maestromac. That makes sense, then. The feature is restricted behind a limited beta, and it looks like the `forem` organization isn't in the beta queue. If you sign up here and ping me, I'll get you squared away so you can start using it again. https://github.com/features/security/advanced-security/signup

maestromac:
@jhutchings1 we signed up. Thank you!

Issue #7798, opened 2020-05-12
Option to delete an organization


alsoknownasdrew:
**Is your feature request related to a problem? Please describe.**

I would like to delete an organization I've created and it seems there's no way to do it from its dashboard

**Describe the solution you'd like**

An option to delete the organization in settings page

atsmith813:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

atsmith813:
@alsoknownasdrew if you'd like to delete an organization in the meantime, you can email yo@dev.to with that request and we'll be happy to delete it for you!

cmgorton:
I believe this issue was fixed with PR #9357. I am going to close this issue but if you are still unable to delete an organization please let me know and we can reopen this issue. Thanks.

PR #9357, opened 2020-07-16
Allow org admins to delete their organizations

Zhao-Andy:
- [x] Feature

## Description
This PR allows users to delete their own organizations, if:
- there is only one member in the organization
- they are an organization admin
- the organization has no articles connected to it

While this isn't a heavily requested feature or anything, it's something we should support. Also, this should be one less support request that @michael-tharrington has to handle.

Related to #7798

Screen Shot 2020-07-16 at 12 56 13 PM
Screen Shot 2020-07-16 at 1 06 39 PM

1. Create an organization
2. Add a user via console:
```ruby
OrganizationMembership.create(user: User.first, organization: Organization.last, type_of_user: "member")
```
3. See that you can't delete it
4. Remove the user (either via settings or via the same script above, except `.destroy`)
5. Try to delete your organization via settings

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

Nope

"It's about damn time."


michael-tharrington:
Woot woot! We may be able to close this related request here - xhttps://github.com/thepracticaldev/tech-private/issues/403, although I think the notes under additional context are somewhat helpful if in the future we want to make it so folks can delete active (>1 member, >0 posts) organizations.

app/controllers/organizations_controller.rb
```diff
@@ -97,6 +112,10 @@ def organization_params
       end
   end
 
+  def destroyable_organization?(organization)
+    organization.organization_memberships.count == 1 && organization.articles.count.zero?
+  end
```
Zhao-Andy, line 117: I use `.count` here twice, but it might not be necessary.

msarit: I think it's fine 🤷‍♀️

app/controllers/organizations_controller.rb
```diff
@@ -49,6 +49,21 @@ def update
     end
   end
 
+  def destroy
+    organization = Organization.find_by(id: params[:id])
+    authorize organization
+    if destroyable_organization?(organization)
```
rhymes, line 55: I like this, though I would put it in the model as `organization.destroyable?` or even better, put it as a condition for the authorization, `:destroy` and remove the if. After all that's what we're saying to the user: you're not authorized to destroy this organization because this and that.

app/controllers/organizations_controller.rb
rhymes: I think so too

app/views/users/_org_admin.html.erb
```diff
@@ -198,3 +198,42 @@
...
+      <% if @org_organization_memberships.size === 1 && @organization.articles_count.zero? %>
+        <% onsubmit = "return confirm('Are you absolutely sure you want to delete your organization: #{@organization.name}?');" %>
+        <%= form_tag organization_path(@organization.id), method: :delete, onsubmit: onsubmit do %>
+          <button class="crayons-btn crayons-btn--danger crayons-btn--icon-left" type="submit">
+            Delete "<%= @organization.name %>"
+          </button>
+        <% end %>
+      <% else %>
+        <p>
+          Your organization currently does not meet the above requirements.
```
benhalpern, line 228: I really like this flow.

app/controllers/organizations_controller.rb
Zhao-Andy: Hmm so if I put the condition in the authorization, would I rescue the error? For example:
```ruby
def destroy
    organization = Organization.find_by(id: params[:id])
    authorize organization
    organization.destroy
    current_user.touch(:organization_info_updated_at)
    CacheBuster.bust_user(current_user)
    flash[:settings_notice] = "Your organization: \"#{organization.name}\" was successfully deleted."
    redirect_to user_settings_path(:organization)
  rescue Pundit::NotAuthorizedError
    flash[:settings_notice] = "Your organization was not deleted; you must be an admin,
       the only member in the organization, and  no articles connected to the organization."
    redirect_to user_settings_path(:organization, id: organization.id)
  end
```

That seems a bit counterintuitive but maybe I'm approaching it incorrectly?

PR #9366, opened 2020-07-17
Rubocop: enable and fix Style/Send and Style/Next

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Activated and corrected the following cops:

- [Style/Send](https://rubystyle.guide/#prefer-__send__) which uses [BasicObject#_ _ send _ _](https://rubyapi.org/2.7/o/basicobject#method-i-__send__) instead of `Kernel#send` to avoid any possible conflict with methods called `send()` . I've also replaced `send` with `public_send` in some places.

- [Style/Next](https://rubystyle.guide/#no-nested-conditionals) to use the classic Ruby style of exiting early

## Added tests?

- [ ] yes
- [ ] no, because I need help


rhymes:
Not sure how to fix that flaky failing spec

PR #9370, opened 2020-07-17
Refactor: Remove Block model and All Related Code

mstruve:

- [x] Refactor

## Description
I have started to look at PKs(Primary Keys) that need to be changed to bigint's in our database. While looking at the blocks table I noticed we only had 7 entries and they were all from 2017. I think we can safely assume this code is no longer used and can be removed. Following the code removal I will remove the database table.

alt_text


rhymes:
Yes! :D

PR #9371, opened 2020-07-17
[DO NOT MERGE] Add alternate carrierwave config

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add the ability to differentiate Heroku config from other environments.

I'm not 100% sure this is correct, but proposing it for an alternate scenario where we're using IAM instead of passing around keys.

### DO NOT MERGE

Even though this logic doesn't break anything, we don't fully understand what the "alt" version should be. We're messing around with variations.


benhalpern:
Closing for https://github.com/forem/forem/pull/9544

PR #9375, opened 2020-07-18
Fix hidden footnote content regression

emgoto:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This bug was originally fixed in [this pull request](https://github.com/forem/forem/pull/8298).

However there may have been some changes to the names of classes - I couldn't find an `article` class or `body` class when I was inspecting an article.

This fixes issue #7760 

footnote

## Added tests?

- [ ] yes
- [x] no, because I need help 

Would be cool to have this tested so it doesn't break again, however not sure if there is a good way to go about it.

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

alt_text

Going to admit it was a bit of a struggle setting up Forem locally, and a bit overwhelming to try and find a ticket that I could work on!

### DIFF
--- a/app/assets/stylesheets/article-show.scss
+++ b/app/assets/stylesheets/article-show.scss
@@ -26,7 +26,7 @@
 }
 
 body.default-navbar-config {
-  .article .body [id] {
+  .crayons-article .crayons-article__body [id] {
     scroll-margin-top: var(--header-height);
   }
 }


emgoto:
> Was there anything in particular you found overwhelming about finding a ticket? Is there a flow you normally use

I'm very new to open source so I didn't really previously have a flow. I think it's just the sheer number of tickets you have to click through to try and find something that isn't already taken, that you feel comfortable doing (i.e. you know where/what in the code you need to touch) or has enough information that you can do it. Also sometimes some of the tickets are no longer reproducible. 

I got a [really great explanation](https://github.com/forem/forem/issues/4686#issuecomment-660953468) of where in the code to get started on one of the issues, I think I will try and reach out more in the future like this because that definitely helps!

Issue #9327, opened 2020-07-15
Proper url parse for images in markdown with cloudinary


crimson-med:
**Describe the bug**

The way the url is parsed to cloudinary creates an issue for images coming from International Domain Names (IDNs)

**To Reproduce**

- Create a new post
- Add the following image: `![graph](https://www.火.dev:8443/user/MedericPixium/graph/sunsetred)`
- Add the following image: `![graph](https://www.xn--vnx.dev:8443/user/MedericPixium/graph/sunsetred)`

https://dev.to/crimsonmed/readme-dynaimic-kn9?preview=a3815214d53d09152a66703707825e0de87863e5bc3a15eeca344c6d5a39b2726af309593795f462e504a6bf0a90cf401073bc0553a6c166fdb570f6

**Expected behavior**

One of the two should be rendered properly as here:
![image](https://www.火.dev:8443/user/MedericPixium/graph/sunsetred)
![image](https://www.xn--vnx.dev:8443/user/MedericPixium/graph/sunsetred)

**Desktop (please complete the following information):**

- Device: Macbook
- OS: MacOS
- Browser: Firefox

**Additional context**

Cloudinary said they support IDNs

> Thanks for reaching out.
>
> We support IDN domain names and here's an example URL using our 'fetch remote image URL' feature:
>
> https://res.cloudinary.com/demo/image/fetch/fl_sanitize/https://www.xn--vnx.dev:8443/user/MedericPixium/graph/sunsetred
>
> And the same URL with additional transformations applied:
> https://res.cloudinary.com/demo/image/fetch/f_jpg,q_auto,e_grayscale/https://www.xn--vnx.dev:8443/user/MedericPixium/graph/sunsetred
>
> In the case of the example that you provided, the 400 image URLs are because of the way the domain was passed in the > URLs - if it was passed as 'Punycode' it would work like the example I provided above
>
> Hope this helps, please let me know if you have any further questions.

However dev.to does not forward the URL to punicode: https://res.cloudinary.com/practicaldev/image/fetch/s--hKwi4Rec--/c_limit%2Cf_auto%2Cfl_progressive%2Cq_auto%2Cw_880/https://www.%25E7%2581%25AB.dev:8443/user/MedericPixium/card

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

[...]

PR #9391, opened 2020-07-19
Fix broken urls with IDNs in Cloudinary helper

bencgreenberg:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This addresses the issue raised in #9327. It adds a new gem to parse unicode URLs and turn them into a format that Cloudinary can work with. Additional specs also added to verify desired behavior.

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed

alt_text


app/helpers/application_helper.rb
```diff
@@ -76,8 +76,12 @@ def icon_url(name)
     "https://res.cloudinary.com/#{ApplicationConfig['CLOUDINARY_CLOUD_NAME']}/image/upload/#{postfix}"
   end
 
+  def idn_domain_fix(url)
```
citizen428, line 79: **thought (non-blocking):** This helper doesn't seem to serve much of a purpose. Maybe for now we can just inline the call to `SimpleIDN` in `cloudinary` and update the test to check the URL in the attribute. I'll accept the PR either way, but if you do change it please request a re-review.

bencgreenberg: That's a good point, @citizen428. My first inclination is to break things out in their own methods for reuse, but you're right, here there isn't much of a reuse potential. I can take another crack at this and revise the specs this evening my time and I'll re-request a review then.

rhymes:
@bencgreenberg I added the missing vendored gem file, remember to commit the result of `bundle install` in the future. Thank you!

PR #9392, opened 2020-07-19
Fix spacing issue on search result page [mobile]

ubaidisaev:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

image

### DIFF
--- a/app/views/articles/search.html.erb
+++ b/app/views/articles/search.html.erb
@@ -11,7 +11,7 @@
   <div class="articles-list" id="articles-list">
     <div id="banner-section"></div>
     <div class="flex-1 flex items-center justify-between">
-      <h2 class="fs-l fw-heavy pl-2">Results</h2>
+      <h2 class="fs-l fw-heavy p-2 m:p-0">Results</h2>
       <nav id="sorting-option-tabs" class="crayons-tabs hidden s:flex">
         <a href="javascript:;" class="crayons-tabs__item <%= "crayons-tabs__item--current" if @current_ordering == :relevance %>">Most Relevant</a>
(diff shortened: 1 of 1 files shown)


rhymes:
Thanks @ubaidisaev!

app/views/articles/search.html.erb
```diff
@@ -11,7 +11,7 @@
   <div class="articles-list" id="articles-list">
     <div id="banner-section"></div>
     <div class="flex-1 flex items-center justify-between">
-      <h2 class="fs-l fw-heavy">Results</h2>
+      <h2 class="fs-l fw-heavy pl-2">Results</h2>
```
ludwiczakpawel, line 14: This will also affect the desktop version of the site, so I'd suggest slightly different approach here:

```suggestion
      <h2 class="fs-l fw-heavy p-2 m:p-0">Results</h2>
```

This would apply `.p-2` (`padding: var(--su-2)`) class by default, and we reach medium breakpoint (this is what `m:` prefix stands for), it would apply `.p-0` (`padding: 0;`)

Also, as you can see my suggestion is to use `p-` instead of `pl-` so we can apply padding on all sides instead only left - this particular area would definitely appreciate more room on all sides :)

mstruve:
Hey @ubaidisaev!

Thank you so much for this contribution! Unfortunately, it looks like this html has actually been removed so I am going to go ahead and close this PR since we no longer need it. Thank you again for being a part of the Forem community! We couldn't do what we do without you!

PR #9395, opened 2020-07-20
Standardize ActiveRecord order clauses

citizen428:

- [X] Refactor
- [X] OCD

## Description

The use of ActiveRecord `order` clauses throughout our codebase was very inconsistent. We used both the hash and string form, didn't always user uppercase keywords in string clauses, etc.

This PR standardizes on the following style:

1. Simple cases are now of the form `order(attribute: direction)`. No more string clauses, no more relying on "ASC" being the default.
2. Nested cases are now consistently of the form `order("other_table.attribute" => direction)`. We previously used a mix of this form and the string form.

Reason: the hash version is slightly smarter/more robust than the string version. It automatically prefixes the table name to the columns, adds an explicit direction of "ASC" when it's missing (we don't do this anymore), etc. It also reads better (IMHO of course) and I think consistency matters.

*checks microphone*
"Is this on? I want to thank ripgrep and Vim's macro system, without whom this PR would not have been possible. This succuess is as much theirs as it is mine."

n/a

If Travis is happy, we are happy.

## Added tests?

## Added to documentation?

- [X] no documentation needed

### DIFF
--- a/app/controllers/chat_channels_controller.rb
+++ b/app/controllers/chat_channels_controller.rb
@@ -175,7 +175,7 @@ class ChatChannelsController < ApplicationController
                                      .where(show_global_badge_notification: true)
                                      .where.not(status: %w[removed_from_channel left_channel])
                                      .includes(%i[chat_channel user])
-                                     .order("chat_channel_memberships.updated_at DESC")
+                                     .order("chat_channel_memberships.updated_at" => :desc)
                                  else
                                    []
                                  end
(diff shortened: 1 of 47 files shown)


rhymes:
Wish there was a Rubocop rule for this :D

citizen428:
> Wish there was a Rubocop rule for this :D

Funny you'd say that because I definitely thought about writing one 😉

PR #9401, opened 2020-07-20
Change from presence to private pusher channels

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We've been using Pusher's "presence" channels, which actually have a max occupancy. So in order to create bigger group channels, we need to move to "private", which is fine. As far as I know we aren't using any of the features of presence anyway. (I think we _may_ have in a prior implementation of some features.

Correct me if I'm wrong about this idea.

Here is the info about presence channels:

https://pusher.com/docs/channels/using_channels/presence-channels

> Presence channels have some limits associated with them: 100 members maximum, 1KB limit for user object, and maximum 128 characters for user id. If you use a numeric user id, remember that the maximum size integer that is representable in JavaScript is 2^53.

And FWIW, here is an article about how one might implement some of the features of "presence" for "large" channels.

https://support.pusher.com/hc/en-us/articles/360019620253-How-can-I-implement-large-presence-channels-on-Channels-

We may want to legit use presence channels or these concepts in the future, but the most likely outcome should be us moving from Pusher to something we can containerize and ship on commodity clouds. So this is probably a step towards that pattern by using less complicated Pusher features.

### DIFF
--- a/app/javascript/chat/chat.jsx
+++ b/app/javascript/chat/chat.jsx
@@ -273,7 +273,7 @@ export default class Chat extends Component {
     );
     this.subscribeChannelsToPusher(
       channels.filter(this.channelTypeFilterFn('invite_only')),
-      (channel) => `presence-channel-${channel.chat_channel_id}`,
+      (channel) => `private-channel-${channel.chat_channel_id}`,
     );
     const chatChannelsList = document.getElementById(
       'chatchannels__channelslist',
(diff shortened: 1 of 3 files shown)


app/javascript/chat/chat.jsx
```diff
@@ -352,7 +352,7 @@ export default class Chat extends Component {
       if (activeChannel.channel_type === 'open')
         this.subscribePusher(`open-channel-${channelId}`);
     }
-    this.subscribePusher(`presence-channel-${channelId}`);
+    this.subscribePusher(`private-channel-${channelId}`);
```
Zhao-Andy, line 355: Will we have to create new channels in Pusher for this?

Zhao-Andy: Ahh okay. Thought it was more of a persistent database object. 👍

leewynne:
Hey there, anything i need to be aware of when implementing this?

PR #5199, opened 2019-12-21
Allow comments from deleted articles to be viewed

twinsfan421:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This allows comments from deleted articles to be viewed as discussed in #5170.  I created a separate deleted_article_comment view that will render from comments#index if no article if found. 
## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
Screen Shot 2019-12-20 at 10 23 33 PM

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

PR #9405, opened 2020-07-20
Rubocop: fix Rails/HasManyOrHasOneDependent by specifying it

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x]  Optimization
- [ ] Documentation Update

## Description

We have a few `has_many` that have no explicit `dependent` clause, and those should be as explicit as possible. I left some comments to each of them to make it easier for reviewer to comment.

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -524,16 +524,6 @@ Performance/StringInclude:
 Rails/FilePath:
   EnforcedStyle: slashes
 
-Rails/HasManyOrHasOneDependent:
-  Description: 'Define the dependent option to the has_many and has_one associations.'
-  StyleGuide: 'https://rails.rubystyle.guide#has_many-has_one-dependent-option'
-  Enabled: false
-
-Rails/SaveBang:
-  Description: 'Identifies possible cases where Active Record save! or related should be used.'
-  StyleGuide: 'https://rails.rubystyle.guide#save-bang'
-  Enabled: false
-
 Rails/SkipsModelValidations:
   Description: >-
                  Use methods that skips model validations with caution.
--- a/app/models/ahoy/visit.rb
+++ b/app/models/ahoy/visit.rb
@@ -2,7 +2,7 @@ module Ahoy
   class Visit < ApplicationRecord
     self.table_name = "ahoy_visits"
 
-    has_many :events, class_name: "Ahoy::Event"
+    has_many :events, class_name: "Ahoy::Event", dependent: :destroy
     belongs_to :user, optional: true
   end
 end
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -19,16 +19,22 @@ class Article < ApplicationRecord
   delegate :name, to: :user, prefix: true
   delegate :username, to: :user, prefix: true
 
-  belongs_to :user
-  belongs_to :organization, optional: true
   # touch: true was removed because when an article is updated, the associated collection
   # is touched along with all its articles(including this one). This causes eventually a deadlock.
   belongs_to :collection, optional: true
+  belongs_to :organization, optional: true
+  belongs_to :user
 
   counter_culture :user
   counter_culture :organization
 
-  has_many :comments, as: :commentable, inverse_of: :commentable
+  has_many :buffer_updates, dependent: :destroy
+  has_many :comments, as: :commentable, inverse_of: :commentable, dependent: :nullify
+  has_many :notification_subscriptions, as: :notifiable, inverse_of: :notifiable, dependent: :destroy
+  has_many :notifications, as: :notifiable, inverse_of: :notifiable, dependent: :delete_all
+  has_many :page_views, dependent: :destroy
+  has_many :profile_pins, as: :pinnable, inverse_of: :pinnable, dependent: :destroy
+  has_many :rating_votes, dependent: :destroy
   has_many :top_comments,
            lambda {
              where(
@@ -39,12 +45,6 @@ class Article < ApplicationRecord
            as: :commentable,
            inverse_of: :commentable,
            class_name: "Comment"
-  has_many :profile_pins, as: :pinnable, inverse_of: :pinnable
-  has_many :buffer_updates, dependent: :destroy
-  has_many :notifications, as: :notifiable, inverse_of: :notifiable, dependent: :delete_all
-  has_many :notification_subscriptions, as: :notifiable, inverse_of: :notifiable, dependent: :destroy
-  has_many :rating_votes
-  has_many :page_views
 
   validates :slug, presence: { if: :published? }, format: /\A[0-9a-z\-_]*\z/,
                    uniqueness: { scope: :user_id }
--- a/app/models/badge.rb
+++ b/app/models/badge.rb
@@ -2,8 +2,8 @@ class Badge < ApplicationRecord
   mount_uploader :badge_image, BadgeUploader
   resourcify
 
-  has_many :badge_achievements
-  has_many :tags
+  has_many :badge_achievements, dependent: :destroy
+  has_many :tags, dependent: :nullify
   has_many :users, through: :badge_achievements
 
   validates :badge_image, presence: true
--- a/app/models/broadcast.rb
+++ b/app/models/broadcast.rb
@@ -2,7 +2,7 @@ class Broadcast < ApplicationRecord
   VALID_BANNER_STYLES = %w[default brand success warning error].freeze
   resourcify
 
-  has_many :notifications, as: :notifiable, inverse_of: :notifiable
+  has_many :notifications, as: :notifiable, inverse_of: :notifiable, dependent: :destroy
 
   validates :title, uniqueness: { scope: :type_of }, presence: true
   validates :type_of, :processed_html, presence: true
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -1,5 +1,5 @@
 class Collection < ApplicationRecord
-  has_many :articles
+  has_many :articles, dependent: :nullify
   belongs_to :user
   belongs_to :organization, optional: true
 
--- a/app/models/concerns/user_subscription_sourceable.rb
+++ b/app/models/concerns/user_subscription_sourceable.rb
@@ -4,7 +4,7 @@ module UserSubscriptionSourceable
   # This all assumes there's an association with User under the column user_id.
 
   included do
-    has_many :user_subscriptions, as: :user_subscription_sourceable
+    has_many :user_subscriptions, as: :user_subscription_sourceable, dependent: :destroy
     has_many :sourced_subscribers,
              class_name: "User",
              through: :user_subscriptions,
--- a/app/models/display_ad.rb
+++ b/app/models/display_ad.rb
@@ -1,6 +1,6 @@
 class DisplayAd < ApplicationRecord
   belongs_to :organization
-  has_many :display_ad_events
+  has_many :display_ad_events, dependent: :destroy
 
   validates :organization_id, presence: true
   validates :placement_area, presence: true,
--- a/app/models/html_variant.rb
+++ b/app/models/html_variant.rb
@@ -10,8 +10,8 @@ class HtmlVariant < ApplicationRecord
   validate  :no_edits
 
   belongs_to :user, optional: true
-  has_many :html_variant_trials
-  has_many :html_variant_successes
+  has_many :html_variant_trials, dependent: :destroy
+  has_many :html_variant_successes, dependent: :destroy
 
   before_save :prefix_all_images
 
--- a/app/models/listing_category.rb
+++ b/app/models/listing_category.rb
@@ -3,7 +3,7 @@ class ListingCategory < ApplicationRecord
   # We standardized on the latter, but keeping the table name was easier.
   self.table_name = "classified_listing_categories"
 
-  has_many :listings, inverse_of: :listing_category
+  has_many :listings, inverse_of: :listing_category, dependent: :nullify
 
   before_validation :normalize_social_preview_color
 
--- a/app/models/listing_category.rb
+++ b/app/models/listing_category.rb
@@ -3,7 +3,7 @@ class ListingCategory < ApplicationRecord
   # We standardized on the latter, but keeping the table name was easier.
   self.table_name = "classified_listing_categories"
 
-  has_many :listings, inverse_of: :listing_category, dependent: :nullify
+  has_many :listings, inverse_of: :listing_category, dependent: :restrict_with_exception
 
   before_validation :normalize_social_preview_color
 
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -12,15 +12,14 @@ class Organization < ApplicationRecord
   acts_as_followable
 
   has_many :api_secrets, through: :users
-  has_many :articles
-  has_many :listings
-  has_many :collections
-  has_many :credits
-  has_many :display_ads
-  has_many :notifications
+  has_many :articles, dependent: :nullify
+  has_many :listings, dependent: :nullify
+  has_many :collections, dependent: :nullify
+  has_many :credits, dependent: :restrict_with_exception
+  has_many :display_ads, dependent: :destroy
+  has_many :notifications, dependent: :destroy
   has_many :organization_memberships, dependent: :delete_all
-  has_many :profile_pins, as: :profile, inverse_of: :profile
-  has_many :sponsorships
+  has_many :sponsorships, dependent: :nullify
   has_many :unspent_credits, -> { where spent: false }, class_name: "Credit", inverse_of: :organization
   has_many :users, through: :organization_memberships
 
--- a/app/models/podcast.rb
+++ b/app/models/podcast.rb
@@ -1,8 +1,8 @@
 class Podcast < ApplicationRecord
   resourcify
 
-  has_many :podcast_episodes
   belongs_to :creator, class_name: "User", inverse_of: :created_podcasts, optional: true
+  has_many :podcast_episodes, dependent: :destroy
 
   mount_uploader :image, ProfileImageUploader
   mount_uploader :pattern_image, ProfileImageUploader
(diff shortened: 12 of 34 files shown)


.rubocop.yml
```diff
@@ -524,16 +524,6 @@ Performance/StringInclude:
 Rails/FilePath:
   EnforcedStyle: slashes
 
-Rails/HasManyOrHasOneDependent:
-  Description: 'Define the dependent option to the has_many and has_one associations.'
-  StyleGuide: 'https://rails.rubystyle.guide#has_many-has_one-dependent-option'
-  Enabled: false
-
-Rails/SaveBang:
```
rhymes, line 532: disabled by default, no need to keep it here

app/models/ahoy/visit.rb
```diff
@@ -2,7 +2,7 @@ module Ahoy
   class Visit < ApplicationRecord
     self.table_name = "ahoy_visits"
 
-    has_many :events, class_name: "Ahoy::Event"
+    has_many :events, class_name: "Ahoy::Event", dependent: :destroy
```
rhymes, line 5: an event belonging to a visit should be destroyed if the visit gets destroyed

app/models/article.rb
```diff
@@ -19,16 +19,22 @@ class Article < ApplicationRecord
   delegate :name, to: :user, prefix: true
   delegate :username, to: :user, prefix: true
 
-  belongs_to :user
-  belongs_to :organization, optional: true
   # touch: true was removed because when an article is updated, the associated collection
   # is touched along with all its articles(including this one). This causes eventually a deadlock.
   belongs_to :collection, optional: true
+  belongs_to :organization, optional: true
+  belongs_to :user
```
rhymes, line 26: just reordering

app/models/article.rb
```diff
@@ -19,16 +19,22 @@ class Article < ApplicationRecord
...
 
   counter_culture :user
   counter_culture :organization
 
-  has_many :comments, as: :commentable, inverse_of: :commentable
+  has_many :buffer_updates, dependent: :destroy
```
rhymes, line 31: reordering

app/models/article.rb
```diff
@@ -19,16 +19,22 @@ class Article < ApplicationRecord
...
+  has_many :comments, as: :commentable, inverse_of: :commentable, dependent: :nullify
```
rhymes, line 32: a comment belonging to an article should become an orphan when the article is destroyed

app/models/article.rb
```diff
@@ -19,16 +19,22 @@ class Article < ApplicationRecord
...
+  has_many :notification_subscriptions, as: :notifiable, inverse_of: :notifiable, dependent: :destroy
+  has_many :notifications, as: :notifiable, inverse_of: :notifiable, dependent: :delete_all
```
rhymes, line 34: reordering

app/models/article.rb
```diff
@@ -19,16 +19,22 @@ class Article < ApplicationRecord
...
+  has_many :page_views, dependent: :destroy
```
rhymes, line 35: a page belonging to an article should be destroyed if the article is destroyed

app/models/article.rb
```diff
@@ -19,16 +19,22 @@ class Article < ApplicationRecord
...
+  has_many :profile_pins, as: :pinnable, inverse_of: :pinnable, dependent: :destroy
```
rhymes, line 36: A profile pin of an article should be destroyed if the article is destroyed

app/models/article.rb
```diff
@@ -19,16 +19,22 @@ class Article < ApplicationRecord
...
+  has_many :rating_votes, dependent: :destroy
```
rhymes, line 37: Rating votes of an article should be destroyed if an article is destroyed

app/models/badge.rb
```diff
@@ -2,8 +2,8 @@ class Badge < ApplicationRecord
   mount_uploader :badge_image, BadgeUploader
   resourcify
 
-  has_many :badge_achievements
-  has_many :tags
+  has_many :badge_achievements, dependent: :destroy
```
rhymes, line 5: If a badge disappears, its achievements should also disappear

app/models/badge.rb
```diff
@@ -2,8 +2,8 @@ class Badge < ApplicationRecord
...
+  has_many :tags, dependent: :nullify
```
rhymes, line 6: If a badge disappears, a tag relationship to it should just disappear

app/models/broadcast.rb
```diff
@@ -2,7 +2,7 @@ class Broadcast < ApplicationRecord
   VALID_BANNER_STYLES = %w[default brand success warning error].freeze
   resourcify
 
-  has_many :notifications, as: :notifiable, inverse_of: :notifiable
+  has_many :notifications, as: :notifiable, inverse_of: :notifiable, dependent: :destroy
```
rhymes, line 5: Destroying broadcasts also means destroyed notifications generated by it

app/models/collection.rb
```diff
@@ -1,5 +1,5 @@
 class Collection < ApplicationRecord
-  has_many :articles
+  has_many :articles, dependent: :nullify
```
rhymes, line 2: Destroying a collection simply means removing the articles from the collection, not destroying the articles

app/models/concerns/user_subscription_sourceable.rb
```diff
@@ -4,7 +4,7 @@ module UserSubscriptionSourceable
   # This all assumes there's an association with User under the column user_id.
 
   included do
-    has_many :user_subscriptions, as: :user_subscription_sourceable
+    has_many :user_subscriptions, as: :user_subscription_sourceable, dependent: :destroy
```
rhymes, line 7: Destroying the object of a subscription, means also destroying the subscriptions attached to it. If the author removes an articles, subscription should go as well

app/models/display_ad.rb
```diff
@@ -1,6 +1,6 @@
 class DisplayAd < ApplicationRecord
   belongs_to :organization
-  has_many :display_ad_events
+  has_many :display_ad_events, dependent: :destroy
```
rhymes, line 3: events belonging to an ad should be tied to it and disappear on destruction

app/models/html_variant.rb
```diff
@@ -10,8 +10,8 @@ class HtmlVariant < ApplicationRecord
   validate  :no_edits
 
   belongs_to :user, optional: true
-  has_many :html_variant_trials
-  has_many :html_variant_successes
+  has_many :html_variant_trials, dependent: :destroy
+  has_many :html_variant_successes, dependent: :destroy
```
rhymes, line 14: trials and successes are strongly coupled with the variants, so they should just go

app/models/listing_category.rb
```diff
@@ -3,7 +3,7 @@ class ListingCategory < ApplicationRecord
   # We standardized on the latter, but keeping the table name was easier.
   self.table_name = "classified_listing_categories"
 
-  has_many :listings, inverse_of: :listing_category
+  has_many :listings, inverse_of: :listing_category, dependent: :restrict_with_exception
```
rhymes, line 6: A category can't be removed if it still has listings attached to it

app/models/organization.rb
```diff
@@ -12,15 +12,14 @@ class Organization < ApplicationRecord
   acts_as_followable
 
   has_many :api_secrets, through: :users
-  has_many :articles
-  has_many :listings
-  has_many :collections
-  has_many :credits
-  has_many :display_ads
-  has_many :notifications
+  has_many :articles, dependent: :nullify
```
rhymes, line 15: Removing an organization should leave the user's articles intact

app/models/organization.rb
```diff
@@ -12,15 +12,14 @@ class Organization < ApplicationRecord
...
+  has_many :listings, dependent: :nullify
```
rhymes, line 16: Removing an organization should leave the user's listings intact

app/models/organization.rb
```diff
@@ -12,15 +12,14 @@ class Organization < ApplicationRecord
...
+  has_many :collections, dependent: :nullify
```
rhymes, line 17: Removing an organization should leave the user's collections intact

app/models/organization.rb
```diff
@@ -12,15 +12,14 @@ class Organization < ApplicationRecord
...
+  has_many :credits, dependent: :restrict_with_exception
```
rhymes, line 18: An org should never be removed if it has credits

app/models/organization.rb
```diff
@@ -12,15 +12,14 @@ class Organization < ApplicationRecord
...
+  has_many :display_ads, dependent: :destroy
```
rhymes, line 19: Ads should go if the org goes

app/models/organization.rb
```diff
@@ -12,15 +12,14 @@ class Organization < ApplicationRecord
...
+  has_many :notifications, dependent: :destroy
```
rhymes, line 20: Orgs notifications should go if the user goes. I don't think there's a case where a notification is the same object for a user and and org, correct?

app/models/organization.rb
```diff
@@ -12,15 +12,14 @@ class Organization < ApplicationRecord
...
   has_many :organization_memberships, dependent: :delete_all
-  has_many :profile_pins, as: :profile, inverse_of: :profile
-  has_many :sponsorships
+  has_many :sponsorships, dependent: :nullify
```
rhymes, line 22: Org's sponsorships should go with it

app/models/podcast.rb
```diff
@@ -1,8 +1,8 @@
 class Podcast < ApplicationRecord
   resourcify
 
-  has_many :podcast_episodes
   belongs_to :creator, class_name: "User", inverse_of: :created_podcasts, optional: true
+  has_many :podcast_episodes, dependent: :destroy
```
rhymes, line 5: Bye podcast, by episodes

app/models/podcast_episode.rb
```diff
@@ -15,7 +15,7 @@ class PodcastEpisode < ApplicationRecord
   delegate :published, to: :podcast
 
   belongs_to :podcast
-  has_many :comments, as: :commentable, inverse_of: :commentable
+  has_many :comments, as: :commentable, inverse_of: :commentable, dependent: :nullify
```
rhymes, line 18: Removing a podcast episode should orphan the comments, not remove them. Comments have unique URLs that can survive the destruction of the object they are attached to

app/models/poll.rb
```diff
@@ -4,9 +4,9 @@ class Poll < ApplicationRecord
   serialize :voting_data
 
   belongs_to :article
-  has_many :poll_options
-  has_many :poll_skips
-  has_many :poll_votes
+  has_many :poll_options, dependent: :destroy
+  has_many :poll_skips, dependent: :destroy
+  has_many :poll_votes, dependent: :destroy
```
rhymes, line 9: Bye bye polls, bye bye poll stuff

app/models/user.rb
```diff
@@ -90,7 +90,7 @@ class User < ApplicationRecord
   has_many :identities_enabled, -> { enabled }, class_name: "Identity", inverse_of: false
   has_many :mentions, dependent: :destroy
   has_many :messages, dependent: :destroy
-  has_many :notes, as: :noteable, inverse_of: :noteable
+  has_many :notes, as: :noteable, inverse_of: :noteable, dependent: :destroy
```
rhymes, line 93: User notes should get destroyed if the user goes

app/models/article.rb
citizen428: Are there places that chain methods unto this (e.g. `comment.commetable.title` or whatever) that may need extra guards?

rhymes: Not really, we already render comments when they don't have a commentable anymore: https://github.com/forem/forem/pull/5199

app/models/poll_option.rb
```diff
@@ -1,9 +1,8 @@
 class PollOption < ApplicationRecord
   belongs_to :poll
-  has_many :poll_votes
+  has_many :poll_votes, dependent: :destroy
```
rhymes, line 3: Removing a poll option should remove its votes as well

app/services/articles/destroyer.rb
```diff
@@ -3,12 +3,14 @@ module Destroyer
...
-      article.destroy!
       Notification.remove_all_without_delay(notifiable_ids: article.id, notifiable_type: "Article")
+
       if article.comments.exists?
-        Notification.remove_all(notifiable_ids: article.comments.pluck(:id),
-                                notifiable_type: "Comment")
+        Notification.remove_all(notifiable_ids: article.comments.pluck(:id), notifiable_type: "Comment")
       end
+
+      article.destroy!
```
rhymes, line 12: moved this as the last step mainly for a reason: related objects should be removed before the object itself is deleted

app/models/user.rb
```diff
@@ -90,7 +90,7 @@ class User < ApplicationRecord
   has_many :identities_enabled, -> { enabled }, class_name: "Identity", inverse_of: false
   has_many :mentions, dependent: :destroy
   has_many :messages, dependent: :destroy
-  has_many :notes, as: :noteable, inverse_of: :noteable
+  has_many :notes, as: :noteable, inverse_of: :noteable, dependent: :nullify
```
rhymes, line 93: Apparently we keep notes around, not sure why but: 

https://github.com/forem/forem/blob/2fe4a36271106f22a19ef996f5432e58d249b2b9/spec/services/users/delete_spec.rb#L91-L99

config/database.yml
```diff
@@ -66,6 +66,8 @@ test:
   <<: *default
   url: <%= ENV['DATABASE_URL_TEST'] || ENV['DATABASE_URL'] || ''%>
   database: PracticalDeveloper_test<%= ENV['TEST_ENV_NUMBER'] %>
+  variables:
+      statement_timeout: <%= ENV['STATEMENT_TIMEOUT'] || 10_000 %>
```
rhymes, line 70: trying to fix DB statement timeout errors

app/models/article.rb
mstruve: @citizen428 we have already fixed most of those thanks to Honeybadgers and orphaned comments already existing.

app/models/concerns/user_subscription_sourceable.rb
atsmith813: Great catch! I don't think we would want to add this `dependent: :destroy`.

The subscriptions should remain even if an author destroyed one of their articles. We would end up displaying it as a subscription to an unpublished/deleted article.  Since it's the author's content, that's sort of on them.

I'm also thinking of how this could look in the future. If we set aside what we know about the liquid tag currently using this, we can use subscriptions in other areas. We could add a button to a user's profile where another user could click "subscribe", for example (also set aside the fact that would be kind of confusing what the difference between subscribing to and following a user is). These subscriptions are logging the bridge between 2 users and where the bridge came from originally. Even if the origin (Article, Comment, User profile, etc) were to be deleted, the subscription itself is still relevant.

Technically, we could do subscriptions with no source at all which is what it ultimately becomes if the source is deleted. It's less useful to the author because they can't segment their audience. In the context of deleting the source though, the source was once present before the author chose to delete their content, so they can still segment and use that power before they delete it.

We also "soft delete" many objects too which helps with this.

rhymes: @atsmith813 thanks for the explanation. I'm going to make the following changes then:

- set this as `dependent: nullify` so that they are automatically set to null if the source gets destroyed
- set `belongs_to :user_subscription_sourceable, polymorphic: true` in the `UserSubscription` model to `optional: true` as the `UserSubscription` can be valid without it as well. 

Would that work?

app/models/organization.rb
rhymes: @Zhao-Andy is working on adding the ability to destroy organizations here https://github.com/forem/forem/pull/9357

See https://github.com/forem/forem/pull/9357/files#diff-33bef3083bae5b2e2bea316eab44c9fdR55 especially

app/models/organization.rb
rhymes: Yeah

app/models/article.rb
citizen428: Thanks @rhymes and @mstruve, it's just one of these cases where I'd rather ask one time too many.

app/models/concerns/user_subscription_sourceable.rb
atsmith813: @rhymes that would work, I like that! I wonder if there's a quick win to validate the presence of `user_subscription_sourceable` only on create 🤔 . A subscription shouldn't be created without it, but it can be updated to be removed.

Thinking out loud...something we could explore in the future (definitely not now because it doesn't even exist yet), is somehow persisting the `user_subscription_sourceable_type` value so we could then say display _Deleted Comment_, _Deleted Article_, etc. Since we only use this in Articles and we don't even have a use case/view for this scenario anyways, we can re-visit the idea if/when we expand subscriptions.

What do you think?

mstruve:
It would probably be a good idea to do a cleanup of data that we will no longer consider valid. If you give me a list of what needs cleaning up I can run the deletes in a console.

rhymes:
@mstruve I think I'm going to unpack this in separate PRs, it's too big if we want to track each cleanup, it's too big nonetheless to be fair :)

mstruve:
GREAT! Hit me up for reviews!

app/models/organization.rb
Zhao-Andy: I ended up adding this line to my PR for deleting orgs: https://github.com/forem/forem/pull/9357

rhymes:
We can retire this now that we are almost done with the dependent clauses. There are still some FKs to be added and optimization to be performed but totally unrelated to this PR.

So long

PR #9421, opened 2020-07-21
Display word Word Count while writing #5932

### LINKED ISSUE #5932: Display word Word Count while writing
It would be nice to have a word count displayed while the user is writing/editing an article. I feel that this would improve the entire dev.to experience, because it would keep people on dev.to for longer. I got the idea from [200wordsaday](https://cowriters.app), now *cowriters.app*.

sahaldos1:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This pull request attempts to solve issue #5932 by displaying the word count as a user writes in the editor. The counter itself is a small box that doesn't distract or take away from the rest of the page design and is intended as a helpful little bit of info for users curious to see what their word count is.

Closes #5932 

To test these changes simply attempt to create a new post and begin typing. As you type, the word counter on the bottom left of the editor will adjust, as seen below.

Screen Shot 2020-07-21 at 4 36 23 PM

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/app/assets/stylesheets/article-form-needed-legacy.scss
+++ b/app/assets/stylesheets/article-form-needed-legacy.scss
@@ -579,11 +579,7 @@
   margin-bottom: -60px;
 }
 
-#word-count-container {
-  width: 20%;
-  margin-top: 30px;
-}
-
 #word-count-text {
-  color: grey;
+  padding-left: 24%;
+  color: black;
 }
--- a/app/assets/stylesheets/article-form-needed-legacy.scss
+++ b/app/assets/stylesheets/article-form-needed-legacy.scss
@@ -578,3 +578,12 @@
   text-align: center;
   margin-bottom: -60px;
 }
+
+#word-count-container {
+  width: 20%;
+  margin-top: 30px;
+}
+
+#word-count-text {
+  color: grey;
+}
--- a/app/javascript/article-form/components/EditorActions.jsx
+++ b/app/javascript/article-form/components/EditorActions.jsx
@@ -110,6 +110,11 @@ export class EditorActions extends Component {
             Revert <span className="hidden s:inline">new changes</span>
           </Button>
         )}
+
+        <div id="word-count-text">
+          {' '}
+          Words: <span id="count"> </span>
+        </div>
       </div>
     );
   }
--- a/app/javascript/article-form/components/EditorBody.jsx
+++ b/app/javascript/article-form/components/EditorBody.jsx
@@ -44,15 +44,6 @@ export const EditorBody = ({
         }}
         name="body_markdown"
       />
-
-      <div class="crayons-card" id="word-count-container">
-        <div class="crayons-card crayons-card--secondary">
-          <div id="word-count-text">
-            {' '}
-            Words: <span id="count"> </span>
-          </div>
-        </div>
-      </div>
     </div>
   );
 };
--- a/app/javascript/article-form/components/EditorBody.jsx
+++ b/app/javascript/article-form/components/EditorBody.jsx
@@ -15,7 +15,8 @@ export const EditorBody = ({
 
   if (inputText) {
     inputText.addEventListener('keyup', function () {
-      let words = inputText.value.match(/\b[-?(\w+)?]+\b/gi);
+      let words = inputText.value.match(/([\p{L}\p{N}-]+)/gu);
+
       if (words) {
         wordCount.innerHTML = words.length;
       } else {
--- a/app/javascript/article-form/components/EditorBody.jsx
+++ b/app/javascript/article-form/components/EditorBody.jsx
@@ -9,6 +9,21 @@ export const EditorBody = ({
   switchHelpContext,
   version,
 }) => {
+  let inputText = document.getElementById('article_body_markdown');
+
+  let wordCount = document.getElementById('count');
+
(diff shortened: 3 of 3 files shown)


citizen428:
**praise:** I like this for its simplicity.

sahaldos1:
Hi @rhymes, thank you for the input. 
 
     In regard to the regex, other languages were something I didn't take into account so thank you for catching that. I managed to find a regex that accurately counts word in practically every language, so far I've tested with english, german, russian, italian, spanish, and even persian haha. The word count is now accurate and covers the kind of non ascii characters issue you brought up. The recent commit takes care of that.

     In regards to the counter placement, the counter as it is currently adjusts to the length of what the user pastes and is always a line or so below the last line the user wrote. If i understand what you want correctly, I think then the display for the counter would be in another component, specifically the editor actions component to stay in line with the buttons as you indicated?

sahaldos1:
So from what I understand after looking through the files, articleForm is the parent and contains form and editorActions. Form also contains editorBody which is where word count currently resides. To display the word count in editorActions I'm guessing I'd have to somehow send the word count up from editor body to form and then articleForm, then send that down to editorActions?

sahaldos1:
@rhymes, just pushed the final version with all the changes you requested. The word count now accounts for the quirks of other languages and their non-ascii characters and it's below and to the right of the body editor body as you requested. 

Screen Shot 2020-07-22 at 5 03 12 PM

ludwiczakpawel:
@sahaldos1 Let me get back to you in regards to this feature in general. We want to be sure we actually need/want to have this in the product because I personally think for most people it's not very useful piece of information. So I'll try to have some internal debate about this with our Product team and circle back here by the end of the week.

sahaldos1:
Sounds good @ludwiczakpawel, thank you!

b3u:
> So I'll try to have some internal debate about this with our Product team and circle back here by the end of the week.

@ludwiczakpawel friendly ping.

ludwiczakpawel:
@b3u I apologize it took so much time - we've started internal debated like exactly when you brought this issue, then we got derailed with CodeLand conf and some other internal discussions. Nevertheless you shouldn't been waiting that long so I apologize one more time!

Anyway, brining that to the main screen of the editor is not really something we wanna do - we want to keep that screen as simple as possible and with as little distractions as possible.  That being said, word counter could potentially be very useful thing, so we were thinking that maybe we could show that in the sidebar but only in the preview mode?

Eventually, we may want to "utilize" the sidebar in preview mode as a place where we add some stats and info about a post, for example:
- words count
- maybe displaying list of links user has entered in the post
- maybe being a little redundant with showing tags user has added so user can double check if these are the right ones
- and so on...

Let me know what do you think.

b3u:
@ludwiczakpawel 
**a)** I don't see the point of showing the word count in preview mode. The word count is much more useful while editing.
**b)** I can't think of many items which should go in the proposed sidebar. As you said, tags would be a bit redundant. A list of links sounds cool, but it's not so necessary. If the only important item in the sidebar is the word count, you're taking up a lot more space than you need to.

ludwiczakpawel:
@b3u Thanks for your thoughts, I appreciate it! 

> I don't see the point of showing the word count in preview mode. The word count is much more useful while editing.

I understand your point but I don't necessarily agree with it. I think it's likely useful for very small portion of users. But majority of users will not really appreciate this feature/information. And when you're writing a new post, the experience shouldn't be disrupted by anything that users don't really care about. That's one of the reasons why post editor looks so much different from other parts of DEV - no common header, footer etc. It's because user is consciously focused on specific action - so this should be as smooth and as non-disruptive experience as possible.


> I can't think of many items which should go in the proposed sidebar. As you said, tags would be a bit redundant. A list of links sounds cool, but it's not so necessary. If the only important item in the sidebar is the word count, you're taking up a lot more space than you need to.

I imagine the sidebar in the preview mode will be some sort of a _health check_ for your post - as mentioned - words count is a one thing that could go there.. maybe also reading length (ie "_9 mins read_"). As for tags, when you type them in editor you often just quickly enter them without looking at tag's description. Showing tags in this sidebar would be indeed redundant, but it would also help post author double check that the tags are correct (I imagine we would show tag WITH its description in the sidebar)...

Except "health check" this could also be a place for tips - for example when you have pretty long post with no headers, we may give a tip there like "hey it's worth splitting such a long post in smaller sections with headlines so it's easier for people to read it through" - you know what I mean?

We could also show some issues that we detect like "hey, this image doesn't have ALT attribute - you sure about that?" or "seems like you have markdown syntax issue there"..

That's just from top of my head. I'm not saying we're gonna ship everything - just giving examples of what could potentially go into that area. So things that are not necessarily crucial for the writing experience but more for the proof reading.

cmgorton:
Hey @ludwiczakpawel Any more thoughts on this PR? Wondering if we should close this and issue #5932 or if this is something we are still thinking of implementing. Thanks.

ludwiczakpawel:
@cmgorton I don't think we wanna implement this as of today without more investigation. So let me close this one, but I'll leave conversation unlocked so we can still discuss it further if needed.

PR #9422, opened 2020-07-21
Add new reserved words

benhalpern:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

We need to proactively protect the top level path early on because once we have many forems, it would be really hard to do new things off the top of the path.

We automatically protect username signups in cases of the Rails route tree, plus we have this misc list for other reserved words.

### DIFF
--- a/config/initializers/reserved_words.rb
+++ b/config/initializers/reserved_words.rb
@@ -12,7 +12,6 @@ class ReservedWords
     advertising
     amp
     anal
-    async_info
     analysis
     analytics
     answers
@@ -20,6 +19,9 @@ class ReservedWords
     app
     article
     articles
+    asset
+    assets
+    async_info
     binary
     bit
     bits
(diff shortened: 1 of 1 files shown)


config/initializers/reserved_words.rb
```diff
@@ -212,6 +212,17 @@ class ReservedWords
...
     yes
+    images
+    assets
+    meta
+    guides
+    guide
+    image
+    asset
+    config
+    styles
```
benhalpern, line 223:
```suggestion
```

config/initializers/reserved_words.rb
```diff
@@ -212,6 +212,15 @@ class ReservedWords
     welcome
     work
     yes
+    images
```
citizen428, line 215: **nitpick (non-blocking):** Should we keep this list in alphabetical order? Apart from being really good for my OCD, it would also make it easier to scan the list when adding new words later.

PR #9430, opened 2020-07-22
Rubocop: Enable and fix Style/OptionHash

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR enabled `Style/OptionHash` to remind the dev to possibly avoid generic `params` arguments and instead use keyword arguments for explicitness.

### DIFF
--- a/app/controllers/concerns/fastly_headers.rb
+++ b/app/controllers/concerns/fastly_headers.rb
@@ -8,13 +8,20 @@ module FastlyHeaders
   # Surrogate-Control is stripped at the cache, Cache-Control persists (in case of other caches in front of fastly)
   # Defaults are:
   #  Cache-Control: 'public, no-cache'
-  #  Surrogate-Control: 'max-age: 86400' - 30 days in seconds
+  #  Surrogate-Control: 'max-age: 86400' - 1 day in seconds
   # custom config example:
   #  {cache_control: 'public, no-cache, maxage=xyz', surrogate_control: 'max-age: 100'}
-  def set_cache_control_headers(max_age = 1.day.to_i, opts = {})
+  def set_cache_control_headers(
+    max_age = 1.day.to_i, cache_control: "public, no-cache",
+    surrogate_control: nil, stale_while_revalidate: nil, stale_if_error: 26_400
+  )
     request.session_options[:skip] = true # no cookies
-    response.headers["Cache-Control"] = opts[:cache_control] || "public, no-cache"
-    response.headers["Surrogate-Control"] = opts[:surrogate_control] || build_surrogate_control(max_age, opts)
+
+    response.headers["Cache-Control"] = cache_control
+
+    response.headers["Surrogate-Control"] = surrogate_control.presence || build_surrogate_control(
+      max_age, stale_while_revalidate: stale_while_revalidate, stale_if_error: stale_if_error
+    )
   end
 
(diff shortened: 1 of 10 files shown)


config/database.yml
```diff
@@ -66,6 +66,8 @@ test:
   <<: *default
   url: <%= ENV['DATABASE_URL_TEST'] || ENV['DATABASE_URL'] || ''%>
   database: PracticalDeveloper_test<%= ENV['TEST_ENV_NUMBER'] %>
+  variables:
+      statement_timeout: <%= ENV['STATEMENT_TIMEOUT'] || 10_000 %>
```
citizen428, line 70: **suggestion (non-blocking):** Maybe we should use `fetch` more:

```suggestion
      statement_timeout: <%= ENV.fetch('STATEMENT_TIMEOUT', 10_000) %>
```

Note: either version will return either a string (if the env var is set) or integers. It's not uncommon to do `ENV.fetch('STATEMENT_TIMEOUT', 10_000).to_i)` in cases like this, since it's just returns the number for integers:

> Since int is already an Integer, returns self.

app/controllers/concerns/fastly_headers.rb
```diff
@@ -8,13 +8,20 @@ module FastlyHeaders
   # Surrogate-Control is stripped at the cache, Cache-Control persists (in case of other caches in front of fastly)
   # Defaults are:
   #  Cache-Control: 'public, no-cache'
-  #  Surrogate-Control: 'max-age: 86400' - 30 days in seconds
+  #  Surrogate-Control: 'max-age: 86400' - 1 day in seconds
   # custom config example:
   #  {cache_control: 'public, no-cache, maxage=xyz', surrogate_control: 'max-age: 100'}
-  def set_cache_control_headers(max_age = 1.day.to_i, opts = {})
+  def set_cache_control_headers(
+    max_age = 1.day.to_i, cache_control: "public, no-cache",
```
citizen428, line 15: **thought (non-blocking):** This signature with 1 normal arg 4 kwargs, all with defaults and 3 of which get passed on to `build_surrogate_control` makes me believe that a controller concern maybe wasn't the right abstraction for this. Once this PR is done we may want to revisit this topic.

config/initializers/slack_notifier.rb
```diff
@@ -1,7 +1,9 @@
 class NoOpHTTPClient
+  # rubocop:disable Style/OptionHash
   def self.post(uri, params = {})
```
citizen428, line 3: **nitpick (non-blocking):** Would using `**kwargs` work here instead of `params = {}`? Then we wouldn't have to disable the cop.

app/controllers/concerns/fastly_headers.rb
atsmith813: @citizen428 @rhymes just dropping in with some quick context on this area specifically...

This was introduced to move off the since deprecated [`fastly-rails` gem](https://github.com/fastly/fastly-rails). We basically extracted the chunks we were using and brought it "in-house". We didn't go too far in trying to refactor and modify the code since it's sensitive and we're in the process of moving off of Fastly.

> a controller concern maybe wasn't the right abstraction for this

I see what you mean. That's how the gem had done it so we did a "1-for-1" copy to ensure it was still working. When we go to move off of Fastly, I'll keep this code in mind and we may improve it.

PR #9431, opened 2020-07-22
Rubocop: Move from OpenStruct to Struct in Organization

citizen428:

- [X] Refactor

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [ ] no documentation needed

alt_text

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -16,7 +16,6 @@ require:
 Performance/OpenStruct:
   Exclude:
     - 'app/models/article.rb'
-    - 'app/models/organization.rb'
     - 'app/services/notifications/reactions/send.rb'
     - 'spec/models/github_repo_spec.rb'
     - 'spec/requests/github_repos_spec.rb'
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -9,6 +9,8 @@ class Organization < ApplicationRecord
     reserved_word: "%<value>s is a reserved word. Contact site admins for help registering your organization."
   }.freeze
 
+  CachedOrganization = Struct.new(:name, :username, :slug, :profile_image_90, :profile_image_url)
+
   acts_as_followable
 
   has_many :api_secrets, through: :users
@@ -122,14 +124,14 @@ class Organization < ApplicationRecord
   def update_articles
     return unless saved_change_to_slug || saved_change_to_name || saved_change_to_profile_image
 
-    cached_org_object = {
-      name: name,
-      username: username,
-      slug: slug,
-      profile_image_90: profile_image_90,
-      profile_image_url: profile_image_url
-    }
-    articles.update(cached_organization: OpenStruct.new(cached_org_object))
+    cached_org_object = CachedOrganization.new(
+      name,
+      username,
+      slug,
+      profile_image_90,
+      profile_image_url,
+    )
+    articles.update(cached_organization: cached_org_object)
   end
 
   def bust_cache


spec/lib/data_update_scripts/update_articles_cached_organization_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      profile_image_url: org.profile_image_url
+    }
+
+    # rubocop:disable Performance/OpenStruct
+    article = create(:article, cached_organization: OpenStruct.new(cached_org))
+    # rubocop:enable Performance/OpenStruct
+
+    expect do
+      described_class.new.run
+    end.to change { article.reload.cached_organization.class }.from(OpenStruct).to(Organization::CachedOrganization)
```
rhymes, line 21: TIL `.from.to`

lib/data_update_scripts/20200723070918_update_articles_cached_organization.rb
```diff
@@ -0,0 +1,15 @@
...
+  class UpdateArticlesCachedOrganization
+    ATTRIBUTES = %i[name slug username profile_image_90 profile_image_url].freeze
+
+    def run
+      Article.find_each do |article|
+        next if article.cached_organization.blank?
+
+        old_cached_org = article.cached_organization
+        new_cached_org = Organization::CachedOrganization.new(*old_cached_org.to_h.values_at(*ATTRIBUTES))
+        article.update(cached_organization: new_cached_org)
```
rhymes, line 11: should this raise an exception?

spec/lib/data_update_scripts/update_articles_cached_organization_spec.rb
citizen428: 😃 I use that a lot, at least where it makes semantic sense for the spec, but also to e.g. make sure something is now `nil` because we cleared it:

```ruby
from(something).to(nil)
```

lib/data_update_scripts/20200723070918_update_articles_cached_organization.rb
citizen428: I'm not sure I understand the question. Does it raise an exception for you? It shouldn't.

citizen428:
Back to draft as @rhymes rightfully pointed out that there's another place in the model that writes to `cached_organization`. Will fix tomorrow.

lib/data_update_scripts/20200723070918_update_articles_cached_organization.rb
```diff
@@ -0,0 +1,15 @@
+module DataUpdateScripts
+  class UpdateArticlesCachedOrganization
+    ATTRIBUTES = %i[name slug username profile_image_90 profile_image_url].freeze
+
+    def run
+      Article.find_each do |article|
```
mstruve, line 6: To make this run faster we should filter for only articles that have organizations.

citizen428: We don't seem to index articles on `organization id`, only `user_id` so this would still be a seq scan (from Blazer):

```sql
Seq Scan on articles  (cost=0.00..207773.65 rows=23879 width=2252) (actual time=0.013..307.515 rows=17537 loops=1)
  Filter: (organization_id IS NOT NULL)
  Rows Removed by Filter: 313960
Planning Time: 0.191 ms
Execution Time: 308.365 ms
```

It would however save us from instantiating unnecessary objects and that's definitely a plus, so thanks! The new version of this script now filters by presence of  `cached_user` or `cached_organization`. Are there any statement timeouts on the live DB I need to worry about? I do hit them on Blazer, but I think that's a Blazer limit.

app/models/articles/cached_entity.rb
```diff
@@ -0,0 +1,14 @@
+module Articles
```
citizen428, line 1: I had this in `app/services` at first, but it really isn't a service. It's a simple model with a secondary constructor and during the discussions in the `Campaign` model PR ( #9300) people seemed ok with having them in `app/models`.

app/models/article.rb
```diff
@@ -585,22 +585,12 @@ def create_password
 
   def update_cached_user
     if organization
-      self.cached_organization = OpenStruct.new(set_cached_object(organization))
+      self.cached_organization = Articles::CachedEntity.from_object(organization)
```
citizen428, line 588: We now have a simple domain model with a speclalized constructor for turning anything with the right attributes (including the currently saved ostructs) into the structs that get cached.

citizen428:
Special thanks to @rhymes whose feedback based on an previous attempt to fix this was invaluable.

image

app/models/articles/cached_entity.rb
rhymes: Can't wait for the reorganization of the `services` folder :D

citizen428:
>  I still think we should migrate to JSONB columns here

I don't disagree, but I wanted to take this step by step.

rhymes:
@citizen428 oh yeah, definitely not something to be added to this PR :D one step at a time!

spec/lib/data_update_scripts/update_articles_cached_entities_spec.rb
```diff
@@ -0,0 +1,36 @@
+require "rails_helper"
+require Rails.root.join("lib/data_update_scripts/20200723070918_update_articles_cached_entities.rb")
+
+describe DataUpdateScripts::UpdateArticlesCachedEntities do
```
joshpuetz, line 4: Spec for an update script?! 👏

citizen428: It wasn't the first one in `spec/lib/data_update_scripts`, just following the good example set by @mstruve  😃

Issue #3244, opened 2019-06-20
Turn off notifications for hearts


emgoto:
**Is your feature request related to a problem? Please describe.**
I would like to turn off being notified (by the number icon in the top right corner) when I have received a heart/like on a comment or article.

**Describe the solution you'd like**

When you go to the notification settings page, there's a checkbox option to turn off receiving notifications for hearts / likes. Ideally I'd like to be notified when someone comments / replies to a comment but nothing else.

Zhao-Andy:
Thanks for the feature request! We'll discuss this as a team and leave another comment.

maestromac:
Very similar to https://github.com/thepracticaldev/dev.to/issues/3171 .

nicolas-amabile:
Hello everyone!
Here is my first try to implement this feature: https://github.com/thepracticaldev/dev.to/pull/3391
@rhymes' comment makes sense. I would like to discuss the approach and to get some help with some missing points on my PR.
- Does the same feature apply for organizations? I think it makes sense. I added the `fetch` and default value [here](https://github.com/thepracticaldev/dev.to/pull/3391/files#diff-029bf5e69e07f2f3c4eef3c94769633fR21) to handle the missing field, but I guess I need to add it to the organization as well.
- I would like to get some help with:
   - Migration script for adding `post_reaction_notifications` field
   - Tests
   - UI. Check the gif on the PR (#3391) description for the current implementation

rhymes:
@nicolas-amabile 

> Does the same feature apply for organizations? I think it makes sense. I added the fetch and default value here to handle the missing field, but I guess I need to add it to the organization as well.

It does, if you look at [ReactionsController#create](https://github.com/thepracticaldev/dev.to/blob/master/app/controllers/reactions_controller.rb#L41) method, you'll see that both the user and the organization can receive a reaction notification:

https://github.com/thepracticaldev/dev.to/blob/727d2fadd049cca6f0cda8168488d40b9c615793/app/controllers/reactions_controller.rb#L65-L66

The [fetch](https://github.com/thepracticaldev/dev.to/pull/3391/files#diff-029bf5e69e07f2f3c4eef3c94769633fR21) you added in that case wouldn't work because as we said in the PR, the field in not in the `receiver_data` hash, but on the `receiver` model

> Migration script for adding post_reaction_notifications field

You can take a look at the guide on how to create a migration for Rails: https://guides.rubyonrails.org/active_record_migrations.html#creating-a-standalone-migration

[...]

PR #9435, opened 2020-07-22
Add ability to turn off reaction notifications in notification settings

emgoto:

- [x] Feature

## Description

Adds new "General Notification Settings" section to UI, with option to turn off reaction notifications. It will be on by default.

When this checkbox is unchecked, users will no longer receive notifications for reactions.

#3244 

image

Open to changing the UI / location of the checkbox, of course!

Are notifications batched? I signed in on two separate DEV accounts on separate browsers, however I couldn't get this react notifications reliably working. i.e. I "react" to a comment and then jump onto my other browser and refresh the page, the notification doesn't always show up. Not sure if this is expected behaviour, or something is wrong with the code that I wrote!

## Added tests?

- [x] yes

## Added to documentation?

- [x] no documentation needed

### DIFF
--- a/app/views/users/_notifications.html.erb
+++ b/app/views/users/_notifications.html.erb
@@ -57,6 +57,16 @@
 
   <div class="crayons-card mb-6 grid gap-6 p-6">
     <header>
+      <h3 class="mb-2">General Notification Settings</h3>
+    </header>
+    <div class="crayons-field crayons-field--checkbox">
+      <%= f.check_box :article_reaction_notifications, class: "crayons-checkbox" %>
+      <%= f.label :article_reaction_notifications, "Send notifications when someone reacts to my comments or posts", class: "crayons-field__label" %>
+    </div>
+  </div>
+
+  <div class="crayons-card mb-6 grid gap-6 p-6">
+    <header>
       <h3 class="mb-2">Mobile Notification Settings (Beta)</h3>
       <p>Additional settings will be rolled out as new notification features are made available.</p>
     </header>
--- a/app/workers/notifications/new_reaction_worker.rb
+++ b/app/workers/notifications/new_reaction_worker.rb
@@ -13,7 +13,14 @@ module Notifications
       return unless %w[User Organization].include?(receiver_klass)
 
       receiver = receiver_klass.constantize.find_by(id: receiver_data.fetch(:id))
-      Notifications::Reactions::Send.call(reaction_data, receiver) if receiver
+      Notifications::Reactions::Send.call(reaction_data, receiver) if should_send?(receiver, receiver_klass)
+    end
+
+    def should_send?(receiver, receiver_klass)
+      return false unless receiver
+      return true if receiver_klass == "Organization"
+
+      !receiver.article_reaction_notifications
     end
   end
 end
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -1,5 +1,6 @@
 class UserPolicy < ApplicationPolicy
   PERMITTED_ATTRIBUTES = %i[
+    article_reaction_notifications
     available_for
     behance_url
     bg_color_hex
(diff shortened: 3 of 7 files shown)


rhymes:
Thanks a lot @emgoto!

> Are notifications batched? I signed in on two separate DEV accounts on separate browsers, however I couldn't get this react notifications reliably working. i.e. I "react" to a comment and then jump onto my other browser and refresh the page, the notification doesn't always show up. Not sure if this is expected behaviour, or something is wrong with the code that I wrote!

IIRC yeah, not all notifications are displayed singularly, some are aggregated. I tagged a couple of team members that likely know this logic better than I do.

cc. @ludwiczakpawel for the copy as I don't know if we want to call it "General Notification Settings" or something else

app/workers/notifications/new_reaction_worker.rb
```diff
@@ -13,7 +13,14 @@ def perform(reaction_data, receiver_data)
...
       receiver = receiver_klass.constantize.find_by(id: receiver_data.fetch(:id))
-      Notifications::Reactions::Send.call(reaction_data, receiver) if receiver
+      Notifications::Reactions::Send.call(reaction_data, receiver) if should_send?(receiver, receiver_klass)
+    end
+
+    def should_send?(receiver, receiver_klass)
+      return false unless receiver
+      return true if receiver_klass == "Organization"
+
+      !receiver.article_reaction_notifications
```
mstruve, line 23: I don't think you want the bang (!) here. By default this is true so if `receiver.article_reaction_notifications` is true then I would think `should_send?` would be true as well.

app/views/users/_notifications.html.erb
```diff
@@ -55,6 +55,16 @@
     </div>
   </div>
 
+  <div class="crayons-card mb-6 grid gap-6 p-6">
+    <header>
+      <h3 class="mb-2">General Notification Settings</h3>
+    </header>
+    <div class="crayons-field crayons-field--checkbox">
+      <%= f.check_box :article_reaction_notifications, class: "crayons-checkbox" %>
+      <%= f.label :article_reaction_notifications, "Send notifications when someone reacts to my comments or posts", class: "crayons-field__label" %>
```
vaidehijoshi, line 64:
```suggestion
      <%= f.check_box :reaction_notifications, class: "crayons-checkbox" %>
      <%= f.label :article_reaction_notifications, "Send notifications when someone reacts to my comments or posts", class: "crayons-field__label" %>
```

Since the description for this attribute includes both comments _and_ posts, I think that `reaction_notifications` might be a better name for this column, rather than the more specific `article_reaction_notifications`. That name makes it a bit more clear that the attribute applies to _all kinds_ of "reaction" notifications, IMO. Let me know what you think about the idea to rename this column :)

app/workers/notifications/new_reaction_worker.rb
emgoto: Ah yes, you're right! Thanks.

app/views/users/_notifications.html.erb
emgoto: Yep, makes sense to me.

rhymes:
Hi @emgoto, I took a look at the failing specs.

- https://travis-ci.com/github/forem/forem/jobs/371886062 - this can only be solved by merging master in (or rebasing your branch on top of master). It was solved by https://github.com/forem/forem/pull/9778

- https://travis-ci.com/github/forem/forem/jobs/371886061 - I believe this was flaky as well, thus I restarted it. If it fails again please run the failing tests locally and if they don't fail, please let us know so that we can figure out a plan to get this build green ;)

Thanks again for the contribution!

emgoto:
Hi @rhymes , I ran `notification_spec` locally on my branch and it didn't fail, but it looks like it's failing remotely.

rhymes:
@emgoto I resolved the schema.db conflict - let's see how the build fares this time

Issue #5475, opened 2020-01-13
Allow "New Navigation SideBar Links" to be defined by a site config or /internal admin


leewynne:
**Is your feature request related to a problem? Please describe.**

As a future admin of a generalised community i'd like the ability to add to the navigation sidebar links with additional community products such as books, movies etc.

**Describe the solution you'd like**

An addition to the Internal/Pages siteadmin view that allows for the creation of extra links in the navigation sidebar including:
1. Friendly name
2. Icon (the podcast icon as an example).
3. The destination page url
4. A value that sets whether the link should be shown in the sidebar 'pre auth' or not.

**Additional context**

See suggested screenshot.

<img width="209" alt="Screenshot 2020-01-13 at 09 04 15" src="https://user-images.githubusercontent.com/12694270/72243374-b3495f80-35e3-11ea-8ec7-91991ebaa1f6.png">

Similar issue here but for the key links sidebar item below it . 

https://github.com/thepracticaldev/dev.to/issues/4661

cmgorton:
Looks like this feature was added per Ridhwana's PR. If not let me know and I will open this back up. Thanks!

Issue #4661, opened 2019-10-30
Allow "key links" section to be defined in some non-hardcoded way


benhalpern:
Currently we have a couple "key links" areas. (left sidebar on home page, site footer, key links page.

They do not all show all the links (should they?). Either way, we should be able to define what these links are at the app level, both for generalization purposes and also possibly for dev.to administration simplification.

I'm not sure what the right implementation for this is.

bolariin:
Are you considering using a file for generalization? If you are considering representing the information in a file, you could use a yaml file. With a yaml file, you can represent both lists and key value pairs.

```yml
links:
  - title: About
    link: /about
  - title: Sponsors
    link: /sponsors
````

rhymes:
This could also be implemented as site configuration once this is in https://github.com/thepracticaldev/dev.to/pull/4729

benhalpern:
@bolariin I love the idea of being able to visualize and possibly edit all the config via YAML, even if we've started with the UI-first approach (which I think made sense). But in the future if we could display, edit, and diff config if people wanted to see things that way I think it would be great.

Ridhwana:
Related to https://github.com/thepracticaldev/dev.to/issues/5475

mikeydorje:
Reviewed navigation link implementation.

The `NavigationLink` model exists with full admin UI at `/admin/customization/navigation_links`.

Implemented features:
- Database model created (2020)
- Admin panel for managing links (PR #9444)
- Dynamic sidebar navigation
- Footer links
- Configurable sections, positioning, icons, visibility rules

Sidebar, footer, and navigation all pull from `NavigationLink.from_subforem`. This was fully implemented and is in production. Closing as completed.

PR #9444, opened 2020-07-22
Admin Panel for Configurable Navigational Links

Ridhwana:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
_(Backend Only, Frontend PR in a subsequent PR)_

Currently we have a couple "key links" areas. (left sidebar on home page, site footer, key links page.

We should be able to define what these links are at the app level, both for generalization purposes and also possibly for dev.to administration simplification. Admins of Forems  should have the ability to add to the navigation sidebar links with additional community products such as books, movies etc., whilst being able to remove the links that are specific to DEV or those that they do not want.

The configurable elements of a navigation item is:
- Friendly name
- Icon (the podcast icon as an example).
- The destination page url
- A value that sets whether the link should be shown in the sidebar 'pre auth' or not.

You'll see that this PR was a journey and a **huge learning experience** for me:🛣
 At first I tried to fit the navigation links within the Site Config page and model, however as I started working with the more complicated actions like "update", I began to understand @citizen428's foresighted early-on comments (see below for comments). The navigation links are not simple scalar values or key-value pairs, but it is essentially an array of hashes which adds unnecessary complexity to work with within the Site Config. 
Hence it made more sense to pull the configurable Navigation Links into it's own page (in the future we'll probably move more sections out of the site config as its getting too complicated) and backed by its own table. In this way we could also add validations on the model without needing to go through a service.

This PR covers the admin portion of adding the database table + the admin page to configure the links. A subsequent PR will remove the hardcoded links and replace them with the values form the table. I chose to separate out the PR's so we can prep the database for DEV before we update the frontend :) 

https://github.com/forem/forem/issues/4661
https://github.com/forem/forem/issues/5475

1. Go to http://localhost:3000/admin/navigation_links
2. Play around with the CRUD
3. Check that it saves correctly to the db 
 
**View:**
Screenshot 2020-10-01 at 16 07 26

**Create:**
Screenshot 2020-10-01 at 16 37 55

**Update:**
Screenshot 2020-10-01 at 16 37 48

**Delete:**
There will be a modal to verify deletion.

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Not right now

alt_text


app/models/site_config.rb
```diff
@@ -39,6 +41,11 @@ class SiteConfig < RailsSettings::Base
     members: ApplicationConfig["DEFAULT_EMAIL"]
   }
 
+  field :navigation, type: :array, default: [
+    { name: 'Reading List', url: 'abc.com', icon: LISTING_ICON },
+    { name: 'Listing', url: 'def.com', icon: MIC_ICON }
```
rhymes, line 46: can the URLs be empty?

Ridhwana: This is a good question, technically they shouldn't be empty but it would not break the interface if they are. We will have a set of default urls and navigation elements as a start.  

What are your thoughts?

rhymes: maybe we can default them to the actual `/readinglist` and `/listings` ? 

ps. `Listing` should be plural: `Listings`

Ridhwana: Haha yes definitely, this PR was just a spike to show that in theory hooking all this up works. I will definitely use sensible names, values and urls. The defaults will be as per the current dev website, not my random values  😅

Ridhwana:
> **thought (non-blocking):** While I'm ok with this approach, I'm not entirely sure how I feel about putting more than scalar values or simple key-value pairs into `SiteConfig`.
> 
> We lose the ability to easily add validations and similar things, so maybe this could also be backed by an actual table (especially if we want to add more attributes) and we do the configuration through `/internal/navigation` (similar to `/internal/pages`).

@citizen428 I had to mull over this for a bit, I agree with your thought process that we lose the ability to easily add validations, and that's actually a pretty big deal. I'm not so sure though about doing the config through `/internal/navigation`, because then we have some config living on the internal/config and others on other routes. We'd need a way to bring it all back together again. I'm conflicted on whether to move forward with this PR and what approach to use.

rhymes:
@Ridhwana what if we add https://api.rubyonrails.org/classes/ActiveModel/Validations.html ? would that work?

Ridhwana:
> @Ridhwana what if we add https://api.rubyonrails.org/classes/ActiveModel/Validations.html ? would that work?

@rhymes I don't think it would work on this current PR implementation where we have an array of hashes.

rhymes:
@Ridhwana maybe with the method `validates_each` ? IIRC `ActiveModel::Validations` has been extracted exactly for custom validations. I haven't tried though. If we were to switch to JSONB we could also use JSON schema to validate but that's advanced (see https://dev.to/kputra/rails-postgresql-jsonb-part-1-4ibg#chapter-4)

Ridhwana:
> @Ridhwana maybe with the method `validates_each` ? IIRC `ActiveModel::Validations` has been extracted exactly for custom validations for fields that can't use ActiveRecord validations. I haven't tried though. If we were to switch to JSONB we could also use JSON schema to validate but that's advanced (see https://dev.to/kputra/rails-postgresql-jsonb-part-1-4ibg#chapter-4)

Hmm I see what you mean, let me try it out tomorrow morning and feedback. Thank you @rhymes !

citizen428:
@Ridhwana 

> then we have some config living on the `internal/config` and others on other routes.

I'm not sure that's a problem, `/internal/config` is already getting very long as is and I'd personally have no objections to splitting it up. We're also going to merge `/admin` and `/internal`, so there may be a bigger restructuring in the future anyway.

Regarding `ActiveModel::Validations`: I just made a PR that uses them in case you need inspiration: https://github.com/forem/forem/pull/9534 I'd advise against trying to fit this into `SiteConfig` though. I'd probably add some sort of "service object" for this that takes care of all validations etc. and then persists the whole array when everything is ok. Let me know if you need help with the `ActiveModel` stuff, I've used it quite a bit in the past (e.g. include `ActiveModel::Naming` so form helpers etc. can infer names, etc.).

benhalpern:
@Ridhwana Since the general vibe here is that the comments are "non-blocking", I propose we move forward and get this shipped.

In the vein of low-hanging fruit as we get new forems live, this strikes me as a good thing to get into prod. Let me know if I can be helpful pushing this over the finish line.

Ridhwana:
@benhalpern agreed, I had a chat with @citizen428 during cooldown and we had determined a way forward. I changed priority due to the start of the cycle, but I'll pop on this on my queue next week.  👍🏽

mstruve:
Hey @Ridhwana! What are your thoughts on this PR? Is this something we want to get pushed out or should we close it for now?

citizen428:
@mstruve This is mostly done. It was already mostly done before the last cycle started, but then was "parked" since it's a cooldown project. Given that people inquired about it twice during the previous cycle, we shoul push it over the finish line instead of closing it. I'll see if I'll get to it while @Ridhwana is away next week, if not hopefully she can finish it during the last week of the cooldown.

citizen428:
Looking back at the discussion above, `validates_each` as suggested by @rhymes isn't what we're looking for. According to [the documentation](http://api.rubyonrails.org/classes/ActiveModel/Validations/ClassMethods.html#method-i-validates_each), its purpose is to validate a series of attributes against a block, not multiple objects:

```ruby
Validates each attribute against a block.
```

@Ridhwana I propose you ignore my comment above for now, and we merge a bare-bones first working version of this and then incrementally improve it.

app/services/site_configs/update_navigation.rb
```diff
@@ -0,0 +1,53 @@
+module SiteConfigs
+  class UpdateNavigation
+    class NavigationLink
```
citizen428, line 3: Using this class to have a home for the actual validations.

app/services/site_configs/update_navigation.rb
```diff
@@ -0,0 +1,53 @@
...
+      include ActiveModel::Validations
+
+      URI_REGEXP = URI::DEFAULT_PARSER.make_regexp(%w[http https]).freeze
```
citizen428, line 6: This regex could be better, e.g. `https://` is valid according to it since almost all parts are optional.

app/services/site_configs/update_navigation.rb
```diff
@@ -0,0 +1,53 @@
...
+      SVG_REGEXP = /\A<svg .*>/i.freeze
+
+      validates :name, :url, :icon, presence: true
+      validates :url, format: URI_REGEXP
+      validates :icon, format: SVG_REGEXP
+
+      def self.from_hash(hash)
```
citizen428, line 13: Custom initializer to simplify creating these objects.

app/services/site_configs/update_navigation.rb
```diff
@@ -0,0 +1,53 @@
...
+      @navigation_links = navigation_links
+      @errors = {}
+    end
+
+    def call
+      @navigation_links.each_with_index do |attributes, index|
+        navigation_link = NavigationLink.from_hash(attributes)
+        next if navigation_link.valid?
+
+        @errors[index] = navigation_link.errors.to_a
```
citizen428, line 42: The errors are key with their array index, e.g. the errros of the second element will go to index 1.

app/views/admin/profile_fields/_add_group_modal.html.erb
```diff
@@ -1,4 +1,4 @@
-<button class="btn btn-secondary" type="button" onclick="showAddGroupModal()">
+<button class="btn btn-primary" type="button" onclick="showAddGroupModal()">
```
mstruve, line 1: This seems a little random, was this change intended?

db/migrate/20201001095312_create_navigation_links.rb
```diff
@@ -0,0 +1,12 @@
+class CreateNavigationLinks < ActiveRecord::Migration[6.0]
+  def change
+    create_table :navigation_links do |t|
```
mstruve, line 3:
```
validates :name, :url, :icon, presence: true
```
Can we add these validations at the database level as well?

mstruve (commented):
CONGRATS on coming to a simpler solution, I have been down that road before, its always nice to see it all pay off in the end once you have overhauled your strategy. this looks really clean!

app/views/admin/profile_fields/_add_group_modal.html.erb
Ridhwana: Yes 🙈, I just slipped it in sneakily because I noticed that all the other add buttons on the admin interface had a primary class name.

db/migrate/20201001095312_create_navigation_links.rb
Ridhwana: @mstruve I've updated the migration here to allow for validation and I've just combined the two migration files into one whilst nobody has really pulled down the branch as yet and altered their local db.  Does this look good?

Ridhwana:
> CONGRATS on coming to a simpler solution, I have been down that road before, its always nice to see it all pay off in the end once you have overhauled your strategy. this looks really clean!

benhalpern:
Woohoo, congrats on getting the PR open, taking a deeper look now. I just re-ran the test which seems like an unrelated failure.

mstruve (approved):
pulled down and tested and this works great!

app/models/navigation_link.rb
```diff
@@ -0,0 +1,8 @@
+class NavigationLink < ApplicationRecord
```
rhymes, line 1: this is missing one validator: `validates :requires_auth, inclusion: { in: [true, false]}`

Learned recently that is important :D

app/controllers/admin/navigation_links_controller.rb
```diff
@@ -0,0 +1,49 @@
+module Admin
+  class NavigationLinksController < Admin::ApplicationController
+    ALLOWED_PARAMS = %i[
+      name url icon requires_auth
+    ].freeze
+    layout "admin"
+
+    def index
+      @navigation_links = NavigationLink.all.order(:name)
```
rhymes, line 9: FYI `.all` is redundant in this case:

```ruby
[5] pry(main)> NavigationLink.all.order(:name).to_sql
=> "SELECT \"navigation_links\".* FROM \"navigation_links\" ORDER BY \"navigation_links\".\"name\" ASC"
[6] pry(main)> NavigationLink.order(:name).to_sql
=> "SELECT \"navigation_links\".* FROM \"navigation_links\" ORDER BY \"navigation_links\".\"name\" ASC"
```

app/controllers/admin/navigation_links_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      else
+        flash[:error] = "Error: #{navigation_link.errors_as_sentence}"
+      end
+      redirect_to admin_navigation_links_url
+    end
+
+    private
+
+    def navigation_link_params
+      allowed_params = ALLOWED_PARAMS
```
rhymes, line 45: this alias is probably redudant, you can pass the constant to `.PERMIT`

app/views/admin/navigation_links/_add_navigation_link_modal.html.erb
```diff
@@ -0,0 +1,54 @@
...
+                <%= form.label :name %>
+                <%= form.text_field :name, class: "form-control" %>
+              </div>
+              <div class="form-group">
+                <%= form.label :url %>
+                <%= form.text_field :url, class: "form-control" %>
+              </div>
+              <div class="form-group">
+                <%= form.label :icon %>
+                <%= form.text_field :icon, class: "form-control" %>
```
rhymes, line 30: I think we should say that the icon should be a SVG in the explanation, as I obviously used a png and had to start from scratch when presented the error :( 
This should also be a text area I guess...

app/models/navigation_link.rb
```diff
@@ -0,0 +1,8 @@
...
+  URI_REGEXP = URI::DEFAULT_PARSER.make_regexp(%w[http https]).freeze
+  SVG_REGEXP = /\A<svg .*>/i.freeze
```
rhymes, line 3: this is not going to work in all cases, as SVGs can have the XML tags at the start:

```xml
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg>
```

and still be valid. also They are also still valid if they start with a newline:

```xml

<svg ... />
```

It's probably better to just check if the svg tag is in there :D

app/models/navigation_link.rb
```diff
@@ -0,0 +1,8 @@
...
+
+  validates :name, :url, :icon, presence: true
+  validates :url, format: URI_REGEXP
```
rhymes, line 6: why not use the URI validator 😊 ?

For example:

```ruby
validates :url, url: { schemes: %w[https http] }
```

app/views/admin/navigation_links/_add_navigation_link_modal.html.erb
```diff
@@ -0,0 +1,54 @@
...
+              </div>
+              <div class="form-group">
+                <%= form.check_box :requires_auth %>
+                <%= form.label :requires_auth %>
```
rhymes, line 34: A question: what does requiring authentication mean? Is it authentication or authorization?

app/views/admin/navigation_links/_edit_navigation_link_modal.html.erb
```diff
@@ -0,0 +1,51 @@
+<div onclick="toggleEditNavigationLinkModal('<%= link[:id] %>')" class="color-accent-warning-darker cursor-pointer">
```
rhymes, line 1: it looks like this is exactly like add except for the title adn the onclick, maybe they can be generalized?

spec/models/navigation_link_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+RSpec.describe NavigationLink, type: :model do
+  describe "validations" do
+    describe "presence validations" do
+      it { is_expected.to validate_presence_of(:name) }
+      it { is_expected.to validate_presence_of(:url) }
+      it { is_expected.to validate_presence_of(:icon) }
+    end
+
+    describe "regex vaidations" do
```
rhymes, line 11: typo

rhymes (commented):
I have a generic question: is there a reason why the creation page is a modal and not a standard page? I tried to add an invalid value and I had to start from scratch when I was presented the error.

Thanks for kickstarting this @Ridhwana. I left a bunch of notes in the review!

app/models/navigation_link.rb
Ridhwana: ✅

app/controllers/admin/navigation_links_controller.rb
Ridhwana: Urgh I've done this before and you corrected it, and I did it again 🤦🏽‍♀️ ✅

app/controllers/admin/navigation_links_controller.rb
Ridhwana: ✅

app/views/admin/navigation_links/_add_navigation_link_modal.html.erb
Ridhwana: Definitely, in the PR I'm working on currently, I've added an explanation for the page and for the fields. I wanted to get the foundation working in this PR. And thanks, changed to text_area ✅

app/views/admin/navigation_links/_add_navigation_link_modal.html.erb
Ridhwana: its authentication, I'll change this to requires_sign_in, it seems more user friendly to an admin. I'm jsut gonna rollback my migration and update the current one instead of creating a new one as its still in development, you may need to do the same if you've pulled down this branch :( sorry

app/models/navigation_link.rb
Ridhwana: ✅

app/views/admin/navigation_links/_edit_navigation_link_modal.html.erb
Ridhwana: I'll do this in a subsequent PR. I've made a note of it.

spec/models/navigation_link_spec.rb
Ridhwana: 🙈✅

app/models/navigation_link.rb
Ridhwana: Changed ✅

Screenshot 2020-10-02 at 13 26 07

Ridhwana:
> I have a generic question: is there a reason why the creation page is a modal and not a standard page? I tried to add an invalid value and I had to start from scratch when I was presented the error.
> 
> Thanks for kickstarting this @Ridhwana. I left a bunch of notes in the review!

Hey @rhymes thanks so much for the review, I updated the PR, can you please review again? 😊 I'm sorry if i messed up your db possibly, I didn't want to add unnecessary migration files so I just rolled it back and created it again. 

The reason I kept it to a modal is because we did something similar for the profile fields and it felt like that was the direction we moving towards. I like the UX of a modal because a user never needs to leave the page to CRUD, but you make a valid point about the errors, its frustrating to lose your input. I wonder if there's some way around that...I'll investigate

rhymes (approved):
Looks nice! I didn't remember the profile fields page was done with modals, I just expected the page to work like the rest of the admin with the good old pages :P 

Just a note, do you think, a11y wise, that the modals are fine? As @nickytonline has been spending time improving a11y of the front website I think we should keep that in mind when we add features to the backoffice as well

Ridhwana:
> Looks nice! I didn't remember the profile fields page was done with modals, I just expected the page to work like the rest of the admin with the good old pages :P
> 
> Just a note, do you think, a11y wise, that the modals are fine? As @nickytonline has been spending time improving a11y of the front website I think we should keep that in mind when we add features to the backoffice as well

@rhymes this is something I will double check and also speak to Nick about :) Thanks!

PR #9447, opened 2020-07-22
Bug Fix:Validate User Password Length is Between 8-100 characters

mstruve:
- [x] Bug Fix

## Description
Validate that a user's password is between 8 and 100 characters long. While setting this up I noticed that submitting an invalid input on the account page of settings would cause a 500 since `@tab` is not set on that page. Rather than try to dig into why I just put in a little workaround. 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -61,7 +61,13 @@ class UsersController < ApplicationController
     else
       Honeycomb.add_field("error", @user.errors.messages.reject { |_, v| v.empty? })
       Honeycomb.add_field("errored", true)
-      render :edit, status: :bad_request
+
+      if @tab
+        render :edit, status: :bad_request
+      else
+        flash[:error] = @user.errors.full_messages.join(", ")
+        redirect_to "/settings"
+      end
     end
   end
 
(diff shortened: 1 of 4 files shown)


app/controllers/users_controller.rb
```diff
@@ -61,7 +61,13 @@ def update
     else
       Honeycomb.add_field("error", @user.errors.messages.reject { |_, v| v.empty? })
       Honeycomb.add_field("errored", true)
-      render :edit, status: :bad_request
+
+      if @tab
+        render :edit, status: :bad_request
+      else
+        flash[:error] = @user.errors.full_messages.join(", ")
```
rhymes, line 68: You can use `@user.errors_as_sentence` here

app/controllers/users_controller.rb
```diff
@@ -61,7 +61,13 @@ def update
...
+        redirect_to "/settings"
```
rhymes, line 69: you can use `settings_path` here

app/controllers/users_controller.rb
mstruve: I knew there was a better shortcut!!!

citizen428: @mstruve Rails offers `to_sentence` as an alternative to `join(", ")`. During a previous cooldown I wrapped it as  `ApplicationRecord#errors_as_sentence` so we standardize on a snigle format. :smiley:

PR #9458, opened 2020-07-23
Validate password confirmation when updating password

citizen428:
## What type of PR is this?

- [x] Bug Fix

## Description

We recently added passwords to our `User` model. However, we never validated the password confirmation when updating the password, so users could accidentally set them to unexpected values:

```ruby
user.reset_password("newpassword", "nupassword")
#=> true
```

This PR fixes this by adding a validation:

```ruby
user.reset_password("newpassword", "nupassword")
#=> false
user.errors_as_sentence
#=> "Password doesn't match password confirmation"
```

Note: I originally tried to use `Devise::Models::Validatable`, however, this expects the password to be non-blank, despite the following:

```ruby
  validates :password, length: { in: 8..100 }, allow_nil: true
```

I thought we can fix this by overriding `password_required?`, but as [the documentation states](https://rubydoc.info/github/plataformatec/devise/master/Devise%2FModels%2FValidatable:password_required%3F):

> Passwords are always required if it's a new record, or if the password or confirmation are being set somewhere.

1. Go to `/settings/account` in your local dev environment.
2. Set the password and confirmation to different values and submit the form.
3. You should see the following error:
    Screen Shot 2020-07-23 at 11 20 14
4. Now set the passwords to the same value and submit the form again.
5. You should see a success message:
    Screen Shot 2020-07-23 at 11 21 09

## Added tests?

- [x] yes

## Added to documentation?

- [x] no documentation needed

image

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -160,6 +160,7 @@ class User < ApplicationRecord
   validate :validate_mastodon_url
   validate :can_send_confirmation_email
   validate :update_rate_limit
+  validate :password_matches_confirmation, if: :encrypted_password_changed?
 
   alias_attribute :public_reactions_count, :reactions_count
   alias_attribute :subscribed_to_welcome_notifications?, :welcome_notifications
@@ -644,5 +645,11 @@ class User < ApplicationRecord
     rate_limiter.check_limit!(:user_update)
   rescue RateLimitChecker::LimitReached => e
     errors.add(:base, "User could not be saved. #{e.message}")
+  end
+
+  def password_matches_confirmation
+    return true if password == password_confirmation
+
+    errors.add(:password, "doesn't match password confirmation")
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -160,6 +160,8 @@ class User < ApplicationRecord
   validate :validate_mastodon_url
   validate :can_send_confirmation_email
   validate :update_rate_limit
+  # NOTE: when updating the password on a Devise enabled model, the :encrypted_password
+  # field will be marked as dirty, not :password.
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -160,6 +160,7 @@ class User < ApplicationRecord
   validate :validate_mastodon_url
   validate :can_send_confirmation_email
   validate :update_rate_limit
+  validate :password_matches_confirmation, if: :encrypted_password_changed?
```
citizen428, line 163: This is an implementation detail of how Devise works, we have to check for `encrypted_password_changed?` not `password_changed?`:

```ruby
user.password = 'muchsecure'
user.password_changed?
# NoMethodError: undefined method `password_changed?' for #<User:0x000000011e7bf100>
user.encrypted_password_changed?
#=> true
```

rhymes: can you add a comment there to explain this?

app/models/user.rb
```diff
@@ -645,4 +646,10 @@ def update_rate_limit
   rescue RateLimitChecker::LimitReached => e
     errors.add(:base, "User could not be saved. #{e.message}")
   end
+
+  def password_matches_confirmation
+    return true if password == password_confirmation
```
rhymes, line 651: can we use secure compare here? 

https://github.com/forem/forem/blob/44b8d0894d00ceb78aa30a35c6bfe62536e3c90b/app/controllers/api/v0/api_controller.rb#L71-L73

citizen428: `secure_compare` prevents timing attacks when a user-provided value is hashed and compared to a stored hash. This is not the case here, both values are user provided and we compare them against each other, not against something stored in the DB.

PR #9469, opened 2020-07-23
Redirect override

fdoxyz:
## Codeland PR

Connect redirect issue when following invites


fdoxyz:
Just testing out locally for a sec before merging 🚨

rhymes:
The cause for this is that we have two different paths mapped on the same helper:

https://github.com/forem/forem/blob/9bdef4d4ae0b41612001a62c2409121b654bf71f/config/routes.rb#L287-L288

PR #9476, opened 2020-07-23
Optimization: Change Large Table Primary Keys to Bigints

mstruve:
- [x] Optimization

## Description
Plan:
After running all of these migrations one by one and [gathering timing information](https://gist.github.com/mstruve/1600b29a30b39b9f4edc0c8d79d9ba91) we have decided to run them all during a low traffic period and display a banner on the site notifying users that they may experience degraded performance for specific parts of the application. Here is the timing breakdown with performance implications
```
# ahoy_messages
# seconds: 1385.38
# Nothing in the UI seemed to break

# articles
# seconds: 201.74
# everything breaks except connect and notifications pages still work. Can still see banner! 

# notifications
# seconds: 2295.07
# Comments created but submitting breaks so it looks like it was not created
# notification pages break
```
Example Banner that we used for another set of migrations except we will explicitly call out for example when notifications are going that people should expect the notification page to be offline. 

Screen Shot 2020-08-06 at 6 41 22 PM

You will also notice I am removing the hypershield views before running the migrations! I have noticed in the past and while doing some of the manual migrations(without the change_column helper) that the additional views caused the migrations to fail or timeout. When the migration is complete hypershield will be refreshed anyways so those views will get added back.

## QA/Review

PLEASE double check that I did not miss any integer columns on these large tables bc I would hate to have to do this twice! I actually missed the feedback_message_id on ahoy messages on my first pass. During my last pass through the schema I noticed I missed a few other integer IDs but all are for small tables that I can easily pick off in another PR. Here are what will remain after this PR
```
"api_secrets" - "user_id"
"badge_achievements" - "rewarder_id"
"broadcasts" - "broadcastable_id"
"display_ads" - "organization_id"
"feedback_messages" - "affected_id", "offender_id", "reporter_id"
"html_variant_successes" - "article_id", "html_variant_id"
"html_variant_trials" - "article_id", "html_variant_id"=
"users_roles" - "id", "user_id"
```

alt_text

### DIFF
--- a/db/migrate/20200806193438_large_table_primary_keys_to_bigints.rb
+++ b/db/migrate/20200806193438_large_table_primary_keys_to_bigints.rb
@@ -1,6 +1,6 @@
 class LargeTablePrimaryKeysToBigints < ActiveRecord::Migration[6.0]
   def up
-    puts "ahoy_messages"
+    puts "migrating ahoy_message PKs to bigints"
     ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.ahoy_messages")
     ActiveRecord::Base.connection.execute(
       <<-SQL
@@ -11,7 +11,7 @@ class LargeTablePrimaryKeysToBigints < ActiveRecord::Migration[6.0]
       SQL
     )
 
-    puts "articles"
+    puts "migrating article PKs to bigints"
     ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.articles")
     ActiveRecord::Base.connection.execute(
       <<-SQL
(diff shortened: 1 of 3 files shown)


db/migrate/20200806193438_large_table_primary_keys_to_bigints.rb
```diff
@@ -0,0 +1,74 @@
+class LargeTablePrimaryKeysToBigints < ActiveRecord::Migration[6.0]
+  def up
+    puts "ahoy_messages"
```
mstruve, line 3: The last migration I ran with custom statements output nothing and it killed me not knowing where we were in the progress. This will allow me to follow along and update the site banner as needed.

citizen428: Not that it matters much, but maybe we can add `if Rails.env.production?` so people don't wonder where this messages come from in their dev environemnts?

mstruve: Better yet, I made them more explicit `migrating ahoy_message PKs to bigints`

PR #9477, opened 2020-07-23
Optimization:Change Mentions, Notes, Podcasts, Roles Related Ids To Bigints

mstruve:
- [x] Optimization

## Description
Update a few more keys on these smaller tables to bigints. All of these tables have 10k or fewer rows which we have seen can be updated very quickly inline. 

You will also notice I am removing the hypershield views before running the migrations! I have noticed in the past and while doing some of the manual migrations(without the `change_column` helper) that the additional views caused the migrations to fail or timeout. When the migration is complete hypershield will be refreshed anyways so those views will get added back. 

alt_text

### DIFF
--- a/db/migrate/20200723203155_change_mentions_notes_podcasts_roles_related_ids_to_bigints.rb
+++ b/db/migrate/20200723203155_change_mentions_notes_podcasts_roles_related_ids_to_bigints.rb
@@ -1,0 +1,41 @@
+class ChangeMentionsNotesPodcastsRolesRelatedIdsToBigints < ActiveRecord::Migration[6.0]
+  def up
+    ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.mentions")
+    safety_assured { change_column :mentions, :user_id, :bigint }
+
+    ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.notes")
+    ActiveRecord::Base.connection.execute(
+      <<-SQL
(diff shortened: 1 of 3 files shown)


db/migrate/20200723203155_change_mentions_notes_podcasts_roles_related_ids_to_bigints.rb
```diff
@@ -0,0 +1,41 @@
+class ChangeMentionsNotesPodcastsRolesRelatedIdsToBigints < ActiveRecord::Migration[6.0]
+  def up
+    ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.mentions")
+    safety_assured { change_column :mentions, :user_id, :bigint }
+
+    ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.notes")
+    ActiveRecord::Base.connection.execute(
```
citizen428, line 7: Somewhat bizarrely Ruby also accepts the following way of formatting heredocs as arguments:

```ruby
ActiveRecord::Base.connection.execute(<<-SQL)
  ALTER TABLE notes
    ALTER COLUMN author_id TYPE bigint,
    ALTER COLUMN noteable_id TYPE bigint
SQL

#<PG::Result:0x00000001154e5320 status=PGRES_COMMAND_OK ntuples=0 nfields=0 cmd_tuples=0>
```

It's always been my preferred form since I find it reads cleaner. 

No need to change was only meant as an FYI.

PR #9491, opened 2020-07-24
Optimization:Index Chat Channel Memberships Async on Message Creation

mstruve:
- [x] Optimization

## Description
When a message is sent to a channel that has a very large number of members it can take us seconds to loop through all of those and enqueue indexing jobs for them which means it takes seconds for us to complete creating the message. The only thing being updated for the memberships is the last message sent at. This value on the memberships helps us order the chat channels in connect and I don't think we need it updated in realtime which is why I created a job to update those records asynchronously. 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -27,7 +27,7 @@ class Message < ApplicationRecord
 
   def update_chat_channel_last_message_at
     chat_channel.touch(:last_message_at)
-    chat_channel.chat_channel_memberships.each(&:index_to_elasticsearch)
+    ChatChannels::IndexesMembershipsWorker.perform_async(chat_channel.id)
   end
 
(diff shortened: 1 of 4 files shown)


app/models/message.rb
```diff
@@ -195,7 +195,7 @@ def determine_user_validity
   def channel_permission
     errors.add(:base, "Must be part of channel.") if chat_channel_id.blank?
 
-    channel = ChatChannel.find(chat_channel_id)
+    channel = chat_channel || ChatChannel.find(chat_channel_id)
```
mstruve, line 198: Snuck in this little optimization to avoid redundantly looking up the chat_channel for existing messages

PR #5460, opened 2020-01-12
Feature🚀: Ability to mention anyone in open chats and notify them

sarthology:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Pheww! After working and tweaking for more than a week this feature is finally out. This is description will be a bit longer, as this PR has some bug fixes as well. So Let's get started 

#### 1. Ability to mention someone in open group

mention suggestion

#### 2. Get a notification if you got mentioned

ezgif com-gif-maker-2

#### 3. Ability to mention all in the group and Notify them
ezgif com-video-to-gif-4

#### 4. Mentions get converted into a clickable link to open info in Sidecar
ezgif com-video-to-gif-5

#### 5. Message count will increase if there is a new message in the open group.
*You will get notified only if mentioned in the group*

ezgif com-video-to-gif-6

#### 6. Message count will be decreased only if the unread message is read
ezgif com-video-to-gif-7

#### 7. Message chat will increase even if you are on the connect page 
*But no notification will be shown*
ezgif com-video-to-gif-8

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

## [optional] What gif best describes this PR or how it makes you feel?

alt_text

PR #9493, opened 2020-07-24
Optimization/Bug Fix:Only Send Mentioned User Notifications, not all for every message

mstruve:
- [x] Optimization
- [x] Bug fix

## Description
I am pretty sure this is a bug but let me know if I am missing something! In [this PR](https://github.com/forem/forem/pull/5460) by @sarthology we set up the ability to notify all users if `@all` is mentioned in a channel. This works by using [getMentionedUsers](https://github.com/forem/forem/blob/9bdef4d4ae0b41612001a62c2409121b654bf71f/app/javascript/chat/chat.jsx#L1562) method to set `mentioned_user_id` to an array of all the users in a channel if `@all` is detected. 

However, if we then go into the controller we notice that instead of just sending notifications for `mentioned_user_id`, we are sending notifications to ALL users every single time. 
```ruby
      notify_users(@message.chat_channel.channel_users_ids, "all")
      notify_users(@mentioned_users_id, "mention")
```
I removed the first line since our javascript already handles the `@all` scenario for us. This will save us A LOT of notifications and dramatically speed up message sending and creation. 

alt_text


mstruve:
@rhymes and @nickytonline additional specs have been added

PR #9513, opened 2020-07-25
Optimization:Change Multiple Small Table Column IDs to Bigints

mstruve:
- [x] Optimization

## Description
Changing the last of the small table keys from ints to Bigints. All of these tables when I ran these migrations on a copy of prod ran in a second or under. The last few small table migrations I have to run beyond these are around 5-10 second migrations. Then last but not least we have long migrations(#9476)  which I plan to schedule downtime for. 
Prod Data Sizes:
collections - 7k
buffer_updates - 55k ([1 second to update id, .5 seconds to update the other columns](https://gist.github.com/mstruve/1600b29a30b39b9f4edc0c8d79d9ba91))
github_repos - 35k
html_variants - 128
mentions - 6k
organizations - 2k
tweets - 11k

For more information on the spec fix checkout #9512 

alt_text

### DIFF
--- a/spec/system/articles/user_visits_an_article_spec.rb
+++ b/spec/system/articles/user_visits_an_article_spec.rb
@@ -1,8 +1,8 @@
 require "rails_helper"
 
 RSpec.describe "Views an article", type: :system do
-  let_it_be(:user) { create(:user) }
-  let_it_be_changeable(:article) do
+  let(:user) { create(:user) }
+  let(:article) do
     create(:article, :with_notification_subscription, user: user)
   end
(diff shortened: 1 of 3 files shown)


spec/system/articles/user_visits_an_article_spec.rb
```diff
@@ -1,8 +1,8 @@
 require "rails_helper"
 
 RSpec.describe "Views an article", type: :system do
-  let_it_be(:user) { create(:user) }
-  let_it_be_changeable(:article) do
+  let(:user) { create(:user) }
```
mstruve, line 4: For more information on this spec fix checkout #9512

PR #9528, opened 2020-07-26
homepageFeed: Add null checks in cases where there's no featured story present

kirananto:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

During the initial setup, if there are no featuredArticles, the UI breaks. so to avoid this particular scenario I'm adding null checks at appropriate places.

N/A

Try setting up a brand new dev environment without any featured article, to reproduce this particular bug

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

N/A

##  What gif best describes this PR or how it makes you feel?

Undefined error

### DIFF
--- a/app/javascript/packs/homePageFeed.jsx
+++ b/app/javascript/packs/homePageFeed.jsx
@@ -82,21 +82,23 @@ export const renderFeed = (timeFrame) => {
 
         const [featuredStory, ...subStories] = feedItems;
         const feedStyle = JSON.parse(document.body.dataset.user).feed_style;
-
-        sendFeaturedArticleAnalytics(featuredStory.id);
+        if(featuredStory) {
+          sendFeaturedArticleAnalytics(featuredStory.id);
+        }
 
         // 1. Show the featured story first
         // 2. Podcast episodes out today
         // 3. Rest of the stories for the feed
         return (
           <div>
+            {featuredStory && 
             <Article
               {...commonProps}
               article={featuredStory}
               isFeatured
               feedStyle={feedStyle}
               isBookmarked={bookmarkedFeedItems.has(featuredStory.id)}
-            />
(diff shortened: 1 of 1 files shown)


citizen428:
> no, because they aren't needed

For bug fixes I'd very much like to see a regression spec.

kirananto:
I'm still new to the codebase, It'd be nice if you could direct me to where the spec files are for the homepageFeed.jsx
I couldn't find one for the homepage, Frontend part.

citizen428:
Tbh, I'm not overly familiar with the JS part of the app myself, it was more a general comment. @nickytonline should know better, or maybe not even want a spec for this.

nickytonline:
There is supposed to always be a featured article. Even with the seed data we always end up with one. I'm trying to remember the criteria in the backend for this though. @joshpuetz would know.

joshpuetz:
That's right: we also set the featured story to SOMETHING on the backend:

https://github.com/forem/forem/blob/master/app/services/articles/feed.rb#L22

I'm certainly not opposed to checks on the client though: I'm curious if a missing featured story is a hypothetical situation, or something you're seeing running locally @kirananto ?

kirananto:
@joshpuetz  Nope, it is definitely something that I see on my local, There was no seed data on initial setup, So I manually created a post, but then it was not showing up due to this breaking... 

I suggest you have this check just to make sure UI doesn't break at any point. I usually try to make clients as fail-proof as possible, and in such cases where you expect it to be there and it doesn't we can always push an analytics/sentry data to inform the concerned team right? So that we can handle it better next time 👍

PR #9529, opened 2020-07-26
Flaky Spec Fix:Remove let_it_be usage from Comment Model Spec

mstruve:
- [x] Flaky Spec Fix

## Description
These specs I saw fail a couple of times in the dependabot update branches so I went ahead and removed the problematic `let_it_be` statements. For more information checkout #9512 

I think one of the dependabot updates must have missed vendoring a gem since after updating with master and bundling I seem to have an extra file. 

alt_text

### DIFF
--- a/spec/models/comment_spec.rb
+++ b/spec/models/comment_spec.rb
@@ -1,9 +1,9 @@
 require "rails_helper"
 
 RSpec.describe Comment, type: :model do
-  let_it_be(:user) { create(:user) }
-  let_it_be(:article) { create(:article, user: user) }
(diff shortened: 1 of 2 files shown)


rhymes:
@mstruve the reason why you have this gem is because dependabot runs on Linux, but the libv8 is platform dependant, so the first among us, using a macOS machine, that runs `bundle install` is bound to have this new `libv8-darwin.gem` file and has to add it to Git :D

mstruve:
@rhymes OH! Thank you for the insight!

PR #9535, opened 2020-07-27
Activate Devise::Lockable

citizen428:

- [X] Feature

## Description

It was recently brought to our attention that we're not locking out users who attempt to conduct dictionary and brute force attacks against user accounts. This PR activates `Devise::Lockable` to address this.

1. Make sure your user has a password that you know:
    ```ruby
    user.update(password: "...", password_confirmation: "...")
   ```
2. Go to `/users/sign_in` or `/enter` in your local environment.
3. Enter the wrong password a couple of times and see your account get locked out.
4. You can unlock it in the console with `user.unlock_access!`.

## Added tests?

- [X] yes

## Added to documentation?

- [X] no documentation needed

### DIFF
--- a/app/services/authentication/authenticator.rb
+++ b/app/services/authentication/authenticator.rb
@@ -127,6 +127,7 @@ module Authentication
 
     def update_user(user)
       user.tap do |model|
+        user.unlock_access! if user.access_locked?
         user.assign_attributes(provider.existing_user_data)
 
         update_profile_updated_at(model)
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -206,7 +206,7 @@ Devise.setup do |config|
   # Defines which strategy will be used to lock an account.
   # :failed_attempts = Locks an account after a number of failed attempts to sign in.
   # :none            = No lock strategy. You should handle locking by yourself.
-  # config.lock_strategy = :failed_attempts
+  config.lock_strategy = :failed_attempts
 
   # Defines which key will be used when locking and unlocking an account
   # config.unlock_keys = [:email]
@@ -216,14 +216,14 @@ Devise.setup do |config|
   # :time  = Re-enables login after a certain amount of time (see :unlock_in below)
   # :both  = Enables both strategies
   # :none  = No unlock strategy. You should handle unlocking by yourself.
-  # config.unlock_strategy = :both
+  config.unlock_strategy = :both
 
   # Number of authentication tries before locking an account if lock_strategy
   # is failed attempts.
-  # config.maximum_attempts = 20
+  config.maximum_attempts = 3
 
   # Time interval to unlock the account if :time is enabled as unlock_strategy.
-  # config.unlock_in = 1.hour
+  config.unlock_in = 1.hour
 
   # Warn on the last attempt before the account is locked.
   # config.last_attempt_warning = true
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -220,7 +220,7 @@ Devise.setup do |config|
 
   # Number of authentication tries before locking an account if lock_strategy
   # is failed attempts.
-  config.maximum_attempts = 3
+  config.maximum_attempts = 5
 
   # Time interval to unlock the account if :time is enabled as unlock_strategy.
   config.unlock_in = 1.hour
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -472,6 +472,7 @@ ActiveRecord::Schema.define(version: 2020_07_27_163200) do
     t.datetime "updated_at", null: false
     t.bigint "user_id"
     t.datetime "verified_at"
+    t.index ["user_id"], name: "index_email_authorizations_on_user_id"
   end
 
   create_table "events", force: :cascade do |t|
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -472,7 +472,6 @@ ActiveRecord::Schema.define(version: 2020_07_27_163200) do
     t.datetime "updated_at", null: false
     t.bigint "user_id"
     t.datetime "verified_at"
-    t.index ["user_id"], name: "index_email_authorizations_on_user_id"
   end
 
   create_table "events", force: :cascade do |t|
@@ -1197,6 +1196,7 @@ ActiveRecord::Schema.define(version: 2020_07_27_163200) do
     t.boolean "export_requested", default: false
     t.datetime "exported_at"
     t.string "facebook_url"
+    t.integer "failed_attempts", default: 0
     t.boolean "feed_admin_publish_permission", default: true
     t.datetime "feed_fetched_at", default: "2017-01-01 05:00:00"
     t.boolean "feed_mark_canonical", default: false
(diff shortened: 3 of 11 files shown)


config/initializers/devise.rb
```diff
@@ -216,14 +216,14 @@
   # :time  = Re-enables login after a certain amount of time (see :unlock_in below)
   # :both  = Enables both strategies
   # :none  = No unlock strategy. You should handle unlocking by yourself.
-  # config.unlock_strategy = :both
+  config.unlock_strategy = :both
```
citizen428, line 219: This together with `config.unlock_in = 1.hour` means the account will automatically unlock after 1 hour. We can change that if people feel like it, but since our main motivation for this was deterring brute-force attacks it seemed like an ok setting.

config/initializers/devise.rb
```diff
@@ -216,14 +216,14 @@
...
 
   # Number of authentication tries before locking an account if lock_strategy
   # is failed attempts.
-  # config.maximum_attempts = 20
+  config.maximum_attempts = 3
```
citizen428, line 223: 20 attempts seemed high. Let me know if you think 3 is too low.

spec/system/user_logs_in_with_password_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "rails_helper"
+
+RSpec.describe "Authenticating with a password" do
```
citizen428, line 3: I think we had no specs for password logins so far, so I added a whole bunch, not only the ones for account locking.

config/initializers/devise.rb
Ridhwana: Personal but not strong opinion: I know that it's common for banks to use 3 but I feel like we can be a bit more lenient and maybe use 5.

spec/system/user_logs_in_with_password_spec.rb
Ridhwana: Nice!🎉

spec/system/user_logs_in_with_password_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      submit_login_form(user.email, "wr0ng")
+      expect(page).to have_text("You have one more attempt before your account is locked.")
+    end
+
+    it "displays a message when the user got locked out" do
+      allow(User).to receive(:maximum_attempts).and_return(1)
+
+      submit_login_form(user.email, "wr0ng")
+      expect(page).to have_text("Your account is locked.")
```
Ridhwana, line 39: This is not necessarily related to your PR, but I was curious to know can we configure this to specify how long your account will be locked for. i.e. "You account is locked. Please try again in an hour". This will reduce the amount of support requests and panic we get from users.

spec/system/user_logs_in_with_password_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+  end
+
+  context "when logging in with the correct credentials" do
+    it "allows the user to sign in with the correct password" do
+      submit_login_form(user.email, password)
+      expect(page).to have_current_path("/dashboard?signin=true")
+    end
+  end
```
Ridhwana, line 48: Question: Is it possible and useful to test that the text "Your account is locked" is removed in an hour (once the account is unlocked) by doing some test time magic, or do you think that's not necessary to test because it's encompassed by devise?

rhymes:
Hi @justinlarson, can you please take a look at this? Thank you!

justinlarson:
Using the 1 hour lockout is ok if you allow valid users to unlock the account through email or logging in with github/twitter. If you don't allow valid users to unlock their account then an attacker _could_ essentially perform an account Denial of Service and lockout that user for at least that hour.

citizen428:
@justinlarson That's how `config.unlock_strategy = :both` for Devise works. The user will receive an email with an unlock token (similar to how password reset works in Devise), but the lock will also automatically expire.

config/initializers/devise.rb
citizen428: Sure, I had 5 at first, then lowered it for local testing, happy to increase it again. Would like to see what others think.

spec/system/user_logs_in_with_password_spec.rb
citizen428: It's the option `config.unlock_in` in `config/initializers/devise.rb`. Note that time-based unlocking is actually a convenience feature, normally the account would just be locked until you click the unlock link in the email.

spec/system/user_logs_in_with_password_spec.rb
citizen428: The text only appears in a flash message after the failed login attempt, not permanently on the page.

config/initializers/devise.rb
mstruve: I like 5, 3 seems too low to me as well.

citizen428:
@mstruve Sorry if this was unclear, but that's not how `:both` is supposed to work since `:lockable` has no knowledge about `omniauthable`. As per the initializer, the two supported ways for unlocking are:

1. Via clicking a link in the email
2. Time based

Good call on an adding this to `/internal`.

mstruve:
> Via clicking a link in the email

How do you get that link?

citizen428:
> How do you get that link?

Devise should email it when the account gets locked. I'll verify that and potentially add a spec for it.

> One more question, could a person get locked out if they didn't have a password set?

Technically I think all our accounts have passwords, as we set them to `Devise.friendly_token(20)` during social login. I would expect users who can't remember ever setting a password to use the forgot password functionality, but worst case, yes, they can get locked out.

Anyway, I had already set this PR back to draft and am working on some related things now, will re-request a review when I'm done.

citizen428:
Made the following changes now:

1. Increased max attempts to 5 before locking out the user.
2. Verified that an email gets sent and added a spec for it.
    Screen Shot 2020-07-29 at 10 32 17
3. Updated our `Authenticator` class so social logins now unlock locked accounts automatically. Tried locally and added a spec for it.
4. Show the lock status on the internal user show page.
    Screen_Shot_2020-07-29_at_10_33_14
5. Added a button for unlocking an account and added a spec for it.
    Screen Shot 2020-07-29 at 10 45 23
    Screen Shot 2020-07-29 at 10 45 31

I'll re-request reviews from everyone.

db/schema.rb
```diff
@@ -472,7 +472,6 @@
     t.datetime "updated_at", null: false
     t.bigint "user_id"
     t.datetime "verified_at"
-    t.index ["user_id"], name: "index_email_authorizations_on_user_id"
```
citizen428, line 475: This keeps popping up for me locally, does anyone know anything about it?

mstruve: I had this issue and updated my database manually via a console to get it to go away. I always figure it came from a branch I tested that never ended up in master

app/services/authentication/authenticator.rb
```diff
@@ -127,6 +127,7 @@ def default_user_fields
 
     def update_user(user)
       user.tap do |model|
+        user.unlock_access! if user.access_locked?
```
mstruve, line 130: 😍

db/schema.rb
citizen428: Yes, that fixed it. I just tried to avoid a similar situation to what we had a few months back were different PRs alternatively added and then dropped a specific index. 😆

citizen428:
Cleaned up the schema now.

PR #9537, opened 2020-07-27
Rubocop: enable and fix new Rails cops

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

https://github.com/forem/forem/pull/9526 introduced quite a few new `Rails/*` cops and best practices.
This PR fixes all violations.

I've also fixed a couple manually. Will leave comments in the code.

https://github.com/forem/forem/pull/9526

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -482,7 +482,6 @@ Performance/IoReadlines:
   Description: 'Use `IO.each_line` (`IO#each_line`) instead of `IO.readlines` (`IO#readlines`).'
   Reference: 'https://docs.gitlab.com/ee/development/performance.html#reading-from-files-and-other-data-sources'
   Enabled: true
-  VersionAdded: '1.7'
 
 Performance/OpenStruct:
   Description: 'Use `Struct` instead of `OpenStruct`.'
@@ -516,18 +515,85 @@ Performance/StringInclude:
 # Rails cops from rubocop-rails
 # https://github.com/rubocop-hq/rubocop-rails/blob/master/config/default.yml
 # TODO: [@thepracticaldev/oss] enable these Rails cops
+
+Rails/ActiveRecordCallbacksOrder:
+  Description: 'Order callback declarations in the order in which they will be executed.'
+  StyleGuide: 'https://rails.rubystyle.guide/#callbacks-order'
+  Enabled: true
+
+Rails/DefaultScope:
+  Description: 'Avoid use of `default_scope`.'
+  StyleGuide: 'https://rails.rubystyle.guide#avoid-default-scope'
+  Enabled: true
+
 Rails/FilePath:
   EnforcedStyle: slashes
+
+Rails/FindById:
+  Description: >-
+                Favor the use of `find` over `where.take!`, `find_by!`, and `find_by_id!` when you
+                need to retrieve a single record by primary key when you expect it to be found.
+  StyleGuide: 'https://rails.rubystyle.guide/#find'
+  Enabled: true
 
 Rails/HasManyOrHasOneDependent:
   Description: 'Define the dependent option to the has_many and has_one associations.'
   StyleGuide: 'https://rails.rubystyle.guide#has_many-has_one-dependent-option'
   Enabled: false
 
+Rails/Inquiry:
+  Description: "Prefer Ruby's comparison operators over Active Support's `Array#inquiry` and `String#inquiry`."
+  StyleGuide: 'https://rails.rubystyle.guide/#inquiry'
+  Enabled: true
+
+Rails/MailerName:
+  Description: 'Mailer should end with `Mailer` suffix.'
+  StyleGuide: 'https://rails.rubystyle.guide/#mailer-name'
+  Enabled: true
+
+Rails/MatchRoute:
+  Description: >-
+                  Don't use `match` to define any routes unless there is a need to map multiple request types
+                  among [:get, :post, :patch, :put, :delete] to a single action using the `:via` option.
+  StyleGuide: 'https://rails.rubystyle.guide/#no-match-routes'
+  Enabled: true
+
+Rails/NegateInclude:
+  Description: 'Prefer `collection.exclude?(obj)` over `!collection.include?(obj)`.'
+  Enabled: true
+
+Rails/Pluck:
+  Description: 'Prefer `pluck` over `map { ... }`.'
+  Enabled: true
+
+Rails/PluckId:
+  Description: 'Use `ids` instead of `pluck(:id)` or `pluck(primary_key)`.'
+  StyleGuide: 'https://rails.rubystyle.guide/#ids'
+  Enabled: false
+
+Rails/PluckInWhere:
+  Description: 'Use `select` instead of `pluck` in `where` query methods.'
+  Enabled: true
+
+Rails/RenderInline:
+  Description: 'Prefer using a template over inline rendering.'
+  StyleGuide: 'https://rails.rubystyle.guide/#inline-rendering'
+  Enabled: true
+
+Rails/RenderPlainText:
+  Description: 'Prefer `render plain:` over `render text:`.'
+  StyleGuide: 'https://rails.rubystyle.guide/#plain-text-rendering'
+  Enabled: true
+
 Rails/SaveBang:
   Description: 'Identifies possible cases where Active Record save! or related should be used.'
   StyleGuide: 'https://rails.rubystyle.guide#save-bang'
   Enabled: false
+
+Rails/ShortI18n:
+  Description: 'Use the short form of the I18n methods: `t` instead of `translate` and `l` instead of `localize`.'
+  StyleGuide: 'https://rails.rubystyle.guide/#short-i18n'
+  Enabled: true
 
 Rails/SkipsModelValidations:
   Description: >-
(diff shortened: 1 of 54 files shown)


app/controllers/articles_controller.rb
```diff
@@ -285,7 +285,7 @@ def handle_notifications(updated)
       Notification.remove_all_by_action_without_delay(notifiable_ids: @article.id, notifiable_type: "Article",
                                                       action: "Published")
       if @article.comments.exists?
-        Notification.remove_all(notifiable_ids: @article.comments.pluck(:id),
+        Notification.remove_all(notifiable_ids: @article.comments.ids,
```
rhymes, line 288: `pluck(:id)` is literally the same as `ids` which is a builtin method in Rails

https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-ids

app/controllers/social_previews_controller.rb
```diff
@@ -5,7 +5,7 @@ class SocialPreviewsController < ApplicationController
 
   def article
     @article = Article.find(params[:id])
-    @tag_badges = Badge.where(id: Tag.where(name: @article.decorate.cached_tag_list_array).pluck(:badge_id))
+    @tag_badges = Badge.where(id: Tag.where(name: @article.decorate.cached_tag_list_array).select(:badge_id))
```
rhymes, line 8: Rails is smart enough to build a SQL subquery when given an ActiveRecord relation in a `.where()`. Unfortunately `.pluck` is an eager method, which results in 2 queries (inner pluck, plus external where) instead of 1 SQL query.

Thus `.select()` is preferable here over `.pluck()` (and we also save some memory in case of huge arrays)

See https://github.com/rubocop-hq/rubocop-rails/issues/246

app/models/article.rb
```diff
@@ -72,13 +72,14 @@ class Article < ApplicationRecord
   validates :video_source_url, url: { allow_blank: true, schemes: ["https"] }
 
   before_validation :evaluate_markdown, :create_slug
-  before_create :create_password
+  before_save :update_cached_user
```
rhymes, line 75: The new `Rails/ActiveRecordCallbacksOrder` sorts callbacks in models following the order Rails will actually execute them, see https://github.com/rubocop-hq/rubocop-rails/pull/285

app/models/broadcast.rb
```diff
@@ -30,7 +30,7 @@ def single_active_announcement_broadcast
     first_broadcast = active_broadcasts.order(id: :asc).limit(1)
     return unless active &&
       type_of == "Announcement" &&
-      ![nil, id].include?(first_broadcast.pick(:id))
+      [nil, id].exclude?(first_broadcast.pick(:id))
```
rhymes, line 33: "not include" equals "exclude"

app/models/data_update_script.rb
```diff
@@ -3,15 +3,14 @@ class DataUpdateScript < ApplicationRecord
   NAMESPACE = "DataUpdateScripts".freeze
   STATUSES = { enqueued: 0, working: 1, succeeded: 2, failed: 3 }.freeze
 
-  default_scope { order(file_name: :asc) }
```
rhymes, line 6: `default_scope` is mostly never a good idea, we could have probably kept it in this case since we don't use a `WHERE` condition on it, but within the spirit of building best practices and discouraging things I went ahead and fixed this violation.

For more details on why `default_scope` in Rails is not a good idea see https://github.com/rubocop-hq/rubocop-rails/issues/76

app/models/sponsorship.rb
```diff
@@ -35,14 +35,14 @@ class Sponsorship < ApplicationRecord
 
   def validate_tag_uniqueness
     return unless self.class.where(sponsorable: sponsorable, level: :tag)
-      .where("expires_at > ? AND id != ?", Time.current, id.to_i).exists?
+      .exists?(["expires_at > ? AND id != ?", Time.current, id.to_i])
```
rhymes, line 38: `where(params).exists?` is the same as `.exists?(params)`

app/models/user.rb
```diff
@@ -278,7 +277,7 @@ def cached_followed_tag_names
         id: Follow.where(
           follower_id: id,
           followable_type: "ActsAsTaggableOn::Tag",
-        ).pluck(:followable_id),
+        ).select(:followable_id),
```
rhymes, line 280: same thing about `pluck` vs `select` in subqueries

spec/system/articles/user_visits_an_article_spec.rb
```diff
@@ -79,7 +79,7 @@
         visit articles.first.path
 
         elements = page.all(:xpath, articles_selector)
-        paths = elements.map { |e| e[:href] }
+        paths = elements.pluck(:href)
```
rhymes, line 82: Enumerable in Rails have the `pluck` method: https://api.rubyonrails.org/classes/Enumerable.html#method-i-pluck

.rubocop.yml
```diff
@@ -516,26 +515,97 @@ Performance/StringInclude:
 # Rails cops from rubocop-rails
 # https://github.com/rubocop-hq/rubocop-rails/blob/master/config/default.yml
 # TODO: [@thepracticaldev/oss] enable these Rails cops
+
+Rails/ActiveRecordCallbacksOrder:
+  Description: 'Order callback declarations in the order in which they will be executed.'
```
mstruve, line 520: This is the only one I am a little hesitant on. 

I think sometimes it can be helpful and more clear to group `save`, `create`, and `destroy` callbacks together rather then in the order they will be executed. For example, putting `before_destroy` up by `save` calls always throws me off bc each callback is being run under different conditions. Sometimes I think those conditions that cause a callback  to trigger can be more clear to group the callbacks by than the order they run since sometimes they dont run at all.

app/controllers/social_previews_controller.rb
fdoxyz: Nice optimization! 😃

app/services/mentions/create_all.rb
```diff
@@ -53,7 +53,7 @@ def authored_by?(user, notifiable)
 
     def delete_mentions_removed_from_notifiable_text(users)
       mentions = @notifiable.mentions.where.not(user_id: users).destroy_all
-      Notification.remove_all(notifiable_ids: mentions.pluck(:id), notifiable_type: "Mention") if mentions.present?
+      Notification.remove_all(notifiable_ids: mentions.pluck(:id), notifiable_type: "Mention") if mentions.present? # rubocop:disable Rails/PluckId
```
fdoxyz, line 56: Absolutely not a problem/blocker but to avoid the disable message in this particular case maybe `mentions.map(&:id)` could be used?

app/workers/field_tests/prune_old_experiments_worker.rb
```diff
@@ -6,7 +6,7 @@ class PruneOldExperimentsWorker
     def perform
       five_precent_membership_count = FieldTest::Membership.count / 20
       memberships = FieldTest::Membership.first(five_precent_membership_count)
-      FieldTest::Event.where(field_test_membership_id: memberships.pluck(:id)).delete_all
+      FieldTest::Event.where(field_test_membership_id: memberships.pluck(:id)).delete_all # rubocop:disable Rails/PluckInWhere
```
fdoxyz, line 9: Again not requesting the change but same suggestion as above, maybe `memberships.map(&:id)` can save us the rubocop disable annotation?

spec/requests/api/v0/followers_spec.rb
```diff
@@ -61,7 +61,7 @@
       it "order results for reverse following date" do
         follower2.follow(user)
 
-        follows = user.followings.order(id: :desc).last(2).pluck(:id)
+        follows = user.followings.order(id: :desc).last(2).pluck(:id) # rubocop:disable Rails/PluckId
```
fdoxyz, line 64: One more that **I think** could save us the annotation (`.map(&:id)`) 😄

fdoxyz:
On a separate note: Is it okay to have `vendor/cache/libv8-8.4.255.0-x86_64-darwin-19.gem` added/updated in the changes? Not sure if this was intentional

app/services/mentions/create_all.rb
rhymes: let me check if that works

rhymes:
@fdoxyz yeah, that's intentional. The first one of us who compiles `libv8` for their own system is going to get that file added to their PR. We have 2 versions currently: one for Linux, one for Darwin 18 (the one before Catalina I've already forgot the name of), and this is for Darwin 19 (Catalina)

app/services/mentions/create_all.rb
rhymes: it worked, thanks! 

see https://github.com/forem/forem/pull/9537/commits/f4b5adf18f49f60636b8446e5e2631e7f1cf2497 and https://github.com/forem/forem/pull/9537/commits/22ec65c5af8db8ab966b141631ba8df6b54fe395

.rubocop.yml
citizen428: Callbacks are a lot like default scopes, most of them should never be added. For the ones we do add, I prefer the full order, because there are other groups (`_validation`, `_initialize`, `_commit`) and even after all these years I'm still not entirely sure in which order they all run.

app/models/data_update_script.rb
```diff
@@ -3,15 +3,14 @@ class DataUpdateScript < ApplicationRecord
...
   enum status: STATUSES
   validates :file_name, uniqueness: true
 
   class << self
     def scripts_to_run
-      ids = load_script_ids
-      enqueued.where(id: ids)
+      insert_new_scripts
+
+      enqueued.where(id: ids).order(file_name: :asc)
```
citizen428, line 13: @rhymes The assignment to `ids` got removed but it's still used in this `where` clause. Maybe this?

```ruby
ids = insert_new_scripts
enqueued.where(id: ids).order(file_name: :asc)
```

PR #9538, opened 2020-07-27
Rubocop: address feedback of #9430 and use kwargs 

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR is to address most of the feedback left by @citizen428 in this PR review https://github.com/forem/forem/pull/9430#pullrequestreview-454005749 which got unaddressed inline unfortunately

https://github.com/forem/forem/pull/9430

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/app/services/authentication/paths.rb
+++ b/app/services/authentication/paths.rb
@@ -1,16 +1,15 @@
-# rubocop:disable Style/OptionHash
 module Authentication
   # These are meant to be called from the specific providers
   module Paths
     # Returns the authentication path for the given provider
-    def self.authentication_path(provider_name, params = {})
+    def self.authentication_path(provider_name, **kwargs)
       Rails.application.routes.url_helpers.public_send(
-        "user_#{provider_name}_omniauth_authorize_path", params
+        "user_#{provider_name}_omniauth_authorize_path", **kwargs
       )
     end
 
     # Returns the sign in URL for the given provider
-    def self.sign_in_path(provider_name, params = {})
+    def self.sign_in_path(provider_name, **kwargs)
       url_helpers = Rails.application.routes.url_helpers
 
       callback_url_helper = "user_#{provider_name}_omniauth_callback_path"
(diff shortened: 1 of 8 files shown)


rhymes:
@mstruve I thought about the name but in the spirit of being "Rubyesque" I realized it's pretty standard Ruby: https://github.com/search?q=%22kwargs%22+extension%3Arb&type=Code&l=&l= (coming from Python, keywords arguments are name `kwargs` there too)

I also ended up on https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/ which talks about `**kwargs` and how the behavior changes in Ruby 3.

Let's remember that Ruby 3 recommends the following:

> However, this style is not recommended in new code, unless you are often passing a Hash as a positional argument, and are also using keyword arguments. Otherwise, use double splat:

```ruby
def foo(**kwargs)
  kwargs
end
foo(k: 1) #=> {:k=>1}
```

(we totally silenced keyword argument violations in a previous PR so that's why we aren't bothered by the `params = {}` style)

That said, doesn't mean we can't make up our own rubocop rule to rename `kwargs` to something else, I fear that without an automatic tool we'll end up having 3 different names instead of 1 so I'd rather keep the standard one for now

mstruve:
All for the automatic checks, just wondering if spelling it out might be more approachable.

app/services/authentication/paths.rb
```diff
@@ -1,25 +1,23 @@
-# rubocop:disable Style/OptionHash
 module Authentication
   # These are meant to be called from the specific providers
   module Paths
     # Returns the authentication path for the given provider
-    def self.authentication_path(provider_name, params = {})
+    def self.authentication_path(provider_name, **kwargs)
       Rails.application.routes.url_helpers.public_send(
-        "user_#{provider_name}_omniauth_authorize_path", params
+        "user_#{provider_name}_omniauth_authorize_path", **kwargs
```
qojrag-muCroj-4demru, line 7: @mstruve

rhymes:
@mstruve I understand your point of having a more verbose "keyword_args" varable but I also think we should keep our code as standard as possible if we can. It's also used in very few places...

PR #9539, opened 2020-07-27
Removes intrusive mobile video autoplay & adds ahoy for visibility

fdoxyz:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Autoplay can be a little intrusive on mobile because it does not let the user read the article associated, which can even be a Series like [BaseCS](https://dev.to/devteam/basecs-intro-to-binary-2j09).

So instead of having the video take over we let the user choose to play. For visibility also added an Ahoy event

Using iOS app in production or Android beta release we can see the videos are played automatically. The web video should not be affected at all.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

play pause

### DIFF
--- a/app/assets/javascripts/initializers/initializeVideoPlayback.js
+++ b/app/assets/javascripts/initializers/initializeVideoPlayback.js
@@ -13,10 +13,12 @@
 /* eslint no-param-reassign: 0 */
 /* eslint no-useless-escape: 0 */
 /* global jwplayer */
+/* global ahoy */
 
 function initializeVideoPlayback() {
   var nativeBridgeMessage;
   var currentTime = '0';
+  var deviceType = 'web';
 
   function getById(name) {
     return document.getElementById(name);
@@ -116,6 +118,9 @@ function initializeVideoPlayback() {
       url: metadata.video_source_url,
       seconds: currentTime,
     });
+
+    var properties = { article: metadata.id, deviceType: deviceType };
+    ahoy.track('Article video play', properties);
   }
 
   function handleVideoMessages(mutation) {
@@ -145,6 +150,7 @@ function initializeVideoPlayback() {
     var metadata = videoMetadata(videoSource);
 
     if (isNativeIOS()) {
+      deviceType = 'iOS';
       nativeBridgeMessage = function (message) {
         try {
           window.webkit.messageHandlers.video.postMessage(message);
@@ -153,6 +159,7 @@ function initializeVideoPlayback() {
         }
       };
     } else if (isNativeAndroid()) {
+      deviceType = 'Android';
       nativeBridgeMessage = function (message) {
         try {
           AndroidBridge.videoMessage(JSON.stringify(message));
@@ -170,7 +177,7 @@ function initializeVideoPlayback() {
     playerElement.addEventListener('click', requestFocus);
 
     playerElement.classList.add('native');
-    getById('pause-butt').classList.add('active');
+    getById('play-butt').classList.add('active');
 
     var mutationObserver = new MutationObserver(function (mutations) {
       mutations.forEach(function (mutation) {
@@ -180,11 +187,6 @@ function initializeVideoPlayback() {
     mutationObserver.observe(videoSource, { attributes: true });
 
     currentTime = `${seconds}`;
-    nativeBridgeMessage({
-      action: 'play',
-      url: metadata.video_source_url,
-      seconds: currentTime,
-    });
   }
 
   // If an video player element is found initialize it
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/initializers/initializeVideoPlayback.js
```diff
@@ -116,6 +118,9 @@ function initializeVideoPlayback() {
       url: metadata.video_source_url,
       seconds: currentTime,
     });
+
+    var properties = { article: metadata.id, deviceType: deviceType };
+    ahoy.track('Article video play', properties);
```
vaidehijoshi, line 123: This one makes more sense to me since it looks like it should only fire when [the player element is clicked](https://github.com/forem/forem/blob/2b219d420a97c55220c25ec64e1d7bccc03cf11a/app/assets/javascripts/initializers/initializeVideoPlayback.js#L170)! 👍 

My only suggestion is that we change the event name to match what we're doing elsewhere in the codebase with ahoy, for example [here](https://github.com/forem/forem/blob/8425c847c3003caa2516dec36a62b1d06206aaf9/app/views/notifications/index.html.erb#L38). So, the equivalent version would be something like:
```suggestion
    ahoy.track('Clicked Video Player', properties);
```

app/assets/javascripts/initializers/initializeVideoPlayback.js
```diff
@@ -163,14 +170,16 @@ function initializeVideoPlayback() {
     } else {
       // jwplayer is initialized and no further interaction is needed
       initWebPlayer(seconds, metadata);
+      var properties = { article: metadata.id, deviceType: deviceType };
+      ahoy.track('Article video play', properties);
```
vaidehijoshi, line 174: Hm, it's a little weird to me that we're calling `ahoy.track` within the `initializePlayer`. That means that we'd be creating a new `Ahoy::Event` every single time the player was initialized, which I think isn't quite what we want to do when we track events? I think it makes sense to just remove this given where it is being called, but maybe I am missing some context here, so please feel free to fill me in if I'm missing something obvious :)

fdoxyz: That's a great question as it's not obvious from the `init/initialized` naming used. The idea behind the event was to know when users "watch a video" for all platforms (web/Android/iOS). 

For both mobile platforms the event is triggered on a click/tap, so the new name for the event suggested there (`Clicked Video Player`) makes a lot of sense. On the web browser the video player is initialized with autoplay, so there will be no click event to track when the video starts to play.

__Follow-up question:__ @vaidehijoshi would it be better to use an event name like `Video Player Started`/`Started Streaming Video`/etc? Or maybe just remove this event from the web entirely? I wouldn't want to get rid of the web tracking so we can compare across all platforms, but if it's not the intended use for Ahoy I can remove it.

Naming things is hard 😆

vaidehijoshi: @fdoxyz Agree re: naming things being hard 🙈 

This additional context is super helpful, thanks for providing it! 😃  We could definitely name the event differently (I like `Video Player Streaming Started` or something like you suggested!), if that is something we still want to track on the web side and if we can't actually "track" a click in that case (it sounds like we can't since the web video player plays automatically).

This makes me wonder if we'd ever want to know when someone on the web platform pauses/stops a video? I'm thinking about this because, if web videos are just automatically streamed without the user triggering a click event to start them, do we want to know how many of those "started" videos get stopped or paused immediately?

fdoxyz: This is great feedback thank you @vaidehijoshi! 

Looking back at the Podcast Player [here](https://github.com/forem/forem/blob/master/app/assets/javascripts/initializers/initializePodcastPlayback.js#L371) __we do track play/pause events__ (next in line for a refactor to remove GA in favor of Ahoy). I took another deep look at `jwplayer` and found an event that we can hook onto in order to get these updates (with a small workaround).

Just sent in a new commit with the event name `Video Player Streaming` and the properties payload now look like this:

```
{"article":29,"deviceType":"Android","action":"pause"}
```

I was also considering `Video Player Interaction` for the name, but me know what you think 🙂

app/assets/javascripts/initializers/initializeVideoPlayback.js
```diff
@@ -64,6 +67,24 @@ function initializeVideoPlayback() {
     }
   }
 
+  function videoPlayerEvent(isPlaying) {
+    // jwtplayer tends to send multiple 'play' actions. This check makes sure
+    // we're not tracking repeated 'play' events for a single interaction.
+    var eventName = isPlaying ? 'play' : 'pause';
+    if (lastEvent == eventName) {
```
vaidehijoshi, line 74: I actually think it probably works fine as is, but just curious if there's a reason to _not_ use strict equality comparison here (`===` rather than `==`?)

fdoxyz: No reason to use it but also no reason not to use it! I included it for the sake of sticking with the better standard

PR #9542, opened 2020-07-27
Update instantClick's script clone logic

maestromac:
- [x] Bug Fix

## Description
This PR updates a portion of InstantClick's `instantanize()` method to ignore any script on the page that has gone missing. This error is responsible for a lot of JS exception we're getting on honeybadger. The chunk of the code is copied directly from InstantClick's repository with very little change.

- Current latest copy chunk of code is from here https://github.com/dieulot/instantclick/blob/master/src/instantclick.js#L296
- The initial commit for the change https://github.com/dieulot/instantclick/commit/9fb9634c1d400e8b7ce7209c14f870cc0b8ee131
- [Basically 90% of all](https://app.honeybadger.io/projects/67192/faults?sort=last_seen_desc&q=-is%3Aresolved%E2%80%83-is%3Aignored%E2%80%83Unexpected+token) the `Unexpected token ':'` we've been getting on HoneyBadger.

Previous attempt: https://github.com/forem/forem/pull/9245

1. Browse around locally with browser console open.
2. You should not see any exceptions.

## Added tests?

## Added to documentation?
- [x] no documentation needed

I'll closely monitor production

### DIFF
--- a/app/assets/javascripts/base.js.erb
+++ b/app/assets/javascripts/base.js.erb
@@ -394,7 +394,6 @@ var instantClick
       for (i = 0; i < scriptElementsToCopy.length; i++) {
         originalElement = scriptElementsToCopy[i]
         if (!originalElement) { // Might have disappeared, see previous comment
-          console.log("gone")
           continue
         }
         if (originalElement.hasAttribute('data-no-instant')) {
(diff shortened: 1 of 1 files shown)


maestromac:
@nickytonline our "copy" is many commits behind and we also made some other modification. It certainly would be nice but I don't want to possibly introduce other issues because we don't test to rely on. I want to only resolve this issue that has been plaguing us for now.

PR #9543, opened 2020-07-27
Optimization:Change Comment, Ad Events, and Github ID Code Ints to Bigints

mstruve:
- [x] Optimization

## Description
Comments and Display Ad Events are medium size tables that need to have integers updated to bigints. Each table takes between 5-10 seconds to update. I have increased the STATEMENT_TIMEOUT on production to 15 seconds to help deal with these minor delays.
[Here is the timing data against our production fork](https://gist.github.com/mstruve/1600b29a30b39b9f4edc0c8d79d9ba91)
```
# 7 seconds
"comments"
# -- change_column(:comments, :id, :bigint)
#    -> 7.1299s
# => []
# irb(main):217:1* Benchmark.realtime {
# irb(main):218:2*   ActiveRecord::Base.connection.execute(
# irb(main):219:0"     <<-SQL
# irb(main):220:0"     ALTER TABLE comments
# irb(main):221:0"     ALTER COLUMN commentable_id TYPE bigint,
# irb(main):222:0"     ALTER COLUMN user_id TYPE bigint
# irb(main):223:2*     SQL
# irb(main):224:1*     )
# irb(main):225:0> }
# => 7.011359310010448

# 6 seconds
"display_ad_events"
# irb(main):235:1* ActiveRecord::Migration[6.0].safety_assured {
# irb(main):236:1*   ActiveRecord::Migration[6.0].change_column :display_ad_events, :display_ad_id, :bigint
# irb(main):237:0> }
# -- change_column(:display_ad_events, :display_ad_id, :bigint)
#    -> 6.0153s
# => []
# irb(main):238:1* ActiveRecord::Migration[6.0].safety_assured {
# irb(main):238:1* ActiveRecord::Migration[6.0].safety_assured {
# irb(main):239:1*   ActiveRecord::Migration[6.0].change_column :display_ad_events, :user_id, :bigint
# irb(main):240:0> }
# -- change_column(:display_ad_events, :user_id, :bigint)
#    -> 5.4779s
```

`github_id_code` for GithubRepos should be a non-issue, that table is small and updates in under a second.

alt_text

### DIFF
--- a/db/migrate/20200727163200_change_comment_ad_events_github_code_ints_to_bigints.rb
+++ b/db/migrate/20200727163200_change_comment_ad_events_github_code_ints_to_bigints.rb
@@ -1,0 +1,49 @@
+class ChangeCommentAdEventsGithubCodeIntsToBigints < ActiveRecord::Migration[6.0]
+  def up
+    ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.display_ad_events")
+    ActiveRecord::Base.connection.execute(
+      <<-SQL
+        ALTER TABLE display_ad_events
+          ALTER COLUMN id TYPE bigint,
+          ALTER COLUMN display_ad_id TYPE bigint,
+          ALTER COLUMN user_id TYPE bigint
+      SQL
+    )
+
+    ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.comments")
+    ActiveRecord::Base.connection.execute(
+      <<-SQL
+        ALTER TABLE comments
+          ALTER COLUMN id TYPE bigint,
+          ALTER COLUMN user_id TYPE bigint,
+          ALTER COLUMN commentable_id TYPE bigint
+      SQL
+    )
+
+    safety_assured { change_column :github_repos, :github_id_code, :bigint }
+  end
+
+  def down
+    ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.display_ad_events")
+    ActiveRecord::Base.connection.execute(
+      <<-SQL
+        ALTER TABLE display_ad_events
+          ALTER COLUMN id TYPE int,
+          ALTER COLUMN display_ad_id TYPE int,
+          ALTER COLUMN user_id TYPE int
+      SQL
+    )
+
+    ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.comments")
+    ActiveRecord::Base.connection.execute(
+      <<-SQL
+        ALTER TABLE comments
+          ALTER COLUMN id TYPE int,
+          ALTER COLUMN user_id TYPE int,
+          ALTER COLUMN commentable_id TYPE int
+      SQL
+    )
+
+    safety_assured { change_column :github_repos, :github_id_code, :int }
+  end
+end
--- a/db/migrate/20200727163200_change_comment_ad_events_github_code_ints_to_bigints.rb
+++ b/db/migrate/20200727163200_change_comment_ad_events_github_code_ints_to_bigints.rb
@@ -20,6 +20,7 @@ class ChangeCommentAdEventsGithubCodeIntsToBigints < ActiveRecord::Migration[6.0
       SQL
     )
 
+    ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.github_repos")
     safety_assured { change_column :github_repos, :github_id_code, :bigint }
   end
 
@@ -44,6 +45,7 @@ class ChangeCommentAdEventsGithubCodeIntsToBigints < ActiveRecord::Migration[6.0
       SQL
     )
 
+    ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.github_repos")
     safety_assured { change_column :github_repos, :github_id_code, :int }
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_07_26_215928) do
+ActiveRecord::Schema.define(version: 2020_07_27_163200) do
 
   # These are extensions that must be enabled in order to support this database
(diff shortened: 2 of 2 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_07_26_215928) do
+ActiveRecord::Schema.define(version: 2020_07_27_163200) do
```
rhymes, line 13: It looks that the schema is changing much more than comments, ad events and github repos though

mstruve: Shoot! good catch, too many migration branches in the air.

db/migrate/20200727163200_change_comment_ad_events_github_code_ints_to_bigints.rb
```diff
@@ -0,0 +1,51 @@
+class ChangeCommentAdEventsGithubCodeIntsToBigints < ActiveRecord::Migration[6.0]
+  def up
+    ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.display_ad_events")
```
vaidehijoshi, line 3: I'm curious: why do we need to explicitly `DROP VIEW` for the hypershield equivalent of the table? Does hypershield look to our DB and if we modify a column type, it blows up?

Which leads me to my second question: do we need to "recreate" the hypershield view for this table once we alter the column type?

These questions are really just me trying to understand how hypershield (blazer?) interacts with our DB and the order of operations here 😃

mstruve: 2 GREAT questions

> I'm curious: why do we need to explicitly DROP VIEW for the hypershield equivalent of the table? 

I am not totally sure WHY but when I started doing these migrations I found that even tiny migrations would timeout with this error and if you scroll to the bottom of the backtrace you can see if goes back to hypershield removing the view. I also found when testing migrations on a fork of production that sometimes columns could not be updated bc they had hypershield views that relied on them. With these two things in mind I now drop the views before running the updates.
```
Caused by:
PG::QueryCanceled: ERROR:  canceling statement due to statement timeout
/app/vendor/bundle/ruby/2.7.0/gems/activerecord-6.0.3.2/lib/active_record/connection_adapters/postgresql/database_statements.rb:92:in `exec'
/app/vendor/bundle/ruby/2.7.0/gems/activerecord-6.0.3.2/lib/active_record/connection_adapters/postgresql/database_statements.rb:92:in `block (2 levels) in execute'
/app/vendor/bundle/ruby/2.7.0/gems/activesupport-6.0.3.2/lib/active_support/dependencies/interlock.rb:48:in `block in permit_concurrent_loads'
/app/vendor/bundle/ruby/2.7.0/gems/activesupport-6.0.3.2/lib/active_support/concurrency/share_lock.rb:187:in `yield_shares'
/app/vendor/bundle/ruby/2.7.0/gems/activesupport-6.0.3.2/lib/active_support/dependencies/interlock.rb:47:in `permit_concurrent_loads'
/app/vendor/bundle/ruby/2.7.0/gems/activerecord-6.0.3.2/lib/active_record/connection_adapters/postgresql/database_statements.rb:91:in `block in execute'
/app/vendor/bundle/ruby/2.7.0/gems/activerecord-6.0.3.2/lib/active_record/connection_adapters/abstract_adapter.rb:722:in `block (2 levels) in log'
/app/vendor/bundle/ruby/2.7.0/gems/activesupport-6.0.3.2/lib/active_support/concurrency/load_interlock_aware_monitor.rb:26:in `block (2 levels) in synchronize'
/app/vendor/bundle/ruby/2.7.0/gems/activesupport-6.0.3.2/lib/active_support/concurrency/load_interlock_aware_monitor.rb:25:in `handle_interrupt'
/app/vendor/bundle/ruby/2.7.0/gems/activesupport-6.0.3.2/lib/active_support/concurrency/load_interlock_aware_monitor.rb:25:in `block in synchronize'
/app/vendor/bundle/ruby/2.7.0/gems/activesupport-6.0.3.2/lib/active_support/concurrency/load_interlock_aware_monitor.rb:21:in `handle_interrupt'
/app/vendor/bundle/ruby/2.7.0/gems/activesupport-6.0.3.2/lib/active_support/concurrency/load_interlock_aware_monitor.rb:21:in `synchronize'
/app/vendor/bundle/ruby/2.7.0/gems/activerecord-6.0.3.2/lib/active_record/connection_adapters/abstract_adapter.rb:721:in `block in log'
/app/vendor/bundle/ruby/2.7.0/gems/activesupport-6.0.3.2/lib/active_support/notifications/instrumenter.rb:24:in `instrument'
/app/vendor/bundle/ruby/2.7.0/gems/activerecord-6.0.3.2/lib/active_record/connection_adapters/abstract_adapter.rb:712:in `log'
/app/vendor/bundle/ruby/2.7.0/gems/activerecord-6.0.3.2/lib/active_record/connection_adapters/postgresql/database_statements.rb:90:in `execute'
/app/vendor/bundle/ruby/2.7.0/gems/hypershield-0.2.0/lib/hypershield.rb:146:in `execute'
/app/vendor/bundle/ruby/2.7.0/gems/hypershield-0.2.0/lib/hypershield.rb:27:in `block in drop_view'
/app/vendor/bundle/ruby/2.7.0/gems/hypershield-0.2.0/lib/hypershield.rb:26:in `each'
```

> Which leads me to my second question: do we need to "recreate" the hypershield view for this table once we alter the column type?

This is done automatically thanks to #9546. After we run the migrations which are kicked off by `db:prepare` then `hypershield:refresh` is run which will repopulate the views.

vaidehijoshi: This is super informative, thanks for the explanation, @mstruve ! 🙌

rhymes: I'm not 100% as to why it locks, but I believe it's because PostgreSQL needs to have the same exact data types as the underlying table, here we do change it from `int` to `bigint` so as the views are readonly they need to be recreated from scratch

PR #9544, opened 2020-07-27
Get carrierwave+aws set up properly for container deploy

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently carrierwave and fog is set up primarily for our Heroku approach with explicit declaring of keys in the app instead of the `use_iam_profile` with private fog and reverse proxy.

I added `SYSTEMSNOTE` for anything that might be more WIP as we push out to the edge in our systems set up.

I decided to make a new PR to supplant #9371 in order to make it clear that this has not yet been approved, but may now be ready to go.

This PR now includes a monkeypatch of the carrierwave url helpers to always include the "public" version for display purposes.

For the current Heroku story, we want that because it's the default anyway, and for the version where we have `fog_public` we still want the generic "public" path to pair with our new asset host. I believe this is logically sound.


config/initializers/carrierwave.rb
```diff
@@ -14,14 +15,24 @@
...
+        aws_access_key_id: ApplicationConfig["AWS_ID"],
+        aws_secret_access_key: ApplicationConfig["AWS_SECRET"],
+        region: region
+      }
+      config.fog_directory = ApplicationConfig["AWS_BUCKET_NAME"]
+    else #SYSTEMSNOTE Non-Heroku for our WIP container
+      config.fog_credentials = {
+        provider: "AWS",
+        use_iam_profile: true,
+        region: "us-east-2"
```
mstruve, line 32: Any reason not to use `region` here as we do above since they are the same?

config/initializers/s3_direct_upload.rb
```diff
@@ -1,7 +1,12 @@
-S3DirectUpload.config do |c|
-  c.access_key_id = ApplicationConfig["AWS_S3_VIDEO_ID"] # your access key id
-  c.secret_access_key = ApplicationConfig["AWS_S3_VIDEO_KEY"] # your secret access key
-  c.bucket = ApplicationConfig["AWS_S3_INPUT_BUCKET"] # your bucket name
-  c.region = nil # region prefix. _Required_ for non-default AWS region, eg. "eu-west-1"
-  c.url = nil # S3 API endpoint (optional), eg. "https://#{c.bucket}.s3.amazonaws.com"
+# SYSTEMSNOTE: We only can load this if we everything working with AWS.
```
mstruve, line 1: Not sure what you meant to say here but the comment is a little jumbled.

config/initializers/carrierwave.rb
```diff
@@ -14,14 +15,24 @@
...
+    if ENV["HEROKU_APP_ID"].present?
+      region = ApplicationConfig["AWS_UPLOAD_REGION"].presence || ApplicationConfig["AWS_DEFAULT_REGION"]
+      config.fog_credentials = {
+        provider: "AWS",
+        aws_access_key_id: ApplicationConfig["AWS_ID"],
+        aws_secret_access_key: ApplicationConfig["AWS_SECRET"],
+        region: region
+      }
+      config.fog_directory = ApplicationConfig["AWS_BUCKET_NAME"]
+    else #SYSTEMSNOTE Non-Heroku for our WIP container
```
mstruve, line 28: Should we go with the convention of using the Github team name for all of these comments? 
```ruby
# @forem/systems: ....
```

benhalpern: Cool that makes sense

config/initializers/carrierwave.rb
benhalpern: I was a little unclear about what we had available via ENV vars and what we didn't, so I just left this hardcoded. I was going to follow up with this to get more clarity, but I changed it now with this comment being clarity.

mstruve:
Looks like that failure might be legit
``` 
1) Videos GET /videos/new when authorized allows authorized users
     Failure/Error: <%= s3_uploader_form callback_url: "/videos",
     ActionView::Template::Error:
       no implicit conversion of nil into String
     # ./app/views/videos/new.html.erb:15:in `_app_views_videos_new_html_erb___3395816909514778428_6252720'
     # ./spec/requests/videos_spec.rb:38:in `block (4 levels) in <top (required)>'
     # ./spec/rails_helper.rb:134:in `block (3 levels) in <top (required)>'
     # ./spec/rails_helper.rb:134:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # TypeError:
     #   no implicit conversion of nil into String
     #   ./app/views/videos/new.html.erb:15:in `_app_views_videos_new_html_erb___3395816909514778428_6252720'
```

mstruve:
Might want to put this in Draft mode as we figure it out to prevent any accidental merging.

PR #9546, opened 2020-07-27
Bug Fix:Run Hypershield Refresh AFTER db:prepare

mstruve:
- [x] Bug Fix

## Description
Recently I noticed that our hypershield refresh task was running before our migrations which is a problem bc then our views are not getting updated with new migrations.
```
[hypershield] Refreshing schemas
[hypershield] Success!
== 20200726215928 ChangeTagIdsToBigints: migrating ============================
== 20200726215928 ChangeTagIdsToBigints: migrated (4.9509s) ===================
```
Now, who is ready to get their mind blown? The reason is bc of the way we were invoking/using `enhance`. 
Turns out when you use [`enhance`](https://ruby-doc.org/stdlib-2.0.0/libdoc/rake/rdoc/Rake/Task.html#method-i-enhance) like this and pass it an argument
```ruby
Rake::Task["db:prepare"].enhance(["hypershield:refresh"])
```
The task that is your argument becomes a **prerequisite** and is run **BEFORE** the rake task you are enhancing. This is a problem for us bc we want our hypershield views updated AFTER we run migrations.

If you pass a block to `enhance` like this
```ruby
Rake::Task["db:prepare"].enhance do
  Rake::Task["hypershield:refresh"]
end
```
It will run the rake task in the block **AFTER** the Rake task that is being enhanced. 🤯 

https://github.com/forem/forem/projects/9#card-42626051

alt_text

### DIFF
--- a/lib/tasks/app_initializer.rake
+++ b/lib/tasks/app_initializer.rake
@@ -15,5 +15,7 @@ namespace :app_initializer do
 end
 
 if ENV["ENABLE_HYPERSHIELD"].present?
-  Rake::Task["db:prepare"].enhance(["hypershield:refresh"])
+  Rake::Task["db:prepare"].enhance do
+    Rake::Task["hypershield:refresh"]
+  end
 end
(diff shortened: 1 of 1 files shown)


mstruve:
@vaidehijoshi GREAT call on a comment! I actually had a block, saw argument syntax and figured that was more refactored. Totally easy for someone to make that mistake again.

PR #9547, opened 2020-07-27
Show article year if article published last year

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently we're not showing the date year when articles were published last year, when we should be.

Screen Shot 2020-07-27 at 6 32 43 PM

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -317,7 +317,7 @@ class Article < ApplicationRecord
   end
 
   def readable_publish_date
-    relevant_date = crossposted_at.presence || published_at
+    relevant_date = displayable_published_at
     if relevant_date && relevant_date.year == Time.current.year
       relevant_date&.strftime("%b %e")
     else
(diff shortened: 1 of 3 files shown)


app/models/article.rb
```diff
@@ -317,7 +317,7 @@ def readable_edit_date
   end
 
   def readable_publish_date
```
citizen428, line 319: **thought (non-blocking):** Looking at this it seems to be a display concern, so in the spirit of cleaning up our models a bit, could it make sense to move this (and the new `displayable_published_at` method) into a decorator?

app/views/articles/show.html.erb
```diff
@@ -150,7 +150,7 @@
 
               <span class="fs-s mb-4 s:mb-0">
                 <% if @article.published_timestamp.present? %>
-                  <%= local_date(@article.published_timestamp, show_year: false) %>
+                  <%= local_date(@article.published_timestamp, show_year: @article.displayable_published_at.year != Time.current.year) %>
```
citizen428, line 153: **thought (non-blocking):** If you go with the decorator suggestion above, we could also add a `show_year?` method to that:

```ruby
local_date(@article.published_timestamp, show_year: @article.show_year?)
```

Or, potentially even better, move all of this into the decorator and this line becomes

```ruby
<%= @article.localized_timestamp %>
```

(or whatever other method name makes sense).

PR #9556, opened 2020-07-28
Flaky Spec Fix:Remove let_it_be Test Prof Helper

mstruve:
- [x] Spec Fix

## Description
These `let_it_be` statements have been wreaking havoc on our builds lately causing a lot of spec failures when the data that these statements are meant to persist gets deleted. We are not sure exactly why the data gets removed but after a few hours of debugging I figured it would probably be easier to just remove the complexity of it altogether. Not persisting data in between specs will also make our suite more resilient and flexible bc specs can literally be run in ANY order even within a file.

Does this increase our spec run time? Yes, a little bit. I ran this branch 4 times and the average build times came out to around 10-11 min where before I think we sat closer to between 9-10 min. I think there is plenty of other ways we can shave that extra min off without needing to persist data between specs. I also noticed the containers not running as close in time as normal, I think that is bc spec timing is now slightly different so Knapsack needs time to process that new timing data. In addition, bc these are causing so many issues I think overall we are winning in time savings bc people wont need to be rerunning their flaky specs over and over.

@snackattas  

Let me know what you all think!   

alt_text


rhymes:
@rhymes I would keep the dependency in as it helps in case one needs to debug locally, things like `TEST_RUBY_PROF`, `TEST_STACK_PROF` and `EVENT_PROF`

PR #9557, opened 2020-07-28
Search: replace replaceState with pushState to fix the back button

### LINKED ISSUE #9433: Back button in the search results page is broken when filters are used
**Describe the bug**

After searching for a results, and filtering the results, the back button is broken

**To Reproduce**

1. Go to https://dev.to/search?q=python
1. Choose one of the filters
1. Hit the browser's back button
1. The browser will land back on the homepage, not on the non filtered results page

**Expected behavior**

I expect the back button not to be broken and going back resulting in going in each of the filtered results page

**Screenshots**

back_button

**Additional context**

I think the issue lies in how we build filters on click handlers and the usage of `window.replaceState` instead of `window.pushState`. See 

https://github.com/forem/forem/blob/b84f04baa5a061790b5a48c48554ceca6392cb0b/app/views/articles/_search.html.erb#L85-L98

PrimeNoodles:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Made changes such that user is now able to click through filters in the search bar while clicking the back button does not immediately take them to the home page, but to the previous search filter. Fixes #9433.

FilterBackButton

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/views/articles/_search.html.erb
+++ b/app/views/articles/_search.html.erb
@@ -48,10 +48,10 @@
         var sortString = buildSortString(sortBy, sortDirection);
 
         if (filters) {
-          window.history.replaceState(null, null, "/search?q=" + query + "&filters=" + filters + sortString);
+          window.history.requestState(null, null, "/search?q=" + query + "&filters=" + filters + sortString);
           search(query, filters, sortBy, sortDirection);
         } else {
-          window.history.replaceState(null, null, "/search?q=" + query + sortString);
+          window.history.requestState(null, null, "/search?q=" + query + sortString);
           search(query, "", sortBy, sortDirection);
         }
 
@@ -83,7 +83,7 @@
           return;
         }
         var filters = e.target.dataset.filter;
-        window.history.replaceState(null, null, "/search?q=" + query + "&filters=" + filters + sortString);
+        window.history.pushState(null, null, "/search?q=" + query + "&filters=" + filters + sortString);
         var className = e.target.className;
         for (var i = 0; i < filterButts.length; i++) {
           filterButts[i].classList.remove("selected");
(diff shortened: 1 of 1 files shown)


app/views/articles/_search.html.erb
```diff
@@ -48,10 +48,10 @@
         var sortString = buildSortString(sortBy, sortDirection);
 
         if (filters) {
-          window.history.replaceState(null, null, "/search?q=" + query + "&filters=" + filters + sortString);
+          window.history.requestState(null, null, "/search?q=" + query + "&filters=" + filters + sortString);
```
rhymes, line 51: what's `requestState` ?

PrimeNoodles: @rhymes It should be requestState, my bad, apparently the commit didn't use my current save file. I'm committing the more updated file. As for using replaceState on line 51 and 54 didn't seem to make a difference between that and pushState. I left it as it is just in case it may create more unseen bugs, but for consistency I'll change it to pushState. As for the replaceState on lines 93 and 96 it seems to cause a weird bug where the user is required to click back twice for every single switch on the filters. Leaving it as replaceState doesn't produce this bug.

nickytonline:
I had tried this with `replacesState` the other day and it is the way to go, but see my comment in the associated issue @rhymes and @PrimeNoodles, https://github.com/forem/forem/issues/9433#issuecomment-664669980.

rhymes:
@nickytonline do you think it's a valid interim fix? I wonder if double back is better or worse than no back button

nickytonline:
@rhymes, the double back is probably better for now and we can address the double back in another PR.

rhymes:
@nickytonline cool, let's merge this and we'll figure that out afterwards, thanks!

nickytonline:
Thanks so much for the PR @PrimeNoodles! 👏

PR #9558, opened 2020-07-28
Attempt to get carrierwave configured correctly for containers and tests passing

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This maintains the same functionality we have in the base path but gives us extra options for testing alternate deployments and image configs.

I tested that this branch seems to work correctly in https://community.benhalpern.com/ so I think this is good to go. I just want to get the `if/else` statements safely into master so that we can begin hacking on the `else` part.


config/initializers/carrierwave.rb
```diff
@@ -6,22 +6,40 @@
...
 
+# rubocop:disable Metrics/BlockLength
 CarrierWave.configure do |config|
   if Rails.env.test?
     config.storage = :file
     config.enable_processing = false
   elsif Rails.env.development?
+    # config.asset_host = "https://#{ApplicationConfig['APP_DOMAIN']}"
+    config.storage = :file
+  elsif ENV["FILE_STORAGE_LOCATION"] == "file" # @forem/systems production version of file store
```
mstruve, line 17: 🔥

Issue #7408, opened 2020-04-21
Word adjacency understanding in search.


benhalpern:
If I search `Can anyone explain how to use Ajax to connect to a C`

These are my results...

<img width="961" alt="Screen Shot 2020-04-20 at 8 38 03 PM" src="https://user-images.githubusercontent.com/3102842/79812415-d738e180-8346-11ea-8c2a-33485968f04c.png">

Adding quotations... `"Can anyone explain how to use Ajax to connect to a C"` gets me results relating to the post I knew existed...

<img width="986" alt="Screen Shot 2020-04-20 at 8 38 46 PM" src="https://user-images.githubusercontent.com/3102842/79812448-f9cafa80-8346-11ea-9568-92ebd91a9834.png">

Even though it's nice that we respect the quotations as a standard, we should do a better job at plucking out this kind of phrase in the results in the first place.

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

Rafi993:
Hello there, Is this issue up for grabs if so I would like to help.

rhymes:
Sure @Rafi993, go ahead and thanks! I'm going to formally assign you to the issue :) Let us know if you get stuck or need any help.

Rafi993:
Awesome. Thanks @rhymes

Rafi993:
I could not reproduce this in local. It does happen in [production](https://dev.to/search?q=Can%20anyone%20explain%20how%20to%20use%20Ajax%20to%20connect%20to%20a%20C) though I even copied those articles to my local still could not reproduce it.
![article0](https://user-images.githubusercontent.com/8092120/81155400-85c55f00-8fa2-11ea-8d7a-58f4a9f005a1.png)

these articles appear much lower when I scroll down the search result

![article1](https://user-images.githubusercontent.com/8092120/81155387-8231d800-8fa2-11ea-8548-25c1347ec309.png)

Any help on how should I go about re-producing it in local would be awesome.

Rafi993:
[...]

PR #9562, opened 2020-07-28
Feature:Boost Scores For Adjacent Words

mstruve:
- [x] Feature

## Description
I found a "workaround" for boosting the search scores for articles with words that are adjacent by using a `should` clause in our [`bool` query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html#query-dsl-bool-query). The way this works is I have added a [`match_phrase`](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query-phrase.html) query which focuses on matching phrases. How well you want the phrase matched depends on your `slop` value. A value of 0 means you want to match the phrase exactly. We are using a value of 0. A greater slop value means you are ok with some words in between the words of the phrase you searched. Adding this query ensures that any time we find the exact phrase we are looking for, the article with the phrase will get scored higher by Elasticsearch. 

However, we don't want to eliminate articles that have all the words even if they are not all next to each other. That is where the `should` part comes in. `should` in a bool clause means we want some combination of the queries in that should block to match. How many of them we want to match depends on our [`minimum_should_match`](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-minimum-should-match.html) parameter. We have set that to 0, which means, if a phrase matches, the article's score gets bumped. If a phrase does not match, it won't count against the article in any way. 

closes https://github.com/forem/forem/issues/7408

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/spec/services/search/feed_content_spec.rb
+++ b/spec/services/search/feed_content_spec.rb
@@ -88,6 +88,18 @@ RSpec.describe Search::FeedContent, type: :service do
         doc_ids = feed_docs.map { |t| t.dig("id") }
         expect(doc_ids).to include(article1.id, article2.id)
       end
+
+      it "bumps scores for words closer together" do
+        allow(article1).to receive(:body_text).and_return("Ruby I dont know maybe is cool")
+        allow(article2).to receive(:body_text).and_return("Ruby is cool")
+        index_documents([article1, article2])
+        query_params = { size: 5, search_fields: "ruby is cool" }
+
+        feed_docs = described_class.search_documents(params: query_params)
+        expect(feed_docs.count).to eq(2)
+        doc_ids = feed_docs.map { |t| t.dig("id") }
+        expect(doc_ids).to include(article2.id, article1.id)
+      end
     end
 
     context "with a filter term" do
--- a/spec/services/search/feed_content_spec.rb
+++ b/spec/services/search/feed_content_spec.rb
@@ -91,14 +91,14 @@ RSpec.describe Search::FeedContent, type: :service do
 
       it "bumps scores for words closer together" do
         allow(article1).to receive(:body_text).and_return("Ruby I dont know maybe is cool")
-        allow(article2).to receive(:body_text).and_return("Ruby is cool")
+        allow(article2).to receive(:body_text).and_return("Ruby is cool I dont know maybe")
         index_documents([article1, article2])
         query_params = { size: 5, search_fields: "ruby is cool" }
 
         feed_docs = described_class.search_documents(params: query_params)
         expect(feed_docs.count).to eq(2)
         doc_ids = feed_docs.map { |t| t.dig("id") }
-        expect(doc_ids).to include(article2.id, article1.id)
+        expect(doc_ids).to eq([article2.id, article1.id])
       end
     end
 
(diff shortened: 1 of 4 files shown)


spec/services/search/feed_content_spec.rb
```diff
@@ -88,6 +88,18 @@
         doc_ids = feed_docs.map { |t| t.dig("id") }
         expect(doc_ids).to include(article1.id, article2.id)
       end
+
+      it "bumps scores for words closer together" do
```
citizen428, line 92: @mstruve I checked out your branch and reverted `Search::QueryBuilders::FeedContent` back to its original form. However, this spec remains green, probably because the search phrase matches more of the overall document. It also stays green when I change the body text of the second article to "Ruby is cool I dont know maybe" though.

Here are my local changes:

```diff
diff --git a/app/services/search/query_builders/feed_content.rb b/app/services/search/query_builders/feed_content.rb
index 2d695800c..891619296 100644
--- a/app/services/search/query_builders/feed_content.rb
+++ b/app/services/search/query_builders/feed_content.rb
@@ -101,13 +101,14 @@ module Search
       def build_queries
         @body[:query] = { bool: {} }
         @body[:query][:bool][:filter] = filter_conditions if filter_keys_present?
-        return unless query_keys_present?
-
-        @body[:query][:bool][:must] = query_conditions
-        # Boost the score of queries that match these conditions but if they dont match any,
-        # minimum_should_match: 0, then that is OK
-        @body[:query][:bool][:should] = match_phrase_conditions
-        @body[:query][:bool][:minimum_should_match] = 0
+        @body[:query][:bool][:must] = query_conditions if query_keys_present?
+        # return unless query_keys_present?
+
+        # @body[:query][:bool][:must] = query_conditions
+        # # Boost the score of queries that match these conditions but if they dont match any,
+        # # minimum_should_match: 0, then that is OK
+        # @body[:query][:bool][:should] = match_phrase_conditions
+        # @body[:query][:bool][:minimum_should_match] = 0
       end

       def filter_conditions
diff --git a/spec/services/search/feed_content_spec.rb b/spec/services/search/feed_content_spec.rb
index a8d0e1ea3..192f4480e 100644
--- a/spec/services/search/feed_content_spec.rb
+++ b/spec/services/search/feed_content_spec.rb
@@ -91,7 +91,7 @@ RSpec.describe Search::FeedContent, type: :service do

       it "bumps scores for words closer together" do
         allow(article1).to receive(:body_text).and_return("Ruby I dont know maybe is cool")
-        allow(article2).to receive(:body_text).and_return("Ruby is cool")
+        allow(article2).to receive(:body_text).and_return("Ruby is cool I dont know maybe")
         index_documents([article1, article2])
         query_params = { size: 5, search_fields: "ruby is cool" }
```

benhalpern:
> How can this be best QAed? Thank you!

IMO tweaks to our search algorithm could be tested in prod, all else equal.

I also feel like we could use more extensive regression testing to ensure new tests get better and all the scenarios we hope will return one result over another remain solid when we add new things. If we throw enough different scenarios at our search and attempt to get it to give us "the best" results every time we'll be in good shape. In the meantime I think the current state is imperfect enough that we can bias towards pushing code and then eyeballing results.

spec/services/search/feed_content_spec.rb
mstruve: Good catch!

The problem was the assertion, we were looking that both articles were included rather than what their order was. I updated article 2 with the additional text and then set the array assertion to equal and now if you comment out my new code it fails.

citizen428: Ah, thanks @mstruve, that makes much more sense. 👍

PR #9349, opened 2020-07-16
Listings redesign

ludwiczakpawel:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR is very first step towards making Listings prettier. 

## QA Instructions, Screenshots, Recordings

image

image

## Added tests?

- [ ] yes
- [ ] no, because they aren't needed
- [x] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

PR #9563, opened 2020-07-29
Make <Modal /> component more accessible

nickytonline:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I came across this while looking at #9349. Fixed our `<Modal />` component so that it uses the `<dialog />` HTML element.

The modal component should render the same as before. You can compare the modal in this [PR's storybook](https://deploy-preview-9563--storybookdevto.netlify.app/?path=/story/3-components-modals--default) to the [main storybook](https://storybook.forem.com/?path=/story/3-components-modals--default).

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Snail without a shell riding a skateboard

### DIFF
--- a/app/javascript/crayons/Modal/Modal.jsx
+++ b/app/javascript/crayons/Modal/Modal.jsx
@@ -17,6 +17,21 @@ function getAdditionalClassNames({ size, className }) {
   return additionalClassNames;
 }
 
+const CloseIcon = () => (
+  <svg
+    width="24"
+    height="24"
+    viewBox="0 0 24 24"
+    className="crayons-icon"
+    xmlns="http://www.w3.org/2000/svg"
+    role="img"
+    aria-labelledby="714d29e78a3867c79b07f310e075e824"
+  >
+    <title id="714d29e78a3867c79b07f310e075e824">Close</title>
+    <path d="M12 10.586l4.95-4.95 1.414 1.414-4.95 4.95 4.95 4.95-1.414 1.414-4.95-4.95-4.95 4.95-1.414-1.414 4.95-4.95-4.95-4.95L7.05 5.636l4.95 4.95z" />
+  </svg>
+);
+
 export const Modal = ({
   children,
   size = 'default',
@@ -25,21 +40,6 @@ export const Modal = ({
   overlay,
   onClose,
 }) => {
-  const Icon = () => (
-    <svg
-      width="24"
-      height="24"
-      viewBox="0 0 24 24"
-      className="crayons-icon"
-      xmlns="http://www.w3.org/2000/svg"
-      role="img"
-      aria-labelledby="714d29e78a3867c79b07f310e075e824"
-    >
-      <title id="714d29e78a3867c79b07f310e075e824">Close</title>
-      <path d="M12 10.586l4.95-4.95 1.414 1.414-4.95 4.95 4.95 4.95-1.414 1.414-4.95-4.95-4.95 4.95-1.414-1.414 4.95-4.95-4.95-4.95L7.05 5.636l4.95 4.95z" />
-    </svg>
-  );
-
   return (
     <div
       className={`crayons-modal${getAdditionalClassNames({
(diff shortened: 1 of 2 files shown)


app/javascript/crayons/Modal/Modal.jsx
```diff
@@ -17,6 +17,21 @@ function getAdditionalClassNames({ size, className }) {
   return additionalClassNames;
 }
 
+const CloseIcon = () => (
+  <svg
```
nickytonline, line 21: This doesn't need to be in the render of the function component.

app/javascript/crayons/Modal/Modal.jsx
```diff
@@ -25,34 +40,19 @@ export const Modal = ({
...
-
   return (
     <div
       className={`crayons-modal${getAdditionalClassNames({
         size,
         className,
       })}`}
     >
-      <div className="crayons-modal__box">
+      <dialog open className="crayons-modal__box">
```
nickytonline, line 50: Using the [\<dialog /\>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) HTML element. `open` is always present because of the way we currently render modals. If this component is rendered, it means the modal is open.

nickytonline:
> I believe both Safari and Firefox don't support the full <dialog> API thus I think introducing this element, thus semantically rich, might further restrict the number of browsers we support? Let me know if I'm wrong here

We don’t use the JS API right now, so it was more just using the HTML element. If you look at the test instructions, the modal should still render.

If we were to use the dialog JS API then yes it would break things in other browsers.

Looking at https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html,   it probably makes sense to keep it as a dive and give it a `role=“dialog”` and `aria-modal=“true”`.

rhymes:
@nickytonline sure, my preoccupation here is that by introducing it we might encourage people to start using the API as well thus breaking things. Side question: is the modal closable with the X on Safari iOS? Does that require the JS API?

rhymes:
BTW I also found this that might help in the future if we want to go "full on dialog" not having to deal with compatbility issues ourselves https://github.com/edenspiekermann/a11y-dialog

nickytonline:
> @nickytonline sure, my preoccupation here is that by introducing it we might encourage people to start using the API as well thus breaking things. Side question: is the dialog based modal closable with the `X` close button on Safari iOS? Does that require the JS API?

The close is just an onclick event handler not related ditto the JS API.

nickytonline:
@rhymes, I made the changes I mentioned and also realized this component did not have tests. 🙃

app/javascript/crayons/Modal/Modal.jsx
```diff
@@ -25,44 +40,32 @@ export const Modal = ({
...
   return (
     <div
+      data-testid="modal-container"
       className={`crayons-modal${getAdditionalClassNames({
         size,
         className,
       })}`}
     >
-      <div className="crayons-modal__box">
+      <div role="dialog" aria-modal="true" className="crayons-modal__box">
```
nickytonline, line 51: Just going with this for now as per [wai-aria dialog](https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html) recommendations.

app/javascript/crayons/Modal/Modal.jsx
ludwiczakpawel: Out of curiosity: what's the difference in terms of where is it placed?

nickytonline: It gets created every time if it in the function component. The way it is now, it is created once (the JS variable, not the Preact component), and then it is always referenced when being used as the `icon` prop.

PR #9565, opened 2020-07-29
Rubocop: fix Rails/HasManyOrHasOneDependent for Ahoy::Visit and Badge

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

I've decided to split https://github.com/forem/forem/pull/9405 in multiple PR with a smaller scope/surface.

This PR tackles `Rails/HasManyOrHasOneDependent` for `Ahoy::Visit` and `Badge`.

1. `Ahoy::Visit` has a `has_many :events` relation with `Ahoy::Events` (thus sadly it has not foreign key to it, but we can add it in a future PR). As `Ahoy::Event` can't be built or is valid without a correspondent `Ahoy::Visit` we can safely destroy `events` when `visits` are destroyed. 

1. `Badge` has two relations we're interested in: `badge_achievements` and `tags`
  - achievements in the achievement of a badge by a particular user. Initially in #9405 I thought "let's remove those if a badge is destroyed" but I think it'd be more transparent to the operator to be told: "you can't remove this badge because it's referenced in achievements" which is what the clause `dependent: :restrict_with_error` does
 - tags can be assigned to a badge, through their `badge_id`. Initially in #9405 I thought "let's set it automatically to null" but as with achievements, I'd prefer to notify the operator that they are trying to remove a badge which some tags point to. 

The result of the restrictions on badge removals are summarize with this:

```ruby
[7] pry(main)> badge.errors_as_sentence
=> "Cannot delete record because dependent tags exist and Cannot delete record because dependent badge achievements exist"
```

Let me know if any of this makes sense or you'd prefer the previous combination of "dependent clauses".

https://github.com/forem/forem/pull/9405

## Added tests?

- [x] yes
- [ ] no, because I need help

I've added a data update script to remove orphaned `Ahoy::Event`s
There is no chance we have orphaned badge achievements because there's currently a foreign key that forbids destruction of any badge that has achievements:

https://github.com/forem/forem/blob/8b882dd5123889e2e4c8afbd23dd2106cb5d1056/db/schema.rb#L1327

(foreign keys FTW)

I haven't added a script to nullify tags referencing non existing badges because the likelihood of that, given that badges can't really be destroyed right now it's really small. Let me know if you'd prefer I'd add a script anyway.


app/models/ahoy/visit.rb
```diff
@@ -2,7 +2,7 @@ module Ahoy
   class Visit < ApplicationRecord
     self.table_name = "ahoy_visits"
 
-    has_many :events, class_name: "Ahoy::Event"
+    has_many :events, class_name: "Ahoy::Event", dependent: :destroy
```
vaidehijoshi, line 5: Hm, I feel like this is something the ahoy gem _should_ just do for us? It's weird to me that it [doesn't](https://github.com/ankane/ahoy/blob/master/test/internal/app/models/ahoy/visit.rb#L4)?

rhymes: It might be because they want the implementor to decide what to do? But yeah, it's weird to me they don't automatically add the foreign key between the two tables at least

PR #9567, opened 2020-07-29
Add edge caching logic to series pages

benhalpern:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

I noticed these pages were a little slow. Adding edge caching headers and small optimization.

Also skipping displaying collections with no articles.

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -1,12 +1,17 @@
 class CollectionsController < ApplicationController
+  before_action :set_cache_control_headers, only: %i[index show]
+
   def index
     @user = User.find_by!(username: params[:username])
-    @collections = @user.collections.order(created_at: :desc)
+    @collections = @user.collections.includes(:articles).order(created_at: :desc)
+    @articles = Article.published.where(collection_id: @collections.pluck(:id))
+    set_surrogate_key_header Article.table_key, @articles.map(&:record_key)
   end
 
   def show
     @collection = Collection.find(params[:id])
     @user = @collection.user
     @articles = @collection.articles.published.order(Arel.sql("COALESCE(crossposted_at, published_at) ASC"))
+    set_surrogate_key_header Article.table_key, @articles.map(&:record_key)
   end
 end
--- a/app/views/collections/index.html.erb
+++ b/app/views/collections/index.html.erb
@@ -12,6 +12,7 @@
   <div class="grid gap-2 m:gap-4 l:gap-6 m:grid-cols-2 l:grid-cols-3 px-2 m:px-0">
     <% if @collections.present? %>
       <% @collections.each do |collection| %>
+        <% next if collection.articles.published.empty? %>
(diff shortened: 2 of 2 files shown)


app/controllers/collections_controller.rb
```diff
@@ -1,12 +1,17 @@
 class CollectionsController < ApplicationController
+  before_action :set_cache_control_headers, only: %i[index show]
+
   def index
     @user = User.find_by!(username: params[:username])
-    @collections = @user.collections.order(created_at: :desc)
+    @collections = @user.collections.includes(:articles).order(created_at: :desc)
+    @articles = Article.published.where(collection_id: @collections.pluck(:id))
```
citizen428, line 7: **suggestion (non-blocking):** Based on @rhymes recent PR, I think you may be able to use `@collections.ids` here.

app/views/collections/index.html.erb
```diff
@@ -12,6 +12,7 @@
   <div class="grid gap-2 m:gap-4 l:gap-6 m:grid-cols-2 l:grid-cols-3 px-2 m:px-0">
     <% if @collections.present? %>
       <% @collections.each do |collection| %>
+        <% next if collection.articles.published.empty? %>
```
citizen428, line 15: **thought (non-blocking):** I'm wondering if these could already be filtered out in the controller, i.e. should `@collections` only include collections with at least one published article?

app/controllers/collections_controller.rb
rhymes: Yeah, theoretically rubocop should have flagged this

rhymes: Shouldn't we use `Collection.table_key` here? The collections are going to be updated when the article is updated:

https://github.com/forem/forem/blob/6ac637b9ef4c74e09201065abe662e2d1e9e8952/app/models/article.rb#L90

https://github.com/forem/forem/blob/6ac637b9ef4c74e09201065abe662e2d1e9e8952/app/models/article.rb#L672-L674

app/controllers/collections_controller.rb
```diff
@@ -1,12 +1,17 @@
...
+    set_surrogate_key_header Article.table_key, @articles.map(&:record_key)
   end
 
   def show
     @collection = Collection.find(params[:id])
     @user = @collection.user
     @articles = @collection.articles.published.order(Arel.sql("COALESCE(crossposted_at, published_at) ASC"))
+    set_surrogate_key_header Article.table_key, @articles.map(&:record_key)
```
rhymes, line 15: Here as well, I think we can manage by just using the collection key

msarit:
@benhalpern Hey Ben! 👋🏾  Just drawing your attention back to this PR re: comments 😄

mstruve:
Hey! Since this still has some failing specs and conflicting files and has not been active for a while I am going to close it. I think if we run into issues in the future with these pages we can revisit this approach.

Issue #6994, opened 2020-04-01
Generalize specific user's profile fields


lightalloy:
We have several fields in the `User` model that are specific to developers, like `currently_hacking_on`, or work-related ones, like `employment_title` or `employer_name`.

The fields are:
- set up on the user's profile form (`app/views/users/_profile.html.erb`)
- displayed on the user's profile page (`app/views/users/_sidebar.html.erb`)
- displayed on the article sidebar: (`app/views/articles/_user_metadata.html.erb`)

We may want to make the fields that are set by a user and displayed on the user's and article's pages.

lightalloy:
To our awesome contributors, please note that this issue will be worked on by a DEV team member and is not up for grabs.

Ridhwana:
@lightalloy wanted to find out what your thoughts were on implementation details for the above?

lightalloy:
@Ridhwana I don't have a complete solution in my mind right now but here are some thoughts:
We could have a `jsonb` field in `users` (or a related table) to store community-specific fields. The community-specific fields themselves could be set up in the `SiteConfig`.
All of these fields will be set up in the user form, so there is no need for a configuration here.
But we'll need a configuration to set up which fields will be displayed in the article sidebar and on the user profile page. This could be another configuration in `SiteConfig` (could be set up together with the list of the community-specific fields, but for now it seems like a separate one, should be more clear while working on it)

However, we should keep in mind that this approach will require existing data migration, and think of the cases when the configuration changes (if it will cause problems or not).

benhalpern:
I think we should move to get this implemented.

I think we should use the `user_optional_fields` so each user can have a `has_many` relationship with their fields. We can use `SiteConfig` to manage which user fields exist, but then dynamically generate them if they are added in settings or onboarding.

We can have an `onboarding_profile_fields` config to dictate which are included.

...And then finally, the primary reason this table was added was to literally include _optional_ fields, so I think we could add the ability to define some custom fields such as pronouns etc. But I think that could be secondary to getting this working for generalization/onboarding.

citizen428:
I had a look at this now. `user_optional_fields` were added in April (#7246) to accommodate custom pronouns (#2365). As part of the discussion on that issue, it was [suggested](https://github.com/thepracticaldev/dev.to/issues/2365#issuecomment-555044836) that

(a) custom fields should be implemented in a separate model and
(b) limited in number (the current maximum is 3, see `validate_quota` in `app/models/user_optional_field.rb`﻿﻿).

[...]

PR #9580, opened 2020-07-31
Add ProfileField model

citizen428:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR introduces a new `ProfileField` model which will be used to store the list of available profile fields.

Notes:

* The list of columns is most likely not complete yet, so this table may get updated in future PRs.
* Integers should be fine as primary keys for this model since I don't expect any Forem to have more than 2 billion profile fields (2,147,483,647 is the max for int IDs).
* For the same reason, there are no indexes apart from the unique index on `label`. Fields numbers should be low enough that the query planner would use a seq scan over an index scan anyway.

Part of #6994

Not much to QA yet, this just adds the model.

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/app/models/profile_field.rb
+++ b/app/models/profile_field.rb
@@ -1,0 +1,12 @@
+class ProfileField < ApplicationRecord
+  # Key names follow the Rails form helpers
+  enum input_type: {
+    text_field: 0,
+    text_area: 1,
+    check_box: 2,
+    color_field: 3
+  }
+
+  validates :label, presence: true, uniqueness: { case_sensitive: false }
+  validates :active, presence: true
+end
--- a/app/models/profile_field.rb
+++ b/app/models/profile_field.rb
@@ -8,5 +8,5 @@ class ProfileField < ApplicationRecord
   }
 
   validates :label, presence: true, uniqueness: { case_sensitive: false }
-  validates :active, presence: true
+  validates :active, inclusion: { in: [true, false] }
 end
--- a/spec/factories/profile_fields.rb
+++ b/spec/factories/profile_fields.rb
@@ -1,0 +1,8 @@
+FactoryBot.define do
+  factory :profile_field do
+    label { "Email" }
+    input_type { :text_field }
+    placeholder_text { "john.doe@example.com" }
+    active { true }
+  end
+end
--- a/spec/factories/profile_fields.rb
+++ b/spec/factories/profile_fields.rb
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :profile_field do
-    label { "Email" }
+    sequence(:label) { |n| "Email #{n}" }
     input_type { :text_field }
     placeholder_text { "john.doe@example.com" }
     active { true }
--- a/db/migrate/20200731041554_create_profile_fields.rb
+++ b/db/migrate/20200731041554_create_profile_fields.rb
@@ -1,0 +1,14 @@
+class CreateProfileFields < ActiveRecord::Migration[6.0]
+  def change
+    create_table :profile_fields do |t|
+      enable_extension("citext")
+
+      t.citext :label, null: false, index: { unique: true }
+      t.integer :input_type, null: false, default: 0
+      t.string :placeholder_text
+      t.boolean :active, null: false, default: true
+
+      t.timestamps
+    end
+  end
+end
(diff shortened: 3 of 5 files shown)


app/models/profile_field.rb
```diff
@@ -0,0 +1,12 @@
+class ProfileField < ApplicationRecord
+  # Key names follow the Rails form helpers
+  enum input_type: {
+    text_field: 0,
+    text_area: 1,
+    check_box: 2,
+    color_field: 3
+  }
+
+  validates :label, presence: true, uniqueness: { case_sensitive: false }
```
citizen428, line 10: This column uses Postgres' case-insensitive text type (`citext`) so the case insensitive uniqueness is also ensured when we don't go through the model:

```
PracticalDeveloper_development=# select label from profile_fields;
 label
-------
 Email
(1 row)

PracticalDeveloper_development=# insert into profile_fields (label, created_at, updated_at) values ('email', now(), now());
ERROR:  duplicate key value violates unique constraint "index_profile_fields_on_label"
DETAIL:  Key (label)=(email) already exists.
```

This seems better than a functional index and is also supported by the schema DSL. Rails natively supports `citext` since 4.2 btw, so this is not some new experimental feature.

citizen428:
Force pushed after rebasing against `master` to resolve conflict in `schema.rb`.

citizen428:
Marking this as ready for review, since I'd rather merge it before we get more merge conflicts on `schema.rb`. All follow up work like admin UI and seed data can be done in follow-up PRs.

app/models/profile_field.rb
```diff
@@ -0,0 +1,12 @@
...
+  validates :active, presence: true
```
Ridhwana, line 11: @citizen428 

Since [presence uses .blank?](https://edgeguides.rubyonrails.org/active_record_validations.html#presence) and since false.blank? is true, if we want to validate the presence of a boolean field should we not be using:
```validates : active, inclusion: { in: [true, false] }```

otherwise, trying to set active to false results in the ffg rollback:
```
[22] pry(main)> p.update(active: false)
   (0.3ms)  BEGIN
  ProfileField Exists? (0.4ms)  SELECT 1 AS one FROM "profile_fields" WHERE LOWER("profile_fields"."label") = LOWER($1) AND "profile_fields"."id" != $2 LIMIT $3  [["label", "Name"], ["id", 1], ["LIMIT", 1]]
   (0.2ms)  ROLLBACK
=> false


p.update(active: 0)
   (0.3ms)  BEGIN
  ProfileField Exists? (0.4ms)  SELECT 1 AS one FROM "profile_fields" WHERE LOWER("profile_fields"."label") = LOWER($1) AND "profile_fields"."id" != $2 LIMIT $3  [["label", "Name"], ["id", 1], ["LIMIT", 1]]
   (0.2ms)  ROLLBACK
=> false
```

File: app/models/profile_field.rb (hunk 1 above)
Comment (mstruve) [reply app/models/profile_field.rb:10]: TIL!

File: spec/factories/profile_fields.rb [hunk 3]
```diff
@@ -0,0 +1,8 @@
+FactoryBot.define do
+  factory :profile_field do
+    label { "Email" }
```
mstruve, line 3: Since we are validating uniqueness for this field I think we should use an incrementer to set the value like we do for emails and such to ensure these records are always valid for tests.

app/models/profile_field.rb
mstruve: nice catch!

citizen428: @Ridhwana You're right, it's also in the [documentation of the uniqueness validator](https://api.rubyonrails.org/classes/ActiveRecord/Validations/ClassMethods.html#method-i-validates_presence_of):

> If you want to validate the presence of a boolean field (where the real values are true and false), you will want to use `validates_inclusion_of :field_name, in: [true, false]`.

Not the first time this got me and probably won't be the last 😉

spec/factories/profile_fields.rb
citizen428: Good idea, done.

citizen428:
Thanks @Ridhwana and @mstruve, changes made.

PR #9586, opened 2020-07-31
Improve accessibility of welcome message

marcysutton:

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As a new user, I get a welcome message that includes a button with no discernible text (the close X) and a group of items that should be marked up as a list. This PR introduces those changes!

One thing to note, I couldn't get `erb-lint` to run in my local environment so I used the `--no-verify` flag to keep this work moving along. You might want to double-check it!

The display should be the same, as there is CSS matching UL elements with classes to remove list styles, margins, and padding. The Close Button label can be verified in the Chrome DevTools accessibility inspector or the [axe extension](https://www.deque.com/axe/).

## Added tests?

I don't think this part of the code is tested yet, hence no tests.

- [ ] yes
- [X] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [X] no documentation needed

Spongebob doing a flip with his mop

### DIFF
--- a/app/views/onboardings/_task_card.html.erb
+++ b/app/views/onboardings/_task_card.html.erb
@@ -17,7 +17,7 @@
 </style>
 
 <div class="onboarding-task-card">
-  <button class="close" onclick="closeTaskCard()">
+  <button class="close" onclick="closeTaskCard()" aria-label="Close Welcome message">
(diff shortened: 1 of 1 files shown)


marcysutton:
Thanks all, CLA signed!

PR #9605, opened 2020-08-03
Seed profile fields

citizen428:

- [X] Feature

## Description

This PR explores how to add profile fields to new Forems. There are two constraints:

1. We want to seed different profile fields so admins don’t have to create everything from scratch. 
2. We don't want to add too many fields by default that admins have to remove/deactivate.

The possible solution explored in this PR creates different service classes for groups of fields. These get used in `db/seeds.rb` (where we probably just add all fields for local development), but also in other contexts like e.g. an internal admin controller.

Part of #6994

1. Ensure you currently have no `ProfileField`s (in the DB): `ProfileField.destroy_all`.
2. Run `rails db:seed`.
3. You know should have 29 `ProfileField` records.

Alternatively, you can also call single classes directly, e.g. `ProfileFields::AddBaseFields.call`.

## Added tests?

- [x] yes

## Added to documentation?

- [x] no documentation needed

alt_text


app/services/profile_fields/add_base_fields.rb
```diff
@@ -0,0 +1,8 @@
+module ProfileFields
+  class AddBaseFields < AddFields
+    field "Name", :text_field, "John Doe"
```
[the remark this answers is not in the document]
Ridhwana: Love how clean and isolated this is.

bin/setup
```diff
@@ -33,7 +33,8 @@ chdir APP_ROOT do
   puts "\n== Preparing Test Elasticsearch =="
   system! 'RAILS_ENV="test" bin/rails search:setup'
 
-  system! "bin/rails app_initializer:setup"
+  puts "\n== Initializing the application =="
+  system! "bin/rails app_initializer:setup forem:setup"
```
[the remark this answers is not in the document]
Ridhwana: I prefer this, we can probably seed the mandatory fields here as well.

db/seeds.rb
```diff
@@ -492,6 +492,13 @@ def create_if_none(klass, count = nil)
 end
 
 ##############################################################################
+
+seeder.create_if_none(ProfileField) do
+  ProfileFields::AddBaseFields.call
```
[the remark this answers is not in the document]
Ridhwana: Would we seed the links here too for development, using the AddLinkFields call?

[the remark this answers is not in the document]
citizen428: Yes, we would. I just didn't add it here yet, this whole PR is still very much WIP.

bin/setup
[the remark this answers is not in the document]
citizen428: Great point @Ridhwana, this is much more concrete than my vague feeling that this is probably going to come in handy in the future 😄

20200804035648_add_explanation_to_profile_fields.rb
```diff
@@ -0,0 +1,5 @@
+class AddExplanationToProfileFields < ActiveRecord::Migration[6.0]
+  def change
+    add_column :profile_fields, :explanation, :string
```
citizen428, line 3: Like I mentioned in #9580, I'll keep adding columns to this table as the need arises.

app/services/profile_fields/add_base_fields.rb
```diff
@@ -0,0 +1,11 @@
+module ProfileFields
+  class AddBaseFields
+    include FieldDefinition
```
citizen428, line 3: For people who reviewed the draft PR previously: I switched this from inheritance to a mixin, as this isn't really "X is a Y" but "X behaves like a Y".

citizen428:
@Ridhwana I did quite a bit of work on this today, so marking it as ready for review.

db/seeds.rb
```diff
@@ -492,6 +492,17 @@ def create_if_none(klass, count = nil)
...
+  ProfileFields::AddLinkFields.call
+  ProfileFields::AddWorkFields.call
+  ProfileFields::AddCodingFields.call
```
Ridhwana, line 500: I just want to clarify a follow up question that I had whilst going through the services. 

Will these specific services `AddWorkFields` and `AddCodingFields` be used only in our local development seeds or is this something we're planning on seeding the forem prod database with as well (to show up on the admin ui)?

20200804035648_add_explanation_to_profile_fields.rb
```diff
@@ -0,0 +1,5 @@
+class AddExplanationToProfileFields < ActiveRecord::Migration[6.0]
```
Ridhwana, line 1: issue (blocking): @citizen428 oops this file is in the wrong place, it's in the root folder instead of in db/migrate

20200804035648_add_explanation_to_profile_fields.rb
citizen428: Ah yes, that is a better name.

20200804035648_add_explanation_to_profile_fields.rb
citizen428: Haha, I had temporarily moved it away when fixing some things locally, then moved it back into the wrong location 🤦 Thanks for spotting this!

db/seeds.rb
citizen428: This is a good question. I think we can add the work fields as a preset for other Forems, but I'm not sure how useful the coding related ones will be.

Ridhwana:
@citizen428 The migration and mixin changes look good, however all the services are still using the explanation field.

Ridhwana:
Ready to be reviewed 🎉

app/services/profile_fields/add_coding_fields.rb
```diff
@@ -0,0 +1,21 @@
+module ProfileFields
+  class AddCodingFields
```
mstruve, line 2: Given some of these are very community-specific I really don't think these belong in the codebase. 

I understand that adding fields from scratch might be tedious but I think a better solution would be a bulk import feature using a CSV or something like that and then in our Forem documentation or playbooks we offer different CSVs that have fields tailored to different communities. For example one CSV might have technical fields for a tech community, another might be geared to Forems for corporations, another for pastime Forems.

What are your thoughts on that?

citizen428: Just to clarify, where you mostly talking about the coding related fields? Here's how I currently think about the presets:

* base fields we want for every community, so I call them in the installer.
* The work-related fields seem generic enough that other communities may want them.
* Links are already a bit debatable since they are heavily skewed towards development and design.
* The coding fields are not particularly useful for anyone, I mentioned that above.

Let me give a CSV importer a shot, that's usually very easy in Ruby.

citizen428: Ok, I have a nicely working basic CSV importer now, will add some more tests around it. Thanks for the idea @mstruve!

app/services/profile_fields/import_from_csv.rb
```diff
@@ -0,0 +1,21 @@
+module ProfileFields
+  class ImportFromCsv
+    HEADERS = %i[label input_type placeholder_text description].freeze
```
citizen428, line 3: The CSV files we import don't need a header row. I also made myself a note to properly document all of this, but first I want to get the implementation nailed down.

lib/data/coding_profile_fields.csv
```diff
@@ -0,0 +1,4 @@
+Skills/Languages,text_area,,What tools and languages are you most experienced with? Are you specialized or more of a generalist?
```
citizen428, line 1: I added this file here so we can use it in `db/seeds.rb`, so our team members can easily get a profile as close to the current DEV one as possible. The file is tiny and we have other things in `lib/` that are DEV specific, e.g. old data update scripts.

app/services/profile_fields/import_from_csv.rb
```diff
@@ -0,0 +1,11 @@
...
+
+    def self.call(file)
+      CSV.foreach(file, headers: HEADERS, skip_blanks: true) do |row|
```
citizen428, line 6: We can tweak the CSV options if needed. I generally prefer semicolons or tabs as separators, since not all programs implement quoting correctly (see [RFC4180](https://tools.ietf.org/html/rfc4180,) sections 2.4 and 2.5), but I left it at comma for now.

spec/fixtures/files/profile_fields.csv
```diff
@@ -0,0 +1,4 @@
+Name,text_field,John Doe,
```
citizen428, line 1: Example CSV used in the spec. Includes a blank line (which will be skipped) as well as a correctly quoted string with comma inside.

spec/services/profile_fields/import_from_csv_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
+
+RSpec.describe ProfileFields::ImportFromCsv do
+  # rubocop:disable RSpec/BeforeAfterAll
```
citizen428, line 5: The CSV import is somewhat slow since it reads from disk and writes to the DB. I didn't feel like doing this 4 times in a row, so I only import once and manually clean up after myself. I also just realized that this would make sense as a source code comment.

citizen428:
Changes made:

1. Added a CSV importer.
2. Removed `ProfileFields::AddCodingFields` service object.
3. Added CSV for those fields in `lib/data` and modified `seeds.rb` to use that.
4. Added some more code comments.

I added more comments inline.

Re-requesting review @Ridhwana @mstruve

citizen428:
@Ridhwana I now also added groups here, should at least be enough to help you with the admin UI.

bin/setup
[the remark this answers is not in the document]
mstruve: Yeah I want to keep everything as central and in one place as possible. Makes @jdoss's life easier bc there is only one rake tasks he has to worry about calling to setup the entire application.

lib/tasks/forem.rake
```diff
@@ -0,0 +1,8 @@
+namespace :forem do
+  desc "Performs basic setup for new Forem instances"
+  task setup: :environment do
+    puts "\n== Setting up profile fields =="
+    ProfileField.destroy_all
```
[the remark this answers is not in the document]
mstruve: FYI the initializer task is run in production so as of right now we will always be resetting our profile fields on every deploy. That seems a little excessive. We might want to guard clause this based on the environment

Issue #2370, opened 2019-04-11
Add Google Lighthouse to Travis CI build


rhymes:
**Is your feature request related to a problem? Please describe.**

[Google Lighthouse](https://developers.google.com/web/tools/lighthouse/) is a great tool to keep an eye on the performance and quality of web pages. But, like all tools, it needs to be regularly checked. I propose to activate the [Google Lighthouse Bot](https://github.com/GoogleChromeLabs/lighthousebot) on PRs through Travis CI. 

An example output:

![](https://user-images.githubusercontent.com/238208/46586553-dfffe600-ca34-11e8-9eea-3846b41a4360.png)

**Describe the solution you'd like**

The bot can be activated and it can run in two modes:

* show the summary for human consumption
* show the summary AND fail a PR if a predefined metrics falls under a given threshold.

I'd activate the first step initially, and maybe consider failing PRs in the future once a better understanding of those metrics is acquired as all tools need to be taken with a grain of salt and weighted against specific goals.

aspittel:
Wow! This is an awesome idea! @maestromac and @benhalpern may have more ideas, but I think this sounds so cool! Thanks for thinking of it @rhymes!

maestromac:
I like this idea but we might have to set up a pr-staging server first.

benhalpern:
I'm for this

jessleenyc:
I don't think we can do this right away because of the current overhead it'll take to set up. Hopefully a headless option comes out soon that'll make it easier for us to incorporate.

rhymes:
Hi @jessleenyc, got it :-) I thought it was headless already. It does require an API key to talk to Google servers though :(

abraham:
There are options that don't use Google's service. I've used [lighthouse-ci](https://www.npmjs.com/package/lighthouse-ci) on Travis with just pointing it to a localhost server in the test environment.

Glennmen:
Google has no released an official [Lighthouse CI](https://github.com/GoogleChrome/lighthouse-ci)
They announced it on [Chrome Dev Summit](https://twitter.com/ChromiumDev/status/1193966926214483968?s=20).

rhymes:
@maestromac this [Lighthouse CI](https://github.com/GoogleChrome/lighthouse-ci) looks cool but we'd still need to find a way to start a web server in the build, which with Rails startup speeds seems it's going to be super slooooow

cmgorton:
[...]

PR #9608, opened 2020-08-03
Ensure all elements are contained by landmarks in mod center

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

All elements should be contained by a [landmark element](https://dequeuniversity.com/rules/axe/3.5/region).

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/views/moderations/_mod_sidebar_left.html.erb
+++ b/app/views/moderations/_mod_sidebar_left.html.erb
@@ -1,4 +1,4 @@
-<div class="crayons-layout__sidebar-left">
+<aside class="crayons-layout__sidebar-left">
   <h1 class="fw-heavy fs-3xl mod-center-heading">
     Mod Center
   </h1>
@@ -50,4 +50,4 @@
       <span>Please share in your Connect Channel!</span>
     </div>
   <% end %>
-</div>
+</aside>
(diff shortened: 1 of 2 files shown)


rhymes:
I was wondering if you had any idea of how we could work towards having regression tests for these changes?

nickytonline:
> I was wondering if you had any idea of how we could work towards having regression tests for these changes?

Not really sure how we'd test for this regression aside from running an a11y tool during CI.

rhymes:
@nickytonline yeah, that's what I was thinking. There was an effort back then to run Lightouse-CI which has a11y checks but I don't even know if it would catch stuff like this. Maybe aXe or WAVE have github actions?

nickytonline:
> @nickytonline yeah, that's what I was thinking. There was an effort back then to run Lightouse-CI which has a11y checks but I don't even know if it would catch stuff like this. Maybe aXe or WAVE have github actions?

Definitely worth looking in to.

rhymes:
This is the old issue about adding Lighthouse https://github.com/forem/forem/issues/2370 - I also found this https://github.com/marketplace/axe-linter

I'm going to add a card to our Team's backlog project

PR #9610, opened 2020-08-03
Profile Admin UI Setup Screen (v1)

Ridhwana:

- [ ] Refactor
- [x]  Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

An admin screen to apply CRUD to the profile fields. 
This version is _the absolute_ minimal interface and will be improved upon. 
I haven't yet applied groups to the interface either.

Part of https://github.com/forem/forem/issues/6994 

Go to http://localhost:3000/internal/profile_fields and try any CRUD method. 

_Note: when you update or add an item it will go to the bottom, will tackle that in another PR_. 

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

No 

alt_text


app/models/profile_field.rb
```diff
@@ -1,11 +1,13 @@
 class ProfileField < ApplicationRecord
   # Key names follow the Rails form helpers
-  enum input_type: {
+  INPUT_TYPES = {
```
citizen428, line 3: I don't think we need this change, Rails `enum` exposes all options by default:

```ruby
ProfileField.input_types
{
    "check_box" => 2,
  "color_field" => 3,
    "text_area" => 1,
   "text_field" => 0
}

File: app/controllers/internal/profile_fields_controller.rb [hunk 2]
```diff
@@ -0,0 +1,33 @@
+module Internal
+  class ProfileFieldsController < Internal::ApplicationController
+    layout "internal"
+
+    def index
+      @profile_fields = ProfileField.all
+    end
+
+    def update
+      @profile_fields = ProfileField.find(params[:id])
```
citizen428, line 10: **nitpick (non-blocking):** Since this is the update action, this should probably be singular, i.e. `profile_field`. It probably doesn't need to be an instance variable either, as we're not rendering in `update`.

app/controllers/internal/profile_fields_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+      @profile_fields.update!(profile_field_params)
+      redirect_to internal_profile_fields_path
```
citizen428, line 12: *suggestion (non-blocking):** I know this is only a first pass at the admin UI, but can we set a success flash to let the user know the `update` worked (same for `create` and `destroy`).

app/controllers/internal/profile_fields_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    def destroy
+      @profile_field = ProfileField.find(params[:id])
+      @profile_field.destroy
+      redirect_to internal_profile_fields_path
+    end
+
+    private
+
+    def profile_field_params
+      allowed_params = %i[input_type label active placeholder_text description]
```
citizen428, line 29: **nitpick (non-blocking):** Since this doesn't change, we can pull it out into a (private) constant to avoid recreating the array.

app/views/internal/profile_fields/index.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="crayons-card grid p-6 mb-6 gap-1">
+  <%= form_for [:internal, ProfileField.new], html: { class: "inline-form" } do |form| %>
+    <div class="form-group">
```
citizen428, line 3: **nitpick (non-blocking):** The indentation for this `div` seems off. OCD: triggered 😉

config/routes.rb
```diff
@@ -70,6 +70,7 @@
         delete :remove_admin
       end
     end
+    resources :profile_fields, only: %i[index update create destroy]
```
citizen428, line 73: **suggestion (non-blocking):** You could use the shorter `except` version here.

```suggestion
    resources :profile_fields, except: %i[new edit show]
```

spec/requests/internal/profile_fields_spec.rb
```diff
@@ -0,0 +1,95 @@
+require "rails_helper"
+
+RSpec.describe "/internal/profile_fields", type: :request do
+  include ActiveJob::TestHelper
```
citizen428, line 4: **thought (non-blocking):** I don't think we need this here. Actually we shouldn't need it anywhere anymore, as we're no longer using `ActiveJob`.

spec/requests/internal/profile_fields_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+  let(:user) { create(:user) }
+
+    before do
```
citizen428, line 7: **nitpick (non-blocking):** OCD triggered again. This `before` block seems wrongly indented.

spec/requests/internal/profile_fields_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+      user.add_role(:super_admin)
```
citizen428, line 8: **suggestion (non-blocking):** We have a trait in the factory for this.

```ruby
let(:user) { create(:user, :super_admin) }
```

spec/requests/internal/profile_fields_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+  describe "GET /internal/profile_fields" do
+    let(:profile_field) { create(:profile_field) }
+
+    it "renders successfully" do
+      get internal_profile_fields_path
+      expect(response).to be_successful
+    end
+
+    it "lists the profile fields" do
+      profile_field
```
citizen428, line 21: **suggestion (non-blocking):** You should be able to avoid doing this by using `let!` over `let` above. Or just don't use `let` at all, since this is the only place where you use it in this block anyway.

```suggestion
      profile_field = create(:profile_field)
```

Ridhwana:
Thanks for the feedback @citizen428  🎉 I'll merge this PR and address it in a follow up.

spec/requests/internal/profile_fields_spec.rb
Ridhwana: Removed it from the app and double checked that the specs still pass

config/routes.rb
Ridhwana: @citizen428 this is the only suggestion I didn't implement. I feel that being more verbose (instead of using the negation) is beneficial here.

citizen428: No problem. There are 7 actions, so I usually use whichever form leads to less than 4 in the array. But I know many people prefer to list the available actions instead.

PR #9613, opened 2020-08-03
Feature:Show Status For DataUpdateScript Enum Column

mstruve:
- [x] Feature

## Description
Show status enum string value in Blazer for Data Update Scripts so people dont have to look in the code to figure out what status is what. 
Screen Shot 2020-08-03 at 12 29 34 PM

alt_text

### DIFF
--- a/config/blazer.yml
+++ b/config/blazer.yml
@@ -30,7 +30,7 @@ data_sources:
       # user_id: "/admin/users/{value}"
 
     smart_columns:
-      # user_id: "SELECT id, name FROM users WHERE id IN {value}"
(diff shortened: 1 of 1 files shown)


config/blazer.yml
```diff
@@ -30,7 +30,7 @@ data_sources:
       # user_id: "/admin/users/{value}"
 
     smart_columns:
-      # user_id: "SELECT id, name FROM users WHERE id IN {value}"
+      status: {0: "enqueued", 1: "working", 2: "succeeded", 3: "failed"}
```
citizen428, line 33: 🔥

PR #9615, opened 2020-08-03
Update Users, Follows, Identities, and Reaction Pks to Bigints

mstruve:
- [x] Optimization

## Description
Below is the timing data for each of the migration update statements. Since these migrations are relatively fast I am proposing we just run them and take the brief outage hit. Then the longer table migrations (notifications, ahoy_messages, and articles) we pair with a warning broadcast banner about what exactly is happening. For this group of migrations, since they are so short, I think we could use a single banner stating service might be degraded for a couple of minutes.
```
# follows
# seconds:65.44
# follow buttons break
# User dashboard breaks

# identities
# seconds:27.37964816301246
# Can't log in

# reactions
# seconds:30.13
# Ability to react and see counts breaks

# users
# seconds:19.704
# everything breaks
```
https://github.com/orgs/forem/projects/6#card-43004918

alt_text

### DIFF
--- a/db/migrate/20200803193841_update_user_follow_identity_reaction_pks_to_bigints.rb
+++ b/db/migrate/20200803193841_update_user_follow_identity_reaction_pks_to_bigints.rb
@@ -1,0 +1,70 @@
+class UpdateUserFollowIdentityReactionPksToBigints < ActiveRecord::Migration[6.0]
+  def up
+    ActiveRecord::Base.connection.execute("DROP VIEW IF EXISTS hypershield.follows")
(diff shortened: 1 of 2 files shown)


benhalpern:
Yup, this plan works for me.

If possible I'd like to do do the big one during low-traffic times, so maybe we could plan an odd hour time we might be available to pull it off @mstruve?

mstruve:
@benhalpern for the final migration I am thinking Saturday evening. Seems to be a solid low point in traffic for us 
Screen Shot 2020-08-06 at 2 24 39 PM

PR #9621, opened 2020-08-03
Refactor:Remove ENVied Gem, Use dotenv Instead for Test and Development

mstruve:
- [x] Refactor

## Description
ENVied is a gem that is not really necessary for Forem to run. It has the nice feature of "requiring" ENV variables to be present in production but that has caused us more headaches than it seems to be worth. It [used to provide us default values](https://github.com/eval/envied#what-happened-to-default-values) but they have since changed that in the gem in favor of other best practices.

This PR removes the ENVied gem and replaces it with [dotenv](https://github.com/bkeepers/dotenv) for test and development environments. This gem allows us to use a `.env` file to declare all of our environment variables. To begin, there is a rake task that will take a user's `application.yml` file and convert it to a `.env` file. After a couple of weeks I will remove that rake task. 

I also took the opportunity to make our initializers more resilient. Now if Pusher or AWS creds are missing the whole app won't crash, we just simply won't set those values. This will of course likely lead to broken features but it at least gives application users the ability to run the app without them. 

TODO: remove `sample_application.yml` and Figaro gem

https://github.com/orgs/forem/projects/6#card-43004935

- Pull down the branch
- Run bin/setup
- Run the app and check that a new .env file has been created with everything from your application.yml file. 

## Added tests?
- [x] yes

alt_text


config/initializers/0_application_config.rb
```diff
@@ -1,10 +1,8 @@
-ENVied.require(*ENV["ENVIED_GROUPS"] || Rails.groups)
-
 class ApplicationConfig
   URI_REGEXP = %r{(?<scheme>https?://)?(?<host>.+?)(?<port>:\d+)?$}.freeze
 
   def self.[](key)
-    ENVied.public_send(key)
+    ENV[key]
```
citizen428, line 5: **suggestion (non-blocking):** Should we use `ENV.fetch(key)` here? The previous code raises a `NameError` for invalid keys, the current one returns `nil` instead. Changing to `fetch` would raise a `KeyError` instead.

app/views/internal/configs/show.html.erb
```diff
@@ -888,13 +888,11 @@
...
           </thead>
           <tbody>
-            <% ENVied.config.variables.each do |v| %>
-            <% if v.group.in?(ENVied.required_groups) %>
-            <tr>
-              <td><%= v.name %></td>
-              <td style="max-width: 750px;"><%= ApplicationConfig[v.name] %></td>
-            </tr>
-            <% end %>
+            <% ENV.sort.each do |key, value| %>
```
citizen428, line 891: **question (non-blocking):** Just out of curiosity, will this list be much longer or were most values required anyway so there's barely a difference?

citizen428:
Thanks @mstruve, I didn't get around to look into this during the last cooldown, but it's been bothering me for a while 👏

app/views/internal/configs/show.html.erb
mstruve: We will double the keys we have. I do think this will be better however bc we will have a clear idea of ALL our ENV variables. There have been a couple that have popped up that I didn't realize we even had in Heorku. 
```
irb(main):001:0> ENVied.config.variables.count
=> 72
irb(main):002:0> ENV.keys.count
=> 168
```

config/initializers/0_application_config.rb
mstruve: One of the complaints I have about ENVied is that the entire app crashes if an ENV variable is missing when in reality letting it be nil might not be a huge deal which makes me want to let it be nil here rather than raise an error. Not married to that idea but that's why I did it this way

citizen428: The main problem I had when using ENV directly, were hard to track down bugs due to the fact that there's no indication if the value is missing on purpose or due to e.g. a typo, especially when combined with `.present?` checks:

```ruby
ENV['RAILS_LOG_TO_STDOUT'].present?
#=> true
ENV['RAILS_LOG_TO_STODUT'].present?
#=> false
```

A typo like the second example can go unnoticed for quite a while in my experience.

Maybe a sort of compromise could work:

```ruby
def self.[](key)
  if ENV.has_key?(key)
    ENV[key]
  else
    # Record missing access in log, DD, or wherever
    Rails.logger.warn("Unset ENV variable: #{key}.")
  end
end
```

This doesn't crash the app but at least leaves a message when trying to access a missing value, so we can perform a  sanity check. Alternatively, this could be achieved with a sentinel value, which at least distinguishes it from an accidental `nil`:

```ruby
def self.[](key)
  ENV.fetch(key, :missing_variable)
end

ApplicationConfig["RAILS_LOG_TO_STODUT"]
#=> :missing_variable
```

app/views/internal/configs/show.html.erb
citizen428: That makes sense, thanks for clarifying.

rhymes:
Don't we still load a library to load the environment variables so that Rails can see them?

I have a proposal here, why don't we switch to the dotenv file format?

It's quite standard, it can be read both by shell tools (zsh supports it for example), by direnv and by Ruby with [dotenv](https://github.com/bkeepers/dotenv), basically any language out there has support for dotenv, it doesn't really ties us to YAML and as we leave `ENVied` we're going to lose the data validation check anyway.

Also: it's picked up automatically by [foreman](https://ddollar.github.io/foreman/). 

Basically instead of having a yaml "sample" file, we'd have a dotenv "sample" file and `.env` would replace `application.yml`

cc @jdoss which might have insights

config/initializers/0_application_config.rb
mstruve: LOVE the logging idea and then we could set up a monitor in Datadog to detect those logs so we get alerted but the whole app won't crash.

mstruve:
@rhymes yeah we still need the figaro gem to read and convert the application.yml file. I like the idea of taking it all the way and using a .env file. 

The last piece if anyone has thoughts on it is that I need a way to smoothly transition dev environments from using ENVied to .env files.

rhymes:
oh, I forgot about figaro, I thought everything was done by Envied. So many Rails specific things :D

rhymes:
Regarding transition... we probably need to hook up a script in `bin/setup` that converts the `application.yml` to a `.env` that can only be run once (so it never overwrites) through `bin/setup`. On boot we either keep both for an interim deprecation period or quit if the `.env` isn't there

mstruve:
Cool, that is what I was thinking @rhymes!

config/initializers/0_application_config.rb
citizen428: That sounds great @mstruve 🎉

.gitignore
```diff
@@ -41,6 +41,7 @@ latest.dump
 /node_modules
 yarn-debug.log*
 .yarn-integrity
+.env
```
citizen428, line 44: 🔥

config/initializers/0_application_config.rb
```diff
@@ -1,15 +1,17 @@
...
+      Rails.logger.warn("Unset ENV variable: #{key}.")
+      nil
+    end
   end
 
   def self.app_domain_no_port
     app_domain = self["APP_DOMAIN"]
-    match = app_domain.match(URI_REGEXP)
-    match[:host]
+    app_domain&.match(URI_REGEXP)&.[](:host)
```
citizen428, line 15: **suggestion (non-blocking):** I think an early return would be easier to read than the `&.[](:host)` form:

```suggestion
    return unless app_domain
    
    app_domain.match(URI_REGEXP)[:host]
```

docs/faqs.md
```diff
@@ -21,7 +21,7 @@ By default Rails logs to `log.development.log`.
 If, instead, you wish to log to `STDOUT` you can add the variable:
 
 ```yaml
```
citizen428, line 23: This code block is not YAML anymore.

docs/troubleshooting.md
```diff
@@ -96,8 +96,8 @@ refer to
 ## CORS
 
 If you are experiencing CORS issues locally or need to display more information
-about the CORS headers, add the following variable to your `application.yml`:
+about the CORS headers, add the following variable to your `.env`:
 
 ```yml
```
citizen428, line 101: Also not a YAML code block anymore.

lib/tasks/temporary/env_variables.rake
```diff
@@ -0,0 +1,17 @@
+# Create a .env file from your application.yml file
+task create_dot_env_file: :environment do
+  env_file = File.open(".env", "w")
+
+  File.open("config/application.yml", 'r') do |file|
+    file.each_line do |line|
```
citizen428, line 6: **suggestion (non-blocking):** Instead of `open` + `each_line` you can also use `foreach`:

```ruby
File.foreach("config/application.yml", 'r') do |line|
  new_line = line.gsub(": ", "=")
  ...
end
```

You can use `open` for `env_file` though:

```ruby
File.open(".env", "w") do |env_file|
  File.foreach("config/application.yml", 'r') do |line|
    new_line = line.gsub(": ", "=")
    ...
  end
end
```

mstruve: `File.foreach` here does not split it up by line, it kinda of haphazardly splits it up so when I tried it I got exports in the middle of lines everywhere. I am going to stick with my inner block but then use a block to create the new env file like you suggest

.travis.yml
```diff
@@ -24,8 +24,6 @@ env:
     # Dummy values needed to verify the app boots via "rails runner"
     - APP_PROTOCOL=http://
     - APP_DOMAIN=localhost:3000
-    - SENDGRID_USERNAME_ACCEL=dummy
```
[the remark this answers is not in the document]
citizen428: Nice!

lib/tasks/temporary/env_variables.rake
```diff
@@ -0,0 +1,16 @@
+# Create a .env file from your application.yml file
+task create_dot_env_file: :environment do
```
rhymes, line 2: this breaks the second time is run because `config/application.yml` doesn't exist anymore:

```text
Errno::ENOENT: No such file or directory @ rb_sysopen - config/application.yml
/Users/rhymes/Development/forem/lib/tasks/temporary/env_variables.rake:4:in `initialize'
/Users/rhymes/Development/forem/lib/tasks/temporary/env_variables.rake:4:in `open'
```

as it's contained in `bin/setup` which can be called more than once locally (we designed it to be idempotent), we should probably bail from the script if there's no `config/application.html` in the first place

docs/backend/configuration.md
```diff
@@ -6,7 +6,6 @@ title: Configuration
 
 We currently use the following gems for configuring the application:
 
-- [ENVied](https://github.com/eval/envied)
```
rhymes, line 9: can we say we use dotenv-rails now?

lib/tasks/temporary/env_variables.rake
mstruve: Oooooh really good find! Thanks!

config/initializers/carrierwave.rb
```diff
@@ -27,14 +27,18 @@
...
       config.fog_credentials = {
         provider: "AWS",
         aws_access_key_id: ApplicationConfig["AWS_ID"],
         aws_secret_access_key: ApplicationConfig["AWS_SECRET"],
         region: region
       }
+    else
+      # Fallback on file storage if AWS creds are not present
+      config.asset_host = "https://#{ApplicationConfig['APP_DOMAIN']}/images"
+      config.storage = :file
```
maestromac, line 41: Thanks for adding this!

PR #9628, opened 2020-08-04
[Moderation] Allow admins to delete comments

msarit:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR allows `admins` and `tech_admins` to delete comments on the comment-moderation page.
~**I'm having the following issues:**~
- ~The flash notice/alert does not appear after the controller actions~
- ~The confirm dialog does NOT appear after clicking the "Delete Button"~

1. Make sure your user has `admin`, `super_admin` or `tech_admin` role. 
2. Fire up `localhost`, locate a comment, select "Moderate" from the three-dot menu.
3. On the comment moderation page, scroll down and click "Delete Comment".
4. You should see the comment moderation page, and the comment snippet replaced with `[deleted]`.

Screen Shot 2020-08-03 at 3 52 41 PM

Screen Shot 2020-08-03 at 3 52 08 PM

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Woman Looking Worried


app/policies/comment_policy.rb
```diff
@@ -78,4 +82,8 @@ def user_is_blocked?
   def user_is_commentable_author?
     record.commentable.present? && record.commentable.user_id == user.id
   end
+
+  def user_is_admin?
+    user.any_admin? || user.tech_admin?
+  end
```
Zhao-Andy, line 88: Let's not do tech admin for this, since tech admin is for viewing our technical services.

app/policies/comment_policy.rb
```diff
@@ -39,6 +39,10 @@ def unhide?
     user_is_commentable_author?
   end
 
+  def admin_delete?
+    user_is_admin?
```
Zhao-Andy, line 43: This can be refactored to use the inherited method `user_admin?` from `application_policy.rb`.

app/javascript/packs/modPage.js
```diff
@@ -0,0 +1,100 @@
...
+    formData.append('reactable_id', thisButton.dataset.reactableId);
+
+    getCsrfToken()
+      .then(sendFetch('reaction-creation', formData))
+      .then(function (response) {
+        if (response.status === 200) {
+          response.json().then(successCb);
+        }
+      });
+  };
```
Zhao-Andy, line 44: I feel like you can refactor this :) maybe in a timebox of an hour? If it takes too long we don't need to ship with the refactored version.

app/views/moderations/mod.html.erb
```diff
@@ -1,5 +1,7 @@
 <% title "[Moderate] " + @moderatable.title %>
 
+<%= javascript_packs_with_chunks_tag "modPage", defer: true %>
```
Zhao-Andy, line 3: Could we call this `commentModPage` instead? We have other moderation pages we still use, so IMO it's better to be specific.

app/javascript/packs/modPage.js
msarit: Yup I shall do so in separate PR 😄

app/controllers/comments_controller.rb
```diff
@@ -230,6 +230,20 @@ def unhide
...
   end
 
+  def admin_delete
+    @comment = Comment.find(params[:comment_id])
+    authorize @comment
+    @comment.deleted = true
+
+    if @comment.save
+      redirect = @comment.commentable&.path
+      redirect_to URI.parse(redirect).path, notice: "Comment was successfully deleted."
```
mstruve, line 240: `URI.parse` here will throw an error if redirect is nil. We might want an index fallback path here if that is the case. 
```
[11] pry(main)> URI.parse(nil)
URI::InvalidURIError: bad URI(is not URI?): nil
from /Users/mollystruve/.rvm/rubies/ruby-2.7.1/lib/ruby/2.7.0/uri/rfc3986_parser.rb:18:in `rescue in split'
Caused by NoMethodError: undefined method `to_str' for nil:NilClass
from /Users/mollystruve/.rvm/rubies/ruby-2.7.1/lib/ruby/2.7.0/uri/rfc3986_parser.rb:16:in `split'
```

citizen428: This code path won't execute if `redirect` is `nil` because of `@comment.save && redirect`. That said, I don't think we need to parse at all:

```ruby
URI.parse(redirect).path == redirect
#=> true
```

`redirect` is already a path (due to `&.path`), so parsing it and then calling `.path` on it will result in the same value.

app/controllers/comments_controller.rb
```diff
@@ -230,6 +230,22 @@ def unhide
...
   end
 
+  def admin_delete
+    @comment = Comment.find(params[:comment_id])
+    authorize @comment
+    @comment.deleted = true
+
+    if @comment.save
+      redirect_url = @comment.commentable&.path
+      if redirect
```
mstruve, line 240: I think you meant `redirect_url` here. Also, what is the fallback if redirect_url is nil?

msarit: Good point Molly, will address 😄

app/controllers/comments_controller.rb
```diff
@@ -232,10 +232,32 @@ def unhide
...
 
+  def admin_delete
+    @comment = Comment.find(params[:comment_id])
+    authorize @comment
+    @comment.deleted = true
+
+    if @comment.save
+      redirect_url = @comment.commentable&.path
+      if redirect_url
+        redirect_to redirect_url, notice: "Comment was successfully deleted."
```
msarit, line 243: My `notice` and `alert` are STILL not firing. To be clear, I'm expecting a banner to appear at the top of the page, with the specified message, after the redirect happens.

mstruve: I think what you want here is
```
flash[:success] = "Comment was successfully deleted."
```

Then for failures
```
flash[:danger] = "Something went wrong..."
```

msarit: 🙏🏽

msarit: Doesn't seem to be working Molly. I looked into the mechanism for the messages confirming changes under Settings, but I do not think it uses the same approach as here.

msarit: I think I have it now 👍

app/controllers/comments_controller.rb
```diff
@@ -232,10 +232,34 @@ def unhide
...
 
   # Use callbacks to share common setup or constraints between actions.
   def set_comment
     @comment = Comment.find(params[:id])
   end
+
+  def redirect_to_comment_path
+    redirect_url = @comment.path
+    flash[:error] = "Something went wrong; Comment NOT deleted."
+    redirect_to "#{redirect_url}/mod"
```
Zhao-Andy, line 263: I think you can make this simpler by doing `redirect_to "#{@comment.path}/mod`. No need to make a one time variable.

app/views/moderations/mod.html.erb
```diff
@@ -96,6 +104,16 @@
       This is the equivalent of vomiting on <b>all</b> of this user's articles.
     </p>
   </div>
+  <% if (current_user.any_admin? || current_user.tech_admin?) && @moderatable.class.name == "Comment" && !@moderatable.deleted %>
```
Zhao-Andy, line 107: Let's make sure this is only available to `current_user.any_admin?` and not tech admins.

app/controllers/comments_controller.rb
msarit: I did that initially Andy, but then I got a CodeClimate error about `Possible unprotected redirect`

Zhao-Andy:
Sorry could you merge master again? I think your build is failing because we made a recent commit that should fix the failure.

PR #9633, opened 2020-08-04
Auto-generate CSS utility classes documentation in Storybook

nickytonline:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update
- [x] Developer experience

## Description

CSS utility classes are generated by the https://github.com/forem/forem/blob/master/app/assets/stylesheets/config/_generator.scss file. This PR generates CSS from that file and parses the CSS style sheet to auto-generate Storybook stories to provide documentation for CSS utility classes.

CSS utility classes are grouped by the property they modify. In some cases, some utility classes modify more than one CSS property. In cases like that, the utility class is grouped in each CSS property it modifies.

CSS utility classes will typically have values set with the `!important` attribute so as to work with pre-design system CSS.

Also noteworthy is CSS properties in the auto-generated documentation point to MDN.

image

image

To test:
- Navigate to the [Storybook](https://deploy-preview-9633--storybookdevto.netlify.app/?path=/story/5-css-utility-classes-align-content--content-start) for this PR.
- Section 5 of the Storybook should contain the auto-generated Storybook stories for all the CSS properties that the CSS utility classes modify.
- CSS utilty classes that modify more than one CSS property should appear under each of those CSS properties (the way the Storybook stories are classified for CSS utility classes). For example, the `px-0` utility class appears under `padding-left` and `padding-right`.
  - https://deploy-preview-9633--storybookdevto.netlify.app/?path=/story/5-css-utility-classes-padding-left--px-0
  - https://deploy-preview-9633--storybookdevto.netlify.app/?path=/story/5-css-utility-classes-padding-right--px-0
- The auto-generatated documentation should contain links that redirect to MDN for the CSS properties that are modified by the CSS utility class.

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

Mr. Krabs from Sponge Bob saying "I'll show you automated!"


.gitignore
```diff
@@ -79,3 +79,6 @@ docs/.static/api/
 db_data
 es_data
 gem_cache
+
+# Auto-generated Storybook stories
+app/javascript/generated_stories/
```
nickytonline, line 84: This folder is a build artifact so it should never be committed. It's in the `app/jaavascript` folder so that the stories get picked up by Storybook.

package.json
```diff
@@ -15,10 +15,10 @@
   "scripts": {
     "api-docs:lint": "spectral lint -F hint -v docs/api_v0.yml && lint-openapi -e docs/api_v0.yml",
     "api-docs:serve": "redoc-cli serve docs/api_v0.yml --options.pathInMiddlePanel --options.jsonSampleExpandLevel=all --options.menuToggle -t docs/api_template.hbs --watch",
-    "generate-themes-storybook": "node-sass app/assets/stylesheets/themes -o app/javascript/storybook-static/themes",
+    "storybook-prerequisites": "node-sass app/assets/stylesheets/themes -o app/javascript/storybook-static/themes && node app/javascript/.storybook/generate-utility-classes-documentation.js",
```
nickytonline, line 18: The scripts are starting to get a little busy. I might consider refactoring this in a subsequent PR.

package.json
```diff
@@ -101,6 +101,7 @@
     "babel-jest": "^26.2.2",
     "babel-loader": "^8.1.0",
     "css-loader": "^4.2.0",
+    "cssom": "^0.4.4",
```
nickytonline, line 104: Required so we can load a style sheet object from CSS generated from a SASS file. The style sheet is used to auto-generate Storybook stories.

.vscode/launch.json
```diff
@@ -4,6 +4,13 @@
   // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
   "version": "0.2.0",
   "configurations": [
+    {
```
nickytonline, line 7: Handy for debugging node processes. I used it while writing this PR.

nickytonline:
I just need to write some tests and then this will be ready for review.

app/javascript/utilities/documentation/index.js
```diff
@@ -0,0 +1,160 @@
...
+    )}' };
+    `);
+  }
+
+  return storybookStories.join('');
+}
+
+async function generateUtilityClassesDocumentation(
+  styleSheet,
+  fileWriter = file.writeFile,
```
nickytonline, line 135: Just a handy way to use the file system but also to be able to test things by passing in a mock file writer.

app/javascript/utilities/generate-utility-classes-documentation.js
```diff
@@ -0,0 +1,42 @@
...
+} = require('./documentation');
+
+const folderExists = util.promisify(fs.exists);
+const mkdir = util.promisify(fs.mkdir);
+const stylesheetsDirectory = path.resolve(
+  __dirname,
+  '../../assets/stylesheets',
+);
+
+(async () => {
```
nickytonline, line 18: The current version of node.js that we are using does not support top level await, hence this async [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE).

app/javascript/utilities/documentation/index.js
```diff
@@ -0,0 +1,160 @@
...
+ * @param {string} file The file to load as a style sheet.
+ *
+ * @returns {CSSStyleSheet} The stylesheet for the given file.
+ */
+async function getStyleSheet(file) {
+  const { css: bytes } = await renderCss({
+    file,
+  });
+  const utilityClassesContent = new TextDecoder('utf-8').decode(bytes);
+  const styleSheet = CSSOM.parse(utilityClassesContent);
```
nickytonline, line 27: I got the idea to use cssom after looking up parsing a CSS abstract syntax tree ([AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree)) on [astexplorer.net](https://astexplorer.net)

app/javascript/utilities/documentation/index.js
```diff
@@ -0,0 +1,160 @@
+const fs = require('fs');
```
nickytonline, line 1: This code is not for the app, but for tooling, so still not sure if it belongs in this folder. To be clear though, it won't get bundled via webpack or sprockets.

nickytonline: @rhymes @citizen428, maybe `bin/utitilies`?

citizen428: This may not be the most popular opinion, but unless you use this outside `bin/generate-css-utility-classes-docs.js` too, I'd probably just move it in there. I'd rather have everything in the same place if there's no reuse anyway, and that we're not quite sure where to put this makes me think that even more.

nickytonline: Yeah, sounds good. Some stuff could potentially be reused, but I'm not going to YAGNI. I'll just rework things a bit so that it can still be tested. It's not production code, but it does affect the output of Storybook.

bin/generate-css-utility-classes-docs.js
```diff
@@ -43,4 +187,20 @@ const stylesheetsDirectory = path.resolve(
   } catch (error) {
     throw new Error('Error generating the CSS utilty class Storybook stories');
   }
-})();
+}
+
+if (process.env.NODE_ENV !== 'test') {
```
nickytonline, line 192: When this runs before Storybook is built, the node environment is always undefined, so I'm checking for `"test"` which jest provides for a node environment.

bin/generate-css-utility-classes-docs.js
```diff
@@ -43,4 +187,20 @@ const stylesheetsDirectory = path.resolve(
...
+  // we're running unit tests so do not run the function to generate docs.
+  generateDocumentation();
+}
+
+/*
+  These variables and functions are only used within this script, but they are being exported
+  so that the core of the doc generation tool can be tested. Normally you would not export
+  things deemed private.
+*/
+module.exports = {
```
nickytonline, line 202: Last question before me merge @ben, @citizen428 and @rhymes. I export these as mentioned in the script just to test codegen core functionality. The tests are co-located in `__tests__` in the `bin` folder. Are y'all OK with a `__tests__` folder in the `bin`? If not I can move it to some folder under `app/javascript`.

nickytonline:
This PR works as expected, but there was some additional configuration for Netlify that needed to be done in #9753. Thanks @Zhao-Andy!

PR #9636, opened 2020-08-05
Add foreign keys to Ahoy models

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

As a follow up to https://github.com/forem/forem/pull/9565 I'm adding FKs to Ahoy::Models

By looking at the DB I found:

- 36 visits assigned to non existent users
- 1 event assigned to a non existent users
- couldn't run the query for messages as it times out, see:

```sql
select user_id from ahoy_messages where user_id is not null and user_id not in (select id from users);
```

I added scripts to cleanup the data, please double check the logic is correct


lib/data_update_scripts/20200729120730_remove_orphaned_ahoy_events.rb
```diff
@@ -1,11 +1,7 @@
 module DataUpdateScripts
   class RemoveOrphanedAhoyEvents
     def run
-      Ahoy::Event.find_each do |event|
-        next if event.visit.blank?
-
-        event.destroy
-      end
+      # Deleted the script as it was not necessary
```
rhymes, line 4: this was removed as it contained the wrong logic

lib/data_update_scripts/20200805102636_remove_orphaned_ahoy_rows.rb
```diff
@@ -0,0 +1,16 @@
+module DataUpdateScripts
+  class RemoveOrphanedAhoyRows
+    def run
+      user_ids = User.ids
+
+      # Delete all Ahoy::Events belonging to a user that does not exist anymore
+      Ahoy::Event.where.not(user_id: nil).where.not(user_id: user_ids).find_each(&:destroy)
+
+      # Delete all Ahoy::Messages belonging to a user that does not exist anymore
+      Ahoy::Message.where.not(user_id: nil).where.not(user_id: user_ids).find_each(&:destroy)
```
mstruve, line 10: There are 40255 Ahoy Messages that belong to missing users. I was able to run this in a console
```
ids = ActiveRecord::Base.connection.execute("select id from ahoy_messages where user_id is not null an
d user_id not in (select id from users);").map(&:itself);
```
However, using the user_ids with Rails notation like this in an array causes Postgres to search that whole array and even for Ahoy::Events, the query took 5+ seconds just for a count. I think to ensure this script completes we need to use straight SQL to get the job done. Then we can also ensure that our STATEMENT_TIMEOUT is bumped a little before we deploy to give the requests a little extra time to finish.

rhymes: Make sense! I'll rewrite them using SQL :-)

rhymes: Done! I wrote three SQL queries.

I wish PostgreSQL's SQL supported CTEs like this but alas it doesn't:

```sql
WITH user_ids AS (
  SELECT id from users
)

DELETE FROM ahoy_events
WHERE user_id IS NOT NULL
AND user_id NOT IN (SELECT ID FROM user_ids)

DELETE FROM ahoy_messages
WHERE user_id IS NOT NULL
AND user_id NOT IN (SELECT ID FROM user_ids)

DELETE FROM ahoy_visits
WHERE user_id IS NOT NULL
AND user_id NOT IN (SELECT ID FROM user_ids)
```

db/migrate/20200805102249_validate_add_missing_foreign_keys_to_ahoy_models.rb
```diff
@@ -0,0 +1,8 @@
+class ValidateAddMissingForeignKeysToAhoyModels < ActiveRecord::Migration[6.0]
+  def change
+    validate_foreign_key :ahoy_events, :ahoy_visits
+    validate_foreign_key :ahoy_events, :users
```
mstruve, line 4: Since we have invalid data in the database, won't this fail? I think we may need to separate the validating into a separate PR.

rhymes: Maybe we could create a PR that cleans the DB and then merge this one. What do you think?

mstruve: That sounds good to me!

lib/data_update_scripts/20200805102636_remove_orphaned_ahoy_rows.rb
citizen428: Not that it matters, but you should be able to do this with variable instead of a CTE:

```sql
DO $$
  DECLARE user_ids bigint[];
BEGIN
  user_ids := ARRAY(SELECT id from users);
  DELETE from ahoy_events
  WHERE user_id IS NOT NULL
  AND user_id != ALL(user_ids);
  
  ...
END $$;
```

rhymes: Oh nice, I didn't think of the `DO` block! TIL :D

rhymes:
To be merged after https://github.com/forem/forem/pull/9710 is merged and deployed

db/migrate/20200805102249_validate_add_missing_foreign_keys_to_ahoy_models.rb
maestromac: This is good now

PR #9639, opened 2020-08-05
Move `/internal` to `/admin

joshpuetz:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR moves interfaces from `/internal` to `/admin`, as well as updates documentation. It's part of the overall plan to consolidate all of our admin views under `/admin`: https://github.com/orgs/forem/projects/1

There's a LOT of files changes here, but they are mostly renaming instances of `internal` to `admin`. There are a few things I DIDN'T change:

- The class `InternalPolicy` was kept, as there's already an `AdminPolicy` object 
- References in `app/views/layouts/admin.html.erb` where kept with the name `internal` (javascript and css files) as when I changed them to `admin` I got tons of errors from the asset pipeline.

closes https://github.com/forem/InternalProjectPlanning/issues/5

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [x] docs.dev.to
- [ ] readme
- [ ] no documentation needed

alt_text


joshpuetz:
CodeClimate is complaining about `Unescaped model attribute`, but the lines in question are all escaped with `html_safe`. Any thoughts?

citizen428:
> the lines in question are all escaped with `html_safe`

`html_safe` doesn't escape, it marks a string as trusted. From [the documentation](http://api.rubyonrails.org/classes/String.html#method-i-html_safe):

> Marks a string as trusted safe. It will be inserted into HTML with no additional escaping performed.

This is exactly what CodeClimate complains about. Looking at the 4 reported violations, I'd say at least 3 of them (the `SiteConfig` ones) are fine, since they can only get set by admins. The buffer one is probably fine too, but I don't know the full code path.

Ridhwana:
@joshpuetz I'm super sorry, I noticed this PR after I merged https://github.com/forem/forem/pull/9610/files, we may need to update that route.

Ridhwana:
@joshpuetz I'm not sure if this is my environment and just some caching or something else, but I'm seeing the styles on the moved pages rendering a bit weird. Here's an example of the diff:  

Screenshot 2020-08-06 at 11 07 00

Screenshot 2020-08-06 at 11 01 47

Again, not sure if this is just my local env caching weirdly, but I thought it's worth putting here.

joshpuetz:
> @joshpuetz I'm super sorry, I noticed this PR after I merged https://github.com/forem/forem/pull/9610/files, we may need to update that route.

No worries @Ridhwana: it's on me to catch this brach up!

joshpuetz:
> > the lines in question are all escaped with `html_safe`
> 
> `html_safe` doesn't escape, it marks a string as trusted. From [the documentation](http://api.rubyonrails.org/classes/String.html#method-i-html_safe):
> 
> > Marks a string as trusted safe. It will be inserted into HTML with no additional escaping performed.
> 
> This is exactly what CodeClimate complains about. Looking at the 4 reported violations, I'd say at least 3 of them (the `SiteConfig` ones) are fine, since they can only get set by admins. The buffer one is probably fine too, but I don't know the full code path.


Thanks for the explanation @citizen428! I agree these should be just fine: there doesn't seem to be a way to ignore brakeman warnings inline like Rubocop, so it sounds like I'll need an admin merge for this.

joshpuetz:
> @joshpuetz I'm not sure if this is my environment and just some caching or something else, but I'm seeing the styles on the moved pages rendering a bit weird. Here's an example of the diff:
> 
> Screenshot 2020-08-06 at 11 07 00
> 
> Screenshot 2020-08-06 at 11 01 47
> 
> Again, not sure if this is just my local env caching weirdly, but I thought it's worth putting here.

Thanks, will look into it!

joshpuetz:
Confirmed from @pkfrank that what I'm seeing locally is what it's in production, so I'm calling this ok!

config/initializers/audit_events.rb
```diff
@@ -4,6 +4,6 @@
 # Put here all custom listener names, for later usage in Audit Instrumentation
 #
 # Example:
-# Audit::Subscribe.listen :admin, :quest_user
+# Audit::Subscribe.listen :internal, :quest_user
```
mstruve, line 7: Why are we changing from `admin` to `internal` here instead of the other way around?

joshpuetz: This is actually just fixing the comment so it aligns with the code! I did not rename the internal (pun intended) audit trail name

PR #9641, opened 2020-08-05
Cleanup preferred_languages language_settings for users

atsmith813:
- [x] Refactor

## Description
Currently, in production, we store a user's `language_settings` in different ways. This PR consolidates the preferred languages piece into a single format - a single `preferred_languages` Array. This will affect ~150k records.

https://github.com/orgs/forem/projects/3#card-42702165
https://github.com/orgs/forem/projects/3#card-42702170

## Added tests?

## Added to documentation?
- [x] no documentation needed

Double-check the data looks good via Blazer.

now_you_are_talking_my_language_gif

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -245,21 +245,10 @@ class User < ApplicationRecord
     end
   end
 
-  # handles both old (prefer_language_*) and new (Array of language codes) formats
   def preferred_languages_array
     return @preferred_languages_array if defined?(@preferred_languages_array)
 
-    if language_settings["preferred_languages"].present?
-      @preferred_languages_array = language_settings["preferred_languages"].to_a
-    else
-      languages = []
-      language_settings.each_key do |setting|
-        to_split = language_settings[setting] && setting.include?("prefer_language_")
-        languages << setting.split("prefer_language_")[1] if to_split
-      end
-      @preferred_languages_array = languages
-    end
-    @preferred_languages_array
+    @preferred_languages_array = language_settings["preferred_languages"]
   end
 
   def processed_website_url
--- a/lib/data_update_scripts/20200805171911_clean_up_language_settings.rb
+++ b/lib/data_update_scripts/20200805171911_clean_up_language_settings.rb
@@ -1,0 +1,19 @@
+module DataUpdateScripts
+  class CleanUpLanguageSettings
+    def run
+      User.where("(language_settings->'preferred_languages') IS null").find_each do |user|
+        language_settings = user.language_settings
+        language_settings["preferred_languages"] = []
+
+        language_settings["preferred_languages"] << "en" if language_settings["prefer_language_en"]
+        language_settings["preferred_languages"] << "es" if language_settings["prefer_language_es"]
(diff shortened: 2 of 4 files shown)


app/models/user.rb
```diff
@@ -245,21 +245,10 @@ def cached_following_podcasts_ids
...
-    else
-      languages = []
-      language_settings.each_key do |setting|
-        to_split = language_settings[setting] && setting.include?("prefer_language_")
-        languages << setting.split("prefer_language_")[1] if to_split
-      end
-      @preferred_languages_array = languages
-    end
-    @preferred_languages_array
+    @preferred_languages_array = language_settings["preferred_languages"]
```
atsmith813, line 251: I think we clean this up more later on by updating all references to `preferred_languages_array` to be `language_settings["preferred_languages"]` directly. I chose not to do it in this PR because I suspect we might want to tweak language settings further. Leaving it wrapped up in this method name will allow us to make more changes without updating a bunch of references again.

atsmith813:
@mstruve good call! It would've broken `/settings/misc`. Just updated the code so that won't happen 🚀 .

app/models/user.rb
vaidehijoshi, line 251: Not a blocker for this PR, but looking at the pre-existing code, I wonder if we could do `@preferred_languages_array ||= ` here and return the variable if it exists, rather than `return @preferred_languages_array if defined?(@preferred_languages_array)` ?

atsmith813: Agreed! My plan is to sweep that up in a future round once we know more about how we're going to use this Array after we scope out the project more. If we add back in some logic, then the existing setup would need to remain/be put back in.

atsmith813:
I will merge/deploy this tomorrow since this touches prod data - it would better to have people around in case it goes...off the Rails :trollface:.

rhymes:
Set to draft to avoid accidental merges.

lib/data_update_scripts/20200805171911_clean_up_language_settings.rb
```diff
@@ -0,0 +1,19 @@
...
+      User.where("(language_settings->'preferred_languages') IS null").find_each do |user|
+        language_settings = user.language_settings
+        language_settings["preferred_languages"] = []
+
+        language_settings["preferred_languages"] << "en" if language_settings["prefer_language_en"]
+        language_settings["preferred_languages"] << "es" if language_settings["prefer_language_es"]
+        language_settings["preferred_languages"] << "fr" if language_settings["prefer_language_fr"]
+        language_settings["preferred_languages"] << "it" if language_settings["prefer_language_it"]
+        language_settings["preferred_languages"] << "ja" if language_settings["prefer_language_ja"]
+        language_settings["preferred_languages"] << "pt" if language_settings["prefer_language_pt"]
```
rhymes, line 13: Just one question, is the order in the array going to have an impact on the UX?

atsmith813: Good question! I had to double-check that - no, the order for the UX is defined by a constant:

`<% Languages::LIST.each do |code, name| %>`

PR #9645, opened 2020-08-05
[DO NOT MERGE] Replace `no-cache` in Cache-Control header to support Nginx caching

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

In https://github.com/forem/forem/pull/9631, we conditionally removed the `no-cache` option from the `Cache-Control` headers that are sent along to Fastly + Nginx via our `CachingHeaders` concern, as a way of experimenting to see if Fastly still caches the same way. We did some experiments, and saw that Fastly still picked up on our change, and still caches the way that we'd expect.

You can play with our [Fastly fiddle snippet](https://fiddle.fastlydemo.net/fiddle/62626efe) if you'd like, but the TL;DR can be summed up by these specific headers/response from Fastly:
```bash
HTTP/2 200

cache-control: max-age=86400, public
x-cache-hits: 0, 1, 1
x-cache: MISS, HIT, HIT
x-served-by: cache-den19627-DEN, cache-wdc5547-WDC, cache-wdc5562-WDC
```

Now that we've confirmed that this _doesn't_ break Fastly caching, this PR uses the same exactly Cache-Control headers for _both_ Fastly **and** Nginx:
```ruby
response.headers["Cache-Control"] = "public, max-age=#{max_age}"
```

Once https://github.com/forem/lfss/pull/1 is merged, we expect that [forem.dev](https://forem.dev/) should Just Work™️  with Nginx caching enabled.

Make sure the tests pass! If you are set up with the LFSS repo, you can pull this branch down and QA it to test out whether caching really _does_ work without the explicitly set headers in the `nginx.conf`, which https://github.com/forem/lfss/pull/1 removes :)

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

We need to keep an eye on this PR when it deploys and watch to make sure we're not seeing a bunch of busted caches; ideally, caching should work as normal and we shouldn't see any impacts.

slowly smiling sloth

### DIFF
--- a/docs/design/language.md
+++ b/docs/design/language.md
@@ -1,0 +1,41 @@
+# Human Language
+
+Forem is built on concepts, and those concepts have names. Here are some language concepts to be aware of and some known "gotchas" where there may be multiple terms currently being used.
+
+This is an incomplete first take. It is also written down _so that it can be changed when needed._ No language concept is set in stone. And no concept is exclusively _english_. As we internationalize we need to evolve our notion of this, and it will remain important that we settle on unified terms so that we limit conceptual sprawl.
+
+# Post
+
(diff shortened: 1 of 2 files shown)


vaidehijoshi:
Looks like I have some test failures to address; will investigate and fix them up tomorrow!

PR #9648, opened 2020-08-06
Add profile model

citizen428:

- [X] Feature

## Description

This PR introduces the `Profile` model, which has a 1:1 relation with `User`. Profile data is serialized into a [JSONB column](https://www.postgresql.org/docs/12/datatype-json.html).

Part of #6994

There's not much to QA here yet as this mostly adds a model and some tests.

## Added tests?

- [x] yes

## Added to documentation?

- [x] no documentation needed

under construction


app/models/profile.rb
```diff
@@ -0,0 +1,14 @@
+class Profile < ApplicationRecord
+  belongs_to :user
+
+  # This method generates typed accessors for all active profile fields
+  def self.define_store_accessors!
```
citizen428, line 5: Originally I executed the body directly, but making it a method and calling it right away makes this testable and is also useful when trying things out in a `rails console` session etc.

app/models/profile.rb
```diff
@@ -0,0 +1,14 @@
...
+    ProfileField.active.find_each do |field|
+      store_attribute :data, field.attribute_name, field.type
+    end
+  end
+
+  define_store_accessors!
+
+  validates :data, presence: true
```
citizen428, line 13: The column is non-nullable with a default value of `{}`, which normally greatly simplifies working with JSONB data.

app/models/user.rb
```diff
@@ -49,6 +49,7 @@ class User < ApplicationRecord
   acts_as_followable
   acts_as_follower
 
+  has_one :profile, dependent: :destroy
```
citizen428, line 52: Apart from profiles being inherently tied to users, this was also added to satisfy current specs and the foreign key constraint.

db/migrate/20200806052718_create_profiles.rb
```diff
@@ -0,0 +1,10 @@
+class CreateProfiles < ActiveRecord::Migration[6.0]
+  def change
+    create_table :profiles do |t|
+      t.belongs_to :user, null: false, foreign_key: { on_delete: :cascade }
```
citizen428, line 4: Under certain conditions (internal admin tools) we use `delete` instead of `destroy` on users, so apart from `dependent: destroy` on the model layer I also added cascading to the DB.

db/migrate/20200806052718_create_profiles.rb
```diff
@@ -0,0 +1,10 @@
...
+      t.jsonb :data, null: false, default: {}
```
citizen428, line 5: Ensure we always deal with a JSON object. This is IMHO the most convenient way to deal with JSONB columns.

app/models/profile.rb
```diff
@@ -0,0 +1,14 @@
+class Profile < ApplicationRecord
+  belongs_to :user
+
+  # This method generates typed accessors for all active profile fields
+  def self.define_store_accessors!
+    ProfileField.active.find_each do |field|
+      store_attribute :data, field.attribute_name, field.type
```
citizen428, line 7: This gives us nicely typed accessors types accessors, including the normal `ActiveRecord` type coercions.

spec/models/profile_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    let(:profile) { described_class.new }
+
+    it "defines accessors for active profile fields", :aggregate_failures do
+      expect(profile).to respond_to(:test1)
+      expect(profile).to respond_to(:test2)
+      expect(profile).not_to respond_to(:test3)
+    end
+
+    it "performs ActiveRecord typecasting for profile fields", :aggregate_failures do
```
citizen428, line 22: While this spec feels a little bit like retesting the used gem (which it does) I also added it for documentation purposes.

db/migrate/20200806052718_create_profiles.rb
rhymes: Agree. It's much easier to remove nulls and leave a default :)

citizen428:
Resolved `schema.rb` conflicts with a rebase against `master`, hence the force push.

citizen428:
The spec seems to legitimately fail because the cascading in the migration DSL apparently doesn't work like I thought it does.

citizen428:
Somehow I messed up my local DB quite a bit during experiments yesterday, so I had to force-push again after fixing everything and then fixing the cascading, sorry for that.

citizen428:
> Is the plan to move some of the existing fields on the user currently off into this new table?

@mstruve  Yes, that's exactly the plan, we're just taking this slowly and step by step. But eventually, the user table will lose almost 30 columns.

rhymes:
> the user table will lose almost 30 columns.

"Is this the real life, is this just fantasy? " 

Yes!

PR #9649, opened 2020-08-06
Profile Admin UI Setup Screen Refactor (v1.5) 

Ridhwana:
- [x] Refactor

## Description
Implement some cleanup suggestions from https://github.com/forem/forem/pull/9610
@citizen428 thanks for some of the suggestions.

Part of https://github.com/forem/forem/issues/6994

Go to http://localhost:3000/internal/profile_fields and try any CRUD method.
Note: when you update or add an item it will go to the bottom, will tackle that in another PR.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

No 

alt_text


app/controllers/internal/profile_fields_controller.rb
```diff
@@ -1,32 +1,50 @@
...
     def index
       @profile_fields = ProfileField.all
     end
 
     def update
-      @profile_fields = ProfileField.find(params[:id])
-      @profile_fields.update!(profile_field_params)
+      profile_field = ProfileField.find(params[:id])
+      if profile_field.update(profile_field_params)
+        flash[:success] = "Profile Field updated"
```
rhymes, line 15: I was thinking that instead of saying "profile field updated" you might want to say: `Profile field #{label} updated` so one will know which of the fields has been updated

app/controllers/internal/profile_fields_controller.rb
```diff
@@ -1,32 +1,50 @@
...
+      else
+        flash[:error] = "Profile Field error: #{profile_field.errors_as_sentence}"
```
rhymes, line 17: It's probably redudant to tell the user "Profile Field" as a prefix of the error message. That's the only thing they can edit in this page

app/controllers/internal/profile_fields_controller.rb
```diff
@@ -1,32 +1,50 @@
...
+      end
       redirect_to internal_profile_fields_path
     end
 
     def create
-      @profile_field = ProfileField.create(profile_field_params)
+      profile_field = ProfileField.new(profile_field_params)
+      if profile_field.save
+        flash[:success] = "Profile Field created"
```
rhymes, line 25: same as the note on "update"

app/controllers/internal/profile_fields_controller.rb
```diff
@@ -1,32 +1,50 @@
...
+      end
       redirect_to internal_profile_fields_path
     end
 
     def destroy
-      @profile_field = ProfileField.find(params[:id])
-      @profile_field.destroy
+      profile_field = ProfileField.find(params[:id])
+      if profile_field.destroy
+        flash[:success] = "Profile Field destroyed"
```
rhymes, line 35: same as the note on "update" but I would say "deleted" as "destroy" is a very technical concept here

spec/rails_helper.rb
```diff
@@ -67,7 +67,6 @@
 
   config.include ApplicationHelper
   config.include ActionMailer::TestHelper
-  config.include ActiveJob::TestHelper
```
Ridhwana, line 70: I did do a check on where we using this, but just want to highlight that I'm removing the ActiveJob Test helper.

Ridhwana:
> Nice! Works well. Left some notes on wordings

Thanks for the great tips @rhymes. Updated as per suggestions 😊

app/models/profile_field.rb
```diff
@@ -1,13 +1,11 @@
 class ProfileField < ApplicationRecord
   # Key names follow the Rails form helpers
-  INPUT_TYPES = {
+  enum input_type: {
```
mstruve, line 3: If you want to make these more human-readable in Blazer you can add this column name and enum descriptions [here to our Blazer.yml file](https://github.com/forem/forem/blob/master/config/blazer.yml#L33)

app/controllers/internal/profile_fields_controller.rb
```diff
@@ -1,32 +1,50 @@
...
       redirect_to internal_profile_fields_path
     end
 
     private
 
+    private_constant :ALLOWED_PARAMS
+
     def profile_field_params
-      allowed_params = %i[input_type label active placeholder_text description]
+      allowed_params = ALLOWED_PARAMS
```
juliannatetreault, line 47: I just used the `ALLOWED_PARAMS`-related code as inspiration for something I'm currently working on--this is exactly what I needed! 😄

app/controllers/internal/profile_fields_controller.rb
```diff
@@ -1,32 +1,50 @@
 module Internal
   class ProfileFieldsController < Internal::ApplicationController
+    ALLOWED_PARAMS = %i[
+      input_type label active placeholder_text description
```
citizen428, line 4: **comment (blocking):** We merged the PR that adds `group` already, so I think we should add that here too.

```suggestion
      input_type label active placeholder_text description group
```

Additional suggestion:

When things are still in flux, I often use something like this:

```ruby
ALLOWED_PARAMS = (ProfileField.column_names - %w[id created_at updated_at]).freeze
```
you can always changed it to the explicit list in the final PR.

app/models/profile_field.rb
citizen428: Oh yes, let's please do this. 

Rails enums can also be backed by real [Postgres enums](https://www.postgresql.org/docs/current/datatype-enum.html#id-1.5.7.15.7) but we don't use that anywhere else in the app. The one drawback is that you need to add a migration for new enum values (`ALTER TYPE name ADD VALUE ...`), but otherwise I prefer that over meaningless integer columns.

spec/rails_helper.rb
citizen428: Thanks @Ridhwana! This should probably have been removed a couple of months ago when we finished moving everything over to Sidekiq.

app/controllers/internal/profile_fields_controller.rb
Ridhwana: Thanks @citizen428, correct - I'm leaving everything group related for the next PR where I tackle that. 

Also, thats a pretty cool tip thank you :)

app/models/profile_field.rb
Ridhwana: Done ✅

I do have a very noob question on this after looking through the documentation though. I found it odd that we don't have a table name attached to to the smart column. What if there's a column name in a different table whose enums are different?

Ridhwana:
Note: I'll wait until https://github.com/forem/forem/pull/9639 to merge this PR so that @joshpuetz doesn't have to keep updating his branch 😊

Ridhwana:
I wanted to get this PR in today but I'm struggling to figure out why this spec is failing. I've restarted the build and it's failing locally too. I'll pick this up on Monday morning.

benhalpern:
Tests passed, I deployed.

PR #9650, opened 2020-08-06
Typography fix & Optimization

ludwiczakpawel:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR fixes an issue which made our fonts picker stop working... Users were not able to choose a specific font in `/settings/ux` for like 1 month, surprisingly there was no reports :D 

Additionaly, I've cleaned up a bit of CSS for fonts and improved few things along the way.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

alt_text


ludwiczakpawel:
@benhalpern oh, ok then - yea, I think it's a good idea to make default font a sans-serif one. But in that case I need help with that since it needs some changes in backend / database as well? I'm not sure...

For that to happen, I've changed naming of variables in CSS so we now use more universal names. That's why this PR got so much bigger with my last commit - it's mainly just swapping variables and class names across many files, nothing else.

- `ff-serif`
- `ff-sans-serif` 
- `ff-monospace`

app/assets/stylesheets/base/typography.scss
```diff
@@ -0,0 +1,51 @@
...
+  font-style: italic;
+}
+
+// Content fonts
+body {
+  --content-font-family: var(--ff-serif);
+
+  &.serif-article-body {
+    --content-font-family: var(--ff-serif);
+  }
```
ludwiczakpawel, line 34: This class name doesn't exist but in spirit of what @benhalpern said re letting admin set their own "default" font I assumed this class might be needed.

fdoxyz:
Hey @ludwiczakpawel, I tested locally and the default font with a clean install (DB reset) on a newly registered user appears as serif for me, not sure if there's something off from my end though.

The first place I can see where the logic will probably need to change is in `app/views/users/_font_selector.html.erb` because `default` is matched as `serif`. With the custom configuration of the forem instance "default font" in mind we'll probably want to add a selector in the Configuration page. We're expecting to see #9639 merged this morning so it's probably best to wait until then to avoid conflicts.

I'll take a closer later in the day (first time diving in this area of the app for me) unless someone else chips in before I get to loop back here 🙂

ludwiczakpawel:
@fdoxyz We used to have serif font being a default one, but we now want sans-serif font to be the default one. The thing you mentioned in `_font_selector.html.erb` is definitely one place that needs to be fixed but it also needs some database/backend change that I don't know how to implement. So yea, if you wanna take it for a ride, go for it definitely!

ludwiczakpawel:
@fdoxyz any updates? :)

fdoxyz:
Sorry @ludwiczakpawel I got a bit stuck last week with a PR I've been trying to finish off for a while now. It's now waiting for review so I should be able to chip in here, if not today I hope I can push something tomorrow morning without a doubt!

fdoxyz:
@benhalpern @ludwiczakpawel a couple questions here:

1. I heard a concern from @rhymes that makes sense: Creators that choose "comic sans" as default font would have a "negative impact" on their instance (readability). With the current changes all fonts are available to choose as default for admins, do we prefer instead to limit them to only set "serif" or "san serif" as defaults?
1. The default config is only affecting the articles (and other content using the "<font>-article-body" class). Is this what we want or should this span over more throughout the site?

nickytonline:
> ## Description
> This PR fixes an issue which made our fonts picker stop working... Users were not able to choose a specific font in `/settings/ux` for like 1 month, surprisingly there was no reports :D
>
> Additionaly, I've cleaned up a bit of CSS for fonts and improved few things along the way.

@ludwiczakpawel, can this PR be split up into two PRs for the stuff it addresses, i.e. can the CSS clean up be in a separate PR?

spec/decorators/user_decorator_spec.rb
```diff
@@ -105,10 +105,26 @@
     end
   end
 
+  describe "#config_font_name" do
+    it "replaces 'default' with font configured for the site in SiteConfig" do
+      expect(user.config_font).to eq("default")
```
nickytonline, line 110: Shouldn't this be `serif`?

```suggestion
      expect(user.config_font).to eq("serif")
```

fdoxyz: User's `config_font` would still be `default` string in the DB (otherwise we would have to update all the existing users column in the DB).

When requesting the `config_font_name` from the decorator it should render their font of choice or swap out `default` for the chosen default font set by the instance admin. In the second case (user still has `default`) the tests checks against all possible fonts down in following lines of this same spec.

Still in doubt about whether we should allow all fonts though ([question here](https://github.com/forem/forem/pull/9650#issuecomment-673519126))

ludwiczakpawel: > Still in doubt about whether we should allow all fonts though

Nah, I think we should only allow the following:

- serif
- sans-serif
- dyslexia one

Once we have better theming infrastructure we could let ppl choose their own default fonts, but for now we should limit the choices to the three fonts above.

fdoxyz: Done! These will be the fonts available to choose from in `/admin/config`:

Screen Shot 2020-08-13 at 10 56 00

I did found out that comments are not using the selected font... Maybe this is because of the changes in the CSS? Can you take a look and see why is this hapenning @ludwiczakpawel? I'm not sure I follow very well what changed that affected this

ludwiczakpawel: @fdoxyz yea, i skipped it intentionally since I'm working on Comments section anyway, so it will be tackled in a separate PR.

PR #9652, opened 2020-08-06
Implement ImageResizer module

maestromac:
- [x] Refactor
- [x] Feature

## Description
This is to pave the way to other future image resizing solution. There will be part 2 to this PR to extract the rest of `cl_image_path` in the codebase

n/a

Navigate around the site locally and see that images are rendered properly.

## Added tests?
- [x] yes, and updated some existing one

## Added to documentation?
- [x] no documentation needed

n/a

### DIFF
--- a/app/lib/image_resizer.rb
+++ b/app/lib/image_resizer.rb
@@ -1,15 +1,21 @@
 module ImageResizer
-  def self.call(img_src, height:, width:, crop:, quality:, flags:)
-    options =  {
-      type: "fetch",
-      height: height,
-      width: width,
-      sign_url: true,
-      crop: crop || "limit",
-      quality: quality || "auto",
-      fetch_format: "auto",
-      flags: flags || "progressive"
-    }.reject { |_, v| v.blank? }
+  DEFAULT_CL_OPTIONS = {
+    type: "fetch",
+    height: nil,
+    width: nil,
+    crop: "limit",
+    quality: "auto",
+    flags: "progressive",
+    fetch_format: "auto",
+    sign_url: true
+  }.freeze
+
+  def self.call(img_src, **kwargs)
+    options = DEFAULT_CL_OPTIONS.merge(kwargs).reject { |_, v| v.blank? }
+
+    if img_src&.include?(".gif")
+      options[:quality] = 66
+    end
 
     ActionController::Base.helpers.cl_image_path(img_src, options)
   end
--- a/app/lib/image_resizer.rb
+++ b/app/lib/image_resizer.rb
@@ -1,0 +1,16 @@
+module ImageResizer
+  def self.call(img_src, height:, width:, crop:, quality:, flags:)
+    options =  {
+      type: "fetch",
+      height: height,
+      width: width,
+      sign_url: true,
+      crop: crop || "limit",
+      quality: quality || "auto",
+      fetch_format: "auto",
+      flags: flags || "progressive"
+    }.reject { |_, v| v.blank? }
+
+    ActionController::Base.helpers.cl_image_path(img_src, options)
+  end
+end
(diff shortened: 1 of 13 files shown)


app/lib/image_resizer.rb
```diff
@@ -0,0 +1,22 @@
...
+    width: nil,
+    crop: "limit",
+    quality: "auto",
+    flags: "progressive",
+    fetch_format: "auto",
+    sign_url: true
+  }.freeze
+
+  def self.call(img_src, **kwargs)
+    options = DEFAULT_CL_OPTIONS.merge(kwargs).reject { |_, v| v.blank? }
```
citizen428, line 14: Unless you are worried about empty strings, you can replace `.reject { |_, v| v.blank? }` with `compact`:

```ruby
{ width: nil, height: nil, crop: "limit" }.compact
#=> {:crop=>"limit"}
```

But this also raises the question: why include the `nil` options in `DEFAULT_CL_OPTIONS` when we then remove them here? Is it mostly for documentation purposes so people know what they can pass in?

app/lib/image_resizer.rb
```diff
@@ -0,0 +1,22 @@
+module ImageResizer
```
rhymes, line 1: I would imagine this to be in `app/services` :)
Something like `Images::Resize.call`

app/lib/image_resizer.rb
```diff
@@ -0,0 +1,22 @@
...
+    height: nil,
+    width: nil,
+    crop: "limit",
+    quality: "auto",
+    flags: "progressive",
+    fetch_format: "auto",
+    sign_url: true
+  }.freeze
+
+  def self.call(img_src, **kwargs)
```
rhymes, line 13: should we make it explicit that this requires an `image_url` ? Otherwise one might be tempted to pass a path to an asset as well

app/lib/image_resizer.rb
maestromac: Thanks for this suggestion! I'm going to do this on the next PR

app/lib/image_resizer.rb
maestromac: Oh, I did not consider this. I will make it raise an exception.

app/lib/image_resizer.rb
maestromac: I haven't thoroughly tested this out but based on Cloudinary's documentation, depending on the `crop` option given, you don't need to specify both width and height. I'm definitely not worried about empty strings so I will change it to `compact`.

app/lib/image_resizer.rb
citizen428: @maestromac We recently started adding more models that aren't backed by `ActiveRecord` to `app/models`. This often raises the question of where something should go. Something that served me well previously is following a  separation between actions (verbs) and things (nouns). So the question becomes does this object represent an action or a thing? In the first case, I'd put it in `app/services/` and call it `ResizeImage` (sometimes referred to as the function object pattern) in the second case I'd put it in `app/models` and keep the current name. In this specific instance, I'd probably go for a "verb" in services.

You don't have to follow this of course, but thinking about this way usually helps me clarify where I want to put something.

maestromac: Thanks you for that @citizen428 !

I've always been a bit confused between what to put in services vs models. There is a plan to implement an `Image` model so my plan is to eventually move it to `app/models` but you just gave me a new perspective. This is an action and should remain `services`.

citizen428: You're welcome @maestromac 😃 If you already know you are going to add an `Image` model, consider namespacing the actions in `app/services`, e.g. `Images::Resize`

PR #9655, opened 2020-08-06
Add X-Accel-Expires header to support Nginx caching

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

**PLEASE READ UPDATED DESCRIPTION AT THE BOTTOM OF THIS PR**

~In https://github.com/forem/forem/pull/9631, we conditionally removed the `no-cache` option from the `Cache-Control` headers that are sent along to Fastly + Nginx via our `CachingHeaders` concern, as a way of experimenting to see if Fastly still caches the same way. We did some experiments, and saw that Fastly still picked up on our change, and still caches the way that we'd expect.~

~You can play with our [Fastly fiddle snippet](https://fiddle.fastlydemo.net/fiddle/62626efe) if you'd like, but the TL;DR can be summed up by these specific headers/response from Fastly:~
```bash
HTTP/2 200

cache-control: max-age=86400, public
x-cache-hits: 0, 1, 1
x-cache: MISS, HIT, HIT
x-served-by: cache-den19627-DEN, cache-wdc5547-WDC, cache-wdc5562-WDC
```

~Now that we've confirmed that this _doesn't_ break Fastly caching, this PR uses the same exactly Cache-Control headers for _both_ Fastly **and** Nginx:~
```ruby
response.headers["Cache-Control"] = "public, max-age=#{max_age}"
```

Once https://github.com/forem/lfss/pull/1 is merged, we expect that [forem.dev](https://forem.dev/) should Just Work™️  with Nginx caching enabled.

Make sure the tests pass! If you are set up with the LFSS repo, you can pull this branch down and QA it to test out whether caching really _does_ work without the explicitly set headers in the `nginx.conf`, which https://github.com/forem/lfss/pull/1 removes :)

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

We need to keep an eye on this PR when it deploys and watch to make sure we're not seeing a bunch of busted caches; ideally, caching should work as normal and we shouldn't see any impacts.

slowly smiling sloth

### DIFF
--- a/app/controllers/concerns/caching_headers.rb
+++ b/app/controllers/concerns/caching_headers.rb
@@ -13,29 +13,20 @@ module CachingHeaders
   #  {cache_control: 'public, no-cache, maxage=xyz', surrogate_control: 'max-age: 100'}
   def set_cache_control_headers(
     max_age = 1.day.to_i,
-    surrogate_control: nil, stale_while_revalidate: nil, stale_if_error: 26_400
+    surrogate_control: nil,
+    stale_while_revalidate: nil,
+    stale_if_error: 26_400
   )
-
-    # TODO: [@forem/delightful]: Once we've tested that removing `no-cache` from the Cache-Control headers
-    # does not adversely affect fastly, we should remove this specific check entirely.
-    article_requested = request.env["REQUEST_PATH"].present? &&
-      request.env["REQUEST_PATH"].include?("the-3-biggest-misconceptions-about-diversity-3hkm")
-    cache_control = if article_requested
-                      "public, max-age=86400"
-                    else
-                      "public, no-cache"
-                    end
-
     request.session_options[:skip] = true # no cookies
-    response.headers["Cache-Control"] = cache_control
 
+    response.headers["Cache-Control"] = "public, max-age=#{max_age}"
     response.headers["Surrogate-Control"] = surrogate_control.presence || build_surrogate_control(
       max_age, stale_while_revalidate: stale_while_revalidate, stale_if_error: stale_if_error
     )
   end
 
   # Sets Surrogate-Key HTTP header with one or more keys strips session data
-  # from the request
+  # from the request.
   def set_surrogate_key_header(*surrogate_keys)
     request.session_options[:skip] = true # No Set-Cookie
     response.headers["Surrogate-Key"] = surrogate_keys.join(" ")
--- a/app/controllers/concerns/caching_headers.rb
+++ b/app/controllers/concerns/caching_headers.rb
@@ -19,7 +19,8 @@ module CachingHeaders
   )
     request.session_options[:skip] = true # no cookies
 
-    response.headers["Cache-Control"] = "public, max-age=#{max_age}"
+    response.headers["Cache-Control"] = "public, no-cache"
+    response.headers["X-Accel-Expires"] = max_age.to_s
     response.headers["Surrogate-Control"] = surrogate_control.presence || build_surrogate_control(
       max_age, stale_while_revalidate: stale_while_revalidate, stale_if_error: stale_if_error
     )
--- a/app/controllers/concerns/caching_headers.rb
+++ b/app/controllers/concerns/caching_headers.rb
@@ -19,8 +19,8 @@ module CachingHeaders
   )
     request.session_options[:skip] = true # no cookies
 
-    response.headers["Cache-Control"] = "public, no-cache"
-    response.headers["X-Accel-Expires"] = max_age.to_s
+    response.headers["Cache-Control"] = "public, no-cache" # Used only by Fastly.
+    response.headers["X-Accel-Expires"] = max_age.to_s # Used only by Nginx.
     response.headers["Surrogate-Control"] = surrogate_control.presence || build_surrogate_control(
       max_age, stale_while_revalidate: stale_while_revalidate, stale_if_error: stale_if_error
     )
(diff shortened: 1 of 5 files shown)


spec/support/initializers/capybara.rb
```diff
@@ -2,7 +2,7 @@
 require "capybara/rspec"
 require "webdrivers/chromedriver"
 
-Webdrivers.cache_time = 86_400
+Webdrivers.cache_time = 0
```
vaidehijoshi, line 5: i actually don't think that this does anything 😩 according to [these docs](https://github.com/titusfortner/webdrivers#caching-drivers), i think it's actually how we tell the webdrivers to update itself/look for updates?

rhymes: Yeah, that's just about self update

benhalpern:
Do we want to update `Cache-Control` here? Is this not the cache that speaks to the browser and we _wouldn't_ typically want to cache non-asset pages in the browser, right?

Like, if I go to `forem.dev/ben` I want the `200` version, not the `304` version, because something might have changed. We want it cached _in nginx_ so it can be purged, but not necessarily in the browser.

Am I thinking about that the right way? And if so could that be the issue with the tests? That these results are being cached between tests so they don't change when they should? And therefor the tests are kind of doing their jobs?

I might be off base but let me know!

atsmith813:
@benhalpern how would you verify the browser is caching it or not when making these changes?

vaidehijoshi:
After some investigation, we've determined that using the `Cache-Control` header for Nginx caching was causing our browser/capybara tests to fail repeatedly. Interestingly, however, we could not actually replicate the "cached browser" behavior that we were seeing in the _tests_ within _production_. But, the fact that the tests were failing were an indication that the `Cache-Control` header was having _some_ kind of impact.

This PR is now taking a different approach: it uses the `X-Accel-Expires` header to tell Nginx how _long_ to cache something for (similar concept to the `max-age` directive that Fastly passes along within _its_ `Cache-Control` headers).

According to [this](https://medium.com/@leshchuk/http-cache-on-rails-nginx-stack-950fee2f8eef) post:
> `X-Accel-Expires` is custom header respected by nginx cache and ignored by Rack::Cache.
> Since X-Accel-Expires have priority on Cache-Control we can manage nginx expiration cache from rails, bypassing Rack::Cache.
> If you want to control nginx cache expiration time from rails you can use
X-Accel-Expires header.

So, this PR adds functionality + tests around us sending the `X-Accel-Expires` header along to Nginx:
```ruby
response.headers["X-Accel-Expires"] = max_age.to_s
```

In [this corresponding LFSS PR](https://github.com/forem/lfss/pull/2), we will completely now ignore the `Cache-Control` headers coming from Rails, which will allow only the `X-Accel-Expires` to pass through; this should tell Nginx how _long_ to cache for:

```bash
proxy_ignore_headers Cache-Control;
```

But how will Fastly interpret this change? Well, [this Fastly documentation](https://docs.fastly.com/en/guides/configuring-caching) does not mention anything about respecting the `X-Accel-Expires` header, so adding this header should _not_ affect Fastly caching. The `X-Accel-Expires` is a Nginx-specific thing only, so it should only be adding support for that, and not impact our current Fastly caching infrastructure.


**TO QA THIS PR:** 
If you want to test this PR out, pull down this branch and pull down the changes to the `nginx.conf` file in the [corresponding LFSS branch](https://github.com/forem/lfss/pull/2) branch, and ensure that cache `HIT`s and `PURGE`s work with openresty/nginx configured to cache! 🚀

app/controllers/concerns/caching_headers.rb
```diff
@@ -13,29 +13,21 @@ module CachingHeaders
...
-                      "public, max-age=86400"
-                    else
-                      "public, no-cache"
-                    end
-
     request.session_options[:skip] = true # no cookies
-    response.headers["Cache-Control"] = cache_control
 
+    response.headers["Cache-Control"] = "public, no-cache" # Used only by Fastly.
+    response.headers["X-Accel-Expires"] = max_age.to_s # Used only by Nginx.
```
mstruve, line 23: Nice solution. Question, bc we took the `max-age=86400` out of the cache_control headers for articles will that affect how we are caching them now? They will still expire, correct? Sorry if that is a dumb question 😅

vaidehijoshi: @mstruve Hm, the only article that we experimented on was [that one article](https://github.com/forem/forem/pull/9655/files#diff-09e7aa268f812ebaaa34e1ff19996cebL21-L23) of @atsmith813's right? Otherwise, we didn't touch any other article's headers 🤔 

`max-age` will still get sent through to Fastly [via the Surrogate-Control header](https://github.com/forem/forem/blob/36bb896caf94e48e644b5be55904f66ef9f16f4e/app/controllers/concerns/caching_headers.rb#L39), so I don't think any endpoints should be affected with regards to caching on Fastly

(Let me know if I am understanding your question correctly, I have a feeling I might not be? 🙃)

mstruve: OH right, nvm ignore me. It all just came together in my head

vaidehijoshi: 🙌

PR #9656, opened 2020-08-06
Grant admin permissions through the UI (part 1)

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We've been operating in a scenario where we can only grant permissions through the console, a scenario we obviously want to do away with.

This lets us add roles the same way we add "trusted", etc. It is only visible and available to super admins.

Screen Shot 2020-08-06 at 6 00 45 PM
Screen Shot 2020-08-06 at 6 00 57 PM

You cannot _remove_ these privileges at the moment, so that will be a follow-on PR. If anything is broken or incorrectly implemented I think we should block the PR. If it's a missing thing, I can bundle it with the next PR. 😄


app/services/moderator/manage_activity_and_roles.rb
```diff
@@ -70,6 +70,18 @@ def handle_user_status(role, note)
...
+        remove_negative_roles
+        user.add_role :admin
+      when "Super Admin"
+        check_super_admin
+        remove_negative_roles
+        user.add_role :super_admin
+      when /^(Resource Admin: )/
+        check_super_admin
+        remove_negative_roles
+        user.add_role(:single_resource_admin, role.split("Resource Admin: ").last.safe_constantize.new.class)
```
citizen428, line 84: After `safe_constantize` you already have the class, no need for `.new.class`:

```ruby
"User".safe_constantize == "User".safe_constantize.new.class
#=> true
```

```suggestion
        user.add_role(:single_resource_admin, role.split("Resource Admin: ").last.safe_constantize)
```

app/views/internal/users/edit.html.erb
```diff
@@ -34,7 +34,9 @@
     <%= form_for(@user, url: user_status_internal_user_path(@user), html: { class: "mt-3" }) do |f| %>
       <div class="form-group">
         <%= f.label "Select new user status", class: "mr-3" %>
-        <%= f.select(:user_status, options_for_select(["Warn", "Comment Ban", "Ban", "Regular Member", "Trusted", "Pro"]), include_blank: true) %>
+        <% roll_select = ["Warn", "Comment Ban", "Ban", "Regular Member", "Trusted", "Pro"] %>
```
citizen428, line 37: How about moving this into `app/lib/constants.rb` as `ROLES` and `SPECIAL_ROLES`, then changing this to 

```ruby
roll_select = current_user.has_role?(:super_admin) ? ROLES + SPECIAL_ROLES : ROLES
```

benhalpern: Good call

app/lib/constants.rb
```diff
@@ -258,4 +258,21 @@ module Constants
...
     # Dynamic values ommitted: configurable_rate_limits and social_media_handles
   }.freeze
+
+  BASE_ROLES = ["Warn",
+                "Comment Ban",
+                "Ban",
+                "Regular Member",
+                "Trusted", "Pro"].freeze
+
+  SPECIAL_ROLES = ["──────────",
```
citizen428, line 268: **nitpick (non-blocking):** Seeing this makes me think that maybe instead of adding a separator manually we could use [`grouped_options_for_select`](http://api.rubyonrails.org/classes/ActionView/Helpers/FormOptionsHelper.html#method-i-grouped_options_for_select). This would mean the following change in the view:

```ruby
<% options = { "Base Roles" => Constants::BASE_ROLES %>
<% options["Special Roles] = Constants::SPECIAL_ROLES if current_user.has_role?(:super_admin) %>
<%= f.select(:user_status, grouped_options_for_select(options), include_blank: true) %>
```

spec/requests/admin/users_manage_spec.rb
```diff
@@ -137,6 +137,28 @@ def call_ghost
...
+    it "does not allow non-super-admin to doll out admin" do
+      super_admin.remove_role(:super_admin)
+      super_admin.add_role(:super_admin)
+      super_admin.reload
+
+      params = { user: { user_status: "Super Admin", note_for_current_role: "they deserve it for some reason" } }
+      patch "/admin/users/#{user.id}/user_status", params: params
+
+      expect(user.has_role?(:super_admin)).not_to be false
+    end
```
Zhao-Andy, line 160: Sorry, what's going on in this test? Are you trying to test that a super admin cannot give a user `super_admin`? Or are you trying to test that a non-super admin cannot give the admin role -- which I think _should_ be allowed, right?

app/lib/constants.rb
```diff
@@ -0,0 +1,277 @@
+module Constants
+  ALLOWED_MASTODON_INSTANCES = [
```
Zhao-Andy, line 2: So we have a file similar to this `app/lib/constants/mastodon.rb` now, that I think was added after you made this PR. We can probably remove this constant, and the `SiteConfig` constant.

app/lib/constants.rb
```diff
@@ -0,0 +1,277 @@
...
+    "social.targaryen.house",
+    "social.tchncs.de",
+    "switter.at",
+    "todon.nl",
+    "toot.cafe",
+    "wikitetas.club",
+    "xoxo.zone",
+  ].freeze
+
+  SITE_CONFIG_DETAILS = {
```
Zhao-Andy, line 73: And same with this.

app/lib/constants.rb
```diff
@@ -0,0 +1,277 @@
...
+      placeholder: "List of valid tags: comma separated, letters only e.g. help,discuss,explainlikeimfive,meta"
+    },
+    feed_style: {
+      description: "Determines which default feed the users sees (rich content, more minimal, etc.)",
+      placeholder: "basic, rich, or compact"
+    }
+    # Dynamic values ommitted: configurable_rate_limits and social_media_handles
+  }.freeze
+
+  BASE_ROLES = ["Warn",
```
Zhao-Andy, line 262: Do we want to follow the (new) pattern we have of having it in a namespaced file?

spec/requests/admin/users_manage_spec.rb
```diff
@@ -137,6 +137,28 @@ def call_ghost
       expect(user.roles.last.name).to eq("comment_banned")
     end
 
+    it "selects super admin role when user was banned" do
```
Zhao-Andy, line 140: _Nitpick -- non-blocking:_ I think maybe being more specific here and say:
```
it "grants super admin role even if the user was banned"
```

app/views/admin/users/edit.html.erb
```diff
@@ -34,7 +34,9 @@
     <%= form_for(@user, url: user_status_admin_user_path(@user), html: { class: "mt-3" }) do |f| %>
       <div class="form-group">
         <%= f.label "Select new user status", class: "mr-3" %>
-        <%= f.select(:user_status, options_for_select(["Warn", "Comment Ban", "Ban", "Regular Member", "Trusted", "Pro"]), include_blank: true) %>
+        <% options = { "Base Roles" => Constants::BASE_ROLES } %>
+        <% options["Special Roles"] = Constants::SPECIAL_ROLES if current_user.has_role?(:super_admin) %>
+        <%= f.select(:user_status, grouped_options_for_select(options), include_blank: true) %>
```
Zhao-Andy, line 39: I think if we could make these roles look prettier here that would be nice too. For example:
Screen Shot 2020-08-18 at 10 52 24 AM

The "Current Roles" don't distinguish which `single_resource_admin` roles they are. That said they do in `/admin/permissions` but I feel like it's more helpful in this single user view.

app/controllers/admin/configs_controller.rb
```diff
@@ -53,6 +53,7 @@ def config_params
         github_secret
         facebook_key
         facebook_secret
+        allow_email_password_registration
```
mstruve, line 56: Did you intend for these email registration changes to be in this PR?

mstruve:
Looks like a few of these changes belong to https://github.com/forem/forem/pull/10039 😅  should probably remove those

benhalpern:
> Looks like a few of these changes belong to #10039 😅 should probably remove those

Yeah @mstruve I noticed that mistake, my thoughts are to first merge the other PR then deal with this as the simplest remediation.

PR #9662, opened 2020-08-07
Rubocop: Fix Rails/HasManyOrHasOneDependent for Broadcast and Collection

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Follow up to https://github.com/forem/forem/pull/9565

I've decided to split #9405 in multiple PR with a smaller scope/surface.

This PR tackles `Rails/HasManyOrHasOneDependent` for `Broadcast` and `Collection`.

I could neither find any orphaned notification belonging to non existing broadcasts (mostly because we don't delete broadcasts, we just deactivate them) nor I could find any article attached to a collection that doesn't exist (mostly because we don't allow people to delete collections yet ;-)).

Thus I didn't add a data update script

#9405 

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/models/broadcast.rb
+++ b/app/models/broadcast.rb
@@ -2,7 +2,7 @@ class Broadcast < ApplicationRecord
   VALID_BANNER_STYLES = %w[default brand success warning error].freeze
   resourcify
 
-  has_many :notifications, as: :notifiable, inverse_of: :notifiable
+  has_many :notifications, as: :notifiable, inverse_of: :notifiable, dependent: :destroy
 
   validates :title, uniqueness: { scope: :type_of }, presence: true
   validates :type_of, :processed_html, presence: true
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -1,5 +1,6 @@
 class Collection < ApplicationRecord
-  has_many :articles
+  has_many :articles, dependent: :nullify
+
   belongs_to :user
   belongs_to :organization, optional: true
 
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -24,8 +24,6 @@ Performance/OpenStruct:
 Rails/HasManyOrHasOneDependent:
   Exclude:
     - 'app/models/article.rb'
-    - 'app/models/broadcast.rb'
-    - 'app/models/collection.rb'
     - 'app/models/concerns/user_subscription_sourceable.rb'
     - 'app/models/display_ad.rb'
     - 'app/models/html_variant.rb'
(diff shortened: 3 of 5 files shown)


rhymes:
One build succeded, don't know two were forked

rhymes:
Closing and reopening to see if we can tell Travis to fork only one build

app/models/broadcast.rb
```diff
@@ -2,7 +2,7 @@ class Broadcast < ApplicationRecord
   VALID_BANNER_STYLES = %w[default brand success warning error].freeze
   resourcify
 
-  has_many :notifications, as: :notifiable, inverse_of: :notifiable
+  has_many :notifications, as: :notifiable, inverse_of: :notifiable, dependent: :destroy
```
nickytonline, line 5: I read up on this briefly, but this just ensures that associated data (via foreign keys) is destroyed (cleaned up) right? https://stackoverflow.com/a/29560982/77814

app/models/collection.rb
```diff
@@ -1,5 +1,6 @@
 class Collection < ApplicationRecord
-  has_many :articles
+  has_many :articles, dependent: :nullify
```
nickytonline, line 2: And for this one, it just sets associated datas foreign keys to null, correct? https://guides.rubyonrails.org/association_basics.html

app/models/broadcast.rb
rhymes: Exactly!

This means that if we call `broadcast.destroy` all `broadcast.notifications` will be destroyed as well, otherwise they'll be there hanging in the DB and invalid

app/models/collection.rb
rhymes: Yep! It means that if we call `collection.destroy` the `collection_id` values in `collection.articles` will be set to null

app/models/broadcast.rb
vaidehijoshi, line 5: The only thing I wonder is whether we should alert admins right before they delete a `Broadcast` and let them know [in the delete method of the controller](https://github.com/forem/forem/blob/master/app/controllers/admin/broadcasts_controller.rb#L49-L59) that the action they're about to take will also delete all associated `Notifications`? I can totally see admins playing with this feature and not necessarily knowing that deleting one model destroys the others 🤷

rhymes: We can replace this with `:restrict_with_error` to make sure notifications are explicitly destroyed before ? Is it possible to destroy notifications from the admin panel?

vaidehijoshi: @rhymes AFAIK, you can't destroy notifications from the admin panel (?), but I do like the idea of us either forcing the admin to explicitly destroy them (somehow) or alerting them that they will be destroyed as a result of the broadcast being deleted.

rhymes: @vaidehijoshi should we make an issue for option 2? Maybe "destroy broadcast" can have a confirmation step where the user is notified that N notifications are going to be deleted and then they can decide whether to confirm or bail?

PR #9667, opened 2020-08-07
Make read next area work for smaller communities

benhalpern:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently the "Read Next" area mostly doesn't work for naissant communities. This should make the article suggestion more viable for these situations. 😄

### DIFF
--- a/app/labor/article_suggester.rb
+++ b/app/labor/article_suggester.rb
@@ -50,7 +50,7 @@ class ArticleSuggester
       .first(max)
   end
 
-  def calculate_offset
+  def offset
     @total_articles_count > 1000 ? 200 : (@total_articles_count / 10)
   end
 
--- a/app/labor/article_suggester.rb
+++ b/app/labor/article_suggester.rb
@@ -1,6 +1,7 @@
 class ArticleSuggester
   def initialize(article)
     @article = article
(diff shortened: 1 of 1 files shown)


app/labor/article_suggester.rb
```diff
@@ -39,14 +44,20 @@ def other_suggestions(max: 4, ids_to_ignore: [])
...
       .first(max)
   end
 
   def offset
-    Rails.env.production? ? 200 : 0
+    @total_articles_count > 1000 ? 200 : (@total_articles_count / 10)
+  end
+
+  def tag_suggestion_query
+    # Fore big communities like DEV we can look at organic page views for indicator.
```
joshpuetz, line 58: Perhaps avoid mention of DEV specifically:

```suggestion
    # For communities with a large number of existing articles we can look at organic page views for indicator.
```

benhalpern: I thought DEV actually was a useful reference as literally the only community that will fall under this umbrella for a while.

I feel like this implementation is naive enough that it should probably be revisited again as we have more active communities in different situations and configs.

PR #9670, opened 2020-08-07
Update docs with details on where to install Elasticsearch

vaidehijoshi:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

I recently ran into some weird elasticsearch issues and fixed them by uninstalling ES and then re-installing. Once I did that, Git picked up on that as a diff:
```bash
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        elasticsearch-7.5.2/
```

However, I don't think we want to commit large ES installation files to the repo, so it made sense to me to add it to `.gitignore`, so that massive tar files don't accidentally sneak into a PR or commit.

N/A

Make sure that installing ES locally doesn't cause a `git diff` 😉 

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,9 @@ latest.dump
 yarn-debug.log*
 .yarn-integrity
 
+# Ignore elasticsearch installations
+/elasticsearch-7.5.2
+
 # Ignore storybook static site generation
 app/javascript/storybook-static/
 yarn-error.log
--- a/.gitignore
+++ b/.gitignore
@@ -42,9 +42,6 @@ latest.dump
 yarn-debug.log*
 .yarn-integrity
 
-# Ignore elasticsearch installations
-/elasticsearch-7.5.2
-
 # Ignore storybook static site generation
 app/javascript/storybook-static/
 yarn-error.log
--- a/docs/installation/mac.md
+++ b/docs/installation/mac.md
@@ -70,9 +70,9 @@ archive. We recommend installing from archive on Mac.
 ### Installing Elasticsearch from the archive
 
 We recommend that you **do not** install Elasticsearch in the app directory.
-Instead, we recommend installing it at the local root level directory. (This
-also ensures that we don't accidentally commit Elasticsearch code to the
-project's repository!)
+Instead, we recommend installing it in your home directory (for example,
+`cd $HOME`). (This also ensures that we don't accidentally commit Elasticsearch
+code to the project's repository!)
 
 The following directions were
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -42,6 +42,9 @@ latest.dump
 yarn-debug.log*
 .yarn-integrity
 
+# Ignore elasticsearch installations
+/elasticsearch-7.5.2
```
maestromac, line 46: What about ignoring _all_ future versions too?

```suggestion
/elasticsearch*
```

vaidehijoshi: @maestromac ah, great call!

benhalpern:
I'd probably wait on a sanity check from @mstruve

vaidehijoshi:
> We definitely don't want elasticsearch code to be committed but it should never be in the main application directory in the first place. I worry that adding it here might be misleading that it should be in the main application directory when it shouldn't. 

@mstruve yeah, that was something that @rhymes brought up separately to me (and makes sense). I think maybe we ought to update the documentation for our elasticsearch _docs_ to explicitly say that it should be installed at your local root level (rather than installing it in your main app directory, which I know both me and others have done!). Let me know how that sounds instead!

mstruve:
Oh, I like that documentation update idea much better!!! Feel free to ping me in the PR when it's ready.

vaidehijoshi:
UPDATE: I've gone ahead and updated our documentation to explain _where_ to install Elasticsearch, and removed it from our `.gitignore`, as per @mstruve's suggestion.

docs/installation/mac.md
```diff
@@ -69,6 +69,11 @@ archive. We recommend installing from archive on Mac.
 
 ### Installing Elasticsearch from the archive
 
+We recommend that you **do not** install Elasticsearch in the app directory.
+Instead, we recommend installing it at the local root level directory. (This
+also ensures that we don't accidentally commit Elasticsearch code to the
+project's repository!)
+
```
rhymes, line 76: A small detail, what's root here? because `cd /` is a read only folder in macOS. My suggestion is to change this to "your home directory, for example `cd $HOME`" or something like that

vaidehijoshi: That's a great point, thanks for bringing that up! Latest commit takes your suggestion into account 👍

PR #9677, opened 2020-08-07
API: retrieve podcast episodes comments

### LINKED ISSUE #6526: API: allow retrieval of comments belonging to a podcast episode

Right now the API only returns all comments belonging to an article, but users can also leave comments on podcast episodes.

Enable users to retrieve podcast episode's comments by adding a parameter to `/api/comments` that would indicate the user is requesting the comments of a podcast episode. Someting like `/api/comments?p_id=123`

`/api/podcast_episodes/{id}/comments` but it would look odd in the current v0 

**Additional context**

The controller `Api::V0::CommentsController` and its action `index` are the place to start. Need to make sure that the surrogate key is distinct from that of articles and that the episodes are correctly purged when updated.

No need to alter the `show` action as it's already possible to retrieve a single comment, even if it belongs to a podcast episode

jkrsn98:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Modified the index in comments controller which previously only handled look for a_id parameter, to having it look for a_id or p_id parameter which allow for accessing podcast episode comments. If a_id key has value, it looks for article with that id, if a_id has no value, then it must be that user is trying to query with p_id, so it looks for podcast episode.

Fixes #6526 

Currently you can only request comments related to articles with parameter a_id
requesting with a id

With the line I modified it now uses parameter p_id as well, which will return all the comments related to the podcast episode with that id, same behaviour as was for articles
For example, with the following podcast episode with id 1, I comment on it then try to query with that id, and it works as follows
requesting with p id

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [x] docs.forem.com
- [ ] readme
- [ ] no documentation needed


app/controllers/api/v0/comments_controller.rb
```diff
@@ -9,7 +9,7 @@ class CommentsController < ApiController
       private_constant :ATTRIBUTES_FOR_SERIALIZATION
 
       def index
-        article = Article.find(params[:a_id])
+        article = params[:a_id] ? Article.find(params[:a_id]) : PodcastEpisode.find(params[:p_id])
```
rhymes, line 12: Small nitpick: can we rename the variable to `commentable`?

jkrsn98: Hi @rhymes, I have changed variable name to commentable and tried to add a test, but I am not sure if it is as in depth as you would like (I have never written one before, so mostly just copied what was with article) only I don't know if it should have all the same checks for comments as article does, e.g child comments, grandchild comments, etc... I assume since the search by article id and search by podcast ep id use the same code it doesn't matter practically speaking, but yes with respect to the test I am not sure

jkrsn98:
I tried to do the rebase but I don't think this is it.. :pensive:

rhymes:
Thanks @jkrsn98, I'll let you know when it's live :)

rhymes:
@jkrsn98 it is live :)

PR #9681, opened 2020-08-08
Add new support admin role

Zhao-Andy:
- [x] Feature

## Description
This adds a new support admin role to handle a specific DEV use-case. Ideally we have a better solution for this in the context of Forem, but for now this should do.

This allows anyone with the `support_admin` and `single_resource_admin, User` roles to banish users.

## Added tests?
- [x] no, because they aren't needed. This is a quick fix that should be replaced by a longer term solution.

## Added to documentation?
- [x] no documentation needed

### DIFF
--- a/spec/models/role_spec.rb
+++ b/spec/models/role_spec.rb
@@ -9,7 +9,7 @@ RSpec.describe Role, type: :model do
     it "contains the correct values" do
       expected_roles = %w[
         admin banned chatroom_beta_tester codeland_admin comment_banned podcast_admin pro restricted_liquid_tag
-        single_resource_admin super_admin tag_moderator mod_relations_admin tech_admin support_admin trusted
+        single_resource_admin super_admin tag_moderator mod_relations_admin support_admin tech_admin trusted
         warned workshop_pass
       ]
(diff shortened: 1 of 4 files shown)


spec/models/role_spec.rb
```diff
@@ -9,7 +9,7 @@
     it "contains the correct values" do
       expected_roles = %w[
         admin banned chatroom_beta_tester codeland_admin comment_banned podcast_admin pro restricted_liquid_tag
-        single_resource_admin super_admin tag_moderator mod_relations_admin tech_admin trusted
+        single_resource_admin super_admin tag_moderator mod_relations_admin tech_admin support_admin trusted
```
juliannatetreault, line 12: I took a look at the Travis failure and recreated it locally and what I found was that the spec is failing due to the order of this array--if you place`support_admin` before `tech_admin` to mirror the array in `role.rb`, this will pass!:
```
expected_roles = %w[
        admin banned chatroom_beta_tester codeland_admin comment_banned podcast_admin pro restricted_liquid_tag
        single_resource_admin super_admin tag_moderator mod_relations_admin support_admin tech_admin trusted
        warned workshop_pass
      ]
```

PR #8311, opened 2020-06-06
Added Copy Permalink Functionality to Comment Dropdown

nizarmah:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR adds a *Copy Permalink* button to the dropdown list.

Copy functionality is similar to the way 'copy article link' is implemented. It checks for the android bridge first, then the navigator, then fallsback to the default document copy command.

Since the codes were really similar, I moved the copy functionality to a function of its own with callbacks and fallbacks as lambdas, and triggered it with the copyArticleLink and the copyPermalink.

This PR is needed for the feature request on the [DEV-Android](https://github.com/thepracticaldev/DEV-Android/issues/66). Two suggestions were proposed, with the help of @fdoxyz:
- Copying the Permalink instead of redirecting when pressing the *Permalink* anchor tag.
- Creating a *Copy Permalink*.

And here we are :smile: 

## Related Tickets & Documents
thepracticaldev/DEV-Android#66

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
image
image

## Added tests?

- [ ] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

## What gif best describes this PR or how it makes you feel?

hell yeaaaaaah

PR #9688, opened 2020-08-09
Add copy code functionality to code block (WIP)

### LINKED ISSUE #4686: I'd love a [copy code] button on code blocks

emgoto:
This is a WIP PR!

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

What this PR does:

* Adds a "Copy" button to the inside of each code block. Clicking this will copy the code block to your clipboard

What this PR does not yet do:

* Doesn't use the `<clipboard-copy>` web component (more explanation below)
* No designs - it's just a plain-text copy button
* No indication that the copy button has been clicked (should there be one / do you have any thoughts on what it should look like?)
* The code inside of `copyCode.js` is mostly a copy-paste of some of whats inside of `initializeCommentDropdown.js`. There are some differences (e.g. the class names it will copy the value from) but let me know if you think it's worth refactoring so that the code is shared between the two (and if so, where that code should live).

I took a look at the way the "copy article link" feature was implemented, and what I noticed was that it uses the `<copy-clipboard>` web component, as well as some copy functionality inside of [initializeCommentDropdown.js](https://github.com/forem/forem/blob/master/app/assets/javascripts/initializers/initializeCommentDropdown.js). I copied this code, and was able to implement the functionality using that. 

However I'm not sure what the `<copy-clipboard>` is doing because it still works fine without me using it (and if I try and use it on its own, it doesn't do anything). Would love some clarification around that, as well as any comments on if the current approach I'm taking is the correct one. Thanks!

Closes #4686 

## Added tests?

- [ ] yes
- [x] no, because I need help

Will hold off on tests until we have confirmed that the current approach is correct

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/javascript/packs/copyCode.js
+++ b/app/javascript/packs/copyCode.js
@@ -17,7 +17,7 @@ function execCopyText() {
 }
 
 function copyText(event) {
-    const inputValue = event.target.nextElementSibling.getElementsByTagName('code')[0].innerText;
+    const inputValue = event.target.parentNode.nextElementSibling.getElementsByTagName('code')[0].innerText;
     if (isNativeAndroidDevice()) {
         AndroidBridge.copyToClipboard(inputValue);
     } else if (isClipboardSupported()) {
@@ -31,8 +31,7 @@ function copyText(event) {
     }
 }
 
-const clipboardCopyElements = document.getElementsByClassName('copy-icon');
-
+const clipboardCopyElements = document.getElementsByClassName('copy-code-icon');
 for (let element of clipboardCopyElements) {
     element.addEventListener('click', copyText);
 }
--- a/app/javascript/packs/copyCode.js
+++ b/app/javascript/packs/copyCode.js
@@ -1,0 +1,38 @@
+function isClipboardSupported() {
+    return (
+      typeof navigator.clipboard !== 'undefined' && navigator.clipboard !== null
+    );
+  }
+
+  function isNativeAndroidDevice() {
+    return (
+      navigator.userAgent === 'DEV-Native-android' &&
+      typeof AndroidBridge !== 'undefined' &&
+      AndroidBridge !== null
+    );
+  }
+
+function execCopyText() {
+    document.execCommand('copy');
+}
+
+function copyText(event) {
+    const inputValue = event.target.nextElementSibling.getElementsByTagName('code')[0].innerText;
+    if (isNativeAndroidDevice()) {
+        AndroidBridge.copyToClipboard(inputValue);
+    } else if (isClipboardSupported()) {
+        navigator.clipboard
+        .writeText(inputValue)
+        .catch(() => {
+            execCopyText();
+        });
+    } else {
+        execCopyText();
+    }
+}
+
+const clipboardCopyElements = document.getElementsByClassName('copy-icon');
+
+for (let element of clipboardCopyElements) {
+    element.addEventListener('click', copyText);
+}
--- a/app/views/articles/_actions.html.erb
+++ b/app/views/articles/_actions.html.erb
@@ -33,7 +33,7 @@
         <clipboard-copy for="article-copy-link-input" aria-live="polite" aria-controls="article-copy-link-announcer" class="dropdown-link-row">
           <div class="flex items-center">
             <input type="text" id="article-copy-link-input" value="<%= article_url(@article) %>" class="crayons-textfield" readonly />
-            <%= inline_svg_tag("copy.svg", aria: true, id: "article-copy-icon", class: "crayons-icon mx-2 shrink-0", title: "Notifications") %>
+            <%= inline_svg_tag("copy.svg", aria: true, id: "article-copy-icon", class: "crayons-icon mx-2 shrink-0", title: "Copy Article") %>
           </div>
           <div id="article-copy-link-announcer" role="alert" class="crayons-notice crayons-notice--success my-2 p-1" hidden>Copied to Clipboard</div>
         </clipboard-copy>
--- a/app/assets/images/clipboard.svg
+++ b/app/assets/images/clipboard.svg
@@ -1,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
+    <path fill="white" d="M7 6V3a1 1 0 011-1h12a1 1 0 011 1v14a1 1 0 01-1 1h-3v3c0 .552-.45 1-1.007 1H4.007A1 1 0 013 21l.003-14c0-.552.45-1 1.007-1H7zm2 0h8v10h2V4H9v2zm-2 5v2h6v-2H7zm0 4v2h6v-2H7z"/>
+</svg>
--- a/app/assets/stylesheets/article-show.scss
+++ b/app/assets/stylesheets/article-show.scss
@@ -739,3 +739,36 @@ article {
   100% {
   }
 }
+
+.copy-code {
+  height: 0;
+  position: relative;
+  z-index: 2;
+}
+
+.copy-code-icon {
+  visibility: hidden;
+  content: url(/assets/clipboard.svg);
+  background: #4f575f;
+  width: var(--su-7);
+  height: var(--su-7);
+  opacity: 0.9;
+  margin-left: auto;
+  padding: var(--su-2);
+  border-bottom-left-radius: var(--radius);
+  border-top-right-radius: var(--radius);
+
+  &:hover {
+    opacity: 1;
+  }
+}
+
+.highlight-wrapper {
+  margin-top: var(--article-rythm);
+}
+
+.highlight-wrapper:hover {
+  .copy-code-icon {
+    visibility: visible;
+  }
+}
--- a/app/assets/stylesheets/article-show.scss
+++ b/app/assets/stylesheets/article-show.scss
@@ -744,6 +744,7 @@ article {
   height: 0;
   position: relative;
   z-index: 2;
+  top: var(--article-rythm);
 }
 
 .copy-code-icon {
@@ -761,10 +762,6 @@ article {
   &:hover {
     opacity: 1;
   }
-}
-
-.highlight-wrapper {
-  margin-top: var(--article-rythm);
 }
 
 .highlight-wrapper:hover {
--- a/app/assets/stylesheets/syntax.scss
+++ b/app/assets/stylesheets/syntax.scss
@@ -29,7 +29,7 @@ div.highlight {
   }
 
   @media (min-width: $breakpoint-m) {
-    padding: var(--su-6);
+    padding: var(--su-7) var(--su-6);
   }
 }
 
--- a/app/assets/stylesheets/views/article.scss
+++ b/app/assets/stylesheets/views/article.scss
@@ -242,9 +242,12 @@
     box-shadow: inset 0 -1px 0 #959da5;
   }
 
-  pre:not(.highlight),
-  div.highlight {
+  pre:not(.highlight) {
     margin: var(--article-rythm) 0;
+  }
+
+  div.highlight {
+    margin: 0 0 var(--article-rythm);
   }
 
   img {
--- a/app/assets/stylesheets/views/article.scss
+++ b/app/assets/stylesheets/views/article.scss
@@ -242,12 +242,9 @@
     box-shadow: inset 0 -1px 0 #959da5;
   }
 
-  pre:not(.highlight) {
-    margin: var(--article-rythm) 0;
-  }
-
+  pre:not(.highlight),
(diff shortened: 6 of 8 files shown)


nickytonline:
@ludwiczakpawel, do we have any design for this?

ludwiczakpawel:
@nickytonline we did not have any design for that, but now we do! :)

image

cc @emgoto let me know if it makes sense and if you need some assistance with frontend.

Some additional thoughts:
- would be nice to have `title=""` attribute so user can see a tooltip when hovering on the button.. even though it's pretty recognizable symbol it's worth adding titles and alts for accessibility reasons... some day we will turn it into actual "tooltip" component :)
- I don't think we need any visual feedback for when user clicks this button. i mean, would be nice maybe to utilize "tooltip" with a label "copied!" but as mentioned above - we don't have tooltips yet.
- when working on this, please make sure it looks good with longer pieces of code (especially those horizontally scorllable)

rhymes:
@ludwiczakpawel 

> I don't think we need any visual feedback for when user clicks this button. i mean, would be nice maybe to utilize "tooltip" with a label "copied!" but as mentioned above - we don't have tooltips yet.

A suggestion: could the icon change color briefly? or background color?

ludwiczakpawel:
@rhymes ok, so here's the ideal state: https://d.pr/v/zvQbnd - if we can get as close as possible to this state, then it would be awesome :) anything in between will be good too :)

nickytonline:
> @rhymes ok, so here's the ideal state: [d.pr/v/zvQbnd](https://d.pr/v/zvQbnd) - if we can get as close as possible to this state, then it would be awesome :) anything in between will be good too :)

Love the animation in the video.

emgoto:
> cc @emgoto let me know if it makes sense and if you need some assistance with frontend.

Hi @ludwiczakpawel  I definitely need some help! Please see the description of my pull request - I'm actually pretty stuck at the moment. I don't know what the purpose of the `<clipboard-copy>` web component is.

ludwiczakpawel:
@emgoto I don't really know much about web components but I think @nickytonline might be right person to ask, Nick?

fdoxyz:
Hi @emgoto! So I did see the `<clipboard-copy>` component a while ago when working with other copy to clipboard features, it was already commented out because it wasn't working back then (I can recall this one #6169 and maybe others).

I think it's not necessary at all and can be removed since it's not actually in use, as you're mentioning it works without the component but not on its own. I couldn't find any reference of it in the source code over a quick search though. Also this other implementation takes advantage of the `AndroidBridge` feature used in the native Android app. I'm sorry I hadn't noticed the component question until now. Thank you for taking the time into tackling this!

I haven't tried it locally with the Android app yet, but unless anyone feels differently I think this approach looks good.

nickytonline:
> Hi @emgoto! So I did see the `<clipboard-copy>` component a while ago when working with other copy to clipboard features, it was already commented out because it wasn't working back then (I can recall this one #6169 and maybe others).
> 
> I think it's not necessary at all and can be removed since it's not actually in use, as you're mentioning it works without the component but not on its own. I couldn't find any reference of it in the source code over a quick search though. Also this other implementation takes advantage of the `AndroidBridge` feature used in the native Android app. I'm sorry I hadn't noticed the component question until now. Thank you for taking the time into tackling this!
> 
> I haven't tried it locally with the Android app yet, but unless anyone feels differently I think this approach looks good.

It's used internally in the `<ClipboardButton />` which is used in the `<ImageUploader />` component and in a few other places, but the web component never gets loaded. It looks like importing the clipboard-copy web component was commented out 5 months ago by @benhalpern.

```javascript
// import('@github/clipboard-copy-element');
// Temporarily removed due to problems
```

See https://github.com/forem/forem/blob/2c54add4b853eb9414c672b2506f77cf66c969bc/app/javascript/packs/clipboardCopy.js#L24

fdoxyz:
Hi @emgoto just checking in again here, a small update on the method `isNativeAndroidDevice()` was merged recently but shouldn't make much of a difference, it would now need to call `Runtime.isNativeAndroidDevice()` instead.

I'm confident we can get rid of the `<clipboard-copy>` element as it's not serving any purpose (replace it with normal div if necessary). The approach looks good and it's a great contribution that can be reused in other areas of the app. Let me know if you need any further help on this

emgoto:
Hi @fdoxyz thanks for checking in! I've been a bit busy lately but will hopefully be able to pick up this PR again soon.

emgoto:
The feature isn’t complete yet, but I decided to push what I had done so that I can get any feedback on the work so far (mainly CSS changes):

code-block

What’s left to do:
* The animation and success tick

Other notes/questions:
* Is there a SVG file available for the clipboard icon?
* I’ve found that by adding the SVG via `content` in CSS, the `fill` proper does nothing. I’ve had to manually add `fill="white"` to the SVG file
* I haven’t made the Android change yet, but if we are having an “on hover show the clipboard” approach, would it even be possible to copy code on a mobile since there is no way to hover?
* I’ve had to add an extra bit of padding to the code blocks because otherwise the clipboard icon overlaps with the text
*  When the width of the screen gets small enough the padding on the code blocks disappears. This means the clipboard icon completely overlaps with the text.

Screen Shot 2020-09-13 at 8 40 41 pm

app/views/articles/_actions.html.erb
```diff
@@ -33,7 +33,7 @@
         <clipboard-copy for="article-copy-link-input" aria-live="polite" aria-controls="article-copy-link-announcer" class="dropdown-link-row">
           <div class="flex items-center">
             <input type="text" id="article-copy-link-input" value="<%= article_url(@article) %>" class="crayons-textfield" readonly />
-            <%= inline_svg_tag("copy.svg", aria: true, id: "article-copy-icon", class: "crayons-icon mx-2 shrink-0", title: "Notifications") %>
+            <%= inline_svg_tag("copy.svg", aria: true, id: "article-copy-icon", class: "crayons-icon mx-2 shrink-0", title: "Copy Article") %>
```
emgoto, line 36: I'm pretty sure the title is wrong (since this tooltip shows up when you hover over the "copy article URL" icon). I think it was mistakenly copy-pasted.

ludwiczakpawel:
> Is there a SVG file available for the clipboard icon?

Try this one below. It's not in the codebase yet so you'd have to add it as a new asset `clipboard.svg`

```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24"><path d="M0 0h24v24H0z"/><path d="M7 4V2h10v2h3.007c.548 0 .993.445.993.993v16.014a.994.994 0 01-.993.993H3.993A.994.994 0 013 21.007V4.993C3 4.445 3.445 4 3.993 4H7zm0 2H5v14h14V6h-2v2H7V6zm2-2v2h6V4H9z"/></svg>
```

ludwiczakpawel:
> Is there a SVG file available for the clipboard icon?

Try this one below. It's not in the codebase yet so you'd have to add it as a new asset `clipboard.svg`

```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24"><path d="M0 0h24v24H0z"/><path d="M7 4V2h10v2h3.007c.548 0 .993.445.993.993v16.014a.994.994 0 01-.993.993H3.993A.994.994 0 013 21.007V4.993C3 4.445 3.445 4 3.993 4H7zm0 2H5v14h14V6h-2v2H7V6zm2-2v2h6V4H9z"/></svg>
```

> I’ve found that by adding the SVG via content in CSS, the fill property does nothing. I’ve had to manually add fill="white" to the SVG file

We prefer to include SVG icons right in the markup, as `<%= include_svg_tag( ..., class: "crayons-icon" ) %>` instead of adding them through CSS. If you do that, the icon will inherit its parent color.

> I haven’t made the Android change yet, but if we are having an “on hover show the clipboard” approach, would it even be possible to copy code on a mobile since there is no way to hover?

I think we can drop the functionality for mobile - it's very unlikely anyone would ever wanted to copy any code while browsing Forem on mobile (it's just doesn't make much sense). In other words: you can keep it as it is, without supporting mobile.

> When the width of the screen gets small enough the padding on the code blocks disappears. This means the clipboard icon completely overlaps with the text.

If it's only on hover, then that's alright. At least it's good enough to be shipped. If we ever find it annoying, we can revisit.

emgoto:
> We prefer to include SVG icons right in the markup, as <%= include_svg_tag( ..., class: "crayons-icon" ) %> instead of adding them through CSS. If you do that, the icon will inherit its parent color.

Does the `<%= ...` work outside of `html.erb` files? I can add the icon via `copyCode.js` or`markdown_parser.rb`. 

I tried to copy how `application_helper.rb` uses `inline_svg_tag()` but I found that it gave me an "undefined method" error. Do I need to import it somewhere?

ludwiczakpawel:
@emgoto maybe @rhymes could jump in and help answer this one? i can, for sure, say it won't work in .js files - in that case you'd unfortunately have to just add this SVG code inline.

As for markdown_parser.rb file - as mentioned, Rhymes may have better answer here. But I'm wondering if there's any reason why you're building this "copy code" markup functionality outside .html.erb views? I was thinking the markup would be inside .html.erb and rest of "JS magic" could be placed in the JS... But that's just my guess - I'm not an expert therefore I summon @forem/frontend  :D

rhymes:
Regarding .js I defer to @nickytonline - I remember that another contributor was asking how to call Ruby code from webpack based code but I don't remember exactly. IIRC there's no easy way for have them communicate.

Regarding how to generate a SVG tag in Markdown Parser, you should be able to call `inline_svg_tag` in there by including the `InlineSvg::ActionView::Helpers` module, see https://github.com/jamesmartin/inline_svg/blob/23dcb648f697833649520435d080f0e883de2744/lib/inline_svg/action_view/helpers.rb#L7

As to what's the best strategy, keeping in context also #8311 which is trying to build the copy function for comment dropdowns, I think @fdoxyz is going to weigh in as to the best way to do it

app/javascript/packs/copyCode.js
```diff
@@ -0,0 +1,37 @@
...
+        AndroidBridge.copyToClipboard(inputValue);
+    } else if (isClipboardSupported()) {
+        navigator.clipboard
+        .writeText(inputValue)
+        .catch(() => {
+            execCopyText();
+        });
+    } else {
+        execCopyText();
+    }
```
fdoxyz, line 31: The alternative when the Runtime PR is merged would be:

```suggestion
    Runtime.copyToClipboard(inputValue);
```

Also all three `isClipboardSupported`, `isNativeAndroidDevice` and `execCopyText` could be removed too.

fdoxyz (commented):
@emgoto taking in consideration this and other PRs (as mentioned by @rhymes to be kept in context) I just sent in a PR to centralize the copy to clipboard code (#10323). I think this will simplify the JS implementation and I also hope it will help unblock you and other contributors 🙂

I left the suggestion inline, though you'll need to get your branch up to date once that one is merged for it to be available.

-------------

Regarding the mobile support for this feature (cc @ludwiczakpawel): 

If we're showing the SVG on `hover` we can maybe make it always visible (without hover) for smaller screens with a media selector? I agree this is a feature not commonly used on mobile, but it would work as long as the button is visible on touch screen devices (thinking tablets even).

I don't feel strongly about this, so if we drop it for those platforms I'm okay with it too.

nickytonline:
I'm going to look at @fdoxyz's #10323 first.

app/javascript/packs/copyCode.js
nickytonline: @emgoto. @fdoxyz's PR has been merged.

ludwiczakpawel:
> If we're showing the SVG on hover we can maybe make it always visible (without hover) for smaller screens with a media selector? I agree this is a feature not commonly used on mobile, but it would work as long as the button is visible on touch screen devices (thinking tablets even).

I'm still not sold whether we need this functionality at all on mobile devices, but if you feel strongly about it, I'd recommend using this [nifty media query](https://css-tricks.com/solving-sticky-hover-states-with-media-hover-hover/) to keep things clean:

```
@media (hover: hover) {
    ...
}
```

emgoto:
So sorry, I haven't had the time to get back to this one!

I can see there was a fullscreen mode recently added in #10784 which I think will change the approach / design of this one a little bit. I'm going to close this WIP PR off to allow someone else to pick this one up.

ludwiczakpawel:
@emgoto We haven't really changed the design too much, in fact, in this PR we worked having "copy code" functionality in mind and there are updated designs that do not deviate from our original idea :). So if you still wanna work on this, go for it :)

image

PR #9709, opened 2020-08-10
Enable Rubocop 0.89 cops and fix Lint/ violations

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Rubocop 0.89 [added a bunch of cops](https://github.com/forem/forem/pull/9699), here we enable and correct all `Lint/` violations. I'll send separate additional PRs for `Style/` issues as this already contains many files.

https://github.com/forem/forem/pull/9699

### DIFF
--- a/app/liquid_tags/null_tag.rb
+++ b/app/liquid_tags/null_tag.rb
@@ -1,5 +1,5 @@
 class NullTag < Liquid::Block
-  def initialize(tag_name, _markup, _options)
+  def initialize(tag_name, _markup, _options) # rubocop:disable Lint/MissingSuper
     raise StandardError, "Liquid##{tag_name} tag is disabled"
   end
 end
--- a/app/services/moderator/merge_user.rb
+++ b/app/services/moderator/merge_user.rb
@@ -6,7 +6,7 @@ module Moderator
 
     attr_reader :keep_user, :admin, :delete_user_id
 
-    def initialize(admin:, keep_user:, delete_user_id:)
+    def initialize(admin:, keep_user:, delete_user_id:) # rubocop:disable Lint/MissingSuper
       @keep_user = keep_user
       @admin = admin
       @delete_user = User.find(delete_user_id.to_i)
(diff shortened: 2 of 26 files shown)


app/services/moderator/merge_user.rb
```diff
@@ -6,7 +6,7 @@ def self.call(admin:, keep_user:, delete_user_id:)
 
     attr_reader :keep_user, :admin, :delete_user_id
 
-    def initialize(admin:, keep_user:, delete_user_id:)
+    def initialize(admin:, keep_user:, delete_user_id:) # rubocop:disable Lint/MissingSuper
```
nickytonline, line 9: I'm sure there is a good reason, but I don't know why the rule is disable here and in a bunch of other places in the PR.

app/liquid_tags/null_tag.rb
```diff
@@ -1,5 +1,5 @@
 class NullTag < Liquid::Block
-  def initialize(tag_name, _markup, _options)
+  def initialize(tag_name, _markup, _options) # rubocop:disable Lint/MissingSuper
```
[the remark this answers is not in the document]
rhymes: Exactly!

app/services/moderator/merge_user.rb
rhymes: The point of this rule is to respect classic OOP inheritance in the construction of objects. You have a base class with an initializer and then when you inherit and "override" the initializer you, in theory, should delegate the construction of the final object to the parent's initializer, adding the new state that's related only to this class's object. But this is not what we're doing it, we're constructing a whole new initializer, thus the need for the rule disable. See how the base class initializer is different from this one

https://github.com/forem/forem/blob/edd0f2c630e062c1c60094b9fd21837943723e09/app/services/moderator/manage_activity_and_roles.rb#L5-L9

PR #9710, opened 2020-08-10
Cleanup orphaned Ahoy rows in the DB

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR was excised from https://github.com/forem/forem/pull/9636 so that we can clean up the DB from Ahoy orphaned rows before merging https://github.com/forem/forem/pull/9636

See https://github.com/forem/forem/pull/9636#discussion_r467582799

https://github.com/forem/forem/pull/9636


rhymes:
Marking this as draft until we restore the deleted ahoy rows

rhymes:
@Zhao-Andy trying not to phisically delete scripts but keeping them empty as it can possibly mess our logic to decide which scripts have to be run if we delete the file on the disk

Issue #9647, opened 2020-08-06
[Dev.to] - Profile header height too small for infobox


zfett:
**Describe the bug**

On my dev.to profile (@zfett), my infobox on the right of my header has a vertical scroll due to the height. Using Firefox Version 79.0 64-bit. Tested in Edge with the same result.

**To Reproduce**

1. Fill out enough information in your infobox, try filling out your email, work, location, and education.
2. Visit your profile page
3. Height should overflow on Firefox.

**Expected behavior**

.user-profile-header should resize in height with .user-metadata-details to prevent vertical overflow on the default breakpoint.

**Screenshots**

![screenshot](https://imgur.com/JJEKrnX.jpg)

**Desktop (please complete the following information):**

- OS: Windows 10 Home, 10.0.18363 Build 18363
- Browser: Firefox
- Version: 79.0

**Additional context**

This only happens when the page width is at or above the 1216px breakpoint. The minute you resize it below that point and the content reflows, it no longer suffers from this issue.

zfett:
I'd also like to note, I loaded my page in a private window where I was not logged in, and got the same result. It seems to be exclusively a CSS issue.

Amorpheuz:
I would like to work on this issue!

rhymes:
@Amorpheuz sure! I suspect our designers @ludwiczakpawel and @lisasy might have directions here. I might be wrong but having an ever growing header height might be an issue if people start putting lots of text there.

Amorpheuz:
Great! I was thinking of getting their opinions on the same too before I started working on this. Will wait for their guidance!

ludwiczakpawel:
@rhymes @Amorpheuz this is something we're literally looking into at this very moment - we're working on little fixes for profile and this is one of the things on my plate right now. So this time I may just claim this one and assign myself. Is that ok?

Amorpheuz:
That's great, I don't mind if you claim it! I was personally encountering the same issue when I enabled too many things on my profile too 😬. Excited to see the fix for the same. ♥

PR #9712, opened 2020-08-10
Fixed bug in [user-metadata-details], top portion clipped when scrollbar is visible

esnz:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This RP fixes the problem in the user profile page, where the top of the **user-metadata-details** section gets clipped and goes off bounds.

As you can see in the first screenshot, the label **EMAIL** is barely visible and there is no way you can scroll there, this will cause scrolling problems as the height of the **user-metadata-details** grows and obviously more part of it is becomes hidden.

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/app/assets/stylesheets/user-profile-header.scss
+++ b/app/assets/stylesheets/user-profile-header.scss
@@ -231,10 +231,12 @@
         padding: 15px 0px;
       }
       @media screen and (min-width: 1100px) {
+        display: flex;
+        flex-direction: column;
+        justify-content: center;
         position: absolute;
-        top: 45%;
-        transform: translateY(-45%);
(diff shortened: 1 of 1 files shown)


ludwiczakpawel:
This seem like a duplicate of #9647 - and this is something [we're working on already](https://github.com/forem/forem/issues/9647#issuecomment-669786429):

> this is something we're literally looking into at this very moment - we're working on little fixes for profile and this is one of the things on my plate right now.

benhalpern:
Hey I think we can close this because it was taken care of with the recent merge.

PR #9714, opened 2020-08-10
Grab email from all_emails object instead of info object for GitHub

Zhao-Andy:
- [x] Refactor
- [x] Bug Fix

## Description
This PR updates the GitHub authentication to grab the GitHub user's primary email, and relies on GitHub's `all_emails` object from the authentication payload.

I'm not sure if this will fix the issue where some users sign up with GitHub and have a blank email, but it's a start. I'll monitor incoming sign ups via GitHub to see if this helps.

1. Sign up via GitHub
2. see that your account has an email address attached to it

## Added tests?

## Added to documentation?
- [x] no documentation needed


app/services/authentication/providers/github.rb
```diff
@@ -9,9 +9,10 @@ class Github < Provider
 
       def new_user_data
         name = raw_info.name.presence || info.name
+        email = auth_payload.extra.all_emails.detect(&:primary).email
```
rhymes, line 12: I think this doesn't work as `all_emails` is an array of hashes whose `primary` key is a string. BTW as far as I know `.info` always contains the primary email if there is one. 

See the implementation here 

https://github.com/omniauth/omniauth-github/blob/d001ed274276bf50e1e756bf38bce75509aa25ec/lib/omniauth/strategies/github.rb#L58-L61

and

https://github.com/omniauth/omniauth-github/blob/d001ed274276bf50e1e756bf38bce75509aa25ec/lib/omniauth/strategies/github.rb#L50-L52

and 

https://github.com/omniauth/omniauth-github/blob/d001ed274276bf50e1e756bf38bce75509aa25ec/lib/omniauth/strategies/github.rb#L28-L39

Zhao-Andy: Hmm I think it does work because the array is a special array, `Hashie::Array`:
Screen Shot 2020-08-10 at 12 48 15 PM

That said my implementation doesn't cover for `verified` which the omniauth gem does. I think I know the problem now though!

rhymes: Oh interesting! Sorry I grabbed the JSON payload without sending it through Hashie :D

PR #9716, opened 2020-08-10
Redirect /internal to /admin

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I noticed how a bunch of my bookmarks to Blazer queries are broken because they point to `/internal` which is now called `/admin`. Prompted also by @michael-tharrington's post [here on forem.dev](https://forem.dev/michaelmadcat/renaming-things-and-broken-links-48j1) I added redirections for the old `/internal` links.

- Fire up the branch
- point it to http://localhost:3000/internal/blazer which should redirect to http://localhost:3000/admin/blazer or http://localhost:3000/internal which will redirect to http://localhost:3000/admin which will redirect to http://localhost:3000/admin/articles :)

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -430,6 +430,11 @@ Style/ClassAndModuleChildren:
   Enabled: true
   SafeAutoCorrect: true
 
+Style/FormatStringToken:
+  Description: 'Use a consistent style for format string tokens.'
+  Exclude:
+    - 'config/routes.rb'
+
 Style/OptionalBooleanParameter:
   Description: 'Use keyword arguments when defining method with boolean argument.'
   StyleGuide: '#boolean-keyword-arguments'
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -298,6 +298,8 @@ Rails.application.routes.draw do
   delete "/messages/:id" => "messages#destroy"
   patch "/messages/:id" => "messages#update"
   get "/live/:username" => "twitch_live_streams#show"
+  get "/internal", to: redirect("/admin")
+  get "/internal/:path", to: redirect("/admin/%{path}")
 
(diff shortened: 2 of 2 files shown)


rhymes:
@joshpuetz your points are super valid. I think it's because the support team has bookmarks and procedures tied to the old URLs. In general I abide to the idea of possibly never breaking URLs and as this doesn't cost us much we can keep it. Maybe in the future we can review all redirection URLs (some even point to DEV's posts, which is not going to fly for other Forems).

I remember there was an attempt to add dynamic redirections, probably by @atsmith813?, but in the meantime, I wouldn't break the links on purpose. 

Does it make sense?

joshpuetz:
> @joshpuetz your points are super valid. I think it's because the support team has bookmarks and procedures tied to the old URLs. In general I abide to the idea of possibly never breaking URLs and as this doesn't cost us much we can keep it. Maybe in the future we can review all redirection URLs (some even point to DEV's posts, which is not going to fly for other Forems).
> 
> I remember there was an attempt to add dynamic redirections, probably by @atsmith813?, but in the meantime, I wouldn't break the links on purpose.
> 
> Does it make sense?

Thanks @rhymes, I can certainly appreciate that! My comments aren't blocking, so I'm an approve as well.

If nothing else, this speaks to a rough edge that was not found in the shaping process for this feature!

michael-tharrington:
> Why not just edit your bookmarks?

In some cases, I think that totally makes sense. But in others cases, the links built into our system seem to be off. For instance, I recently wrote this issue up: https://github.com/thepracticaldev/tech-private/issues/460

PR #9717, opened 2020-08-10
Fix Netlify catch all redirections

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

http://docs.dev.to/api and anything else like http://docs.dev.to/ruby-doc all redirect to http://docs.forem.com/* instead of capturing the path.

See https://docs.netlify.com/routing/redirects/redirect-options/#splats for the correct syntax

### DIFF
--- a/docs/netlify.toml
+++ b/docs/netlify.toml
@@ -7,6 +7,8 @@
 
 [build]
   command = "make"
+
+# See <https://docs.netlify.com/routing/redirects/redirect-options/> for more options
 [[redirects]]
 from = "http://docs.dev.to"
 to = "http://docs.forem.com"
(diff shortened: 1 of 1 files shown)


docs/netlify.toml
```diff
@@ -7,26 +7,16 @@
...
 from = "http://docs.dev.to"
 to = "http://docs.forem.com"
 status = 301
 force = true
 
-[[redirects]]
-from = "https://docs.dev.to"
-to = "https://docs.forem.com"
-status = 301
-force = true
```
rhymes, line 20: this looks duplicate to me

docs/netlify.toml
```diff
@@ -7,26 +7,16 @@
...
-
 [[redirects]]
 from = "http://docs.dev.to/*"
-to = "http://docs.forem.com/*"
-status = 301
-force = true
```
rhymes, line 26: this looks duplicate to me

PR #9720, opened 2020-08-10
Migrate ImageResizer to Images::Optimizer

maestromac:
- [x] Refactor
- [x] Feature

## Description
A continuation of https://github.com/forem/forem/pull/9652 and following feedbacks.

https://github.com/forem/forem/pull/9652

1. Startup the server locally
2. Browse around to make ensure all images are loaded properly.

## Added tests?
- [x] yes, updated the specs

## Added to documentation?
- [x] no documentation needed

Keep close attention to production


app/helpers/application_helper.rb
```diff
@@ -77,14 +77,9 @@ def icon_url(name)
...
   def cloudinary(url, width = "500", quality = 80, format = "auto")
-    cl_image_path(url || asset_path("#{rand(1..40)}.png"),
-                  type: "fetch",
-                  width: width,
-                  crop: "limit",
-                  quality: quality,
-                  flags: "progressive",
-                  fetch_format: format,
-                  sign_url: true)
+    image_url = url || asset_path("#{rand(1..40)}.png")
```
rhymes, line 80: `url.presence` is probably more forward thinking as empty string would not result in the second clause to be called

maestromac:
@rhymes re-requesting review because I've added additional changes. Sorry for the noise.

spec/services/images/optimizer_spec.rb
```diff
@@ -0,0 +1,29 @@
+require "rails_helper"
+
+RSpec.describe Images::Optimizer, type: :service do
+  include CloudinaryHelper
```
mstruve, line 4: Why do we have to include this in the spec but not in the actual image optimizer file?

spec/services/images/optimizer_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  it "performs exactly like cl_image_path" do
+    cloudinary_url = cl_image_path(image_url,
+                                   type: "fetch",
+                                   width: 50, height: 50,
+                                   crop: "imagga_scale",
+                                   quality: "auto",
+                                   flags: "progressive",
+                                   fetch_format: "auto",
+                                   sign_url: true)
+    expect(described_class.call(image_url, width: 50, height: 50, crop: "imagga_scale")).to eq(cloudinary_url)
```
mstruve, line 17: Are we talking to a stubbed cloudinary object here to get these URL results?

spec/services/images/optimizer_spec.rb
maestromac: This is so I call `cl_image_path` as supposed to  `ActionController::Base.helpers.cl_image_path`. 

I'm trying to slim down available methods by removing `include CloudinaryHelper` from the app but I figure it doesn't too much in spec (which also let me access other view helpers like `asset_path`). This spec file is my current playground to verify we're still generating the proper URL. I can remove this in upcoming PRs to reduce indirection.

spec/services/images/optimizer_spec.rb
maestromac: This is not a stubbed method. This is the gem provided method that generates a URL. 

This spec is to make sure the combination of provided args and `DEFAULT_CL_OPTIONS` would yield the same output as how we previously use `cl_image_tag` directly.

mstruve: Ah great, wasnt sure if it was making an external call or something to generate the URL.

benhalpern:
This looks really brilliantly done!

PR #9724, opened 2020-08-11
Start using profile model

citizen428:

- [X] Refactor
- [X] Feature

## Description

This PR

* refactors `ProfileField` (attribute names are now generated differently and persisted),
* updates some field labels and placeholders and
* adds a Rake task for creating the existing DEV profile fields.

Note to whoever it may concern (but cc @benhalpern and @ludwiczakpawel so someone feels responsible): This introduces some slight copy changes to `/settings/profile`:

* "Dribble profile URL", "Behance profile URL", etc. will lose the word "profile" in the middle. It was a bit redundant anyway.
*  "I'm getting into" -> "Currently learning" (better reflects the DB column we store this in)
* "My projects and hacks" -> "Currently hacking on" (better reflects the DB column we store this in)
 
Part of #6994

1. Run `bundle exec rails db:migrate`
1. Run `rails temporary:create_profile_fields` 
1. Verify that `ProfileField.count == 29`.
1. Sanity check the generated profile fields.
1. Take a few minutes to calm down after this thrilling ride.

## Added tests?

- [X] yes

## Added to documentation?

- [X] no documentation needed

* [ ] Run `rails temporary:create_profile_fields`

### DIFF
--- a/app/models/profile_field.rb
+++ b/app/models/profile_field.rb
@@ -1,4 +1,8 @@
 class ProfileField < ApplicationRecord
+  before_create :generate_attribute_name
+
+  WORD_REGEX = /\w+/.freeze
+
   # Key names follow the Rails form helpers
   enum input_type: {
     text_field: 0,
@@ -9,16 +13,19 @@ class ProfileField < ApplicationRecord
 
   validates :label, presence: true, uniqueness: { case_sensitive: false }
   validates :active, inclusion: { in: [true, false] }
+  validates :attribute_name, presence: true, on: :update
 
   scope :active, -> { where(active: true) }
-
-  def attribute_name
-    label.titleize.delete(" ").underscore
-  end
 
   def type
     return :boolean if check_box?
 
     :string
+  end
+
+  private
+
+  def generate_attribute_name
+    self.attribute_name = label.titleize.scan(WORD_REGEX).join.underscore
   end
 end
--- a/app/services/profile_fields/add_link_fields.rb
+++ b/app/services/profile_fields/add_link_fields.rb
@@ -3,17 +3,17 @@ module ProfileFields
     include FieldDefinition
 
     group "Links" do
-      field "Facebook profile URL", :text_field, placeholder: "https://facebook.com/..."
+      field "Facebook URL", :text_field, placeholder: "https://facebook.com/..."
       field "Youtube URL", :text_field, placeholder: "https://www.youtube.com/channel/..."
-      field "StackOverflow profile URL", :text_field, placeholder: "https://stackoverflow.com/users/..."
-      field "LinkedIn profile URL", :text_field, placeholder: "https://www.linkedin.com/in/..."
-      field "Behance profile URL", :text_field, placeholder: "https://..."
-      field "Dribble profile URL", :text_field, placeholder: "https://dribble.com/..."
-      field "Medium profile URL", :text_field, placeholder: "https://..."
-      field "GitLab profile URL", :text_field, placeholder: "https://..."
-      field "Instagram profile URL", :text_field, placeholder: "https://..."
-      field "Mastodon profile URL", :text_field, placeholder: "https://..."
-      field "Twitch profile URL", :text_field, placeholder: "https://..."
+      field "StackOverflow URL", :text_field, placeholder: "https://stackoverflow.com/users/..."
+      field "LinkedIn URL", :text_field, placeholder: "https://www.linkedin.com/in/..."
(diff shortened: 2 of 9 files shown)


app/models/profile_field.rb
```diff
@@ -1,4 +1,8 @@
 class ProfileField < ApplicationRecord
+  before_create :generate_attribute_name
```
citizen428, line 2: This replaces the previous `attribute_name` method with a more sensible approach and pesists the name. Note that this name will never change again even if the field's label gets updated. This is so we can avoid rewriting every single user's stored profile for a wording change. @Ridhwana this means we're **not** going to expose `attribute_name` to admins.

app/services/profile_fields/add_link_fields.rb
```diff
@@ -3,17 +3,17 @@ class AddLinkFields
...
-      field "StackOverflow profile URL", :text_field, placeholder: "https://stackoverflow.com/users/..."
-      field "LinkedIn profile URL", :text_field, placeholder: "https://www.linkedin.com/in/..."
-      field "Behance profile URL", :text_field, placeholder: "https://..."
-      field "Dribble profile URL", :text_field, placeholder: "https://dribble.com/..."
-      field "Medium profile URL", :text_field, placeholder: "https://..."
-      field "GitLab profile URL", :text_field, placeholder: "https://..."
-      field "Instagram profile URL", :text_field, placeholder: "https://..."
-      field "Mastodon profile URL", :text_field, placeholder: "https://..."
-      field "Twitch profile URL", :text_field, placeholder: "https://..."
+      field "StackOverflow URL", :text_field, placeholder: "https://stackoverflow.com/users/..."
```
citizen428, line 8: For consistency's sake I decided to remove the word "profile" from most of these labels. It also makes the profile migration code a bit easier. Apart from that I added some better placeholders.

lib/data/coding_profile_fields.csv
```diff
@@ -1,4 +1,4 @@
 Skills/Languages,text_area,,What tools and languages are you most experienced with? Are you specialized or more of a generalist?,Coding
-I'm getting into,text_area,,What are you learning right now? What are the new tools and languages you're picking up right now?,Coding
-My projects and hacks,text_area,,What projects are currently occupying most of your time?,Coding
+Currently learning,text_area,,What are you learning right now? What are the new tools and languages you're picking up right now?,Coding
```
citizen428, line 2: Two more fields that got their label renamed to better reflect the stored attribute name.

app/models/profile_field.rb
Ridhwana: Sounds good!

citizen428:
@rhymes Thanks, I had a bigger branch but decided to split it into 2, apparently that wasn't as seamless as I thought.

rhymes:
@citizen428 I think it's fine, just the QA instructions were missing the first step: `bundle exec rails db:migrate` :-)

After that it worked fine!

citizen428:
Thanks for looking into it @rhymes, I'll mark it was ready for review again.

lib/tasks/temporary/create_profile_fields.rake
```diff
@@ -0,0 +1,7 @@
+namespace :temporary do
+  desc "Create profile fields based on DEV profile" 
+  task create_profile_fields: :environment do
```
mstruve, line 3: Is this something we should also put in the seed file or are we planning to do that later?

mstruve:
LMAO @rhymes you commented before I did and I still blindly went down my own setup path as well and ran into the same error https://github.com/forem/forem/pull/9724#pullrequestreview-467066131

Great minds fail alike ???? 😂

lib/tasks/temporary/create_profile_fields.rake
citizen428: The seeds files uses a mix right now:

```ruby
  ProfileFields::AddBaseFields.call
  ProfileFields::AddLinkFields.call
  ProfileFields::AddWorkFields.call
  coding_fields_csv = Rails.root.join("lib/data/coding_profile_fields.csv")
  ProfileFields::ImportFromCsv.call(coding_fields_csv)
  ProfileFields::AddBrandingFields.call
```

This was originally done when I wrote the various `ProfileFields::*` service objects to exercise both them and the CSV importer. I can change it in a future PR. The set of generated fields is exactly the same anyway, as `dev_profile_fields.csv` was generated by exporting the `profile_fields` table to CSV after running this seed task.

PR #9735, opened 2020-08-11
Create option for calculating score without external black box

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We presently use a "private/external endpoint" for calculating score, using AWS Lambda. As we get up and running, this is not entirely necessary. I think inserting a basic "is this post good?" calc if the SDK keys are not present should be fine and we can revisit the functionality down the road.

### DIFF
--- a/app/black_box/black_box.rb
+++ b/app/black_box/black_box.rb
@@ -16,10 +16,7 @@ class BlackBox
         reaction_points = (reaction_points * 0.8).to_i # watercooler posts shouldn't get as much love in feed
       end
 
-      article_hotness = function_caller.call(
-        "blackbox-production-articleHotness",
-        { article: article, user: article.user }.to_json,
-      ).to_i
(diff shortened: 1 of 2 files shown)


app/black_box/black_box.rb
```diff
@@ -38,6 +35,8 @@ def comment_quality_score(comment)
     def calculate_spaminess(story, function_caller = FunctionCaller)
       # accepts comment or article as story
       return 100 unless story.user
+      return 0 if ENV["AWS_SDK_KEY"].blank? # Skip this if we don't have a private spam score for now
+      return 0 if ENV["AWS_SDK_KEY"] == "foobarbaz" # Also skip if placeholder
```
benhalpern, line 39: This is just because at the moment I think we _need_ to have this set in prod, and are using `foobarbaz` at the moment in these cases.

PR #9738, opened 2020-08-11
Don't call Fastly if not present

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

If we don't have Fastly in an environment we can skip this stuff, eh?

### DIFF
--- a/app/labor/cache_buster.rb
+++ b/app/labor/cache_buster.rb
@@ -19,6 +19,8 @@ module CacheBuster
     #
     # https://github.com/fastly/fastly-ruby#efficient-purging
     return unless Rails.env.production?
+    return if ENV["FASTLY_API_KEY"].blank?
+    return if ENV["FASTLY_API_KEY"] == "foobarbaz"
 
     HTTParty.post("https://api.fastly.com/purge/https://#{ApplicationConfig['APP_DOMAIN']}#{path}",
                   headers: { "Fastly-Key" => ApplicationConfig["FASTLY_API_KEY"] })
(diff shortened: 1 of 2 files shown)


app/labor/cache_buster.rb
```diff
@@ -19,6 +19,8 @@ def self.bust(path)
     #
     # https://github.com/fastly/fastly-ruby#efficient-purging
     return unless Rails.env.production?
+    return if ENV["FASTLY_API_KEY"].blank?
+    return if ENV["FASTLY_API_KEY"] == "foobarbaz"
```
rhymes, line 23: is there a reason we're hardcoding this `foobarbaz` string 🤣 ?

jdoss: I can just unset the `FASTLY_API_KEY` or totally remove them so we are not hardcoding things. That might be the better path here.

vaidehijoshi: I don't know that we want to return without fastly set because busting with nginx will involve changes to CacheBuster and the Purgeable concern and if we short-circuit without fastly, then nginx won't work as the default in the fallback? If we did this, then I suspect that we'll have to do a bunch of conditional checks to see if nginx is "configured" which maybe is fine eventually, but for now, until we have nginx work done, I wonder if we actually really need to do these short-circuits at all?

vaidehijoshi: Wanted to also add this TL;DR: unless there is a strong reason to add these checks (am I maybe missing some important piece of info? Let me know if so!), I'd vote to not short-circuit/check for fastly ENV vars.

I'll be in this code soon in order to handle nginx purging, and IMHO, that seems like a better time to figure out ~whether we want to do~ how to handle these checks. :)

PR #9741, opened 2020-08-11
 🚀 Ability to allow tag moderators to create chat channels

sarthology:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
A non admin user can make a group and invite others as well. 

Screen Recording 2020-08-11 at 08 26 PM gif

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

## Post-deployment tasks?

* [ ] (Added by @citizen428) The chat channel with id 119143 now has an invalid name (empty string) that won't pass the newly added validation. There don't seem to be any chat channel memberships for this, so the channel should be removed.


sarthology:
Branch is up to dated it was some other issue that @nickytonline helped me resolve 👍 We just need to hard reset our master instead of normal merge. 😊

app/assets/stylesheets/chat.scss
```diff
@@ -106,10 +106,12 @@
 
 .chat__channels input {
   width: 100%;
+  outline: none;
 }
```
nickytonline, line 110: If the outline is removed, there needs to be some other visual indicator for keyboard users/accessibility.

app/javascript/chat/components/createChatModal.jsx
```diff
@@ -0,0 +1,65 @@
...
+const CreateChatModal = ({
+  toggleModalCreateChannel,
+  handleCreateChannelSuccess,
+}) => {
+  const [channelName, setchannelName] = useState('');
+  const [userNames, setUserNames] = useState('');
+
+  const handleCreateChannel = async (e) => {
+    e.preventDefault();
+    const res = await createChannel(channelName, userNames);
```
nickytonline, line 15: Prefer complete names, e.g. `result` instead of `res`.

nickytonline:
> Branch is up to dated it was some other issue that @nickytonline helped me resolve 👍 We just need to hard reset our master instead of normal merge. 😊

Justt to clarify @mstruve. They were merging upstream master into their forks master and basing branches off of their forked master. So now they do a pull of upstream master to get latest or a hard reset to upstream master onto their fork’s master.

citizen428:
We talked a bit about the security implications of this. At first, this will only be rolled out to moderators/trusted users, so we get a friendly group of testers.

Then we can work out a more comprehensive strategy to reduce the spam potential of this feature. One initial idea was to not make this feature available to brand new accounts (say less than 1 or 2 weeks old for example), so they can't sign up, open a channel, add a bunch of people and dump in some spam links.

app/policies/chat_channel_policy.rb
```diff
@@ -51,6 +51,10 @@ def joining_invitation_response?
     record.present?
   end
 
+  def create_channel?
+    record.present? && user.id
```
citizen428, line 55: **suggestion (blocking):** As dicussed yesterday, let's change this to mods/trusted users for now.

app/services/send_channel_invitation_service.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def send_invitations
+    usernames = invitation_usernames.split(",").map do |username|
+      username.strip.delete("@")
+    end
+    if usernames.present?
+      users = User.where(username: usernames)
+      invitations_sent = chat_channel.invite_users(users: users, membership_role: "member", inviter: current_user)
+      message = if invitations_sent.zero?
+                  "No invitations sent. Check for username typos."
```
citizen428, line 18: **suggestion (non-blocking):** Maybe in a future PR, we can add some form of auto-complete on username, at least for mutual follows.

citizen428 (changes requested):

narender2031:
@citizen428 please re-start the build. un-related tests are failing.

nickytonline:
> @citizen428 please re-start the build. un-related tests are failing.

I restarted it for you @narender2031

app/policies/chat_channel_policy.rb
```diff
@@ -51,6 +51,10 @@ def joining_invitation_response?
     record.present?
   end
 
+  def create_channel?
+    record.present? && user.has_role?(:mod_relations_admin)
```
citizen428, line 55: There is only 1 user with this role. Maybe we should use `:tag_moderator`, there are 274 of them and we have a tag mod newsletter where we can inform them about this new feature so they actually test it. /cc @michael-tharrington

citizen428:
Sorry this feel off my radar 😞 The PR looks almost good to go now, but I think we should change the role that will have permission to do this @narender2031, see comment. Thank you!

citizen428 (approved):

app/controllers/chat_channels_controller.rb
```diff
@@ -172,6 +174,41 @@ def channel_info
     render json: { error: "not found", status: 404 }, status: :not_found
   end
 
+  def create_channel
```
citizen428, line 177: To make CodeClimate happy — and because this is a pretty long method — we could potentially introduce a `ChatChannels::Create` service.

narender2031: @citizen428  I was thinking about the same approach but still, there will be 21 methods.

narender2031: @citizen428 I fixed the code-climate issues. I found a few methods in the controller that can be moved under the helpers.

citizen428: We are getting into personal preference territory here, but I personally strongly dislike helpers because the became a messy global dumping ground. If this is for building JSON data specifically, it sounds like a job for a serializer.

app/controllers/chat_channel_memberships_controller.rb
```diff
@@ -138,7 +134,7 @@ def request_details
     user_chat_channels = ChatChannel.includes(:chat_channel_memberships).where(
       chat_channel_memberships: { user_id: current_user.id, role: "mod", status: "active" },
     )
-    @memberships = user_chat_channels.map(&:requested_memberships).flatten
+    @memberships = user_chat_channels.map(&:requested_memberships)&.flatten
```
citizen428, line 137: Ruby has a combination of `map` + `flatten` called `flat_map`:

```suggestion
    @memberships = user_chat_channels.flat_map(&:requested_memberships)
```

app/helpers/chat_channel_memberships_helper.rb
```diff
@@ -15,4 +15,30 @@ def formatted_membership_user(memberships)
       }
     end
   end
+
+  def membership_users(memberships)
```
citizen428, line 19: Where is this used? I can't seem to find it elsewhere in this PR.

app/helpers/chat_channels_helper.rb
```diff
@@ -0,0 +1,48 @@
...
+  def render_unopened_json_response
+    @chat_channels_memberships = if session_current_user_id
+                                   ChatChannelMembership.where(user_id: session_current_user_id)
+                                     .where(has_unopened_messages: true)
+                                     .where(show_global_badge_notification: true)
+                                     .where.not(status: %w[removed_from_channel left_channel])
+                                     .includes(%i[chat_channel user])
+                                     .order("chat_channel_memberships.updated_at" => :desc)
+                                 else
+                                   []
```
citizen428, line 12: Non-blocking: It would be slight more Rails-y to use `ChatChannelMembership.none` here, so `@chat_channels_memberships` is always an instance of `ChatChannelMembership::ActiveRecord_Relation`.

app/helpers/chat_channels_helper.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def render_pending_json_response
+    @chat_channels_memberships = if current_user
+                                   current_user
+                                     .chat_channel_memberships.includes(:chat_channel)
+                                     .where(status: "pending")
+                                     .order("chat_channel_memberships.updated_at" => :desc)
+                                 else
+                                   []
```
citizen428, line 23: Same as above.

app/helpers/chat_channels_helper.rb
```diff
@@ -0,0 +1,48 @@
+# rubocop:disable Rails/HelperInstanceVariable
```
citizen428, line 1: Blocking: We should absolutely not ignore this warning. Instance variables have no place in helpers since they indicate that this is not a generic helper, but something that relies on a specific context.

citizen428 (changes requested):

app/helpers/chat_channel_memberships_helper.rb
narender2031: it is used in this file `app/views/chat_channel_memberships/chat_channel_info.json.jbuilder`. I was just shifting the common code under the one file.

narender2031:
@citizen428 all suggestions are updated.

app/helpers/chat_channels_helper.rb
```diff
@@ -0,0 +1,46 @@
...
+                                                               left_channel]).pluck(:chat_channel_id)
+  end
+
+  def render_joining_request_json_response
+    requested_memberships_id = current_user
+      .chat_channel_memberships
+      .includes(:chat_channel)
+      .where(chat_channels: { discoverable: true }, role: "mod")
+      .pluck(:chat_channel_id)
+      .map { |membership_id| ChatChannel.find_by(id: membership_id).requested_memberships }
```
citizen428, line 38: This `map` + `flatten` + `map` combination look like it could be simplified. Maybe something like

```ruby
      .pluck(:chat_channel_id)
      .flat_map { |membership_id| ChatChannel.find_by(id: membership_id).requested_memberships.ids
```

app/helpers/chat_channels_helper.rb
```diff
@@ -0,0 +1,46 @@
+module ChatChannelsHelper
+  def render_unopened_json_response
```
citizen428, line 2: None of these methods technically render anything, they return records from the DB.

citizen428 (commented):
This seems to shape up nicely from the Ruby side, maybe @nickytonline has some more thoughts on JS.

citizen428:
I also restarted the build since it was failing due to a flaky spec.

narender2031:
@citizen428 or @nickytonline  can you please restart the build. There are some non-related specs are failing

nickytonline:
Gonna check this out today @narender2031 .

citizen428:
@narender2031 Restarted.

app/javascript/chat/actions/chat_channel_setting_actions.js
```diff
@@ -173,3 +173,24 @@ export async function updateMembershipRole(membershipId, chatChannelId, role) {
...
+export async function createChannel(channelName, userNames) {
+  const response = await request(`/create_channel`, {
+    method: 'POST',
+    body: {
+      chat_channel: {
+        channel_name: channelName,
+        invitation_usernames: userNames,
+      },
+    },
+    credentials: 'same-origin',
```
nickytonline, line 192: refactor (non-blocking): This header is not necessary as the `request` utility function handles this by default. See https://github.com/forem/forem/blob/master/app/javascript/utilities/http/request.js#L47

app/javascript/chat/chat.jsx
```diff
@@ -52,12 +53,14 @@ export default class Chat extends Component {
     chatChannels: PropTypes.string.isRequired,
     chatOptions: PropTypes.string.isRequired,
     githubToken: PropTypes.string.isRequired,
+    tagModerator: PropTypes.object.isRequired,
```
nickytonline, line 56: refactor (blocking): Use [PropTypes.shape](https://reactjs.org/docs/typechecking-with-proptypes.html) instead of the generic `PropTypes.object`.

```suggestion
    tagModerator: PropTypes.shape({ isTagModerator: PropTypes.bool.isRequired }).isRequired,
```

app/javascript/chat/components/createChatModal.jsx
```diff
@@ -0,0 +1,65 @@
+import { h } from 'preact';
```
nickytonline, line 1: nitpick: Component filenames should be Pascal case.

app/javascript/chat/components/createChatModal.jsx
```diff
@@ -0,0 +1,65 @@
...
+import { useState } from 'preact/hooks';
+import PropTypes from 'prop-types';
+import { Modal } from '../../crayons/Modal';
+import { createChannel } from '../actions/chat_channel_setting_actions';
+const CreateChatModal = ({
```
nickytonline, line 6: documentation (blocking): Document the component as it's exported.

nitpick: Arrow functions are only necessary when you need to deal with scope, i.e. `this`. In the case of function components, there is no this. Everything is functions.

app/javascript/chat/components/createChatModal.jsx
```diff
@@ -0,0 +1,65 @@
+import { h } from 'preact';
+import { useState } from 'preact/hooks';
+import PropTypes from 'prop-types';
+import { Modal } from '../../crayons/Modal';
```
nickytonline, line 4: refactor (non-blocking): You can use the webpack alias for crayons instead.

```suggestion
import { Modal } from '@crayons/Modal';
```

app/javascript/chat/components/createChatModal.jsx
```diff
@@ -0,0 +1,65 @@
...
+const CreateChatModal = ({
+  toggleModalCreateChannel,
+  handleCreateChannelSuccess,
+}) => {
+  const [channelName, setchannelName] = useState('');
+  const [userNames, setUserNames] = useState('');
+
+  const handleCreateChannel = async (e) => {
+    e.preventDefault();
+    const result = await createChannel(channelName, userNames);
```
nickytonline, line 15: refactor (blocking): There should be a check for an empty channel name as the initial value of `channelName` is an empty string and the user could also just submit an empty channel name. One way to handle this client-side as well would be to have the button to create the channel disabled until there is a valid channel name.

I'm assuming this is currently handled on the backend in terms of validation @citizen428?

app/javascript/chat/components/createChatModal.jsx
```diff
@@ -0,0 +1,65 @@
...
+          value={userNames}
+          onChange={(e) => setUserNames(e.target.value)}
+        />
+
+        <button
+          href="#"
+          className="crayons-btn"
+          onClick={handleCreateChannel}
+          style="margin-top:20px"
+        >
```
nickytonline, line 52: refactor (blocking): Use the design system buttons and as well, a button does not have an `href` attribute. See the [Button component](https://storybook.forem.com/?path=/story/3-components-buttons--default) in Storybook.

nickytonline (changes requested):
Just some change requests @sarthology.

app/javascript/chat/components/createChatModal.jsx
citizen428: Funnily enough I don't see a validation for `channel_name` in `app/models/chat_channel.rb`. If you try to create a chat channel from a Rails console without any attributes, you'll only see errors for channel_type and slug:

```ruby
cc = ChatChannel.create
cc.save
cc.errors_as_sentence
#=> "Channel type can't be blank, Channel type is not included in the list, and Slug can't be blank"
```

So probably this should be added too.

narender2031: @nickytonline & @citizen428 Added the client-side and backend validation for channel name text field. 
client-side:
- button is disabled until you have no chat channel name. 
backend: 
- added a model validation for this field.

narender2031:
@nickytonline and @citizen428 PR suggestions are added also file name changed to PascalCase.

app/javascript/chat/chat.jsx
```diff
@@ -1341,6 +1329,45 @@ export default class Chat extends Component {
                 ) : null}
               </span>
             </button>
+            {this.state.isTagModerator ? (
+              <button
```
nickytonline, line 1333: refactor (blocking): Use the design system [Button](https://storybook.forem.com/?path=/story/3-components-buttons--default) component.

app/javascript/chat/chat.jsx
```diff
@@ -1341,6 +1329,45 @@ export default class Chat extends Component {
...
+                className="chat__channelssearchtoggle "
+                aria-label="Toggle request manager"
+                onClick={this.toggleModalCreateChannel}
+                data-content="sidecar-joining-request-manager"
+              >
+                <span
```
nickytonline, line 1339: refactor (blocking): This span acting as a button appears to be redundant given that it's contained within a `<button />`.

app/javascript/chat/components/ChannelButton.jsx
```diff
@@ -0,0 +1,123 @@
+import { h, Component, createRef } from 'preact';
+import PropTypes from 'prop-types';
+
+class ChannelButton extends Component {
```
nickytonline, line 4: refactor (blocking): use a function component as this is new code.

app/javascript/chat/components/ChannelButton.jsx
```diff
@@ -0,0 +1,123 @@
...
+    if (isActiveChannel) {
+      this.buttonRef.current.click();
+    }
+  }
+
+  renderChannelImage = () => {
+    const { channel, newMessagesIndicator, discoverableChannel } = this.props;
+
+    return (
+      <span
```
nickytonline, line 19: refactor (non-blocking): Is the span necessary or can all these data-attributes be on the ``?

app/javascript/chat/components/ChannelButton.jsx
```diff
@@ -0,0 +1,123 @@
...
+      handleSwitchChannel,
+      otherClassname,
+      newMessagesIndicator,
+      isUnopened,
+      discoverableChannel,
+      triggerActiveContent,
+    } = this.props;
+
+    return (
+      <button
```
nickytonline, line 53: refactor (blocking): Use the design system [Button](https://storybook.forem.com/?path=/story/3-components-buttons--default) component.

app/javascript/chat/components/ChannelButton.jsx
```diff
@@ -0,0 +1,123 @@
...
+        onClick={
+          discoverableChannel ? triggerActiveContent : handleSwitchChannel
+        }
+        data-content="sidecar-channel-request"
+        data-channel-id={channel.chat_channel_id}
+        data-channel-slug={channel.channel_modified_slug}
+        data-channel-status={channel.status}
+        data-channel-name={channel.channel_name}
+      >
+        <span
```
nickytonline, line 67: refactor (non-blocking): Is this span necessary? Can all the data attributes and styling be on the button?

nickytonline (changes requested):
@narender2031 @sarthology just some more comments and request changes.

app/javascript/chat/components/ChannelButton.jsx
narender2031: This span is necessary for another styling for the message indicator.

app/javascript/chat/components/createChatModal.jsx
citizen428: Thank you @narender2031!

@mstruve I did a quick check with Blazer, we have one chat channel where the channel name is an empty string which would now be invalid. There are no `ChatChannelMemberships` for this channel, so I think it can safely be deleted. I'll add this is a post-deployment task for someone with console access.

app/javascript/chat/chat.jsx
```diff
@@ -1878,7 +1860,7 @@ export default class Chat extends Component {
 
   renderChannelBackNav = () => {
     return (
-      <button
+      <Button
         className="crayons-btn crayons-btn--icon-rounded crayons-btn--ghost active-channel__back-btn"
         onClick={this.navigateToChannelsList}
         onKeyUp={(e) => {
```
nickytonline, line 1866: @reobin or @link2twenty, this could be refactored in a future PR to use the new Keyboard component/hook.

app/javascript/chat/components/ChannelButton.jsx
```diff
@@ -0,0 +1,74 @@
+import { h, createRef } from 'preact';
+import { useEffect } from 'preact/hooks';
+import PropTypes from 'prop-types';
+import { defaultChannelPropTypes } from '../../common-prop-types/channel-list-prop-type';
+import ChannelImage from './ChannelImage';
+import { Button } from '@crayons';
+
+export default function ChannelButton(props) {
```
nickytonline, line 8: refactor (blocking): Comment the component via JSDoc documentation.

nickytonline:
@citizen428, @mstruve, this is more of a product question, but do we cap the number of channels a user can create? Also do we have a throttling mechanism in place if someone were maliciously to create thousands of channels?

nickytonline (changes requested):
refactor (blocking): Document the component via JSDoc documentation.

mstruve:
> do we cap the number of channels a user can create? Also do we have a throttling mechanism in place if someone were maliciously to create thousands of channels?

We do not have a throttle around chat channel creation yet but that would pretty easy to add. [Here is an example of throttling being added to feedback messages](https://github.com/forem/forem/pull/7832)

The file you will most likely want to add it to is [rate_limit_checker.rb](https://github.com/forem/forem/blob/master/app/services/rate_limit_checker.rb).

citizen428:
@narender2031 @sarthology Can you add throttling as described by @mstruve in a follow-up PR?

mstruve:
> @narender2031 @sarthology Can you add throttling as described by @mstruve in a follow-up PR? I made an issue for this: #11351

There is a lot going on in this PR. I think it would make our job as reviewers a bit easier if we split up the throttling PR from the create feature PR. What do you all think?

narender2031:
> > @narender2031 @sarthology Can you add throttling as described by @mstruve in a follow-up PR? I made an issue for this: #11351
> 
> There is a lot going on in this PR. I think it would make our job as reviewers a bit easier if we split up the throttling PR from the create feature PR. What do you all think?

Yeah Let's split the throttling part from the PR. I will create a separate PR for this, once this PR is merged with.

nickytonline (commented):
There appears to be a bunch of changes unrelated to this PR. Is this a merge issue?

narender2031:
> There appears to be a bunch of changes unrelated to this PR. Is this a merge issue?

yes

citizen428 (changes requested):
There are still too many changes in here that shouldn't be part of this PR, please fix them first before re-requesting a review.

narender2031:
@citizen428 @nickytonline PR is updated. There are some non-related PR changes These appear due to es-lint fix and rubocop fixes before committing the code. I hope that's fine at the moment.

app/helpers/chat_channel_memberships_helper.rb
```diff
@@ -15,4 +15,30 @@ def formatted_membership_user(memberships)
...
+    memberships.includes(:user).map do |membership|
+      {
```
citizen428, line 21: Shouldn't this block use the `format_membership` helper method instead of duplicating everything?

app/helpers/chat_channels_helper.rb
```diff
@@ -0,0 +1,44 @@
+module ChatChannelsHelper
+  def unopened_json_response
```
citizen428, line 2: These names always bothered me, because they sound like they should really be serializers. **But** they don't actually return a JSON response, but `ActiveRecord::Relation`s and helpers are definitely not the place to make DB complex  DB queries.

config/routes.rb
```diff
@@ -347,6 +347,7 @@
 
     # Chat channel
     patch "/chat_channels/update_channel/:id" => "chat_channels#update_channel"
+    post "/create_channel" => "chat_channels#create_channel"
```
citizen428, line 350: Just out of curiosity, the standard POST to `/chat_channels` isn't an option here?

citizen428 (approved):

app/helpers/chat_channels_helper.rb
narender2031: I was thinking to move them to the services or model layer.  In coming up PR we will refactor our controller also. Because our controllers are fat.

nickytonline:
THere's a bunch of files in this PR that are just prettier changes. I put up PR #11586 to help with that.

nickytonline:
I'm logged on as a non-admin user, but I do not see the `+` button to create a channel. Is there a step I'm missing in regards to getting set up? There were no test instructions aside from the animated gif which show the `+` button present. If it makes any difference, even as an admin user, I do not see the `+` button.

narender2031:
> I'm logged on as a non-admin user, but I do not see the `+` button to create a channel. Is there a step I'm missing in regards to getting set up? There were no test instructions aside from the animated gif which show the `+` button present. If it makes any difference, even as an admin user, I do not see the `+` button.

This feature is only available for `tag_moderator` at the moment. To test this you need to add the role `tag_moderator` for the user.

nickytonline:
> > I'm logged on as a non-admin user, but I do not see the `+` button to create a channel. Is there a step I'm missing in regards to getting set up? There were no test instructions aside from the animated gif which show the `+` button present. If it makes any difference, even as an admin user, I do not see the `+` button.
> 
> This feature is only available for `tag_moderator` at the moment. To test this you need to add the role `tag_moderator` for the user. 

I'm confused as to why only a tag_moderator can do this, but that's a separate topic. I was able to make myself a tag moderator, but I still do not see the <kbd>+</kbd> button.

@narender2031, can you document the steps to test this in the PR description? I'd really like to close the loop on this PR.

These are the steps I've taken with the latest from your branch.

1. Made myself a super admin
1. From the Tags admin page made myself a tag moderator for the #discuss tag, http://localhost:3000/admin/tags/13

image

1. Navigated to Connect, http://localhost:3000/connect/, and see that I am now a part of the #discussion mods group

image

1. There is no <kbd>+</kbd> button to create a channel.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -33,7 +33,9 @@ as any relevant images for UI changes._
 
 ### UI accessibility concerns?
 
-_If your PR includes UI changes, please replace this line with details on how accessibility is impacted and tested. For more info, check out the [Forem Accessibility Docs](https://docs.forem.com/frontend/accessibility)._
+_If your PR includes UI changes, please replace this line with details on how
```
nickytonline, line 36: @narender2031, if you merge latest from master, this file and several others will no longer be part of your PR. See https://github.com/forem/forem/pull/9741#issuecomment-732482975

I'd do it myself, but I do not appear to have access to push to your branch.

app/javascript/chat/components/ChannelButton.jsx
```diff
@@ -0,0 +1,91 @@
+import { h, createRef } from 'preact';
+import { useEffect } from 'preact/hooks';
+import PropTypes from 'prop-types';
+import { defaultChannelPropTypes } from '../../common-prop-types/channel-list-prop-type';
+import ChannelImage from './ChannelImage';
+import { Button } from '@crayons';
+
+/**
+ * Render the cahnnel button
```
nickytonline, line 9: spelling mistake:

```suggestion
 * Render the channel button
```

nickytonline (changes requested):

citizen428:
> I'm confused as to why only a tag_moderator can do this, but that's a separate topic.

@nickytonline The answer to that is relatively simple: small rollout to a trusted user group so we can test this feature in the wild without opening the abuse floodgates just yet. Note: this decision predates abuse reports in Connect.

.github/PULL_REQUEST_TEMPLATE.md
narender2031: I will do that.

app/views/chat_channels/index.html.erb
```diff
@@ -14,7 +14,8 @@
       data-pusher-key="<%= ApplicationConfig["PUSHER_KEY"] %>"
       data-github-token="<%= @github_token %>"
       data-chat-channels="<%= [] %>"
-      data-chat-options="<%= { showChannelsList: true, showTimestamp: true, activeChannelId: @active_channel&.id, currentUserId: current_user.id }.to_json %>">
+      data-chat-options="<%= { showChannelsList: true, showTimestamp: true, activeChannelId: @active_channel&.id, currentUserId: current_user.id }.to_json %>"
+      data-tag-moderator=<%= { isTagModerator: current_user.has_role?(:tag_moderator) }.to_json %>>
```
citizen428, line 18: This seems to be incorrect which is also why @nickytonline can't see the + button despite making his user a tag moderator locally.

```suggestion
      data-tag-moderator=<%= { isTagModerator: current_user.tag_moderator?) }.to_json %>>
```

I used a Rails console to verify this:

```ruby
user.has_role?(:tag_moderator)
#=> false
user.tag_moderator?
#=> true
```

If you look into `user.rb`, you'll see that the method is defined as follows:

```ruby
roles.where(name: "tag_moderator").any?
```

citizen428 (changes requested):

narender2031:
testing docs:
- add a new `tag_moderator` role for the logged-in user.
- go to connect `http://localhost:3000/connect`
- Click on the `+` icon. A modal will pop up
- Create the channel.

citizen428:
I now see the modal but when I click "Create" nothing happens and I see the following error in the console:

> Uncaught (in promise) SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data

This is how my `data-tag-moderator` attribute looks like, I assume the HTML entity is the culprit here: `data-tag-moderator={&quot;isTagModerator&quot;:true}>`

narender2031:
> I now see the modal but when I click "Create" nothing happens and I see the following error in the console:
> 
> > Uncaught (in promise) SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
> 
> This is how my `data-tag-moderator` attribute looks like, I assume the HTML entity is the culprit here: `data-tag-moderator={&quot;isTagModerator&quot;:true}>`

checking this

narender2031:
> Uncaught (in promise) SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data

@citizen428 I am not able to reproduce the json issue. can you please share me the steps or video.

app/views/chat_channels/index.html.erb
```diff
@@ -14,7 +14,8 @@
       data-pusher-key="<%= ApplicationConfig["PUSHER_KEY"] %>"
       data-github-token="<%= @github_token %>"
       data-chat-channels="<%= [] %>"
-      data-chat-options="<%= { showChannelsList: true, showTimestamp: true, activeChannelId: @active_channel&.id, currentUserId: current_user.id }.to_json %>">
+      data-chat-options="<%= { showChannelsList: true, showTimestamp: true, activeChannelId: @active_channel&.id, currentUserId: current_user.id }.to_json %>"
+      data-tag-moderator=<%= { isTagModerator: current_user.tag_moderator? }.to_json %>>
```
nickytonline, line 18: refactor (blocking): This needs to be in double quotes as it renders as an HTML attribute with a value. As @citizen428 [mentioned](https://github.com/forem/forem/pull/9741/#issuecomment-733511751), this is the culprit.

```suggestion
      data-tag-moderator="<%= { isTagModerator: current_user.tag_moderator? }.to_json %>">
```

nickytonline (changes requested):

citizen428:
When I try to create a channel in see the following message in my Rails log:

> Pundit::NotAuthorizedError - not allowed to create_channel? this ChatChannel

This is because the policy still uses `user.has_role?(:tag_moderator)` instead of `user.tag_moderator?`, which we established isn't the right method. Once I made the following change everything seems to work:

```diff
diff --git a/app/policies/chat_channel_policy.rb b/app/policies/chat_channel_policy.rb
index 87206dad0..7d89f75d2 100644
--- a/app/policies/chat_channel_policy.rb
+++ b/app/policies/chat_channel_policy.rb
@@ -52,7 +52,7 @@ def joining_invitation_response?
   end

   def create_channel?
-    record.present? && user.has_role?(:tag_moderator)
+    record.present? && user.tag_moderator?
   end

   private
```

nickytonline (changes requested):
UX (blocking): There are some UX gaps that need to be handled.

- When the channel is created, it does not appear in the list. You need to select a different channel filter (all, direct, group) and then come back to all or the group filter to see the new channel.

createchannel1

- To be able to create a channel, you need to give focus to the Invite Users text field so that the Create channel button becomes enabled.

createchannel3

- Channels can have the same name. I imagine we don’t want this? I'm guessing this should be a constraint in the database that the channel name be unique? cc: @citizen428 

image

- Pressing the escape key should close the modal. I believe this is a separate PR though as it should be part of our modal component.

citizen428:
> To be able to create a channel, you need to give focus to the Invite Users text field so that the Create channel button becomes enabled.

I was under the impression that this is intended behavior, i.e. you should invite at least one user right away.

> Channels can have the same name. I imagine we don’t want this? 

Yes, I thought we had already mentioned this earlier in this PR.

nickytonline:
> > To be able to create a channel, you need to give focus to the Invite Users text field so that the Create channel button becomes enabled.
> 
> I was under the impression that this is intended behavior, i.e. you should invite at least one user right away.
> 

@citizen428, it enables the submit button if you focus in the Invite Users text box, but the field is not required as I can submit without inviting someone. If we require at least one invite, there should be validation for this on the client-side and server-side.

citizen428:
Thanks for clarifying @nickytonline, @sarthology probably knows more about the intended behavior.

I double-checked the validations in `ChatChannel`, we only validate slugs for uniqueness, not names:

```ruby
validates :slug, uniqueness: true, presence: true
validates :channel_name, presence: true
```

It would probably make sense to add a uniqueness validation for names too @sarthology

sarthology:
> It would probably make sense to add a uniqueness validation for names too @sarthology

Names can be the same, that's why `channel_name` not kept unique in the schema.
`Uniqueness validation should be with a unique index.`

 I think the use case for that is in future people can have the same group name as in messenger or WhatsApp, as long as slug is different.

cc @benhalpern

sarthology:
> When the channel is created, it does not appear in the list. You need to select a different channel filter (all, direct, group) and then come back to all or the group filter to see the new channel.

Not able to regenerate this issue, working fine at my end. I suspect that's a pusher issue, setting up pusher will resolve this. @nickytonline

citizen428:
> Names can be the same, that's why channel_name not kept unique in the schema. Uniqueness validation should be with a unique index.

We can always add that constraint in a migration after checking how many duplicated we have in production (if any).

> I think the use case for that is in future people can have the same group name as in messenger or WhatsApp, as long as slug is different.

Some names may be popular ("React") and some people may end up in more than one of them. Or once we open this up to everyone people with bad intentions could reuse the name of "official" channels like "Contributors". Not everyone will always check the slug.

nickytonline:
@sarthology I've tested this locally and I still have the blocking UX issues I mentioned in https://github.com/forem/forem/pull/9741#pullrequestreview-540206360, i.e.:

- You still have to give focus to the invite users text box to enable the <kbd>Create</kbd> button though. I know you're supposed to invite someone, but it still isn't required, so I would either enable the Create button as long  as there is a channel name or add the requirement in the frontend and backend that at least one person must be invited. I'm leaning towards option 1 where they can create the channel without inviting someone and revisit requiring at least one invite in future work if that's the behaviour we want long term.

- The channel still doesn't get added to the list for me when I create it. As mentioned I need to switch filters and come back to see it. You mentioned to enable Pusher, but I already have Pusher configured. I have Pusher and the beam set up.

- For the duplicate names I will go with what @citizen428 mentios in https://github.com/forem/forem/pull/9741#issuecomment-736272102. I get that folks might want to create non-unique names which could make sense if they were private channels, but if the channel is public, we should consider disallowing this. Maybe the unique constraint is the slug, name and whether or not the channel is private.

sarthology:
@nickytonline 

> 1. You still have to give focus on the invite user's text box to enable the Create button though.

Done

> 2. The channel still doesn't get added to the list for me when I create it.

Done

> 3. For the duplicate names, I will go with what @citizen428 mentions in #9741 (comment). 

I believe as @citizen428 said, we should add this in a follow-up PR.

nickytonline:
- @sarthology, the Create channel button is now enabled if only a channel name is included, thanks.
- when the channel is created though, it is still not added to the list unless I toggle filters still.
- one thing I noticed as well while testing is a regression in the UI. If there are a lot of channels, you can not scroll to the last channel or two as they are covered by the settings panel in the bottom left of Connect.

Local with this Branch

image

Production

image

There are also some rspec tests failing, but I think they may be resolved by merging latest from master.

citizen428:
Yes, this PR has been going on for a while, so I'd like to wrap it up and do things like uniqueness validations in a follow-up. I was unexpectedly busy these past few days, I'll have another look on Monday.

citizen428 (approved):
In the spirit of our agreement I think we can go ahead and merge this and do follow-ups in separate PRs.

sarthology:
Yes, please.

nickytonline (approved):
> In the spirit of our agreement I think we can go ahead and merge this and do follow-ups in separate PRs.

As @citizen428 mentioned, subsequent PRs for the follow up. 🚀

citizen428:
@sarthology I restarted the build on Travis, it's green now. Given that we have two approvals I'll merge this now so you can get started on the follow-up PRs 🙂

PR #9750, opened 2020-08-12
Migrate profile data

citizen428:

- [X] Refactor
- [X] Feature

## Description

Useful context: v1 of this feature tries to lay all the groundwork for generalized profiles for other communities, while not breaking DEV. This means that some of the code here sets up compatibility methods, which will eventually be removed.

This PR migrated profile data from `User` to `Profile`.  The idea is not to switch everything immediately but instead to the following:

1. Set up synchronisation between `User` and `Profile`.
    1. All new users will automatically have a filled profile. 
    1. Changes to profile related columns will still be stored on `User` so we can easily roll back, but they will also propagate to profiles. This shouldn't be too much of a problem and I'm trying to keep updates to the necessary minimum, but still making sure @mstruve sees this.
2. A Rake task for backfilling profile data for all users.

Part of #6994.

* Run `rails temporary:migrate_profiles` and check the backfilled profiles.
* Create a new user and see that they automatically have a filled profile.
* Make changes to a user and ensure they propagate to the profile.

## Added tests?

- [X] yes

## Added to documentation?

- [X] no documentation needed

We need to run `rails temporary:migrate_profiles` at one point, but the code is set up so nothing relies on that being done at a specific point in time. If we don't feel like kickstarting the task for all users, what we really need to do is run the following for each user once:

```ruby
Profile.create(user: user, data: Profiles::ExtractData.call(user))
```


app/models/profile.rb
```diff
@@ -1,6 +1,8 @@
 class Profile < ApplicationRecord
   belongs_to :user
 
+  validates :user_id, uniqueness: true
```
citizen428, line 4: Every user should only have a single profile.

app/models/profile.rb
```diff
@@ -9,6 +11,4 @@ def self.define_store_accessors!
   end
 
   define_store_accessors!
-
-  validates :data, presence: true
```
citizen428, line 13: I removed (a) there's a default of `{}` set on the DB level and (b) this default doesn't play well with the `presence?` checl that this validation performs.

app/models/profile_field.rb
```diff
@@ -1,4 +1,8 @@
 class ProfileField < ApplicationRecord
```
citizen428, line 1: Part of  #9724, this will go away when that other branch was merged.

app/models/user.rb
```diff
@@ -5,6 +5,50 @@ class User < ApplicationRecord
   include Searchable
   include Storext.model
 
+  # NOTE: @citizen428 This is temporary code during profile migration and will
+  # be removed.
+  # rubocop:disable Metrics/BlockLength
+  concerning :ProfileMigration do
```
citizen428, line 11: `concerning` defines an anonymous concern and immediately includes it. This is my preferred way for concerns that aren't actually shared since it allows me to group related things without Rubocop complaining, but also doesn't introduce any indirection. And since most of this will be removed relatively soon, I'd rather give it more visibiltiy.

app/models/user.rb
```diff
@@ -5,6 +5,50 @@ class User < ApplicationRecord
...
+    included do
+      PROFIE_FIELDS =
+        (Profiles::ExtractData::DIRECT_ATTRIBUTES + Profiles::ExtractData::MAPPED_ATTRIBUTES.values).freeze
+
+      # All new users should automatically have a profile
+      after_create_commit { Profile.create(user: self, data: Profiles::ExtractData.call(self)) }
+
+      # Keep saving changes locally for the time being, but propagate them to profiles.
+      after_update_commit do
+        if (previous_changes.keys.map(&:to_sym) & User::PROFIE_FIELDS).present?
```
citizen428, line 21: Only update the profile if the recent changes actually included a profile field.

app/models/user.rb
```diff
@@ -5,6 +5,50 @@ class User < ApplicationRecord
...
+          profile.update(data: Profiles::ExtractData.call(self))
+        end
+      end
+
+      # Define wrapped getters for profile attributes. We first try to get the
+      # value from the profile and if it doesn't exist there we move retrieve it
+      # from here.
+      Profiles::ExtractData::DIRECT_ATTRIBUTES.each do |attribute|
```
citizen428, line 29: These getters allow us to run the profile backfilling at our convenience without any rush.

app/services/profile_fields/add_link_fields.rb
```diff
@@ -3,17 +3,17 @@ class AddLinkFields
     include FieldDefinition
```
citizen428, line 3: Part of  #9724, this will go away when that other branch was merged.

app/services/profiles/extract_data.rb
```diff
@@ -0,0 +1,47 @@
+module Profiles
+  module ExtractData
+    DIRECT_ATTRIBUTES = %i[
```
citizen428, line 3: Note that this hard-coded for DEV values. We'll keep this in prod for a bit until all profiles have been migrated, then generalize this.

db/migrate/20200811044202_add_attribute_name_to_profile_field.rb
```diff
@@ -0,0 +1,5 @@
+class AddAttributeNameToProfileField < ActiveRecord::Migration[6.0]
```
citizen428, line 1: Part of  #9724, this will go away when that other branch was merged.

lib/data/coding_profile_fields.csv
```diff
@@ -1,4 +1,4 @@
 Skills/Languages,text_area,,What tools and languages are you most experienced with? Are you specialized or more of a generalist?,Coding
```
citizen428, line 1: Part of  #9724, this will go away when that other branch was merged.

lib/data/dev_profile_fields.csv
```diff
@@ -0,0 +1,29 @@
+Display email on profile,check_box,,,Basic
```
citizen428, line 1: Part of  #9724, this will go away when that other branch was merged.

spec/factories/profiles.rb
```diff
@@ -1,6 +0,0 @@
-FactoryBot.define do
```
citizen428, line 1: We don't really need this anymore. User will automatically have attached profiles and profiles without users don't make sense. If you need a profile in a spec, `create(:user).profile` is the way to go.

spec/factories/users.rb
```diff
@@ -26,8 +26,6 @@
     bg_color_hex                 { Faker::Color.hex_color }
     text_color_hex               { Faker::Color.hex_color }
 
-    after(:create) { |user| create(:profile, user: user) }
```
citizen428, line 29: This is now done by the model itself.

spec/models/profile_field_spec.rb
```diff
@@ -10,6 +10,14 @@
       it { is_expected.to validate_presence_of(:label) }
```
citizen428, line 10: Part of  #9724, this will go away when that other branch was merged.

app/models/user.rb
```diff
@@ -5,6 +5,52 @@ class User < ApplicationRecord
...
+  # rubocop:disable Metrics/BlockLength
+  concerning :ProfileMigration do
+    included do
+      PROFIE_FIELDS =
+        (Profiles::ExtractData::DIRECT_ATTRIBUTES + Profiles::ExtractData::MAPPED_ATTRIBUTES.values).freeze
+
+      default_scope { includes(:profile) } # rubocop:disable Rails/DefaultScope
+
+      # All new users should automatically have a profile
+      after_create_commit { Profile.create(user: self, data: Profiles::ExtractData.call(self)) }
```
citizen428, line 19: Rails 6 introduce more granular callbacks, which I'm using here to distinguish between creates and updates.

app/models/user.rb
```diff
@@ -5,6 +5,52 @@ class User < ApplicationRecord
...
 
+  # NOTE: @citizen428 This is temporary code during profile migration and will
+  # be removed.
+  # rubocop:disable Metrics/BlockLength
+  concerning :ProfileMigration do
+    included do
+      PROFIE_FIELDS =
+        (Profiles::ExtractData::DIRECT_ATTRIBUTES + Profiles::ExtractData::MAPPED_ATTRIBUTES.values).freeze
+
+      default_scope { includes(:profile) } # rubocop:disable Rails/DefaultScope
```
citizen428, line 16: In this PR I apparently use everything I usually rant about. 😒 That said, we do want to always load profiles with users and for the time being and given the temporary nature of this code, this seems to be an ok solution. I'll look into how to tie this into Devise/Warden/`current_user` eventually.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,8 +6,8 @@
      commits if your PR already received reviews or comments.
```
citizen428, line 6: This should not be here at all, I'll clean it up before moving this out of draft.

app/models/user.rb
mstruve: I'm cool with this bc its temporary. Lets get the job done quickly and smoothly then we can simply remove the code.

citizen428: Cool @mstruve, thanks for confirming!

> Lets get the job done quickly

That's the goal. Though I'm not going to rush this, as I don't want to risk messing up live data on DEV. 

I had some time to think about this over the weekend (cc @Ridhwana): relatively soon we want to add code for updating profile data through the profile page but using the new model. At that point, we can reverse the direction of the sync.

Before: Changes update `User` first, syncs to `Profile`.
After: Changes update `Profile` first, but syncs back to `User`.

Once we're confident that everything works as expected, we can remove the sync and drop the related columns from `users`, but I generally wait for a bit before performing this sort of destructive action on production data.

I have some ideas how to deal with users vs profiles, preloading, etc, after, but first I want to see how our new use case looks like since according to a discussion with @benhalpern on Forem.dev (see [here](https://forem.dev/ben/comment/9p) we may remove some things from the UI to simplify things overall.

citizen428:
Regarding the `default_scope`: it makes me uneasy. So far I've explored/spiked the following options:

1. `default_scope`
    ```ruby
    default_scope { includes(:profile) } # rubocop:disable Rails/DefaultScope
    ```
    This works reasonably well, but a lot of specs fail because of unused eager loading. We should temporarily disable this.

     **Pros:**
       * it's a small and self-contained change
     **Cons:**
       * It loads the profile in many cases where we don't need it.
       * It makes many specs fail (easily fixable)
2. A dedicated scope plus a new method `current_user_with_profile`
    ```ruby
    scope :with_profile, -> { includes(:profile) }

    def current_user_with_profile
      @current_user_with_profile ||= current_user.tap do |user|
        ::ActiveRecord::Associations::Preloader.new.preload(user, :profile)
      end
    end
    ```
 
   **Pros:**
     * The cleanest solution
     * Forces call-sites to be more explicit about what they need
     * Only loads what's needed
    Cons:
      * Needs a lot of code spec changes right now and I'm not sure that's what we should spend time one at this stage.
3. Doing nothing and living with the occasional extra query to load a profile. We should temporarily disable the Bullet error in specs for this specific case.
    **Pros:**
      * Easiest approach at this point in time
    **Cons:**
      * We need to disable bullet warnings for missing includes for this case.
      * We will run extra database queries under certain conditions.

My gut feeling: always including profiles (solution #1, where will also include profiles in all the contexts where we don't need them) will generate more extra queries than not including them at all (solution 3, where we occasionally have an extra query for the profile). I'm inclined to — but not set on — going with option 3 for the time being, but I'd like some input (cc @Ridhwana @mstruve @rhymes).

Ridhwana:
> Regarding the `default_scope`: it makes me uneasy. So far I've explored/spiked the following options (I left in all the commits for the time being so you can actually look at things):
> 
> 1. `default_scope`
>    ```ruby
>    default_scope { includes(:profile) } # rubocop:disable Rails/DefaultScope
>    ```
>    
>    
>    This works reasonably well, but a lot of specs fail because of unused eager loading. We should temporarily disable this.
>    
>    
>    
>    Pros
>    Cons
>    
>    
>    
>    
>    it's a small and self-contained change
>    It loads the profile in many cases where we don't need it
>    
>    
>    
>    it makes many specs fail (easily fixable)
> 2. A dedicated scope plus a new method `current_user_with_profile`
>    ```ruby
>    scope :with_profile, -> { includes(:profile) }
>    
>    def current_user_with_profile
>      @current_user_with_profile ||= current_user.tap do |user|
>        ::ActiveRecord::Associations::Preloader.new.preload(user, :profile)
>      end
>    end
>    ```
>    
>    
>    
>    
>    
>    Pros
>    Cons
>    
>    
>    
>    
>    the cleanest solution
>    Needs a lot of code spec changes right now and I'm not sure that's what we should spend time on at this stage.
>    
>    
>    Forces call-sites to be more explicit about what they need
>    
>    
>    
>    Only loads what's needed
> 3. Doing nothing and living with the occasional extra query to load a profile. We should temporarily disable the Bullet error in specs for this specific case.
>    
>    
>    
>    Pros
>    Cons
>    
>    
>    
>    
>    Easiest approach at this point in time
>    We will run extra database queries under certain conditions.
>    
>    
>    
>    We need to disable bullet warnings for missing includes for this case.
> 
> My gut feeling: always including profiles (solution #1, where will also include profiles in all the contexts where we don't need them) will generate more extra queries than not including them at all (solution 3, where we occasionally have an extra query for the profile). I'm inclined to — but not set on — going with option 3 for the time being, but I'd like some input (cc @Ridhwana @mstruve @rhymes).

I'm in favour of option 3. We can always improve on the solution as we chug along with this feature, so for now I'd vote for using the simplest approach and improving on it as we learn more.

rhymes:
My vote goes for option 3 as well

mstruve:
I'm in for option 3, dont over complicate things until we have to!

citizen428:
Thanks everyone, really appreciate both the feedback and the pragmatism! 😃

config/environments/development.rb
```diff
@@ -128,6 +128,8 @@
     # Supress incorrect warnings from Bullet due to included columns: https://github.com/flyerhzm/bullet/issues/147
     Bullet.add_whitelist(type: :unused_eager_loading, class_name: "Article", association: :top_comments)
     Bullet.add_whitelist(type: :unused_eager_loading, class_name: "Comment", association: :user)
+    # NOTE: @citizen428 Temporarily ignoring this while working out user - profile relationship
+    Bullet.add_whitelist(type: :n_plus_one_query, class_name: "User", association: :profile)
```
citizen428, line 132: Alas @rhymes' PR for hasn't been merged upstream yet, so this is stil `whitelist` 😞

app/models/user.rb
rhymes: TIL

app/services/profiles/extract_data.rb
```diff
@@ -0,0 +1,47 @@
...
+      available_for
+      behance_urL
```
rhymes, line 5: is this case sensitive? because the `L` is here capitalized

app/services/profiles/extract_data.rb
```diff
@@ -0,0 +1,47 @@
...
+      medium_url
+      mostly_work_with
+      name
+      summary
+      twitch_url
+      website_url
+      youtube_url
+    ].freeze
+
+    MAPPED_ATTRIBUTES = {
```
rhymes, line 27: love this mapping :D

config/environments/development.rb
rhymes: np

db/migrate/20200813031851_add_unique_index_to_profiles.rb
```diff
@@ -0,0 +1,8 @@
+class AddUniqueIndexToProfiles < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def change
+    remove_index :profiles, column: :user_id, algorithm: :concurrently
+    add_index :profiles, :user_id, unique: true, algorithm: :concurrently
```
rhymes, line 6: Yes! :)

app/services/profiles/extract_data.rb
mstruve: Didnt want you to mistake it for an "i" 😂

db/migrate/20200813031851_add_unique_index_to_profiles.rb
mstruve: Will it cause any problems that these are both running concurrently or does postgres know to remove one and then add the next? Or is it fine if both end up running at the same time? Just curious I am not super educated on Postgres internals :)

lib/tasks/temporary/create_profile_fields.rake
```diff
@@ -1,7 +1,8 @@
 namespace :temporary do
-  desc "Create profile fields based on DEV profile" 
-  task create_profile_fields: :environment do
-    csv = Rails.root.join("lib/data/dev_profile_fields.csv")
-    ProfileFields::ImportFromCsv.call(csv)
+  desc "Migrate existing profile data"
+  task migrate_profiles: :environment do
+    User.find_each do |user|
+      Profile.create(user: user, data: Profiles::ExtractData.call(user))
```
mstruve, line 5: Should this be in a data update script so it can run on all environments or are we not quite there yet?

db/migrate/20200813031851_add_unique_index_to_profiles.rb
rhymes: Great question!

> Regular index builds permit other regular index builds on the same table to occur simultaneously, but only one concurrent index build can occur on a table at a time. In either case, schema modification of the table is not allowed while the index is being built. Another difference is that a regular CREATE INDEX command can be performed within a transaction block, but CREATE INDEX CONCURRENTLY cannot.

from https://www.postgresql.org/docs/11/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY

as I interpret it, the second index is queued and has to wait.

As the index is done outside of a transaction and is entered a catalog before it is actually built, I think we're safe here:

> In a concurrent index build, the index is actually entered into the system catalogs in one transaction, then two table scans occur in two more transactions. Before each table scan, the index build must wait for existing transactions that have modified the table to terminate. After the second scan, the index build must wait for any transactions that have a snapshot (see Chapter 13) predating the second scan to terminate. Then finally the index can be marked ready for use, and the CREATE INDEX command terminates. Even then, however, the index may not be immediately usable for queries: in the worst case, it cannot be used as long as transactions exist that predate the start of the index build.

If we weren't, we could not even do this in different files. We'd have to wait for one index to be finished before submitting the second one (can't wait for PostgreSQL 12 so we can build a mini admin panel to monitor concurrent index building and its status: https://www.depesz.com/2019/04/18/waiting-for-postgresql-12-report-progress-of-create-index-operations/)

lib/tasks/temporary/create_profile_fields.rake
citizen428: A data update script automatically gets scheduled after deployment, I'd much rather have someone start this by hand at a good time or run `Profile.create(user: user, data: Profiles::ExtractData.call(user))` for smaller batches as they see fit.

db/migrate/20200813031851_add_unique_index_to_profiles.rb
citizen428: @rhymes Thanks for looking this up.

Additionally, I'd like to add that `CONCURRENTLY` doesn't refer to indexing operations (as quoted you can't actually have 2 different `CREATE INDEX CONCURRENTLY` on the same table at the same time) but update operations on the indexed table, avoiding potential locks. The `DROP` and `CREATE` will most likely happen in sequence, `CONCURRENTLY` just ensures that we don't lock the table, which is a bit of a moot point anyway, since the table will be empty before this PR gets deployed.

app/services/profiles/extract_data.rb
citizen428: Thanks @rhymes, fixed now.

lib/tasks/temporary/create_profile_fields.rake
mstruve: How do you plan on making sure this is run on all instances of Forem then? We have Lee, Forem.dev and who knows what other communities out there that will need this run.

citizen428: @mstruve I figured we can delay this until the feature is completely done but you just changed my mind on this 👍

app/models/user.rb
```diff
@@ -5,6 +5,59 @@ class User < ApplicationRecord
...
+  # NOTE: @citizen428 This is temporary code during profile migration and will
+  # be removed.
+  # rubocop:disable Metrics/BlockLength
+  concerning :ProfileMigration do
+    included do
+      PROFIE_FIELDS =
+        (Profiles::ExtractData::DIRECT_ATTRIBUTES + Profiles::ExtractData::MAPPED_ATTRIBUTES.values).freeze
+
+      # NOTE: this allows us to skip the profile create callback in a thread-safe way
+      attr_accessor :_skip_creating_profile
```
citizen428, line 17: Tired: using `skip_callback` which modifies global state and can lead to hard to track down bugs since it's not thread-safe.
Wired: using an instance-level `attr_accessor` so we don't add more flakiness to our tests.

citizen428:
@mstruve Thanks for pushing me to turn the temporary Rake tasks into a data update script! While doing so I added specs, which made me finally sit down and figure out the best way to skip the profile creation callback so we can have a dedicated profile factory again and don't need the `create(:user).profile` workaround. 

Even though this PR has already received two approvals I re-requested your review and maybe @rhymes and @Ridhwana want to have a look at the recent changes too.

app/models/user.rb
rhymes: ahha neat trick! though It's odd to have the underscore. Is it only going to be used in the tests?

lib/data_update_scripts/20200819025131_migrate_profile_data.rb
```diff
@@ -0,0 +1,13 @@
+module DataUpdateScripts
+  class MigrateProfileData
+    def run
+      User.find_each do |user|
+        # NOTE: no production users have profiles yet, but we want this script
+        # to be idempotent.
+        next if user.profile.present?
+
+        Profile.create(user: user, data: Profiles::ExtractData.call(user))
```
rhymes, line 9: Is there a chance that this will fail?
I know we have some weird data in some early users, I wonder if we should use `create!` to bubble up the issues, or send the failed errors somewhere

rhymes (approved):
Left a coulpe of notes

app/models/user.rb
citizen428: Yes, it's only supposed to be used in tests, the underscore is meant to be a hint to that. It's not a completely uncommon practice btw, I've seen the same pattern in other codebases.

lib/data_update_scripts/20200819025131_migrate_profile_data.rb
citizen428: I want to say that it can't fail, but computers, can't trust 'em. Here's what this does:

1. Take a user's attributes as hash.
2. Extract relevant keys from it.
3. Create a profile for the user with the data in a JSONB column.

There are no validations on the JSONB keys, only type coercions. What do you think?

rhymes: Let's cross fingers :D

mstruve: Given the high number of users we have, unless you deploy this during light hours there is a good chance the job gets killed before it can finish. A problem I have on my list to fix for Sidekiq so we can have long-running jobs. @citizen428 when you deploy this make sure to check Blazer to ensure this jobs completes. If it does not, ping me and I can run it in a console.

app/models/user.rb
mstruve: TIL! I'm going to guess not many people know about this. Might be worth a little blog post :) 

Should we put a comment here that the purpose of this is for tests? I had to go back through the PR conversation to figure out why it was here. I could see someone else having the same question

lib/data_update_scripts/20200819025131_migrate_profile_data.rb
rhymes: @mstruve what is intended by light hours?

Unfortunately given the time zone difference, usually @citizen428 is offline when the bulk of the US team is operational. Given that, can the SRE team deploy this when you see fit? Otherwise we can do that a member of team OSS will ping a SRE team member tomorrow UTC afternoon and see if it's mergeable by then. 

cc @Ridhwana here as she has the same timezone as I do and she's more knowledgeable than I on the profile generalization code

mstruve: I mean when we are not deploying a lot since a deploy restarts the Sidekiq Workers so honestly, @citizen428 doing it during his hours is probably the best.

app/models/user.rb
citizen428: @mstruve Yes, I already made a note to write up a post. I'll update the comment now.

lib/data_update_scripts/20200819025131_migrate_profile_data.rb
citizen428: Thanks, @mstruve, I'm deploying this not at 4:18 UTC which seems like it would be one of our lightest times (11:18pm Chicago area, past midnight in NY and 6:18am in Central Europe/South Afriace).

PR #9758, opened 2020-08-12
Restore #9621 with new migration script

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This is https://github.com/forem/forem/pull/9621 plus https://github.com/forem/forem/commit/4cf10f286fada6efea731a70eb580d4c52ff7f5c

that was added for the following reasons: 

- as some of us ran the script locally we were left with neither of the file files (.env or config/application.yml) after running bin/setup
- if the yaml file contains empty keys, those will result in a parsing error when copied over the .env
- the default values are only in the sample file but are not copied over in the migration script

thus:

- now we do not delete the application.yml so the developer can manually check things have been copied over
- the default keys are added
- fixed the copying of empty keys

Keeping this in draft

## QA

- checkout the PR
- run `/bin/setup`
- start the app and see that stuff works without key errors


rhymes:
Superseded by https://github.com/forem/forem/pull/9767

PR #9762, opened 2020-08-13
Update styling of authentication feed card for logged out users

lisasy:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
Update authentication feed card as part of initiative to improve UX & visual design of existing login experiences.

Closes https://github.com/forem/InternalProjectPlanning/issues/17

### Current

Small screen:
Screen Shot 2020-08-12 at 18 28 54

Large screen:
Screen Shot 2020-08-12 at 10 10 23 AM

### Updated versions:
Small screen:
Screen Shot 2020-08-12 at 17 56 04

Large screen:
Screen Shot 2020-08-12 at 6 12 33 PM

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [ ] no documentation needed

### DIFF
--- a/app/views/stories/_sign_in_invitation.html.erb
+++ b/app/views/stories/_sign_in_invitation.html.erb
@@ -4,8 +4,8 @@
           alt="Secondary <%= community_name %> logo" />
     <div class="authentication-feed__content">
       <h2 class="authentication-feed__title">
-community_members_label
-                                                                %>
+<a href="/"><%= community_name %></a> is a community of <br />
+        <%= number_with_delimiter User.registered.estimated_count %> amazing <%= community_members_label %>
       </h2>
       <p class="authentication-feed__description">
         Log in to customize your experience and get involved.
--- a/app/views/stories/_sign_in_invitation.html.erb
+++ b/app/views/stories/_sign_in_invitation.html.erb
@@ -1,24 +1,24 @@
-<div class="single-article single-article-small-pic feed-cta" id="in-feed-cta">
-  <div class="cta-container" id="cta-content">
-    <img class="secondary-logo" src="<%= cloudinary(safe_logo_url(SiteConfig.secondary_logo_url), 190) %>"
-         alt="Secondary <%= community_name %> logo" />
-
-    <h2>
-      <a href="/"><%= community_name %></a> is a community of <br />
-      <%= number_with_delimiter User.registered.estimated_count %> amazing <%= community_members_label %>
-    </h2>
-
-    <h3>
-      Create your profile to customize your experience and get involved.
-    </h3>
-
-    <div class="button-container">
-      <a href="/enter" class="cta cta-button" aria-label="Sign up.">
-        Join Now
-      </a>
-      <div class="feed-cta-sub">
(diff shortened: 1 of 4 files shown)


app/views/stories/_sign_in_invitation.html.erb
```diff
@@ -1,24 +1,24 @@
...
-        <em>We require social login to prevent abuse.</em>
-      </div>
+<div class="authentication-feed__card" id="in-feed-cta">
+  <div class="authentication-feed__container">
+    <img class="authentication-feed__logo" src="<%= cloudinary(safe_logo_url(SiteConfig.secondary_logo_url), 190) %>"
+          alt="Secondary <%= community_name %> logo" />
+    <div class="authentication-feed__content">
+      <h2 class="authentication-feed__title">
+community_members_label
+                                                                %>
```
fdoxyz, line 8: I think a portion of the welcome message may have been lost along the way? Should it maybe look something like this instead?

```suggestion
<a href="/"><%= community_name %></a> is a community of <br />
        <%= number_with_delimiter User.registered.estimated_count %> amazing <%= community_members_label %>
```

So the card can render on the feed like this:

Screen Shot 2020-08-13 at 11 23 12

lisasy: Whoops! Let me look into this.

lisasy:
> I think that updating the query selectors in the failing tests should do most of the trick to make them pass, but also left an question inline 🙂

How do I run the tests in the CLI to check which query-selectors are failing? @fdoxyz

fdoxyz:
From [Travis logs](https://travis-ci.com/github/forem/forem/jobs/371480194) there's 3 failing specs from 2 different files:

Screen Shot 2020-08-13 at 12 35 12

So I would personally run them and try to fix them one by one like this:

- `bundle exec rspec spec/views/stories/sign_in_invitation.html.erb_spec.rb`
- `bundle exec rspec  spec/system/articles/user_visits_articles_by_timeframe_spec.rb`

The second one it's looking for element `.cta-feed` which was renamed, and for the first one I actually think that the text typo might have fixed it already :D

app/views/stories/_sign_in_invitation.html.erb
```diff
@@ -1,24 +1,24 @@
...
     </div>
   </div>
+
+  <div class="authentication-feed__actions">
+    <a href="/enter" class="crayons-btn" aria-label="Create new account">
+      Create new account
+    </a>
+    <a href="/enter" class="crayons-btn crayons-btn--ghost-brand" aria-label="Log in">
+      Log in
+    </a>
```
ludwiczakpawel, line 22: Why two separate buttons taking to the same place?

lisasy: This is the fix for now to make sure that these actions address people's different intentions, but we're doing this in a phased way, and eventually this will lead people to a split experience.

PR #9766, opened 2020-08-13
Bump storybook to version 6

### LINKED ISSUE #9763: Upgrade to Storybook 6

Storybook 6 has been released. We will need to make some changes going forward in regards to Storybook. See https://dev.to/emma/storybook-v6-0-released-1cg4 

<details>
<summary>Details</summary>

There are some breaking/new changes. See https://dev.to/emma/storybook-v6-0-released-1cg4

</details>

Upgrade to Storybook 6 by making the required upgrade path changes.

**Additional context**

emgoto:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Library version bump

## Description

Following this [migration guide](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md) I bumped Storybook to v6.0.

Notable changes:
* `@storybook/addon-notes` has been deprecated (there is no v6.0 available for it, as there are currently no maintainers). But we can still use it for now. ~~I have removed it and replaced it with `@storybook/addon-docs`. This puts the notes up the top of the page under the "docs" tab. It's not a 1-for-1 replacement - looking at the docs tab shows all the storybooks in that folder as well so it is a bit weird (let me know if you'd prefer to stick with v5.3 of the docs for now)~~
* The `withA11y` decorator is [no longer needed](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#removed-witha11y-decorator) (it will still work without it) so I have removed it
* We were using the import `@storybook/addon-knobs/react`. This no longer works (Storybook will fail to start up with it). I couldn't find specific mention of it in the docs, but I changed those usages to `@storybook/addon-knobs` and the knobs still seem to work fine.

I haven't done a _thorough_ story by story check, but overall it still seems like everything that I looked at functions the same.

If you want to see  the updates to Storybook in action, check out the [PR's preview deploy](https://deploy-preview-9766--storybookdevto.netlify.app/)

Closes #9763 

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Things like this:
```
export const Basic = () => <Button />
Basic.story = {
  name: 'foo',
  parameters: { ... },
  decorators: [ ... ],
};
```

Can be replaced with:

```
export const Basic = () => <Button />
Basic.storyName = 'foo';
Basic.parameters = { ... };
Basic.decorators = [ ... ];
```
It won't necessarily save any lines of code, looking at the codebase, but it may be worth switching to the new approach for consistency's sake (and so that as people add more storybooks, they won't be confused into using the old approach when looking at other code for inspiration).

There is a codemod that will make this change easy to do [here](https://github.com/storybookjs/storybook/blob/next/lib/codemod/README.md#csf-hoist-story-annotations).


app/javascript/.storybook/main.js
```diff
@@ -8,8 +8,8 @@ module.exports = {
     '@storybook/addon-knobs',
     '@storybook/addon-actions',
     '@storybook/addon-links',
-    '@storybook/addon-a11y/register',
-    '@storybook/addon-notes/register-panel',
+    '@storybook/addon-a11y',
+    '@storybook/addon-docs',
```
nickytonline, line 12: We can look at switching to `@storybook/addon-docs` in another PR. The docs add on preserves our existing documentation from notes, but we may need to do some work on the rendered components as it includes decorators like the theme selector in it.

Let's do the switch to addon-docs this for another PR.

Before with addon-notes

image

After with addon-docs

image

app/javascript/.storybook/preview.js
```diff
@@ -66,9 +65,16 @@ const themeSwitcherDecorator = (storyFn) => {
 };
 
 addDecorator(themeSwitcherDecorator);
-addDecorator(withA11y);
 
 addParameters({
+  docs: {
```
nickytonline, line 70: I believe this is related to switching to the addon-docs, so let's pull this out for now and as mentioned above, let's do this in another PR.

nickytonline:
> ## [optional] Are there any post deployment tasks we need to perform?
> Things like this:
> 
> ```
> export const Basic = () => <Button />
> Basic.story = {
>   name: 'foo',
>   parameters: { ... },
>   decorators: [ ... ],
> };
> ```
> 
> Can be replaced with:
> 
> ```
> export const Basic = () => <Button />
> Basic.storyName = 'foo';
> Basic.parameters = { ... };
> Basic.decorators = [ ... ];
> ```
> 
> It won't necessarily save any lines of code, looking at the codebase, but it may be worth switching to the new approach for consistency's sake (and so that as people add more storybooks, they won't be confused into using the old approach when looking at other code for inspiration).
> 
> There is a codemod that will make this change easy to do [here](https://github.com/storybookjs/storybook/blob/next/lib/codemod/README.md?rgh-link-date=2020-08-13T11%3A23%3A30Z#csf-hoist-story-annotations).

If you'd like to do this as well @emgoto, let me know. I think it makes sense to go with the newer style of writing Storybook stories if that's what they're recommending.

emgoto:
> If you'd like to do this as well @emgoto, let me know.

For sure, would be happy to give it a go.

PR #9767, opened 2020-08-13
Remove ENVied Gem, Use dotenv Take 2

mstruve:
- [x] Refactor

## Description
This is a followup to #9621. The difference between the original and this is the migration script. Now the migration 
- **does NOT remove the application.yml file.** This allows users that may have issues with the migration not to lose their keys which @rhymes and I(yeah I did it too when I creating this) were both victims to. 
- It **sets default values** that may not be present in the application.yml file. These default however have been added to the `.env_sample` file so going forward they should always be present for users starting fresh.

TODO: Remove sample application yml file and Figaro gem

## QA Instructions
You dont have to do all of these but if you want to give it a thorough test. 

- Pull down branch
- run `bin/setup`
- Ensure app still runs
- run `bin/setup` again to make sure no errors
- Ensure app still runs
- force kill app and boot it back up, should still run

alt_text


lib/tasks/temporary/env_variables.rake
```diff
@@ -0,0 +1,43 @@
...
+  File.open(".env", "w") do |env_file|
+    # add default values only if they are not already present in the
+    # application.yml file which would result in a duplicate key
+    REQUIRED_VALUES.each_pair do |variable, value|
+      env_file.write("export #{variable}=#{value}\n") unless application_yml.include?(variable.to_s)
+    end
+
+    File.open("config/application.yml", 'r') do |file|
+      file.each_line do |line|
+        new_line = line.gsub(": ", "=")
```
mstruve, line 35: @rhymes I found with your script the colons within the default URLs got swapped out for `=` which we dont want so I switched back to this way. I also couldn't find the commit even after fetching all so I went ahead and copied your code into my own commit.

rhymes: Perfect! :-)

mstruve:
@citizen428 can you poke around the container setup and let me know if you see anything not working? Please and THANK YOU!

mstruve:
Yeah @maestromac you have to live with those warnings until application.yml is removed manually or when we remove the Figaro gem

PR #9769, opened 2020-08-13
Add Options to "Invalidate" and "Mark as Valid" Vomit Reactions

juliannatetreault:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR adds two buttons, "Invalidate" and "Mark as Valid," to `_abuse_reports.html.erb` that gives Admins the option to un-confirm a confirmed vomit reaction and reopen an invalidated vomit reaction. To accomplish this, this PR adds a conditional button to `_abuse_reports.html.erb` and a couple of functions to handle the un-confirming and reopening of a vomit reaction in `reaction_controller.js`.

- To QA these changes, ensure that you are a `super_admin` and navigate to `/admin/reports` in your browser:
Screen Shot 2020-08-13 at 9 49 56 AM

- Once you have navigated there, click "confirm" on one of the vomit reactions located under "Vomit Reactions" -> "Open/Unresolved":
Screen Shot 2020-08-13 at 9 51 18 AM

- The confirmed vomit can then be found under the "Resolved" tab:
Screen Shot 2020-08-14 at 2 50 30 PM

- Click "Invalidate" on the confirmed vomit reaction. This will "invalidate" the vomit, placing it under the "Invalid" tab:
Screen Shot 2020-08-17 at 2 15 51 PM

- From there, go to the "Invalid" tab and click "Mark as Valid". This will place the vomit reaction previously marked as "invalid" underneath the "Resolved" tab, providing an Admin with more flexibility in case something is wrongly marked:
Screen Shot 2020-08-17 at 2 21 11 PM

- Try to break things! ⚒️ 
 
## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

N/A

erase

### DIFF
--- a/app/javascript/admin/controllers/reaction_controller.js
+++ b/app/javascript/admin/controllers/reaction_controller.js
@@ -2,7 +2,7 @@
 import { Controller } from 'stimulus';
 
 export default class ReactionController extends Controller {
-  static targets = ['invalid', 'confirmed'];
+  static targets = ['invalid', 'confirmed', 'valid'];
 
   // eslint-disable-next-line class-methods-use-this
   /* eslint no-alert: "error" */
@@ -41,6 +41,10 @@ export default class ReactionController extends Controller {
     this.updateReaction(this.reactionId, this.invalidStatus);
   }
 
+  updateReactionValid() {
+    this.updateReaction(this.reactionId, this.validStatus);
+  }
+
   reactableUserCheck() {
     if (this.reactableType === 'user') {
       // eslint-disable-next-line no-restricted-globals
@@ -66,5 +70,9 @@ export default class ReactionController extends Controller {
 
   get invalidStatus() {
     return this.invalidTarget.dataset.altstatus;
+  }
+
+  get validStatus() {
+    return this.validTarget.dataset.altstatus;
   }
 }
--- a/app/javascript/admin/controllers/reaction_controller.js
+++ b/app/javascript/admin/controllers/reaction_controller.js
@@ -2,7 +2,7 @@
 import { Controller } from 'stimulus';
 
 export default class ReactionController extends Controller {
-  static targets = ['invalid', 'confirmed', 'valid'];
+  static targets = ['invalid', 'confirmed'];
 
   // eslint-disable-next-line class-methods-use-this
   /* eslint no-alert: "error" */
@@ -41,8 +41,8 @@ export default class ReactionController extends Controller {
     this.updateReaction(this.reactionId, this.invalidStatus);
   }
 
-  updateReactionValid() {
-    this.updateReaction(this.reactionId, this.validStatus);
+  updateReactionConfirmed() {
+    this.updateReaction(this.reactionId, this.confirmedStatus);
   }
 
   reactableUserCheck() {
@@ -70,9 +70,5 @@ export default class ReactionController extends Controller {
 
   get invalidStatus() {
     return this.invalidTarget.dataset.altstatus;
-  }
-
-  get validStatus() {
-    return this.validTarget.dataset.altstatus;
   }
 }
--- a/app/models/feedback_message.rb
+++ b/app/models/feedback_message.rb
@@ -15,7 +15,7 @@ class FeedbackMessage < ApplicationRecord
             }
   validates :status,
             inclusion: {
-              in: %w[Open Invalid Resolved]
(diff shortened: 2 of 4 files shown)


juliannatetreault:
> LGTM! 👍
> 
> It's been a little bit sense I messed with reactions. Does this undo the effect of confirming a vomit on an article or user?

Thanks for the review, @jacobherrington, and great question! From what I understand, the user and their article will be unaffected by the vomit reaction by doing this, but I can circle back around to this with a better answer once I have one, as I am still trying to get acquainted with this part of the codebase as well. 😄

mstruve:
The logic doesnt seem to add up for me here. 
- Vomit reaction status is resolved (meaning the user is affected by the reaction from what I can tell)
- Vomit reaction status is updated via an "unresolved" button which changes the status to valid (this I find confusing, why not go back to status "open"?)
- A vomit reaction status being valid to me would mean that it's true, resolved, and should affect the user but it seems like the opposite is true.

Maybe I am a little in the weeds here but would love more clarification about why we need another status rather than going back to open.

juliannatetreault:
> LGTM! 👍
> 
> It's been a little bit sense I messed with reactions. Does this undo the effect of confirming a vomit on an article or user?

Circling back to your question, @jacobherrington: when an Admin clicks the "INVALIDATE" button to un-confirm a vomit reaction, there will be no effect on the user's post or article. It is _only_ when an Admin clicks the "CONFIRM" button that there will be an effect on the user and/or their article. 😄

juliannatetreault:
> The logic doesnt seem to add up for me here.
> 
> * Vomit reaction status is resolved (meaning the user is affected by the reaction from what I can tell)
> * Vomit reaction status is updated via an "unresolved" button which changes the status to valid (this I find confusing, why not go back to status "open"?)
> * A vomit reaction status being valid to me would mean that it's true, resolved, and should affect the user but it seems like the opposite is true.
> 
> Maybe I am a little in the weeds here but would love more clarification about why we need another status rather than going back to open.

This is a great question and to be completely honest, I was totally stumped by this as well. After poking around at the code and asking around for clarification, I think I have a sufficient answer for you @mstruve! 

So, apparently, in this context, both `open` and `valid` mean the same thing. The difference is that the `FeedbackMessage` report uses the status of `open`, while a `Reaction` uses the status of `valid` to convey the same thing. I hope this answers your question and clarifies some of the confusion around it all, rather than making it more confusing!

I should also mention that after speaking with my team, we agreed that this should probably be renamed to `open` in the future to avoid future confusion (because it's definitely a bit confusing!), but for now, it's a bit out of scope for the task at hand.

juliannatetreault:
I'm getting a `/black_box/black_box_spec.rb` failure again (this happened to me yesterday) both locally and on `master`: `BlackBox#article_hotness_score returns the lower correct value if article tagged with watercooler`. Any thoughts on whether I should mark this test as pending as well? 🤔

mstruve:
Thanks for the clarification!

>  we agreed that this should probably be renamed to open in the future to avoid future confusion (because it's definitely a bit confusing!), but for now, it's a bit out of scope for the task at hand.

It seems like switching to using open would be easy since we haven't even really implemented this feature yet or are there are difficulties that come along with that?

jacobherrington:
@mstruve IIRC "valid" is used in a lot of places pertaining to Reactions, so it might be a little messy to change it to open.

juliannatetreault:
> Thanks for the clarification!
> 
> > we agreed that this should probably be renamed to open in the future to avoid future confusion (because it's definitely a bit confusing!), but for now, it's a bit out of scope for the task at hand.
> 
> It seems like switching to using open would be easy since we haven't even really implemented this feature yet or are there are difficulties that come along with that?

@mstruve: I'm not entirely sure of difficulties that come along with making the change to using "open" in place of "valid" (I think @jacobherrington might be right, though), but I did propose making a new `status` for this work when I first started scoping it, but my team and I agreed that this feature is being implemented for an edge case that rarely occurs and needs to get out the door to make it easier for support to handle the un-confirming of vomits. As I said, we recognize that this is confusing and do not disagree that it should be changed, but when I proposed changing it yesterday evening after seeing your comment, we again, came to the consensus that this should be dealt with as a separate task and PR! 😊

Issue #9719, opened 2020-08-10
Add a "spoiler tag" to optionally hide content


pkfrank:
**Is your feature request related to a problem? Please describe.**
As an author and commenter, I might want to hide content that can be optionally viewed when the viewer is ready.  Such as solutions to a problem, etc.

**Describe the solution you'd like**
As a writer — an intuitive, markdown-based ways to include Spoiler text and images in content.

As a reader — the spoiler content is initially hidden, and I have to click to view it.

**Additional context**
Reddit uses the `>!Spoiler Content!<` syntax as seen [here](https://www.reddit.com/wiki/commenting#wiki_posting).

CC @maestromac based on the comment [here](https://github.com/forem/forem/issues/3785#issuecomment-525417377)

rhymes:
Possibly related to https://github.com/forem/forem/issues/947 as well, see the last sentence in the description:

>  This could also be exposed using a {% spoiler %} liquid tag.

Issue #947, opened 2018-10-18
Allow the <details> tag in posts and comments


bigblind:
**Is your feature request related to a problem? Please describe.**
In posts that contain questions for the reader to answer, or challenges for them to solve, it's useful to be able to hide the answer.

**Describe the solution you'd like**
I'd like to be able to use the `<details>` and <summary> tag to hide certain elements in a post. For more information on these tags, see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details).

An example usage would be:

```html
<details>
    <summary>What does it mean for a method to be idempotent?</summary>
    It means that calling the method repeatedly with the same input has the same result as calling it once. The next calls don't change the state of the application.
</details>
```

GitHub already supports them, so here's what that looks like:

<details>
    <summary>What does it mean for a method to be idempotent?</summary>
    It means that calling the method repeatedly with the same input has the same result as calling it once. The next calls don't change the state of the application.
</details>

**Describe alternatives you've considered**
This could also be exposed using a `{% spoiler %}` liquid tag.

bigblind:
Looking at the browser support chart on MDN, we'd need a polyfill for Edge.

maestromac:
Hey @bigblind, thank you for this issue! This is up for grab if anyone wants to give it a whirl!

Link2Twenty:
I guess we're thinking something like 

```
{% spoiler Spoiler header %}
  Spoiler content goes in here
{% endspoiler %}
```
becoming
```html
<details>
  <summary>Spoiler header</summary>
  Spoiler content goes in here
</details>
```

rhymes:
unstale

blunket:
**I'd like to work on this one!!!** :raised_hands:

One thought I have: is the actualy word `spoiler` really what we want to use for this tag? Maybe it's just me, but `spoiler` makes me think of blacked-out text that a user must interact with to reveal.

On Discord, for example:
![image](https://user-images.githubusercontent.com/17580512/65999064-3f36e180-e462-11e9-8061-08530e81efaa.png)

Or on [Stack Overflow](https://meta.stackexchange.com/questions/72877/whats-the-exact-syntax-for-spoiler-markup):
![image](https://user-images.githubusercontent.com/17580512/65999117-5d044680-e462-11e9-99d5-af762c44931e.png)

Maybe a more fitting tag name for `<details>` would be something like `collapse`, especially in case the above types of spoiler tags might be introduced later?

benhalpern:
[...]

PR #9777, opened 2020-08-13
Create details liquid tag

atsmith813:
- [x] Feature

## Description
This PR creates a new liquid tag that enables an author to include a `details` HTML element which helps to hide content behind a click.

This concept has been discussed in a few places and I started by registering the tag under `details`, `collapsible`, and `spoiler` so that any/all of these would work and do the same thing. If we want to separate the functionality, like make `spoiler` behave similarly to how [Reddit handles it](https://www.reddit.com/wiki/commenting#wiki_posting), let's decide that now so we can avoid having to re-save a bunch of articles in the future.

I figured since the default styling looks okay we can start there, see some real use cases in the wild, and then update the styling with the design team later.

While this liquid tag is bare-bones by putting text into a `details` and `summary` element, one of the advantages of implementing this functionality as a liquid tag is it gives the option to expand on it later (add custom JavaScript, styles, etc.) with less restriction. We could update this tag to be a fully custom implementation of hiding/showing content on click with crazy JavaScript animations. It would require a re-save to any existing articles using the tag and then it would "just work" :trollface:.

Closes https://github.com/forem/forem/issues/9719
Closes https://github.com/forem/forem/issues/947

1. Fire up the app locally.
2. Go to `/new` (aka **Write a post**).
3. Use the new liquid tag.
4. Click preview and/or publish to see it in action 🎉 .

```
{% details What is the best community platform? %}
Forem 🎉 
{% enddetails %}
```
**Documentation**
Screen Shot 2020-08-13 at 4 49 08 PM

**Using the liquid tag**
Screen Shot 2020-08-13 at 4 51 49 PM

**How it renders in an article by default**
Screen Shot 2020-08-13 at 4 51 54 PM

**How it looks when clicked**
Screen Shot 2020-08-13 at 4 52 05 PM

_Note: this feature uses [the `details` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) and is not supported by Internet Explorer._
## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed - updated liquid tag documentation which could _definitely_ use some feedback to make it more concise 🙈 

details_gif

### DIFF
--- a/app/liquid_tags/details_tag.rb
+++ b/app/liquid_tags/details_tag.rb
@@ -1,0 +1,25 @@
+class DetailsTag < Liquid::Block
+  PARTIAL = "liquids/details".freeze
+
+  def initialize(_tag_name, summary, _parse_context)
+    super
+    @summary = summary.strip
+  end
+
+  def render(_context)
+    content = Nokogiri::HTML.parse(super)
+    parsed_content = content.xpath("//html/body").text.strip
+
+    ActionController::Base.new.render_to_string(
+      partial: PARTIAL,
+      locals: {
+        summary: @summary,
+        parsed_content: parsed_content
+      },
+    )
+  end
+end
+
+Liquid::Template.register_tag("collapsible", DetailsTag)
+Liquid::Template.register_tag("details", DetailsTag)
+Liquid::Template.register_tag("spoiler", DetailsTag)
--- a/app/liquid_tags/details_tag.rb
+++ b/app/liquid_tags/details_tag.rb
@@ -1,14 +1,16 @@
 class DetailsTag < Liquid::Block
+  include ActionView::Helpers::SanitizeHelper
+
   PARTIAL = "liquids/details".freeze
 
   def initialize(_tag_name, summary, _parse_context)
     super
-    @summary = summary.strip
+    @summary = sanitize(summary.strip)
   end
 
   def render(_context)
     content = Nokogiri::HTML.parse(super)
-    parsed_content = content.xpath("//html/body").text.strip
+    parsed_content = sanitize(content.xpath("//html/body").inner_html)
 
     ActionController::Base.new.render_to_string(
       partial: PARTIAL,
--- a/app/views/pages/_editor_guide_text.html.erb
+++ b/app/views/pages/_editor_guide_text.html.erb
@@ -55,11 +55,11 @@
     </p>
     <h3><strong><%= community_name %> Article/Post Embed</strong></h3>
     <p>All you need is the full link of the article:</p>
-    <code>{% link https://dev.to/kazz/boost-your-productivity-using-markdown-1be %}</code>
+    <code>{% link <%= app_url("/kazz/boost-your-productivity-using-markdown-1be") %> %}</code>
     <p>You can also use the slug like this:</p>
     <code>{% link kazz/boost-your-productivity-using-markdown-1be %}</code>
     <p>You can also use the alias post instead of link like this:</p>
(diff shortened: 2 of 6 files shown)


app/views/pages/_editor_guide_text.html.erb
```diff
@@ -82,6 +82,15 @@
     <p>You can also use the category and slug like this:</p>
     <code>{% listing collabs/dev-is-open-source-823 %}</code>
     <p>Note: Expired listings will raise an error. Make sure the listing is published or recently bumped.</p>
+    <h3><strong>Details Embed</strong></h3>
+    <p>You can embed a details HTML element by using details, spoiler, or
+    collapsible. The <em>summary</em> will be what the dropdown title displays.
+    The <em>content</em> will be the text hidden behind the dropdown. This is
+    great for when you want to hide text (i.e. answers to questions) behind a
+    user action/intent (i.e. a click).</p>
```
mstruve, line 90: When I went to test this out locally I wrote some text and then put an image in the details tag and noticed the image will not load inside the tag. Should we mention something about that here?

atsmith813: @mstruve That's a great point 🤔 . I could see where images or code snippets would be super useful to include in the details tag. I just updated the code to support this. I don't _think_ it's a security risk since it falls into the same processing path?

app/liquid_tags/details_tag.rb
```diff
@@ -0,0 +1,27 @@
...
+  PARTIAL = "liquids/details".freeze
+
+  def initialize(_tag_name, summary, _parse_context)
+    super
+    @summary = sanitize(summary.strip)
+  end
+
+  def render(_context)
+    content = Nokogiri::HTML.parse(super)
+    parsed_content = sanitize(content.xpath("//html/body").inner_html)
```
atsmith813, line 13: I don't _think_ this is a security issue to use `inner_html`, but please let me know if I missed something here!

app/views/pages/_editor_guide_text.html.erb
mstruve: Hmmm looks like there are some security considerations we should think about here: https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML

> here are ways to execute JavaScript without using <script> elements, so there is still a security risk whenever you use innerHTML to set strings over which you have no control. For that reason, it is recommended that you do not use innerHTML when inserting plain text; instead, use Node.textContent. This doesn't parse the passed content as HTML, but instead inserts it as raw text. 

> Warning: If your project is one that will undergo any form of security review, using innerHTML most likely will result in your code being rejected. For example, if you use innerHTML in a browser extension and submit the extension to addons.mozilla.org, it will not pass the automated review process.

atsmith813: I was thinking something similar 🤔 .

Do you happen to know how doing this inside the liquid tag is different than a user typing the same malicious content outside of it?

Either way, the content goes through several steps of sanitization before it's in its final form. Is this liquid tag doing anything different that opens up the risk more?

mstruve: I think we are good here bc we sanitize it before hand. I was trying to trigger the exploit but could not get it to work

app/liquid_tags/details_tag.rb
citizen428: Maybe @justinlarson can weigh in on this?

maestromac: It won't be because when you throw it into `ActionController::Base.new.render_to_string`, the content will be escaped

PR #9788, opened 2020-08-14
Allow canonical urls that have not been taken by a published article

### LINKED ISSUE #9759: False positives with error: "canonical_url: has already been taken"
**Describe the bug**

Occasionally users try to input a canonical url for their post and get hit with the following warning: 
`Whoops, something went wrong: canonical_url: has already been taken`

This warning is meant to block people from reposting an article that was already posted, but it's currently causing issues. The main problem is that the warning appears even if the article that first used the canonical is only in draft mode (unpublished), which occasionally causes issues for folks in the same org.

This generally occurs because organizations don't currently have a way to share RSS, though there is a request for this - https://github.com/forem/forem/issues/4373 / https://github.com/forem/forem/issues/3363. We occasionally end up with a situation where a user in an org will set up the feature Publishing from RSS (https://dev.to/settings/publishing-from-rss) and input the RSS for their org blog, loading in all of their org member's posts into their dashboard. When loading in posts this way, the canonical URL for these posts is set automatically. Often, the person loading in posts this way will just leave the drafts that aren't theirs unpublished in their user dashboard. This is how we end up with drafted posts blocking canonicals.

**Expected behavior**

One short term fix would be to make it so that unpublished posts can't block canonical_urls. I can't think of any downsides to this.

In the long-term, I think we should also allow members of the same org to have access to a shared RSS ( https://github.com/forem/forem/issues/4373 / https://github.com/forem/forem/issues/3363)

**Screenshots**

Here's a screenshot of the warning

Screenshot 2020-08-10 at 17 19 47

**Additional context**

This could be a separate request, but I think we should also consider adapting the error message to offer more info for the recipient. 

The current message says:

`Whoops, something went wrong: canonical_url: has already been taken`

But it doesn't provide any info about which post used the canonical url. If it linked to the post instead:

`Whoops, something went wrong: The canonical_url you designated is already in use on the following post https://dev.to/michaeltharrington/1234. If you think there is a mistake, please email yo@dev.to.`

... then the user could potentially help us to triage an issue where one person is claiming the canonical URL for another person's post.

eliblocks:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Replace uniqueness validation with custom method to only check published articles for the uniqueness of a new article canonical url

Closes #9759 

1. Create a draft article with a canonical url
2. Create another article with the same canonical url
3. The second article should update without any errors
4. Update the first article to be published.
5. Make any update to the second article. Now it should show the error "Canonical url has already been taken"

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

alt_text

Note: A potential disadvantage of this change is that draft  articles that were valid when saved can become invalid once another article is published using the same canonical url.

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -52,10 +52,9 @@ class Article < ApplicationRecord
                     length: { maximum: 128 }
   validates :user_id, presence: true
   validates :feed_source_url, uniqueness: { allow_blank: true }
-  validates :canonical_url,
-            url: { allow_blank: true, no_local: true, schemes: %w[https http] }
-  # uniqueness: { scope: :published, allow_blank: true }
-  validate :canonical_url_must_be_unique
+  validates :canonical_url, url: { allow_blank: true, no_local: true, schemes: %w[https http] }
+  validates :canonical_url, uniqueness: { allow_blank: true },
+                            if: -> { Article.published.exists?(canonical_url: canonical_url) }
   validates :body_markdown, length: { minimum: 0, allow_nil: false }, uniqueness: { scope: %i[user_id title] }
   validate :validate_tag
   validate :validate_video
@@ -566,12 +565,6 @@ class Article < ApplicationRecord
     return unless canonical_url.to_s.match?(/[[:space:]]/)
 
     errors.add(:canonical_url, "must not have spaces")
-  end
-
-  def canonical_url_must_be_unique
-    return unless canonical_url && Article.published.exists?(canonical_url: canonical_url)
-
-    errors.add(:canonical_url, "has already been taken")
   end
 
   def create_slug
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -53,8 +53,9 @@ class Article < ApplicationRecord
   validates :user_id, presence: true
   validates :feed_source_url, uniqueness: { allow_blank: true }
   validates :canonical_url,
-            url: { allow_blank: true, no_local: true, schemes: %w[https http] },
-            uniqueness: { allow_blank: true }
+            url: { allow_blank: true, no_local: true, schemes: %w[https http] }
+  # uniqueness: { scope: :published, allow_blank: true }
+  validate :canonical_url_must_be_unique
   validates :body_markdown, length: { minimum: 0, allow_nil: false }, uniqueness: { scope: %i[user_id title] }
   validate :validate_tag
   validate :validate_video
@@ -565,6 +566,12 @@ class Article < ApplicationRecord
     return unless canonical_url.to_s.match?(/[[:space:]]/)
 
     errors.add(:canonical_url, "must not have spaces")
+  end
+
+  def canonical_url_must_be_unique
+    return unless canonical_url && Article.published.exists?(canonical_url: canonical_url)
+
+    errors.add(:canonical_url, "has already been taken")
   end
 
   def create_slug
--- a/spec/models/article_spec.rb
+++ b/spec/models/article_spec.rb
@@ -14,7 +14,6 @@ RSpec.describe Article, type: :model do
   it_behaves_like "UserSubscriptionSourceable"
 
   describe "validations" do
-    it { is_expected.to validate_uniqueness_of(:canonical_url).allow_blank }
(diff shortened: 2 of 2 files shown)


eliblocks:
Thanks @rhymes I see your point! I refactored it and the tests are passing.

spec/models/article_spec.rb
```diff
@@ -136,6 +136,31 @@ def build_and_validate_article(*args)
...
+        another_article.canonical_url = url
+
+        expect(another_article).not_to be_valid
+      end
+
+      it "is valid when the url has been taken by a draft article" do
+        another_article = build(:article)
+        another_article.canonical_url = url
+
+        expect(another_article).to be_valid
```
rhymes, line 160: I think there's a bug in this test now that I look at it closely. The article's still published and in memory. To create a draft article I think you have to use `create(:article, published: false)`

eliblocks: You're right. I fixed the tests so that articles are created correctly, and the spec fails without the fix.

app/models/article.rb
```diff
@@ -52,9 +52,9 @@ class Article < ApplicationRecord
                     length: { maximum: 128 }
   validates :user_id, presence: true
   validates :feed_source_url, uniqueness: { allow_blank: true }
-  validates :canonical_url,
-            url: { allow_blank: true, no_local: true, schemes: %w[https http] },
-            uniqueness: { allow_blank: true }
+  validates :canonical_url, url: { allow_blank: true, no_local: true, schemes: %w[https http] }
+  validates :canonical_url, uniqueness: { allow_blank: true },
+                            if: -> { Article.published.exists?(canonical_url: canonical_url) }
```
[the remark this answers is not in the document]
citizen428: An index on `articles(published, canonical_url)` would be covering this query, so even if we hit the DB, it'd be an index-only scan, not reading rows from disk. Additionally, it would also be used by the `published` scope since the `published` column comes first in the index definition.

In production the query plan for `select * from articles where published = true and canonical_url = '...';` currently looks like this (URL removed for brevity):

```
Seq Scan on articles  (cost=0.00..78699.79 rows=1 width=2199) (actual time=113.317..402.869 rows=1 loops=1)
  Filter: (published AND ((canonical_url)::text = '...'::text))
  Rows Removed by Filter: 361680
Planning Time: 2.820 ms
Execution Time: 402.924 ms
```

Experimenting locally with the index leads to a much friendlier plan:

```
 Limit  (cost=0.14..5.91 rows=1 width=4) (actual time=0.081..0.082 rows=1 loops=1)
   ->  Index Only Scan using articles_published_canonical_url_idx on articles  (cost=0.14..5.91 rows=1 width=4) (actual time=0.080..0.080 rows=1 loops=1)
         Index Cond: ((published = true) AND (canonical_url = '...'::text))
         Heap Fetches: 1
 Planning Time: 0.595 ms
 Execution Time: 0.118 ms
```
Of course, the amount of data is very different, but it shows that this index turns the whole condition into an access predicate for the query planner.

I think we can consider adding this, no matter if we only validate published articles or drafts too.

spec/models/article_spec.rb
```diff
@@ -136,6 +135,32 @@ def build_and_validate_article(*args)
...
+      end
+
+      it "is not valid" do
+        body_markdown = "---\ntitle: Title\ncanonical_url: #{url}\npublished: true\n---\n\n"
+        create(:article, body_markdown: body_markdown)
+        another_article = build(:article, body_markdown: body_markdown)
+
+        expect(another_article).not_to be_valid
```
citizen428, line 152: I prefer being somewhat more explicit in my assertions (i.e. checking for a specific error condition), so this test can't accidentally remain green.

rhymes:
Hi @eliblocks, in the midst of our DB cleanup we ended up adding a unique index (and paired validator) on all `canonical_url`(s) - see https://github.com/forem/forem/pull/10161 - right after cleaning up all duplicates in the DB - see https://github.com/forem/forem/pull/10138 - thus I don't think we should go ahead with this PR, sorry for the mishap

PR #9789, opened 2020-08-14
Feature:Use Sidekiq to Run Scheduled Jobs

mstruve:
- [x] Feature

## Description
[The sidekiq-cron gem](https://github.com/ondrejbartas/sidekiq-cron) eliminates the need for us to include additional infrastructure to run scheduled jobs. Instead this allows us to do it through Sidekiq using Sidekiq's polling mechanism. You can read more about the [under the hood details here](https://github.com/ondrejbartas/sidekiq-cron#under-the-hood).

Another gem that was considered was [sidekiq-scheduler](https://github.com/Moove-it/sidekiq-scheduler). While the scheduler gem came with some fancy features like dynamic schedules I decided to go with sidekiq-cron because it was much more maintained. Plus, the extra fluff sidekiq-scheduler has are not things we really _need_.

Once this is merged and I have confirmed all things are working properly in production I will start porting over the remaining scheduled jobs. 

TODO: Remove log_worker_queue_stats from Heorku Scheduler. 

https://github.com/orgs/forem/projects/6#card-43680382

- pull down the branch
- restart your app
- navigate to http://localhost:3000/sidekiq/cron and see the first scheduled cron job

## Added tests?

alt_text


config/schedule.yml
```diff
@@ -0,0 +1,3 @@
+log_worker_queue_stats:
+  cron: "*/10 * * * *"
```
maestromac, line 2: This is every 10 minute?

mstruve: Yep! Matches our scheduler 
Screen Shot 2020-08-14 at 2 43 40 PM

PR #9793, opened 2020-08-14
Reduce arguments method buff! at model buffer_update

brunnohenrique:

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Refactoring the model `buffer_update.rb` to reduce the arguments in the method `buff!`

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/app/labor/bufferizer.rb
+++ b/app/labor/bufferizer.rb
@@ -21,7 +21,7 @@ class Bufferizer
       if text.length < 250 && SiteConfig.twitter_hashtag
         text = text.gsub(" #{SiteConfig.twitter_hashtag}", " #{SiteConfig.twitter_hashtag} ##{tag.name}")
       end
-      BufferUpdate.buff!(@article.id, text, tag.buffer_profile_id_code, { tag_id: tag.id, admin_id: @admin_id })
+      BufferUpdate.buff!(@article.id, text, tag.buffer_profile_id_code, tag_id: tag.id, admin_id: @admin_id)
     end
     @article.update(last_buffered: Time.current)
   end
@@ -34,9 +34,9 @@ class Bufferizer
 
   def facebook_post!
     BufferUpdate.buff!(@article.id, fb_buffer_text, ApplicationConfig["BUFFER_FACEBOOK_ID"],
-                       { social_service_name: "facebook", admin_id: @admin_id })
+                       social_service_name: "facebook", admin_id: @admin_id)
     BufferUpdate.buff!(@article.id, fb_buffer_text + social_tags, ApplicationConfig["BUFFER_LINKEDIN_ID"],
-                       { social_service_name: "linkedin", admin_id: @admin_id })
+                       social_service_name: "linkedin", admin_id: @admin_id)
     @article.update(facebook_last_buffered: Time.current)
   end
 
--- a/app/labor/bufferizer.rb
+++ b/app/labor/bufferizer.rb
@@ -21,22 +21,21 @@ class Bufferizer
       if text.length < 250 && SiteConfig.twitter_hashtag
         text = text.gsub(" #{SiteConfig.twitter_hashtag}", " #{SiteConfig.twitter_hashtag} ##{tag.name}")
       end
-      BufferUpdate.buff!(@article.id, text, tag.buffer_profile_id_code, tag_id: tag.id, admin_id: @admin_id)
+      BufferUpdate.buff!(@article.id, text, buffer_profile_id_code: tag.buffer_profile_id_code, tag_id: tag.id,
+                                            admin_id: @admin_id)
     end
     @article.update(last_buffered: Time.current)
   end
 
   def main_tweet!
-    BufferUpdate.buff!(@article.id, twitter_buffer_text, ApplicationConfig["BUFFER_TWITTER_ID"],
-                       { admin_id: @admin_id })
+    BufferUpdate.buff!(@article.id, twitter_buffer_text, admin_id: @admin_id)
     @article.update(last_buffered: Time.current)
   end
 
   def facebook_post!
-    BufferUpdate.buff!(@article.id, fb_buffer_text, ApplicationConfig["BUFFER_FACEBOOK_ID"],
-                       social_service_name: "facebook", admin_id: @admin_id)
-    BufferUpdate.buff!(@article.id, fb_buffer_text + social_tags, ApplicationConfig["BUFFER_LINKEDIN_ID"],
-                       social_service_name: "linkedin", admin_id: @admin_id)
+    BufferUpdate.buff!(@article.id, fb_buffer_text, social_service_name: "facebook", admin_id: @admin_id)
+    BufferUpdate.buff!(@article.id, fb_buffer_text + social_tags, social_service_name: "linkedin", admin_id: @admin_id)
+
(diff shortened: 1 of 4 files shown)


app/labor/bufferizer.rb
```diff
@@ -21,21 +21,22 @@ def satellite_tweet!
       if text.length < 250 && SiteConfig.twitter_hashtag
         text = text.gsub(" #{SiteConfig.twitter_hashtag}", " #{SiteConfig.twitter_hashtag} ##{tag.name}")
       end
-      BufferUpdate.buff!(@article.id, text, tag.buffer_profile_id_code, "twitter", tag.id, @admin_id)
+      BufferUpdate.buff!(@article.id, text, tag.buffer_profile_id_code, { tag_id: tag.id, admin_id: @admin_id })
```
rhymes, line 24: It's odd that rubocop didn't flag the parenthesis here as they are redudant 🤔

app/models/buffer_update.rb
```diff
@@ -5,9 +5,11 @@ class BufferUpdate < ApplicationRecord
   validate :validate_body_text_recent_uniqueness, :validate_suggestion_limit
   validates :status, inclusion: { in: %w[pending sent_direct confirmed dismissed] }
 
-  def self.buff!(
-    article_id, text, buffer_profile_id_code, social_service_name = "twitter", tag_id = nil, admin_id = nil
-  )
+  def self.buff!(article_id, text, buffer_profile_id_code, **options)
```
rhymes, line 8: can we call this `**kwargs` for consistency?

app/models/buffer_update.rb
```diff
@@ -5,9 +5,11 @@ class BufferUpdate < ApplicationRecord
...
+    social_service_name = options.fetch(:social_service_name, "twitter")
+    tag_id = options.fetch(:tag_id, nil)
+    admin_id = options.fetch(:admin_id, nil)
```
rhymes, line 11: I don't think you need to use fetch here. `kwargs[:tag_id]` is perfectly fine. As it's going to be nil anyway if the value is not given

app/labor/bufferizer.rb
wenderjean, line 24: The `tag` object itself has been used twice at the same calling, it suggests to me that you could just pass the object itself for the `buff!` method.

app/labor/bufferizer.rb
```diff
@@ -21,21 +21,22 @@ def satellite_tweet!
...
     end
     @article.update(last_buffered: Time.current)
   end
 
   def main_tweet!
-    BufferUpdate.buff!(@article.id, twitter_buffer_text, ApplicationConfig["BUFFER_TWITTER_ID"], "twitter", nil,
-                       @admin_id)
+    BufferUpdate.buff!(@article.id, twitter_buffer_text, ApplicationConfig["BUFFER_TWITTER_ID"],
```
wenderjean, line 30: Considering you're using the `buff!` method on different places I believe you could consider something like:

```ruby
class Bufferizer
  ...
  def facebook_post!
    buffer_params = build_buffer_params(text: fb_buffer_text, social_service_name: "facebook")

    BufferUpdate.buff!(buffer_params)
  end

  def build_buffer_params(text:, social_service_name:)
    BufferUpdate::Params.new(
      @article, @tag, @admin, text, social_service_name
    )
  end
  ...
end
```

app/labor/bufferizer.rb
```diff
@@ -21,21 +21,22 @@ def satellite_tweet!
...
+                       { admin_id: @admin_id })
     @article.update(last_buffered: Time.current)
   end
 
   def facebook_post!
-    BufferUpdate.buff!(@article.id, fb_buffer_text, ApplicationConfig["BUFFER_FACEBOOK_ID"], "facebook", nil, @admin_id)
-    BufferUpdate.buff!(@article.id, fb_buffer_text + social_tags, ApplicationConfig["BUFFER_LINKEDIN_ID"], "linkedin",
-                       nil, @admin_id)
+    BufferUpdate.buff!(@article.id, fb_buffer_text, ApplicationConfig["BUFFER_FACEBOOK_ID"],
```
wenderjean, line 36: Following my last suggestion, you could create on `ApplicationConfig` an accessor to return that ID using the `social_service_name` and extract it directly inside de `BufferUpdate.buff!`

app/models/buffer_update.rb
```diff
@@ -1,13 +1,24 @@
 class BufferUpdate < ApplicationRecord
+  SOCIAL_SERVICE = {
```
rhymes, line 2: should this be plural?

app/models/buffer_update.rb
```diff
@@ -54,6 +65,10 @@ def self.twitter_default_text(article)
     ].compact.join.strip
   end
 
+  def self.wich_social_service(service_name)
+    SOCIAL_SERVICE.fetch(service_name)
+  end
```
rhymes, line 70: side from the typo, I don't think we need the method at all, you can simply call `SOCIAL_SERVICES[social_service_name]` in the `buff!` method

app/models/buffer_update.rb
brunnohenrique: How about `SOCIAL_MEDIA`? My next commit will change it.

rhymes: sure :-)

app/labor/bufferizer.rb
brunnohenrique: I've tried to do a `params_builder` but I think it was too complex. It seems like that:

```rb
  def build_buffer_params(text, social_service_name:, tag:)
    @article.id,
    text,
    Hash.new
      .merge({social_service_name: social_service_name}) if social_service_name
      .merge({buffer_profile_id_code: tag.buffer_profile_id_code}) if tag && tag.buffer_profile_id_code
      .merge({tag_id: tag.id}) if tag
      .merge({admin_id: @admin.id}) if @admin
  end
```

What did you think about it?

brunnohenrique:
@rhymes @wenderjean thanks a lot for the reviews :champagne: 
If there is something else to do, just ping me.

app/models/buffer_update.rb
```diff
@@ -1,11 +1,11 @@
 class BufferUpdate < ApplicationRecord
-  SOCIAL_SERVICE = {
+  SOCIAL_MEDIA = {
     "twitter" => ApplicationConfig["BUFFER_TWITTER_ID"],
     "facebook" => ApplicationConfig["BUFFER_FACEBOOK_ID"],
     "linkedin" => ApplicationConfig["BUFFER_LINKEDIN_ID"]
   }.freeze
 
-  DEFAULT_SOCIAL_SERVICE = "twitter".freeze
+  DEFAULT_SOCIAL_MIDIA = "twitter".freeze
```
rhymes, line 8: @brunnohenrique there's a small typo here :-)

brunnohenrique: adjusted with `git push -f` to don't to pollute the commits

juliannatetreault:
TIL `**kwargs`! Thanks for taking the time to refactor this - it all looks good to me! 👍 👍

PR #9801, opened 2020-08-15
🐞Bug Fix: Chat doesn't extend to its frame on Firefox iOS [ Fixes #4024 ]

### LINKED ISSUE #4024: Chat doesn't extend to its frame on Firefox iOS
**Describe the bug**

DEV Connect doesn't occupy the entire window frame on Firefox iOS

**To Reproduce**

1. Go to https://dev.to/connect
2. Notice how the chat doesn't fully extend to the entire page, like it had a margin bottom preventing it to fully use the browser window. 

This doesn't happen on Firefox for Android 

**Expected behavior**

Expect the chat to not be restricted to the upper part of the frame 

**Screenshots**

118F7848-E86F-4BBC-BF90-C27EF3C660F7

- Device: iPhone 8
- OS: iOS 12.4.1
- Browser: Firefox 
- Version: 19 build 16042

sarthology:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
So after doing more research into it. I found few things, `navigator.userAgent.match(/iPhone/i) && !navigator.userAgent.match('CriOS')` this logic as added for IOS app which have this tabs
image

But user agent in case of iOS app is custom and equal to "DEV-Native-ios" that's how I discover that this request never fulfill and classes never got added. So When I removed it everything as working fine.  
Fixes [#4024](https://github.com/forem/forem/issues/4024)
Screen Recording 2020-08-15 at 04 08 PM gif
image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

alt_text

### DIFF
--- a/app/assets/stylesheets/chat.scss
+++ b/app/assets/stylesheets/chat.scss
@@ -297,7 +297,7 @@
   @media screen and (max-width: 426px) {
     padding: 10px;
     min-width: 50%;
-    width: calc(100% - 16px);
+    width: calc(100%);
     position: fixed;
     height: 100%;
     margin-left: 0;
--- a/app/assets/stylesheets/chat.scss
+++ b/app/assets/stylesheets/chat.scss
@@ -12,13 +12,7 @@
 }
 
 .live-chat {
-  height: calc(100vh - 56px);
-
-  &.live-chat--iossafari {
-    height: calc(100vh - 190px);
-  }
-
-  overflow-y: hidden;
+  height: calc(100 * var(--vh) - var(--header-height));
 }
 
 .live-chat {
@@ -28,14 +22,10 @@
 
 .chat {
   display: flex;
-  height: calc(100vh - var(--header-height));
+  height: calc(100 * var(--vh) - var(--header-height));
   position: relative;
   overflow-x: hidden;
   overflow-y: hidden;
-
-  &.chat--iossafari {
-    height: calc(100vh - 201px);
-  }
 }
 
 .chat__channels {
--- a/app/javascript/chat/chat.jsx
+++ b/app/javascript/chat/chat.jsx
@@ -1830,11 +1830,6 @@ export default class Chat extends Component {
 
   render() {
     const { state } = this;
-    const detectIOSSafariClass =
-      navigator.userAgent.match(/iPhone/i) &&
-      !navigator.userAgent.match('CriOS')
-        ? ' chat--iossafari'
-        : '';
     let channelHeader = <div className="active-channel__header">&nbsp;</div>;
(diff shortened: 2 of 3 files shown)


sarthology:
image
Also removed this ☝🏼

fdoxyz:
Thank you for this fix @sarthology! This seems to make things work great on Firefox browser and it actually improves the UX in others. 

There's a small problem on both Safari & Chrome with both the input field & top navigation bar not visible when scrolling on iOS only, Android seems solid. Maybe this has something to do with this change can be addressed too now that this is being fixed?

I'm leaving GIFs (may seen a bit slow FPS when converted to GIF) on all three of them, if you have an idea why this is happening it would be great. Otherwise this can be worked on separately

## Firefox iOS (looks good 💯 )

connect_firefox

## Chrome iOS (scroll issue)

connect_chrome

## Safari iOS (scroll issue)

connect_safari_480

sarthology:
@fdoxyz Thanks for pointing this out, so apparently this is part of bigger issue regarding `vh` in mobiles. They consider the entire mobile screen as `vh` so for us using innerheight is better than `vh`. Just added a hack that appears to be working fine. Please let me know if it works now 😅

fdoxyz:
I see, it's odd that Firefox browser on iOS works great but Safari and Chrome have this weird behavior. On Android things work great as well. I'm adding another GIF to show what's going on (sorry slow FPS GIF please be patient 😅)

I remember we encountered a similar issue when working with Codeland's sidecar @lisasy do you have any suggestions or ideas how we can get these heights to match? Maybe @ludwiczakpawel if you have any ideas would be greatly appreciated!

connect_sliding_problem_na480

sarthology:
@fdoxyz btw I fixed it.

fdoxyz:
The last GIF I uploaded was taken using the most recent commit `0b3dff0`. I think the height of the content is matching the height of the actual viewport, so that's fixed compared to the previous iOS Safari GIF, but it's scrolling and leaving that bottom space most notably when typing and sending a message.

At the very end of the GIF loop when scrolling up and down, the Connect top navigation bar shows/hides underneath the main top navigation. Not sure if this can be fixed in place (the second scroll that causes the `<` button to hide away from sight).

For now I just wanted to ask if anyone has an idea for how to get rid of that scroll, which would make the experience throughout iOS browsers really great (specially since Safari is the default and popular one).

sarthology:
This should solve the problem @fdoxyz 😅

PR #9834, opened 2020-08-17
Create optimized_image_tag helper

maestromac:
- [x] Refactor
- [x] Feature

## Description
Created a helper method to replace the use of `cl_image_tag`.

Shoutout to @citizen428 's for his suggestion.
n/a
Browse around locally (including podcast page) to make sure image are loading properly.

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

n/a


app/views/articles/_bottom_content.html.erb
```diff
@@ -7,17 +7,8 @@
...
-                           crop: "imagga_scale",
-                           width: 100,
-                           height: 100,
-                           quality: "auto",
-                           flags: "progressive",
-                           fetch_format: "auto",
-                           loading: "lazy",
-                           sign_url: true,
-                           alt: "#{article.cached_user.username} profile image") %>
+          <% image_url = Images::Optimizer.call(article.cached_user.profile_image_url, crop: "imagga_scale", width: 100, height: 100) %>
```
citizen428, line 10: **suggestion (non-blocking):** I wonder if it could make sense to wrap this up in a combined helper, something like

```ruby
def optimized_image_tag(url, optimizer_options: {}, image_options: {})
  image_option[:width] ||= optimizer_options[:width]
  image_option[:height] ||= optimizer_options[:height]
  image_url = Images::Optimizer.call(url, optimizer_options)

  image_tag(image_url, image_options)
end
```

Which would be called like this:

```ruby
<%= optimized_image_url(article.cached_user.profile_image_url, 
                        optimizer_options: { crop: "imagga_scale", width: 100, height: 100 },
                        image_options: { loading: "lazy", alt: "#{article.cached_user.username} profile image" })
%>
```

(This is untested, just meant to illustrate the idea)

maestromac: 🤯  Thank you so much for this suggestion! 

This is exactly what I wanted. I was making things harder for myself because I thought we shouldn't be using more than than one `option` hash as args. Updating now.

app/liquid_tags/podcast_tag.rb
```diff
@@ -33,7 +33,7 @@ def initialize(_tag_name, link, _parse_context)
   end
 
   def render(_context)
-    ActionController::Base.new.render_to_string(
+    ApplicationController.render(
```
maestromac, line 36: Had to make this change to this portion can have access to helpers. I'm going to make this change to the rest of liquid tag in a different PR.

app/helpers/application_helper.rb
```diff
@@ -82,6 +82,14 @@ def cloudinary(url, width = "500", quality = 80, format = "auto")
     Images::Optimizer.call(SimpleIDN.to_ascii(image_url), width: width, quality: quality, fetch_format: format)
   end
 
+  def optimized_image_tag(image_url, optimizer_options: {}, image_options: {})
+    image_options[:width] ||= optimizer_options[:width]
```
mstruve, line 86: Whats the point of having `width` and `height` in the  optimizer_options and then transferring them to the image_options? Why not set them in the image_options to begin with?

This also seems pretty similar to the method above in terms of its function. Any reason not to combine them so we have a single method handling `Images::Optimizer.call` in all the views? Then if we update the Optimizer we only have to change it in one place.

maestromac: Width and Height will be used by both the optimizer (`cl_image_path`) and the `image_tag`. Technically the value will be used by the optimizer first that's why I did it this way. I can see the scenario where we provide a different W/H for `image_option` and/or not provide W/H to `optimizer_options` (see note below). Perhaps I can change `image_options` to `image_tag_options` so it's more clear?

The method above, `#cloudinary`, is quite similar to barebone `Images::Optimizer` and I was putting it aside when I was consolidating `cl_image_tag. I can do that refactor in a different PR. I know it's already in the codebase, but Is it a bit weird to be calling service object from the view?

misc note: `cl_image_tag` will only take width and height into account when `crop` isn't nil but these values will always be applied to the image tag's property. We don't have `crop: nil` anywhere so that doesn't affect us.

mstruve: > I can see the scenario where we provide a different W/H for image_option and/or not provide W/H to optimizer_options

Gotcha, thanks for the explanation! 

> but Is it a bit weird to be calling service object from the view?

I wouldn't call the optimizer service from the view but I would consolidate the helper methods that call it into one single helper method so its all consolidated. But like you mentioned a good refactor PR once you iron out all the functionality you need.

citizen428: > This also seems pretty similar to the method above in terms of its function. Any reason not to combine them 

When I suggested this helper in [this comment](https://github.com/forem/forem/pull/9834#discussion_r471910795), I only used the context of this PR, so didn't notice there's a related helper around. Sometimes you just get stuck in a local maximum 😉 But 👍 for combining them in a refactor PR.

app/liquid_tags/podcast_tag.rb
citizen428: **suggestion (non-blocking):** I think `ApplicationController.render` reads nicer than `ActionController::Base.new.render_to_string`, but if you don't want to change all Liquid tags, you should be able to `include ApplicationHelper` in the base class (`LiquidTagBase`) and be done with it. Just to give you an option for the refactor PR.

citizen428 (approved):
@maestromac Based on the discussions in this PR, I thought about the signature of the helper a bit: since the width and height are reused, maybe we can pull them out of the options hashes altogether:

```ruby
def optimized_image_tag(image_url, width:, height: width, optimizer_options: {}, tag_options: {})
```

This defaults `height` to `width` so you don't have to specify both for square images and also renames `image_options` to `tag_options`, which I think is clear enough given the helper is called `optimized_image_tag`. Just some input for the potential refactoring PR you mentioned.

PR #9836, opened 2020-08-17
Refactor:Move Yearly Badge Awarder to Sidekiq Cron

mstruve:
- [x] Refactor
- [x] Feature

## Description
This moves the yearly badge awarding method into a Sidekiq Cron job using the already existing `BadgeAchievements::BadgeAwardWorker` which I tweaked. I could make a separate worker for each badge or a different worker for "predefined" badges but I was not sure what to call it and keeping all the badge rewarding in one place seemed like a better idea. 

Rather than doing all badge awarding in a single cron I decided to split them out. This will make the jobs run faster which is what Sidekiq is really made for and it will make them more resilient. If we break something for one badge the rest won't fail bc of it. 

Let me know what you all think of this approach! If you approve then I will continue with the rest of the badges. If not we can discuss and tweak things. 

- Pull down branch
- Restart app
- visit http://localhost:3000/sidekiq/cron and you should see the new job
- If you want to run it create a One Year Club badge then click "Enqueue Now" on the Sidekiq Cron dashboard and you should see the work run in your logs. You will also see it fail on the two year club badge.  

## Added tests?

- [x] yes

alt_text

### DIFF
--- a/app/labor/badge_rewarder.rb
+++ b/app/labor/badge_rewarder.rb
@@ -1,5 +1,5 @@
 module BadgeRewarder
-  YEARS = { 1 => "one", 2 => "two", 3 => "three" }.freeze
+  YEARS = { 1 => "one", 2 => "two", 3 => "three", 4 => "four", 5 => "five", 6 => "six", 7 => "seven" }.freeze
   REPOSITORIES = ["thepracticaldev/dev.to", "thepracticaldev/DEV-ios", "thepracticaldev/DEV-Android"].freeze
 
   LONGEST_STREAK_WEEKS = 16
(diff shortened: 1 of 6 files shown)


app/labor/badge_rewarder.rb
```diff
@@ -1,5 +1,5 @@
 module BadgeRewarder
-  YEARS = { 1 => "one", 2 => "two", 3 => "three" }.freeze
+  YEARS = { 1 => "one", 2 => "two", 3 => "three", 4 => "four", 5 => "five", 6 => "six", 7 => "seven" }.freeze
```
mstruve, line 2: Decided to expand the size of this array. Ideally we probably want a better way to generate this. One possibility is [a gem like this numbers_in_words](https://github.com/markburns/numbers_in_words) gem but I figured that was outside the scope of this PR.

app/labor/badge_rewarder.rb
```diff
@@ -10,12 +10,13 @@ module BadgeRewarder
   MINIMUM_QUALITY = -25
 
   def self.award_yearly_club_badges
-    (1..3).each do |i|
+    total_years = Time.current.year - ApplicationConfig["COMMUNITY_COPYRIGHT_START_YEAR"].to_i
```
mstruve, line 13: Decided while I was here to make this a little more dynamic

app/labor/badge_rewarder.rb
benhalpern: This is great. I've been meaning to add four in.

This change buys us at least three years before we need to consider something new. 😋

app/labor/badge_rewarder.rb
benhalpern: Looks good to me. This is notorious off-by-one territory so I had to count really carefully. 😄

PR #9837, opened 2020-08-17
Social Auth generalization refactor

fdoxyz:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR takes the Generalization refactor from #7952 into it's own separate PR in order to simplify the complexity in an effort to ship new Authentication Providers.

Kudos to @rhymes for the original implementation over in #7952 of the generalization logic, this PR just cherry picks from there.

#7952

This PR should have no effect in the authentication capabilities from master (GitHub + Twitter OAuth integrations). It only refactors the way Authentication Providers are added, so in order to test this I'd recommend:

1. Boot up locally with GitHub & Twitter Auth ENV variables setup
1. Check `/admin/config` inside `All Site Configuration` > `Authentication` to make sure both GitHub & Twitter providers are enabled
1. Make sure both Authentication providers continue to work as expected

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Must check and be sure both Authentication Providers (Twitter + GitHub) are enabled & configured correctly in `/admin/config`

step one


app/models/user.rb
```diff
@@ -146,14 +145,25 @@ class User < ApplicationRecord
...
   validates :username, uniqueness: { case_sensitive: false }, if: :username_changed?
   validates :website_url, :employer_url, url: { allow_blank: true, no_local: true }
   validates :website_url, length: { maximum: 100 }, allow_nil: true
   validates :youtube_url, length: { maximum: 1000 }, format: YOUTUBE_URL_REGEXP, allow_blank: true
 
+  # add validators for provider related usernames
+  Authentication::Providers.username_fields.each do |username_field|
+    # make sure usernames are not empty string, to be able to use the database unique index
+    clean_provider_username = proc do |record|
+      cleaned_username = record.attributes[username_field.to_s].presence
```
mstruve, line 159: This is 🔥

app/services/authentication/authenticator.rb
```diff
@@ -143,14 +144,26 @@ def update_profile_updated_at(user)
...
       range = 1.week.ago.beginning_of_day..Time.current
       range.cover?(user_identity_age)
     end
 
+    def extract_created_at_from_payload(logged_in_identity)
+      raw_info = logged_in_identity.auth_data_dump.extra.raw_info
+
+      if raw_info.created_at.present?
+        Time.zone.parse(raw_info.created_at)
+      elsif raw_info.auth_time.present?
```
mstruve, line 160: Did we find some other edge cases that caused these additional booleans here?

app/services/broadcasts/welcome_notification/generator.rb
```diff
@@ -92,7 +92,7 @@ def commented_on_welcome_thread?
       end
 
       def authenticated_with_all_providers?
-        identities.count == SiteConfig.authentication_providers.count
+        identities.count == Authentication::Providers.enabled.size
```
mstruve, line 95: 🎉

app/services/authentication/authenticator.rb
fdoxyz: I don't think so, as in I never ran into another odd scenario. In this case the `else` handles when Providers (Apple is the one I can recall) don't provide the complete data in their payload which then uses the `Time.current` default

rhymes:
Awesome! and thanks for cherry-picking the commits. 😭

PR #9842, opened 2020-08-18
Rubocop: Enable and fix Style/SingleArgumentDig

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Enabling Rubocop 0.89's `Style/SingleArgumentDig` to replace `hash.dig(:foo)` with `hash[:foo]` which is the most common syntax for single key access

- #9699
- #9709

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -486,7 +486,7 @@ Style/Send:
 
 Style/SingleArgumentDig:
   Description: 'Avoid using single argument dig method.'
-  Enabled: pending
+  Enabled: true
 
 Style/SingleLineBlockParams:
   Description: 'Enforces the names of some block params.'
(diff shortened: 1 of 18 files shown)


rhymes:
@mstruve ahha got it. I see your point. The reason why I prefer `hash[]` is because it's the default method to access hashes in lots of languages, Ruby included. 

Fun fact `dig()` calls `hash[]` if the number of keys is one https://github.com/ruby/ruby/blob/a0c7c23c9cec0d0ffcba012279cd652d28ad5bf3/hash.c#L4396-L4404

PR #9843, opened 2020-08-18
Automatically refresh profile accessors

citizen428:

- [X] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR adds two service objects, `ProfileFields::Add` and `ProfileFields::Remove` which take care of adding and removing profile fields and keeping the corresponding store accessors in sync. For now we do not actually modify any data when a profile field is removed, I'm still spiking ideas how to do this without killing our DB. I also updated `Admin::ProfileFieldsController` to make use of the new service objects.

Related to #6994

1. Use the service objects in a Rails console session and ensure that the store accessors automatically update.
2. Use the admin page to add/remove profile fields and ensure that the store accessors automatically update.

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -2,13 +2,13 @@ class Profile < ApplicationRecord
   belongs_to :user
 
   # This method generates typed accessors for all active profile fields
-  def self.define_store_accessors!
+  def self.refresh_store_accessors!
     ProfileField.active.find_each do |field|
       store_attribute :data, field.attribute_name, field.type
     end
   end
 
-  define_store_accessors!
+  refresh_store_accessors!
 
   validates :data, presence: true
 end
--- a/app/services/profile_fields/remove.rb
+++ b/app/services/profile_fields/remove.rb
@@ -1,0 +1,32 @@
+# This service removes a profile field and ensures that the corresponding store
+# accessor also gets removed immediately.
+module ProfileFields
+  class Remove
+    def self.call(id)
+      new(id).call
+    end
+
+    attr_reader :profile_field, :error_message
+
+    def initialize(id)
+      @id = id
+      @success = false
+    end
+
+    def call
+      @profile_field = ProfileField.find(@id)
+      if @profile_field.destroy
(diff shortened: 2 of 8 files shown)


app/models/profile.rb
```diff
@@ -2,13 +2,13 @@ class Profile < ApplicationRecord
   belongs_to :user
 
   # This method generates typed accessors for all active profile fields
-  def self.define_store_accessors!
+  def self.refresh_store_accessors!
```
citizen428, line 5: This is just a better name.

app/services/profile_fields/remove.rb
```diff
@@ -0,0 +1,32 @@
...
+    def initialize(id)
+      @id = id
+      @success = false
+    end
+
+    def call
+      @profile_field = ProfileField.find(@id)
+      if @profile_field.destroy
+        accessor = profile_field.attribute_name.to_s
+        Profile.undef_method(accessor) if accessor.in?(Profile.stored_attributes[:data])
```
Ridhwana, line 20: **Question**: TIL about undef_method. Do we call it instead of remove_method because we don't want it to search superclasses and mixed-in modules? and undef_method, no matter where it exists, will just not respond to this accessor.

Ridhwana (approved):
🔥

app/services/profile_fields/remove.rb
citizen428: Oh, no particular reason. `undef_method` was the one I remembered. Remove method is semantically more correct though, so I'll change it. Thanks!

citizen428: Hm, changing this actually made the existing specs fails. Since I'm not concerned about mixings or superclasses here, I'll just change it back to `undef_method`. Just FYI @Ridhwana.

rhymes (approved):
Works great!

PR #9846, opened 2020-08-18
Add relations and foreign keys between HtmlVariant related models

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR is step 2 of 2, step 1 is https://github.com/forem/forem/pull/9845 which has been merged and deployed first.

It contains foreign keys and depedent relations between `HtmlVariant` related models (successes, trails, articles)

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -41,6 +41,8 @@ class Article < ApplicationRecord
            class_name: "Comment"
   has_many :profile_pins, as: :pinnable, inverse_of: :pinnable
   has_many :buffer_updates, dependent: :destroy
+  has_many :html_variant_successes, dependent: :nullify
+  has_many :html_variant_trials, dependent: :nullify
   has_many :notifications, as: :notifiable, inverse_of: :notifiable, dependent: :delete_all
   has_many :notification_subscriptions, as: :notifiable, inverse_of: :notifiable, dependent: :destroy
   has_many :rating_votes
--- a/app/models/html_variant.rb
+++ b/app/models/html_variant.rb
@@ -1,15 +1,17 @@
 class HtmlVariant < ApplicationRecord
   GROUP_NAMES = %w[article_show_below_article_cta badge_landing_page campaign].freeze
 
+  belongs_to :user, optional: true
+
+  has_many :html_variant_successes, dependent: :destroy
+  has_many :html_variant_trials, dependent: :destroy
+
+  validates :group, inclusion: { in: GROUP_NAMES }
   validates :html, presence: true
   validates :name, uniqueness: true
-  validates :group, inclusion: { in: GROUP_NAMES }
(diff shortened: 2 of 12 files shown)


app/models/article.rb
```diff
@@ -41,6 +41,8 @@ class Article < ApplicationRecord
            class_name: "Comment"
   has_many :profile_pins, as: :pinnable, inverse_of: :pinnable
   has_many :buffer_updates, dependent: :destroy
+  has_many :html_variant_successes, dependent: :nullify
+  has_many :html_variant_trials, dependent: :nullify
```
rhymes, line 45: As the relation is optional between those two, I thought it best to nullify the key. Let me know if they should be destroyed instead

app/models/html_variant.rb
```diff
@@ -1,15 +1,17 @@
 class HtmlVariant < ApplicationRecord
   GROUP_NAMES = %w[article_show_below_article_cta badge_landing_page campaign].freeze
 
+  belongs_to :user, optional: true
+
+  has_many :html_variant_successes, dependent: :destroy
+  has_many :html_variant_trials, dependent: :destroy
```
rhymes, line 7: Added `dependent: destroy` here, the rest is reordering

app/models/user.rb
```diff
@@ -62,6 +62,8 @@ class User < ApplicationRecord
                            foreign_key: :resource_owner_id, inverse_of: :resource_owner, dependent: :delete_all
   has_many :affected_feedback_messages, class_name: "FeedbackMessage",
                                         inverse_of: :affected, foreign_key: :affected_id, dependent: :nullify
+  has_many :ahoy_events, class_name: "Ahoy::Event", dependent: :destroy
+  has_many :ahoy_visits, class_name: "Ahoy::Visit", dependent: :destroy
```
rhymes, line 66: I snuck this in as I forgot to add the inverse relation for ahoy models previously

db/migrate/20200818101637_add_missing_foreign_keys_to_html_variants_models.rb
```diff
@@ -0,0 +1,9 @@
+class AddMissingForeignKeysToHtmlVariantsModels < ActiveRecord::Migration[6.0]
+  def change
+    add_foreign_key :html_variant_successes, :html_variants, column: :html_variant_id, on_delete: :cascade, validate: false
+    add_foreign_key :html_variant_successes, :articles, column: :article_id, on_delete: :nullify, validate: false
```
rhymes, line 4: html variant success to variant itself is enforced in cascade mode
html variant success to article is enforced in nullify mode

same goes for trials below

Issue #3682, opened 2019-08-11
Chat - Sound Effect When Messaging


omenking:
**Is your feature request related to a problem? Please describe.**

When someone messages I have to manually check chat which causes a delay in conversation.

**Describe the solution you'd like**

- [ ] When someone notifies me I should hear a sound.
- [ ] I should have a toggle to globally turn on and off notification sounds

**Describe alternatives you've considered**

**Additional context**

jessleenyc:
This makes sense to me, and we definitely want to have the user setting available to opt-in/out of right away 😅.

Figuring out a sound would be fun 😁.

theomelo:
@sarthology @jessleenyc I assume this PR fixes the issue? Do you think it satisfies as a solution?

cmgorton:
It looks like this may be addressed in the above PR if not feel free to open up a discussion around this feature in [forem.dev](forem.dev) Andrew!

 There we can get more eyes on potential features and highlight any we really would love the community to contribute too. Thanks!

PR #9847, opened 2020-08-18
🚀 Feature: Give Tab Notification on New Message

sarthology:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Going through old issues I found this [#3682](https://github.com/forem/forem/issues/3682). Thought this will be an interesting issue to work on. So when new message comes, your tab will  alert you that you received a message and if you are using chrome or Firefox then you will receive sound notification as well. Having few Permission issues in Safari as added this in draft so that someone can suggest what to do.
Will add following things in upcoming days.
- Add ability to disable sound notification in settings.
- Add ability to see to messaged new in tab itself.
- Add this notify feature on other pages as well.

ezgif com-resize

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

alt_text


benhalpern:
This is pretty cool! Probably makes sense for the notification sound to be default off until we're confident in the whole thing and then make it easy to point people to how they might modify the setting.

theomelo:
@sarthology do you think this PR can be promoted out of Draft? Do you think there's still work to do on it?

config/webpacker.yml
```diff
@@ -31,6 +31,7 @@ default: &default
     - .ttf
     - .woff
     - .woff2
+    - .mp3
```
citizen428, line 34: Do we need this? It seems like you are not using mp3s for now.

config/webpacker.yml
```diff
@@ -48,6 +48,7 @@ default: &default
     - .gif
     - .jpeg
     - .jpg
+    - .mp3
```
citizen428, line 51: I guess it also needs to be removed here, no?

narender2031: @citizen428 Done

citizen428 (approved):

citizen428:
I re-ran the specs for you, they're green now.

app/javascript/utilities/connect/index.js
```diff
@@ -1,2 +1,3 @@
 export * from './getUnopenedChannels';
 export * from './pusher';
+export * from './newMessageNotify';
```
nickytonline, line 3: Refactor (blocking):  the `./newMessageNotify` import only exports a default function, I do not see the need to include this in this [barrel file](https://basarat.gitbook.io/typescript/main-1/barrel).

Consider having a named import in `'./newMessageNotify` if you want to have it in the barrel file.

Also, I do not see this being used anywhere, e.g. `import notifyUser from 'utilities/connect';`

nickytonline (changes requested):
A small change request and also do existing tests cover these changes? If not, please add some.

app/javascript/utilities/connect/index.js
sarthology: Done

app/javascript/utilities/connect/newMessageNotify.js
```diff
@@ -0,0 +1,19 @@
+export default function notifyUser() {
+  modifyTitle();
+  // May need it's own big feature
```
nickytonline, line 3: refactor (blocking): If this is dead code, please remove it.

app/javascript/utilities/connect/newMessageNotify.js
```diff
@@ -0,0 +1,19 @@
...
+  // const audio = new Audio('../../../assets/sound/notification.mp3');
+  // audio.play();
+}
+
+const modifyTitle = () => {
```
nickytonline, line 8: I'm not sure I follow the logic here.

- Every 2 seconds, check if the title of the document is the initial value of the document title and if so change it to "New Message 👋🏻", otherwise set it back to the original title.
- And let that all run for 12 seconds and then stop it.

Any reason we do 12 seconds and also, why do we stop? Don't we want to keep notifying users that they have new messages if new ones come in?

sarthology: > Any reason we do 12 seconds and also, why do we stop? Don't we want to keep notifying users that they have new messages if new ones come in?

So 12 sec only because I thought it will be too disturbing to keep doing it. It's like it's shouting for your attention even if I have seen it. If there is a new message the notification will start again for 12 sec. But if you think that we should keep it going that's also fine.

nickytonline: Ahh, that's the part I misunderstood. I didn't realize it was 12 seconds for only when a new message comes in. In that case, instead of hardcoding `12000` milliseconds, consider making it a constant explaining what it is.

nickytonline:
Just restarting the build @sarthology.

nickytonline:
@sarthology, I see there is an mp3 file in the PR. Where does it get referenced? And how is it silenced by default as @benhalpern mentions?

citizen428:
I think the MP3 is no longer used @nickytonline, can someone please confirm and remove it from the PR, thanks!

sarthology:
Correct ! Not need of it for this PR, but that Sound Notification is something that is in the roadmap.

sarthology:
@nickytonline @citizen428 Removed !

nickytonline (approved):
Thanks for removing the mp3! LGTM.

nickytonline:
One of the build nodes failed. I just restarted it.

PR #9853, opened 2020-08-18
Rubocop: Fix ListingCategory has_many relation to Listing

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Currently the relation between `ListingCategory` has two issues:

- it does not work because it's not able to infer the name of the foreign key, you can verify it this by loading the console in master:

```ruby
category = ListingCategory.first
category.listings # this doesn't return the array of listings
```

- it doesn't define a `dependent` clause. I chose `:restrict_with_error` because removing a category should fail if there are listings assigned to that category.

The foreign key is already in place so no need to add that.

```ruby
category = ListingCategory.first
category.destroy
puts category.errors
```

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/models/listing_category.rb
+++ b/app/models/listing_category.rb
@@ -3,7 +3,9 @@ class ListingCategory < ApplicationRecord
   # We standardized on the latter, but keeping the table name was easier.
   self.table_name = "classified_listing_categories"
 
-  has_many :listings, inverse_of: :listing_category
+  has_many :listings, inverse_of: :listing_category,
+                      foreign_key: :classified_listing_category_id,
+                      dependent: :restrict_with_error
 
   before_validation :normalize_social_preview_color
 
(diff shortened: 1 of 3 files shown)


rhymes:
@fdoxyz added an empty commit to restart it

app/models/listing_category.rb
```diff
@@ -3,7 +3,9 @@ class ListingCategory < ApplicationRecord
   # We standardized on the latter, but keeping the table name was easier.
   self.table_name = "classified_listing_categories"
 
-  has_many :listings, inverse_of: :listing_category
+  has_many :listings, inverse_of: :listing_category,
```
citizen428, line 6: Thanks @rhymes! That we didn't notice this until now seems to indicate that we never use this direction of the association.

PR #9857, opened 2020-08-18
Only bust_fastly_cache if fastly is enabled

vaidehijoshi:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [ ] no documentation needed

alt_text


mstruve:
Don't forget the Purgeable module! I just noticed that is now the point of failure for the jobs on forem.dev

vaidehijoshi:
> Don't forget the Purgeable module! I just noticed that is now the point of failure for the jobs on forem.dev

Oh I gotchu @mstruve. It now checks to see if Fastly is configured, and will either return a service when you call `fastly` on the module/concern, or it'll return `false` 🙌

app/models/concerns/purgeable.rb
```diff
@@ -23,13 +24,14 @@ def table_key
     end
 
     def fastly
-      return if Rails.env.development?
+      return false if Rails.env.development? ||
+        !(ApplicationConfig["FASTLY_API_KEY"].present? || ApplicationConfig["FASTLY_SERVICE_ID"].present?)
```
mstruve, line 28: This might be a little easier to read if you broke it up into 2 lines and used `.blank?` instead of `!(.present?)` 
```ruby
return false if Rails.env.development?
return false if ApplicationConfig["FASTLY_API_KEY"].blank? || ApplicationConfig["FASTLY_SERVICE_ID"].blank?
```

app/labor/cache_buster.rb
```diff
@@ -19,17 +22,55 @@ def self.bust(path)
...
+    )
+  end
+
+  def self.bust_nginx_cache(path)
+    uri = URI.parse("#{ApplicationConfig['OPENRESTY_PROTOCOL']}#{ApplicationConfig['OPENRESTY_DOMAIN']}#{path}")
+    http = Net::HTTP.new(uri.host, uri.port)
+    response = http.request Net::HTTP::NginxPurge.new(uri.request_uri)
+
+    # Explicitly use `fail` here since we are not catching and re-raising the exception.
+    fail Error, response.body unless response.is_a?(Net::HTTPSuccess) # rubocop:disable Style/SignalException
```
mstruve, line 69: TIL about fail!
Does Honeybadger deal with this the same way it does raised exceptions?

app/labor/cache_buster.rb
```diff
@@ -19,17 +22,55 @@ def self.bust(path)
...
-    HTTParty.post("https://api.fastly.com/purge/https://#{ApplicationConfig['APP_DOMAIN']}#{path}",
-                  headers: { "Fastly-Key" => ApplicationConfig["FASTLY_API_KEY"] })
-    HTTParty.post("https://api.fastly.com/purge/https://#{ApplicationConfig['APP_DOMAIN']}#{path}?i=i",
-                  headers: { "Fastly-Key" => ApplicationConfig["FASTLY_API_KEY"] })
+    if fastly_enabled?
+      bust_fastly_cache(path)
+    elsif nginx_enabled?
+      bust_nginx_cache(path)
+    else
+      raise ConfigurationError, "You cannot bust a cache without a caching service! Please enable a caching service."
```
mstruve, line 31: What about Forem's that are not using any caching? I think we need to be able to silently exit here without any errors for those Forems choosing not to configure any caching beyond the rails app.

.env_sample
```diff
@@ -2,6 +2,10 @@
 export APP_DOMAIN="localhost:3000"
 export APP_PROTOCOL="http://"
 
+# Openresty domain + Protocol setting for development
+export OPENRESTY_DOMAIN="localhost:9090"
```
mstruve, line 6: Oh hey we can add ENV variables and not have to REQUIRE them 🎉

app/labor/cache_buster.rb
vaidehijoshi, line 31: Let me know if this feels too aggressive; my thought here was that we don't want to try to _ever_ call `bust` on a cache if we don't have _some_ kind of caching service configured, but maybe I'm wrong about that. 🤷

app/labor/cache_buster.rb
```diff
@@ -206,3 +247,12 @@ def self.bust_article_comment(commentable)
...
 end
+
+# Creates our own purge method for an HTTP request,
+# which is used by Nginx to bust a cache.
+# See Net::HTTPGenericRequest for attributes/methods.
+class Net::HTTP::NginxPurge < Net::HTTPRequest # rubocop:disable Style/ClassAndModuleChildren
+  METHOD = "PURGE".freeze
+  REQUEST_HAS_BODY = false
+  RESPONSE_HAS_BODY = true
+end
```
vaidehijoshi, line 258: We can't use `HTTParty` here the way that we do in `bust_fastly_cache` because `PURGE` is not supported by that client (nor is it defined in the HTTP RFCs, apparently!!).

Took a cue from [`fastly-ruby`](https://github.com/fastly/fastly-ruby), which basically creates their [own version of a `PURGE` Net::HTTPRequest class](https://github.com/fastly/fastly-ruby/blob/159ae26e4fe41bcdef2fc8f45c25d7dd4f0a5cbe/lib/fastly/client.rb#L180-L185). I just named it differently since we know we're only ever going to use it for purging nginx 😉

spec/labor/cache_buster_spec.rb
```diff
@@ -11,6 +11,31 @@
...
+      allow(cache_buster).to receive(:bust_fastly_cache).and_call_original
+      allow(ApplicationConfig).to receive(:[]).with("APP_DOMAIN").and_return("fake-key")
+      allow(ApplicationConfig).to receive(:[]).with("FASTLY_API_KEY").and_return("fake-key")
+      allow(ApplicationConfig).to receive(:[]).with("FASTLY_SERVICE_ID").and_return("localhost:3000")
+    end
+
+    it "can bust a fastly cache when configured" do
+      cache_buster.bust_fastly_cache("/#{user.username}")
+    end
+  end
```
vaidehijoshi, line 37: Ideally, I would have liked to have just tested `CacheBuster.bust` directly here, but because we short-circuit out of that method when we're [not in production](https://github.com/forem/forem/pull/9857/files#diff-6e58540a6172a3c1054d1ede7134dd00R24), I decided to test the individual caching service methods directly. 🤷

app/labor/cache_buster.rb
vaidehijoshi: @mstruve yeah, I was wondering whether this will be too aggressive, see this comment:
https://github.com/forem/forem/pull/9857#discussion_r474330025

vaidehijoshi: OK, I think actually exiting silently makes sense; I didn't realize that `CacheBuster.bust` would ever get called implicitly (as a side-effect of some other action), but now that I'm looking at the code again, it looks like we call that from tons of places and very basic controller actions. So what you're saying now makes a lot more sense, I'll update that.

app/labor/cache_buster.rb
vaidehijoshi: @mstruve I believe that `fail`, internally, [will `raise` a `RuntimeError`](https://apidock.com/ruby/Kernel/fail), so I would think that if Honeybadger _should_ be able to deal with it? (I couldn't find anything specific in the Honeybadger docs about any exceptions around `fail`).

To be 100% honest, I saw this usage [happening in `fastly-ruby`](https://github.com/fastly/fastly-ruby/blob/a2c5f2cf3247a07ee9c39fa7115ab8281e4b6239/lib/fastly/client.rb#L115), which is actually how _I_ learned about it! I hadn't heard of it or used it before either, and it seems like people are pretty divided on whether or not to use it. I kept it in because Fastly does that, and also because of [this](https://avdi.codes/jim-weirich-on-exceptions) blog post. But on the other hand, Rubocop does [_not_ seem to like or recommend this](https://github.com/rubocop-hq/ruby-style-guide/issues/512), so maybe we should just be `raise`ing here instead?

I'm down to change it if that's what people prefer, I kinda just threw it in here after seeing Fastly doing it 😛

mstruve: Oh, thanks for sharing those resources and that blog post! I think for the sake of code cohesion we should use raise here. I could see a younger dev coming in and seeing this and mimic it again with another fail and before you know it we have 50/50 fails and raises sprinkled throughout the codebase. Not a blocker but that is the way I would lean.

vaidehijoshi: > I could see a younger dev coming in and seeing this and mimic it again with another fail and before you know it we have 50/50 fails and raises sprinkled throughout the codebase.

Hm, I think that this is reason enough to convince me to use that instead 👍 Have switched this from `fail` to `raise` in the latest commit, which Honeybadger should catch for us should Nginx go down. Also threw in a nicer error message since we are `raise`ing the error ourselves now.

app/labor/cache_buster.rb
```diff
@@ -19,17 +19,50 @@ def self.bust(path)
...
     DatadogStatsClient.increment("cache_buster.invalid_uri", tags: ["path:#{path}"])
   end
 
+  def self.fastly_enabled?
+    ApplicationConfig["FASTLY_API_KEY"].present? && ApplicationConfig["FASTLY_SERVICE_ID"].present?
+  end
+
+  def self.nginx_enabled?
+    ApplicationConfig["OPENRESTY_PROTOCOL"].present? && ApplicationConfig["OPENRESTY_DOMAIN"].present?
+  end
```
rhymes, line 39: these two probably deserved to be surfaced somewhere global so that you can both here and in Purgeable check `Cache.fastly_enabled?`

Definitely not blocking

app/labor/cache_buster.rb
rhymes: I initially thought one could do extend the HTTParty module with an initializer:

by creating `config/initializers/httparty.rb`:

```ruby
class Net::HTTP::Purge < Net::HTTPRequest # rubocop:disable Style/ClassAndModuleChildren
  METHOD = "PURGE".freeze
  REQUEST_HAS_BODY = false
  RESPONSE_HAS_BODY = true
end

module HTTPartyPurge
  extend ActiveSupport::Concern

  class_methods do
    def purge(path, options = {}, &block)
      perform_request Net::HTTP::Purge, path, options, &block
    end
  end
end

HTTParty.send(:include, HTTPartyPurge)
```

and then you could do:

```ruby
HTTParty.purge("https://dev.to/api")
```

unfortunately the `perform_request` method is private, not sure if there's a way to call it anyway:

https://github.com/jnunemaker/httparty/blob/0d808c7f3a68b8d091506e22003fc00dd84bf6a6/lib/httparty.rb#L556-L561

In the meantime, could we at least put the `Net::HTTP::Purge` class a `net_http.rb` initializer ?

All non blocking obviously :D

app/labor/cache_buster.rb
```diff
@@ -206,3 +239,12 @@ def self.bust_article_comment(commentable)
     end
   end
 end
+
+# Creates our own purge method for an HTTP request,
+# which is used by Nginx to bust a cache.
+# See Net::HTTPGenericRequest for attributes/methods.
+class Net::HTTP::NginxPurge < Net::HTTPRequest # rubocop:disable Style/ClassAndModuleChildren
```
rhymes, line 246: is there a reason why it's called `NginxPurge` and not just Purge? after all the HTTP method is `PURGE` and this can become a generic extension to be used anywhere that method is supported

PR #9862, opened 2020-08-18
Set Stripe dynamically instead of at boot

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This is an example of setting stripe from dynamically, so the keys can be pulled from a storage source instead of `ENV`.... We can do this in several other places as well in order to move away from dependency on boot-time logic which requires us to restart the app if a user were to configure it.

An admin can now go into _their own_ stripe account and put the keys into their forem.... Which really is the way things should be. If we want to modify how storage is done in the future, we could, but the way this all works for now makes sense. I think we can keep plopping more into `SiteConfig` until we're done getting stuff out of `ENV`.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -128,10 +128,6 @@ class ApplicationController < ActionController::Base
   end
 
   def initialize_stripe
-    Rails.configuration.stripe = {
-      publishable_key: "Set dynamically instead of once at boot."
-    }
-
-    Stripe.api_key = "Also set dynamically instead of once at boot."
+    Stripe.api_key = SiteConfig.stripe_api_key
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -126,4 +126,12 @@ class ApplicationController < ActionController::Base
   def api_action?
     self.class.to_s.start_with?("Api::")
   end
+
+  def initialize_stripe
+    Rails.configuration.stripe = {
+      publishable_key: "Set dynamically instead of once at boot."
+    }
+
+    Stripe.api_key = "Also set dynamically instead of once at boot."
+  end
 end
--- a/app/controllers/credits_controller.rb
+++ b/app/controllers/credits_controller.rb
@@ -1,5 +1,6 @@
 class CreditsController < ApplicationController
   before_action :authenticate_user!
+  before_action :initialize_stripe
 
   def index
     @user_unspent_credits_count = current_user.credits.unspent.size
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -149,7 +149,7 @@ module Constants
         placeholder: "sk_live_...."
       },
       stripe_publishable_key: {
-        description: "Public stripe key for receiving payments. " \
+        description: "Public Stripe key for receiving payments. " \
         "See: https://stripe.com/docs/keys",
         placeholder: "pk_live_...."
       },
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -131,6 +131,16 @@ module Constants
         "See: https://github.com/thepracticaldev/dev.to/pull/6345",
         placeholder: "$pay.somethinglikethis.co/value"
       },
+      stripe_api_key: {
+        description: "Secret stripe key for receiving payments. " \
+        "See: https://stripe.com/docs/keys",
+        placeholder: "sk_live_...."
+      },
+      stripe_publishable_key: {
+        description: "Public stripe key for receiving payments. " \
+        "See: https://stripe.com/docs/keys",
+        placeholder: "pk_live_...."
+      },
       mailchimp_newsletter_id: {
         description: "Main Newsletter ID",
         placeholder: ""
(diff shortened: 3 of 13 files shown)


app/controllers/application_controller.rb
```diff
@@ -126,4 +126,12 @@ def anonymous_user
   def api_action?
     self.class.to_s.start_with?("Api::")
   end
+
+  def initialize_stripe
+    Rails.configuration.stripe = {
```
citizen428, line 131: If I understand the point of this PoC correctly, we probably want to make sure that both of these operations are thread-safe.

benhalpern: I decided to just remove this one, because it's not really needed. It's just a value that can be plopped straight into the view.

app/controllers/credits_controller.rb
```diff
@@ -1,5 +1,6 @@
 class CreditsController < ApplicationController
   before_action :authenticate_user!
+  before_action :initialize_stripe
```
benhalpern, line 3: Putting these in the controller works fine, but it'd probably be more DRY and reliable in the `Stripe::` object or whatever.... I'm just not sure what the most ideal situation is.

Ridhwana (approved):

app/controllers/application_controller.rb
```diff
@@ -126,4 +126,8 @@ def anonymous_user
   def api_action?
     self.class.to_s.start_with?("Api::")
   end
+
+  def initialize_stripe
+    Stripe.api_key = SiteConfig.stripe_api_key
+  end
```
rhymes, line 132: I wonder if this could be extracted in a concern in https://github.com/forem/forem/tree/master/app/controllers/concerns, not blocking though

app/lib/constants/site_config.rb
```diff
@@ -131,6 +131,16 @@ module SiteConfig
         "See: https://github.com/thepracticaldev/dev.to/pull/6345",
         placeholder: "$pay.somethinglikethis.co/value"
       },
+      stripe_api_key: {
+        description: "Secret stripe key for receiving payments. " \
```
rhymes, line 135: Stripe should be capitalized

app/lib/constants/site_config.rb
```diff
@@ -131,6 +131,16 @@ module SiteConfig
...
+        "See: https://stripe.com/docs/keys",
+        placeholder: "sk_live_...."
+      },
+      stripe_publishable_key: {
+        description: "Public stripe key for receiving payments. " \
```
rhymes, line 140: Stripe should be capitalized

app/controllers/credits_controller.rb
vaidehijoshi: I think this is pretty Rails-y and okay for now, but the next time we come back to Stripe code, I think extracting it into a Stripe service that can initialize itself when it is created might be a nice approach. That way, all you have to do is call on `Stripe::Service` (or whatever we call it), and it'll initialize itself whenever it's being invoked.

But again, not necessary until we actually are in this code again, IMO.

config/initializers/stripe.rb
```diff
@@ -1,9 +1,3 @@
-Rails.configuration.stripe = {
-  publishable_key: ApplicationConfig["STRIPE_PUBLISHABLE_KEY"]
-}
-
-Stripe.api_key = ApplicationConfig["STRIPE_SECRET_KEY"]
-
 if Rails.env.development? && Stripe.api_key.present?
```
rhymes, line 1: we should move this in `initialize_stripe`, as `Stripe.api_key` is not set here this code will never be executed

app/views/credits/new.html.erb
```diff
@@ -144,7 +144,7 @@
 </div>
   <script>
 
-    var stripe = Stripe('<%= ApplicationConfig["STRIPE_PUBLISHABLE_KEY"] %>');
+    var stripe = Stripe('<%= SiteConfig.stripe_publishable_key %>');
```
vaidehijoshi, line 147: If an admin hasn't set `SiteConfig.stripe_publishable_key` (since it's not required) , this JavaScript will go boom when try to `createCardElement`, right? Do we want to warn admins about this at all, or are we okay letting [this line](https://github.com/forem/forem/pull/9862/files#diff-919744afb1f282f44b4ebee18b262ef3R206) explode with a `can't call createToken on undefined` error?

benhalpern: Yeah, the clean way to do this is to ensure this path never gets hit if this doesn't get shipped. But this is essentially already broken so it's no less broken this way.

app/lib/constants/site_config.rb
benhalpern: Fixed

app/controllers/application_controller.rb
benhalpern: Yeah I think this is ideal, but I think I'll leave this for a follow-on PR to make sure it gets the proper testing etc. At least this is stupid simple by putting it directly in the traffic path.

app/lib/constants/site_config.rb
benhalpern: Fixed

PR #9863, opened 2020-08-18
Move community_copyright_start_year to SiteConfig

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
No reason the environment and our systems teem needs to know anything about this value.

jumanji

### DIFF
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -43,7 +43,7 @@ module Constants
       },
       community_copyright_start_year: {
         description: "Used to mark the year this forem was started.",
-        placeholder: "2020"
+        placeholder: Time.zone.today.year.to_s
       },
       tagline: {
         description: "Used in signup modal.",
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -41,6 +41,10 @@ module Constants
         description: "Used to determine the action of community e.g coding, reading, training etc.",
         placeholder: ""
       },
+      community_copyright_start_year: {
+        description: "Used to mark the year this forem was started.",
+        placeholder: "2020"
+      },
       tagline: {
         description: "Used in signup modal.",
         placeholder: "We're a place where coders share, stay up-to-date and grow their careers."
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -31,7 +31,7 @@ class SiteConfig < RailsSettings::Base
   field :community_action, type: :string
   field :tagline, type: :string
   field :community_copyright_start_year, type: :integer,
-    default: ApplicationConfig["COMMUNITY_COPYRIGHT_START_YEAR"] || 2020
+                                         default: ApplicationConfig["COMMUNITY_COPYRIGHT_START_YEAR"] || 2020
 
   # Emails
   field :email_addresses, type: :hash, default: {
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -31,7 +31,8 @@ class SiteConfig < RailsSettings::Base
   field :community_action, type: :string
   field :tagline, type: :string
   field :community_copyright_start_year, type: :integer,
-                                         default: ApplicationConfig["COMMUNITY_COPYRIGHT_START_YEAR"] || 2020
+                                         default: ApplicationConfig["COMMUNITY_COPYRIGHT_START_YEAR"] ||
+                                           Time.zone.today.year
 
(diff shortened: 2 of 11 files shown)


benhalpern:
@mstruve I think the answer is _generally_ yes, but I'd be 99% sure nobody currently live will be negatively affected by `2020` as a default. We'll have to get used to thinking in these terms more often, but here I think we'll be fine without.

benhalpern:
@mstruve on second thought _we_ will need to update this, and why leave it to a race condition when we could safely add it into the code.

I added it in as the default if available and in the future we can remove it if we feel like it.

app/models/site_config.rb
```diff
@@ -30,6 +30,8 @@ class SiteConfig < RailsSettings::Base
   field :community_member_label, type: :string, default: "user"
   field :community_action, type: :string
   field :tagline, type: :string
+  field :community_copyright_start_year, type: :integer,
+    default: ApplicationConfig["COMMUNITY_COPYRIGHT_START_YEAR"] || 2020
```
mstruve, line 34: 🔥

benhalpern:
Oh right... Added view code.

app/models/site_config.rb
```diff
@@ -30,6 +30,8 @@ class SiteConfig < RailsSettings::Base
   field :community_member_label, type: :string, default: "user"
   field :community_action, type: :string
   field :tagline, type: :string
+  field :community_copyright_start_year, type: :integer,
+                                         default: ApplicationConfig["COMMUNITY_COPYRIGHT_START_YEAR"] || 2020
```
citizen428, line 34: Same as above, maybe replace 2020 with `Date.today.year`?

app/lib/constants/site_config.rb
```diff
@@ -41,6 +41,10 @@ module SiteConfig
         description: "Used to determine the action of community e.g coding, reading, training etc.",
         placeholder: ""
       },
+      community_copyright_start_year: {
+        description: "Used to mark the year this forem was started.",
+        placeholder: "2020"
```
citizen428, line 46: Instead of hard-coding 2020, how about `Date.today.year.to_s`?

app/models/site_config.rb
citizen428: On second thought, the problem with this is always that if the users never update the value explicitly the default will be wrong eventually. If we hardcode 2020 and forget to update it on Jan 1st it will be in the past and if we use my approach it will automatically move forward.

app/views/admin/configs/show.html.erb
```diff
@@ -243,6 +243,15 @@
...
               <div class="alert alert-info">Used in signup modal.</div>
             </div>
+
+            <div class="form-group">
+              <%= admin_config_label :community_copyright_start_year %>
+              <%= f.text_field :community_copyright_start_year,
+                               class: "form-control",
+                               value: SiteConfig.community_copyright_start_year,
+                               placeholder: "2020" %>
+              <div class="alert alert-info">Used in the footer to note what year the forem started.</div>
```
Ridhwana, line 253: It looks like some of the optimizations using constants in this file either got reverted or just not committed on master at some point 🤷🏽‍♀️, but I'll look into this it in another PR.  

Anyways, you'll see in some parts of this file, e.g. https://github.com/forem/forem/blob/cc021307438a126b96d55fba2cb38207f87ec34b/app/views/admin/configs/show.html.erb#L598-L603 we're moving towards using Constants for the description and placeholder, which looks cleaner and makes it easier to reuse the descriptions across sections as we progress. 

So, in this case we would want to change the description to `Constants::SiteConfig::DETAILS[config_key][:description]` and placeholder to `Constants::SiteConfig::DETAILS[config_key][:placeholder]` and add it to this file https://github.com/forem/forem/blob/master/app/lib/constants/site_config.rb

Ridhwana (commented):
We should try and use the constants module for descriptions and placeholders

app/views/admin/configs/show.html.erb
benhalpern: Ah that makes sense.... I added it in for this PR, and we can fix this holistically in another.

app/models/site_config.rb
benhalpern: Using current year makes sense. This is just the starter value so it should indeed always be current.

Ridhwana (approved):

PR #9864, opened 2020-08-19
Add community_name to site config

benhalpern:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In an effort to minimize what systems needs to care about, let's move this variable to what is available to forem admins.

water ski

### DIFF
--- a/app/controllers/email_subscriptions_controller.rb
+++ b/app/controllers/email_subscriptions_controller.rb
@@ -1,5 +1,4 @@
 class EmailSubscriptionsController < ApplicationController
-
   def unsubscribe
     verified_params = Rails.application.message_verifier(:unsubscribe).verify(params[:ut])
 
--- a/app/controllers/email_subscriptions_controller.rb
+++ b/app/controllers/email_subscriptions_controller.rb
@@ -1,10 +1,10 @@
 class EmailSubscriptionsController < ApplicationController
   PREFERRED_EMAIL_NAME = {
-    email_digest_periodic: "#{ApplicationConfig['COMMUNITY_NAME']} digest emails",
+    email_digest_periodic: "#{SiteConfig.community_name} digest emails",
     email_comment_notifications: "comment notifications",
     email_follower_notifications: "follower notifications",
     email_mention_notifications: "mention notifications",
-    email_connect_messages: "#{ApplicationConfig['COMMUNITY_NAME']} connect messages",
+    email_connect_messages: "#{SiteConfig.community_name} connect messages",
     email_unread_notifications: "unread notifications",
     email_badge_notifications: "badge notifications"
   }.freeze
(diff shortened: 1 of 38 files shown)


app/controllers/email_subscriptions_controller.rb
```diff
@@ -1,10 +1,10 @@
 class EmailSubscriptionsController < ApplicationController
   PREFERRED_EMAIL_NAME = {
-    email_digest_periodic: "#{ApplicationConfig['COMMUNITY_NAME']} digest emails",
+    email_digest_periodic: "#{SiteConfig.community_name} digest emails",
```
benhalpern, line 3: I just did find and replace, but I suspect this may cause a problem in terms of `SiteConfig` being available yet here in this context, not entirely sure. Will check back on this tomorrow.

app/views/html_variants/_form.html.erb
```diff
@@ -20,7 +20,7 @@
     <%= f.text_field :name, placeholder: "Unique, descriptive name" %>
   <% end %>
   <%= f.label :html %>
-  <%= f.text_area :html, placeholder: "HTML to be shown. Make sure all CSS is properly scoped, and all HTML tags are closed, etc. Manditory field." %>
+  <%= f.text_area :html, placeholder: "HTML to be shown. Make sure all CSS is properly scoped, and all HTML tags are closed, etc. Mandatory field." %>
```
benhalpern, line 23: Unrelated typo fix.

app/controllers/email_subscriptions_controller.rb
citizen428: Once the app is initialized you should be ok.

benhalpern: It did indeed cause the error... `/home/travis/build/forem/forem/vendor/bundle/ruby/2.7.0/gems/rails-settings-cached-2.2.0/lib/rails-settings/base.rb:129:in `_all_settings': You cannot use settings before Rails initialize. (RuntimeError)`

I think this is because constants first get called before `SiteConfig` is initialized and available. So in these cases I needed to move it into a method.... It was just a couple instances where this was a problem.

app/views/moderations/_mod_sidebar_right.html.erb
```diff
@@ -3,7 +3,7 @@
     <div class="crayons-card__body mod-saying-hello">
       <p class="fw-medium fs-2xl">Hello! 👋</p>
       <p>
-        Thank you for helping to keep <%= ApplicationConfig["COMMUNITY_NAME"] %> safe! ❤️
+        Thank you for helping to keep <%= SiteConfig.community_name %> safe! ❤️
```
Ridhwana, line 6: **Comment (not-blocking):**This probably should have been using the method name community_name from the application helper. We probably missed it in the first pass.

Ridhwana (approved):
This looks good to me 🎉

app/controllers/email_subscriptions_controller.rb
Ridhwana: @benhalpern I don't fully understand what you did to solve this, can you clarify?

PR #9867, opened 2020-08-19
Use ApplicationController for liquid render

maestromac:
- [x] Refactor

## Description
I just found out that `Application.render` is a feature introduced in Rails 5 to [render erb outside of the request-response cycle](https://evilmartians.com/chronicles/new-feature-in-rails-5-render-views-outside-of-actions). Aside from being cleaner looking, it also provides view helpers without additional tinkering.
n/a
1. Start up server locally
2. Navigate to `/new`
3. Try any liquid tag and click preview/save.
4. Content should render normally.
## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a

### DIFF
--- a/app/liquid_tags/asciinema_tag.rb
+++ b/app/liquid_tags/asciinema_tag.rb
@@ -8,7 +8,7 @@ class AsciinemaTag < LiquidTagBase
   end
 
   def render(_context)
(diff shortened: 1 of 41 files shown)


rhymes (approved):
LGTM!

PR #9873, opened 2020-08-19
[Admin Tooling] Allow admins delete empty Connect Channels

msarit:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR allows admins to delete channels that have no users from `/admin/chat_channels`.

1. Navigate to `localhost:300/admin/chat_channels` as an `:admin` or `:super_admin` user
2. If there are channels with users, you should see the text _**"Cannot delete; channel has users"**_ under their "Delete Channel" column
3. To create a user-less channel, create a test channel at the top. Your created channel should appear in the list, and its `Delete Channel` button should be visible
4. Delete your test channel to observe functionality

https://www.loom.com/share/ea26f5ddfd394faaab7dcd56eb45d78e

## Added tests?

- [X] yes, and quite proud of them 😊 
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

Girl collapsing on floor


Zhao-Andy:
Looks like there are merge conflicts now 😅

app/views/admin/chat_channels/index.html.erb
```diff
@@ -29,34 +29,46 @@
...
+</div>
+<%= paginate @group_chat_channels %>
+<table class="table">
+  <thead>
+    <tr>
+      <th scope="col">Channel Name</th>
+      <th scope="col">Users</th>
+      <th scope="col">Add Users</th>
+      <th scope="col">Remove User</th>
+      <% if current_user.any_admin? %>
```
mstruve, line 41: I noticed this method is called twice here in this view. I went in a console to see if `has_role?` was doing any local caching and it was not. If we want to improve performance a little here we could add an instance variable for `any_admin?` in the user model like this

```ruby
  def any_admin?
    @any_admin ||= (has_role?(:super_admin) || has_role?(:admin))
  end
```

This would allow us to only hit the database once, instead of twice for this view. It would also improve the performance of a couple of other views I found that do the same. Totally outside the scope of this PR but I wanted to point it out bc I think it's a little peek into how an SRE approaches a PR vs someone looking for correctness. Let me know if you plan to do it, otherwise, I will probably throw up a quick PR for it.

app/controllers/admin/chat_channels_controller.rb
```diff
@@ -29,6 +29,17 @@ def remove_user
       redirect_back(fallback_location: "/admin/chat_channels")
     end
 
+    def destroy
+      @chat_channel = ChatChannel.find(params[:id])
+      if @chat_channel.users.count.zero?
+        @chat_channel.destroy
+        flash[:success] = "Channel was successfully deleted."
+      else
+        flash[:alert] = "Channel NOT deleted!"
```
mstruve, line 38: Should we add something about it not being able to be deleted bc it has users?

app/views/admin/chat_channels/index.html.erb
msarit: @mstruve Thanks for this Molly! I will do so 😄

app/models/user.rb
```diff
@@ -298,7 +298,8 @@ def admin?
   end
 
   def any_admin?
-    has_role?(:super_admin) || has_role?(:admin)
+    # done this way for performance improvement
+    @any_admin ||= (has_role?(:super_admin) || has_role?(:admin))
```
mstruve, line 302: 😍

app/models/user.rb
```diff
@@ -297,6 +297,7 @@ def admin?
     has_role?(:super_admin)
   end
 
+  # this method will be optimized in a seperate PR
```
mstruve, line 300: You can kill this too if you want, I opened up an optimization PR that way you can keep the scope of this nice and narrow :)

msarit: awesome - thanks Molly! ❤️

app/controllers/admin/chat_channels_controller.rb
```diff
@@ -29,6 +29,17 @@ def remove_user
       redirect_back(fallback_location: "/admin/chat_channels")
     end
 
+    def destroy
+      @chat_channel = ChatChannel.find(params[:id])
+      if @chat_channel.users.count.zero?
```
rhymes, line 34: small optimization: `.count` is going to issue a SQL query no matter how many times it's called in a request.
As this is called both in the controller and in the view, I think we can save one SQL query by caching this:

```
@chat_channel_users_count = @chat_channel.users.count
if @chat_channel_users_count.zero?
```

and then in the view you can use the same variable

app/controllers/admin/chat_channels_controller.rb
```diff
@@ -29,6 +29,17 @@ def remove_user
...
+        @chat_channel.destroy
+        flash[:success] = "Channel was successfully deleted."
+      else
+        flash[:alert] = "Channel NOT deleted, because it still has users."
```
Ridhwana, line 38: **Suggestion (non-blocking)**: Maybe change this to `Channel cannot be deleted, because it still has users`

app/views/admin/chat_channels/index.html.erb
```diff
@@ -29,34 +29,46 @@
...
+          <% if current_user.any_admin? %>
             <td>
-              <%= form_with url: remove_user_admin_chat_channel_url(channel), model: [:admin, channel], method: :delete, local: true do |f| %>
-                <%= f.label :usernames_string, "Users to Remove", class: "sr-only" %>
-                <%= f.text_field :username_string, id: "usersToRemoveFromChannel#{channel.id}", placeholder: "Username to remove", aria: { label: "Usernames" } %>
-                <%= f.submit "Remove From channel", data: { confirm: "Are you sure you want to remove this user?" } %>
+              <% if channel.users.count.zero? %>
+                <%= button_to "Delete Channel", { action: "destroy", id: channel.id }, method: :delete, data: { confirm: "Are you sure?" }, class: "crayons-btn crayons-btn--danger" %>
+              <% else %>
+                Cannot delete; channel has users
```
Ridhwana, line 68: **Suggestion (nit-pick, non-blocking)**: this reads a little bit awkward to me with the ; in the middle of the sentence.  Perhaps we can change it to `Channel with users cannot be deleted`

Ridhwana (approved):
This looks good @msarit 🎉. I left two non-blocking suggestions on text updates 😊

rhymes (commented):
The functional works well!

The UI has a few issues, let me know if it's okay for me to add a commit, as I noticed it's a bit complicated to add suggestions related to HTML in the PR review tooling

app/controllers/admin/chat_channels_controller.rb
rhymes: nvm, this would apply to a potential `index` action which counts both in the controller and in the view. There's no view here :D

PR #9874, opened 2020-08-19
Update top-header for logged-out user

lisasy:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
Update logged-out experience by:
- Removing unnecessary links to /notifications and /connect in top-header
- Adding clear CTA to create account or log in (both go to same route; in future, they'll be different)
- Restructuring top header for mobile & desktop experience
- Updating sidebar widget card for authentication to match these updates

Closes https://github.com/forem/InternalProjectPlanning/issues/16

#### Proposed updates
Screen Shot 2020-08-19 at 11 16 07

Screen Shot 2020-08-19 at 11 16 13

Screenshot_2020-08-19 DEV(local) Community 👩‍💻👨‍💻

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed


app/assets/stylesheets/top-bar.scss
```diff
@@ -75,6 +93,9 @@
   display: flex;
   align-items: center;
   margin-left: auto;
+
+  &--logged-out {
```
lisasy, line 97: Delete these empty lines.

app/views/articles/_sidebar_additional.html.erb
```diff
@@ -3,11 +3,31 @@
   <aside class="side-bar sidebar-additional showing grid gap-4" id="sidebar-additional">
     <% unless user_signed_in? || !any_selfserve_auth? %>
       <section class="crayons-card crayons-card--secondary signin-cta-widget">
-        <header class="crayons-card__header">
-          <h3 class="crayons-card__header__headline">Join <%= community_name %></h3>
-        </header>
-        <div class="p-4 grid gap-2">
-          <%= render partial: "shared/authentication/providers_sidebar" %>
+        <div class="authentication-widget__container">
```
lisasy, line 6: I feel like at some point in the future, I could bring this code block into a partial since I reference this in multiple places.

app/views/layouts/_top_bar.html.erb
```diff
@@ -99,5 +99,60 @@
...
+          </a>
+
+          <a href="/connect" id="connect-link" class="top-bar__link hidden" aria-label="Connect">
+            <%= inline_svg_tag("connect.svg", aria: true, class: "crayons-icon", title: "Connect") %>
+            <span class="crayons-indicator crayons-indicator--accent hidden" id="connect-number"></span>
+          </a>
+          <a href="/notifications" id="notifications-link" class="top-bar__link hidden" aria-label="Notifications">
+            <%= inline_svg_tag("bell.svg", aria: true, class: "crayons-icon", title: "Notifications") %>
+            <span class="crayons-indicator crayons-indicator--critical hidden" id="notifications-number"></span>
+          </a>
```
lisasy, line 149: For logged-out user, I initially removed this entire codeblock, but removing it didn't make the `top-bar__menu__trigger` open because of the way these interactions are set-up in `initializeSpecialNavigationFunctionality.js`. If anyone can help with refactoring this (either in this PR or a separate one), that would be much appreciated.

ludwiczakpawel:
Having the hamburger icon to trigger signup dropdown is a bit misleading - hamburger icon is usually used to open regular navigation - which we currently do by tapping this: 

image

I think we will, very soon, update the way we display navigation and header actions to reflect the new layouts (grid) we've been working on in last couple months. I tried to play with your idea and what we're going to ship soon here: 

https://www.figma.com/file/pzLExApyx00xaBnGyTwYBQ/Mobile-Header?node-id=0%3A1

lisasy:
> Having the hamburger icon to trigger signup dropdown is a bit misleading - hamburger icon is usually used to open regular navigation - which we currently do by tapping this:
> 
> image
> 
> I think we will, very soon, update the way we display navigation and header actions to reflect the new layouts (grid) we've been working on in last couple months. I tried to play with your idea and what we're going to ship soon here:
> 
> https://www.figma.com/file/pzLExApyx00xaBnGyTwYBQ/Mobile-Header?node-id=0%3A1

The hamburger menu in the top-header is *already* what we're using today to show authentication options, and I'm continuing this idea for this PR. I like organization you've started around the site-wide navigation and look forward to getting there eventually!

app/assets/stylesheets/top-bar.scss
```diff
@@ -48,6 +48,18 @@
 .top-bar--logo {
   display: flex;
 
+  &--logged-out {
+    width: auto;
+
+    @media (min-width: $breakpoint-m) {
+      width: 2fr;
```
ludwiczakpawel, line 55: This won't work - `fr` units work only within grid.

app/assets/stylesheets/widgets.scss
```diff
@@ -66,36 +66,7 @@
...
-        margin: 8px;
-        background-color: $bold-blue;
-        padding: 16px 0;
-        border-radius: 50px;
-        color: $white;
-        text-align: center;
-        font-size: 1.1em;
-      }
-    }
-  }
```
ludwiczakpawel, line 98: Nice! ♥️

app/views/layouts/_top_bar.html.erb
```diff
@@ -1,57 +1,57 @@
 <div class="navigation-progress" id="navigation-progress">
 </div>
-<header class="top-bar" id="top-bar" role="banner">
-  <a href="#articles-list" class="skip-content-link">Skip to content</a>
-  <div class="top-bar__container">
-    <a href="/" class="top-bar--logo" aria-label="<%= community_name %> Home"><%= logo_svg %></a>
+<% if user_signed_in? %>
```
ludwiczakpawel, line 3: We should somehow refactor this part - looks like there's a lot of the same code in both IF statements (whether user is logged in or logged out) - also the caching issue that @joshpuetz mentioned could be (not sure) related.

app/assets/stylesheets/top-bar.scss
```diff
@@ -67,6 +79,12 @@
   @media (min-width: $breakpoint-s) {
     margin: 0 var(--su-4);
   }
+
+  &--logged-out {
+    @media (min-width: $breakpoint-s) {
+      max-width: $breakpoint-s;
+    }
+  }
```
ludwiczakpawel, line 87: Don't you think it looks a bit off when search is different when logged in vs logged out? Of course its just my personal opinion but why do we do that in the first place?

app/assets/stylesheets/top-bar.scss
```diff
@@ -48,6 +48,18 @@
...
 
+  &--logged-out {
+    width: auto;
+
+    @media (min-width: $breakpoint-m) {
+      width: 2fr;
+    }
+
+    @media (min-width: $breakpoint-l) {
+      width: 240px;
```
ludwiczakpawel, line 59: This adds the space between logo and the search BUT it makes the space clickable.

ludwiczakpawel (changes requested):

app/views/layouts/_top_bar.html.erb
lisasy: 💯 agree refactor because this feels really messy — I think I'll need someone who is a developer to help me out here :)

app/assets/stylesheets/top-bar.scss
lisasy: I think it looks nicer if there's space in the center to have the space bar align with then main feed column — less jumpy eye movement :)

app/assets/stylesheets/top-bar.scss
lisasy: Noted!

fdoxyz:
I just pushed the changes mentioned earlier today @lisasy (the commit was actually from last Friday). The Tests will continue to fail because they're mostly expecting the buttons with text "Sign in with..."  to be there in the main feed but I can help take care of that later on when we get to a behavior we like :)

app/assets/stylesheets/nav-menu.scss
```diff
@@ -0,0 +1,47 @@
+@import 'config/import';
```
benhalpern, line 1: This file should be added to ~ line 30 of `app/views/layouts/_styles.html.erb` so that it renders inline.

app/views/layouts/_top_bar.html.erb
```diff
@@ -41,63 +41,85 @@
...
         </button>
         <%= render "layouts/nav_menu" %>
       </div>
     </div>
   </div>
-  <% if user_signed_in? %>
-    <script async>
-      // Here we have some scripts we want to get working on before
-      // waiting for the page to stream in.
+  <script async>
```
benhalpern, line 78: I don't think this needs to be changed.... There may have been some confusion about this internally eh?

benhalpern:
Design looks like a big improvement.... Looks like there are some last mile things we'll need to figure out.

app/views/layouts/_top_bar.html.erb
fdoxyz: It was indeed a confusion. The bug we saw before was unrelated to this logic. I'm working on getting the last few specs to pass and also taking it back to the previous logic 👍

app/assets/stylesheets/nav-menu.scss
fdoxyz: I just added the reference as mentioned in line ~30 but I have a question though, shouldn't this be added on all/most other `if-else` too? Similar to how `"top-bar.css"` is added in 8 different locations in that same file.

For example, at the current location it will be added inline for pages that have `@article_index = true`. But this is the CSS of the nav menu and it be presented all throughout the app. Unless I don't fully understand the inline optimization  of `layouts/_styles.html.erb` which is possible 😅

app/controllers/registrations_controller.rb
```diff
@@ -3,8 +3,11 @@ class RegistrationsController < Devise::RegistrationsController
 
   def new
     if user_signed_in?
-      redirect_to dashboard_path
+      redirect_to root_path(signin: "true")
     else
+      if URI(request.referer || "").host == URI(request.base_url).host
+        store_location_for(:user, request.referer)
+      end
```
fdoxyz, line 10: Makes use of Devise's `store_location_for` to redirect the users back using the referer (if valid)

spec/system/authentication/redirects_using_referer_spec.rb
```diff
@@ -0,0 +1,58 @@
+require "rails_helper"
```
fdoxyz, line 1: This test ensures the redirect after authentication brings the user back to where they started the process. Only applies to login because when registering they will go through the onboarding (already covered by other tests).

app/controllers/application_controller.rb
```diff
@@ -72,7 +72,7 @@ def customize_params
   # the user to after a successful log in
   def after_sign_in_path_for(resource)
     if current_user.saw_onboarding
-      path = request.env["omniauth.origin"] || stored_location_for(resource) || dashboard_path
+      path = stored_location_for(resource) || request.env["omniauth.origin"] || root_path(signin: "true")
```
fdoxyz, line 75: Moved up in priority (if available) the use of Devise's `stored_location_for` for redirecting back to the correct location

app/assets/stylesheets/nav-menu.scss
ludwiczakpawel: Hmm this file doesn't really contain any real styling. It does contain some SCSS variables and mixins that are referenced elsewhere (it is also required when you're setting media queries in your CSS because it contains our variables for breakpoints), but I don't think it has to be imported in the `_styles.html.erb`

lisasy: It seems like it's already added as line 37 of `app/views/layouts/_styles.html.erb` so I'll resolve this.

lisasy: @fdoxyz @ludwiczakpawel 

What I think I can do is bring the content of this file to `top-bar.scss` given their relevance to one another. That way, it's better contained in the right place.

ludwiczakpawel:
hey @fdoxyz I've made little updates with regards to the login/loggedout states - would you mind giving a look at these - I just wanna be sure I haven't screwed up anything..

lisasy:
@joshpuetz 

This PR is failing tests related to the new Facebook sign-up, which you can see starting on [line 1809 here in the Travis build](https://travis-ci.com/github/forem/forem/jobs/378132682). 

Below is what the proposed design looks like. The Facebook sign-in button would appear *only* on `/enter`, and is called in `app/views/shared/authentication/_providers_registration_form.html.erb`. **Could you help me update the tests to reflect these updates?**

Screen Shot 2020-08-27 at 9 07 03 AM

rhymes:
@ludwiczakpawel I can't re-request a review but GitHub signals you as currently having requested changes, so you might need to re-review this PR to approve or not approve. Thank you!

ludwiczakpawel:
Yeap, going to approve but would love to get green light from @fdoxyz re: https://github.com/forem/forem/pull/9874#issuecomment-681828392

ludwiczakpawel (approved):

app/controllers/application_controller.rb
rhymes: I think we lost some functionality now. As a logged out user check https://dev.to/new and http://localhost:3000/new (I actually think we should protect a route like /new and then redirect to it once logged in)

rhymes (commented):
Left a comment

app/controllers/application_controller.rb
fdoxyz: Thank you for this @rhymes! That's true, the `localhost:3000/new` used to redirect directly to the editor as a follow up from a logged-out user clicking on "Write a post".

I think we were pushing for a unified CTA in the navigation bar to make it easier (fewer buttons) for users to register/login. So right now I don't think there's a way to get to `/new` unless you type it directly in the URL, right?

Asking @lisasy @benhalpern about their thoughts on this because we can either remove unused logic (from [this if-else](https://github.com/forem/forem/blob/master/app/views/devise/registrations/_registration_form.html.erb#L2-L15)) or I can add the necessary changes to make sure we redirect to the editor correctly after logging in from `/new`

rhymes: Sure, there's no link to it right now, but people might have boomarked the link or end up going there while typing on the address bar of the browser. I wonder what happens if you submit an article while you're logged out :eyes:

fdoxyz: I just pushed a change that makes sure the redirect (after signing in/signing up) from `/new` goes to back the editor to avoid loosing the functionality.

I think the conversation around whether or not keeping this in the long run is valid but maybe good to have it elsewhere and ship this PR if it looks good for now 🙂

rhymes (approved):

db/schema.rb
```diff
@@ -1180,6 +1181,8 @@
   end
 
   create_table "users", force: :cascade do |t|
+    t.datetime "apple_created_at"
+    t.string "apple_username"
```
benhalpern, line 1185:
```suggestion
```

db/schema.rb
```diff
@@ -941,6 +941,7 @@
   end
 
   create_table "profile_fields", force: :cascade do |t|
+    t.boolean "active", default: true, null: false
```
benhalpern, line 944:
```suggestion
    t.boolean "active", default: true, null: false
```

db/schema.rb
```diff
@@ -1321,6 +1324,7 @@
     t.boolean "welcome_notifications", default: true, null: false
     t.datetime "workshop_expiration"
     t.string "youtube_url"
+    t.index ["apple_username"], name: "index_users_on_apple_username"
```
benhalpern, line 1327:
```suggestion
```

PR #9877, opened 2020-08-19
Remove method not used at artciles controller

brunnohenrique:

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Remove method not used at articles_controller, and resolve 2 issues at codeclimate

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -292,20 +292,6 @@ class ArticlesController < ApplicationController
     end
   end
 
-  def redirect_after_creation
-    @article.decorate
-    if @article.persisted?
-      redirect_to @article.current_state_path, notice: "Article was successfully created."
-    else
-      if @article.errors.to_h[:body_markdown] == "has already been taken"
-        @article = current_user.articles.find_by(body_markdown: @article.body_markdown)
-        redirect_to @article.current_state_path
-        return
-      end
-      render :new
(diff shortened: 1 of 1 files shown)


app/controllers/articles_controller.rb
```diff
@@ -254,7 +254,7 @@ def article_params_json
     if params["article"]["series"].present?
       collection = Collection.find_series(params["article"]["series"], @user)
       params["article"]["collection_id"] = collection.id
-    elsif params["article"]["series"] == "" # reset collection?
+    else # reset collection?
```
citizen428, line 257: This a logic change. Having an `elsif` means that this whole construct could be a no-op, which is not true with an `else` since there will always be one branch executing. I don't feel particularly good with this change unless we make sure there's excellent test coverage around this.

brunnohenrique: It really isn't the same. I'll remove this change and open another PR to treat the exceeded lines.

citizen428 (approved):
Thanks for the update @brunnohenrique! 😺

rhymes (approved):

PR #9841, opened 2020-08-18
Listing endorsements: models setup

Oghenebrume50:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Followign #3062  issue on creating endorsements for listing and my work progress on #9553  this PR is a smaller one showing the model relationship for this implementation

`bundle exec rspec spec/models/listing_endorsement_spec.rb` will test this model, other relevant models to test would be 
`bundle exec rspec spec/models/listing_spec.rb` and `bundle exec rspec spec/models/user_spec.rb`

## Added tests?

- [x] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

PR #9878, opened 2020-08-19
Flaky Spec Fix:Use Comment created_at in Expectation

mstruve:
- [x] Flaky Spec Fix

## Description
I noticed while reviewing https://github.com/forem/forem/pull/9841 that it had the following flaky spec. When I looked closer at the spec I realized that we were comparing the comment created_at in the response to the article created_at which is incorrect. I updated the spec with the proper root_comment created_at. 
```
       -"created_at" => "2020-08-19T15:39:53Z",
       +"body_html" => "<p>Neutra kinfolk distillery chartreuse small batch cliche. Polaroid mustache cornhole direct trade 
        ....
       +"created_at" => "2020-08-19T15:39:54Z",
```

alt_text

### DIFF
--- a/spec/requests/api/v0/comments_spec.rb
+++ b/spec/requests/api/v0/comments_spec.rb
@@ -100,7 +100,7 @@ RSpec.describe "Api::V0::Comments", type: :request do
     it "returns date created" do
       get api_comments_path(a_id: article.id)
       expect(find_root_comment(response)).to include(
-        "created_at" => article.created_at.utc.iso8601,
+        "created_at" => root_comment.created_at.utc.iso8601,
       )
     end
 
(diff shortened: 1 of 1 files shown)


spec/requests/api/v0/comments_spec.rb
```diff
@@ -167,7 +167,9 @@ def find_child_comment(response, action = :index)
     context "when getting by podcast episode id" do
       let(:podcast) { create(:podcast) }
       let(:podcast_episode) { create(:podcast_episode, podcast: podcast) }
-      let!(:comment) { create(:comment, commentable: podcast_episode) }
+      let(:comment) { create(:comment, commentable: podcast_episode) }
```
mstruve, line 170: Rubocop complained about `let!` when I tried to commit this so I fixed it.

vaidehijoshi: Ugh yeah, i kind of wish rubocop wouldn't because you end up having to invoke the dang thing just to make it not mad!

PR #9886, opened 2020-08-20
Add display columns to ProfileField

citizen428:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR adds two more colums to the `ProfileField` model: 

* `display_area`: an enum that for now has 2 options, "header" or "left_sidebar".
* `show_in_onboarding`": Whether or not the profile field wil be shown during the onboarding flow.

Part of #6994

Not much to test yet, this is mostly about unblocking @Ridhwana for the work on the new profile page.

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/app/models/profile_field.rb
+++ b/app/models/profile_field.rb
@@ -11,9 +11,15 @@ class ProfileField < ApplicationRecord
     color_field: 3
   }
 
+  enum display_area: {
+    header: 0,
+    left_sidebar: 1
+  }
+
   validates :label, presence: true, uniqueness: { case_sensitive: false }
   validates :active, inclusion: { in: [true, false] }
   validates :attribute_name, presence: true, on: :update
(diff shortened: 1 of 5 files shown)


rhymes (commented):
Nothing wrong with this PR, just me asking if there's a better way not have to presentation logic embedded in the DB.

What do you think?

citizen428:
> Nothing wrong with this PR, just me asking if there's a better way not have to presentation logic embedded in the DB.

Unless we go with simplistic heuristics like links -> header, rest -> sidebar we need to store it _somewhere_. I'm open to alternative suggestions, but for now I'd rather just keep this moving along.

Ridhwana (approved):

PR #9888, opened 2020-08-20
Add relations and foreign keys between Podcast related models

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

`PodcastEpisode` and `Podcast` are missing a few clauses and one foreign key.

There's should be no need for a cleanup PR as there are no orphan rows in the DB 

See 

- https://dev.to/admin/blazer/queries/194-orphan-podcast-episodes
- https://dev.to/admin/blazer/queries/195-orphan-podcasts

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/models/podcast_episode.rb
+++ b/app/models/podcast_episode.rb
@@ -15,7 +15,7 @@ class PodcastEpisode < ApplicationRecord
   delegate :published, to: :podcast
 
   belongs_to :podcast
-  has_many :comments, as: :commentable, inverse_of: :commentable
+  has_many :comments, as: :commentable, inverse_of: :commentable, dependent: :nullify
 
   mount_uploader :image, ProfileImageUploader
   mount_uploader :social_image, ProfileImageUploader
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -26,8 +26,6 @@ Rails/HasManyOrHasOneDependent:
     - 'app/models/article.rb'
     - 'app/models/concerns/user_subscription_sourceable.rb'
     - 'app/models/organization.rb'
-    - 'app/models/podcast.rb'
-    - 'app/models/podcast_episode.rb'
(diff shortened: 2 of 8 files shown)


app/models/podcast_episode.rb
```diff
@@ -15,7 +15,7 @@ class PodcastEpisode < ApplicationRecord
   delegate :published, to: :podcast
 
   belongs_to :podcast
-  has_many :comments, as: :commentable, inverse_of: :commentable
+  has_many :comments, as: :commentable, inverse_of: :commentable, dependent: :nullify
```
Ridhwana, line 18: **Question**: I've never really used nullify before, but I read up on it and wanted to understand more. When a Podcast is deleted we destroy the depending podcast_episodes, so why do we nullify the comments instead of destroying them?

Ridhwana (approved):

app/models/podcast_episode.rb
mstruve: A little while ago we determined that comments really belong to the user that wrote them and that they might contain pertinent information even without the context of the original commentable so we decided to keep them even when their commentable is removed. 

On a separate note, going forward comments with commentable will have their commentable ID set to null, should we backfill null values for those comments that have already had their commentable removed?

rhymes: @mstruve I haven't found any related to podcast episodes, see the third clause here https://dev.to/admin/blazer/queries/194-orphan-podcast-episodes

I agree though :) There are if we check the articles table, but I left `Article`, `Organization` and `User` to the end of this "cleanup sprint" as they are the most connected: https://dev.to/admin/blazer/queries/196-comments-with-an-invalid-article-id - 8,916 comments with an invalid article id

PR #9891, opened 2020-08-20
Use and organize constants for the config page 

Ridhwana:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Replace the rest of the placeholders and descriptions from the constants file.
Also, just _loosely_ alphabetized the hash.
 
No Visual Changes. 
You can just go to http://localhost:3000/admin/config to ensure that the descriptions and placeholders show. 

 
## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

alt_text

### DIFF
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -5,13 +5,17 @@ module Constants
         description: "How can users sign in?",
         placeholder: ""
       },
-      campaign_hero_html_variant_name: {
-        description: "Hero HtmlVariant name",
-        placeholder: ""
-      },
       campaign_articles_require_approval: {
         description: "",
         placeholder: "Campaign stories show up on sidebar with approval?"
+      },
+      campaign_featured_tags: {
+        description: "Posts with which tags will be featured in the campaign sidebar (comma separated, letters only)",
+        placeholder: "List of campaign tags: comma separated, letters only e.g. shecoded,theycoded"
+      },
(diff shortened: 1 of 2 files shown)


app/lib/constants/site_config.rb
```diff
@@ -5,14 +5,18 @@ module SiteConfig
...
-      campaign_hero_html_variant_name: {
-        description: "Hero HtmlVariant name",
-        placeholder: ""
-      },
       campaign_articles_require_approval: {
         description: "",
         placeholder: "Campaign stories show up on sidebar with approval?"
       },
+      campaign_featured_tags: {
+        description: "Posts with which tags will be featured in the campaign sidebar (comma separated, letters only)",
```
rhymes, line 13: what do you think about `Posts with these tags` or `Tags whose posts...`

rhymes (approved):
non blocking as I've noticed it's the test we already have in place

app/lib/constants/site_config.rb
Ridhwana: @rhymes this sounds like better grammar. My next pass is going to fix the language and make this more descriptive and concise for new communities. I'll tackle it then. Thanks 😊

PR #9896, opened 2020-08-20
Add the Ability to Remove Badge Achievements From Users

juliannatetreault:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR gives Admins the ability to remove badges from users and separates `badge_achievement`-related code from `badge`-related code. To accomplish this, this PR does the following:
- Adds an `Admin::BadgeAchievement::Controller`
- Adds `#index` and `#destroy` actions to the `Admin::BadgeAchievements::Controller` and ports over `#award` and `#award_badges` from the `Admin::Badges::Controller`
- Removes `#award` and `#award_badges` from the `Admin::Badges::Controller`
- Adds routes for `badge_achievements` and redirects to ensure that Admins to not encounter errors potentially caused by bookmarking the old `badge_achievement` routes
- Adds an `index` and `award` view for Badge Achievements
- Adds a warning to the `index` view stating: "Note: Badges that are automatically rewarded cannot and will not be removed from users!"
- Adds a search by `user_id` bar to the `index` view
- Limits the number of badges displayed on the `index` view to 15, order by descending `created_at`
- Adds specs around awarding and deleting `BadgeAchievements`
- Removes `badge_achievement`-related tests from the `badges_spec.rb`
- Adds a Badge Achievements menu item to the Admin nav bar

Relates to a private issue.

**To QA these changes, first ensure that you are a `super_admin` with badges seeded, then please do the following:**
- Navigate to `/admin/badge_achievements` by typing out the URL or by navigating to `/admin` and then clicking the "Badge Achievements" navigation tab:
Screen Shot 2020-08-21 at 12 22 43 PM
- Click the "Award Badges" button to navigate to the newly moved badge award form and click the "Back to Badge Achievements" button to return to the `index` page:
Screen Shot 2020-08-20 at 11 43 20 AM
- Once you're back on the `index` view, test out the search by `user_id` functionality by searching for a user who has a Badge Achievement by ID:
Screen Shot 2020-08-21 at 12 23 48 PM
Screen Shot 2020-08-21 at 12 23 59 PM
- Finally, remove a badge from a user (yourself, another user, or both!). You will be redirected back to the `index` view and
 should see a pop-up confirming the removal and then a confirmation message of the removal after clicking "Remove":
Screen Shot 2020-08-21 at 12 24 29 PM
Screen Shot 2020-08-21 at 12 24 38 PM
- To ensure that everything worked properly, navigate to the user's page and double-check that the badge has been removed from their profile:
Screen Shot 2020-08-21 at 12 25 32 PM
- Try to break things! 🛠️ 

## Added tests?

- [x] yes, and refactored existing specs
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

N/A

backstreet boys


app/views/admin/badge_achievements/index.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  <tbody>
+   <% @badge_achievements.each do |badge_achievement| %>
+    <tr>
+      <td><%= badge_achievement.user.username %></td>
+      <td><%= badge_achievement.badge.title %></td>
+      <h5>
+        <td class="justify-content-center">
+          <img class="mx-auto mt-3" width="40" height="40" src="<%= badge_achievement.badge.badge_image %>" alt="badge image">
+        </td>
+        <td><%= link_to "Delete", url_for(action: :destroy, id: badge_achievement.id), method: :delete, data: { confirm: "Are you sure?" }, class: "btn btn-danger" %></td>
```
juliannatetreault, line 23: Should this button read "Delete", "Destroy" or "Remove"? I think that both "Delete" and "Remove" are the best options, but I'm not sure which of the two is more intuitive. Thoughts? 🤔

app/views/admin/badge_achievements/index.html.erb
```diff
@@ -0,0 +1,30 @@
...
+          <img class="mx-auto mt-3" width="40" height="40" src="<%= badge_achievement.badge.badge_image %>" alt="badge image">
+        </td>
+        <td><%= link_to "Remove", url_for(action: :destroy, id: badge_achievement.id), method: :delete, data: { confirm: "Are you sure?" }, class: "btn btn-danger" %></td>
+      </h5>
+    </tr>
+    <% end %>
+  </tbody>
+</table>
+
+<%= paginate @badge_achievements %>
```
juliannatetreault, line 30: I added pagination to this page because I am fearful that there will be a _ton_ of Badge Achievements on production, but I also fear that this isn't enough. Should I try to add a search bar of some sort in this view so that Admins can search via `username`?

app/controllers/admin/badge_achievements_controller.rb
```diff
@@ -0,0 +1,44 @@
+module Admin
+  class BadgeAchievementsController < Admin::ApplicationController
+    layout "admin"
+
+    def index
+      @badge_achievements = BadgeAchievement.all.page(params[:page]).order(created_at: :desc).per(10)
```
mstruve, line 6: Using `.all` here is not going to be practical since we have 354k of these in the DEV database.

app/controllers/admin/badge_achievements_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+      if @badge_achievement.destroy
+        flash[:success] = "Badge has been deleted!"
+      else
+        flash[:danger] = @badge_achievement.errors_as_sentence
+      end
+      redirect_to admin_badge_achievements_path
+    end
+
+    def award
+      @badge = Badge.all
```
mstruve, line 21: We don't have as many badges currently but we want this all to be very scalabe so we should try and avoid using `.all` whenever possible and use SQL paging

app/views/admin/badge_achievements/index.html.erb
Zhao-Andy: Hmm yeah I think a search function would be useful. I think probably `user_id` first is fine, since `username` might be a bit more complicated.

@jacobherrington had a good search implementation throughout `/admin` if you want to look at some examples. It's the [Ransack gem](https://github.com/activerecord-hackery/ransack).

juliannatetreault: Okay, I'll take a look @Zhao-Andy! I actually halfway implemented search for this but ran into some issues and ended up switching gears since it wasn't part of the MVP. 😄

config/routes.rb
```diff
@@ -117,8 +117,13 @@
     resources :webhook_endpoints, only: :index
     resource :config
     resources :badges, only: %i[index edit update new create]
-    get "/badge_achievements/award_badges", to: "badges#award"
-    post "/badge_achievements/award_badges", to: "badges#award_badges"
+    # These redirects serve as a safegaurd to prevent 404s for any Admins
+    # who have the old badge_achievement URLs bookmarked.
+    get "/badges/badge_achievements", to: redirect("/admin/badge_achievements")
```
mstruve, line 122: 🔥  thank you for saving us all from ourselves 😉

juliannatetreault: Any time! 😄

juliannatetreault:
> Couple performance comments. What kind of badges will this be used to remove? I ask bc we have badges like "Beloved Comment" badge which if removed will simply be re-added when the next time that job is run.

@mstruve: Thanks for the review! 🚀  This PR is specifically for Badge Achievements, so badges like the Hacktoberfest badge that users were awarded for participating last year would fall under this category. The `resource_admin/badge_achievements` panel is going to be deprecated in the nearer future, so this will give Admins the ability to perform the actions they would there, here. I'm not sure what a comment badge is, but I don't believe that this has anything to do with that! 😄

mstruve:
> This PR is specifically for Badge Achievements, so badges like the Hacktoberfest badge that users were awarded for participating last year would fall under this category. 

Any badge, whether it is Hacktoberfest or Beloved Comment or Weekly Streak is given to a person via a BadgeAchivement. That means your PR will give admins the ability to remove any badge achievement which will remove a badge from a user. Here is the beloved comment badge code. As you can see, it gives the user a badge through badge achievements. In this case though, even if we remove that badge achievement, it will simple get readded the next time this chunk of code runs. 
```ruby
  def self.award_beloved_comment_badges(comment_count = 24)
    badge_id = Badge.find_by!(slug: "beloved-comment").id
    Comment.includes(:user).where("public_reactions_count > ?", comment_count).find_each do |comment|
      message = "You're famous! [This is the comment](#{URL.comment(comment)}) for which you are being recognized. 😄"
      achievement = BadgeAchievement.create(
        user_id: comment.user_id,
        badge_id: badge_id,
        rewarding_context_message_markdown: message,
      )
      comment.user.save if achievement.valid?
    end
  end
```

juliannatetreault:
> > This PR is specifically for Badge Achievements, so badges like the Hacktoberfest badge that users were awarded for participating last year would fall under this category.
> 
> Any badge, whether it is Hacktoberfest or Beloved Comment or Weekly Streak is given to a person via a BadgeAchivement. That means your PR will give admins the ability to remove any badge achievement which will remove a badge from a user. Here is the beloved comment badge code. As you can see, it gives the user a badge through badge achievements. In this case though, even if we remove that badge achievement, it will simple get readded the next time this chunk of code runs.
> 
> ```ruby
>   def self.award_beloved_comment_badges(comment_count = 24)
>     badge_id = Badge.find_by!(slug: "beloved-comment").id
>     Comment.includes(:user).where("public_reactions_count > ?", comment_count).find_each do |comment|
>       message = "You're famous! [This is the comment](#{URL.comment(comment)}) for which you are being recognized. 😄"
>       achievement = BadgeAchievement.create(
>         user_id: comment.user_id,
>         badge_id: badge_id,
>         rewarding_context_message_markdown: message,
>       )
>       comment.user.save if achievement.valid?
>     end
>   end
> ```

Well then... back to the drawing board. Thank you so much for clarifying this @mstruve! 😅

Zhao-Andy:
@mstruve The idea is that we want to be able to remove badges for certain one-off instances. For example:
1. we have a user that needs to merge their account, but the merge is blocked because both accounts have the 1 year badge
2. we have a user who was accidentally granted the wrong one-off badge (contest winner vs contest participant) and need to remove their badge achievement

I see your point about allowing admins to remove _any_ badge achievement, and was something @pkfrank voiced, too -- for slightly different reasons. I personally think it's okay if this happens anyway, since if the wrong badge achievement was removed it would be re-added by the chron job. Also, I don't think an admin should be trying to remove a chron job related badge achievement. I suppose we could implement a way to prevent that from happening -- maybe adding a `type_of` to the `BadgeAchievement` table -- but I think that would probably be adding some complexity that we didn't need (for this PR at least).

@juliannatetreault now that I think about it, maybe the index page doesn't have to list all the badge achievements at all, and can simply be a search page. Then we won't have to worry about the performance of the page at all, since an admin would search for a badge achievement by a certain user ID and then go from there. What do you two think?

mstruve:
@Zhao-Andy I am not worried about the wrong badge being removed, I am worried about an admin _thinking_ they can remove a badge and that it's gone forever, when in reality that is not the case for many badges. I could see a Forem admin thinking "Lets remove that beloved comment badge" Then they come in the next day and see its back. They try to remove it again but another day goes by and its back. Now they are emailing us bc they think something is broken. Wether we prevent that from happening with code or just improve our messaging around certain badges, I dont really have a preference but I think its something that should be addressed.

app/controllers/admin/badge_achievements_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      if @badge_achievement.destroy
+        flash[:success] = "Badge has been deleted!"
+      else
+        flash[:danger] = @badge_achievement.errors_as_sentence
+      end
+      redirect_to admin_badge_achievements_path
+    end
+
+    def award
+      @badge = Badge.page(params[:page]).per(Badge.count)
```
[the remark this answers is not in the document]
juliannatetreault: So, my only hesitation with changing this to a fixed number, like 15 or 20, is that I think the reason why this loads all of the badges for the Admin is so that they are able to select any badge from the badges that exist and award it to a user at any given time (_I believe_). 

As for changing the variable name here to be something more explicit, I was able to do this without an issue and went ahead and changed it to `@all_badges`. How does this sound @mstruve? 😄

app/views/admin/badge_achievements/award.html.erb
```diff
@@ -0,0 +1,21 @@
+<main>
+  <h2>Award Badges</h2>
+  <div class="alert alert-warning"><strong>Warning: Usernames are case-sensitive and must be all lowercase! This task will fail silently when supplied with incorrect
```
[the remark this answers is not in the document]
juliannatetreault: I added an additional downcase on `usernames` in my last badge-related PR to catch any edge cases with capitalized `usernames` when an Admin fills out this form, so I don't think this is _completely_ necessary. The super explicit message here is more for the user and less about it being an issue that Admins experience at this point!

app/controllers/admin/badge_achievements_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+        .order(created_at: :desc)
+        .ransack(params[:q])
+      @badge_achievements = @q.result.page(params[:page] || 1).per(15)
+    end
+
+    def destroy
+      @badge_achievement = BadgeAchievement.find(params[:id])
+
+      if @badge_achievement.destroy
+        flash[:success] = "Badge has been deleted!"
```
Zhao-Andy, line 18: Nitpick: Could this be changed to "Badge achievement has been deleted!"?

app/views/admin/badges/index.html.erb
```diff
@@ -2,7 +2,6 @@
   <div class="col">
     <div class="row justify-content-end">
       <%= link_to "Create Badge", new_admin_badge_path, class: "btn btn-primary" %>
-      <%= link_to "Award Badge", admin_badge_achievements_award_badges_path, style: "margin-left: 20px;", class: "btn btn-primary" %>
```
Zhao-Andy, line 5: Hmm I actually think it's nice to keep this here in `/admin/badges` 😅

app/controllers/admin/badge_achievements_controller.rb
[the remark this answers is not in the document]
mstruve: >  I think the reason why this loads all of the badges for the Admin is so that they are able to select any badge

By using `page` and `per` here you are setting up the view for pagination so I would assume the admin could page through all of the badges to find the one they are looking for. 

> I was able to do this without an issue and went ahead and changed it to @all_badges

WOOT! 🙌

app/controllers/admin/badge_achievements_controller.rb
juliannatetreault: Great call @Zhao-Andy ! I've made this change.

app/views/admin/badges/index.html.erb
juliannatetreault: Okay, I've added this back! 😄

juliannatetreault:
> Working great @juliannatetreault! Others have made several comments that I will not duplicate.
> 
> I had a potential out-of-scope question though:
> Assuming community-listing credits are issued for each badge awarded (5 points in the screenshot), does this mean that rescinding the badge also removes these credits?
> 
> Screen Shot 2020-08-24 at 11 35 12 AM

Thanks for the review and great question @msarit! I just dropped into my console and tested this out and it does not look like the removal of a Badge Achievement affects a User's credits.

app/controllers/admin/badge_achievements_controller.rb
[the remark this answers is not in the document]
juliannatetreault: That's a fair point @mstruve! I just sanity checked myself to make sure that we needed and wanted access to all of the Badges here and it seems like that is the case. That being said, I've changed this to `@all_badges = Badge.all.select(:title, :slug)` to optimize the query a little more and removed the SQL paging since we do not need pagination here and it's adding unnecessary code to the controller. How do you feel about this?

PR #9898, opened 2020-08-20
Refactor:Move Listing and Support Tag Tasks to Sidekiq

mstruve:
- [x] Refactor
- [x] Optimization

## Description
This moves the `resave_supported_tags` and `expire_old_listings` rake tasks to Sidekiq Crons. While I was here I made a little optimization to the listing task by adding published to our scope and by looking for published articles with `expires_at` less than the current day. This will ensure that listings get expired even if we run into problems and possibly skip a day for one reason or another. 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/spec/workers/tags/resave_supported_tags_worker_spec.rb
+++ b/spec/workers/tags/resave_supported_tags_worker_spec.rb
@@ -1,0 +1,20 @@
+require "rails_helper"
+
+RSpec.describe Tags::ResaveSupportedTagsWorker, type: :worker do
+  let(:worker) { subject }
+
+  include_examples "#enqueues_on_correct_queue", "low_priority"
+
+  describe "#perform" do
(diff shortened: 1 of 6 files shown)


spec/workers/tags/resave_supported_tags_worker_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  describe "#perform" do
+    it "resaves supported tags" do
+      Timecop.freeze do
+        old_update_timestamp = 1.week.ago
+        tag = create(:tag, supported: true, updated_at: old_update_timestamp)
+
+        worker.perform
+
+        expect(tag.updated_at).to be > old_update_timestamp
```
maestromac, line 16: TIL! Don't think i've ever used `>` in expectation before.

PR #9899, opened 2020-08-20
Add Twitter and GitHub keys to SiteConfig

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

Remove the need to set social keys in `ENV` and instead move it to config that an admin can do.

Admins need to go and log into the appropriate portal to get the key, so it makes sense that they'd paste it into the app and not relay to systems team or otherwise over-complicate this.

This makes use of the Omniauth "setup phase" which exists because of scenarios like multi-tenant apps, etc. which is the same principles for why we need to do it this way.

Our docs used `GITHUB_KEY` etc. as examples for ENV stuff, so swapped with Cloudinary (which will also be gone soon!)

**To note:**  I think we could be do for a slight re-categorization of the Admin config dashboard, but I think it can be out of the scope of this PR.


config/.env_sample
```diff
@@ -0,0 +1,134 @@
+# App domain + Protocol setting for development
```
mstruve, line 1: This file should be removed since it is now located in the root directory

app/views/admin/configs/show.html.erb
```diff
@@ -130,9 +130,40 @@
                              class: "form-control selectpicker" %>
               <div class="alert alert-info"><%= Constants::SiteConfig::DETAILS[:authentication_providers][:description] %></div>
             </div>
+            <div class="form-group">
+              <%= admin_config_label :twitter_key %>
+              <%= f.text_field :twitter_key,
+                               class: "form-control",
+                               value: SiteConfig.twitter_key,
+                               placeholder: Constants::SiteConfig::DETAILS[:twitter_key][:placeholder] %>
```
mstruve, line 138: As is, these values are still going to require an app restart to take affect. Should we mention something about that here so people dont assume as soon as they are changed they are being used?

benhalpern: Ah... I didn't realize it still worked this way. I was thinking that the instructions were passed in to execute at runtime, even if it was instructed at boot.

I added a notice about this for now, but I still think we can make further adjustments to make sure this just works. Need to read through the omniauth instructions more clearly 😄

joshpuetz:
@benhalpern Is the auth setup documentation at https://docs.dev.to/backend/auth-twitter/ and https://docs.dev.to/backend/auth-github/ still correct given these changes?

PR #9900, opened 2020-08-20
Bug Fix:Skip Busting Edge Caches if Fastly API Key is not Present

mstruve:
- [x] Bug Fix

## Description
On Forem.dev we are not currently using any cacheing. This causes our CacheBuster to fail with the following error whenever it is run
```
Fastly::Error: {"msg":"Provided credentials are missing or invalid"}
```
To prevent this from happening I added a guard clause to ensure we dont try to hit the Fastly API unless we have credentials for it. 

https://github.com/orgs/forem/projects/6#card-43527526

alt_text

### DIFF
--- a/app/labor/cache_buster.rb
+++ b/app/labor/cache_buster.rb
@@ -19,7 +19,7 @@ module CacheBuster
     #
     # https://github.com/fastly/fastly-ruby#efficient-purging
     return unless Rails.env.production?
-    return unless ApplicationConfig["FASTLY_API_KEY"]
+    return if ApplicationConfig["FASTLY_API_KEY"].blank?
 
     HTTParty.post("https://api.fastly.com/purge/https://#{ApplicationConfig['APP_DOMAIN']}#{path}",
                   headers: { "Fastly-Key" => ApplicationConfig["FASTLY_API_KEY"] })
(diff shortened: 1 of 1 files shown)


atsmith813:
@vaidehijoshi / @forem/delightful is in the process of tackling this as well with https://github.com/forem/forem/pull/9857.

@vaidehijoshi what do you think?

mstruve:
For a little background, I currently have a Datadog alert muted bc of the noise this is causing. This doesn't have to be the final solution but it would at least allow me to turn that alert back on so we can be notified when jobs die.

vaidehijoshi:
> @vaidehijoshi / @forem/delightful is in the process of tackling this as well with #9857.
> 
> @vaidehijoshi what do you think?

Confirm, I've got a draft PR that will be ready soon (pending new tests) [here](https://github.com/forem/forem/pull/9857/files#diff-6e58540a6172a3c1054d1ede7134dd00R38) that does a version of this. I'm fine to merge what @mstruve has but just with the caveat that we'll be changing it very shortly!

mstruve:
Whew! yeah @vaidehijoshi that is a bit more robust than my patch 😅

PR #9904, opened 2020-08-20
Replace HEROKU_SLUG_COMMIT with custom footprint (naive)

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We use `HEROKU_SLUG_COMMIT` to know when one release is later than another to bust caches and other duties.

This is a Heroku dependency we need to get rid of. This PR introduces a minute-precision timestamp which will be good enough right this moment, but it could be different on a process-by-process basis. For basic cache-busting and the such that is fine, but ideally we can computer a value here which will be more consistent.

If anybody has an idea for the most straightforward way to do that, feel free to drop an idea. Otherwise, I'd say we ship this as "good enough for right now".

### DIFF
--- a/app/controllers/async_info_controller.rb
+++ b/app/controllers/async_info_controller.rb
@@ -35,7 +35,7 @@ class AsyncInfoController < ApplicationController
     set_surrogate_key_header "shell-version-endpoint"
     # shell_version will change on every deploy.
     # *Technically* could be only on changes to assets and shell, but this is more fool-proof.
-    shell_version = ApplicationConfig["HEROKU_SLUG_COMMIT"]
+    shell_version = ApplicationConfig["RELEASE_FOOTPRINT"]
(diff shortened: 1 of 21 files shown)


benhalpern:
@mstruve yeah something like that totally works for me.

Ideally I'd think we can do something with minimal moving pieces.

mstruve:
@jdoss can we hook something like that up in our deploy process? Where we grab the SHA from the container and write it to a file before boot?

jdoss:
@mstruve shouldn't be too hard. I'll look into it tomorrow.

benhalpern:
Changed the code... @jdoss will add a `RELEASE_FOOTPRINT` to our Forem Cloud environment which will increment before boot.

benhalpern:
Dealing with fixing the tests for new scenario.

PR #9909, opened 2020-08-21
Expand Images::Optimizer to support imgproxy

maestromac:
- [x] Feature

## Description
Added Imgproxy support and enable it to the only one area: the mascot that would appear when the user has 0 posts on their dashboard.

This *won't* enable Imgproxy unless both Imgproxy's key and endpoint vars are provided.

https://github.com/orgs/forem/projects/6#card-43004903

To test out imgproxy, you'll have to do the following

1. You will need to install imgproxy locally. You can do so with `brew install imgproxy`
2. You need to generate a key/salt pair for this to work. In a terminal, rung `echo $(xxd -g 2 -l 64 -p /dev/random | tr -d '\n')` twice and copy those values to your `.env` in the next step
3. In your `.env` set the following
```
export IMGPROXY_ENDPOINT= 'http://localhost:8080'
export IMGPROXY_KEY='1b1c9aae804e070b0864f2547fba7ce8ff31bf7..........'
export IMGPROXY_SALT='8c6d449d4fc2cada5bab538826cae709d2ade9f9286af8b83573346bfb7.........'
```
4. In a separate terminal tab, start imgproxy server by running `IMGPROXY_KEY='your key' IMGPROXY_SALT='your salt' imgproxy`
5. Start rails server normally.
6. Just for this demonstration, please change line 40 of `app/views/dashboards/show.html.erb` to `<% if false %>`. We're pretending that you have 0 articles.
7. Startup rails console on another terminal window, and set your mascot URL with `SiteConfig.mascot_image_url = "https://thepracticaldev.s3.amazonaws.com/i/bh2wmpcltaybu1xsnico.png"`
8. Navigate to `/dashboard`, the dev icon you see should have a link similar to `http://localhost:8080/Utx2iCpXzy8qyckZNCeVL5F7mHtUvUiqEE1gbeys_1g/w:300/aHR0cHM6Ly90aGVw/cmFjdGljYWxkZXYu/czMuYW1hem9uYXdz/LmNvbS9pL2JoMndt/cGNsdGF5YnUxeHNu/aWNvLnBuZw` and not a Cloudinary url.

Please let me know if any of these steps need clarification.
## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed


app/services/images/optimizer.rb
```diff
@@ -1,5 +1,13 @@
 module Images
   module Optimizer
+    def self.call(img_src, service: :cloudinary, **kwargs)
+      if imgproxy_enabled?
+        public_send(service, img_src, kwargs)
```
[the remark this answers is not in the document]
maestromac: Yep that totally would!

spec/views/comments/_comment.html.erb_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "rails_helper"
```
maestromac, line 1: This spec is unrelated to the feature introduced in this PR but still necessary to test the low-quality feature that I broke on my previous PR.

app/services/images/optimizer.rb
```diff
@@ -1,5 +1,11 @@
 module Images
   module Optimizer
+    def self.call(img_src, service: :cloudinary, **kwargs)
+      return cloudinary(img_src, kwargs) unless imgproxy_enabled?
+
+      public_send(service, img_src, kwargs)
+    end
```
maestromac, line 7: I decided to go with @rhymes suggestion to add a guard clause here to override the arg injection. Ideally, I would most likely move the responsibility of this override to a view helper but this achieves what I need with less code and minimal indirection.

app/services/images/optimizer.rb
```diff
@@ -1,5 +1,11 @@
 module Images
   module Optimizer
+    def self.call(img_src, service: :cloudinary, **kwargs)
+      return cloudinary(img_src, kwargs) unless imgproxy_enabled?
```
mstruve, line 4: Should we be more explicit here and say
```ruby
return cloudinary(img_src, kwargs) unless imgproxy_enabled? && service == :imgproxy
```

config/initializers/imgproxy.rb
```diff
@@ -0,0 +1,20 @@
...
+  # Hex-encoded signature key
+  config.hex_key = ApplicationConfig["IMGPROXY_KEY"]
+  # Hex-encoded signature salt
+  config.hex_salt = ApplicationConfig["IMGPROXY_SALT"]
+
+  # Base64 encode all URLs
+  config.base64_encode_urls = true
+
+  # Always escape plain URLs
+  # config.always_escape_plain_urls = true
```
mstruve, line 19: Why do we have this commented out rather than using it?

rhymes (approved):
So cool! 

Great work @maestromac!

imgproxy also gives us [webp support for free](https://docs.imgproxy.net/#/configuration?id=webp-support-detection) and we can also decide TTL for the assets 🔥 

We can also [have presets](https://docs.imgproxy.net/#/presets) 😭

app/services/images/optimizer.rb
maestromac: yep I can do that

config/initializers/imgproxy.rb
maestromac: I missed this. From the look of it, this will only help us. Going to enable it

PR #9910, opened 2020-08-21
Stop using ProfileField#active Part 1

citizen428:

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When creating the new `ProfileField` model I originally added a column called `active`. My rationale was as follows: While admins experiment with setting up their Forem instance the may try out different fields and see how the profile page looks. Deleting and recreating is tiring, so I wanted to give them an option to disable it. However, after discussing this with @Ridhwana we decided that the potential confusion is not worth it and this is a case where less flexibility makes for a better product experience. 

There will be a follow-up PR to remove the column itself.

Part of #6994

Nothing, if the specs are green, we should be good to go.

## Added tests?

- [x] yes (updated, not added)
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -3,9 +3,9 @@ class Profile < ApplicationRecord
 
   validates :user_id, uniqueness: true
 
-  # This method generates typed accessors for all active profile fields
+  # This method generates typed accessors for all profile fields
   def self.refresh_store_accessors!
-    ProfileField.active.find_each do |field|
+    ProfileField.find_each do |field|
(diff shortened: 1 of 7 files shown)


citizen428:
@mstruve @Ridhwana Fixed all the specs now, going to merge and deploy this now. Thanks for pre-emptively approving so I don't get blocked on my Monday morning, much appreciated 🙏

PR #9912, opened 2020-08-21
Add Profiles::Update service object

citizen428:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Closes https://github.com/forem/forem/issues/10135

This PR introduces a service object (`Profiles::Update`) for making changes to profiles. This will most likely be used in a new `ProfilesController` backing `/settings/profile`, but there may also be other places where this is useful. The service also cleans out stale profile fields and sync data back to `User` (which will be eventually removed).

The easiest way to test this is using the service object in a Rails console and observing the changes to both the profile and user.

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

image


app/models/profile.rb
```diff
@@ -3,12 +3,47 @@ class Profile < ApplicationRecord
 
   validates :user_id, uniqueness: true
 
-  # This method generates typed accessors for all profile fields
-  def self.refresh_store_accessors!
+  USER_MIXIN = Module.new
```
citizen428, line 8: We're building this mixin up dynamically, see https://dev.to/appsignal/configurable-ruby-modules-the-module-builder-pattern-4483 for more info. I'll add a comment later.

app/models/profile.rb
```diff
@@ -3,12 +3,47 @@ class Profile < ApplicationRecord
...
+  }.with_indifferent_access.freeze
+
+  # Generates typed accessors for all currently defined profile fields.
+  def self.refresh_attributes!
     ProfileField.find_each do |field|
-      store_attribute :data, field.attribute_name, field.type
+      attribute_name = field.attribute_name
+      store_attribute :data, attribute_name, field.type
+
+      getter = MAPPED_ATTRIBUTES.fetch(attribute_name, attribute_name).to_s
```
citizen428, line 29: Here we add to the dynamic module defined above. This works because Ruby is extremly late binding and method lookup doesn't happen until runtime.

Simplied example:

```ruby
# Define a test module and class
module TestMixin; end
class Test; include TestMixin; end

# There is no hello method yet
test1 = Test.new
test1.hello
# NoMethodError: undefined method `hello' for #<Test:0x0000000110b92100>

# Let's add one
TestMixin.instance_eval { define_method(:hello) { "👋" } }

# New test instances now have a hello method
Test.new.hello
#=> "👋"

# BUT: test1 also now has a hello method, since method lookup has changed
test1.hello
#=> "👋"
```

app/models/profile.rb
```diff
@@ -3,12 +3,47 @@ class Profile < ApplicationRecord
...
+    display_email_on_profile: :email_public,
+    display_looking_for_work_on_profile: :looking_for_work_publicly,
+    git_lab_url: :gitlab_url,
+    linked_in_url: :linkedin_url,
+    recruiters_can_contact_me_about_job_opportunities: :contact_consent,
+    stack_overflow_url: :stackoverflow_url
+  }.with_indifferent_access.freeze
+
+  # Generates typed accessors for all currently defined profile fields.
+  def self.refresh_attributes!
```
citizen428, line 24: I changed the name from `refresh_store_accessors!` to `regresh_attributes!` since the former was reflecting an implementation detail while the new name reflects the domain event.

app/models/profile.rb
```diff
@@ -3,12 +3,47 @@ class Profile < ApplicationRecord
...
+        alias_method(attribute_name, getter) if attribute_name != getter
+      end
     end
   end
 
-  refresh_store_accessors!
+  refresh_attributes!
+
+  # Returns an array of all currently defined `store_attribute`s on `data`.
+  def self.attributes
```
citizen428, line 48: I'm using this now outside the class instead of directly accessing `Profile.stored_attributes[:data]`. The old way broke encapsulation as it relied on an implementation detail of `Profile`.

app/models/profile.rb
```diff
@@ -3,12 +3,55 @@ class Profile < ApplicationRecord
...
+    stored_attributes[:data]
+  end
+
+  # NOTE: @citizen428 This is a temporary mapping so we don't break DEV during
+  # profile migration/generalization work.
+  def self.mapped_attributes
```
citizen428, line 54: This will eventually be removed, but for now it's used for the sync from user to profile.

app/models/user.rb
```diff
@@ -20,45 +18,17 @@ class User < ApplicationRecord
       attr_accessor :_skip_creating_profile
 
       # All new users should automatically have a profile
-      after_create_commit :create_profile, unless: :_skip_creating_profile
+      after_create_commit -> { Profile.create(user: self, data: Profiles::ExtractData.call(self)) },
```
citizen428, line 21: Inlining this for now to keep this inline concern short. If we still need this in a bit we can move it into the class proper and give it a dedicated method again if we feel like it.

app/models/user.rb
```diff
@@ -20,45 +18,17 @@ class User < ApplicationRecord
...
+                          unless: :_skip_creating_profile
 
       # Keep saving changes locally for the time being, but propagate them to profiles.
       after_update_commit do
-        if (previous_changes.keys.map(&:to_sym) & User::PROFIE_FIELDS).present?
+        if previous_changes.keys.any? { |attribute| attribute.in?(Profile.mapped_attributes) }
```
citizen428, line 26: I'm not thrilled about this, but it's better than duplicating all this information in a constant here as it was previously.

citizen428:
@Ridhwana I made a lot of progress here today but I want to add more tests tomorrow, so I'm still leaving this as draft for now.

app/models/profile.rb
Ridhwana: Haha I was so confused that I almost googled what `regresh` means but I see you meant refresh 😂

maestromac:
closing and re-opening to re-trigger travis.

app/models/user.rb
```diff
@@ -7,58 +7,26 @@ class User < ApplicationRecord
...
           profile.update(data: Profiles::ExtractData.call(self))
         end
       end
-
-      # Define wrapped getters for profile attributes. We first try to get the
```
citizen428, line 32: After talking to @Zhao-Andy we decided to remove the getters now. They were always meant to be temporary and they proofed useful while verifying that we are syncing profiles, but they have outlived their usefulness for now. Going forward we will read and write from `User`, but sync to `Profile`. Then at one point we need to make a hard switch and stop acessing these values through `User` completely.

app/services/profiles/extract_data.rb
```diff
@@ -1,45 +1,12 @@
...
-      recruiters_can_contact_me_about_job_opportunities: :contact_consent,
-      stack_overflow_url: :stackoverflow_url
-    }.freeze
-
     def self.call(user)
       user_attributes = user.attributes.symbolize_keys!
 
-      direct_data = user_attributes.extract!(*DIRECT_ATTRIBUTES)
-      mapped_data = MAPPED_ATTRIBUTES.keys.zip(user_attributes.values_at(*MAPPED_ATTRIBUTES.values)).to_h
+      mapped_attributes = Profile::MAPPED_ATTRIBUTES
```
citizen428, line 6: This got simplified and many of the responsibilities moved into `Profile`. This also has the benefit of calling `refresh_attributes` through `Profile.attributes`.

app/models/profile.rb
```diff
@@ -3,12 +3,35 @@ class Profile < ApplicationRecord
...
+  }.with_indifferent_access.freeze
+
+  # Generates typed accessors for all currently defined profile fields.
+  def self.refresh_attributes!
     ProfileField.find_each do |field|
       store_attribute :data, field.attribute_name, field.type
     end
   end
 
-  refresh_store_accessors!
```
citizen428, line 13: We do no longer call this automatically. It breaks things in horrible ways when no DB exists yet and we try to create one (e.g. in CI or whenever setting up the app from scratch). Instead, we now call `refresh_attributes!` explicitly in the service object and other relevant places.

app/models/user.rb
```diff
@@ -7,58 +7,32 @@ class User < ApplicationRecord
...
-
       # NOTE: There are rare cases were we want to skip this callback, primarily
       # in tests. `skip_callback` modifies global state, which is not thread-safe
       # and can cause hard to track down bugs. We use an instance-level attribute
       # instead. See `spec/factories/profiles.rb` for an example.
       attr_accessor :_skip_creating_profile
 
+      # NOTE: used for not sync-ing back data to profiles when the profile got
+      # updated first. This will eventually be removed:
+      attr_accessor :_skip_profile_sync
```
citizen428, line 20: This allows us to skip the sync from user to profile when updating doing it through the service object (in which case the sync direction is profile -> user). 

I was debating removing this sync direction entirely, but 

1. I wanted to be able to deploy this before we change the `/settings/profile` page to use the service object and
2. ensure that there are no other places where this gets set, so we don't risk getting out of sync.

app/models/profile.rb
citizen428: Well, this was a nice idea and worked, but we got rid of the proxying for now, so 👋

app/models/profile.rb
Zhao-Andy: This would be overwriting ActiveRecord's `.attributes` right? Is this intended?

citizen428: AR's `attributes` is an instance method, this is a class method. The name was chosen intentionally, to reflect the relationship (i.e. a user's instance variable attributes are the attributes of their profile).

```ruby
User.attributes
# NoMethodError: undefined method `attributes' for #<Class:0x0000000114140a90>
User.first.attributes
#  User Load (2.3ms)  SELECT "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT $1  [["LIMIT", 1]]
# {
#                          "articles_count" => 5,
#                           "available_for" => nil,
```

citizen428:
This branch was open for quite a while so I merged back `master` just in case.

app/models/profile.rb
Zhao-Andy: Ahhh duh haha, thanks for clarifying.

app/services/profiles/extract_data.rb
```diff
@@ -1,45 +1,12 @@
...
     def self.call(user)
-      user_attributes = user.attributes.symbolize_keys!
+      user_attributes = user.attributes
 
-      direct_data = user_attributes.extract!(*DIRECT_ATTRIBUTES)
-      mapped_data = MAPPED_ATTRIBUTES.keys.zip(user_attributes.values_at(*MAPPED_ATTRIBUTES.values)).to_h
+      mapped_attributes = Profile::MAPPED_ATTRIBUTES
+      direct_attributes = Profile.attributes! - mapped_attributes.keys
+      direct_data = user_attributes.extract!(*direct_attributes)
+      mapped_data = mapped_attributes.keys.zip(user_attributes.values_at(*mapped_attributes.values)).to_h
```
mstruve, line 9: The `mapped_attributes.values` here need to be converted to strings otherwise you wont be able to find them in user_attributes which results in nothing getting set in the Profile. 
```ruby
mapped_data = mapped_attributes.keys.zip(user_attributes.values_at(*mapped_attributes.values.map(&:to_s))).to_h
```

I figured this out by pulling the branch down and testing it and seeing that fields were not getting updated in the Profile from the user. I think we should write at least 1 end-to-end test where we check if a mapped user value is updated, after saving the profile is also updated.

citizen428: It seems to be working fine for me:

[6] forem(main)>
[6] forem(main)> user.name
"Hi Molly 👋"
[7] forem(main)> user.profile.name
"Hi Molly 👋"
[8] forem(main)> user.update(name: "New name, who dis?")
...
[9] forem(main)> user.name
"New name, who dis?"
[10] forem(main)> user.profile.name
"New name, who dis?"
```

There also is a test for this:

```ruby
    it "propagates changes to the profile model", :aggregate_failures do
      expect do
        user.update(available_for: "profile migrations")
      end.to change { user.profile.reload.available_for }.from(nil).to("profile migrations")

      # Changes were also persisted in the users table
      expect(user.reload.available_for).to eq "profile migrations"
    end
```

citizen428: Thanks @mstruve! Fixed the code and added a test for a mapped attribute.

citizen428:
Had to re-run Travis to get the specs green.

Ridhwana (approved):

citizen428:
Making this mergeable again, since #10119 has now been deployed.

Zhao-Andy:
Merged and tracking deploy 👍

PR #9930, opened 2020-08-21
Refactor:Remove save_nil_hotness_scores Task

mstruve:
- [x] Refactor

## Description
Per some discussion in slack it was brought to my attention that this rake task was originally intended to patch a bug where Articles had nil hotness scores. Given this happened a long time ago we are going to try removing this task to see if the problem pops up again. If it does I will work to find the root cause so we dont need a rake task like this to patch it. 

alt_text


mstruve:
3 days and still no articles with nil hotness score so I think this is safe to remove. Will continue to monitor for the next couple of days.

rhymes (approved):

PR #9932, opened 2020-08-22
Add dependent clauses to relations for Organization and add cleanup scripts

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR adds many missing `dependent` clauses to `has_many` relations of the `Organization` model.
It also adds 7 cleanup scripts to either delete orphaned data or nullify the `*_id` column if the organization doesn't exist. The script reflect the type of relation between the two entities.

Please double check each script while reviewing.

We do have orphaned rows in our DB (and likely in other Forems DB's), see https://dev.to/admin/blazer/queries/197-organization-counts

The following PR will add foreign keys where necessary.

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/lib/data_update_scripts/20200821103125_nullify_orphaned_articles_by_organization.rb
+++ b/lib/data_update_scripts/20200821103125_nullify_orphaned_articles_by_organization.rb
@@ -1,0 +1,15 @@
+module DataUpdateScripts
+  class NullifyOrphanedArticlesByOrganization
+    def run
+      # Nullify organization_id for all Articles linked to a non existing Organization
+      ActiveRecord::Base.connection.execute(
+        <<~SQL,
+          UPDATE articles
+          SET organization_id = NULL
+          WHERE organization_id IS NOT NULL
+          AND organization_id NOT IN (SELECT id FROM organizations);
+        SQL
+      )
+    end
+  end
+end
--- a/lib/data_update_scripts/20200821103405_remove_orphaned_credits_by_organization.rb
+++ b/lib/data_update_scripts/20200821103405_remove_orphaned_credits_by_organization.rb
@@ -1,0 +1,14 @@
+module DataUpdateScripts
+  class NullifyOrphanedCollectionsByOrganization
+    def run
+      # Delete all User less Credits belonging to Organizations that don't exist anymore
+      ActiveRecord::Base.connection.execute(
+        <<~SQL,
+          DELETE FROM credits
+          WHERE user_id IS NULL
+          AND organization_id NOT IN (SELECT id FROM organizations);
+        SQL
+      )
+    end
+  end
+end
--- a/lib/data_update_scripts/20200821103405_remove_orphaned_credits_by_organization.rb
+++ b/lib/data_update_scripts/20200821103405_remove_orphaned_credits_by_organization.rb
@@ -1,6 +1,15 @@
 module DataUpdateScripts
-  class NullifyOrphanedCollectionsByOrganization
+  class RemoveOrphanedCreditsByOrganization
     def run
+      # Apparently we have a bunch of Credits that don't belong to either user or org
+      ActiveRecord::Base.connection.execute(
+        <<~SQL,
+          DELETE FROM credits
+          WHERE user_id IS NULL
+          AND organization_id IS NULL
+        SQL
+      )
+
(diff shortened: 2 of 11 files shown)


lib/data_update_scripts/20200822082229_remove_orphaned_notifications_by_organization.rb
```diff
@@ -0,0 +1,13 @@
+module DataUpdateScripts
+  class RemoveOrphanedNotificationsByOrganization
+    def run
+      # Delete all Notifications belonging to Organizations that don't exist anymore
+      ActiveRecord::Base.connection.execute(
+        <<~SQL,
+          DELETE FROM notifications
+          WHERE organization_id NOT IN (SELECT id FROM organization_id);
```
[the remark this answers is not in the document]
mstruve:
```suggestion
          WHERE organization_id NOT IN (SELECT id FROM organizations);
```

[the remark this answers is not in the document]
rhymes: LOL

lib/data_update_scripts/20200821103405_remove_orphaned_credits_by_organization.rb
```diff
@@ -0,0 +1,23 @@
+module DataUpdateScripts
+  class RemoveOrphanedCreditsByOrganization
+    def run
+      # Apparently we have a bunch of Credits that don't belong to either user or org
+      ActiveRecord::Base.connection.execute(
+        <<~SQL,
+          DELETE FROM credits
```
nickytonline, line 7: Is this because users/orgs were banned or are these orphaned credits a result of something else or no idea? 🙃

spec/models/organization_spec.rb
```diff
@@ -8,15 +8,15 @@
...
       it { is_expected.to have_many(:api_secrets).through(:users) }
-      it { is_expected.to have_many(:articles) }
-      it { is_expected.to have_many(:listings) }
-      it { is_expected.to have_many(:collections) }
-      it { is_expected.to have_many(:credits) }
-      it { is_expected.to have_many(:display_ads) }
-      it { is_expected.to have_many(:notifications) }
+      it { is_expected.to have_many(:articles).dependent(:nullify) }
+      it { is_expected.to have_many(:collections).dependent(:nullify) }
+      it { is_expected.to have_many(:credits).dependent(:restrict_with_error) }
```
nickytonline, line 13: Any reason this is not `:destroy`? Any reason `:restrict_with_error` handle is used? From what I read about it, it adds an error to the dependent object, but I'm not really understanding why it's only needed for credits.

lib/data_update_scripts/20200821103125_nullify_orphaned_articles_by_organization.rb
```diff
@@ -0,0 +1,15 @@
+module DataUpdateScripts
+  class NullifyOrphanedArticlesByOrganization
+    def run
+      # Nullify organization_id for all Articles linked to a non existing Organization
+      ActiveRecord::Base.connection.execute(
+        <<~SQL,
+          UPDATE articles
+          SET organization_id = NULL
+          WHERE organization_id IS NOT NULL
```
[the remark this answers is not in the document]
rhymes: Exactly, it's so that PostgreSQL selects less rows in the query and locks less rows in the update. A little trick

lib/data_update_scripts/20200821103405_remove_orphaned_credits_by_organization.rb
rhymes: I think they were just created for test: https://dev.to/admin/blazer/queries/199-orphan-credits - They don't have a cost nor a user nor an org. Probably before validations were added

spec/models/organization_spec.rb
rhymes: The logic here is: credits were purchased, so we want to avoid removing them automatically. The operator should be notified that credits exist associated to such an org and act accordingly. `:restrict_with_error` is helpful to build a UI on top of it so that you can tell the human what's going on.

PR #9961, opened 2020-08-23
Display all authentication providers in mailer footer

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently we display either Twitter or GitHub in the mailer footer to nudge a person to sign in with the correct provider. There are a few problems with that:

- a person can be signed in with more than one provider, so it's arbitrary to choose one or the other in the email
- soon we'll have more than one providers
- either or all of these two providers might be disabled in future Forems

This PR makes the footer a little bit more aware

It goes from:

>  Reminder: You used Twitter to authenticate your account, so use that to sign in if prompted. 

to

> Reminder: You used <b>GitHub and Twitter</b> to authenticate your account, so use any of those to sign in if prompted

if more than one is enabled.

1. start the PR
1. sign in with any of the authentication methods
1. check the footer of the following email: http://localhost:3000/rails/mailers/digest_mailer/digest_email

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/helpers/authentication_helper.rb
+++ b/app/helpers/authentication_helper.rb
@@ -15,13 +15,7 @@ module AuthenticationHelper
     end
   end
 
-  # Returns the list of enabled providers for the given user, defaults to `current_user`
-  def authentication_enabled_providers_for_user(user = nil)
-    user ||= current_user
-
-    providers = Authentication::Providers.enabled & user.identities.pluck(:provider).map(&:to_sym)
-    providers.sort.map do |provider_name|
-      Authentication::Providers.get!(provider_name)
-    end
+  def authentication_enabled_providers_for_user(user = current_user)
+    Authentication::Providers.enabled_for_user(user)
   end
 end
--- a/app/helpers/authentication_helper.rb
+++ b/app/helpers/authentication_helper.rb
@@ -15,8 +15,11 @@ module AuthenticationHelper
     end
   end
 
-  def current_user_authentication_enabled_providers
-    providers = Authentication::Providers.enabled & current_user.identities.pluck(:provider).map(&:to_sym)
+  # Returns the list of enabled providers for the given user, defaults to `current_user`
+  def authentication_enabled_providers_for_user(user = nil)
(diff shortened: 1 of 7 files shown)


app/helpers/authentication_helper.rb
```diff
@@ -15,8 +15,11 @@ def authentication_enabled_providers
     end
   end
 
-  def current_user_authentication_enabled_providers
-    providers = Authentication::Providers.enabled & current_user.identities.pluck(:provider).map(&:to_sym)
+  # Returns the list of enabled providers for the given user, defaults to `current_user`
+  def authentication_enabled_providers_for_user(user = nil)
```
citizen428, line 19: Ruby's default arguments are quite flexible, you can call other methods there. This way you don't need manually default to `current_user` on the next line.

```suggestion
  def authentication_enabled_providers_for_user(user = current_user)
```

app/helpers/authentication_helper.rb
```diff
@@ -15,8 +15,11 @@ def authentication_enabled_providers
...
+    user ||= current_user
+
+    providers = Authentication::Providers.enabled & user.identities.pluck(:provider).map(&:to_sym)
```
citizen428, line 22: I'm starting to wonder if this is really a helper, it doesn't seem to be directly related to view concerns. How about moving this into `Authentication::Providers.enabled_for_user(user)`? You could call that directly in `app/views/layouts/mailer.html.erb`, `app/views/users/_account.html.erb` and `app/views/users/_account_providers_settings.html.erb`, or wrap it up in helper that doesn't do more than pass through `current_user`, e.g.

```ruby
def authentication_enabled_providers_for_user(user = current_user)
  Authentication::Providers.enabled_for_user(user)
end
```

rhymes: done!

citizen428: Thanks @rhymes!

PR #9966, opened 2020-08-24
Mod Button placement fix on mobile

ludwiczakpawel:

- [x] Bug Fix

## Description

This PR fixes a position of mod actions button on article view on mobile - it will no longer cover the action bar at the bottom.

image

^ left: before. right: after.

## Added tests?

## Added to documentation?

- [x] no documentation needed

### DIFF
--- a/app/assets/stylesheets/moderators.scss
+++ b/app/assets/stylesheets/moderators.scss
@@ -163,13 +163,12 @@
 }
 
 .mod-actions-menu-btn {
-  transition: 0.3s all ease;
   position: fixed;
   bottom: var(--su-7);
+  margin-bottom: env(safe-area-inset-bottom);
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/moderators.scss
```diff
@@ -163,13 +163,12 @@
 }
 
 .mod-actions-menu-btn {
-  transition: 0.3s all ease;
```
jacobherrington, line 166: Is this transition property removed on purpose?

ludwiczakpawel: yeap: it wasn't really doing anything useful (or anything at all) (unless i missed something) and it was only causing a little visual jump on mobile.

PR #9967, opened 2020-08-24
Change to same_site: lax for session store

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We currently have configuration such that our cookies have the property of `SameSite=None`, which is more appropriate for widgets/embeds, etc.

This modifies our session cookie to have `SameSite=Lax`, which is the appropriate setting for our site's context.

### DIFF
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -14,6 +14,6 @@ Rails.application.config.session_store :redis_store,
                                        servers: servers,
                                        expire_after: expires_after,
                                        signed: true,
-                                       secure: true,
+                                       secure: ENV["FORCE_SSL_IN_RAILS"] == "true",
                                        same_site: :lax,
                                        httponly: true
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -13,5 +13,5 @@ Rails.application.config.session_store :redis_store,
                                        servers: servers,
                                        expire_after: expires_after,
                                        signed: true,
-                                       secure: ApplicationConfig["FORCE_SSL_IN_RAILS"] == "true",
+                                       secure: true,
                                        httponly: true
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -14,5 +14,5 @@ Rails.application.config.session_store :redis_store,
                                        servers: servers,
                                        expire_after: expires_after,
                                        signed: true,
-                                       secure: true,
+                                       secure: false,
                                        httponly: true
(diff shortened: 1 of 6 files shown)


config/initializers/session_store.rb
```diff
@@ -10,8 +10,10 @@
 servers = ApplicationConfig["REDIS_SESSIONS_URL"] || ApplicationConfig["REDIS_URL"]
 Rails.application.config.session_store :redis_store,
                                        key: ApplicationConfig["SESSION_KEY"],
+                                       domain: ApplicationConfig["APP_DOMAIN"],
                                        servers: servers,
                                        expire_after: expires_after,
                                        signed: true,
-                                       secure: ApplicationConfig["FORCE_SSL_IN_RAILS"] == "true",
+                                       secure: ENV["FORCE_SSL_IN_RAILS"] == "true",
```
mstruve, line 17: Why change from App config to ENV here?

config/initializers/session_store.rb
```diff
@@ -10,8 +10,10 @@
 servers = ApplicationConfig["REDIS_SESSIONS_URL"] || ApplicationConfig["REDIS_URL"]
 Rails.application.config.session_store :redis_store,
                                        key: ApplicationConfig["SESSION_KEY"],
+                                       domain: ApplicationConfig["APP_DOMAIN"],
```
mstruve, line 13: Will this affect the already stored keys at all?

config/initializers/session_store.rb
benhalpern: Oh I think I just changed that without really thinking about it because I was trying a bunch of different scenarios. I just changed it back to avoid an change that doesn't affect anything.

config/initializers/session_store.rb
benhalpern: I tested the deployment scenarios a bunch on https://community.benhalpern.com/ and things were fine. I'm pretty sure this just flips one attribute on the cookies without otherwise affecting behavior.

config/initializers/session_store.rb
```diff
@@ -10,8 +10,10 @@
...
                                        servers: servers,
                                        expire_after: expires_after,
                                        signed: true,
                                        secure: ApplicationConfig["FORCE_SSL_IN_RAILS"] == "true",
+                                       same_site: :lax,
```
rhymes, line 18: should this be set only if secure is true?

rhymes (commented):
Crossing fingers :D

config/initializers/session_store.rb
benhalpern: What I'm reading is that

> You must ensure that you pair SameSite=None with the Secure attribute

The same doesn't apply for `SameSite=Lax` eh?

PR #9972, opened 2020-08-24
Refactor:Move Credit Counter Cache Sync to Sidekiq

mstruve:
- [x] Refactor

## Description
I was going to dive in and see if we even still need this but decided not to waste time on it [since even the gem docs suggest running a sync periodically](https://github.com/magnusvk/counter_culture#manually-populating-counter-cache-values). To me that seems like a major weakness in the gem but one we can live with for now. 

## Added tests?
- [x] yes

alt_text


mstruve:
Not sure why storybook is failing since this doesn't have anything to do with it.

rhymes (approved):

PR #9978, opened 2020-08-24
Remove AWS SDK and private endpoints completely

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

DEV calls out to "private" endpoints for some score calculation and spam mitigation, but this is a PR where I'm pulling that logic into the app itself. It's not one-for-one for the slightly more complicated logic we had in our separate service— and how we catch spam might change or not be as smart _right away_, but I think getting the logic into the app and out of the closed endpoint will help us with the ongoing development of this feature— and simplify our infrastructure.

We had been doing this with an AWS Lambda call, which is a little over-engineering.

We wanted to "obfuscate the numbers" so that it's not clear to a potential spammer what to prey on, but I think with the work we've done to empower `trusted` users to flag content, and the _better_ work we can do if this code is all in the open, I think it's all around better to have this functionality in the open source as we look to spring up more Forems.

I kept the calculations in the same approximate realm of what they'd previously been, but a little _higher_ on avergae— in order to ensure old scores don't rise to the top if we have a new baseline for calculation. Score calc like this is a little arbitrary, but I hope with it in the open we can become a little more sophisticated and scientific working with the open source community.

I think the AWS SDK was a pretty heavy-weight gem, so we'll get a bit of a boost on boot up memory savings. More light-weight work :)


app/black_box/black_box.rb
```diff
@@ -32,14 +32,17 @@ def comment_quality_score(comment)
...
+      user = story.user
+      return 100 unless user
+      return 100 if user.trusted
+      return 100 if user.badges_count.positive?
 
-      function_caller.call("blackbox-production-spamScore",
-                           { story: story, user: story.user }.to_json).to_i
+      base_spaminess = 0
+      base_spaminess += 25 if
+        user.registred_at < ((user.github_created_at || user.twitter_created_at || 3.days.ago) + 2.days) &&
```
rhymes, line 43: there's a typo here in the name of the attribute `registered_at`

I wonder if you could use `user.registered_at.between?(min, max)` for readability

app/black_box/black_box.rb
```diff
@@ -50,20 +53,13 @@ def calculate_bonus_score(body_markdown)
...
-      else
-        # Simple calculation that takes in published at time and scores
-        # Same order of magnitude calculation as existing private function
-        # Gives credit to new articles and articles which score well from users and mods
-        article.published_at.to_i / 10_000 +
-          (article.score * 5000) +
-          (article.comment_score * 5000)
-      end
+    def last_mile_hotness_calc(article)
+      epoch_number = "2010-01-01 00:00:01".to_time.to_i
```
rhymes, line 57: as this never changes, maybe we can put it in a constant.

rhymes (commented):
Cool that we can remove one dependency 🔥

app/black_box/black_box.rb
benhalpern: Good call.

benhalpern:
Yeah @rhymes in trying to explain and coordinate the different scenarios where we may or may not have the outside dependency, I saw it as a major reason to just work to get rid of that.

Any communication burden _now_ is a complexity burden later.

app/black_box/black_box.rb
benhalpern: This is actually asking for the relationship between registered at and the recency of the social log in— which probably speaks to the readability issues in the first place.

rhymes: Also, I think we should change this as `github_created_at` and `twitter_created_at` might not exist for all Forems.

You should be able to do something like:

```ruby
enabled_providers = Authentication::Providers.enabled_for_user(user)
enabled_providers.map(&:user_created_at_field).map { |f| user.public_send(f) }
```

when we merge https://github.com/forem/forem/pull/9961

benhalpern:
Should be good to go

PR #9984, opened 2020-08-24
Refactor:Fetch New Github Repos From Sidekiq Cron

mstruve:
- [x] Refactor
- [x] Optimization

## Description
This moves the `github_repo_fetch_all` rake task to Sidekiq. I also took the opportunity to optimize the `update_to_latest` method for GithubRepo. 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/workers/podcasts/enqueue_get_episodes_worker.rb
+++ b/app/workers/podcasts/enqueue_get_episodes_worker.rb
@@ -1,0 +1,13 @@
+module Podcasts
+  class EnqueueGetEpisodesWorker
+    include Sidekiq::Worker
+
+    sidekiq_options queue: :medium_priority, retry: 10
+
+    def perform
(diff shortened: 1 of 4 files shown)


app/models/github_repo.rb
```diff
@@ -27,9 +27,8 @@ def self.upsert(user, **params)
   end
 
   def self.update_to_latest
-    where("updated_at < ?", 1.day.ago).find_each do |repo|
-      user = User.find_by(id: repo.user_id)
-      next unless user
+    where("updated_at < ?", 26.hours.ago).includes(:user).find_each do |repo|
```
mstruve, line 30: eager load users and extend lookback to 26 hours just in case an outage occurs, we have a little overlap. Plus it wont run at the exact same time each day since we poll for cron jobs every 10 seconds so without the overlap there is a chance that a repo could fall in that small window and get missed.

PR #9990, opened 2020-08-25
Add dependent clauses to Article and fix Comment validation

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR adds a few `dependent` clauses to `has_many` and the correspondent cleanup scripts.

I also fixed `Comment` validation as we currently have 8,987 with invalid article IDs and those comments should be valid and survive even if the article doesn't exist.

As usual, the PR with the foreign keys will be submitted separately

- we have orphaned buffer updates, comments with invalid ids, orphaned notifications subscriptions, pins and rating votes https://dev.to/admin/blazer/queries/197-articles-counts
- we have orphaned notifications https://dev.to/admin/blazer/queries/201-orphan-notifications-by-article

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -28,7 +28,16 @@ class Article < ApplicationRecord
   counter_culture :user
   counter_culture :organization
 
-  has_many :comments, as: :commentable, inverse_of: :commentable
+  has_many :buffer_updates, dependent: :destroy
+  has_many :comments, as: :commentable, inverse_of: :commentable, dependent: :nullify
+  has_many :html_variant_successes, dependent: :nullify
+  has_many :html_variant_trials, dependent: :nullify
+  has_many :notification_subscriptions, as: :notifiable, inverse_of: :notifiable, dependent: :destroy
+  has_many :notifications, as: :notifiable, inverse_of: :notifiable, dependent: :delete_all
+  has_many :page_views, dependent: :destroy
+  has_many :polls, dependent: :destroy
+  has_many :profile_pins, as: :pinnable, inverse_of: :pinnable, dependent: :destroy
+  has_many :rating_votes, dependent: :destroy
   has_many :top_comments,
            lambda {
              where(
@@ -39,15 +48,6 @@ class Article < ApplicationRecord
            as: :commentable,
            inverse_of: :commentable,
            class_name: "Comment"
-  has_many :profile_pins, as: :pinnable, inverse_of: :pinnable
-  has_many :buffer_updates, dependent: :destroy
-  has_many :html_variant_successes, dependent: :nullify
-  has_many :html_variant_trials, dependent: :nullify
-  has_many :notifications, as: :notifiable, inverse_of: :notifiable, dependent: :delete_all
-  has_many :notification_subscriptions, as: :notifiable, inverse_of: :notifiable, dependent: :destroy
-  has_many :rating_votes
-  has_many :page_views
-  has_many :polls, dependent: :destroy
 
   validates :slug, presence: { if: :published? }, format: /\A[0-9a-z\-_]*\z/,
                    uniqueness: { scope: :user_id }
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -14,9 +14,11 @@ class Comment < ApplicationRecord
   TITLE_HIDDEN = "[hidden by post author]".freeze
 
   belongs_to :commentable, polymorphic: true, optional: true
-  counter_culture :commentable
   belongs_to :user
+
+  counter_culture :commentable
   counter_culture :user
+
   has_many :mentions, as: :mentionable, inverse_of: :mentionable, dependent: :destroy
   has_many :notifications, as: :notifiable, inverse_of: :notifiable, dependent: :delete_all
   has_many :notification_subscriptions, as: :notifiable, inverse_of: :notifiable, dependent: :destroy
@@ -26,17 +28,19 @@ class Comment < ApplicationRecord
   before_create :adjust_comment_parent_based_on_depth
   after_create :after_create_checks
   after_create :notify_slack_channel_about_warned_users
-  after_update :update_descendant_notifications, if: :deleted
   after_update :remove_notifications, if: :deleted
+  after_update :update_descendant_notifications, if: :deleted
   before_destroy :before_destroy_actions
   after_destroy :after_destroy_actions
+
   after_save :synchronous_bust
   after_save :bust_cache
-  validate :permissions, if: :commentable
+
+  validate :published_article, if: :commentable
   validates :body_markdown, presence: true, length: { in: BODY_MARKDOWN_SIZE_RANGE }
   validates :body_markdown, uniqueness: { scope: %i[user_id ancestry commentable_id commentable_type] }
-  validates :commentable_id, presence: true
-  validates :commentable_type, inclusion: { in: COMMENTABLE_TYPES }
+  validates :commentable_id, presence: true, if: :commentable_type
+  validates :commentable_type, inclusion: { in: COMMENTABLE_TYPES }, if: :commentable_id
   validates :user_id, presence: true
 
   after_create_commit :record_field_test_event
@@ -255,7 +259,7 @@ class Comment < ApplicationRecord
     self.markdown_character_count = body_markdown.size
   end
 
-  def permissions
+  def published_article
     errors.add(:commentable_id, "is not valid.") if commentable_type == "Article" && !commentable.published
   end
 
--- a/spec/models/comment_spec.rb
+++ b/spec/models/comment_spec.rb
@@ -8,14 +8,19 @@ RSpec.describe Comment, type: :model do
   include_examples "#sync_reactions_count", :article_comment
 
   describe "validations" do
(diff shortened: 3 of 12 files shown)


app/models/article.rb
```diff
@@ -28,7 +28,16 @@ class Article < ApplicationRecord
   counter_culture :user
   counter_culture :organization
 
-  has_many :comments, as: :commentable, inverse_of: :commentable
+  has_many :buffer_updates, dependent: :destroy
```
rhymes, line 31: most of this is me sorting lines

rhymes: this clause was already there BTW

app/models/article.rb
```diff
@@ -28,7 +28,16 @@ class Article < ApplicationRecord
...
+  has_many :comments, as: :commentable, inverse_of: :commentable, dependent: :nullify
```
rhymes, line 32: this is important. When articles are deleted we leave `commentable_id` and `commentable_type` with the previous value, but those 2 values are invalid. The reason why we haven't noticed until now is that our validation is counter intuitive.

See my comment in the `comment.rb` file for further info

app/models/article.rb
```diff
@@ -28,7 +28,16 @@ class Article < ApplicationRecord
...
+  has_many :html_variant_successes, dependent: :nullify
```
rhymes, line 33: it was already there

app/models/article.rb
```diff
@@ -28,7 +28,16 @@ class Article < ApplicationRecord
...
+  has_many :html_variant_trials, dependent: :nullify
```
rhymes, line 34: it was already there

app/models/article.rb
```diff
@@ -28,7 +28,16 @@ class Article < ApplicationRecord
...
+  has_many :notification_subscriptions, as: :notifiable, inverse_of: :notifiable, dependent: :destroy
```
rhymes, line 35: it was already there

app/models/article.rb
```diff
@@ -28,7 +28,16 @@ class Article < ApplicationRecord
...
+  has_many :notifications, as: :notifiable, inverse_of: :notifiable, dependent: :delete_all
```
rhymes, line 36: it was already there and it's already enforced by a FK

app/models/article.rb
```diff
@@ -28,7 +28,16 @@ class Article < ApplicationRecord
...
+  has_many :page_views, dependent: :destroy
```
rhymes, line 37: bye bye page views (this is already enforced at the DB level by a foreign key)

app/models/article.rb
```diff
@@ -28,7 +28,16 @@ class Article < ApplicationRecord
...
+  has_many :polls, dependent: :destroy
```
rhymes, line 38: it was already there

app/models/article.rb
```diff
@@ -28,7 +28,16 @@ class Article < ApplicationRecord
...
+  has_many :profile_pins, as: :pinnable, inverse_of: :pinnable, dependent: :destroy
```
rhymes, line 39: bye bye pins

app/models/article.rb
```diff
@@ -28,7 +28,16 @@ class Article < ApplicationRecord
...
+  has_many :rating_votes, dependent: :destroy
```
rhymes, line 40: bye bye ratings

app/models/comment.rb
```diff
@@ -26,17 +28,19 @@ class Comment < ApplicationRecord
...
   after_save :bust_cache
-  validate :permissions, if: :commentable
+
+  validate :published_article, if: :commentable
   validates :body_markdown, presence: true, length: { in: BODY_MARKDOWN_SIZE_RANGE }
   validates :body_markdown, uniqueness: { scope: %i[user_id ancestry commentable_id commentable_type] }
-  validates :commentable_id, presence: true
-  validates :commentable_type, inclusion: { in: COMMENTABLE_TYPES }
+  validates :commentable_id, presence: true, if: :commentable_type
+  validates :commentable_type, inclusion: { in: COMMENTABLE_TYPES }, if: :commentable_id
```
rhymes, line 43: We declare `commentable` as `optional: true` on line 16 but then, until now, we requested `commentable_id` to be present.

We should check its presence only if `commentable_type` is present as well and we should check the valid type only if `commentable_id` is there

app/models/comment.rb
```diff
@@ -255,7 +259,7 @@ def set_markdown_character_count
     self.markdown_character_count = body_markdown.size
   end
 
-  def permissions
+  def published_article
```
rhymes, line 262: a bit clearer

spec/models/comment_spec.rb
```diff
@@ -28,14 +33,55 @@
...
-    it { is_expected.to validate_inclusion_of(:commentable_type).in_array(%w[Article PodcastEpisode]) }
 
-    it "is invalid if commentable is unpublished article" do
-      # rubocop:disable RSpec/NamedSubject
-      subject.commentable = build(:article, published: false)
-      expect(subject).not_to be_valid
-      # rubocop:enable RSpec/NamedSubject
+    # rubocop:disable RSpec/NamedSubject
+    describe "commentable" do
+      it "is invalid if commentable is an unpublished article" do
```
mstruve, line 39: Interestingly, I just found out that if someone unpublishes a post then you can no longer view those comments through the permalink. However, we do allow comments to be viewed from deleted posts. Wonder if this is something we should change. 🤔

rhymes: > Let me know if you need help with the notifications. Also if you give me a heads up before this goes out I could increase our STATEMENT_TIMEOUT to try and get the job to just do it all successfully.

Sure, let's sync up tomorrow UTC afternoon

> Interestingly, I just found out that if someone unpublishes a post then you can no longer view those comments through the permalink. However, we do allow comments to be viewed from deleted posts. Wonder if this is something we should change. 🤔

I don't know TBH, it's probably more a product decision. In a way it would be odd to look at a comment belonging to something that has been unpublished, but on the other side as you mentioned, we keep comments for deleted articles because they intrinsically belong to the commenter, not the article owner. 

There's a catch we haven't explored in either situation: what if a comment quotes a part of the article and the article owner doesn't want its content to be indexed anymore? Thus the unpublish or the delete. Food for thought

app/models/concerns/user_subscription_sourceable.rb
```diff
@@ -4,7 +4,7 @@ module UserSubscriptionSourceable
   # This all assumes there's an association with User under the column user_id.
 
   included do
-    has_many :user_subscriptions, as: :user_subscription_sourceable
+    has_many :user_subscriptions, as: :user_subscription_sourceable, dependent: :destroy
```
rhymes, line 7: I added this in this PR as it's the only model using the `user_subscriptions`. Those should go if the article goes away

rhymes:
@mstruve added this, https://github.com/forem/forem/pull/9990#pullrequestreview-475246751,  so I re-requested your review

app/models/concerns/user_subscription_sourceable.rb
mstruve: Should we have a warning about this? I feel like if I collected a bunch of emails from an article and then decided to remove that article I might be a little caught off guard when my list of emails also suddenly disappears. @atsmith813 any thoughts?

atsmith813: I don't think we want to destroy subscriptions. While we don't currently support displaying subscriptions for a deleted article in a view for an author, we still would want to retain the data. The subscriptions could still have value to the author, even on a deleted Article. In the future, we can show the subscriptions for _Deleted Article_ when we introduce more author analytics.

Here are some more thoughts on this: https://github.com/forem/forem/pull/9405/files/b01d5fec24bd3b07c48f23b4b2e5d3891896219c#diff-41409b89978eea2120acf9882a9b2ccf

rhymes: Ah ah I totally forgot that conversation and apparently never replied to your last question.

We currently have a "null: false" on the column, so there's no possibility for subscriptions to sit in the DB with `user_subscription_sourceable_id` as null.

I think we should switch to your proposal here:

>  I wonder if there's a quick win to validate the presence of user_subscription_sourceable only on create 🤔 . A subscription shouldn't be created without it, but it can be updated to be removed.

What do you both think?

atsmith813: I'm down with that approach 💯 .

rhymes: Ok, I'm going to set this back to draft and send a PR to change that

rhymes: https://github.com/forem/forem/pull/10033

rhymes:
@atsmith813 I switched to `nullify` after we merged and deployed https://github.com/forem/forem/pull/10033 :)

rhymes:
Set this to draft as it should be merge-deployed by the SRE team, given this comment https://github.com/forem/forem/pull/9990#pullrequestreview-474806347

PR #9992, opened 2020-08-25
🚀 Feature: Ability to upload images to connect by drag and drop

sarthology:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Github flavored drag and drop image uploader in connect. Just a small experiment. 😅

ezgif com-resize-2

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

alt_text

### DIFF
--- a/app/javascript/chat/chat.jsx
+++ b/app/javascript/chat/chat.jsx
@@ -8,6 +8,7 @@ import { h, Component } from 'preact';
 import PropTypes from 'prop-types';
 import { setupPusher } from '../utilities/connect';
 import debounceAction from '../utilities/debounceAction';
+import { dragDrop } from '../utilities/dragAndUpload';
 import { addSnackbarItem } from '../Snackbar';
 import {
   conductModeration,
@@ -1409,7 +1410,35 @@ export default class Chat extends Component {
       .getElementById('jumpback_button')
       .classList.remove('chatchanneljumpback__hide');
   };
-
+  handleImageDrop = (e) => {
+    document.getElementById('messagelist').style.opacity = 1;
+    dragDrop(e, this.handleImageSuccess, this.handleImageFailure);
+  };
+  handleImageSuccess = (res) => {
+    const mLink = `![alt text](${res.links[0]})`;
+    const el = document.getElementById('messageform');
+    const start = el.selectionStart;
+    const end = el.selectionEnd;
+    const text = el.value;
+    let before = text.substring(0, start);
+    before = text.substring(0, before.lastIndexOf('@') + 1);
+    const after = text.substring(end, text.length);
+    el.value = `${before + mLink} ${after}`;
+    el.selectionStart = start + mLink.length + 1;
+    el.selectionEnd = start + mLink.length + 1;
+    el.focus();
+  };
+  handleImageFailure = (e) => {
+    addSnackbarItem({ message: e.message, addCloseButton: true });
+  };
+  handleDragHover(e) {
+    e.preventDefault();
+    document.getElementById('messagelist').style.opacity = 0.3;
+  }
+  handleDragExit(e) {
+    e.preventDefault();
+    document.getElementById('messagelist').style.opacity = 1;
+  }
   renderActiveChatChannel = (channelHeader) => {
     const { state, props } = this;
 
@@ -1424,6 +1453,9 @@ export default class Chat extends Component {
               this.scroller = scroller;
             }}
             id="messagelist"
+            onDragOver={this.handleDragHover}
+            onDragExit={this.handleDragExit}
+            onDrop={this.handleImageDrop}
           >
             {this.renderMessages()}
             <div className="messagelist__sentinel" id="messagelist__sentinel" />
--- a/app/javascript/chat/chat.jsx
+++ b/app/javascript/chat/chat.jsx
@@ -1411,11 +1411,17 @@ export default class Chat extends Component {
       .classList.remove('chatchanneljumpback__hide');
   };
   handleImageDrop = (e) => {
-    document.getElementById('messagelist').style.opacity = 1;
-    dragDrop(e, this.handleImageSuccess, this.handleImageFailure);
+    e.preventDefault();
+    const { files } = e.dataTransfer;
+
+    const messageArea = document.getElementById('messagelist');
+    messageArea.classList.remove('opacity-25');
+    messageArea.classList.add('opacity-100');
+    dragDrop(files, this.handleImageSuccess, this.handleImageFailure);
   };
   handleImageSuccess = (res) => {
-    const mLink = `![alt text](${res.links[0]})`;
+    const { links, image } = res;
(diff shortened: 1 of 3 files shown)


app/javascript/utilities/dragAndUpload.js
```diff
@@ -0,0 +1,12 @@
+import { generateMainImage } from '../article-form/actions';
+import { validateFileInputs } from '../packs/validateFileInputs';
+
+export function dragDrop(e, handleImageSuccess, handleImageFailure) {
+  e.preventDefault();
+  const files = e.dataTransfer.files;
```
nickytonline, line 6:
```suggestion
  const { files } = e.dataTransfer;
```

app/javascript/chat/chat.jsx
```diff
@@ -1409,7 +1410,35 @@ export default class Chat extends Component {
       .getElementById('jumpback_button')
       .classList.remove('chatchanneljumpback__hide');
   };
-
+  handleImageDrop = (e) => {
+    document.getElementById('messagelist').style.opacity = 1;
+    dragDrop(e, this.handleImageSuccess, this.handleImageFailure);
+  };
+  handleImageSuccess = (res) => {
+    const mLink = `![alt text](${res.links[0]})`;
```
nickytonline, line 1418: Consider a complete name as this variable will get minified by Terser in webpack. I'm assuming `mLink` means markdown link?

app/javascript/chat/chat.jsx
```diff
@@ -1409,7 +1410,35 @@ export default class Chat extends Component {
...
+    el.selectionStart = start + mLink.length + 1;
+    el.selectionEnd = start + mLink.length + 1;
+    el.focus();
+  };
+  handleImageFailure = (e) => {
+    addSnackbarItem({ message: e.message, addCloseButton: true });
+  };
+  handleDragHover(e) {
+    e.preventDefault();
+    document.getElementById('messagelist').style.opacity = 0.3;
```
nickytonline, line 1436: Consider toggling [opacity utility CSS classes](https://storybook.forem.com/?path=/story/5-css-utility-classes-opacity--opacity-0).

app/javascript/chat/chat.jsx
```diff
@@ -1409,7 +1410,35 @@ export default class Chat extends Component {
...
+    let before = text.substring(0, start);
+    before = text.substring(0, before.lastIndexOf('@') + 1);
+    const after = text.substring(end, text.length);
+    el.value = `${before + mLink} ${after}`;
+    el.selectionStart = start + mLink.length + 1;
+    el.selectionEnd = start + mLink.length + 1;
+    el.focus();
+  };
+  handleImageFailure = (e) => {
+    addSnackbarItem({ message: e.message, addCloseButton: true });
```
nickytonline, line 1432: Snackbar! 😎

app/javascript/chat/chat.jsx
```diff
@@ -1409,7 +1410,35 @@ export default class Chat extends Component {
       .getElementById('jumpback_button')
       .classList.remove('chatchanneljumpback__hide');
   };
-
+  handleImageDrop = (e) => {
+    document.getElementById('messagelist').style.opacity = 1;
```
nickytonline, line 1414: Instead of using getElementById, leverage (P)React paradigms and consider using the [useRef](https://reactjs.org/docs/hooks-reference.html#useref) hook. You can see an example of this in action in the Forem code base here. 👇🏻 https://github.com/forem/forem/blob/2f09256713d6da36619ff761e887ec01687d8c60/app/javascript/packs/flagUserModal.jsx#L100.

app/javascript/utilities/dragAndUpload.js
```diff
@@ -0,0 +1,12 @@
+import { generateMainImage } from '../article-form/actions';
+import { validateFileInputs } from '../packs/validateFileInputs';
+
+export function dragDrop(e, handleImageSuccess, handleImageFailure) {
```
nickytonline, line 4: Although a small function, some tests can be added. It's easy to mock the Event object and the `handleImageSuccess` and `handleImageFailure` arguments can be easily mocked via `jest.fn()`.

app/javascript/chat/chat.jsx
sarthology: Yes, markdown link.

app/javascript/chat/chat.jsx
sarthology: ?

rhymes:
Looks great! Should we set a max width to the image so that it won't fill the entire width of the chat window?

app/javascript/chat/chat.jsx
rhymes, line 1418: is there a way we can generate the alt text from the file name? It's better than "alt text" and since JS has access to the filename beore it gets uploaded, we should be able to do that, right?

https://developer.mozilla.org/en-US/docs/Web/API/File

rhymes (commented):
Love how relatively simple this is! :D

app/javascript/chat/chat.jsx
nickytonline: I just meant that you were using the `<Snackbar />` component. 😄

app/javascript/chat/chat.jsx
sarthology: Can be done.

app/javascript/chat/chat.jsx
sarthology: Tried implementing it, giving some problems in case of hover. I think Creatref is not necessary here.
```
There are a few good use cases for refs:

    Managing focus, text selection, or media playback.
    Triggering imperative animations.
    Integrating with third-party DOM libraries.

Avoid using refs for anything that can be done declaratively.
```
Source: [React Docs](https://reactjs.org/docs/refs-and-the-dom.html)

app/javascript/chat/chat.jsx
sarthology: Done

sarthology:
Check if it's good to go now. I have one more small feature to add to this. 😅

app/javascript/chat/chat.jsx
```diff
@@ -1409,7 +1410,44 @@ export default class Chat extends Component {
       .getElementById('jumpback_button')
       .classList.remove('chatchanneljumpback__hide');
   };
+  handleImageDrop = (e) => {
+    e.preventDefault();
+    const { files } = e.dataTransfer;
 
+    const messageArea = document.getElementById('messagelist');
```
nickytonline, line 1417: This can be improved by using a `ref` as mentioned previously, but not a blocker.

nickytonline:
Thanks for the PR @sarthology! A short post on Forem would be awesome about this feature if you have time.

sarthology:
Definitely 😊

PR #10001, opened 2020-08-25
Include APP_NAME in Pusher channels

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

By including the `APP_NAME` in Pusher channels, it allows a single pusher credential set to work across many apps if the owner of multiple forems wants to use all the same keys for simplicity. Since the keys are private, it's technically fine to do this. If a forem chooses to use different keys in different apps this won't get in the way of that scenario.

Pusher channels and keys are not persisted in any way, so we can change these keys as often as we want. There is a bit of lack of DRY here, but I felt like the find/replace approach was fine just to make sure this scenario is fine.


.env_sample
```diff
@@ -1,6 +1,7 @@
 # App domain + Protocol setting for development
 export APP_DOMAIN="localhost:3000"
 export APP_PROTOCOL="http://"
+export APP_NAME="forem_local"
```
rhymes, line 4: what if we use the hostname or app domain instead of a colloquial name?

benhalpern:
> Will this in any way affect existing channels and things like showing past messages?

@mstruve Nope. Channels get spun up and down as needed. They only exist as a live pipe to subscribe to and unsubscribe from, no persistence or memory.

The only possible issue this could cause is that live channels could be inconsistent during the time of deployment, but it would be fixed on the next page change so wouldn't be a problem.

.env_sample
benhalpern: It doesn't _really_ matter because we could probably change this if we made that change, but the way we're using `APP_NAME` in production as a unique identifier which sticks with the forem forever whereas the domain could hypothetically be a bit more fluid, or one app could have multiple domains (not something we support in any capacity now, but in some scenario like `smile.amazon.com` functionality or whatever, I just feel like systems-wise this is the proper thing.

FYI `APP_NAME` might be new to local, but had already been an environment variable that existed in Heroku land and we adopted as a variable in Forem Cloud universe, in case that info is not totally clear as well.

rhymes: thanks for the explanation, makes sense!

Issue #9994, opened 2020-08-25
Generalise: Site mascot footer image


Link2Twenty:
**Describe the bug**

The site mascot currently has a fixed size (52x120) this means if a new forem changes their mascot to something that doesn't fit it will looks stretched

**To Reproduce**

Scroll to the footer on https://forem.dev
Look at Sloan

**Expected behavior**

The forem maintainer should be able to have any image and not have to match a specific size

**Screenshots**

![poor sloan](https://user-images.githubusercontent.com/3534427/91195220-bd424d80-e6f0-11ea-9f35-48b29b095397.png)

**Desktop (please complete the following information):**

- OS, version: Windows 10 - 18363.1016
- Browser, version: Chrome 84.0.4147.135

**Additional context**

Perhaps removing the height would be enough?
https://github.com/forem/forem/blob/0e2fc110b2a48002d6115e64c60a8129a4c27fdf/app/views/layouts/_footer.html.erb#L50-L55

jacobherrington:
@Link2Twenty I'll work on changing this since generalization is currently a focus area for us.

Do you feel like the size values should be configured by an admin user (either in the configuration UI or the image's size), or do you think it should be generated dynamically?

PR #10004, opened 2020-08-26
Allow greater flexibility for footer mascot image

jacobherrington:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This change allows admins to use footer mascot images of any size without coercing them to the default values. They can manage the default values in site config.

I think ideally this would be removed and replaced with some sort of custom CSS/theming system down the road, but I'm not sure we should anticipate that until we are ready to dedicate resources to it. When that time comes this will be easy to yank out!

Closes https://github.com/forem/forem/issues/9994

image

Adds these too fields!

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

*note:* adds site config values

### DIFF
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -144,11 +144,11 @@ module Constants
         placeholder: "https://image.url"
       },
       mascot_footer_image_width: {
-        description: "The footer image will be resized to this value, defaults to 52",
+        description: "The footer mascot width will resized to this value, defaults to 52",
         placeholder: ""
       },
       mascot_footer_image_height: {
-        description: "The footer image will be resized to this value, defaults to 120",
+        description: "The footer mascot height will be resized to this value, defaults to 120",
         placeholder: ""
       },
       mascot_image_description: {
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -143,6 +143,14 @@ module Constants
         description: "Special cute mascot image used in the footer.",
         placeholder: "https://image.url"
       },
+      mascot_footer_image_width: {
+        description: "The footer image will be resized to this value, defaults to 52",
+        placeholder: ""
+      },
+      mascot_footer_image_height: {
+        description: "The footer image will be resized to this value, defaults to 120",
+        placeholder: ""
+      },
       mascot_image_description: {
         description: "Used as the alt text for the mascot image",
         placeholder: ""
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -135,6 +135,8 @@ module Admin
         mascot_image_description
         mascot_image_url
         mascot_footer_image_url
+        mascot_footer_image_width
+        mascot_footer_image_height
         mascot_user_id
       ]
     end
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -75,6 +75,8 @@ class SiteConfig < RailsSettings::Base
   field :mascot_image_url, type: :string
   field :mascot_image_description, type: :string, default: "The community mascot"
   field :mascot_footer_image_url, type: :string
+  field :mascot_footer_image_width, type: :integer, default: 52
+  field :mascot_footer_image_height, type: :integer, default: 120
 
   # Meta keywords
   field :meta_keywords, type: :hash, default: {
--- a/app/views/admin/configs/show.html.erb
+++ b/app/views/admin/configs/show.html.erb
@@ -590,6 +590,22 @@
             </div>
 
             <div class="form-group">
+              <%= admin_config_label :mascot_footer_image_width %>
+              <%= f.text_field :mascot_footer_image_width,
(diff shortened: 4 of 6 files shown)


rhymes (approved):
Looks good to me, I pinged @ludwiczakpawel in case there's something we're missing in the whole design scheme

app/lib/constants/site_config.rb
```diff
@@ -143,6 +143,14 @@ module SiteConfig
         description: "Special cute mascot image used in the footer.",
         placeholder: "https://image.url"
       },
+      mascot_footer_image_width: {
+        description: "The footer image will be resized to this value, defaults to 52",
```
ludwiczakpawel, line 147: Shouldn't that say something like:

> The footer **mascot width** will be resized to this value, defaults to 52

?

Same for the height couple lines below

ludwiczakpawel (approved):
Left one comment in the code, not a blocker though.

Overall I think this is a very good addition and I wish it was easier to do this kind of a thing with CSS  only (because we don't want to add too many granular settings to the config). I'll try to come up with some more "automated" solution, but in the meantime I think this is perfect.

Ridhwana:
@jacobherrington I'm just thinking through this, and I'm not sure we should add widths and heights of images to the Site Config page. It feels like it wouldn't scale well and I'm not sure that heights and widths of images is something that should be configured.

Some questions:
@ludwiczakpawel is there no css solution that will help us to proportion the mascot better without hardcoding heights and widths in the config? Would using an svg instead of a png help with the scales and proportions?

jacobherrington:
@Ridhwana I agree wholeheartedly!

> I think ideally this would be removed and replaced with some sort of custom CSS/theming system down the road, but I'm not sure we should anticipate that until we are ready to dedicate resources to it. When that time comes this will be easy to yank out!

My reasons for including this as a config value: 1) Someone has requested this in an issue, implying that their looking for an immediate solution. 2) I have no idea what the product vision for "Forem themes" looks like just yet.

I'd *really* like to talk about how we build a platform that is deeply customizable by a non-engineer or (at minimum) with nothing more than knowledge of CSS. I feel like this is a suitable solution in the meantime as it gives the functionality to the user who requested it, in a way that is easily reversible (deprecating this would be a quick two-part process of warning the user, then removing a few lines of code), without making assumptions about a major feature that doesn't exist yet.

Does this make sense or am I putting the cart before the horse? We might benefit a lot from *not* merging this and instead working on the product vision concerning customization of individual instances.

app/lib/constants/site_config.rb
jacobherrington: Yes, it could be either, since this is the "mascot image." I think your suggestion is more coherent, so I'll make that change!

Ridhwana:
> @Ridhwana I agree wholeheartedly!
> 
> > I think ideally this would be removed and replaced with some sort of custom CSS/theming system down the road, but I'm not sure we should anticipate that until we are ready to dedicate resources to it. When that time comes this will be easy to yank out!
> 
> My reasons for including this as a config value: 1) Someone has requested this in an issue, implying that they're looking for an immediate solution. 2) I have no idea what the product vision for "Forem themes" looks like just yet.
> 
> I'd _really_ like to talk about how we build a platform that is deeply customizable by a non-engineer or (at minimum) with nothing more than knowledge of CSS. I feel like this is a suitable solution in the meantime as it gives the functionality to the user who requested it, in a way that is easily reversible (deprecating this would be a quick two-part process of warning the user, then removing a few lines of code), without making assumptions about a major feature that doesn't exist yet.
> 
> Does this make sense or am I putting the cart before the horse? We might benefit a lot from _not_ merging this and instead working on the product vision concerning customization of individual instances.

@jacobherrington thanks for the thoughtful response. I think as you pointed out we probable should start looking into the theming customization, but I do agree with you that this is easily reversible, so let's merge this but try and reverse it as soon as we can. I'll bring up the theming in the next forem meeting :)

Ridhwana (approved):

ludwiczakpawel:
@Ridhwana I'll make some research... Requiring SVG for footer mascot would definitely help. Width & Height are especially needed for PNG/JPG when you upload, for example, 200x200px image but you define the size 100x100px to make it look good on retina. 

So if we required SVG, then that ^ wouldn't be a problem. But that gonna need a bit more research before I let you quote me on that :D I'm happy to do so though.

Ridhwana:
> @Ridhwana I'll make some research... Requiring SVG for footer mascot would definitely help. Width & Height are especially needed for PNG/JPG when you upload, for example, 200x200px image but you define the size 100x100px to make it look good on retina.
> 
> So if we required SVG, then that ^ wouldn't be a problem. But that gonna need a bit more research before I let you quote me on that :D I'm happy to do so though.

Perfect, thanks @ludwiczakpawel! If the research turns out positive, I think it's acceptable to require svg uploads for these logo's.

PR #10007, opened 2020-08-26
Introduce ProfileFieldGroup model

citizen428:

- [X] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Based on recent discussions we decided to make the profile groups their own model. This PR 

1. adds the model,
2. updates `ProfileField` to use the new model,
3. Updates/adds specs for the same, and
4. adds a data update script to migrate the existing profile fields.

Part of #6994

1. Run the migrations locally
2. Run the data update script
3. Verify that all profile fields are now in the correct groups

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed


app/models/profile_field_group.rb
```diff
@@ -0,0 +1,5 @@
+class ProfileFieldGroup < ApplicationRecord
+  has_many :profile_fields, dependent: :nullify
```
citizen428, line 2: I went for `nullify` here because I wasn't 100% sure if all fields always need to belong to a group.

app/services/profile_fields/field_definition.rb
```diff
@@ -8,8 +8,8 @@ def call
         new.add_fields
       end
 
-      def group(group = nil)
-        @group = group
+      def group(description = nil)
```
citizen428: I'm glad I made this little DSL, so I didn't have to go and make too many changes elsewhere.

lib/data_update_scripts/20200826075937_migrate_profile_field_groups.rb
```diff
@@ -0,0 +1,13 @@
+module DataUpdateScripts
+  class MigrateProfileFieldGroups
+    def run
+      profile_field_groups = Hash.new do |hash, key|
+        hash[key] = ProfileFieldGroup.find_or_create_by(name: key)
+      end
+
+      ProfileField.find_each do |profile_field|
+        profile_field.update(profile_field_group: profile_field_groups[profile_field.group])
```
citizen428, line 9: Question regarding data update scripts: We will want to remove the `group` from `ProfileField` soon, which means this script will no longer work. Have we encountered a situation like this in the past? If yes, how did we deal with it. If no, is a simple guard like `return unless "group".in?(ProfileField.column_names)` at the top of `call` sufficient?

app/views/admin/profile_fields/index.html.erb
```diff
@@ -1,36 +1,20 @@
 <main id="profileFields">
-  <article class="crayons-card grid p-6 mb-6 gap-1">
```
Ridhwana, line 2: I've removed the creating of new fields and groups temporarily. We can currently edit what we have on the interface. The creation of these will require some rework + implementation of the updated designs (which caters for this new model) which I don't believe needs to hold this PR back. https://www.figma.com/file/oo9B9YSe8aWkuLKPBLc0g6/Profile-Generalization?node-id=29%3A718

Ridhwana (approved):

Ridhwana:
I believe that this is ready to be reviewed. @citizen428 I'll let you move it out of Draft as I'm not sure if you need to add any more instructions. I've update the admin page to show the groups within its new model. However, I've removed the creating of new fields and groups temporarily. We are currently just able to edit what we have on the interface. The creation of these will require some rework + implementation of the updated designs (which caters for this new model) which I don't believe needs to hold this PR back. https://www.figma.com/file/oo9B9YSe8aWkuLKPBLc0g6/Profile-Generalization?node-id=29%3A718

lib/data_update_scripts/20200826075937_migrate_profile_field_groups.rb
mstruve: We have dealt with it a couple of different ways. you can [add a guard clause](https://github.com/forem/forem/blob/master/lib/data_update_scripts/20200226193303_backfill_column_for_broadcasts.rb#L4) or you can [comment it out](https://github.com/forem/forem/blob/master/lib/data_update_scripts/20200326145114_re_index_feed_content_to_elasticsearch.rb#L4) or you could [even delete it](https://github.com/forem/forem/blob/master/lib/data_update_scripts/20200729120730_remove_orphaned_ahoy_events.rb). I personally like having the history so I would go for option 1 or 2.

.travis.yml
```diff
@@ -57,6 +57,7 @@ script:
   - bundle config set path 'vendor/bundle'
   - bin/ci-bundle
   - yarn install --frozen-lockfile
+  - createdb PracticalDeveloper_test
```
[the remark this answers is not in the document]
citizen428: Yes, I'll remove it again now. Desperate times...

lib/data_update_scripts/20200826075937_migrate_profile_field_groups.rb
citizen428: Thanks @mstruve! I'm almost always in favor of preserving history, so guard clause it is.

citizen428:
Thanks @Ridhwana! Leaving it in draft was ok, I still made same minor changes this morning.

lib/data_update_scripts/20200826075937_migrate_profile_field_groups.rb
citizen428: Guard clause added.

db/migrate/20200826072722_add_profile_field_group_to_profile_field.rb
```diff
@@ -0,0 +1,7 @@
+class AddProfileFieldGroupToProfileField < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :profile_fields, :profile_field_group, index: { algorithm: :concurrently }
```
rhymes, line 5: BTW: this does not add the foreign key, only the column, you shouldp probably add a FK as well to the `profile_field_group_id`

rhymes (commented):
Haven't tested it yet but found a thing

db/migrate/20200826072722_add_profile_field_group_to_profile_field.rb
citizen428: Excellent point, will do. Thanks!

citizen428: @rhymes I added the foreign key now. The index is probably useless as we only have 10s of profile fields with a handful of groups. We may potentially remove it in the future.

citizen428:
Added the foreign key as requested and rebased against master to resolve the conflict with `db/schema.rb`. I'll deploy this when specs are green.

PR #10008, opened 2020-08-26
Add "required" to the search form

J2TEAM:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Add `required` to the search form to prevent this:

image

image

image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/views/layouts/_top_bar.html.erb
+++ b/app/views/layouts/_top_bar.html.erb
@@ -19,7 +19,7 @@
 
     <div class="top-bar--search" id="top-bar--search">
       <form acceptCharset="UTF-8" method="get" action="/search">
-        <input class="top-bar--search-input crayons-textfield" type="text" name="q" id="nav-search" placeholder="Search..." autoComplete="off" />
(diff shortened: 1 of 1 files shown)


ludwiczakpawel:
I think it would make a bit more sense to actually submit that form and take user to search results page - even though no query was entered, user will see some results there. Especially we're planning to improve empty search results page a bit to be a bit more of a "discovery" type of page.. Think of [Twitter's explore](https://twitter.com/explore) page.

J2TEAM:
> I think it would make a bit more sense to actually submit that form and take user to search results page - even though no query was entered, user will see some results there. Especially we're planning to improve empty search results page a bit to be a bit more of a "discovery" type of page.. Think of [Twitter's explore](https://twitter.com/explore) page.

Oh, it's good! Thanks for your info!

PR #10014, opened 2020-08-26
Make confirmation text a required field

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Right now, you can submit the config form without including the required confirmation text. When you do this, it's a bit frustrating because it's such a large form and you can lose your inputs. 

~Also, this returns an unauthorized status which isn't really true because you can hit this error with all the necessary permissions, so I changed it to return a 422 (unprocessable entity) because most accurately (imo) you're missing a required parameter.~

This change now includes both frontend validation and more specific error messaging for the backend validation.

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -67,14 +67,14 @@ module Admin
       )
     end
 
-    def confirmation_text_error
-      render json: { Error: "confirmation text did not match" }, status: :unprocessable_entity
+    def raise_confirmation_mismatch_error
+      raise ActionController::BadRequest.new, "The confirmation key does not match"
     end
 
     def extra_authorization_and_confirmation
       not_authorized unless current_user.has_role?(:single_resource_admin, Config) # Special additional permission
       confirmation_message = "My username is @#{current_user.username} and this action is 100% safe and appropriate."
-      confirmation_text_error if params[:confirmation] != confirmation_message
+      raise_confirmation_mismatch_error if params.require(:confirmation) != confirmation_message
     end
 
     def clean_up_params
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -4,6 +4,10 @@ module Admin
 
     before_action :extra_authorization_and_confirmation, only: [:create]
 
+    def show
+      @confirmation_text = confirmation_text
+    end
+
     def create
       clean_up_params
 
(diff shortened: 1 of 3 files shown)


Ridhwana (changes requested):
@jacobherrington when I click Submit, it's showing me the json error as per my screenshot. Are you seeing different behaviour?

Screenshot 2020-08-26 at 15 02 48

jacobherrington:
@Ridhwana Does this happen when you input the confirmation text correctly, or if you remove the required attribute and submit?

Ridhwana:
> @Ridhwana Does this happen when you input the confirmation text correctly, or if you remove the required attribute and submit?

It happens when I enter the confirmation text incorrectly.

jacobherrington:
@Ridhwana Yeah! It looks like there is a spec failure regarding the same issue. I'll look into it. thanks :smiley:

jacobherrington:
@Ridhwana I just pushed a commit that will still fail CI, but I'm curious if you think this approach is better? It raises more specific errors that describe the missing or incorrect parameter.

Ridhwana:
> @Ridhwana I just pushed a commit that will still fail CI, but I'm curious if you think this approach is better? It raises more specific errors that describe the missing (I used params.require() which should handle that case) or incorrect parameter (I used a BadRequest error, which may not be perfect, but it's pretty close without defining a custom error for this one case).

I think this is much better than what we had before this PR.

Ridhwana:
A win in terms of user experience for me would be to make it more dynamic and treat it just like a field validation. i.e. when I enter the correct confirmation text I should be able to stay on the same page with all my info still entered (if possible) and a little red notice pop up to say that I entered the confirmation text incorrectly. But, that seems more complicated and I feel like this is good for now.

jacobherrington:
@Ridhwana Updated to do exactly that!

It should now also pass specs and the improved error messages will help if frontend validation fails.

Ridhwana (approved):
@jacobherrington this solution looks and works great 😊, its so much more intuitive *and* the user doesn't lose their information or get navigated away from the page 🎉

The one point that I do want to mention is that previously they would have to provide their username, but now we're showing them their username by default, so they can essentially just copy and paste the sentence as is. I'm not sure if it was by design to not show their username in that prompt initially? but I don't see any real harm in this and I've approved.

PR #10015, opened 2020-08-26
Adding Crayons to Admin

ludwiczakpawel:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This is a quite big PR but only at first glance... It adds a bit of Crayons styling to the entire Admin section (that's why there's so many files affected) but I tried to not screw anything  and limit myself mostly to adding a bit of classes and wrappers.

image

image

image

image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [x] no documentation needed


app/assets/stylesheets/internal/layout.scss
```diff
@@ -1,112 +1,112 @@
-@import '../crayons';
-@import '../scaffolds';
-@import '../variables';
-
-#siteConfigBodyContainer {
```
Ridhwana, line 5: @ludwiczakpawel removing these styles breaks the ui. 
For example on the site config page `.site-config__required` is used for the required tag, etc. Same with the others.
It doesn't look like they've been replaced in some other way?

Zhao-Andy:
😍 😍 😍 can't wait!!

app/views/admin/comments/index.html.erb
```diff
@@ -1,59 +1,57 @@
-<main>
```
Ridhwana, line 1: Perhaps @jacobherrington can comment on this, but I remember reviewing [this PR](https://github.com/forem/forem/pull/9927/files) where <main> was added to fix general accessibility warnings

app/views/admin/mods/index.html.erb
```diff
@@ -1,63 +1,58 @@
...
 
-  <%= paginate @mods, theme: "internal" %>
+<%= paginate @mods, theme: "internal" %>
 
-  <table class="table">
-    <thead>
+<table class="crayons-table" width="100%">
+  <thead>
+    <tr>
+      <th>ID</th>
```
Ridhwana, line 28: **(question)**: @ludwiczakpawel I've seen scope=col removed, I understand those are for screen readers, so does adding it to the thead remove the need for the scope attr ?

Ridhwana (commented):
I reviewed about halfway thus far. Some comments on it as I review so that they're not bunched up.

app/views/admin/comments/index.html.erb
nickytonline: The `<main />` tag is used to identify the primary content of the page. It can be the `<main />` tag or `role='main'`. See the [Main Landmark](https://www.w3.org/TR/wai-aria-practices-1.1/examples/landmarks/main.html) documentation for more info. Having this tag or role helps assistive technologies navigate a page.

Ridhwana: Thanks Nick 😊 `Sounds like we can add the `role='main'` back at least so as to not reverse the work done in the PR to fix the general accessibility warnings in admin. I don't remember seeing these warnings, when/where/how do we see them?

ludwiczakpawel: Hmm, I do have this tag but somewhere else now - in [general layout](https://github.com/forem/forem/pull/10015/files#diff-c861d9ec1cc861e6b11cab02955eaaeaR42) - do you think that's wrong?

app/views/admin/mods/index.html.erb
ludwiczakpawel: These shouldn't be removed. I'll bring them back.

app/assets/stylesheets/internal/layout.scss
ludwiczakpawel: Let me double check that.

app/views/admin/mods/index.html.erb
Ridhwana: Cool I think they're in a couple of places 😊

ludwiczakpawel: okeeeey, these should be back now..

app/views/admin/comments/index.html.erb
Ridhwana: Awesome, I'm not sure exactly how this works without doing more research, so maybe @jacobherrington or @nickytonline can verify that the changes make sense.

app/assets/stylesheets/internal/layout.scss
ludwiczakpawel: You were right, good catch. I thought it would cause a harm. Anyway, I brought some of the styling back to the [admin.scss](https://github.com/forem/forem/pull/10015/files#diff-111bd41e9b25b9eaafc015184274028d) file and will gradually remove/replace that with Crayons stuff.

Ridhwana: @ludwiczakpawel great but there's still more that we're using, these I know are being used offhand without ay further research:
```
#siteConfigBodyContainer {
  padding: 20px;
}

#siteConfig .card-header {
  cursor: pointer;
}

#siteConfig .card-header__toggle {
  font-size: 12px;
  color: var(--secondary);
}

#siteConfig .notice {
  margin: 20px 0;
  font-weight: bold;
  color: $dark-purple;
  font-size: 14px;
}

#profileFields .card-header {
  cursor: pointer;
}

.profile__group-toggle {
  text-decoration: underline;
  cursor: pointer;
  font-size: 12px;
  color: grey;
  margin: 1em 0;
  text-align: end;
}
```
Perhaps we should add them all back to avoid regressing the UI and then in a subsequent PR we can remove those that we're sure are not being used and move others to crayons if thats possible.

ludwiczakpawel:
```
#siteConfigBodyContainer {
  ...
}
```

This has been replaced with some new styling already. It should still look good.

```
#siteConfig .card-header {
  ...
}

#profileFields .card-header {
  ...
}
```

We have button that toggles the card body, so even if the whole header is clickable it doesn't need a pointer cursor. I quite intentionally removed it, because eventually ONLY button should toggle the content, especially in some other cases we may have more buttons than just a "toggle" (remember our designs for Profile Fields? I know it's a different view, but the same UI pattern eventually)

```
#siteConfig .notice {
  ...
}
```

This has been replaced with `.crayons-notice` element.

.profile__group-toggle {
  ...
}

You're right - I'll bring that back in.

app/views/admin/configs/show.html.erb
```diff
@@ -1,952 +1,928 @@
...
-                               :official_name,
-                               authentication_enabled_providers.map(&:provider_name),
-                             ),
-                             multiple: true,
-                             class: "form-control selectpicker" %>
-              <div class="alert alert-info"><%= Constants::SiteConfig::DETAILS[:authentication_providers][:description] %></div>
-            </div>
-            <div class="notice">
-              Notice: Changing authentication keys will not take affect until devops restarts the app (about once per day). This will soon be instant.
-            </div>
```
Ridhwana, line 135: @ludwiczakpawel looks like this may have been reverted back to an <h4> when you merged, hence I was not seeing the relevant styles being applied. 

Can you please revert to this, and maybe change notice to .crayons-notice as you mention [here]([https://github.com/forem/forem/pull/10015#discussion_r477358925])

app/assets/stylesheets/internal/layout.scss
Ridhwana: @ludwiczakpawel sounds good. Thanks. Regarding the toggles, it makes complete sense especially with the new profile design. One thing to note though is we don't use a button on the inner card headers, hence perhaps we should add a cursor pointer to at least the text?

Screenshot 2020-08-26 at 16 59 45
 
Happy with this being in a subsequent PR though. 

Also made a note elsewhere in the code about .notice and a git merge :) https://github.com/forem/forem/pull/10015/files#r477367526

nickytonline:
Any chance we can break up this PR @ludwiczakpawel? It's not user/public facing except for administrators, so they can probably temporarily live with the admin UI being converted bit by bit to crayons.

ludwiczakpawel:
@nickytonline 



how many PRs do you want? :d

nickytonline:
> @nickytonline
>
> how many PRs do you want? :d

Doesn't need to be a specific number. 😄 If this can be broken up a bit, it will be easier to test/review.

ludwiczakpawel:
@nickytonline I'm not really sure how to split it into smaller PRs - if you can help me out with that, I could do it... 

But honestly, even though I know it's a big PR, 85-90% of it is basically adding/removing some wrappers and classes - not really changing any logics or rails code.

nickytonline:
I'll take a peek at this this morning.

app/views/admin/configs/show.html.erb
ludwiczakpawel: Done.

app/assets/stylesheets/internal/layout.scss
ludwiczakpawel: Yeap, agree - I'll make it a separate PR later on.

FYI: the changes I've made with current PR are basically very first step for Admin visual improvements :) More are coming!

db/schema.rb
```diff
@@ -484,7 +484,6 @@
     t.datetime "updated_at", null: false
```
fdoxyz, line 484: I think these changes shouldn't be included unless I'm missing right? Side effect of jumping between branches every now and then for me 😄

ludwiczakpawel: uhh, yea - you're right. i'll revert that one... i'd love this file to be in gitignore :) - i catch myself adding this by mistake too often :/

ludwiczakpawel: Done. Resolving.

benhalpern:
I'm cool with this being one giant PR, TBH.

@ludwiczakpawel you'll need to fix the conflicts, but we can try to get this shipped ASAP and deal with any small fallouts from it being such a big PR as we see them. I think ultimately we'll be in a definitively better place once we have this out.

app/views/admin/configs/show.html.erb
```diff
@@ -991,8 +990,8 @@
         <table class="table">
           <thead>
             <tr>
-              <th>Variable</th>
-              <th>Value</th>
+              <th scope="col">Variable</th>
```
nickytonline, line 993: TIL about the [scope](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th#attr-scope) attribute.

app/views/admin/listings/edit.html.erb
```diff
@@ -1,36 +1,42 @@
...
+      <%= form.label :title %>
+      <%= form.text_field :title, size: 100, maxlength: 128, class: "form-control" %>
+    </div>
+    <div class="form-group">
+      <%= form.label :body_markdown %>
+      <%= form.text_area :body_markdown, cols: 40, rows: 15, maxlength: 400, class: "form-control" %>
+    </div>
+    <div class="form-group">
+      <%= form.label :tag_list %>
+      <i style="font-size: 15px;">Comma separated, one space, 8 tags max</i>
```
nickytonline, line 29: Can we use a CSS class instead of an `inline` style for the `font-size`?

app/views/admin/organizations/show.html.erb
```diff
@@ -1,58 +1,61 @@
...
+  <div class="ml-auto">
+    <%= link_to "View @#{@organization.name}", "/#{@organization.slug}", class: "btn btn-primary", target: "_blank", rel: "noopener" %>
+  </div>
+</header>
+
+<div class="crayons-card p-6 mb-6">
+  <h3 class="mb-4">General Info</h3>
+  <dl>
+    <dt>ID:</dt>
+    <dd><%= @organization.id %></span></dd>
```
nickytonline, line 16: Looks like there is an orphaned closing `<span />` tag here.

app/views/admin/configs/show.html.erb
ludwiczakpawel: (me too :D)

app/views/admin/listings/edit.html.erb
ludwiczakpawel: consider it done.

ludwiczakpawel:
All reported issues should be fixed now.

db/schema.rb
Ridhwana: @ludwiczakpawel just checking in on this - it doesn't look like the schema has been resolved as yet, as there are still changes to the schema showing up on your PR?

Ridhwana:
> All reported issues should be fixed now.

@ludwiczakpawel can you double check the schema? It looks like there are still changes to the schema on your branch.

ludwiczakpawel:
@Ridhwana ugh that's because I'm a moron... Should be good now. :)

Ridhwana (approved):
I pulled down the branch and tested the interface and went through the code. There may be some things that I have missed because the PR was large, but for the most part it looks like it's in a good state. @ludwiczakpawel there's a couple of tests that are failing due to text removals/edits in the UI. Approving pending passing tests.

app/views/admin/organizations/show.html.erb
nickytonline, line 16: There is this orphaned `</span>` still.

```suggestion
    <dd><%= @organization.id %</dd>
```

ludwiczakpawel: uhhh, good catch, thanks! fixed!

app/views/admin/badge_achievements/award.html.erb
```diff
@@ -1,21 +1,21 @@
...
-  <%= f.text_area :usernames, placeholder: "username1, username2, username3", required: true, size: "40x10", class: "form-control" %>
-</div>
-<div class="form-group">
-  <%= f.label :message_markdown, "Override Default Message (Supports Markdown)" %>
-  <%= f.text_area :message_markdown, size: "40x10", class: "form-control" %>
-</div>
-<%= f.submit "Award Badges", class: "btn btn-primary mb-4" %>
-<%= link_to "Back to Badge Achievements", admin_badge_achievements_path, class: "btn btn-primary float-right" %>
-<% end %>
+<h2 class="fs-2xl s:fs-3xl mb-6">Award Badges</h2>
```
rhymes, line 1: IIRC `<main>` was added by @jacobherrington to fix some a11y issues, is it okay if it disappears? Same for all the other pages

rhymes (commented):
Nice!

Is there a scenario where this crayonsification keeps the full width? One thing that I like about the current admin is that it uses all the horizontal real estate instead of boxing in the center column.

app/views/admin/badge_achievements/award.html.erb
nickytonline: @rhymes, there is a `<main />` tag in the admin pages. It was moved elsewhere is all. https://github.com/forem/forem/pull/10015#discussion_r479418367

rhymes: got it, thanks!

PR #10018, opened 2020-08-26
Add missing foreign keys to models related to users

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Adds all the missing FKs to models related to users

Part 2 of https://github.com/forem/forem/pull/10011

### DIFF
--- a/spec/factories/rating_votes.rb
+++ b/spec/factories/rating_votes.rb
@@ -4,7 +4,7 @@ FactoryBot.define do
     group { "experience_level" }
     rating { rand(1.0..8.0) }
     after(:build) do |vote|
-      vote.article ||= create(:article, user: vote.user)
+      vote.article ||= create(:article)
     end
   end
 end
--- a/spec/services/users/delete_spec.rb
+++ b/spec/services/users/delete_spec.rb
@@ -115,10 +115,6 @@ RSpec.describe Users::Delete, type: :service do
       create_associations(direct_associations.select { |a| kept_association_names.include?(a.name) })
     end
 
-    before do
-      allow(user).to receive(:trusted).and_return(true)
-    end
-
     def create_associations(names)
       associations = []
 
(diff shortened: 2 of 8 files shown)


rhymes:
@vaidehijoshi @mstruve I made a mistake. Rating votes can't be currently nullified as the `user_id` in that column is not nullable (fortunately a test alerted me of that :D): 

https://github.com/forem/forem/blob/83c2e1427c0a9933e41a8599e62ab14dcaf80ef3/db/schema.rb#L979-L990

Also, as you can see, rating vote doesn't have an optional relation with `User`

https://github.com/forem/forem/blob/83c2e1427c0a9933e41a8599e62ab14dcaf80ef3/app/models/rating_vote.rb#L1-L3

I'm going to set this to draft and send a PR to adjust "nullability" in the `rating.user_id` column and adapt the model fields and validators as a consequence of that.

spec/factories/rating_votes.rb
```diff
@@ -4,7 +4,7 @@
     group { "experience_level" }
     rating { rand(1.0..8.0) }
     after(:build) do |vote|
-      vote.article ||= create(:article, user: vote.user)
+      vote.article ||= create(:article)
```
rhymes, line 7: for `Users::Delete` magical specs to work, the rating vote creator can't be the same as the article creator. This is because deleting a user will delete their articles and thus the rating votes attached to such article. So we shouldn't make a user vote on themselves in tests when it's already not permitted in the validation ;)

spec/services/users/delete_spec.rb
```diff
@@ -3,7 +3,7 @@
 RSpec.describe Users::Delete, type: :service do
   before { omniauth_mock_github_payload }
 
-  let(:user) { create(:user, :with_identity, identities: ["github"]) }
+  let(:user) { create(:user, :trusted, :with_identity, identities: ["github"]) }
```
rhymes, line 6: These tests contain a lot of magical reflection built by @lightalloy, instead of creating a "if association is rating votes do something slightly different" I just decided to upgrade power up the user so that the rating vote validator will let them create a row in the DB

rhymes (commented):
Updated the validators and fixed a specs

Ridhwana (approved):
🎉

citizen428 (approved):

PR #10021, opened 2020-08-26
Create routes for i18n

atsmith813:
- [x] Feature

## Description
This PR is starting to lay the groundwork to support i18n and viewing Forem in various languages. This introduces the routing structure we want to use in the app by nesting all routes under an optional scope. This will allow routes such as `/lang-fr-ca/username`. As we build on i18n, this is how users will access Forem in various languages (such as French Canadian).

Setting this up as a "sub-folder" enables us to better approach SEO, caching, and server-side code in the future. We are doing it as `/lang-fr-ca/` as opposed to the more widely recognized `/fr/ca` approach to give ourselves more flexibility on how we want to handle future changes.

The loose gameplan here is to lay the groundwork (routes, caching, etc) so that the app has the ability to understand the language preference to build on top of later.

**TODO**
- [ ] Add validations to Organizations and pages to prevent the use of `lang-`.
- [ ] Add note about moving routes outside of locale
- [ ] Fix specs
- [ ] Add a few specs around locale
- [ ] Add documentation

1. Fire up the app locally.
3. Add `/lang-fr-ca/` to the front of some routes and see that it still works (i.e. http://localhost:3000/lang-fr-ca/dashboard).
4. Try to update your username to `lang-anything` and verify it shows an error.
5. Visit a user's profile to ensure the route still works.

That's it 🎉 

## Added tests?
- [ ] yes
- [ ] no, because I need help

## Added to documentation?
- [x] no documentation needed - I'm going to do a "state of i18n" writeup after this PR to explain the current state, where we're trying to go, and how the community can help.

spinning_globe

### DIFF
--- a/app/views/users/_org_admin.html.erb
+++ b/app/views/users/_org_admin.html.erb
@@ -79,7 +79,7 @@
 <div class="crayons-card grid gap-6 p-6 mb-6">
   <h2>Organization details</h2>
 
-  <%= form_for @organization, html: { class: "grid gap-6" } do |f| %>
+  <%= form_for @organization, url: organization_path(params[:locale], @organization), html: { class: "grid gap-6" } do |f| %>
     <%= f.hidden_field :id %>
     <div class="crayons-field">
       <%= f.label :name, class: "crayons-field__label" %>
@@ -219,7 +219,7 @@
       </p>
       <% if @org_organization_memberships.size === 1 && @organization.articles_count.zero? %>
         <% onsubmit = "return confirm('Are you absolutely sure you want to delete your organization: #{@organization.name}?');" %>
-        <%= form_tag organization_path(@organization.id), method: :delete, onsubmit: onsubmit do %>
+        <%= form_tag organization_path(params[:locale], @organization.id), method: :delete, onsubmit: onsubmit do %>
           <button class="crayons-btn crayons-btn--danger crayons-btn--icon-left" type="submit">
             Delete "<%= @organization.name %>"
           </button>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,8 +1,12 @@
 # rubocop:disable Metrics/BlockLength
 
 Rails.application.routes.draw do
+  # Devise does not support scoping omniauth callbacks under a dynamic segment
+  # so this lives outside our i18n scope.
   devise_for :users, only: :omniauth_callbacks, controllers: { omniauth_callbacks: "omniauth_callbacks" }
 
+  # [@forem/delightful] - all routes are nested under this optional scope to
+  # begin supporting i18n.
   scope "(:locale)", locale: /lang-.*/ do
     use_doorkeeper do
       controllers tokens: "oauth/tokens"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,508 +1,511 @@
 # rubocop:disable Metrics/BlockLength
 
 Rails.application.routes.draw do
-  use_doorkeeper do
-    controllers tokens: "oauth/tokens"
-  end
+  devise_for :users, only: :omniauth_callbacks, controllers: { omniauth_callbacks: "omniauth_callbacks" }
 
-  devise_for :users, controllers: {
-    omniauth_callbacks: "omniauth_callbacks",
-    registrations: "registrations",
-    invitations: "invitations",
-    sessions: "sessions"
-  }
-
-  devise_scope :user do
-    get "/enter", to: "registrations#new", as: :sign_up
-    get "/confirm-email", to: "devise/confirmations#new"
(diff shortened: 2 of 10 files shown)


config/routes.rb
```diff
@@ -1,506 +1,514 @@
...
-  use_doorkeeper do
-    controllers tokens: "oauth/tokens"
-  end
+  # Devise does not support scoping omniauth callbacks under a dynamic segment
+  # so this lives outside our i18n scope.
+  devise_for :users, only: :omniauth_callbacks, controllers: { omniauth_callbacks: "omniauth_callbacks" }
+
+  # [@forem/delightful] - all routes are nested under this optional scope to
+  # begin supporting i18n.
+  scope "(:locale)", locale: /lang-.*/ do
```
atsmith813, line 10: While the diff looks rather large, these lines are the only real change to this file. All of the existing routes were just nested under this scope.

config/routes.rb
```diff
@@ -1,506 +1,514 @@
...
+      controllers tokens: "oauth/tokens"
+    end
 
-  devise_for :users, controllers: {
-    omniauth_callbacks: "omniauth_callbacks",
-    registrations: "registrations",
-    invitations: "invitations"
-  }
+    devise_for :users, skip: :omniauth_callback, controllers: {
+      registrations: "registrations",
```
atsmith813, line 16: I moved the `omniauth_callbacks` higher up outside of the `locale` scope.

config/routes.rb
```diff
@@ -1,506 +1,515 @@
...
-  use_doorkeeper do
-    controllers tokens: "oauth/tokens"
-  end
+  # Devise does not support scoping omniauth callbacks under a dynamic segment
+  # so this lives outside our i18n scope.
+  devise_for :users, only: :omniauth_callbacks, controllers: { omniauth_callbacks: "omniauth_callbacks" }
+
+  # [@forem/delightful] - all routes are nested under this optional scope to
+  # begin supporting i18n.
+  scope "(:locale)", locale: LOCALE_LANG_REGEXP do
```
atsmith813, line 11: While the diff looks rather large, these lines are the only real change to this file. All of the existing routes were just nested under this scope.

config/routes.rb
```diff
@@ -1,506 +1,515 @@
...
+    get "/:username/:slug/manage" => "articles#manage"
+    get "/:username/:slug/edit" => "articles#edit"
+    get "/:username/:slug/delete_confirm" => "articles#delete_confirm"
+    get "/:username/:slug/stats" => "articles#stats"
+    get "/:username/:view" => "stories#index",
+        :constraints => { view: /comments|moderate|admin/ }
+    get "/:username/:slug" => "stories#show"
+    get "/:sitemap" => "sitemaps#show",
+        :constraints => { format: /xml/, sitemap: /sitemap-.+/ }
+    get "/:locale" => "stories#index", :constraints => { locale: LOCALE_LANG_REGEXP }
```
atsmith813, line 508: Added this line to catch future routes such as `https://forem.dev/lang-fr-ca`.

benhalpern: We actually _do_ allow `-` for orgs and pages which share the top level domain... We... So maybe we should add a validation so they can't start with `lang-`?

app/views/users/_org_admin.html.erb
```diff
@@ -79,7 +79,7 @@
 <div class="crayons-card grid gap-6 p-6 mb-6">
   <h2>Organization details</h2>
 
-  <%= form_for @organization, html: { class: "grid gap-6" } do |f| %>
+  <%= form_for @organization, url: organization_path(params[:locale], @organization), html: { class: "grid gap-6" } do |f| %>
```
benhalpern, line 82: Will we need to add this to all helpers across the site? Also will it matter in non `GET` methods?

config/routes.rb
atsmith813: Ohhh barnacles. We'll have to add validations from here forward to anything else that does is in the future.  I wonder if there's a good way to remind us this validation is needed in those cases 🤔

app/views/users/_org_admin.html.erb
atsmith813: I'm not sure yet 🤔 . I'm still trying to find a way to not have to do this once I get the other specs fixed.

I don't think it should matter in non `GET` methods. We may need it with non `GET` too because otherwise, the server may interpret that the response should be in the default locale without one being provided.

config/routes.rb
benhalpern: Best is probably a page in documentation and/or comments explaining the general concerns and issues with the top level domain in general.

I think the best way to not mess this up is to have an acute awareness of the problems presented by using the same namespace for so many things.

PR #10029, opened 2020-08-26
Optimization:Add Option to Throttle SQL tracking in Sidekiq Workers

mstruve:
- [x] Optimization

## Description
In #9998 I mentioned one reason I wanted to rollback the change was that the worker was flooding Honeycomb with events due to looping through ALL of our users. Since there is no option to disable Honeycomb for single actions I went with the approach of throttling the events considerably through our `Honeycomb::NoiseCancellingSampler` at a rate to 100_000 which means 1 out of every 100_000 SQL events will get sent to Honeycomb. 

While here I went ahead and moved everything from `/lib` to `app/lib`

https://github.com/orgs/forem/projects/6#card-44303910

My plan is to use this for the RssReaderWorker when I create it. In the meantime, if you want to test this you can 
1) add `throttle_honeycomb_tracking: true` to the `sidekiq_options` for any worker.
2) add a `puts` statement to the NoiseCancellingSampler right above `rate = 100_000`
3) restart your server and run your job. You should see your puts statement in the logs. 

## Added tests?
- [x] yes

alt_text


mstruve:
Going to pump the breaks on this one. I think if we can break down our Sidekiq workers enough we can eliminate a lot of these giant traces.

PR #10035, opened 2020-08-27
Remove transferring user fields to profile related code

Zhao-Andy:
- [x] Refactor
- [x] Bug Fix

## Description
This PR does two things:
- remove code that is no longer necessary, as the code was intended to verify that the profile migration was successful (which it was!)
- fix a bug where users with certain `nil` fields would bypass our validations, and therefore a user could submit an invalid field. Invalid fields on a user prevent them from signing in.

1. go to `/settings/profile`
2. try to update a field with an invalid value, like `website_url = 'test'`
3. see that it does not submit
4. try to update it with a valid value, like `website_url = 'https://dev.to'`
5. see that it submits and saves successfully

## Added tests?
- [x] no, because they aren't needed. We'll be removing validations eventually.

## Added to documentation?
- [x] no documentation needed

None, for now...

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -10,7 +10,7 @@ class User < ApplicationRecord
   # rubocop:disable Metrics/BlockLength
   concerning :ProfileMigration do
     included do
-      PROFIE_FIELDS =
+      PROFILE_FIELDS =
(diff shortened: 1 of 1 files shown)


Zhao-Andy:
Resolved by https://github.com/forem/forem/pull/9912

PR #10036, opened 2020-08-27
Delete all cookies when signing out

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This is designed to help an issue where we have some cookie inconsistencies which need to be remedied on signout, but it's also just pretty decent hygiene in terms of ensuring we're not enabling any extra tracking in an unintended way. This cleans the slate when someone needs to leave.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,0 +1,9 @@
+class SessionsController < Devise::SessionsController
+  def destroy
+    # Let's say goodbye to all the cookies when someone signs out.
+    cookies.each do |cookie|
+      cookies.delete cookie[0]
+    end
+    super
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -8,7 +8,8 @@ Rails.application.routes.draw do
   devise_for :users, controllers: {
     omniauth_callbacks: "omniauth_callbacks",
     registrations: "registrations",
-    invitations: "invitations"
+    invitations: "invitations",
+    sessions: "sessions"
   }
 
   devise_scope :user do


app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,9 @@
+class SessionsController < Devise::SessionsController
+  def destroy
+    # Let's say goodbye to all the cookies when someone signs out.
+    cookies.each do |cookie|
```
rhymes, line 4: should we add a test for this? you can access `response.cookies` in a request spec

benhalpern:
@rhymes I tried and couldn't figure out how to get the test right, but manually tested a bunch and feel pretty good about this.

But if you want to take a hack at adding the test for regression that would be very great.

rhymes:
@benhalpern 

I tried as well, but it doesn't really work because we actually set cookies right after again, in the redirect.

You can see it production. Open the console in the storage cookies panel, make a mental note of the values, go to sign out while keep it open. You'll notice the cookies don't disappear, they just become invalid. Which means that right after you clear them, something is setting them again as soon as you're redirected to the homepage.

It happens in the tests as well: 

```ruby
require "rails_helper"

RSpec.describe "UserSignOut", type: :request do
  let(:user) { create(:user) }

  before do
    sign_in user
  end

  it "removes all cookies when signing out" do
    get root_path
    expect(cookies.to_hash.size.positive?).to be(true)

    delete destroy_user_session_path
    # here you have a new session cookie, the only difference is that remember_user_token is empty
    expect(cookies.to_hash.size.positive?).to be(false) # this fails
  end
end
```

PR #10038, opened 2020-08-27
[Admin Tooling] Allow admins unpublish articles from the Mod Actions Panel

msarit:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
For admins to unpublish articles, the current workflow is:
- navigate to `<article_path>/edit`
- change 'published' status to 'false' (v1 editor) or click 'Unpublish Post' button (v2 editor)

This PR adds an 'Unpublish Article' button to the Mod ActionsPanel, for a one-click flow.

### Note:
I keep getting Pusher-related errors repeatedly (upwards of 40 or 50 times) in the browser console. Here are 2 examples:

> Access to XMLHttpRequest at 'https://sockjs-us2.pusher.com/pusher/app//937/pgrbam4g/xhr_streaming?protocol=7&client=js&version=7.0.0&t=1598903280151&n=15' from origin 'http://localhost:3000' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.

> Refused to connect to 'wss://ws-us2.pusher.com/app/?protocol=7&client=js&version=7.0.0&flash=false' because it violates the following Content Security Policy directive: "connect-src 'self' https: http://localhost:3035 ws://localhost:3035".

These errors do not seem to impede functionality per se, but I'd like to understand why they are being generated.

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

Black man speaking


app/controllers/articles_controller.rb
```diff
@@ -185,6 +185,19 @@ def stats
...
+    authorize @article
+    if @article.has_frontmatter?
+      @article.body_markdown.sub!("\npublished: true\n", "\npublished: false\n")
+    else
+      @article.published = false
+    end
+
+    return unless @article.save
+
+    redirect_to "#{URL.url@article.current_state_path}"
```
juliannatetreault, line 198: I'm not sure if this solution is exactly what you're looking for, but this redirect opens the article editor in the Mod Actions Panel giving the Admin the ability to edit the article from there once it is unpublished. I found this redirect in the #update method in this controller (`Articles::Controller`) for reference! 😄 
```suggestion
    redirect_to("#{@article.path}/edit")
```

msarit: Thanks @juliannatetreault! I'm not sure if the article's `edit` path is where we want to end up though.
The bigger issue is getting the redirect to happen within the actual article page, not within the iframe 😓 
Thank you though!

citizen428: @msarit Iframes can't update the location of the embedding site by default for security reasons. You need to add an option to your `<iframe>` tag:

```
sandbox="allow-top-navigation"
```

app/controllers/articles_controller.rb
```diff
@@ -185,6 +185,21 @@ def stats
     @organization_id = @article.organization_id
   end
 
+  def admin_unpublish
+    authorize @article
+    if @article.has_frontmatter?
+      @article.body_markdown.sub!("\npublished: true\n", "\npublished: false\n")
```
rhymes, line 191: This regexp will fail if `published:true`, or `published: true ` or other combinations that are valid Markdown frontamtters.

I think we should replace it with the following:


```suggestion
      @article.body_markdown.sub!(%r{\npublished:\s*true\s*\n}, "\npublished: false\n")
```

this takes into account optional spaces between `published:` and `true` and after `true` and before the `newline`

app/javascript/actionsPanel/initializeActionsPanelToggle.js
```diff
@@ -1,7 +1,7 @@
 /** This initializes the mod actions button on the article show page (app/views/articles/show.html.erb). */
 export default function initializeActionsPanel(user, path) {
   const modActionsMenuHTML = `
-    <iframe id="mod-container" src=${path}/actions_panel title="Moderation panel actions">
+    <iframe id="mod-container" src=${path}/actions_panel sandbox="allow-top-navigation allow-scripts allow-same-origin allow-modals" title="Moderation panel actions">
```
rhymes, line 4: I'm not sure I see the connection between the Pusher error and this sandboxed iframe. BTW, MDN strongly discourages using both `allow-script` and `allow-same-origin` as the script can remove the sandboxing:

> When the embedded document has the same origin as the embedding page, it is strongly discouraged to use both allow-scripts and allow-same-origin, as that lets the embedded document remove the sandbox attribute — making it no more secure than not using the sandbox attribute at all.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

If the goal is the to redirect the parent page, I was able to make this work without `sandbox`...

app/javascript/actionsPanel/actionsPanel.js
```diff
@@ -143,6 +143,31 @@ async function updateExperienceLevel(currentUserId, articleId, rating, group) {
...
+      method: 'PATCH',
+      body: JSON.stringify({ id, username, slug }),
+      credentials: 'same-origin',
+    });
+
+    const outcome = await response.json();
+
+    /* eslint-disable no-restricted-globals */
+    if (outcome.message == "success") {
+      window.top.location.href = window.location.origin + outcome.path;
```
rhymes, line 158:
```suggestion
      window.top.location.assign(`${window.location.origin}${outcome.path}`);
```

this for me works without sandboxing, both in Firefox and in Chrome

rhymes (commented):
Had a look, was able to make it work without sandboxing using `window.top.location.assign`. 

Let me know!

rhymes (approved):
Works great @msarit !

PR #10039, opened 2020-08-27
Create initial "register by email and password" beta/MVP in prep for "setup mode"

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This pull request is primarily to **ready us for a scenario where we can allow _the first user_ to sign up for a new Forem** to then become a super admin to seed the rest of the info needed to get a Forem going.

Since you need to sign in as an admin to set up _any_ kind of config, we need to enable this to create that first user. I decided that as a component of the yak shave to help enable that _first user scenario_, it would make sense to set ourselves up to enable sign in with email as a beta option in general.

So this is the first pass, with minimal functionality needed to make this all properly happen.

It would allow a Forem to toggle this functionality on if they so choose, but does not provide the views linked to anywhere, and the initial follow up here would be to mostly just enable this functionality when a Forem is in "setup mode".

What this PR does not include: Beautiful views, fully fleshed out UX or anything like that.


app/controllers/registrations_controller.rb
```diff
@@ -8,4 +8,19 @@ def new
...
     end
   end
+
+  def create
+    not_authorized unless SiteConfig.allow_email_password_registration
+
+    build_resource(sign_up_params)
+    resource.saw_onboarding = false
+    resource.editor_version = "v2"
+    resource.save if resource.email.present?
```
fdoxyz, line 18: Would it make sense to add a check for the number of registered users and if it happens to be the first user in the instance maybe make the user admin?

This may be too much of a hack so take it a grain of salt (to say the least). It's just an idea to streamline the process in this beta period for creators.

benhalpern: > Would it make sense to add a check for the number of registered users and if it happens to be the first user in the instance maybe make the user admin?

This is approximately how I'm thinking about it, or just have forems live in a `SiteConfig.setup_mode = true` or something like that.

This pass doesn't add that specific "should I make this person a super admin logic", I'll add that after we get this in.

fdoxyz: Cool! Following up on that concept of setup mode: We already have the `setup_completed?` helper, which "means" the instance hasn't been configured yet (requires an admin) which could save us adding another SiteConfig.

Anyways, just sharing ideas here because of course doing that over a SiteConfig has its own set of implications.

PR #10042, opened 2020-08-27
Optimization:Dont Send Multiple Digest Emails in a Day

mstruve:
- [x] Optimization

## Description
I am working on breaking up the SendEmailDigestWorker into individual workers for each user Digest bc the looping through all users is taking hours. Prior to that being ready, this is a small optimization that will save us some time if that long-running job restarts in the middle of its run. **This assumes we would never want to send more than 1 digest email in a day which seems pretty reasonable to me.** 

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/labor/email_logic.rb
+++ b/app/labor/email_logic.rb
@@ -28,31 +28,36 @@ class EmailLogic
   def get_articles_to_send
     fresh_date = get_fresh_date
 
-    articles = if user_has_followings?
-                 experience_level_rating = (@user.experience_level || 5)
-                 experience_level_rating_min = experience_level_rating - 3.6
-                 experience_level_rating_max = experience_level_rating + 3.6
+    if fresh_date < 23.hours.ago
+      articles = if user_has_followings?
+                   experience_level_rating = (@user.experience_level || 5)
+                   experience_level_rating_min = experience_level_rating - 3.6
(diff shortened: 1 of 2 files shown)


mstruve:
Well that happened faster than I thought but I still like this optimization https://github.com/forem/forem/pull/10043

app/labor/email_logic.rb
```diff
@@ -28,31 +28,36 @@ def should_receive_email?
...
-                   .where("published_at > ?", fresh_date)
-                   .where(featured: true, email_digest_eligible: true)
-                   .where.not(user_id: @user.id)
-                   .where("score > ?", 25)
-                   .order(score: :desc)
-                   .limit(8)
-               end
-
-    @ready_to_receive_email = false if articles.length < 3
+    if fresh_date < 23.hours.ago
```
citizen428, line 31: Random comment: Rails adds a couple of nice helpers (in the module `DateAndTime::Calculations`), so this can also be written as 

```ruby
fresh_date.before?(23.hours.ago)
```

which some people may find a bit more readable.

citizen428 (approved):

app/labor/email_logic.rb
mstruve: 🔥 TIL! Thanks!

PR #10043, opened 2020-08-27
Optimization:Break Send User Digest Emails in Single Jobs

mstruve:
- [x] Optimization

## Description
The SendEmailDigest job is a beast right now bc it loops through nearly 300k users to send email digests. This breaks it down so that the article collection and email sending is handled in individual jobs, one for each user. All the main job, now called `Email::EnqueueDigestWorker` has to do is enqueue those short fast-running jobs. 

The reason we want lots of fast running jobs rather than one long running one is because Sidekiq is built for short running jobs. Also, when we deploy, we shutdown sidekiq and all the jobs that are processing get restarted. This means if you have a long running job and you deploy a lot, it will never get a chance to finish. Its better to have lots of short running jobs. Ideally ones that can finish in under 30 seconds since that is how much grace time Sidekiq gives before forcefully killing processes during a deploy. 

https://github.com/orgs/forem/projects/6#card-44446358

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/labor/email_digest.rb
+++ b/app/labor/email_digest.rb
@@ -8,19 +8,8 @@ class EmailDigest
   end
 
   def send_periodic_digest_email
-    @users.find_each do |user|
-      user_email_heuristic = EmailLogic.new(user).analyze
-      next unless user_email_heuristic.should_receive_email?
-
-      articles = user_email_heuristic.articles_to_send
-      begin
-        next unless user.email_digest_periodic?
-
-        DigestMailer.with(user: user, articles: articles).digest_email.deliver_now
-      rescue StandardError => e
-        Honeybadger.context({ user_id: user.id, article_ids: articles.map(&:id) })
-        Honeybadger.notify(e)
-      end
+    @users.ids.each do |user_id|
+      Email::SendUserDigestWorker.perform_async(user_id)
     end
   end
 
--- a/app/workers/email/enqueue_digest_worker.rb
+++ b/app/workers/email/enqueue_digest_worker.rb
@@ -1,0 +1,11 @@
+module Email
+  class EnqueueDigestWorker
+    include Sidekiq::Worker
+
+    sidekiq_options queue: :medium_priority, retry: 15
+
+    def perform
+      EmailDigest.send_periodic_digest_email
+    end
+  end
+end
--- a/app/workers/email/send_user_digest_worker.rb
+++ b/app/workers/email/send_user_digest_worker.rb
@@ -1,0 +1,23 @@
+module Email
+  class SendUserDigestWorker
+    include Sidekiq::Worker
+
+    sidekiq_options queue: :low_priority, retry: 15, lock: :until_executing
+
+    def perform(user_id)
+      user = User.find_by(id: user_id)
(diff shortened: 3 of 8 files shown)


app/workers/email/send_user_digest_worker.rb
```diff
@@ -0,0 +1,23 @@
...
+      user = User.find_by(id: user_id)
+      return unless user&.email_digest_periodic? && user&.registered?
+
+      user_email_heuristic = EmailLogic.new(user).analyze
+      return unless user_email_heuristic.should_receive_email?
+
+      articles = user_email_heuristic.articles_to_send
+      begin
+        DigestMailer.with(user: user, articles: articles).digest_email.deliver_now
+      rescue StandardError => e
```
mstruve, line 17: I kept this logic from the rake task. We could probably remove it but for now I just ported it over.

app/workers/email/send_user_digest_worker.rb
```diff
@@ -0,0 +1,23 @@
+module Email
+  class SendUserDigestWorker
+    include Sidekiq::Worker
+
+    sidekiq_options queue: :low_priority, retry: 15, lock: :until_executing
```
mstruve, line 5: There will be 300k of these enqueued at once, I wanted them on the low priority queue to ensure they didn't block anything else that was more time sensitive

citizen428 (approved):

app/workers/email/enqueue_digest_worker.rb
```diff
@@ -0,0 +1,11 @@
+module Email
```
rhymes, line 1: as most of the workers follow the pattern of `"#{plural resource}::worker name" ` maybe we can rename this as `Emails::`

app/labor/email_digest.rb
```diff
@@ -8,19 +8,8 @@ def initialize(users = [])
...
-      articles = user_email_heuristic.articles_to_send
-      begin
-        next unless user.email_digest_periodic?
-
-        DigestMailer.with(user: user, articles: articles).digest_email.deliver_now
-      rescue StandardError => e
-        Honeybadger.context({ user_id: user.id, article_ids: articles.map(&:id) })
-        Honeybadger.notify(e)
-      end
+    @users.ids.each do |user_id|
```
rhymes, line 11: Question: do you think there's a benefit in loading ids in batches? Something like `User.select(:id).in_batches` ?

It's maybe a premature optimization though

app/workers/email/send_user_digest_worker.rb
rhymes: make sense. We should eventually move all this logic rich workers to services. Haven't reached a solid opinion on that :D

rhymes (approved):

mstruve:
> creates a growing backlog that our server can't work through in time before the next time the job get enqueued.

@maestromac the jobs will definitely be enqueued faster than they can be processed but that is why I put them on the low priority queue, they can hang out until a worker is ready for them and we dont have to worry about them blocking more time sensitive jobs

mstruve:
@rhymes will followup with a cleanup PR! we have 2 of these workers running now that I want to get killed so going to deploy as is

PR #10051, opened 2020-08-28
Fix userData as undefined error

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This is the error in the console

Screenshot 2020-08-28 at 3 02 03 PM

As you can see the `user` is returned as undefined. I think it's because one of the cookies sent by the browser is already expired. I thought it was due to #10036 but this already happened me twice since that's been in production and both times I cleaned out cookies and storage.

### DIFF
--- a/app/assets/javascripts/utilities/userData.js
+++ b/app/assets/javascripts/utilities/userData.js
@@ -3,5 +3,11 @@
 function userData() {
   const { user = null } = document.body.dataset;
 
+  // [@forem/oss]: there's an edge case in which `user` has the value of "undefined",
+  // this results in is a JSON syntax error (`user` should be a hash, not a string).
+  if (user === 'undefined') {
+    return {};
+  }
+
   return JSON.parse(user);
 }


app/assets/javascripts/utilities/userData.js
```diff
@@ -3,5 +3,11 @@
 function userData() {
   const { user = null } = document.body.dataset;
 
+  // [@forem/oss]: there's an edge case in which `user` has the value of "undefined",
+  // this results in is a JSON syntax error (`user` should be a hash, not a string).
+  if (user === 'undefined') {
```
nickytonline, line 8: Odd that it is a string `'undefined'`. The default value of `null` handles the case when user is `undefined` so that `JSON.parse(user)` works all the time.

It'd be good to know why it's a  `'undefined'` instead of actually `undefined`

rhymes:
Going to close this as it's clearly not the issue here

PR #10052, opened 2020-08-28
Delete cookies correctly by specifying the domain

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Cookie deletion should specify the domain as well:

> Please note that if you specify a :domain when setting a cookie, you must also specify the domain when deleting the cookie:

```ruby
cookies[:name] = {
  value: 'a yummy cookie',
  expires: 1.year,
  domain: 'domain.com'
}

cookies.delete(:name, domain: 'domain.com')
```

from https://api.rubyonrails.org/v6.0.3.2/classes/ActionDispatch/Cookies.html

#10036 

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,9 +1,9 @@
 class SessionsController < Devise::SessionsController
   def destroy
     # Let's say goodbye to all the cookies when someone signs out.
-    cookies.each do |cookie|
-      cookies.delete cookie[0]
-    end
+    domain = Rails.env.production? ? ApplicationConfig["APP_DOMAIN"] : nil
+    cookies.delete(domain: domain)
+
     super
   end
 end


app/controllers/sessions_controller.rb
```diff
@@ -1,9 +1,9 @@
 class SessionsController < Devise::SessionsController
   def destroy
     # Let's say goodbye to all the cookies when someone signs out.
-    cookies.each do |cookie|
-      cookies.delete cookie[0]
-    end
+    domain = Rails.env.production? ? ApplicationConfig["APP_DOMAIN"] : nil
```
fdoxyz, line 4: Interesting, I think we would need this on development too. Not sure if `localhost:3000` behaves differently but I can reproduce the problem in development with a custom domain (ngrok or xip.io for mobile development)

rhymes: We don't see the domain in development: 

https://github.com/forem/forem/blob/6c2357af48e801b4d403accbbb685958d24a8f79/config/initializers/session_store.rb#L12-L13

fdoxyz: Oh cool! Defenitely didn't see that

PR #10054, opened 2020-08-28
Refactor:Cleanup Naming and Optimize Email Digest Code

mstruve:
- [x] Refactor
- [x] Optimization

## Description
- Move Logic intensive Email Digest code into Services, we are deprecating labor
- Optimization: Grab user_ids in batches
- Use plural Emails instead of Email to match with convention

alt_text


app/workers/emails/enqueue_digest_worker.rb
```diff
@@ -1,4 +1,4 @@
-module Email
+module Emails
```
rhymes, line 1: IIRC when the name of a worker is changed we should add an alias for a while, right?

app/services/email_digest.rb
```diff
@@ -8,8 +8,10 @@ def initialize(users = [])
   end
 
   def send_periodic_digest_email
-    @users.ids.each do |user_id|
-      Email::SendUserDigestWorker.perform_async(user_id)
+    @users.select(:id).in_batches do |batch|
```
rhymes, line 11: :-)

rhymes (commented):
neat

app/workers/emails/enqueue_digest_worker.rb
mstruve: Nah, this will be fine and I actually want this right now bc I need to kill the job running.

mstruve:
Admin merged to kill the current job running which is overwhelming our database

PR #10055, opened 2020-08-28
Migrate ProfileImage to Images::Profile

maestromac:

- [x] Refactor

## Description
Prefer Module over Class and update the name from "profile image" to "avatar".

n/a
Spec should've caught all the issues.

1. Boot up the server locally.
2. You should be able to browse all pages normally.

## Added tests?

- [x] yes, updated existing tests. I broke so many test 😩 

## Added to documentation?
- [x] no documentation needed

n/a


maestromac:
@mstruve it's shorter and seeing the word `image` being repeated twice in `Images::ProfileImage` feels a bit off, and `Images::Profile` is straight up confusing. 

I'm not married to it though. if @benhalpern thinks there's a better word or we should revert please let me know.

mstruve:
Works for me, I was just curious, to me ProfileImage seemed very clear but I do see how the redundancy could be annoying

maestromac:
After looking around the codebase, my change here might add more confusion than I initially thought. I'd rather be consistent so I'm no longer going with Avatar.

rhymes:
@maestromac what about `Images::Profile.call()` ? The suffix in `ProfileImage` might be redundant

PR #10058, opened 2020-08-28
Add Recent Reports to Users' Admin User Page

juliannatetreault:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR adds a "Recent Reports" section to a User's Admin User page (`/admin/users/:id/edit`). The reports section displays a User's most recent `FeedbackMessages` with useful data (`user_types`, `category`, `status`, `created_at`, and `message`). By adding this data to the User's Admin User page, Admins now have the ability to more easily view which reports that a User has and any related detail for the reports, like if the User was Affected, the Offender, the Reporter, or all of the above, allowing Admins to refer back to this data if need be. Additionally, when a report is clicked on, the Admin will be brought to the `/admin/reports/:id`, where they will find the relevant report, giving them more context as to why the report was made in the first place. 

Relates to a private issue and relates to PR #10059 

**To QA this PR, first ensure that you have seeded your DB so that you have `FeedbackMessage`s (if you don't have any `FeedbackMessage`s seeded, report some content via `/admin/reports`!) and have `super_admin` permissions, then please do the following:**
- Navigate to `/admin/users/:id/edit` in your browser:
Screen Shot 2020-08-31 at 11 28 29 AM

- Click the "Toggle" button next to "Recent Reports" to see a list of all of the reports for the particular user:
Screen Shot 2020-09-03 at 11 05 20 AM

- The `user_type` should be displayed next to the report, displaying whether the user was the person affected, the offender, the reporter, or all of the above:
Screen Shot 2020-09-03 at 10 08 53 AM

- If the user does not have any reports, this message should be displayed:
Screen Shot 2020-08-31 at 11 29 10 AM

- Click on an individual report to navigate to `/admin/reports/:id`, where the relevant report can be found:
Screen Shot 2020-09-03 at 10 15 39 AM

- Try to break things! ⚒️ 

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

N/A

Kitten that wants reports

### DIFF
--- a/app/models/feedback_message.rb
+++ b/app/models/feedback_message.rb
@@ -26,4 +26,12 @@ class FeedbackMessage < ApplicationRecord
   def capitalize_status
     self.status = status.capitalize if status.present?
   end
+
+  def user_types(user_id)
+    types = []
+    types << "Affected" if user_id == affected_id
+    types << "Offender" if user_id == offender_id
+    types << "Reporter" if user_id == reporter_id
+    types
+  end
 end
--- a/app/views/admin/users/_reports.html.erb
+++ b/app/views/admin/users/_reports.html.erb
@@ -1,14 +1,14 @@
 <div class="row my-3">
   <div class="col-12">
     <h2 class="d-inline">Recent Reports</h2>
-    <button type="button" data-toggle="collapse" data-target="#reports-row" class="btn btn-secondary float-right">Toggle</button>
+    <button type="button" data-toggle="collapse" data-target="#reactions-row" class="btn btn-secondary float-right">Toggle</button>
   </div>
-  <div class="col-12 collapse" id="reports-row">
+  <div class="col-12 collapse" id="reactions-row">
     <div class="list-group-flush">
-    <% unless @reports.nil? %>
-      <% @reports.order(created_at: :desc).limit(50).each do |report| %>
-        <a href="/admin/reports/<%= report.id %>" class="list-group-item list-group-item-action">
-        <%= report.feedback_type %>
+    <% unless @related_reports.empty? %>
+      <% @related_reports.each do |report| %>
+        <a href="<%= report.path %>" class="list-group-item list-group-item-action">
+        <%= report.feedback_type %>: <%= reaction.reported_url %> (<%= report.category.capitalize %>)
         <em> - <%= report.created_at&.strftime("%b %e '%y") %></em>
         </a>
       <% end %>
--- a/app/views/admin/users/_reports.html.erb
+++ b/app/views/admin/users/_reports.html.erb
@@ -7,9 +7,11 @@
     <div class="list-group-flush">
     <% unless @related_reports.empty? %>
       <% @related_reports.each do |report| %>
-        <a href="<%= report.path %>" class="list-group-item list-group-item-action">
-        <%= report.feedback_type %>: <%= reaction.reported_url %> (<%= report.category.capitalize %>)
+        <a href="<%= report.reported_url %>" class="list-group-item list-group-item-action">
+        <%= report.feedback_type.titlecase %>: <%= report.category.capitalize %> (<%= report.status %>)
         <em> - <%= report.created_at&.strftime("%b %e '%y") %></em>
+        <br>
+        <%= report.message %>
         </a>
       <% end %>
       <% else %>
--- a/app/views/admin/users/_reports.html.erb
+++ b/app/views/admin/users/_reports.html.erb
@@ -6,7 +6,7 @@
   <div class="pt-6 collapse" id="reactions-row">
     <% unless @related_reports.empty? %>
       <% @related_reports.each do |report| %>
-        <a href="<%= report.reported_url %>" class="list-group-item list-group-item-action action px-5 d-flex justify-content-between">
+        <a href="<%= admin_feedback_messages_path %>" class="list-group-item list-group-item-action action px-5 d-flex justify-content-between">
         <span><%= report.category.capitalize %> (<%= report.status %>)</span>
         <span><strong><%= report.feedback_type.titlecase %></strong>: <%= report.message %></span>
         <em> <%= report.created_at&.strftime("%b %e '%y") %></em>
--- a/app/views/admin/users/_reports.html.erb
+++ b/app/views/admin/users/_reports.html.erb
@@ -7,11 +7,10 @@
     <div class="list-group-flush">
     <% unless @related_reports.empty? %>
       <% @related_reports.each do |report| %>
-        <a href="<%= report.reported_url %>" class="list-group-item list-group-item-action">
-        <%= report.feedback_type.titlecase %>: <%= report.category.capitalize %> (<%= report.status %>)
-        <em> - <%= report.created_at&.strftime("%b %e '%y") %></em>
-        <br>
-        <%= report.message %>
+        <a href="<%= report.reported_url %>" class="list-group-item list-group-item-action action px-5 d-flex justify-content-between">
+        <span><%= report.category.capitalize %> (<%= report.status %>)</span>
+        <span><strong><%= report.feedback_type.titlecase %></strong>: <%= report.message %></span>
+        <em> <%= report.created_at&.strftime("%b %e '%y") %></em>
         </a>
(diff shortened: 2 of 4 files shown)


app/views/admin/users/_reports.html.erb
```diff
@@ -0,0 +1,22 @@
+<div class="row my-3">
+  <div class="col-12">
+    <h2 class="d-inline">Recent Reports</h2>
+    <button type="button" data-toggle="collapse" data-target="#reactions-row" class="btn btn-secondary float-right">Toggle</button>
+  </div>
+  <div class="col-12 collapse" id="reactions-row">
+    <div class="list-group-flush">
+    <% unless @related_reports.empty? %>
```
juliannatetreault, line 8: Would it be worth it for me to use `counter_culture` here or should I stick to `.empty?` to follow the convention of the other partials like `_notes.html.erb`?

app/views/admin/users/_reports.html.erb
```diff
@@ -0,0 +1,22 @@
...
+      <% @related_reports.each do |report| %>
+        <a href="<%= report.reported_url %>" class="list-group-item list-group-item-action">
```
msarit, line 10: In my local app, `<%= report.reported_url %>` returns `nil` for some reports. For these reports, clicking on the link just reloads the current page. Is this acceptable?

juliannatetreault: I'll look into this @msarit! I might just remove this link and/or adjust it to link to something else.

juliannatetreault:
> One more minor, non-blocking change @juliannatetreault
> 
> When the Reports section is toggled open, I think adding a space between the Toggle button and the list of reports is good 👍
> 
> Screen Shot 2020-08-31 at 11 04 05 AM

I agree with you @msarit, but all of the toggle buttons are built with that padding, so to follow convention I think it makes sense to leave this as-is since the others in Admin are like this! 😄  How do you feel about leaving this as is?

app/views/admin/users/_reports.html.erb
```diff
@@ -0,0 +1,21 @@
+<div class="row my-3">
+  <div class="col-12">
+    <h2 class="d-inline">Recent Reports</h2>
+    <button type="button" data-toggle="collapse" data-target="#reactions-row" class="btn btn-secondary float-right">Toggle</button>
+  </div>
+  <div class="col-12 collapse" id="reactions-row">
+    <div class="list-group-flush">
+    <% unless @related_reports.empty? %>
+      <% @related_reports.each do |report| %>
+        <a href="<%= report.reported_url %>" class="list-group-item list-group-item-action action px-5 d-flex justify-content-between">
```
msarit, line 10: Hey! So I checked the format of `reported_url` in production, and the URLs are absolute (eg `https://dev.to/weber360/comment/cble`).

However, I notice in this PR that `<%= report.reported_url %>` is generating a relative URL.

Assume `report.reported_url` is `example.com`. The `href` above is generating `localhost:3000/admin/users/:id/example.com` instead of `example.com`, and of course, the report's link is blowing up when clicked.

Screen Shot 2020-08-31 at 11 54 45 AM

I'm still exploring why, but wanted to give this heads-up. 👍

msarit:
@juliannatetreault re: Toggle Button padding, yup let's leave as-is if they are that way everywhere 😅

app/views/admin/users/_reports.html.erb
juliannatetreault: Yeah, I'm unsure of what to use here or if I should remove the link entirely -- I was waiting for everyone's opinions when they review this, so sounds good! Thanks, @msarit 👍

juliannatetreault:
> @juliannatetreault re: Toggle Button padding, yup let's leave as-is if they are that way everywhere 😅

Design implemented some changes to the Admin UI this morning, so I'm going to see if I can apply those changes to the toggle button now @msarit! 😄

app/views/admin/users/_reports.html.erb
juliannatetreault: I've since changed this and now have each individual report linking to the `Feedback Messages` page, `/admin/feedback_messages`, which should give Admins the context they need per our discussion during our team sync today! Let me know what you think @msarit. 😄

app/views/admin/users/_reports.html.erb
```diff
@@ -0,0 +1,19 @@
+<div class="crayons-card p-6">
+  <div>
+    <h2 class="d-inline">Recent Reports</h2>
+    <button type="button" data-toggle="collapse" data-target="#reactions-row" class="btn btn-secondary float-right">Toggle</button>
+  </div>
+  <div class="pt-6 collapse" id="reactions-row">
+    <% unless @related_reports.empty? %>
+      <% @related_reports.each do |report| %>
+        <a href="<%= admin_feedback_messages_path %>" class="list-group-item list-group-item-action action px-5 d-flex justify-content-between">
```
Zhao-Andy, line 9: Let's use `admin_report_path(report.id)` instead here, to link to the individual report. We technically could do `admin_feedback_message_path(report.id)` too but `/admin/reports/:id` is a bit more popular.

app/views/admin/users/_reports.html.erb
```diff
@@ -0,0 +1,19 @@
...
+        <span><%= report.category.capitalize %> (<%= report.status %>)</span>
+        <span><strong><%= report.feedback_type.titlecase %></strong>: <%= report.message %></span>
```
Zhao-Andy, line 11: So we technically only have one `feedback_type` now, and it's become an unused column.

Instead, I think it's more useful to show if it's a reporter, offender, or affected report. I think the easiest way to show this is to do `report.reporter_id == user.id`. You might want to make a method for the instance in the model, like:

```ruby
# feedback_message.rb
def user_type(user_id)
  # returns the string "Reporter, Offender, and Affected" if all three IDs are true
end
```

app/views/admin/users/_reports.html.erb
juliannatetreault: Great suggestion!

msarit:
Looking good @juliannatetreault 💃 

Didn't uncover any more issues (other than those already mentioned)

app/models/feedback_message.rb
```diff
@@ -26,4 +26,12 @@ def abuse_report?
   def capitalize_status
     self.status = status.capitalize if status.present?
   end
+
+  def user_types(user_id)
```
Zhao-Andy, line 30: Awesome! This is the same way I thought of too :D

Zhao-Andy (approved):
This looks great!

msarit (approved):
🚀

app/views/admin/users/edit.html.erb
```diff
@@ -48,7 +48,7 @@
   </div>
 
 <%= render "notes" %>
-<%= render "negative_reactions" %>
```
fdoxyz, line 51: I can't see this partial referenced anywhere else. If it's no longer in use maybe worth getting rid of it?

fdoxyz (approved):
I took this for a spin and it's great to see all that info in there with the `user_types` context 🎉 

It all looks good to me. I'm not that experienced with this part of the codebase but left a non-blocking comment/suggestion inline.

app/views/admin/users/edit.html.erb
juliannatetreault: Great question, @fdoxyz ! This partial is used to display all of the `vomit` reactions that a particular user has on the `/admin/users/:id/edit` page just like the `reports` partial displays all of the recent `FeedbackMessage`s for a particular user on that page. This code is (unfortunately) showing up in this PR's history due to a rebase--I originally started building out the `_negative_reactions` partial in this PR and then pulled it out into a separate one which has since been merged/deployed. 😅

fdoxyz: Got it, thanks for clarifying!

juliannatetreault: Absolutely! Thanks for the review. 😄

PR #10059, opened 2020-08-28
Add Recent Vomit Reaction Reports to Users' Admin Users Page

juliannatetreault:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR adds a "Recent Reactions" section to a User's Admin User page (`/admin/users/:id/edit`). The reactions section displays a User's most recent `Reactions` with useful data (`path`, `reactable_type`, `title`, `category`, and `created_at`). By adding this data to the User's Admin User page, Admins now have the ability to more easily view which reactions that a User has and any related detail for the reactions, allowing Admins to refer back to this data if need be.

Relates to a private issue and relates to PR #10058 

**To QA this PR, first ensure that you have seeded your DB so that you have `vomit` reactions (if you don't have any reactions seeded, `vomit` on some content!) and have `super_admin` permissions, then please do the following:**
- Navigate to `/admin/users/:id/edit` in your browser:
Screen Shot 2020-08-28 at 11 25 29 AM

- Click the "Toggle" button next to "Recent Reactions" to see a list of all of the `vomit` reactions for the particular user:
Screen Shot 2020-08-28 at 2 40 23 PM

- If the user has been vomited on, the reaction will display as shown below:
Screen Shot 2020-08-28 at 2 40 37 PM

- If the user does not have any `vomit` reactions, this message should be displayed:
Screen Shot 2020-08-28 at 2 00 49 PM

- Click the `vomit` reaction to navigate to content related to the `vomit` reaction:
Screen Shot 2020-08-28 at 11 26 08 AM

- Try to break things! 🛠️ 

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

N/A

Dog in the office


app/views/admin/users/_activity.html.erb
```diff
@@ -15,7 +15,7 @@
     <p>(view full history in notes below)</p>
     <ul>
       <% @user.roles.each do |role| %>
-        <li><%= role.name %> <em><%= "- " + Tag.find(role.resource_id).name if role.resource_id.present? %></em></li>
+        <li><%= role.name %> <em><%= "- #{Tag.find(role.resource_id).name}" if role.resource_id.present? %></em></li>
```
juliannatetreault, line 18: I believe this is here because I changed the filename from `.erb` to `.html.erb` so these changes carried over. 👀

app/views/admin/users/_reactions.html.erb
```diff
@@ -0,0 +1,20 @@
+<div class="row my-3">
+  <div class="col-12">
+    <h2 class="d-inline">Recent Reactions</h2>
+    <button type="button" data-toggle="collapse" data-target="#reactions-row" class="btn btn-secondary float-right">Toggle</button>
+  </div>
+  <div class="col-12 collapse" id="reactions-row">
+    <div class="list-group-flush">
+    <% unless @related_vomit_reactions.empty? %>
```
juliannatetreault, line 8: Would it be worth it for me to use `counter_culture` here or should I stick to `.empty?` to follow the convention of the other partials like `_notes.html.erb`?

app/views/admin/users/_reactions.html.erb
```diff
@@ -0,0 +1,20 @@
...
+      <% @related_vomit_reactions.each do |reaction| %>
+        <a href="<%= reaction.reactable.path %>" class="list-group-item list-group-item-action">
```
msarit, line 10: This layout is simple and it works, but I wonder if it wouldn't be more helpful to Forem admins to have this information be more **visual**, kinda like the `Vomit Reactions` table under `/admin/reports`?

So if it is a vomit record, it displays a 🤢 emoji. I'm not sure if it's only vomits we are tracking?

Also since the record is a link, have it appear in a link color (blue).

And I think we should include the author name for the article (or whatever was vommitted on)

Screen Shot 2020-08-28 at 2 04 31 PM

app/controllers/admin/users_controller.rb
```diff
@@ -170,6 +171,16 @@ def remove_org_credits
...
     end
 
+    def set_related_reactions
+      user_article_ids = @user.articles.ids
+      user_comment_ids = @user.comments.ids
+      @related_vomit_reactions = Reaction.where(reactable_type: "Comment", reactable_id: user_comment_ids,
+                                                category: "vomit")
+        .or(Reaction.where(reactable_type: "Article", reactable_id: user_article_ids, category: "vomit"))
+        .includes(:reactable)
+        .order(created_at: :desc).limit(15)
```
msarit, line 181: Does `limit(15)` mean we are limiting the results to the first 15? If so, why? Or another way to ask is, could Forem admins have access to records not contained in this 15?

juliannatetreault: Yes, the `limit(15)` limits the records to 15 records per user. I have done this for scalability reasons and to follow the convention of the other sections similar to this in `/admin/users` like the "Recent Emails" and "Recent Notes" section, but I'm open to suggestions!

app/views/admin/users/_reactions.html.erb
juliannatetreault: These are nice suggestions @msarit! I actually styled everything this way per @Zhao-Andy's request on the issue itself, so maybe Andy would have some more insight on whether or not these stylistic changes are necessary. As for adding the author name, I'm not sure if that would add unnecessary data to the view -- we want to be able to see what was vomited on and who the vomit was for, but I don't know if we need to include the author name, since theoretically, the Admin should be able to click through to the content that was vomited on and gather extra data there!

app/controllers/admin/users_controller.rb
```diff
@@ -170,6 +171,16 @@ def remove_org_credits
       Credit.remove_from(org, amount)
     end
 
+    def set_related_reactions
+      user_article_ids = @user.articles.ids
+      user_comment_ids = @user.comments.ids
+      @related_vomit_reactions = Reaction.where(reactable_type: "Comment", reactable_id: user_comment_ids,
+                                                category: "vomit")
+        .or(Reaction.where(reactable_type: "Article", reactable_id: user_article_ids, category: "vomit"))
```
Zhao-Andy, line 179: Sorry, I should've included this on the example code I gave, but I think we'll also want an `.or` for vomit reactions on the user, too.

app/controllers/admin/users_controller.rb
Zhao-Andy: I think limiting to 15 is fine -- we might want to paginate it in a future PR.

app/views/admin/users/_reactions.html.erb
Zhao-Andy: I think `.empty?` is fine! I don't think it would make sense to use `counter_culture` here since you've already loaded these reactions already.

app/views/admin/users/_reactions.html.erb
Zhao-Andy: I think @msarit's style suggestions are nice! I think it would be worth updating in this PR.

I agree we don't need the author name since these reactions are for the author already.

msarit:
Just to be clear, this PR accomplishes what going to `/admin/privileged_reactions` and filtering by username would as well? Showing all the vomits that a user has made?

@juliannatetreault @Zhao-Andy

app/views/admin/users/edit.html.erb
```diff
@@ -46,6 +46,7 @@
 </div>
 
 <%= render "notes" %>
+<%= render "reactions" %>
```
Zhao-Andy, line 49: Maybe call the file `negative_reactions`?

msarit:
Last comment 😄 : This might be a caching issue, but when I generate vomits as a `trusted` user, then check back in that user's profile, I do not see their vomits under `Recent Reactions` (although I see them right away under `Priviledged Reactions`).

Zhao-Andy:
> Just to be clear, this PR accomplishes what going to `/admin/privileged_reactions` and filtering by username would as well? Showing all the vomits that a user has made?

To clarify, this shows the vomits that a user has received, and not the vomits they have made, so the two pages are doing two different things. :)

app/controllers/admin/users_controller.rb
juliannatetreault: No worries! I almost added that in, but I thought that when you `vomit` on a User, it also vomits on their Articles, so those `vomit`s should fall within the query for Articles too, correct @Zhao-Andy?
Screen Shot 2020-08-28 at 12 58 10 PM

app/views/admin/users/edit.html.erb
juliannatetreault: I'll make this change now!

app/controllers/admin/users_controller.rb
Zhao-Andy: It actually just lowers the score of their articles, instead of creating vomit reactions for all of their articles. Maybe we should update the language in that message.

juliannatetreault: Okay, good to know! I think we definitely should -- should we make a card for this and add it to the project board or should I address this in this PR?

Zhao-Andy: Let's make a card in the project 👍

juliannatetreault: Will do!

app/views/admin/users/_reactions.html.erb
juliannatetreault: Agreed and making these now, as well as taking into account the logic for `reactable_type: "User"`!

rhymes:
> @msarit To clarify, this shows the vomits that a user has received, and not the vomits they have made, so the two pages are doing two different things. :)

I definitely didn't understand this! :D I thought it was the vomits the user made, not received. Thanks for clarifying team!
I'm not sure how to leave vomit reactions TBH so I ended up using the console but I can't see them in the panel :( 

What's the best way to create the content to test this?

juliannatetreault:
> > @msarit To clarify, this shows the vomits that a user has received, and not the vomits they have made, so the two pages are doing two different things. :)
> 
> I definitely didn't understand this! :D I thought it was the vomits the user made, not received. Thanks for clarifying team!
> I'm not sure how to leave vomit reactions TBH so I ended up using the console but I can't see them in the panel :(
> 
> What's the best way to create the content to test this?

I would suggest creating them in the console so that you can see the IDs of the User you're creating the vomits for -- this makes it easier to then check the vomits against the `user_id`s in `admin/users` (I think, at least!). That being said, to create _most_ of my vomits for myself and other seeded users I did the following:
```
me = User.last
me.vomitted_on?    # if false, then proceed to create content
article1 = Article.create(...)
vomit1 = Reaction.create(reactable_id: 27, reactable_type: "Article", user_id: 11, category: "vomit")
# reactable_id is the Article ID and user_id matches my User's ID
```
Additionally, you can vomit on comments and/or write a post as your User and have other Users vomit on it and your comments (via the console) to create the reactions that you need. Let me know if you have any further questions @rhymes -- hopefully this helps! 😄

rhymes (approved):
Thanks @juliannatetreault! 

I was able to create the content needed to be displayed there.

LGTM

PR #10067, opened 2020-08-28
Optimization:Ignore Docs for Heroku Slug

mstruve:
- [x] Optimization

## Description
Our Heroku Slug size has been steadily growing and is currently at 470MB of the allotted 500MB. One way to decrease slug size is to ignore files that are not needed for the app to run on Heroku. Since we serve our docs through Netlify I am pretty sure there is no reason to have them sent to Heroku. 

https://github.com/orgs/forem/projects/6#card-44452848

alt_text


Zhao-Andy:
@mstruve updated my comment above because I hit enter too quickly 🙃

rhymes (approved):

PR #10073, opened 2020-08-29
Allow initial user to sign up and configure app

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Gives users the ability to be the first user and automatically become super users and move into start mode.

Now when a new forem is created, it will start with this screen....

start up mode

(This is just the MVP, and I figured having a cute gif _for now_ is fun)

Also adds new "mode" functionality to seed data and changes default for new forem to "private"... which makes sense as the default I think, especially pre-configuration.

### DIFF
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -11,6 +11,8 @@ class SiteConfig < RailsSettings::Base
   STACK_ICON = File.read(Rails.root.join("app/assets/images/stack.svg")).freeze
   LIGHTNING_ICON = File.read(Rails.root.join("app/assets/images/lightning.svg")).freeze
 
+  field :waiting_on_first_user, type: :boolean, default: !User.exists?
+
   # API Tokens
   field :health_check_token, type: :string
 
@@ -148,7 +150,7 @@ class SiteConfig < RailsSettings::Base
   field :feed_style, type: :string, default: "basic"
   # a non-public forem will redirect all unauthenticated pages to the registration page.
   # a public forem could have more fine-grained authentication (listings ar private etc.) in future
-  field :public, type: :boolean, default: 1
+  field :public, type: :boolean, default: 0
   # The default font for all users that have not chosen a custom font yet
   field :default_font, type: :string, default: "sans_serif"
 
--- a/app/views/shared/authentication/_email_registration_form.html.erb
+++ b/app/views/shared/authentication/_email_registration_form.html.erb
@@ -1,61 +1,55 @@
-<% if SiteConfig.allow_email_password_registration %>
-  <div id="sign-in-password-form" class="mt-8 mb-6 crayons-card p-7 align-left mx-auto" style="max-width:580px;">
-    <%= form_for(User.new, as: :user, url: registration_path(:user)) do |f| %>
-      <% if resource.errors.any? %>
-        <div class="crayons-card crayons-card--secondary crayons-notice crayons-notice--danger">
-          <div class="crayons-card__header">
-            <h3 class="crayons-card__headline">
-              Whoops, we found <%= pluralize(resource.errors.size, "problem") %>
-            </h3>
-          </div>
-          <div class="crayons-card__body">
-            <ul>
-              <% resource.errors.full_messages.each do |message| %>
-                <li><%= message %></li>
-              <% end %>
-            </ul>
-          </div>
+<div id="sign-in-password-form" class="mt-8 mb-6 crayons-card p-7 align-left mx-auto" style="max-width:580px;">
+  <%= form_for(User.new, as: :user, url: registration_path(:user)) do |f| %>
+    <% if defined?(resource) && resource&.errors&.any? %>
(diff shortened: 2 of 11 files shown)


app/models/site_config.rb
```diff
@@ -11,6 +11,8 @@ class SiteConfig < RailsSettings::Base
   STACK_ICON = File.read(Rails.root.join("app/assets/images/stack.svg")).freeze
   LIGHTNING_ICON = File.read(Rails.root.join("app/assets/images/lightning.svg")).freeze
 
+  field :waiting_on_first_user, type: :boolean, default: !User.exists?
```
benhalpern, line 14: The "default" will be set when the app is first spun up and it will be based on whether there are any users.

app/models/site_config.rb
```diff
@@ -148,7 +150,7 @@ class SiteConfig < RailsSettings::Base
   field :feed_style, type: :string, default: "basic"
   # a non-public forem will redirect all unauthenticated pages to the registration page.
   # a public forem could have more fine-grained authentication (listings ar private etc.) in future
-  field :public, type: :boolean, default: 1
+  field :public, type: :boolean, default: 0
```
benhalpern, line 153: Changed the default state here from public to private. This makes private forem the default but can obviously be changed at config. Existing forems will already have a value here and the default won't need to apply.

app/views/shared/authentication/_email_registration_form.html.erb
```diff
@@ -1,61 +1,55 @@
...
+      <%= f.file_field :profile_image, accept: "image/*", class: "crayons-card crayons-card--secondary p-3 flex items-center flex-1 w-100", required: true %>
+    </div>
 
-      <div class="crayons-field mt-2">
-        <%= f.label :name, class: "crayons-field__label" %>
-        <%= f.text_field :name, class: "crayons-textfield" %>
-      </div>
+    <div class="crayons-field mt-2">
+      <%= f.label :name, class: "crayons-field__label" %>
+      <%= f.text_field :name, class: "crayons-textfield", required: true %>
```
benhalpern, line 28: Didn't change much in this file, but marked stuff as required for better UX

db/seeds.rb
```diff
@@ -30,6 +30,21 @@ def create_if_none(klass, count = nil)
...
+# Default development site config if different from production scenario
+
+SiteConfig.public = true
+SiteConfig.waiting_on_first_user = false
+
+##############################################################################
+
+# Put forem into "starter mode"
+
+if ENV["MODE"] == "STARTER"
```
benhalpern, line 42: Adds first new "mode", I imagine we could create more different scenarios this way for better developer experience.

Like for spinning up different configs, or putting us in different scenarios of follow behavior, types of listings created, chat channel situations, etc.

mstruve:
How are Forems in this `waiting_on_first_user` state protected from a bad actor taking over the Forem?

benhalpern:
@mstruve the current protection would be the obfuscation of the address of the forem I'd think.

`forem_b58jkls87345d.forem.cloud` could essentially act as the password, and the stakes are pretty low for the mishap with this because there's not much to take over with a blank forem.... But we could alternatively also spin up a secret which could be handed to the user to input upon signing up to further verify their authenticity.

Since the stakes are low I could see it making sense to ship this now to help improve new creator UX and then get to working on the secret-generation as a follow-on and coordinate with systems to get the forem secret environment variable spun up alongside new forems.

I imagine the secret could be randomly generated as a new environment variable when a forem is first spun up to be added as a signup field and we could hand that off to the user along with the URL as part of the standard procedure... And of course we'd 100% need something like that when this is in self-serve.

@jdoss thoughts?

benhalpern:
To follow on with that thought.... @mstruve @jdoss 

Adding basic auth would be pretty straightforward as the very first layer if we needed this now, but I think it might be better to add that layer as its own PR after merging this, just to give that solution room to breath instead of cramming it in here.

```ruby
      if ENV["FOREM_OWNER_SECRET"].present? && SiteConfig.waiting_on_first_user
          authenticate_or_request_with_http_basic "Application" do |name, password|
              name == ENV["FOREM_NAME"] && password == ENV["FOREM_OWNER_SECRET"]
          end
          return
      end
```

jdoss:
If it is not hard to do, having the `FOREM_OWNER_SECRET` auto generated on setup is a good idea.

benhalpern:
Sounds good. @jdoss let's get `FOREM_OWNER_SECRET` as a random secret that gets spun up along with new forems, and I'll follow up this PR with another one which adds that as another field in the form if `FOREM_OWNER_SECRET` is present in the env variables (e.g. if systems doesn't set this value we don't look for it or ask for it, so we're good either way)

benhalpern:
Alright, I'm going to merge this to limit the scope of this PR and quickly follow up with the `FOREM_OWNER_SECRET` option.

PR #10094, opened 2020-08-30
Fixes unsafe reflection warning in dashboards_controller

diogoosorio:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This one is purely to make CodeClimate/brakeman happy as the code that was already there is equivalent (as the parameter being `constantize`'d was validated before the reflection part kicked in).

I've switch things around and do a lookup on the `UserSubscription::ALLOWED_TYPES` for the subscription `source_type`, which makes brakeman understand that type comes from an allowed list of types (instead the user input).

#3739

Guarantee that no regression was introduced (the code is indeed equivalent):

1. The `requests/dashboard_spec.rb` covers these scenarios

---

1. Visit: http://localhost:3000/dashboard/subscriptions?source_type=Article&source_id=1
1. See that the page loads the subscriptions of the correct article

---

1. Visit: http://localhost:3000/dashboard/subscriptions?source_type=Comment&source_id=1
1. See that that a 404 page is rendered

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

alt_text

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -78,7 +78,7 @@ class DashboardsController < ApplicationController
   def set_source
     source_type = UserSubscription::ALLOWED_TYPES.detect { |allowed_type| allowed_type == params[:source_type] }
 
-    not_found if source_type.nil?
+    not_found unless source_type
 
     source = source_type.constantize.find_by(id: params[:source_id])
     @source = source || not_found
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -76,8 +76,9 @@ class DashboardsController < ApplicationController
   private
 
   def set_source
-    source_type = params[:source_type]
-    not_found unless UserSubscription::ALLOWED_TYPES.include? source_type
+    source_type = UserSubscription::ALLOWED_TYPES.detect { |allowed_type| allowed_type == params[:source_type] }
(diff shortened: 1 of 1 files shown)


citizen428:
Thanks for this PR! I'm not a big fan of changing code to please a tool, especially since the `detect` version is just writing out explicitly what `include?` does. But I have no problem merging this if others are ok with it.

app/controllers/dashboards_controller.rb
```diff
@@ -76,8 +76,9 @@ def subscriptions
   private
 
   def set_source
-    source_type = params[:source_type]
-    not_found unless UserSubscription::ALLOWED_TYPES.include? source_type
+    source_type = UserSubscription::ALLOWED_TYPES.detect { |allowed_type| allowed_type == params[:source_type] }
+
+    not_found if source_type.nil?
```
rhymes, line 81:
```suggestion
    not_found unless source_type
```

rhymes (commented):
I'd change the `if` to an `unless` but other than that I agree that CC is overzealous here but as it's not breaking anything, we can merge it

rhymes (approved):
Thanks @diogoosorio

PR #10097, opened 2020-08-30
Allow admins to set brand color

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This feature allows admins to set their own main brand color....

Screen Shot 2020-08-30 at 5 00 14 PM
Screen Shot 2020-08-30 at 4 59 54 PM
Screen Shot 2020-08-30 at 4 58 44 PM

This color needs to contrast properly with white, so I added the wcag contrast gem. I think this is a good gem to add, because we can start using this elsewhere, like user profile colors, tags, etc. Currently we let users set non-contrasting colors and basically just moderate it. This will help us ensure a minimum of 4.5:1 contrast for readability and accessibility for all.

The gem has not recently been updates, but it is a fairly simple low-bloat utility gem. The gem author is active and I feel like we can be comfortable that we could merge a change if we needed to.

Also added some basic validations in the controller. I figured we can start adding some validations in this area and extract these to a better place in a future refactor once we establish more of these.


app/controllers/admin/configs_controller.rb
```diff
@@ -100,6 +109,17 @@ def bust_relevant_caches
       CacheBuster.bust("/tags/onboarding")
     end
 
+    # Validations
+    def brand_contrast_too_low
+      hex = params[:site_config][:primary_brand_color_hex]
+      hex.present? && WCAGColorContrast.ratio(hex.delete("#"), "ffffff") < 4.5 # a11y
```
citizen428, line 115: I've mentioned this before, but directly sprinkling dependencies into the codebase is often not great (I know this is only used in one place for now). It's often better to create a wrapper object and use that, so if  `WCAGColorContrast` really disappears/goes stale/etc we only have to update the wrapper, not every single call-site. This way you can also encapsulate some of the logic a bit better.

```ruby
class Color
  MIN_CONTRAST = 4.5

  def self.low_contrast?(color, base_color = "ffffff")
    return unless(color.present?)

    WCAGColorContrast.ratio(color.delete("#"), base_color.delete("#")) < MIN_CONTRAST
   end
end
```

Then this just becomes

```ruby
      hex = params[:site_config][:primary_brand_color_hex]
      Color.low_contrast?(hex)
```

which is really all the reader of this validation needs to know 😃

app/controllers/admin/configs_controller.rb
```diff
@@ -100,6 +109,17 @@ def bust_relevant_caches
...
+    end
+
+    def brand_color_not_hex
+      hex = params[:site_config][:primary_brand_color_hex]
+      hex.present? && !hex.match(/\A#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})\z/)
```
citizen428, line 120: There's a predicate version of `match` called `match?`.  There are also relevant character classes defined for hex digits:

```
/\h/ - A hexdigit character ([0-9a-fA-F])
/\H/ - A non-hexdigit character ([^0-9a-fA-F])
/[[:xdigit:]]/ - Digit allowed in a hexadecimal number (i.e., 0-9a-fA-F)
```

Using both we'd end up with this
```suggestion
      hex.present? && !hex.match?(/\A#(\h{6}|\h{3})\z/)
```

app/views/shell/_top.html.erb
```diff
@@ -51,7 +51,16 @@
...
-        <div id="body-styles"></div>
+        <div id="body-styles">
+          <style>
+            :root {
+              --accent-brand: <%= SiteConfig.primary_brand_color_hex %>;
+              --accent-brand-darker: <%= HexComparer.new([SiteConfig.primary_brand_color_hex]).brightness(0.85) %>;
+              --accent-brand-lighter: <%= HexComparer.new([SiteConfig.primary_brand_color_hex]).brightness(1.1) %>;
+              --accent-brand-a10: #{rgba(<%= SiteConfig.primary_brand_color_hex %>, 0.1)};
+            }
+          </style>
```
ludwiczakpawel, line 62: I'm not sure if that's correct but it seems like THIS will be replaced in [certain cases](https://github.com/forem/forem/blob/d86b9d9ce5b62535c613343b57262262fc1d125d/app/views/layouts/_user_config.html.erb#L9-L15).  Is that intended/expected?

ludwiczakpawel: Another thing: are we sure that updating the brightness is ENOUGH to make lighter/darker variants look good? Because this will look different when your brand color is #000 vs some shade of purple vs some shade of light blue.

ludwiczakpawel (commented):
Left few comments, not blockers though.

app/assets/stylesheets/components/buttons.scss
```diff
@@ -261,6 +265,18 @@
...
 
+.crayons-btn--brand-facebook {
+  --bg: var(--brand-facebook-bg);
+  --bg-hover: var(--brand-facebook-bg-hover);
+  --color: var(--brand-facebook-color);
+  --color-hover: var(--brand-facebook-color);
+  --bg-inverted: var(--brand-facebook-bg);
+  --bg-inverted-hover: var(--brand-facebook-bg-hover);
+  --color-inverted: var(--brand-facebook-color);
+  --color-inverted-hover: var(--brand-facebook-color);
```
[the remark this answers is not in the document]
benhalpern: @ludwiczakpawel they actually do belong here. I noticed in doing this that Facebook's styles aren't described anywhere so that button defaults to brand color. I think that went unnoticed because our blue-purple button is decently close to Facebook brand blue.

app/views/shell/_top.html.erb
benhalpern: Yeah, my thoughts are that the default theme can be branded but the other themes don't need the brand colors because they may clash.

That makes sense about the brightness needing to vary depending on the hex.... We could likely programmatically determine the darkness of a hex and adjust our adjustment based on the root darkness. Given that we're only accepting one admin-generated field here and then working from it I'd say it's okay if we ship it without having all the accents worked out and then seek to get that right after the fact?

app/assets/stylesheets/components/buttons.scss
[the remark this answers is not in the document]
ludwiczakpawel: gotcha.. well, then good catch! :)

app/views/shell/_top.html.erb
ludwiczakpawel: makes sense 👍

app/controllers/admin/configs_controller.rb
benhalpern: Great call. I added a new `Color::Accessibility` class, and it's definitely cleaner.

app/services/color/accessibility.rb
```diff
@@ -0,0 +1,11 @@
+module Color
```
citizen428, line 1: Usually, we tend to use plurals for service name spaces, but don't let that stop this PR, I can change it during cooldown and potentially move the `HecComparer` into the `Colors` namespace too, keeping related things together and all.

citizen428 (approved):
Nice, thanks for making the changes @benhalpern 😃

PR #10098, opened 2020-08-30
Purge core edge/fragment caches on config update

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Added a few cache busts which can help ensure the relevant areas of the site change as soon as config is changed. We could get more granular with this in the future, but config won't change very often, so not a big deal to not have that optimization right now.

`delete_matched` is not currently in our codebase, so I hope it works as expected with Redis. This looks for any caches that match `RELEASE_FOOTPRINT` which are the caches we expect might need to change when we push new code. This is a similar scenario of new content like logos, etc.

### DIFF
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -100,7 +100,7 @@ module Admin
       CacheBuster.bust("/shell_top") # Cached at edge, sent to service worker.
       CacheBuster.bust("/shell_bottom") # Cached at edge, sent to service worker.
       CacheBuster.bust("/onboarding") # Page is cached at edge.
-      Rails.cache.delete_matched(ApplicationConfig["RELEASE_FOOTPRINT"]) # Clear relevant fragment caches tied to this key.
+      Rails.cache.delete_matched(ApplicationConfig["RELEASE_FOOTPRINT"]) # Delete all caches tied to this key.
     end
 
     def campaign_params
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -97,6 +97,9 @@ module Admin
 
     def bust_relevant_caches
       CacheBuster.bust("/tags/onboarding") # Needs to change when suggested_tags is edited.
+      CacheBuster.bust("/shell_top") # Cached at edge, sent to service worker.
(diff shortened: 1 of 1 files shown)


app/controllers/admin/configs_controller.rb
```diff
@@ -96,8 +96,11 @@ def clean_up_params
     end
 
     def bust_relevant_caches
-      # Needs to change when suggested_tags is edited.
-      CacheBuster.bust("/tags/onboarding")
+      CacheBuster.bust("/tags/onboarding") # Needs to change when suggested_tags is edited.
+      CacheBuster.bust("/shell_top") # Cached at edge, sent to service worker.
+      CacheBuster.bust("/shell_bottom") # Cached at edge, sent to service worker.
+      CacheBuster.bust("/onboarding") # Page is cached at edge.
+      Rails.cache.delete_matched(ApplicationConfig["RELEASE_FOOTPRINT"]) # Delete all caches tied to this key.
```
citizen428, line 103: Can  `RELEASE_FOOTPRINT` contain any glob patterns? If yes, you need to escape them with `\` to match verbatim. The glob patterns are `?`, `*`, character classes (`[ab]`) and negated character classes (`[^ab]`). Single `.` as in version numbers are ok:

```
❯ redis-cli
127.0.0.1:6379> MSET foo 1 fao 2 f.o 3
OK
127.0.0.1:6379> KEYS f.o
1) "f.o"
```

citizen428 (approved):

app/controllers/admin/configs_controller.rb
benhalpern: It's set by the environment 

`export RELEASE_FOOTPRINT=$(date -u +'%Y-%m-%dT%H:%M:%SZ')` in `entrypoint.sh`... So no we don't have any glob patterns... Someone could set this otherwise in their environment but that seems like an unwise choice on their part.

I think we should be good on all this.

citizen428: Thanks @benhalpern! I sort of remembered what `RELEASE_FOOTPRINT` looks like, the comment was also for refreshing my own memory of how matching on Redis keys works.

PR #10103, opened 2020-08-31
Bug Fix:Destroy GithubRepos with Suspended or Blocked Access

mstruve:
- [x] Bug Fix

## Description
We have some failing `GithubRepos::RepoSyncWorker` due to the following errors. In the even that a user account is suspended or if they no longer have access to a repo we should destroy the repo. 

Fixes: https://app.honeybadger.io/fault/66984/6cb5b7e34a5b6c538664cfd6135f2c97

https://github.com/orgs/forem/projects/6#card-44619717

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/workers/github_repos/repo_sync_worker.rb
+++ b/app/workers/github_repos/repo_sync_worker.rb
@@ -29,6 +29,11 @@ module GithubRepos
         repo.user&.touch(:github_repos_updated_at)
       rescue Github::Errors::NotFound, Github::Errors::Unauthorized
         repo.destroy
+      rescue Github::Errors::ClientError => e
+        msg = e.message
+        raise e unless msg.include?("Your account was suspended") || msg.include?("Repository access blocked")
+
+        repo.destroy
       end
     end
   end
--- a/spec/workers/github_repos/repo_sync_worker_spec.rb
+++ b/spec/workers/github_repos/repo_sync_worker_spec.rb
@@ -33,5 +33,39 @@ RSpec.describe GithubRepos::RepoSyncWorker, type: :worker do
         expect(old_updated_at).not_to eq(GithubRepo.find(repo.id).updated_at)
       end
     end
+
+    it "destroys unfound repos" do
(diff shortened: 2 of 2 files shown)


app/workers/github_repos/repo_sync_worker.rb
```diff
@@ -28,6 +28,11 @@ def perform(repo_id)
         )
         repo.user&.touch(:github_repos_updated_at)
       rescue Github::Errors::NotFound, Github::Errors::Unauthorized
+        repo.destroy
+      rescue Github::Errors::ClientError => e
+        msg = e.message
+        raise e unless msg.include?("Your account was suspended") || msg.include?("Repository access blocked")
```
rhymes, line 34: Great catch! BTW this is because we're not wrapping the two correct exceptions for these two statuses:

`Octokit::RepositoryUnavailable` and `Octokit::AccountSuspended`. See https://github.com/octokit/octokit.rb/blob/a69f5eab0a2fb2763fcda9fadcb0c961df96a373/lib/octokit/error.rb#L265 and https://github.com/octokit/octokit.rb/blob/a69f5eab0a2fb2763fcda9fadcb0c961df96a373/lib/octokit/error.rb#L257

If you add two exceptions called `RepositoryUnavailable` and `AccountSuspended` to https://github.com/forem/forem/blob/master/app/services/github/errors.rb you should be able to catch handle them without inspecting the message

rhymes (approved):
Left a note but it can be done in a refactoring

app/workers/github_repos/repo_sync_worker.rb
mstruve: Oh yeah for sure that would make life easier. going to merge as is and leave that for a refactor. Maybe a good little issue for Hacktoberfest :)

rhymes: Going to make an issue out of the comment, otherwise we'll forget about it :D

PR #10105, opened 2020-08-31
Admin Reports view hot fix

ludwiczakpawel:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR fixes an issue that was breaking layout on `/admin/reports` page. It was caused by me putting piece of code in a wrong place with my admin updates couple hours ago.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/views/admin/feedback_messages/_latest_articles.html.erb
+++ b/app/views/admin/feedback_messages/_latest_articles.html.erb
@@ -9,8 +9,8 @@
     <div id="articlesBodyContainer" class="collapse hide" aria-labelledby="articleHeader" data-parent="#latestArticles">
       <div class="card-body" style="overflow: scroll; max-height: 500px;">
         <% @new_articles.each do |article| %>
+          <% next if article.user.badge_achievements_count > 2 %>
           <div class="flex py-2">
-            <% next if article.user.badge_achievements_count > 2 %>
             <a href="<%= article.path %>"><%= article.title %></a>
 
             <div class="ml-auto shrink-0">


vaidehijoshi:
Do we have any tests around this/can we add some? I don't think it needs to happen as part of _this_ PR since this is a hotfix, but ideally we'd add a test to ensure that we don't accidentally break this in the same way again :)

rhymes:
@vaidehijoshi oops sorry, hadn't seen your comment. I don't think we have a test for this currently. 

@ludwiczakpawel if you add self explanatory commit messages then when the PR is merged the commit message won't just be `fix` or `.` ;-) The title of the PR doesn't really count, though we can change the message on merge, it might be just easier to keep a "talking" message right from the start

vaidehijoshi:
> @vaidehijoshi oops sorry, hadn't seen your comment. I don't think we have a test for this currently.

If this area of the codebase is being worked on currently, I think adding in a _new_ test would be a good call here! For context, I'm basically trying to figure out how we didn't catch it when it was merged initially and trying to ensure that we have some safeguards in place to avoid that. Even just a rudimentary "I expect `x` number of reports to render" test case would fail here and would catch this in the future

ludwiczakpawel:
@rhymes good point! 🙌

> "I expect x number of reports to render"

I think the fail here was mostly frontend related. But overall, as much as I hate tests and specs because they are pain in the ass and not very YOLO (that's Paweł's Personal opinion, not Paweł's Professional opinion), I do think having them would help catch these kind of things.

PR #10108, opened 2020-08-31
Add logic for asking for FOREM_OWNER_SECRET to be present on first user sign up if set by system

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This adds an extra field to require brand new forems to pass in a value generated by systems stack to ensure a publicly generated forem can't be coopted if someone has the URL.

Screen Shot 2020-08-31 at 1 31 24 PM

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,7 +17,14 @@ class ApplicationController < ActionController::Base
     error_too_many_requests(exc)
   end
 
-  PUBLIC_CONTROLLERS = %w[shell async_info ga_events service_worker omniauth_callbacks registrations].freeze
+  PUBLIC_CONTROLLERS = %w[shell
+                          async_info
+                          ga_events
+                          service_worker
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -17,7 +17,14 @@ class ApplicationController < ActionController::Base
...
   end
 
-  PUBLIC_CONTROLLERS = %w[shell async_info ga_events service_worker omniauth_callbacks registrations].freeze
+  PUBLIC_CONTROLLERS = %w[shell
+                          async_info
+                          ga_events
+                          service_worker
+                          omniauth_callbacks
+                          registrations
+                          confirmations
```
benhalpern, line 26: Added confirmations and passwords as controllers that shouldn't have been in the _not allowed_ camp in the first place. Not really related to PR.

PR #10115, opened 2020-08-31
Add cache-control to CarrierWave when using S3

maestromac:
- [x] Feature

## Description
We don't have any cache-control configuration in place for uploaded images to S3. We need to have it in place or else nginx won't cache image from s3.

https://github.com/carrierwaveuploader/carrierwave#using-amazon-s3

QA on production!

## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a, but ideally at somepoint we update all the images on our s3 to have a cache-control


benhalpern:
> Will this have any effect on apps not using Nginx?

I suspect not, but can't hurt.

PR #10118, opened 2020-08-31
Comment out Flaky Reaction Search Sort Spec

mstruve:
- [x] Spec Fix

## Description
I know exactly what the problem is with this spec, it has to do with how we are storing this ID field in Elasticsearch making accurately sorting by it not really possible. The fix for it is a bigger PR but in the meantime I dont want this to be a blocker for anyone. 

alt_text


benhalpern:
If the tests don't pass it can literally only be because of another flaky test. So I think we're good to merge before waiting.

PR #10123, opened 2020-09-01
Add unique index to articles slug and feed_source_url

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

In an effort to unpack https://github.com/forem/forem/pull/8072 I'm going to start with the easy part: adding a unique index to columns that have no duplicates in the table:

- `slug` + `user_id` to make sure that slugs are indeed unique per index, not just at the Rails level but also at the DB level
- `feed_source_url` to make sure that each article has that unique as well

Currently there are no duplicates for either: 

- https://dev.to/admin/blazer/queries/78-duplicate-articles-by-slug-and-user_id
- https://dev.to/admin/blazer/queries/79-duplicate-articles-feed_source_url

https://github.com/forem/forem/pull/8072

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -49,30 +49,29 @@ class Article < ApplicationRecord
            inverse_of: :commentable,
            class_name: "Comment"
 
-  validates :slug, presence: { if: :published? }, format: /\A[0-9a-z\-_]*\z/,
-                   uniqueness: { scope: :user_id }
-  validates :title, presence: true,
-                    length: { maximum: 128 }
-  validates :user_id, presence: true
-  validates :feed_source_url, uniqueness: { allow_blank: true }
-  validates :canonical_url,
-            url: { allow_blank: true, no_local: true, schemes: %w[https http] },
-            uniqueness: { allow_blank: true }
   validates :body_markdown, length: { minimum: 0, allow_nil: false }, uniqueness: { scope: %i[user_id title] }
-  validate :validate_tag
-  validate :validate_video
-  validate :validate_collection_permission
-  validate :past_or_present_date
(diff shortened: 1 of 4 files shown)


app/models/article.rb
```diff
@@ -49,30 +49,29 @@ class Article < ApplicationRecord
...
-  validate :validate_tag
-  validate :validate_video
-  validate :validate_collection_permission
-  validate :past_or_present_date
-  validate :canonical_url_must_not_have_spaces
-  validates :video_state, inclusion: { in: %w[PROGRESSING COMPLETED] }, allow_nil: true
   validates :cached_tag_list, length: { maximum: 126 }
+  validates :canonical_url, url: { allow_blank: true, no_local: true, schemes: %w[https http] },
+                            uniqueness: { allow_blank: true }
+  validates :feed_source_url, uniqueness: { allow_nil: true }
```
rhymes, line 56: this is the only change, we can allow nils as the column is nullable, we shouldn't allow empty strings

db/migrate/20200901074007_add_unique_index_to_articles_slug_and_feed_source_url.rb
```diff
@@ -0,0 +1,33 @@
+class AddUniqueIndexToArticlesSlugAndFeedSourceUrl < ActiveRecord::Migration[6.0]
```
rhymes, line 1: as there's a tiny chance that a migration breaks but one of the indexes passes through we better be idempotent as much as possible here ;)

Issue #10106, opened 2020-08-31
Refactor: improve error handling in GithubRepos::RepoSyncWorker


rhymes:
Currently we're inspecting the error message to determine if a user's account was suspended or if the repository has become unavailable so that we can destroy it

https://github.com/forem/forem/blob/25e20bdc1ce3e66282298a88cdf55f89b1b1e9a5/app/workers/github_repos/repo_sync_worker.rb#L32-L37

`Ockokit` (the GitHub library we use) has two specific errors for those two cases: `Octokit::RepositoryUnavailable` and `Octokit::AccountSuspended`.

It'd be nice to replace those messages inspection with the correct error classes.

See https://github.com/octokit/octokit.rb/blob/a69f5eab0a2fb2763fcda9fadcb0c961df96a373/lib/octokit/error.rb#L265 and https://github.com/octokit/octokit.rb/blob/a69f5eab0a2fb2763fcda9fadcb0c961df96a373/lib/octokit/error.rb#L257

By adding two exceptions similarly named, `RepositoryUnavailable` and `AccountSuspended`, to https://github.com/forem/forem/blob/master/app/services/github/errors.rb one should be able to handle them correctly in the worker, without inspecting the message and thus keeping the same behaviour

_Originally posted by @rhymes in https://github.com/forem/forem/pull/10103#discussion_r480245028_

diogoosorio:
FYI I'll give this one a crack. 😄

rhymes:
@diogoosorio are you still working on this? Thank you

diogoosorio:
Hum, I kind of already did (??):

https://github.com/forem/forem/pull/10131/files#diff-1dca539dd1dfbbc75ce747cbd737698d

I haven't looked into it, it was either reverted for some reason or there was some sort of a concurrency issue which lead to git merge this incorrectly. I'll look into the version control history and publish my findings here. 😄

diogoosorio:
Nevermind, this is in the state I was expecting on master:

https://github.com/forem/forem/blob/master/app/workers/github_repos/repo_sync_worker.rb#L30

Unless I entirely misunderstood the problem, I think this is done. If I have misunderstood something, please ELI5 it to me. 🙏

rhymes:
You're right! Thanks @diogoosorio

PR #10131, opened 2020-09-01
Rely on error types to determine when a repo should be removed on the RepoSyncWorker

diogoosorio:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Makes the `RepoSyncWorker` rely on the error types and not their messages, to make the decision of wether to remove the Github repository reference from the database.

#10106

TBH I don't quite know how have a repository reach that state (have an account suspended or a repository deemed "unavailable"), so I wasn't able to do a proper functional test.

I tried to cover the full codepath with tests as an alternative. But beware, I'm missing a proper functional test.

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

alt_text

### DIFF
--- a/app/services/github/errors.rb
+++ b/app/services/github/errors.rb
@@ -17,5 +17,11 @@ module Github
 
     class InvalidRepository < ArgumentError
     end
+
+    class AccountSuspended < ClientError
+    end
+
+    class RepositoryUnavailable < ClientError
+    end
   end
 end
--- a/app/workers/github_repos/repo_sync_worker.rb
+++ b/app/workers/github_repos/repo_sync_worker.rb
@@ -27,12 +27,10 @@ module GithubRepos
           info_hash: fetched_repo.to_hash,
         )
         repo.user&.touch(:github_repos_updated_at)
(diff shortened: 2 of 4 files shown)


diogoosorio:
Thanks @mstruve! I think that this block of code does that implicitly:

https://github.com/forem/forem/blob/7b5bc89b7c162912e41b3fc94ce8cd61befef079/app/services/github/oauth_client.rb#L81-L83

I was unsure as well when I was looking at it, that's why I actually ended up adding those 2 specific tests to the OAuth client class (to make sure I hadn't mis-interpreted how those errors were being mapped). 😄

rhymes:
@mstruve @diogoosorio is correct, the magic is already done in there. If there's a a class it raises that, otherwise it raises the generic one.

Thanks @diogoosorio!

rhymes (approved):
LGTM! :)

PR #10137, opened 2020-09-01
WIP - Add routes for i18n using optional locale

atsmith813:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [ ] no documentation needed

alt_text

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,4 @@
 # rubocop:disable Metrics/BlockLength
-LOCALE_LANG_REGEXP = /lang-.*/i.freeze
 
 Rails.application.routes.draw do
   use_doorkeeper do
@@ -23,7 +22,8 @@ Rails.application.routes.draw do
 
   # [@forem/delightful] - all routes are nested under this optional scope to
   # begin supporting i18n.
-  scope "(:locale)", locale: LOCALE_LANG_REGEXP, defaults: { locale: nil } do
+  scope "(/locale/:locale)", defaults: { locale: nil } do
+    get "/locale/:locale" => "stories#index"
     require "sidekiq/web"
     require "sidekiq_unique_jobs/web"
     require "sidekiq/cron/web"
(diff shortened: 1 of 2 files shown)


rhymes:
Hi @atsmith813, thanks for the work you've been doing!

It's such a pity we can't use `dev.to/en-us/content` because we allow dashes for organizations and podcasts as they both sit at the top of the path. 

I have a couple of ideas, that might have been already tried and ruled out so let me know:

- Keep `dev.to/en-us/` and use a regexp which matches among all possible combinations (we can compile this on app load if needed). We just need to make sure that there's no organization or podcast called `en-us` or the other combinations of language code plus country code and add a validation rule for those so they won't be in the future. This would allow us to keep a cleaner URL and matching a regexp in the Stories Controller is going to be fast anyway

- If that's not possible, has `dev.to/lang/en-us/rhymes/slug` been ruled out? Instead of `dev.to/lang-en-us/rhymes/slug` ? Just an aestethic preference though. There might be a SEO penalty for the extra path?

atsmith813:
We ultimately landed on `/lang-us-en` because it's more straight forward to implement and makes it easier to change in the future as we expand our i18n implementation.

Currently, the routes are a little tricky with dynamic top-level routes that we have to account for (usernames, organization names, and pages). By doing `/lang-us-en`:
- We don't have to worry (yet) about maintaining the list of supported languages and subregions to support the routes file.
- The regex is simpler. We match anything with `lang-`. It's slightly easier to use the region or not as we need it `lang-en` and `lang-en-us`.
- It makes the future validations we have to maintain more straightforward by blocking `lang-*` names.
- We want to continue to support 2 character names for users across Forems.

Once i18n has more traction, I could see us changing up the routes. For example, in order to get edge caching working the way we'd like it to, the edge needs to be aware of the languages that the currently supports. So at some point, we'll very likely have a list of what's supported making some other pieces easier. It's easier to make that change to routes entirely when they all live under `/lang-`. 

There are some SEO considerations as well (translating the URL, choosing the correct language path as the canonical URL, etc.). In terms of this first step, it's more about setting up the "sub-folder" approach to build on, with the finer SEO points coming into play later on. Given the nature of Forem, doing sub/domains is not an approach we want to take due to the scale and complexity of essentially infinite domains to manage. We've also come to learn that in terms of SEO, that seems to matter much less these days than other areas of translation.

Overall, what you've outlined is certainly possible. Since i18n is a large undertaking, we took a more nimble and flexible approach to start as we still have other large pieces to implement that we'll learn from once we get in the weeds.

cc: @benhalpern

PR #10138, opened 2020-09-01
Remove draft articles with a duplicate canonical_url

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

To be able to close the loop on https://github.com/forem/forem/pull/8072 and add a unique index on `canonical_url` we need first to be remove duplicates that have the same `canonical_url`. 

This PR contains a scripts that does two things:

- removes all drafts that in a partition have the same body. A partition is determined as a group of articles with the same canonical url
- removes all drafts that in a partition have different bodies by keeping the most recent one. A partition is determined as a group of articles with the same canonical url

Please read the SQL query closely.

If you want to take a look at the data:

- https://dev.to/admin/blazer/queries/212-duplicate-articles-canonical_url-with-the-same-body
- https://dev.to/admin/blazer/queries/214-duplicate-articles-canonical_url-with-different-bodies

https://github.com/forem/forem/pull/8072

Ideally this should be tested on a copy of the production DB. To make sure it worked I downloaded the sample of duplicates from https://dev.to/admin/blazer/queries/216-duplicate-articles-canonical_url-with-the-same-body-ids - imported them locally and ran the scripts to make sure they worked. They obviously present different results as it's a script run on the subset of duplications (so it works because in some cases we have 2 or more duplicates of the same article).

If you want to test it nonetheless:

- download a CSV from https://dev.to/admin/blazer/queries/216-duplicate-articles-canonical_url-with-the-same-body-ids
- open the SQL console with `psql PracticalDeveloper_development`
- type `\d articles`
- note the names of the foreign key constraints
- drop the foreign key constraints on the `articles` table, otherwise the `user_id` and `organization_id` won't work:

```sql
alter table articles drop constraint fk_rails_3d31dad1cc;
alter table articles drop constraint fk_rails_7809a1a57d;
```

remember that the constraint names might be different than mine. Please double check with `\d articles` in the SQL console and replace them with yours
- import the data from the downloaded CSV:

```sql
copy articles(id,user_id,title,body_html,slug,created_at,updated_at,main_image,description,published,published_at,featured,processed_html,featured
 _number,social_image,body_markdown,organization_id,canonical_url,collection_id,show_comments,main_image_background_hex_color,receive_notifications,approved,published_from_feed,c
 omments_count,reactions_count,video,video_code,video_source_url,video_thumbnail_url,video_closed_caption_track_url,hotness_score,feed_source_url,second_user_id,third_user_id,las
 t_buffered,positive_reactions_count,edited_at,crossposted_at,spaminess_rating,language,facebook_last_buffered,cached_tag_list,path,cached_user_name,cached_user_username,last_com
 ment_at,boost_states,video_state,page_views_count,previous_positive_reactions_count,score,reading_time,comment_template,video_duration_in_seconds,experience_level_rating,experie
 nce_level_rating_distribution,last_experience_level_rating_at,rating_votes_count,originally_published_at,organic_page_views_count,organic_page_views_past_month_count,organic_pag
 e_views_past_week_count,cached_user,cached_organization,archived,any_comments_hidden,nth_published_by_author,comment_score,search_optimized_title_preamble,search_optimized_descr
 iption_replacement,public_reactions_count,previous_public_reactions_count,user_subscriptions_count) from '/path/to/file.csv.csv' DELIMITER ',' csv header;
```

don't forget to replace the path of the file with the full path of the file you downloaded

- count the number of articles, just to have a quick idea if the script has worked
- run the data update script with `rails data_updates:run` if it's the first time or manually in the console

```ruby
require Rails.root.join("lib/data_update_scripts/20200901085230_remove_draft_articles_with_duplicate_canonical_url.rb")
DataUpdateScripts::RemoveDraftArticlesWithDuplicateCanonicalUrl.new.run
```

- re-count the number of articles, it should be less ;-)
- run https://dev.to/admin/blazer/queries/212-duplicate-articles-canonical_url-with-the-same-body locally to make sure they are gone

## Added tests?

- [x] yes
- [ ] no, because I need help


citizen428 (approved):
Very nice! 

I see you removed the `digest` call from the query now. 😃  The one question I already had yesterday when we talked about this in Slack: can there be no duplicates with a lag > 1, i.e. will the duplicates always have strictly increasing ids and not be of the form:

* id: duplicate 1
* id + 1: unrelated articled
* id +2: duplicate 2

rhymes:
@mstruve could you triple check if there are any leftovers with the following generic query?

```sql
select canonical_url, count(*)
from articles
where canonical_url is not null
group by canonical_url
having count(*) > 1
order by count(*) desc
```

Window functions are indeed magical, these are nice resources: https://tapoueh.org/blog/2013/08/understanding-window-functions/ and https://mjk.space/advanced-sql-window-functions/

@citizen428 yeah I removed the digest as you were right, it's noticeably faster in a large DB and also, that query was just for my "manual checking" needs :D I see your question, it's probably not impossible I guess, that's why I want to triple check with the group by to see if there's any fallacy in the logic. Although most of these posts are double posts or have somehow created twice by the RSS reader. That's still a mistery but with the unique index I'm confident we'll get to the bottom of this

mstruve:
@rhymes triple checked, there are none left 😄

rhymes:
@mstruve omg I'm the happiest. Is it fine if we send all the deleted ids to DD like that? My plan is to record them and if we find out we deleted an article that needs to be restored at least we can get the row out of a backup. Those `.event()` entries in DD don't have a short expiration date, correct?

mstruve:
13 month retention on Datadog events so I think we should be good. 
Screen Shot 2020-09-02 at 2 41 49 PM

Also, I have never sent events to Datadog before from the code, I like that idea!

rhymes:
@mstruve ok perfect. Let's admin merge this and then merge https://github.com/forem/forem/pull/10161 when you're back online.

mstruve:
I just got around to looking at this PR from my notifications, I will plan on merging this tomorrow(Friday) as soon as I am online after riding, and followup with the db one immediately after.

PR #10139, opened 2020-09-01
Remove ghostify from app

Zhao-Andy:
- [x] Removal

## Description
This removes the ghostify option from the app. We really don't use this feature much, and it's more confusing to use and keep around than to make it worthwhile.

https://github.com/forem/InternalProjectPlanning/issues/31

1. Search for `ghostify` or `ghost` or `Ghostify` in the app

## Added tests?

## Added to documentation?
- [x] no documentation needed

No

Casper the ghost walks slowly and sadly...

### DIFF
--- a/app/services/moderator/delete_user.rb
+++ b/app/services/moderator/delete_user.rb
@@ -1,47 +1,15 @@
 module Moderator
   class DeleteUser < ManageActivityAndRoles
-    attr_reader :user, :admin, :user_params
+    attr_reader :user, :user_params
 
-    def self.call(admin:, user:, user_params:)
-      if user_params[:ghostify] == "true"
-        new(user: user, admin: admin, user_params: user_params).ghostify
-      else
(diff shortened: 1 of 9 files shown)


rhymes (approved):
Looks great @Zhao-Andy!

I think it's going to fail because of https://github.com/forem/forem/blob/d8178f1c07dc1d76256250bd88dc4ab2c551734d/spec/requests/user/user_settings_spec.rb#L59 and related lines

app/services/moderator/delete_user.rb
```diff
@@ -1,47 +1,15 @@
 module Moderator
   class DeleteUser < ManageActivityAndRoles
-    attr_reader :user, :admin, :user_params
+    attr_reader :user, :user_params
```
Zhao-Andy, line 3: So we don't need `admin` here anymore as a variable. I think this is fine since we don't add a `note` whenever to delete an account, unlike the other `Moderator` module actions.

PR #10142, opened 2020-09-01
Index Reaction created_at Timestamp

mstruve:
- [x] Bug Fix Part 1

## Description
We store reactions in Elasticsearch and the default way we sort them is by ID desc which will give us the most recent reactions first. HOWEVER, this is not quite the case. IDs are stored in Elasticsearch as strings which means when you are doing a value comparison you end up with weird behavior such as 99 > 100. The means Reactions are not being sorted correctly all the time. This PR introduces `created_at` to Elasticsearch for reactions. Once this is merged and the data is backfilled I will then issue a PR to default sort by `created_at` 

https://github.com/orgs/forem/projects/6#card-44644200

- pull down branch
- run bin/setup
- open a console
- `Reaction.readinglist.last.elasticsearch_doc` should return a document with `created_at` in it

## Added tests?
- [x] updated

alt_text

### DIFF
--- a/app/serializers/search/reaction_serializer.rb
+++ b/app/serializers/search/reaction_serializer.rb
@@ -1,6 +1,6 @@
 module Search
   class ReactionSerializer < ApplicationSerializer
-    attributes :id, :category, :status, :user_id
+    attributes :id, :created_at, :category, :status, :user_id
 
     attribute :reactable do |reaction|
       reactable = reaction.reactable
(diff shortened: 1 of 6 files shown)


rhymes:
> IDs are stored in Elasticsearch as strings which means when you are doing a value comparison you end up with weird behavior such as 99 > 100.

:-(

rhymes (approved):

citizen428 (approved):

PR #10136, opened 2020-09-01
🚀 Feature: Draw Command for connect with image annotations 

sarthology:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
While making the drag and drop feature for image upload, I thought about this idea and started working on it. I believe this will empower teams using connect like for mods this can be extremely useful. Apart from that, it's a fun command for Connect. 😄

## QA Instructions, Screenshots, Recordings

ezgif com-gif-maker-2

## Added tests?

- [ ] yes
- [ ] no, because they aren't needed
- [x] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

## [optional] What gif best describes this PR or how it makes you feel?

alt_text

PR #10145, opened 2020-09-01
Drag and drop an image in the editor

nickytonline:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds drag and drop image functionality to the v2 Editor. You can drop an image in the cover image section and it will upload. You can also drop an image in the content body of the post. This feature works exactly like image drag and drop on GitHub.

#9992, #10203 

### To test

Ensure #9992 works still

#### Drag/drop a cover image

cover-image-drop

1. Drag a valid image file that we support and drop it on to the cover image area of the editor.
1. Once the file is dropped, it will behave like a file selection upload.
	1. The spinner will appear
	1. The image will upload successfully and a cover image will appear

#### Drag/drop a post body image

body-image-drop

1. Drag a valid image file that we support and drop it on to the post body area of the editor.
1. The post body editable area will change to an opacity of 25%
1. The file uploads and inserts the markdown for the uploaded file into the last known cursor position of the post body editable area.
1. Opacity returns to normal.

#### Errors

- If more than one file is dropped on either of the droppable regions mentioned above, a snackbar item will appear notifying the user that they can only drop one file.

image

- If an invalid file is dropped on either of the droppable regions mentioned above, a snackbar item will appear notifying the user that the file is an unsupported image type.

image

Also, the editor help has been updated to take image drag and drop into account.

image

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Jim from the Office looking at the camera and saying "Yes!"


nickytonline:
I still need to write some tests @ludwiczakpawel and @lisasy, but I was wondering about styling for dragging over and dropping images. Currently I add/remove the CSS utility class `opactiy-25` like it does in the chat image drag and drop.

benhalpern:
Hell yeah

app/javascript/utilities/dragAndDrop.js
```diff
@@ -0,0 +1,91 @@
...
+ */
+export function DragAndDropZone({ children, onDragOver, onDragExit, onDrop }) {
+  const { setElement } = useDragAndDrop({ onDragOver, onDragExit, onDrop });
+  const dropZoneRef = useRef(null);
+
+  if (dropZoneRef.current) {
+    setElement(dropZoneRef.current);
+  }
+
+  return cloneElement(children, {
```
nickytonline, line 88: You can't add props as they are read-only, but you can clone an element and add additional props that way. In our case, we add a ref prop so that we can access the related DOM element to add and remove drag events.

app/javascript/article-form/components/ArticleCoverImage.jsx
```diff
@@ -17,14 +19,16 @@ export class ArticleCoverImage extends Component {
...
+  handleMainImageUpload = (event) => {
+    event.preventDefault();
 
     this.setState({ uploadingImage: true });
     this.clearUploadError();
 
     if (validateFileInputs()) {
-      const payload = { image: e.target.files, wrap_cloudinary: true };
+      const { files: image } =
+        event instanceof DragEvent ? event.dataTransfer : event.target;
```
nickytonline, line 30: Need to see what type of event it is so we can grab the appropriate `files` property.

ludwiczakpawel:
@nickytonline this is awesome <3. Don't worry about the `opacity-25` - I have an idea for some styling, but that can easily be a separate PR and i'll be more than happy to jump into it.

ludwiczakpawel (approved):

ludwiczakpawel:
whoops, I've approved that premature..

rhymes:
@nickytonline is this based off https://github.com/forem/forem/pull/10136 ?

nickytonline:
> @nickytonline is this based off https://github.com/forem/forem/pull/10136 ?

@rhymes, no. It's based off of master which contained the initial drag and drop functionality for Connect.

app/javascript/utilities/dragAndDrop.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {JSX.Element} props.children The React element that will register it's DOM counterpart with drag and drop events.
+ * @param {Function} props.onDragOver The handler that runs when the dragover event is fired.
+ * @param {Function} props.onDragExit The handler that runs when the dragexit/dragleave events are fired.
+ * @param {Function} props.onDrop The handler that runs when the drop event is fired.
+ */
+export function DragAndDropZone({ children, onDragOver, onDragExit, onDrop }) {
+  const { setElement } = useDragAndDrop({ onDragOver, onDragExit, onDrop });
+  const dropZoneRef = useRef(null);
+
+  if (dropZoneRef.current) {
```
citizen428, line 84: This may be a silly question, but since we're initializing the reference to `null` just above this line, can the condition be true? I guess I expected some closure that gets attached to something, but maybe I'm just confused 😃

nickytonline: When the initial state is set to `null`, it set `dropZoneRef.current` to `null`. `dropZoneRef` is always defined and on subsequent rerender, the ref gets its value set to the reference to the DOM node,i.e. `dropZoneRef.current`. For more info see the ref documentation, https://reactjs.org/docs/hooks-reference.html#useref .

citizen428: I read the documentation before asking the question but in the example, they capture the ref in a closure which gets attached as an event handler in the returned JSX, which makes more intuitive sense. 

But I think I get it now, the binding of ref here happens "outside" this component via `useDragAndDrop`. So the two lines in the fragment in the doc are basically split up here.

msarit:
Works like a charm @nickytonline 💎 

This may be a Product question, but would it make sense to include information about the ability to drag-drop images?

Screen Shot 2020-09-03 at 11 29 16 AM

nickytonline:
> Works like a charm @nickytonline 💎
> 
> This may be a Product question, but would it make sense to include information about the ability to drag-drop images?

Definitely. Also, product design will be working on drag and drop look and feel in another PR as mentioned by @ludwiczakpawel above.

app/javascript/article-form/components/Help.jsx
```diff
@@ -235,6 +235,13 @@ export class Help extends Component {
             </a>{' '}
             to add rich content such as Tweets, YouTube videos, etc.
           </li>
+          <li>
```
nickytonline, line 238: Not sure about the wording for this. Any improvements welcome.

app/javascript/article-form/components/dragAndDropHelpers.js
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/**
+ * DragExit handler for the editor
+ *
+ * @param {DragEvent} event the drag event.
+ */
+export function onDragExit(event) {
+  event.preventDefault();
+  event.currentTarget.classList.remove('opacity-25');
```
nickytonline, line 59: Note that as mentioned by @ludwiczakpawel, styling updates are coming in a subsequent PR, #10203.

app/javascript/article-form/components/ArticleCoverImage.jsx
```diff
@@ -53,55 +58,75 @@ export class ArticleCoverImage extends Component {
...
+        message: 'Only one image can be dropped at a time.',
+        addCloseButton: true,
+      });
+      return;
+    }
+
+    this.handleMainImageUpload(event);
+  };
+
   render() {
```
Link2Twenty, line 75: Are we worried that the render now goes over the 50 lines limit?

nickytonline: Not too much honestly. The line bust is because of prettier formatting. It could be broken up a bit though. I'll take a peek.

nickytonline: I took a peek and there some other refactorings that could go on in there, but I would probably leave it for a subsequent PR.

app/javascript/article-form/components/ArticleCoverImage.jsx
```diff
@@ -53,55 +58,75 @@ export class ArticleCoverImage extends Component {
     });
   };
 
+  onDropImage = (event) => {
+    onDragExit(event);
+
+    if (event.dataTransfer.files.length > 1) {
```
rhymes, line 64: is there a reason for this limitation? I'm referring to Github's editor which allows multiple ones. Non blocking obvs, but I was wondering why we can't allow more than 1 and use promises to upload them concurrently. We could set an arbitrary max. 

I just uploaded 25 in thix box and it worked. The UX isn't great as it attaches the markdown in the position of the cursor which means that if you move the cursor during update it'll create invalid tags mashed with each other.

app/javascript/article-form/components/ArticleCoverImage.jsx
```diff
@@ -53,55 +58,75 @@ export class ArticleCoverImage extends Component {
...
-                  data-max-file-size-mb="25"
-                />
-              </Button>
-              {mainImage && (
-                <Button
-                  variant="ghost-danger"
-                  onClick={this.triggerMainImageRemoval}
-                >
-                  Remove
+      <DragAndDropZone
```
rhymes, line 81: can't wait to have this in the comments as well :D

app/javascript/article-form/components/ArticleCoverImage.jsx
```diff
@@ -53,55 +58,75 @@ export class ArticleCoverImage extends Component {
...
+              className="crayons-article-form__cover__image"
+              width="250"
+              height="105"
+              alt="Post cover"
+            />
+          )}
+          <div className="flex items-center">
+            {uploadingImage ? (
+              <span class="lh-base pl-1 border-0 py-2 inline-block">
+                <Spinner /> Uploading...
```
rhymes, line 99: It'd be cool if we could see something like this for the article images as well but I read that a UX refactoring PR is in the pipeline

app/javascript/article-form/components/EditorBody.jsx
```diff
@@ -1,14 +1,61 @@
...
+import { useDragAndDrop } from '@utilities/dragAndDrop';
+
+function handleImageSuccess(textAreaRef) {
+  return function (response) {
+    // Function is within the component to be able to access
+    // textarea ref.
+    const editableBodyElement = textAreaRef.current.base;
+    const { links, image } = response;
+
+    const markdownImageLink = `![${image[0].name}](${links[0]})`;
```
rhymes, line 21: can we remove the extension from the `alt` text? So "screenshot" instead of "screenshot.png"

app/javascript/article-form/components/EditorBody.jsx
```diff
@@ -1,14 +1,61 @@
...
+    const { selectionStart, selectionEnd, value } = editableBodyElement;
+    const before = value.substring(0, selectionStart);
+    const after = value.substring(selectionEnd, value.length);
+
+    editableBodyElement.value = `${before + markdownImageLink} ${after}`;
```
rhymes, line 26: as we don't really support inline images, should the insertion have a "newline" as well?

I tried to upload two images and the tags were kept on the same line

app/javascript/article-form/components/Help.jsx
rhymes: "In addition to images for the post's content, you can also drag and drop a cover image"

Maybe something short like this?

app/javascript/packs/articleForm.jsx
```diff
@@ -11,6 +12,13 @@ HTMLDocument.prototype.ready = new Promise((resolve) => {
 });
 
 function loadForm() {
+  // The Snackbar for the article page
+  const snackZone = document.getElementById('snack-zone');
```
rhymes, line 16: snack zone 🤣  best place to be in the DOM

rhymes (commented):
Looks cool! Left some suggestions :)

app/javascript/article-form/components/EditorBody.jsx
nickytonline: I'm fine with adding a new line. Coming up!

app/javascript/article-form/components/EditorBody.jsx
nickytonline: Done.

app/javascript/article-form/components/ArticleCoverImage.jsx
nickytonline: Yeah exactly. For the cover image I basically follow the same flow as an image upload via the native file selector. For the post content it is a little different. @ludwiczakpawel is addressing it in #10203.

app/javascript/article-form/components/ArticleCoverImage.jsx
nickytonline: From what I can tell right now, our backend is only equipped to receive one image in the POST request. We can definitely allow for multiple though. I'll revisit this in the cooldown period.

Link2Twenty:
Next we should look at pasting an image in directly 😉.

In GitHub you can have an image in your clipboard and when you paste you get a nice little indicator and it does it for you.
I'm a stickler for sending images this way 😁

image

rhymes (approved):
Works great!

PR #10147, opened 2020-09-01
Optimization:Select Article Columns for Digest Email, Limit 6, Use User Attributes

mstruve:
- [x] Optimization

## Description
Few improvements here to the Digest Mailer
- rather than `limit(8)` then only taking the `first(6)` in the mailer, lets just make it `limit(6)`
- Articles are big models with a lot of data, we only need title, description, and path in order to populate the body of the email so limit the columns to those. 
- Replace Redis lookup `@user.cached_following_users_ids` with reading a user attribute `@user.following_users_count`
- Replace Postgres db email hits with user attributes: `@user.follows.size` with `@user.following_users_count` and `@user.articles.published` with `@user.articles_count`. Now `follows.size` includes tags but the messaging in the email is for following users so I figured the replacement made sense using the users count.
- Refactor: replaced `published: true` with our scope. 

https://github.com/orgs/forem/projects/6#card-44508249

## Added tests?

alt_text

### DIFF
--- a/app/views/mailers/digest_mailer/digest_email.html.erb
+++ b/app/views/mailers/digest_mailer/digest_email.html.erb
@@ -27,7 +27,7 @@
       Visit <b><a href="<%= app_url(user_settings_path(:misc)) %>" style="text-decoration: none;">your settings</a></b> to enter your experience level on a scale of 1-10.
       <br /><br />
       You can change it any time in the future. Happy <%= SiteConfig.community_action %>!
-    <% elsif @user.follows.size == 0 %>
+    <% elsif @user.following_users_count == 0 %>
       <b>
         <%= community_name %> Digest is a new periodic email featuring the best posts from our community of <%= community_members_label %>.
       </b>
@@ -45,7 +45,7 @@
       <% elsif tip_rand == 3 %>
         You're a better <%= SiteConfig.community_member_label %> today than you were yesterday.
       <% elsif tip_rand == 4 %>
-        <% if @user.articles.published.any? %>
+        <% if @user.articles_count > 0 %>
           Can't wait to see <b>your</b> next <%= community_name %> post!
         <% else %>
           Can't wait to see <b>your</b> first <%= community_name %> post!
--- a/spec/workers/emails/send_user_digest_worker_spec.rb
+++ b/spec/workers/emails/send_user_digest_worker_spec.rb
@@ -20,29 +20,29 @@ RSpec.describe Emails::SendUserDigestWorker, type: :worker do
       before { user.follow(author) }
 
       it "send digest email when there are at least 3 hot articles" do
-        articles = create_list(:article, 3, user_id: author.id, public_reactions_count: 20, score: 20)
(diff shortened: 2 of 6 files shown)


spec/workers/emails/send_user_digest_worker_spec.rb
```diff
@@ -20,29 +20,29 @@
       before { user.follow(author) }
 
       it "send digest email when there are at least 3 hot articles" do
-        articles = create_list(:article, 3, user_id: author.id, public_reactions_count: 20, score: 20)
+        create_list(:article, 3, user_id: author.id, public_reactions_count: 20, score: 20)
 
         worker.perform(user.id)
 
-        expect(DigestMailer).to have_received(:with).with(user: user, articles: articles)
+        expect(DigestMailer).to have_received(:with).with(user: user, articles: Array)
```
mstruve, line 27: This was giving me serious issues after adding the `select` statement in. I even regathered the articles here with the select statement for this assertion and despite it looking the exact same in the debugger, Rspec was still reporting them differently. I am sure it has something to do with the AR objects not being the same. Rather than fight it, after about an hour of playing around, I settled for a general assertion here with the array.

app/views/mailers/digest_mailer/digest_email.html.erb
```diff
@@ -45,7 +45,7 @@
       <% elsif tip_rand == 3 %>
         You're a better <%= SiteConfig.community_member_label %> today than you were yesterday.
       <% elsif tip_rand == 4 %>
-        <% if @user.articles.published.any? %>
+        <% if @user.articles_count > 0 %>
```
rhymes, line 48: Does this only include published articles? Because I don't see anything special in the `Article` model to suggest the drafts are not included in this count. Is that okay?

spec/workers/emails/send_user_digest_worker_spec.rb
rhymes: It migth yeah, don't know. Let's keep it like that :)

rhymes (commented):
Nice!

app/views/mailers/digest_mailer/digest_email.html.erb
mstruve: This includes drafts. I figured based on the wording and that this is only chosen 1/5 of the time including drafts in this count is not a huge deal. It's also the very bottom of the email so I don't think it is consumed very much and the difference between first and next is not much in my mind.

rhymes: Fine by me! Let's see if Vaidehi comes up with a compelling reason not to use it. Otherwise we'll have to add a new column for the published articles and point counter culture to it 👎

rhymes (approved):

app/views/mailers/digest_mailer/digest_email.html.erb
vaidehijoshi: i think it's probably okay to do this; i would vote that, if anything, we change the wording rather than add a new column for this. but imho, i think the wording here works 👍

PR #10150, opened 2020-09-02
Add ga tracking to config

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We want admins to be able to set a Google Analytics tracking code so it's not an ENV var.

### DIFF
--- a/app/assets/javascripts/initializers/initializeBaseTracking.js.erb
+++ b/app/assets/javascripts/initializers/initializeBaseTracking.js.erb
@@ -12,7 +12,6 @@ function initializeBaseTracking() {
     wait++;
     if (window.ga && ga.create) {
       ga('create', document.body.dataset.gaTracking, 'auto');
-      console.log(document.body.dataset.gaTracking)
       ga('set', 'anonymizeIp', true);
       ga('send', 'pageview', location.pathname + location.search);
       clearInterval(waitingOnGA);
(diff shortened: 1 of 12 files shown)


app/lib/constants/site_config.rb
```diff
@@ -113,8 +113,8 @@ module SiteConfig
         description: "The \"Client Secret\" portion of the GitHub Oauth Apps portal",
         placeholder: ""
       },
-      ga_view_id: {
```
mstruve, line 116: Was this just unused?

benhalpern: Yeah, this was part of the functionality for using the Google Analytics API for tallying total page views, which we no longer use. So this is just dead functionality.

app/assets/javascripts/initializers/initializeBaseTracking.js.erb
```diff
@@ -11,7 +11,7 @@ function initializeBaseTracking() {
     addedGA = true;
     wait++;
     if (window.ga && ga.create) {
-      ga('create', '<%= ApplicationConfig["GA_TRACKING_ID"] %>', 'auto');
+      ga('create', document.body.dataset.gaTracking, 'auto');
```
benhalpern, line 14:
```suggestion
      ga('create', '<%= ApplicationConfig["GA_TRACKING_ID"] %>', 'auto');
```

benhalpern:
I modified this to actually _not yet_ use this value in the document, because there _could_ be a scenario where we roll this out and it's not yet available in the view but the functionality is changed in the JS....

So I'm going to do the change in JS as an additional step 24 hours after merging this (aka once all caches would have cleared)

PR #10152, opened 2020-09-02
Attempt to fix facebook login

joshpuetz:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently in production (Forem.dev and MMALife) Facebook login fails with the error: `Could not authenticate you from Facebook because "Invalid credentials".` Based on guidance from https://github.com/heartcombo/devise/wiki/OmniAuth:-Overview, this PR adds a parsing option to the OmniAuth-Facebook setup in an attempt to fix this error.

(Of course this error doesn't occur in development...grr)

https://github.com/forem/forem/pull/9922

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -9,11 +9,6 @@ GITHUB_OMNIUATH_SETUP = lambda do |env|
   env["omniauth.strategy"].options[:scope] = "user:email"
 end
 
-FACEBOOK_OMNIAUTH_SETUP = lambda do |env|
-  env["omniauth.strategy"].options[:client_id] = SiteConfig.facebook_key
-  env["omniauth.strategy"].options[:client_secret] = SiteConfig.facebook_secret
-end
-
 Devise.setup do |config|
   # The secret key used by Devise. Devise uses this key to generate
   # random tokens. Changing this key will render invalid all existing
@@ -299,9 +294,9 @@ Devise.setup do |config|
   # up on your models and hooks.
 
   # Fun fact, if this is reordered to have Twitter first, it doesn't work for some reason.
-  config.omniauth :facebook, setup: FACEBOOK_OMNIAUTH_SETUP
   config.omniauth :github, setup: GITHUB_OMNIUATH_SETUP
   config.omniauth :twitter, setup: TWITTER_OMNIAUTH_SETUP
+  config.omniauth :facebook, SiteConfig.facebook_key, SiteConfig.facebook_secret, token_params: { parse: :json }
 
   # ==> Warden configuration
   # If you want to use other strategies, that are not supported by Devise, or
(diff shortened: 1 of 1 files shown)


rhymes (commented):
Let's see if this works!

There's only a problem: `SiteConfig` is not available in the initializers, so you'll have to replace this with the environment variable, at least until we figure out if the fix works and then explore a lasting solution that takes into account the ability for admin to change keys without restarting the server.

As we don't declare `FACEBOOK_KEY` and `FACEBOOK_SECRET` anywhere you can just access `ENV["FACEBOOK_KEY"]` and `ENV["FACEBOOK_SECRET"]` and then notify the admins of Forem.dev and ThisMMALife to set them in their server to test

config/initializers/devise.rb
```diff
@@ -9,11 +9,6 @@
   env["omniauth.strategy"].options[:scope] = "user:email"
 end
 
-FACEBOOK_OMNIAUTH_SETUP = lambda do |env|
-  env["omniauth.strategy"].options[:client_id] = SiteConfig.facebook_key
-  env["omniauth.strategy"].options[:client_secret] = SiteConfig.facebook_secret
```
fdoxyz, line 14: Adding the config this way worked for me locally:

```
env["omniauth.strategy"].options[:token_params][:parse] = 'json'
```

Although I'm not 100% sure if it makes a difference since it worked locally for me with or without it

joshpuetz: How did you verify that the syntax for the parse setting is correct? I couldn't find a way to inspect that setting programmatically during runtime, which made me a bit nervous!

joshpuetz: Ok: playing around and setting token_parse to `text` results in an error (in fact the one in product), so I'm confident this syntax is working!

config/initializers/devise.rb
```diff
@@ -298,7 +299,7 @@
   # Add a new OmniAuth provider. Check the wiki for more information on setting
   # up on your models and hooks.
 
-  # Fun fact, if this is reordered to have Twitter first, it doesn't work for some reason.
+  # Fun fact, unless Twitter is last, it doesn't work for some reason.
```
joshpuetz, line 302: TIL!

config/initializers/devise.rb
fdoxyz: > Ok: playing around and setting token_parse to text results in an error (in fact the one in product), so I'm confident this syntax is working!

Oh nice smart move to test it that way! I hadn't verified tbh, it was an eyeball config from my end but good to hear it's being set correctly then :)

rhymes (approved):
LGTM!

PR #10153, opened 2020-09-02
Extracts DisplayAds from Administrate into admin

fdoxyz:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Extracts Display Ads dashboard from `/resource_admin` (Administrate) into `/admin` dashboard

CRUD capabilities to manage Display Ads should be available in `/admin/display_ads`

Screen Shot 2020-09-01 at 23 15 57

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

N/A

dog swimming along

### DIFF
--- a/app/controllers/admin/display_ads_controller.rb
+++ b/app/controllers/admin/display_ads_controller.rb
@@ -1,0 +1,71 @@
+module Admin
+  class DisplayAdsController < Admin::ApplicationController
+    layout "admin"
+
+    def index
+      @display_ads = DisplayAd.order(id: :desc)
+        .select("display_ads.*")
+        .joins(:organization)
+        .includes([:organization])
+        .page(params[:page]).per(50)
+
+      return if params[:search].blank?
+
+      @display_ads = @display_ads.where("organizations.name ILIKE :search", search: "%#{params[:search]}%")
+    end
+
+    def new
+      @display_ad = DisplayAd.new
+    end
+
+    def edit
+      @display_ad = DisplayAd.find(params[:id])
+    end
+
+    def create
+      @display_ad = DisplayAd.new(display_ad_params)
+
+      if @display_ad.save
+        flash[:success] = "Display Ad has been created!"
+        redirect_to admin_display_ads_path
+      else
+        flash[:danger] = @display_ad.errors.full_messages.to_sentence
+        render new_admin_display_ad_path
+      end
+    end
+
+    def update
+      @display_ad = DisplayAd.find(params[:id])
+
+      if @display_ad.update(display_ad_params)
+        flash[:success] = "Display Ad has been updated!"
+        redirect_to admin_display_ads_path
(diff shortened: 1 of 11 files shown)


app/lib/constants/role.rb
```diff
@@ -16,6 +16,7 @@ module Role
                      "Resource Admin: Page",
                      "Resource Admin: FeedbackMessage",
                      "Resource Admin: Config",
-                     "Resource Admin: Broadcast"].freeze
+                     "Resource Admin: Broadcast",
+                     "Resource Admin: DisplayAd"].freeze
```
fdoxyz, line 20: I'm not 100% sure this had to be added, I think it's helpful but open to take it off if anyone feels differently :)

fdoxyz:
@ludwiczakpawel since you were already pushing for Crayons in admin I sprinkled a bit of it in here, in case you wanted to take a look

ludwiczakpawel:
This is awesome!

app/controllers/admin/display_ads_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    def create
+      @display_ad = DisplayAd.new(display_ad_params)
+
+      if @display_ad.save
+        flash[:success] = "Display Ad has been created!"
+        redirect_to admin_display_ads_path
+      else
+        flash[:danger] = @display_ad.errors.full_messages.to_sentence
```
rhymes, line 32: you can use `@display_ad.errors_as_sentence` which is a shortcute we have

rhymes (approved):
Nice! Works great!

app/views/admin/display_ads/_form.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  <%= text_field_tag :organization_id, @display_ad.organization_id, class: "form-control" %>
+</div>
+
+<div class="form-group">
+  <%= label_tag :body_markdown, "Body Markdown:" %>
+  <%= text_area_tag :body_markdown, @display_ad.body_markdown, size: "100x10", class: "form-control" %>
+</div>
+
+<div class="form-group">
+  <%= label_tag :placement_area, "Placement area:" %>
```
juliannatetreault, line 12: _Quick, non-blocking suggestion_: I think this would look nice titlecased since all of the other labels are. 😄 
```suggestion
  <%= label_tag :placement_area, "Placement Area:" %>
```

app/lib/constants/role.rb
juliannatetreault: I'm curious about whether this needs to be added or not, too! 😄

ludwiczakpawel (approved):

PR #10155, opened 2020-09-02
Enable CSP in report only mode and use Honeybadger to track violations

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

#10144 added a CSP but the website is too big to be able to encompass all cases ahead of time. I've seen violations related to Pusher and to inline images in the chat for example. Also, the Blazer admin page index is not working for me.

For this reason I propose we use a different approach: enable CSP in report only mode.

This will help us track violations without blocking content, we keep an eye on the rules that need fixing/expanding and then when we have close to no violations we can switch off the reporting mode and keep the report URI for whatever we might have missed.

The alternative is to turn off report only mode but keep the report URI and then keep changing the CSP in the following days until we get close to zero.

We just can't rely on our own browser's console.

I'm using Honeybadger for the report URI 

@mstruve: the HB documentation suggests to create a new app just for the reports as they are going to be numerous. We can sync on that and I can amend the PR when we'll have the new key, unless you think it's fine to keep them in them in the global HB app. See https://docs.honeybadger.io/lib/ruby/integration-guides/rails-exception-tracking.html#content-security-policy-reports

#10144

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -5,24 +5,23 @@
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
 
 Rails.application.config.content_security_policy do |policy|
-  if Rails.env.production?
-    policy.default_src :self, :https
-    policy.font_src    :self, :https, :data
-    policy.img_src     :self, :https, :data
-    policy.object_src  :none
-    policy.script_src  :self, :https, :unsafe_inline
-    policy.style_src   :self, :https, :unsafe_inline
+  policy.default_src :self, :https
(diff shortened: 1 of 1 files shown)


config/initializers/content_security_policy.rb
```diff
@@ -5,24 +5,23 @@
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
 
 Rails.application.config.content_security_policy do |policy|
-  if Rails.env.production?
```
rhymes, line 8: no need for this

PR #10161, opened 2020-09-02
Add unique index on articles canonical_url

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Adding the unique index on articles `canonical_url`.

This is in conflict with https://github.com/forem/forem/pull/9788

Thus we need to decide which of these two options:

1. `canonical_url` has to be unique across all articles (drafts and published)
2. `canonical_url` has to be unique only within the scope of published articles (thus this index needs to add `published` in its column list)

I'm not 100% sure what we want here so I though we'd start with this drarft PR

ps. this should be deployed right after https://github.com/forem/forem/pull/10138 - possibly closely after to avoid further duplicates ;)

cc @mstruve @citizen428 due to, respectively, https://github.com/forem/forem/pull/9788#pullrequestreview-472925409 and https://github.com/forem/forem/pull/9788#discussion_r480880463

- https://github.com/forem/forem/pull/8072
- https://github.com/forem/forem/pull/10138

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -51,8 +51,8 @@ class Article < ApplicationRecord
 
   validates :body_markdown, length: { minimum: 0, allow_nil: false }, uniqueness: { scope: %i[user_id title] }
   validates :cached_tag_list, length: { maximum: 126 }
-  validates :canonical_url, url: { allow_blank: true, no_local: true,
-                                   schemes: %w[https http] }, uniqueness: { allow_blank: true }
+  validates :canonical_url, uniqueness: { allow_nil: true }
+  validates :canonical_url, url: { allow_blank: true, no_local: true, schemes: %w[https http] }
   validates :feed_source_url, uniqueness: { allow_blank: true }
   validates :main_image, url: { allow_blank: true, schemes: %w[https http] }
   validates :main_image_background_hex_color, format: /\A#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})\z/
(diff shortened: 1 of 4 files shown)


mstruve:
> 1. canonical_url has to be unique across all articles (drafts and published)

I vote #1 all the way! I think that is the most simple solution and eliminates the problem of someone thinking they have a valid unpublished article and then not being able to publish it.

citizen428:
I think 1. is preferable here, as it's more straight-forward.

app/models/article.rb
```diff
@@ -51,8 +51,8 @@ class Article < ApplicationRecord
 
   validates :body_markdown, length: { minimum: 0, allow_nil: false }, uniqueness: { scope: %i[user_id title] }
   validates :cached_tag_list, length: { maximum: 126 }
-  validates :canonical_url, url: { allow_blank: true, no_local: true,
-                                   schemes: %w[https http] }, uniqueness: { allow_blank: true }
+  validates :canonical_url, uniqueness: { allow_nil: true }
+  validates :canonical_url, url: { allow_blank: true, no_local: true, schemes: %w[https http] }
```
rhymes, line 55: split the validators in 2 lines and replaced `allow_blank` with `allow_nil` to avoid the empty string ;)

mstruve (approved):

rhymes:
Had to force push to fix the schema conflict

Zhao-Andy (approved):

mstruve:
Confirmed there are 0 duplicates in the database after the update script

PR #10167, opened 2020-09-02
Introduce cached_user on comments for memory allocation and overall speed gains.

benhalpern:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

[Here](https://blog.heroku.com/solving-n-plus-one-queries#manually-building-count-data-in-hashes) is a post not _entirely_ related to the contents of this PR, but thematically similar with this really nice big, bold header....
 
Screen Shot 2020-09-02 at 11 38 17 AM

This pull request alleviates the need for us to include `:user` when we pull in comments. I suspect this could have a big impact on how memory is allocated and grows within our application. Expensive active record queries are easily the top one or two things that causes memory bloat in an app.

I benchmarked to conclude pretty definitively that these are improvements in both memory and performance, though due to our good practices with eager loading, the bigger hypothetical improvements will be to memory allocation vs speed.

The trade off is stuff we store vs call on demand which intros its own complexity, but I think since we already do this via articles and this generally opens us up to more freedom when calling comments, it should be a good thing all around.

The north star this also opens up is a scenario which we use `pluck` as the underlying mechanism for fetching comments instead of pulling in entire active record objects. That modification doesn't quite seem worth it now, but could be something we move towards in the future.

The last time we modified the cached user functionality we ran into problems with the data updates, and I think this solution is better than what we had done.

Instead of looping through the comments, I loop through the users with at least one comment, and bulk update their comments, with no callbacks triggered. So with `find_each` that loops through the users, this should finish reasonably promptly.... We should definitely merge and be on stand by to fix the issues, but I feel good about the feasibility of this for DEV, and other communities will have no problems.

If the code passes muster I think this will be an overall improvement in app performance and scalability.


app/views/comments/_comment.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if comment && comment.user %>
+<% if comment && comment.cached_user.present? %>
```
benhalpern, line 1: Because we were already guarding against the absence of a user for any reason, we have this safety clause if any component of this change fails on deployment.

app/models/comment.rb
```diff
@@ -190,6 +193,12 @@ def calculate_score
     Comments::CalculateScoreWorker.perform_async(id)
   end
 
+  def update_cached_user
+    return unless user
+
+    self.cached_user = Articles::CachedEntity.from_object(user)
```
joshpuetz, line 199: `Articles::CachedEntity` seems like an odd name now that this is a cached User in use outside of Article!

mstruve:
> I benchmarked to conclude pretty definitively that these are improvements in both memory and performance,

Can you quantify what these improvements are?

I am very hesitant to add another cached entity like this bc the added complexity of keeping it in sync and now when users are updated we have to update all of their comments. While this may not seem like a big deal now, I have worked with relationships like this in the past and they become huge bottlenecks bc every single user update now has to trigger tons of comment updates.

benhalpern:
@mstruve 100% hear you on that.... Technically migrating _away_ from this would not be so difficult if we ever found it more trouble than it would be worth, though of course it's never _that_ easy...

Here was my benchmark.....

```ruby
a = Article.find_by_slug "a-glass-of-blessings-voluptatem-et-2pa9"
report_with_includes = MemoryProfiler.report do
  a.comments.includes(:user).each do |c|
    c.user.name
  end
end

report_with_cached = MemoryProfiler.report do
  a.comments.each do |c|
    c.cached_user.name
  end
end

report_with_cached_plucked = MemoryProfiler.report do
  a.comments.pluck(:cached_user).each do |c|
    c.name
  end
end

report_with_includes.pretty_print
Total allocated: 8198779 bytes (87157 objects)
Total retained:  1219926 bytes (6536 objects)

allocated memory by gem
-----------------------------------
   3964829  activemodel-6.0.3.2
   1995944  activerecord-6.0.3.2
   1063449  bootsnap-1.4.8
    515536  activesupport-6.0.3.2
    228760  binding_of_caller-0.8.0
     95470  store_attribute-0.7.1
     73632  ddtrace-0.39.0
     73608  better_errors-2.7.1
     65033  honeybadger-4.7.2
     34432  dev.to/app
     29122  json-2.3.0
     18840  delegate
     12016  amazing_print-1.2.1
      4368  set
      4336  jsonapi-serializer-2.0.0
      4024  ancestry-3.1.0
      3680  concurrent-ruby-1.1.7
      2756  logger
      2704  forwardable
      2304  bullet-6.1.0
      1464  counter_culture-2.6.0
      1272  rolify-5.3.0
       432  mutex_m
       408  zeitwerk-2.4.0
       240  weakref
        80  other
        40  after_commit_action-1.1.0


report_with_cached.pretty_print
Total allocated: 339334 bytes (3816 objects)
Total retained:  134323 bytes (1151 objects)

allocated memory by gem
-----------------------------------
    169114  psych
    103121  activerecord-6.0.3.2
     46238  activemodel-6.0.3.2
      4986  honeybadger-4.7.2
      4655  activesupport-6.0.3.2
      4408  other
      4384  ddtrace-0.39.0
      1560  delegate
       336  set
       212  logger
       208  forwardable
       168  bullet-6.1.0



report_with_cached_plucked.pretty_print
Total allocated: 213368 bytes (2672 objects)
Total retained:  1210 bytes (13 objects)

allocated memory by gem
-----------------------------------
    169114  psych
     14414  activemodel-6.0.3.2
     12921  activerecord-6.0.3.2
      4940  honeybadger-4.7.2
      4408  other
      2979  activesupport-6.0.3.2
      2276  ddtrace-0.39.0
      1560  delegate
       336  set
       212  logger
       208  forwardable
```


The north star we possibly enable in this cached user universe is that we could pluck the whole object as an array (possibly using `pluck_to_hash` gem for better interface) and then we dramatically reduce the footprint of our memory bloat.

....... Of course a big reason for the bloat is that we even use the `user` model in the first place here, as opposed to a stripped down table that's _only_ for bare bones public-facing attributes. And _maybe_ a multiple-table pluck may be the best of both worlds...... 



@citizen428 and @Ridhwana could weigh in here about how our user profile redesign could fit in with this conversation in terms of how we might reduce our memory footprint.

citizen428:
> @citizen428 and @Ridhwana could weigh in here about how our user profile redesign could fit in with this conversation in terms of how we might reduce our memory footprint.

For once, the `users` table will eventually lose 29 columns, that's a lot less data will get materialized for each user instance.

> Of course a big reason for the bloat is that we even use the user model in the first place here, as opposed to a stripped down table that's only for bare bones public-facing attributes. 

We should probably discuss this whole idea a bit more and come up with a proper code representation for this (i.e. a `PublicProfile` subclass that just exposes certain columns of a user's profile or something like that).

mstruve:
@jdoss is going to try out JEMALLOC https://github.com/forem/forem/pull/10195, I would like to see how far that gets us. If it works like it did on Heroku I think we could forego this optimization

mstruve (commented):
I slept on this overnight and I think because this is a little bit our pattern(with articles caching users and orgs) I am fine with this for now with a couple of tweaks to this PR.  

In general, I would like to lean away from caching data in a primary data store like Postgres and instead make more use of Redis for things like this where we are storing simple JSON data structures. I think another step in the right direction would be if we could send more simple data structures to our frontend instead of ActiveRecord models to populate the views. For example, it would be awesome if comments wasn't an array of AR objects but instead hashes. If we want to lean more into Elasticsearch this would force us to head in that direction since we would be returning all hashes from there. All ideas for another day but wanted to throw them out there to get people thinking :)

citizen428:
> I would like to lean away from caching data in a primary data store like Postgres and instead make more use of Redis for things like this where we are storing simple JSON data structures.

I don't necessarily disagree, but it means two independent roundtrips to separate datastores, which may partially offset the benefit of caching.

>  I think another step in the right direction would be if we could send more simple data structures to our frontend instead of ActiveRecord models to populate the views.

The problem is not AR vs hash here, but that we generally select all columns, independent of which ones we need.

```ruby
ObjectSpace.memsize_of(user)
#=> 144
ObjectSpace.memsize_of(user.attributes)
#=> 7264
ObjectSpace.memsize_of(user.as_json)
#=> 3424
```

Note: In the first case the object is very small because it essentially just holds a collection of pointers to the actual data. The size of an "empty" object itself is 40 bytes, independent of what it is:

```ruby
ObjectSpace.memsize_of(Object.new)
#=> 40
ObjectSpace.memsize_of({})
#=> 40
```

If we are concerned with reducing the amount of allocated memory, we need to do this on the ORM level so the unwanted attributes don't get materialized in the first place. If we load them and then extract a hash out of an AR instance we end up allocating some of the data twice.

> If we want to lean more into Elasticsearch this would force us to head in that direction since we would be returning all hashes from there.

Based on my previous experience people generally go the other way and wrap ES results in structs or `[Hashie::Mash](https://www.rubydoc.info/github/intridea/hashie/Hashie/Mash)` instances. Having objects allows for leveraging OO (class specific decorators, delegation, etc), which is not quite as easy with hashes, so you end up with lots of ad-hoc code.

mstruve:
I have used `Hashie::Mash` in the past with ES data and it worked great for views! I also think some of the very complex queries we are sending against Postgres would be better suit and easier for Elasticsearch to handle. Especially when we are joining over tables like tags.

citizen428:
> I have used Hashie::Mash in the past with ES data and it worked great for views!

😃 We already have the Hashie gem as a transitive dependency, so we can also promote it to the `Gemfile` and start using it more purposefully.

> I also think some of the very complex queries we are sending against Postgres would be better suit and easier for Elasticsearch to handle. Especially when we are joining over tables like tags.

I tackled this in the past by indexing more complex documents into ES that don't have a direct mapping to a single `ActiveRecord` (e.g. directly embedding associations as nested documents, etc.). I guess this is something we best tackle query-by-query as even relatively complex joins are not necessarily performance bottlenecks if the right indexes for a hash join are in place.

PR #10169, opened 2020-09-02
Session cookie domain hotfix for login and logout

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

TLDR; we have cases of double cookies that have the same name but differ on domain, regardless on the value of `SameSite` (both were changed at the same time). 

This hotfix should work both for those who can't login and those who can't logout

#9967 

Cross fingers and hope for the best 🤞

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -44,6 +44,14 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
     )
 
     if user_persisted_and_valid?
+      # Deleting the session cookie with the previous app domain, the one without the leading dot.
+      # This should fix the double cookie scenario
+      domain = Rails.env.production? ? ApplicationConfig["APP_DOMAIN"] : nil
+      if domain&.starts_with?(".")
+        domain_without_leading_dot = domain.gsub(/\A./, "")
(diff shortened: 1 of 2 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -1,9 +1,33 @@
 class SessionsController < Devise::SessionsController
+  def create
```
rhymes, line 2: this is where email / password login goes through

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -44,6 +44,14 @@ def callback_for(provider)
     )
 
     if user_persisted_and_valid?
+      # Deleting the session cookie with the previous app domain, the one without the leading dot.
```
rhymes, line 47: this is where SSO login goes through

vaidehijoshi (approved):
This conceptually all makes sense to me. The only thing I can think to add here is a comment about when we want to remove this bandaid (we probably don't need to keep this code around after the fix for the old cookie has been live for some time, right?)

atsmith813 (approved):

maestromac (approved):

rhymes:
@vaidehijoshi I added a generic "2 weeks from deployment"

benhalpern (approved):
I'll merge this and be on the lookout for behavior in prod.

PR #10171, opened 2020-09-02
Bug Fix:Sort Reactions by created_at

mstruve:
- [x] Bug Fix (Part 2 to https://github.com/forem/forem/pull/10142)

## Description
We store reactions in Elasticsearch and the default way we sort them is by ID desc which will give us the most recent reactions first. HOWEVER, this is not quite the case. IDs are stored in Elasticsearch as strings which means when you are doing a value comparison you end up with weird behavior such as 99 > 100. The means Reactions are not being sorted correctly all the time. This PR switches from using the ID to sort reactions to created_at

https://github.com/orgs/forem/projects/6#card-44644200

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/services/search/query_builders/reaction.rb
+++ b/app/services/search/query_builders/reaction.rb
@@ -22,7 +22,7 @@ module Search
       }.freeze
 
       DEFAULT_PARAMS = {
-        sort_by: "id",
+        sort_by: "created_at",
         sort_direction: "desc",
         size: 0
       }.freeze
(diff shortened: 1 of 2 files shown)


maestromac (approved):

atsmith813 (approved):

mstruve:
Putting this in draft until reactions are done indexing to prevent accidental merge

citizen428 (approved):
Interesting that ES has no built-in way to do a natural sort on numeric strings. You can do it by [defining your own analyzer](https://jprante.github.io/elasticsearch-analysis-naturalsort/#collation-keys) but I think `created_at` will be fine for our use case.

PR #10175, opened 2020-09-02
Optimization:Add Index for User ID and Mailer to Ahoy Messages

mstruve:
- [x] Optimization

## Description
In our `EmailDigestArticleCollector` we look up a user's most recent ahoy messages to help us determine if we should be sending them a Digest Email. 
```ruby
@last_user_emails ||= @user.email_messages.where(mailer: "DigestMailer#digest_email").limit(10)
```
The average for the lookup time for this query is almost 2 seconds, with it being closer to 10 when the database is under higher load. This should help speed things up a bit. 
Screen Shot 2020-09-02 at 3 49 47 PM

https://github.com/orgs/forem/projects/6#card-44508249


maestromac (approved):

rhymes (approved):
YES!

PR #10179, opened 2020-09-02
Revert removal of unneeded eager load (if needed)

benhalpern:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

After merging this PR we've seen a slow down in requests. #10168

Which makes _zero_ sense to me because the affects of this optimization should be very minimal one way or another as they are behind the fragment cache and generally minimal either way... 

But I figured I'd toss this reversal up unless there's another explanation about activity going on which would cause slower response times.

.... In looking more closely at this code, there are some much bigger optimizations that could be made here either way, but I don't know if this is actually something affecting responses or a coincidence.

### DIFF
--- a/app/labor/sticky_article_collection.rb
+++ b/app/labor/sticky_article_collection.rb
@@ -22,6 +22,7 @@ class StickyArticleCollection
 
   def tag_articles
     @tag_articles ||= Article.published.tagged_with(article_tags, any: true)
+      .includes(:user)
       .where("public_reactions_count > ? OR comments_count > ?", reaction_count_num, comment_count_num)
       .where.not(id: article.id).where.not(user_id: article.user_id)
       .where("featured_number > ?", 5.days.ago.to_i)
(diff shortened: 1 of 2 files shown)


benhalpern:
In looking at the traces, it doesn't seem like these queries are actually the thing slowing much of anything down...

`ArticleSuggester` are responsible for much more of the expensive queries in everything I looked at.

mstruve:
Here is what I am seeing and we saw the same thing yesterday when Mac deployed his ProfileImage changes. An increase in the 50th percentile response time. I have our alerts bumped from 1500ms to 2000ms to keep them from going off
Screen Shot 2020-09-02 at 6 54 56 PM

And we saw a decrease in the number of story endpoint requests taking less than 100ms https://ui.honeycomb.io/thepracticaldev/datasets/rails/result/vXXXwGt1zbi
Screen Shot 2020-09-02 at 6 56 28 PM

mstruve:
I have no idea why but response times are going back to normal. 
Screen Shot 2020-09-02 at 8 49 55 PM
Screen Shot 2020-09-02 at 8 50 09 PM

PR #10180, opened 2020-09-02
Add routes for i18n

atsmith813:
- [x] Feature

## Description
This PR is starting to lay the groundwork to support i18n and viewing Forem in various languages. This introduces the routing structure we want to use in the app by nesting all routes under an optional scope `/locale/:locale` or `/locale/fr-ca`. This will allow routes such as `/locale/lang-fr-ca/username`. As we build on i18n, this is how users will access Forem in various languages (such as French Canadian).

Setting this up as a "sub-folder" enables us to better approach SEO, caching, and server-side code in the future. We are doing it as `/locale/fr-ca/` as opposed to the more widely recognized `/fr/ca` approach to give ourselves more flexibility on how we want to handle future changes. Our routing is also complicated with several clashing top-level, dynamic domains.

The loose gameplan here is to lay the groundwork (routes, caching, etc) so that the app has the ability to understand the language preference to build on top of later. **Check out the documentation added in this PR for more information.**

**TODO**
- [x] Add note about moving routes outside of locale
- [x] Fix specs
- [x] Add a few specs around locale
- [ ] Add documentation

1. Fire up the app locally.
3. Add `/locale/fr-ca/` to the front of some routes and see that it still works (i.e. http://localhost:3000/locale/fr-ca/dashboard).
4. Try to update your username to `lang-anything` and verify it shows an error.
5. Visit a user's profile to ensure the route still works (http://localhost:3000/locale/fr-ca/username.

That's it 🎉 

## Added tests?
- [x] yes

## Added to documentation?
- [x] yes

spinning_globe

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,6 +5,8 @@ Rails.application.routes.draw do
     controllers tokens: "oauth/tokens"
   end
 
+  # Devise does not support scoping omniauth callbacks under a dynamic segment
+  # so this lives outside our i18n scope.
   devise_for :users, controllers: {
     omniauth_callbacks: "omniauth_callbacks",
     registrations: "registrations",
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -18,496 +20,502 @@
     delete "/sign_out", to: "devise/sessions#destroy"
   end
 
-  require "sidekiq/web"
-  require "sidekiq_unique_jobs/web"
-  require "sidekiq/cron/web"
+  # [@forem/delightful] - all routes are nested under this optional scope to
+  # begin supporting i18n.
+  scope "(/locale/:locale)", defaults: { locale: nil } do
```
atsmith813, line 25: There are several routes here that we could move outside of the `/locale/:locale` scope because we don't need them internationalized. I figured by starting with them all nested, we can unnest them over time. As long as the default URLs still work, it doesn't hurt to have them here.

atsmith813:
This might be a tricky PR to merge because the routes file changes so often and introduces conflicts 😞

PR #10187, opened 2020-09-03
Adds ability to create new sponsorships from admin dashboard

fdoxyz:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds the ability to create sponsorships in `/admin/sponsorships` and also adds tabs for filtering by status on `/admin/sponsorships`

## Related Tickets & Documents

Same topic as #10153: Trying to extract missing features from `/resource_admin` to `/admin`

## QA Instructions, Screenshots, Recordings

You should now be able to create new sponsorships from within `/admin/sponsorships`

Screen Shot 2020-09-03 at 06 07 37

Screen Shot 2020-09-03 at 06 12 57

## Added tests?

- [x] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

## [optional] What gif best describes this PR or how it makes you feel?

dog comforting

PR #10189, opened 2020-09-03
Add sponsorships to seed data

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization ?
- [ ] Documentation Update

## Description

Noticed this would have been helpful when I checked out #10187

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -530,6 +530,20 @@ end
 
 ##############################################################################
 
+seeder.create_if_none(Sponsorship) do
+  organizations = Organization.take(3)
+  organizations.each do |organization|
+    Sponsorship.create!(
+      organization: organization,
+      user: User.order(Arel.sql("RANDOM()")).first,
+      level: "silver",
+      blurb_html: Faker::Hacker.say_something_smart,
+    )
+  end
+end
+
+##############################################################################
+
 puts <<-ASCII
 
   ```````````````````````````````````````````````````````````````````````````


db/seeds.rb
```diff
@@ -530,6 +530,19 @@ def create_if_none(klass, count = nil)
 
 ##############################################################################
 
+seeder.create_if_none(Sponsorship) do
+  5.times do
+    Sponsorship.create!(
+      organization: Organization.order(Arel.sql("RANDOM()")).first,
+      user: User.order(Arel.sql("RANDOM()")).first,
+      level: "silver",
```
rhymes, line 538: there's a tiny chance this breaks if `random()` returns the same organization twice as levels have to be unique per org

rhymes (commented):
Cool!

rhymes (approved):

db/seeds.rb
jacobherrington: Good point! The minimum number of organizations is three, so to avoid this issue I'll just make one for each of the first three organizations. :+1: We can make this generate more diverse data later if we feel it will make life easier.

maestromac (approved):
LGTM!

That failed spec is a flaky spec.

PR #10190, opened 2020-09-03
Move deletion of legacy cookie higher in omniauth callback

vaidehijoshi:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

A continuation on https://github.com/forem/forem/pull/10169. As mentioned in that PR, we weren't entirely sure if this was going to work. We had @lisasy test this out on production and we _suspect_ that when signing in with SSO, the cookie might be being deleted too late, so this PR moves it up higher in the `OmniauthCallbacksController#callback_for`.

I also updated the current `NOTE` to a `TODO` with a specific date when we should remove this temporary code :)  

Please see https://github.com/forem/forem/pull/10169 for more context!

Not really possible to test this unless you are in production _and_ if you have the double cookie instance 🙃 

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Need to have @lisasy test this out on production once it has been deployed.

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -34,6 +34,15 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   private
 
   def callback_for(provider)
+    # Deleting the session cookie with the previous app domain, the one without the leading dot.
+    # This should fix the double cookie scenario
+    # NOTE: this code is a hotfix, and shall be removed soon (around 2 weeks from deployment)
+    domain = Rails.env.production? ? ApplicationConfig["APP_DOMAIN"] : nil
+    if domain&.starts_with?(".")
+      domain_without_leading_dot = domain.gsub(/\A./, "")
+      cookies.delete(ApplicationConfig["SESSION_KEY"], domain: domain_without_leading_dot)
+    end
+
     auth_payload = request.env["omniauth.auth"]
     cta_variant = request.env["omniauth.params"]["state"].to_s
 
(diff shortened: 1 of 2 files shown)


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -34,6 +34,12 @@ def failure
   private
 
   def callback_for(provider)
+    # Deleting the session cookie with the legacy app domain, which does NOT include a preceding dot.
+    # This should fix the double cookie scenario.
+    # TODO: this code is a hotfix, we should remove it after 09/18/2020.
+    legacy_cookie_domain = Rails.env.production? ? ApplicationConfig["APP_DOMAIN"] : nil
```
vaidehijoshi, line 40: Since the cookie we want to delete does _not_ include the preceding dot, we can just pull it directly from `ApplicationConfig["APP_DOMAIN"]` and delete it.

app/controllers/sessions_controller.rb
```diff
@@ -6,29 +6,28 @@ def create
...
     Rails.env.production? ? ApplicationConfig["APP_DOMAIN"] : nil
   end
 
-  # NOTE: this code is a hotfix, and shall be removed soon (around 2 weeks from deployment)
-  def delete_legacy_cookie
-    domain = cookie_domain
-
-    return unless domain&.starts_with?(".")
+  def cookie_domain
+    ".#{legacy_cookie_domain}"
```
vaidehijoshi, line 24: I think that eventually, we can just change this to be `".#{ApplicationConfig["APP_DOMAIN"]}"`.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -44,15 +50,6 @@ def callback_for(provider)
     )
 
     if user_persisted_and_valid?
-      # Deleting the session cookie with the previous app domain, the one without the leading dot.
-      # This should fix the double cookie scenario
-      # NOTE: this code is a hotfix, and shall be removed soon (around 2 weeks from deployment)
-      domain = Rails.env.production? ? ApplicationConfig["APP_DOMAIN"] : nil
-      if domain&.starts_with?(".")
```
vaidehijoshi, line 51: This didn't work in #10169 because the app domain was `dev.to`, so we never actually made it into this conditional!

benhalpern (approved):

atsmith813 (approved):

PR #10193, opened 2020-09-03
Add routes for i18n

atsmith813:
- [x] Feature

## Description
This PR is starting to lay the groundwork to support i18n and viewing Forem in various languages. This introduces the routing structure we want to use in the app by nesting all routes under an optional scope `/locale/:locale` or `/locale/fr-ca`. This will allow routes such as `/locale/fr-ca/username`. As we build on i18n, this is how users will access Forem in various languages (such as French Canadian).

Setting this up as a "sub-folder" enables us to better approach SEO, caching, and server-side code in the future. We are doing it as `/locale/fr-ca/` as opposed to the more widely recognized `/fr/ca` approach to give ourselves more flexibility on how we want to handle future changes. Our routing is also complicated with several clashing top-level, dynamic domains.

The loose gameplan here is to lay the groundwork (routes, caching, etc) so that the app has the ability to understand the language preference to build on top of later. **Check out the documentation added in this PR for more information.**

1. Fire up the app locally.
3. Add `/locale/fr-ca/` to the front of some routes and see that it still works (i.e. http://localhost:3000/locale/fr-ca/dashboard).
4. Visit a user's profile to ensure the route still works (http://localhost:3000/locale/fr-ca/username.
5. Visit an article to ensure the route still works (http://localhost:3000/locale/fr-ca/username/article-slug).

That's it 🎉 

## Added tests?
- [x] yes

## Added to documentation?
- [x] yes

spinning_globe

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,6 +5,8 @@ Rails.application.routes.draw do
     controllers tokens: "oauth/tokens"
   end
 
+  # Devise does not support scoping omniauth callbacks under a dynamic segment
+  # so this lives outside our i18n scope.
   devise_for :users, controllers: {
     omniauth_callbacks: "omniauth_callbacks",
     registrations: "registrations",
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -18,497 +20,502 @@
...
-
-  authenticated :user, ->(user) { user.tech_admin? } do
-    Sidekiq::Web.set :session_secret, Rails.application.secrets[:secret_key_base]
-    Sidekiq::Web.set :sessions, Rails.application.config.session_options
-    Sidekiq::Web.class_eval do
-      use Rack::Protection, origin_whitelist: [URL.url] # resolve Rack Protection HttpOrigin
+  # [@forem/delightful] - all routes are nested under this optional scope to
+  # begin supporting i18n.
+  scope "(/locale/:locale)", defaults: { locale: nil } do
+    get "/locale/:locale" => "stories#index"
```
atsmith813, line 26: This file diff is much scarier than it really is. We nested all routes and added these 2 lines (plus some comments above).

There are several routes here that we could move outside of the `/locale/:locale` scope because we don't need them internationalized. I figured by starting with them all nested, we can unnest them over time. As long as the default URLs still work, it doesn't hurt to have them here.

Nesting internationalization routes under `/locale` (open to changing this to `/language` or anything else) makes things cleaner. We don't have to fight so many dynamic top-level routes and we don't have to add further validations to things like organizations, pages, usernames, etc.

mstruve (commented):
Thanks for all the amazing information in this PR!

Question, is there a way we can make `/locale/en/a-path` redirect to `/a-path`? I ask bc if someone starts to use `locale/en` we are going to end up with a completely new set of monitoring data for English content which I would like to avoid and keep consolidated how we already have it.

benhalpern:
@mstruve the intended behavior is probably going to depend on a few UX concerns we're sort of punting on, but I think it's going to be a while before _users_ are using this functionality outside of specific testing, so I'd say we should keep this scenario in mind for the finished state and ship without that.

(I'm not _entirely_ sure what you're requesting specifically, but I think the answer either way is probably to tackle that a bit later.

atsmith813:
@mstruve that's included in the end goal - should be doable!

We're thinking of this in terms of default locale. So if a user visits the DEV instance of Forem where the default locale is `en-us`, they won't really be able to "hit" `/locale/en-us`. Within the app, it'll know to use the "root" routes and we'll look into a redirect for if someone tried to type `/locale/en-us` into the address bar manually.

We also are thinking about what that looks like for international Forems. Say an instance has a default locale of `fr-ca` (French - Canada). We'd want to provide a similar behavior, but for `/locale/fr-ca` at the root route. The concept of a default locale touches a few things (the feed algorithm, site configs, etc) that will be ironed out.

When it comes to paths without the region code, that's a little TBD on what we're going to do. `/en-us` is nice because locales like `/us-gb` are slightly different (currencies, date formats, etc.). The routes here will support both (`/locale/en` and `/locale/en-us`) as we build on the logic. Later on, we'll add logic into places like the application controllers to interpret and set the locale as needed. I imagine we'll focus on languages and translations first (`/locale/fr`) and then add region support with currencies, date formats, etc. after (`/locale/fr-ca`).

vaidehijoshi:
ProTip for anyone review this PR: if you hide whitespace changes, the diff is _much_ smaller to review :)

Screen Shot 2020-09-03 at 1 57 25 PM

docs/backend/internationalization.md
```diff
@@ -0,0 +1,148 @@
+---
+title: Internationalization (i18n)
+---
+
+# Internationalization (i18n)
+
+## What is internationlization (i18n)?
```
vaidehijoshi, line 7:
```suggestion
## What is internationalization (i18n)?
```

docs/backend/internationalization.md
```diff
@@ -0,0 +1,148 @@
...
+
+To over simplify the concept a bit, internalization (i18n for short) is the
+process of making the platform more user-friendly in various languages for
+people around the globe. This includes, but is certainly not limited to, things
+like making the site available in different languages, changing currency values
```
vaidehijoshi, line 12:
```suggestion
people around the globe. This includes, but is certainly not limited to, making the site available in different languages, changing currency values
```

docs/backend/internationalization.md
```diff
@@ -0,0 +1,148 @@
...
+The following is a high level outline of an approach to internationalization.
+This is by no means set in stone.
+
+We encourage you to open a pull request (PR) to this documentation or to
+contribute to internationalization with your ideas - we're
+[open-source](https://github.com/forem/forem/pulls)!
+
+## Routing
+
+We have logic for routes setup. You can visit a page and add `/locale/:locale`
```
vaidehijoshi, line 41:
```suggestion
We have logic for routes set up. You can visit a page and add `/locale/:locale`
```

mstruve (approved):
Sounds good @atsmith813!

Tested it locally and it works great!

docs/backend/internationalization.md
```diff
@@ -0,0 +1,148 @@
...
+to the beginning of the path. For example, if you visit the homepage, you can
+add `/locale/fr-ca` for French, Canadian where `fr` is the language code and
```
vaidehijoshi, line 43:
```suggestion
add `/locale/fr-ca/` for French, Canadian where `fr` is the language code and
```
Suggesting this only because I wonder if devs will just copypasta this to the end and not actually read and see that it's meant to be prepended to the _beginning_, as you describe in the docs. Not a blocker though, just an idea :)

docs/backend/internationalization.md
```diff
@@ -0,0 +1,148 @@
...
+`ca` is the region code.
+
+Setting up languages under this "sub-folder" approach helps with Search Engine
+Optimization (SEO), routing, and more.
+
+_Currently, the various language routes will not do anything - it will stil show
```
vaidehijoshi, line 49:
```suggestion
_Currently, the various language routes will not do anything - it will still render
```

benhalpern (approved):
Looks good to me. `/locale/fr-ca` passed the param as expected and I didn't see anything broken. Lots of functionality _to go_, but everything expected here looks good to me.

vaidehijoshi:
@atsmith813 This looks good, I just had some comments around the docs but looks like this got merged in the middle of my review! 😕

PR #10199, opened 2020-09-03
Include locale in fragment cache key

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Only some fragment caches make use of this method...... But it's a starter towards accounting for fragment caches across different locales.

This doesn't account for all fragments (and some we can't just change without thinking through the deployment concerns of cold hits), but this one change should be pretty harmless and let us account for at least this scenario for now.

Let me know what you think.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -184,7 +184,7 @@ module ApplicationHelper
     release_footprint = ApplicationConfig["RELEASE_FOOTPRINT"]
     return path if release_footprint.blank?
 
-    "#{path}-#{release_footprint}"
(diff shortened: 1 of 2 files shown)


atsmith813 (approved):

vaidehijoshi (approved):
seems okay to change this one since it looks like it anyways is busted every time the `RELEASE_FOOTPRINT` is updated

vaidehijoshi:
looks like a failing spec related to the cache key change, but once CI goes green, i think it should be okay 🚀

PR #7246, opened 2020-04-13
Add optional fields to user

achasveachas:
- this provides the back-end implementation for #2365
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR provides a back-end implementation for adding optional fields to user profiles (e.g. pronouns etc.) as discussed in #2365

## Related Tickets & Documents
#2365

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
N/A

## Added tests?

- [x] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

## [optional] What gif best describes this PR or how it makes you feel?

You can't just assume everyone's pronouns buddy!

PR #10202, opened 2020-09-04
Add custom profile fields

citizen428:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

1. Originally discussed in #2365
2. First implementation added in #7246 (never used)
3. Removed in  #9579 to prepare for profile generalization
4. Re-added in this PR as part of the ongoing profile generalization work

There's not much to test here yet, except maybe that you can't create more than 3 profile fields per user.

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed


app/models/concerns/acts_as_profile_field.rb
```diff
@@ -0,0 +1,16 @@
+concern :ActsAsProfileField do
```
citizen428, line 1: This is a [shortcut method](http://api.rubyonrails.org/classes/Module/Concerning.html#method-i-concerning) defined in `Module::Concerning` which is equivalent to the following:

```ruby
module ActsAsProfileField
  extend ActiveSupport::Concern

   ...
end
```

app/models/profile_field.rb
```diff
@@ -1,9 +1,7 @@
 class ProfileField < ApplicationRecord
-  self.ignored_columns = ["group"]
-
-  before_create :generate_attribute_name
+  include ActsAsProfileField
 
-  WORD_REGEX = /\w+/.freeze
+  self.ignored_columns = ["group"]
```
citizen428, line 4: This will eventually get removed, but I left it in for now as it has nothing to do with this PR.

db/migrate/20200904040009_create_custom_profile_fields.rb
```diff
@@ -0,0 +1,18 @@
+class CreateCustomProfileFields < ActiveRecord::Migration[6.0]
+  def change
+    safety_assured do
```
citizen428, line 3: Since this is a brand new table adding the foreign key is harmless. If people prefer I can remove the `safety_assured` block and do it in two migrations instead.

Ridhwana (approved):
Thanks for the comments here, it helped me to follow the code. 🚀

app/models/custom_profile_field.rb
```diff
@@ -0,0 +1,16 @@
+class CustomProfileField < ApplicationRecord
+  include ActsAsProfileField
+
+  belongs_to :profile
+
+  validate :validate_maximum_count
+
+  private
+
+  # We allow a maximum of 3 custom profile fields per user
```
mstruve, line 10: This seems low for a max. I feel like 5 might be a little less restrictive.

mstruve (commented):
Do I have this right? We have pre-defined profile fields that are set by the site admin. These are in addition to those and are set by the user themselves.

citizen428:
> Do I have this right? We have pre-defined profile fields that are set by the site admin. These are in addition to those and are set by the user themselves.

@mstruve Exactly! 😃 It allows the users a bit of self-expression, by adding more info about themselves. The original issue brought up personal pronouns as a possible use case.

app/models/custom_profile_field.rb
citizen428: @mstruve This is the value the original implementation in #7246 had. My main goal was reimplementing what we previously had (and never used and then removed) in the context of the new generalized profiles. I have no strong preference in regards to the numbers, so since you expressed one 5 it shall be.

citizen428:
@mstruve Updated the limit and resolved the merge conflict in `schema.rb`.

mstruve (approved):
🚀

PR #10206, opened 2020-09-04
Enable admin-configured app_domain

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In order to allow a user to start with a scenario like `temp-forem-red-fish.forem.cloud` to be migrated to `taylorswiftfanclub.com` through settings, we need to override certain conventions that currently exist.

.... But Rails mostly is friendly to this scenario. But it takes a few hacks here and there to make sure it all works.

This is functionality which will enable a smooth self-serve Forem Cloud experience.

The biggest areas of scrutiny needed is the stuff related to cookies... Stuff we already saw some issues with a prior migration, and then everything related to our existing hack to ensure `remember_user_token` is present for logged in users (aka the cookie we use to send to the edge to detect logged in/logged out edge cache in both Heroku and Forem Cloud)


app/controllers/concerns/valid_request.rb
```diff
@@ -10,7 +10,7 @@ def valid_request_origin?
     return if Rails.env.test?
 
     if request.referer.present?
-      request.referer.start_with?(URL.url)
+      request.referer.start_with?(ApplicationConfig["APP_PROTOCOL"] + SiteConfig.app_domain)
```
citizen428, line 13: Why are we changing this away from `URL.url`, wouldn't the method still work?

app/controllers/admin/tools_controller.rb
```diff
@@ -40,8 +40,8 @@ def handle_article_cache
     end
 
     def bust_link(link)
-      if link.starts_with?("https://#{ApplicationConfig['APP_DOMAIN']}")
-        link.sub!("https://#{ApplicationConfig['APP_DOMAIN']}",
+      if link.starts_with?("https://#{SiteConfig.app_domain}")
+        link.sub!("https://#{SiteConfig.app_domain}",
```
citizen428, line 44: Should this use `URL.url` (and maybe update that method if necessary)?

app/controllers/buffered_articles_controller.rb
```diff
@@ -16,6 +16,6 @@ def buffered_articles_urls
                end
 
     paths = relation.pluck(:path)
-    paths.map { |path| "https://#{ApplicationConfig['APP_DOMAIN']}#{path}" }
+    paths.map { |path| "https://#{SiteConfig.app_domain}#{path}" }
```
citizen428, line 19: Since we're already changing code here, maybe this should be `URL.url(path)`?

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  private
+
+  def legacy_cookie_domain
+    Rails.env.production? ? SiteConfig.app_domain : nil
+  end
+
+  def cookie_domain
+    ".#{legacy_cookie_domain}"
```
citizen428, line 24: `legacy_cookie_domain` can be `nil`, should we guard against this or do we want it to return `"."` in this case?

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    # Let's say goodbye to all the cookies when someone signs out.
+    cookies.clear(domain: cookie_domain)
+
+    super
+  end
+
+  private
+
+  def legacy_cookie_domain
+    Rails.env.production? ? SiteConfig.app_domain : nil
```
citizen428, line 20: This seems cleaner, especially if we add a guard clause to cookie domain (as asked below):

```suggestion
    Rails.env.production? && SiteConfig.app_domain
```

If there won't be a guard clause we can't make this change.

app/decorators/article_decorator.rb
```diff
@@ -22,7 +22,7 @@ def cached_tag_list_array
   end
 
   def url
-    "https://#{ApplicationConfig['APP_DOMAIN']}#{path}"
+    "https://#{SiteConfig.app_domain}#{path}"
```
citizen428, line 25: This should probably also become `URL.url(path)`.

app/lib/url.rb
```diff
@@ -7,7 +7,7 @@ def self.protocol
   end
 
   def self.domain
-    ApplicationConfig["APP_DOMAIN"]
+    Rails.application&.initialized? ? SiteConfig.app_domain : ApplicationConfig["APP_DOMAIN"]
```
citizen428, line 10: I'm not sure how I feel about this. Maybe things that rely on the environment variable before Rails is fully initialized should use it directly instead of through the `URL` module?

citizen428 (commented):
I only left some very general comments as I miss some of the context (e.g. the details of the recent cookie issues we experienced, etc.)

app/middlewares/set_cookie_domain.rb
```diff
@@ -0,0 +1,12 @@
+class SetCookieDomain
```
rhymes, line 1: Is this needed because the domain is now user configured?

config/initializers/devise_monkeypatch.rb
```diff
@@ -0,0 +1,15 @@
+module Devise
```
rhymes, line 1: Should this have a reason/explanation for this monkey patch? I'm not sure what's going on here...

jdoss (approved):
LGTM!

app/controllers/application_controller.rb
```diff
@@ -144,6 +146,17 @@ def initialize_stripe
...
   end
 
+  def remember_cookie_sync
+    # Set remember cookie token in case not properly set.
+    if user_signed_in? && cookies[:remember_user_token].blank?
+      current_user.remember_me = true
+      current_user.remember_me!
+      remember_me(current_user)
+    elsif !user_signed_in?
+      cookies.delete :remember_user_token, domain: ".#{SiteConfig.app_domain}"
```
vaidehijoshi, line 156: Just leaving a note here to ensure this actually works; I recently learned that `cookies.delete`[ doesn't necessarily work the way you'd expect](https://github.com/forem/forem/pull/10211).

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def cookie_domain
```
vaidehijoshi, line 23: Did we intentionally add this back in? This has been removed on `master`: https://github.com/forem/forem/pull/10211/files#diff-d5241d488259f32ecbe2f636133e5ddaL23

spec/helpers/application_helper_spec.rb
```diff
@@ -76,6 +76,7 @@
     before do
       allow(ApplicationConfig).to receive(:[]).with("APP_PROTOCOL").and_return("https://")
       allow(ApplicationConfig).to receive(:[]).with("APP_DOMAIN").and_return("dev.to")
+      SiteConfig.app_domain = "dev.to"
```
vaidehijoshi, line 79: Do we need to unset this so that it doesn't leak into other tests?

benhalpern: Ideally all functionality is independently flexible such that it can take a different `SiteConfig.app_domain`, so as long as the test itself doesn't test the specific value here without first setting it, it's probably okay if it changes from test to test.

... But that might not be ideal in terms of our overall test suite consistency. I'm not sure if it's better to unset values like this or to be resilient to changing values 🤷

app/controllers/sessions_controller.rb
benhalpern: Yeah, I meant to pull latest, I guess I didn't properly resolve things. Good call.

benhalpern:
I addressed a few of the questions, but I want to look and see if there is any of this I can remove. Part of my work here may not be 100% necessary as I was trying a bunch of stuff. Anything I can remove from this I will.

To @rhymes' comment... The devise monkeypatch explicitely adds the domain to this function where it wasn't before, but I'm actually not _sure_ it's needed. I'll need to work with this further. The last thing I want to do is mess up anything with cookies again 😬

app/controllers/admin/tools_controller.rb
```diff
@@ -40,8 +40,8 @@ def handle_article_cache
     end
 
     def bust_link(link)
-      if link.starts_with?("https://#{ApplicationConfig['APP_DOMAIN']}")
-        link.sub!("https://#{ApplicationConfig['APP_DOMAIN']}",
+      if link.starts_with?(URL.url)
+        link.sub!(URL.url,
```
citizen428, line 44: Seems the formatting got a bit funky here, the second argument was moved to the next line.

app/controllers/admin/feedback_messages_controller.rb
```diff
@@ -107,7 +107,7 @@ def generate_message(params)
       <<~HEREDOC
         *New note from #{params['author_name']}:*
         *Report status: #{params['feedback_message_status']}*
-        Report page: https://#{ApplicationConfig['APP_DOMAIN']}/admin/reports/#{params['noteable_id']}
+        Report page: https://#{SiteConfig.app_domain}/admin/reports/#{params['noteable_id']}
```
citizen428, line 110: We could also use a path or url helper here, e.g. 

```ruby
admin_report_path(id: params['noteable_id'])
```

This seems like it would generally be a more robust way to generate internal links.

config/application.rb
```diff
@@ -44,6 +44,10 @@ class Application < Rails::Application
     config.autoload_paths += Dir["#{config.root}/lib"]
     config.eager_load_paths += Dir["#{config.root}/lib"]
 
+    Dir["./app/middlewares/*.rb"].sort.each do |file|
```
citizen428, line 47: Do we really need to require these explicitly? Generally, subdirectories of `app/` are autoloaded, but if they aren't yet at this point, so a comment would maybe help.

spec/helpers/application_helper_spec.rb
citizen428: I'd probably replace it with a mock:

allow
```suggestion
      alllow(SiteConfig).to receive(:app_domain).and_return("dev.to")
```

This is not only about our codebase being resilient to changes, but also our test environment, i.e. will this work with different Travis workers running different sets of tests due to Knapsack and changing this value non-deterministically?

app/controllers/application_controller.rb
benhalpern: Yeah, this was just a bunch of trial and error. Not needed in the end.

benhalpern:
Okay, this seems to be working as expected. The TLDR is that we need to modify some code to check for the `SiteConfig` version, otherwise all the config is set to check what was set at boot up time.

app/controllers/admin/tools_controller.rb
benhalpern: Fixed

config/application.rb
benhalpern: Added a comment

app/controllers/admin/feedback_messages_controller.rb
benhalpern: I'm going to treat this as out of scope here just to cut off the little changes here and focus mostly on the core functional changes.

app/lib/url.rb
benhalpern: I ran into some less-important situations right now that technically include this before boot, but in unimportant ways which could change so I think we can keep this and circle back to it. All current forems will have the same value for env and site config stuff right now.

vaidehijoshi:
Is there a way to safely test this change on a different forem? (forem.dev?) This just feels like a potential breaking change and I am not 100% confident about what it's actually doing to really confidently review it...which makes me want to have some safety nets around deploying it

benhalpern:
Yeah @vaidehijoshi I've been testing on `community.benhalpern.com` which can also be accessed by `https://benhalpern-community-herokuapp-com.global.ssl.fastly.net`.

Signing in will only work with the version that `SiteConfig` is configured for, which is currently the `community.benhalpern.com` domain.

I think this is our best bet for validating that we won't have a regression on `dev.to` if we deploy.

benhalpern:
I also tested it on the `https://coffee-and-plants.herokuapp.com/` test app and it works fine. I think it's good to ship.

app/middlewares/set_cookie_domain.rb
```diff
@@ -0,0 +1,15 @@
+# Since we must explicetly set the cookie domain in session_store before SiteConfig is available,
```
vaidehijoshi, line 1:
```suggestion
# Since we must explicitly set the cookie domain in session_store before SiteConfig is available,
```

config/environments/production.rb
```diff
@@ -110,13 +110,13 @@
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
 
-  config.app_domain = ENV["APP_DOMAIN"] || "localhost:3000"
+  # config.app_domain = ENV["APP_DOMAIN"] || "localhost:3000"
```
vaidehijoshi, line 113: hm, is this a temporary comment or unintentional and meant to be removed? if it's intentional, i'd add a TODO here to explain what it is there for/when it can be removed

config/initializers/devise_monkeypatch.rb
```diff
@@ -0,0 +1,24 @@
...
+  module Controllers
+    module Rememberable
+      # We need to use SiteConfig.app_domain instead of default Rails config on boot
+      def remember_cookie_values(resource)
+        options = { httponly: true }
+        options.merge!(forget_cookie_values(resource))
+        options.merge!(
+          value: resource.class.serialize_into_cookie(resource),
+          expires: resource.remember_expires_at,
+          domain: ".#{SiteConfig.app_domain}",
```
vaidehijoshi, line 13: At some point (a separate PR?) we may want to extract this out rather than appending `.` in front of the app domain. Too scared/scared from that double-cookie domain scenario last week and afraid we might accidentally forget to add this at some point 😬

config/initializers/devise_monkeypatch.rb
```diff
@@ -0,0 +1,24 @@
...
+        )
+      end
+
+      def self.cookie_values
+        # Default: Rails.configuration.session_options.slice(:path, :domain, :secure)
+        # We need to use SiteConfig.app_domain instead of default Rails config on boot
+        { domain: ".#{SiteConfig.app_domain}", secure: ApplicationConfig["FORCE_SSL_IN_RAILS"] == "true" }
```
vaidehijoshi, line 20: Same comment here re: cookie domain

config/environments/production.rb
benhalpern: Deleted this code, thanks.

config/initializers/devise_monkeypatch.rb
benhalpern: I think we're safe from getting this wrong because the cookie isn't going to be set if the request doesn't match the request domain.

But 1000% understand the scarring issue. It's the right concern. I'm pretty sure this is safe, but we could follow up with more safeguards and refactors on this all in future work.

PR #10211, opened 2020-09-04
Remove all attempts to delete cookies

maestromac:
- [x] Bug Fix

## Description
@atsmith813 and I figured out that attempting to delete legacy cookies when user sign-in/out would not work because

>A cookie can only be overwritten (or deleted) by a subsequent cookie exactly matching the name, path and domain of the original cookie. Even though a cookie with domain “.example.org” set by www.example.org is perfectly valid, it will not overwrite a previous cookie of the same name which was set against “www.example.org”. Instead, both cookies will be stored, and on subsequent requests, only one will be sent.

The proper solution would be to update our session key. I'm removing these codes because they may cause unintended issues.

Undo https://github.com/forem/forem/pull/10190, https://github.com/forem/forem/pull/10169, and https://github.com/forem/forem/pull/10036

refs:
https://sam-low.com/ruby/rails/cookie/changing-rails-session-cookie-domain.html
https://www.sitepoint.com/3-things-about-cookies-you-may-not-know/
This PR doesn't remove or add any new features. You should be able to sign in and sign out without difficulty.

## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a


mstruve (approved):
Thanks for the cleanup! 🚀

atsmith813 (approved):

vaidehijoshi (approved):
Great work figuring this out 👍🏽

rhymes (approved):
Thanks a lot Mac!

PR #10215, opened 2020-09-04
Optimization:Increase refresh_interval For Possible High Indexing Indexes

mstruve:
- [x] Optimization

## Description
On 9/3 we had some of our "Slow Elasticsearch Query" alerts go off indicating that Elasticsearch response times were increasing. When I looked further into the issue I noticed that the leading spike came from indexing requests and we did have an increase in indexing at the time. 
Screen Shot 2020-09-04 at 4 35 23 PM

Another thing I noticed when digging into the Elasticsearch stats was that our refresh time had also spiked during this time. [By default, Elasticsearch will refresh an index every second](https://www.elastic.co/guide/en/elasticsearch/reference/current/tune-for-indexing-speed.html#_unset_or_increase_the_refresh_interval). Refreshing an index takes a considerable amount of resources which can cause indexing to slow down. To alleviate this pressure, if you don't need near-realtime searching then increasing the refresh interval can help. 

Screen Shot 2020-09-04 at 4 37 09 PM

This PR increases the refresh interval on 3 of our indexes, Feed Content, Users, and Reactions. 
- Feed Content and Users -> 5 seconds. We execute these updates in the background so we already dont expect them to be near real-time bc of the worker delay. 5 seconds seemed high enough to give us indexing gains but still low enough to be really snappy. 
- Reactions -> 2 seconds. When a user clicks to save an article to their readinglist, they might then just pop over to the reading list to see what is on it. For this reason, I kept this refresh shorter but still we cut the number of refreshes in half for a little more indexing power. 

For more information about what a Refresh Interval is, check out the "Refresh Interval" section in this blog post https://dev.to/molly_struve/scaling-elasticsearch-part-1-how-to-speed-up-indexing-2pel

CodeClimate can be ignored 

https://github.com/orgs/forem/projects/6#card-44926128

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/spec/services/search/base_spec.rb
+++ b/spec/services/search/base_spec.rb
@@ -3,6 +3,13 @@ require "rails_helper"
 RSpec.describe Search::Base, type: :service do
   let(:document_id) { 123 }
 
+  def recreate_tag_index
+    Search::Tag.delete_index if described_class.index_exists?
+    Search::Tag.create_index
+    Search::Tag.add_alias
+    Search::Tag.update_mappings
+  end
+
   before do
     # Need to use an existing index name to ensure proper data cleanup
     stub_const("#{described_class}::INDEX_NAME", "tags_#{Rails.env}")
@@ -69,7 +76,11 @@ RSpec.describe Search::Base, type: :service do
     end
   end
 
-  describe "::create_index", elasticsearch_reset: true do
+  describe "::create_index" do
+    before { recreate_tag_index }
+
+    after { recreate_tag_index }
+
     it "creates an elasticsearch index with INDEX_NAME" do
       described_class.delete_index
       expect(Search::Client.indices.exists(index: described_class::INDEX_NAME)).to eq(false)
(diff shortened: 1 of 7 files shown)


spec/services/search/base_spec.rb
```diff
@@ -3,6 +3,13 @@
 RSpec.describe Search::Base, type: :service do
   let(:document_id) { 123 }
 
+  def recreate_tag_index
```
mstruve, line 6: By only recreating a single index for these tests instead of all of them using `elasticsearch_reset: true` this spec's runtime went from 20+ seconds to 8 seconds.

atsmith813 (approved):
**suggestion (non-blocking):** Should we consider, someday, moving this sort of thing into an "advanced config" on a per Forem basis? I'm thinking new Forems won't need this optimization for a long time and it currently only applies to the DEV instance. Or vice versa, a Forem grows super large and they want to adjust this further for their scale.

mstruve:
> should we consider, someday, moving this sort of thing into an "advanced config" 

That is definitely something to think about. We would probably put that kind of config in ENV variables since advanced users would be able to set those themselves assuming they have control over their own environments. Probably a bit beyond where we are now since we are still trying to get most basic things into SiteConfig but the premise is a good one. Will be interesting to see what kinds of Forems in terms of hands off vs hands on get spun up in the next 6 months and how that shapes our code going forward.

maestromac (approved):
LGTM!

Issue #10164, opened 2020-09-02
Display date of listings


rhymes:
**Is your feature request related to a problem? Please describe.**

Right now both https://dev.to/listings and the single page for a listing do not display any kind of date.

**Describe the solution you'd like**

Listings are currently sorted by reverse `bumped_at` in the index, so we should probably display that but I also think the creation date is meaningful

**Describe alternatives you've considered**

Show either creation or bump date

**Additional context**

Dates should be localized for the user that's viewing them

Rafi993:
Hello there, If this is not taken I would like to help :)

Rafi993:
I just realised this feature does not have tag "ready for dev". I'll take it up if it is "ready for dev".

rhymes:
@Rafi993 this is definitely ready for dev, thank you!

Rafi993:
Awesome thanks @rhymes

PR #10217, opened 2020-09-05
Showing date in listing tile

Rafi993:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
It shows date in listing tile

closes https://github.com/forem/forem/issues/10164

1. https://dev.to/listings
2. See listings with date

I have updated the newer screenshots in the comments and I just left the old one here so that it would help during review

listing_date

listing_modal

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed


Rafi993:
I displayed the year hear too. But I think I should remove the year and use make it similar to what you see when you go inside the post 

date_format

any thoughts?

rhymes (commented):
Hi @Rafi993, thanks again for this!

A few proposed changes:

- I agree with you, I think we should keep the same "algorithm" that we have in posts and comments: if the year is the current one, display day and month, otherwise the full date

- I think we should both display the bump date and the creation date, otherwse it'd be weird to have listings without date if they haven't been bumped yet

- Please use the `<time>` element to add semantics to the date

- It'd be great if the `title` attribute of the `<time>` element would display the full localized date and time like we do for comments and articles.

Unfortunately we have a bit of duplication around the time code but these are two examples:

You can find some useful helpers here https://github.com/forem/forem/blob/master/app/assets/javascripts/utilities/localDateTime.js

Hope this helps!

app/javascript/listings/__tests__/SingleListing.test.jsx
```diff
@@ -24,6 +27,12 @@ const listing = {
 };
 
 describe('<SingleListing />', () => {
+  beforeEach(() => {
+    window.timestampToLocalDateTimeLong = () =>
```
Rafi993, line 31: I couldn't define common mock directory (for javascript) to put this mock function

app/assets/javascripts/utilities/localDateTime.js
```diff
@@ -25,18 +25,44 @@ function timestampToLocalDateTime(timestamp, locale, options) {
...
     month: 'long',
     day: 'numeric',
     hour: 'numeric',
     minute: 'numeric',
     second: 'numeric',
-  };
+  });
+}
+
+function timestampToLocalDateTimeShort(timestamp) {
```
Rafi993, line 42: I have added this util function which uses ```timestampToLocalDateTime``` to give shorter version of date similar to what you see in post (the helper used by post is rails helper). This could be used to replace duplicate date display logic in other places.

app/assets/javascripts/utilities/localDateTime.js
```diff
@@ -25,18 +25,44 @@ function timestampToLocalDateTime(timestamp, locale, options) {
   }
 }
 
-function addLocalizedDateTimeToElementsTitles(elements, timestampAttribute) {
+function timestampToLocalDateTimeLong(timestamp) {
```
Rafi993, line 28: I have added this util function which uses ```timestampToLocalDateTime``` defined above to give longer version of date.

app/serializers/search/listing_serializer.rb
```diff
@@ -6,6 +6,7 @@ class ListingSerializer < ApplicationSerializer
                :category,
                :contact_via_connect,
                :expires_at,
+               :created_at,
```
Rafi993, line 9: Listing API was not passing created_at date now so I added it.

Rafi993: This API  https://dev.to/search/listings?category=&listing_search=&page=0&per_page=75&tag_boolean_mode=all

Rafi993:
> Hi @Rafi993, thanks again for this!
> 
> A few proposed changes:
> 
> * I agree with you, I think we should keep the same "algorithm" that we have in posts and comments: if the year is the current one, display day and month, otherwise the full date
> * I think we should both display the bump date and the creation date, otherwse it'd be weird to have listings without date if they haven't been bumped yet
> * Please use the `<time>` element to add semantics to the date
> * It'd be great if the `title` attribute of the `<time>` element would display the full localized date and time like we do for comments and articles.
> 
> Unfortunately we have a bit of duplication around the time code but these are two examples:
> 
> You can find some useful helpers here https://github.com/forem/forem/blob/master/app/assets/javascripts/utilities/localDateTime.js
> 
> Hope this helps!

This is how it looks now after the changes

listing_tiles

listing_modal

It hides the year if it is the current year

app/assets/javascripts/utilities/localDateTime.js
```diff
@@ -13,7 +13,7 @@
   for more information.
 */
 function timestampToLocalDateTime(timestamp, locale, options) {
-  if (timestamp === '') {
+  if (!timestamp) {
```
Rafi993, line 16: I handled the case if timestamp is undefined.

app/assets/javascripts/utilities/localDateTime.js
Rafi993: I should have added these new functions at the end so that it would have shown better diff.

app/javascript/listings/__tests__/SingleListing.test.jsx
Rafi993: Also I'm making this function to return string without any actual logic here in the mock since the test can fail if the user has different locale settings in their system.

mstruve (commented):
In order for this to work we need to create a DataUpdateScript in order to reindex all listings to ensure they have the new created_at field present to display in the view. 

With that said, I am not sure if created_at is the best date here. If you want to indicate how old the listing is this date may be misleading. A user could create a listing but let it sit as a draft for a while before publishing it it. In that case, the listing would look older than it actually was in terms of how long it had been published.

Rafi993:
> In order for this to work we need to create a DataUpdateScript in order to reindex all listings to ensure they have the new created_at field present to display in the view.
> 
> With that said, I am not sure if created_at is the best date here. If you want to indicate how old the listing is this date may be misleading. A user could create a listing but let it sit as a draft for a while before publishing it it. In that case, the listing would look older than it actually was in terms of how long it had been published.

But when a draft becomes published won't it have bumped_at date?

Rafi993:
It seems that when a listing becomes published from draft its bumped_at date remains same

[app/controllers/concerns/listings_toolkit.rb](https://github.com/forem/forem/blob/acaeba0b3efd09f47236c5d3fb5f7d128cfc43f4/app/controllers/concerns/listings_toolkit.rb#L10)

```ruby 
 module ListingsToolkit
   def publish_listing
     @listing.update(published: true)
   end
 ...
 end
```

Rafi993:
and when creating listing for the first time bumped_at does get set (so bumped_at could never be null but I might be missing something)

https://github.com/forem/forem/blob/acaeba0b3efd09f47236c5d3fb5f7d128cfc43f4/app/controllers/concerns/listings_toolkit.rb#L90-L102

Rafi993:
I think if bumped_at date gets set when listing becomes published from draft state we could just use the bumped_at date (since bumped_at date won't be null for a published listing anymore) and avoid passing created_at date. Any thoughts in this?

citizen428:
@Rafi993 I don't think we should use `bumped_at` for this. Bumping a listing requires credits and is done to extend the lifetime of a listing past the original expiry date which is semantically different from promoting a draft to published.

Rafi993:
I think I'll extract out the date formatting util functions from here to solve this https://github.com/forem/forem/issues/10233. Since this PR is starting to doing too many things now. Is that ok?

rhymes:
@Rafi993 that's a great idea, let's start with the data formatting functions so we can review those separately. There are a few places where they are used (listingDate, modCenter/singleArticle/util.js at least).

Rafi993:
> @Rafi993 that's a great idea, let's start with the data formatting functions so we can review those separately. There are a few places where they are used (listingDate, modCenter/singleArticle/util.js at least).

Awesome

Rafi993:
> @Rafi993 I don't think we should use `bumped_at` for this. Bumping a listing requires credits and is done to extend the lifetime of a listing past the original expiry date which is semantically different from promoting a draft to published. To be clear: I saw your comment above, I'm also questioning if setting it at creation like we're currently doing is the right thing to do.

so would adding a new column published_at solve this?

citizen428:
@Rafi993 I guess it would. I'd be more explicit about the name though, `originally_published_at` or `first_published_at`. I'm usually a bit wary about adding more columns, but this one seems justified as it allows us to fully decouple the domain logic related to dates (`bumped_at`, `expires_at`, `first_published_at`) from the lifecycle of the DB record (`created_at` and `updated_at`).

rhymes:
@citizen428 how do we backfill the field for all the existing ones? Do we assume `created_at`?

citizen428:
@rhymes This seems like the most straight-forward approach, as we're probably already implicitly assuming that right now.

Rafi993:
> @rhymes This seems like the most straight-forward approach, as we're probably already implicitly assuming that right now.

Awesome

Rafi993:
Hello there, should I move the Pr to draft make the changes and request for review again?

Rafi993:
If this has already been done or there are other blocking issues now please feel free to change assignee or close this pull request.

citizen428:
@Rafi993 The proposed workflow (back to draft, fix, re-request reviews) works fine, thanks!

citizen428:
> Write database migrations (assume existing created_at as value for originally_published_at)

@Rafi993 Please also do the backfilling with a data update script, not directly in the migration. Just clarifying this since you specficially mentioned it for re-indexing, but not backfilling.

Rafi993:
> > Write database migrations (assume existing created_at as value for originally_published_at)
> 
> @Rafi993 Please also do the backfilling with a data update script, not directly in the migration. Just clarifying this since you specficially mentioned it for re-indexing, but not backfilling.

Awesome thanks @citizen428

lib/data_update_scripts/20200917125741_re_index_listing_with_originally_published_at.rb
```diff
@@ -0,0 +1,7 @@
+module DataUpdateScripts
+  class ReIndexListingWithOriginallyPublishedAt
+    def run
+      Listing.find_each(&:index_to_elasticsearch_inline)
```
Rafi993, line 4: when I run this I'm having the following error

```bash
$ rails data_updates:run

2020-09-17 19:37:10 +0530: PUT http://localhost:9200/classified_listings_development_alias/_doc/1 [status:400, request:0.005s, query:N/A]
2020-09-17 19:37:10 +0530: > {"id":1,"body_markdown":"## Explicabo\nTempora reiciendis dolorem. Est facilis in. Aut consequuntur impedit.\n## Sint\nSint maxime cum. Hic accusantium velit. Qui qui et.\n0. Fuga. \n1. Nemo. \n","bumped_at":"2020-09-17T12:22:38.734Z","category":"cfp","contact_via_connect":true,"expires_at":null,"originally_published_at":"2020-09-17T12:22:38.762Z","location":"Schimmelside","processed_html":"\n  <a href=\"#explicabo\" class=\"anchor\">\n  </a>\n  Explicabo\n\n\n<p>Tempora reiciendis dolorem. Est facilis in. Aut consequuntur impedit.</p>\n\n  <a href=\"#sint\" class=\"anchor\">\n  </a>\n  Sint\n\n\n<p>Sint maxime cum. Hic accusantium velit. Qui qui et.</p>\n\n<ol>\n<li>Fuga. </li>\n<li>Nemo. </li>\n</ol>\n","published":true,"slug":"ipsum-animi-aut-facere-5796","title":"Ipsum animi aut facere.","user_id":1,"tags":["python","discuss"],"author":{"username":"org8016","name":"Bream-Hall","profile_image_90":"/uploads/organization/profile_image/2/0d24410d-3fe9-48a6-ae67-9d1b1deb4464.png"},"last_indexed_at":"2020-09-17T14:07:10.866Z"}
2020-09-17 19:37:10 +0530: < {"error":{"root_cause":[{"type":"strict_dynamic_mapping_exception","reason":"mapping set to strict, dynamic introduction of [originally_published_at] within [_doc] is not allowed"}],"type":"strict_dynamic_mapping_exception","reason":"mapping set to strict, dynamic introduction of [originally_published_at] within [_doc] is not allowed"},"status":400}
2020-09-17 19:37:10 +0530: [400] {"error":{"root_cause":[{"type":"strict_dynamic_mapping_exception","reason":"mapping set to strict, dynamic introduction of [originally_published_at] within [_doc] is not allowed"}],"type":"strict_dynamic_mapping_exception","reason":"mapping set to strict, dynamic introduction of [originally_published_at] within [_doc] is not allowed"},"status":400}

```
any thoughts on how to go about solving this @forem/oss  ? And I feel like I should have done the updating index in batches.

Rafi993: I think this error is because of strict indexing.

app/views/layouts/admin.html.erb
```diff
@@ -25,8 +25,8 @@
 
 </head>
 <body class="admin"
-  data-honybadger-key="<%= ApplicationConfig['HONEYBADGER_JS_API_KEY'] %>"
-  data-release-footprint="<%= ApplicationConfig['RELEASE_FOOTPRINT'] %>"
+  data-honybadger-key="<%= ApplicationConfig["HONEYBADGER_JS_API_KEY"] %>"
```
Rafi993, line 28: I think prettier ran on this file when I merged master into this branch

lib/data_update_scripts/20200917125741_re_index_listing_with_originally_published_at.rb
mstruve: Have you run `bin/setup` recently to update your mappings? This indicates we are trying to index a field that is not in our mapping which is likely because you have not updated them recently

app/controllers/concerns/listings_toolkit.rb
```diff
@@ -134,7 +135,7 @@ def update
       process_after_unpublish
       return
     elsif listing_params[:action] == "publish"
-      unless @listing.bumped_at?
+      unless @listing.originally_published_at?
```
Rafi993, line 138: I changed bumped_at to originally_published_at to see if it is first publish

Rafi993:
apparently there is some test that is failing let me fix that

lib/data_update_scripts/20200917115821_backfill_listing_published_date.rb
```diff
@@ -0,0 +1,13 @@
+module DataUpdateScripts
+  class BackfillListingPublishedDate
+    def run
+      return unless Listing.column_names.include?("originally_published_at")
+
+      Listing.find_each do |listing|
+        if listing.published
```
citizen428, line 7: **suggestion (non-blocking):** Instead of using an `if` here let's just not fetch unpublished listings from the DB:

```ruby
Listing.where(published: true).find_each do |listing|
```

The question is, do we need `ActiveRecord` callbacks to run here? You have a separate indexing script that runs after this one anyway, so maybe we can just do everything in a single update statement without instantiating any Ruby objects:

```ruby
Listing.where(published: true).update_all('originally_published_at = created_at')
```

Maybe we should unify the two scripts into one:

```ruby
def run
  return unless Listing.column_names.include?("originally_published_at")

  Listing.where(published: true).update_all('originally_published_at = created_at')
  Listing.find_each(&:index_to_elasticsearch_inline)
end
```

citizen428 (approved):

lib/data_update_scripts/20200917125741_re_index_listing_with_originally_published_at.rb
citizen428, line 4: We only need to reindex the published listings, as they are the ones that get updated by the other script. Maybe we should actually combine them:

```
def run
  return unless Listing.column_names.include?("originally_published_at")

  published_listings = Listing.where(published: true).select(:id)
  published_listings.in_batches(of: 100) do |batch|
       batch.update_all('originally_published_at = created_at')
       Search::BulkIndexWorker.set(queue: :default).perform_async("Listing", batch.ids)
  end
end
```

I think I like this last version the best.

lib/data_update_scripts/20200917115821_backfill_listing_published_date.rb
mstruve: If we use `update!` here then that will trigger all of the listings to be reindexed and then we can remove the  `ReIndexListingWithOriginallyPublishedAt` DataUpdateScript below. I think that is the way to go

Rafi993: I have used 
```ruby
Listing.where(published: true).find_each do |listing|

```
with `update!`. I did not use `update_all` since I did not know if it will trigger re-indexing and I have removed separate datascript for re-indexing elastic search. Thank you @mstruve and @citizen428

mstruve (approved):

Issue #872, opened 2018-10-09
Typeahead liquid tag dropdown in new editor


benhalpern:
**Is your feature request related to a problem? Please describe.**
As a user, it can be hard to remember what exactly the commands for liquid tags in the editor are. In the way I might use this feature in my code, I'd like to type `{% ` and get a list of options for autocomplete.

**Additional context**
The editor is built in Preact, and ideally this could be done without new dependencies, but if we need to use [preact-compat](https://github.com/developit/preact-compat), we could.

I believe the editor will need to be written as a content-editable div, and we'd listen for the typing etc.

Ultimately it should work similarly to how I use typeahead in VSCode etc. 🙂

This is what it might look like:
![](https://thepracticaldev.s3.amazonaws.com/i/kftuy3gk6rvt9p6tgl70.png)

bigblind:
Would there be an endpoint where the client can ask the server for a list of liquid tags, or would the list on the front end be updated manually whenever a new tag is added?

benhalpern:
@bigblind For the purpose of the dropdown we'd probably ship it from the server with the page since there are only about a dozen.

So something like `<div data-liquidtags="[github,glitch,link,etc]"></div>`

And then we'd `JSON.parse` it to start playing with it.

theoutlander:
This is cool. I had made a similar request on the comment box thread (https://github.com/thepracticaldev/dev.to/issues/354). I could tackle this one after the @ mention auto-suggest feature.

In addition, these features will need to be ported over to the editor. Ideally, we should be able to reuse what I'm building, but first we will need to turn it into a reusable library.

cmgorton:
Since this one has been around for a while without much traction I am going to close it and encourage anyone who would like to see this feature in the future to open a discussion in [forem.dev](forem.dev) . There we can get more eyes on potential features and highlight any we really would love the community to contribute too. Thanks!

PR #10241, opened 2020-09-08
Add LiquidTagsController

citizen428:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR adds an internal API controller for receiving a list of our custom Liquid tags. This can then be used by the frontend to display a typeahead liquid tag dropdown in new editor.

#872

It's pretty much all covered by the specs.

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/app/controllers/liquid_tags_controller.rb
+++ b/app/controllers/liquid_tags_controller.rb
@@ -1,0 +1,13 @@
+class LiquidTagsController < ApplicationController
+  before_action :authenticate_user!
+
+  FILTER_REGEX = /^(?:NullTag|Liquid::)/.freeze
+
+  def index
+    custom_tags = Liquid::Template.tags.filter_map do |name, tag|
+      name unless tag.match?(FILTER_REGEX)
+    end.sort
+
+    render json: { liquid_tags: custom_tags }
+  end
+end
--- a/spec/requests/liquid_tags_request_spec.rb
+++ b/spec/requests/liquid_tags_request_spec.rb
@@ -6,7 +6,7 @@ RSpec.describe "LiquidTags", type: :request do
       it "returns a list of all custom Liquid tags" do
         get liquid_tags_path
 
-        expect(response.status).to eq 401
+        expect(response).to have_http_status(:unauthorized)
       end
     end
 
--- a/spec/requests/liquid_tags_request_spec.rb
+++ b/spec/requests/liquid_tags_request_spec.rb
@@ -1,4 +1,4 @@
-require 'rails_helper'
+require "rails_helper"
 
 RSpec.describe "LiquidTags", type: :request do
   describe "GET /liquid_tags" do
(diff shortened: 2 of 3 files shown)


app/controllers/liquid_tags_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+  before_action :authenticate_user!
+
+  FILTER_REGEX = /^(?:NullTag|Liquid::)/.freeze
+
+  def index
+    custom_tags = Liquid::Template.tags.filter_map do |name, tag|
+      name unless tag.match?(FILTER_REGEX)
+    end.sort
+
+    render json: { liquid_tags: custom_tags }
```
citizen428, line 11: @nickytonline I'm not sure about you, but I prefer JSON APIs that always return objects, so I decided to return a response that looks like this:

```json
{
  "liqid_tags": [
    "asciinema",
    "blogcast",
    ...
  ]
}
```

app/controllers/liquid_tags_controller.rb
```diff
@@ -0,0 +1,13 @@
+class LiquidTagsController < ApplicationController
+  before_action :authenticate_user!
```
citizen428, line 2: Since this is an internal API and we want to show the result in the editor I decided to only make it accessible to authenticated users.

spec/requests/liquid_tags_request_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    context "when signed in" do
+      let(:user) { create(:user) }
+      before { sign_in(user) }
+
+      it "returns an array of all custom Liquid tags", :aggregate_failures do
+        get liquid_tags_path
+
+        expect(response.status).to eq 200
```
rhymes, line 20: you can use `expect(response).to have_http_status(:ok)`

spec/requests/liquid_tags_request_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe "LiquidTags", type: :request do
+  describe "GET /liquid_tags" do
+    context "when not signed in do" do
+      it "returns a list of all custom Liquid tags" do
+        get liquid_tags_path
+
+        expect(response.status).to eq 401
```
rhymes, line 9: you can use `expect(response).to have_http_status(:unauthorized)`

spec/requests/liquid_tags_request_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    context "when signed in" do
+      let(:user) { create(:user) }
+      before { sign_in(user) }
+
+      it "returns an array of all custom Liquid tags", :aggregate_failures do
+        get liquid_tags_path
+
+        expect(response.status).to eq 200
+        json = JSON.parse(response.body, symbolize_names: true)
```
rhymes, line 21: what about `response.parsed_body` (plus `.with_indifferent_access` if you want to use symbols) ?

rhymes (commented):
Cool! I wonder if we could find a way in the future to send expected arguments as well to the JS

spec/requests/liquid_tags_request_spec.rb
citizen428: Ah, yes. I guess I'm jut used to the codes, but this is probably nicer for many people.

spec/requests/liquid_tags_request_spec.rb
citizen428: `HashWithIndifferentAccess` is quite the ugly hack, I changed it to `response.parsed_body` and string keys.

rhymes (approved):

Link2Twenty:
That's a throwback issue 😮

citizen428:
@Link2Twenty We roughly follow the [Shape Up](https://basecamp.com/shapeup) methodology here, which alternates cycles of project work with cooldown periods. A while ago I went through our open issues and made a list of some that I am interested to tackle during one of the latter and here we are 😃

benhalpern:
Cool!

mstruve (approved):
AWESOME! Love this idea bc I always have to dig into the docs to remember what tags we have.

PR #10245, opened 2020-09-08
Add feed_strategy config

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR adds a `feed_strategy` configuration option, with either "basic" or "optimized", and more possible options in the future.

This solves the issue of the DEV feed really being "optimized" for _a lot_ of users, when a typical forem doesn't really need that algorithm (arguably DEV doesn't either), but this gives us some room to offer different approaches for different types of forems, while not exposing too many knobs to the admin-facing API which will let us work with the underlying algorithms over time.

This PR does not include all the possible code that could fall under a choice, but I wanted to provide the basics to ensure new forems have a usable feed right away and the choice in config. We can adjust more of the underlying functionality, including further refactoring in the future.

**Post deploy task: DEV should be updated to "optimized", other forems should get on the default "basic" version. No need for data update script**


app/services/feeds/optimized.rb
```diff
@@ -0,0 +1,279 @@
...
+
+    # Timeframe values from Timeframer::DATETIMES
+    def top_articles_by_timeframe(timeframe:)
+      published_articles_by_tag.where("published_at > ?", Timeframer.new(timeframe).datetime)
+        .order(score: :desc).page(@page).per(@number_of_articles)
+    end
+
+    def latest_feed
+      published_articles_by_tag.order(published_at: :desc)
+        .where("featured_number > ? AND score > ?", 1_449_999_999, -20)
```
mstruve, line 45: For my own education, what does this featured_number represent or define?

app/services/feeds/optimized.rb
```diff
@@ -0,0 +1,279 @@
...
+      stories
+    end
+
+    def more_experience_level_weight_experiment
+      @experience_level_weight = 3
+      _featured_story, stories = default_home_feed_and_featured_story(user_signed_in: true)
+      stories
+    end
+
+    def mix_of_everything_experiment
```
mstruve, line 118: We have been running these experiments for a while now. I think its time to determine which one or few are the most ideal and stick with those. That will allow us to hon in and optimize for those feed cases. What is the plan for doing that?

app/services/feeds/optimized.rb
```diff
@@ -0,0 +1,279 @@
+module Feeds
+  class Optimized
```
mstruve, line 2: "Optimized" feels a little vague here. Is this optimized for performance? Is it optimized for user experience? user preference? etc. 

Maybe something along these lines that just gives a little more color as to what we are optimizing for.
`HighVolumeOptimized` `Articles::VolumeOptimized`

I would also vote to keep `Articles` in the namespace. I think that makes it very clear what is in the Feed. In the future I could see having a user feed, event feed, comment feed etc and then we would have to redo this.

mstruve (commented):
Few general comments but I am cool with this if/else approach for now.

app/services/feeds/optimized.rb
benhalpern: `self.featured_number = Time.current.to_i` It's basically a number where the bigger it is the more recently it's been published. The name comes from early days when we'd more manually feature something so the default will be the time, but it's more just a number to move up if something should move up. Basically the higher the featured number the more likely it is to be featured in the home feed.

app/controllers/stories/feeds_controller.rb
```diff
@@ -27,23 +27,49 @@ def show
...
+                  signed_out_base_feed
                 end
       ArticleDecorator.decorate_collection(stories)
     end
 
-    def ab_test_user_signed_in_feed(feed)
+    def signed_in_base_feed
+      if SiteConfig.feed_strategy == "basic"
+        Feeds::Basic.new(user: current_user, page: @page, tag: params[:tag]).feed
+      else # optimized
```
vaidehijoshi, line 45:
```suggestion
      else
```
I think that the method name describes this well enough that the comment is redundant here, and on line 53

app/services/feeds/optimized.rb
vaidehijoshi: 100% agree with @mstruve on the naming of this service here

app/services/feeds/optimized.rb
vaidehijoshi, line 45: Where did we come up with `1_449_999_999` and `-20`? They feel a bit magic-numbery to me. Also, keeping in line with Forem generalization work, is this something that we want to define "higher up" as a constant (or maybe make configureable at some point?), the way that we do with `RANDOM_OFFSET_VALUES`?

app/services/feeds/optimized.rb
```diff
@@ -0,0 +1,279 @@
...
+    RANDOM_OFFSET_VALUES = [0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 2, 3, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13].freeze
```
vaidehijoshi, line 3: It's not clear to me why we have multiple instances of the same number in this constant, do you happen to know why?

app/services/feeds/optimized.rb
vaidehijoshi: I also am curious if we have any metrics around these experiments; if not, do we even have enough information/data to know which of these actually _sticks_?

joshpuetz (changes requested):
Definitely would like to see documentation updated for this (if we have admin docs somewhere).

app/services/feeds/optimized.rb
citizen428: I asked this before: we generate a random index into this array, so this skews the probabilities in favor of certain numbers.

app/controllers/stories/feeds_controller.rb
```diff
@@ -27,23 +27,49 @@ def show
...
                 else
-                  feed.default_home_feed(user_signed_in: user_signed_in?)
+                  signed_out_base_feed
                 end
       ArticleDecorator.decorate_collection(stories)
     end
 
-    def ab_test_user_signed_in_feed(feed)
+    def signed_in_base_feed
+      if SiteConfig.feed_strategy == "basic"
```
citizen428, line 43: I'd really prefer if we could do something like this in the long run (can be a follow-up PR):

1. move the top-level methods into the service classes: `Feeds::Basic.signed_in_base_feed`, `Feeds::Optimized.signed_out_base_feed`, etc.
2. Add a method here for picking the correct strategy, e.g. 
    ```ruby
    def strategy
      Feed.const_get(SiteConfig.feed_strategy.classify)
    end
   ```
3. Use the strategies here:
    ```ruby
    def  signed_in_base_feed
      strategy.signed_in_base_feed
    end

    # or even
    delegate :signed_in_base_feed, :signed_out_base_feed, to: :strategy
    ```

This way we can add more strategies without having to touch this controller.

app/lib/constants/site_config.rb
```diff
@@ -97,6 +97,10 @@ module SiteConfig
         description: "Used as the site favicon",
         placeholder: "https://image.url"
       },
+      feed_strategy: {
+        description: "Determines the main feed algorithm approach the app takes (basic or optimized)",
```
citizen428, line 101: This message might be confusing in the admin UI since people have no idea what basic and optimized mean. If you give me these options I'd definitely pick optimized without further info. 😉 

image

app/services/feeds/basic.rb
```diff
@@ -0,0 +1,42 @@
...
+  class Basic
+    def initialize(user: nil, number_of_articles: 25, page: 1, tag: nil)
+      @user = user
+      @number_of_articles = number_of_articles
+      @page = page
+      @tag = tag
+    end
+
+    def feed
+      articles = Article.published.order("hotness_score DESC")
```
citizen428, line 11: Since we're already breaking this over multiple lines, maybe we should do the same for `order`? Also: 

```suggestion
      articles = Article.published.order(hotness_score: :desc)
```

app/services/feeds/basic.rb
```diff
@@ -0,0 +1,42 @@
...
+        .where("score >= 0")
```
citizen428, line 12: If we want to lean fully into the AR DSL:

```suggestion
        .where(score: 0..)
```

app/services/feeds/optimized.rb
```diff
@@ -0,0 +1,279 @@
...
+      @tag_weight = 1 # default weight tags play in rankings
+      @comment_weight = 0 # default weight comments play in rankings
+      @experience_level_weight = 1 # default weight for user experience level
+    end
+
+    def self.find_featured_story(stories)
+      featured_story =  if stories.is_a?(ActiveRecord::Relation)
+                          stories.where.not(main_image: nil).first
+                        else
+                          stories.detect { |story| !story.main_image.nil? }
```
citizen428, line 20: IMHO this reads better than the negated version:

```suggestion
                          stories.detect { |story| story.main_image.present? }
```

app/services/feeds/optimized.rb
citizen428: 👍 for making them constants, ideally with descriptive names and comments if necessary.

app/services/feeds/optimized.rb
```diff
@@ -0,0 +1,279 @@
...
+    def user_following_org_ids
+      @user_following_org_ids ||= (@user&.cached_following_organizations_ids || [])
+    end
+
+    def user_following_users_ids
+      @user_following_users_ids ||= (@user&.cached_following_users_ids || [])
+    end
+
+    def first_half(array)
+      array[0...(array.length / 2)]
```
citizen428, line 272: We can lean a bit more on Ruby here:

```suggestion
      array.first(array.length / 2)
```

app/services/feeds/optimized.rb
```diff
@@ -0,0 +1,279 @@
...
+    end
+
+    def last_half(array)
+      array[(array.length / 2)..array.length]
```
citizen428, line 276: We don't need to explicitly specify the end anymore:

```suggestion
      array[(array.length / 2)..]
```

benhalpern:
I made a number of changes based on feedback, including a page in documentation.

I'd still like to limit the scope of this PR and follow up with more work as long as it is passable, because I think it would be a good thing to get this concept into Forems in general and we can update the functionality for the different ones thereafter and iterate on this idea.

I think the #1 priority for the feed has to be continual change, vs ever being "done", and there are definitely some opportunities for change on everything we have and to pluck some winning concepts from the a/b testing and introduce new ideas, but I think it's best to get this in first and then grow from there.

app/services/feeds/optimized.rb
mstruve: @benhalpern what is the plan going forward for these experiments?

citizen428 (approved):

atsmith813 (approved):

app/services/articles/feeds/basic.rb
```diff
@@ -0,0 +1,45 @@
...
+        @user = user
+        @number_of_articles = number_of_articles
+        @page = page
+        @tag = tag
+      end
+
+      def feed
+        articles = Article.published
+          .order(hotness_score: :desc)
+          .where(score: 0..)
```
vaidehijoshi, line 14: TIL: ruby has an infinite range!

vaidehijoshi (approved):

docs/technical-overview/readme.md
```diff
@@ -2,6 +2,7 @@
 items:
   - stack.md
   - architecture.md
+  - feed_algorithm.md
```
benhalpern, line 5:
```suggestion
  - feed.md
```

PR #10251, opened 2020-09-08
Feature:Add ENV to Skip Fastly Config Update

mstruve:
- [x] Feature

## Description
In the event of a Fastly outage we need to be able to skip this config update step to allow deploys to happen.

### DIFF
--- a/lib/tasks/fastly.rake
+++ b/lib/tasks/fastly.rake
@@ -1,19 +1,21 @@
 namespace :fastly do
   desc "Update Fastly configs"
   task update_configs: :environment do
-    fastly_credentials = %w[
-      FASTLY_API_KEY
-      FASTLY_SERVICE_ID
-    ]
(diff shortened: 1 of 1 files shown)


maestromac (approved):

atsmith813 (approved):

vaidehijoshi (approved):

lib/tasks/fastly.rake
```diff
@@ -1,19 +1,21 @@
 namespace :fastly do
   desc "Update Fastly configs"
   task update_configs: :environment do
-    fastly_credentials = %w[
-      FASTLY_API_KEY
-      FASTLY_SERVICE_ID
-    ]
+    unless ENV["SKIP_FASTLY_CONFIG_UPDATE"] == "true"
```
vaidehijoshi, line 4: we may want to document that this variable can be used to skip through fastly

mstruve: Good call, I will throw it in the on-call docs.

PR #10252, opened 2020-09-08
Give Admins the Ability to Add Co-Authors to Posts

juliannatetreault:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR gives admins the ability to add co-authors to posts via a user's ID within `/admin/articles`. To accomplish this, this PR adds `second_user_id` and `third_user_id`s to the `article_params` in the `Admin::Articles::Controller`, adds `second_user_id` and `third_user_id` to `:limited_columns_internal_select` within `article.rb`, adds "Second Author ID" and "Third Author ID" form fields to `_individual_article.html.erb` to match the current "Author ID" form field, and adds tests around adding an additional co-author/additional co-authors to `admin/articles_spec.rb`.

Relates to a private issue.

To QA this PR, first ensure that you have `:super_admin` permissions, then do the following:

- Navigate to `/admin/articles` in your browser:
Screen Shot 2020-09-08 at 1 44 22 PM

- Add a "Second Author ID" and a "Third Author ID" to an article of your choice and click "Submit":
Screen Shot 2020-09-08 at 1 44 55 PM

- Click on the article title to view the article. Once on the article show page, ensure that the correct users associated with the IDs that you provided in the "Second Author ID" and "Third Author ID" form fields show up:
Screen Shot 2020-09-08 at 1 45 45 PM

- Click on the author names to ensure that you are brought to the correct user's profile:
Screen Shot 2020-09-08 at 1 45 58 PM
Screen Shot 2020-09-08 at 1 46 10 PM

- Try to break things! 🛠️ 

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

N/A

Pretending to write something down

### DIFF
--- a/app/controllers/admin/articles_controller.rb
+++ b/app/controllers/admin/articles_controller.rb
@@ -45,6 +45,8 @@ module Admin
       article.boosted_additional_articles = article_params[:boosted_additional_articles].to_s == "true"
       article.boosted_dev_digest_email = article_params[:boosted_dev_digest_email].to_s == "true"
       article.user_id = article_params[:user_id].to_i
+      article.second_user_id = article_params[:second_user_id].to_i
+      article.third_user_id = article_params[:third_user_id].to_i
       article.update!(article_params)
       render body: nil
     end
@@ -129,6 +131,8 @@ module Admin
                           main_image_background_hex_color
                           featured_number
                           user_id
+                          second_user_id
+                          third_user_id
                           last_buffered]
       params.require(:article).permit(allowed_params)
     end
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -144,7 +144,7 @@ class Article < ApplicationRecord
            :video_thumbnail_url, :video_closed_caption_track_url, :social_image,
            :published_from_feed, :crossposted_at, :published_at, :featured_number,
            :last_buffered, :facebook_last_buffered, :created_at, :body_markdown,
-           :email_digest_eligible, :processed_html)
+           :email_digest_eligible, :processed_html, :second_user_id, :third_user_id)
   }
 
   scope :boosted_via_additional_articles, lambda {
--- a/app/views/admin/articles/_individual_article.html.erb
+++ b/app/views/admin/articles/_individual_article.html.erb
@@ -85,6 +85,16 @@
         <input id="author_id_<%= article.id %>" class="form-control" size="6" name="article[user_id]"
           value="<%= article.user_id %>">
       </div>
+      <div class="form-group col">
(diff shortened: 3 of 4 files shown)


Zhao-Andy (approved):
Looks great!

app/controllers/admin/articles_controller.rb
```diff
@@ -45,6 +45,8 @@ def update
       article.boosted_additional_articles = article_params[:boosted_additional_articles].to_s == "true"
       article.boosted_dev_digest_email = article_params[:boosted_dev_digest_email].to_s == "true"
       article.user_id = article_params[:user_id].to_i
+      article.second_user_id = article_params[:second_user_id].to_i
+      article.third_user_id = article_params[:third_user_id].to_i
```
vaidehijoshi, line 49: I don't think this is in the scope of _this_ PR, but seeing these changes here seems like a indicator that we probably want to move away from having explicit columns for each co-author id (`second_user_id`, `third_user_id`, etc.) and opt for a more flexible solution: `co_author_ids`. It would be nice for the controller to send along params like this:
```ruby
params: { co_author_ids: [2, 3] }
```
so that we can ask the article for its coauthors like so:
```ruby
article.co_author_ids
=> [2, 3]
```

which is more flexible than adding a column each time we want to add a co-author id :) it would also mean that we don't need to send along 2 params for effectively the same concept like we are doing in this PR.

You can imagine that the moment we want to allow authors or admins to add a _fourth_ co-author, this is going to be painful to do the way that we currently have it set up.

juliannatetreault: This is a great suggestion @vaidehijoshi , but definitely a bit out of scope for this feature since the need for it is pretty high. I've created an issue for this on my team's cool-down project board so that we can circle back around to it, though. 😊

fdoxyz: I really like this refactor idea

vaidehijoshi: > I've created an issue for this on my team's cool-down project board so that we can circle back around to it

@juliannatetreault Yep, I think that's the right call; again, not a blocker, as I mentioned in my comment, but worth refactoring in the near term since we're already in the code.

fdoxyz (approved):
I think the refactor suggested sounds great but also agree it's out of scope for this PR. I tried it out locally and works great thank you @juliannatetreault!

PR #10256, opened 2020-09-09
Article refactoring

citizen428:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I originally came to this model to move some things into the decorator, but that turned into a dead end since several methods were calling into each other and with some of them also being called in the search serializer it was too hard to untangle them. So instead I decided to just clean up some code.

Cooldown fun.

Nothing specific. This is a refactoring, so I just make sure that we have specs for code I'm changing and that they stay green.

## Added tests?

- [ ] yes
- [X] no, because they aren't needed (see above)
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -40,10 +40,8 @@ class Article < ApplicationRecord
   has_many :rating_votes, dependent: :destroy
   has_many :top_comments,
            lambda {
-             where(
-               "comments.score > ? AND ancestry IS NULL and hidden_by_commentable_user is FALSE and deleted is FALSE",
-               10,
-             ).order("comments.score" => :desc)
+             where(comments: { score: 11.. }, ancestry: nil, hidden_by_commentable_user: false, deleted: false)
+               .order("comments.score" => :desc)
            },
            as: :commentable,
            inverse_of: :commentable,
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -353,13 +353,11 @@ class Article < ApplicationRecord
   end
 
   def video_duration_in_minutes
-    minutes = video_duration_in_minutes_integer
-    seconds = video_duration_in_seconds.to_i % 60
-    seconds = "0#{seconds}" if seconds.to_s.size == 1
+    duration = ActiveSupport::Duration.build(video_duration_in_seconds.to_i).parts
+    minutes_and_seconds = format("%<minutes>02d:%<seconds>02d", duration)
+    return minutes_and_seconds if duration[:hours] < 1
 
-    hours = (video_duration_in_seconds.to_i / 3600)
-    minutes = "0#{minutes}" if hours.positive? && minutes < 10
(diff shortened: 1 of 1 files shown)


app/models/article.rb
```diff
@@ -355,13 +353,11 @@ def cloudinary_video_url
   end
 
   def video_duration_in_minutes
-    minutes = video_duration_in_minutes_integer
-    seconds = video_duration_in_seconds.to_i % 60
-    seconds = "0#{seconds}" if seconds.to_s.size == 1
+    duration = ActiveSupport::Duration.build(video_duration_in_seconds.to_i).parts
```
citizen428, line 356: This takes a value in seconds and turns it into a useful hash:

```ruby
ActiveSupport::Duration.build(4567).parts
#=> { hours: 1, minutes: 16, seconds: 7 }
```

app/models/article.rb
```diff
@@ -355,13 +353,11 @@ def cloudinary_video_url
...
+    minutes_and_seconds = format("%<minutes>02d:%<seconds>02d", duration)
```
citizen428, line 357: This is standard Ruby string formatting, but using some of the more advanced options:

1. `"%02d"` means "take a number and format it as a 2 digit string, with leading zero added if necessary".
    ```ruby
    format("%02d", 1)
    #=> "01"
    format("%02d", 12)
    #=> "12"
    ```
2. Apart from individual arguments, format also allows a hash. Names in angle brackets refer to hash keys, so  `"%<minutes>02d` formats the value of `duration[:minutes]` as described above.

citizen428:
I guess we can ignore CodeClimate, as the 2 errors have nothing to do with the contents of this PR.

rhymes:
@citizen428 I guess brakeman got updated on their end, I wonder if we could fix them nonetheless because I reckon they are going to appear on contributor's PRs at some point as well.

I believe we can "fast track" this PR

app/models/article.rb
```diff
@@ -355,13 +353,11 @@ def cloudinary_video_url
...
+    return minutes_and_seconds if duration[:hours] < 1
 
-    hours = (video_duration_in_seconds.to_i / 3600)
-    minutes = "0#{minutes}" if hours.positive? && minutes < 10
-    hours < 1 ? "#{minutes}:#{seconds}" : "#{hours}:#{minutes}:#{seconds}"
+    "#{duration[:hours]}:#{minutes_and_seconds}"
   end
 
   def video_duration_in_minutes_integer
```
rhymes, line 363: can we move this in the serializer? It's only called there... The serializer already contains some serialization logic.

rhymes (approved):
Left a note about moving `video_duration_in_minutes_integer` but other than that 🆗

benhalpern (approved):

app/models/article.rb
citizen428: Since this got merged already, I'll check that tomorrow.

PR #10257, opened 2020-09-09
Remove profile validations from user

citizen428:

- [X] Refactor

## Description

This PR removes all validations for profile fields on `User`. Context for those who are not closely following along on the generalization effort:

1. Validating these on user no longer makes sense, as they now belong to profile (also a source of recent bugs, which was addressed in #9912.
2. In the new system, a validation would belong to a specific `ProfileField`. However, it's not clear how we would expose this to admins without building a complex validation generation UI. 
3. For now, we decided to just forgo these validations completely (see discussions in the `#profile-generalization` Slack channel). 

Part of #6994

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed


mstruve (approved):
I actually really support this. I think the URL validation is a little bit superfluous in that the person is only hurting themselves if they put in an invalid URL so probably not something we should worry about. 

What's your thought on things like `text_color_hex` and `bg_color_hex` since we actually have to read and use those ones? Fallback on something that is valid if the hex is not?

citizen428:
> What's your thought on things like text_color_hex and bg_color_hex since we actually have to read and use those ones? Fallback on something that is valid if the hex is not?

In this specific case we can just use the correct input type (`<input type="color">`) since `color_field` is a dedicated `ProfileField` type. For other cases, we are still exploring/gathering ideas.

jacobherrington (approved):
This makes sense to me after reading up on the context. :+1:

PR #10263, opened 2020-09-09
Reintroduce sticky_nav changes (ab81f36 & 7f92855)

maestromac:
- [x] Refactor

## Description
https://github.com/forem/forem/pull/10168 and https://github.com/forem/forem/pull/10055 were both previously reverted because they invalidate the sticky_nav fragment cache which causes latency. Now that we know the cause, we know that the latency will smoothen itself out eventually. we're going to re-add it back in and I'll actively monitor the latency.

Should be able to navigate site normally.

## Added tests?

## Added to documentation?
- [x] no documentation needed

Will need active monitoring


mstruve (approved):
🔥

PR #10267, opened 2020-09-09
Proper cache delete_matched

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We had a prior functionality which is meant to delete the core fragment caches that change on every deployment, or in this case, when we update config and want a fresher cache.

This updates it to work as intended (and also tacked on another page purge)...

The original PR did not break anything, but did not appropriately follow the `delete_matched` functionality which requires wildcards to work properly like this. The included test is a basic regression test, not _directly_ exercise this functionality, but at least adds a bit more coverage.


Zhao-Andy (approved):
LGTM!

PR #10268, opened 2020-09-10
Move video_duration_in_minutes_integer to serializer

citizen428:

- [X] Refactor
- [X] Bug Fix

## Description

In #10256 I refactored `video_duration_in_minutes`. Since this now no longer relies on `video_duration_in_minutes_integer` it was [suggested to move the latter into the search serializer](https://github.com/forem/forem/pull/10256#discussion_r485383795), since that's the only place where it's used.

While doing so I also realized that the method was wrong. Here's the old definition:

```ruby
(video_duration_in_seconds.to_i / 60) % 60
```

This will be wrong for any duration > 1 hour, because of the modulo operation:

```ruby
duration = (1.hour + 1.minute).to_i # should be 61 minutes
#=> 3660
(duration / 60) % 60 # modulo discards the hours
#=> 1 
```

To go from seconds to minutes we only need to divide by 60, so I removed the modulo.

n/a

Nothing specific

## Added tests?

- [X] yes

## Added to documentation?

- [X] no documentation needed

### DIFF
--- a/lib/data_update_scripts/20200911045602_reindex_articles_with_videos.rb
+++ b/lib/data_update_scripts/20200911045602_reindex_articles_with_videos.rb
@@ -1,0 +1,8 @@
+module DataUpdateScripts
+  class ReindexArticlesWithVideos
+    def run
+      articles = Article.where.not(video: nil).or(Article.where.not(video: ""))
+      articles.find_each(&:index_to_elasticsearch_inline)
+    end
+  end
+end
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -360,10 +360,6 @@ class Article < ApplicationRecord
     "#{duration[:hours]}:#{minutes_and_seconds}"
   end
 
-  def video_duration_in_minutes_integer
-    (video_duration_in_seconds.to_i / 60) % 60
-  end
-
   def update_score
     new_score = reactions.sum(:points) + Reaction.where(reactable_id: user_id, reactable_type: "User").sum(:points)
     update_columns(score: new_score,
--- a/app/serializers/search/article_serializer.rb
+++ b/app/serializers/search/article_serializer.rb
@@ -12,7 +12,9 @@ module Search
     # however, it really is a string in the format 00:00 which is why we
     # added an extra field to handle that string
     attribute :video_duration_string, &:video_duration_in_minutes
-    attribute :video_duration_in_minutes, &:video_duration_in_minutes_integer
+    attribute :video_duration_in_minutes do |article|
+      article.video_duration_in_seconds.to_i / 60
+    end
 
     attribute :readable_publish_date_string, &:readable_publish_date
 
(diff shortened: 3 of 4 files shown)


rhymes (commented):
Great catch! Should we reindex all articles with a video to fix the length?

mstruve (approved):
NICE! would probably be a good idea to add a little data update script to reindex articles. If there's a way to only reindex those with videos that would be ideal and save us some work. 

As an aside, we should add a reindex button to admin to reindex things like articles or users that way every time we tweak a doc like this we dont necessarily have to do a data update script if we dont think many forems would be affected and issue a changelog to reindex for those forems that our. Might be more manual process than we want but just a thought :)

rhymes:
> NICE! would probably be a good idea to add a little data update script to reindex articles. If there's a way to only reindex those with videos that would be ideal and save us some work.

this Blazer query about finding 0 length videos might help https://dev.to/admin/blazer/queries/6-published-articles-with-0-length-videos

lib/data_update_scripts/20200911045602_reindex_articles_with_videos.rb
```diff
@@ -0,0 +1,8 @@
+module DataUpdateScripts
+  class ReindexArticlesWithVideos
+    def run
+      articles = Article.where.not(video: nil).or(Article.where.not(video: ""))
```
citizen428, line 4: @rhymes Do you think this condition is sufficient?

citizen428:
I added a data update script now. I got a bit stuck on the spec and it's Friday afternoon, so I decided to just commit. I think the data update script is straight-forward enough to not need one, if you feel differently I can add it on Monday.

lib/data_update_scripts/20200911045602_reindex_articles_with_videos.rb
rhymes: Wonder if the query plan is different from `Article.where.not(video: ["", nil])` but yeah! that's fine.

Our `.with_video` scope is a bit more specific but I think this should cover also those that aren't shown

https://github.com/forem/forem/blob/df245e7e84b38260be4b7fe4ef359ee539b8a40b/app/models/article.rb#L184-L189

rhymes (approved):
👍

lib/data_update_scripts/20200911045602_reindex_articles_with_videos.rb
citizen428: The generated queries are slightly different, but logically equivalent:

`WHERE ("articles"."video" IS NOT NULL OR "articles"."video" != $1)` vs `WHERE NOT (("articles"."video" = $1 OR "articles"."video" IS NULL))`

I decided against the scope because I figured we want to reindex this for all articles with videos.

rhymes: This PR should be safe for deployment, we can check on Blazer later if it has fixed the issues

mstruve (approved):
This query ran quick and no problem in a console. Should be good to go. 
```
irb(main):002:0>  Article.where.not(video: nil).or(Article.where.not(video: "")).count
=> 1622
```

PR #10272, opened 2020-09-10
WIP: /profile/settings page rendering

Ridhwana:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The profile settings page now renders dynamically from the `ProfileFieldGroup` model and `ProfileField` model. 
Updates to `ProfilesController` are sent to `/profiles/:id`. We re-use the `Profiles::Update` service to split the user and the generalized profile fields to update the relevant models. 

Need to:

- [ ] Add tests 
- [x] Fix the checkboxes and styling on the ui
- [x] Fix the checkboxes, it's currently not working 
- [ ] Fix the upload of the profile image (move the validation etc. as well)
- [x] Show feedback on the page when success or errors occurs
- [x] Need to check the success of both the profile updates and the user updates
- [ ] and maybe some more...

Part of #6994

_Will add this afterwards_

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

alt_text


app/services/profiles/update.rb
```diff
@@ -35,8 +48,8 @@ def call
       self
     end
 
-    def success?
-      @success
+    def update_user
+      @success = true if @user.update(@updated_user_attributes.to_h)
```
Ridhwana, line 52: This won't work, need to take into account the success of both updates on the profile sync and on the user model directly. Was busy tackling this.

Ridhwana (commented):
@citizen428 I still need to do some stuff on this PR, namely:  

- [ ] Add tests
- [ ]  Fix the checkboxes and styling on the ui
- [ ]  Fix the checkboxes, it's currently not working
- [ ]  Fix the upload of the profile image (move the validation etc. as well)
- [ ]  Show feedback on the page when success or errors occurs
- [ ]  Need to check the success of both the profile updates and the user updates
- [ ]  clean up code...

I wanted to make sure this is the approach that you suggested, i.e. regarding service, looks okay before I continue further

app/controllers/profiles_controller.rb
```diff
@@ -1,13 +1,16 @@
...
   def update
-    Profiles::Update.call(current_user.profile, update_params)
+    Profiles::Update.call(current_user, update_params)
   end
 
   private
 
   def update_params
-    params.require(:profile).permit(*Profile.attributes)
+    params.permit(profile: Profile.attributes!, user: ALLOWED_PARAMS)
```
citizen428, line 14: I was just guessing in which form the params will arrive, but this makes sense. I'd probably rename `ALLOWED_PARAMS` to `ALLOWED_USER_PARAM<S` to be more clear.

app/services/profile_fields/add_link_fields.rb
```diff
@@ -8,7 +8,7 @@ class AddLinkFields
       field "StackOverflow URL", :text_field, placeholder: "https://stackoverflow.com/users/..."
       field "LinkedIn URL", :text_field, placeholder: "https://www.linkedin.com/in/..."
       field "Behance URL", :text_field, placeholder: "https://www.behance.net/..."
-      field "Dribble URL", :text_field, placeholder: "https://dribble.com/..."
+      field "Dribbble URL", :text_field, placeholder: "https://dribbble.com/..."
```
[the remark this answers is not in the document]
citizen428: The most correct way would be a data update script that idempotently updates the existing field.

app/services/profiles/update.rb
citizen428: Should we update the user at all if the profile update didn't succeed? While they live on different models, the update process should be atomic and either succeed for both or neither. When doing a normal update everything gets rolled back when one field is invalid, I think the same should be true here too. 

We can potentially still do the following though:
1. assign the attributes
2. run validations
3. collect the errors to return to the user

Does this make sense? If not I can work on it on Monday.

app/views/users/_profile.html.erb
```diff
@@ -43,158 +35,30 @@
...
-      <%= f.url_field :mastodon_url, class: "crayons-textfield", placeholder: "https://..." %>
-    </div>
-
-    <div class="crayons-field">
-      <%= f.label :twitch_url, "Twitch profile URL", class: "crayons-field__label" %>
-      <%= f.url_field :twitch_url, class: "crayons-textfield", placeholder: "https://..." %>
-    </div>
-  </div>
+  <% @profile_field_groups.each do |group| %>
+    <% unless group.profile_fields.empty? %>
```
citizen428, line 45: We can potentially add a scope for this on on `ProfileFieldGroup` and use that in the controller in order to never even return empty groups from the DB. 

I added an empty group called `Test` locally for testing this: 

```ruby
ProfileFieldGroup.count
#=> 6
ProfileFieldGroup.includes(:profile_fields).count
#=> 6

# Using an inner join to filter out field-less groups
groups = ProfileFieldGroup.joins(:profile_fields).includes(:profile_fields).uniq
groups.count
#=> 5
# Fields are preloaded
groups.first.profile_fields
# no query here
#=> [
  [0] #<ProfileField:0x000000011e997770> {
            attribute_name: "website_url",
```

app/views/users/_profile.html.erb
```diff
@@ -43,158 +35,30 @@
...
+      <div class="crayons-card mb-6 grid grid-flow-row gap-6 p-6">
+        <h2><%= group.name %></h2>
+        <% if group.description %>
```
citizen428, line 48: I think we better use `group.description.present?` here since empty strings are truthy in Ruby.

citizen428 (approved):

citizen428 (changes requested):
Sorry, I had accidentally removed this previously ☕

app/controllers/profiles_controller.rb
Ridhwana: ✅

app/services/profiles/update.rb
Ridhwana: I tried something @citizen428, maybe you can have a look and correct where necessary or provide suggestions? Even though i added a flash for the error and success and its passing through the variables correctly, I couldn't get the flash to show up on the page, maybe coz i'm rendering through the users controller and updating through the profiles controller. I dunno, I'm looking into it and learning more.

app/views/users/_profile.html.erb
Ridhwana: ✅

app/views/users/_profile.html.erb
Ridhwana: ✅

citizen428:
@Ridhwana

* We're now passing `current_user` to the service object, but the argument wasn't renamed to reflect this. This is a bit confusing:

    ```ruby
      class Update
        def self.call(profile, updated_attributes = {})
          new(profile, updated_attributes).call
        end
    
        # BUT
        def initialize(user, updated_attributes)
          @user = user
          @profile = user.profile
          ...
    ```
    I attached a patch for this.
2. I also refactored the whole service object a bit (independent of your changes), I also updated the specs (reflecting the profile/user attributes separation introduced in this PR), but you may want to add some more. This is the second patch.

Note on patches:

1. Download [pr-10272-patches.zip](https://github.com/forem/forem/files/5222359/pr-10272-patches.zip) (GH doesn't allow to attach `.patch` files directly, even though they are just txt).
2. Extract them in your source directory
3. Apply them:
    ```
    git apply 0001-Rename-argument.patch
    git apply 0002-Refactor-service-object.patch
    ```

app/services/profiles/update.rb
citizen428: Re the flash: you're setting it, but not rendering anything after (this is a patch request). This should follow the usual Rails redirect on success and re-render on error idiom:

```ruby
    if update_result.success?
      flash[:success] = "Your profile has been updated"
      redirect_to ...
    else
      flash[:error] = "Error: #{update_result.error_message}"
      render ...
    end
```

citizen428:
@Ridhwana I still can't push directly to your branch. I now configured my vim plugin correctly to ask for username/password in case of HTTPS clone URLs, but that's not it, it also doesn't work from the CLI:

```
Username for 'https://github.com': citizen428
Password for 'https://citizen428@github.com':
remote: Invalid username or password.
fatal: Authentication failed for 'https://github.com/Ridhwana/dev.to.git/'
```

The username and password are definitely correct, they come from 1password.

Normally when you open a PR against someone else’s repo, there’s a checkbox that says something like “Allow edits from maintainers” or similar. Do you see this when you open a PR from your fork or do they skip that since you’re also a member of the Forem org?

citizen428:
Anyway, I'll just provide patches again. 

* `0001-Fix-CodeClimate-warning-and-rearrange-code.patch`: Fixes the CodeClimate issues and re-arranges some code in the service object.
* `0002-Fix-service-object-add-missing-attribute.patch`: fixes a bug in the service object and adds `skills_languages: :mostly_work_with` to `Profile::MAPPED_ATTRIBUTES`.  This probably means that since we merged #9912 we did no longer sync this correctly, so we may need a data update script for it.
* `0003-Fix-flash-and-add-redirect.patch`: fixes the flash (we need `:settings_notice` instead of `:success` and adds the redirect to `user_settings_path`. The flashes now render correctly:

Screen Shot 2020-09-16 at 11 48 19
Screen Shot 2020-09-16 at 11 53 42

Here's the patch ZIP:
[pr-10272-patches-2.zip](https://github.com/forem/forem/files/5229764/pr-10272-patches-2.zip)

Ridhwana:
> @Ridhwana I still can't push directly to your branch. I now configured my vim plugin correctly to ask for username/password in case of HTTPS clone URLs, but that's not it, it also doesn't work from the CLI:
> 
> ```
> Username for 'https://github.com': citizen428
> Password for 'https://citizen428@github.com':
> remote: Invalid username or password.
> fatal: Authentication failed for 'https://github.com/Ridhwana/dev.to.git/'
> ```
> 
> The username and password are definitely correct, they come from 1password.
> 
> Normally when you open a PR against someone else’s repo, there’s a checkbox that says something like “Allow edits from maintainers” or similar. Do you see this when you open a PR from your fork or do they skip that since you’re also a member of the Forem org?

This is so weird, I made sure that box was checked. 
Screenshot 2020-09-16 at 08 29 57. I'm not sure what else it could be 🤔

citizen428:
> I'm not sure what else it could be 🤔

Probably me, no worries.

citizen428:
For the record, since I confused this several times in the past: `git am` fill apply patches generated with `git format-patch` as if the patch author committed directly to the repo, `git apply` (as I erronously suggested above) will not.

citizen428:
I think functionality-wise this is in a good place now, if I get around to it tomorrow I'll look into the specs, otherwise it'll have to wait for Monday morning.

citizen428:
Note: I now disabled `spec/system/user_selects_looking_for_work_spec.rb`. I first tried to fix it, then realized this is called from `UsersController#update`, which the current form doesn't hit. I also added a comment about this method in general and started a related Slack discussion:

```
# TODO: [@forem/oss] This doesn't make sense for non DEV communities. We
# should either remove it or special case it.
```

@Ridhwana I noticed there are some other things happening in `UsersController#update` which we probably need to move, e.g. 

```ruby

if @user.export_requested?
  notice += " The export will be emailed to you shortly."
    ExportContentWorker.perform_async(@user.id)
   end
end

cookies.permanent[:user_experience_level] = @user.experience_level.to_s if @user.experience_level.present?
```

We should talk about this a bit more, I don't think everything needs to go into the `Profile::Update` service, i.e. content export could be its own miniform that triggers a different service.

citizen428:
I marked "Fix the checkboxes, it's currently not working" as done, since they seemed to work fine when I tried them locally.

citizen428:
I added two data update scripts now:

1. `FixDribbbleUrlProfileField`: this fixes the Dribbble-related fields by adding the third 'b' and makes sure that all users that have a Dribble URL get them correctly added to their profiles. I checked with Blazer, we currently have ~1.7k users like that:
    ```sql
    select count(*) from users where not (dribbble_url = '' or dribbble_url is null);
    
    count
    --
   1,674
    ```

2. `BackfillProfileSkillsLanguages`: this field was missing from the mapping and since we merged #9912 we may not have synced it correctly. In order not to lose data we backfill the values now. This affects ~61k users:

    ```sql
    select count(*) from users where not (mostly_work_with = '' or mostly_work_with is null);
    
    count
    --
    60,703
    ```

app/services/profile_fields/add_link_fields.rb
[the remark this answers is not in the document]
citizen428: A data update script was added for this.

citizen428:
@Ridhwana I'm wasn't quite sure about the following two todo items:

* "Fix the checkboxes, it's currently not working":  They seem to work fine when I tried them locally, so I marked this as done. Please confirm.
* "Fix the upload of the profile image (move the validation etc. as well)": this also does seem to work, but I left it as for now since I don't fully understand what you wanted to move.

citizen428:
@Ridhwana I  made the following changes now:

1. Make `sync_to_user` conditional on `update_profile`: `update_profile` now directly returns the result from `@profile.save` and the two are combined by `update_profile && sync_to_user` so the sync only happens if the profile update succeeds. Also `sync_to_user` now returns a boolena too (the value of `@success`).
2. The combined result of `update_profile && sync_to_user` is used to determine if we call the new method `perform_after_update_actions` or report errors to Honeycomb (this part was taken and slightly modified from `UsersController#update`).
3. `perform_after_update_actions` does two things for now: enqueue the RSS update job and touch the `profile_updated_at` attribute of the user.
4. I added setting the experience level in the cookie to `ProfilesController`.

Things in `UsersController#update` that are not taken care of (yet):

* Some tab handling logic because I'm not entirely sure how this works
* The `unless valid_image?` profile image check
* The `config_changed?` check (which just outputs a different notice if the theme was changed). 
* The export. I still think we should make this a different section in the UI and handle it from a different controller action.

Ridhwana:
Yesterday I had to setup my forem fork again due to some laptop issues, so looking at this PR, it's trying to merge from an unknown repository because my fork had been reset. I'm hoping that it works fine when we merge, but I just wanted to highlight it.

citizen428:
Closing this in favor of #10733

PR #10239, opened 2020-09-07
Remove draft articles with duplicate feed source URL and cleanup invalid data: take 2

rhymes:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

**this PR is a clone of https://github.com/forem/forem/pull/10205 that we ended up reverting as it was merged prematurely**

Similar to https://github.com/forem/forem/pull/10138 but based on `feed_source_url`.

To be able to close the loop on #8072 and add a unique index on feed_source_url we need first to be remove duplicates that have the same feed_source_url.

This PR contains scripts to accomplish the following things:

1. remove all drafts that in a partition have the same body. A partition is determined as a group of articles with the same feed source URL
2. remove all drafts that in a partition have different bodies by keeping the most recent one. A partition is determined as a group of articles with the same feed source URL
3. fixes invalid data for `feed_source_url`. Currently, unlike with `canonical_url`, we don't have a URL validator on the field, thus I found a bunch of invalid URLs in production data, see https://dev.to/admin/blazer/queries/220-articles-with-an-invalid-feed-source-url

Please read the SQL queries closely.

If you want to take a look at the data:

- https://dev.to/admin/blazer/queries/218-duplicate-articles-feed_source_url-with-the-same-body
- https://dev.to/admin/blazer/queries/219-duplicate-articles-feed_source_url-with-different-bodies
- https://dev.to/admin/blazer/queries/220-articles-with-an-invalid-feed-source-url

## Related Tickets & Documents

#10138 
#8072 

## QA Instructions, Screenshots, Recordings

Ideally this should be tested on a copy of the production DB. To make sure it worked I downloaded the sample of duplicates https://dev.to/admin/blazer/queries/221-duplicate-articles-feed_source-url-with-the-same-body-ids - imported them locally and ran the scripts to make sure they worked. They obviously present different results as it's a script run on the subset of duplications (so it works because in some cases we have 2 or more duplicates of the same article).

If you want to test it nonetheless:

- download a CSV from https://dev.to/admin/blazer/queries/221-duplicate-articles-feed_source-url-with-the-same-body-ids
- open the SQL console with `psql PracticalDeveloper_development`
- type `\d articles`
- note the names of the foreign key constraints
- drop the foreign key constraints on the `articles` table, otherwise the `user_id` and `organization_id` won't work:

```sql
alter table articles drop constraint fk_rails_3d31dad1cc;
[...]

PR #10273, opened 2020-09-10
Cleanup articles with duplicate user_id, title and body_markdown

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR contains two scripts

1. the first one deletes duplicate draft articles that have the same combination of `user_id`, `title` and `body_markdown`. The algorithm is the same of https://github.com/forem/forem/pull/10138 and https://github.com/forem/forem/pull/10239

1. the second one "merges" published articles:
  - basically we divide the articles in groups (some articles have 1 duplicate, some have more)
  - of all the groups we take the first article
  - we copy over the data from the others onto this first one
  - we destroy the others

This is where the data is:

- https://dev.to/admin/blazer/queries/232-duplicate-draft-articles-body_markdown-user_id-title
- https://dev.to/admin/blazer/queries/235-duplicate-published-articles-body_markdown-user_id-title
- https://dev.to/admin/blazer/queries/236-published-articles-of-users-with-duplicates

Ideally this should be tested on a copy of the production DB. To make sure it worked I downloaded the sample of duplicates from https://dev.to/admin/blazer/queries/236-published-articles-of-users-with-duplicates - imported them locally and ran the scripts to make sure they worked. 

If you want to test it:

- download a CSV from https://dev.to/admin/blazer/queries/236-published-articles-of-users-with-duplicates
- open the SQL console with `psql PracticalDeveloper_development`
- type `\d articles`
- note the names of the foreign key constraints
- drop the foreign key constraints on the `articles` table, otherwise the `user_id` and `organization_id` won't work:

```sql
alter table articles drop constraint fk_rails_3d31dad1cc;
alter table articles drop constraint fk_rails_7809a1a57d;
```

remember that the constraint names might be different than mine. Please double check with `\d articles` in the SQL console and replace them with yours
- import the data from the downloaded CSV:

```sql
copy articles(id,user_id,title,body_html,slug,created_at,updated_at,main_image,description,published,published_at,featured,processed_html,featured
 _number,social_image,body_markdown,organization_id,canonical_url,collection_id,show_comments,main_image_background_hex_color,receive_notifications,approved,published_from_feed,c
 omments_count,reactions_count,video,video_code,video_source_url,video_thumbnail_url,video_closed_caption_track_url,hotness_score,feed_source_url,second_user_id,third_user_id,las
 t_buffered,positive_reactions_count,edited_at,crossposted_at,spaminess_rating,language,facebook_last_buffered,cached_tag_list,path,cached_user_name,cached_user_username,last_com
 ment_at,boost_states,video_state,page_views_count,previous_positive_reactions_count,score,reading_time,comment_template,video_duration_in_seconds,experience_level_rating,experie
 nce_level_rating_distribution,last_experience_level_rating_at,rating_votes_count,originally_published_at,organic_page_views_count,organic_page_views_past_month_count,organic_pag
 e_views_past_week_count,cached_user,cached_organization,archived,any_comments_hidden,nth_published_by_author,comment_score,search_optimized_title_preamble,search_optimized_descr
 iption_replacement,public_reactions_count,previous_public_reactions_count,user_subscriptions_count) from '/path/to/file.csv.csv' DELIMITER ',' csv header;
```

don't forget to replace the path of the file with the full path of the file you downloaded

- count the number of articles, just to have a quick idea if the script has worked
- replace `.destroy_all` with `.delete_all` on line https://github.com/forem/forem/pull/10273/files#diff-ae11923370b79ce32bb9abbeebca964dR43 (the reason for this is that you won't have the same user objects locally, but we need to pass through Rails to make sure everything left attached to articles is actually destroyed)
- run the data update script with `rails data_updates:run` if it's the first time or manually in the console

```ruby
require Rails.root.join("lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb")
DataUpdateScripts::CleanupPublishedArticlesWithDuplicateUserIdTitleBodyMarkdown.new.run
```

- re-count the number of articles, it should be less ;-)
- run https://dev.to/admin/blazer/queries/235-duplicate-published-articles-body_markdown-user_id-title locally to make sure they are gone
- also run the two group by queries - https://dev.to/admin/blazer/queries/81-duplicate-draft-articles-by-body_markdown-user_id-title and https://dev.to/admin/blazer/queries/227-duplicate-published-articles-by-body_markdown-user_id-title to double check

### DIFF
--- a/lib/data_update_scripts/20200910135958_remove_draft_articles_with_duplicate_user_id_title_body_markdown.rb
+++ b/lib/data_update_scripts/20200910135958_remove_draft_articles_with_duplicate_user_id_title_body_markdown.rb
@@ -1,0 +1,32 @@
+module DataUpdateScripts
+  class RemoveDraftArticlesWithDuplicateUserIdTitleBodyMarkdown
+    def run
+      # Currently there are duplicate draft articles in the DB with the same body_markdown, title, user_id combination
+
+      # This statement deletes all draft articles in excess found to be duplicate over those three columns
+      ActiveRecord::Base.connection.execute(
+        <<~SQL.squish,
+          WITH duplicates_draft_articles AS
+              (SELECT id
+               FROM
+                   (SELECT id,
+                           published,
+                           body_markdown,
+                           title,
+                           user_id,
+                           ROW_NUMBER() OVER(PARTITION BY user_id, title, body_markdown
+                                             ORDER BY id ASC) AS row_number
+                    FROM articles) duplicates
+               WHERE duplicates.row_number > 1
+                   AND published = 'f' -- drafts
+           )
+          DELETE
+          FROM articles
+          WHERE id IN
+                  (SELECT id
+                   FROM duplicates_draft_articles) RETURNING id;
+        SQL
+      )
+    end
+  end
+end
--- a/lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
+++ b/lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
@@ -62,28 +62,27 @@ module DataUpdateScripts
       article_id = article_to_keep.id
       articles_to_graft_ids = articles_to_graft.map { |a| a["id"] }
 
+      # NotificationSubscription, Notification and RatingVote rows will be removed
+      # Poll is ignored because it's related to the liquid tag inside the article, also user's can't use polls
+      # TagAdjustment is ignored as there's likely no reason for article to have an adjustment moved over
       models_with_a_direct_relation = [
         BufferUpdate,
         HtmlVariantSuccess,
         HtmlVariantSuccess,
         HtmlVariantTrial,
         PageView,
-        # Poll is ignored because it's related to the liquid tag inside the article,
-        # also user's can't use polls
-        RatingVote,
-        # TagAdjustment is ignored as there's likely no reason for article to have an adjustment moved over
       ]
       models_with_a_direct_relation.each do |model_class|
         model_class.where(article_id: articles_to_graft_ids).update_all(article_id: article_id)
       end
 
-      Comment.where(commentable_type: "Article",
-                    commentable_id: articles_to_graft_ids).update_all(commentable_id: article_id)
-      NotificationSubscription.where(notifiable_type: "Article",
-                                     notifiable_id: articles_to_graft_ids).update_all(notifiable_id: article_id)
-      Notification.where(notifiable_type: "Article",
-                         notifiable_id: articles_to_graft_ids).update_all(notifiable_id: article_id)
-      ProfilePin.where(pinnable_type: "Article", pinnable_id: articles_to_graft_ids).update_all(pinnable_id: article_id)
+      Comment
+        .where(commentable_type: "Article", commentable_id: articles_to_graft_ids)
+        .update_all(commentable_id: article_id)
+
+      ProfilePin
+        .where(pinnable_type: "Article", pinnable_id: articles_to_graft_ids)
+        .update_all(pinnable_id: article_id)
 
       # we add tags that are not already present in the article from the others
       # we don't really need to graft these as the tag objects belonging to the soon to be
--- a/lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
+++ b/lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
@@ -1,0 +1,104 @@
+module DataUpdateScripts
+  class CleanupPublishedArticlesWithDuplicateUserIdTitleBodyMarkdown
+    def run
+      # load all published duplicates
+      rows = ActiveRecord::Base.connection.execute(
+        <<~SQL.squish,
+          SELECT *
+          FROM
+              (SELECT *,
+                      ROW_NUMBER() OVER(PARTITION BY user_id, title, body_markdown) AS row_number,
+                      COUNT(*) OVER(PARTITION BY user_id, title, body_markdown) AS num_rows_in_partition
+               FROM articles) duplicates
+          WHERE published = 't'
+              AND num_rows_in_partition > 1 -- we select all rows in each partition with duplicates
+        SQL
+      )
+
+      articles_to_delete_ids = []
+      ActiveRecord::Base.transaction do
+        # now that we have all rows in the partition, we group them by the partition columns
+        # so we can have small groups referring to the same article and graft data of surplus articles
+        # onto the first of the partition (it doesn't really matter which one we're going to use as the "mold")
+        groups = rows.group_by { |row| [row["user_id"], row["title"], row["body_markdown"]] }
+
+        # for each group we select the first, and then graft the data of the other ones on to it
+
+        groups.each_value do |group|
+          article_to_keep = group.first
+          articles_to_graft = group[1..]
+
+          article = Article.find(article_to_keep["id"])
+
+          graft_articles(article, articles_to_graft)
+
+          # save all the magic
+          article.save
+          article.index_to_elasticsearch_inline
+
+          articles_to_delete_ids += articles_to_graft.map { |a| a["id"] }
+        end
+
+        # we shouldn't need to call Articles::Destroy as all the data has been moved over by now
+        Article.where(id: articles_to_delete_ids).destroy_all
+      end
+
+      return if articles_to_delete_ids.blank?
+
+      # Store deleted IDs temporarily in Redis for safe keeping
+      Rails.cache.write(
+        "DataUpdateScripts::CleanupPublishedArticlesWithDuplicateUserIdTitleBodyMarkdown",
+        articles_to_delete_ids,
+        expires_in: 2.weeks,
+      )
+    end
+
+    private
+
+    # Main goal here is copy and merge everything from the duplicates to the main one
+    # NOTE: I'm intentionally ignore the case that two duplicates belong to different orgs
+    # or different collections as it's highly unlikely
+    def graft_articles(article_to_keep, articles_to_graft)
+      article_id = article_to_keep.id
+      articles_to_graft_ids = articles_to_graft.map { |a| a["id"] }
+
+      models_with_a_direct_relation = [
+        BufferUpdate,
+        HtmlVariantSuccess,
+        HtmlVariantSuccess,
+        HtmlVariantTrial,
+        PageView,
+        # Poll is ignored because it's related to the liquid tag inside the article,
+        # also user's can't use polls
+        RatingVote,
+        # TagAdjustment is ignored as there's likely no reason for article to have an adjustment moved over
+      ]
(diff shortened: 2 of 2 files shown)


lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
```diff
@@ -0,0 +1,104 @@
+module DataUpdateScripts
+  class CleanupPublishedArticlesWithDuplicateUserIdTitleBodyMarkdown
```
rhymes, line 2: In an internal discussion on Slack we've decided to go ahead and "merge" duplicate articles

cc @mstruve

lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
```diff
@@ -0,0 +1,104 @@
...
+    private
+
+    # Main goal here is copy and merge everything from the duplicates to the main one
+    # NOTE: I'm intentionally ignore the case that two duplicates belong to different orgs
+    # or different collections as it's highly unlikely
+    def graft_articles(article_to_keep, articles_to_graft)
+      article_id = article_to_keep.id
+      articles_to_graft_ids = articles_to_graft.map { |a| a["id"] }
+
+      models_with_a_direct_relation = [
```
rhymes, line 65: I went through `schema.rb` and found those tables with `article_id` in them

lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
```diff
@@ -0,0 +1,104 @@
...
+        # Poll is ignored because it's related to the liquid tag inside the article,
+        # also user's can't use polls
+        RatingVote,
+        # TagAdjustment is ignored as there's likely no reason for article to have an adjustment moved over
+      ]
+      models_with_a_direct_relation.each do |model_class|
+        model_class.where(article_id: articles_to_graft_ids).update_all(article_id: article_id)
+      end
+
+      Comment.where(commentable_type: "Article",
```
rhymes, line 80: these are the tables that are polymorphic, I think we have all the meaningful ones

mstruve:
Forking production then will run all of this fun stuff against it :)

lib/data_update_scripts/20200910135958_remove_draft_articles_with_duplicate_user_id_title_body_markdown.rb
```diff
@@ -0,0 +1,32 @@
+module DataUpdateScripts
+  class RemoveDraftArticlesWithDuplicateUserIdTitleBodyMarkdown
+    def run
+      # Currently there are duplicate draft articles in the DB with the same body_markdown, title, user_id combination
+
+      # This statement deletes all draft articles in excess found to be duplicate over those three columns
+      ActiveRecord::Base.connection.execute(
```
mstruve, line 7: This statement works perfectly!

lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
```diff
@@ -0,0 +1,104 @@
...
+        expires_in: 2.weeks,
+      )
+    end
+
+    private
+
+    # Main goal here is copy and merge everything from the duplicates to the main one
+    # NOTE: I'm intentionally ignore the case that two duplicates belong to different orgs
+    # or different collections as it's highly unlikely
+    def graft_articles(article_to_keep, articles_to_graft)
```
mstruve, line 61: This transaction block has a few issues. 
1) Duplicate Notification Subscriptions when we try to roll them over `duplicate key value violates unique constraint "idx_notification_subs_on_user_id_notifiable_type_notifiable_id"` 
2) Duplicate Rating votes when we try to roll them over `duplicate key value violates unique constraint "index_rating_votes_on_user_id_and_article_id_and_context"`
3) Duplicate Notifications when we try to roll them over `duplicate key value violates unique constraint "index_notifications_on_user_notifiable_and_action_not_null"` - Honestly, I think we don't roll these over and we let them be destroyed. 

I tried to fix the first couple by doing this within the transaction block. 
```ruby
  NotificationSubscription.where(
    notifiable_type: "Article",
    notifiable_id: articles_to_graft_ids
  ).each do |notification_sub|
    notification_sub.update(notifiable_id: article_id)
  rescue ActiveRecord::RecordNotUnique
  end
```
However, that does not work bc as it turns out if Postgres raises an exception, it poisons future transactions on the same connection and you end up with this error
```
ActiveRecord::StatementInvalid (PG::InFailedSqlTransaction: ERROR:  current transaction is aborted, commands ignored until end of transaction block)
```

TL;DR We need a way to avoid duplicates when we are rolling over records and my 6pm brain is not able to figure it out right now ;)

mstruve (commented):
Couple things to iron out!

lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
citizen428: > if Postgres raises an exception, it poisons future transactions on the same connection

This is a setting called `ON_ERROR_ROLLBACK`.

When it's off statements after the error will not be executed:

```
PracticalDeveloper_development# begin;
BEGIN
PracticalDeveloper_development*# select 1/0;
ERROR:  division by zero
PracticalDeveloper_development!# select 1;
ERROR:  current transaction is aborted, commands ignored until end of transaction block
PracticalDeveloper_development!# rollback;
ROLLBACK
```

When it's on, statements continue executing (this implicitly uses [`ROLLBACK TO SAVEPOINT`](https://www.postgresql.org/docs/12/sql-rollback-to.html)):

PracticalDeveloper_development# \set ON_ERROR_ROLLBACK
PracticalDeveloper_development# begin;
BEGIN
PracticalDeveloper_development*# select 1/0;
ERROR:  division by zero
PracticalDeveloper_development*# select 1;
┌──────────┐
│ ?column? │
╞══════════╡
│        1 │
└──────────┘
(1 row)

PracticalDeveloper_development*# commit;
COMMIT
```

That said, option 3 seems sensible to me.

File: lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb [hunk 6]
```diff
@@ -0,0 +1,104 @@
...
+                      ROW_NUMBER() OVER(PARTITION BY user_id, title, body_markdown) AS row_number,
+                      COUNT(*) OVER(PARTITION BY user_id, title, body_markdown) AS num_rows_in_partition
+               FROM articles) duplicates
+          WHERE published = 't'
+              AND num_rows_in_partition > 1 -- we select all rows in each partition with duplicates
+        SQL
+      )
+
+      articles_to_delete_ids = []
+      ActiveRecord::Base.transaction do
```
citizen428, line 19: The default transaction isolation level is read committed. I assume that's what we want here, but double checking doesn't hurt.

lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
```diff
@@ -0,0 +1,104 @@
...
+      models_with_a_direct_relation = [
+        BufferUpdate,
+        HtmlVariantSuccess,
+        HtmlVariantSuccess,
+        HtmlVariantTrial,
+        PageView,
+        # Poll is ignored because it's related to the liquid tag inside the article,
+        # also user's can't use polls
+        RatingVote,
+        # TagAdjustment is ignored as there's likely no reason for article to have an adjustment moved over
```
citizen428, line 74: Will this potentially leave tag adjsutments for articles that no longer exist? If yes, should we clean them up?

lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
rhymes: Thanks a lot @mstruve for the test! So glad unique indexes in the DB are working, we wouldn't have never caught these duplicates if we'd only have the Rails layer :D All this work we've been doing is validated by these errors.

I feel we have 2 options here: 

1. use a similar technique as with tags (check if they already exists, if they don't add them, otherwise skip)
1. ignore these three models and forget about it

I'll set a time slot to see if I can come up with a simple logic, otherwise we move to plan B, though being able to merge those as well should be helpful if we decide to move this script behind an admin page in the end

lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
rhymes: They are going to be deleted when we call `.destroy_all` afterwards. They are just not migrated.

To be 100% nitpicky we should probably migrate those adjustments belonging to tags that the original article didn't have as we merge the tags but I'm not really sure it's worth it

lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
rhymes: Decided to ignore NotificationSubscription, Notification and RatingVote :)

lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
rhymes: Would you suggest otherwise? I'm not sure we need a different one

rhymes:
Added the PR to add the index https://github.com/forem/forem/pull/10284

mstruve (approved):
Ran the new script for merging articles and it worked like a charm! 20 articles were removed. I then ran the blazer SQL query again to verify they were all gone and they were.

lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
citizen428: Agreed @rhymes, it's not worth it.

lib/data_update_scripts/20200910140109_cleanup_published_articles_with_duplicate_user_id_title_body_markdown.rb
citizen428: No, I think it's fine. I just wanted to double check, all good 👍

citizen428 (approved):

PR #10274, opened 2020-09-10
Update data exporter to handle admin send

Zhao-Andy:
- [x] Feature

## Description
This updates the data exporter functionality to handle being able to send to admins. We want to support this use-case in the event that someone needs to be deleted from the admin's end and we want to send them their data before removing them.

Other technical changes include:
- updating the code so that you _always_ send an email -- previously the code had the option to not send an email, but I don't think that was ever used
- adding a new test for sending to the admin

Screen Shot 2020-09-10 at 12 39 48 PM

QA Steps:
1. Go to any user `/admin/users/1/edit`
2. Hit send to admin or user
3. Go to `localhost:3000/sidekiq` to see if a job was queued

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

Nope, just need to make sure it works in production.


spec/services/exporter/service_spec.rb
```diff
@@ -71,20 +71,36 @@ def extract_zipped_exports(buffer)
...
 
         exports = extract_zipped_exports(zipped_export)
         expect(exports).to eq(extract_zipped_exports(StringIO.new(attachment)))
       end
     end
 
+    context "when emailing an admin" do
+      it "delivers one email to the default admin email" do
+        admin_email = "admin@example.com"
+        SiteConfig.email_addresses = { default: admin_email }
```
Zhao-Andy, line 97: There might be a cleaner way of updating an ENV var within a test?

app/services/exporter/service.rb
```diff
@@ -26,7 +26,7 @@ def export(send_email: false, config: {})
 
       zipped_exports = zip_exports(exports)
 
-      send_exports_by_email(zipped_exports) if send_email
+      send_exports_by_email(zipped_exports, send_to_admin: send_to_admin)
```
Zhao-Andy, line 29: Mentioning again that this removes the option to _not_ send an email. I don't think we ever had a case where we did that, except within console.

spec/services/exporter/service_spec.rb
maestromac: I would use stubbing instead. You are setting env var that could bleed into other test.
```suggestion
        allow(SiteConfig).to receive(:email_addresses).and_return({ default: admin_email })
```

maestromac (commented):
When I try to QA this, I got
```
undefined method `admin_users_edit_path' for #<Admin::UsersController:0x00007f95c6a29758>
Did you mean?  admin_users_path
               admin_user_path
```
did I miss a step in the QA setup?

app/controllers/admin/users_controller.rb
```diff
@@ -49,6 +49,14 @@ def user_status
       redirect_to "/admin/users/#{@user.id}/edit"
     end
 
+    def export_data
+      user = User.find(params[:id])
+      send_to_admin = JSON.parse(params[:send_to_admin])
```
mstruve, line 54: Why do you use `JSON.parse` here? I would think the param would be a simple true/false.

Zhao-Andy: The form sends the param as a string, and I want to pass along a boolean instead.

Zhao-Andy:
@maestromac fixed it, my path variable was wrong 😅

app/controllers/admin/users_controller.rb
mstruve: I did not know JSON.parse would return a boolean for you from a "true"/"false" string. NEAT! TIL!

Zhao-Andy: Yeah! Better than trying to a write a custom method IMO :)

app/controllers/admin/users_controller.rb
```diff
@@ -49,6 +49,14 @@ def user_status
...
+      ExportContentWorker.perform_async(user.id, send_to_admin: send_to_admin)
+      flash[:success] = "Data exported to the #{send_to_admin ? 'admin' : 'user'}. The job will complete momentarily."
```
mstruve, line 56: Rather than propagating this boolean variable all the way through this logic flow, how about we use it here to set the email we want to send to and then we send that email to the worker. I think 
```ruby
ExportContentWorker.perform_async(user.id, email_address)
```
is a bit cleaner and more reusable for other scenarios.

mstruve (commented):
LOVE this feature!

app/controllers/admin/users_controller.rb
citizen428: @Zhao-Andy While `JSON.load` works, it's slightly confusing for the reader. Rails has built-in typecasting that you can use for this (it works with more than just simple "true"/"false", e.g. "on"/"off"):

```ruby
ActiveModel::Type::Boolean.new.cast(params[:send_to_admin])
```

It's a bit more verbose, but also a lot less confusing. Tbh I often wrap this up in a string monkey-patch or utility function:

```class
# somewhere, e.g. `app/refinements`:
module StringRefinements
  refine String do
    def to_bool
      ActiveModel::Type::Boolean.new.cast(self)
    end
  end
end

# Then
"false".to_bool
NoMethodError: undefined method `to_bool' for "false":String
using StringRefinements
"false".to_bool
#=> false
```

citizen428: There's of course also the good old `send_to_admin = params[:send_to_admin] == "true"` approach. In this specific case, I think having `to_bool` available could be worse the extra effort since based on past experience there are different hacks used in different places and standardizing one doesn't hurt.

app/controllers/admin/users_controller.rb
```diff
@@ -49,6 +49,14 @@ def user_status
       redirect_to "/admin/users/#{@user.id}/edit"
     end
 
+    def export_data
```
citizen428, line 52: Just pinging @lightalloy because this method (and especially the worker it triggers) might be useful to know about for her upcoming GDPR work.

app/controllers/admin/users_controller.rb
rhymes: Agree, it looks like we only use the email of the recipient down in the emailer and the email templates so, we should probably just probably just send the ID of the user to export and the email address of the recipient.
It'd simplify a lot of code actually :)

app/controllers/admin/users_controller.rb
```diff
@@ -49,6 +49,14 @@ def user_status
...
+      user = User.find(params[:id])
+      send_to_admin = JSON.parse(params[:send_to_admin])
+      ExportContentWorker.perform_async(user.id, send_to_admin: send_to_admin)
```
rhymes, line 55: BTW this is going to fail as Sidekiq doesn't support keyword arguments.

If you schedule the job you'll notice the following in Sidekiq:

Screenshot_2020-09-11  DEVELOPMENT  Sidekiq

rhymes (changes requested):
Nice work Andy! 

Added a request for change as there's bug :)

app/controllers/admin/users_controller.rb
Zhao-Andy: 🤯 🤯 🤯 wow, capital TIL!!! I'll add the refinement -- I definitely think we will have cases where we'll use `.to_bool`.

maestromac: TIL 🤯

app/controllers/admin/users_controller.rb
```diff
@@ -49,6 +49,16 @@ def user_status
       redirect_to "/admin/users/#{@user.id}/edit"
     end
 
+    def export_data
+      user = User.find(params[:id])
+      using StringToBool
```
msarit, line 54: This line is blowing up in my local env; I'm getting the error:
```
undefined method `using' for #<Admin::UsersController:0x00007fa43cb00990>
```

This is happening for both `Export to Admin` and `Export to User` buttons.

app/refinements/string_to_bool.rb
```diff
@@ -0,0 +1,7 @@
+module StringToBool
```
rhymes, line 1: should we call this `StringToBoolean` ?

app/controllers/admin/users_controller.rb
rhymes: Great catch @msarit. I don't think `using` can be inside a method call. According to https://docs.ruby-lang.org/en/2.7.0/Module.html#method-i-using it can be called inside a module or a class definition.

In this case I think it should be:

```ruby
module Admin
  class UsersController < Admin::ApplicationController
    using StringToBool

    layout "admin"

    # ...
  end
end
```

app/controllers/admin/users_controller.rb
```diff
@@ -49,6 +49,16 @@ def user_status
...
+      email_to_send = params[:send_to_admin].to_bool ? SiteConfig.email_addresses[:default] : user.email
+      ExportContentWorker.perform_async(user.id, email_to_send)
+      flash[:success] =
+        "Data exported to the#{params[:send_to_admin].to_bool ? 'admin' : 'user'}. The job will complete momentarily."
```
rhymes, line 58: as you already check `params[:send_to_admin].to_bool` up on line 55, maybe you can refactor to:

```ruby
send_to_admin = params[:send_to_admin].to_bool
if send_to_admin
  email = SiteConfig.email_addresses[:default]
  receiver = "admin"
else
  email = user.email
  receiver = "user
end
```

ps. there's a missing space between `the` and the category of user

spec/refinements/string_to_bool_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "rails_helper"
+
+RSpec.describe StringToBool, type: :refinement do
+  describe "#to_bool" do
+    using described_class
+    it "converts a string to a boolean" do
+      expect("true".to_bool).to eq true
```
rhymes, line 7: this should probably be `.to be(true)`

app/controllers/admin/users_controller.rb
citizen428: Yes, that's correct. My example was a copy/paste from an IRb session, where you're implicitly in the scope of the `main` module, so `using` just works.

app/refinements/string_to_bool.rb
citizen428: I like that it reflect the method that gets added, i.e.`String#to_bool` so if we rename it, we should rename the method too.

app/controllers/admin/users_controller.rb
Zhao-Andy: Ahh I see. Thanks for catching this!

spec/mailers/notify_mailer_spec.rb
```diff
@@ -429,12 +429,6 @@ def create_badge_achievement(user, badge, rewarder)
     it "includes the tracking pixel" do
       expect(email.html_part.body).to include("open.gif")
     end
-
-    it "includes UTM params" do
-      expect(email.html_part.body).to include(CGI.escape("utm_medium=email"))
-      expect(email.html_part.body).to include(CGI.escape("utm_source=notify_mailer"))
-      expect(email.html_part.body).to include(CGI.escape("utm_campaign=export_email"))
-    end
```
Zhao-Andy, line 437: I removed these tests because we can't track without `@user` set. I think we don't need to track this email but open to changing it!

rhymes (approved):
Works well!

A consideration for a future PR: we could auditing (through the audit log) so that we can internally track who requested a data export, sent to whom and when

citizen428 (approved):

PR #10275, opened 2020-09-10
Generalizing date formatting in moderation article

Rafi993:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
Removing duplicate code related to date formatting

date_time

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed


app/assets/javascripts/utilities/localDateTime.js
```diff
@@ -77,10 +77,21 @@ function timestampToLocalDateTimeLong(timestamp) {
 function timestampToLocalDateTimeShort(timestamp) {
   // example: "10 Dec 2018" if it is not the current year
   // example: "6 Sep" if it is the current year
+  // example: "22:43" if the date is the current date
 
   if (timestamp) {
-    const currentYear = new Date().getFullYear();
-    const givenYear = new Date(timestamp).getFullYear();
+    const dateToday = new Date();
```
Rafi993, line 83: There is one change in the util function used here. It displays the time if the given date is the current date else it functions the same way

Rafi993: I was not sure if I define another formatting function here that extends the current one or just edit the current one to include that change. I edited the current function for now to add that behaviour. any thoughts on this ?

Rafi993: This behaviour will propagate in other places this function is used (Right now listing dashboard for creation date)

rhymes: cc @ludwiczakpawel as it's probably a product/design decision as well

rhymes (commented):
Thanks @Rafi993 I tagged @ludwiczakpawel so that we can think about this out loud and reason on the implications around the UX

app/assets/javascripts/utilities/localDateTime.js
ludwiczakpawel: Ideally it would be nice to use relative time like:

> Just now, 2 mins ago, 1h ago, 5h ago, Yesterday, 2 days ago, Last week, ...

...but I imagine it's gonna be much more work, right? In that case, I think I'd default to use "Today" label IF given date is current date. BUT, as an addition we should add at least `title` attribute with full date and time like.

> March 17th, 4:30pm

Thoughts?

Rafi993: I think it displays full date with time in title attribute now in the format "Friday, 11 September 2020, 07:59:08"

Rafi993: But should it omit the time part and display the label "Today" if the date is current date?

rhymes: GitHub and Basecamp use this library (see our comments in here) <https://github.com/github/time-elements>.

It has localized components like `<time-ago>`, `<time-until>`, and the generic `<local-time>`.

I wonder if we should just include it and start using it everywhere @ludwiczakpawel @nickytonline - this is the size https://bundlephobia.com/result?p=@github/time-elements@3.1.1

Rafi993: But I do agree that relative date labels looks nice (but it might be useful or cause confusion depending on the context it is rendered in).

Rafi993: There is a util function there to already display time in relative labels in **app/assets/javascripts/utilities/timeAgo.js**

ludwiczakpawel: > I wonder if we should just include it and start using it everywhere

Hell yes!

rhymes: @Rafi993 their time ago algorithm seems a bit more sofisticated than ours https://github.com/github/time-elements/blob/main/src/relative-time.js

Rafi993: @rhymes you are right it does handle lot of cases. Also since it is a web component we can use it easily in preact and other places where there is js without preact (without a global util function which does need some extra things to be done when testing). It could possibly used in erb files too.

Rafi993: I mean using in erb files without any additional setup.

Rafi993: can I 
1. Alter this Pr to use https://github.com/github/time-elements in that one place and make sure it is ok 
2. Replace `dateTime` component with `time-element` in the next PR
3. Replace other places in subsequent PR

is that ok?

msarit:
@rhymes @ludwiczakpawel Hey fellas! 👋🏾 
Mind giving this PR a second look?

ludwiczakpawel:
Hey @Rafi993 do you think you could find some time to resolve conflicts and merge latest master?

Sorry it took so long for us to get to it, apparently this one slipped out of our attention :(

Rafi993:
> Hey @Rafi993 do you think you could find some time to resolve conflicts and merge latest master?
> 
> Sorry it took us so long to get to it, apparently this one slipped out of our attention :(

I have fixed the conflicts now but I think there was a discussion of using `time-elements` web component by Github to generalise  date formatting https://github.com/forem/forem/pull/10275#discussion_r487043021

citizen428:
@nickytonline Do you maybe want to weigh in on the discussion re using `time-elements` (https://github.com/forem/forem/pull/10275#discussion_r487043021)?

app/assets/javascripts/utilities/localDateTime.js
nickytonline: It will add to our bundle size, but I imagine we'll also be removing code if use `time-elements` custom elements everywhere. As well, we won't need to worry about polyfills as the browsers we support support them. Let's try it out and we can go from there. Thanks for your patience @Rafi993. 😅

msarit:
@nickytonline @rhymes Hey folks - just placing this PR on your radar 🙏🏾

nickytonline:
I weighed in on the time-elements in the referenced PR. I will take shorter peek at this though.

Rafi993:
> I weighed in on the time-elements in the referenced PR. I will take shorter peek at this though.

Hi @nickytonline if time-elements is going to be used can I close this PR since it may not be relevant in that case.

aitchiss:
> > I weighed in on the time-elements in the referenced PR. I will take shorter peek at this though.
> 
> Hi @nickytonline if time-elements is going to be used can I close this PR since it may not be relevant in that case.

Hey @Rafi993 @nickytonline - looks like [from the comments above](https://github.com/forem/forem/pull/10275#discussion_r487043021) that it was agreed we'd like to try out `time-elements` rather than move forward with the approach in this PR, so I'm going to go ahead and close this for now.

If I'm wrong in any way, please weigh in and we can always re-open. Thanks @Rafi993 for your patience and contributions with the ongoing discussions!

PR #9911, opened 2020-08-21
Stop using ProfileField#active Part 2

citizen428:
__This needs #9910 to be merged first!__

## What type of PR is this? (check all applicable)

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

See #9910 for details. tl;dr: we decided against using an `active` attribute on profile fields, so we are removing. Following the practices recommended by the [`strong_migrations` gem](https://github.com/ankane/strong_migrations) we do this in 2 steps. 

**Note:** The diff here will get smaller once the other PR has been merged and we update this PR accordingly. I just wanted to prepare everything in one go.

## Related Tickets & Documents

Part of 6994.

## QA Instructions, Screenshots, Recordings

Specs green? Good to go.

## Added tests?

- [ ] yes
- [X] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

PR #10279, opened 2020-09-10
Bug Fix:Add New Migration to Remove Active from Profile Fields

mstruve:
- [x] Bug Fix

## Description
This migration initially didn't run on production bc of a [little typo in the original PR](https://github.com/forem/forem/pull/9911/files#diff-96fd763dbe823dbb9971e988ecbb3390R3). This adds a new migration to do the job. 

https://github.com/orgs/forem/projects/6#card-45085888

alt_text

### DIFF
--- a/db/migrate/20200910205316_remove_active_from_profile_fields2.rb
+++ b/db/migrate/20200910205316_remove_active_from_profile_fields2.rb
@@ -1,0 +1,13 @@
+class RemoveActiveFromProfileFields2 < ActiveRecord::Migration[6.0]
+  def up
+    return unless column_exists?(:profile_fields, :active)
+
+    safety_assured { remove_column :profile_fields, :active }
+  end
+
+  def down
+    return if column_exists?(:profile_fields, :active)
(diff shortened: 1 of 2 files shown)


vaidehijoshi (approved):

vaidehijoshi:
honestly i am surprised that rails didn't complain about or catch this!

citizen428:
> honestly i am surprised that rails didn't complain about or catch this!

@vaidehijoshi Alas, there's not much it can do here. Defining methods inside other methods is valid Ruby [1] and since some people use migrations for data migrations you can't assume that a new migration means there has to be a schema change.

[1] Which is a bit bizarre, since they behave different from netsted methods in almost all other languages I've seen  that support them. A Ruby method just executes its body, so when it encounters a `def`, it just defines a new method in the same scope as the surrounding method.

Screen Shot 2020-09-11 at 09 21 40

citizen428 (approved):
Thank you @mstruve!

PR #10287, opened 2020-09-11
Check that nginx is available before pinging it

vaidehijoshi:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Currently, we don't have a good way of knowing whether Openresty is available before we try to bust an Nginx cache via Openresty's server.

This means that we can potentially make a request to Openresty even when it is unavailable, which causes the app to not be able to boot up when this code path is called.

This PR adds a check to make sure that Openresty is **both** _configured_ **and** _available_ before trying to bust an Nginx cache or before making additional requests to Openresty. Unfortunately, from what I can tell, the only way to know if Openresty is "up" and available is to ping it (make a request to it). In this PR, I'm making a basic `GET` request to the `openresty_path` to just see if it responds with a success or not.

This PR also improves the pre-existing tests and adds some new ones! 😉 

Closes https://github.com/forem/InternalProjectPlanning/issues/38.

In your local `.env` file, set some `OPENRESTY_` ENV variables:
```bash
export OPENRESTY_DOMAIN=localhost:9090
export OPENRESTY_PROTOCOL=http://
```

Start up your app:
```ruby
bin/startup
```

Your app should boot without any issues.

Then, go into your `rails console`.
Check that the ENV variables are what you expect/set property. Then, trigger a cache bust:
```ruby
➜  forem git:(vaidehijoshi/check-for-openresty-connection) ✗ rails c
Loading development environment (Rails 6.0.3.3)
[1] pry(main)> ApplicationConfig["OPENRESTY_DOMAIN"]
=> "localhost:9090"
[2] pry(main)> CacheBuster.bust("/vaidehijoshi")
Could not connect to Openresty via http://localhost:9090!
=> nil
```
You should see an error logged out, which is caught by the `nginx_available?` method [here](https://github.com/forem/forem/compare/vaidehijoshi/check-for-openresty-connection?expand=1#diff-6e58540a6172a3c1054d1ede7134dd00R49).

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Nope!

emotionally unavailable gif

### DIFF
--- a/app/labor/cache_buster.rb
+++ b/app/labor/cache_buster.rb
@@ -44,11 +44,14 @@ module CacheBuster
     uri = URI.parse(openresty_path)
     http = Net::HTTP.new(uri.host, uri.port)
     response = http.get(uri.request_uri)
-  rescue StandardError
-    # If we can't connect to openresty, alert ourselves that it is unavailable
-    Rails.logger.error("Could not connect to Openresty via #{openresty_path}!")
 
-    response.is_a?(Net::HTTPSuccess)
+    return true if response.is_a?(Net::HTTPSuccess)
+  rescue StandardError
+    # If we can't connect to openresty, alert ourselves that
+    # it is unavailable and return false.
+    Rails.logger.error("Could not connect to Openresty via #{openresty_path}!")
+    DatadogStatsClient.increment("cache_buster.service_unavailable", tags: ["path:#{openresty_path}"])
+    false
   end
 
   def self.bust_fastly_cache(path)
--- a/app/labor/cache_buster.rb
+++ b/app/labor/cache_buster.rb
@@ -20,7 +20,7 @@ module CacheBuster
     # https://github.com/fastly/fastly-ruby#efficient-purging
     if fastly_enabled?
       bust_fastly_cache(path)
-    elsif nginx_enabled?
+    elsif nginx_enabled? && nginx_available?
       bust_nginx_cache(path)
     end
   rescue URI::InvalidURIError => e
@@ -34,6 +34,21 @@ module CacheBuster
 
   def self.nginx_enabled?
     ApplicationConfig["OPENRESTY_PROTOCOL"].present? && ApplicationConfig["OPENRESTY_DOMAIN"].present?
+  end
+
+  def self.openresty_path
+    "#{ApplicationConfig['OPENRESTY_PROTOCOL']}#{ApplicationConfig['OPENRESTY_DOMAIN']}"
+  end
+
+  def self.nginx_available?
+    uri = URI.parse(openresty_path)
+    http = Net::HTTP.new(uri.host, uri.port)
+    response = http.get(uri.request_uri)
+  rescue StandardError
+    # If we can't connect to openresty, alert ourselves that it is unavailable
+    Rails.logger.error("Could not connect to Openresty via #{openresty_path}!")
+
+    response.is_a?(Net::HTTPSuccess)
   end
 
   def self.bust_fastly_cache(path)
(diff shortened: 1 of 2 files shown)


atsmith813:
**question:** Why does this code get triggered by `rake app_initializer:setup`? Is there a way to only do this check in the rake task or higher up? 

**issue:** We'll likely hit this again with other services in the future. From what I understand, we are trying to avoid this error only during the bootup, but it should error out otherwise. Doing this check higher up will help us to not have to do a manual check on each service in each place that service is being called across the app.

jdoss (approved):

app/labor/cache_buster.rb
```diff
@@ -36,6 +36,21 @@ def self.nginx_enabled?
     ApplicationConfig["OPENRESTY_PROTOCOL"].present? && ApplicationConfig["OPENRESTY_DOMAIN"].present?
   end
 
+  def self.openresty_path
+    "#{ApplicationConfig['OPENRESTY_PROTOCOL']}#{ApplicationConfig['OPENRESTY_DOMAIN']}"
+  end
```
atsmith813, line 41: **suggestion: ** Should we move this to [URL](https://github.com/forem/forem/blob/master/app/lib/url.rb) or another helper so we can use it elsewhere in the future?

benhalpern (approved):

vaidehijoshi:
> **question:** Why does this code get triggered by `rake app_initializer:setup`? Is there a way to only do this check in the rake task or higher up?
> 
> **issue:** We'll likely hit this again with other services in the future. From what I understand, we are trying to avoid this error only during the bootup, but it should error out otherwise. Doing this check higher up will help us to not have to do a manual check on each service in each place that service is being called across the app.

I had the same question because I also couldn't figure out _where_ exactly the rake task is trying to connect to openresty. I chatted briefly with @jdoss to see if I could get a trace on where it was coming from, but he also mentioned that he didn't know exactly who was triggering it...

atsmith813:
cc: @mstruve

I remember when we did Fastly we intentionally set it up so that a deploy would fail if we couldn't connect to Fastly. IIRC, our thinking was we'd rather have the old code remain than ship a deploy with broken Fastly code. And I know [we changed that recently because Fastly itself was having issues, not our code/Fastly config, to be optional behind an ENV var and skip updating Fastly config, in general](https://github.com/forem/forem/pull/10251).

**question:** I'm wondering how to handle situations like this - where we can't connect to services we either need (Elasticsearch, Postgres, etc.) or would really like to have (Caching, Redis, etc.). In this case, a Forem instance would stand up without Openresty and we wouldn't be able to bust any caches without addressing the error if something were actually wrong and not just the app booting up. Is the game plan to set up a monitor in the future for when these errors occur to address when this is triggered?

In other words, I'm worried that in the event something is actually wrong with Openresty, we wouldn't know unless users started reporting weird bugs that were caused by stale caches since we couldn't bust them due to the error.

atsmith813:
Is this causing the issue?

https://github.com/forem/forem/blob/3a469a99e20285b9fffcad55832187d6373fea26/config/initializers/initialization_cachebust.rb#L1-L5

Calling `CacheBuster` in this initializer? If it is, maybe we can modify this code?

mstruve:
> Calling CacheBuster in this initializer? If it is, maybe we can modify this code?

Hot damn, we should probably try to avoid busting cache's on initialization and instead offload that to our release task. The app should definitely be able to stand up even if an external service is down. I am good with this fix but I am making a ticket to remove that 

https://github.com/orgs/forem/projects/6#card-45328906

vaidehijoshi:
Per @atsmith813's suggestion, I updated this PR to add a DataDog monitor so that we won't be out of the loop if Openresty silently fails. It might be a bit annoying at first because we'll be alerted of "unavailable" cache services until that initializer code is moved out of the "boot up" process. But, I figured that this would be okay to do because at least we'll know that it is set up correctly to catch Openresty instances that have died.

cc @atsmith813 @mstruve @jdoss

atsmith813 (approved):

PR #10288, opened 2020-09-11
Spike/WIP: Add Site Config to public API

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Of course, this will only be accessible for those with appropriate permissions (of course), but this will enable tooling for UX that exists _outside_ of Forems, as well as a way that _support_ could potentially interact with a Forem programmatically if they have appropriate permission to do so.

The first few files are just an idea. I think there could be other ways we do this.


app/controllers/api/v0/api_controller.rb
```diff
@@ -47,6 +47,17 @@ def authenticate!
         end
       end
 
+      def authenticate_super_admin!
```
citizen428, line 50: This feels like we're mixing authentication and authorization. Using a policy in `SiteConfigController` feels more in line with what we usually do.

PR #10291, opened 2020-09-11
Update URL.domain not to use SiteConfig

maestromac:
- [x] Bug fix

## Description
For reason beyond me, you can not evoke SiteConfig from within `app/lib/url.rb` because it causes `Siteconfig.authentication_providers` to return `[]`, breaking our sign-in process. 

n/a

Try signing out and signing back in with either Github or Twitter auth. You should be able to see those options available to you.

## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a

### DIFF
--- a/app/lib/url.rb
+++ b/app/lib/url.rb
@@ -7,7 +7,7 @@ module URL
   end
 
   def self.domain
-    Rails.application&.initialized? ? SiteConfig.app_domain : ApplicationConfig["APP_DOMAIN"]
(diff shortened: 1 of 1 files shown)


mstruve (commented):
Looks like a couple of specs need to get updated also @benhalpern we need to figure out how we want to handle this issue so app domain can be set by SiteConfig.

maestromac:
Superseded by https://github.com/forem/forem/pull/10296

PR #10292, opened 2020-09-11
Refactor:Remove unused ReadingList Labor methods

mstruve:
- [x] Refactor

## Description
While setting up the reading list to be served from the feed content index I noticed this labor class and that it had a couple unused methods so I removed them. 

alt_text


maestromac (approved):

rhymes (approved):
Didn't even know we had this class :D

Issue #10130, opened 2020-09-01
Add a notification when resetting password


mrmir:
**Is your feature request related to a problem? Please describe.**
When you currently enter an email address on the forgot password page, it redirects you to the login page without any notification showing up

**Describe the solution you'd like**
Some sort of flash or notification saying 'An email has been sent' or similar showing up above the login box

**Describe alternatives you've considered**
Some sort of flash or notification showing up in the password page itself, maybe?

mrmir:
I'll work on this

rhymes:
Thank you @mrmir, I've assigned you!

tuang3142:
nicely done @mrmir 
@rhymes I think we can close this

rhymes:
I think this is still vlaid @tuang3142

mrmir:
@rhymes Can you reassign it to me, please? I was too new to Rails then, but I can handle it now and would love to give it a shot.

rhymes:
Done @mrmir, thank you!

mrmir:
On it

mrmir:
@rhymes @tuang3142 Do you know what I can do about this while setting up the app locally? 

![image](https://user-images.githubusercontent.com/7458152/136058423-c3469af5-a31b-4665-94cf-bd8eb7c7fb6e.png)

rhymes:
Hi @mrmir, sorry I dropped the ball on this. 

The error you're experiencing is usually because `APP_DOMAIN` in your local environment isn't correctly set. On development it should be `localhost:3000`, also check if `APP_PROTOCOL` is set to `http://`

Regarding the style of notification. I think just adding a regular Rails redirect notice should suffice, the controller should be able to pick it up and show it, we have a `global_notice` key if the regular notice doesn't work:

https://github.com/forem/forem/blob/cceb175d5d8e09bab9df945f8973a9231e21a4a3/app/views/layouts/application.html.erb#L86-L90

PR #10296, opened 2020-09-13
Fix URL.domain in development by correctly loading authentication providers

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently login is broken in development mode.

Thanks to @maestromac we found a weird interconnection between Zeitwerk's ([Rails 6 autoloader](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html)) loading order, the expectation that `SiteConfig` is ready when it should be inside `URL.domain` and the fact that `SiteConfig.authentication_providers` depends on the providers to be already loaded at runtime.

The solution is to basically move 

```ruby
Rails.application.config.to_prepare do
  Dir[Rails.root.join("app/services/authentication/**/*.rb")].each do |f|
    require_dependency(f)
  end
end
```

from the initializer to the `Authentication::Providers` file. The reason why we can't simply remove the [to_prepare](https://api.rubyonrails.org/classes/Rails/Railtie/Configuration.html#method-i-to_prepare) wrapping callback and require those dependency is that Zeitwerk forbids autoloading during initialization (the reason why I used the callback in the first place). I tried to use [after_initialize](https://api.rubyonrails.org/classes/Rails/Railtie/Configuration.html#method-i-after_initialize) but that also doesn't guarantee that the classes are loaded when we needed them.

https://github.com/forem/forem/pull/10206
Supersedes https://github.com/forem/forem/pull/10291

1. Load `master` in incognito, notice you can't login with SSO options
1. Load this branch, login and logout and see it works

## Added tests?

- [ ] yes
- [ ] no, because I need help

### DIFF
--- a/app/services/authentication/providers.rb
+++ b/app/services/authentication/providers.rb
@@ -1,3 +1,12 @@
+# We require all authentication modules to make sure providers
+# are correctly preloaded and ready to be used at this point as the loading
+# order is important
+require_dependency Rails.root.join("app/services/authentication/providers/provider.rb")
+
+Dir[Rails.root.join("app/services/authentication/**/*.rb")].each do |f|
+  require_dependency(f)
+end
+
 module Authentication
   module Providers
     # Retrieves a provider that is both available and enabled
--- a/config/application.rb
+++ b/config/application.rb
@@ -45,8 +45,8 @@ module PracticalDeveloper
     config.eager_load_paths += Dir["#{config.root}/lib"]
 
     # Middlewares folder is not otherwise autorequired.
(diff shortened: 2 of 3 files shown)


mstruve (approved):
I kinda like this better, keep all the auth code in a single service file. I think the spec is a flaky one so I restarted it 🚀

mstruve:
Looks like this might be another option https://github.com/forem/forem/pull/10310, any thoughts?

vaidehijoshi (approved):

rhymes:
@mstruve I think we should merge #10130 as well :)

PR #10310, opened 2020-09-13
Fix SiteConfig's default option of authentication_providers field

takmar:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Though I tried logging in by my GitHub account in my local environment, I found no login button there. So I looked into the code getting authentication providers, and I found that was `authentication_enabled_providers` helper method. Inside of it, `SiteConfig.authentication_providers` is called, and it returned an empty array.

After that, I inspected `SiteConfig.authentication_providers` method, and I found `Authentication::Providers.available` was set to default option. But we should pass Proc to the default option to be made lazy evaluation.

| before | after |
|---|---|
| Screen Shot 2020-09-13 at 19 31 35 | Screen Shot 2020-09-13 at 19 29 42 |

### DIFF
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -20,7 +20,7 @@ class SiteConfig < RailsSettings::Base
 
   # Authentication
   field :allow_email_password_registration, type: :boolean, default: false
-  field :authentication_providers, type: :array, default: Authentication::Providers.available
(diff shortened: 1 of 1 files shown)


joshpuetz (approved):

rhymes (approved):
Great find @takmar, thanks!

PR #10311, opened 2020-09-13
Optimization:Move Initialization CacheBust to Background Worker

mstruve:
- [x] Optimization

## Description
It is not ideal to talk to an external service in order to bust caches while initializing the application code. This moves that cache-busting out of the app initializer and into a background job that is kicked off by our app_initializer script which runs after the app code is loaded.
 
https://github.com/orgs/forem/projects/6#card-45328906

alt_text

### DIFF
--- a/config/initializers/initialization_cachebust.rb
+++ b/config/initializers/initialization_cachebust.rb
@@ -1,8 +1,0 @@
-# Trigger cache purges for globally-cached endpoints that could have changed
-CacheBuster.bust("/shell_top")
-CacheBuster.bust("/shell_bottom")
-CacheBuster.bust("/async_info/shell_version")
-CacheBuster.bust("/onboarding")
-
-# We will set RELEASE_FOOTPRINT in our Forem Cloud environment, or use HEROKU_SLUG_COMMIT if set (e.g. Heroku env)
-ENV["RELEASE_FOOTPRINT"] ||= ENV["HEROKU_SLUG_COMMIT"]
--- a/config/initializers/set_release_footprint.rb
+++ b/config/initializers/set_release_footprint.rb
@@ -1,0 +1,2 @@
+# We will set RELEASE_FOOTPRINT in our Forem Cloud environment, or use HEROKU_SLUG_COMMIT if set (e.g. Heroku env)
+ENV["RELEASE_FOOTPRINT"] ||= ENV["HEROKU_SLUG_COMMIT"]
--- a/lib/tasks/app_initializer.rake
+++ b/lib/tasks/app_initializer.rake
@@ -10,6 +10,9 @@ namespace :app_initializer do
     puts "\n== Updating Data =="
     Rake::Task["data_updates:enqueue_data_update_worker"].execute
 
+    puts "\n== Bust Caches =="
+    Rake::Task["cache:enqueue_path_bust_workers"].execute
+
     SiteConfig.health_check_token ||= SecureRandom.hex(10)
   end
 end
--- a/lib/tasks/cache.rake
+++ b/lib/tasks/cache.rake
@@ -1,0 +1,10 @@
+namespace :cache do
+  desc "Enqueue BustCachePathWorker"
+  task enqueue_path_bust_workers: :environment do
+    # Trigger cache purges for globally-cached endpoints that could have changed
+    BustCachePathWorker.set(queue: :high_priority).perform_in(10.minutes, "/shell_top")
+    BustCachePathWorker.set(queue: :high_priority).perform_in(10.minutes, "/shell_bottom")
+    BustCachePathWorker.set(queue: :high_priority).perform_in(10.minutes, "/async_info/shell_version")
+    BustCachePathWorker.set(queue: :high_priority).perform_in(10.minutes, "/onboarding")
+  end
+end
--- a/lib/tasks/cache.rake
+++ b/lib/tasks/cache.rake
@@ -2,9 +2,9 @@ namespace :cache do
   desc "Enqueue BustCachePathWorker"
   task enqueue_path_bust_workers: :environment do
     # Trigger cache purges for globally-cached endpoints that could have changed
-    BustCachePathWorker.set(queue: :high_priority).perform_in(10.minutes, "/shell_top")
-    BustCachePathWorker.set(queue: :high_priority).perform_in(10.minutes, "/shell_bottom")
-    BustCachePathWorker.set(queue: :high_priority).perform_in(10.minutes, "/async_info/shell_version")
-    BustCachePathWorker.set(queue: :high_priority).perform_in(10.minutes, "/onboarding")
+    BustCachePathWorker.set(queue: :high_priority).perform_in(5.minutes, "/shell_top")
+    BustCachePathWorker.set(queue: :high_priority).perform_in(5.minutes, "/shell_bottom")
+    BustCachePathWorker.set(queue: :high_priority).perform_in(5.minutes, "/onboarding")
+    BustCachePathWorker.set(queue: :high_priority).perform_in(5.minutes, "/async_info/shell_version")
   end
 end
--- a/lib/tasks/cache.rake
+++ b/lib/tasks/cache.rake
@@ -2,9 +2,9 @@ namespace :cache do
   desc "Enqueue BustCachePathWorker"
   task enqueue_path_bust_workers: :environment do
     # Trigger cache purges for globally-cached endpoints that could have changed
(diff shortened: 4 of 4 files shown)


rhymes (commented):
As much as I agree with the PR, I remember @benhalpern mentioning something about needing to do this before the app boots.

I'm not sure about the implications of deferring this by 10 minutes after deployment as the shell is the container of the app. 

These are questions without a specific "designated person" to answer

- would this mean that we'll serve a stale version of the app for 10 minutes after deploy?
- if so, is there an objective issue in doing that in case something changes in the shell?

atsmith813:
@benhalpern has the full context. I'll jot down how I understand it here.

This is tricky. I think it's possible that we can put this in a background/delayed worker. If we did, we'd have to re-think some additional code logic. This is due to how we have service workers, caching, and these shell views setup.

Currently, the [service worker checks for the `shell_version`](https://github.com/forem/forem/blob/f0b124cc26fda652f06dc32169dced9f00b97048/app/views/service_worker/index.js.erb#L155-L162) in [`async_info_controller#shell_version`](https://github.com/forem/forem/blob/f0b124cc26fda652f06dc32169dced9f00b97048/app/views/service_worker/index.js.erb#L155-L162). The `shell_version` is set based on `ApplicationConfig["RELEASE_FOOTPRINT"]` [which is basically the `HEROKU_SLUG_COMMIT`](https://github.com/forem/forem/blob/f0b124cc26fda652f06dc32169dced9f00b97048/config/initializers/initialization_cachebust.rb#L8).

The service worker hits the endpoint to check if the version number has changed. If it has, it fetches the new shell. In other words, the service worker needs to know in what cases it needs to fetch something new. As a quick fix, we have logic that basically instructs the service worker to fetch the new shell on every deploy - since it's possible a deploy could include an update to the shell. You'll see a nod to this [here](https://github.com/forem/forem/blob/f0b124cc26fda652f06dc32169dced9f00b97048/app/controllers/async_info_controller.rb#L30-L31):
> shell_version will change on every deploy.
> *Technically* could be only on changes to assets and shell, but this is more fool-proof.

And the way we've been telling the service worker that a new deploy has occurred is setting `RELEASE_FOOTPRINT` to be the `HEROKU_SLUG_COMMIT`, which changes on each deploy.

The challenge with moving this to a background/delayed worker is it would cause service workers to fetch stale content. **We would need to update the versioning logic of the shell if we want to move this logic to the background.** Here's a scenario. If we move this to the background, the app deploys, a user hits the site _before_ the background job completes. The service worker will see the version number has changed and fetch the new shell. However, the background job hasn't busted the cache for the old shell yet, so the service worker still gets the stale/old/previous shell back on that request. Those users who get the stale shell cache in that window before the worker completes wouldn't get the latest until the _next_ deploy (assuming no additional shell code change) when the version increments again telling the service worker to re-fetch. At that point, the latest shell would be in cache so even if the worker doesn't complete there, it would "fix" itself.

mstruve:
> Here's a scenario. Let's say we move this to the background, the app deploys, and a user hits the site before the background job completes. The service worker will see the version number has changed and fetch the new shell. However, the background job hasn't busted the cache for the old shell yet, so the service worker still gets the stale/old/previous shell back on that request. 

I believe this is already what is happening. Consider this, when the app deploys and initializes it busts the service worker shell cache. HOWEVER, we dont actually start serving the new code until a few minutes later. This means that even though we busted the cache when booting up the app, it very very likely is reset to the old shell while we are waiting for our workers to cycle from the old ones to the new ones on Heroku.

atsmith813:
I think that's spot on - it's one of the causes of our "that's a weird caching/service worker issue, but it works now!" cases we see.

Wouldn't that (the current setup/issue) be a smaller window than explicitly delaying 10 minutes in the background? [It looks like Herku overlaps two versions of code for up to 3 minutes.](https://devcenter.heroku.com/articles/preboot#caveats)

I think if we change how we're versioning (not necessarily right now or in this PR) it would fix this without having to worry about the app deploy pipeline at all - the shell cache is busted on demand like our other caches. It would also play nicer with other instances of Forem.

mstruve:
> Wouldn't that (the current setup/issue) be a smaller window than explicitly delaying 10 minutes in the background?

I am open to making it smaller. I went with 10 bc that is what we use for data update workers and its a super safe bet that the new code is out there. Though, I have noticed sidekiq workers turn over to new code pretty immediately after a deploy so we could make this 5 min. I'm game to try whatever to get this out of our app boot cycle.

atsmith813:
Ahhhh I think I see what you mean. Your goal is to move this code out of the boot cycle.

Once we move this to the background, the issue will still be there (possibly worse), but at least it's out of the app boot cycle. An additional few minutes may not be as bad as I'm fearing! Maybe it pushes us to improve the logic sooner!

jdoss (approved):

mstruve:
@benhalpern this is continuing to pop up and create little issues. What are your thoughts on this?

atsmith813:
**Idea that builds on this**
_Some pseudo code here just to demonstrate my idea_

What if instead of basically saying "hey, service workers, update the shell content on every deploy (version number change)" we update that to be, "hey, service workers, update the shell content on every deploy once we've busted the caches"?

Here's how that could look.

First, we add some "global piece of state" (YAML file, ENV var, DB) so the app knows when the cache has been busted. Let's say we used an ENV var. In an initializer, we'll have `ENV["SERVICE_WORKER_CACHES_BUSTED"] = false`. So on every deploy, this value defaults to `false`. Then, at the end of `lib/tasks/cache.rake` we set this value to `true` which basically tells the entire application, "hey, the service worker caches have been busted, we're good". 

From there, we update the following:
https://github.com/forem/forem/blob/0f6030262fb64beeb9e324ba8556faf1b1fb0c71/app/controllers/async_info_controller.rb#L28-L34

To be something like:
```ruby
def shell_version
  set_surrogate_key_header "shell-version-endpoint"
  # shell_version will change on every deploy.
  # *Technically* could be only on changes to assets and shell, but this is more fool-proof.
  shell_version = ApplicationConfig["RELEASE_FOOTPRINT"]
  service_worker_caches_busted = ApplicationConfig["SERVICE_WORKER_CACHES_BUSTED"]
  version = Rails.env.production? ? shell_version : rand(1000)
  render json: { version: version, service_worker_caches_busted: service_worker_caches_busted }.to_json
end
```

This basically tells the service worker that the version number has changed and whether or not the caches have been busted.

And, finally, we update the service worker logic [here](https://github.com/forem/forem/blob/0f6030262fb64beeb9e324ba8556faf1b1fb0c71/app/views/service_worker/index.js.erb#L155-L179) to basically say "if the version number has changed **_and_** the service worker caches have been busted, _then_ fetch the latest".

On the next deploy, the aforementioned initializer changes `ENV["SERVICE_WORKER_CACHES_BUSTED"]` back to `false`, run it all back.

_If we decide to go this route, or similar, we'll want to make sure we remain conscious of busting the cache for `/async_info/shell_version` **after** we update `SERVICE_WORKER_CACHES_BUSTED` to `true`._

mstruve:
The one gotcha here is that if we change the `ENV["SERVICE_WORKER_CACHES_BUSTED"]` value, it is only changed for the process we updated it. It does not change it globally across all processes and servers. In order to do that we would have to store the value in likely our SiteConfig so its in a central database. This would then mean adding an additional db call everytime we initiate a service worker and I am not sure if that is worth it for fixing this small bug which technically already exists in one form in our application. 

Thoughts?

atsmith813:
I was hoping the DB call wouldn't be bad since we cache that endpoint

https://github.com/forem/forem/blob/0f6030262fb64beeb9e324ba8556faf1b1fb0c71/app/controllers/async_info_controller.rb#L29

and we only bust it in 2 circumstances - a deploy and once the cache buster finishes. I'm thinking it would only hit the DB twice on any given deploy with caching, our average deploy frequency, etc.

If we did it in the DB, the code change would be simpler - we could use the version number and store it. It would only change once the cache buster completes.

mstruve:
Oh, yeah, bc its cached on Fastly we only hit our backend endpoint after the cache has been busted. Duh, Molly 🤦‍♀️

atsmith813:
Ignore me, vacation next week 😅

atsmith813 (approved):

mstruve:
I looked at recent deploy data of when the deploy hits and when we restart servers and since it looks like 8-9 min I switched the delay back to 10. 
Screen Shot 2020-09-17 at 10 46 49 AM
Screen Shot 2020-09-17 at 10 46 56 AM

benhalpern (approved):
Yup, looks good to me.

Issue #9683, opened 2020-08-08
dev.to announcements banner hides "chat invitation" notifications in chat window


nitya:
**Describe the bug**

**Context**
When a user is invite to a private chat channel, they are alerted to invite via a visual notification.
This is delivered via an indicator (shown by a users icon) in the chat session _sidebar_ menu (top left)
Users need to see and accept the invite in order to join the channel (and this is the ONLY notification)

**Bug**
 When the announcement banner is shown, it effectively masks the entire chat session sidebar menu including the filters (All-Direct-Group) and the notifications alert (users icon) there. This meant that invited users would remain unaware of pending invites. Further, since many are new to the chat feature, they might not even know this sidebar menu existed (folks tend to look at the main notifications alert at top right of page)

**To Reproduce**

1. Visit the chat page https://dev.to/connect/
2. Verify that the sidebar menu (All-Direct-Group filter, and users-icon notification) is visible
3. Trigger the announcements banner in some way (not in end user control)
4. Check if the sidebar menu is still visible (on https://dev.to/connect/)

**Expected behavior**

Notifications should always remain visible at all times - but when announcement banner is live, these get hidden behind it. (Note: the only solution is to close the banner by clicking the X - but many users are unaware of this)

**Screenshots**
The triggering of announcements is not in my control. I first observed this at CodelandConf - and didn't capture screenshots then. However in a [recent twitter exchange ](https://twitter.com/vaidehijoshi/status/1291784064337248257) I realized the same announcement banner was used in other notifications (e.g, maintenance updates) so I am sharing a screenshot of that instead.

Shows page location of announcement banner.
![banner-location](https://user-images.githubusercontent.com/279083/89710735-94c9fd80-d953-11ea-9e45-505ffdee5b37.png)

Shows the equivalent chat page (without banner) showing the chat menu sidebar (top left) - with All-Direct-Group-AlertIcon options. **I don't have a screenshot showing the exact page with a banner on it and I cannot reproduce that since banner creation is not in our control, but hope the side by side images show how the banner location occludes the alert menu**
![chatmenu-location](https://user-images.githubusercontent.com/279083/89710738-98f61b00-d953-11ea-8f45-43d47709e701.png)

**Desktop (please complete the following information):**

- OS: MacOS (Catalina)
- Browser: Edge
- Version: Version 84.0.522.52

[Twitter conversation that prompted me to file this.](https://twitter.com/nitya/status/1291524673650528256)
[...]

PR #10312, opened 2020-09-13
Hide active broadcasts from /connect routes

vaidehijoshi:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Fixes the issue described in https://github.com/forem/forem/issues/9683, wherein an active announcement broadcast would obfuscate the nav bar on `/connect` routes. Since the chat channel code seems to force scrolling down to the latest message, it made the most sense (in my opinion) to completely hide the announcement bar from view on `/connect` (similar to what we do on the `articleForm` component, aka the `/new` routes).

While doing this, I realized that some of this logic could be cleaned up quite nicely. Previously, we had broadcast-related code in both `initializeBroadcast` _and_ within the `articleForm` component. But it would be nicer to have it all live in one place so that we don't need to keep track of that logic in multiple places.

So, to that end, I refactored the `initializeBroadcast` function to handle all broadcast-related logic using regular expressions, and removed all broadcast checks from the `articleForm` component. 🎉  

Fixes https://github.com/forem/forem/issues/9683.

1. In your local environment, create an active announcement broadcast:
Screen Shot 2020-09-13 at 3 46 47 PM

2. Next, ensure that the broadcast is showing up throughout the site:
Screen Shot 2020-09-13 at 3 46 58 PM

3. Navigate to the `/connect` route (view a chat channel). You should not see the broadcast. Do a hard-refresh, and check that it _still_ does **not** show a broadcast:
Screen Shot 2020-09-13 at 3 47 15 PM

4. Check that the same happens on the article form (`/new` route). You should not see the broadcast. Do a hard-refresh, and check that it _still_ does **not** show a broadcast:
Screen Shot 2020-09-13 at 3 47 05 PM

5. Navigate back to anywhere else on the site, and the broadcast should render as expected! ✅ 

## Added tests?

- [ ] yes
- [x] no, because ~they aren't needed~ we don't have testing set up yet for our JavaScript initializers (although at some point we should definitely add them!! 😬)
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] ~no documentation needed~ updated inline documentation 

N/A

happy animal

### DIFF
--- a/app/assets/javascripts/initializers/initializeBroadcast.js
+++ b/app/assets/javascripts/initializers/initializeBroadcast.js
@@ -81,20 +81,27 @@ function renderBroadcast(broadcastElement, data) {
 
 /**
  * A function to determine if a broadcast should render.
- * Does not render a broadcast on the `/new` route or in an iframe.
+ * Does not render a broadcast in an iframe, or on `/connect` and `/new` routes.
  * Does not render a broadcast if the current user has opted-out,
  * if the broadcast has already been inserted, or if the key for
  * the broadcast's title exists in localStorage.
  *
- * If the broadcast exists in the DOM but was hidden by the articleForm,
- * the function will re-display it again by adding a class.
- *
  * @function initializeBroadcast
  */
 function initializeBroadcast() {
+  const shouldHideBroadcast = window.location.pathname.match(
+    /^(?:\/connect|\/new)/,
+  );
+
   // Iframes will attempt to re-render a broadcast, so we want to explicitly
   // avoid initializing one if we are within `window.frameElement`.
-  if (window.frameElement || window.location.pathname === '/new') {
+  if (window.frameElement || shouldHideBroadcast) {
+    const broadcast = document.getElementById('active-broadcast');
+
+    // Hide the broadcast if it exists and we are on a path where it should be hidden.
+    if (broadcast) {
+      broadcast.classList.remove('broadcast-visible');
+    }
     return;
   }
 
--- a/app/javascript/packs/articleForm.jsx
+++ b/app/javascript/packs/articleForm.jsx
@@ -39,22 +39,7 @@ function loadForm() {
   });
 }
 
-/**
- * A function to hide an active broadcast if it exists
- * by removing a `broadcast-visible` class from it.
- *
- * @function hideActiveBroadcast
- */
-function hideActiveBroadcast() {
-  const broadcast = document.getElementById('active-broadcast');
-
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/initializers/initializeBroadcast.js
```diff
@@ -81,20 +81,27 @@ function renderBroadcast(broadcastElement, data) {
...
  * the broadcast's title exists in localStorage.
  *
- * If the broadcast exists in the DOM but was hidden by the articleForm,
- * the function will re-display it again by adding a class.
- *
  * @function initializeBroadcast
  */
 function initializeBroadcast() {
+  const shouldHideBroadcast = window.location.pathname.match(
+    /^(?:\/connect|\/new)/,
```
vaidehijoshi, line 93: oh what's that oh it's just vaidehi beefing up her regex chops 💪 

the nice thing about this is that we should always be able to check this regex to see which pathnames we do _not_ want to render an announcement broadcast on! at some point maybe we want to extract this out into a constant, but since it's small, i figured it would be fine to leave it as is for now.

app/javascript/packs/articleForm.jsx
```diff
@@ -39,22 +39,7 @@ function loadForm() {
...
-function hideActiveBroadcast() {
-  const broadcast = document.getElementById('active-broadcast');
-
-  if (broadcast) {
-    broadcast.classList.remove('broadcast-visible');
-  }
-}
-
 document.ready.then(() => {
-  hideActiveBroadcast();
```
vaidehijoshi, line 57: it was pretty easy to miss/forget that the broadcast element's class was being modified here. i also realized that we were:
1. initializing a broadcast in `initializeBroadcast`, and inserting it into the DOM with an active class
2. removing the active class here in the `articleForm` component
3. potentially re-adding a class back in the `initializeBroadcast` code [here](https://github.com/forem/forem/pull/10312/files#diff-a99b7a23c274fa6b57e161e372ad2e28L89-L90)

i don't know why i wrote the code like this originally 🤦‍♀️ but in any case, i realized that we now can simplify things a lot and just do it all directly inside of the `initializeBroadcast` 🚀

mstruve (approved):
Adding a feature and net deleting code, I fully support this. Super clean!!!

citizen428 (approved):

rhymes (approved):
🔥

PR #10313, opened 2020-09-14
Changes to profile field indexing

citizen428:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR is part of the ongoing profile generalization work. It specifically deals with how we index the new profile fields in Elasticsearch:

1. Adds two new [nested fields](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html), `profile_fields` and `custom_profile_fields` to the users mapping.
2. Updates `Search::UserSerializer` to update the two new fields.
3. Adds a data update script for reindexing users.

These are the first 3 steps described in [our Elasticsearch documentation](https://docs.dev.to/backend/elasticsearch/#how-to-add-a-new-field-to-elasticsearch).

Part of #6994

1. Check out this PR
2. Update your mappings, e.g. with `bin/setup`
3. Index a user, e.g. `user.index_to_elasticsearch_inline`
4. Check that the ES document contains the nested attributes:
    ```ruby
    user.reload.elasticsearch_doc.dig("_source", "profile_fields")
    user.reload.elasticsearch_doc.dig("_source", "custom_profile_fields")
    ```
## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -658,6 +658,8 @@ class User < ApplicationRecord
   end
 
   def index_roles(_role)
+    return unless persisted?
+
     index_to_elasticsearch_inline
   end
 
--- a/app/serializers/search/user_serializer.rb
+++ b/app/serializers/search/user_serializer.rb
@@ -1,5 +1,8 @@
 module Search
   class UserSerializer < ApplicationSerializer
+    CUSTOM_ATTRIBUTES = "custom_attributes".freeze
+    HASH_TRANSFORM = ->(key, value) { { name: key, value: value } }
+
     attributes :id,
                :available_for,
                :comments_count,
@@ -17,6 +20,14 @@ module Search
 
     attribute :roles do |user|
       user.roles.map(&:name)
+    end
+
+    attribute :profile_fields do |user|
+      user.profile.data.except(CUSTOM_ATTRIBUTES).map(&HASH_TRANSFORM)
+    end
+
+    attribute :custom_profile_fields do |user|
+      user.profile.data[CUSTOM_ATTRIBUTES].map(&HASH_TRANSFORM)
     end
   end
 end
--- a/app/serializers/search/user_serializer.rb
+++ b/app/serializers/search/user_serializer.rb
@@ -23,11 +23,11 @@ module Search
     end
 
     attribute :profile_fields do |user|
-      user.profile.data.except(CUSTOM_ATTRIBUTES).map(&HASH_TRANSFORM)
+      user.profile.data.except(CUSTOM_ATTRIBUTES).map(&HASH_TRANSFORM) if user.profile
     end
 
     attribute :custom_profile_fields do |user|
-      user.profile.data[CUSTOM_ATTRIBUTES].map(&HASH_TRANSFORM)
+      user.profile.data[CUSTOM_ATTRIBUTES].map(&HASH_TRANSFORM) if user.profile
     end
   end
 end
--- a/config/elasticsearch/mappings/users.json
+++ b/config/elasticsearch/mappings/users.json
@@ -59,6 +59,28 @@
     },
     "username": {
       "type": "keyword"
+    },
+    "profile_fields": {
+      "type": "nested",
+      "properties": {
+        "name": {
+          "type": "keyword"
+        },
+        "value": {
+          "type": "keyword"
+        }
+      }
+    },
+    "custom_profile_fields": {
+      "type": "nested",
+      "properties": {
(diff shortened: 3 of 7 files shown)


app/serializers/search/user_serializer.rb
```diff
@@ -1,5 +1,8 @@
 module Search
   class UserSerializer < ApplicationSerializer
+    CUSTOM_ATTRIBUTES = "custom_attributes".freeze
```
citizen428, line 3: I decided to freeze this so we don't re-allocate the key every time we serialize a user for search.

app/serializers/search/user_serializer.rb
```diff
@@ -1,5 +1,8 @@
...
+    HASH_TRANSFORM = ->(key, value) { { name: key, value: value } }
```
citizen428, line 4: I needed the exact same transformation in two places and a lambda was convenient for this purpose. If we end up needing this elsewhere I will move it into a refinement or something.

config/elasticsearch/mappings/users.json
```diff
@@ -59,6 +59,28 @@
     },
     "username": {
       "type": "keyword"
+    },
+    "profile_fields": {
```
citizen428, line 63: @mstruve I think the only things we're really interested in are field names and values, the ID and group id aren't something we search by.

citizen428:
Moving this back to draft for now. Here's what's happening:

1. Lots of spec failures. 
2. I checked the factories, created users always have profiles.
3. BUT: if we try to create a user with a role in the spec, this will cause an exception. Here's why:
    * The factory does something like `after(:build) { |user| user.add_role(:super_admin) }`
    * The `add_role` triggers the search serializer which leads to an error since the profile is generated in the `after_create_commit` callback.

I could of course add something like `if profile.data` to the serializer, but that seems off since it would solely be there to make the specs happy. I don't with we use `add_role` on unpersisted users elsewhere in the app.

app/models/user.rb
```diff
@@ -658,6 +658,8 @@ def destroy_follows
   end
 
   def index_roles(_role)
+    return unless persisted?
```
citizen428, line 661: This change is explained in detail here https://github.com/forem/forem/pull/10313#issuecomment-691849151

citizen428:
There were still several specs failing after the previous fix. I could address this, but I'm afraid that this is just going to make it harder to write specs for others in the future (you need to be aware whether or not something may trigger an ES indexing and if the user has already been persisted at this point), so I begrudgingly added the conditionals to `Search::UserSerializer` after all. 😞

config/elasticsearch/mappings/users.json
mstruve: Do users or admins have the ability to change the name of profile fields? If they do then we are going to want to make sure we store and search by the ID otherwise every time a profile field name is updated we would have to go and find all users with that profile field old name and update them.

citizen428: Admins have the ability to change them, BUT the attribute name will never change, exactly for this reason. Say you create a profile field with a description of "Twitter Handle", the attribute will automatically be called "twitter_handle". If you later rename it to "Twitter nickname", the display in the forms will change, but never the attribute. We never directly expose the attribute to admins, so they can't change it.

citizen428: I found the following:

* `20200305201627_index_users_to_elasticsearch.rb`: already commented out
* `20200406213152_re_index_users_to_elasticsearch.rb`: already commented out
* `20200519142908_re_index_feed_content_and_users_to_elasticsearch.rb`: I now commented out the section about users and added a link to this discussion.

I also updated my current script to use the `BulkIndexWorker`, like so:

```ruby
      User.select(:id).in_batches(of: 200) do |batch|
        Search::BulkIndexWorker.set(queue: :default).perform_async("User", batch.ids)
      end
```

I wonder if there is a way to make data update scripts that deal with ES idempotent, going back and commenting out old ones does not seem like the right thing to do long term.

mstruve: Ah description is what everyone sees and we can edit, I like that! 

> I wonder if there is a way to make data update scripts that deal with ES idempotent, 

I'll put some thought against that.

mstruve (approved):

benhalpern (approved):

PR #10321, opened 2020-09-14
Add missing foreign keys to ahoy_messages and articles

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Thanks to [active_record_doctor](https://github.com/gregnavis/active_record_doctor) I found a bunch of missing FKs around the database.

I'll split the migrations in a few PRs.

This is about the following relations:

- `ahoy_messages` and `feedback_messages` in a 1:1 relation (an email message is attached to a feedback message and vice versa)
- `articles` and `users` on `second_user_id` and `third_user_id` in a 1:N (a user can be second author of many articles, same thing as a third user)

None of them has orphaned rows or invalid values: https://dev.to/admin/blazer/queries/242-orphaned-rows-ahoy_messages-articles

1. Load the console
2. Check `.feedback_message` works for `EmailMessage` and `.email_message` works for `FeedbackMessage`
3. Check `.second_user` and `.third_user` works for `Article` and `.articles_as_second_user` and `.articles_as_third_user` work for `User`

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -24,8 +24,8 @@ class Article < ApplicationRecord
   belongs_to :collection, optional: true
 
   belongs_to :organization, optional: true
-  belongs_to :second_user, optional: true
-  belongs_to :third_user, optional: true
+  belongs_to :second_user, class_name: "User", inverse_of: :articles_as_second_user, optional: true
+  belongs_to :third_user, class_name: "User", inverse_of: :articles_as_third_user, optional: true
   belongs_to :user
 
   counter_culture :user
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -71,8 +71,11 @@ class Article < ApplicationRecord
 
   validate :canonical_url_must_not_have_spaces
   validate :past_or_present_date
+  validate :validate_additional_authors_must_not_be_the_same
   validate :validate_collection_permission
+  validate :validate_second_user
   validate :validate_tag
+  validate :validate_third_user
   validate :validate_video
 
   before_validation :evaluate_markdown, :create_slug
@@ -551,6 +554,28 @@ class Article < ApplicationRecord
     return unless collection && collection.user_id != user_id
 
     errors.add(:collection_id, "must be one you have permission to post to")
+  end
+
+  def validate_second_user
+    return unless second_user_id
+    return unless second_user_id == user_id
+
+    errors.add(:second_user_id, "must not be the same user as the author")
+  end
+
+  def validate_third_user
+    return unless third_user_id
+    return unless third_user_id == user_id
+
+    errors.add(:third_user_id, "must not be the same user as the author")
+  end
+
+  def validate_additional_authors_must_not_be_the_same
+    return unless second_user_id
+    return unless third_user_id
+    return if second_user_id != third_user_id
+
+    errors.add(:base, "second and third user cannot be the same")
   end
 
   def past_or_present_date
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -19,11 +19,14 @@ class Article < ApplicationRecord
   delegate :name, to: :user, prefix: true
   delegate :username, to: :user, prefix: true
 
-  belongs_to :user
-  belongs_to :organization, optional: true
   # touch: true was removed because when an article is updated, the associated collection
   # is touched along with all its articles(including this one). This causes eventually a deadlock.
   belongs_to :collection, optional: true
+
+  belongs_to :organization, optional: true
+  belongs_to :second_user, optional: true
+  belongs_to :third_user, optional: true
+  belongs_to :user
 
   counter_culture :user
   counter_culture :organization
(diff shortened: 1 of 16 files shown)


app/models/article.rb
```diff
@@ -19,12 +19,15 @@ class Article < ApplicationRecord
...
 
-  belongs_to :user
-  belongs_to :organization, optional: true
   # touch: true was removed because when an article is updated, the associated collection
   # is touched along with all its articles(including this one). This causes eventually a deadlock.
   belongs_to :collection, optional: true
 
+  belongs_to :organization, optional: true
+  belongs_to :second_user, class_name: "User", inverse_of: :articles_as_second_user, optional: true
+  belongs_to :third_user, class_name: "User", inverse_of: :articles_as_third_user, optional: true
```
rhymes, line 28: added this two

app/models/article.rb
```diff
@@ -550,6 +556,28 @@ def validate_collection_permission
...
+    errors.add(:third_user_id, "must not be the same user as the author")
+  end
+
+  def validate_additional_authors_must_not_be_the_same
+    return unless second_user_id
+    return unless third_user_id
+    return if second_user_id != third_user_id
+
+    errors.add(:base, "second and third user cannot be the same")
+  end
```
rhymes, line 579: we don't have this three validations and now that we'll have a FK I decided to add them in. The second and third authors can technically be the same or be equal to the article's author. We don't want that

app/models/email_message.rb
```diff
@@ -1,4 +1,6 @@
 class EmailMessage < Ahoy::Message
+  belongs_to :feedback_message, optional: true
```
rhymes, line 2: If I understood the code correctly an email message can optionally only belong to a feedback message

app/models/feedback_message.rb
```diff
@@ -4,6 +4,8 @@ class FeedbackMessage < ApplicationRecord
   belongs_to :offender, class_name: "User", optional: true, inverse_of: :offender_feedback_messages
   belongs_to :reporter, class_name: "User", optional: true, inverse_of: :reporter_feedback_messages
   belongs_to :affected, class_name: "User", optional: true, inverse_of: :affected_feedback_messages
+
+  has_one :email_message, dependent: :nullify
```
rhymes, line 8: If i understood the code correctly a feedback message can have only one optional associated email message

app/models/user.rb
```diff
@@ -91,6 +91,10 @@ def sync_profile
   has_many :ahoy_visits, class_name: "Ahoy::Visit", dependent: :destroy
   has_many :api_secrets, dependent: :destroy
   has_many :articles, dependent: :destroy
+  has_many :articles_as_second_user, class_name: "Article", foreign_key: :second_user_id,
+                                     inverse_of: :second_user, dependent: :nullify
+  has_many :articles_as_third_user, class_name: "Article", foreign_key: :third_user_id,
+                                    inverse_of: :third_user, dependent: :nullify
```
rhymes, line 97: so we can fetch the articles a user is "second user" or "third user" of

app/models/feedback_message.rb
mstruve: What do we use these EmailMessages for? Is there a reason to keep them around even if the feedback message that created it is destroyed?

rhymes: AFAIK this is used for conflict resolution purposes. In case a violation occurs we send an email. I think for audit purposes we should keep a copy of such email

mstruve (approved):

app/models/article.rb
Zhao-Andy: Oh hmmmmmm so I think we have a PR soon that will change this to an array column and deprecate `article.second_user_id` and `article.third_user_id`. Might be better to hold off on changing them for now?

cc @juliannatetreault who is working on refactoring the column.

juliannatetreault: Correct, @Zhao-Andy! I have a draft PR open right now that removes `second_user_id` and `third_user_id` in favor of `co_author_ids` that should be ready for review in the _very_ near future (I ran into some major DB problems last night into this morning that held the PR up).

nickytonline (approved):
🚀

app/models/article.rb
rhymes: Ah, didn't know :( I'll remove those from the PR then

Zhao-Andy: Thanks, sorry for the extra work!

juliannatetreault: So, so sorry, @rhymes! I should have communicated this better. 😞

rhymes: ahha no problem :) I'll take a look at your PR when it's ready @juliannatetreault!

juliannatetreault: Just added you as a reviewer, @rhymes! 😎

PR #10322, opened 2020-09-14
Refactor:Remove ReadingList Labor Class, Replace with user Method

mstruve:
- [x] Refactor

## Description
We were using an entire labor class to fetch cached reading list IDs for users. For all other cached list of IDs we have simple methods on the user model for them so I updated the code to match that pattern. 

Extra refactor: I removed `saw_onboarding` and `checked_code_of_conduct` from the async info cache key because both of those attributes are updated via `user.save` meaning that `updated_at` will be updated when they are so they are redundant in the cache key. 

## Added tests?
- [x] yes

alt_text


vaidehijoshi (approved):

spec/models/user_spec.rb
```diff
@@ -1057,6 +1057,24 @@ def provider_username(service_name)
...
     end
+
+    def create_reaction(user, article)
+      Reaction.create!(
+        user_id: user.id,
+        reactable_id: article.id,
+        reactable_type: "Article",
+        category: "readinglist",
+      )
+    end
```
rhymes, line 1068: Is there a reason for this method over the classic `create(:reading_reaction)` factory?

rhymes (approved):
Cool!

spec/models/user_spec.rb
mstruve: Nope! That was a lazy copy/paste, updated! :)

Issue #10174, opened 2020-09-02
Cannot unfollow tags 


michael-tharrington:
**Describe the bug**

I'm unable to unfollow tags that I'm currently following. In other words, when I click on the "following" button for a tag, the button changes to "follow" making it appear that I've unfollowed it, but when I refresh the page it says "following" again.

**To Reproduce**

1. Go to 'https://dev.to/dashboard/following_tags' -->
2. Use this page to see what tags you're following and navigate to a tag's landing page (`dev.to/t/nameoftag`) -->
3. Click on "following" next to the tag's name
4. See it change to "follow"
5. Refresh the page and you'll see it saying "following" again.

**Expected behavior**

When I click on "following" next to a tag's name, it should change to "follow" and remain that way even when I refresh the page.

**Screenshots**

Here's a link to a video of the issue in action - https://share.getcloudapp.com/bLuRy8yB

**Additional context**

A DEV user reported this issue and I've confirmed it is happening to me too.

It appears that another user was experiencing a similar (but different) issue here previously - https://github.com/forem/forem/issues/5397

madarsbiss:
Works fine on Chrome 85.0.4183.83 (64-bit)

michael-tharrington:
I'm out of date! 😅

I'm on Chrome 84.0.4147.135 (64-bit)

Let me see if updating solves the problem for me. Thanks for the suggestion!

michael-tharrington:
I'm officially updated, but still experiencing issues.

So I returned to https://dev.to/dashboard/following_tags and it looks like #jokes is no longer an option there, so maybe I succesfully unfollowed it? I returned to https://dev.to/t/jokes and it still says "following". The pattern continues — I click to change it to "follow" -> refresh -> it's back to "following".

I thought that maybe this is some sort of caching issue so repeated the steps with a different tag: #history. I navigated to https://dev.to/t/history -> clicked on "following" -> it changed to "follow" -> I refresh -> it returns to saying "following" ... but, when I go back to https://dev.to/dashboard/following_tags "history" is gone.

madarsbiss:
I would try to logout, clear cache, disable all extensions (for testing), login and see if the problem persists

citizen428:
I experience the exact same behavior as @michael-tharrington on the latest Firefox Developer Edition.

Link2Twenty:
You're not really following it anymore but the `userData()` function hasn't updated. You can write *this* 

```js
[...]

PR #10324, opened 2020-09-14
Fix unable to unfollow tags

takmar:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
There are two cache methods related to the reason the following tags are not updated, as pointed out in [this comment](https://github.com/forem/forem/issues/10174#issuecomment-687067914). So I fixed code relating to these two, which are cached_followed_tags and user_data methods.

Closes https://github.com/forem/forem/issues/10174

## Added tests?
no, because the test environment default cache setting is off.

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

alt_text

### DIFF
--- a/app/decorators/user_decorator.rb
+++ b/app/decorators/user_decorator.rb
@@ -19,7 +19,8 @@ class UserDecorator < ApplicationDecorator
   ].freeze
 
   def cached_followed_tags
-    follows_map = Rails.cache.fetch("user-#{id}-#{last_followed_at&.rfc3339}/followed_tags", expires_in: 20.hours) do
+    follows_map = Rails.cache.fetch("user-#{id}-#{following_tags_count}-#{last_followed_at&.rfc3339}/followed_tags",
+                                    expires_in: 20.hours) do
       Follow.follower_tag(id).pluck(:followable_id, :points).to_h
     end
 
--- a/app/models/follow.rb
+++ b/app/models/follow.rb
@@ -27,7 +27,8 @@ class Follow < ApplicationRecord
   scope :follower_tag, ->(id) { where(follower_id: id, followable_type: "ActsAsTaggableOn::Tag") }
 
   counter_culture :follower, column_name: proc { |follow| COUNTER_CULTURE_COLUMN_NAME_BY_TYPE[follow.followable_type] },
-                             column_names: COUNTER_CULTURE_COLUMNS_NAMES
+                             column_names: COUNTER_CULTURE_COLUMNS_NAMES,
+                             touch: true
   after_create :send_email_notification
   before_destroy :modify_chat_channel_status
   after_save :touch_follower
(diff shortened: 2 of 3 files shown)


app/decorators/user_decorator.rb
```diff
@@ -19,7 +19,8 @@ class UserDecorator < ApplicationDecorator
   ].freeze
 
   def cached_followed_tags
-    follows_map = Rails.cache.fetch("user-#{id}-#{last_followed_at&.rfc3339}/followed_tags", expires_in: 20.hours) do
+    follows_map = Rails.cache.fetch("user-#{id}-#{following_tags_count}-#{last_followed_at&.rfc3339}/followed_tags",
```
takmar, line 22: I added following_tags_count in order to update the cache when we unfollow tags, as the code below does.
https://github.com/forem/forem/blob/83d37df4d79700d3daaad14df560f221689c8226/app/models/user.rb#L313

app/models/follow.rb
```diff
@@ -27,7 +27,8 @@ class Follow < ApplicationRecord
   scope :follower_tag, ->(id) { where(follower_id: id, followable_type: "ActsAsTaggableOn::Tag") }
 
   counter_culture :follower, column_name: proc { |follow| COUNTER_CULTURE_COLUMN_NAME_BY_TYPE[follow.followable_type] },
-                             column_names: COUNTER_CULTURE_COLUMNS_NAMES
+                             column_names: COUNTER_CULTURE_COLUMNS_NAMES,
+                             touch: true
```
takmar, line 31: I added a touch option in order to update the user's updated_at, which purpose is to update the [user_data cache](https://github.com/forem/forem/blob/master/app/controllers/async_info_controller.rb#L48) in unfollowing a tag. But I'm not sure that updating the user's update_at affects anything else.  if this way is not good, I'll find another way.

Zhao-Andy: I think it might be safer to updated `user.last_followed_at` instead than the entire `user.updated_at`, since the cache key you linked uses `user.last_followed_at`.

app/decorators/user_decorator.rb
Zhao-Andy: Good catch!

Zhao-Andy (commented):
One comment -- not sure if it's _needed_ though. Thanks for the PR!

app/models/follow.rb
mstruve: I think this could lead to a lot of unnecessary updates bc of all the callbacks on user. I would rather update the `user_cache_key` in the async controller to maybe include the `@user.cached_followed_tags.count`

When you do this though, be aware that `cached_followed_tags` will hit Redis so ideally we will want to save that to a local variable before creating that async info hash and then use that local variable for the hash and the count for the cache key

mstruve (commented):
Nice!!! I think we should leave out the added `touch` and instead update our cache key for user data. Let me know what you think!

app/models/follow.rb
takmar: @Zhao-Andy @mstruve 
Thanks for your comments!

I understood that there was a possibility that updating `user.updated_at` causes a problem when a user follows a tag. So I removed `touch`.

And I considered updating `user.last_followed_at` and adding `@user.cached_followed_tags.count` to `user_cache_key` like you two advised me.
However, I ended up adding `current_user&.following_tags_count` to `user_cache_key` because that needs no additional update or cache and solve the issue.

mstruve (approved):

citizen428 (approved):

PR #10331, opened 2020-09-14
Optimization:Store Podcast Ep Data in Redis, Rather than Sending as Sidekiq Argument

mstruve:
- [x] Optimization

## Description
Take it from someone who has made this mistake before, it is a really bad idea to use large Sidekiq arguments. If you end up enqueueing a ton of jobs it will eat up your Redis resources extremely fast from a memory standpoint and from an I/O standpoint bc of all the polling Sidekiq does against the queues. This PR updates our Podcast::CreateWorker so that we store the large `item_data` hash in Redis and send a hashed key representing it instead as an argument to our worker. The worker then looks up the key in Redis and fetches the item data to send to the CreateEpisode service.

https://github.com/orgs/forem/projects/6#card-43005072

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/services/podcasts/get_episode.rb
+++ b/app/services/podcasts/get_episode.rb
@@ -12,7 +12,9 @@ module Podcasts
       if episode
         try_update_media_url(episode: episode, item_data: item_data, force_update: force_update)
       else
-        PodcastEpisodes::CreateWorker.perform_async(podcast.id, item_data.to_h)
+        episode_cache_key = cache_key(item_data)
+        cache_episode_data(episode_cache_key, item_data)
+        PodcastEpisodes::CreateWorker.perform_async(podcast.id, episode_cache_key)
       end
     end
 
@@ -35,6 +37,14 @@ module Podcasts
       episode.save
     rescue ArgumentError, NoMethodError => e
       Rails.logger.error("not a valid date: #{e}")
+    end
+
+    def cache_episode_data(episode_cache_key, item_data)
+      Rails.cache.write(episode_cache_key, item_data.to_h)
+    end
+
+    def cache_key(item_data)
+      Digest::SHA1.hexdigest(item_data.to_s)
     end
   end
 end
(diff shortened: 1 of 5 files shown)


app/services/podcasts/get_episode.rb
```diff
@@ -36,5 +38,13 @@ def update_published_at(episode, item_data)
     rescue ArgumentError, NoMethodError => e
       Rails.logger.error("not a valid date: #{e}")
     end
+
+    def cache_episode_data(episode_cache_key, item_data)
+      Rails.cache.write(episode_cache_key, item_data.to_h)
```
citizen428, line 43: I think this only need to have `item_data` as an argument, since it can call `cache_key` by itself, which would simplify the call-site.

In case we want to keep the flexibility of being able to override the cache key (though it seems we only have a single caller), we can make it a keyword argument with a default value, otherwise I'd just inline it:

```ruby
# Ruby keyword arguments can use other arguments even in method calls
def cache_episode_data(item_data, episode_cache_key: cache_key(item_data))
  Rails.cache.write(episode_cache_key, item_data.to_h)
end

# or
def cache_episode_data(item_data)
  episode_cache_key = Digest::SHA1.hexdigest(item_data.to_s)
  Rails.cache.write(episode_cache_key, item_data.to_h)
end
```

mstruve: The problem with this is that we need the cache key to send to the worker above in line 17 so we need access to it in the `call` method

citizen428: Ah, I missed that. Good morning, Michael. I guess in that case I'd just inline the `Rails.cache.write`, the method doesn't seem to offer any benefit over that.

```ruby
        episode_cache_key = cache_key(item_data)
        Rails.cache.write(episode_cache_key, item_data.to_h)
        PodcastEpisodes::CreateWorker.perform_async(podcast.id, episode_cache_key)
```

citizen428 (approved):

benhalpern (approved):

PR #10332, opened 2020-09-14
Optimization:Use delete_by when Destroying HtmlVariant Trials and Successes

mstruve:
- [x] Optimization

## Description
`HtmlVariantSuccess` and `HtmlVariantTrial`'s are pretty simple objects that have 0 callbacks or relationships we need to worry about cleaning up when we destroy them. For this reason, rather than using [`destroy_by`](https://apidock.com/rails/v6.0.0/ActiveRecord/Relation/destroy_by) which will go through and delete them all 1 by 1 executing all callback, let's switch to using a quick [`delete_by`](https://apidock.com/rails/v6.0.0/ActiveRecord/Relation/delete_by) which will delete them all in one SQL statement. 

This will eliminate a lot of spans and speed this puppy up quite a bit
Screen Shot 2020-09-14 at 3 37 06 PM

## Added tests?

alt_text

### DIFF
--- a/app/workers/html_variants/remove_old_data_worker.rb
+++ b/app/workers/html_variants/remove_old_data_worker.rb
@@ -5,8 +5,8 @@ module HtmlVariants
     sidekiq_options queue: :low_priority, retry: 15
 
     def perform
-      HtmlVariantTrial.destroy_by("created_at < ?", 2.weeks.ago)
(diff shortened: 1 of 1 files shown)


Ridhwana (approved):

Zhao-Andy (approved):
Nice!

PR #10337, opened 2020-09-15
Remove redirect on comments#update

atsmith813:
- [x] Refactor
- [x] Bug Fix

## Description
This refactors `comments#update` to avoid doing a redirect causing a race condition when busting cache.

https://github.com/forem/forem/issues/10338

1. Fire up the app locally and sign in.
2. Leave a comment on an article.
3. Update the comment you just made and make sure it doesn't error out. It should be the same UX as before just without a redirect - it goes back to the comment index view and displays the updated comment content.
4. Leave a comment on a podcast.
5. Repeat step 3.
6. Add yourself to an organization in your local DB and publish an article under an organization.
7. Leave a comment on the article.
8. Repeat step 3.

## Added tests?

## Added to documentation?
- [x] no documentation needed

dog_racing_human_in_push_car


app/controllers/comments_controller.rb
```diff
@@ -141,7 +141,26 @@ def update
     authorize @comment
     if @comment.update(permitted_attributes(@comment).merge(edited_at: Time.zone.now))
       Mention.create_all(@comment)
-      redirect_to URI.parse(@comment.path).path, notice: "Comment was successfully updated."
+
+      # The following sets variables used in the index view. We render the index
+      # view directly to avoid having to redirect which leads to caching issues.
```
rhymes, line 146: It works! I would link [this explanation](https://github.com/forem/forem/issues/10338#issuecomment-693401481) in the comment, mostly to clarify "caching issues"

rhymes (approved):
I'm a bit sad we have to go this path removing the redirect but it looks like there's no better way. Thanks Alex!

app/controllers/comments_controller.rb
```diff
@@ -141,7 +141,33 @@ def update
...
+      case @commentable_type
+      when "PodcastEpisode"
+        @user = @commentable&.podcast
+      when "Article"
+        # user could be a user or an organization
+        @user = @commentable&.user
+        @article = @commentable
+      else
+        @user = @commentable&.user
+      end
```
vaidehijoshi, line 168: non-blocker: it might be nice to extract this into a method and out of the controller action itself.

vaidehijoshi (approved):

app/controllers/comments_controller.rb
atsmith813: Agreed! I'm thinking how we do comments as a whole could use a refactor 🤔  I may circle back to clean up `index` and `update`.

PR #10339, opened 2020-09-15
Add co_author_ids to Articles

juliannatetreault:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR replaces `second_user_id` and `third_user_id` with a `co_author_ids` column on the `Articles` table to make adding co-authors to posts more flexible in the future. Additionally, this PR adds two methods to the `article_decorater`, `#co_authors`, and `#co_author_name_and_path`, that are used to display the linked co-author names in a human-readable format in the `articles/show.html.erb` view, replaces `second_user_id` and `third_user_id` with `co_author_ids` when necessary in the codebase, refactors existing specs to use `co_author_ids` in place of `second_user_id` and `third_user_id`, adds a `data_update` script to update the `co_author_ids` column with the existing `second_user_id` and `third_user_id` data, and adds validations around co-author IDs not being the same as the author's ID or as a co-author's ID.

Relates to a private issue and relates to PR #10252 . The second part to this PR (the removal of `second_user_id` and `third_user_id` from the Articles table) can be found here: PR #10383 

**To QA this refactor, ensure that you have `super_admin` permissions and then do the following:**

- Navigate to `/admin/articles` and add `co_author_ids` to the post of your choice (_heads-up_: I created my own post for testing purposes) and click "submit" to save your changes. Upon submit, you should be redirected to the article show page, should see a flash messge indicating a successful update, and you should now see the co-author IDs listed in the view:
Screen Shot 2020-09-15 at 2 57 29 PM
Screen Shot 2020-09-15 at 2 57 53 PM
Screen Shot 2020-09-17 at 1 43 36 PM

- Click on the post with the newly added co-authors and ensure that they are displayed correctly (in a readable sentence) and that the correct path is associated with each co-author (you should be taken to each individual user's profile upon click):
Screen Shot 2020-09-17 at 1 44 28 PM
Screen Shot 2020-09-17 at 1 45 42 PM

- If you're feeling ambitious, you can add even _more_ co-authors to the post by returning to `/admin/articles` and following the first two QA steps again:
Screen Shot 2020-09-17 at 1 47 29 PM
Screen Shot 2020-09-17 at 1 47 54 PM
Screen Shot 2020-09-17 at 1 48 03 PM
Screen Shot 2020-09-17 at 1 48 22 PM

- To test the co-author-related validations, attempt to add the author of the post as a co-author of the post. If you try to make the author of the post a co-author as well, you should be shown a flash message stating that you cannot do this. Additionally, test that you cannot add the same co-author to a particular article twice and that adding text for the co_author_id is rendered as invalid. You should again, be shown a flash message stating that you cannot do this:
Screen Shot 2020-09-17 at 1 49 42 PM
Screen Shot 2020-09-17 at 1 54 38 PM
Screen Shot 2020-09-17 at 2 03 35 PM
Screen Shot 2020-09-17 at 2 03 45 PM
Screen Shot 2020-09-30 at 11 14 25 AM
Screen Shot 2020-09-30 at 12 57 41 PM

- Finally, try to break things! 🔨 

**_Additionally, please make sure that all of the specs pass and that the app looks as expected._**

## Added tests?

- [x] yes and refactored existing specs to account for changes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Yes: I have added a `data_update` script to update the `co_author_ids` column with the existing `second_user_id` and `third_user_id` data.

typing

### DIFF
--- a/app/controllers/admin/articles_controller.rb
+++ b/app/controllers/admin/articles_controller.rb
@@ -46,8 +46,12 @@ module Admin
       article.boosted_dev_digest_email = article_params[:boosted_dev_digest_email].to_s == "true"
       article.user_id = article_params[:user_id].to_i
       article.co_author_ids = article_params[:co_author_ids].split(",").map(&:strip)
-      article.save!
-      render body: nil
+      if article.save
+        flash[:success] = "Article saved!"
+      else
+        flash[:danger] = article.errors_as_sentence
+      end
+      redirect_to admin_article_path(article.id)
     end
 
     private
--- a/app/controllers/admin/articles_controller.rb
+++ b/app/controllers/admin/articles_controller.rb
@@ -45,9 +45,8 @@ module Admin
       article.boosted_additional_articles = article_params[:boosted_additional_articles].to_s == "true"
       article.boosted_dev_digest_email = article_params[:boosted_dev_digest_email].to_s == "true"
       article.user_id = article_params[:user_id].to_i
-      article.second_user_id = article_params[:second_user_id].to_i
-      article.third_user_id = article_params[:third_user_id].to_i
-      article.update!(article_params)
+      article.co_author_ids = article_params[:co_author_ids].split(/\s*,\s*/)
+      article.save!
       render body: nil
     end
 
@@ -131,8 +130,7 @@ module Admin
                           main_image_background_hex_color
                           featured_number
                           user_id
-                          second_user_id
-                          third_user_id
+                          co_author_ids
                           last_buffered]
       params.require(:article).permit(allowed_params)
     end
--- a/app/controllers/admin/articles_controller.rb
+++ b/app/controllers/admin/articles_controller.rb
@@ -45,7 +45,7 @@ module Admin
       article.boosted_additional_articles = article_params[:boosted_additional_articles].to_s == "true"
       article.boosted_dev_digest_email = article_params[:boosted_dev_digest_email].to_s == "true"
       article.user_id = article_params[:user_id].to_i
-      article.co_author_ids = article_params[:co_author_ids].split(/\s*,\s*/)
+      article.co_author_ids = article_params[:co_author_ids].split(",").map(&:strip)
       article.save!
       render body: nil
     end
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -294,7 +294,7 @@ class StoriesController < ApplicationController
     end
 
     @comments_to_show_count = @article.cached_tag_list_array.include?("discuss") ? 50 : 30
-    assign_second_and_third_user
+    assign_co_authors
     set_article_json_ld
     @comment = Comment.new(body_markdown: @article&.comment_template)
   end
@@ -303,11 +303,10 @@ class StoriesController < ApplicationController
     !@article.published && params[:preview] != @article.password
   end
 
-  def assign_second_and_third_user
-    return if @article.second_user_id.blank?
+  def assign_co_authors
+    return if @article.co_author_ids.blank?
 
-    @second_user = User.find(@article.second_user_id)
-    @third_user = User.find(@article.third_user_id) if @article.third_user_id.present?
+    @co_authors = User.find(@article.co_author_ids)
   end
 
   def assign_user_comments
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -295,11 +295,18 @@ class StoriesController < ApplicationController
 
     @comments_to_show_count = @article.cached_tag_list_array.include?("discuss") ? 50 : 30
     set_article_json_ld
+    assign_co_authors
     @comment = Comment.new(body_markdown: @article&.comment_template)
   end
 
   def permission_denied?
     !@article.published && params[:preview] != @article.password
+  end
+
+  def assign_co_authors
+    return if @article.co_author_ids.blank?
+
+    @co_author_ids = User.find(@article.co_author_ids)
   end
 
   def assign_user_comments
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -294,19 +294,12 @@ class StoriesController < ApplicationController
     end
 
     @comments_to_show_count = @article.cached_tag_list_array.include?("discuss") ? 50 : 30
-    assign_co_authors
     set_article_json_ld
     @comment = Comment.new(body_markdown: @article&.comment_template)
   end
 
   def permission_denied?
     !@article.published && params[:preview] != @article.password
-  end
-
-  def assign_co_authors
-    return if @article.co_author_ids.blank?
-
-    @co_authors = User.find(@article.co_author_ids)
   end
 
   def assign_user_comments
--- a/app/dashboards/article_dashboard.rb
+++ b/app/dashboards/article_dashboard.rb
@@ -10,8 +10,7 @@ class ArticleDashboard < Administrate::BaseDashboard
   ATTRIBUTE_TYPES = {
     user: Field::BelongsTo,
     user_id: UserIdField,
-    second_user_id: Field::Number,
-    third_user_id: Field::Number,
+    co_author_ids: Field::Number,
     organization: Field::BelongsTo,
     id: Field::Number,
     title: Field::String,
@@ -65,8 +64,7 @@ class ArticleDashboard < Administrate::BaseDashboard
   # on the model's form (`new` and `edit`) pages.
   FORM_ATTRIBUTES = %i[
     user_id
-    second_user_id
-    third_user_id
+    co_author_ids
     organization
     title
     search_optimized_title_preamble
--- a/app/dashboards/article_dashboard.rb
+++ b/app/dashboards/article_dashboard.rb
@@ -10,7 +10,7 @@ class ArticleDashboard < Administrate::BaseDashboard
   ATTRIBUTE_TYPES = {
     user: Field::BelongsTo,
     user_id: UserIdField,
-    co_author_ids: Field::Number,
+    co_author_ids: Field::Select.with_options(collection: []),
     organization: Field::BelongsTo,
     id: Field::Number,
     title: Field::String,
@@ -64,7 +64,6 @@ class ArticleDashboard < Administrate::BaseDashboard
   # on the model's form (`new` and `edit`) pages.
   FORM_ATTRIBUTES = %i[
     user_id
-    co_author_ids
     organization
     title
     search_optimized_title_preamble
--- a/app/decorators/article_decorator.rb
+++ b/app/decorators/article_decorator.rb
@@ -88,7 +88,7 @@ class ArticleDecorator < ApplicationDecorator
   end
 
   def co_authors
-    User.select(:name, :username).where(id: co_author_ids)
+    User.select(:name, :username).where(id: co_author_ids).order(created_at: :asc)
   end
 
   def co_author_name_and_path
--- a/app/decorators/article_decorator.rb
+++ b/app/decorators/article_decorator.rb
@@ -86,4 +86,14 @@ class ArticleDecorator < ApplicationDecorator
   def long_markdown?
     body_markdown.present? && body_markdown.size > LONG_MARKDOWN_THRESHOLD
   end
+
+  def co_authors
+    User.where(id: co_author_ids)
+  end
+
+  def co_author_name_and_path
+    co_authors.map do |user|
+      "<b><a href=\"#{user.path}\">#{user.name}</a></b>"
+    end.to_sentence
+  end
 end
--- a/app/decorators/article_decorator.rb
+++ b/app/decorators/article_decorator.rb
@@ -88,7 +88,7 @@ class ArticleDecorator < ApplicationDecorator
   end
 
   def co_authors
-    User.where(id: co_author_ids)
+    User.select(:name, :username).where(id: co_author_ids)
   end
 
   def co_author_name_and_path
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -72,9 +72,9 @@ class Article < ApplicationRecord
   validate :validate_collection_permission
   validate :validate_tag
   validate :validate_video
-  validate :validate_co_authors
-  validate :validate_co_authors_must_not_be_the_same
-  validate :validate_co_authors_exist
+  validate :validate_co_authors, unless: -> { co_author_ids.blank? }
+  validate :validate_co_authors_must_not_be_the_same, unless: -> { co_author_ids.blank? }
+  validate :validate_co_authors_exist, unless: -> { co_author_ids.blank? }
 
   before_validation :evaluate_markdown, :create_slug
   before_save :update_cached_user
@@ -556,19 +556,19 @@ class Article < ApplicationRecord
   end
 
   def validate_co_authors
-    return if co_author_ids.blank? || co_author_ids.exclude?(user_id)
+    return if co_author_ids.exclude?(user_id)
 
     errors.add(:co_author_ids, "must not be the same user as the author")
   end
 
   def validate_co_authors_must_not_be_the_same
-    return if co_author_ids.blank? || co_author_ids.uniq.count == co_author_ids.count
+    return if co_author_ids.uniq.count == co_author_ids.count
 
     errors.add(:base, "co-author IDs must be unique")
   end
 
   def validate_co_authors_exist
-    return if co_author_ids.blank? || User.where(id: co_author_ids).count == co_author_ids.count
+    return if User.where(id: co_author_ids).count == co_author_ids.count
 
     errors.add(:co_author_ids, "must be valid user IDs")
   end
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -74,6 +74,7 @@ class Article < ApplicationRecord
   validate :validate_video
   validate :validate_co_authors
   validate :validate_co_authors_must_not_be_the_same
+  validate :validate_co_authors_exist
 
   before_validation :evaluate_markdown, :create_slug
   before_save :update_cached_user
@@ -567,6 +568,13 @@ class Article < ApplicationRecord
     return if co_author_ids.uniq.count == co_author_ids.count
 
     errors.add(:base, "co-author IDs must be unique")
+  end
+
+  def validate_co_authors_exist
+    valid_co_authors = User.where(id: co_author_ids).count == co_author_ids.count
+    return if valid_co_authors
+
+    errors.add(:co_author_ids, "must be valid user IDs containing only numbers")
   end
 
   def past_or_present_date
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -556,27 +556,25 @@ class Article < ApplicationRecord
   end
 
   def validate_co_authors
-    return unless co_author_ids
+    return if co_author_ids.blank?
+
     return unless co_author_ids.include?(user_id)
 
     errors.add(:co_author_ids, "must not be the same user as the author")
   end
 
   def validate_co_authors_must_not_be_the_same
-    return if co_author_ids.blank?
-
-    return if co_author_ids.uniq.count == co_author_ids.count
+    return if co_author_ids.blank? || co_author_ids.uniq.count == co_author_ids.count
 
     errors.add(:base, "co-author IDs must be unique")
   end
 
   def validate_co_authors_exist
-    return unless co_author_ids
+    return if co_author_ids.blank?
 
-    valid_co_authors = User.where(id: co_author_ids).count == co_author_ids.count
-    return if valid_co_authors
+    return if User.where(id: co_author_ids).count == co_author_ids.count
 
-    errors.add(:co_author_ids, "must be valid user IDs containing only numbers")
+    errors.add(:co_author_ids, "must be valid user IDs")
   end
 
   def past_or_present_date
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -571,6 +571,8 @@ class Article < ApplicationRecord
   end
 
   def validate_co_authors_exist
+    return unless co_author_ids
+
     valid_co_authors = User.where(id: co_author_ids).count == co_author_ids.count
     return if valid_co_authors
 
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -556,9 +556,7 @@ class Article < ApplicationRecord
   end
 
   def validate_co_authors
-    return if co_author_ids.blank?
-
-    return unless co_author_ids.include?(user_id)
+    return if co_author_ids.blank? || co_author_ids.exclude?(user_id)
 
     errors.add(:co_author_ids, "must not be the same user as the author")
   end
@@ -570,9 +568,7 @@ class Article < ApplicationRecord
   end
 
   def validate_co_authors_exist
-    return if co_author_ids.blank?
-
-    return if User.where(id: co_author_ids).count == co_author_ids.count
+    return if co_author_ids.blank? || User.where(id: co_author_ids).count == co_author_ids.count
 
     errors.add(:co_author_ids, "must be valid user IDs")
   end
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -564,10 +564,9 @@ class Article < ApplicationRecord
   def validate_co_authors_must_not_be_the_same
     return unless co_author_ids
 
-    co_author = co_author_ids.map { |id| id }
-    return if co_author_ids.include?(co_author)
+    return if co_author_ids.uniq.count == co_author_ids.count
 
-    errors.add(:base, "must not be the same user as the co-author")
+    errors.add(:base, "co-author IDs must be unique")
   end
 
   def past_or_present_date
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -144,7 +144,7 @@ class Article < ApplicationRecord
            :video_thumbnail_url, :video_closed_caption_track_url, :social_image,
            :published_from_feed, :crossposted_at, :published_at, :featured_number,
            :last_buffered, :facebook_last_buffered, :created_at, :body_markdown,
-           :email_digest_eligible, :processed_html, :second_user_id, :third_user_id)
+           :email_digest_eligible, :processed_html, :co_author_ids)
   }
 
   scope :boosted_via_additional_articles, lambda {
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -562,7 +562,7 @@ class Article < ApplicationRecord
   end
 
   def validate_co_authors_must_not_be_the_same
-    return unless co_author_ids
+    return if co_author_ids.blank?
 
     return if co_author_ids.uniq.count == co_author_ids.count
 
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -565,7 +565,7 @@ class Article < ApplicationRecord
     return unless co_author_ids
 
     co_author = co_author_ids.map { |id| id }
-    return if co_author_ids != co_author
+    return if co_author_ids.include?(co_author)
 
     errors.add(:base, "must not be the same user as the co-author")
   end
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -72,6 +72,8 @@ class Article < ApplicationRecord
   validate :validate_collection_permission
   validate :validate_tag
   validate :validate_video
+  validate :validate_co_authors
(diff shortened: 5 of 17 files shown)


app/controllers/admin/articles_controller.rb
```diff
@@ -45,9 +45,8 @@ def update
       article.boosted_additional_articles = article_params[:boosted_additional_articles].to_s == "true"
       article.boosted_dev_digest_email = article_params[:boosted_dev_digest_email].to_s == "true"
       article.user_id = article_params[:user_id].to_i
-      article.second_user_id = article_params[:second_user_id].to_i
-      article.third_user_id = article_params[:third_user_id].to_i
-      article.update!(article_params)
+      article.co_author_ids = article_params[:co_author_ids].split(/\s*,\s*/)
```
rhymes, line 48: I think we should change this

`"1,2, 3 ".split(/\s*,\s*/)` results in `["1", "2", "3 "]` which brings along both a 3 with a space

What about:

```ruby
ids.split(",").map(&:strip)
```

This way we remove the leading and trailing spaces for each value and let the validation occur

vaidehijoshi:
> I could see there being a possibility of having to backfill the new co_author_ids column with any existing second_user_ids and third_user_ids, but I'm not sure if any co-authors have been added to posts yet and/or there are any ids associated with these columns currently.

You want to check this on Blazer. I ran some quick queries (feel free to edit it!) on Blazer so you see which/how many users are currently a `second_user_id` for some article [here](https://dev.to/admin/blazer/queries/243-articles-with-second_user_id), as well as a `third_user_id` for some article [here](https://dev.to/admin/blazer/queries/244-articles-with-third_user_id).

You'll definitely need to copy over the ids that currently are `second_user_id`s and `third_user_id`s before removing the column, otherwise you'll lose that data.

db/migrate/20200909163415_remove_columns_from_articles.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveColumnsFromArticles < ActiveRecord::Migration[6.0]
+  def change
+    safety_assured do
+      remove_column :articles, :second_user_id, :bigint
+      remove_column :articles, :third_user_id, :bigint
+    end
+  end
+end
```
vaidehijoshi, line 8: We can't remove these columns until we've backfilled every article that currently a `second_user_id` and `third_user_id` and copied those `id`s over to `co_author_ids`

app/dashboards/article_dashboard.rb
```diff
@@ -10,8 +10,7 @@ class ArticleDashboard < Administrate::BaseDashboard
   ATTRIBUTE_TYPES = {
     user: Field::BelongsTo,
     user_id: UserIdField,
-    second_user_id: Field::Number,
-    third_user_id: Field::Number,
+    co_author_ids: Field::Number,
```
rhymes, line 13: should this be `Field::Select` ?

app/views/articles/show.html.erb
```diff
@@ -155,11 +155,12 @@
...
                 <% end %>
-                <% if @second_user.present? %>
-                  <em>with <b><a href="<%= @second_user.path %>"><%= @second_user.name %></a></b></em>
-                <% end %>
-                <% if @third_user.present? %>
-                  <em> and <b><a href="<%= @third_user.path %>"><%= @third_user.name %></a></b></em>
+                <% if @article.co_author_ids.present? %>
+                  <% if @article.co_author_ids.count == 1 %>
+                    <em>with <b><a href="<%= @article.co_authors.first.path %>"><%= @article.co_authors.first.name %></a></b></em>
+                  <% elsif @article.co_author_ids.count > 1 %>
```
vaidehijoshi, line 161:
```suggestion
                  <% else %>
```
Since you already have checked if `@article.co_author_ids.present?`, and have an `if` for when the `count` is exactly `1`, you can just do a plain `else` here, because you can be sure that it will be greater than `1` but also not `0`.

app/decorators/article_decorator.rb
```diff
@@ -86,4 +86,14 @@ def video_metadata
   def long_markdown?
     body_markdown.present? && body_markdown.size > LONG_MARKDOWN_THRESHOLD
   end
+
+  def co_authors
+    User.where(id: co_author_ids)
```
rhymes, line 91: small optimization, as it seems you're only using `path` and `name`, maybe you can select only the needed fields: `username` and `name`. 

```ruby
 User.select(:name, :username).where(id: co_author_ids)
```

this so we don't load the entire row for only what we need

app/views/articles/show.html.erb
juliannatetreault: Good call!

app/views/articles/show.html.erb
```diff
@@ -155,11 +155,12 @@
...
-                <% if @second_user.present? %>
-                  <em>with <b><a href="<%= @second_user.path %>"><%= @second_user.name %></a></b></em>
-                <% end %>
-                <% if @third_user.present? %>
-                  <em> and <b><a href="<%= @third_user.path %>"><%= @third_user.name %></a></b></em>
+                <% if @article.co_author_ids.present? %>
+                  <% if @article.co_author_ids.count == 1 %>
+                    <em>with <b><a href="<%= @article.co_authors.first.path %>"><%= @article.co_authors.first.name %></a></b></em>
+                  <% else %>
+                    <em>with <%= @article.co_author_name_and_path.html_safe %></em>
```
rhymes, line 162: why make a distinction here?

db/migrate/20200909163415_remove_columns_from_articles.rb
rhymes: I was going to say the same, the correct approach here is to hide the columns as suggested by https://github.com/ankane/strong_migrations#removing-a-column

otherwise we lose the data.

This PR needs a data update script that backfills the data.

app/models/article.rb
```diff
@@ -143,7 +143,7 @@ class Article < ApplicationRecord
            :video_thumbnail_url, :video_closed_caption_track_url, :social_image,
```
rhymes, line 143: I think we need the following validators at least:

- co author ids can't contain the user id
- the array needs to be a set essentially (1, 1) can't be added
- ideally we should also check that those IDs actually exist (which was the goal of the foreign key in the other PR)

rhymes (changes requested):
Hi @juliannatetreault, there are a few details to streamline, especially on data preservation, but the PR is on the the right path :)

app/decorators/article_decorator.rb
juliannatetreault: Great suggestion!

app/dashboards/article_dashboard.rb
juliannatetreault: I actually have no idea. I think `Field::Number` makes sense given the description for `ATTRIBUTE_TYPES`, but again, I'm not sure. Why do you think that this field should possibly be `Field::Select` instead @rhymes? 🤔

rhymes: Because I think `::Number` is for single values, `::Select` is to select multiple values. I haven't tested though, maybe see in `localhost:3000/resource_admin` which is the old admin :D

app/views/articles/show.html.erb
juliannatetreault: Great question! I thought this distinction was necessary for formatting reasons and was needed to create a coherent sentence, but now I'm realizing that the way I set the logic up in the decorator makes so that I don't even need the distinction anymore, so I've since removed this conditional. 😄

mstruve (commented):
OMG YES!!!! I love this!!! I just reviewed @rhymes's PR applying foreign keys to these second and third id columns and was cringing the whole time THANK YOU!

app/controllers/stories_controller.rb
```diff
@@ -294,7 +294,6 @@ def assign_article_show_variables
     end
 
     @comments_to_show_count = @article.cached_tag_list_array.include?("discuss") ? 50 : 30
-    assign_second_and_third_user
```
vaidehijoshi, line 297: Just looking at this again, and I'm curious: why does this PR remove this step of assigning these users? I would think that refactor would still keep around this step of assigning coauthors (but move from `second_user_id` and `third_user_id` to instead be `co_author_ids`) here too, right? Unless there is something I am not seeing in which case please do point it out to me! :)

app/controllers/admin/articles_controller.rb
```diff
@@ -45,10 +45,13 @@ def update
...
-      article.third_user_id = article_params[:third_user_id].to_i
-      article.update!(article_params)
-      render body: nil
+      article.co_author_ids = article_params[:co_author_ids].split(",").map(&:strip)
+      if article.save
+        flash[:success] = "Article saved!"
+      else
+        flash[:danger] = article.errors_as_sentence
+      end
+      redirect_to admin_article_path(article.id)
```
juliannatetreault, line 54: These changes were necessary for a few reasons:
- if `article.update!(article_params)` is used here, the article gets overwritten during `#update`, while using `article.save` updates the article in as expected.
- in order to get the flash messages and redirect to work in the `#update` action, I had to remove the green highlighting that occurred on submit in the form in `_individual_article.html.erb` and add `local: true`. Prior to these changes, it did not appear that errors were rendering for Articles at all and no redirects took place. 😱

app/models/article.rb
juliannatetreault: Agreed and updated!

app/dashboards/article_dashboard.rb
juliannatetreault: I'm still trying to figure this out.. after poking around at this for a _while_ yesterday, I found a solution (using `Field::Select` with the `:collection` option that defaults to an empty array) that should work, but I was unable to get it to work properly. I'm circling back around to this now that I've made other changes and hopefully with a fresh set of eyes, I will be able to make it work @rhymes ! 😄

app/controllers/stories_controller.rb
juliannatetreault: Great question, @vaidehijoshi! At the time that I had removed this, it looked like `co_author_ids` was assigned elsewhere, but over the past two days as I've been addressing outstanding PR comments, I'm unsure of whether or not what I had initially thought is true -- I'm unable to find where they would be assigned. 🙈 All of that being said, I'm reverting this change and adding back the assignment of `co_author_ids` to the `Stories::Controller`. 😊

juliannatetreault:
Okay, I think I have addressed everyone's concerns and only have the outstanding failing spec (that I cannot, for the life of me, figure out!) left to tackle. To recap all of the changes I've made per your requests, I've:

- Added a `data_update` script to update the `co_author_ids` column with the existing `second_user_id` and `third_user_id`.
- Added back the `second_user_id` and `third_user_id` columns and deleted the migration file that removed the columns from this PR. The removal of these columns will be handled here: PR #10383 
- Added validations around `co_author_ids` to ensure that `co_author_ids` are not the same as the author ID and that `co_author_ids` cannot be duplicated (e.g. the same co-author cannot be added as a co-author to a particular article twice).
- Added tests around `co_author_ids` validations in `article_spec.rb`
- Optimized the query in the Article model to use `.select()`.
- Replaced the regex in the `Articles::Controller` with `.split()`.
- Changed `Field::Number` to `Field::Select.with_options(collection: [])` in `article_dashboard.rb` to handle the array of `co_author_ids`.
- Reverted the changes made to the code removed from the `Stories::Controller`.
- Cleaned up some superfluous code in the Articles `show.html.erb`

Please let me know if I'm missing anything or if you have further change requests, otherwise, I will continue to attempt to fix the failing `stories_show_spec.rb` while awaiting further review! 😅 

cc: @vaidehijoshi and @rhymes 😄

app/models/article.rb
```diff
@@ -552,6 +554,22 @@ def validate_collection_permission
...
+    return unless co_author_ids.include?(user_id)
+
+    errors.add(:co_author_ids, "must not be the same user as the author")
+  end
+
+  def validate_co_authors_must_not_be_the_same
+    return unless co_author_ids
+
+    co_author = co_author_ids.map { |id| id }
+    return if co_author_ids.include?(co_author)
```
fdoxyz, line 568: > Added validations around co_author_ids to ensure that co_author_ids are not the same as the author ID and that co_author_ids cannot be duplicated (e.g. the same co-author cannot be added as a co-author to a particular article twice).

If I'm following along correctly I think this may be the problem. Both `co_author_ids` and `co_author` are arrays and the `include?` method won't work for this. Not sure if it's the best way but I think this suggestion may help validate there are no duplicate IDs in the array

```suggestion
return if co_author_ids.uniq.count == co_author_ids.count
```

The error message here should also maybe read something along the lines of "coauthor ids must be unique"

fdoxyz (commented):
I've been in favor of all the suggestions and the PR looks almost ready! Chipping in now I think I have a suggestion that may help for when you're back from vacation 🙂

app/models/article.rb
juliannatetreault: 🤦‍♀️ you are a lifesaver, @fdoxyz! I have been soooo stumped by why this was failing and any time I dropped into `pry` I found no useful information. I've made both of these great suggestions. Thank you times a million! 😄

fdoxyz: Happens all the time with these small things! Glad it helped 😄

mstruve (approved):
Looks like we have 59 articles with additional authors so this should be an easy update.

lib/data_update_scripts/20200916202343_backfill_co_author_ids_for_articles.rb
```diff
@@ -0,0 +1,10 @@
+module DataUpdateScripts
+  class BackfillCoAuthorIdsForArticles
+    def run
+      articles = Article.where.not(second_user_id: nil).or(Article.where.not(third_user_id: nil))
+      articles.find_each do |article|
```
citizen428, line 5: Here we can update all articles in a single SQL statement, without instantiating any `ActiveRecord` objects:

```ruby
articles.update_all("co_author_ids = array[second_user_id, third_user_id]")
```

This will issue the following statement:

```sql
UPDATE "articles" 
SET co_author_ids = array[second_user_id, third_user_id] 
WHERE ("articles"."second_user_id" IS NOT NULL OR "articles"."third_user_id" IS NOT NULL);
```

I think we should let the database do the heavy lifting here since we don't actually need `ActiveRecord` instances for this.

mstruve (commented):
Thinking ahead for the future. You might want to add to your roadmap for this to add these additional users to Elasticsearch. That way if someone searches for a co author they will also find the article

app/models/article.rb
```diff
@@ -72,6 +72,8 @@ class Article < ApplicationRecord
   validate :validate_collection_permission
   validate :validate_tag
   validate :validate_video
+  validate :validate_co_authors
+  validate :validate_co_authors_must_not_be_the_same
```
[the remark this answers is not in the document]
rhymes: I thought about it but I'd somehow liked the idea of preserving insertion order so that "co-authored by alice and bob" would always be in this order, even if another alice is added at the end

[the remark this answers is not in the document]
juliannatetreault: @rhymes when you say "preserving insertion order" are you hoping that the names are always alphabetically ordered even if the co-author names are not added in alphabetical order (Bob is added as a co-author before Alice is, but Alice's name appears first on the article) or are you hoping for something else? 😄

lib/data_update_scripts/20200916202343_backfill_co_author_ids_for_articles.rb
juliannatetreault: This is an **excellent** suggestion and I've since made this change @citizen428!

app/models/article.rb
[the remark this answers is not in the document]
rhymes: No, not the alphabetical order. If today I add Bob and tomorrow I add Alice, being this an array, it should stay "bob and alice", never "alice and bob". But I don't feel strongly about this, whatever you choose is fine :)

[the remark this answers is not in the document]
juliannatetreault: Ahh, I see! Thank you for the clarification @rhymes. 😄

juliannatetreault:
I _believe_ that I've addressed all of the outstanding PR comments that came in last week aside from @mstruve's Elasticsearch comment (which I'm still looking into -- I've been reading through our Elasticsearch docs to get a better idea of how to approach the suggested roadmap), so I'd appreciate another (hopefully final 😅 ) review from my reviewers. Please and thank you so much in advance! cc: @rhymes, @vaidehijoshi, @fdoxyz, @Zhao-Andy, and @citizen428 (I know you're busy so no need to respond, but I'm tagging you anyway since you had left me feedback!)

mstruve:
@juliannatetreault FWIW I think you should do the Elasticsearch work in a different PR which is why I approved this one as is 😊. Keep this focused on just adding the fields.

juliannatetreault:
> @juliannatetreault FWIW I think you should do the Elasticsearch work in a different PR which is why I approved this one as is 😊. Keep this focused on just adding the fields.

OH! Sweet. Thank you for the clarification @mstruve!

fdoxyz (changes requested):
I gave this another run and it's looking great, almost there! 🙌🏼

I did find a bug though. When passing in text instead of integers (as `co_author_ids` are meant to be) they are parsed to integers and Ruby turns the text into a `0`, which then breaks the Article. 

https://share.getcloudapp.com/4guJDRXb

This in turn made me think anyone can add a number that doesn't match with an existing User. I'm on the fence about how to validate this because from how I see it there's two ways to go here:

1. We could be okay with having administrators add any number they want, they're in control of what they're adding as long as they're valid integers (greater than zero). In this case a validator that checks for something like this `co_author_ids.map(&:to_i).include?(0)` would be enough to make sure gibberish is not being turned into zeros. But it would be ideal too that this won't break the Article view when performing the fetch of those IDs (in case someone added a huge number that doesn't match any existing User in the DB)
1. We could add a validator that checks all co_author_ids match with an existing User in the DB. Something along the lines of `User.where(id: co_author_ids).count == co_author_ids.count` could work and this should make the problem go away.

In either case though, I'm not sure how the co_author fetch would behave if a user is deleted (that is a co_author of an Article). Maybe adding a spec for this situation to make sure the render of the Article won't be affected by this makes sense?

I don't feel too strongly about one approach over the other (or other ones I haven't thought about), but I do prefer **2.** as it would be a bit more robust and would avoid unexpectedly breaking an Article page because of a typo 🙂

juliannatetreault:
> I gave this another run and it's looking great, almost there! 🙌🏼
> 
> I did find a bug though. When passing in text instead of integers (as `co_author_ids` are meant to be) they are parsed to integers and Ruby turns the text into a `0`, which then breaks the Article.
> 
> https://share.getcloudapp.com/4guJDRXb
> 
> This in turn made me think anyone can add a number that doesn't match with an existing User. I'm on the fence about how to validate this because from how I see it there's two ways to go here:
> 
> 1. We could be okay with having administrators add any number they want, they're in control of what they're adding as long as they're valid integers (greater than zero). In this case a validator that checks for something like this `co_author_ids.map(&:to_i).include?(0)` would be enough to make sure gibberish is not being turned into zeros. But it would be ideal too that this won't break the Article view when performing the fetch of those IDs (in case someone added a huge number that doesn't match any existing User in the DB)
> 2. We could add a validator that checks all co_author_ids match with an existing User in the DB. Something along the lines of `User.where(id: co_author_ids).count == co_author_ids.count` could work and this should make the problem go away.
> 
> In either case though, I'm not sure how the co_author fetch would behave if a user is deleted (that is a co_author of an Article). Maybe adding a spec for this situation to make sure the render of the Article won't be affected by this makes sense?
> 
> I don't feel too strongly about one approach over the other (or other ones I haven't thought about), but I do prefer **2.** as it would be a bit more robust and would avoid unexpectedly breaking an Article page because of a typo 🙂

AWESOME catch @fdoxyz ! I added a validation to handle this and an error message that I feel is descriptive enough for the Admin (please let me know if you feel otherwise). I'm working on the spec now and will have the changes pushed up ASAP.

vaidehijoshi (commented):
Not sure if it's just me seeing this or if anyone else can repro this, but I noticed a random `<` on the `/admin/articles` route while switching to your branch to QA this PR 🤔 

On `master`:
Screen Shot 2020-09-30 at 10 26 36 AM

On this PR:
Screen Shot 2020-09-30 at 10 28 25 AM

Do you see the same, @juliannatetreault ?

juliannatetreault:
> Not sure if it's just me seeing this or if anyone else can repro this, but I noticed a random `<` on the `/admin/articles` route while switching to your branch to QA this PR 🤔
> 
> On `master`:
> Screen Shot 2020-09-30 at 10 26 36 AM
> 
> On this PR:
> Screen Shot 2020-09-30 at 10 28 25 AM
> 
> Do you see the same, @juliannatetreault ?

Yes @vaidehijoshi , I see that and addressed it in a separate PR, which was closed while I was on vacation because it has already been addressed on `master` -- it's still showing on my branch because I have not rebased on `master` since the fix for it was merged! 😊

app/models/article.rb
```diff
@@ -552,6 +555,30 @@ def validate_collection_permission
...
+    errors.add(:base, "co-author IDs must be unique")
+  end
+
+  def validate_co_authors_exist
+    return unless co_author_ids
+
+    valid_co_authors = User.where(id: co_author_ids).count == co_author_ids.count
+    return if valid_co_authors
+
+    errors.add(:co_author_ids, "must be valid user IDs containing only numbers")
```
vaidehijoshi, line 579: I'm a bit confused about this validation + error message.

We check this: `User.where(id: co_author_ids).count == co_author_ids.count`, which checks to see that the `count` of Users that exist in the database with the `co_author_ids` of the article is the same as the `count` of the `co_author_ids` on the Article. But I'm not sure how that is ensuring that `co_author_ids` are _only_ going to be numbers? We could instead validate that the `co_author_ids` are actually only numbers here if we wanted to use this error message.

spec/models/article_spec.rb
```diff
@@ -52,6 +52,39 @@ def build_and_validate_article(*args)
...
+      end
+
+      it "is invalid if the co_author is entered as a text value rather than an integer" do
+        article.co_author_ids = [user.id, "abc"]
+
+        expect(article).not_to be_valid
+      end
+
+      it "is invalid if the co_author ID is not greater than 0" do
+        article.co_author_ids = [user.id, "0"]
```
vaidehijoshi, line 76: Should this be `0` instead of `"0"`? Looking at it right now, I'm not sure if this test is not valid because its a string or because of the fact that the id is `0`

vaidehijoshi (commented):
This is looking good! I left a few comments throughout and had 2 validation-related questions :)

spec/models/article_spec.rb
juliannatetreault: It should be `0`... that's what I get for copy-pasta-ing. 🤦‍♀️ The test was passing because passing a string is _also_ invalid. Good catch, @vaidehijoshi!

app/models/article.rb
```diff
@@ -552,6 +555,30 @@ def validate_collection_permission
     errors.add(:collection_id, "must be one you have permission to post to")
   end
 
+  def validate_co_authors
+    return unless co_author_ids
+    return unless co_author_ids.include?(user_id)
```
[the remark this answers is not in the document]
juliannatetreault: Nope! Good catch. 😊

app/models/article.rb
```diff
@@ -552,6 +555,30 @@ def validate_collection_permission
...
+    return if co_author_ids.uniq.count == co_author_ids.count
+
+    errors.add(:base, "co-author IDs must be unique")
+  end
+
+  def validate_co_authors_exist
+    return unless co_author_ids
+
+    valid_co_authors = User.where(id: co_author_ids).count == co_author_ids.count
+    return if valid_co_authors
```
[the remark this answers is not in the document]
juliannatetreault: Done! ✅

app/models/article.rb
juliannatetreault: Soooo, I kind of just got ahead of myself and have since changed this by removing the "containing only numbers" portion of the error message. 😅 😬

app/models/article.rb
```diff
@@ -72,6 +72,9 @@ class Article < ApplicationRecord
   validate :validate_collection_permission
   validate :validate_tag
   validate :validate_video
+  validate :validate_co_authors
+  validate :validate_co_authors_must_not_be_the_same
+  validate :validate_co_authors_exist
```
vaidehijoshi, line 77: Idea: all of these validations check `if co_author_ids.blank?` as the very first thing within the validation method. If you want to tidy up these validations _even further_, you could extract that out into a [method](https://edgeguides.rubyonrails.org/active_record_validations.html#using-a-symbol-with-if-and-unless), or even better, a [proc](https://edgeguides.rubyonrails.org/active_record_validations.html#using-a-proc-with-if-and-unless), and then remove that repeated check from all three of these. Something like this:

```ruby
validates : validate_co_authors, unless: -> { co_author_ids.blank? }
validates : validate_co_authors_must_not_be_the_same, unless: -> { co_author_ids.blank? }
validates : validate_co_authors_exist, unless: -> { co_author_ids.blank? }
```

vaidehijoshi (approved):

vaidehijoshi:
This is a big improvement compared to the 2 columns we had previously, and it'll be much more extensible in the future! Nice work :)

app/controllers/stories_controller.rb
```diff
@@ -294,20 +294,19 @@ def assign_article_show_variables
...
   end
 
-  def assign_second_and_third_user
-    return if @article.second_user_id.blank?
+  def assign_co_authors
+    return if @article.co_author_ids.blank?
 
-    @second_user = User.find(@article.second_user_id)
-    @third_user = User.find(@article.third_user_id) if @article.third_user_id.present?
+    @co_author_ids = User.find(@article.co_author_ids)
```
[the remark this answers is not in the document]
rhymes: ahah TIL. I always use `.where(id: [])`

app/decorators/article_decorator.rb
```diff
@@ -86,4 +86,14 @@ def video_metadata
   def long_markdown?
     body_markdown.present? && body_markdown.size > LONG_MARKDOWN_THRESHOLD
   end
+
+  def co_authors
+    User.select(:name, :username).where(id: co_author_ids).order(created_at: :asc)
```
rhymes, line 91: 🔥

lib/data_update_scripts/20200916202343_backfill_co_author_ids_for_articles.rb
```diff
@@ -0,0 +1,8 @@
+module DataUpdateScripts
+  class BackfillCoAuthorIdsForArticles
+    def run
+      articles = Article.where.not(second_user_id: nil).or(Article.where.not(third_user_id: nil))
+      articles.update_all("co_author_ids = array[second_user_id, third_user_id]")
```
rhymes, line 5: There's a problem in this script:

- load the console
- take an article, set `.second_user_id` to something and save it
- run the two lines here above manually
- reload the article, the result will be something like:

```ruby
[6] pry(main)> a.second_user_id
=> 3
[7] pry(main)> a.co_author_ids
=> [3, nil]
```

I don't think we want nil in that array. 

I think we should iterate on the result of line 4 and build the array in Ruby, remove the nils with `.compact` and then save it to the column with `update_columns(co_author_ids: array)` so to skip callbacks. There might be a way to use SQL but I wound't bother as it's only 59 articles on DEV: https://dev.to/admin/blazer/queries/247-articles-with-coauthors

rhymes (changes requested):
Great job @juliannatetreault, I think once we fixed the script this can go live!

rhymes (approved):
Great work @juliannatetreault, thanks for the changes! :-)

juliannatetreault:
> Great work @juliannatetreault, thanks for the changes! :-)

Thank you and absolutely -- happy to make them! Thank YOU for all of the back-and-forth review @rhymes . 😄

rhymes:
@juliannatetreault I think @fdoxyz needs to re-review before you can merge as he requested changes earlier

fdoxyz (approved):
🙈

juliannatetreault:
> Great! @juliannatetreault I think @fdoxyz needs to re-review before you can merge as he requested changes earlier

I'm not sure why there is still an outstanding check that hasn't passed... should I just rerun the build or is this something that an admin merge would solve? @rhymes 🤔

PR #10340, opened 2020-09-15
Bug Fix/Optimization:Clear Github Repo HttpCache Key Daily

mstruve:
- [x] Bug Fix
- [x] Optimization

## Description
Our Github Oauth client uses [Faraday::HttpCache](https://github.com/forem/forem/blob/master/app/services/github/oauth_client.rb#L114) to cache request results which can be served in event of an error. The problem with this is every time we fetch Github repos we add another request to our Redis key AND we reset the expiration back to 24 hours. This can cause the key in Redis to grow uncontrollably. When it gets really large we start to timeout trying to add to it. You can see this problem in the response time Honeycomb graph below. You can also see the time spent writing to Redis in the next image which is of a Honeycomb trace. To prevent the key from getting too large we delete it once a day.
Screen Shot 2020-09-14 at 4 01 16 PM
Screen Shot 2020-09-15 at 11 03 52 AM

The other option that I first thought of was to set the expiration time for the HttpCache keys to something shorter so they would expire. However, that would mean we would have to ensure that at some point no one would hit the github repos index endpoint so the key could expire and I didn't think that was feasible as we grow so I went with the manual clear. I tried to use the gems code paths as much as possible to avoid issues if the gems change. Unfortunately, it still led to me calling one private method to get the `Faraday::HttpCache` storage object.

https://github.com/orgs/forem/projects/6#card-45548773

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/workers/github_repos/clear_http_cache_worker.rb
+++ b/app/workers/github_repos/clear_http_cache_worker.rb
@@ -1,0 +1,24 @@
+module GithubRepos
+  class ClearHttpCacheWorker
+    include Sidekiq::Worker
+    GITHUB_REPOS_URL = "https://api.github.com/user/repos?per_page=100".freeze
+
+    sidekiq_options queue: :medium_priority, retry: 10
+
+    def perform
+      return unless SiteConfig.github_key.present? && SiteConfig.github_secret.present?
+
+      # We need a dummy client to access Faraday::HttpCache storage class
+      # this client never actually makes any github requests so credentials are not needed
+      client = Github::OauthClient.new(client_id: "placeholder")
+      middleware_cache = client.middleware.handlers.detect { |h| h == Faraday::HttpCache }
+
+      # Our Github Oauth client uses Faraday::HttpCache (https://github.com/forem/forem/blob/master/app/services/github/oauth_client.rb#L114)
+      # to cache request results which can be served in event of an error.
+      # The problem with this is every time we fetch github repos we add another request to
+      # our Redis key AND we reset the expiration back to 24 hours. This can cause the key in Redis
+      # to grow uncontrollably. To prevent the key from getting too large we delete it once a day.
+      middleware_cache.build.__send__("storage").delete(GITHUB_REPOS_URL)
+    end
+  end
+end
--- a/config/schedule.yml
+++ b/config/schedule.yml
@@ -110,4 +110,7 @@ get_podcast_episodes:
 update_latest_github_repos:
   cron: "30 16 * * *" # daily at 4:30 pm UTC
   class: "GithubRepos::UpdateLatestWorker"
+bust_http_github_repo_cache_key:
+  cron: "0 1 * * *" # daily at 1 am UTC
+  class: GithubRepos::ClearHttpCacheWorker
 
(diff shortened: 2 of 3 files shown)


app/workers/github_repos/clear_http_cache_worker.rb
```diff
@@ -0,0 +1,24 @@
+module GithubRepos
+  class ClearHttpCacheWorker
+    include Sidekiq::Worker
+    GITHUB_REPOS_URL = "https://api.github.com/user/repos?per_page=100".freeze
```
mstruve, line 4: I tried to figure out a way I could get the URL for the repos endpoint through the Oktokit gem but it is created in the request so I hardcoded it here. I know its not ideal but its the best way I could think to handle this.

app/workers/github_repos/clear_http_cache_worker.rb
```diff
@@ -0,0 +1,24 @@
...
+      # this client never actually makes any github requests so credentials are not needed
+      client = Github::OauthClient.new(client_id: "placeholder")
+      middleware_cache = client.middleware.handlers.detect { |h| h == Faraday::HttpCache }
+
+      # Our Github Oauth client uses Faraday::HttpCache (https://github.com/forem/forem/blob/master/app/services/github/oauth_client.rb#L114)
+      # to cache request results which can be served in event of an error.
+      # The problem with this is every time we fetch github repos we add another request to
+      # our Redis key AND we reset the expiration back to 24 hours. This can cause the key in Redis
+      # to grow uncontrollably. To prevent the key from getting too large we delete it once a day.
+      middleware_cache.build.__send__("storage").delete(GITHUB_REPOS_URL)
```
maestromac, line 21: might be a stupid question here but is this just a single cache used in every request that is causing all the trouble?

mstruve: Yep! And it will grow infinitely unless it expires https://github.com/sourcelevel/faraday-http-cache/blob/02274d8639f80ec3821d8efa9a824213882443ab/lib/faraday/http_cache/storage.rb#L45 or a PUT/POST/DELETE request is sent to the same URL and does not receive an error message.

citizen428:
Another possible solution, though it's not necessarily prettier, just less reliant on `Faraday::HttpCache` internals (though hardcoding some other assumptions):

1. This caching mechanism uses the Rails cache.
2. We only use this cache for this specific request (it's the only place where I see the middleware added).
3.  So if the key name is stable/easily generatable, we could use `Rails.cache.delete(key)` and just wipe the whole Faraday cache completely in one go.

If I'm not mistaken, for us the cache key would be (based on [this](https://github.com/sourcelevel/faraday-http-cache/blob/02274d8639f80ec3821d8efa9a824213882443ab/lib/faraday/http_cache/storage.rb#L152-L155)):

```ruby
Digest::SHA1.hexdigest("Marshal#{GITHUB_REPOS_URL}")
```

This way the worker would essentially just be this:
```ruby
CACHE_KEY = Digest::SHA1.hexdigest("Marshalhttps://api.github.com/user/repos?per_page=100").freeze

def perform
  Rails.cache.delete(CACHE_KEY)
```

mstruve:
@citizen428 We could totally do that, I tried using the gem's methods instead of just hard coding the cache key incase the gem changes the cache key at some point.

mstruve:
Going to close and try to go the route of removing the cache.

PR #10343, opened 2020-09-16
fix: Start podcast record spinning animation from where it paused

reobin:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When pausing the podcast through the spinning record, its position used to be reset back to its initial position.

By using the [animation-play-state](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-play-state) CSS property, it is possible to keep the animation defined in both playing and paused states, but playing/pausing it accordingly.

N/A

1. Navigate to any podcasts
2. Play/pause the podcast using the podcast record image

### Before

before

### After

after

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

opening doors


app/assets/stylesheets/config/_variables.scss
```diff
@@ -94,4 +94,10 @@
 
   --lh-tight: 1.25; // 20px for 16px font-size
   --lh-base: 1.5; // 24px for 16px font-size
+
+  ////////////////////////////////////////////////////////////////////////////////////
+
+  // Podcasts
+
+  --podcast-spinning-animation: spin 20s linear infinite;
```
ludwiczakpawel, line 102: Can we keep this one outside of `config/_variables.scss`? This file is more for _very global_ variables and this particular variable feels like something very specific...

ludwiczakpawel (changes requested):

app/assets/stylesheets/config/_variables.scss
reobin: Quick related questions:
- If I was to add a common CSS class that would handle the spin animation between the two places it's used, would the liquid tag of existing articles be updated with the new class?
- How do you manage migration for updates to liquid tags?
- Since the processed HTML of an article is stored in the database, does it ever get processed again?

Thanks!

reobin: Actually, that was a question for another time. The liquid tag doesn't need an update for this right now.

ludwiczakpawel (approved):

fdoxyz (approved):
Thank you for the contribution @reobin

PR #10344, opened 2020-09-16
Optimization:Use Locally Bundled Gems in Containers

mstruve:
- [x] Optimization

## Description
I noticed when trying to use Docker with BuildKite that we were downloading and installing all of our gems rather than using our local vendor'ed cache. This updates our container bundle script to use our local gem cache which will save us a bit of time. 

alt_text


jdoss (approved):
LGTM. I can't believe I vendored the gems but forgot this! :facepalm:

citizen428 (approved):

PR #10345, opened 2020-09-16
Feature:Introduce Buildkite pipeline.yml file

mstruve:
- [x] Feature

## Description
I am in the process of getting [Buildkite](https://buildkite.com/) set up for us to use. I recently enabled a Github webhook which will trigger on pushes to this repo. This adds the file necessary for that step to return green right now despite not being fully configured. The file will only trigger an actual pipeline build for my Buildkite branches and anything else will automatically return green since the branch name is not a match. 

This step is not required but if you are like me you like to see that green dot on your PRs, this will keep them nice and clean 😉 

NOTE: This build shows the buildkite step passing. I deactivated the Webhook after seeing it fail and block another branch from being merged. I will enable it after this is merged. 

alt_text

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,13 @@
+steps:
+  - name: ":rspec:"
+    command: "scripts/ci/parallel_specs.sh"
+    artifact_paths: "log/**/*"
+    env:
+      RAILS_ENV: test
+    plugins:
+      - docker-compose#v3.5.0:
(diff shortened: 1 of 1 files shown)


rhymes (approved):
Let's try this!

mstruve:
Figured out how to trigger builds for my branch from Buildkite and figured out you can deactivate a webhook without removing it. So many learnings today. Closing this until the first pipeline is ready :)

PR #10349, opened 2020-09-16
Remove need for ENV vars during asset precompile

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

On the systems side of things, it is freeing to be able to run `RAILS_ENV=production rails assets:precompile` in a context without environment variables present. However, we require a few ENV vars just to kick off any Rails task. This pull request removes the few times we need environment variables to compile assets.

### Bonus enabling functionality ✨

By introducing the sensible constraint that assets cannot include environment variables, we create a situation where all Forems can share the same static assets, so all Forem Cloud customers can user `https://forem.cloud/assets/whatever-1493723.js` etc. in order to introduce major efficiencies across the network while also not doing anything to stop individual forems to run with assets hosted on that forem's domain.

This philosophy of progressive enhancement where all Forems _can_ live on an island, but there are efficiencies gained if they share certain components, is pretty cool. We're doing similar things at other layers of the stack. This PR does not include the full functionality to make that situation a reality, but it enables that once we want to make it happen.


app/controllers/open_search_controller.rb
```diff
@@ -0,0 +1,8 @@
+class OpenSearchController < ApplicationController
```
mstruve, line 1: For my own education, what is this for?

mstruve (approved):
🔥 🔥 🔥

app/controllers/open_search_controller.rb
benhalpern: We currently have `search.xml` in our assets, and this moves it to a controller for the purpose of this PR, but in general  it's a file to be compatible with this: https://developer.mozilla.org/en-US/docs/Web/OpenSearch ...which allows users to search through a site within their browser. It's not the most important thing in the world but it's a protocol we support.

Screen Shot 2020-09-16 at 11 22 19 AM

mstruve: Neat, TIL! Thanks!

jdoss (approved):

PR #10350, opened 2020-09-16
Bug Fix/Optimization:Remove HttpCache from Github OauthClient

mstruve:
- [x] Bug Fix
- [x] Optimization

## Description
Our Github Oauth client uses [Faraday::HttpCache](https://github.com/forem/forem/blob/master/app/services/github/oauth_client.rb#L114) to cache request results which can be served in event of an error. The problem with this is every time we fetch Github repos we add another request to our Redis key AND we reset the expiration back to 24 hours. This can cause the key in Redis to grow uncontrollably. When it gets really large we start to timeout trying to add to it. You can see this problem in the response time Honeycomb graph below. You can also see the time spent writing to Redis in the next image which is of a Honeycomb trace. To prevent this I am removing the cache. 
Screen Shot 2020-09-14 at 4 01 16 PM
Screen Shot 2020-09-15 at 11 03 52 AM

https://github.com/orgs/forem/projects/6#card-45548773

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/services/github/oauth_client.rb
+++ b/app/services/github/oauth_client.rb
@@ -111,7 +111,6 @@ module Github
       # <https://github.com/octokit/octokit.rb#caching>
       # and <https://github.com/octokit/octokit.rb/blob/master/lib/octokit/default.rb>
       Faraday::RackBuilder.new do |builder|
-        builder.use Faraday::HttpCache, store: Rails.cache, serializer: Marshal, shared_cache: false
         builder.use Faraday::Request::Retry, exceptions: [Octokit::ServerError]
         builder.use Octokit::Middleware::FollowRedirects
         builder.use Octokit::Response::RaiseError


rhymes (approved):
Let's do it! :D 🔥

maestromac (approved):

app/services/github/oauth_client.rb
```diff
@@ -111,7 +111,6 @@ def faraday_middleware_stack
       # <https://github.com/octokit/octokit.rb#caching>
       # and <https://github.com/octokit/octokit.rb/blob/master/lib/octokit/default.rb>
       Faraday::RackBuilder.new do |builder|
-        builder.use Faraday::HttpCache, store: Rails.cache, serializer: Marshal, shared_cache: false
```
citizen428, line 114: I think this is the only place where we use it, so we should probably also remove it from the `Gemfile`.

mstruve: ooooh good call! Will make a follow up and see what breaks, if nothing 🔪

PR #10351, opened 2020-09-16
Refactor:Remove Unused Faraday Http Cache Gem

mstruve:
- [x] Refactor

## Description
Remove Unused Faraday Http Cache Gem. Followup to #10350. Going to let #10350 bake for a couple days before removing the gem which is why I am keeping this in Draft mode for now. 

alt_text


citizen428 (approved):
I'll pre-approve this. If we decide not to bring the cache back, we definitely want to remove the gem since it's not used elsewhere.

rhymes (approved):
Adding that if we want to bring the cache back for some reason we could consider switching to https://github.com/lostisland/faraday_middleware/wiki/Caching#simple-response-body-caching which has a simpler implementation

PR #10352, opened 2020-09-16
Fix case sensitive published bug for basic markdown posts

Zhao-Andy:
- [x] Bug Fix

## Description
This fixes a bug where articles that are using front matter (also known as v1 articles or basic markdown articles) with `Published: true` will fail to publish. This is because the `published` attribute is case sensitive in the `front_matter_parser` gem we use, and therefore needs to be properly lowercased before sending it to the gem for processing.

1. Go to `localhost:3000/settings/ux` and update your article version to "basic markdown"
2. Click "Write a Post"
3. Try to publish a post with `Published: true`
4. See that it publishes

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

An elephant shows the capital and lowercase letter P.

### DIFF
--- a/app/labor/markdown_fixer.rb
+++ b/app/labor/markdown_fixer.rb
@@ -4,7 +4,7 @@ class MarkdownFixer
   class << self
     def fix_all(markdown)
       methods = %i[
-        add_quotes_to_title add_quotes_to_description
+        add_quotes_to_title add_quotes_to_description lowercase_published
         modify_hr_tags convert_new_lines split_tags underscores_in_usernames
       ]
       methods.reduce(markdown) { |acc, elem| public_send(elem, acc) }
@@ -33,6 +33,12 @@ class MarkdownFixer
     def modify_hr_tags(markdown)
       markdown.gsub(/-{3}.*?-{3}/m) do |front_matter|
         front_matter.gsub(/^---/).with_index { |match, i| i > 1 ? "#{match}-----" : match }
+      end
+    end
+
+    def lowercase_published(markdown)
+      markdown.gsub(/-{3}.*?-{3}/m) do |front_matter|
+        front_matter.gsub(/^published: /i, "published: ")
       end
     end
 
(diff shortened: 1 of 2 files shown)


app/labor/markdown_fixer.rb
```diff
@@ -36,6 +36,12 @@ def modify_hr_tags(markdown)
       end
     end
 
+    def lowercase_published(markdown)
+      markdown.gsub(/-{3}.*?-{3}/m) do |front_matter|
+        front_matter.gsub(/^published: /i, "published: ")
```
Zhao-Andy, line 41: This is:
`/^published: /i`
`/newline,published,colon,1-whitespace-character/case-insensitive-flag`

mstruve (approved):
NICE!

app/labor/markdown_fixer.rb
```diff
@@ -36,6 +36,12 @@ def modify_hr_tags(markdown)
       end
     end
 
+    def lowercase_published(markdown)
```
joshpuetz, line 39: Suggestion: are any front-matter tags allowed to be capitalized? Should we lowercase ALL of them?

spec/labor/markdown_fixer_spec.rb
```diff
@@ -7,6 +7,7 @@ def front_matter(title: "", description: "")
     <<~HEREDOC
       ---
       title: #{title}
+      published: false
```
joshpuetz, line 10: Suggestion: Is there a end-to-end test (or test for `fix-all`) this could be set to `Published: false` in?

joshpuetz (approved):
Some suggestions made, none are blocking. 👍🏻

app/labor/markdown_fixer.rb
vaidehijoshi: Yeah, I would second this. Just thinking that, unless we have a good enough reason not to, we may as well be consistent in terms of what we allow through markdown (seems like we're heading in the direction of being more forgiving re: case).

vaidehijoshi (approved):

benhalpern (approved):

spec/labor/markdown_fixer_spec.rb
```diff
@@ -98,6 +99,16 @@ def front_matter(title: "", description: "")
     end
   end
 
+  describe "::lowercase_published" do
```
benhalpern, line 102: Good test 👍

PR #10355, opened 2020-09-16
Redesigned Comment Form

ludwiczakpawel:

- Refactor
- Feature

## Description

This PR introduces new styling for Comments Form.

[Demo](https://d.pr/v/3vpQ5M).

image

image

image

## Added tests?

- no, because I need help

## Added to documentation?

- no documentation needed

## ToDos

- [x] Issue with Templates when user tries to use them in a thread reply. cc @Zhao-Andy 
- [x] Editing comments
- [x] Inserting templates on Comment edit view.
- [x] Specs


ludwiczakpawel:
@Zhao-Andy Because I've changed some frontend code I have some issues with [Response Templates code](https://github.com/forem/forem/blob/master/app/javascript/responseTemplates/responseTemplates.js#L255-L277).. It happens in two scenarios:

1. When I hit "Reply" on existing comment and try to insert a template - nothing happens.
2. When I hit "Reply" on existing comment and then try to "Dismiss" there's JS error in console.

Is this something you could help me with?

ludwiczakpawel:
I have issues with specs / tests.. Is this something @rhymes or @nickytonline could help me with?

Zhao-Andy:
Okay, figured out #1 re: response templates. If you can change the `callback` code (starting at line 266 in `app/javascript/responseTemplates/responseTemplates.js` to this:

```js
  const callback = (mutations) => {
    const form = Array.from(mutations[0].addedNodes).filter(node => node.nodeName === 'FORM');
    if (form.length > 0) {
      prepareOpenButton(form[0]);
    }
  };
```

That should work!

<hr>

Re: the dismiss button, I didn't see any errors in the console when I hit "Reply" then "Dismiss", but when I dismissed a comment reply, hitting "Reply" again would do nothing.

ludwiczakpawel:
@Zhao-Andy sweet, thanks it works!

Re: the dismiss button, ok i've just figured it was my fault and fixed it, thanks! :)

lisasy:
Looking great! There are some small UX bugginess here, and I think we can resolve them together.

1. **"Submit" button is at the default enabled state even when the textarea is empty.** Could we add some conditional here so that it is disabled until there is text content in the textarea?
2. Related to #1, "Preview" is also at its default enabled state even when the textarea is empty.
3. Once you open "Templates", it is persistently visible until you close it. However, this creates some strange UX with if you decide to "preview" the content because the elements swap positions; see this video here: https://share.getcloudapp.com/bLuw8JAJ
4. Once you're in "Preview" mode, you have to select "Edit" button to go back to editing mode... When I was trying this out, something felt *off* to me with the language of these toggles. Could a label like "Continue editing" help connect Preview and Edit mode? Perhaps improving the UX in #3 will make this "issue" seem less jarring.

app/views/comments/_form.html.erb
```diff
@@ -28,73 +27,90 @@
...
         <div class="personal-responses-container">
           <%# filled by JS %>
         </div>
         <div class="moderator-responses-container hidden">
           <%# filled by JS %>
         </div>
-        <a target="_blank" rel="noopener nofollow" href="<%= user_settings_path(tab: "response-templates") %>" class="mod-response-create-new">
+
+        <a target="_blank" rel="noopener nofollow" href="<%= user_settings_path(tab: "response-templates") %>">
           Create new template
```
lisasy, line 91: I think that instead of saying "Create new template", we can just say "Create template". There is no need for the adjective "new" here.

app/views/comments/_form.html.erb
```diff
@@ -28,73 +27,90 @@
...
-          <%= inline_svg_tag("external-link-logo.svg", aria: true) %>
         </a>
         <p>Create templates to quickly answer FAQs or store snippets for re-use.</p>
```
lisasy, line 93: Rather than start this sentence by repeating the line directly above it, we could say something like:

"Templates let you quickly answer FAQs or store snippets for re-use."

app/views/comments/_form.html.erb
ludwiczakpawel: Very good suggestion! I like short labels :)

ludwiczakpawel:
Thanks for feedback @lisasy ! Just FYI this is still draft PR :)

> "Submit" button is at the default enabled state even when the textarea is empty. Could we add some conditional here so that it is disabled until there is text content in the textarea?

~~Yeap, that's should be simple thing to do.~~ It took me few hours to figure it out actually but it's done :D 


> Once you open "Templates", it is persistently visible until you close it. However, this creates some strange UX with if you decide to "preview" the content because the elements swap positions; see this video here:

Yea, I see.. Unfortunately this will have to be fixed later on. Seems like a bigger issue here and I don't want to increase size of this PR anymore. FWIW that UX is similar to what we have currently. Also, I'm not worried about postponing fixing it since not that many people use templates anyway.

> Once you're in "Preview" mode, you have to select "Edit" button to go back to editing mode... When I was trying this out, something felt off to me with the language of these toggles. Could a label like "Continue editing" help connect Preview and Edit mode? Perhaps improving the UX in #3 will make this "issue" seem less jarring.

Makes sense, label fixed.

---

Also, your copy suggestion are also implemented.

app/views/comments/_form.html.erb
lisasy: Me too!

lisasy:
> Yea, I see.. Unfortunately this will have to be fixed later on. Seems like a bigger issue here and I don't want to increase size of this PR anymore. FWIW that UX is similar to what we have currently. Also, I'm not worried about postponing fixing it since not that many people use templates anyway.

That makes a lot of sense to me!

rhymes (approved):
Works great! I wasn't able to test the reply comment feature in notifications.

ludwiczakpawel:
@rhymes I haven't touched this one but I have tested it locally and seemed working (meaning: i didn't break it :D)

lisasy:
Re-reviewing and it's lookin' good! The small nit I saw were the different button sizes here. 

Screen Shot 2020-10-02 at 4 10 11 PM

lisasy (approved):
Other than the different button label sizes, it looks good!

I'll approve of this and you can make that small fix.

Issue #10341, opened 2020-09-15
Unable to Remove Hearts on Comment Chain; Heart is Duplicated Instead


michael-tharrington:
**Describe the bug**

The logic behind hearts (❤️ reactions) appears to be off in certain situations — e.g. when a user is liking comments in a comment chain. Occasionally in comment chains, users can't remove a heart by clicking on it and instead the heart doubles.

**To Reproduce**

1. Make a comment on a test post.
2. Heart the comment and reply to your first comment with a second comment.
3. If you try to remove the heart on the first comment, the heart actually gets duplicated and appears as 2 hearts.
5. Try un-hearting it, it doesn't go away.

**Expected behavior**

Clicking your hearts should always make them go away and never duplicate.

**Video**

Here's a linked video of this in action - https://share.getcloudapp.com/NQurPZ5k

**Additional context**

A user originally reported this to us as a security concern, but we felt this was more of a bug than a security flaw.

souldzin:
Hey! Friendly heads up that we're going to try to recreate this and put together a PR for it :wave:

michael-tharrington:
Rock on, thanks for the notice!

souldzin:
Hey! We created https://github.com/forem/forem/pull/10357 but it seems like now in it's current state, it only addresses part of this related problem. I don't necessarily have capacity to contribute more to this so I'm going to gracefully unassign myself 😅

michael-tharrington:
All good @souldzin, appreciate the efforts!!

mikeydorje:
Code review suggests this may be resolved or no longer applicable due to changes in this area since the issue was filed. Closing due to age and low activity. Please reopen with current repro steps if this persists.

PR #10357, opened 2020-09-16
Fix bug that shows all self-authored as reacted

souldzin:
Co-authored with @lauraxm

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR addresses https://github.com/forem/forem/issues/10341, by removing some lines that were severely confusing the Frontend when creating and reacting to self-created comments. 

Previously (on `master`):
- When the FE loads **any** comment that was self created, it is rendered as "liked" even if I specifically unliked it.

Now (in this PR):
- We use the API response to determine what is "reacted" or not.

Closes https://github.com/forem/forem/issues/10341

1. Make a comment on a test post
2. Refresh the page.
3. Unlike the comment.
4. Refresh the page.
5. The comment should show as unliked (currently on `master` it will show as liked)

## Added tests?

- [ ] yes
- [x] no, because I need help

It probably makes the most sense to add something to `spec/system/comments/user_fills_out_comment_spec.rb`. I can do this but will need to get rspec running locally :smile:

## Added to documentation?

I'm not sure if there's any preexisting documentation for this "auto like your own comments" feature :shrug:

- [ ] docs.forem.com
- [ ] readme
- [ ] no documentation needed

uncertain

## Edits

There was some confusion on the requirements here. Originally, this MR also changed: 

<details>

Previously (on `master`):
- When a comment is created, we render it as "liked" on the frontend (FE), but the backend (BE) actually doesn't recognize this.

Now (in this PR):
- When a comment is created, we do not render it as "liked" so that it is consistent with the BE.

We believe this new behavior is also consistent with user expectations (it can come across a little haughty to like your own comments).

## QA Instructions, Screenshots, Recordings (Old)

In this MR's previous state, you could test the changes by:

https://youtu.be/WhYLclxpLz0

Instructions taken from https://github.com/forem/forem/issues/10341

1. Make a comment on a test post. The comment should be unhearted by default (matches BE behavior).
2. Heart the comment and reply to your first comment with a second comment.
3. If you try to remove the heart on the first comment, the heart is removed.

</details>

### DIFF
--- a/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
+++ b/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
@@ -48,7 +48,6 @@ function initializeCommentsPage() {
                     userActionsEl.innerHTML = '<a data-no-instant href="' + userActionsEl.parentNode.dataset.path + '/delete_confirm" class="edit-butt">DELETE</a>\
                                                 <a href="' + userActionsEl.parentNode.dataset.path + '/edit">EDIT</a>'
                     userActionsEl.style.display = 'inline-block';
-                    document.getElementById('button-for-comment-' + allNodes[i].dataset.commentId).classList.add('reacted');
                   }
                 }
               }


app/assets/javascripts/initializers/initializeCommentsPage.js.erb
```diff
@@ -48,7 +48,6 @@ function initializeCommentsPage() {
                     userActionsEl.innerHTML = '<a data-no-instant href="' + userActionsEl.parentNode.dataset.path + '/delete_confirm" class="edit-butt">DELETE</a>\
                                                 <a href="' + userActionsEl.parentNode.dataset.path + '/edit">EDIT</a>'
                     userActionsEl.style.display = 'inline-block';
-                    document.getElementById('button-for-comment-' + allNodes[i].dataset.commentId).classList.add('reacted');
```
souldzin, line 51: **Is it safe to remove this?**

This is the bit that was walking through every comment that was self created and rendering it as "liked". This creates inconsistencies with the BE which then causes us to enter weird states when the BE tells us `result: "create"`

app/assets/javascripts/utilities/buildCommentHTML.js.erb
```diff
@@ -95,10 +95,9 @@ function actions(comment) {
 
 function reactions(comment) {
   if (comment.newly_created) {
-    return '<button class="reaction-button reacted" id="button-for-comment-'+comment.id+'" data-comment-id="'+comment.id+'">\
```
souldzin, line 98: **What is  this change doing?**

The BE does **not** treat newly created comments as "reacted" so it is misleading when we render it as such. It can also force the BE to get into weird states (which is another issue altogether...)

souldzin:
@lauraxm here's the PR we worked on together 😄

app/assets/javascripts/initializers/initializeCommentsPage.js.erb
fdoxyz: I agree that this is most likely a problem (always render all self created comments as "liked"). The AJAX response  however (`responseObj`) does include useful data that could possibly be used to put this statement behind an `if` clause.

Adding a debug similar to `console.log("REACTIONS RES:", responseObj)` can help understand the body of the response. Also from the network tab in the devtools you can parse the response from the request and see something like this:

Screen Shot 2020-09-17 at 10 41 28

You can see that the payload includes the id of the `current_user` up top and all the reaction IDs on the Article. So this new `if` statement to protect the rendering of all self created comments as "liked" should only be true when this `reactions` array  includes the comment with id `allNodes[i].dataset.commentId` and the `user_id` from that reaction matches the `current_user`'s ID.

fdoxyz (changes requested):
Thank you both for taking a stab at this issue! This feels like a tricky one, you've made good progress but I left some inline comments to provide a bit more context.

> It probably makes the most sense to add something to spec/system/comments/user_fills_out_comment_spec.rb. I can do this but will need to get rspec running locally 😄

Excellent idea, it will definitely be a valuable test to include. I haven't tried it yet so I'm not certain if Sidekiq will kick in and create that first reaction from within the test suite. Just something to keep an eye out when working on this spec.

app/assets/javascripts/utilities/buildCommentHTML.js.erb
souldzin: > There's an after_create_commit callback here that executes the following:

Ah! I see.

Is there a way to make this transactional such that the comment and reaction are created at the same time? Is that a change you think will be helpful?

There's a small performance tradeoff, but I imagine creating a reaction is a pretty small task that it's negligible :shrug:

app/assets/javascripts/initializers/initializeCommentsPage.js.erb
souldzin: > The AJAX response however (responseObj) does include useful data that could possibly be used to put this statement behind an if clause.

I'll look into this, but I'm pretty sure this is already being handled here:

https://github.com/osfn/forem/blob/11e9e20e4f839cf512b54b688c3e2c9f2d53ddfc/app/assets/javascripts/utilities/buildCommentHTML.js.erb#L102

In our testing, comments that we actually liked still showed the full heart even on refresh.

souldzin: > but I'm pretty sure this is already being handled here:

Oh wait! It's actually looks like it's being handled here 😅:

https://github.com/osfn/forem/blob/11e9e20e4f839cf512b54b688c3e2c9f2d53ddfc/app/assets/javascripts/initializers/initializeCommentsPage.js.erb#L25

fdoxyz: > Oh wait! It's actually looks like it's being handled here 

You're right, I missed that too. Looks like this may be an unnecessary line after all AFAICT

app/assets/javascripts/utilities/buildCommentHTML.js.erb
fdoxyz: At the moment I think adding better handling of the AJAX response when reacting could fix the issue. The JSON payload includes `"result"` which is either `create`/`destroy`. This means a new reaction is created or an existing reaction is destroyed, there's no in between [in the controller](https://github.com/fdoxyz/forem/blob/master/app/controllers/reactions_controller.rb#L68-L80).

What this means is that even if the race condition happens in a production environment (i.e. high load on work queue) and two reactions are created, once they're both created then the destroy should happen twice and it should go back to normal (no more than 1 reaction from the comment author).

I will get back to you on this when I get a chance to dive a bit deeper on the topic again, because it's also a bug I can't seem to replicate as easily.

souldzin: > This means a new reaction is created or an existing reaction is destroyed, there's no in between in the controller.

This seems like we could be committing to something even more complex, but I trust your experience on this topic 😅 💪

souldzin: Moved this change out of this PR so that it's not blocking the other bug fix which is happening in `app/assets/javascripts/initializers/initializeCommentsPage.js.erb `

souldzin:
@fdoxyz, would you mind reviewing this PR again? It sounds like the FE and BE sync will be a bit involved which I won't have capacity for.

I removed that part out and left the line removal that fixes the issue where all self-authored comments are viewed as reacted even if I intentionally unliked one.

fdoxyz:
Hi @souldzin , thanks for simplifying the changes and following up with the PR!

I'm away for the moment but I'll without a doubt double check and test this locally again when I'm back next week.

I'll also try to think of a way we can cover this behavior, most likely using system tests.

souldzin:
> I'm away for the moment but I'll without a doubt double check and test this locally again when I'm back next week.

Thanks so much! :smile:

souldzin:
Updated this PR title and description to match it's current scope.

fdoxyz:
Thank you for submitting the PR @souldzin and sorry we weren't able to confirm this properly in time. The `initializeCommentsPage.js.erb` was more thoroughly refactored with a recent update to the comments UI and this doesn't seem to be a problem anymore. 

Comments are indeed reacted by default, that's the expected behavior (sharing a recording of a test I did). If you still encounter this problem or think there's something wrong please let me know and we can re-review the situation but for now I'm closing the PR

https://share.getcloudapp.com/4guOb54l

souldzin:
> The initializeCommentsPage.js.erb was more thoroughly refactored with a recent update to the comments UI and this doesn't seem to be a problem anymore.

Nice! I see that the line this PR removed was also removed in the refactor :+1: 

Thanks for the update! :wave:

PR #10361, opened 2020-09-17
Add opacity to HexComparer

citizen428:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As requested by @ludwiczakpawel, this PR adds an opacity method to `HexComparer` which turns a hex color into an RGBA value:

```ruby
HexComparer.new(['#123456']).opacity(0.5)
#=> "rgba(18, 52, 86, 0.50)"
```

Following the lead of `brightness`, this uses the smallest of the hex values if more than one was provided. Maybe we should rethink if this class really is a `HexComparor` or if we should call it something else, e.g. `ColorTools`. In the meantime I just refactored the code a bit.

Your friendly neighborhood Slack request.

Nothing special, you can play arouind with it in a Rails console to make sure it behaves as expected.

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

alt_text

### DIFF
--- a/app/labor/hex_comparer.rb
+++ b/app/labor/hex_comparer.rb
@@ -1,45 +1,34 @@
 class HexComparer
+  RGB_REGEX = /^#?(?<r>..)(?<g>..)(?<b>..)$/.freeze
+  ACCENT_MODIFIERS = [1.14, 1.08, 1.06, 0.96, 0.9, 0.8, 0.7, 0.6].freeze
+
   def initialize(hexes, amount = 1)
-    @hexes = hexes
+    @hexes = hexes.sort
     @amount = amount
   end
 
-  def order
-    hexes.sort
-  end
-
   def smallest
-    order.first
+    hexes.first
   end
 
   def biggest
-    order.last
+    hexes.last
   end
 
   def brightness(amount = 1)
-    rgb = smallest.delete("#").scan(/../).map(&:hex).map { |color| color * amount }.map(&:round)
-    format("#%<r>02x%<g>02x%<b>02x", r: rgb[0], g: rgb[1], b: rgb[2])
+    rgb = smallest.match(RGB_REGEX).named_captures.map do |key, color|
+      [key.to_sym, (color.hex * amount).round]
+    end.to_h
+    format("#%<r>02x%<g>02x%<b>02x", rgb)
   rescue StandardError
     smallest
   end
 
+  # Returns the first valid hex string it finds (# + 6 digits)
   def accent
-    if brightness(1.14**amount).size == 7
-      brightness(1.14**amount)
-    elsif brightness(1.08**amount).size == 7
-      brightness(1.08**amount)
-    elsif brightness(1.06**amount).size == 7
-      brightness(1.06**amount)
-    elsif brightness(0.96**amount).size == 7
-      brightness(0.96**amount)
-    elsif brightness(0.9**amount).size == 7
-      brightness(0.9**amount)
-    elsif brightness(0.8**amount).size == 7
-      brightness(0.8**amount)
-    elsif brightness(0.7**amount).size == 7
-      brightness(0.7**amount)
-    elsif brightness(0.6**amount).size == 7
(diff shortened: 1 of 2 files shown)


app/labor/hex_comparer.rb
```diff
@@ -1,49 +1,49 @@
 class HexComparer
+  RGB_REGEX = /^#?(?<r>..)(?<g>..)(?<b>..)$/.freeze
```
citizen428, line 2: This is now a regex with named captured groups, so we don't have to manually assemble rgb hashes.

app/labor/hex_comparer.rb
```diff
@@ -1,49 +1,49 @@
...
+  ACCENT_MODIFIERS = [1.14, 1.08, 1.06, 0.96, 0.9, 0.8, 0.7, 0.6].freeze
+
   def initialize(hexes, amount = 1)
-    @hexes = hexes
+    @hexes = hexes.sort
```
citizen428, line 6: We never really use the `order` method from the outside, so inlining the sorting and removing the method seemed sensible.

app/labor/hex_comparer.rb
```diff
@@ -1,49 +1,49 @@
...
     @amount = amount
   end
 
-  def order
-    hexes.sort
-  end
-
   def smallest
-    order.first
+    hexes.first
```
citizen428, line 11: `hexes` is already sorted now.

app/labor/hex_comparer.rb
```diff
@@ -1,49 +1,49 @@
...
   end
 
   def biggest
-    order.last
+    hexes.last
```
citizen428, line 15: `hexes` is already sorted now.

app/labor/hex_comparer.rb
```diff
@@ -1,49 +1,49 @@
...
   end
 
   def brightness(amount = 1)
-    rgb = smallest.delete("#").scan(/../).map(&:hex).map { |color| color * amount }.map(&:round)
-    format("#%<r>02x%<g>02x%<b>02x", r: rgb[0], g: rgb[1], b: rgb[2])
+    rgb = hex_to_rgb_hash(smallest).transform_values do |color|
```
citizen428, line 19: Leaning a bit more on Ruby here. `scan` followed by 3 map operations wasn't particularly efficient, this seems easier and I can reuse `hex_to_rgb_hash` in `opacity`.

app/labor/hex_comparer.rb
```diff
@@ -1,49 +1,49 @@
...
-      brightness(0.96**amount)
-    elsif brightness(0.9**amount).size == 7
-      brightness(0.9**amount)
-    elsif brightness(0.8**amount).size == 7
-      brightness(0.8**amount)
-    elsif brightness(0.7**amount).size == 7
-      brightness(0.7**amount)
-    elsif brightness(0.6**amount).size == 7
-      brightness(0.6**amount)
+    ACCENT_MODIFIERS.each do |modifier|
```
citizen428, line 29: The previous solution was (a) verbose and (b) calculated the final brightness twice, once for comparing the length, once for returning it. I instead use `each` + `break` with a return value.

app/labor/hex_comparer.rb
```diff
@@ -1,49 +1,49 @@
 class HexComparer
+  RGB_REGEX = /^#?(?<r>..)(?<g>..)(?<b>..)$/.freeze
+  ACCENT_MODIFIERS = [1.14, 1.08, 1.06, 0.96, 0.9, 0.8, 0.7, 0.6].freeze
```
rhymes, line 3: 🔥

rhymes (approved):
Cool!

ludwiczakpawel (approved):
I know nothing about Ruby but I'll rely on @rhymes approval and give a green light too :D

PR #10364, opened 2020-09-17
Hide column backup_data.json_data from Blazer via Hypershield

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This column contains PII and should be hidden via Hypershield: `json_data` is the JSON dump of an `Identity` object


mstruve (approved):
I think we should remove backup_data completely re https://github.com/forem/forem/pull/10365 but in the meantime, lets lock this down

PR #3377, opened 2019-07-03
Add functionality to remove and recover identity

Zhao-Andy:
## What type of PR is this? (check all applicable)

- [x] Feature

## Description
This PR adds the ability for admins to remove identities, as well as recover them. There are instructions in the view that explain the situation(s) when an identity should be removed.

This also adds the groundwork for backing up data via a database table, which is essentially a table like this:

| instance_type | instance_id | instance_user_id | json_data |
| ------------- | ------------- | ------------- | ------------- |
| identity | 1 | 1 | `{ id: 1, uid: 1, auth_data_dump: { lots_of_data } }` |

## Mobile & Desktop Screenshots/Recordings (if there are UI changes)
Screen Shot 2019-07-03 at 8 21 25 AM

PR #10365, opened 2020-09-17
Cleanup orphan data from buffer_updates

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Prior to add foreign keys to the following 5 columns - https://dev.to/admin/blazer/queries/245-orphan-rows - I need to cleanup the rows. 

This PR removes backup data that refers to identities of users that don't exist anymore and nullifies `composer_user_id` for those rows that have a value that ceased to be valid


lib/data_update_scripts/20200917135827_remove_orphan_rows_from_backup_data_by_instance_user_id.rb
```diff
@@ -0,0 +1,13 @@
+module DataUpdateScripts
+  class RemoveOrphanRowsFromBackupDataByInstanceUserId
+    def run
+      # Delete all BackupData rows belonging to Users that don't exist anymore
+      ActiveRecord::Base.connection.execute(
+        <<~SQL.squish,
+          DELETE FROM backup_data
```
mstruve, line 7: I am all for cleaning this up but I am wondering, this is a pretty old feature https://github.com/forem/forem/pull/3377, is this something we still feel is necessary? I think if a user deletes their identity it should be gone, we shouldn't be storing that data in a backup. What do you think about removing this table and code altogether?

cc @Zhao-Andy

mstruve (commented):
buffer_updates looks good, had a thought on the backup data table.

lib/data_update_scripts/20200917135827_remove_orphan_rows_from_backup_data_by_instance_user_id.rb
Zhao-Andy: Hmm yeah the feature itself definitely needs a revisit. It was definitely built to handle some edge cases at the time that we don't really run into anymore, or just have better ways of solving the same issue.

@michael-tharrington is the delete identity feature something you still use?

mstruve: I think the question is not quite do we delete identities, bc that is something we still do for sure and allow users to do. The question is do we need these backups?

rhymes: This functionality is used only in the admin, so no users are allowed to access their backups:

https://github.com/forem/forem/blob/19c8d219bc79b3d9b5476e8fe685f8a840a4b292/app/controllers/admin/users_controller.rb#L83-L87

https://github.com/forem/forem/blob/19c8d219bc79b3d9b5476e8fe685f8a840a4b292/app/controllers/admin/users_controller.rb#L97-L101

Zhao-Andy: Ah @mstruve @rhymes good points. Yeah these backups are really just for "just in case", and in that case, it might be better to scrap this table and feature. If we do ever want it again in the future, we can build it to be more flexible -- like allowing users to access backup data.

rhymes: Let's wait for @michael-tharrington feedback before we proceed either way. I'm all for removing this table but maybe it's needed due to our problems with merging identities

michael-tharrington: I used to use the remove identity function regularly for folks with login issues but I haven't had to use it in the past couple of months. Hopefully the instance in which I needed to use it previously has been fixed. In any case, I _think_ this would be fine to remove.

mstruve: 🙌  Lets 🔪  it!

rhymes:
@mstruve rebased this on master and removed the script for the table that was removed :)

lib/data_update_scripts/20200917135827_remove_orphan_rows_from_backup_data_by_instance_user_id.rb
rhymes: Done!

mstruve (approved):

Zhao-Andy (approved):

PR #10368, opened 2020-09-17
Add and apply crayons link header style

lisasy:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

- Add `crayons-link--header` style as optional headers for navigation link-blocks
- Unify styling of header with `crayons-link--header` in Notification and Mod Center

Note:
The styling of the link blocks in Notification is slightly incorrect, but I will update that in a separate PR where we apply `crayons-layout` to the view.

Proposed:
Screen Shot 2020-09-17 at 10 06 55 AM
Screen Shot 2020-09-17 at 10 06 51 AM

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

alt_text


app/views/moderations/_mod_sidebar_left.html.erb
```diff
@@ -3,7 +3,7 @@
     Mod Center
   </h1>
   <nav class="mod-center--nav" aria-label="Mod center inbox navigation">
-    <h3 class="fw-bold fs-s mod-center--nav__title">
+    <h3 class="crayons-link--header">
```
ludwiczakpawel, line 6: Since it's not a link, should it be called "crayons-link"?

I imagine it could be used in different scenarios in sidebar not only in context of links (think of form filters etc..)

Maybe simply `crayons-header`? Or actually `crayons-subheader`  since we may turn all page titles into its own component at some point too?

app/assets/stylesheets/components/navigation.scss
```diff
@@ -54,6 +54,14 @@
   }
 }
 
+.crayons-link--header {
+  color: var(--link-color);
```
ludwiczakpawel, line 58: I think we can safely go with `var(--base-90)` to not be links-oriented.

app/views/moderations/_mod_sidebar_left.html.erb
lisasy: Good catch!

If it is `crayons-subheader`, which file do you think would be an appropriate one of put this? 
And re: `crayons-header`, I think it could be handy to define this as well, since we want to have the same consistent type styling for all page titles :)

ludwiczakpawel: I think you can create new component file: `components/headers.scss` and put your code for `.crayons-subheader`. I can then (in a separate PR) add styling for `crayons-header` too :)

app/views/notifications/_nav_menu.html.erb
```diff
@@ -1,9 +1,9 @@
 <nav class="notification-nav">
   <% if @organizations.present? %>
-    <header class="pl-2 mb-2">
-      <h4>
+    <header>
```
ludwiczakpawel, line 3: One last extra piece of feedback: IF `<header>` doesn't have any extra styling, we can drop it from the markup.

ludwiczakpawel (approved):

juliannatetreault (approved):
This looks good to me! 👍 👍

PR #10370, opened 2020-09-17
Removing ElasticSearch Brew instructions

cmgorton:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description
If you try to install brew as directed you get this Error:
Error: Calling Installation of elasticsearch-oss from a GitHub commit URL is disabled! Use 'brew extract elasticsearch-oss' to stable tap on GitHub instead.

Brew/ElasticSearch no longer allows you to tap as we previously have suggested in the doc. Also we can no longer grab the 7.5.2 version with brew.

We have decided to remove the instructions for now until we upgrade ElasticSearch if possible.
Screen Shot 2020-09-17 at 2 10 18 PM
## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [x] docs.forem.com
- [ ] readme
- [ ] no documentation needed

### DIFF
--- a/docs/installation/mac.md
+++ b/docs/installation/mac.md
@@ -20,7 +20,7 @@ Please refer to their [installation guide](https://yarnpkg.com/en/docs/install).
 
 ### PostgreSQL
 
-Forem requires PostgreSQL version 11 or higher to be running.
+Forem requires PostgreSQL version 11 or higher to run.
 
 The easiest way to get started is to use
 [Postgres.app](https://postgresapp.com/). Alternatively, check out the official
@@ -38,7 +38,7 @@ You can install ImageMagick with `brew install imagemagick`.
 
 ### Redis
 
-Forem requires Redis version 4.0 or higher to be running.
+Forem requires Redis version 4.0 or higher to run.
 
 We recommend using [Homebrew](https://brew.sh):
 
@@ -61,7 +61,7 @@ redis-cli ping
 
 ### Elasticsearch
 
-Forem requires a version of Elasticsearch 7.5 or higher to be running. We recommend version 7.5.2.
+Forem requires Elasticsearch 7.x to run. We recommend version 7.5.2.
 
 You have the option of installing Elasticsearch with Homebrew or through an archive. We **recommend** installing from archive on Mac.
 
--- a/docs/installation/mac.md
+++ b/docs/installation/mac.md
@@ -12,8 +12,7 @@ title: macOS
    [rbenv](https://github.com/rbenv/rbenv). Please follow their
    [installation guide](https://github.com/rbenv/rbenv#installation).
 2. With the Ruby version manager, install the Ruby version listed on our badge.
-   (i.e. with rbenv: `rbenv install *latest badge version*`) 
-    example: `rbenv install 2.7.1`
+   (i.e. with rbenv: `rbenv install $(cat .ruby-version)`)
 
 ### Yarn
 
@@ -62,10 +61,9 @@ redis-cli ping
 
 ### Elasticsearch
 
-Forem requires a version of Elasticsearch between 7.1 and 7.5 to be running. Version 7.6 is
-not supported. We recommend version 7.5.2.
+Forem requires a version of Elasticsearch 7.5 or higher to be running. We recommend version 7.5.2.
 
-We recommend installing from archive on Mac.
+You have the option of installing Elasticsearch with Homebrew or through an archive. We **recommend** installing from archive on Mac.
 
 ### Installing Elasticsearch from the archive
 
@@ -106,6 +104,26 @@ To start elasticsearch as a daemonized process:
 ```shell
 ./bin/elasticsearch -d
 ```
+
+### Installing Elasticsearch with Homebrew
+
+To install Elasticsearch with Homebrew we will use the following commands to: 
+- tap the Elastic Homebrew repository
+- install the latest OSS distribution
+- pin the latest OSS distribution. 
+
+```shell
+brew tap elastic/tap
+brew install elastic/tap/elasticsearch-oss
+brew pin elasticsearch-oss
+```
+
+After installation you can manually test if the Elasticsearch server starts by
+issuing the command `elasticsearch` in the shell. You can then start the server
+as a service with `brew services start elasticsearch-oss`.
+
+You can find further info on your local Elasticsearch installation by typing
+`brew info elastic/tap/elasticsearch-oss`.
 
 #### Troubleshooting startup issues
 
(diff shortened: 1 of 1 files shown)


docs/installation/mac.md
```diff
@@ -12,15 +12,16 @@ title: macOS
    [rbenv](https://github.com/rbenv/rbenv). Please follow their
    [installation guide](https://github.com/rbenv/rbenv#installation).
 2. With the Ruby version manager, install the Ruby version listed on our badge.
-   (i.e. with rbenv: `rbenv install $(cat .ruby-version)`)
+   (i.e. with rbenv: `rbenv install *latest badge version*`) 
+    example: `rbenv install 2.7.1`
```
nickytonline, line 16: ``(i.e. with rbenv: `rbenv install $(cat .ruby-version)`)`` should stay as it is reading the `.ruby-version` file from the forked Forem repository.

If this happens before the docs mention grabbing the repository, we should rework that too.

nickytonline (commented):
Looks good, just one comment.

docs/installation/mac.md
cmgorton: Yes it is happening before there is mention of forking the repo but we can mention to fork the repository first instead and keep `(i.e. with rbenv: `rbenv install $(cat .ruby-version)`)`

docs/installation/mac.md
```diff
@@ -61,11 +62,10 @@ redis-cli ping
 
 ### Elasticsearch
 
-Forem requires a version of Elasticsearch between 7.1 and 7.5. Version 7.6 is
+Forem requires a version of Elasticsearch between 7.1 and 7.5 to be running. Version 7.6 is
 not supported. We recommend version 7.5.2.
```
rhymes, line 66: I think we should update this. In production we run 7.5.2 (though it technically runs with 7.9 as well but it's not officially supported), we don't really know if it still works on 7.1 to 7.4.

Maybe we should mention 7.5 or higher, like we do for PostreSQL?

docs/installation/mac.md
```diff
@@ -107,25 +107,6 @@ To start elasticsearch as a daemonized process:
 ./bin/elasticsearch -d
 ```
 
-### Installing Elasticsearch with Homebrew
```
rhymes, line 110: If we switch to "it runs on 7.5 or higher" we could in theory keep these instructions, after all the official docs still support ES via Homebrew: https://www.elastic.co/guide/en/elasticsearch/reference/current/brew.html

(I'm using `elasticsearch-oss` 7.9.1 via `brew tap elastic/tap` in development)

The risk is that someone will install ES 8 when it comes out and maybe it won't work so I'm also in favor of scrapping them.

If we keep them I'd replace them with something like:

```
brew tap elastic/tap
brew install elastic/tap/elasticsearch-oss
brew pin elasticsearch-oss
```

The rest of the post installation instructions would be the same.

rhymes (commented):
Great find @cmgorton, seems like our instructions are a bit outdated :( 

Left notes about how we could avoid scraping all the homebrew instructions by instead updating them

docs/installation/mac.md
joshpuetz: Given my recent conversations with @mstruve and @jdoss, it seems like "unofficially" supporting 7.5 and above is a good step. I'd MUCH rather keep in the homebrew installation instructions if we can!

(@cmgorton and I went 10 rounds with `brew` yesterday and there's just no good/easy way to install older versions of Elasticsearch anymore)

mstruve: I'm cool with bumping it to 7.x version that works.

cmgorton: Ok, maybe I should close this PR for now and create a Draft PR until we all come to a consensus on brew and Elasticsearch. Thoughts?

joshpuetz: I'd be comfortable just changing this PR to a) mentioned we use Elasticsearch 7.5 and up, and b) changing the home-brew instructions to pull right from the latest version like @rhymes mentioned at https://github.com/forem/forem/pull/10370#discussion_r490709809!

cmgorton: Hey I am circling back on this as I know I let it lapse. I had a community member who was going to pair with Nick this week unable to get to get the installation working because of elasticsearch. Should I recommend he pull from the latest version like @rhymes and @joshpuetz mentioned?

If we are still ok with that I can also update this PR to do the same.

joshpuetz: I'd vote for specifying the latest version of Elasticsearch!

mstruve: Any 7.x version of Elasticsearch will work and can be recommended

cmgorton:
I made updates per everyones recommendation:

Added back line 15 per Nicks recommendation  _(i.e. with rbenv: `rbenv install $(cat .ruby-version)`)_

Line 65  updated to  7.5 or higher for Elasticsearch per Rhymes recommendation
_Forem requires a version of Elasticsearch 7.5 or higher to be running.  We recommend version 7.5.2._

Line 110 update:
Added back in the Homebrew instructions but updated commands to `install elastic/tap/elasticsearch-oss`

docs/installation/mac.md
```diff
@@ -61,11 +61,9 @@ redis-cli ping
 
 ### Elasticsearch
 
-Forem requires a version of Elasticsearch between 7.1 and 7.5. Version 7.6 is
-not supported. We recommend version 7.5.2.
+Forem requires a version of Elasticsearch 7.5 or higher to be running. We recommend version 7.5.2.
```
mstruve, line 64:
```suggestion
Forem requires Elasticsearch 7.x to run. We recommend version 7.5.2.
```

docs/installation/mac.md
```diff
@@ -38,7 +38,7 @@ You can install ImageMagick with `brew install imagemagick`.
 
 ### Redis
 
-Forem requires Redis version 4.0 or higher.
+Forem requires Redis version 4.0 or higher to be running.
```
mstruve, line 41:
```suggestion
Forem requires Redis version 4.x to run.
```

mstruve (approved):
LGTM! I left a couple of wording suggestions for the versions to keep our major versions consistent but otherwise I like it!

docs/installation/mac.md
cmgorton: Yeah, I like this suggestions! Thank you Molly.

cmgorton: Updated with these suggestions along with the section on PostgreSQL that also had "to be running." Thanks for the suggestions @mstruve

joshpuetz (approved):

PR #10372, opened 2020-09-17
Replace need for GitHub ENV vars on bootup

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We still had this scenario where production (via eager loading of classes) won't boot without GitHub keys.

Currently we're forced to include dummy github keys in ENV because the `GitHub` module gets loaded at boot and triggers the sequence of events that leads to the error being thrown if those credentials are not set.

This changes the code to instead call an instance where the `SiteConfig` is available.


mstruve (approved):
Few more calls to Redis for that SiteConfig but I like this!

benhalpern:
Yeah, @mstruve it's a little less efficient but none of this is high-traffic work.

I think there could be space for future optimizations of how we use `SiteConfig` all around.

PR #10375, opened 2020-09-18
More detailed notification when setup is not completed

lightalloy:

- [x] Feature

## Description
Added information about missing fields for admins when the setup is not completed

изображение

## Added tests?
- [x] yes

### DIFF
--- a/app/controllers/concerns/verify_setup_completed.rb
+++ b/app/controllers/concerns/verify_setup_completed.rb
@@ -28,7 +28,11 @@ module VerifySetupCompleted
   end
 
   def setup_completed?
-    MANDATORY_CONFIGS.all? { |config| SiteConfig.public_send(config).present? }
+    missing_configs.empty?
+  end
+
+  def missing_configs
+    MANDATORY_CONFIGS.reject { |config| SiteConfig.public_send(config).present? }
   end
 
   private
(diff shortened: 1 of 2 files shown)


rhymes (approved):
LOVE this. The only issue I can think of is that for a new installation this notification might be listing too many values? If so, maybe we can set a cap? Something like "missing a, b, c and others" =

app/controllers/concerns/verify_setup_completed.rb
```diff
@@ -28,7 +28,11 @@ module VerifySetupCompleted
   end
 
   def setup_completed?
-    MANDATORY_CONFIGS.all? { |config| SiteConfig.public_send(config).present? }
+    missing_configs.empty?
```
citizen428, line 31: This way we're going to iterate all the `MANDATORY_CONFIGS` twice if the setup hasn't been completed yet, once for this check, then another time for generating the message. I wonder if there's a better way to do this, e.g. getting rid of `setup_completed?` renaming `missing_configs` to `missing_mandatory_configs` and changing `verify_setup_completed` like this:

```ruby
missing_configs = missing_mandatory_configs
return if config_path? || missing_configs.empty? || SiteConfig.waiting_on_first_user

link = ...
 missing_configs_text = missing_configs.map...
```

app/controllers/concerns/verify_setup_completed.rb
```diff
@@ -37,8 +41,9 @@ def verify_setup_completed
     return if config_path? || setup_completed? || SiteConfig.waiting_on_first_user
 
     link = helpers.link_to("the configuration page", admin_config_path, "data-no-instant" => true)
+    missing_configs_text = missing_configs.map { |c| c.to_s.tr("_", " ") }.join(", ")
```
citizen428, line 44: We can maybe address @rhymes concern here (but this would require making the other suggested change, otherwise we check the mandatory configs too many times):

```ruby
display_missing = missing_configs.size > 3 ? missing_configs.first(3) + ["others"] : missing_configs
missing_configs_text = display_missing.map { |c| c.to_s.tr("_", " ") }.to_sentence
```

This would generate the following texts:

1 element: "community description"
2 elements: "community description and community action"
3 elements: "community description, community action, and tagline"
More than 3 elements: "community description, community action, tagline, and others"

citizen428 (approved):

lightalloy:
@rhymes @citizen428 thanks for the feedback, I will implement your suggestions :+1:

PR #10384, opened 2020-09-19
Refactoring user_controller separating some classes to service

brunnohenrique:

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Remove some methods at `controllers/admin/users_controllers.rb` to solve a codeclimate issue: over than 20 methods

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

alt_text


app/services/users/manage_credits.rb
```diff
@@ -0,0 +1,50 @@
+module Users
+  class ManageCredits
+    def self.call(*args)
+      new(*args).call
+    end
+
+    def initialize(user, user_params)
+      @user = user
+      @organization_id = user_params[:organization_id]
```
jonatasrancan, line 9: Instead of using all this variables, what do you think of use just `@user_params = user_params`

And in each place you access what you need.

app/services/users/manage_credits.rb
```diff
@@ -0,0 +1,50 @@
...
+      @add_credits_param = user_params[:add_credits]
+      @remove_credits_param = user_params[:remove_credits]
+      @add_org_credits_param = user_params[:add_org_credits]
+      @remove_org_credits_param = user_params[:remove_org_credits]
+    end
+
+    def call
+      add_credits if add_credits_param
+      remove_credits if remove_credits_param
```
jonatasrancan, line 18: Since you handle credits for the `user` and credits for `orgs`.
Lets make this more explicitly like this:

```ruby
def call
  manage_user_credits
  manage_org_credits
end
```
And inside of each one of this new methods, you call the correct ones.

app/services/users/manage_credits.rb
```diff
@@ -0,0 +1,50 @@
...
+      add_org_credits if add_org_credits_param
+      remove_org_credits if remove_org_credits_param
+    end
+
+    private
+
+    attr_reader :user, :add_credits_param, :add_org_credits_param, :remove_org_credits_param,
+                :remove_credits_param, :organization_id
+
+    def add_credits
```
jonatasrancan, line 28: Instead of calling this methods with the inline `if` to check if they can run.

What do you think of using `early returns`?

```ruby
def add_credits
  return unless user_params[:add_credits]
  ....
end
```

app/services/users/manage_credits.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    def add_org_credits
+      org = Organization.find(organization_id)
+      amount = add_org_credits_param.to_i
+      Credit.add_to(org, amount)
+    end
+
+    def remove_org_credits
+      org = Organization.find(organization_id)
```
jonatasrancan, line 45: You are `loading` the organization twice.

Lets load it once. Something like this:


```ruby
def organization
  @organization ||= Organization.find(user_params[:organization_id])
end
```

spec/services/users/manage_credits_spec.rb
```diff
@@ -0,0 +1,64 @@
+require "rails_helper"
+
+RSpec.describe Users::ManageCredits, type: :service do
+  describe "When the the service is called" do
+    subject(:invoke_service) { described_class.call(user, user_params) }
+
+    let(:user) { create(:user) }
+    let(:organization) { create(:organization) }
+
+    context "when user add the proper amount of credits" do
```
jonatasrancan, line 10: I think just `adding credits to the user`, is a proper context description.

spec/services/users/manage_credits_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      let(:user_params) do
+        { add_credits: "4" }
+      end
+
+      before { invoke_service }
+
+      it { expect(user.reload.credits_count).to eq 4 }
```
jonatasrancan, line 17: A better way to test this is to check if the `credits_count` was updated correctly, something like this:

```ruby
it "adds the correct amount of credits to the user" do
 expect {
   invoke_service
 }.to change { user.reload }.by(user_params[:add_credits])
end
```

or

```ruby
it "adds the correct amount of credits to the user" do
 expect {
   invoke_service
 }.to change { user.reload }.to(user_params[:add_credits])
end
``

If it's really necessary to check `unspent_credits_count`, then do it in the same way

File: spec/services/users/manage_credits_spec.rb [hunk 7]
```diff
@@ -0,0 +1,64 @@
...
+        { remove_credits: "4" }
+      end
+
+      before do
+        Credit.add_to(user, 10)
+        invoke_service
+      end
+
+      it { expect(user.reload.credits_count).to eq 6 }
+      it { expect(user.reload.unspent_credits_count).to eq 6 }
```
jonatasrancan, line 32: Same comment as `adding credits to the user` on how to test this.

spec/services/users/manage_credits_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+        { add_credits: "4" }
+      end
+
+      before { invoke_service }
+
+      it { expect(user.reload.credits_count).to eq 4 }
+      it { expect(user.reload.unspent_credits_count).to eq 4 }
+    end
+
+    context "when user removes the proper amount of credits" do
```
jonatasrancan, line 21: A better description: `removing credits from the user`

spec/services/users/manage_credits_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+        { remove_org_credits: "6", organization_id: organization.id }
+      end
+
+      before do
+        create(:organization_membership, user: user, organization: organization, type_of_user: "admin")
+        Credit.add_to(organization, 10)
+        invoke_service
+      end
+
+      it { expect(organization.reload.credits_count).to eq 4 }
```
jonatasrancan, line 46: Same comment as previous on how to test this.

spec/services/users/manage_credits_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      let(:user_params) do
+        { add_org_credits: "7", organization_id: organization.id }
+      end
+
+      before do
+        create(:organization_membership, user: user, organization: organization, type_of_user: "admin")
+        invoke_service
+      end
+
+      it { expect(organization.reload.credits_count).to eq 7 }
```
jonatasrancan, line 60: Same comment as previous on how to test this.

app/services/users/manage_credits.rb
brunnohenrique: In this case, I don't know if it can return after the first verification. To do this I need more information from the business case.
For example on the older way, it did verify all the methods and didn't return.

app/controllers/admin/users_controller.rb
```diff
@@ -33,7 +33,7 @@ def show
 
     def update
       @user = User.find(params[:id])
-      manage_credits
+      Users::ManageCredits.call(@user, user_params)
```
citizen428, line 36: Since this handles user and org credits, I'd prefer `Credits::Manage` as a name.

app/services/users/manage_credits.rb
```diff
@@ -0,0 +1,50 @@
+module Users
+  class ManageCredits
+    def self.call(*args)
```
citizen428, line 3: I know you just copied a pattern we use elsewhere, but I'm not a big fan of obscuring the params of the main entry point of a service object. 

```ruby
def self.call(user, user_params)
  new(user, user_params).call
end
```

This also allows you to have different signatures between `.call` and `#initialize`, e.g.

```ruby
def self.call(user, user_params)
  new(
    user,
    user_params.slice(:add_credits, :remove_credits),
    user_params.slice(:organization_id, :dd_org_credits, :remove_org_credits)
  )
end

def initiliaze(user, user_params, org_params)
  ...
end
```

rhymes (approved):
Love it! I think you've addressed all of the requested changes

Thanks @brunnohenrique

app/services/credits/manage.rb
```diff
@@ -54,7 +54,7 @@ def remove_org_credits
     end
 
     def organization
-      Organization.find(org_params[:organization_id])
+      @organization ||= Organization.find(org_params[:organization_id])
```
brunnohenrique, line 57: Just an improvement appointed by @jonatasrancan 
Ty!

citizen428 (approved):

PR #10412, opened 2020-09-21
Update Imgproxy to use predetermined endpoint

maestromac:
- [x] Feature

## Description
Instead of dedicating an endpoint env var for Imgproxy, we will always default to `/images` to the domain of the Forem instance it belongs to. The also change how we check if Imgproxy should be enabled; now it will check for both KEY and SALT instead of KEY and Endpoint.

https://github.com/orgs/forem/projects/11#card-44712228

This does not change any existing behavior and only applicable on production.
## Added tests?
- [x] updated existing tests

## Added to documentation?
- [x] no documentation needed

n/a

### DIFF
--- a/config/initializers/imgproxy.rb
+++ b/config/initializers/imgproxy.rb
@@ -4,8 +4,13 @@ Imgproxy.configure do |config|
   # Full URL to where your imgproxy lives.
   #
   config.endpoint = if Rails.env.production?
-                      URL.url("images")
+                      # Use /images with the same domain on Production as
+                      # our default configuration
+                      URL.url("images") # ie. https://forem.dev/images
                     else
+                      # On other environments, rely on ApplicationConfig for a
+                      # more flexible configuration
+                      # ie. default imgproxy endpoint is localhost:8080
                       ApplicationConfig["IMGPROXY_ENDPOINT"]
                     end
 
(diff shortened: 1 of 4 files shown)


jdoss (approved):

config/initializers/imgproxy.rb
```diff
@@ -2,7 +2,12 @@
   # imgproxy endpoint
   #
   # Full URL to where your imgproxy lives.
-  config.endpoint = ApplicationConfig["IMGPROXY_ENDPOINT"]
+  #
+  config.endpoint = if Rails.env.production?
+                      URL.url("images")
```
mstruve, line 7: Why only for production? Why not use this for all environments?

maestromac: Having imgproxy set up with same domain as the rails app under `/images` require special openresty configuration. This gives more flexibility to local setups.

mstruve: Gotcha, it would be awesome if we could remove this somehow but since that might be hard we should add a comment about why it is here. It is always a little bit of a red flag to me when I see `if Rails.env.production?` statements in the code but at least a comment will explain a bit what is going on.

mstruve (approved):
small suggestion otherwise LGTM!

rhymes (approved):
Cool! So glad we're getting closer to having imgproxy all the way :D

PR #10417, opened 2020-09-22
Custom dependencies with more detailed debug logs

fdoxyz:
For Debugging/Testing pursposes only


fdoxyz:
Debugging succesful:

```ruby
OmniAuth.config.full_host = URL.url
```

PR #10419, opened 2020-09-22
Sets Omniauth full_host config to match the environment

fdoxyz:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The reason Facebook Authentication provider hasn't been working was due to a mismatch in the redirect URL sent in the OAuth process. 

This PR explicitly adds to the Omniauth initializer so it will rely on `URL.url` to build the redirect URL for callbacks.

## Related Tickets & Documents

fixes #10200 and #10102

## QA Instructions, Screenshots, Recordings

All authentication providers in the local development should continue to work as expected. This is a recording of the Facebook authentication working in https://thismmalife.com

https://share.getcloudapp.com/WnuJLJq2

## Added tests?

- [ ] yes
- [x] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

## [optional] What gif best describes this PR or how it makes you feel?

Visual representation of the debugging process (aka rabbit hole):
rabbit hole

PR #10423, opened 2020-09-22
Revert "Sets Omniauth full_host config to match the environment"

mstruve:
Reverts forem/forem#10419

Just in case


fdoxyz (approved):
I'm okay with reverting if we can't check DEV's GitHub OAuth app config

PR #10428, opened 2020-09-24
Crayonisification of "Reading list"

ludwiczakpawel:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR brings a bit of Crayons styling to the "Reading list" view. It also renames "Reading list" to "Saved" in various places to be more in line with how we really want to treat that page.

This PR does **not** address two things though:

1. It doesn't bring responsiveness but that's ok, it will come with another PR.
2. ~~It doesn't rename any backend stuff from "readinglist" to "saved" because I don't have enough skills to do so :D~~ [Click](https://github.com/forem/forem/pull/10428#issuecomment-698754128)

image

## Added tests?

- [ ] yes
- [x] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed


citizen428:
> It doesn't rename any backend stuff from "readinglist" to "saved"

To be honest, I'm not a big fan of this. I went through a lot of trouble once to rename everything on the backend to make the names for listings consistent between frontend and backend, so deliberately doing this here rubs me the wrong way. I'd much prefer if we kee this PR to style changes, then discuss and decide whether we really want to rename this and do it in one go for frontend and backend.

ludwiczakpawel:
@citizen428 Thanks for feedback! TBH renaming is something we've been having few internal conversations about, so we were kinda leaning towards making that change. BUT this hasn't been discussed more broadly with rest of the team so I'll revert that renaming change and we can bring it back once we're all on the same page. Thanks for pointing this out!

citizen428:
Thanks, @ludwiczakpawel. 😺 I agree with the renaming in principle. I just don't want to do it for the frontend now and then we somehow don't get around to do the backend and we end up having two distinct terms for the same concept.

mstruve (approved):
Tested this out and works and looks great!

app/views/layouts/_nav_menu.html.erb
```diff
@@ -12,7 +12,7 @@
     <div class="border-0 border-b-1 border-solid border-base-20 p-1">
       <a href="<%= dashboard_path %>" class="crayons-link crayons-link--block">Dashboard</a>
       <a href="<%= new_path %>" class="crayons-link crayons-link--block">Write a Post</a>
-      <a href="<%= readinglist_path %>" class="crayons-link crayons-link--block">Reading List</a>
+      <a href="<%= readinglist_path %>" class="crayons-link crayons-link--block">Reading list</a>
```
rhymes, line 15: I think we should keep the capitalization here, as all the other entries have it

rhymes (commented):
Hi @ludwiczakpawel, great work! Aside from the capitalization detail, the load more button looks a bit odd: 

Screenshot_2020-09-28 Reading List - DEV(local) Community 👩‍💻👨‍💻

Should it be so attached to the last item?

app/views/layouts/_nav_menu.html.erb
ludwiczakpawel: We don't have it in the main nav and I think our tendency is to keep only first letter capitalized.. But truth to be told - I don't care that much right now as I hope to rename that whole place to be "Saved" :D

ludwiczakpawel:
@rhymes spacing issue fixed, thanks!

benhalpern (approved):

PR #10429, opened 2020-09-24
Crayonsification of Listings form

ludwiczakpawel:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR introduces new Crayons-like design for adding and editing listings.

image

image

## Added tests?

- [ ] yes
- [x] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed


app/views/listings/_form.html.erb
```diff
@@ -37,59 +38,71 @@
...
       <% else %>
-        <p>You need at least one credit to create a listing.</p>
-        <a href="/credits/new" class="cta cta-main-listing-form" data-no-instant>Purchase Credits Now</a>
-      <% end %>
-    </div>
-    <p><em><%= community_name %> admins may modify tags and categories to ensure accuracy. We may also take down and refund inappropriate listings.</em></p>
-    <p><em>Listings expire 30 days after they are published or "bumped".</em></p>
-    <div class="listings-current-credits">
-      <% if @credits.size == 0 && (@organizations.blank? || @organizations.sum(:unspent_credits_count).zero?) %>
-        <h3>Listings Require Credits</h3>
```
fdoxyz, line 76: The failing spec seems to be caused by a check for this String to appear visible. Maybe `You need at least one credit to create a listing` would be the best replacement for this check 🙂

fdoxyz (commented):
Looking fresh with Crayons!

mstruve:
I pulled this down to test it and after running `bin/setup` and completely stopping and starting my server I am still seeing some weird behavior. I load the listings page and the listings show for a second and then disappear. No errors or anything on the page.

ludwiczakpawel:
@mstruve that's weird - I haven't touched (intentionally) Listings `Index` or `Show` views.. I have only worked on `New` and `Edit` views 🤔

But I tried to do the same: `bin/setup`, `bin/startup` and still everything seems fine, including `Index`, `Show`, `New` and `Edit` views.

Can you maybe provide some screenshots / recording?

app/javascript/listings/__tests__/ContactViaConnect.test.jsx
```diff
@@ -16,26 +16,20 @@ describe('<ContactViaConnect />', () => {
...
+    const { getByText } = render(
       <ContactViaConnect onChange={onChange} checked />,
     );
 
-    const checkbox = getByLabelText(
-      'Allow Users to Message Me Via In-App Chat (DEV Connect)',
-    );
+    const checkbox = getByText('Allow Users to message me via Connect.');
 
     fireEvent.input(checkbox, { target: { checked: true } });
```
fdoxyz, line 34: I saw the HTML was a bit different. In practice it works is the same way, but with the new structure the query doesn't return the element for some reason (or the event doesn't fire correctly). This worked for me to make `yarn test` pass locally though:

```suggestion
    fireEvent.click(checkbox);
```

mstruve:
Apparently, it happens for me on master as well so please ignore!

fdoxyz (approved):

ludwiczakpawel:
@lisasy can you try to clear your cache? also, can you see any errors in the console?

rhymes (commented):
Couple of details:

- the bump action has no super visible feedback. When I press bump listing the page refreshes but as it's scrolled down, I can't immediately see anything and it's not clear to me if the button is not working, or the action is too fast or whatever. I think we should find a way to clarify the feedback for that button.

- not sure if it's related but when i create or update a listing and get redirected to http://localhost:3000/listings - I can't see it. I see it for a split second and then it disappears in a flash. It also happens on master which might be related to the other PR that got merged? #10428

ludwiczakpawel:
> the bump action has no super visible feedback. When I press bump listing the page refreshes but as it's scrolled down, I can't immediately see anything and it's not clear to me if the button is not working, or the action is too fast or whatever. I think we should find a way to clarify the feedback for that button.

Does it have visible feedback for you on master? I couldn't see it so that's why it is the way it is in my branch. If you can't see it in master either, let's work on it in a separate PR.

> not sure if it's related but when i create or update a listing and get redirected to http://localhost:3000/listings - I can't see it. I see it for a split second and then it disappears in a flash. It also happens on master which might be related to the other PR that got merged? #10428

Why do you think this could be related to that Reading list PR? These are completely separate views. Did I miss something? :)

rhymes:
> Does it have visible feedback for you on master? I couldn't see it so that's why it is the way it is in my branch. If you can't see it in master either, let's work on it in a separate PR.

Nope, but maybe it's an easy fix 👍 

> Why do you think this could be related to that Reading list PR? These are completely separate views. Did I miss something? 

Sorry, I thought that PR was for the listings. I always conflate listings with reading list. It's the word list :D

lisasy:
> @lisasy can you try to clear your cache? also, can you see any errors in the console?

@ludwiczakpawel Phew, this works now.

rhymes (approved):
LGTM!

Let's circle back on the smaller issues in a separate PR. Thanks @ludwiczakpawel!

lisasy:
In terms of this PR being focused on applying Crayons styling to this form, it overall looks good. I think that there are some opportunities to improve the UX/language of this to communicate that credits are required to make a listing, but I'd say that's outside the scope of this.

ludwiczakpawel:
I agree with everything above!

PR #10431, opened 2020-09-24
Bug Fix:Update bottom content cache on releases

mstruve:
- [x] Bug Fix

## Description
We recently updated the `bottom_content` CSS which ended up breaking this partial cache bc the cache did not bust since there was no HTML change. To prevent this from happening I updated the cache key to include the release footprint. I'm not sure if we want this long term or if it will bust this cache too much to make it useful but at least for now it will fix production. Within 18 hours all old cache's with old styling will have expired and we could revert this. 

Me when it comes to cache bugs 👇 
alt_text

### DIFF
--- a/app/views/articles/show.html.erb
+++ b/app/views/articles/show.html.erb
@@ -215,7 +215,7 @@
 
       <%= render "articles/full_comment_area" %>
 
-      <% cache("article-bottom-content-#{@article.id}-#{user_signed_in?}", expires_in: 18.hours) do %>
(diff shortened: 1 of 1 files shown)


jdoss (approved):

ludwiczakpawel (approved):
🙌

mstruve:
I busted all of the partial Redis cache's manually in a console, I am now busting the Fastly caches for the article show pages. With that done this should be solved and we dont need this code so I am going to close for now.

PR #10432, opened 2020-09-24
Fixes Omniauth redirect_uri when SiteConfig.app_domain is updated

fdoxyz:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When `SiteConfig.app_domain` is updated in a existing (running) environment we don't get the correct `redirect_uri` generated for OAuth providers. This happens not because we rely on `ApplicationConfig["APP_DOMAIN"]` (since we use `URL.url` for this) but instead because the `full_host` Omniauth config is set in the initializer and it doesn't change if we update `SiteConfig.app_domain` unless we reboot (not what we want).

The fix consists of making `Omniauth.config.full_host` use a Proc instead of a String generated at boot time so the generated value will always use the up-to-date `SiteConfig.app_domain`

1. Boot the app locally and login with any OAuth provider (successfully)
1. Sign out (otherwise after executing the next step you'll be unable t logout bc of cookie domain settings)
1. Open a rails console and edit `SiteConfig.app_domain = "bad.domain.com"
1. Try to login with any OAuth provider and **it should fail** because the `redirect_uri` is now using an unauthorized URI
1. Revert `SiteConfig.app_domain` (to `localhost:3000` or whatever value you're using locally)
1. It should work again

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

None

no names

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -1,2 +1,2 @@
 OmniAuth.config.logger = Rails.logger
-OmniAuth.config.full_host = URL.url
+OmniAuth.config.full_host = proc { URL.url }
--- a/spec/system/authentication/omniauth_redirect_uri_spec.rb
+++ b/spec/system/authentication/omniauth_redirect_uri_spec.rb
@@ -1,0 +1,33 @@
+require "rails_helper"
+
+RSpec.describe "Omniauth redirect_uri", type: :system do
+  let!(:test_app_domain) { SiteConfig.app_domain }
+
+  # Avoid messing with other tests by resetting back SiteConfig.app_domain
+  after { SiteConfig.app_domain = test_app_domain }
+
+  def provider_redirect_regex(provider_name)
+    %r{
+      /users/auth/#{provider_name}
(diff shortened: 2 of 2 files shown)


mstruve (approved):
I was knee deep in the OmniAuth gem source code. I got [to here](https://github.com/octokit/octokit.rb/blob/61751879ecf178f0643f5926b38896d8c5c0ea1b/lib/octokit/client/authorizations.rb#L165) but then couldn't figure out the rest. This looks good and ca confirm on our broken env that the value is wrong for `OmniAuth.config.full_host`

fdoxyz:
Oh absolutely! It boils down all the way to the `callback_url` being built in Omniauth, which uses `full_host` [here](https://github.com/omniauth/omniauth/blob/master/lib/omniauth/strategy.rb#L423-L428) and is where we would now be using the Proc definition 😄

jdoss (approved):

PR #10435, opened 2020-09-25
Bug Fix:Count the Health Checks Controller as Public

mstruve:
- [x] Bug Fix

## Description
We need to be able to hit the health check API with only the health check token and no additional authentication. This allows us to do that for private forems

alt_text


jdoss:
:partying_face: Woooo!

jdoss (approved):

citizen428 (approved):

PR #10436, opened 2020-09-25
Feature:Rake Task to Return Health Check Token

mstruve:
- [x] Feature

## Description
Return health check token for @jdoss to use for Forem things.

### DIFF
--- a/lib/tasks/forem.rake
+++ b/lib/tasks/forem.rake
@@ -6,4 +6,8 @@ namespace :forem do
     ProfileField.destroy_all
     ProfileFields::AddBaseFields.call
   end
+
+  task health_check_token: :environment do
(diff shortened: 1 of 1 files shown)


jdoss:
MRW @mstruve makes me new toys...

mstruve:
@jdoss was also thinking, I could write this value to a file as well using this rake task if that would be easier to ingest for you. let me know :)

jdoss:
> @jdoss was also thinking, I could write this value to a file as well using this rake task if that would be easier to ingest for you. let me know :)

Kicking it out to stdout is perfect for now. If I need to write it to a file I can on my end easily.

jdoss (approved):

jacobherrington (approved):
:+1:

PR #10468, opened 2020-09-28
A11y: Add keyboard navigation to article feed

### LINKED ISSUE #596: Accessibility: Add keyboard navigation to feed
## Feature Request or Task

Add listeners for `j` & `k` to scroll up one or down a post in the feed. This would align it more with other feeds (Twitter, Tumblr, etc), and is a nice way to scroll through content.

This also goes along with https://github.com/thepracticaldev/dev.to/issues/121 and https://github.com/thepracticaldev/dev.to/issues/394.

### Definition of Done

Pressing `j` or `k` will initiate keyboard navigation, which will apply the `focus` state to the first fully-visible element in the feed (skipping any that are partially scrolled past/visible), and bring it to the top of the screen. Then, any subsequent presses of those keys will move one element up or down, respectively.

reobin:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This adds keyboard navigation to the article feed (and potentially more stuff).

Closes #596

### Keys

1. `j` goes down
2. `k` goes up

### Behavior

If no article is focused on (or anything inside an article), the first article that is visible in the feed is focused on

### Demo

demo

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

JSDoc added

## How it works

The hook can be called in any functional component.

```javascript
useGlobalListNavigation(
  'article[id=featured-story-marker],article[id^=article-]',
  'a[id^=article-link-]',
  'div.paged-stories',
);
```

### 3 params

1. `itemContainerSelector`: The selector for the highest level of the list item
2. `focusableSelector`: What should actually be focused on in the list item container
3. `waterfallItemContainerSelector` (**optional**): selector of the waterfall container if the list of items uses a waterfall-like architecture, like the article feed.

### Flow

It uses the new global key event listener hook (which is now used by the search bar for `/` and `0`) to listen for `j` and `k`.

On the first keypress, it focuses on the first visible article. Then it follows one of the next paths:

#### Happy path:

1. Get closest parent article of the currently focused-on element
2. Get next or previous sibling
3. Get the link of sibling
4. Focus on link

#### Dealing with the article feed waterfall

What I refer to as the waterfall:

```html
<article></article>
<article></article>
<article></article>
<div class="paged-stories">
  <article></article>
  <article></article>
  <article></article>
  <div class="paged-stories">
    <article></article>
    <article></article>
    <article></article>
  </div>
</div>
```

##### Going down the waterfall

1. Get closest parent article of the currently focused-on element
2. Get next sibling (is a paged-stories div)
3. Get the next article link in DOM (article link of the first article in the div)
4. Focus on the article link

##### Going up the waterfall

1. Get closest parent article of the currently focused-on element
2. Get previous sibling (there is none)
3. Get parent paged-stories div
4. Get previous sibling (is an article)
5. Get the article link
6. Focus on the article link


app/javascript/articles/Feed.jsx
```diff
@@ -1,47 +1,44 @@
...
 import PropTypes from 'prop-types';
 
+import useNavigation from '../utilities/hooks/useNavigation';
+
 /* global userData sendHapticMessage showModal buttonFormData renderNewSidebarCount */
 
-export class Feed extends Component {
-  constructor(props) {
-    super(props);
+export const Feed = ({ timeFrame, renderFeed }) => {
```
reobin, line 9: The `Feed` component was refactored to become a functional component.

app/javascript/utilities/hooks/useNavigation.js
```diff
@@ -0,0 +1,51 @@
...
+      let nextWrapper = !closestWrapper
+        ? getFirstElement()
+        : direction === 'up'
+        ? getPreviousElement(closestWrapper)
+        : getNextElement(closestWrapper);
+
+      const nextFocusable = nextWrapper?.querySelector(focusableSelector);
+      if (nextFocusable) {
+        if (EVENTFUL_KEYS.includes(event.key)) {
+          event.preventDefault();
```
reobin, line 43: Avoid screen scrolling when using arrow keys

app/javascript/articles/Feed.jsx
```diff
@@ -1,47 +1,44 @@
...
-  }
+  useEffect(() => {
+    setPodcastEpisodes(getPodcastEpisodes());
+  }, []);
 
-  componentDidMount() {
-    const { timeFrame } = this.props;
+  useEffect(() => {
+    const fetchFeedItems = async () => {
+      const feedItems = await getFeedItems(timeFrame);
```
reobin, line 24: This should be wrapped into a try catch. What should be done with an error on fetch?

reobin:
Tried the hook with the listings for fun, and it works well:

```js
import useNavigation from '../../utilities/hooks/useNavigation';

useNavigation('div[id^=single-listing-]', 'a.crayons-link');
```

app/javascript/utilities/hooks/useNavigation.js
```diff
@@ -0,0 +1,51 @@
+import { isInViewport } from '../viewport';
+import useKeyEventListener from './useKeyEventListener';
+
+const NAVIGATION_UP_KEYS = ['k', 'ArrowUp'];
+const NAVIGATION_DOWN_KEYS = ['j', 'ArrowDown'];
+const EVENTFUL_KEYS = ['ArrowUp', 'ArrowDown'];
+
+const useNavigation = (
```
reobin, line 8: `useNavigation` should be renamed to hold more meaning

app/javascript/utilities/hooks/useGlobalKeyEventListener.js
```diff
@@ -0,0 +1,32 @@
+import { useEffect } from 'preact/hooks';
+
+export const registerGlobalKeyEventListener = (keys, callback) => {
```
reobin, line 3: This is exported to support the use of global key event listener on class-based components

reobin:
I noticed a bit late that the article feed is quite different while the user is not logged in. I implemented the key navigation for that too but left it quite specific as it will probably never be reused.

When the logged out article feed is refactored to use preact, if it ever is, the specific code can be removed.

app/javascript/articles/Feed.jsx
fdoxyz: An option that comes to mind is using the Snackbar to tell the User there was a problem fetching the feed

fdoxyz (approved):
I tested this and it's working well thank you for the contribution @reobin!

As far as the implementation goes I'll defer to @nickytonline as he's more aware of the bigger picture approach in the frontend codebase. In a general sense it LGTM

app/javascript/articles/Feed.jsx
reobin: I'll take a look at how the Snackbar is currently used elsewhere, thanks!

app/views/stories/_main_stories_feed.html.erb
```diff
@@ -19,3 +19,89 @@
     <div class="single-article-small-pic" id="article-index-hidden-div" style="display:none"></div>
   <% end %>
 </div>
+
+<script>
```
benhalpern, line 23: Hmmmmm I'm not sure about plopping all of this in the HTML like this.

If so, let's add `defer` just because we typically defer JS tags, but yeah @nickytonline if you want to give more thoughts on the general arrangement here?

reobin: 100% agree on deferring the tag.

I would honestly rather have this code elsewhere, but it's so specific to the article feed (while logged off), and none of the code should be reused.

Feedback is welcome on how to handle the differences between:
1. logged off article feed in Rails
2. logged in article feed in preact
3. There are also differences in the structure of the HTML between the two

app/javascript/articles/Feed.jsx
reobin: @fdoxyz I experimented a bit with the snackbar and I'm not sure it 100% fits our needs here. Please don't hesitate if I'm in the wrong here.

If an error happens while fetching the feed, then that error is not temporary like a snackbar item would be. I have the feeling that the message that shows up in such a case should stay there until the page is refreshed, or until the user changes page. A snackbar item will disappear after 5 seconds, which would leave the user in an unknown state.

I thought about using a danger notice instead where the article feed would go:

Screen Shot 2020-09-30 at 17 56 58

What do you think?

rhymes (commented):
Haven't reviewed the code in detail but I see we're working with a global key event listener so that got me thinking if it'd make sense to consider using https://github.com/github/hotkey as well (cc @nickytonline)

I have an a11y related question: isn't this going to override the default behavior of navigating through tabbable elements? For example, I now can't use the arrows to navigate tags like I can do on dev.to:

tags

Now if I'm in the tags div and press arrow down it goes to the articles:

tags2

As you can see, I can't scroll anymore using the arrows, just the tab.

Is this a problem? I'm asking as I'm not entirely sure.

reobin:
That's a good point @rhymes !

I tested around and it turns out sites like twitter or facebook don't listen for the arrow keys, just `j` and `k`.

Perhaps it should be the case for forem as well, as the arrows imply a lot in terms of accessibility.

app/javascript/articles/Feed.jsx
```diff
@@ -50,19 +52,20 @@ export class Feed extends Component {
    *
    * @returns {Promise} A promise containing the JSON response for the feed data.
    */
-  static getFeedItems(timeFrame = '', page = 1) {
-    return fetch(`/stories/feed/${timeFrame}?page=${page}`, {
+  const getFeedItems = async (timeFrame = '', page = 1) => {
```
nickytonline, line 55: syntax (non-blocking): There is no need for these to be arrow functions, they can be regular functions.

app/javascript/utilities/hooks/useGlobalKeyEventListener.js
```diff
@@ -0,0 +1,35 @@
+import { useEffect } from 'preact/hooks';
+
+const FOCUSED_TAG_EXCLUDE_LIST = ['INPUT', 'TEXTAREA'];
+
+export const registerGlobalKeyEventListener = (keys, callback) => {
```
nickytonline, line 5: syntax (non-blocking): There is no need for these to be arrow functions (here and below), they can be regular functions.

API documentation (blocking): The function requires documentation as it is being exported. This provides documentation in the file directly, but the JSDoc can also provide Intellisense information to a developer, e.g. VS Code. For an example of JSDoc comments in action in the project, see https://github.com/forem/forem/blob/master/app/javascript/utilities/http/request.js

app/javascript/utilities/hooks/useGlobalKeyEventListener.js
```diff
@@ -0,0 +1,35 @@
...
+
+    callback(event);
+  };
+
+  window.addEventListener('keydown', eventListener);
+
+  return eventListener;
+};
+
+export default (keys, callback) => {
```
nickytonline, line 28: syntax (non-blocking): It's not documented in the project, but `default` exports are usually not preferred as it can prevent refactoring via an editor like VS Code. For more context, see the following posts.

* [Why I've stopped exporting defaults from my JavaScript modules](https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module)
* [Avoid Export Default](https://basarat.gitbook.io/typescript/main-1/defaultisbad)

app/views/stories/_main_stories_feed.html.erb
```diff
@@ -19,3 +19,83 @@
     <div class="single-article-small-pic" id="article-index-hidden-div" style="display:none"></div>
   <% end %>
 </div>
+
+<script defer>
```
nickytonline, line 23: syntax (blocking): Avoid inline JavaScript in ERB templates for a few reasons:

* One it is inline and won't be cached in the browser (well potentially, but as part of an HTML fragment in our server-side caching)
* This code can not be tested via unit/integration tests in Jest

Prefer to create a webpacker pack file, which can import most the required functions and run them. See the Forem [webpacker](https://docs.forem.com/frontend/webpacker/#webpacker) documentation as well as the webpacker [folder structure](https://github.com/rails/webpacker/blob/master/docs/folder-structure.md) documentation.

nickytonline (changes requested):
Thanks so much for the PR @reobin. Things are starting to look good. Just some comments and a change request.

app/views/stories/_main_stories_feed.html.erb
reobin: Wow! completely missed this.

Refactored to reuse the code of the list navigation list, but with different values, since the logged out feed is different. It felt so good to remove that specific unreusable code 👍

reobin:
I appreciate the feedback @nickytonline !

(A lot of) documentation was added. Feedback is more than welcome on it.

The navigation for the logged out feed has been moved to the webpack pack files, and it now uses the hook code.

reobin:
Note: Refactored (again 🌞) to use a jsx element to register the various event listeners.

It makes it easier and less error-prone to use the functionality in class-based components or erb templates.

reobin:
WIP while #10713 is settled

reobin:
@nickytonline #10713 has been merged, so this is ready for another review 🌞

What was done since the last review:
* Remove the global key event listener custom hook (did the same as the new keyboard shortcuts hook)
* Use the new keyboard shortcuts hook in the list navigation hook and the `/` and `0` shortcuts of the header
* Add tests covering all the cases from the article feed
* Add JSDoc

rhymes (commented):
Hi @reobin, it seems to be working nicely.

I'm unsure about the fact that visually we highlight the title of the article and not the entire box, I'd imagine the entire rectangle of the article item to become highlighted. What do you think @ludwiczakpawel ?

ps. there's a failing test here https://travis-ci.com/github/forem/forem/jobs/404108634#L1830

ludwiczakpawel:
How about adding following:

```
&:focus-within {
  --card-border: var(--accent-brand);
  box-shadow: 0 0 0 2px var(--card-border);
}
```

in [`/stories.scss` around line 16](https://github.com/forem/forem/blob/master/app/assets/stylesheets/components/stories.scss#L16)? This should add an outline around story card when something inside is being focused, like this:

image

Even though this extra outline around the title itself isn't ideal, it also doesn't bring any visual harm...

reobin:
That's a cool idea @ludwiczakpawel 

Would we want to explore the idea of the article being focusable itself? Like twitter does, for example.

Some details would have to be checked out for accessibility, but I think it would be doable. I'll gladly take the time to explore some implementations if we decide to go in this direction.

Example of a focused tweet using `j` and `k`:
Screen Shot 2020-10-23 at 07 51 17

How they chose to make a non-link being focusable:
Screen Shot 2020-10-23 at 07 51 35

reobin:
I tested it out and I have to say that this looks amazing @ludwiczakpawel :

demo

rhymes:
@reobin @ludwiczakpawel love it! :D

ludwiczakpawel (approved):

reobin:
With the new changes (see comments above), the whole works and look a lot better. I made sure it was usable with a screen reader as well.

With that said, adding a `role`, `onKeydown`, etc. to an `article` element violates accessibility rules, and make the `axe` tests fail.

Pros: functionality, looks
Cons: Breaks eslint and accessibility rules

That's option no.1, the one that is pushed to the branch right now.

**Option no.2**
I considered an alternative which is to focus on the article title like I did initially, and to have a focus style only on the card, and not the title.

Similar to [the card inclusive component demo](https://heydon.github.io/Inclusive-Components/cards-pseudo-content-author-link/), by [inclusive-components.design](https://inclusive-components.design/cards/). It works well, but the fact the title is not the first element in the card creates undesired behaviors.

One example of an undesired behavior is if you use the `tab` key inside a card, you focus first on the avatar, name, etc. Meanwhile, the focus style on the card is activated. Then, you focus on the title, and it doesn't have a focus style, so it looks like a bug.

Also, the focused image doesn't look great because the outlines overlap.

See the demo:

demo

Pros: does not break any eslint or accessibility rules
Cons: Visually less nice, confusion in the tab order

Did I leave out an option, here? Any ideas welcome

Link2Twenty:
If you want to replicate the keep to top functionality twitter has you could do something like

```js
window.scrollTo(0, ($0.offsetTop - 60))
```

where `$0` is the selected element and `64` is nav height (56) + shadow width (2) + extra margin (6)

app/javascript/articles/Article.jsx
```diff
@@ -50,6 +50,18 @@ export const Article = ({
       id={isFeatured ? 'featured-story-marker' : `article-${article.id}`}
       data-content-user-id={article.user_id}
       data-testid={isFeatured ? 'featured-article' : `article-${article.id}`}
+      role="link"
+      tabIndex="0"
+      aria-aria-labelledby={`article-link-${article.id}`}
```
Link2Twenty, line 55: This should just be 
```
aria-labelledby={`article-link-${article.id}`}
```

reobin: good catch!

reobin:
Works great @Link2Twenty ! 

demo

I'm a fan of going center as well:

demo

Feels a bit less crammed a the top

Link2Twenty:
If you wanna be really fancy you can do something like

```js
window.scrollTo({
  top: $0.offsetTop - 60,
  behavior: 'smooth',
});
```

Which is the same but it won't snap directly to the point. I guess that's more of a UX question though.

Link2Twenty:
I would say there is a case to be made for each article link not having an article tag.

On twitter they use an article tag because each tweet is a fully contained thing but on forem each item in the feed is just a link to an article.

> The HTML `<article>` element represents a self-contained composition in a document, page, application, or site, which is intended to be independently distributable or reusable (e.g., in syndication). Examples include: a forum post, a magazine or newspaper article, or a blog entry, a product card, a user-submitted comment, an interactive widget or gadget, or any other independent item of content.

___

That being said there is an alternative that is best of both worlds, maybe...

https://github.com/forem/forem/blob/737bb166ac0a99b6d80ab75cd5ff4fac8a7306c4/app/javascript/articles/Article.jsx#L46-L65

becomes

```jsx
<article
  className={`crayons-story cursor-pointer${
    isFeatured ? ' crayons-story--featured' : ''
  }`}
  id={isFeatured ? 'featured-story-marker' : `article-${article.id}`}
  data-content-user-id={article.user_id}
  data-testid={isFeatured ? 'featured-article' : `article-${article.id}`}
>
<a href={article.path} tabIndex="-1" aria-labelledby={`article-link-${article.id}`}></a>
```

and this

https://github.com/forem/forem/blob/61b4b00898fec109c68ed1903c0810ac5fa9555d/app/assets/stylesheets/components/stories.scss#L18-L22

becomes

```scss
&:focus {
  outline: none;
}

&:focus-within {
  --card-border: var(--accent-brand);
  box-shadow: 0 0 0 2px var(--card-border);
}
```

_Or we have an `onFocus` and `onBlur` on the anchor to set a class on the article_

Now we just set the focus to that anchor (which is not in the tab list) when we tab on it will go to the next item in the tab list so tabbing makes sense. Also as it's an anchor the `ctrl + return` will just work

reobin:
Thanks for taking the time to do this @Link2Twenty !

I think the option to explore here would be the second one because making the article an anchor would create the illegal "nested links" situation anyway.

I'll check it out today, I think you're on to something!

reobin:
The flow works really well with the hidden navigation link that I left at `tabindex="0"`.

It makes the whole card with `tab` focusable before entering it to focus on more specific elements. I think in terms of UX, that's the best-case scenario.

Thanks @Link2Twenty for this door into the best of both worlds 🎉

best of both worlds

app/assets/stylesheets/components/stories.scss
```diff
@@ -33,6 +40,16 @@
     }
   }
 
+  &__hidden-navigation-link {
+    position: absolute;
+    opacity: 0;
+    z-index: -1;
```
Link2Twenty, line 46: Personal preference I prefer

```css
pointer-events: none;
```

https://caniuse.com/pointer-events

reobin: Good point

Link2Twenty:
If you wanted the experience to be more like twitter (again not sure you do) you could do something like this, meaning the outer border goes once you have tabbed within the element.

```scss
&__hidden-navigation-link {
  pointer-events: none;
  position: absolute;
  border-radius: var(--radius);
  color: rgba(0,0,0,0);
  top: 0;
  right: 0;
  bottom: 0;
  left: 0;

  &:focus {
    outline: none;
    --card-border: var(--accent-brand);
    box-shadow: 0 0 0 2px var(--card-border);
  }
}
```

reobin:
I actually like it this way, but both are nice, honestly.

@ludwiczakpawel Do you want to chip in on this? Should we keep the focus border once the user focuses into the card (author, comments, etc.),

ludwiczakpawel:
@reobin I'd definitely keep the border :)

reobin:
@nickytonline waiting on your go :)

Excited to disable vimium on dev.to!

rhymes (approved):
Forgot to approve ahhaha

Link2Twenty:
@nickytonline any update on this? I'm really looking forward to seeing it in action 😁

nickytonline:
> @nickytonline any update on this? I'm really looking forward to seeing it in action 😁

Looking at it now. I got your other one reviewed earlier today. 😄

app/javascript/utilities/viewport.js
```diff
@@ -0,0 +1,24 @@
+/**
```
nickytonline, line 1: Refactor (non-blocking): Not in this PR, but in a future PR, this can be replaced with the [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) API it's corresponding [polyfill](https://github.com/w3c/IntersectionObserver/tree/master/polyfill) for Firefox on Android.

nickytonline:
@reobin, I've run this locally and updated your branch with the latest from the main branch. Navigating via `j` or `k` works, but it is slow. I can press `j` or `k`, but then I need to wait about a second and press again before it works again. I don't know if you've exhibited this as well.

reobin:
@nickytonline as you can see in the gifs I posted above, it is pretty snappy for me. 🤔

I'll pull the changes and check it out

nickytonline:
I haven't checked it out, but I believe there is a 500ms timeout by default for the changes @Link2Twenty put in place. Maybe it's that?

reobin:
You are right @nickytonline, `useKeyboardShortcuts` has changed. I'll update my use of it with the new params

reobin:
I'm not sure having a 500ms delay by default is the right call, though.

It means that the other shortcuts we implemented in #10878 and #11116  also now have that delay (unless they were updated as well, not sure)

Link2Twenty:
The time out should only be for chained keys not single presses (I'll have a check in the morning to see if there's something I can do)

nickytonline:
> I'm not sure having a 500ms delay by default is the right call, though.
> 
> It means that the other shortcuts we implemented in #10878 and #11116 also now have that delay (unless they were updated as well, not sure)

Agree, I think this just wasn't caught. That's my bad. Is that the issue though? It was just a guess on my part.

nickytonline:
@reobin, @Link2Twenty, yeah, it's the default timeout of 500ms that's the issue. I'll put up a PR for that. I think we can probably move the timeout to just combined keystrokes. I'll look into that.

reobin:
@nickytonline Yeah, I added the params locally for a timeout of zero, and it works fine.

Didn't realize it was a mistake, I'll wait for your fix 👍

nickytonline:
You can set it to 0 ms in your branch though, because it will just get merged without that change as it will be in master already.

reobin:
Done. Also changed the names to `.js` instead of `.jsx`, based on one of your comments in #11113

app/assets/javascripts/utilities/buildArticleHTML.js
```diff
@@ -267,13 +267,27 @@ function buildArticleHTML(article) {
         '</div></a>';
     }
 
-    return (
-      '<article class="crayons-story" data-article-path="' +
+    var navigationLink = `
```
nickytonline, line 270: refactor (non-blocking): You can use [template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

nickytonline (approved):
Thanks so much for the PR @reobin! 🚀

Link2Twenty:
@reobin if you want, doing a post on dev, tagged as meta, about this new feature might be a good idea 😊

Issue #8457, opened 2020-06-12
RSS/HTML import doesn't handle newlines


binford2k:
**Describe the bug**

HTML does not distinguish whitespace. This means that text like so

```
This is a line
with some
newline characters.
```

and this

`This is a line with some newline characters.`

should be treated exactly the same. Unfortunately, the RSS importer just strips the newline character, so the resulting string becomes this:

`This is a linewith somenewline characters.`

This makes syndicated content a GIANT pain to post.

In my case, I publish via Jekyll and GitHub pages and I also wrap my markdown source. See an example at https://raw.githubusercontent.com/binford2k/binford2k.com/master/_posts/2020-06-10-writing-a-new-metrics-plugin.md

Jekyll publishes the text with embedded newlines, which is rendered properly by every browser that correctly implements the HTML spec, including the browser in which you're reading this issue.

``` html
<p>I’ve been working on the Dropsonde telemetry framework for the Puppet ecosystem
for a while. If you’ve followed any of its development, you likely already know
that the main focus is on providing community value and maintaining privacy and
transparency.</p>
``` 

ends up imported as

> I’ve been working on the Dropsonde telemetry framework for the Puppet ecosystemfor a while. If you’ve followed any of its development, you likely already knowthat the main focus is on providing community value and maintaining privacy andtransparency.

**To Reproduce**

1. Publish content with newline wrapped text.
1. RSS import it.

**Expected behavior**

Newline characters should be treated like whitespace, following the HTML spec.

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

rhymes:
@binford2k would the following comment - https://github.com/thepracticaldev/dev.to/issues/7923#issuecomment-630399659 - help?

AFAIK we don't support markdown in the RSS feed, the content should be HTML

citizen428:
@rhymes The feed doesn't seem to include Markdown, but HTML:

http://binford2k.com/atom.xml

binford2k:
[...]

PR #10476, opened 2020-09-29
Fix newlines being chomped in RSS import

hugopeixoto:
The CustomText converter was monkey patched to fix a bug related to
preserve_tags. Somewhere in that process, remove_inner_newlines was
changed to strip newlines completely, instead of replacing them with a
single space. This restores that functionality.

In development mode, files are not eager loaded. That means that the
converters were not being required, so it was not possible to reproduce
that issue in development mode (since it was falling back to the
original converter, which does not have the bug). This commit also adds
an initializer that requires every converter.

This fixes half of #8457.

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/lib/reverse_markdown/converters/custom_text.rb
+++ b/app/lib/reverse_markdown/converters/custom_text.rb
@@ -40,7 +40,7 @@ module ReverseMarkdown
       end
 
       def remove_inner_newlines(text)
-        text.tr("\n\t", "")
+        text.tr("\r\n\t", " ").squeeze(" ")
       end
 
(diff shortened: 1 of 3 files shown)


maestromac (approved):
Thank you for this PR. LGTM!

Zhao-Andy (approved):

Issue #10096, opened 2020-08-30
Link no longer available


omenking:
The link does exist
<img width="692" alt="Screen Shot 2020-08-30 at 4 03 53 PM" src="https://user-images.githubusercontent.com/7776/91668405-81d1c580-eada-11ea-8be7-297308c1f755.png">

But it displays this
<img width="773" alt="Screen Shot 2020-08-30 at 4 03 48 PM" src="https://user-images.githubusercontent.com/7776/91668403-7f6f6b80-eada-11ea-8d6c-c3da6d5c365d.png">

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

- OS, version:
- Browser, version:

**Smartphone (please complete the following information):**

- Device:
- OS, version:
- Browser, version:

**Additional context**

rhymes:
Hi @omenking, this is definitely a bug but it's also a communication failure. The `link` liquid tag only works for `dev.to` articles

Rafi993:
Hello there, If this still up for taking. I can help with this bug.

rhymes:
@Rafi993 sure, go ahead! thanks

Rafi993:
> @Rafi993 sure, go ahead! thanks

Thank you @rhymes

Rafi993:
When you use url   
```
{% link https://www.freecodecamp.org/news/ %}
```
it displays the correct error message

![link_error](https://user-images.githubusercontent.com/8092120/94501112-aaa2c300-021e-11eb-9b8f-290b8c70654d.png)

but when I add the url with another '/' like

```
{% link https://www.freecodecamp.org/news/ultimate-dynamodb-2020-cheatsheet/ %}
```
it causes the problem
![link_error](https://user-images.githubusercontent.com/8092120/94501166-d3c35380-021e-11eb-95c2-77d71a219351.png)

It is probably issue in 

https://github.com/forem/forem/blob/1b8a497054cd03e60fafecad7c10a2b5d8c22816/app/liquid_tags/link_tag.rb#L23-L31

PR #10437, opened 2020-09-25
Handling empty gist link error in liquid tag

Rafi993:
## What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Better error message when user does not provide link to the Gist in liquid tag

## Related Tickets & Documents

closes https://github.com/forem/forem/issues/10410

## QA Instructions, Screenshots, Recordings

empty_gist_link

## Added tests?

- [x] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

PR #10479, opened 2020-09-29
Handling links from different domain

Rafi993:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Shows error message when user uses links from different domain in **link** liquid tag

closes https://github.com/forem/forem/issues/10096

1. Go to comment section in article
2. Paste the comment 
```
{% link https://www.freecodecamp.org/news/ultimate-dynamodb-2020-cheatsheet/ %}
```
3. Click preview it should that link is not valid article link

link_error

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/liquid_tags/link_tag.rb
+++ b/app/liquid_tags/link_tag.rb
@@ -21,7 +21,15 @@ class LinkTag < LiquidTagBase
   end
 
   def article_hash(slug)
-    path = Addressable::URI.parse(slug).path
+    url = Addressable::URI.parse(slug)
+    domain = url.port ? "#{url.host}:#{url.port}" : url.host
+    path = url.path
+
+    # If domain is present in url check if it belongs to the app
+    unless domain.blank? || domain&.casecmp?(SiteConfig.app_domain)
+      raise StandardError, "This URL is not an article link: {% link #{slug} %}"
+    end
+
     path.slice!(0) if path.starts_with?("/") # remove leading slash if present
     path.slice!(-1) if path.ends_with?("/") # remove trailing slash if present
     extracted_hash = Addressable::Template.new("{username}/{slug}").extract(path)&.symbolize_keys
--- a/app/liquid_tags/link_tag.rb
+++ b/app/liquid_tags/link_tag.rb
@@ -27,13 +27,13 @@ class LinkTag < LiquidTagBase
 
     # If domain is present in url check if it belongs to the app
     unless domain.blank? || domain&.casecmp?(SiteConfig.app_domain)
-      raise StandardError, "This URL is not an article link: {% link #{slug} %}"
+      raise StandardError, "This link does not point to an article: {% link #{slug} %}"
     end
 
     path.slice!(0) if path.starts_with?("/") # remove leading slash if present
     path.slice!(-1) if path.ends_with?("/") # remove trailing slash if present
     extracted_hash = Addressable::Template.new("{username}/{slug}").extract(path)&.symbolize_keys
-    raise StandardError, "This URL is not an article link: {% link #{slug} %}" unless extracted_hash
+    raise StandardError, "This link does not point to an article: {% link #{slug} %}" unless extracted_hash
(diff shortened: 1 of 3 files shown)


spec/liquid_tags/link_tag_spec.rb
```diff
@@ -19,12 +19,12 @@
     create(:article, user_id: user.id, title: "Hello & Hi <3 <script>", tags: "tag")
   end
 
-  def generate_new_liquid(slug)
+  def generate_new_liquid(slug:)
```
Rafi993, line 22: I converted it to symbol to make it similar to change done in https://github.com/forem/forem/pull/10437#discussion_r495221635

app/liquid_tags/link_tag.rb
```diff
@@ -21,7 +21,15 @@ def get_article(slug)
   end
 
   def article_hash(slug)
-    path = Addressable::URI.parse(slug).path
+    url = Addressable::URI.parse(slug)
+    domain = url.port ? "#{url.host}:#{url.port}" : url.host
+    path = url.path
+
+    # If domain is present in url check if it belongs to the app
```
Rafi993, line 28: This comment might be redundant but I added it anyway,

app/liquid_tags/link_tag.rb
```diff
@@ -21,7 +21,15 @@ def get_article(slug)
...
+    unless domain.blank? || domain&.casecmp?(SiteConfig.app_domain)
+      raise StandardError, "This URL is not an article link: {% link #{slug} %}"
```
Rafi993, line 30: I have a feeling there is a better way to deal with this issue any thoughts?

Link2Twenty:
Maybe a question for a different PR but why are some errors like this

Gist error

and others like this

new link error

It might be confusing for users to have to guess where the error will appear.

app/liquid_tags/link_tag.rb
rhymes: slug might not be a URL right? Then the message is slightly misleading

rhymes (commented):
Looks fine to me, I'd only update the messaging to be a little more generic so to encompass both slug and URL

app/liquid_tags/link_tag.rb
Rafi993: when you use
```
{% link https://www.freecodecamp.org/news/ultimate-dynamodb-2020-cheatsheet/ %}
```
variable slug has value https://www.freecodecamp.org/news/ultimate-dynamodb-2020-cheatsheet/

and when you do
```
{% link abbottkarey/surprised-by-joy-et-assumenda-jgn %}
```
slug has the value **abbottkarey/surprised-by-joy-et-assumenda-jgn**

I think the variable name slug might be the cause of confusion. I was not sure what to rename it to so I just left it same.

Rafi993: Probably I should rename slug to **url** or **url_string** 
(since there is ```    url = Addressable::URI.parse(slug)``` on the next line) ? any thoughts?

rhymes: I'm not talking about the variable though, I meant the wording of the error message ehhe

`/abbottkarey/surprised-by-joy-et-assumenda-jgn/` is not a URL so telling the user "the URL is not an article link" is misleading. Also because it's just a slug, not a link either.

Maybe something like "The article you're looking for does not exist" ?

Rafi993: ok. I'll change it.

Rafi993:
There seems to be some test failing let me fix it

rhymes (approved):
LGTM! :)

benhalpern (approved):

PR #10485, opened 2020-09-29
Redis namespace (for Redis multitenancy)

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This pull request adds the ability to namespace redis keys, which is a native option built into all of our Redis configuration.

This means, hypothetically, the system could allow many Forem instances to use the same Redis keys, possibly simplifying orchestration in some scenarios and/or cost savings and performance improvements.

As far as I can tell, functionality when namespaces exist will automatically account for them in almost all cases. There are some commands which do not apply 1:1 described in the [redis-namespaced gem](https://github.com/resque/redis-namespace) but I don't think any of this will be an issue for us, as users of our multitenant environment cannot directly run commands, we'll always be able to account for these things.

The Redis data model is simple such that multi-tenancy here seems more straightforward than Postgresql where there would be _much more_ to consider in terms of pros and cons. This functionality allows us to take the most effective approach for the big picture of Redis cloud hosting.

**🚨 `SKIP_REDIS_NAMESPACE` should be set to `true` for DEV so that namespaces are not applied because we do not want to invalidate all keys and cause deployment issues. All other Forems should probably be fine with at current traffic levels.**


mstruve (commented):
Pulled this down to test and the app crashes looks like we need the namespace gem to make it work
```
ERROR: Your Redis configuration uses the namespace 'dev_local' but the redis-namespace gem is not included in the Gemfile.Add the gem to your Gemfile to continue using a namespace. Otherwise, remove the namespace parameter.
```

benhalpern:
@mstruve I added the gem and tested a little locally as far as I could tell, but I don't know if I fully understand how to reproduce that. Let me know if things look good on your machine or if we could add some tests.

Gemfile
```diff
@@ -81,6 +81,7 @@ gem "ransack", "~> 2.3" # Searching and sorting
 gem "recaptcha", "~> 5.5", require: "recaptcha/rails" # Helpers for the reCAPTCHA API
 gem "redcarpet", "~> 3.5" # A fast, safe and extensible Markdown to (X)HTML parser
 gem "redis", "~> 4.2.2" # Redis ruby client
+gem "redis-namespace", "~> 1.8" # Allows for namespacing in Redis
```
mstruve, line 84: The [blind pass through](https://github.com/resque/redis-namespace#blind-passthrough) issue worries me a little bit that it might cause some collisions between apps.

config/environments/production.rb
```diff
@@ -75,7 +80,7 @@
   redis_url = ENV["REDISCLOUD_URL"]
   redis_url ||= ENV["REDIS_URL"]
   default_expiration = 24.hours.to_i
-  config.cache_store = :redis_cache_store, { url: redis_url, expires_in: default_expiration }
+  config.cache_store = :redis_cache_store, { namespace: ENV["REDIS_NAMESPACE"], url: redis_url, expires_in: default_expiration }
```
mstruve, line 83: For testing purposes, I think we need to have this also set in our development environment for our Redis setup to keep it as consistent as possible.

config/initializers/sidekiq.rb
```diff
@@ -38,7 +38,7 @@ def not_past_scheduled_time?(current_time)
   sidekiq_url = ApplicationConfig["REDIS_SIDEKIQ_URL"] || ApplicationConfig["REDIS_URL"]
   # On Heroku this configuration is overridden and Sidekiq will point at the redis
   # instance given by the ENV variable REDIS_PROVIDER
-  config.redis = { url: sidekiq_url }
+  config.redis = { url: sidekiq_url, namespace: ENV["REDIS_NAMESPACE"] }
```
mstruve, line 41: I am seeing the namespace on most of our Rails traffic but I am not seeing it for some of our Sidekiq keys and in consequence Sidekiq is not working. Below are a couple of examples. 
```
 "dev_local:views/articles/_full_comment_area:de39d75373c05859b3e87229f9867dcf/whole-comment-area-8-2020-09-18 00:09:23 UTC-true",
 "cron_job:send_email_digest:enqueued",
 "cron_job:sitemap_refresh:enqueued",
 "dev_local:stat:failed",
 "queues",
 "cron_job:award_sixteen_week_streak_badge:enqueued",
 "uniquejobs:bfd4f2184500998cceb6328c5dbe617a:GRABBED",
 "cron_jobs",
 "uniquejobs:cdef8e603a2ce66c9c194fefa72687e5:EXISTS",
```

config/initializers/session_store.rb
```diff
@@ -20,4 +20,5 @@
                                        signed: true,
                                        secure: ApplicationConfig["FORCE_SSL_IN_RAILS"] == "true",
                                        same_site: :lax,
-                                       httponly: true
+                                       httponly: true,
+                                       namespace: ENV["REDIS_NAMESPACE"]
```
mstruve, line 24: I'm also not sure if this is working since I can see my session Redis key locally and it does not have a namespace.

mstruve (changes requested):
Does not appear to be working with Sidekiq or sessions locally from my testing. We should definitely get the prod setup into the development.rb file so it's easy to replicate locally.

mstruve:
Was just catching up on the forem.team thread for this, what is the reason we decided to go with namespacing rather than using Redis databases per @citizen428's suggestion? I agree that I think databases(which I have also used in the past successfully) would be a tad more robust than namespacing.

Gemfile
citizen428: I've used this gem in production before, and I'm not a big fan. We had lots of problems with things that aren't namespace aware, e.g. this as mentioned in the documentation:

> The effects of some redis commands cannot be limited to a particular namespace (e.g., `FLUSHALL`, which literally truncates all databases in your redis server, regardless of keyspace). 

This e.g. caused us to lose all background jobs once when we wanted to evict all keys from a specific cache. So the way to do this correctly is first looking for the correct keys yourself (with something like `KEYS  <namespace>-*`, then iterating over them and deleting them. This is an O(n) operation and was very slow for our use case (a heavily used cache with periodic full eviction).

citizen428:
> which is a native option built into all of our Redis configuration

I don't necessarily agree with this assessment. It's a quite clever hack that sits on top of Redis and gives you the illusion of namespaces, but they don't exist as an actual concept inside the data store.

Gemfile
rhymes: > The blind pass through issue worries me a little bit that it might cause some collisions between apps.

As Redis keeps improving each version, this might be a problem which would lock us on a specific version of Redis. Now we're only constrained by which versions the cloud providers support

I'm also quite frightened by:

> Redis::Namespace provides a namespaced interface to Redis by keeping an internal registry of the method signatures in Redis provided by the redis-rb gem; we keep track of which arguments need the namespace added, and which return values need the namespace removed.

It honestly feels like a tool useful to people that have small apps and want to separate different parts of it inside Redis without namespacing it themselves

config/initializers/session_store.rb
rhymes: Not sure why but I also noticed this https://github.com/redis-store/redis-store/issues/339 - not all commands are namespaces, same issue redis-namespace has

config/initializers/sidekiq.rb
rhymes: This is probably because we'd have to configure both the client and the server connections, here we're only configuring the server, not the client https://dev.to/rbglod/redis-namespace-when-one-sidekiq-process-is-not-enough-39am

rhymes (commented):
I second @mstruve, @citizen428 and @jdoss sentiment of skepticism about using this gem to segment different apps on the same Redis instance. I think we should explore something like multiple DBs while keeping in mind that we might eventually want to configure a transient caching Redis differently than a persistent storage Redis or a Redis with sessions or queues. 

A possibility could be to have, for example, different types of Redis servers:

1 - redis-cache-1 which contains 16 (I think the max is 16 Redis DBs) caches of 16 forems
2 - redis-queue-1 which contains 16 sidekiq queues of 16 forems 
3 - redis-sessions-1 ...

All of these can be configured indipendently with different eviction policies and storage allocation policies (I'd imagine redis cache being bigger than redis sessions, so that divided by 16 similar Forems would allow us to understand how much to allocate per each copy of "Redis cache N").

This way we don't namespace, we just point each app to redis-cache-1, redis-sessions-1 and redis-queue-1 each with a different DB number, when those are filled, we start a a new set of redis servers redis-cache-2 and so on and keep going.

Thoughts?

Issue #10061, opened 2020-08-28
Bug: Comments that Point to Another Comment via a Liquid Tag Are Invalid if Liquid Tag referenced Comment is Deleted 


mstruve:
**Describe the bug**

If you have a comment that references another comment and that comment is deleted, the original comment becomes invalid. For example:
Comment A Markdown: `Checkout {% devcomment 1b %}`
If Comment with id_code `1b` is deleted, Comment A becomes invalid and can no longer be saved. This prevents others from reacting to it amongst other things. 

**To Reproduce**

1. Create a comment with a comment liquid tag that points to a second comment. 
2. Delete the comment referenced by the liquid tag.
3. Attempt to save the original comment with the liquid tag. You will get the following error. `StandardError, "Invalid comment ID or comment does not exist"`

**Expected behavior**

Comments that reference other comments should remain valid even if the other comment is deleted. I would suspect that instead of displaying the comment for the liquid tag it would display a message along the lines of "Comment has been removed"

msarit:
I'll grab this! 😄

PR #10486, opened 2020-09-29
Fix bug that arises when  a previously tagged comment is deleted

msarit:

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
See [Issue #10061](https://github.com/forem/forem/issues/10061). If a user edits their comment that contained a Comment Tag to a comment that no longer exists, upon saving their edits, the save-action fails silently; navigating to the user's original comment also 404's.
This bug fix returns the message `"Comment Not Found"` if a previously-tagged comment no longer exists.

**Additional Considerations:**
We may want to fix the silent failure of the save-action, perhaps display some sort of error message. This is beyond the scope of this PR. 

1. Fire up your dev environment and sign in using two different profiles (use incognito windows).
2. As Profile1, create a Comment (1) and note its permalink id_code
3. As Profile2, create a Comment (2) and Tag Comment1 in it
4. Delete Comment1
5. Edit Comment2 and save. You should now see the message "Comment Not Found" inside Comment1's liquid tag

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

No

Bug Spray


app/liquid_tags/comment_tag.rb
```diff
@@ -14,9 +14,7 @@ def render(_context)
   end
 
   def find_comment(id_code)
-    Comment.find(id_code.to_i(26))
-  rescue ActiveRecord::RecordNotFound
-    raise StandardError, "Invalid comment ID or comment does not exist"
+    Comment.find_by(id_code: id_code)
```
msarit, line 17: I have a question about how I am finding the Comment here.

Originally, the code was
```ruby
Comment.find(id_code.to_i(26))
```

Please advise me if the conversion of the `id_code` to Integer Base 26 is essential. Cos that would mean my approach above is inappropriate 🤷‍♀️

rhymes: I think it's fine, as `id_code` is the serialization of `id.to_i(26)` as indicated here 

https://github.com/forem/forem/blob/53deac75cde12bca16f680369424dfd187705f85/app/models/comment.rb#L90-L94

If the liquid tag tests are still working then it's working ;)

rhymes (approved):
Great job @msarit!

benhalpern (approved):

PR #10487, opened 2020-09-29
fix: Clear chat input after Enter key submit

reobin:
- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The chat input value was handled and reset outside of the component. While that is not necessarily a problem, the auto resized Textarea component acted weird when the value was not "controlled" by the `Textarea` component.

The solution was indeed to hold a state inside the compose component, but also to make sure the only place where the input of the textarea is controlled was inside the component itself, and nowhere else. 

Also in the PR:
* Compose becomes a functional component, which made it easier to manage this particular state and props flow
* Use 1 textarea instance for both add and edit mode

## Related Tickets & Documents

Closes #10480 

## QA Instructions, Screenshots, Recordings

* Open chat
* Make sure the input is cleared when sent with the Enter key
* Make sure the input is cleared when sent with the send button
* Make sure the textarea is resized when the input overflows

All of the above should be true in edit mode as well.

## Added tests?

- [x] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

The tests check for an empty textarea field after the following events:

* Pressing Enter
* Clicking Send
* Clicking Save edit
* Clicking Close edit

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

PR #10493, opened 2020-09-30
fix mention in chat

narender2031:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
- In chat, Mention POP not display and through the error. 
Steps to reproduce: -
- go dev.to
- click on chat icon
- type `@` in default;t open chat. It will not work. 

Screenshot 2020-09-30 at 12 28 35 PM

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

alt_text


reobin:
This won't be compatible with #10487 because of the change on the `compose.jsx` file, which I'm not sure is necessary for this PR (might be wrong).

The `value` state is only used by the Textarea component, so removing its value field makes holding the state useless in this situation. That said, the textarea resize doesn't always work properly, and controlling its input with the state (or at least having a `setState` after each input, which forces a rerender) will make sure it does, so I get why the state is left there.

I refactored the Compose component's input state management in #10487 which fixes a bug when using the Enter key to send the message. Both PRs are compatible and complete each other, but the first one merged will leave a conflict for the other before it's merged.

narender2031:
In that case, I will remove `compose.js` changes. This PR fixes the mention user issues. There is one more that is when user clicks on the mentioned username. Text input does not update, That  I fixed by removing value from `compose.js`. I think your will solve this specific issue. So will remove undo my changes for `compose.js`.

narender2031:
@citizen428 please restart the build

ludwiczakpawel:
@narender2031 I've restarted the job for you.

narender2031:
any updates for this PR @nickytonline

citizen428 (approved):

ludwiczakpawel:
@narender2031 Do you have any tips on how to test it locally? Since this requires interaction between two accounts to be tested AND you can really have only one account locally, it's a bit black magic for me :D And even internally we don't have a good and established patterns to handle cases like that locally. So I'm wondering if maybe you had some good idea around this?

citizen428:
> AND you can really have only one account locally (because of social auth, unless you have multiple github or twitter account),

Don't forget email login.

ludwiczakpawel:
Oh right, we do have it now! Good point!

app/javascript/chat/chat.jsx
```diff
@@ -253,7 +253,7 @@ export default class Chat extends Component {
           activeChannel ||
           this.filterForActiveChannel(channels, activeChannelId),
       });
-      this.setupChannel(activeChannelId);
+      this.setupChannel(activeChannelId, activeChannel);
```
nickytonline, line 256: @narender2031, are there tests to test this change or does it require some new tests?

ludwiczakpawel (changes requested):
Hmm.. it kinda works and kinda doesn't work.. so the popups shows up but isn't very useful since I can't really hit [tab] or [return] to autofill the username.. It's just an autocomplete popup but it doesn't really autocomplete anything :)  https://d.pr/v/aRAtpv

I also occasionally see some console error as well: https://d.pr/v/baSYi1

narender2031:
> Hmm.. it kinda works and kinda doesn't work.. so the popups shows up but isn't very useful since I can't really hit [tab] or [return] to autofill the username.. It's just an autocomplete popup but it doesn't really autocomplete anything :) https://d.pr/v/aRAtpv
> 
> I also occasionally see some console error as well: https://d.pr/v/baSYi1

I will check

citizen428:
@narender2031 This seems to work when I click on a user name in the popup, but not when I select it with the cursor keys and press enter. I think we should support both, what do you say?

narender2031:
> @narender2031 This seems to work when I click on a user name in the popup, but not when I select it with the cursor keys and press enter. I think we should support both, what do you say?

Yes I think we need to add support for those.

narender2031:
@citizen428 There is some issue to add enter support. Because there is a focus on a text field and on enter, it works differently. If you have any suggestions please add suggestions.
Either, we can add this later. At the moment we can go with this.

citizen428:
@narender2031 Agreed, we can add confirmation via Enter later.

ludwiczakpawel (approved):
Seems to be working now, cheers!

app/javascript/chat/chat.jsx
nickytonline: Just following up on this @narender2031.

citizen428 (changes requested):
@narender2031 @sarthology I think this is still missing tests as requested by @nickytonline, so I'll request changes to this PR now so we don't accidentally merge before they were added.

app/javascript/chat/chat.jsx
narender2031: @nickytonline  There are no existing test cases for mention. we need to add new test cases for the mentions.

narender2031:
> @narender2031 Do you have any tips on how to test it locally? Since this requires interaction between two accounts to be tested AND you can really have only one account locally (because of social auth, unless you have multiple github or twitter account), it's a bit black magic for me :D And even internally we don't have a good and established patterns to handle cases like that locally. So I'm wondering if maybe you had some good idea around this?

I have an two github account. One is my personal and another is company provided.

app/javascript/chat/chat.jsx
nickytonline: Just following up on this.

nickytonline:
@narender2031, it looks like there is a merge issue as there are 1171 file changes.

narender2031:
> @narender2031, it looks like there is a merge issue as there are 1171 file changes.

Yes. I am trying to revert these changes and commit to old state.

app/javascript/chat/chat.jsx
nickytonline: > @nickytonline  @citizen428 we require some test cases for this change. Currently, we don't have any test cases for this.  

I see the bad merge has been fixed @sarthology. Thanks. Just following up on the missing tests.

citizen428:
The giant diff is gone now, thanks @sarthology. We still need specs for this branch though.

narender2031:
By removing the merge issue we lost some file changes and that’s by part of issue still there due to lose of changes. I am fixing this PR again.

app/javascript/chat/chat.jsx
```diff
@@ -1634,6 +1646,10 @@ export default class Chat extends Component {
     const { activeChannel } = this.state;
     const mention = e.keyCode === 64;
     if (mention && activeChannel.channel_type !== 'direct') {
+      const memberListElemt = document.getElementById('mentionList');
+      memberListElemt.focus();
+      // console.log('Hello')
+      // document.getElementById('messageform').blur();
```
nickytonline, line 1652: Refactor (blocking): Remove commented code.

app/javascript/chat/chat.jsx
```diff
@@ -1634,6 +1646,10 @@ export default class Chat extends Component {
     const { activeChannel } = this.state;
     const mention = e.keyCode === 64;
     if (mention && activeChannel.channel_type !== 'direct') {
+      const memberListElemt = document.getElementById('mentionList');
```
nickytonline, line 1649: nitpick: Name this `memberList` or `memberListElement`. There is no need for the abbreviated variable name as it get minified when bundled.

citizen428 (approved):

app/javascript/chat/chat.jsx
```diff
@@ -353,16 +355,20 @@ export default class Chat extends Component {
...
+    const { messages, messageOffset, appDomain } = this.state;
     if (
       !messages[channelId] ||
       messages[channelId].length === 0 ||
       messages[channelId][0].reception_method === 'pushed'
     ) {
       getAllMessages(channelId, messageOffset, this.receiveAllMessages);
     }
-    if (activeChannel && activeChannel.channel_type !== 'direct') {
+    if (
```
nickytonline, line 367: Question: It's not clear to me what this logic is. Is it checking that it's not a DM and that the current channel is indeed the current channel?

app/javascript/chat/__tests__/chat.test.jsx
```diff
@@ -226,4 +226,19 @@ describe('<Chat />', () => {
       }),
     ).toBeDefined();
   });
+
+  it('should show mention pop-up on keyUp @', async () => {
```
nickytonline, line 230: Missing tests (blocking): This is great that it tests that the mention list opens when you type the `@` character, but it would also be good to have tests for a couple of other things:

-  test that the current user does not appear in the list.
- test that the list filters as the user types. For example, if your list contains nickytonline and sarthology, if I type `@n`, only `nickytonline` should be in the the list.

app/javascript/chat/chat.jsx
```diff
@@ -1745,13 +1762,16 @@ export default class Chat extends Component {
       channelUsers,
       memberFilterQuery,
     } = this.state;
+
     const filterRegx = new RegExp(memberFilterQuery, 'gi');
+
     return (
       <div
```
nickytonline, line 1769: Semantically, this would make more sense as an unordered list, i.e. `<ul><li>...</li>...</ul>` and that real buttons are used instead of a `<div />` with a `role` attribute with the value `"button"`, but I will create a separate issue for that as it is not related to the bug fix. It only caught my attention because of the newly added `data-testid` attribute.

nickytonline (changes requested):
We're getting there @narender2031. Some more request changes.

app/javascript/chat/chat.jsx
narender2031: It is checking the active channel is not DM.

nickytonline (approved):
🚀

narender2031:
@nickytonline Few Tests are failing. Those are not related to PR.

citizen428:
@narender2031 I reran the build for you until it passed.

PR #10495, opened 2020-09-30
Add missing presence validators to models

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

**Disclaimer:** the PR is a little bigger than usual but it's only presence validators.

Basically we're missing a ton of `validates :field, presence: true` as all fields that in the DB are declared as `null: false` should have a presence validator.

Ran `bundle exec rake active_record_doctor:missing_presence_validation` to intercept them, see https://github.com/gregnavis/active_record_doctor#detecting-missing-presence-validations

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -9,7 +9,9 @@ class Tag < ActsAsTaggableOn::Tag
   # This model doesn't inherit from ApplicationRecord so this has to be included
   include Purgeable
   include Searchable
+
   ALLOWED_CATEGORIES = %w[uncategorized language library tool site_mechanic location subcommunity].freeze
+  HEX_COLOR_REGEXP = /\A#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})\z/.freeze
 
   belongs_to :badge, optional: true
   belongs_to :mod_chat_channel, class_name: "ChatChannel", optional: true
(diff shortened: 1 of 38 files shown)


app/mailers/verification_mailer.rb
```diff
@@ -5,7 +5,7 @@ class VerificationMailer < ApplicationMailer
 
   def account_ownership_verification_email
     @user = User.find(params[:user_id])
-    email_authorization = EmailAuthorization.create(user: @user, type_of: "account_ownership")
+    email_authorization = EmailAuthorization.create!(user: @user, type_of: "account_ownership")
```
rhymes, line 8: Caught this little bug in the process, as the mailer should fail if something goes wrong here

app/models/tag.rb
```diff
@@ -9,7 +9,9 @@ class Tag < ActsAsTaggableOn::Tag
   # This model doesn't inherit from ApplicationRecord so this has to be included
   include Purgeable
   include Searchable
+
   ALLOWED_CATEGORIES = %w[uncategorized language library tool site_mechanic location subcommunity].freeze
+  HEX_COLOR_REGEXP = /\A#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})\z/.freeze
```
rhymes, line 14: grouped the regexp here

app/models/user.rb
```diff
@@ -145,23 +145,43 @@ def sync_profile
   devise :invitable, :omniauthable, :registerable, :database_authenticatable, :confirmable, :rememberable,
          :recoverable, :lockable
 
+  validates :articles_count, presence: true
```
rhymes, line 148: there's a bunch of sorting lines going on here, in addition to the new validators

mstruve (approved):
NICE cleanup!!!!

Zhao-Andy (approved):

Issue #10338, opened 2020-09-15
Remove redirects in update actions


atsmith813:
## What is the problem?
Currently, we have a race condition any time we're using a redirect after an action that updates data causing the cache to be busted and we're directing to the page whose cache was just busted. We usually lose the race - we redirect before the cache can be successfully busted. This results in users seeing stale content.

The best example is updating a comment. Once a comment is updated, we bust the cache for the comment and redirect back to the comments Index action. The redirect happens before the busting of the cache causing the user to not see their new update unless they then refresh the page again manually.

There are a handful of places where this race condition poses an issue. It's mostly in update actions where we update cached data and then redirect to the index view where we're displaying said data right after.
- [x] [Updating a comment](https://github.com/forem/forem/blob/6c3f5d51fe20413682feea411bc40e522c97d90b/app/controllers/comments_controller.rb#L142-L144)
- [x] ~[Tag updates (only admins and tag mods can do this)](https://github.com/forem/forem/blob/6c3f5d51fe20413682feea411bc40e522c97d90b/app/controllers/tags_controller.rb#L22-L24)~
- [ ] [Updating articles under certain conditions](https://github.com/forem/forem/blob/6c3f5d51fe20413682feea411bc40e522c97d90b/app/controllers/articles_controller.rb#L151-L158)
- [ ] [Updating listings](https://github.com/forem/forem/blob/6c3f5d51fe20413682feea411bc40e522c97d90b/app/controllers/concerns/listings_toolkit.rb#L148-L149)
- [ ] Document this quirk

## Solution
To fix this, we'll refactor the code to not redirect.

atsmith813:
To our awesome contributors, please note that this issue will be worked on by a Forem team member and is not up for grabs.

citizen428:
> To fix this, we'll refactor the code to not redirect.

This means changing a pretty common Rails idiom. Pretty much every Rails tutorial on the planet will do the following in their update action:

1. Update the actual object
2. Set a flash message
3. Redirect to the original page

Two prominent examples:

* [The Rails Guides](https://edgeguides.rubyonrails.org/getting_started.html):
    ```ruby
    def update
      @article = Article.find(params[:id])

     if @article.update(article_params)
        redirect_to @article
      else
        render 'edit'
      end
    end
    ```
* [Michael Hartl's Rails tutorial](https://3rd-edition.railstutorial.org/book/updating_and_deleting_users), the most used/recommended Rails tutorial out there:
    ```ruby
[...]

PR #10499, opened 2020-09-30
Remove redirect for listings#update

atsmith813:
- [x] Refactor
- [x] Bug Fix

## Description
This refactors `listings#update` to avoid doing a redirect causing a race condition when busting cache. An alternative approach here is to create a listings "show" page/action to render to avoid issuing queries to populate the necessary data for `index`. I'm open to that as well if folks prefer that option!

https://github.com/forem/forem/issues/10338

1. Fire up the app locally and sign in.
2. Add credits to your user if you don't have any already. You can do this in a console by running `Credit.add_to(user, 1_000)` where `user` is the `User` object of your signed-in user.
3. Create a listing.
4. Update the listing - you shouldn't see any errors and you'll see it render the index view as it did before this change.

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed - I'll do a separate PR at the end that documents this quirk as a whole.

dog_racing_human_in_push_car

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -133,9 +133,7 @@ class ListingsController < ApplicationController
     # cache.
     #
     # https://github.com/forem/forem/issues/10338#issuecomment-693401481
-    published_listings = Listing.where(published: true)
-
-    @listings = published_listings
+    @listings = Listing.where(published: true)
       .order(bumped_at: :desc)
       .includes(:user, :organization, :taggings)
       .limit(12)
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -33,8 +33,7 @@ class ListingsController < ApplicationController
   # rubocop:enable Rails/LexicallyScopedActionFilter
 
   def index
-    published_listings = Listing.where(published: true)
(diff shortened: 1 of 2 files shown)


app/controllers/listings_controller.rb
```diff
@@ -124,7 +124,23 @@ def process_unsuccessful_update
...
+    # The following sets variables used in the index view. We render the
+    # index view directly to avoid having to redirect.
+    #
+    # Redirects lead to a race condition where we redirect to a cached view
+    # after updating data and we don't bust the cache fast enough before
+    # hitting the view, therefore stale content ends up being served from
+    # cache.
+    #
+    # https://github.com/forem/forem/issues/10338#issuecomment-693401481
+    @listings = Listing.where(published: true)
```
mstruve, line 136: Might be nice to take this "default scope" for listings which is also used in the `index` view and create a method for it. That way if we want to change the index view we dont forget about this straggler down here.

mstruve (approved):
Tested, works great!!!

app/controllers/listings_controller.rb
atsmith813: Great idea! Updated!

vaidehijoshi (approved):

PR #10500, opened 2020-09-30
Return Cache Bust Response When Busting a Cached Path

mstruve:
- [x] Refactor

## Description
When we updated this code to bust the cached paths we went from returning the cache bust response to returning the object we were using to do the busting. This PR adds the cache bust response to that object so we can view it which can be very helpful if you have to manually bust stuff and want to confirm that it was successful or not. Let me know what you all think! 

https://github.com/orgs/forem/projects/6#card-45783449

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/spec/services/edge_cache/bust_spec.rb
+++ b/spec/services/edge_cache/bust_spec.rb
@@ -4,6 +4,21 @@ RSpec.describe EdgeCache::Bust, type: :service do
   let(:user) { create(:user) }
   let(:path) { "/#{user.username}" }
 
+  describe "#call" do
+    before do
+      configure_fastly
+    end
+
+    let(:cache_bust_service) { described_class.new(path) }
+
+    it "returns cache bust response" do
+      allow(cache_bust_service).to receive(:bust_fastly_cache).and_return("success")
+
+      cache_bust_service.call
+      expect(cache_bust_service.response).to eq("success")
+    end
+  end
+
   describe "#bust_fastly_cache" do
     context "when fastly is not configured" do
       before do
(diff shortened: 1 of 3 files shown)


spec/services/edge_cache/bust_spec.rb
```diff
@@ -4,6 +4,21 @@
...
+      configure_fastly
+    end
+
+    let(:cache_bust_service) { described_class.new(path) }
+
+    it "returns cache bust response" do
+      allow(cache_bust_service).to receive(:bust_fastly_cache).and_return("success")
+
+      cache_bust_service.call
+      expect(cache_bust_service.response).to eq("success")
```
vaidehijoshi, line 18: I think we ought to move this expectation into the `"when fastly is configured"` context block and add an equivalent expectation to the `"when openresty is configured"` test on line `88` since we're testing the other outcomes of `call` in those tests already. We could also test that `response` is `nil` (since you added logic for that) in the `"when openresty is not configured"` context block.

Just trying to avoid future confusion in this file since it is currently set up to very granularly test fastly and nginx behavior.

vaidehijoshi (commented):
This is great! Good call on adding this. I just had a comment about where we put the test for this, but I'm onboard to add response to the object in general 🙌

spec/services/edge_cache/bust_spec.rb
mstruve: Good call, I actually solved this issue a different way, wrote the spec, then rewrote the code and left the spec here. Makes much more sense to move it.

vaidehijoshi (approved):

atsmith813 (approved):
🚀

Issue #10283, opened 2020-09-11
Remove CAPTCHA for mods


michael-tharrington:
**Is your feature request related to a problem? Please describe.**

Kind of... mods are trusted users and regularly reporting spam/abuse to us. While they have specific mod tools that help to efficiently report comments/articles/users, they occasionally need to use the public https://dev.to/report-abuse route to provide more context about what they are reporting.

We've heard feedback that CAPTCHA is slowing folks down here.

**Describe the solution you'd like**

Remove CAPTCHA on dev.to/report-abuse for mods.

**Describe alternatives you've considered**

I've not actually considered any other alternatives here, but am open to suggestions!

**Additional context**

A few mods brought up that CAPTCHA on this page is annoying and time consuming.

There's another issue here that focuses on upgrading reCAPTCHA - https://github.com/forem/forem/issues/9919 ... the convo there has a number of details that could be relevant:
1) Should we switch from reCAPTCHA to hCAPTCHA?
2) We need to make sure that our CAPTCHA practices are accessibility-friendly.

yellow1912:
Or use the recaptcha v3. I really don't understand why the current captcha keeps showing me picture after picture. There can be much better way to do this for example a hidden credit/trust system.

yellow1912:
![Screenshot_20201003-125157](https://user-images.githubusercontent.com/374134/94984605-e966af00-0577-11eb-83d7-da9dcb3c7cbc.png)
So easy to spam

yellow1912:
![Screenshot_20201003-125757](https://user-images.githubusercontent.com/374134/94984623-15823000-0578-11eb-8925-79ce08bcae7d.png)
And impossible to report, keeps showing me picture after picture.

rhymes:
Hi @yellow1912 have you tried to disable your extensions? I'm not sure but it could be that. What browser are you using?

yellow1912:
Let me try, I was on chrome, had tons of extensions. I will try on edge which does not have any extension and will report back.

yellow1912:
Hello, seems like captcha on chrome works fine, the one I'm having problem with is actually the Android app version of dev.to.

rhymes:
@yellow1912 thank you! If you've installed the DEV Android app and are experiencing issues with it, please open an issue to the app repository so we can better track it: https://github.com/thepracticaldev/DEV-Android

thank you!

Issue #9919, opened 2020-08-21
Upgrade reCAPTCHA to v3


fdoxyz:
**Is your feature request related to a problem? Please describe.**

Based on community reports (#9908) we see a bit of friction on abuse reports secured by reCAPTCHA (and probably any other flows secured by it). The "I'm not a robot" [checkbox (v2)](https://developers.google.com/recaptcha/docs/display) is reliable probably for most users, but once the connection is [using a VPN](https://superuser.com/questions/1337801/why-is-recaptcha-image-fade-in-out-so-slow) or even on custom browsers (like the native apps) the experience seems to degrade quite a bit.

**Describe the solution you'd like**

[reCAPTCHA v3](https://developers.google.com/recaptcha/docs/v3) seems un-intrusive to the user experience. Although it presents a challenge to implement, because the requests will always succeed and will include a "trust score". It would be up to us to discard or accept the reports in this case, which would require a refactor.

**Describe alternatives you've considered**

Maybe abuse reports could just include the "trust score" and a quick button to manually "discard" the report if at first glance doesn't seem like a valid report instead of having the system discard reports that come in with a low score (since report abuse could have a low score because it was submitted using a VPN). This could be helpful to avoid discarding valid reports because reCAPTCHA didn't trust them.

michael-tharrington:
Heyo!

I also wanted to chime in on this one as a DEV user recently wrote in suggesting that we drop reCAPTCHA in favor of hCAPTCHA.

They mentioned that reCAPTCHA can be pretty annoying for the end user, and said "hcaptcha LOOKS like recaptcha, but you can actually solve it pretty quickly, and no disappearing images, or clipped items you have to guess at."

The pointed to this blog post - https://blog.cloudflare.com/moving-from-recaptcha-to-hcaptcha/ for info on changing it.

rhymes:
A lot of people in the comments there seem to be unhappy with hCaptcha :(

bdmorin:
Another point of view: https://dev.to/elmuerte/hcaptcha-instead-of-google-s-recaptcha-2c92

Also - this epically long thread: https://news.ycombinator.com/item?id=22812509

I'm unsure if hcaptcha is the panacea solution; however I'm not seeing a ton of contenders. 
https://www.phpcaptcha.org/ (self hosted)

If you are looking to stop bot specific transactions, you might want to invest in bot detection, then only show the captcha on a high score, karma scoring based on user profiles and past abuse might be useful too.  example:
[...]

PR #10502, opened 2020-09-30
Refactor reCaptcha code and bypass in report abuse for trustworthy users only

fdoxyz:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

There's been plenty of reports that the reCaptcha in the abuse report form is problematic. This is usually the case in certain scenarios, i.e. users that use a VPN usually are "non-trustworthy" from Google's standpoint so it becomes very difficult to get these reports in (see the discussion in #9919).

**After leaving this PR in the ice box for a few weeks it still does the same thing but using a different implementation approach:**

Due to ReCaptcha being used in the Registration Form this PR refactors reCaptcha logic into it's own service 
- Anywhere else reCaptcha wants to be used it can just rely on using a thoroughly tested service
- It helps avoid spreading out reCaptcha code throughout controllers, helpers and duplicated tests
- For this I used #11265 (service generator)

#9919
#10283

1. When logged in with a user that hasn't received a vomit reaction (at least 1 month old) you should be able to fill out the report abuse form without the reCaptcha requirement.
1. Unauthenticated users will always have to complete the reCaptcha test

## Added tests?

- [ ] yes
- [x] no, because I adapted the existing tests
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

these arent the droids youre looking for


fdoxyz:
Just updated the validation because I totally missed #10283, this now takes trusted users in consideration.

mstruve (commented):
Making abuse reports is a big target for our Pentesters which is why I think even though we have the user checks in place we are also going to need some sort of throttling for this if we bypass the recaptcha. Without throttling(and even likely with it), I really think we are going to get spammed.

fdoxyz:
@mstruve makes total sense! Quick question though as I may not be well aware of this:

What is the difference between throttling and the [rate_limit!](https://github.com/fdoxyz/forem/blob/master/app/controllers/feedback_messages_controller.rb#L7) in place?

If throttling is a big change to introduce I can leave only the `.auditable?` check so we can at least bypass only on trusted/mod/admin users ( #10283 )

mstruve:
Rate limiting is just another term for throttling in my head. Both can be accomplished in multiple different ways. We can either use our RateLimitChecker to accomplish it which is what `rate_limit!` does or we can use [Rack::Attack](https://github.com/fdoxyz/forem/blob/master/config/initializers/rack/attack.rb#L4) which is a gem that inserts middleware which is run on every request we get. 

For something like this where its one very specific endpoint I would lead towards using the RateLimitChecker. Rack Attack is great when you need to throttle a wide range of endpoints but if you are doing just 1 and it's not super popular, then the overhead of checking it for EVERY request doesnt really make sense. Rack Attack for context, has to hit Redis to do each throttle check. Redis is fast but we still need to be cognizant of our usage. 

Let me know if you have other questions!

Throttle/RateLimit examples
https://github.com/forem/forem/pull/7683
https://github.com/forem/forem/pull/7832

fdoxyz:
Thanks for the clarification! TIL when to choose one over the other 🙌🏼

In that case I think we're safeguarded from that abuse, because we do have throttling in place for this endpoint (it's just outside the PR diff because it wasn't modified). It was actually added in that last PR you referenced 😄

Only to further clarify: The reCaptcha bypass means it won't be presented to users that meet the criteria, but the `rate_limit!` call remains there.

mstruve (approved):
Ah, you are totally right! I never make the connection between FeedbackMessage and AbuseReport 🙈 

This still makes me a little nervous with Pentesters around on DEV still but that will be fixed once we have moved them to their own Forem.

joshpuetz (approved):
This looks ok to me, only a request for a test (but not blocking)

app/helpers/feedback_messages_helper.rb
```diff
@@ -1,4 +1,11 @@
 module FeedbackMessagesHelper
+  def bypass_recaptcha?
```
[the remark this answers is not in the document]
rhymes: I think that's the only place where we have the captcha on the website

rhymes (commented):
cc @michael-tharrington to make sure the logic in the description is sound :) 

great job @fdoxyz!

app/helpers/feedback_messages_helper.rb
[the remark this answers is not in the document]
fdoxyz: I think the suggestion refers to testing the other cases where `bypass_recaptcha?` returns true, right? As the current test for it only uses the "user without vomits + 1 month old" criteria. 

It wouldn't hurt to add one or two more tests here, I'll push the changes in a bit :)

fdoxyz:
Just a heads up here: I'll be mostly offline for a little over a week.

If anyone feels like merging this feel free to do so. Otherwise I'll sync and do so when I'm back 🙂

rhymes:
@fdoxyz I set this back to draft until we have feedback from @michael-tharrington to make sure we're not missing something in the logic :) Thanks for the effort!

spec/helpers/authentication_helper_spec.rb
```diff
@@ -38,33 +38,6 @@
     end
   end
 
-  describe "#recaptcha_configured_and_enabled?" do
```
fdoxyz, line 41: This test was migrated to `spec/services/re_captcha_spec.rb` (not entirely removed). So the coverage regarding this check remains the same.

fdoxyz:
This PR was approved a while back, but since then some more reCaptcha features were added (to the registration form) and having this logic scattered felt like it could benefit from extracting it from the Helpers/Controllers (also for future places that we may want to use reCaptcha).

Because of the refactor I'm requesting review again to hear thoughts on it. Basically the logic remains the same in both the FeedbackMessage (Abuse Reports) area and also the Email Registration.

Side note: While getting things in-sync with master I ran into a few automated fixes by the Husky pre-hooks, which is why you'll see a few SVGs and other unrelated files changed. I can revert those if anyone feels like that would be best

michael-tharrington:
Y'all, I'm really sorry for missing my pings earlier @rhymes and @fdoxyz.

I do think this addresses the problem mentioned in the actual issue!

My comments were more side-related to other things that may be solved when upgrading reCAPTCHA to v3 or at least seem related to the general talk of CAPTCHA:
1. [someone suggesting to move from reCAPTCHA to hCAPTCHA](https://github.com/forem/forem/issues/9919#issuecomment-684995307)
2. [make sure CAPTCHA is accessible](https://github.com/forem/forem/issues/9919#issuecomment-690343204)
3. [user reports of folks failing CAPTCHA](https://dev.to/yellow1912/i-fail-the-spam-report-captcha-on-dev-3397)

Nothing too necessary to follow up on here!

fdoxyz:
No worries @michael-tharrington thanks a lot for confirming this!

Upgrading to a different reCaptcha version or an alternative is more complex than being a bit more flexible with trustworthy users here. If we ever get spammed by any user we can tweak this approach even despite the rate limit in place. Hopefully this will be enough to get rid of the friction users face with the current implementation :)

app/services/re_captcha.rb
```diff
@@ -0,0 +1,52 @@
...
+#
+# Example: ReCaptcha.call(current_user).enabled? => true/false
+class ReCaptcha
+  include Devise::Controllers::Helpers
+
+  def self.call(current_user = nil)
+    new(current_user).call
+  end
+
+  def initialize(current_user)
```
joshpuetz, line 15: Is supporting both singleton and instance invocation a conscious choice? I'm curious how we feel about that: I like the flexibility, but wonder if it's worth the added complexity?

joshpuetz (approved):
Still a 👍 from me, just a design question (non-blocking)

app/services/re_captcha.rb
fdoxyz: It is a conscious choice in the sense that using the class methods help shorten the syntax for the registration form. This is only because it's known/expected to never need to pass in a `current_user` (because it's a registration feature).

```ruby
ReCaptcha.for_registration_disabled? # current approach

# vs

ReCaptcha.call.for_registration_disabled?
```

In all other situations we do need to pass in the `current_user` available from the Controller/View context.

I don't have strong feelings about the service approach though. I was mostly following the footprints (and the [service generator](https://github.com/forem/forem/pull/11265)) approach of using the `call` method for initialization. An alternative for this would be to replace the `call` initialization with purely class methods like this:

```ruby
ReCaptcha.call(current_user).enabled? # current approach

# vs

ReCaptcha.enabled?(current_user)
```

Maybe @rhymes, @citizen428 or anyone else have any suggestions/opinions on what would be the best approach here?

rhymes: We do have a service generator (`rails generate service` and [documentation on service objects](https://docs.forem.com/backend/service-objects/).

Maybe in a follow up PR we can split that service objects in two? It looks like you only need the following (coming up with names as I write, definitely open to better ones):

- `ReCaptcha::CheckEnabledGlobally.call(current_user)` (returns true or false, no need to have the negative method)
- `ReCaptcha::CheckEnabledForRegistration.call` (returns true or false, no need to have the negative method)

no need to halt this PR though

fdoxyz: Thanks for the input @rhymes I'm merging this and sending in a PR right away to refactor the service in two objects. We can discuss the best naming and ways to restructure there 👍🏼

Issue #10481, opened 2020-09-29
Add ability to manage Listing Categories in /admin


fdoxyz:
**Is your feature request related to a problem? Please describe.**

The Admin dashboard (`/admin`) currently has a Listings section that allows administrators to manage them but we are lacking a way to manage Listing Categories.

**Describe the solution you'd like**

A common Rails CRUD dashboard for Listing Categories (`/admin/listings/categories`). This would require a new controller with at least an `index`, `new`, `create`, `edit` and `update` actions. For reference [there's this PR (#10153) for Display Ads](https://github.com/forem/forem/pull/10153/files) that can be used as blueprint for guidance.

As this can be considered a "nested dashboard" and in order to avoid growing the admin side menu (left) too much, the following is a suggested placement of the link/button that will take the administrators to the Listing Categories dashboard:

<img width="1334" alt="Screen Shot 2020-09-29 at 05 05 45" src="https://user-images.githubusercontent.com/6045239/94551831-357dc080-0213-11eb-8a8c-548b1ef40e5f.png">

**Additional context**

This is part of a larger effort to remove the `/resource_admin` dashboard in favor of `/admin`.

Rafi993:
Hello there, If this up for taking I can help.

rhymes:
Ok @Rafi993, I've assigned this to you

Rafi993:
Thank you @rhymes

PR #10506, opened 2020-10-01
Manage listing categories in dashboard

Rafi993:
## What type of PR is this? 

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

It adds ability to add/update/remove listing categories in admin panel

closes https://github.com/forem/forem/issues/10481

https://vimeo.com/463735841/dfe6705dab

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/controllers/admin/listing_categories_controller.rb
+++ b/app/controllers/admin/listing_categories_controller.rb
@@ -44,7 +44,17 @@ module Admin
       end
     end
 
-    def destroy; end
+    def destroy
+      @listing_category = ListingCategory.find(params[:id])
+
+      if @listing_category.destroy
+        flash[:success] = "Listing Category has been deleted!"
+        redirect_to admin_listing_categories_path
+      else
+        flash[:danger] = "Something went wrong with deleting the Listing Category."
+        render :edit
+      end
+    end
 
     private
 
--- a/app/controllers/admin/listing_categories_controller.rb
+++ b/app/controllers/admin/listing_categories_controller.rb
@@ -9,8 +9,7 @@ module Admin
 
       return if params[:search].blank?
 
-      # @listing_categories = @listing_categories.where("organizations.name ILIKE :search",
-      # search: "%#{params[:search]}%")
+      @listing_categories = @listing_categories.where("name ILIKE :search", search: "%#{params[:search]}%")
     end
 
     def new; end
--- a/app/controllers/admin/listing_categories_controller.rb
+++ b/app/controllers/admin/listing_categories_controller.rb
@@ -1,0 +1,28 @@
+module Admin
+  class ListingCategoriesController < Admin::ApplicationController
+    include ListingsToolkit
+    layout "admin"
+
+    def index
+      @listing_categories = ListingCategory.order(id: :desc)
+        # .joins(:organization)
+        # .includes([:organization])
+        .page(params[:page]).per(50)
+
+      return if params[:search].blank?
+
+      # @listing_categories = @listing_categories.where("organizations.name ILIKE :search",
+      # search: "%#{params[:search]}%")
+    end
+
+    def new; end
+
+    def edit; end
+
+    def create; end
+
+    def update; end
+
+    def destroy; end
+  end
+end
--- a/app/controllers/admin/listing_categories_controller.rb
+++ b/app/controllers/admin/listing_categories_controller.rb
@@ -27,7 +27,7 @@ module Admin
         flash[:success] = "Listing Category has been created!"
         redirect_to admin_listing_categories_path
       else
-        flash[:danger] = @display_ad.errors_as_sentence
+        flash[:danger] = @listing_category.errors_as_sentence
         render new_admin_listing_category_path
       end
     end
@@ -60,6 +60,10 @@ module Admin
 
     def listing_category_params
       params.permit(:name, :cost, :rules, :slug, :social_preview_color, :social_preview_description)
+    end
+
+    def authorize_admin
+      authorize ListingCategory, :access?, policy_class: InternalPolicy
     end
   end
 end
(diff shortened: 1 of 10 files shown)


app/lib/constants/role.rb
```diff
@@ -18,6 +18,7 @@ module Role
                      "Resource Admin: Config",
                      "Resource Admin: Broadcast",
                      "Resource Admin: HtmlVariant",
-                     "Resource Admin: DisplayAd"].freeze
+                     "Resource Admin: DisplayAd",
+                     "Resource Admin: ListingCategory"].freeze
```
Rafi993, line 22: I am not sure if it should be called Listing or ListingCategory

rhymes: as it's a category, I think `ListingCategory` is correct

Rafi993:
I think some unrelated test is failing

rhymes:
@Rafi993 restarted, cross fingers :D

app/controllers/admin/listing_categories_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    def destroy
+      @listing_category = ListingCategory.find(params[:id])
+
+      if @listing_category.destroy
+        flash[:success] = "Listing Category has been deleted!"
+        redirect_to admin_listing_categories_path
+      else
+        flash[:danger] = "Something went wrong with deleting the Listing Category."
```
fdoxyz, line 54: It's best to do the same as the `update` action so the error message will be descriptive

```suggestion
        flash[:danger] = @listing_category.errors.full_messages.to_sentence
```

fdoxyz (changes requested):
Thank you for the PR @Rafi993, great job in following the sample PR from the issue! 

Functionality-wise the PR is working great locally.

I have a small change I think it would be good to include and another non-blocking suggestion. I checked the sample PR from the issue and just now noticed this one slipped in there, so not your fault at all!

With this tiny change in place I'd say this looks good.

Rafi993:
> Thank you for the PR @Rafi993, great job in following the sample PR from the issue! 
> 
> Functionality-wise the PR is working great locally.
> 
> I have a small change I think it would be good to include and another non-blocking suggestion. I checked the sample PR from the issue and just now noticed this one slipped in there, so not your fault at all!
> 
> With this tiny change in place I'd say this looks good.

Thank you @fdoxyz I'll make those changes.

fdoxyz (approved):
Awesome thanks again @Rafi993!

The codeclimate errors can be ignored since they're complaining about similar code in different places. So far we've considered it's best to maintain a similar approach in different admin controllers and maybe refactor later on in a way that would make this go away (or get rid of the codeclimate rule for this particular scenario).

juliannatetreault (commented):
Thank you so much for adding Listing Categories management to `/admin`, @Rafi993! I tested everything out locally and it all worked really well. One thing that I did notice though, is that the placeholder text in the search bar, found under `/admin/listings/categories`, on a 15-inch screen on Google Chrome is _slightly_ cut off (shown in the picture below). That aside, everything looked and worked great! 🎉 
Screen Shot 2020-10-01 at 12 54 38 PM

mstruve (approved):
Tested, works great!

Rafi993:
> Thank you so much for adding Listing Categories management to `/admin`, @Rafi993! I tested everything out locally and it all worked really well. One thing that I did notice though, is that the placeholder text in the search bar, found under `/admin/listings/categories`, on a 15-inch screen on Google Chrome is _slightly_ cut off (shown in the picture below). That aside, everything looked and worked great! 
> Screen Shot 2020-10-01 at 12 54 38 PM

Thank you @juliannatetreault I did try to emulate 15 inch mac with and without retina display in chrome dev tools with following values

* size 2880 x 1800
* I tried device pixel ration 1, 2 and 3 (default value for my screen is 2)
* My screen size is 14 inches hdpi screen

 I also tried couple of other screen resolutions too but was unable to reproduce this issue.

fdoxyz:
I do get the same effect on 13 inch chrome, but also ran into this over with DisplayAds

Screen Shot 2020-10-01 at 20 39 54

I couldn't find a way to make the input grow horizontally there (I tried a few things from https://storybook.forem.com/).

A suggestion: Shortening the placeholder to "Search by name" would make it fit (like in DisplayAds) and it wouldn't be cut out.
Maybe nice to have: A fix for the element to grow horizontally (could also be used to enhance other dashboards like this one)

Rafi993:
> I do get the same effect on 13 inch chrome, but also ran into this over with DisplayAds
> 
> Screen Shot 2020-10-01 at 20 39 54
> 
> I couldn't find a way to make the input grow horizontally there (I tried a few things from https://storybook.forem.com/).
> 
> A suggestion: Shortening the placeholder to "Search by name" would make it fit (like in DisplayAds) and it wouldn't be cut out.
> Maybe nice to have: A fix for the element to grow horizontally (could also be used to enhance other dashboards like this one)

Thank you @fdoxyz I have changed the text to "Search by name" now.

Rafi993:
Thank you @fdoxyz  should I revert the "search by name" placeholder change?

fdoxyz:
So sorry @Rafi993 that's my bad! I was checking in to see if CI had failed on mobile and closed by accident 🙈

I'd love to give the opportunity for other different timezones to weigh in, we may get a suggestion for manipulating the input field a little better. It still has my approval 🙌

ludwiczakpawel:
There's not an easy way (right now!) to manipulate easily the width of the input.. Changing the label to be shorter is a way to go IMO: "Search by name" and call it a day :)

We can of course do some hacky things but I don't think it's worth it :)

juliannatetreault (approved):
Thank you for making the change to the placeholder text, @Rafi993! I think this PR is good-to-go now. Thanks again for doing this! 🎉

Rafi993:
> Thank you for making the change to the placeholder text, @Rafi993! I think this PR is good-to-go now. Thanks again for doing this!

Thank you @juliannatetreault

Issue #10060, opened 2020-08-28
"Tag Mods" link under "/admin/mods" blows up when there are no tag mods


msarit:
**Describe the bug**

When there are no tag moderators, the `Tag Mods` link under `/admin/mods` blows up when clicked.

**To Reproduce**

1. Make sure your current user has the role `admin` or `super_admin`
(to do so, in `rails console`, run the command `User.find(<your_user_id>).add_role(:admin).save!`)
2. Make sure there are no tag moderators in your local app
(in `rails console`, the command `User.with_role("tag_moderator").count` should return `0`)
3. Navigate to `localhost:xxxx/admin/mods`
4. Click on the `Tag Mods` link
5. An `ActiveRecord::RecordNotFound` error should result

**Expected behavior**

There are many alternative behaviors, but I would expect to see a message like `"There are no Tag Moderators"`, maybe followed by short directions for making a user a tag moderator.

**Screenshots**

https://www.loom.com/share/cec177c9c91d453ab88403a2538316f3

**Desktop (please complete the following information):**

- OS, version: macOS Catalina version 10.15.5
- Browser, version: Google Chrome version 85.0.4183.83 (Official Build) (64-bit)

diogoosorio:
I was looking into this and I think that the problem here might be that the role itself (a role named `tag_moderator`) doesn't exist in your local database and that's what's triggering the exception.

If you manipulate the `state=` query string parameter to any non-existing role name, it'll also trigger this problem, which is not a very elegant way to handle that input I guess (see my screenshot's query string).

![image](https://user-images.githubusercontent.com/1484824/91659223-9ad07b80-eac6-11ea-849c-a6d852b00c74.png)

A valid role (any role, not just the ones linked on the sub-navigation of mods page) filters the list of users by that role (screenshot filtering by the `super_admin` role):

![image](https://user-images.githubusercontent.com/1484824/91659675-35ca5500-eac9-11ea-8c3e-8ddad45805f2.png)

Here's the output of filtering by a valid role, but with no results (which I think was what you mentioned on the issue's initial description):

![image](https://user-images.githubusercontent.com/1484824/91663619-e17f9f00-eae1-11ea-9f72-279eb8528db2.png)

So I believe that we're looking at 2 different things:

1. How to properly validate the `state` parameter:

    1. Should we assume that those roles always exist on the database? Or should we disable (or something similar) the mod filter buttons if the corresponding roles are missing from the database?

[...]

PR #10515, opened 2020-10-01
Fixes 500 error when viewing the "Tag Mods" page without having the role on the database

diogoosorio:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I think that the issue (#10060) is self-explanatory. What was happening with @msarit was that she was visiting the `/admin/mods?state=tag_moderator` without having a role named `tag_moderator` on the database.

The query class explicitly "blew up" in this condition. I tried to follow the approach proposed by @Zhao-Andy:

1. The query class now doesn't raise an exception in the scenario above. Instead returns an empty result set

1. The mods page now returns a generic "there are mods matching your search" warning if there are no mods matching the query

#10060 

1. Go to the `/admin/mods` page
1. Search by a random text string which yields no results and see that a "no results" message is rendered

1. Delete the `tag_moderator` role from your database (`Role.find_by(name: "tag_moderator").delete`)
1. Go to `/admin/mods?state=tag_moderator`
1. See that a "no results" message is rendered

image

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/queries/admin/moderators_query.rb
+++ b/app/queries/admin/moderators_query.rb
@@ -24,7 +24,7 @@ module Admin
     end
 
     def self.role_id_for(role)
-      Role.find_by!(name: role).id
+      Role.find_by(name: role)&.id
     end
 
     def self.search_relation(relation, search)
--- a/app/views/admin/mods/index.html.erb
+++ b/app/views/admin/mods/index.html.erb
@@ -19,40 +19,43 @@
     <%= f.submit "Search", class: "btn btn-light" %>
   <% end %>
 </div>
+<% if @mods.empty? %>
+    <div class="alert alert-warning">There are no mods matching your search criteria.</div>
+<% else %>
+  <%= paginate @mods, theme: "internal" %>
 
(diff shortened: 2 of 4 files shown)


app/queries/admin/moderators_query.rb
```diff
@@ -24,7 +24,7 @@ def self.call(relation: User.all, options: {})
     end
 
     def self.role_id_for(role)
-      Role.find_by!(name: role).id
+      Role.find_by(name: role)&.id
```
fdoxyz, line 27: I always forget about this operator, nice 🙌

fdoxyz (approved):
Thank you for the fix @diogoosorio. Tested it locally and seems to be working well

juliannatetreault (approved):
Awesome work, @diogoosorio ! This worked great locally. 🚀

PR #10518, opened 2020-10-01
Optimization:Dont Notify of Channel Leavings

mstruve:
- [x] Optimization

## Description
I find that notifications when people leave a channel, especially big group channels like CodeNewbie are super noisy and not helpful. This would skip sending push notifications for any "removed_from_channel" actions. What do you all think?

alt_text


Zhao-Andy (approved):
Makes sense to me! re: `action == "removed_from_channel"` I think in the future we could add the ability to choose whether we want to tell if someone is being removed from a channel, too.

rhymes (commented):
Good idea! Is there a chance to add a test for this? This way we can build on it and add a clause for user configurability if @Zhao-Andy's suggestion is added

mstruve:
Updates and improved to include left_channel and removed_from_channel statuses! Thanks for holding me accountable @rhymes!

rhymes (approved):
LGTM!

PR #10519, opened 2020-10-01
Spec Fix:Dont Reuse Same Username and Let Factory Set Title

mstruve:
- [x] Spec Fix

## Description
Couple issues I noticed in this spec while working with it in docker.
1) Sometimes I would get invalid user errors bc of the repeated username. Not sure why that would happen with transactions but this ensures that it never will bc the factory will set a unique username for us.
2) Let Factory for article set a unique title instead of us setting it and risking a duplicate. 

alt_text


juliannatetreault (approved):
👍 👍

Zhao-Andy (approved):

PR #10528, opened 2020-10-02
Remove flare tag from title and hightlight it within the tags row

### LINKED ISSUE #10523: Adjust flare tag display on Feed card
## Problem
Right now, the "flare tag" on feed cards sit above the title, but this visual design is clunky and not what we want to keep in long-term.

## Proposed solution
Change the way we display flare tag by highlighting it within the row of tags below the title. 

[Design mockup](https://www.figma.com/file/lMIzNY8ORLej9XInAqx34w/Untitled?node-id=0%3A1)

akshayymahajan:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
This PR does 2 things:
1. Remove the Flare Tag from the post title.
2. Moves the flare tag to the first position within the list of tags and highlights it.

Closes #10523

image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/assets/javascripts/utilities/buildArticleHTML.js
+++ b/app/assets/javascripts/utilities/buildArticleHTML.js
@@ -35,14 +35,12 @@ function buildArticleHTML(article) {
       flareTag =
         "<a href='/t/" +
         article.flare_tag.name +
-        "' class='crayons-tag'>" +
-        "<span class='crayons-story__flare-tag' style='background:" +
+        "' class='crayons-tag' style='background:" +
         article.flare_tag.bg_color_hex +
         ';color:' +
         article.flare_tag.text_color_hex +
         "'><span className='crayons-tag__prefix'>#</span>" +
         article.flare_tag.name +
-        '</span>' +
         '</a>';
     }
     if (article.class_name === 'PodcastEpisode') {
--- a/app/assets/javascripts/utilities/buildArticleHTML.js
+++ b/app/assets/javascripts/utilities/buildArticleHTML.js
@@ -55,7 +55,7 @@ function buildArticleHTML(article) {
     }
 
     var tagString = '';
-    var tagList = article.tag_list || article.cached_tag_list_array;
+    var tagList = article.tag_list || article.cached_tag_list_array || [];
     if (flareTag) {
       tagList = tagList.filter(function (tag) {
         return tag !== article.flare_tag.name;
--- a/app/assets/javascripts/utilities/buildArticleHTML.js
+++ b/app/assets/javascripts/utilities/buildArticleHTML.js
@@ -25,8 +25,45 @@ function buildArticleHTML(article) {
 
   if (article) {
     var container = document.getElementById('index-container');
+
+    var flareTag = '';
+    var currentTag = '';
+    if (container) {
+      currentTag = JSON.parse(container.dataset.params).tag;
+    }
+    if (article.flare_tag && currentTag !== article.flare_tag.name) {
+      flareTag =
+        "<a href='/t/" +
+        article.flare_tag.name +
+        "' class='crayons-tag'>" +
+        "<span class='crayons-story__flare-tag' style='background:" +
+        article.flare_tag.bg_color_hex +
+        ';color:' +
+        article.flare_tag.text_color_hex +
+        "'><span className='crayons-tag__prefix'>#</span>" +
+        article.flare_tag.name +
+        '</span>' +
+        '</a>';
+    }
+    if (article.class_name === 'PodcastEpisode') {
+      flareTag = "<span class='crayons-story__flare-tag'>podcast</span>";
+    }
+    if (article.class_name === 'Comment') {
+      flareTag = "<span class='crayons-story__flare-tag'>comment</span>";
+    }
+    if (article.class_name === 'User') {
+      flareTag =
+        "<span class='crayons-story__flare-tag' style='background:#5874d9;color:white;'>person</span>";
+    }
+
     var tagString = '';
     var tagList = article.tag_list || article.cached_tag_list_array;
+    if (flareTag) {
+      tagList = tagList.filter(function (tag) {
+        return tag !== article.flare_tag.name;
+      });
+      tagString += flareTag;
+    }
     if (tagList) {
       tagList.forEach(function buildTagString(t) {
         tagString =
@@ -51,32 +88,6 @@ function buildArticleHTML(article) {
         '#comments" class="crayons-btn crayons-btn--s crayons-btn--ghost crayons-btn--icon-left "><svg class="crayons-icon" width="24" height="24" xmlns="http://www.w3.org/2000/svg"><path d="M10.5 5h3a6 6 0 110 12v2.625c-3.75-1.5-9-3.75-9-8.625a6 6 0 016-6zM12 15.5h1.5a4.501 4.501 0 001.722-8.657A4.5 4.5 0 0013.5 6.5h-3A4.5 4.5 0 006 11c0 2.707 1.846 4.475 6 6.36V15.5z"/></svg>' +
         commentsCount +
         '<span class="hidden s:inline">&nbsp;comments</span></a>';
-    }
-
-    var flareTag = '';
-    var currentTag = '';
-    if (container) {
-      currentTag = JSON.parse(container.dataset.params).tag;
-    }
-    if (article.flare_tag && currentTag !== article.flare_tag.name) {
-      flareTag =
-        "<span class='crayons-story__flare-tag' style='background:" +
-        article.flare_tag.bg_color_hex +
-        ';color:' +
-        article.flare_tag.text_color_hex +
-        "'>#" +
-        article.flare_tag.name +
-        '</span>';
-    }
-    if (article.class_name === 'PodcastEpisode') {
-      flareTag = "<span class='crayons-story__flare-tag'>podcast</span>";
-    }
-    if (article.class_name === 'Comment') {
-      flareTag = "<span class='crayons-story__flare-tag'>comment</span>";
-    }
-    if (article.class_name === 'User') {
-      flareTag =
-        "<span class='crayons-story__flare-tag' style='background:#5874d9;color:white;'>person</span>";
     }
 
     var rc = article.public_reactions_count;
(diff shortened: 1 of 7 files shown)


app/javascript/articles/components/TagList.jsx
```diff
@@ -1,16 +1,35 @@
...
+      {flare_tag && (
+        <a className="crayons-tag" href={`/t/${flare_tag.name}`}>
+          <span
+            className="crayons-story__flare-tag"
+            style={{
+              background: flare_tag.bg_color_hex,
+              color: flare_tag.text_color_hex,
+            }}
+          >
+            {`#${flare_tag.name}`}
```
ludwiczakpawel, line 20: Can you wrap `#` with `<span className="crayons-tag__prefix">` similar to [how we do it with other tags](https://github.com/forem/forem/pull/10528/files#diff-bcb8e7ee1cc3e3133c3796a5a01689f0R26)?

ludwiczakpawel (changes requested):
Oh damn, I love it! Thank you so much! I have one very tiny change request - other than that it's awesome!

app/javascript/articles/components/TagList.jsx
akshayymahajan: Sure thing!

ludwiczakpawel:
I've just realized, that when you scroll down quite a lot on your feed we actually load feed cards differently (I don't know why...) - so the feed card code is generated in multiple places. That jsx component is one of them and it works great. But we gonna need it in other places too.. Do you think you can look at it as well? You can also spot it in logged out version of the page where we also load feed card differently.

akshayymahajan:
Sure.. I'll look into it..

ludwiczakpawel:
@akshayymahajan awesome! I'm super excited about this one  :)

Btw. here are some references you may find useful: 

- [_single-story.html.erb](https://github.com/forem/forem/blob/master/app/views/articles/_single_story.html.erb)
- [buildArticleHTML.js](https://github.com/forem/forem/blob/master/app/assets/javascripts/utilities/buildArticleHTML.js)

akshayymahajan:
Hey @ludwiczakpawel, I have updated both files and tested in the logged out page and the search results. Can you check again?

app/assets/javascripts/utilities/buildArticleHTML.js
```diff
@@ -25,8 +25,45 @@ function buildArticleHTML(article) {
...
+        article.flare_tag.name +
+        "' class='crayons-tag'>" +
+        "<span class='crayons-story__flare-tag' style='background:" +
+        article.flare_tag.bg_color_hex +
+        ';color:' +
+        article.flare_tag.text_color_hex +
+        "'><span className='crayons-tag__prefix'>#</span>" +
+        article.flare_tag.name +
+        '</span>' +
+        '</a>';
```
ludwiczakpawel, line 46: We should drop this extra container `crayons-story__flare-tag` - it's not needed anymore.

```suggestion
    flareTag = `<a href="/t/${article.flare_tag.name}" class="crayons-tag" style="background: ${article.flare_tag.bg_color_hex}; color: ${article.flare_tag.text_color_hex};">
      <span class="crayons-tag__prefix>#</span>
      ${article.flare_tag.name}
    </a>`;
```

app/assets/stylesheets/components/stories.scss
```diff
@@ -135,7 +135,7 @@
 
   &__flare-tag {
```
ludwiczakpawel, line 136: We should now be able to remove this entire class, right?

app/javascript/articles/components/TagList.jsx
```diff
@@ -1,16 +1,36 @@
...
+  let tagsToDisplay = tags;
+  if (flare_tag) {
+    tagsToDisplay = tagsToDisplay.filter((tag) => tag !== flare_tag.name);
+  }
+  return (
+    <div className="crayons-story__tags">
+      {flare_tag && (
+        <a className="crayons-tag" href={`/t/${flare_tag.name}`}>
+          <span
+            className="crayons-story__flare-tag"
```
ludwiczakpawel, line 14: Similarly to what I suggested in [previous comment](https://github.com/forem/forem/pull/10528/files#diff-419ded8bb2b41fca12af2061c1272094), we should drop `.crayons-stroy__flare-tag`

app/assets/javascripts/utilities/buildArticleHTML.js
ludwiczakpawel: The flare tag is basically now a classic `crayons-tag` with custom background and color.

app/views/articles/_single_story.html.erb
```diff
@@ -51,11 +51,19 @@
...
         <a href="<%= story.path %>" data-preload-image="<%= cloud_cover_url(story.main_image) %>" id="article-link-<%= story.id %>">
-          <%= render "articles/tag_identifier", story: story, tag: @tag %>
+          <%= story.title %>
         </a>
       </h2>
       <div class="crayons-story__tags">
-        <% story.cached_tag_list_array.each do |tag| %>
+        <% flare_tag = FlareTag.new(story, @tag).tag %>
+        <% if flare_tag %>
+          <a href="/t/<%= flare_tag.name %>" class="crayons-tag"><span class="crayons-story__flare-tag" style="background:<%= flare_tag.bg_color_hex %>;color:<%= flare_tag.text_color_hex %>"><span className="crayons-tag__prefix">#</span><%= flare_tag.name %></span></a>
```
ludwiczakpawel, line 60: And same here.

ludwiczakpawel (changes requested):
Awesome! I left few last change request - nothing really major :)

akshayymahajan:
@ludwiczakpawel updated!

app/assets/stylesheets/components/stories.scss
```diff
@@ -133,21 +133,6 @@
...
-    position: relative;
-    top: -2px;
-    line-height: 1em;
-    display: inline-block;
-    border-radius: var(--radius);
-    background: var(--base-60);
-    margin-right: var(--su-1);
-    color: var(--base-inverted);
-  }
-
```
ludwiczakpawel, line 150: Whoops, looks like this is still needed for search results. I apologize for the confusion! Other than that, it's good to go 👍

ludwiczakpawel (approved):
One tiny change and it's ready!

lisasy (approved):
Nice! Thank you :) This looks great.

app/assets/stylesheets/components/stories.scss
akshayymahajan: Added back!

rhymes:
@akshayymahajan it looks like some of the tests are failing, please take a look at the build here: https://github.com/forem/forem/pull/10528/checks?check_run_id=1201884642

akshayymahajan:
Hey @rhymes, I have updated the code to fix the failing test.

ludwiczakpawel (changes requested):
Hey, I've taken it for another ride and there's still something a little bit off, when I scroll down the feed. Here's a [video](https://d.pr/v/Pzx0DE) showing the issue. Looks like on the feed cards that are being loaded with infinity scroll we don't apply appropriate colors/styling to the flare tag.

akshayymahajan:
Hey @ludwiczakpawel, looks like the data being returned from the API has some null values. Not sure if this is related to the changes I made. Would really appreciate some help.

Screenshot 2020-10-05 at 2 55 34 PM

ludwiczakpawel:
Interesting. I can see this working properly on prod dev.to... 

@rhymes I SUMMON YOU.

summon rhymes

rhymes:
Hi @akshayymahajan, that indicates that the tag has no defined colors: 

https://github.com/forem/forem/blob/a943f045c56237f4ec8a4b2e504a6ac3f7fb0db9/app/models/tag.rb#L26-L27

both hex colors can be `nil`

akshayymahajan:
Thanks, @rhymes. @ludwiczakpawel in that case, I think it would be best to add back the `span.crayons-stroy__flare-tag` wrapper in all the places. What do you say?

ludwiczakpawel:
@rhymes are you sure? can you please checkout the screen recording here: https://d.pr/v/Pzx0DE - it's the same tag. one time it has colros applied and the other time does not... which means that this tag in general has colors applied but they are not returned in certain cases.

rhymes:
@ludwiczakpawel I'm not very familiar with how flare tags work but I came up with these two clues:

`article.flare_tag` always calls the `FlareTag` class with just the article object, without the "except tag" parameter:

https://github.com/forem/forem/blob/a943f045c56237f4ec8a4b2e504a6ac3f7fb0db9/app/models/article.rb#L312-L314

https://github.com/forem/forem/blob/a943f045c56237f4ec8a4b2e504a6ac3f7fb0db9/app/services/flare_tag.rb#L18-L21

It looks to me that if `FlareTag` were to be initialized with a second argument, it would send back `nil` even if the `Tag` itself has the colors assigned to it:

https://github.com/forem/forem/blob/a943f045c56237f4ec8a4b2e504a6ac3f7fb0db9/app/services/flare_tag.rb#L37-L40

The clue for what might be going on in your recording is to be found in:

https://github.com/forem/forem/blob/a943f045c56237f4ec8a4b2e504a6ac3f7fb0db9/app/views/articles/_tag_identifier.html.erb#L1-L3

which is passing `@tag` as the "except tag". 

This comes from https://github.com/forem/forem/blob/master/app/views/articles/_single_story.html.erb

I seem to recall that the first page of the feed is rendered one way, subsequent pages are rendered in a different way but I'm not sure how all that works, @joshpuetz or @nickytonline might instantly know what's going on here.

There might be a different behavior between the flare tag appearing in the first batch of articles and the following batches

akshayymahajan:
I tried removing the "except tag" from the call, but it still returns `null` value for colors. I also tried the latest master branch locally and the JSON returned still has `null` values. Looks to me like this is something related to the local development environment.

ludwiczakpawel:
@akshayymahajan let me give it another look tomorrow and I'll get back to you then.

ludwiczakpawel:
Yea, looks like it's being broken on master as well.. So I'll go ahead and approve this PR :) Thanks for all the work you've done, it's really something I've been looking for for a quite long time so I appreciate your effort :)

ludwiczakpawel (approved):

Issue #6755, opened 2020-03-21
Reading list API of an authenticated user


srebalaji:
**Is your feature request related to a problem? Please describe.**
Dev API doesn't have the option to list the authenticated user's reading list 

**Describe the solution you'd like**
Build an API to list user's reading list

srebalaji:
@benhalpern What do you think? If the Dev team approves, I can start working on it.

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

Bhacaz:
@rhymes I would like to work on this.

rhymes:
Cool, I've assigned this to you! Make sure the API spec documentation is also updated if you can, thank you!

cmgorton:
Looks like this feature was added per PR #10540. I am going to close this issue. If this feature is still lacking let me know and we can discuss opening it back up. Thanks.

PR #10540, opened 2020-10-02
API - New endpoint to retrieve the articles in the reading list of the authenticated user

Bhacaz:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

Add a new endpoint in the DEV API to retrieve the reading list articles for the authenticated user. 

`GET api/articles/me/readinglist`

This feature was requested in Issue #6755.

You can try the new API endpoint via a `curl` command with a valid _api-key_.

```sh
curl -H "api-key: API-KEY" http://localhost:3000/api/readinglist 
```

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [x] docs.forem.com/api
- [ ] readme
- [ ] no documentation needed

Cat reading


benhalpern:
Big fan of this! It looks good _at a glance_ but I'll circle back for a more comprehensive look soon.

app/controllers/api/v0/articles_controller.rb
```diff
@@ -104,6 +104,23 @@ def me
           .decorate
       end
 
+      def reading_list
+        per_page = (params[:per_page] || 30).to_i
+        num = [per_page, 1000].min
```
mstruve, line 109: I think the max per_page should be a bit lower here given the complexity of the SQL statement below. How about 100? It also might be a good idea to pull this out and set it as a constant.
```
MAX_READING_LIST_PAGE_SIZE=100 # or something like that
```

mstruve (commented):
Pulled this down and tested it, works great!!!! Love this feature! 

One small suggestion on default per_page size but otherwise LGTM!

rhymes (changes requested):
Hi @Bhacaz, great start! I have a couple of requests stemmed by the same argument: the reading list is and should be its own resource.

Preamble: the API (whose I stands for interface) doesn't necessarily need to reflect the organization of the code or of database models.

In this regard I think we should promote the "reading list" as its own collective resource: a collection of items. These items appear to be articles but they are indeed reactions. Those reactions have metadata that I believe it should be part of the serialization of the "reading list item" sent to the client. 

In the context of the reading list, I think we should send back the date of creation as well.

For this reason I propose the following two changes:

- the URL should be `/api/readinglist`, simple as that :) `/api/articles/me` is a bit of a patch. In an ideal world we should just have a `/api/articles/` and check authorization to choose which articles to send. Nesting the reading list under there is a bit odd.

- the serialized JSON should be something like (using pseudo json just to make this quicker :D):

```json
[
  {
    "type_of": "readinglist",
    "created_at": "2020-10-03T04:49:16Z",
    "article": { 
    } 
   ...
  }
]
```

`article` then can be the rendering of the [`_article` partial](https://github.com/forem/forem/blob/master/app/views/api/v0/articles/_article.json.jbuilder).

Let me know what you think, thank you!

Bhacaz:
@rhymes 

Thanks for the review and the comments.

> the URL should be /api/readinglist

It's a good idea and if for any reason in the future we want to add some CRUD functions this resource will be at a good and significant place (URL).

> Those reactions have metadata that I believe it should be part of the serialization of the "reading list item" sent to the client.

I agree, the payload should return some useful information about the reaction: `created_at` and maybe the `status` if someday handle archived elements from the reading list.

But I have some concern about placing the `article` inside an object in the response body. I think the users of the API will expert to receive directly the `articles`, not a `reaction/readinglist`. May I suggest the following serialized JSON:

```js
{
  "type_of":"article",
  "id":18,
  "title":" The Wings of the Dove Inventore recusandae",
  "description":"Kickstarter normcore tattooed...",
  // Rest of the article attributes
  "readinglist": {
    "type_of": "readinglist"
    "status": "valid"
    "created_at": "2020-10-03T04:49:16Z",
  }
}
```

But I understand your point, the readinglist is in fact a list of reactions with an article `reactable_type`.

I let you have the last word. 😃

rhymes:
@Bhacaz I'm imagining the use case which would be to build a list of favorite articles by a user, in theory by building an "hypertexted" API we shouldn't embed the `article` object but link to the API representation of it. As v0 is not "hypertexted" I think we can get away with embedding the article object. One of of the things I don't like about bookmarks is that I can't easily see when that bookmark has been added so I think metadata should be there, as we both agree. Great call about adding `status`.

What I'm less convinced about is to embed the full representation of the `article` object inside. I was thinking that the article endpoints are cached for a reason (serializing articles can be not super fast), so if we embed their exact representation inside the "list of saved items" (the reading list) we might want to eventually cache that as well (not necessary in this PR, we can adjust depending on real world performance).

If we look at the UI version of the reading list you get the following information:

Screenshot_2020-10-04 Reading List - DEV Community 👩‍💻👨‍💻

The body is not there for example, nor are the stats. What if we adhere to that programmatically?

In conclusion:

- I think the serialization should make it super clear that this is a list of reading list items
- The article should be there but it doesn't necessarily should contain the entire representation of the `index` or `show` endpoints. Maybe we should add only the article data that would suffice to build a list of favorites

What do you think?

cc @benhalpern and @mstruve as well as they provided feedback earlier

Bhacaz:
Change in 568dd93 

* The `readinglist` have now his own _resources_.
* `GET api/readinglist` body return the information about the reaction and an embed `article` (the same information from `GET api/articles`), like it's was previously [suggested](https://github.com/forem/forem/pull/10540#pullrequestreview-501521512)
* I preload the maximum of records to use has must `ActiveRecord::Relation#select` has I can. I'm not sure if it's a commun pattern here.

docs/api_v0.yml
```diff
@@ -829,6 +840,34 @@ components:
               type: string
               enum: [bump, publish, unpublish]
 
+    Readinglist:
```
rhymes, line 843: I think we should name this `ReadingList`

docs/api_v0.yml
```diff
@@ -1616,6 +1655,8 @@ tags:
     description: Listings are classified ads
   - name: podcast-episodes
     description: Podcast episodes
+  - name: readinglist
+    description: User readinglist
```
rhymes, line 1659: `User's reading list` (it's called "reading list" in the copy)

docs/api_v0.yml
```diff
@@ -2790,6 +2831,54 @@ paths:
               -d '{"listing":{"title":"Title"}' \
               https://dev.to/api/listings/{id}
 
+  /readinglist:
+    get:
+      operationId: getReadinglist
+      summary: User's readinglist
```
rhymes, line 2837: `User's reading list`

docs/api_v0.yml
```diff
@@ -2790,6 +2831,54 @@ paths:
...
+      description: |
+        This endpoint allows the client to retrieve a list of the reactions with the type readinglist along with the related article for the authenticated user.
```
rhymes, line 2839: `a list of readinglist reactions`

docs/api_v0.yml
```diff
@@ -2790,6 +2831,54 @@ paths:
...
+
+        "Articles" are all the posts that users create on DEV that typically
+        show up in the feed. They can be a blog post, a discussion question,
+        a help thread etc. but is referred to as article within the code.
```
rhymes, line 2843: I think we can do away with this as we're sendinb back reading list items, not articles

docs/api_v0.yml
```diff
@@ -2790,6 +2831,54 @@ paths:
...
+
+        Reading list will be in reverse chronological order base
+        on the creation if the reaction.
```
rhymes, line 2846: `of the reaction`

docs/api_v0.yml
```diff
@@ -2790,6 +2831,54 @@ paths:
...
+
+        It will return the reading list with his article with pagination.
+        By default a page will contain `30` articles.
```
rhymes, line 2849: `It will return paginated reading list items along with the articles they refer to`
`By default a page will contain 30 items`

docs/api_v0.yml
```diff
@@ -2790,6 +2831,54 @@ paths:
...
+      tags:
+        - readinglist
+      parameters:
+        - $ref: '#/components/parameters/pageParam'
+        - $ref: '#/components/parameters/perPageParam30to100'
+      responses:
+        "200":
+          description: The readinglist with a overwiew of the article
```
rhymes, line 2857: `reading list`

rhymes (commented):
Thanks! Works well, left a few notes and corrections for the copy!

Bhacaz:
@rhymes Thanks for the corrections, English is not my first language. 😃

rhymes (approved):

rhymes:
@Bhacaz thank you! Don't worry about that, commits will be squashed on merge :) Each PR corresponds to a single commit when we merge them in.

mstruve (approved):
This looks like its ready to go, we have a flaky spec that was just fixed in master. Can you update this branch and then we should get a green build and be good to go!

Bhacaz:
Codeclimate found duplicated fragment of code...

mstruve (approved):

PR #10541, opened 2020-10-02
Feature:Notify Devs when Sidekiq is Not Running in Development 

mstruve:
- [x] Feature

## Description
People missing Elasticsearch data seems to be a common issue in development which got me thinking about how we could alert people of an issue. However, usually missing Elasticsearch data is linked to Sidekiq not running which means things are not getting indexed. Rather than trying to compare ES and db counts I decided to add a warning for development if Sidekiq is not running. 

I am not a huge fan of having development specific code but I thought this would be nice to have. Let me know what you all think!

Screen Shot 2020-10-02 at 5 32 09 PM

1. Boot app with sidekiq running and see that no banner is displayed
2. boot app without sidekiq, `rails s` and verify the banner is displayed. 

## Added tests?
Since this is just for development I thought leaving it without tests is fine. Plus the other validation setup methods also did not have specs. 

alt_text


rhymes (commented):
I share your concerns as well as I'm not sure this is something we should embed in the app. What if we do something on boot in `development.rb`, the same way we check for data update scripts?

Another side of me thinks we shouldn't do this at all. If the documentation explicitly states that `bin/startup` is how you start the app, that's the way it is :) 

Anyhow, if we decide to go ahead with this I'd change the messaging towards using `bin/startup` and not running single services manually or to tell them to check the installation documentation

citizen428:
> Another side of me thinks we shouldn't do this at all. If the documentation explicitly states that `bin/startup` is how you start the app, that's the way it is :)

I second that. FWIW, I generally don't start the app this way because I have my own workflow, but I think it's implicit in any project that if you're not following the canonical way you're on your own.

atsmith813:
I love this idea! I like the idea that @rhymes suggested as well with
> What if we do something on boot in development.rb, the same way we check for data update scripts?

@rhymes @citizen428 with this
> Another side of me thinks we shouldn't do this at all. If the documentation explicitly states that bin/startup is how you start the app, that's the way it is :)

The entire development experience is legacy in the sense that the decision to use `bin/setup` and most of everything inside of it (i.e. Foreman) was made before the vast majority of the team joined. It's not something most of the team had a chance to buy-in on or contribute some feedback. In other words, I think it has some room for improvement and I'm hesitant to tie ourselves so closely to that idea for the sake of sticking to documentation (whether the concept of this PR is merged or not).

> it's implicit in any project that if you're not following the canonical way you're on your own

This is also tough. On one hand, this is how I typically feel. On the other hand and to sort of repeat myself here, I think the goal leans more towards enhancing the development experience for the team and contributors. The challenge with `bin/startup` is it's legacy, doesn't play nicely with common debugging practices (i.e. `pry`), unreadable log outputs, and more. Similar to you @citizen428, I don't use `bin/startup` and if I recall correctly, I believe others on the team don't use it either. That also makes me feel like the experience could be updated.

citizen428:
My main concern is that this will add a check for the Rails env to every single request that hits any descendant of `ApplicationController` (the `if` proc), also in production:

```ruby
before_action :verify_sidekiq_running, only: %i[index show], if: proc { Rails.env.development? }
```

I guess I would feel better about it if (a) we move the entire check into a separate concern instead of `VerifySetupCompleted` and (b) conditionally include the concern, e.g. 

```ruby
include DependenciesRunning if Rails.app.development?
```

This way we only do the check on app boot once and not on every request.

mstruve:
Im back and I updated this PR. Let me know what you all think!!! 🙏

app/controllers/concerns/development_dependency_checks.rb
```diff
@@ -0,0 +1,16 @@
...
+    before_action :verify_sidekiq_running, only: %i[index show], if: proc { Rails.env.development? }
+  end
+
+  private
+
+  def verify_sidekiq_running
+    return if Sidekiq::ProcessSet.new.size.positive?
+
+    flash[:global_notice] = "Sidekiq is not running and is needed for the app to function properly. \
+                           Use bin/startup to start the application properly.".html_safe
```
citizen428, line 14: **nitpick (non-blocking):** The `html_safe` here is superfluous, as the string contains nothing that needs escaping.

citizen428 (approved):
Thanks @mstruve! 🚀

rhymes (approved):

atsmith813 (approved):

Issue #4523, opened 2019-10-21
Allow way for authors to pin a comment on top


ananyaneogi:
A lot of times readers add some useful comment relating to the post and the author might want that to be highlighted and not be lost in other comments. Youtube has a feature of pinned comment, so this would be similar to that.

The UI can be similar to how pinned articles look on the profile!

Also I'm not aware if there has been similar feature requests, I didn't find any on my own so I opened this issue. If this is a duplicate or this feature is already in the works then please close this 🙂  

PS: Dev.to is awesome ❤️

Link2Twenty:
Discussion post has been created on Forem

https://forem.dev/thomasbnt/pin-a-message-fpo

benhalpern:
This is a great idea.

We can work on the backend functionality on this without having to  entirely figure out the design right away.

This could end up being a bit of a pain from an implementation perspective....

I'd say we could do this by adding a `pinned` boolean to comments, and then automatically bump up the `score` of the pinned comment to ensure it rises to the top (or do this otherwise, but I'd be a little concerned about having an overly complicated sorting functionality, or needing multiple queries)...

Other thoughts... Should pinning or something be a type of "reaction" which could make this a little extensible... Probably not, just thinking outloud.

Link2Twenty:
Would pinned be a comment flair?

Taniyah-L-Jackson:
Sounds like a good idea.

The setup could be similar to StackOverflow or Medium's UI

nickytonline:
As discussed, I'm assigning this to you @Taniyah-L-Jackson taniyah-l-jackson.

nickytonline:
@benhalpern, @rhymes and Ruby folk. Any reason we couldn't store the pinned comment ID on the article entity? Is it because of caching issues in regards to an article? To me it makes more sense there because it makes the relationship clearer, but `pinned` on a comment entity could make sense as well as @benhalpern mentioned.

I'm going to be pairing on this with @Taniyah-L-Jackson this week, so trying to get a general idea of where this functionality on the backend should live.

rhymes:
My proposal is the following:

I’m looking at `ProfilePin` and I wonder if it can be recycled for this by having `profile` extended to `Article` and `pinnable` extended to `Comment`. Something like:

```ruby
class ProfilePin < ApplicationRecord
  belongs_to :pinnable, polymorphic: true
  belongs_to :profile, polymorphic: true

  validates :profile_id, presence: true
[...]

PR #10542, opened 2020-10-02
Taniyah l jackson/pinned comments on articles

Taniyah-L-Jackson:
Afternoon! I wanted to ask about the possibility of using a scroll feature. Instead of having the comments surface to the top, the page could scroll to the targeted comment based on an event.  The setup for the front-end could look like a pdf word search, except the search bar could be replaced with '1 out of 5 pins' and updates with the event trigger. I'm aware that it may be a detour from the original idea, but hopefully the approach is similar. Would this be safer and user-friendly or difficult and confusing?

### DIFF
--- a/app/models/profile_pin.rb
+++ b/app/models/profile_pin.rb
@@ -1,4 +1,4 @@
-class GenericPins < ApplicationRecord #migrate later
+class ProfilePin < ApplicationRecord 
   belongs_to :pinnable, polymorphic: true
   belongs_to :profile, polymorphic: true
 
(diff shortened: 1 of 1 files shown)


rhymes:
Hi @Taniyah-L-Jackson, if possible can we keep the conversration in https://github.com/forem/forem/issues/4523 ? I understand you've already asked the question in https://github.com/forem/forem/issues/4523#issuecomment-702833324 and unfortunately no one has answered yet but it'd be difficult to split the conversation in multiple places. 

I'll answer there

PR #10544, opened 2020-10-03
Allow glitch id starting with ~

### LINKED ISSUE #10427: Glitch liquid tag does not support ids with ~ in them
**Describe the bug**

Glitch liquid tag does not support ids with `~` in them

**To Reproduce**

1. Create a new post
2. Write the following
	```liquid
	{% glitch ~the-super-tiny-compiler %}
	```
3. At the top, there will be an error message
	> Whoops, something went wrong:
	> - base: Invalid Glitch ID

**Expected behavior**

The glitch id containing `~` (tilde) should render correctly

**Screenshots**

- OS, version: Windows 10 2004
- Browser, version: Firefox 81

- Device:
- OS, version:
- Browser, version:

**Additional context**

irmela:
- [x] Bug Fix

## Description
~This PR improved the regex that is used to validate the glitch id to allow ids starting with `~`.~
This PR improves the `parse_id` method to remove any prefixed `~` before passing it to the template. 

resolves #10427

Create a new blog post, add `{% glitch ~the-super-tiny-compiler %}` and save -> no validation error should appear.

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/liquid_tags/glitch_tag.rb
+++ b/app/liquid_tags/glitch_tag.rb
@@ -2,7 +2,8 @@ class GlitchTag < LiquidTagBase
   attr_accessor :uri
 
   PARTIAL = "liquids/glitch".freeze
-  ID_REGEXP = /\A~*[a-zA-Z0-9\-]{1,110}\z/.freeze
+  ID_REGEXP = /\A[a-zA-Z0-9\-]{1,110}\z/.freeze
+  TILDE_PREFIX_REGEXP = /\A~*/.freeze
   OPTION_REGEXP = /(app|code|no-files|preview-first|no-attribution|file=\w(\.\w)?)/.freeze
   OPTIONS_TO_QUERY_PAIR = {
     "app" => %w[previewSize 100],
@@ -36,6 +37,7 @@ class GlitchTag < LiquidTagBase
 
   def parse_id(input)
     id = input.split(" ").first
+    id.gsub!(TILDE_PREFIX_REGEXP, '')
     raise StandardError, "Invalid Glitch ID" unless valid_id?(id)
 
     id
--- a/app/liquid_tags/glitch_tag.rb
+++ b/app/liquid_tags/glitch_tag.rb
@@ -3,7 +3,7 @@ class GlitchTag < LiquidTagBase
 
   PARTIAL = "liquids/glitch".freeze
   ID_REGEXP = /\A[a-zA-Z0-9\-]{1,110}\z/.freeze
-  TILDE_PREFIX_REGEXP = /\A~*/.freeze
+  TILDE_PREFIX_REGEXP = /\A~/.freeze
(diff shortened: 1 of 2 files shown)


rhymes (commented):
The code looks great @irmela! I was wondering though if you were able to actually make it work.

I noticed the following while testing it manually:

Screenshot_2020-10-04 test

Screenshot_2020-10-04 test - DEV(local) Community 👩‍💻👨‍💻


The markdown is:

```markdown
test glitch

{% glitch earthy-course %}

{% glitch ~remote-hands %}
```

So the first "glitch" has no tilde, the second does. The second never fully loads. I wonder if there's anything we need to change in the template to make those work or if it's just a problem on my end.

irmela:
Good catch @rhymes. I was only looking for _some_ glitch content to appear 🙈 The tilde should not appear in the `src` tag of the iFrame. Will fix it.

irmela:
I updated the PR description:

> ~This PR improved the regex that is used to validate the glitch id to allow ids starting with `~`.~
> This PR improves the `parse_id` method to remove any prefixed `~` before passing it to the template.

app/liquid_tags/glitch_tag.rb
```diff
@@ -3,6 +3,7 @@ class GlitchTag < LiquidTagBase
 
   PARTIAL = "liquids/glitch".freeze
   ID_REGEXP = /\A[a-zA-Z0-9\-]{1,110}\z/.freeze
+  TILDE_PREFIX_REGEXP = /\A~*/.freeze
```
rhymes, line 6: to be 100% correct, this matches "nothing at the beginning of a string", so basically anything you pass. That's because you use the `*`. As you then replace the ID with nothing is not a big deal but in theory this should simply be:

`\A~` which means "begins with tilde"

app/liquid_tags/glitch_tag.rb
```diff
@@ -36,6 +37,7 @@ def valid_id?(input)
 
   def parse_id(input)
     id = input.split(" ").first
+    id.gsub!(TILDE_PREFIX_REGEXP, "")
```
rhymes, line 40: no need to `gsub!` if the regexp uses `\A` since it's just going to scan the beginning of the string, `sub!` is fine: 

```ruby
[2] pry(main)> "~some-id".gsub!(/\A~/, "")
=> "some-id"
[3] pry(main)> "~some-id".sub!(/\A~/, "")
=> "some-id"
```

rhymes (commented):
Thanks for the changes @irmela! I noticed something with the regexp now that you're just going remove the leading tilde from the ID

Great work BTW :)

rhymes (approved):
LGTM! :)

maestromac (approved):
Thank you for this PR @irmela . LGTM!

Issue #10132, opened 2020-09-01
An API endpoint to fetch the profile pic of a user OR organization based on the username alone


Booligoosh:
**Is your feature request related to a problem? Please describe.**
I want to get an avatar URL based on a dev.to username. Since a username can be either for an organization or a user, there is currently no way to do this.

For user profile pics, I can get the profile pic by requesting `https://dev.to/api/users/by_username?url=<username>`. However, there is no corresponding endpoint for organizations at the moment (see #9212).

But even if this endpoint did exist, there would still be no way to get the profile pic from username alone, unless you made two requests, one to the organization endpoint and one to the the user endpoint, and then checked which returned a response.

**Describe the solution you'd like**
Some endpoint along the lines of `/api/profile_pic/by_username?url=<username>` that works with both organization and user usernames.

It could also be an endpoint like `/api/profile_info/by_username?url=<username>` which returns basic profile info fields shared across users and organizations such as name, profile pic and location (this would be the preferred functionality).

**Describe alternatives you've considered**
At the moment, I'm just fetching `/api/users/by_username?url=<username>` and if a 404 is returned, I just display a fallback image.

Note that you _can_ get organization profile pics at the moment if you have the URL of a post, since the articles endpoint returns the organization info as well as user info, but that only works if you know one of the organization's article URLs.

**Additional context**

N/A

rhymes:
I think we should add a `/profile-pictures/{username}` endpoint to the API to solve this

diogoosorio:
Seems simple enough, I can pick this one up! 👍

fdoxyz:
Thank you @diogoosorio I just assigned the issue to you

ankiiitraj:
Hello, I'm new here, this seems an easy task for me to get started. Let me know if I can work on it! 😃

rhymes:
@ankiiitraj, thanks for offering, @diogoosorio already claimed this PR but you're welcome to choose any other issue :) You can take a lok at the rest of the [hacktoberfest pool](https://github.com/forem/forem/issues?q=is%3Aissue+is%3Aopen+label%3Ahacktoberfest) or those tagged with [difficulty: easy](https://github.com/forem/forem/issues?q=is%3Aissue+is%3Aopen+label%3Ahacktoberfest+label%3A%22difficulty%3A+easy%22) as well

Thank you!

PR #10547, opened 2020-10-04
Introduce /api/profile_images/:username endpoint

diogoosorio:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR aims to introduce the operation `GET /api/profile_images/:username` to the API.

I struggled a bit with what the contract for the operation should be (as this isn't a "resource" per se), I ended up keeping it simple and re-using the structure I saw throughout the rest of the API, when dealing with profile images:

```json
{
  "type_of": "profile_image",
  "profile_image": "/uploads/organization/profile_image/be5e87e9-5b77-40ea-81e8-dd10c6d01146.png",
  "profile_image_90": "/uploads/organization/profile_image/be5e87e9-5b77-40ea-81e8-dd10c6d01146.png"
}
```

Having said that, I'm happy to refactor the contract to whatever the maintainers believe it to be the correct way to represent a profile image in this endpoint - I just ask for concrete example of what the response's payload should look like 😄 .

https://github.com/forem/forem/issues/10132

`curl "http://localhost:3000/api/profile_images/:username"`

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [x] docs.forem.com
- [ ] readme
- [ ] no documentation needed


app/controllers/api/v0/profile_images_controller.rb
```diff
@@ -0,0 +1,29 @@
+module Api
+  module V0
+    class ProfileImagesController < ApiController
+      before_action :set_cache_control_headers, only: %i[show]
```
rhymes, line 4: If you set the caching headers, you also need a way to destroy that cache, so you need the surrogate keys for the resource, which in this case it could be a user or an org.

I'm not sure it's worth to cache this endpoint though and in general I think we should lean towards caching endpoints that have either complex fetching queries or complex serialization structures.

This have neither, so I'd lean towards removing the cache altogether

app/controllers/api/v0/profile_images_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      private
+
+      def profile_image_owner
+        user || organization
+      end
+
+      def user
+        @user ||= User.select(:id, :profile_image)
+          .find_by(username: params[:username], registered: true)
```
rhymes, line 20: love that you thought about the registered users. Can I suggest making it even more explicit by using the `.registered` scope?

Something like

```ruby
User.registered.select().find_by()
```

app/views/api/v0/profile_images/show.json.jbuilder
```diff
@@ -0,0 +1,4 @@
+json.type_of "profile_image"
+
+json.profile_image Images::Profile.call(@profile_image_url, length: 640)
+json.profile_image_90 Images::Profile.call(@profile_image_url, length: 90)
```
rhymes, line 4: I read your doubts in the description but I think the serialization is simple enough to be effective, if we need more versions of the image we can add them but this closely resembles what we send back with the user or organization representations in their own partials.

I would add a key that should help the client: is this the image of a user or an org? 

Maybe:

```json
{
  "type_of": "profile_image",
  "image_of: "user"
  ...
}
```

what do you think?

rhymes (commented):
Looks great, left a couple of notes! :-)

app/views/api/v0/profile_images/show.json.jbuilder
diogoosorio: Yup I tinkered a bit with the idea of embedding the owner somehow on the response:

```json
{
  ...
  "owner": {
     "type_of": "user" | "organization",
      ...
  }
}
```

But given the use case (simply retrieving the profile image), it seemed an overkill tbh. I also tinkered with the possibility of trying to model this using [OpenAPI Links](https://swagger.io/docs/specification/links/), but I couldn't find a way to do so. 😄 

Anyway I agree and I ended up implementing your suggestion. 👍

app/controllers/api/v0/profile_images_controller.rb
diogoosorio: Yup, didn't looked into this with much attention - giving the method a quick glance I got the idea that it was just setting a 1 day cache control header (I didn't even realize that there was a cache invalidation mechanism also put in place).

Thanks for noticing this, I've removed the action. 👏

rhymes (approved):
Works well, thanks @diogoosorio!

mstruve (approved):
LGTM!

rhymes:
I fixed the conflict by upping the file version value to `0.8.2`

PR #10594, opened 2020-10-05
Update to Ruby 2.7.2

citizen428:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR updates us to Ruby 2.7.2. Deprecation warnings are now disabled by default and we can re-enable them where we think it's necessary.

> This release contains intentional incompatibility. Deprecation warnings are off by default on 2.7.2 and later. You can turn on deprecation warnings by specifying the -w or -W:deprecated option at the command-line.
https://www.ruby-lang.org/en/news/2020/10/02/ruby-2-7-2-released/

n/a

n/a

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed


citizen428:
@jdoss Requested your review here as we'll need a 2.7.2 tagged image up on quay.io/forem/ruby.

rhymes (approved):
Cool! I wonder if now that deprecation warnings are disabled by default, we should also remove https://github.com/forem/forem/blob/07c69b35ef64237310a7cd86814452862a75384a/config/environments/development.rb#L2-L3

and the same thing in `test.rb`

citizen428:
@rhymes We should probably do this! Where do you think it would make sense to explicitly enable them again? Or do you want to leave it up to individual developers to set their `RUBYOPTS` explicitly?

rhymes:
@citizen428 would you volunteer on trying to run the tests and the console locally with deprecations enabled and see how much cruft they still leave behind? maybe we've already solved them all ™ :D

rhymes:
The issue in having them disabled everywhere is that we don't know at what point of the journey we are :d

citizen428:
@rhymes For the console it looks pretty good:

```
$ RUBYOPT="-W:deprecated" rails c
/Users/michi/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/zeitwerk-2.4.0/lib/zeitwerk/loader.rb:581: warning: constant ::Data is deprecated
Loading development environment (Rails 6.0.3.3)
[1] forem(main)>
```

I can start a test run, but they take quite a while locally, I'll report back.

citizen428:
@rhymes I see some deprecation warnings around keyword parameters when running specs, but it doesn't look terrible. I wonder if we should run them with warning enabled in CI Sure, it makes the output noisier, but at least it would give us a centralized location where everyone can look for deprecations without needing to configure their own environment.

rhymes:
+1 for me

jdoss (commented):
This will have to wait until we have Ruby 2.7.2 building on [Koji](https://koji.fedoraproject.org/koji/packageinfo?packageID=125) and then I can make a `2.7.2` tag.

rhymes (approved):
LGTM!

jdoss (approved):
LGTM!

citizen428:
What should we do about Travis, do we want to enable the deprecation warnings there? A part of me wants to do it so we have a single location for tracking it, but generally writing to STDOUT does slow things down a bit and I'm not sure I want that on our CI /cc @rhymes and @mstruve

rhymes:
@citizen428 I would keep them disabled for now as we have continuous deployment on each merge and those slow downs add up.

citizen428:
@rhymes Agreed, that's partially why I asked the question again because this changed since we last talked about it.

mstruve:
@citizen428 might be a good idea to make a ticket/task for enabling that on an individual branch and then addressing the warnings. Actually might be a good external contributor task.

config/environments/test.rb
```diff
@@ -75,8 +72,6 @@
     # Supress incorrect warnings from Bullet due to included columns: https://github.com/flyerhzm/bullet/issues/147
     Bullet.add_whitelist(type: :unused_eager_loading, class_name: "Article", association: :top_comments)
     Bullet.add_whitelist(type: :unused_eager_loading, class_name: "Comment", association: :user)
-    # NOTE: @citizen428 Temporarily ignoring this while working out user - profile relationship
-    Bullet.add_whitelist(type: :n_plus_one_query, class_name: "User", association: :profile)
```
mstruve, line 79: Does this change relate to the new Ruby version or is it just an additional cleanup?

citizen428: Oops, this should be here at all, thanks for catching it.

mstruve (approved):
🚀

citizen428:
@mstruve Created an issue: https://github.com/forem/forem/issues/11282

PR #10614, opened 2020-10-05
Articles API: filtering by tags (#3654)

### LINKED ISSUE #3654: Articles API: filtering by tags

In response to #3509, we want to allow the API endpoint to filter by tags and include an 'anti-tag' list.

rafaltrojanowski:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Closes #3654

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [x] docs.forem.com
- [ ] readme
- [ ] no documentation needed


spec/services/article_api_index_service_spec.rb
```diff
@@ -0,0 +1,36 @@
+require "rails_helper"
```
rhymes, line 1: I think we'd better transfer these tests into https://github.com/forem/forem/blob/master/spec/requests/api/v0/articles_spec.rb where all the others are instead of testing the service directly. It's not wrong, it's just that all those API tests are integration tests

rhymes (changes requested):
Hi @rafaltrojanowski, thanks for taking the first stab at this. The functionality seems to be working correctly! I would add a test to explictly state that if you pass, for example, "java" in the included tags and "java" in the excluded tags, you'll get back nothing :) 

 I played a bit with it locally and it works. A few details:

- the PR lacks a description, could you at least mention that it closes https://github.com/forem/forem/issues/3654 ?
- *blocking* the PR lacks the documentation, could you please update https://github.com/forem/forem/blob/master/docs/api_v0.yml to add information about the new parameters and how they work? You can read https://docs.dev.to/contributing/api/ to know how to view that locally

ps. don't foget to name the branch `rafaltrojanowski/something` the next time ;)

spec/services/article_api_index_service_spec.rb
rafaltrojanowski: I was a little surprised that there are no service tests yet. Anyway, I moved them, as you suggested.

app/services/article_api_index_service.rb
```diff
@@ -71,6 +75,16 @@ def tag_articles
     articles.page(page).per(per_page || DEFAULT_PER_PAGE)
   end
 
+  def tagged_articles
+    articles = Article.published.includes(:user, :organization)
+    articles = articles.tagged_with(tags) if tags
```
rafaltrojanowski, line 80: I just confirmed that tags do not need to be passed as an array, thanks to:
```
ActsAsTaggableOn.default_parser.new("test1, test2,test3").parse
=> ["test1", "test2", "test3"]
```
https://github.com/mbleigh/acts-as-taggable-on/blob/950c01072b7ab6227d74dc0d0b112f0135281fc2/lib/acts_as_taggable_on/taggable/core.rb#L111

rhymes (commented):
Thanks for the change @rafaltrojanowski! 

The only thing we have to decide if `tags` should be in `AND` like they are now, or in `OR`.

See the discussion here https://github.com/forem/forem/issues/3654#issuecomment-538592335

benhalpern:
@rhymes what if we took the comma-separated version as meaning `AND` (like here), and allowed pipes (`|`) to indicate `OR` (which doesn't allow us to do more complicated boolean logic, but that's probably fine)....

spec/services/article_api_index_service_spec.rb
citizen428: @rafaltrojanowski There are quite a few specs for services in `spec/services`:

```
$ ls -1 spec/services/**/*.rb | wc -l
     115
```

citizen428:
Gut feeling I'd go with `OR` as the default, but as long as it's documented I can live with either.

benhalpern (approved):
This looks good to me. Sensible default, properly documented. ✅

citizen428 (approved):

rhymes (approved):
LGTM!

Thanks @rafaltrojanowski!

PR #10615, opened 2020-10-05
Add ability for admin to add anti-spam terms

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This adds the ability for admins to modify a list of terms which may indicate spam. But unlike past "quiet" spam indicators, this automatically creates a vomit reaction which an admin can later manually reverse or at least be aware of. In the past we have modified spam-related scores but it hasn't really worked effectively into our workflows. I think this reversible action should be how we raise spam automatically in general.

With comments I decided to limit it to newer accounts because we're examining _the whole comment_ and not just the title. But this can be modified over time. If a support admin is seeing false positives with a term they should consider removing those. We can alter the logic over time to ensure as few false-positive scenarios as possible.

### DIFF
--- a/spec/services/notifications/new_comment/send_spec.rb
+++ b/spec/services/notifications/new_comment/send_spec.rb
@@ -27,6 +27,13 @@ RSpec.describe Notifications::NewComment::Send, type: :service do
     expect(notification.json_data["user"]["username"]).to eq(child_comment.user.username)
   end
 
+  it "does not send if comment has negative score already" do
+    described_class.call(child_comment)
+
+    notification = child_comment.notifications.last
+    expect(notification).to be_nil
+  end
+
   it "creates the correct comment data for the notification" do
     described_class.call(child_comment)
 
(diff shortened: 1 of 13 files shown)


spec/services/notifications/new_comment/send_spec.rb
```diff
@@ -27,6 +27,12 @@
     expect(notification.json_data["user"]["username"]).to eq(child_comment.user.username)
   end
 
+  it "does not send if comment has negative score already" do
+    prior_notification_size = Notification.all.size
+    described_class.call(child_comment)
+    expect(Notiifcation.all.size).to eq prior_notification_size
```
Zhao-Andy, line 33: Small typo here:
```suggestion
    expect(Notification.all.size).to eq prior_notification_size
```

Zhao-Andy (approved):
Good to go once CI passes after the typo is fixed!

atsmith813 (approved):

karandpr:
Hey Ben. Thanks for your work. 
A suggestion, 
Instead of comma separated terms, can we have an array of regexes which we can set ?

Issue #10293, opened 2020-09-12
DEV.to header and footer displaying on Rails app with the same port


markmead:
**Describe the bug**

After working on Forem locally and then switching to a new Rails app on the same port, the DEV.to header and footer still show.

**To Reproduce**

1. Start Forem up
2. Close the server
3. Open a Rails app on the same server

**Expected behavior**

Not maintain cache

**Screenshots**

<img width="1425" alt="Screenshot 2020-09-12 at 11 06 11" src="https://user-images.githubusercontent.com/50486078/92993583-9f475c00-f4ea-11ea-977c-e8671e5291a8.png">

**Desktop (please complete the following information):**

- OS, version: Catalina, 10.15.6
- Browser, version: Chrome, 85

**Additional context**

- This doesn't happen on Safari
- Probably isn't limited to Rails apps on the same port

fdoxyz:
Thank you for the report @markmead. This is a known issue and I believe this must be the service worker in place that once it's installed on `localhost:3000` it has no way of knowing whether the local development server has changed.

At the moment I'm not sure if there's a way to go around this. The only solution is to refresh without cache (on macOS `CMD + SHIFT+ R`) could be an easy way to clear out the header & footer, though I haven't tried it myself yet.

I'm leaving the issue open for now to see if anyone has an idea on how to work this out.

markmead:
> Thank you for the report @markmead. This is a known issue and I believe this must be the service worker in place that once it's installed on `localhost:3000` it has no way of knowing whether the local development server has changed.
> 
> At the moment I'm not sure if there's a way to go around this. The only solution is to refresh without cache (on macOS `CMD + SHIFT+ R`) could be an easy way to clear out the header & footer, though I haven't tried it myself yet.
> 
> I'm leaving the issue open for now to see if anyone has an idea on how to work this out.

markmead:
> At the moment I'm not sure if there's a way to go around this. The only solution is to refresh without cache (on macOS `CMD + SHIFT+ R`) could be an easy way to clear out the header & footer, though I haven't tried it myself yet.

On the new app, you can unregister the service worker in the dev tools - that's a more permanent fix than hard-refreshing.

fdoxyz:
Yeah I don't think `onbeforeunload` would work because the URL would still be `localhost:3000`. That would unregister the SW anytime you're visiting external sites, unless I'm not following the idea how it could be used to unregister it.

[...]

PR #10617, opened 2020-10-05
Skip service workers in development by default

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Based on discussions and experience I think _not_ enabling service workers as the default probably makes sense, so that `localhost:3000` does not run the same service workers?

Here is a relevant conversation: https://github.com/forem/forem/issues/10293#issuecomment-691648063

This still makes it easy to have service workers on, but is a pretty straightforward way to turn them off. `ENV` var seems like a decent way to do this, eh?

This still registers service workers, just makes the file empty so no functionality exists. Neither scenario is _perfect_ because either way the user is either dealing with confusion or inconsistency but at least this enables options.

### DIFF
--- a/.env_sample
+++ b/.env_sample
@@ -1,9 +1,11 @@
+# Development options
+export SKIP_SERVICEWORKERS="true"
+
 # App core values
 export APP_DOMAIN="localhost:3000"
 export APP_PROTOCOL="http://"
 export APP_NAME="forem_local"
 export FOREM_OWNER_SECRET="secret"
-
 
 # Openresty domain + Protocol setting for development
 export OPENRESTY_DOMAIN=""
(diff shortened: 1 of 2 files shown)


atsmith813 (approved):
👍 

My only hesitation about doing this is forgetting this is set or having weird behavior where a feature in development works/looks right without service workers, then acts weird with them re-enabled. Maybe we document this switch for clarity? Or add some sort of notification/reminder that the user has it turned off?

vaidehijoshi (approved):
Echoing @atsmith813's comment to say that I think we should add some documentation around this. Might also be nice to include a line to point out when you'd want to run the app _with_ serviceworkers enabled in development explicitly, too.

citizen428 (approved):

PR #10623, opened 2020-10-05
Do not require SiteConfig.tagline

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We have a concept of a required `tagline` as part of our `SiteConfig`, but in _most_ places, we have a fallback check to only render the `tagline` if it exists:
```ruby
<% if SiteConfig.tagline.present? %>
  <%= SiteConfig.tagline %>
<% else %>
  Some other text here!
<% end %>
```

This PR makes `SiteConfig.tagline` _not_ required, and fixes any scenarios where we _don't_ have a `SiteConfig.tagline.present?` check. This change will be super useful as we spin up new Forem instances and set up Forem creator onboarding, as it will be _one less thing_ that a Forem creator will need to set up 😸 

Additionally, this PR:
- [x] Fixes some typos
- [x] Fixes some `.erb` formatting
- [x] Moves `tagline` higher up in the `Community Content` section of the SiteConfig view

Closes https://github.com/forem/InternalProjectPlanning/issues/90.

This PR removes the formerly required `tagline` input from the "required" set of SiteConfig options...
Screen_Shot_2020-10-05_at_2_38_55_PM

...and instead renders it in the `Community Content` section of the SiteConfig view:
Screen_Shot_2020-10-05_at_2_39_48_PM

## Added tests?

- [x] yes _(improved the preexisting test! ❤️)_
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Nope! All Forem instances should continue to work as-is.

girls require gif

### DIFF
--- a/app/controllers/concerns/verify_setup_completed.rb
+++ b/app/controllers/concerns/verify_setup_completed.rb
@@ -7,7 +7,6 @@ module VerifySetupCompleted
     community_name
     community_description
     community_action
-    tagline
 
     main_social_image
     logo_png
--- a/app/views/admin/configs/show.html.erb
+++ b/app/views/admin/configs/show.html.erb
@@ -24,8 +24,8 @@
       <div id="getStartedBodyContainer" class="<%= VerifySetupCompleted.setup_completed? ? "hide collapse" : "show expand" %> p-3" aria-labelledby="siteConfigHeader">
         <%= form_for(SiteConfig.new, url: admin_config_path) do |f| %>
 
-          <!-- we need to list the config as separate fields if the data structure is a Hash -->
           <% VerifySetupCompleted::MANDATORY_CONFIGS.each do |config_key| %>
+            <%# we need to list the config as separate fields if the data structure is a Hash %>
             <% if SiteConfig.public_send(config_key).is_a?(Hash) %>
               <%= admin_config_label config_key.to_s %>
               <div class="form-group">
--- a/app/views/admin/configs/show.html.erb
+++ b/app/views/admin/configs/show.html.erb
@@ -24,8 +24,8 @@
       <div id="getStartedBodyContainer" class="<%= VerifySetupCompleted.setup_completed? ? "hide collapse" : "show expand" %> p-3" aria-labelledby="siteConfigHeader">
         <%= form_for(SiteConfig.new, url: admin_config_path) do |f| %>
 
+          <!-- we need to list the config as separate fields if the data structure is a Hash -->
           <% VerifySetupCompleted::MANDATORY_CONFIGS.each do |config_key| %>
-            <!-- we need to list the config as separate fields if the data structure is a Hash -->
             <% if SiteConfig.public_send(config_key).is_a?(Hash) %>
               <%= admin_config_label config_key.to_s %>
               <div class="form-group">
@@ -39,7 +39,7 @@
                                              placeholder: Constants::SiteConfig::DETAILS[config_key][:placeholder] %>
                       <div class="alert alert-info"><%= "#{key} #{config_key.to_s.tr! '_', ' '}" %></div>
                     </div>
-                    <% end %>
+                  <% end %>
                 <% end %>
               </div>
             <% elsif SiteConfig.public_send(config_key).is_a?(Array) %>
@@ -64,12 +64,12 @@
           <% end %>
 
           <% if current_user.has_role?(:single_resource_admin, Config) %>
-                <div class="form-group">
-                  <%= label_tag "Please type out the sentence below to submit:" %>
-                  <%= text_field_tag :confirmation, nil, class: "form-control", placeholder: "Confirmation text", autocomplete: "off", required: true, pattern: @confirmation_text %>
-                  <div class="alert alert-info">Type the sentence: <%= @confirmation_text %></div>
-                  <%= f.submit "Update Site Configuration", class: "btn btn-danger btn-lg" %>
-                </div>
+            <div class="form-group">
+              <%= label_tag "Please type out the sentence below to submit:" %>
+              <%= text_field_tag :confirmation, nil, class: "form-control", placeholder: "Confirmation text", autocomplete: "off", required: true, pattern: @confirmation_text %>
+              <div class="alert alert-info">Type the sentence: <%= @confirmation_text %></div>
+              <%= f.submit "Update Site Configuration", class: "btn btn-danger btn-lg" %>
+            </div>
           <% end %>
         <% end %>
       </div>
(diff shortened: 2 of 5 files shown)


app/views/admin/configs/show.html.erb
```diff
@@ -24,8 +24,8 @@
       <div id="getStartedBodyContainer" class="<%= VerifySetupCompleted.setup_completed? ? "hide collapse" : "show expand" %> p-3" aria-labelledby="siteConfigHeader">
         <%= form_for(SiteConfig.new, url: admin_config_path) do |f| %>
 
+          <!-- we need to list the config as separate fields if the data structure is a Hash -->
```
vaidehijoshi, line 27: Moved this comment up because it was rendering every time we iterated over a config, which made it hard to read the DOM 😕

app/views/admin/configs/show.html.erb
```diff
@@ -87,910 +87,909 @@
...
+                      <img alt="Campaign sidebar image" class="img-fluid" src="<%= SiteConfig.campaign_sidebar_image %>" />
+                    </div>
+                  </div>
+                <% end %>
+                <%= f.text_field :campaign_sidebar_image,
+                                 class: "form-control",
+                                 value: SiteConfig.campaign_sidebar_image,
+                                 placeholder: Constants::SiteConfig::DETAILS[:campaign_sidebar_image][:placeholder] %>
+                <div class="alert alert-info"><%= Constants::SiteConfig::DETAILS[:campaign_sidebar_image][:description] %></div>
+              </div>
```
vaidehijoshi, line 310: As it turns out, `tagline` was rendered _twice_: once in the "required" section of configs, and once in here, in the "community content" section. I just moved it up so that it renders below the community name, instead of above the copyright year, because that makes a bit more sense, IMO :)

vaidehijoshi:
Most of this PR's line diffs is because of me fixing incorrect indentation. I recommend reviewing it by hiding whitespace changes like so:
Screen Shot 2020-10-05 at 3 16 03 PM

atsmith813 (approved):
🚀

app/controllers/concerns/verify_setup_completed.rb
```diff
@@ -7,7 +7,6 @@ module VerifySetupCompleted
     community_name
     community_description
     community_action
-    tagline
```
citizen428, line 10: :fire:

app/views/admin/configs/show.html.erb
citizen428: **suggestion (non-blocking):** How about we leave it where it is (because it really relates to the `if` statement), but turn it into an ERb comment that doesn't show up in the generated HTML?  

```ruby
<% VerifySetupCompleted::MANDATORY_CONFIGS.each do |config_key| %>
  <%# we need to list the config as separate fields if the data structure is a Hash %>
  <% if SiteConfig.public_send(config_key).is_a?(Hash) %>
   ...
<% end %>
```

app/views/devise/invitations/edit.html.erb
```diff
@@ -2,7 +2,9 @@
 <div class="crayons-layout crayons-layout--limited mb-10" style="max-width:640px">
   <div class="crayons-notice mt-4">
     <h1>Welcome to <%= community_name %></h1>
-    <h3><%= SiteConfig.tagline %></h3>
+    <% if SiteConfig.tagline.present? %>
```
citizen428, line 5: **suggestion (non-blocking):** Including this addition, I think we have this conditional in 5 places now. maybe we could move it into the `ApplicationHelper`?

```ruby
def tagline
  return if SiteConfig.tagline.blank?

  content_tag(:h3, SiteConfig.tagline)
end
```

and then use 

```ruby
<%= tagline %>
```

ERb deals gracefully with `nil`s:

```ruby
template = ERB.new("<%= nil %>")
template.result
#=> ""
```

(The result for the `if` version would be `" \n"`)

spec/views/layouts/signup_modal.html.erb_spec.rb
```diff
@@ -1,8 +1,17 @@
...
+  it "renders the tagline if it is set" do
+    allow(SiteConfig).to receive(:tagline).and_return(tagline_text)
     render
     expect(rendered).to have_text(SiteConfig.tagline)
   end
+
+  it "does not render the tagline if it is not set" do
+    allow(SiteConfig).to receive(:tagline).and_return(nil)
+    render
+    expect(rendered).to have_no_text(tagline_text)
```
citizen428, line 15: **nitpick (non-blocking):** We use the `expect().not_to have_text()` form 8 times throughout our tests, but we never use `to have_no_text`. RSpec is a bit smarter than Minitest here and the two forms are equivalent (example uses `have_xpath`, but same consideration applies):

> Be aware that because of this behaviour, the following two statements are not equivalent, and you should always use the latter!
> 
> !page.has_xpath?('a')
> page.has_no_xpath?('a')
> 
> The former would immediately fail because the content has not yet been removed. Only the latter would wait for the asynchronous process to remove the content from the page.
> 
> Capybara's RSpec matchers, however, are smart enough to handle either form. The two following statements are functionally equivalent:
> 
> expect(page).not_to have_xpath('a')
> expect(page).to have_no_xpath('a')

https://www.rubydoc.info/gems/capybara/3.3.1

app/views/admin/configs/show.html.erb
vaidehijoshi: @citizen428 that's a great suggestion and compromise -- i like it!

spec/views/layouts/signup_modal.html.erb_spec.rb
vaidehijoshi: @citizen428 that's fair! i think i only used `to have_no_text` because rspec suggested that, but i agree with you that we ought to keep it consistent. thanks for pointing that out, i'll update the spec to use this the more consistent method here 👍

app/views/devise/invitations/edit.html.erb
vaidehijoshi: Hm, I might end up doing this in a separate PR since I'll be making some more changes in this portion of the codebase to handle https://github.com/forem/InternalProjectPlanning/issues/87 and https://github.com/forem/InternalProjectPlanning/issues/89...

Since this is a non-blocking suggestions, I'm going to come back to cleaning this (and maybe some more conditional?) up when I finish up those other two issues :)

citizen428: That sounds perfect, totally just a suggestion and it can most definitely wait/be done in a different way/ignored.

Ridhwana (approved):

juliannatetreault (approved):
👍 👍

Issue #9585, opened 2020-07-31
Keyboard access and usability in Onboarding screens


marcysutton:
**Describe the bug**

In the onboarding flow, there's a series of screens welcoming users and prompting them to accept the Code of Conduct, agree to T&Cs, follow tags, add some bio information, and follow users. A lot of these user interface controls are reachable for accessibility due to the use of buttons and real form controls, which is excellent! But they could still use some improvements.

The challenges I ran into were related to navigating with the keyboard alone. I couldn't complete some actions, and had to do unconventional navigating to complete the rest. Some improvements to markup and semantics will make them more usable in screen readers, as well.

For example: 

1. I can't reach the Code of Conduct link or the Terms and Conditions links from the checkboxes–the links inside the labels are not reachable for some reason, and thus can't be opened before agreeing to the terms if you can't use a mouse.

2. From the second screen onward, the action buttons to go to the next screen are above the onboarding actions (such as toggling specific tag buttons), so keyboard users have to go backwards in the tab order to advance. It's pretty unintuitive from a keyboard access perspective, and would be very unusable to someone who can't see the screen.

3. The same tag follow buttons don't have any semantic information marking them as selected, and they are nested in DIV elements so screen reader users won't know how many tag buttons there are, nor when they are or aren't selected.

3. When you do click "Continue" or "Skip for Now", the focus point is put at the end of the next screen as well–so to use a keyboard, you have to go backwards in the tab order again. Ideally, focus would be put at the start of the items for a more ergonomic keyboard experience.

**To Reproduce**

1. Create a new account on Dev, for example by authenticating with GitHub
2. Go through the onboarding flow with the keyboard alone (no cheating with the mouse!)
3. Try to open the Code of Conduct link from the keyboard.
4. Follow some tags with the keyboard using TAB and the Space or Enter keys
5. Try to reach the Continue button. You'll find you have to use Shift + TAB to reach it, which is not ideal.
6. Go to the next screen and try to fill in the profile fields. You'll find you're at the end of the tab order and again have to go backwards.

**Expected behavior**

All interactive items should be reachable and operable from the keyboard, in an intuitive order.

The Continue/Skip for now buttons should come after the UI controls they depend on in the source order AND the tab order. A design update to put those buttons at the bottom would be the best solution, as CSS positioning and tabIndex hacks are not recommended.

Toggle buttons should follow a pattern indicating their state for assistive technologies: https://inclusive-components.design/toggle-button/

[...]

Issue #10610, opened 2020-10-05
a11y: improve keyboard access on modals


lisasy:
# Problem

When I'm looking at a modal and tab through with a keyboard, it goes through everything behind the modal overlay first, making it a really frustrating user experience. 

# Proposed solution

When a modal is open and a user tabs through with their keyboard, have them navigate through the modal first, not what is behind the modal.

Video here: https://share.getcloudapp.com/6quPkxyr

Monty7:
Hello, I would like to claim this issue.

lisasy:
@Monty7 You can go ahead and claim this :)

Happy hacking!

nickytonline:
@Monty7, I've assigned the issue to you.

Link2Twenty:
I'd suggest looking at the [inert polyfill](https://github.com/WICG/inert).

Rafi993:
<img src="https://user-images.githubusercontent.com/8092120/87239167-91615600-c429-11ea-917c-9f02bec477f4.png" width="16" height="16"/> Forwarded from issue https://github.com/thepracticaldev/dev.to/issues/10660

> I think it is related to https://github.com/forem/forem/issues/9585

mstruve:
@Monty7 do you still have plans to work on this? If not we would like to get it prioritized for our internal team to get done.

Rafi993:
There is something similar done by PR https://github.com/forem/forem/pull/12366

PR #10661, opened 2020-10-06
Improving usability on onboarding screen when using keyboard and screen reader

Rafi993:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Improving usability when using keyboard and screen reader

Closes https://github.com/forem/forem/issues/9585

1. Go to url http://localhost:3000/onboarding
2. Try to complete the on-boarding journey with keyboard (with screen reader on)

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Happy


app/assets/stylesheets/preact/onboarding-modal.scss
```diff
@@ -246,7 +246,7 @@
 }
 // modal overflow scrolling
 .onboarding-modal-scroll-container {
-  height: calc(100vh - 289px);
+  height: calc(100vh - 332px);
```
Rafi993, line 249: Since it was causing two scrollbars I changed it to so that it has only one scrollbar

app/javascript/onboarding/components/FollowTags.jsx
```diff
@@ -118,9 +118,9 @@ class FollowTags extends Component {
               <h2 className="subtitle">Follow tags to customize your feed</h2>
             </header>
             <div className="onboarding-modal-scroll-container">
-              <div data-testid="onboarding-tags" className="onboarding-tags">
+              <ul data-testid="onboarding-tags" className="onboarding-tags">
```
Rafi993, line 121: I changed div to ul and li so that I could be read by screen reader

app/javascript/shared/components/focustrap.jsx
```diff
@@ -0,0 +1,30 @@
+import PropTypes from 'prop-types';
+import { h, Fragment } from 'preact';
+import { useEffect, useRef } from 'preact/hooks';
+import { createFocusTrap } from 'focus-trap';
+import { defaultChildrenPropTypes } from '../../common-prop-types';
+
+const FocusTrap = ({ selector, children }) => {
```
Rafi993, line 7: I created this wrapper component that uses library 'focus-trap' to trap focus. Ideally I love to  convert it to a hook rather than a component but since it is used by class components too it can't be a hook.

app/javascript/shared/components/focustrap.jsx
```diff
@@ -0,0 +1,30 @@
+import PropTypes from 'prop-types';
+import { h, Fragment } from 'preact';
+import { useEffect, useRef } from 'preact/hooks';
+import { createFocusTrap } from 'focus-trap';
```
Rafi993, line 4: I added this library https://bundlephobia.com/result?p=focus-trap@6.1.2  It's size is **2.3kb** minified and gzipped is this ok?

app/javascript/shared/components/focustrap.jsx
```diff
@@ -0,0 +1,30 @@
...
+    return () => {
+      focusTrap.current.deactivate();
+    };
+  });
+
+  return <Fragment>{children}</Fragment>;
+};
+
+FocusTrap.defaultProps = {
+  selector: '.crayons-modal',
```
Rafi993, line 22: I made the default selector value `'.crayon-modal'` since this component will be mostly be used by modals (rarely in other places where different selector string could be passed)

app/javascript/onboarding/components/FollowTags.jsx
Rafi993: I changed div to ul and li so that I could be read by screen reader

app/javascript/shared/components/focusTrap.jsx
```diff
@@ -0,0 +1,30 @@
+import PropTypes from 'prop-types';
+import { h, Fragment } from 'preact';
+import { useEffect, useRef } from 'preact/hooks';
+import { createFocusTrap } from 'focus-trap';
+import { defaultChildrenPropTypes } from '../../common-prop-types';
+
+const FocusTrap = ({ selector, children }) => {
```
Rafi993, line 7: I created this wrapper component that uses library 'focus-trap' to trap focus. Ideally I love to convert it to a hook rather than a component but since it is used by class components too it can't be a hook.

app/javascript/shared/components/focusTrap.jsx
```diff
@@ -0,0 +1,30 @@
+import PropTypes from 'prop-types';
+import { h, Fragment } from 'preact';
+import { useEffect, useRef } from 'preact/hooks';
+import { createFocusTrap } from 'focus-trap';
```
Rafi993, line 4: I added this library https://bundlephobia.com/result?p=focus-trap@6.1.2 It's size is 2.3kb minified and gzipped is this ok?

app/javascript/shared/components/focusTrap.jsx
```diff
@@ -0,0 +1,30 @@
...
+    return () => {
+      focusTrap.current.deactivate();
+    };
+  });
+
+  return <Fragment>{children}</Fragment>;
+};
+
+FocusTrap.defaultProps = {
+  selector: '.crayons-modal',
```
Rafi993, line 22: I made the default selector value '.crayon-modal' since this component will be mostly be used by modals (rarely in other places where different selector string could be passed)

app/javascript/onboarding/components/FollowTags.jsx
```diff
@@ -100,84 +101,97 @@ class FollowTags extends Component {
...
+              <header className="onboarding-content-header">
+                <h1 id="title" className="title">
+                  What are you interested in?
+                </h1>
+                <h2 id="subtitle" className="subtitle">
+                  Follow tags to customize your feed
+                </h2>
+              </header>
+              <div className="onboarding-modal-scroll-container">
+                <ul data-testid="onboarding-tags" className="onboarding-tags">
```
Rafi993, line 132: I changed div to ul and li so that I could be read by screen reader

rhymes (commented):
Haven't reviewed the code but I wonder if we can improve the behavior in the third slide

I have to tab 3 times before I can press space to select the "follow" button:

on1

It seems to work nice otherwise. There are a few violations found in the Accessibility panel in Firefox but I think they are beyond the objective of this PR.

Thanks @Rafi993!

app/javascript/shared/components/focusTrap.jsx
Link2Twenty: If you open a second modal over the top of an existing one I think the tabbing will be locked to the first.
That being said I believe, with how you've set it up, you can some something like

```jsx
<FocusTrap selector={reference.current}>
  <div id="locked-in" ref={reference}></div>
</FocusTrap>
```

Link2Twenty:
FocusTrap would be useful for #10610 too.

app/javascript/shared/components/focusTrap.jsx
Link2Twenty, line 7: Is it worth having a `bool` active prop?
Meaning in the future we could add the ability to lock focus if we want to,

Rafi993: I think boolean `active` can be added to **FocusTrap** component and made optional default value being `true`. But right now it is not used here but it could be added later to the component if needed without affecting existing code.

app/javascript/shared/components/focusTrap.jsx
Rafi993: Ref would work too. Incase modal appears another modal it can pass another selector string like id of the modal and it would work.

```
<FocusTrap selector="#secondModal">
  <div>
    second modal
  </div>
</FocusTrap>
```
But on the other hand nested modals might not be used anywhere (but I could be wrong). Having no prop needed to be passed for this component by default makes API much simpler.

Rafi993:
> Haven't reviewed the code but I wonder if we can improve the behavior in the third slide
> 
> I have to tab 3 times before I can press space to select the "follow" button:
> 
> on1
> 
> It seems to work nice otherwise. There are a few violations found in the Accessibility panel in Firefox but I think they are beyond the objective of this PR.
> 
> Thanks @Rafi993!

When "follow" button is selected focus is jumping to the previous element causing this issue. Let me search for reason it is happening.

Rafi993:
> Haven't reviewed the code but I wonder if we can improve the behavior in the third slide
> 
> I have to tab 3 times before I can press space to select the "follow" button:
> 
> on1
> 
> It seems to work nice otherwise. There are a few violations found in the Accessibility panel in Firefox but I think they are beyond the objective of this PR.
> 
> Thanks @Rafi993!

Similar issue is also happening in tag selection too. I couldn't figure out why it is happening. Any thoughts @forem/oss

tags

Thanks for you help

app/javascript/onboarding/components/FollowUsers.jsx
```diff
@@ -140,72 +141,86 @@ class FollowUsers extends Component {
...
+                </h2>
+              </header>
+
+              <ul
+                data-testid="onboarding-users"
+                className="onboarding-modal-scroll-container"
+              >
+                {users.map((user) => (
+                  <li key={user.name}>
+                    <button
```
Link2Twenty, line 178: I think this button should have tabIndex="-1".
I think we only want the follow button to be in the tab order.

Link2Twenty: Here is the flow with the tabindex -1

forem__outer-button

only 1 tab to get to the next follow button

Rafi993: Thank you @Link2Twenty. But there is one issue though the screen reader is not reading out items in the list so the user (the name of the user and the description below now) it just says follow. This might cause confusion for users who are using screen readers since they don't know which user did they just followed. I can add **aria-label** to the follow button saying 
**"Follow Jere Thompson"** it will help but it will still miss out the description but I don't know if that is ok.

Rafi993: When I remove the handler ```this.handleClick``` it is not happening I did try adding both ```event.preventDefault()``` and ```event.stopPropagation()``` they don't seem to affect this. It is something to do with the state change. Let me keep searching.

Rafi993: It is a bug in the ```FocusTrap``` component that is causing this when I remove it is working properly. Let me keep searching further.

Rafi993: I was able to fix this by moving FocusTrap component one level up to **app/javascript/onboarding/Onboarding.jsx** there are some tests failing (even though what the test is looking for is working when tested manually) let me fix those too.

Rafi993: I have fixed the breaking tests too

Rafi993:
> Haven't reviewed the code but I wonder if we can improve the behavior in the third slide
> 
> I have to tab 3 times before I can press space to select the "follow" button:
> 
> on1
> 
> It seems to work nice otherwise. There are a few violations found in the Accessibility panel in Firefox but I think they are beyond the objective of this PR.
> 
> Thanks @Rafi993!

I have resolved this now

tab_key

Rafi993:
I think some unrelated test failed.

rhymes:
@Rafi993 the tests are fine, it's Codeclimate complaining :) We should likely ignore it in the context of this PR

rhymes (approved):
It looks good, thanks @Rafi993, only a question still on the "follow people slide". Is it possible to go down from "2 tabs + space to select follow" to "1 tab plus space"?

Rafi993:
> It looks good, thanks @Rafi993, only a question still on the "follow people slide". Is it possible to go down from "2 tabs + space to select follow" to "1 tab plus space"?

Solution @Link2Twenty suggested above does it https://github.com/forem/forem/pull/10661#discussion_r501755695 . But when I was using screen reader it just read out the text in the "Follow" button alone and it did not read the user name or the description below.

app/javascript/onboarding/components/FollowUsers.jsx
```diff
@@ -140,72 +141,86 @@ class FollowUsers extends Component {
...
+                        <h4 className="user-name">{user.name}</h4>
+                        <p className="user-summary">
+                          {he.unescape(user.summary || '')}
+                        </p>
+                      </div>
+                      <button
+                        data-testid="onboarding-user-following-status"
+                        type="button"
+                        className="user-following-status"
+                        aria-pressed={selectedUsers.includes(user)}
```
Link2Twenty: Here would could add 

```jsx
aria-label={`${selectedUsers.includes(user) ? 'Unfollow' : 'Follow'} ${user.name}`}
```

This will tell the screen reader what to read.

Link2Twenty: This would be to coincide with the `tabIndex` change.

Rafi993: > Here would could add
> 
> ```js-jsx
> aria-label={`${selectedUsers.includes(user) ? 'Unfollow' : 'Follow'} ${user.name}`}
> ```
> 
> This will tell the screen reader what to read.

 But it would miss the description for each user

Link2Twenty: I guess the three options are

- We remove something from the tabIndex be it the inner or outer button
- We accept multiple tabs is a thing that people have to deal with
- We wait for #10468 and use the useListNavigation hook to add j and k navigation support.

nickytonline:
I'm looking to give this a review this week. Hacktoberfest was pretty busy. 😅

app/javascript/onboarding/components/FollowUsers.jsx
marcysutton: To answer your comment regarding the button name, @Rafi993, an `aria-label` would help to uniquely identify the individual follow buttons. That would solve an issue that exists throughout the site where controls are repeated (i.e. "Links with the same name have a similar purpose" in the axe extension).

Rafi993: Thank you @marcysutton I have made the change. I did not know about this before.

package.json
```diff
@@ -155,7 +155,7 @@
     "@honeybadger-io/webpack": "^1.2.0",
     "@rails/webpacker": "5.2.1",
     "@testing-library/user-event": "^12.1.7",
-    "ahoy.js": "^0.3.7",
+    "ahoy.js": "^0.3.8",
```
nickytonline, line 158: refactor (blocking): This looks like a bad merge or maybe ensure you have merged latest as this does not appear to be related to the work being done in this PR.

yarn.lock
```diff
@@ -18039,10 +18039,10 @@ tabbable@^5.1.1:
   resolved "https://registry.yarnpkg.com/tabbable/-/tabbable-5.1.1.tgz#33d6950a615bc78c6eb6828f3953e66544817dc4"
   integrity sha512-qjkrV7I29Lkg///jPQnq26+26q9wQOKTEcckrwPejTsxxl2zdK60x0XmiEELCEParLhgbVQRflyKkpr+K/SmCQ==
 
-table@^5.2.3:
-  version "5.4.6"
-  resolved "https://registry.yarnpkg.com/table/-/table-5.4.6.tgz#1292d19500ce3f86053b05f0e8e7e4a3bb21079e"
-  integrity sha512-wmEc8m4fjnob4gt5riFRtTu/6+4rSe12TpAELNSqHMfF3IqnA+CH37USM6/YR3qRZv7e56kAEAtd6nKZaxe0Ug==
+table@^6.0.4:
```
nickytonline, line 18042: refactor (blocking): This looks like a bad merge or maybe ensure you have merged latest as this does not appear to be related to the work being done in this PR.

nickytonline (changes requested):

Rafi993:
As we discussed on the stream it is better to split this PR into multiple small PR's so that it will be easy to review. I'll start creating separate small PR's.

Rafi993:
Closing this PR in favour of smaller PRs that have been raised.

PR #10662, opened 2020-10-06
Prepare the database for the configurable navigation items 

Ridhwana:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR tackles the following subtasks related to providing admins with a configurable navbar:

1. Adding a field called order that allows an admin to decide on the order he'd like the navigation items to follow ( A future iteration can possibly be a drag and drop)
2. Renames a column more applicably 
3. Creates a data script to seed some default navigation items for all forems on creation
4. Creates a temporary data script that will allow DEV and existing forems like thismma and forem.dev to seed their navigational items.

Follow up PR of on https://github.com/forem/forem/pull/9444
https://github.com/forem/forem/issues/4661
https://github.com/forem/forem/issues/5475

1. Run rake db:migrate 
2. Update the data scripts using `rails data_updates:run`. You should now see 5 items in the database using `NavigationLink.all.length'
3. Run the rake task using `rake temporary:navigation_links:update_links`. You should now see 13 items in the database using `NavigationLink.all.length'
4. You should also be able to add, edit and delete any link from http://localhost:3000/admin/navigation_links

Also triple check urls and names compared to the current site.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

1. Need to ensure that data script runs and check blazer if the NavigationLinks table is populated with 5 items
2. Run the rake task `rake temporary:navigation_links:update_links` and check that the NavigationLinks table is populated with 13 items

alt_text


db/migrate/20201004160728_add_order_to_navigation_links.rb
```diff
@@ -0,0 +1,5 @@
+class AddOrderToNavigationLinks < ActiveRecord::Migration[6.0]
+  def change
+    add_column :navigation_links, :order, :integer
```
Ridhwana, line 3: Added a column called order that will allow the admin to save the order in which he'd like to display the nav items

db/migrate/20201005162540_add_unique_index_to_navigation_links_order.rb
```diff
@@ -0,0 +1,7 @@
+class AddUniqueIndexToNavigationLinksOrder < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :navigation_links, :order, unique: true, algorithm: :concurrently
```
Ridhwana, line 5: Added an index on order so that the admin cant use the same integer twice.

db/migrate/20201005181510_change_display_when_signed_in_column_name.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeDisplayWhenSignedInColumnName < ActiveRecord::Migration[6.0]
+  def change
+    safety_assured { rename_column :navigation_links, :display_when_signed_in, :display_only_when_signed_in }
```
Ridhwana, line 3: I felt like this column name didn't make any sense and so I've renamed it. Since this is a brand new feature we dont have datd for it as yet. I haven't created a frontend as yet.

lib/data_update_scripts/20201001173841_add_navigation_links.rb
```diff
@@ -0,0 +1,26 @@
+module DataUpdateScripts
+  class AddNavigationLinks
```
Ridhwana, line 2: This data update script will seed all forems with some base links. The base links are ReadingList, Code of Conduct, Privacy Policy, Terms of Use, and Contact. If the team feels any should be removed or added, drop a note here.

lib/tasks/temporary/add_dev_navigation_links.rb
```diff
@@ -0,0 +1,57 @@
+# rubocop:disable Metrics/BlockLength
+namespace :temporary do
```
Ridhwana, line 2: This temporary rake task will get existing forems like DEV,  thismmalife and forem.dev to the same point at which they currently are with 13 existing links on their sidebar.

lib/tasks/temporary/add_dev_navigation_links.rake
```diff
@@ -0,0 +1,57 @@
+# rubocop:disable Metrics/BlockLength
+namespace :temporary do
```
Ridhwana, line 2: This temporary rake task will get existing forems like DEV, thismmalife and forem.dev to the same point at which they currently are with 13 existing links on their sidebar.

Ridhwana: We'll remove it once we run this

benhalpern (approved):

lib/data_update_scripts/20201001173841_add_navigation_links.rb
```diff
@@ -0,0 +1,26 @@
...
+    PROTOCOL = ApplicationConfig["APP_PROTOCOL"].freeze
+    DOMAIN = Rails.application&.initialized? ? SiteConfig.app_domain : ApplicationConfig["APP_DOMAIN"]
+    BASE_URL = "#{PROTOCOL}#{DOMAIN}".freeze
+
+    READING_ICON = File.read(Rails.root.join("app/assets/images/twemoji/drawer.svg")).freeze
```
benhalpern, line 7: Thinking outloud here..... I think it's great that we're giving individual forems a lot of control over this section. There's an argument for whether we want to be more restrictive for consistency _across_ forems, but I think this is the right baseline here: Focus on the ways Forems can independently succeed through flexibility in places like this, but in the future I could see a scenario where users of forems could customize this space such that they always see reading list in the same place, or whatever.... This would be in what we now have as `/settings/ux` And that they could port that customization across all their forems to automatically inject their preferences.

Not a blocker, just a thought.

citizen428: I feel like even the possibility of a scenario as portable preferences should be documented somewhere (even if it's as an aspirational item in some sort of backlog) and not get lost in a Github comment. The clarity/extensibility of code often relies on the soundness of the data model and shoehorning new features into existing structures can be very painful. I know you're not the biggest fan of big up front designs, and I'm not advocating for that, but I do have concerns over the long-term quality of the codebase if things stay too ad-hoc.

db/migrate/20201004160728_add_order_to_navigation_links.rb
citizen428: **question (non-blocking):** Can this column be null? If not, you probably should add the constraint. If yes, maybe we should extend the `order` statement in some way (e.g. `order(order: asc, name: :asc)`) to get an explicit consistent ordering? We could potentially also use a proc default to automatically populate this column, e.g.

```ruby
 default: -> { NavigationLink.maximum(:position).next 
```

There should always be few enough rows in this table that `maximum` doesn't become a bottleneck.

app/controllers/admin/navigation_links_controller.rb
```diff
@@ -1,12 +1,12 @@
...
   class NavigationLinksController < Admin::ApplicationController
     ALLOWED_PARAMS = %i[
-      name url icon display_when_signed_in
+      name url icon display_only_when_signed_in order
     ].freeze
     layout "admin"
 
     def index
-      @navigation_links = NavigationLink.order(:name)
+      @navigation_links = NavigationLink.order(order: :asc)
```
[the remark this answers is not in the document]
Ridhwana: Done ✅ This makes more sense

db/migrate/20201004160728_add_order_to_navigation_links.rb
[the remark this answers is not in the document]
Ridhwana: ✅

benhalpern (approved):

citizen428 (approved):

PR #10678, opened 2020-10-06
Remove community_action from SiteConfig

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As part of the Creator Onboarding project, we are removing fields from the SiteConfig that are too DEV-specific and/or that we don't want all Forem creators to have to fill out when they set up a new Forem instance.

Since `community_action` isn't used in too many places (and is currently fairly pretty DEV-specific), we think that it makes more sense to drop this field and update the few instances of where it is used in emails instead.

Fixes https://github.com/forem/InternalProjectPlanning/issues/89.

Before this change:
Screen Shot 2020-10-06 at 11 10 50 AM

After this change:
Screen Shot 2020-10-06 at 11 08 29 AM

## Added tests?

- [ ] yes
- [x] no, ~because they aren't needed~ but i updated the preexisting ones
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Since we are just dropping this field, we shouldn't need to take any major action here.

dropping cardboard on a cat's face

### DIFF
--- a/app/views/mailers/digest_mailer/digest_email.html.erb
+++ b/app/views/mailers/digest_mailer/digest_email.html.erb
@@ -21,12 +21,12 @@
   <em>
     <% if @user.experience_level == nil %>
       <b>
-        You can now add your <%= SiteConfig.community_action %> experience level to your account in order to receive more relevant content.
+        You can now add your experience level to your account in order to receive more relevant content.
       </b>
       <br /><br />
       Visit <b><a href="<%= app_url(user_settings_path(:misc)) %>" style="text-decoration: none;">your settings</a></b> to enter your experience level on a scale of 1-10.
       <br /><br />
-      You can change it any time in the future. Happy <%= SiteConfig.community_action %>!
+      You can change it any time in the future.
     <% elsif @user.following_users_count == 0 %>
       <b>
         <%= community_name %> Digest is a new periodic email featuring the best posts from our community of <%= community_members_label %>.
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -141,7 +141,6 @@ module Admin
         community_name
         community_description
         community_member_label
-        community_action
         community_copyright_start_year
         staff_user_id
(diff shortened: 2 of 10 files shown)


app/views/mailers/digest_mailer/digest_email.html.erb
```diff
@@ -21,12 +21,12 @@
   <em>
     <% if @user.experience_level == nil %>
       <b>
-        You can now add your <%= SiteConfig.community_action %> experience level to your account in order to receive more relevant content.
+        You can now add your experience level to your account in order to receive more relevant content.
```
vaidehijoshi, line 24: Opted to just remove `community_action` here and not replace it with anything since I think "experience level" is self-explanatory (it exists as a field on a user's profile).

app/views/mailers/digest_mailer/digest_email.html.erb
```diff
@@ -21,12 +21,12 @@
...
       </b>
       <br /><br />
       Visit <b><a href="<%= app_url(user_settings_path(:misc)) %>" style="text-decoration: none;">your settings</a></b> to enter your experience level on a scale of 1-10.
       <br /><br />
-      You can change it any time in the future. Happy <%= SiteConfig.community_action %>!
```
vaidehijoshi, line 29: Removed this exclamation here, and opted not to replace it with anything specific here.

juliannatetreault (approved):
Looks good to me! ✨

atsmith813 (approved):
cut_it_out

benhalpern (approved):
> Since community_action isn't used in too many places (and is currently fairly pretty DEV-specific), we think that it makes more sense to drop this field and update the few instances of where it is used in emails instead.

Yup, this makes sense to me.....

I think if we want to inject more delight/customization to emails it can be done with more care and develop more powerful personalisation here.

PR #10684, opened 2020-10-06
Allow pattern match for spam terms

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This makes our spam detection more sophisticated matches by interpreting the string as a regex... It's also creates more potential for mistakes, so we will need to ensure this is used carefully.

This PR also fixes some issues with missing closing `div`s caused by another PR which had an improperly-resolved merge conflict.

### DIFF
--- a/app/views/admin/configs/show.html.erb
+++ b/app/views/admin/configs/show.html.erb
@@ -79,13 +79,13 @@
   <div id="siteConfig">
     <div class="card w-100">
       <div class="card-header" id="siteConfigHeader" data-toggle="collapse"
-        data-target="#siteConfigBodyContainer" aria-expanded="false" aria-controls="siteConfigBodyContainer">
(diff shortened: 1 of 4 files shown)


app/views/admin/configs/show.html.erb
```diff
@@ -1006,33 +984,4 @@
       </div>
     </div>
   </div>
-  <% if current_user.has_role?(:single_resource_admin, Config) && current_user.has_role?(:super_admin) %>
```
benhalpern, line 1009: I've been meaning to remove this for a while, as it doesn't really make sense in the Forem world. Just decided to do it here while cleaning things up.

vaidehijoshi (approved):

PR #10686, opened 2020-10-07
Limit articles created by newer users

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Allows us to treat "newer" users differently when it comes to rate limiting how many articles they can create.

Admins can configure how many articles a 3-day-old user can create within 5 minutes.

N/A

Screen_Shot_2020-10-06_at_5_17_24_PM

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [x] inline documentation
- [ ] docs.forem.com
- [ ] readme
- [ ] no documentation needed

### DIFF
--- a/app/services/rate_limit_checker.rb
+++ b/app/services/rate_limit_checker.rb
@@ -9,6 +9,7 @@ class RateLimitChecker
     listing_creation: { retry_after: 60 },
     organization_creation: { retry_after: 300 },
     published_article_creation: { retry_after: 30 },
+    published_article_antispam_creation: { retry_after: 300 },
     reaction_creation: { retry_after: 30 },
     send_email_confirmation: { retry_after: 120 },
     user_subscription_creation: { retry_after: 30 },
@@ -85,8 +86,15 @@ class RateLimitChecker
   end
 
   def check_published_article_creation_limit
+    # TODO: Vaidehi Joshi - We should make this time frame configurable.
     user.articles.published.where("created_at > ?", 30.seconds.ago).size >
       SiteConfig.rate_limit_published_article_creation
+  end
+
+  def check_published_article_antispam_creation_limit
+    # TODO: Vaidehi Joshi - We should make this time frame configurable.
+    user.articles.published.where("created_at > ?", 5.minutes.ago).size >
+      SiteConfig.rate_limit_published_article_antispam_creation
   end
 
   def check_follow_account_limit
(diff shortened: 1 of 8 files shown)


app/services/rate_limit_checker.rb
```diff
@@ -85,10 +86,17 @@ def check_comment_creation_limit
...
 
   def check_published_article_creation_limit
+    # TODO: Vaidehi Joshi - We should make this time frame configurable.
     user.articles.published.where("created_at > ?", 30.seconds.ago).size >
       SiteConfig.rate_limit_published_article_creation
   end
 
+  def check_published_article_antispam_creation_limit
+    # TODO: Vaidehi Joshi - We should make this time frame configurable.
+    user.articles.published.where("created_at > ?", 5.minutes.ago).size >
```
vaidehijoshi, line 96: We should make this `5.minutes.ago` a configurable timeframe; but for now, just hardcoded it and left a note for myself to do that later :)

benhalpern (approved):
LGTM (once the test is passed)

I think this is good for now, but note to us, we may want to make this message a little more delightful since a user could encounter this by accident and be thrown off.
 
Screen Shot 2020-10-06 at 8 40 02 PM

Let's not worry about this _right now_ because it's an edge case, but we can follow up with something _a bit_ more pleasant, but @joshpuetz @lisasy maybe we could come up with a specialized cuddly modal to deliver more news to users if they hit any kind of "new user" restrictions (hopefully flexible for all such scenarios we may introduce)

spec/requests/admin/configs_spec.rb
```diff
@@ -468,6 +468,13 @@
           end.to change(SiteConfig, :rate_limit_published_article_creation).from(9).to(3)
         end
 
+        it "updates rate_limit_published_article_antispam_creation" do
+          expect do
+            post "/admin/config", params: { site_config: { rate_limit_published_article_antispam_creation: 3 },
+                                            confirmation: confirmation_message }
+          end.to change(SiteConfig, :rate_limit_published_article_antispam_creation).from(1).to(2)
```
benhalpern, line 475: This should change to 3

vaidehijoshi: Lol, told you i was tired 😛

PR #10707, opened 2020-10-07
Drop profile columns from user

citizen428:

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR drops profile related columns from `User` and sets up delegated getters and setters to `Profile`. At one point we will also remove these delegators, but right now there are too many instances throughout the codebase that directly rely on specific profile columns that get accessed through `User`. This PR also removes the code used for syncing changes between users and profiles.

Notes:

This PR also removes two more validations:

* `conditionally_validate_summary`: I moved this into `Profile`, see the inline comment.
* `validate_mastodon_url` (and the instance whitelist): I think this is ok, the instance whitelist wasn't super practical anyway and in slight conflict with the federated nature of Mastodon.

Part of #6994

1. Checkout this PR.
2. Test everything you can think of, e.g.
  * Updating profiles
  * Updating user attributes like (email)
  * Post articles
  * Make comments
...

This change touches lots of things, so the more comprehensively we can test this, the better.

## Added tests?

- [X] yes, mostly modified existing ones

## Added to documentation?

- [X] no documentation needed

image


app/services/profiles/update.rb
```diff
@@ -13,9 +13,6 @@ def initialize(profile, updated_attributes)
     end
 
     def call
-      # Ensure we have up to date attributes
```
citizen428, line 16: We should no longer need to do this. Profile fields are now automatically loaded on app boot and later updated by the `ProfileFields::Add` and `ProfileFields::Remove` service objects.

lib/data/dev_profile_fields.csv
```diff
@@ -8,7 +8,7 @@ Youtube URL,text_field,https://www.youtube.com/channel/...,,Links
 StackOverflow URL,text_field,https://stackoverflow.com/users/...,,Links
 LinkedIn URL,text_field,https://www.linkedin.com/in/...,,Links
 Behance URL,text_field,https://www.behance.net/...,,Links
-Dribble URL,text_field,https://dribble.com/...,,Links
+Dribbble URL,text_field,https://dribbble.com/...,,Links
```
citizen428, line 11: This will be rebased out of here once the other PR is merged, but I needed it to be able to continue.

app/controllers/stories_controller.rb
```diff
@@ -465,17 +465,7 @@ def user_same_as
     [
       @user.twitter_username.presence ? "https://twitter.com/#{@user.twitter_username}" : nil,
       @user.github_username.presence ? "https://github.com/#{@user.github_username}" : nil,
-      @user.mastodon_url,
```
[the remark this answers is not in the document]
citizen428: @cc @juliannatetreault

app/models/profile.rb
```diff
@@ -51,7 +61,16 @@ def self.mapped_attributes
   # the profiles table doesn't exist yet (e.g. when running bin/setup in a new
   # clone). I wish Rails had a hook for code to run after the app started, but
   # for now this is the best I can come up with.
-  refresh_attributes! if ApplicationRecord.connection.table_exists?("profiles")
+  def self.db_ready?
```
citizen428, line 64: This is used in `refresh_attributes`, which is necessary so we can always start the app with profiles that are aware of all their fields. We need this helper so the method can do an early return in scenarios where the DB does not exist yet, e.g. `bin/setup`.

The `begin` - `rescue` - `else` construct is a bit odd, but works well for our use case. There's no way to check for an AR connection except for actually calling it. If there is no DB, it will raise an `ActiveRecord::NoDatabaseError`, which we resuce. If we don't we execute the `else` branch, which ensures that the `profiles` table exists.

I really wish Rails had a callback that I could hook into for this, but for the time being this will have to do.

app/models/profile.rb
```diff
@@ -51,7 +61,16 @@ def self.mapped_attributes
...
+    ActiveRecord::Base.connection
+  rescue ActiveRecord::NoDatabaseError
+    false
+  else
+    ActiveRecord::Base.connection.table_exists?("profiles")
+  end
+  private_class_method :db_ready?
+
+  refresh_attributes!
```
citizen428, line 73: Ruby evaluates class bodies as it reads the class, so you can actually call methods in them (that's how things like `belongs_to` work). Here we call `refresh_attributes!` to set up all the store accessors for the currently defined profile fields.

app/models/user.rb
```diff
@@ -221,6 +252,9 @@ def sync_profile
   before_create :set_default_language
   before_destroy :unsubscribe_from_newsletters, prepend: true
   before_destroy :destroy_follows, prepend: true
+
+  # NOTE: @citizen428 Temporarily while migrating to generalized profiles
+  before_save { |user| user.profile&.save if user.profile&.changed? }
```
citizen428, line 257: This is a bit of a hack but seems to work well: if we assign a new value to a delegated attribute and call `user.save`, the updated value will end up inside the profile's `data` attribute, so `profile.changed?` will be `true`.  I'm now wondering if I should change the condition to `valid? && user.profile&.changed? ` 🤔

app/services/profiles/update.rb
```diff
@@ -35,6 +35,17 @@ def success?
 
     private
 
+    def update_successful?
+      return false unless verify_profile_image
+
+      ActiveRecord::Base.transaction do
```
citizen428, line 41: We either update the profile **and** the user or neither.

app/models/profile.rb
```diff
@@ -60,4 +79,14 @@ def custom_profile_attributes
   def clear!
     update(data: {})
   end
+
+  private
+
+  def conditionally_validate_summary
+    return unless ProfileField.exists?(attribute_name: SUMMARY_ATTRIBUTE) && summary.present?
+    # Grandfather in people who had a too long summary before.
+    return if data_was[SUMMARY_ATTRIBUTE] && data_was[SUMMARY_ATTRIBUTE].size > MAX_SUMMARY_LENGTH
```
mstruve, line 88: TIL about `data_was` !

app/services/profiles/update.rb
```diff
@@ -19,14 +17,16 @@ def initialize(user, updated_attributes)
     end
 
     def call
-      if verify_profile_image && update_profile && sync_to_user
+      if update_successful?
         @user.touch(:profile_updated_at)
-        follow_hiring_tag
+        follow_hiring_tag if SiteConfig.dev_to?
```
mstruve, line 22: I cringe whenever I see environment specific. This feels like something simple that we could get rid of to help unify the behavior of all of the Forems and reduce "one-off"'s and complexity of the code. 

Rather than having the attribute, "looking_for_work", Im thinking we remove that and instead ask if they would like to follow the hiring tag. Using a user attribute to then follow a tag seems a little roundabout to me. While I don't think we need to do it in this PR, I think adding a TODO here to follow up with that cleanup would be super helpful and be a nice code cleanup.

mstruve (commented):
Few initial comments, going to swing back later, pull this puppy down, and give it a good pounding :)

app/models/profile.rb
citizen428: The `data` part is just the name of the column, you can always access the previous value in an update action as `column_name_was`.

app/services/profiles/update.rb
citizen428: This is existing behavior that was moved from `UsersController` to this service object. I wasn't sure if we want to remove it from DEV just yet, hence the conditional.

> Im thinking we remove that and instead ask if they would like to follow the hiring tag.

We can't assume this tag will exist across Forems. Or do you mean only on DEV?

> Using a user attribute to then follow a tag seems a little roundabout to me. 

Agreed, but like I said, I just moved the existing implementation.

fdoxyz (commented):
First of all I think the delegator approach is really cool! A change this big would make this a monster (huge) PR. Doing this data/model migration progressively makes it feasible 🎉 

So I wanted to test what it was like migrating from my local env to this branch (to see what kind of stuff we could run into when making the switch). I'm sharing my process, which is probably not doing things in the right order but may be helpful to understand the order of tasks that will need to be executed (probably for DEV and other Forems).

While testing on my local setup I simply did `bin/setup && bin/startup` and ran into a `NoMethodError` for Profile getters/setters while testing different areas of the site:

Screen Shot 2020-10-28 at 09 54 13

From what I can understand in the PR, we need the seeds locally. By this I mean we need the Profile fields to be generated. So I ran `rake db:seeds` but that gave me this other error:

Screen Shot 2020-10-28 at 09 52 59

I then went down the "nuke" route of `bundle exec rails db:reset && bin/setup && bin/startup` but that gave me this:

Screen Shot 2020-10-28 at 09 58 41

I think this has to do with the `forem.rake` task where, if we don't have any ProfileFields we invoke the custom field definitions here: https://github.com/forem/forem/blob/citizen428/drop-columns-from-user-6994/db/seeds.rb#L552

Does this mean we should reorder `seeds.rb` so we can have these fields generated before trying to create new users? (from the fresh local environment install point of view)

Let me know if there's anything else I'm missing as I may be totally overlooking something I'm doing wrong 😅

app/controllers/stories_controller.rb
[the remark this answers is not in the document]
juliannatetreault: Works for me! 👍 👍

mstruve:
Ran into the same issue that @fdoxyz did but with `employment_title` not being found for me.

citizen428:
Thanks @fdoxyz and @mstruve, it's been a while since I did a full `db:reset` on this branch, I'll look into it.

app/services/profiles/update.rb
citizen428: I added a comment now. Please let me know if you think it's sufficient. Also curious to hear how you think we should handle this in the future re "We can't assume this tag will exist across Forems. Or do you mean only on DEV?"

citizen428:
@fdoxyz and @mstruve I now worked on `db/seeds.rb` a bit more. In the process I added a new method to `Seeder`, `#create`, which just executes the block without any prior checking. This is good here, since I use the CSV importer for bringing in all DEV fields and that is already idempotent by itself. I also moved profile field creation towards the top of the file.

1. `db:reset` seems to work fine now, I did it a couple of times.
2. A normal `db:seed` will still add all missing DEV profile fields and also leave other existing profile fields alone. To verify that run `rails db:reset`, add a new field from the console, delete one of the existing DEV fields and then run `rails db:seed`. You should end up with 30 fields, 29 from DEV and the one you just created.

mstruve:
I still end up with an undefined method error when I go to view an article even after running `bin/setup` Is there a way we can set this up so it doesnt no require a `db:reset` to work properly? I dont think asking everyone to do a `db:reset` is practical for our contributors.

citizen428:
I'll look into this tomorrow. The `db:reset` wasn't meant to fix any app issues, it was just meant to fix, well, `db:reset`, `db:seeds`, and related commands as mentioned above by @fdoxyz.

mstruve:
>  This will obviously also add them for every other Forem, but until we have finally updated all code and can get rid of the delegation these fields are basically required. 

Sounds like then we don't really have a choice and need to add them to every Forem and then once the code is finished we will have to decide whether we want to remove them or not. Adding these profile fields just creates a db record of what is already on the user, correct? In which case, I would assume no one will notice the added Profile Fields to the other Forems since we already display these fields for users, is that also the correct assumption?

citizen428:
Yes @mstruve, all of that is correct. So far we always forced the DEV specific fields on every Forem as columns on the `users` table (and all the code that used those), now we'll also add 29 rows in the `profile_fields` table. From the perspective of other Forems nothing changes. 

I think it may be best to put this PR into draft and do the other things in a separate other PR that we merge first, because I probably also want the DUS to ensure all current user data is **really** present in the profiles before we disable the columns.

spec/fixtures/files/profile_fields.csv
```diff
@@ -1,4 +1,4 @@
-Name,text_field,John Doe,,Basic,header
-Skills/Languages,text_area,,Programming languages,Coding,left_sidebar
+Test name,text_field,John Doe,,Basic,header
+Test lLanguages,text_area,,Programming languages,Coding,left_sidebar
```
mstruve, line 2:
```suggestion
Test languages,text_area,,Programming languages,Coding,left_sidebar
```

mstruve (commented):
small spec typo fyi, probably why the specs are failing

spec/fixtures/files/profile_fields.csv
citizen428: Thanks!

citizen428:
Now that other necessary PRs have been merged and I updated this PR and fixed the specs again, I marked it as ready for review. 

Please, for the love of Sloan, test this as much as you can, there is quite a bit of potential for breakage here. I'm even wondering if we should do a test deploy of this on some smaller community (like Ben's), but I'm not sure how easy to to this is right now.

app/models/user.rb
```diff
@@ -581,39 +615,18 @@ def bust_cache
 
   def core_profile_details_changed?
     saved_change_to_username? ||
-      saved_change_to_name? ||
-      saved_change_to_summary? ||
-      saved_change_to_bg_color_hex? ||
-      saved_change_to_text_color_hex? ||
       saved_change_to_profile_image? ||
+      profile&.changed? || # TODO: @citizen428 This is not ideal, need better solution
```
citizen428, line 619: I want to call out that this may probably needs to be monitored as it could cause too many profile resaves. The alternative is to completely remove this for now and take some time to figure out a better way to do this. WDYT @mstruve?

app/services/profiles/update.rb
citizen428: @benhalpern I'm wondering how you'd feel about ripping of this particular bandaid right away? We managed to avoid going down a hiring/HR feature rabbit hole on DEV and also have a classifieds section, so maybe it's time to just phase out looking for work completely?

fdoxyz (commented):
I took this for a spin and played around with it. I had to re-seed manually to get the DEV profile fields but I'm guessing that won't be a problem for existing Forems thanks to #11246 being released already, right?

I didn't run into any problems before messing with the existing fields, thats great 🎉 

The only problem I see is that Admins can freely edit the existing DEV fields. Since many profile fields are hardcoded a Forem Admin can break the site if they edit fields like `summary`. This is what I encountered after setting it to `check_box`.

Screen Shot 2020-11-19 at 11 37 31

I just want to make sure this is known/intentional, so please correct me if I'm wrong with the following "next steps" thought process:
1. The `:profile_admin` Feature Flag will remain in place, meaning Forem Admins won't be able to break their instances
1. Future PRs will remove the direct use of profile fields on the User model (`user.field = XXXX` will be replaced with `user.profile.field = XXXX`)
1. Profile views will be generalized in future PRs because right now new fields won't ever be visible (i.e. `app/views/users/_sidebar.html.erb`)
1. Before removing the feature flag we should consider setting some validation on "core" profile fields. Something like "readonly" profile fields, because `summary` and others feel like they shouldn't be editable/removable even by Forem admins
1. Finally remove the `:profile_admin` Feature Flag

Does this make sense? If so I think we should be safe in moving this to production. Just left an inline comment on the User `core_profile_details_changed?` method discussion

citizen428:
> I had to re-seed manually to get the DEV profile fields but I'm guessing that won't be a problem for existing Forems thanks to #11246 being released already, right?

Hmm, it also shouldn't have been necessary for you if you actually ran the data update script that we ran on all the Forems. Or maybe you ran `bin/setup` from another branch in between, I think that would still have removed them and only reseeded the base fields. Maybe you can verify that?

> The only problem I see is that Admins can freely edit the existing DEV fields. Since many profile fields are hardcoded a Forem Admin can break the site if they edit fields like summary.

Yes, but that's known and your assumptions are correct.

Re 1. Correct, the feature flag will remain in place for the time being.
Re 2. Removing the hardcoded values is one of the biggest remaining chunks of work.
Re 3. See #11485 
Re 4. This is primarily a product decision. How much freedom do we want to give admins? Maybe a super privacy focussed IT security community wants almost completely empty profiles. Should they be able to do so? I don't know. 🤷 
Re 5. The day can't come soon enough.

Thanks for the other feedback as well, will address now.

app/models/user.rb
citizen428: You are correct, we only need to do this for a few fields, which is why my comment calls out this code as less than ideal. 

The problem is that `core_profile_details_changed?` on `Profile` doesn't necessarily work. To make it reliable it would need to check `data_was` during the profile update, which may already have happened at this point.

But typing out this answer made me realize that this whole method should move into the `Profiles::Update` service object because that's the one place where I have all the necessary info to make this decision in the same place and it's also the canonical way to update users and profiles going forward.

Excelltent rubberducking, thanks!



I was thinking about this yesterday, the problem is that this

mstruve:
Hey @citizen428 its been a long week and this really needs a set of Molly morning eyes, not 6pm Friday eyes. Sorry for the delay but it will be my top priority Monday!!!

citizen428:
@mstruve This is 💯 the right call, no need to apologize at all. With the potential for breakage in this PR, I'd much rather wait an extra day or two.

app/services/profiles/update.rb
mstruve: @citizen428 what do you think about making an Issue for this in the internal repo and putting it on the EOY Priority board in Input Needed for after this is merged? Seems like a good cleanup task to unify Forem behavior

spec/rails_helper.rb
```diff
@@ -95,6 +95,13 @@
   config.before do
     # Worker jobs shouldn't linger around between tests
     Sidekiq::Worker.clear_all
+
+    # NOTE: @citizen428 needed while we delegate from User to Profile to keep
+    # spec changes limited for the time being.
+    ProfileField.destroy_all
```
mstruve, line 101: This data refresh is going to slow down our specs considerably. Does every spec start failing without this? Is there any way we can limit this to a smaller subset of specs?

mstruve (commented):
Comment and one question, pulling down now to test locally.

mstruve:
Going to drop little things here as I find them. I went to save my profile with a summary that was too long and the error message is a bit obfuscated. 
Screen Shot 2020-11-23 at 9 53 56 AM

mstruve (commented):
Everything looks to be saving and persisting correctly for users! I switched to the branch with my given dataset and everything worked great. I also reset my database and started from 0 creating a user and then updating it and everything seems to be working. Created some posts, comments, and reactions. Only issue I ran into was that summary validation error display. If there is anything I missed you want to be tested let me know! 

That organization elasticsearch doc(spec/models/organization_spec.rb) spec is legit failing, I tried looking into it briefly but nothing obvious stood out as the cause. I thought maybe the Sidekiq job wasn't getting executed but that doesnt seem to be the case, something deeper is going on. 

I am concerned about the spec run time having to reset profile fields after every spec. It looks like it adds about 6 min to our spec run times. Could we maybe stub some of those methods on the user factory rather than having to write to the database? 

Overall, this is really close!!! Thanks for all the hard work on this project, its been a long process but I am so excited now that I see the end in sight :D

app/services/profiles/update.rb
citizen428: @mstruve Done ✅ https://github.com/forem/InternalProjectPlanning/issues/366, it's also on the board in "input needed".

spec/rails_helper.rb
citizen428: > Does every spec start failing without this?

No, but too many to fix right now in one go. 

> Is there any way we can limit this to a smaller subset of specs?

That's how I started, by making a shared context. When I realized how many specs I'd need to add this too I went for this instead. Let me try and just move this to the `before(:suite)` block so we only create them once.

citizen428:
> Everything looks to be saving and persisting correctly for users!

Thanks for the feedback @mstruve, that's encouraging to hear.

> Only issue I ran into was that summary validation error display. 

Thanks for spotting this, I'll look into it.

> That organization elasticsearch doc(spec/models/organization_spec.rb)

Hmm, it's hard to see how this would happen because of this branch, but I'll have a look.

> I am concerned about the spec run time having to reset profile fields after every spec.

Let's see how `before(:suite)` is doing. It would only create them once, which should be fast enough.

> Overall, this is really close!!! Thanks for all the hard work on this project, its been a long process but I am so excited now that I see the end in sight :D

At least one of us sees an end in sight 😉

citizen428:
The good news: `before(:suite)` seems to be working, the specs finished in 10 minutes and 46 seconds.

The not quite so good news: `spec/models/organization_spec.rb)` is still failing, but it's the only one.

citizen428:
The summary too long error message has been fixed too now. I had only done it for user errors not profile ones because until I added the summary validation there were no error messages on profiles.

Screen Shot 2020-11-24 at 11 14 10

I did a quick fix at first, which took like 5 minutes. In the end, this led me down a path of completely reworking how we deal with errors in this service object which took a bit. I do think the approach is both cleaner and more in line with how we do things elsewhere, so it was worth it.

mstruve:
Ooooh so fast for the specs! Thank you for taking the time to do that. 

If I have some time later I will try and dig further into that org spec. Maybe take a pry and try stepping through, I agree it seems like it should be completely unrelated. 🤔

fdoxyz:
I'll take a thorough look again today 👍🏼

mstruve (approved):
The organization spec is [failing due to the removal of this line](https://github.com/forem/forem/pull/10707/files#diff-9802ca3c9c4cf89904fd44bc114e35ebdf2c5dd3d5b645491e2b253e1afef29bL227) because the organization spec was incorrectly passing to begin with. I know what the problem is, an `after_commit on: :destroy` is not working properly so feel free to comment out the spec and throw a `#TODO` saying I will fix it. Going to do it in the morning bc there are a few options that I dont feel like deciding between right now :).

citizen428:
Thanks @mstruve, I did this now:

```ruby
    # TODO: This will be fixed by @mstruve
    # https://github.com/forem/forem/pull/10707#pullrequestreview-538071192
    xit "on destroy removes data from elasticsearch" do
```

citizen428:
Travis is green again now that the org specs has been commented out, so feel free to take this for a spin whenever you're ready @fdoxyz. Thanks for the help and reviews (also @mstruve), it's much appreciated! 🙏

fdoxyz (commented):
This feels very solid and it didn't break for me during my tests locally! 

I don't know why I only just realized this now, but should we validate all other custom fields and not just the `summary`? I was able to add a very large string (thousands on chars) for a `text_field` type as there's no DB limit for them (in this case Postgres would fail at some point for `text` or `string` types)

I was able to get this to work with a simple validator method on the profile:

```ruby
def validate_custom_attributes
  ProfileField.all.each do |field|
    # Summary is validated in conditionally_validate_summary
    next if field.attribute_name == SUMMARY_ATTRIBUTE

    case field.input_type
    when 'check_box'
      # no check necessary here bc it will only evaluate as true/false
      next
    when 'color_field'
      # Reusing TAG regex but I think we can also do without this check just like `check_box`
      hex_value = send(field.attribute_name)
      next if hex_value.nil? || Tag::HEX_COLOR_REGEXP =~ hex_value
      errors.add(field.attribute_name, "is not a valid hex color")
    when 'text_area'
      # The `&` and `to_i` are there for fallbacks which I think are necessary
      next if send(field.attribute_name)&.size.to_i <= 200
      errors.add(field.attribute_name, "is too long.")
    when 'text_field'
      # The `&` and `to_i` are there for fallbacks which I think are necessary
      next if send(field.attribute_name)&.size.to_i <= 100
      errors.add(field.attribute_name, "is too long.")
    end
  end
end
```

If I missed a conversation somewhere about why this isn't necessary just let me know, because otherwise this LGTM 🙂

A future nice to have the Profile Field limit could be set by admins but I think this would definitely be outside the scope of this PR

citizen428:
> This feels very solid and it didn't break for me during my tests locally!

@fdoxyz Thanks for the feedback, that's great to hear!

> I don't know why I only just realized this now, but should we validate all other custom fields and not just the summary?  [...] If I missed a conversation somewhere about why this isn't necessary just let me know

It's a good point and we did talk about it in Slack a bit, which brought up custom validations defined by Forem admins and we eventually decided to punt the whole topic. Your idea to add some basic validations is good though, I took it as the basis for my solution. Thanks!

citizen428:
@mstruve @fdoxyz After adding the validations I went ahead and made some other changes, mostly cleanup (removing some no longer necessary methods), refactoring/code reorganization and small fixes (e.g. making sure the CSV importer refreshes attributes at the end of the import). I completely dropped my DB and verified that `bin/setup` works (several times), but if you'd be willing to take this branch for a spin one more time I'd be very grateful.

app/models/user.rb
```diff
@@ -5,7 +5,40 @@ class User < ApplicationRecord
   include Searchable
   include Storext.model
 
-  self.ignored_columns = %w[currently_streaming_on twitch_username]
+  # @citizen428 Preparing to drop profile columns from the users table
+  PROFILE_COLUMNS = %w[
```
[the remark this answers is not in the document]
fdoxyz: Just noticed that `name` needs to be added to this list too because [it's being used in the views](https://github.com/forem/forem/blob/citizen428/drop-columns-from-user-6994/app/views/users/show.html.erb#L52) to render the User's profile name. Updating a user's name didn't reflect in my profile page locally because it's passed in as a Profile param and not a User param. Does this happen to you as well?

Otherwise I think we would need to get rid of `name` from the Basic `ProfileFieldGroup` if we want users to **always** have a name or make it "read only". I'm okay with just adding `name` to the `PROFILE_COLUMNS` so the profile page gets updated correctly, but just wanted to mention this because we had [a similar conversation here](https://github.com/forem/forem/pull/10707#issuecomment-730816702).

fdoxyz (changes requested):
I really like the ProfileValidator approach! This definitely brings peace of mind no one should be able to abuse the DB storage 🙌🏼

I did run into a problem with the validator though, check out this video:

https://share.getcloudapp.com/OAuWOWkr

It's a bit odd because it doesn't happen on every update 🤔 From what I've seen it happens on newly created fields (the `record` doesn't have the field yet). This results in a `NoMethodError` inside the validator, for example when `text_area_valid?` or `text_field_valid?` call on `record.public_send(attribute)`:

Screen Shot 2020-11-27 at 10 02 10

I wish I was able to help debug this a bit further but I'm not sure where to go from here, specially since sometimes you have to go over the process of creating a new custom field and editing the settings of a user a few times. All I know is that the record is coming in **without** the field.

citizen428:
@fdoxyz Thanks again for taking this for a spin!

When you say "newly added field", were they added via `ProfileFields::Add` (which the admin area also uses) or via a Rails console? If the latter, did you call `Profile::refresh_attributes!` after creating the field? This is taken care off by the service and CSV importer but needs to be explicitly called when just adding a field.

fdoxyz:
> When you say "newly added field", were they added via ProfileFields::Add (which the admin area also uses) or via a Rails console?

The new field was added using the admin area (`/admin/profile_fields`), so it was using ProfileFields::Add. The way I was able to run into this was by doing:
1. Add new field (in a new group)
1. Edit settings (any other field but leaving the newly added field blank) and save
1. Repeat # 2 (Edit settings) a few times until the `NoMethodError` appears

The clip (~1min) goes through this process:

> https://share.getcloudapp.com/OAuWOWkr

-----------

From the screenshot above and other debugging this is what I was able to understand:

Screen Shot 2020-11-27 at 10 02 10

1. It is to be expected for the Profile to arrive without all existing ProfileFields, right?
   - If I'm following correctly, the slicing/removing of unused fields gets rid of unused ones
   - From the screenshot above:
     - The record (**red**) does not have the `wild_field` (**yellow** - I'm not good at naming)
     - The call originally comes from ProfileValidator#validate line 25 (**green**) which loops around `ProfileField.all`
2. Based on the above, would it make sense to add a `responds_to?` check before line 25? Something like the following:

```ruby
next unless record.respond_to?(attribute) # skip check if attribute isn't there?
next if __send__("#{field.input_type}_valid?", record, attribute)
```

citizen428:
Yes, when I was thinking about this a bit more this morning I contemplated a `responds_to?` check too, I just don't want it to accidentally mask a more subtle problem. Anyway, I'm going to look at this on Monday with a fresher set of eyes.

citizen428:
@fdoxyz I tried to reproduce what happens in your video, but no luck yet. Tbh, the behavior seen there makes little sense to me, the `store_attribute` is defined on the class level, so it shouldn't randomly disappear (you are correct about the key being removed from the data hash), but that doesn't trigger a `NoMethodError`:

```ruby
Profiles::Update.call(profile_attributes: { test_field: nil })
# at binding.pry breakpoint
record.public_send(attribute)
#=> nil
# back in the normal Rails console
user.profile.test_field
#=> nil
user.profile.respond_to?(:test_field)
#=> true
```

In the meantime, I think I reached a decision, `name` should remain on `User` as I think it's an inherent attribute, just as `username`.

app/validators/profile_validator.rb
```diff
@@ -0,0 +1,62 @@
...
+  MAX_TEXT_AREA_LENGTH = 200
+  MAX_TEXT_FIELD_LENGTH = 100
+
+  HEX_COLOR_REGEXP = /^#?(?:\h{6}|\h{3})$/.freeze
+
+  ERRORS = {
+    color_field: "is not a valid hex color",
+    text_area: "is too long (maximum: #{MAX_TEXT_AREA_LENGTH})",
+    text_field: "is too long (maximum: #{MAX_TEXT_FIELD_LENGTH})"
+  }.with_indifferent_access.freeze
```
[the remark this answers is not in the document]
citizen428: Rails internally uses `HashWithIndifferentAccess` quite a bit, and while I'm a bit ambivalent about it it's idiomatic and occasionally useful.

citizen428:
@fdoxyz I think I reached a decision, `name` should remain on `User` as I think it's an inherent attribute, just as `username`.

citizen428:
@fdoxyz Your error still baffles me as I can't seem to come up with a way to repro this (I only managed once, with a stale object that somehow didn't reload).

1. Load a  profile
    ```ruby
    profile = User.find(12).profile
    ```
2. Add a new field with the service object:
    ```ruby
    ProfileFields::Add.call(label: "So confusing")
    ```
3. The already loaded `profile` will now respond to this new method as will new profiles:
    ```ruby
    profile.respond_to?(:so_confusing)
    #=> true
    Profile.new.respond_to?(:so_confusing)
    #=> true
    ```

fdoxyz:
> @fdoxyz I think I reached a decision, name should remain on User as I think it's an inherent attribute, just as username.

That's great 👍  We'll need to add this text_field to the User fields in the Settings then (like username), right?

Screen Shot 2020-11-30 at 16 39 03

------

> .... the store_attribute is defined on the class level, so it shouldn't randomly disappear (it also shouldn't disappear from data, only deleted fields will as per the comment [1]), but that doesn't trigger a NoMethodError

1. I think that's the exact situation when the problem happens. As soon as the field **X** (name doesn't matter)  is deleted from the Profile `store_attribute` (JSON payload).
1. The next time the ProfileValidator validates it will loop around all existing ProfileFields in the DB, and for each one will send a validation request to `text_field_valid?` (in case it is a text_field).
1. Inside that `text_field_valid?` it calls `record.public_send(attribute)`, which will raise the `NoMethodError`. Which is why I think this check should fix the problem:

```ruby
next unless record.respond_to?(attribute) # skip check if attribute isn't there?
next if __send__("#{field.input_type}_valid?", record, attribute)
```

I think there's no underlying problem if we skip these checks, because it only means the fields are not there (inside the JSON payload).  In the future, if we add more control for admins by giving them the ability to mark certain custom fields as mandatory then we will need to validate those fields exist. But until then this shouldn't be a problem AFAICT

Check out #11687 where I add a system spec that reproduces this exact situation. If you add that `responds_to?` validation the system spec is fixed :D

citizen428:
In my experience, this is not how `store_accessor` works, because that would make it a lot less useful for working with JSON. Any field that's absent from the backing column is treated as `nil`:

```ruby
p = Profile.new
p.data
#=> { "custom_attributes" => {} }
p.respond_to?(:brand_color1)
#=> true
p.brand_color1
#=> nil
```

citizen428:
For anyone who is still following along with this, here's the exciting conclusion to profile generalization season 3: the culprit was caching. Fernando ran his local env with caching enabled, I didn't. The problem is that when an admin adds a new profile field and then goes to edit their own profile, the field won't exist in their cached user/profile. It ShouldWork™ for other users though. This convinced me that adding a` responds_to?` check will mask a more subtle underlying problem, so I'll add this check later.

citizen428:
@fdoxyz I now added the responds_to? check, added name back as user field in the profile settings page and fix the remaining spec. I think this is good to go for one last round of reviews.

mstruve:
Going to take this for a spin. One request, can we tweak the description and title?
> This PR drops profile related columns from User and sets up delegated getters and setters to Profile

When I see drops I think the actual migration to drop the columns from the db is in this PR and I was confused for a second when I didn't see it. Maybe "drops use of profile columns from Active Record" so its clear to anyone in the future.

mstruve (approved):
Still LGTM! 🚀 

Cut over to branch, did lots of things with existing user. Deleted user and re-signed up. Did more stuff, resetting username, name, URL fields, commenting, posts, all worked great.

fdoxyz (approved):
Ran a thorough sweep from my end too. It looks ready for the rocket emoji 🚀

citizen428:
Awesome, thanks for the feedback @mstruve and @fdoxyz. Molly, what do you think about deploying this, I'd feel better if someone with full access is around when we merge this so was thinking my AM your PM. What do you think?

app/models/profile.rb
```diff
@@ -34,30 +37,16 @@ def self.attributes
...
     custom_profile_fields.pluck(:attribute_name)
   end
 
   def clear!
     update(data: {})
   end
+
+  # NOTE: @citizen428 We want to have a current list of profile attributes the
+  # moment the application loads. I wish Rails had a hook for code to run after
+  # the app started...
```
rhymes, line 50: @citizen428 it does: `config.after_initialize`: https://guides.rubyonrails.org/configuring.html#rails-general-configuration

citizen428: I'm aware of that, but it didn't seem to work for this specific case. I may try it again  in the future, but I was just happy to have it run at one point.

Issue #5023, opened 2019-12-06
Add inline markdown syntax highlight and keyboard shortcuts for editor


deepu105:
**Is your feature request related to a problem? Please describe.**

The current Editor is quite basic and is not very user-friendly on the Go, especially when you want to quickly write something up. It would be nice to have at least the below support in the editor

1. Inline Markdown syntax highlight (like in VSCode for example)
2. Keyboard shortcuts(ctrl+s, ctrl+i, ctrl+b and so on similar to [this](https://github.com/yzhang-gh/vscode-markdown) VScode plugin)

**Describe the solution you'd like**

I believe the above can be implemented as a client-side feature using existing libraries and would greatly improve the writing experience on Dev.

The VSCode plugin [here](https://github.com/yzhang-gh/vscode-markdown) can be used as inspiration as its OSS

**Describe alternatives you've considered**

Currently, I write in VSCode and copy-paste to the Dev editor, then I also use Grammarly to correct grammar, so the ability to write directly on the Dev webpage would definitely improve the workflow for anyone who has a similar process.

If there is interest I would be glad to try and implement this with a PR

**Additional context**

Here are some screenshots from the Plugin mentioned

![image](https://github.com/yzhang-gh/vscode-markdown/raw/master/images/gifs/toggle-bold.gif)

![image](https://github.com/yzhang-gh/vscode-markdown/raw/master/images/toc.png)

![image](https://github.com/yzhang-gh/vscode-markdown/raw/master/images/gifs/on-tab-key.gif)

deepu105:
Could be related to https://github.com/thepracticaldev/dev.to/issues/3049

jessleenyc:
Thanks for the request @deepu105, updated the title for clarity.

mvasigh:
Is this something that we would want to use an existing solution for? And maybe this would be something that would be opt-in depending on bundle size impact?

I'd like to work on this!

mudlabs:
This would be a good feature. I was just about to make a request for some StackOverflow style keyboard shortcuts for the markdown editor.

- Except where their `Ctrl+K` adds an indented line, we would add a ``` code block.

Link2Twenty:
Just discovered there are actually keyboard shortcuts on that page (only one).

Pressing CTRL + SHIFT + P toggles the preview. I think the way the code is written at the moment it wouldn't be easy to add many more shortcuts and also it means the component can only be used on that one screen.

I suggest we do a rewrite making it more robust I was thinking something like

```js
[...]

PR #10713, opened 2020-10-07
Allow keyboard shortcuts

Link2Twenty:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Add handler for keyboard shortcuts. This allows custom shortcuts to be set up on a page by page (or even component by component) basis.

[React demo](https://codesandbox.io/s/fervent-williams-53pin)

### Usage

In this example we're going to run a function when the user presses CTRL+SHIFT+P
```jsx
const shortcuts = {
  "ctrl+shift+KeyP":  (e) => {
    e.preventDefault();
    someFunction();
  }
}

<KeyboardShortcuts shortcuts={shortcuts} />
```

related: #5023 #596

This facilitates changes in the future

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Once this is merged I'd like to modify this [current shortcut](https://github.com/forem/forem/blob/master/app/javascript/article-form/components/keyboardShortcutsHandler.jsx) to use this method before thinking about what other shortcuts could/should be added.

seeing the code


rhymes:
Hi @Link2Twenty, interesting! What if we integrate https://github.com/github/hotkey ?

It'd be great if we'd eventually work towards a keyboard shortcuts modal populated by the registered shortcuts on each page, something like this:

Screenshot_2020-10-07 Allow keyboard shortcuts by Link2Twenty · Pull Request #10713 · forem forem

Link2Twenty:
It looks like hotkeys is only for interacting with specific elements. Whereas this way you can call any function you like including, but not limited to, dom interactions.

I liked how they registered their shortcuts, was much cleaner than my way do I've _borrowed_ that.

reobin:
Super clean @Link2Twenty !

I've implemented something similar in #10468 that is called `KeyEventListener`.

It would be nice to merge our efforts to go in the right direction with forem native shortcuts. My component doesn't handle key modifiers like `ctrl` and `shift`, which I think is a really nice touch.

I've already implemented `KeyEventListener` in a couple of places in the PR like the `/` and `0` shortcuts for the header, as well as `j` and `k` navigation for the article feed if you want to take a look.

Tell me what you think, feedback is appreciated! 🌞

Then, as @rhymes said, we could create an issue to work on a store that is fed by the calls to that very component. With that store, build a modal that lists all the available shortcuts for a given page.

Link2Twenty:
Thanks @reobin 😊 I've updated my code to allow it to cope with single key shortcuts now, similar to hotkeys.

@nickytonline is this the sort of direction you'd like to see this go? If so we could get this merged then Reobin and I could look at using this in #10468

citizen428:
@Link2Twenty I really like the work you and @reobin are doing on this! 😄

nickytonline:
Just got out of the stream. Taking a peek now @Link2Twenty.

app/javascript/shared/components/keyboardShortcuts.jsx
```diff
@@ -0,0 +1,66 @@
...
+ * const shortcuts = {
+ *   "ctrl+alt+KeyG": (e) => {
+ *     e.preventDefault(); 
+ *     alert("Control Alt G has been pressed")
+ *   }
+ * }
+ * <KeyboardShortcuts shortcuts={shortcuts} />
+ * 
+ */
+const KeyboardShortcuts = ({ shortcuts, ...props }) => {
```
nickytonline, line 19: This is a component that returns `null` always. That's OK to do and valid, but given that this component is just a `useEffect`, consider making a custom hook instead.

I made a quick proof of concept here, https://codesandbox.io/s/alternate-suggestion-for-httpsgithubcomforemforempull10713-w5fro?file=/src/useKeyboardShortcuts.js

If we need to support class based components as well, you can do what I did for drag and drop, e.g. https://github.com/forem/forem/blob/master/app/javascript/utilities/dragAndDrop.js. I provide a hook and a component.

Thoughts?

app/javascript/shared/components/keyboardShortcuts.jsx
```diff
@@ -0,0 +1,66 @@
...
+      if (e.defaultPrevented) return;
+      // Get special keys
+      const keys = `${e.ctrlKey || e.metaKey ? "ctrl+" : ""}${e.altKey ? "alt+" : ""}${e.shiftKey ? "shift+" : ""}`;
+      // If no special keys are pressed and focus is inside a field return
+      if (e.target instanceof Node && isFormField(e.target) && !keys) return;
+      const shortcut = shortcuts[`${keys}${e.code}`];
+      if (shortcut) shortcut(e);
+    };
+
+    window.addEventListener("keydown", keyEvent);
```
nickytonline, line 47: This may not be required now, so don't [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), but we may want to add keyboard events of any type, e.g. `keyup`, `keypress` etc.

nickytonline (commented):
@Link2Twenty, thanks for the PR! This is looking good. Just some questions.

app/javascript/shared/components/keyboardShortcuts.jsx
reobin: Nice! I like the hook/component combo. With that said, I think the `KeyboardShortcuts` component's difference will be that it should not necessarily "wrap" anything (or obligatory have children) as we want the event listeners on `document` most of the time and not only on the children elements.

Note: I'll WIP #10468 for now while we settle on this, and update it as soon as it's merged

nickytonline:
> Thanks @reobin 😊 I've updated my code to allow it to cope with single key shortcuts now, similar to hotkeys.
> 
> @nickytonline is this the sort of direction you'd like to see this go? If so we could get this merged then Reobin and I could look at using this in #10468

@Link2Twenty I think going with use effect makes sense. I made a suggestion to consider making what you're proposing a custom hook. See PR review comments.

Update 10 minutes after writing first paragraph.😄 I saw @reobin's PR is using a custom hook, https://github.com/forem/forem/pull/10468/files#diff-976b57823cb7a305a55921032b2708aeR45, so I think a custom hook with `useEffect` is the direction to go.

app/javascript/shared/components/keyboardShortcuts.jsx
Link2Twenty: Thanks for the review! I've moved it over to a hook and added a component that uses that hook.

app/javascript/shared/components/keyboardShortcuts.jsx
Link2Twenty: I think it would be easy enough to add an `eventType = "keydown"` down the road without it breaking anything.

Link2Twenty:
@nickytonline any further thought or are we good to go?

app/javascript/shared/components/useKeyboardShortcuts.jsx
```diff
@@ -0,0 +1,97 @@
...
+ * @param {object} shortcuts List of keyboard shortcuts/event
+ * @param {EventTarget} [eventTarget=window] An event target.
+ *
+ */
+export function useKeyboardShortcuts(shortcuts, eventTarget = window) {
+  useEffect(() => {
+    if (!shortcuts) return;
+
+    // Return true if element is a form element
+    const isFormField = (element) => {
```
nickytonline, line 28: refactor (non-blocking): `isFormField` can be a function outside of `useKeyboardShortcuts` since it does not rely on anything from `useKeyboardShortcuts`.

app/javascript/shared/components/useKeyboardShortcuts.jsx
```diff
@@ -0,0 +1,97 @@
...
+      const type = (element.getAttribute("type") || "").toLowerCase();
+      return (
+        name === "select" ||
+        name === "textarea" ||
+        (name === "input" && ["submit", "reset", "checkbox", "radio"].indexOf(type) < 0) ||
+        element.isContentEditable
+      );
+    };
+
+    const keyEvent = (e) => {
```
nickytonline, line 43: refactor (non-blocking): `keyEvent` depends on the argument `shortcuts`, but it does not need to be directly in the `useEffect`, it can be declared outside of the `useEffect` call but still reside in the `useKeyboardShortcuts` function body.

app/javascript/shared/components/useKeyboardShortcuts.jsx
```diff
@@ -0,0 +1,97 @@
...
+ * }
+ *
+ * <KeyboardShortcuts shortcuts={shortcuts} />
+ * <KeyboardShortcuts shortcuts={shortcuts} eventTarget={ref.current} />
+ *
+ * @param {object} shortcuts List of keyboard shortcuts/event
+ * @param {EventTarget} [eventTarget=window] An event target.
+ *
+ */
+export function KeyboardShortcuts({ shortcuts, eventTarget }) {
```
nickytonline, line 83: I imagine this is still here so that class based components can use the hook?

nickytonline (changes requested):
This is looking good @Link2Twenty. I have some suggestions, but they do not block the PR.

One blocker though is we need some tests. If you need to see an example of testing hooks, take a peek at https://github.com/forem/forem/blob/master/app/javascript/utilities/__tests__/dragAndDrop.test.js and if you have any questions, let me know.

app/javascript/shared/components/useKeyboardShortcuts.jsx
Link2Twenty: Don't useEffects prefer a used function to be stated inside the useEffect where possible? If you have it outside, as a const for instance, doesn't that get reset on each render meaning the useEffect is recalled each render?

app/javascript/shared/components/useKeyboardShortcuts.jsx
Link2Twenty: Yep that was the idea

Link2Twenty:
> One blocker though is we need some tests. If you need to see an example of testing hooks, take a peek at https://github.com/forem/forem/blob/master/app/javascript/utilities/__tests__/dragAndDrop.test.js and if you have any questions, let me know.

I'm sure my lack of testing knowledge is more than apparent 😅
I've added some tests but I'm sure they'll need some changes. Also there is one test case I couldn't work out. I'm sending fake keyboard events in order to make sure the functions are called but I also want to test this with the target being a text area (it should prevent the function from being called.

nickytonline:
> I'm sure my lack of testing knowledge is more than apparent 😅
> I've added some tests but I'm sure they'll need some changes. Also there is one test case I couldn't work out. I'm sending fake keyboard events in order to make sure the functions are called but I also want to test this with the target being a text area (it should prevent the function from being called.

Thanks for adding some tests @Link2Twenty. 👏🏻 For the scenario where you want to test a `<textarea />`, you could do something like this:

```jsx
it('should fire a function when keydown is detected', () => {
      const keyPress = jest.fn();
      const event = new KeyboardEvent('keydown', { code: "KeyK" });
      const eventTarget = document.createElement('textarea') // eventTarget set since the default is window

      Window.prototype.addEventListener = jest.fn();

      const { result } = renderHook(() =>
        useKeyboardShortcuts({
          KeyK: keyPress
        },
        eventTarget,
      );
      eventTarget.dispatchEvent(event);

      // tests to handle the use case when it's a text area
    });
```

app/javascript/shared/components/useKeyboardShortcuts.jsx
nickytonline: Use effect will only be called when the dependency array is changed. The functions I'm suggesting to move out would be referenced only and created one time (when the ES module is first instantiated. If they remain in the `useEffect` callback, these functions will be recreated every time the `useEffect` callback is called.

nickytonline:
@Link2Twenty, if you're having issues testing your use case with a form element, let me know and I can push a commit to your branch.

app/javascript/shared/components/useKeyboardShortcuts.jsx
Link2Twenty: Maybe this is my understanding breaking down (or preact differs from React) but I thought.

```jsx
const FakeComponent = ({...props}) => {
  const someFunction = () => {
    // I'm created on every render
  }

  useEffect(()=>{
    someFunction();
    // therefore I run on every render
  },[props.cheese, someFunction])

  return null
}
```

compared to
```jsx
const FakeComponent = ({...props}) => {
  useEffect(()=>{
     const someFunction = () => {
       // I'm created every time the cheese prop changes
     }

    someFunction();
    // therefore I run on every cheese prop change.
  },[props.cheese])

  return null
}
```
and in React I'd do
```jsx
const FakeComponent = useCallback(({...props}) => {
  const someFunction = () => {
    // I'm created the first render only
  }, [])

  useEffect(()=>{
    someFunction();
    // therefore I run on every cheese prop change
  },[props.cheese, someFunction])

  return null
}
```

Link2Twenty:
@nickytonline I'm not entirely sure why the tests are failing.

I've checked my [code](https://codesandbox.io/s/alternate-suggestion-for-httpsgithubcomforemforempull10713-forked-t64iv) and it does actually work so I must be doing something wrong with the tests.

Link2Twenty:
I keep getting random ideas to fix bit and it's working 😮
Just 2 tests left to get working 😁

Link2Twenty:
@nickytonline I think those last two may have stumped me. Any pointers?

app/javascript/shared/components/useKeyboardShortcuts.jsx
nickytonline: You don't even need to use `useCallback`. What I meant was, you can write it like this.

```jsx
const someFunction = () => {
    // I'm declared only once
}

const FakeComponent = ({...props}) => {
  useEffect(()=>{
    someFunction(); // I'm called in use effect, but I'm never redeclared.
    // therefore I run on every render
  }, [props.cheese, someFunction])

  return null
}
```

`props.cheese` 😆

nickytonline: The only time you need to have a function like `someFunction` declared within the function component is if it requires access to variables in the [closure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures).

Link2Twenty: In this bit of code the function `keyEvent` relies on `shortcuts`, so needs to stay somewhere where that is accessible. Unless you'd rather we pass shortcut from useEffect to the function outside?

```jsx
const someFunction = (cheese) => {
    // I'm declared only once
}

const FakeComponent = ({...props}) => {
  useEffect(()=>{
    someFunction(props.cheese); // I'm called in use effect, but I'm never redeclared.
    // therefore I run on cheese prop update
  }, [props.cheese])

  return null
}
```

app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx
```diff
@@ -0,0 +1,115 @@
...
+    });
+
+    it('should add event listener to window', async () => {
+      HTMLDocument.prototype.addEventListener = jest.fn();
+
+      const { rerender } = render(() =>
+        <KeyboardShortcuts eventTarget={document} shortcuts={{ KeyK: null }} />,
+      );
+      
+      rerender(() =>
```
nickytonline, line 91: refactor (blocking): The rerender is not necessary.

app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+    it('should remove event listener when the hook is unmounted', async () => {
+      HTMLDocument.prototype.addEventListener = jest.fn();
+      HTMLDocument.prototype.removeEventListener = jest.fn();
+
+      const { rerender, unmount } = render(() =>
+        <KeyboardShortcuts eventTarget={document} shortcuts={{ KeyK: null }} />,
+      );
+      
+      rerender(() =>
```
nickytonline, line 106: refactor (blocking): The rerender is not necessary.

nickytonline (changes requested):
Figured out why the tests weren't running @Link2Twenty.

app/javascript/shared/components/useKeyboardShortcuts.jsx
nickytonline: Ahh, I must have missed that. My bad. In that case I'd say pass it in, but I'll leave that up to you to decide.

Link2Twenty:
When you know you know 😁
Thanks @nickytonline

nickytonline (approved):
Thanks for being patient as there was a lot of back and forth and thanks so much for the PR @Link2Twenty! 🚀

reobin:
Good job @Link2Twenty !

It seems stable enough. I'll start integrating this in #10468 some time tonight

Link2Twenty:
@nickytonline thanks for the help hopefully I'll have a better understanding of tests now 😅

@reobin thank you, sorry for the hold up 😬

reobin:
Quick question @nickytonline @Link2Twenty : is `shared/components` the right place for this hook? Asking because I'll put my list navigation component in the same place.

nickytonline:
> Quick question @nickytonline @Link2Twenty : is `shared/components` the right place for this hook? Asking because I'll put my list navigation component in the same place.

@reobin, for common components `shared/components` is good place. We might revise this in the future, but for now, pop it there.

Link2Twenty:
Just realised a problem with using `event.code` rather than `event.key.toLowerCase()`
In the uk the @ symbol would appear at `shift+Quote` but in the US it would be `shift+Digit2`.
I'm going to update the code so it can use either, because we may want `crtl+Digit2` to do something but `ctrl+Numpad2` to not. @reobin you shouldn't need to update what you've done I don't think.

reobin:
Pulled the changed and all is well 👍

Link2Twenty:
Hey @Ridhwana any progress on this? 😊

(no rush, of course, just curious)

Ridhwana (approved):
Looks great to me 🚀

PR #10721, opened 2020-10-07
Do not require meta_keywords to be set

vaidehijoshi:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As a part of our ongoing Forem creator onboarding project, we are removing some required fields from the SiteConfig and making them optional. This will help reduce the number of fields that a brand new Forem creator will have to fill out when they set up a new Forem instance.

This PR makes `SiteConfig.meta_keywords` **not required**. These fields can still be configured by Forem creators, but _not_ defining them will not stop them from setting up new instances.

Since a `meta_keyword` may or may not be set, I also did some refactoring in this PR to conditionally render a `<meta />` tag if the appropriate keyword is defined for it. This allows us to avoid having to add these guards in a ton of places in our codebase:
```ruby
<% if SiteConfig.meta_keywords[:default].present? %>
  <meta name="keywords" content="<%= SiteConfig.meta_keywords[:default] %>">
<% end %>
```

and instead allows us to just write this, instead:
```ruby
<%= meta_keywords_default %>
```

The helper will generate and render the `<meta />` tag if the keyword exists, and render nothing if it doesn't! 😉 

🗒️  _A note about future work:_ As part of the onboarding process, we will eventually be asking Forem creators for something called `Community tags`. The end goal is to pre-populate the `meta_keywords` in the SiteConfig with the values of `Community tags` -- but this PR doesn't tackle that, it's just the _set up work_ for that eventual feature (which you can [read more about here](https://github.com/forem/InternalProjectPlanning/issues/113)).

Fixes https://github.com/forem/InternalProjectPlanning/issues/87.

The `meta_keywords` still exist on `/admin/config`, under the `Meta Keywords` section:
Screen Shot 2020-10-07 at 1 19 05 PM

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -216,10 +216,10 @@ module ApplicationHelper
     SiteConfig.community_member_label.pluralize
   end
 
-  def meta_keywords_tag(tag)
+  def meta_keywords_tag(tag_name)
     return if SiteConfig.meta_keywords[:tag].blank?
 
-    tag(:meta, name: "keywords", content: "#{SiteConfig.meta_keywords[:tag]}, #{tag}")
+    tag.meta name: "keywords", content: "#{SiteConfig.meta_keywords[:tag]}, #{tag_name}"
   end
 
   def app_url(uri = nil)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -216,6 +216,12 @@ module ApplicationHelper
     SiteConfig.community_member_label.pluralize
   end
 
+  def meta_keywords_tag(tag)
+    return if SiteConfig.meta_keywords[:tag].blank?
+
+    tag(:meta, name: "keywords", content: "#{SiteConfig.meta_keywords[:tag]}, #{tag}")
+  end
+
   def app_url(uri = nil)
     URL.url(uri)
   end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -216,6 +216,12 @@ module ApplicationHelper
     SiteConfig.community_member_label.pluralize
   end
 
+  def meta_keywords_default
+    return if SiteConfig.meta_keywords[:default].blank?
+
+    tag.meta name: "keywords", content: SiteConfig.meta_keywords[:default]
+  end
+
   def meta_keywords_tag(tag_name)
     return if SiteConfig.meta_keywords[:tag].blank?
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -222,6 +222,18 @@ module ApplicationHelper
     tag.meta name: "keywords", content: SiteConfig.meta_keywords[:default]
   end
 
+  def meta_keywords_article(article_tags = nil)
+    return if SiteConfig.meta_keywords[:article].blank?
+
+    content = if article_tags.present?
+                "#{article_tags}, #{SiteConfig.meta_keywords[:article]}"
+              else
+                SiteConfig.meta_keywords[:article]
+              end
+
+    tag.meta name: "keywords", content: content
+  end
+
   def meta_keywords_tag(tag_name)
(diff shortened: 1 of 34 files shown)


app/views/admin/configs/show.html.erb
```diff
@@ -642,7 +642,7 @@
                 <%= f.fields_for :meta_keywords do |meta_keywords_field| %>
                   <% SiteConfig.meta_keywords.each do |area, keywords| %>
                     <div>
-                      <%= meta_keywords_field.label "#{area}*" %>
```
vaidehijoshi, line 645: I wasn't sure why we had a `*` appended to these, but it sort of makes it look like they're required fields. I removed this since it could potentially be confusing and didn't seem super informative
Screen Shot 2020-10-07 at 1 31 02 PM

app/helpers/application_helper.rb
```diff
@@ -216,6 +216,30 @@ def community_members_label
...
+              end
+
+    tag.meta name: "keywords", content: content
+  end
+
+  def meta_keywords_tag(tag_name)
+    return if SiteConfig.meta_keywords[:tag].blank?
+
+    tag.meta name: "keywords", content: "#{SiteConfig.meta_keywords[:tag]}, #{tag_name}"
+  end
```
vaidehijoshi, line 241: I considered making one single `meta_keywords` helper method, but I actually found that personally much harder to read. I like the idea of having these helpers be keyword-specific (at least to start!).

app/helpers/application_helper.rb
```diff
@@ -216,6 +216,30 @@ def community_members_label
     SiteConfig.community_member_label.pluralize
   end
 
+  def meta_keywords_default
```
atsmith813, line 219: **suggestion (non-blocking)** what if we explored moving these into their own helper that gets included in the `ApplicationHelper`? If we started that trend, then at some point in a separate PR we can follow suit and clean up this file a bit. I could almost see us doing the same for other "groups" of methods that are in this file, like the [`url` methods](https://github.com/forem/forem/blob/1cd89399ea53697e37bb2cc8aa5f00ec923fd350/app/helpers/application_helper.rb#L243-L269). That approach may even give us more flexibility be only including `MetaTagHelper` in areas instead of `ApplicationHelper`.

atsmith813 (approved):

app/helpers/application_helper.rb
citizen428: @atsmith813 Because of how Rails works, this would only reduce the amount of added methods in places where we explicitly include helpers (e.g. service objects). We could however wrap them up in some little helper object:

```ruby
module MetaKeywords
  extend ActionView::Helpers::TagHelper

  def self.default
    return if SiteConfig.meta_keywords[:default].blank?
  
    tag.meta name: "keywords", content: SiteConfig.meta_keywords[:default]
  end

  ...
end

MetaKeywords.default
#=> "<meta name=\"keywords\" content=\"test\">"
MetaKeywords.default.html_safe?
#=> true
```

This completely takes them out of the global helper namespace and therefore reduces the surface arae of `ApplicationHelper`, but is still easy to use in views:

```ruby
<%= MetaKeywords.default %>
```

Zhao-Andy (approved):
LGTM!

app/helpers/application_helper.rb
vaidehijoshi: Hm, that's an interesting idea, @citizen428! FWIW, @atsmith813, I started down this path based on [@citizen428's suggestion here](https://github.com/forem/forem/pull/10623#discussion_r499973586), but I see your point as to how the `ApplicationHelper` could start/has started to become a bit unwieldy. I think I'll be back in this SiteConfig code again later this cycle; maybe at that point, I can investigate how it feels to wrap these into a helper object and see if it's worth the refactor effort :)

PR #10731, opened 2020-10-08
Only show setup completed banner to single resource admins

citizen428:

- [X] Refactor

## Description

Yesterday we deployed a [PR](https://github.com/forem/forem/pull/10671) that made a `SiteConfig` option to the mandatory configs defined in `VerifySetupCompleted`, without ensuring that this value is actually set on DEV. This led to a brief period where all users could see the following banner:

image

This PR updates the verification code to only display the banner to single resource admins. They are the only ones who actually can set the missing values, so the banner offers little benefit to other users since there's nothing actionable for them (except for maybe pinging an admin). 

n/a

* Ensure one of the mandatory configs (e.g. `logo_png` or `logo_svg`) is `nil`.
* Visit the site with a non-admin user, you shouldn't see a banner.
* Visit the site with a single resource admin (`user.add_role(:single_resource_admin)`), you should now see the banner.

## Added tests?

- [X] yes, improved and added

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/app/controllers/concerns/verify_setup_completed.rb
+++ b/app/controllers/concerns/verify_setup_completed.rb
@@ -21,9 +21,9 @@ module VerifySetupCompleted
   ].freeze
 
   included do
-    # rubocop:disable Rails/LexicallyScopedActionFilter
-    before_action :verify_setup_completed, only: %i[index new edit show]
-    # rubocop:enable Rails/LexicallyScopedActionFilter
+    before_action :verify_setup_completed,
+                  only: %i[index new edit show],
+                  if: -> { current_user&.has_role?(:single_resource_admin) }
   end
 
   def setup_completed?
--- a/spec/system/admin/admin_manages_configuration_spec.rb
+++ b/spec/system/admin/admin_manages_configuration_spec.rb
@@ -1,40 +1,70 @@
 require "rails_helper"
 
 RSpec.describe "Admin manages configuration", type: :system do
-  let(:admin) { create(:user, :super_admin) }
+  let(:single_resource_admin) { create(:user, :single_resource_admin) }
 
-  before do
-    sign_in admin
-    visit admin_config_path
-  end
+  context "when on the config page" do
+    before do
+      sign_in single_resource_admin
+      visit admin_config_path
+    end
 
-  # Note: The :meta_keywords are handled slightly differently in the view, so we
-  # can't check them the same way as the rest.
-  (VerifySetupCompleted::MANDATORY_CONFIGS - [:meta_keywords]).each do |option|
-    it "marks #{option} as required" do
-      selector = "label[for='site_config_#{option}']"
-      expect(first(selector).text).to include("Required")
+    # Note: The :meta_keywords are handled slightly differently in the view, so we
+    # can't check them the same way as the rest.
+    (VerifySetupCompleted::MANDATORY_CONFIGS - [:meta_keywords]).each do |option|
+      it "marks #{option} as required" do
+        selector = "label[for='site_config_#{option}']"
+        expect(first(selector).text).to include("Required")
+      end
     end
   end
 
-  context "when mandatory options are missing" do
-    it "does not show the banner on the config page" do
-      allow(SiteConfig).to receive(:tagline).and_return(nil)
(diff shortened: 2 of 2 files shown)


app/controllers/concerns/verify_setup_completed.rb
```diff
@@ -21,9 +21,9 @@ module VerifySetupCompleted
   ].freeze
 
   included do
-    # rubocop:disable Rails/LexicallyScopedActionFilter
-    before_action :verify_setup_completed, only: %i[index new edit show]
-    # rubocop:enable Rails/LexicallyScopedActionFilter
+    before_action :verify_setup_completed,
+                  only: %i[index new edit show],
+                  if: -> { current_user&.has_role?(:single_resource_admin) }
```
rhymes, line 26: Unsure if this is the right role to check for, maybe `admin` as well ?

I don't recall exactly what's the sequence of steps for a new forem

citizen428: The problem is that admins can't edit the config. So we'd show them a CTA for an action they can't perform.

rhymes: ah ok! thought admin/super admins could edit configuration without being single resource admins

rhymes: ah ok! thought admin/super admins could edit configuration without being single resource admins. makes sense to keep the two things separate

rhymes (approved):

app/controllers/concerns/verify_setup_completed.rb
benhalpern: So the way we currently have it is that you need to be a `super_admin` _and_ a `single_resource_admin, "Config"`, but I think that _might_ be overkill. We did it that way out of an abundance of caution, but realistically, `super_admin` should probably be enough. It would probably help to improve visualization of impending changes on an "are you sure" step, along with better safeguards on inputs that are particularly "dangerous".

..... But on this PR, I don't think this approach will work due to the caching of this page such that `current_user` would leak... But what if we moved this functionality into the `broadcast` area, which would DRY up the code anyway and allow more flexibility here, as that banner loads async.

benhalpern (changes requested):

citizen428:
> ..... But on this PR, I don't think this approach will work due to the caching of this page such that current_user would leak...

I'm not sure I understand, can you elaborate a bit? Where are we caching `current_user` in a way that it leaks between requests?

> But what if we moved this functionality into the broadcast area, which would DRY up the code anyway 

Right now it's in the `global_notification` flash. I'm a bit confused what you mean by DRY here, we only have it in one place. 😕

benhalpern:
- @citizen428 unless I'm mistaken by what's goin on, this banner shows up within the html of the initial request, so if that request is cached _at the edge_, we will take whatever the html is output as and return the same html for the next request. So any page with html cached at the edge can't have `current_user`.

citizen428:
@benhalpern The banner itself is a flash message, which does not rely on current_user:

```ruby
    link = helpers.link_to("the configuration page", admin_config_path, "data-no-instant" => true)
    flash[:global_notice] = "Setup not completed yet, missing #{missing_configs_text}. Please visit #{link}.".html_safe
```

But I probably just don't get what you're saying 😞 Anyway, I'm still not sure how the broadcast area would DRY this up (it's only in a single place right now anyway) and am starting to wonder if maybe just leave it as is? Or if you think it's worth making the change, feel free to just push directly into this branch, maybe I'll get it once I see some code.

citizen428:
[This comment](https://github.com/forem/forem/pull/10926#issuecomment-712213076) by @atsmith813 in another PR cleared up the confusion for me, I'll dwell on this for a bit.

msarit:
@rhymes @citizen428 @benhalpern Hey fellas, drawing your attention to this PR.
Did we decide on an approach moving forward?
Please let me (or current OSS on call) know 🙏🏾

citizen428:
@msarit To be honest I haven't quite decided if I want to abandon this or try to solve it, but I think we can close this PR for now I still have a local version of this branch in case I want to revisit this.

PR #10732, opened 2020-10-08
Remove Twitch integration

citizen428:

- [X] Refactor

## Description

During profile generalization work I found two columns on `User` I was unsure about, `currently_streaming_on` and `twitch_username`. According to @benhalpern we can remove this functionality since we never quite got it to work and now have the Twitch Liquid tag instead.

n/a

All the specs should still pass, nothing else.

## Added tests?

- [X] updated/removed as necessary

## Added to documentation?

- [X] no documentation needed

Screen Shot 2020-10-08 at 13 42 49


rhymes (approved):
Cool! Definitely +1 on removing unused features.

Should we ignore the two columns in this PR and then remove them in a future one?

citizen428:
> Should we ignore the two columns in this PR and then remove them in a future one? 

We totally should, I already have the change locally, it's just not pushed yet :-)

benhalpern (approved):

PR #10733, opened 2020-10-08
/profile/settings page rendering (Ridhwana/Michael)

citizen428:
_The original PR with comments was https://github.com/forem/forem/pull/10272, but Ridwhana had some issues with the checkout so we started a new branch._

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The profile settings page now renders dynamically from the `ProfileFieldGroup` model and `ProfileField` model. 
Updates to `ProfilesController` are sent to `/profiles/:id`. We re-use the `Profiles::Update` service to split the user and the generalized profile fields to update the relevant models. 

Need to:

- [x] Add tests 
- [x] Fix the checkboxes and styling on the ui
- [x] Fix the checkboxes, it's currently not working 
- [x] Fix the upload of the profile image (move the validation etc. as well)
- [x] Show feedback on the page when success or errors occurs
- [x] Need to check the success of both the profile updates and the user updates

Part of #6994

1. Go to `/settings/profile`
2. Edit various attributes and observe that the changes are reflected on the page as expected.
3. Optional: use the Rails console to verify that the attributes are synced between users and profiles.

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

alt_text


Ridhwana:
We had to create a new PR as the [previous branch](https://github.com/forem/forem/pull/10272) was orphaned due to a deletion of the original fork. 

I'm dropping a summary of the comments from there that still need some addressing: 

@citizen428 : 

Note: I now disabled `spec/system/user_selects_looking_for_work_spec.rb`. I first tried to fix it, then realized this is called from `UsersController#update`, which the current form doesn't hit. I also added a comment about this method in general and started a related Slack discussion:

```
# TODO: [@forem/oss] This doesn't make sense for non DEV communities. We
# should either remove it or special case it.
```

@Ridhwana I noticed there are some other things happening in `UsersController#update` which we probably need to move, e.g. 

```ruby

if @user.export_requested?
  notice += " The export will be emailed to you shortly."
    ExportContentWorker.perform_async(@user.id)
  end
end

cookies.permanent[:user_experience_level] = @user.experience_level.to_s if @user.experience_level.present?
```

We should talk about this a bit more, I don't think everything needs to go into the `Profile::Update` service, i.e. content export could be its own form that triggers a different service.

__________________________________________________________________________

@Ridhwana I'm wasn't quite sure about the following two todo items:

* "Fix the checkboxes, it's currently not working":  They seem to work fine when I tried them locally, so I marked this as done. Please confirm.
* "Fix the upload of the profile image (move the validation etc. as well)": this also does seem to work, but I left it as for now since I don't fully understand what you wanted to move.

___________________________________________________________________________

@Ridhwana I  made the following changes now:

1. Make `sync_to_user` conditional on `update_profile`: `update_profile` now directly returns the result from `@profile.save` and the two are combined by `update_profile && sync_to_user` so the sync only happens if the profile update succeeds. Also `sync_to_user` now returns a boolena too (the value of `@success`).
2. The combined result of `update_profile && sync_to_user` is used to determine if we call the new method `perform_after_update_actions` or report errors to Honeycomb (this part was taken and slightly modified from `UsersController#update`).
3. `perform_after_update_actions` does two things for now: enqueue the RSS update job and touch the `profile_updated_at` attribute of the user.
4. I added setting the experience level in the cookie to `ProfilesController`.

Things in `UsersController#update` that are not taken care of (yet):

* Some tab handling logic because I'm not entirely sure how this works
* The `unless valid_image?` profile image check
* The `config_changed?` check (which just outputs a different notice if the theme was changed). 
* The export. I still think we should make this a different section in the UI and handle it from a different controller action.
* Following the hiring tag. I think we might be able to just get rid of this, see the comment in the code.

Ridhwana:
> @Ridhwana I noticed there are some other things happening in UsersController#update which we probably need to move, e.g.
> 
> if @user.export_requested?
>   notice += " The export will be emailed to you shortly."
>     ExportContentWorker.perform_async(@user.id)
>   end
> end
> 

@citizen428 the miscellaneous tab has a form that updates /:user/:id. The piece of code below in the user controller update action gets triggered from that form in the miscellaneous tab. 
```
if @user.export_requested?
  notice += " The export will be emailed to you shortly."
    ExportContentWorker.perform_async(@user.id)
  end
end
```
Hence, I don't feel like we need to move this code. I see it behaving in the same manner that it did before we moved the profile form to hit our profile controller. Perhaps I'm missing something though, can you help me understand why you said we should move it?

Ridhwana:
> @Ridhwana I'm wasn't quite sure about the following two todo items:
> 
> "Fix the checkboxes, it's currently not working": They seem to work fine when I tried them locally, so I marked this as done. Please confirm.

@citizen428 they weren't showing up on the UI because I was setting value to true instead of checked. It's fixed now.

> "Fix the upload of the profile image (move the validation etc. as well)": this also does seem to work, but I left it as for now since I don't fully understand what you wanted to move.

I was talking about the valid_image? profile image check. You mention it in the list you made as well.

app/controllers/profiles_controller.rb
```diff
@@ -1,13 +1,27 @@
...
   end
 
   private
 
   def update_params
-    params.require(:profile).permit(*Profile.attributes)
+    params.permit(profile: Profile.attributes!, user: ALLOWED_USER_PARAMS)
+  end
+
+  def update_cookie_experience_level
```
Ridhwana, line 22: @citizen428 why do we update this in the profiles controller. Wouldn't it still need to be updated in the users controller as on the UX tab we still updating the users controller?

citizen428:
@Ridhwana 

> Hence, I don't feel like we need to move this code.

You're probably correct here, it was a long week and I confused myself about the relationship between the two controllers. The way the settings page has several tabs that all use the same controller makes things a bit complicated, but I get it now. This also answers your question about the experience level cookie, same confusion between the two controllers/which tab needs what, it shouldn't be in the profiles controller (and the RSS fetcher shouldn't be in the service object, since `feed_url` is also a different tab.

I'll soon push a commit that does the following:

* Remove the experience level cookie from `ProfilesController`
* Remove the `ReaderFetchUserWorker` from `Profiles::Update`
* Simplify `Profiles::Update` a bit since it now needs to do less again.

app/services/profiles/update.rb
```diff
@@ -12,7 +12,7 @@ def initialize(user, updated_attributes)
       @user = user
       @profile = user.profile
       @updated_profile_attributes = updated_attributes[:profile] || {}
-      @updated_user_attributes = updated_attributes[:user] || {}
+      @updated_user_attributes = updated_attributes[:user].to_h || {}
```
Ridhwana, line 15: @citizen428 I did test this change, but wanted to bring it to your attention that I moved the `to_h` from the `update_user_attributes` method to this line in the initialize. That way we could use  `@updated_user_attributes[:profile_image]` instead of `@updated_user_attributes["profile_image"]` which made the testing on profile images simpler.

app/services/profiles/update.rb
```diff
@@ -1,52 +1,102 @@
...
-    end
-
-    def success?
-      @success
+      @profile.save
     end
 
-    private
-
+    # Propagate changes back to the `users` table
```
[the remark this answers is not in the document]
Ridhwana: @benhalpern this will probably useful for you to see in terms of how we sync the profile data back to the user table. 

In addition, regarding backfilling, @citizen428 wrote a data_update script (which can be found in lib⁩/data_update_scripts /⁩20200819025131_migrate_profile_data.rb) that migrates the profile data for older users. You'll also notice a script BackfillProfileSkillsLanguages in this PR that does the same for a field that was missed.

benhalpern (approved):
Everything seems good here to my eye.

pkfrank:
Looping in @AfroDevGirl, an early Forem Admin.

She is very interested in using this feature as part of setting up her community, and might be able to get involved as an OSS Contributor as well.  She just had this PR merged earlier today: https://github.com/forem/forem/pull/10870

Any particular areas that we need extra help?

citizen428:
@pkfrank Not right now, at least not in a way that's easy to handover. First, we need to get this PR merged, as it's been open for way too long and is blocking some of my other work. Once that's unblocked there will soon(ish) be a point where we can involve the wider community a bit more if we choose to do that.

atsmith813 (approved):
Tested locally as well - let's do it 🚀

Ridhwana (approved):

PR #10736, opened 2020-10-08
Add recaptcha keys to as site-configurable keys

Zhao-Andy:
- [x] Feature

## Description
This adds Google ReCAPTCHA keys into `SiteConfig`, so that Forem admins are able to add ReCAPTCHA without having to go through the `.env` file.

1. Visit /admin/config
2. Find the Google ReCAPTCHA field
3. Update the field to anything
4. Verify that it updated

Screen Shot 2020-10-08 at 9 35 05 AM

## Added tests?
- [x] yes

## Added to documentation?
- [x] to add to forem admin docs

No

Lorelai Gilmore says, "I assuming you have documentation."

### DIFF
--- a/app/controllers/feedback_messages_controller.rb
+++ b/app/controllers/feedback_messages_controller.rb
@@ -9,7 +9,7 @@ class FeedbackMessagesController < ApplicationController
     params = feedback_message_params.merge(reporter_id: current_user&.id)
     @feedback_message = FeedbackMessage.new(params)
 
-    if recaptcha_verified? && @feedback_message.save
+    if (recaptcha_disabled? || recaptcha_verified?) && @feedback_message.save
       Slack::Messengers::Feedback.call(
         user: current_user,
         type: feedback_message_params[:feedback_type],
@@ -30,9 +30,11 @@ class FeedbackMessagesController < ApplicationController
 
   private
 
-  def recaptcha_verified?
-    return true if SiteConfig.recaptcha_site_key.blank? && SiteConfig.recaptcha_secret_key.blank?
+  def recaptcha_disabled?
+    SiteConfig.recaptcha_site_key.blank? && SiteConfig.recaptcha_secret_key.blank?
+  end
 
+  def recaptcha_verified?
     recaptcha_params = { secret_key: SiteConfig.recaptcha_secret_key }
     params["g-recaptcha-response"] && verify_recaptcha(recaptcha_params)
   end
--- a/app/controllers/feedback_messages_controller.rb
+++ b/app/controllers/feedback_messages_controller.rb
@@ -31,7 +31,7 @@ class FeedbackMessagesController < ApplicationController
   private
 
   def recaptcha_verified?
-    recaptcha_params = { secret_key: ApplicationConfig["RECAPTCHA_SECRET"] }
+    recaptcha_params = { secret_key: SiteConfig.recaptcha_secret_key }
     params["g-recaptcha-response"] && verify_recaptcha(recaptcha_params)
   end
 
--- a/app/controllers/feedback_messages_controller.rb
+++ b/app/controllers/feedback_messages_controller.rb
@@ -31,6 +31,8 @@ class FeedbackMessagesController < ApplicationController
   private
 
   def recaptcha_verified?
+    return true if SiteConfig.recaptcha_site_key.blank? && SiteConfig.recaptcha_secret_key.blank?
+
     recaptcha_params = { secret_key: SiteConfig.recaptcha_secret_key }
     params["g-recaptcha-response"] && verify_recaptcha(recaptcha_params)
   end
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -215,6 +215,14 @@ module Constants
         description: "Determines the mininum for the periodic email digest",
         placeholder: 2
       },
+      recaptcha_site_key: {
+        description: "Site key for Google ReCAPTCHA, used for reporting abuse.",
+        placeholder: "..."
+      },
+      recaptcha_secret_key: {
+        description: "Secret key for Google ReCAPTCHA, used for reporting abuse.",
+        placeholder: "..."
+      },
       right_navbar_svg_icon: {
         description: "The SVG icon used to expand the right navbar navigation menu. Should be a max of 24x24px.",
         placeholder: "<svg ...></svg>"
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -220,7 +220,7 @@ module Constants
         placeholder: "..."
       },
       recaptcha_secret_key: {
-        description: "Secret key for Google ReCAPTCHA, used for reporting abuse.",
+        description: "Secret key for Google reCAPTCHA, used for reporting abuse.",
         placeholder: "..."
       },
       right_navbar_svg_icon: {
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -216,7 +216,7 @@ module Constants
         placeholder: 2
       },
       recaptcha_site_key: {
-        description: "Site key for Google ReCAPTCHA, used for reporting abuse.",
+        description: "Site key for Google reCAPTCHA, used for reporting abuse.",
         placeholder: "..."
       },
       recaptcha_secret_key: {
--- a/app/views/admin/configs/show.html.erb
+++ b/app/views/admin/configs/show.html.erb
@@ -474,14 +474,14 @@
                        } %>
             <div id="recaptchaContainer" class="card-body collapse hide" aria-labelledby="recaptchaContainer">
               <div class="form-group">
-                <%= admin_config_label :recaptcha_site_key, "ReCAPTCHA Site Key" %>
+                <%= admin_config_label :recaptcha_site_key, "reCAPTCHA Site Key" %>
                 <%= f.text_field :recaptcha_site_key,
                                  class: "form-control",
                                  value: SiteConfig.recaptcha_site_key %>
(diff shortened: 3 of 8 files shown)


app/views/admin/configs/show.html.erb
```diff
@@ -464,6 +464,32 @@
             </div>
           </div>
 
+          <div class="card mt-3">
```
Zhao-Andy, line 467: I just stuck this under `Google Analytics` because alphabetical, but maybe we want to make it a required field in the first Forem setup steps?

Zhao-Andy:
Converted to draft while I add some additional things.

benhalpern:
Looks good to me so far.... I guess this means recaptcha should probably be _mandatory?_ Because certain parts of the site just won't work if this is wrong.

Let's make sure we cover ample instructions. (E.g. v2 vs v3)

Zhao-Andy:
@benhalpern we started a discussion in Slack and in a different repo, but basically I made it optional and added to the admin docs with instructions.

app/controllers/feedback_messages_controller.rb
```diff
@@ -31,7 +31,9 @@ def create
   private
 
   def recaptcha_verified?
-    recaptcha_params = { secret_key: ApplicationConfig["RECAPTCHA_SECRET"] }
+    return true if SiteConfig.recaptcha_site_key.blank? && SiteConfig.recaptcha_secret_key.blank?
```
Ridhwana, line 34: @Zhao-Andy perhaps its late and i'm tired but isn't recaptcha_verified? supposed to return false if the SiteConfig.recaptcha_site_key or secret key is blank?

Zhao-Andy: All good :)

I want to return true because if they're blank, we're assuming they don't want to use / haven't set recaptcha. This allows them to submit abuse reports without recaptcha.

I was thinking maybe this should be another SiteConfig variable that's just a boolean, like `SiteConfig.recaptcha_enabled` but that seemed unnecessary.

app/lib/constants/site_config.rb
```diff
@@ -215,6 +215,14 @@ module SiteConfig
         description: "Determines the mininum for the periodic email digest",
         placeholder: 2
       },
+      recaptcha_site_key: {
+        description: "Site key for Google ReCAPTCHA, used for reporting abuse.",
```
juliannatetreault, line 219: _Small, absolutely non-blocking nitpick_: I think it would be nice to, anywhere the "r" in reCAPTCHA is capitalized, change it to be lowercase to better follow the capitalization in Google's reCAPTCHA docs and our own internal docs that you wrote. 😊  
```suggestion
        description: "Site key for Google reCAPTCHA, used for reporting abuse.",
```

app/controllers/feedback_messages_controller.rb
Ridhwana: **(non-blocking, suggestion)**: hmm I see what you mean. I think I was confused by the method name vs. what we doing in that specific line of code. When we return true from that line, it's not because it was verified but rather because we don't want to use recaptcha which makes the naming of the function confusing. 

Perhaps there's some way to make this easier to read, like extracting that condition into its own descriptive function.

Zhao-Andy: Updated!

app/views/admin/configs/show.html.erb
Zhao-Andy: Actually I think it would be easier to find as a separate section, and also because I think it's more clear in the docs if these sections are separate. We can revisit in the future though!

app/lib/constants/site_config.rb
```diff
@@ -215,6 +215,14 @@ module SiteConfig
         description: "Determines the mininum for the periodic email digest",
         placeholder: 2
       },
+      recaptcha_site_key: {
+        description: "Site key for Google reCAPTCHA, used for reporting abuse.",
+        placeholder: "..."
+      },
+      recaptcha_secret_key: {
+        description: "Secret key for Google ReCAPTCHA, used for reporting abuse.",
```
juliannatetreault, line 223: _Non-blocking, nitpick_:
```suggestion
        description: "Secret key for Google reCAPTCHA, used for reporting abuse.",
```

app/views/admin/configs/show.html.erb
```diff
@@ -464,6 +464,32 @@
...
+            <%= render partial: "card_header",
+                       locals: {
+                         header: "Google ReCAPTCHA Keys",
+                         state: "collapse",
+                         target: "recaptchaContainer",
+                         expanded: "false"
+                       } %>
+            <div id="recaptchaContainer" class="card-body collapse hide" aria-labelledby="recaptchaContainer">
+              <div class="form-group">
+                <%= admin_config_label :recaptcha_site_key, "ReCAPTCHA Site Key" %>
```
juliannatetreault, line 477: _Non-blocking, nitpick_:
```suggestion
                <%= admin_config_label :recaptcha_site_key, "reCAPTCHA Site Key" %>
```

app/views/admin/configs/show.html.erb
```diff
@@ -464,6 +464,32 @@
...
+                <%= f.text_field :recaptcha_site_key,
+                                 class: "form-control",
+                                 value: SiteConfig.recaptcha_site_key %>
+                <div class="alert alert-info"><%= Constants::SiteConfig::DETAILS[:recaptcha_site_key][:description] %></div>
+              </div>
+              <div class="form-group">
+                <%= admin_config_label :recaptcha_secret_key, "ReCAPTCHA Secret Key" %>
```
juliannatetreault, line 484: _Non-blocking, nitpick_:
```suggestion
                <%= admin_config_label :recaptcha_secret_key, "reCAPTCHA Secret Key" %>
```

app/views/admin/configs/show.html.erb
```diff
@@ -464,6 +464,32 @@
             </div>
           </div>
 
+          <div class="card mt-3">
+            <%= render partial: "card_header",
+                       locals: {
+                         header: "Google ReCAPTCHA Keys",
```
juliannatetreault, line 470: _Non-blocking, nitpick_:
```suggestion
                         header: "Google reCAPTCHA Keys",
```

app/views/admin/configs/show.html.erb
```diff
@@ -464,6 +464,32 @@
...
+            <div id="recaptchaContainer" class="card-body collapse hide" aria-labelledby="recaptchaContainer">
+              <div class="form-group">
+                <%= admin_config_label :recaptcha_site_key, "ReCAPTCHA Site Key" %>
+                <%= f.text_field :recaptcha_site_key,
+                                 class: "form-control",
+                                 value: SiteConfig.recaptcha_site_key %>
+                <div class="alert alert-info"><%= Constants::SiteConfig::DETAILS[:recaptcha_site_key][:description] %></div>
+              </div>
+              <div class="form-group">
+                <%= admin_config_label :recaptcha_secret_key, "ReCAPTCHA Secret Key" %>
```
juliannatetreault, line 484: _Non-blocking, nitpick_:
```suggestion
                <%= admin_config_label :recaptcha_secret_key, "ReCAPTCHA Secret Key" %>
```

app/views/admin/configs/show.html.erb
```diff
@@ -464,6 +464,32 @@
...
+            <%= render partial: "card_header",
+                       locals: {
+                         header: "Google reCAPTCHA Keys",
+                         state: "collapse",
+                         target: "recaptchaContainer",
+                         expanded: "false"
+                       } %>
+            <div id="recaptchaContainer" class="card-body collapse hide" aria-labelledby="recaptchaContainer">
+              <div class="form-group">
+                <%= admin_config_label :recaptcha_site_key, "ReCAPTCHA Site Key" %>
```
juliannatetreault, line 477: _Non-blocking, nitpick_:
```suggestion
                <%= admin_config_label :recaptcha_site_key, "reCAPTCHA Site Key" %>
```

juliannatetreault (commented):
Left a few small, non-blocking comments, but otherwise, everything is looking great!

Ridhwana (approved):
This looks great! Thanks @Zhao-Andy 🚀

juliannatetreault (approved):
Thanks for making these changes and writing the thorough docs for this. I tested everything locally and it worked great! 🎉

app/controllers/feedback_messages_controller.rb
```diff
@@ -30,8 +30,12 @@ def create
 
   private
 
+  def recaptcha_disabled?
+    SiteConfig.recaptcha_site_key.blank? && SiteConfig.recaptcha_secret_key.blank?
+  end
+
   def recaptcha_verified?
-    recaptcha_params = { secret_key: ApplicationConfig["RECAPTCHA_SECRET"] }
+    recaptcha_params = { secret_key: SiteConfig.recaptcha_secret_key }
```
jacobherrington, line 38: 😍

app/controllers/feedback_messages_controller.rb
```diff
@@ -30,8 +30,12 @@ def create
 
   private
 
+  def recaptcha_disabled?
+    SiteConfig.recaptcha_site_key.blank? && SiteConfig.recaptcha_secret_key.blank?
```
jacobherrington, line 34: What happens if one of these is set and not the other? Does recaptcha need both?

jacobherrington (commented):
Looks good! One question, but it's probably not blocking.

app/controllers/feedback_messages_controller.rb
Zhao-Andy: Yeah, it needs both. The `site_key` is used to display the tag, and the `secret_key` is used to validate the request.

jacobherrington: So should this be `SiteConfig.recaptcha_site_key.blank? || SiteConfig.recaptcha_secret_key.blank?`

Issue #10629, opened 2020-10-06
Text goes outside the block (app)


mityaua:
**Describe the bug**

In mobile app (Android 10) text in bio (look like url type) goes outside the block.

**To Reproduce**

Open app on device with 393px viewport.

1. Go to mobile app on Android 10 (device width 393px)
2. Click on any post by https://dev.to/vaibhavkhulbe
3. Scroll down to end of article
4. See error https://prnt.sc/utvn2p

**Screenshots**

https://prnt.sc/utvn2p

**Smartphone (please complete the following information):**

- Device: MI 9 LITE
- OS, version: Android 10 (MIUI 12.0.3)
- Browser, version: App, 1.4.2

Link2Twenty:
Same is true on the mobile web version

PR #10740, opened 2020-10-08
Long text goes outside cards

Link2Twenty:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Add word-break to crayons-card this will allow words to be broken if the word is too long but will wrap normally otherwise.

closes: #10629

forem__resize-bug

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

N/A

fix it with love

### DIFF
--- a/app/assets/stylesheets/components/cards.scss
+++ b/app/assets/stylesheets/components/cards.scss
@@ -5,7 +5,7 @@
   background: var(--card-bg);
   color: var(--card-color);
   box-shadow: 0 0 0 1px var(--card-border);
-  word-break: break-word;
+  overflow-wrap: anywhere;
 
   &--secondary {
     background: var(--card-secondary-bg);
--- a/app/assets/stylesheets/components/cards.scss
+++ b/app/assets/stylesheets/components/cards.scss
@@ -5,6 +5,7 @@
   background: var(--card-bg);
   color: var(--card-color);
   box-shadow: 0 0 0 1px var(--card-border);
+  word-break: break-word;
 
   &--secondary {
     background: var(--card-secondary-bg);


app/assets/stylesheets/components/cards.scss
```diff
@@ -5,6 +5,7 @@
   background: var(--card-bg);
   color: var(--card-color);
   box-shadow: 0 0 0 1px var(--card-border);
+  word-break: break-word;
```
reobin, line 8: Hey @Link2Twenty !

Just dropping [this (mdn word-break docs)](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) here.

Quotes about `word-break: break-word`:
* "This deprecated API should no longer be used, but will probably still work"
* Solution: "Has the same effect as `word-break: normal` and `overflow-wrap: anywhere`, regardless of the actual value of the `overflow-wrap` property."

Link2Twenty: Good catch 😅

ludwiczakpawel:
I'll play with it locally since it touches `.crayons-card` and that being used like everywhere  :D. 

Also out of curiosity, why not `overflow-wrap: break-word;`?

Link2Twenty:
No real reason, all you gain with everywhere is the ability to use `min-content` and `max-content` as widths.

[example](https://jsfiddle.net/link2twenty/Lyeaxmk3/)

ludwiczakpawel (approved):
Cheers!

benhalpern (approved):

PR #10742, opened 2020-10-08
Move Imgproxy endpoint config to Images::Optimizer

maestromac:
- [x] Bug Fix

## Description
We cannot set `Imgproxy.config.endpoint = URL.url("/images")` in the initializer because it will use the hardcoded `ApplicationConfig["APP_DOMAIN"]` instead of `SiteConfig.app_domain`. Instead, I have to move it closer to the surface, where the App (mainly `URL` module) is completely booted.
n/a
- Makes sure the app boots correctly.
- For a thorough imgproxy setup guide, please see https://github.com/forem/forem/pull/9909
## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a

### DIFF
--- a/app/services/images/optimizer.rb
+++ b/app/services/images/optimizer.rb
@@ -35,7 +35,7 @@ module Images
 
     def self.imgproxy(img_src, **kwargs)
       options = DEFAULT_IMGPROXY_OPTIONS.merge(kwargs).reject { |_, v| v.blank? }
-      Imgproxy.config.endpoint ||= set_imgproxy_endpoint
+      Imgproxy.config.endpoint ||= get_imgproxy_endpoint
       Imgproxy.url_for(img_src, options)
     end
 
@@ -43,7 +43,7 @@ module Images
       Imgproxy.config.key.present? && Imgproxy.config.salt.present?
     end
 
-    def self.set_imgproxy_endpoint
+    def self.get_imgproxy_endpoint
       if Rails.env.production?
         # Use /images with the same domain on Production as
         # our default configuration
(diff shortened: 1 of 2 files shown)


app/services/images/optimizer.rb
```diff
@@ -35,12 +35,26 @@ def self.cloudinary(img_src, **kwargs)
...
+      Imgproxy.config.endpoint ||= set_imgproxy_endpoint
       Imgproxy.url_for(img_src, options)
     end
 
     def self.imgproxy_enabled?
       Imgproxy.config.key.present? && Imgproxy.config.salt.present?
     end
+
+    def self.set_imgproxy_endpoint
+      if Rails.env.production?
```
maestromac, line 47: We no longer need to check for `ApplicationConfig["APP_DOMAIN"] && ApplicationConfig["APP_PROTOCOL"]` because at this point the App is completely booted with the proper domain and protocol set. I've previously added that to fix container building issue.

Zhao-Andy (approved):
Hmmmmmm interesting, seems like it works. 👍

app/services/images/optimizer.rb
```diff
@@ -35,12 +35,26 @@ def self.cloudinary(img_src, **kwargs)
...
-
+      Imgproxy.config.endpoint ||= set_imgproxy_endpoint
       Imgproxy.url_for(img_src, options)
     end
 
     def self.imgproxy_enabled?
       Imgproxy.config.key.present? && Imgproxy.config.salt.present?
     end
+
+    def self.set_imgproxy_endpoint
```
citizen428, line 46: **nitpick (non-blocking):** The method doesn't actually set the endpoint, it just returns it.

citizen428 (approved):

app/services/images/optimizer.rb
maestromac: that's a great point. let me update it to `get_imgproxy_endpoint`

PR #10745, opened 2020-10-08
[Email/Password Authentication] Allow Admins to enable email/password login

msarit:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR gives Admins the ability to enable/disable email login on the user-login page. When enabled, the option to login with email and password is displayed. When disabled, email/password login is not present.

1. Using 2 incognito browser windows, login as an `admin` user on window1.
2. Navigate to `/admin/config` and toggle `"Allow email password login"` option ON and save configuration.

enable-email-login

3. In window2, navigate to `http://localhost:xxxx/enter`. You should see the option to login using email and password.

email-login-enabled

4. As `admin` user, toggle `"Allow email password login"` option OFF and save configuration.

disable-email-login

5. In window2, navigate to `http://localhost:xxxx/enter`. You should NOT see the option to login using email and password.

email-login-disabled

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

NO

AGT Contestant Speaking

Note: closes https://github.com/forem/InternalProjectPlanning/issues/114


app/views/devise/registrations/_registration_form.html.erb
```diff
@@ -22,40 +22,14 @@
...
         <div class="registration__actions-email" id="sign-in-password-form">
-          <%# if there are no SSO auths enabled, we don't ask the user if they have a password %>
-          <% if authentication_enabled_providers.any? %>
+          <% if SiteConfig.allow_email_password_login %>
             <div class="registration__hr">
               <span class="registration__hr-label">
                 Have a password? Continue with your email address
               </span>
             </div>
+            <%= render partial: "shared/authentication/email_login_form" %>
```
joshpuetz, line 31: 👏

joshpuetz (approved):

spec/system/articles/user_edits_an_article_spec.rb
```diff
@@ -4,8 +4,15 @@
   let(:template) { file_fixture("article_published.txt").read }
   let(:user) { create(:user) }
   let(:article) { create(:article, user: user, body_markdown: template) }
+  let(:svg_image) { file_fixture("svg_image.txt").read }
 
   before do
+    SiteConfig.main_social_image = "https://i.ibb.co/NYsy5Fm/JTL-Logo.jpg"
```
msarit, line 10: There are two image files in `spec/fixtures/files` called `800x600.png` and `800x600.jpg`, but I do not know how to reference them in a `:let` statement (like I did for the `svg-image`).
So in order to set the necessary `SiteConfig` values, I referenced some online files I hosted. But in my mind this is not good practice ❌ 
So I welcome suggestions for improvement 😄

msarit (commented):
In the `spec/system/articles/user_edits_an_article_spec.rb` file, the last spec `"user unpublished their post", the expectation fails intermittently.

When the expectation is `expect(page).to have_text("Unpublished Post.")`, sometimes the spec fails and within the expected output, you don't see `"Unpublished Post."` but you see `published: false`

When the expectation is `expect(page).to have_text("published: false")`, sometimes the spec fails and within the expected output, you don't see `published: false` but you see `Unpublished Post.`

I am at a loss for why this certain spec flip-flops like this 🤷‍♀️

spec/system/articles/user_edits_an_article_spec.rb
joshpuetz: It looks like at least on existing spec uses dummyimage.com: spec/requests/admin/configs_spec.rb references `https://dummyimage.com/300x300`. I'd go with that!

```suggestion
    SiteConfig.main_social_image = "https://dummyimage.com/800x600.jpg"
```

spec/system/articles/user_edits_an_article_spec.rb
```diff
@@ -4,8 +4,15 @@
...
+    SiteConfig.logo_png = "https://i.ibb.co/5nJvG1X/JTL-Logo.png"
```
benhalpern, line 11: I think these should be called like this:

```ruby
allow(SiteConfig).to receive(: logo_png).and_return("https://i.ibb.co/5nJvG1X/JTL-Logo.png")
```

...This helps _this_ test have the value, but it won't leak into other situations. Technically if you assign `SiteConfig.logo_png = ` directly it will persist between tests and could cause inconsistencies.

msarit: Thank you! 😄

spec/fixtures/files/svg_image.txt
```diff
@@ -0,0 +1 @@
...3 1.916-.566 2.806a7.327 7.327 0 01-1.61 2.379 7.431 7.431 0 01-2.408 1.59c-.901.368-1.867.558-2.842.558H1a1 1 0 00-1 1v5.822a1 1 0 001 1h1.475c1.97 0 3.858.773 5.25 2.148a7.288 7.288 0 012.176 5.186zM152.067 55.244v18.09c0 3.89-1.569 7.62-4.364 10.37A15.018 15.018 0 01137.168 88H97.202v-9.778h39.966a5.007 5.007 0 003.512-1.432 4.852 4.852 0 001.455-3.457V53.29a9.667 9.667 0 011.884-5.732A9.925 9.925 0 01148.958 44a9.925 9.925 0 01-4.939-3.557 9.667 9.667 0 01-1.884-5.732V14.667c0-1.297-.524-2.54-1.455-3.457a5.007 5.007 0 00-3.512-1.432H85.5C83.5 5.867 79 1.63 77 0h60.168c3.952 0 7.741 1.545 10.535 4.296a14.548 14.548 0 014.364 10.37v18.09c0 .963.193 1.916.567 2.806.375.89.923 1.698 1.615 2.379a7.46 7.46 0 002.417 1.59 7.555 7.555 0 002.851.558H161a1 1 0 011 1v5.822a1 1 0 01-1 1h-1.483a7.51 7.51 0 00-5.268 2.148 7.276 7.276 0 00-2.182 5.185z\" fill=\"#08090A\"></path><path d=\"M49.2 0c14.842 0 27.437 9.58 31.886 22.866-7.2 6.092-11.952 14.945-12.61 24.912H63.6c-18.557 0-33.6-14.974-33.6-33.445V1a1 1 0 011-1h18.2zM129 21.111a1 1 0 011 1v10.945c0 17.152-13.968 31.055-31.2 31.055h-12V88h-9.6V54.556l.091-4.778C78.515 33.744 91.98 21.11 108.4 21.11H129z\" fill=\"#1AB3A6\"></path></svg>"
```
vaidehijoshi, line 1: It's a bit odd that this is a `.txt` file when it contains only an `<svg>` (especially given [how it is used](https://github.com/forem/forem/pull/10745/files#diff-88c874426c0a935a4a1eead5be1314ebR7)). Can we put it inside of `spec/fixtures/icons` or `/spec/fixtures/images` and save it as an `.svg`? We probably should format this correctly and make sure it is optimized, too.

Alternatively, you could just inline the `<svg>` tag in the spec file itself.

spec/requests/registrations_spec.rb
```diff
@@ -15,27 +15,36 @@
...
       end
 
-      it "shows the sign in text for password based authentication" do
+      it "only shows the single sign on options if they are present" do
+        allow(Authentication::Providers).to receive(:enabled).and_return([])
+
         get sign_up_path
 
-        expect(response.body).to include("Have a password? Continue with your email address")
+        expect(response.body).not_to include("Continue with")
```
vaidehijoshi, line 23:
```suggestion
        expect(response.body).not_to include("Have a password? Continue with your email address")
```
Any reason we can't include the full text in this assertion? Just from reading the test, expecting `"Continue with"` to _not_ render doesn't make it entirely clear what the behavior should be, IMO.

vaidehijoshi (commented):
This works well for me locally! 

I have a higher-level question, though: what does it mean for one of these to be checked, and not the other?
Screen Shot 2020-10-09 at 1 09 24 PM

In other words, what happens if an admin allows you to register with email/password, but doesn't allow you to log in with it? Similarly, what happens if an admin allows you to log in with email/password, but hasn't allowed you to register with it? It seems like these two configurations really ought to be more intertwined...

spec/fixtures/files/300x100.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="300" height="100" xmlns="http://www.w3.org/2000/svg">
+    <path fill="#dedede" stroke="#555" stroke-width="2" d="M2 2h296v96H2z"/>
```
vaidehijoshi, line 2: I think you want 2 space indentation here instead of 4 spaces here? (that seems to be the convention in our other [html fixtures](https://github.com/forem/forem/blob/master/spec/fixtures/approvals/katextag/render/generates_katex_output.approved.html#L3))

spec/system/articles/user_edits_an_article_spec.rb
vaidehijoshi: I agree with @benhalpern on the suggestion to use `allow(...).to receive(: ...).and_return()` here! We probably can update the [other](https://github.com/forem/forem/pull/10745/files/712fc183de8030fd3897aaa9b69cfa607b525135#diff-b13a1d18331fa835d78ef96e51770e3bR14) specs in this PR to make sure that we're isolating the exact test where we assign these values :)

vaidehijoshi (approved):

PR #10748, opened 2020-10-08
Update [authentication] in /admin/config

lisasy:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

**Problem:**

Currently in the Authentication section in `/admin/config`, the information and form are displayed in a way that is unclear and its UX is poor. If we are going to add additional configs here for email login, it's only to become longer, more unwielding, and more confusing for the person looking at this.

**Proposal:**

Update the UX by:
- Packaging the different auth methods into their own section in an accordion-esque display
- Updating content & language to make it clearer

This design also shows how we could build out some of the email login configs we've discussed for this cycle: enabling captchas, throttling new accounts getting created, etc.

> **Note in this PR**
> 
> I think I want to do this in small steps since there is limited resourcing for eng to assist me in this. In this PR, I'm simply updating the visual a bit. In future iterations, I'd like to deprecate the "Authentication providers" dropdown and bring its function to these rows.

Closes https://github.com/forem/InternalProjectPlanning/issues/120

2020-10-08_Authentication Update First PR

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [ ] no documentation needed

alt_text


app/assets/stylesheets/admin.scss
```diff
@@ -96,3 +96,29 @@
...
+  &--subtitle {
+    font-size: var(--fs-s);
+    color: var(--base-60);
+  }
+}
+
+.config-authentication-form {
+  display: flex;
+  flex-direction: column;
+  grid-gap: var(--su-4);
```
ludwiczakpawel, line 122: `grid-gap` works only with `display: grid`

app/views/admin/configs/show.html.erb
```diff
@@ -134,56 +129,142 @@
...
+                      <%= f.check_box :allow_email_password_registration, checked: SiteConfig.allow_email_password_registration, class: "crayons-checkbox" %>
+                      <div>
+                        <%= admin_config_label :allow_email_password_registration %>
+                        <p class="crayons-field__description">
+                          <%= Constants::SiteConfig::DETAILS[:allow_email_password_registration][:description] %>
+                        </p>
+                      </div>
+                    </div>
+                  </fieldset>
+                </div>
```
ludwiczakpawel, line 156: Headers are quite misaligned with rest of the content below:

image

I'd personally suggest different HTML structure - it would prevent having misalignments and it would also help with accordion effect later on.

Instead of having two rows (one with icon + header and the other with inputs) I'd actually suggest having two columns - one with icon and the other with header and inputs.

```
+--------+-----------------------+
| +----+ | +-------------------+ |
| |icon| | |Twitter            | |
| +----+ | +-------------------+ |
|        | +-------------------+ |
|        | | Form field...     | |
|        | | Form field...     | |
|        | +-------------------+ |
+--------+-----------------------+
```

app/assets/stylesheets/admin.scss
```diff
@@ -96,3 +96,29 @@
...
+  }
+
+  &--title {
+    font-weight: var(--fw-medium);
+  }
+
+  &--subtitle {
+    font-size: var(--fs-s);
+    color: var(--base-60);
+  }
```
ludwiczakpawel, line 116: These two elements feel like won't be used anywhere else, so maybe instead of writing CSS here you could apply appropriate utility classes in the html?

```
<element class="fw-medium">
<element class="fs-s color-base-60">
```

I tend to use utility classes over building own separate classes, especially when it's for very specific thing (your class name `config-authentication` suggest it will be used only here), when I have only one or two styles to apply.

app/assets/stylesheets/admin.scss
```diff
@@ -96,3 +96,29 @@
...
+}
+
+.config-authentication-form {
+  display: flex;
+  flex-direction: column;
```
ludwiczakpawel, line 121: I don't think this actually does anything here?

ludwiczakpawel (changes requested):

app/assets/stylesheets/admin.scss
```diff
@@ -96,3 +96,29 @@
 .admin-navigation-links {
   width: 240px;
 }
+
+.config-authentication__row {
+  &--select {
+    display: grid;
+    grid-template-columns: auto 1fr auto;
```
ludwiczakpawel, line 103: Aren't there actually 2 columns?

app/views/admin/configs/show.html.erb
```diff
@@ -134,56 +129,142 @@
...
-                <div class="alert alert-info"><%= Constants::SiteConfig::DETAILS[:facebook_key][:description] %></div>
-              </div>
-              <div class="form-group">
-                <%= admin_config_label :facebook_secret %>
-                <%= f.text_field :facebook_secret,
-                                 class: "form-control",
-                                 value: SiteConfig.facebook_secret,
-                                 placeholder: Constants::SiteConfig::DETAILS[:facebook_secret][:placeholder] %>
-                <div class="alert alert-info"><%= Constants::SiteConfig::DETAILS[:facebook_secret][:description] %></div>
+                Changing authentication keys will not take affect until devops restarts the app (about once per day). This will soon be instant.
```
joshpuetz, line 267: **Suggestion** Use more neutral language here, and don't promise future features!

```suggestion
                Changing authentication keys will not take affect until this forem instance is restarted.
```

joshpuetz (approved):
One suggestion on text, otherwise this looks great to me. I'll leave the CSS feedback to @ludwiczakpawel

app/assets/stylesheets/admin.scss
lisasy: Ah yes. I wrote this initially because I had also include the button on the right that says "Edit", but I ended up removing that markup anyway to simplify this PR. I'll update this.

app/assets/stylesheets/admin.scss
lisasy: I can update these to utility classes.

Tangent, but I think we could develop some of these typographic styles into Crayons components. For example, a row is a pretty typical UI component and it will be sensible to turn this into a repeatable pattern that we use in multiple places and can reference easily.

app/views/admin/configs/show.html.erb
lisasy: Thanks for this helpful suggestion. Lemme try this with `grid` and see how far along I get...

app/views/admin/configs/show.html.erb
lisasy: Done.

rhymes (approved):
Looks really cool Lisa! Thanks!

benhalpern (approved):
Big improvement.

PR #10756, opened 2020-10-09
Refactor carrierwave config

maestromac:
- [x] Refactor

## Description
This is a refactor to keep Carrierwave config more manageable.
n/a
n/a
## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a


config/initializers/carrierwave.rb
```diff
@@ -6,40 +6,43 @@
...
+                      end
+end
+
+if !Rails.env.production? || ENV["FILE_STORAGE_LOCATION"] == "file"
+  CarrierWave.configure(&disk_storage)
+elsif ENV["FOREM_CONTEXT"] == "forem_cloud" # @forem/systems jdoss's special sauce.
+  CarrierWave.configure do |config|
+    config.storage = :fog
+    config.asset_host = "https://#{ApplicationConfig['APP_DOMAIN']}/remoteimages"
+    config.fog_attributes = { cache_control: "public, max-age=#{365.days.to_i}" }
```
maestromac, line 23: This `config.fog_attributes` is newly added. I forgot to add it previously.

config/initializers/carrierwave.rb
```diff
@@ -6,40 +6,43 @@
...
-      config.fog_attributes = { cache_control: "public, max-age=#{365.days.to_i}" }
-    else
-      # Fallback on file storage if AWS creds are not present
-      config.asset_host = "https://#{ApplicationConfig['APP_DOMAIN']}/localimages"
-      config.storage = :file
-    end
+disk_storage = proc do |config|
+  config.storage = :file
+  config.enable_processing = !Rails.env.test?
+  config.asset_host = if Rails.env.production?
```
citizen428, line 12: I 'd probably use a temporary variable here for better readability:

```ruby
asset_host = "https://#{ApplicationConfig['APP_DOMAIN']}/localimages" if Rails.env.production?
config.asset_host = asset_host
```

config/initializers/carrierwave.rb
```diff
@@ -6,40 +6,43 @@
...
+                        "https://#{ApplicationConfig['APP_DOMAIN']}/localimages"
+                      end
+end
+
+if !Rails.env.production? || ENV["FILE_STORAGE_LOCATION"] == "file"
+  CarrierWave.configure(&disk_storage)
```
citizen428, line 18: Is there a specific reason for `disk_storage` to be a proc while the other branches of this `if` are inlined?

config/initializers/carrierwave.rb
mstruve: Any chance we can add this separate from the refactor? I worry when adding new features during a refactor that if something breaks it makes it hard to figure out if its the new feature or if the refactor had a bug.

maestromac: Yep I can do that!

config/initializers/carrierwave.rb
mstruve: I know the purpose of the proc is to de-duplicate code but I think it would read a little cleaner without it.

config/initializers/carrierwave.rb
```diff
@@ -6,40 +6,42 @@
...
-    end
+disk_storage = proc do |config|
+  config.storage = :file
+  config.enable_processing = !Rails.env.test?
+  config.asset_host = if Rails.env.production?
+                        "https://#{ApplicationConfig['APP_DOMAIN']}/localimages"
+                      end
+end
+
+if !Rails.env.production? || ENV["FILE_STORAGE_LOCATION"] == "file"
```
mstruve, line 17: I'm one of those people that always has to pause when I get to negative assertion statements like this. I think here we could remove the `!Rails.env.production?` and let the final else catch it.

mstruve (commented):
couple small suggestions, let me know what you think!

config/initializers/carrierwave.rb
maestromac: Yeah, I wasn't really happy with this refactor either. I just re-arranged it differently and removed this proc.

config/initializers/carrierwave.rb
maestromac: I removed it and it gave me another idea of how to re-organize this.

mstruve (approved):

citizen428 (approved):

Issue #9770, opened 2020-08-13
Reading list counter


ezio1404:
**Describe the bug**

I have a 200+ reading list unopened, I archive them after I read or save, after I clear out my reading list and go back to the main page the reading list on the sidebar has a counter(unopened or unread post) and when I clicked it it's empty.
Things I've done:
- hard refresh and emptying the cache
- closing the browser and opening it again
- logout and logging in again

**To Reproduce**

**Expected behavior**

**Screenshots**
Here are the screenshots:
![readinglist](https://user-images.githubusercontent.com/34921038/90159486-0b875080-ddc3-11ea-96e8-6525afa7a694.JPG)
![reading list list](https://user-images.githubusercontent.com/34921038/90159505-1215c800-ddc3-11ea-9d6f-8ef4b81eacc7.JPG)

**Desktop (please complete the following information):**

- OS, version:
Windows 10 Home single language 64-bit (10.0,Build 18362)
- Browser, version:
Microsoft Edge Version 84.0.522.59 (Official build) (64-bit)
**Smartphone (please complete the following information):**

- Device:
- OS, version:
- Browser, version:

**Additional context**

rhymes:
Hi @ezio1404 the counter reflects the non archived items, but it's weird that youur list is empty but the counter says 5. Do you have the items in the archive section?

ezio1404:
Hi @rhymes yes I have 290+ archived items and recently I save 7 post and it backed up to 29
![re1](https://user-images.githubusercontent.com/34921038/90162018-94ec5200-ddc6-11ea-84c0-867d298b653a.JPG)
![re2](https://user-images.githubusercontent.com/34921038/90162031-99186f80-ddc6-11ea-9603-f83e121f5f0f.JPG)

Archived:
![re3](https://user-images.githubusercontent.com/34921038/90162053-a2094100-ddc6-11ea-9b0a-4806e296b239.JPG)

takmar:
Could I take this?

rhymes:
@takmar sure, I've assigned it to you, thanks!

rhymes:
Reopening this as it was reverted in https://github.com/forem/forem/pull/10932

rhymes:
Looks like it was then closed by #10933

PR #10763, opened 2020-10-10
Fix reading list counter

takmar:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
I found that a wrong reading list count was caused by caching. The cache of [cached_reading_list_article_ids](https://github.com/forem/forem/blob/10efa25c0a35b69248dbebe6fe828a201607ca45/app/models/user.rb#L290) is changed only when `public_reactions_count` is updated, but we must also recache it when user archives and unarchives an article.

To achieve this, I added the `last_changed_archive_at` column to the users table and cache keys.
I could not come up with an idea to fix this without adding an extra column. However, Let me know If you have any other ideas. I'll give it a try.

https://github.com/forem/forem/issues/9770

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

tenor

### DIFF
--- a/db/migrate/20201009160628_add_last_changed_archive_at_to_user.rb
+++ b/db/migrate/20201009160628_add_last_changed_archive_at_to_user.rb
@@ -1,0 +1,5 @@
+class AddLastChangedArchiveAtToUser < ActiveRecord::Migration[6.0]
+  def change
+    add_column :users, :last_changed_archive_at, :datetime, default: "2017-01-01 05:00:00"
+  end
+end
--- a/spec/requests/reading_list_items_spec.rb
+++ b/spec/requests/reading_list_items_spec.rb
@@ -22,11 +22,13 @@ RSpec.describe "ReadingListItems", type: :request do
     it "returns archives item if no param" do
       put "/reading_list_items/#{reaction.id}"
       expect(reaction.reload.status).to eq("archived")
+      expect(user.last_changed_archive_at).not_to eq(user.reload.last_changed_archive_at)
     end
 
     it "unarchives an item if current_status is passed as archived" do
       put "/reading_list_items/#{reaction.id}", params: { current_status: "archived" }
       expect(reaction.reload.status).to eq("valid")
+      expect(user.last_changed_archive_at).not_to eq(user.reload.last_changed_archive_at)
     end
 
     it "raises NotAuthorizedError if current_user is not the reaction user" do
--- a/app/controllers/async_info_controller.rb
+++ b/app/controllers/async_info_controller.rb
@@ -76,6 +76,7 @@ class AsyncInfoController < ApplicationController
     #{current_user&.last_sign_in_at}__
     #{current_user&.following_tags_count}__
     #{current_user&.last_followed_at}__
+    #{current_user&.last_changed_archive_at}__
     #{current_user&.updated_at}__
     #{current_user&.reactions_count}__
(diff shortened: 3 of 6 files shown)


db/migrate/20201009160628_add_last_changed_archive_at_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddLastChangedArchiveAtToUser < ActiveRecord::Migration[6.0]
+  def change
+    add_column :users, :last_changed_archive_at, :datetime, default: "2017-01-01 05:00:00"
```
mstruve, line 3: I am not a huge fan of adding another "cache" focused column to user but I cant think about a better way to do this. 

What do you think of making this a more general column so we could use it for other caches as well? I was thinking `last_reacted_at` which would fit nicely with our pattern of `last_followed_at`, `last_article_at` etc.

mstruve (commented):
Thanks for tackling this bug. Code looks pretty good, I left one suggestion, let me know what you think!

db/migrate/20201009160628_add_last_changed_archive_at_to_user.rb
takmar: Thank you for your comments.

> I am not a huge fan of adding another "cache" focused column to user

Neither am I. There are already a bunch of cache keys.

> What do you think of making this a more general column so we could use it for other caches as well? I was thinking last_reacted_at which would fit nicely with our pattern of last_followed_at, last_article_at etc.

That’s a new idea for me. At least, I think that it is better than mine as we do not need to add a column whenever something related to a user needs to be cached.

spec/requests/reading_list_items_spec.rb
```diff
@@ -22,11 +22,13 @@
     it "returns archives item if no param" do
       put "/reading_list_items/#{reaction.id}"
       expect(reaction.reload.status).to eq("archived")
+      expect(user.last_reacted_at).not_to eq(user.reload.last_reacted_at)
```
mstruve, line 25: Rather than relying on ActiveRecord to local memoize this here, I think a cleaner approach would be.
```ruby
original_reacted_at = user.last_reacted_at
put "/reading_list_items/#{reaction.id}"
expect(reaction.reload.status).to eq("archived")
expect(original_reacted_at).not_to eq(user.reload.last_reacted_at)
```
Same with the example below.

mstruve (approved):
Small suggestion for the spec but otherwise this LGTM!!!!

spec/requests/reading_list_items_spec.rb
citizen428: We could also use an actual `change` assertion here:

```ruby
expect do 
  put "/reading_list_items/#{reaction.id}"
end.to change { user.reload.last_reacted_at }
expect(reaction.reload.status).to eq("archived")
```

takmar: I have never thought of being able to use `change` by giving `put "..."` to `expect` as a block. Thank you!

atsmith813 (approved):
Looks like there's a small conflict that needs to be resolved. Otherwise, LGTM! Thank you so much, I'm looking forward to this fix 🎉 .

PR #10784, opened 2020-10-12
add fullscreen to code blocks

### LINKED ISSUE #8990: "Lightbox mode" for code snippets
Code often requires more room to take in what we're reading, and with our narrower code blocks it means more side scrolling.

A lot of code is legible in a narrow space, but a lot is not.

I propose an "expand" button on the top right of all code blocks (similar to how they are displayed in window OS)

Screen Shot 2020-06-29 at 1 44 38 PM

This would allow a user to expand the viewing area for any code block.

Minimizing also seems like a decently useful feature for code blocks. If we're looking at other functionality that might go in this area, I think a "copy" button could also make sense.

This is mostly a desktop feature so probably only display on hover. I think this could add functionality and also a pretty slick and distinct look by mimicking OS windows a bit.

The "lightbox" mode should also apply to images if we get that feature shipped. It's been proposed in the past.

Winner95:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I added fullscreen functionality for code blocks via panel, shown on hover. There are some updates for highlight code block, which allow division of concerns in code:

* classnames starting with _js-_ for event-listeners - in order to be independent from markup.
* added 'highlight__panel-action', which allow to add several elements (action-icons, e.g. copy) by the same pattern, which will result in proper styling
* added store of window scroll position and preventing scroll on document on fullscreen mode
* added event-listener for escape key on keyboard for switching off the full-screen
* scss is written in BEM, using some utility classes.
* in order to keep size of `markdown_parser_spec.rb` below 250 lines, divide it into two files.
* added padding for scrolled code on right side

This is how highlight block was before:
before

This is after:

after2

after

after3

Closes #8990 

* add different code blocks to the page
* hover over code blocks to see the fullscreen button
* click on button to enter fullscreen mode
* click on button back / press 'escape' button

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

PS: I hope this pull-request will be helpful:)


citizen428 (approved):
Tried it out locally, works well! Thanks for the contribution 😃

app/assets/stylesheets/components/syntax.scss
```diff
@@ -34,6 +35,100 @@ div.highlight {
...
+  left: 0;
+  z-index: 9999;
+  width: 100%;
+  height: 100vh;
+  background: var(--syntax-background-color);
+  color: var(--syntax-text-color);
+  overflow: auto;
+  visibility: hidden;
+  transform: scale(0.01);
+  transition: transform 0.4s ease-in-out, visibility 0.4s ease-in-out;
```
ludwiczakpawel, line 115: Can we replace `0.4s ease-in-out` with `var(--transition-props)` - won't be a huge difference visually, but will definitely speed up the transition (which I think could much faster).

```suggestion
  transition: transform var(--transition-props), visibility var(--transition-props);
```

app/assets/stylesheets/components/syntax.scss
```diff
@@ -34,6 +35,100 @@ div.highlight {
...
+  background-size: var(--icon-size) var(--icon-size);
+}
+
+.highlight-action-icon--fullscreen {
+  background-image: url(image-src('fullscreen-on.svg'));
+
+  .highlight.is-fullscreen & {
+    background-image: url(image-src('fullscreen-off.svg'));
+  }
+}
```
ludwiczakpawel, line 101: Is there a chance to use inline svg in markup/js instead of background-image in css?

app/assets/stylesheets/components/syntax.scss
```diff
@@ -34,6 +35,100 @@ div.highlight {
...
+
+  & + & {
+    margin-left: 1px;
+    border-bottom-left-radius: 0;
+  }
+}
+
+.highlight-action {
+  display: flex;
+  background: #4f575f;
```
ludwiczakpawel, line 72: Can we use `var(--base-60)` instead that hex value?

ludwiczakpawel (changes requested):
Oh, this works great, lovely job!

I have few tiny requests but nothing really major - left comments in the code..
Also, is there a chance to simplify that markup? It feels a bit overkill for one button:

image

Winner95:
@ludwiczakpawel thank you for your feedback, updated PR,  reduced markup and kept possibility to add any new buttons in action-panel

app/assets/stylesheets/components/syntax.scss
```diff
@@ -34,6 +34,94 @@ div.highlight {
   }
 }
 
+div.highlight {
+  position: relative;
+
+  // this a way to separate scroll of long code & icons in panel
+  .highlight {
+    overflow: auto;
+  }
```
ludwiczakpawel, line 43: There's one last thing we could try to fix... So this style changes a little bit the behavior of how overflow looks like.. On prod, when you have long lines in `highlight` the code is actually from edge to edge when scrolling. 

Because of that additional styling here ^ there's an offset on the left and right side, defined by the padding, so the code is no longer from edge to edge.

I do understand why you have this styling here (so the buttons are sticked to the corner of the code block), although maybe we could try to bring the previous styling back (so code will be edge to edge when scrolling). I think the trick here is to move the padding from the outer `.highlight` to the inner one. But this may need a bit more testing though.

If you need help with it, I'm happy to jump in.

ludwiczakpawel (changes requested):
One issue in the comments, though not a blocker.

But also, I've just realized that old posts do not have an icon in the code blocks:

image

I guess it's because markup is not generated on the fly but stored somehow? I don't know tbh... 

I don't think it matters that much for "older posts" so if there's not an easy way to fix it, I'd at least try to hide it for older posts and make this functionality "available" for new posts only.

rhymes:
@ludwiczakpawel 

> I guess it's because markup is not generated on the fly but stored somehow? I don't know tbh...

Yes, on save an article turns the markdown into HTML, that's stored in the DB and in the cache at the first request. Everytime an article is updated the cache is wiped and the HTML version in the DB is overwritten.

If we need to update all old articles we need an update script to resave all existing articles to update their markup. It's a possibility.

benhalpern (approved):
This looks great to me.

@ludwiczakpawel @rhymes Because it has the `if ...` action built in, this is resilient to the old posts not _immediately_ being compatible....

We'll handle this scenario as a one-off, but yeah we should seek to create a more general framework around backfilling breaking changes. There's no easy answer, but we're definitely outgrowing the old fragile way quickly.

@Winner95 only thing left to do is make sure all the tests are updated. It looks like the only failing tests are approval tests. See info here: https://docs.forem.com/tests/approvals-tests/

I approve this pending the tests passing.

app/assets/stylesheets/components/syntax.scss
Winner95: @ludwiczakpawel,  sure, I will check it!

nickytonline:
Thanks so much for the PR @Winner95. Any chance you can put some screenshots in the PR description showing the functionality in action. Thanks!

app/views/articles/show.html.erb
```diff
@@ -8,9 +8,9 @@
 
 <%= render "shared/webcomponents_loader_script" %>
 <% if user_signed_in? %>
-  <%= javascript_packs_with_chunks_tag "clipboardCopy", "webShare", "articlePage", "articleModerationTools", defer: true %>
+  <%= javascript_packs_with_chunks_tag "clipboardCopy", "webShare", "articlePage", "codeFullscreenModeSwitcher", "articleModerationTools", defer: true %>
```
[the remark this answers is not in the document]
Winner95: Added this change to the scope of PR.

app/assets/stylesheets/components/syntax.scss
Winner95: I move padding to inner element -  .highlight, where overflow is and it worked. Now we also have proper calculation of paddings (screenshot will be attached) for wide scrollable area (before text had no right padding at the end of the line).

nickytonline (approved):
Thanks for making the change @Winner95! 🚀

ludwiczakpawel (approved):
🎉🎉🎉

PR #10788, opened 2020-10-12
new liquid tag - dotnetfiddle

### LINKED ISSUE #10402: Support for dotnetfiddle.net

As a writer that blogs mainly about .NET and C#, I would like to be able to embed dotnetfiddle.net in my posts.

It might be a frustrating experience for users to exit the post to check or run code examples.

I would ask to expand the support that already exists for jsfiddle.net to include dotnetfiddle.net and maybe even other languages.

The current alternative is to use gists to include code snippets and link the users to external dotnetfiddle.net links.

ShubhankarKG:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR adds a new liquid tag, **dotnetfiddle**.

Fixes #10402 

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [x] docs.forem.com
- [ ] readme
- [ ] no documentation needed

PS : Apologies for the delay


ShubhankarKG:
Okay, I would like some help in the tests section :sweat_smile:

citizen428:
@ShubhankarKG How can we help you with the tests? 😃

ShubhankarKG:
My approval test passed on 2 out of 3 builds and failed on 1. The approval test is causing a bit of a problem here. :sob:

citizen428:
Thanks for clarifying, I'll have a look.

citizen428:
The problem was that calling `liquid.render` on line 20 of the spec resulted in an error:

```
"Liquid error: internal"
```

So I started a development server and made a post including the Liquid tag `{% dotnetfiddle https://dotnetfiddle.net/Widget/v2kx9jcd %}` This also raises an exceptiona nd the provided link results in a 404.

I guess you should be able to further debug things from here.

app/views/liquids/_dotnet_fiddle.html.erb
```diff
@@ -0,0 +1,13 @@
+<%# Weird hack due to the fact that embedded links have Widget in between.
+Starting from https:// we end at 24 characters at dotnetfiddle.net %>
+
+<iframe
+  src="<%= link.insert(24, "/Widget").to_s %>"
```
rhymes, line 5: can we keep this "hack" in the liquit tag ruby file? We should also test it :D 

Is there a chance the number of chars will change?

rhymes (commented):
Hi @ShubhankarKG, thanks for the PR. Left a note

app/views/liquids/_dotnet_fiddle.html.erb
citizen428: ➕ for moving this into the actual tag instead of keeping it in the view. I think the 24 is unlikely to change, as it's the lenght of "https://dotnetfiddle.net", but I think a more robust solution with Ruby's `URI` module for building the link would be preferable.

ShubhankarKG: Sure, thanks for the inputs everyone. I had the "hack" (xD) shifted in the controller, but I sticked to using the string.insert method. 

I'll try to make the change through the URI module now. Thanks for the heads up :+1

app/liquid_tags/dotnet_fiddle_tag.rb
```diff
@@ -0,0 +1,49 @@
...
+    else
+      # URI.path gives us strings like "/abcde", use substring to get rid of forward slash
+      # otherwise join won't work
+      URI.join("https://dotnetfiddle.net", "/Widget/", uri.path[1..]).to_s
+    end
+  end
+
+  def valid_link?(link)
+    link_no_space = link.delete(" ")
+    (link_no_space =~ LINK_REGEXP).zero?
```
citizen428, line 45: **nitpick (non-blocking):** 

```suggestion
    link_no_space.match?(LINK_REGEXP)
```

The regex is anchored, so if it matches, it matches starting at anyway.

citizen428 (approved):

benhalpern (approved):

PR #10789, opened 2020-10-12
Liquid tag for sharing pretty code snippets

genderev:
## What type of PR is this? 
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

There are two options I consider when sharing code on dev.to - carbon.now.sh images are pretty but you can't copy code from an image. Github Gists and the built in `<pre>` tag allow users to share code they can copy and paste, but they're not as fun as Carbon. What if you combined nice visuals with code you can copy for your own use? Enter Code Sharer.

A test for the liquid tag can be found [here](https://github.com/genderev/forem/blob/master/spec/liquid_tags/codesharer_tag_spec.rb). 

Here's a screenshot:

screenshot

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [X] docs.forem.com
- [ ] readme
- [ ] no documentation needed

dancing lobster


citizen428:
Hi and thanks for your contribution @genderev. Could you please include a screenshot in the PR description, thanks!

app/liquid_tags/codesharer_tag.rb
```diff
@@ -0,0 +1,54 @@
+class CodesharerTag < LiquidTagBase
+  PARTIAL = "liquids/codesharer".freeze
+  PARAM_REGEXP = /\w|(#|=)/.freeze
+
+  def initialize(_tag_name, link, _parse_context)
+    super
+    stripped_link = ActionController::Base.helpers.strip_tags(link)
+    the_link = stripped_link.split(" ").first
+    @embedded_url = CodesharerTag.embedded_url(the_link)
```
citizen428, line 9: Instead of defining `embedded_url` (and all it's dependencies) on the class, I'd make them private instance methods of the tag class. While it's true that they don't rely on instance state, it's more in line with how our other tags do things.

app/liquid_tags/codesharer_tag.rb
```diff
@@ -0,0 +1,54 @@
...
+    "https://codesharer.netlify.app/#{URI.encode_www_form(parse_link(link))}"
+  end
+
+  def self.parse_link(link)
+    begin
+      url = URI(link)
+    rescue StandardError
+      raise_error
+    end
+    hostname_ok = url.hostname == "codesharer.netlify.app"
```
citizen428, line 31: Please pull `"codesharer.netlify.app"` out into a constant.

app/liquid_tags/codesharer_tag.rb
```diff
@@ -0,0 +1,54 @@
...
+    raise_error unless hostname_ok
+    parse_params(url)
+  end
+
+  def self.parse_params(url)
+    query = url.query.nil? ? [] : URI.decode_www_form(url.query)
+    %i[from to theme readOnly].each do |param|
```
citizen428, line 38: Please turn this params array into a constant, we don't need to recreate it at every method invocation.

app/liquid_tags/codesharer_tag.rb
```diff
@@ -0,0 +1,54 @@
...
+      value = query.assoc(param.id2name)&.last
+      result[param] = valid_param?(value) ? value : ""
+    end
+    result
+  end
+
+  def self.valid_param?(value)
+    !value&.match(PARAM_REGEXP)&.nil?
```
citizen428, line 46: There's a `match?` predicate method, which would be better suited for this task:

```suggestion
    value&.match?(PARAM_REGEXP)
```

app/liquid_tags/codesharer_tag.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def self.parse_link(link)
+    begin
+      url = URI(link)
+    rescue StandardError
+      raise_error
+    end
+    hostname_ok = url.hostname == "codesharer.netlify.app"
+    raise_error unless hostname_ok
```
citizen428, line 32: I don't think the extra variable adds much here.

```ruby
raise_error unless url.hostname == CODESHARER_URL
pare_params(url)
```

citizen428 (commented):
Thanks for the PR @genderev. I left some comments. 

I see that Codesharer runs on Netlify. Is it your project? It does look really good, but I'm slightly reluctant to add Liquid tags for someone's side project, as that (a) puts some burden on us to ensure the site is stil operational and (b) could lead to a proliferation of Liquid tags if users see adding them to DEV as a way to "advertise" their projects.

app/liquid_tags/codesharer_tag.rb
genderev: Thanks so much for the review! 

Code Sharer is my repo. I just want to be able to share nice code snippets in my dev.to posts, not to advertise the project. 

RE: site reliability - I care about SRE a lot. I don't expect that to be an issue. The CI pipeline keeps errors from arriving in production.

citizen428: Thanks for the context @genderev! I hope you enjoy learning Ruby and contributing to the Forem codebase 😃

ludwiczakpawel:
Hey, I tried to test it locally but it doesn't work for me:

image

---

Can you give us some context on when do you expect users to use it (that liquid tag) instead of our default code blocks that look like this:

image

genderev:
Wow, thanks for all the feedback!

Contributers to the Git repo for Code Sharer noted that they would like to see this feature on dev.to, but I need to look into the Safari issues.

mstruve:
Hi @genderev!

Thank you so much for this liquid tag PR! The time and effort you put into crafting this is very clear based on the wonderful code. 

I am the lead SRE here at Forem and had some thoughts on your PR. Given this liquid tag is related to a small very niche service I think it is best if we forego adding it to the repository at this time.  When we make decisions about liquid tags to include we look at things like how many of our customers will benefit and use this tag(within DEV and within other Forems which may include lots of nontechnical folks) and then we balance that with how widely adopted the service the tag is supporting is. The more widely used and adopted a service is the less risk it carries for us to add a liquid tag for it. The reason we have to be so diligent about this is that once a tag is added, we cant easily remove it once it starts getting used as it would break posts. 

I believe that this tag would not be widely used across all our platforms and because it is linked to a very new, young service I view the risk of adding it and maintaining it as high. I also commend you on how striking the code-sharing snippets are. With that said, we do have code highlighting already provided through our markdown parser so I don't think this adds much in terms of additional functionality for those users wanting to share code. 

Let me know if that makes sense and once again THANK YOU for contributing to Forem. We couldn't do what we do and we wouldn't have such an amazing platform without amazing contributors like yourself. We know it takes a lot of time and effort to contribute to open source and that having all your hard work not get included sometimes can be very frustrating. I am always here and happy to answer any questions you might have. 

Cheers!

Issue #10723, opened 2020-10-08
support RSS feed discovery from articles and profile pages


chdsbd:
**Is your feature request related to a problem? Please describe.**

I'd like to be able to subscribe to a user's posts via RSS by pasting their article url or profile url into my feed reader.

**Describe the solution you'd like**

Add `<link rel="alternate" type="application/rss+xml"  href="/feed/<username>" />` tags to articles and user profile pages to support RSS feed discovery.

RSS readers use these tags to discover feeds. I didn't find an official article on this, but this post discusses this feature: https://www.petefreitag.com/item/384.cfm

**Describe alternatives you've considered**

After some searching I found the feed url is `dev.to/feed/<username>`. A link to this URL on the profile page would be useful too.

**Additional context**

query-string:
Hello there! I'd love to work on this one please :)

rhymes:
Assigned to you @query-string, thanks!

cmgorton:
Thank you for opening this request.
Since this has been open for awhile without much progress I am going to close it here. 
The Forem core team now uses an internal RFC ("request for comments") process to assess and prioritize new features. This process is intended to provide a consistent and standardized path for new changes to enter the Forem ecosystem. If you'd like to propose a new feature like this one, please visit [forem.dev](https://forem.dev/) to start a discussion around a new feature! 🙏

PR #10802, opened 2020-10-12
Support RSS feed discovery from articles and profile pages

query-string:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR makes RSS channels for user profile / organization profile / articles pages auto discoverable by RSS readers as it was requested per #10723 Besides – RSS subscription link added to the profile sidebar

Fun fact – RSS discoverability has been in the project [since forever](https://github.com/forem/forem/commit/074d50d2b70be66d800cefa2b11c249bb38725e9#diff-1b2a844dcc6bcd4b8821a51d811fa11aR18), but during the course of code evaluation events `yield(:page_meta)` became available only for robotic crawlers (if I'm gathering `internal_navigation?` mission correctly) thus RSS meta hidden from the rest of the regular users and their browsers. And that's why we have this issue from @chdsbd in the first place...

My gut feeling – RSS meta is not the only `<head>` data that was supposed to be visible only for regular users and not search engines so I decided to introduce `yield(:internal_navigation_page_meta)` area for such cases and put it inside `_top` partial. Please let me know if you think it is an overuse and RSS meta should exist for both, robotic crawlers and regular browsers worlds. But still RSS meta is something that supposed to be inserted in the page header based on conditions so probably it's good to have a dedicated area for it...

#### Current PR making RSS discovery icon visible in FireFox + [Awesome RSS addon](https://addons.mozilla.org/en-US/firefox/addon/awesome-rss/) adddress bar on the user profile page

Screen Shot 2020-10-12 at 18 05 01

#### RSS discovery icon is visible in FireFox + [Awesome RSS addon](https://addons.mozilla.org/en-US/firefox/addon/awesome-rss/) adddress bar on the user profile page in production

Screen Shot 2020-10-12 at 18 04 49

#### RSS subscription link in the profile sidebar

Screen Shot 2020-10-12 at 17 49 31

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/spec/requests/user/user_profile_spec.rb
+++ b/spec/requests/user/user_profile_spec.rb
@@ -62,18 +62,6 @@ RSpec.describe "UserProfiles", type: :request do
       expect(response.body).not_to include("<meta name=\"googlebot\" content=\"noindex\">")
     end
 
-    it "renders rss feed link if any stories" do
-      create(:article, user_id: user.id)
-
-      get "/#{user.username}"
-      expect(response.body).to include("/feed/#{user.username}")
-    end
-
-    it "does not render feed link if no stories" do
-      get "/#{user.username}"
-      expect(response.body).not_to include("/feed/#{user.username}")
-    end
-
     it "renders user payment pointer if set" do
       user.update_column(:payment_pointer, "test-payment-pointer")
       get "/#{user.username}"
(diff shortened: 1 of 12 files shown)


spec/requests/user/user_profile_spec.rb
```diff
@@ -62,18 +62,6 @@
...
-    it "renders rss feed link if any stories" do
-      create(:article, user_id: user.id)
-
-      get "/#{user.username}"
-      expect(response.body).to include("/feed/#{user.username}")
-    end
-
-    it "does not render feed link if no stories" do
-      get "/#{user.username}"
-      expect(response.body).not_to include("/feed/#{user.username}")
```
query-string, line 74: After bringing RSS subscription link into the sidebar this test case was constantly failing because of obvious reasons so I decided to change it a bit so [not it checks](After bringing RSS subscription link into the sidebar this test case ) that page includes `type="application/rss+xml"` string, which seems to be more relevant anyway.

ludwiczakpawel (approved):
I think UI wise it's fine. We will soon revamp the profile view a little and we can then improve it. So right now I wouldn't spend too much efforts on improving the visuals here which are just ok.

cmgorton:
Hi @query-string 

PR #10758, opened 2020-10-09
Allow Forem admins to enable an invite only mode.

jacobherrington:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I had to read a bunch of code, but this seemed too easy. :thinking: 

This should allow a Forem admin to check a box and disable all social authorization providers, which will mean users can only join via invitation.

## Related Tickets & Documents

Closes: https://github.com/forem/InternalProjectPlanning/issues/115

This PR builds on the stuff @benhalpern did here: https://github.com/forem/forem/pull/9294

## QA Instructions, Screenshots, Recordings

Adds a checkbox to config that disables the auth providers input when checked:
image

This is what you see if you try to create an account:
image

## Added tests?

- [x] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

PR #10811, opened 2020-10-12
Remove code setting SiteConfig values in specs

jacobherrington:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

These changes didn't get pushed to https://github.com/forem/forem/pull/10758 before it was merged.

## Related Tickets & Documents

https://github.com/forem/forem/pull/10758

## Added tests?

- [x] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

PR #10812, opened 2020-10-12
Allow all third party auth providers to be deactivated

joshpuetz:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Just spitballing a possible solution to private forems: this PR allows one to turn off all third party authentication providers. Combined with turning off email account creation, this would allow for a private, invite only forem.

Just turn off everything:
Configs

The account creation page becomes quite empty:
Welcome__-_Forem

And you can only log in with an existing email account:
Welcome__-_Forem

## Related PRs

https://github.com/forem/forem/pull/10758 - adds an 'invite only mode', but doesn't address this particular UI issue.

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Members Only


joshpuetz:
LOL: Never mind! https://github.com/forem/forem/pull/10758

joshpuetz:
After talking with @jacobherrington, he pointed out this solves a different problem than https://github.com/forem/forem/pull/10758, so I'm reopening it!

spec/requests/admin/configs_spec.rb
```diff
@@ -79,6 +79,14 @@
           expect(SiteConfig.authentication_providers).to eq([provider])
         end
 
+        it "allows all providers to be disabled" do
+          # We have to send _something_ for an update, so we'll simulate the borwser omitting the auth
+          # providers array when no options are selected
+          post "/admin/config", params: { site_config: { community_description: "test" },
```
jacobherrington, line 85: This test might need to be updated, I think I added a request spec for this in my PR, so we might be able to just drop this one.

There isn't a system spec for it, which is the real issue that this change is addressing, as I understand it.

joshpuetz: So the thing about unselecting all auth providers in the UI is that the browser won't send that key to the backend: the `authentication_providers` key will simply be omitted from params. Any request test that has `authentication_providers: []` as a param is constructing a situation that can't occur!

I tried constructing a test with a blank `site_config` hash, but since it's a require parameter Rails won't let that fly. I've changed this test to have a just `ga_tracking_id`

msarit (approved):
Ui-wise, everything works! 🚀  Great job @joshpuetz

spec/requests/admin/configs_spec.rb
```diff
@@ -79,6 +79,33 @@
...
+          # We have to send _something_ for an update, so we'll simulate the borwser omitting the auth
+          # providers array when no options are selected
+          post "/admin/config", params: { site_config: {ga_tracking_id: "123" },
+                                          confirmation: confirmation_message }
+          expect(SiteConfig.authentication_providers).to be_empty
+        end
+
+        context "when authentication providers are present" do
+          before do
+            SiteConfig.authentication_providers = Authentication::Providers.available.map(&:to_s)
```
jacobherrington, line 92: This needs to be updated, it has been removed in master: https://github.com/forem/forem/pull/10811

I believe nothing after line 89 should show up in the diff.

jacobherrington (changes requested):
Sorry this is a bit confusing because a PR in that spec file got merged with uncommitted changes and had to be updated in another PR.. that seems to be causing a bit of an issue for you :sweat:

app/controllers/admin/configs_controller.rb
```diff
@@ -220,5 +223,11 @@ def job_params
         display_jobs_banner
       ]
     end
+
+    def allowed_empty_enumerables
```
citizen428, line 227: Seems like this could be a constant?

app/controllers/admin/configs_controller.rb
```diff
@@ -14,7 +14,8 @@ def create
 
       config_params.each do |key, value|
         if value.is_a?(Array)
-          SiteConfig.public_send("#{key}=", value.reject(&:blank?)) unless value.empty?
+          SiteConfig.public_send("#{key}=", value.reject(&:blank?)) unless
```
citizen428, line 17: **nitpick (non-blocking):** If we have to break it over two lines it's probably not a good use case of an inline modifier. I'd say there's enough happening in this line that we can move things into a method:

```ruby
def update_siteconfig_with_array(key, value)
  return if value.empty? && allowed_empty_enumerables.exclude?(key.to_sym)

  SiteConfig.public_send("#{key}=", value.reject(&:blank?)
end
```

```suggestion
          update_siteconfig_with_array(key, value)
```

spec/requests/admin/configs_spec.rb
```diff
@@ -79,6 +79,33 @@
...
+          end
+
+          it "allows authentication providers to be unset" do
+            new_provider = Array.wrap(SiteConfig.authentication_providers.pop.to_s)
+            post "/admin/config", params: { site_config: { authentication_providers: new_provider },
+                                            confirmation: confirmation_message }
+            expect(SiteConfig.authentication_providers).to eq(new_provider)
+          end
+
+          it "allows all authentication providers to be unset" do
```
citizen428, line 102: **suggestion (non-blocking):** I think it'd be great to actually see the disabling "in action" with a `change` assertion, e.g

```ruby
expect do
  post "/admin/config", 
        params: { site_config: {ga_tracking_id: "123" },
end.to change(SiteConfig, :authentication_providers).to([])
```

app/controllers/admin/configs_controller.rb
joshpuetz: Great suggestion, they've been constantized!

app/controllers/admin/configs_controller.rb
joshpuetz: Great suggestion, I agree: it's getting too complicated in there!

app/controllers/admin/configs_controller.rb
```diff
@@ -1,39 +1,95 @@
 module Admin
   class ConfigsController < Admin::ApplicationController
-    layout "admin"
+    CAMPAIGN_PARAMS =
```
citizen428, line 3: Nice, so much better. If you ever feel like a follow-up PR @joshpuetz, maybe moving these into a file in `app/constants` could be good, similar to `app/constants/site_config.rb`.

citizen428 (approved):
🚀 Thanks for the changes @joshpuetz

jacobherrington (approved):

PR #10816, opened 2020-10-12
Centralize on "trusted" language: Add new copy and some slight feature adjustments.

### LINKED ISSUE #10734: Rename :trusted role to :moderator

It's not super clear to the casual user (and I did forgot as well :D) that `trusted` is the role that corresponds to moderators. 

I think we should rename `trusted` to the self explanatory `moderator`

Nope, none of the alternatives like `trusted_moderator` seems better.

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This changes the language from "community mod" to "trusted" to align with the language we use _in the role_.... So rather than have the role called `trusted` but the email and on-site information mostly speak about "community moderation", we'd rather centralize on these being "trusted" users... They can still _do_ community moderation, but the name of the role and how it is introduced can centralize around `trusted`....

This also removes `banned` users from the query for users to "potentially" make trusted. We do not want to accidentally "entrust" a banned user.

Closes #10734

### DIFF
--- a/app/queries/admin/moderators_query.rb
+++ b/app/queries/admin/moderators_query.rb
@@ -10,8 +10,10 @@ module Admin
 
       relation = if state.to_s == "potential"
                    relation.where(
-                     "id NOT IN (SELECT user_id FROM users_roles WHERE role_id = ?)",
+                     "id NOT IN (SELECT user_id FROM users_roles WHERE role_id = ?)
+                     AND id NOT IN (SELECT user_id FROM users_roles WHERE role_id = ?)",
                      role_id_for(:trusted),
+                     role_id_for(:banned),
                    ).order("users.comments_count" => :desc)
                  else
                    relation.joins(:roles)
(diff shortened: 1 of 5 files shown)


citizen428:
cc @rhymes who recently opened https://github.com/forem/forem/issues/10734

pkfrank (approved):
Copy-related changes look great to me.

michael-tharrington (approved):
I think this looks excellent. I rarely review PRs in this fashion, so left my comments in the wrong place. 🙃

Only real comment:

I previously proposed an updated version of the automated ~community mod~ trusted user email here - https://github.com/thepracticaldev/tech-private/issues/454. We could use elements of it when updating the mod email now, later, or I can just scrap it. To be clear, this isn't blocking at all if you want to just go forward with how things are!

jacobherrington (approved):
Makes sense to me :+1:

PR #10817, opened 2020-10-12
Skip edge caching for private forems

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently if you use a "private" Forem, the ability to peak into internal pages could be spoofed, because the way we differentiate at the edge is not meant for data that needs to remain private. Privacy on DEV is currently reserved for pages which don't add edge cache headers, but our "high traffic public pages" all have heavy caching....

In order to perform caching _within_ private communities we will either want a solution for something akin to Action Caching in place where edge caching would otherwise be, because this kind of caching can happen _after_ authentication. There may also be ways we could do auth _within_ nginx in a clever way in the future, but that path is less obvious to me.

Either way, this PR makes the tradeoff to enable _truly_ private Forems now while we figure out how to do approach caching. Luckily, private forems are, in general, more specialized and lower traffic. They will be, by definition, more capable of dealing with this transition.

### DIFF
--- a/spec/requests/stories_index_spec.rb
+++ b/spec/requests/stories_index_spec.rb
@@ -104,6 +104,15 @@ RSpec.describe "StoriesIndex", type: :request do
       expect(response.body).to include(CGI.escapeHTML(listing.title))
     end
 
+    it "does not set surrogate headers if private" do
+      allow(SiteConfig).to receive(:public).and_return(false)
+      get "/"
+      expect(response.status).to eq(200)
+
+      expected_surrogate_key_headers = %w[main_app_home_page]
+      expect(response.headers["Surrogate-Key"].split(", ")).not_to match_array(expected_surrogate_key_headers)
+    end
+
     it "sets Fastly Surrogate-Key headers" do
       get "/"
       expect(response.status).to eq(200)
--- a/spec/requests/stories_index_spec.rb
+++ b/spec/requests/stories_index_spec.rb
@@ -104,11 +104,13 @@ RSpec.describe "StoriesIndex", type: :request do
       expect(response.body).to include(CGI.escapeHTML(listing.title))
     end
 
-    it "does not set surrogate headers if private" do
+    it "does not set cache-related headers if private" do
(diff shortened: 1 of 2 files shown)


spec/requests/stories_index_spec.rb
```diff
@@ -104,6 +104,15 @@
       expect(response.body).to include(CGI.escapeHTML(listing.title))
     end
 
+    it "does not set surrogate headers if private" do
+      allow(SiteConfig).to receive(:public).and_return(false)
+      get "/"
+      expect(response.status).to eq(200)
+
+      expected_surrogate_key_headers = %w[main_app_home_page]
+      expect(response.headers["Surrogate-Key"].split(", ")).not_to match_array(expected_surrogate_key_headers)
```
benhalpern, line 113:
```suggestion
      expect(response.headers["Surrogate-Key"]).to eq(nil)
```

spec/requests/stories_index_spec.rb
```diff
@@ -104,6 +104,14 @@
...
     end
 
+    it "does not set surrogate headers if private" do
+      allow(SiteConfig).to receive(:public).and_return(false)
+      get "/"
+      expect(response.status).to eq(200)
+
+      expect(response.headers["Surrogate-Key"]).to eq(nil)
+    end
+
```
atsmith813, line 114: I believe this should check `Cache-Control` and `X-Accel-Expires` headers instead of or in addition to `Surrogate-Key`. The surrogate keys are only used for busting on Fastly. The `Cache-Control` and `X-Accel-Expires` headers are used by Fastly and Nginx (respectively) to dictate what's cached or not.

vaidehijoshi: Yep, can confirm; I think if just add an assertion to check that both `response.headers["X-Accel-Expires"]` and ``response.headers["Cache-Control"]` are `nil`, we should be good to go here 👍

atsmith813 (approved):
Thanks!

PR #10818, opened 2020-10-12
Feature:Add Video States Encoder Key to SiteConfig

mstruve:
- [x] Refactor
- [x] Feature

## Description
Currently the AWS video encoder key we use to encode videos is linked to a single user in our database. This should not be the case since this value is used for the entire application. This moves the encoder value to SiteConfig so that it can easily be set by the site admin. Once this is merged then I will update the VideoStatesController to use the new SiteConfig value rather than the individual user's secret. 

As far as I know @leewynne is the only other Forem using this feature. I will be sure to followup with him on getting this key setup properly on his Forem. 

https://github.com/orgs/forem/projects/6#card-46047741

## Added tests?
- [x] yes

alt_text

### DIFF
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -62,6 +62,7 @@ module Admin
         spam_trigger_terms
         recaptcha_site_key
         recaptcha_secret_key
+        video_encoder_key
       ]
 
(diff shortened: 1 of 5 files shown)


maestromac (approved):
LGTM!

citizen428 (approved):

PR #10820, opened 2020-10-12
Ensure serviceworker bypass only checks path

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This functionality _naively_ checks for which paths to skip over for service workers. The reason it can be pretty naive is that it is not that big a deal if it's a little too aggressive. Occasionally skipping service workers is fine.

But one problem is that `https://forem.dev` _includes `/forem`........ Which is a problem because it _always_ skipped over and did nothing. This changes the code to check only path, which is what it should have been before.

There is definitely still a need to make this more precise, but for now this solves this particular defect.

### DIFF
--- a/app/views/service_worker/index.js.erb
+++ b/app/views/service_worker/index.js.erb
@@ -128,7 +128,7 @@
       '/users/auth',           // Don't run on authentication.
       '/welcome',              // Don't run on welcome reroutes.
       '/workshops',            // redirects
-    ].some(path => url.href.includes(path))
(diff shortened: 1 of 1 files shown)


atsmith813 (approved):

Zhao-Andy (approved):

juliannatetreault (approved):
👍

PR #10828, opened 2020-10-13
Create framework for custom seed files and add notification seed file

Zhao-Andy:
- [x] (Development) Feature

## Description
This PR lays down the framework for creating custom seed files. Custom seed files will allow us to more specific data when needed, beyond the initial `rails db:seed`.

This PR also adds the notification seed file to generate examples of every possible notification we have.

1. Run `rails db:seed:notifications USERNAME="your_username"`
2. Check `/notifications`  to see if the notifications loaded properly

## Added tests?

## Added to documentation?
- [x] docs.forem.com

Dean Winchester from Supernatural jams out excitedly in his car.

### DIFF
--- a/db/custom_seeds/notifications.rb
+++ b/db/custom_seeds/notifications.rb
@@ -11,6 +11,15 @@ if user.blank?
   raise StandardError
 end
 
+if ENV["CLEAN_SETUP"].present?
+  puts "starting with none of your notifications, followers, articles, and comments..."
+  user.notifications.delete_all
+  Follow.where(followable: user).destroy_all
+  Follow.where(follower: user).destroy_all
+  user.articles.destroy_all
+  user.comments.destroy_all
+end
+
 seeder = Seeder.new
 
 puts "0/#{NUM_TOTAL_STEPS} checking requirements..."
--- a/db/custom_seeds/notifications.rb
+++ b/db/custom_seeds/notifications.rb
@@ -5,51 +5,153 @@
 
 NUM_TOTAL_STEPS = 11
 
-user = User.find_by(username: ENV["USERNAME"]) || User.last
+user = User.find_by(username: ENV["USERNAME"])
+if user.blank?
+  puts "Your user not found. Please ensure the username is present or that the command was properly run: rails db:seed:notifications USERNAME='your_username'"
+  raise StandardError
+end
 
-puts "1/#{NUM_TOTAL_STEPS} seeding follower notifications..."
+seeder = Seeder.new
+
+puts "0/#{NUM_TOTAL_STEPS} checking requirements..."
+users = User.where(id: 1..6).count == 6 ? User.where(id: 1..6) : []
+((1..6).to_a - users.pluck(:id)).each do |id|
+  user = User.create!(
+    id: id,
+    name: name,
+    summary: Faker::Lorem.paragraph_by_chars(number: 199, supplemental: false),
+    profile_image: File.open(Rails.root.join("app/assets/images/#{rand(1..40)}.png")),
+    website_url: Faker::Internet.url,
+    twitter_username: Faker::Internet.username(specifier: name),
+    email_comment_notifications: false,
+    email_follower_notifications: false,
+    # Emails limited to 50 characters
+    email: Faker::Internet.email(name: name, separators: "+", domain: Faker::Internet.domain_word.first(20)),
+    confirmed_at: Time.current,
+    password: "password",
+    password_confirmation: "password",
+  )
+  users.push(user)
+end
+
+puts "1/#{NUM_TOTAL_STEPS} seeding broadcast notifications..."
+seeder.create_if_none(Broadcast) do
+  broadcast_messages = {
+    set_up_profile: "Welcome to DEV! 👋 I'm Sloan, the community mascot and I'm here to help get you started. " \
+      "Let's begin by <a href='/settings'>setting up your profile</a>!",
+    welcome_thread: "Sloan here again! 👋 DEV is a friendly community. " \
+      "Why not introduce yourself by leaving a comment in <a href='/welcome'>the welcome thread</a>!",
+    twitter_connect: "You're on a roll! 🎉 Do you have a Twitter account? " \
+      "Consider <a href='/settings'>connecting it</a> so we can @mention you if we share your post " \
+      "via our Twitter account <a href='https://twitter.com/thePracticalDev'>@thePracticalDev</a>.",
+    facebook_connect: "You're on a roll! 🎉  Do you have a Facebook account? " \
+      "Consider <a href='/settings'>connecting it</a>.",
+    github_connect: "You're on a roll! 🎉  Do you have a GitHub account? " \
+      "Consider <a href='/settings'>connecting it</a> so you can pin any of your repos to your profile.",
+    customize_feed: "Hi, it's me again! 👋 Now that you're a part of the DEV community, let's focus on personalizing " \
+      "your content. You can start by <a href='/tags'>following some tags</a> to help customize your feed! 🎉",
+    customize_experience: "Sloan here! 👋 Did you know that that you can customize your DEV experience? " \
+      "Try changing <a href='settings/ux'>your font and theme</a> and find the best style for you!",
+    start_discussion: "Sloan here! 👋 I noticed that you haven't " \
+      "<a href='https://dev.to/t/discuss'>started a discussion</a> yet. Starting a discussion is easy to do; " \
+      "just click on 'Write a Post' in the sidebar of the tag page to get started!",
+    ask_question: "Sloan here! 👋 I noticed that you haven't " \
+      "<a href='https://dev.to/t/explainlikeimfive'>asked a question</a> yet. Asking a question is easy to do; " \
+      "just click on 'Write a Post' in the sidebar of the tag page to get started!",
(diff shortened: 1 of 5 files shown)


lib/tasks/custom_seed.rake
```diff
@@ -0,0 +1,13 @@
+namespace :db do
+  namespace :seed do
+    Dir[Rails.root.join("db/custom_seeds/*.rb")].each do |filename|
+      task_name = File.basename(filename, ".rb")
+      desc "Seed #{task_name}, based on the file with the same name in `db/seeds/*.rb`"
```
rhymes, line 5: `r/seeds/custom_seeds`

rhymes (commented):
I like the simple approach! 👍 

I wonder if we can import "create_if_none" from the seed file as this should be idempotent as much as possible. It also does some assumptions about IDs that might not be true if one uses `rails db:seed:replant` (IDs are not recycled if you truncate).

We also do a lot of `.find()` operations with the same id so those could be cached somewhere.

I think we just need to move the various users used up top and not use hard coded IDs

db/custom_seeds/notifications.rb
```diff
@@ -0,0 +1,269 @@
...
+    user: Notifications.user_data(User.fourth),
+    aggregated_siblings: [
+      Notifications.user_data(User.fourth),
+      Notifications.user_data(User.fifth)
+    ]
+  }
+)
+
+# one follower notification
+follow = User.find(6).follow(user)
```
mstruve, line 34: I would avoid using specific IDs when possible like here. If there is no user with ID 6 the entire script is going to come to a halt. I think by removing specific user IDs the script will be more flexible and can be run on databases in any state, including if someone was testing user deletion and had deleted their ID 6 user. One suggestion for tackling this might be to get the list of ids at the beginning, `User.ids` and then sifting through that and using it to populate some of these.

mstruve (commented):
LOVE THIS!!! One suggestion to help make it a little more robust.

citizen428:
> I wonder if we can import "create_if_none" from the seed 

If we do we should consider moving it to some new module in `lib/` (like `MigrationTools`) and share it between the two places.

> IDs are not recycled if you truncate

You can do a `reset_sequence!` after truncating, but I agree, not relying on specific IDs is a much more robust approach.

lib/tasks/custom_seed.rake
```diff
@@ -0,0 +1,13 @@
+namespace :db do
+  namespace :seed do
+    Dir[Rails.root.join("db/custom_seeds/*.rb")].each do |filename|
```
citizen428, line 3: I prefer the more explicit form, but since `Dir.[]` is an alias for `glob` they are equivalent:

```ruby
Dir.glob(Rails.root.join("db/custom_seeds/*.rb"))
```

Use the one you like better.

lib/tasks/custom_seed.rake
```diff
@@ -0,0 +1,13 @@
...
+      task_name = File.basename(filename, ".rb")
+      desc "Seed #{task_name}, based on the file with the same name in `db/seeds/*.rb`"
+      task task_name.to_sym => :environment do
+        break if Rails.env.production?
```
citizen428, line 7: This will raise an exception:

```
LocalJumpError: break from proc-closure
```

I suggest using `abort` here. It's part of the `Kernel` module, so available everywhere. It also sets a non-zero exit code, indicating failure. 

```suggestion
        abort if Rails.env.production?
```

citizen428 (changes requested):

citizen428:
2 of my comments are just general in nature, but the one re `abort` vs `break` is blocking.

db/custom_seeds/notifications.rb
```diff
@@ -0,0 +1,396 @@
...
+    email: Faker::Internet.email(name: name, separators: "+", domain: Faker::Internet.domain_word.first(20)),
+    confirmed_at: Time.current,
+    password: "password",
+    password_confirmation: "password",
+  )
+  users.push(user)
+end
+
+puts "1/#{NUM_TOTAL_STEPS} seeding broadcast notifications..."
+seeder.create_if_none(Broadcast) do
```
Zhao-Andy, line 49: This is duplicated from `db/seeds.rb` but I'm not sure where else to put this if we didn't want to duplicate it. I think the next best option is in the new `db/seeder.rb` file as a constant?

db/custom_seeds/notifications.rb
```diff
@@ -0,0 +1,396 @@
+# Notification seed file, to mock notifications for a specific user (ideally yourself).
+# To run the command, run:
+# rails db:seed:notifications USERNAME='your_username'
+# you can also add CLEAN_SETUP="y" if you want to start with fresh notifications.
```
Zhao-Andy, line 4: I wish there was a cleaner way to do this -- basically I have to pass in _any_ argument in order for the ENV variable to be `present` :/

db/custom_seeds/notifications.rb
```diff
@@ -0,0 +1,396 @@
...
+    email_comment_notifications: false,
+    email_follower_notifications: false,
+    # Emails limited to 50 characters
+    email: Faker::Internet.email(name: name, separators: "+", domain: Faker::Internet.domain_word.first(20)),
+    confirmed_at: Time.current,
+    password: "password",
+    password_confirmation: "password",
+  )
+  users.push(user)
+end
```
Zhao-Andy, line 46: This could probably be refactored but it works for now 🤷‍♂️

Zhao-Andy (commented):
Here's a preview of all of the notifications if anyone is curious. It's not inline because it is a _long_ screenshot:

https://user-images.githubusercontent.com/17884966/96009084-3b2bf680-0e0e-11eb-8d4c-fe7bb61a3ac0.png

db/custom_seeds/notifications.rb
```diff
@@ -0,0 +1,397 @@
...
+
+seeder = Seeder.new
+
+puts "0/#{NUM_TOTAL_STEPS} checking requirements..."
+users = User.where(id: 1..6).count == 6 ? User.where(id: 1..6) : []
+((1..6).to_a - users.pluck(:id)).each do |id|
+  user = User.create!(
+    id: id,
+    name: name,
```
mstruve, line 32: `name` here needs to be set, it is not defined. Probably want this at the top of the block the `create!`
```ruby
name = Faker::Name.unique.name
```

db/custom_seeds/notifications.rb
```diff
@@ -0,0 +1,397 @@
...
+  Follow.where(followable: user).destroy_all
+  Follow.where(follower: user).destroy_all
+  user.articles.destroy_all
+  user.comments.destroy_all
+end
+
+seeder = Seeder.new
+
+puts "0/#{NUM_TOTAL_STEPS} checking requirements..."
+users = User.where(id: 1..6).count == 6 ? User.where(id: 1..6) : []
```
mstruve, line 28: Here you want to grab and fill in the IDs that are missing. This statement, setting users to an empty array if there are not 6 will cause you to attempt to create all 6 users rather than only the IDs that are missing. I ended up taking a whack at it and got this working
```ruby
puts "0/#{NUM_TOTAL_STEPS} checking requirements..."
((1..6).to_a - User.where(id: 1..6).ids).each do |id|
  name = Faker::Name.unique.name

  user = User.create!(
    id: id,
    name: name,
    summary: Faker::Lorem.paragraph_by_chars(number: 199, supplemental: false),
    profile_image: File.open(Rails.root.join("app/assets/images/#{rand(1..40)}.png")),
    website_url: Faker::Internet.url,
    twitter_username: Faker::Internet.username(specifier: name),
    email_comment_notifications: false,
    email_follower_notifications: false,
    # Emails limited to 50 characters
    email: Faker::Internet.email(name: name, separators: "+", domain: Faker::Internet.domain_word.first(20)),
    confirmed_at: Time.current,
    password: "password",
    password_confirmation: "password",
  )
end
users = User.where(id: 1..6)
```

mstruve (commented):
Works GREAT if you have a fresh database. If you are missing users though there are still some errors that need to be worked out. I offered suggestions for fixing the user creation. 

Once I had the user creation fixed, I removed all notifications that were made in previous runs and tried again. This time we got pretty far down the script but it died with this error. 
```
5/11 seeding single reaction on comment notification...
6/11 seeding comment moderation notification...
rails aborted!
NoMethodError: undefined method `id' for nil:NilClass
/Users/mollystruve/forem/app/services/notifications.rb:27:in `comment_data'
/Users/mollystruve/forem/app/services/notifications/moderation/send.rb:14:in `comment_data'
/Users/mollystruve/forem/app/services/notifications/moderation/send.rb:24:in `public_send'
/Users/mollystruve/forem/app/services/notifications/moderation/send.rb:24:in `call'
/Users/mollystruve/forem/app/services/notifications/moderation/send.rb:11:in `call'
/Users/mollystruve/forem/db/custom_seeds/notifications.rb:298:in `<main>'
```
I poked around a little but nothing stood out as to the obvious cause so you might want to look further into that.

Zhao-Andy:
Not sure why I didn't see these reviews, but will update!

msarit:
Hey @Zhao-Andy! 👋🏾 
Drawing your attention to this PR 😉

Zhao-Andy:
Aha thanks @msarit, going to close and make into an RFC.

PR #10830, opened 2020-10-13
Flaky Spec Fix: Wait for button before trying to click it

mstruve:
- [x] Spec Fix

## Description
It appears that sometimes during this spec we cannot find the Subscription button. I believe this happens when the page loads and we check for the button before the javascript has time to create the tag. I tested this by duplicating the failing spec 10 times and running the file. It would pretty consistently get the error below without waiting for the javascript. 

alt_text

### DIFF
--- a/spec/liquid_tags/user_subscription_tag_spec.rb
+++ b/spec/liquid_tags/user_subscription_tag_spec.rb
@@ -46,6 +46,7 @@ RSpec.describe UserSubscriptionTag, type: :liquid_tag do
     before do
       sign_in subscriber
       visit article_with_user_subscription_tag.path
+      wait_for_javascript
     end
 
     it "shows the signed in UX" do
--- a/spec/liquid_tags/user_subscription_tag_spec.rb
+++ b/spec/liquid_tags/user_subscription_tag_spec.rb
@@ -46,13 +46,16 @@ RSpec.describe UserSubscriptionTag, type: :liquid_tag do
     before do
       sign_in subscriber
       visit article_with_user_subscription_tag.path
-      wait_for_javascript
     end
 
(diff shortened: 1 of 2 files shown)


spec/liquid_tags/user_subscription_tag_spec.rb
```diff
@@ -46,6 +46,7 @@
     before do
       sign_in subscriber
       visit article_with_user_subscription_tag.path
+      wait_for_javascript
```
citizen428, line 49: My first thought when seeing this was that explicit waiting in Capybara tests was always considered to be a bit of an anti-pattern, since Capybara has waiting assertions. So I looked at the `wait_for_javascript` helper, which leads to this [Medium post](https://medium.com/doctolib/hunting-flaky-tests-2-waiting-for-ajax-bd76d79d9ee9). It is indeed where we took this method from, but it's listed under the headline "If you are lazy, blindly wait for ajax". 

I have no problem with merging this, as the flaky specs got quite annoying again recently. But we probably need another concereted effort around improving the quality of our test suite, maybe during the next downtime.

citizen428 (approved):

spec/liquid_tags/user_subscription_tag_spec.rb
mstruve: Agree, I haven't marked this as ready yet bc I was reproducing successfully but then was not able to but your suggestion has given me another idea to try and work in some waiting assertions.

mstruve:
TIL that [`click_button' will actually wait for the button to be visible up to the default_max_wait_time](https://rubydoc.info/github/jnicklas/capybara/master/Capybara/Node/Actions#click_button-instance_method) so this wont help.

PR #10831, opened 2020-10-13
Comment out Flaky Basic Feed Spec

mstruve:
- [x] Spec ignore

## Description
This spec has failed quite a few builds (example: https://travis-ci.com/github/forem/forem/jobs/398925392) and based on what the Basic Feed class actually does I dont think this is even testing the correct thing. Commenting it out to unblock deploys until we can get it fixed.

### DIFF
--- a/spec/services/articles/feeds/basic_spec.rb
+++ b/spec/services/articles/feeds/basic_spec.rb
@@ -9,7 +9,7 @@ RSpec.describe Articles::Feeds::Basic, type: :service do
   let!(:low_scoring_article) { create(:article, score: -1000) }
   let!(:month_old_story) { create(:article, published_at: 1.month.ago) }
 
-  it "returns articles in approximately published order" do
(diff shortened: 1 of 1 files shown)


spec/services/articles/feeds/basic_spec.rb
```diff
@@ -9,7 +9,7 @@
   let!(:low_scoring_article) { create(:article, score: -1000) }
   let!(:month_old_story) { create(:article, published_at: 1.month.ago) }
 
-  it "returns articles in approximately published order" do
+  xit "returns articles in approximately published order" do
```
vaidehijoshi, line 12: is a fix in progress? if so, i think it's fine as is. but if not, we might want to put a `TODO` comment in here since we very intentionally are skipping it for a specific reason

vaidehijoshi (approved):

juliannatetreault (approved):

PR #10834, opened 2020-10-13
Add releasedate meta data

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

It will prove useful to know when a forem was released in whatever context we need to view it.

This could complement well other meta data such as release version, to give something humans and computers can use as needed when crawling meta data.

This also modifies the Heroku version of `RELEASE_FOOTPRINT` to check for the date instead of commit slug because the way we calculate `RELEASE_FOOTPRINT` is by date in Forem Cloud.

I feel like we could use this in the short term for more info for humans to better understand when a forem page was released (or possibly cached) and this is a small thing that could help in our big picture.

### DIFF
--- a/app/views/shell/_top.html.erb
+++ b/app/views/shell/_top.html.erb
@@ -50,6 +50,7 @@
       <meta property="forem:name" content="<%= community_name %>" />
       <meta property="forem:logo" content="<%= optimized_image_url(SiteConfig.logo_png, width: 512, fetch_format: "png") %>" />
       <meta property="forem:domain" content="<%= SiteConfig.app_domain %>" />
+      <meta property="forem:releasedate" content="<%= ENV["RELEASE_FOOTPRINT"] %>" />
     <% end %>
   </head>
   <% unless internal_navigation? %>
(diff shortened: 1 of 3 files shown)


app/views/shell/_top.html.erb
```diff
@@ -50,6 +50,7 @@
       <meta property="forem:name" content="<%= community_name %>" />
       <meta property="forem:logo" content="<%= optimized_image_url(SiteConfig.logo_png, width: 512, fetch_format: "png") %>" />
       <meta property="forem:domain" content="<%= SiteConfig.app_domain %>" />
+      <meta property="forem:releasedate" content="<%= ENV["RELEASE_FOOTPRINT"] %>" />
```
mstruve, line 53: We already have [this data below in our body tag](https://github.com/forem/forem/pull/10834/files#diff-54c7373a1a9cafdb686788688f850f3a4469b23c6b6d48e39751a02b54795caaR64) Duplicating it feels unnecessary, should we remove the one below in favor of this one?

mstruve (commented):
I think if we want to add metadata for the release date/time we should make it separate from the release footprint rather than using it as the release footprint when a unique SHA is not present.

Issue #4807, opened 2019-11-13
Highlight or Prevent Heading Level Errors in Markdown 


aitchiss:
**Is your feature request related to a problem? Please describe.**

This issue is coming out of a discussion on post https://dev.to/s_aitchison/getting-heading-levels-right-including-here-on-dev-439

A substantial proportion of posts on DEV break WCAG guidance on heading levels. The post title automatically becomes the `h1` of the page; from that point on heading levels should only increase by 1. 

What you actually see in reality is a lot of additional `h1` elements, and then quite often non-sequential heading levels beyond that.

It would be a boost for accessibility if we could help limit or prevent this in some way. As a bit of context, the latest WebAIM screen reader user survey indicated that close to 70% of users prefer to navigate by heading level, and the ability to scan content this way is compromised by non-WCAG compliant heading usage.

**Describe the solution you'd like**

When viewing a post in 'preview' mode, it would be good to see heading level issues flagged in the "Heads up" warning that appears at the top of the page. At a minimum, flagging duplicate `h1`s, but ideally linting for other issues in the heading levels (e.g. an `h3` being present but no `h2`)

colabottles:
@nickytonline I can take a look at this if you want to let me have a crack at this one.

techieeliot:
Hey @nickytonline, I'd like to tackle this one tomorrow.

nickytonline:
Sounds good. I've assigned this to you.

techieeliot:
# Does this look like an H1 in github

Note that we're looking at options: 

- markdown-lint  - https://github.com/DavidAnson/markdownlint

techieeliot:
## What we discovered on the live stream

- @nickytonline @cmgorton and @techieeliot we toyed around with the markdown lint demo from the docs
- in the docs we tested out the following code in the articleForm.js at `app/javascript/packs/`: 

```
import markdownlint from 'markdown-lint';

const options = {
  "strings": {
    "good.string": "# good.string\n\nThis string passes all rules.",
    "bad.string": "#bad.string\n\n#This string fails\tsome rules."
  }
};

markdownlint(options, function callback(err, result) {
  if (!err) {
    console.log(result.toString());
  }
});
```

-  we discovered after trial and error that we did not need to import it, so we removed it and swiped the following docs from the source and placed them into the path `app/assets/javascripts` then applied the following into the base.js.erb: 

```
//= require markdown-it.min.js
//= require markdownlint-browser.min.js
```

[...]

PR #10837, opened 2020-10-13
Begin cleaning up and generalizing events

benhalpern:

- [x] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Let's rewind the clock a few years..... DEV decides to put on live online events as a component of a membership program designed to deliver community value and help develop out a component of the business model.

Despite a good deal success, the overhead of the initiative proves a little much given we're also building out the technical platform and we are radically under-resourced.

Fast forward a few years, we're now rolling out Forems, and this feature we never fully said goodbye to is intriguing to several Forem creators— for good reason. Events make up a good deal of what "community" can be. It's also a feature social platforms like Facebook normalized as a component of the overall experience. DEV itself is now back to putting on events and could use this feature IMO.

We said goodbye to _events_ in favor of _listings_, but realistically both can serve a place within a community. Right now as events can be something _the community_ puts on, and the events category of listings can be more ad hoc. This could change over time, but all in all with a certain amount of reflection I'm glad we never let the ship sail on this feature as a component of the ecosystem.

All in all, the basics of events are all in place, but they need to be cleaned up and generalized. This is a small initial PR to get the ball rolling.

See events live in all their glory at https://dev.to/events

......... It's possible events could be _extracted_ into a plug-in. I'm going to keep my eyes on this possibility as I look into this.

### DIFF
--- a/app/views/events/index.html.erb
+++ b/app/views/events/index.html.erb
@@ -28,9 +28,9 @@
     <% end %>
 
     <% if @events.none? %>
-      <h3 class="mt-0 color-base-70">
+      <div class="crayons-notice">
         There are no upcoming events currently scheduled
-      </h3>
+      </div>
     <% end %>
 
     <% if @past_events.any? %>
--- a/app/views/events/index.html.erb
+++ b/app/views/events/index.html.erb
@@ -7,40 +7,37 @@
   <meta property="og:type" content="article" />
   <meta property="og:url" content="<%= app_url("/events") %>" />
   <meta property="og:title" content="<%= community_name %> Events" />
-  <meta property="og:image" content="https://thepracticaldev.s3.amazonaws.com/i/bhavxzyx2e7bdsyvysve.jpg" />
-  <meta property="og:description" content="Talks and workshops designed to help community members level up." />
+  <meta property="og:description" content="Community events." />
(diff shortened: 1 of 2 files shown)


app/views/events/index.html.erb
```diff
@@ -7,40 +7,37 @@
...
 
     <% @events.each do |event| %>
       <%= render "event", event: event %>
     <% end %>
 
-    <h1 style="font-size:1.9em;font-weight:600;text-align:center;">
-      PAST <br> EVENTS
-    </h1>
+    <% if @events.none? %>
+      <h3 class="mt-0 color-base-70">
```
citizen428, line 31: We have an open issue about using heading tags in a way that's beneficial for screen-readers: https://github.com/forem/forem/issues/4807 Is this really a heading or more of a case of using an h3 for styling purposes?

benhalpern: Good call. I changed this to a notice, which is definitely more correct for both a11y and overall guidelines.

Screen Shot 2020-10-14 at 10 23 27 AM

citizen428 (approved):

pkfrank (approved):
Looks great.  A few Forem Admins will be really pumped about the start of this clean up + generalization effort.

PR #10838, opened 2020-10-13
Make allow email password login default true

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related issue

Closes https://github.com/forem/InternalProjectPlanning/issues/91

## Description

Technically regardless of registration mechanism, the _default_ for logging in should be to allow this, because even someone who has signed up with social or invitation can _technically_ log in. If a forem wants to disable that for any reason they could do so.

At least for now I think this makes sense.

### DIFF
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -23,7 +23,7 @@ class SiteConfig < RailsSettings::Base
 
   # Authentication
   field :allow_email_password_registration, type: :boolean, default: false
-  field :allow_email_password_login, type: :boolean, default: false
+  field :allow_email_password_login, type: :boolean, default: true
   field :authentication_providers, type: :array, default: proc { Authentication::Providers.available }
   field :invite_only_mode, type: :boolean, default: false
   field :twitter_key, type: :string, default: ApplicationConfig["TWITTER_KEY"]
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -6,7 +6,7 @@ module Constants
         placeholder: ""
       },
       allow_email_password_login: {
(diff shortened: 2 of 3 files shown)


app/models/site_config.rb
```diff
@@ -23,7 +23,7 @@ class SiteConfig < RailsSettings::Base
 
   # Authentication
   field :allow_email_password_registration, type: :boolean, default: false
-  field :allow_email_password_login, type: :boolean, default: false
+  field :allow_email_password_login, type: :boolean, default: true
```
vaidehijoshi, line 26: is [this description](https://github.com/forem/forem/blob/7ce3b8abd25fb2424e67b07d18d11d9676e1b2a5/app/lib/constants/site_config.rb#L9) still correct? if we default it to being on/`true`, then we're adding it as an additional auth option, but not the _only_ option, right?

wondering if we should update the wording there because someone might be confused why it defaults to being on but other auth providers also show up when you go to sign in

vaidehijoshi:
adding a note that this PR will close out [this issue](https://github.com/forem/InternalProjectPlanning/issues/91)

app/models/site_config.rb
benhalpern: I'll revisit the functionality to make sure I understand it all in the morning.

juliannatetreault (approved):
Thanks for making this change! 👍

app/models/site_config.rb
benhalpern: Okay I looked at this and I think this makes sense and is straightforward. I updated the copy in the description.

PR #10826, opened 2020-10-13
SVG logo image re-sizing

Ridhwana:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The DEV SVG logo was being resized to 20% height and width on the actual uploaded SVG (and not within the codebase). This allows logo's that are of similar size to work really nicely within our site's header. However, when other forems try to paste an SVG in the config that have huge heights and widths, it causes the header to have a meltdown 😅 See [this issue](https://github.com/forem/InternalProjectPlanning/issues/53) for examples of the meltdown 😊

Hence, I've added some css to allow the header to always accommodate the SVG irrespective of the size that of the SVG that was pasted. In addition, I've also made the site config logo svg preview use the same html/css that is being used on the header to allow a more accurate representation.

## Related Tickets & Documents
https://github.com/forem/InternalProjectPlanning/issues/53

## QA Instructions, Screenshots, Recordings

1. Go to http://localhost:3000/admin/config
2. Look for the logo svg
3. Upload any huge or small svg logo (with or without) heights set and it should resize appropriately to fit in the header.
4. In addition, you will see the resized logo being displayed in the config as well. 

**Here are some existing logos that weren't working that now resize properly:**
Screenshot 2020-10-13 at 13 31 35
Screenshot 2020-10-13 at 13 10 36

**Here is a random 48x48 width svg that I've uploaded:**
Screenshot 2020-10-13 at 13 36 11

**Here are some existing logos that are still working with the fix:**
Screenshot 2020-10-13 at 13 39 54

Screenshot 2020-10-13 at 13 32 17

**Here is the config page that shows the resized logo as it would be displayed on the header:**
Screenshot 2020-10-13 at 13 42 44

## Added tests?

- [ ] yes
- [x] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

## [optional] What gif best describes this PR or how it makes you feel?

alt_text

PR #10840, opened 2020-10-14
Logo updates

ludwiczakpawel:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR does few things:

- it helps generalize the logo_svg and drops all DEV-only styling we used to have for our logo
- it takes #10826 a step further and brings similar fix to other places where we use logo (footer and `/new`)
- it **prepares** our codebase to make `logo_svg` **not**  required field: https://github.com/forem/InternalProjectPlanning/issues/83 - in that case it would display a `community_name`. To be more specific: this PR prepares frontend part of our codebase to be able to serve `community_name`. **This PR does not make `logo_svg` optional field** - we'll have another PR for that (see linked issue above).

- #10826 
- https://github.com/forem/InternalProjectPlanning/issues/53
- https://github.com/forem/InternalProjectPlanning/issues/83

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

## 🚨🚨🚨 Are there any post deployment tasks we need to perform? 🚨🚨🚨

Update our `logo_svg` field in `/admin/config` on prod with the following (cc @benhalpern):

```
<svg width="50" height="40" viewBox="0 0 50 40" fill="none" xmlns="http://www.w3.org/2000/svg"><rect width="50" height="40" rx="3" style="fill: currentColor;"/><path d="M19.099 23.508c0 1.31-.423 2.388-1.27 3.234-.838.839-1.942 1.258-3.312 1.258h-4.403V12.277h4.492c1.31 0 2.385.423 3.224 1.27.846.838 1.269 1.912 1.269 3.223v6.738zm-2.808 0V16.77c0-.562-.187-.981-.562-1.258-.374-.285-.748-.427-1.122-.427h-1.685v10.107h1.684c.375 0 .75-.138 1.123-.415.375-.285.562-.708.562-1.27zM28.185 28h-5.896c-.562 0-1.03-.187-1.404-.561-.375-.375-.562-.843-.562-1.404V14.243c0-.562.187-1.03.562-1.404.374-.375.842-.562 1.404-.562h5.896v2.808H23.13v3.65h3.088v2.808h-3.088v3.65h5.054V28zm7.12 0c-.936 0-1.684-.655-2.246-1.965l-3.65-13.758h3.089l2.807 10.804 2.808-10.804H41.2l-3.65 13.758C36.99 27.345 36.241 28 35.305 28z" style="fill: var(--base-inverted);"/></svg>
```

### DIFF
--- a/app/javascript/article-form/articleForm.jsx
+++ b/app/javascript/article-form/articleForm.jsx
@@ -47,7 +47,7 @@ export default class ArticleForm extends Component {
     version: PropTypes.string.isRequired,
     article: PropTypes.string.isRequired,
     organizations: PropTypes.string,
-    logoSvg: PropTypes.string.isRequired,
+    siteLogo: PropTypes.string.isRequired,
   };
 
   static defaultProps = {
@@ -56,7 +56,7 @@ export default class ArticleForm extends Component {
 
   constructor(props) {
     super(props);
-    const { article, version, logoSvg } = this.props;
+    const { article, version, siteLogo } = this.props;
     let { organizations } = this.props;
     this.article = JSON.parse(article);
     organizations = organizations ? JSON.parse(organizations) : null;
@@ -101,7 +101,7 @@ export default class ArticleForm extends Component {
       edited: false,
       updatedAt: this.article.updated_at,
       version,
-      logoSvg,
+      siteLogo,
       helpFor: null,
       helpPosition: null,
       ...previousContentState,
@@ -300,7 +300,7 @@ export default class ArticleForm extends Component {
       version,
       helpFor,
       helpPosition,
-      logoSvg,
+      siteLogo,
     } = this.state;
 
     return (
@@ -316,7 +316,7 @@ export default class ArticleForm extends Component {
           organizations={organizations}
           organizationId={organizationId}
           onToggle={this.handleOrgIdChange}
-          logoSvg={logoSvg}
+          siteLogo={siteLogo}
         />
 
         {previewShowing ? (
--- a/app/views/articles/_v2_form.html.erb
+++ b/app/views/articles/_v2_form.html.erb
@@ -17,7 +17,6 @@
   data-article="<%= article.to_json(only: %i[id title cached_tag_list published body_markdown main_image organization_id canonical_url updated_at], methods: %i[series all_series]) %>"
   data-organizations="<%= organizations&.to_json(only: %i[id name bg_color_hex text_color_hex], methods: [:profile_image_90]) %>"
   data-version="<%= version %>"
-  data-logo-svg='<%= logo_svg %>'
   data-site-logo='<%= render "layouts/logo" %>'>
   <form class="crayons-article-form crayons-article-form--<%= version %>">
 
--- a/app/views/articles/_v2_form.html.erb
+++ b/app/views/articles/_v2_form.html.erb
@@ -17,11 +17,14 @@
   data-article="<%= article.to_json(only: %i[id title cached_tag_list published body_markdown main_image organization_id canonical_url updated_at], methods: %i[series all_series]) %>"
   data-organizations="<%= organizations&.to_json(only: %i[id name bg_color_hex text_color_hex], methods: [:profile_image_90]) %>"
   data-version="<%= version %>"
-  data-logo-svg='<%= logo_svg %>'>
+  data-logo-svg='<%= logo_svg %>'
+  data-site-logo='<%= render "layouts/logo" %>'>
   <form class="crayons-article-form crayons-article-form--<%= version %>">
 
     <div class="crayons-article-form__header">
-      <a href="/" class="crayons-article-form__logo" aria-label="<%= community_name %> Home"><%= logo_svg %></a>
+      <span class="crayons-article-form__logo">
+        <%= render "layouts/logo" %>
+      </span>
 
       <div class="crayons-field__label">
         <span class="hidden s:inline-block mr-2">Write a new post</span>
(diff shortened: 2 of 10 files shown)


app/javascript/article-form/articleForm.jsx
```diff
@@ -47,7 +47,7 @@ export default class ArticleForm extends Component {
     version: PropTypes.string.isRequired,
     article: PropTypes.string.isRequired,
     organizations: PropTypes.string,
-    logoSvg: PropTypes.string.isRequired,
```
Ridhwana, line 50: **(question)**: Was this maybe not named this way to make it easier to differentiate between logo_png and logo_svg that come from the site config (w/o having to trace back to the start)?

app/views/layouts/_logo.html.erb
```diff
@@ -0,0 +1,9 @@
+<a href="/" class="site-logo" aria-label="<%= community_name %> Home">
+  <% if SiteConfig.logo_svg.present? %>
```
Ridhwana, line 2: Should we not be using just `if logo_svg.present?` here?

app/views/articles/_v2_form.html.erb
```diff
@@ -17,11 +17,14 @@
   data-article="<%= article.to_json(only: %i[id title cached_tag_list published body_markdown main_image organization_id canonical_url updated_at], methods: %i[series all_series]) %>"
   data-organizations="<%= organizations&.to_json(only: %i[id name bg_color_hex text_color_hex], methods: [:profile_image_90]) %>"
   data-version="<%= version %>"
-  data-logo-svg='<%= logo_svg %>'>
+  data-logo-svg='<%= logo_svg %>'
```
Ridhwana, line 20: *(question)* Is this variable still being used somewhere or should it be removed?

app/views/layouts/_logo.html.erb
Ridhwana: Ah so logo svg seems to be this method in application_helper.rb:

```
  def logo_svg
    if SiteConfig.logo_svg.present?
      SiteConfig.logo_svg.html_safe # rubocop:disable Rails/OutputSafety
    else
      inline_svg_tag("devplain.svg", class: "logo", size: "20% * 20%", aria: true, title: "App logo")
    end
  end
```

Hence, I see why you used `SiteConfig.logo_svg.present?` instead of `logo_svg.present?` for this if statement.
However, then it would seem like the else statement in logo_svg is useless as we never want it to render devplain, but rather the text. Perhaps you can check and we can remove it? 

In that case also, perhaps we wouldn't need to pass logo_svg all around, but we could just use the SiteConfig variable if we get rid of the else.

app/javascript/article-form/articleForm.jsx
ludwiczakpawel: I don't know the purpose of its original name, but I've decided to rename it since it will no longer be only about logo_svg. The logic we're about to ship soon is following:

- `logo_svg` is not longer required field.
- if you upload `logo_svg`, sweet - this gonna be your loog.
- if you don't upload `logo_svg` we will use site name (actual text label) as placeholder.

And because this can either be `SVG` element **OR just a text label**, the name "_logoSvg_" no longer makes sense.

Ridhwana:
@ludwiczakpawel I always find is useful to have reproduction steps or screenshots in the PR. It helps me get the idea of what the PR is doing more easily 😊

app/views/layouts/_logo.html.erb
ludwiczakpawel: Um, I guess so. Fixing :)

app/views/articles/_v2_form.html.erb
ludwiczakpawel: Nah, I forgot to remove it. Good catch!

ludwiczakpawel:
@Ridhwana but it doesn't really change anything visually so not sure what should be the screenshot? If you're talking about the wordmark (text label) being displayed instead of logo_svg - this won't actually happen because we still **require** logo_svg so this scenario is not possible. This PR only prepares frontend for such a scenario in close future (another PR). If you want to test it locally, you'll have to make code changes manually I believe.

app/views/layouts/_logo.html.erb
ludwiczakpawel: Re application_helper.rb - that's a good catch! I was not aware of this `if` statement in logo_svg. I think you're 100% right - we should drop it. But I believe that means refactoring some other bits where we use it, including all the logic that makes the logo_svg required. And THAT in particular is actually something I don't wanna touch because it's back-end yada yada yada... But this is exactly what we wanna handle separately: https://github.com/forem/InternalProjectPlanning/issues/83

Does that make sense?

Ridhwana:
> @Ridhwana but it doesn't really change anything visually so not sure what should be the screenshot? If you're talking about the wordmark (text label) being displayed instead of logo_svg - this won't actually happen because we still **require** logo_svg so this scenario is not possible. This PR only prepares frontend for such a scenario in close future (another PR). If you want to test it locally, you'll have to make code changes manually I believe.

Ah I see, that would have been something good to know in the reproduction steps 😅

app/views/layouts/_logo.html.erb
Ridhwana: Coolios, happy with it being removed and the code cleaned up in a follow up PR 👍

Ridhwana:
Screenshot 2020-10-14 at 14 30 00
Whoop whoop this works nicely once we remove logo svg as mandatory. Nicely done @ludwiczakpawel

Ridhwana (approved):
Looks good to me 🚀 Some cleanup to be done in a subsequent PR for logo_svg.

benhalpern (approved):
These look good to me. I'll merge this when I'm in a good place to make sure I get the logo svg swapped in for DEV.

PR #10841, opened 2020-10-14
Fix ui for signup with email 

Ridhwana:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
When you try and create an account via email, the UI is broken. The email svg is not getting rendered as html but rather as a string.

Screenshot 2020-10-14 at 11 17 00

https://github.com/forem/InternalProjectPlanning/issues/142

Go to http://localhost:3000/enter?state=new-user, and you should see a fixed version:
Screenshot 2020-10-14 at 11 39 51

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

alt_text

### DIFF
--- a/app/views/shared/authentication/_providers_registration_form.html.erb
+++ b/app/views/shared/authentication/_providers_registration_form.html.erb
@@ -9,7 +9,7 @@
     </a>
   <% end %>
   <% if params[:state] == "new-user" && SiteConfig.allow_email_password_registration %>
-    <%= link_to "#{inline_svg_tag('email.svg', aria: true, class: 'crayons-icon', title: 'email')}Sign up with Email",
+    <%= link_to "#{inline_svg_tag('email.svg', aria: true, class: "crayons-icon", title: "email")}Sign up with Email".html_safe,
                 request.params.merge(state: "email_signup"),
                 class: "crayons-btn crayons-btn--l crayons-btn--brand-email crayons-btn--icon-left whitespace-nowrap",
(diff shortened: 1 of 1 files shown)


ludwiczakpawel (approved):
👍

I do wonder why that doesn't happen on some other communities? For example https://www.thismmalife.com/enter?state=new-user ?

Ridhwana:
> 👍
> 
> I do wonder why that doesn't happen on some other communities? For example https://www.thismmalife.com/enter?state=new-user ?

@ludwiczakpawel this is such a good question, and so I went digging around for the answer. 

It seems that communities like mmalife have not deployed in a couple of days. [The commit that caused this](https://github.com/forem/forem/commit/3377b5b69e36250747adc1a9d658eefa772c8a82) was only committed 14 hours ago. 

The commit simply changed the link from using concatenation to interpolation which is a common thing to do. I would not have expected it to break either. Hence, I feel like this PR using interpolation and html safe with it (instead of reverting it to concatenation) will safeguard us from someone down the line reverting to just interpolation again and mistakenly breaking it.

ludwiczakpawel:
Whoa, nice investigation! :D.

joshpuetz (approved):

juliannatetreault (approved):
👍 👍

PR #10851, opened 2020-10-14
Optimization:Break up SiteConfig Updates by Groups

mstruve:
- [x] Optimization

## Description
Every time when we would save a SiteConfig we would update every single config whether or not it actually got changed. This led to 20+ second response times and thousands of unnecessary hits to our databases which you can see in the Honeycomb trace screenshot below. To fix this I split each section into its own form so we only update those values in each section instead of every single one. The view could likely be refactored further and we could get more technical about what is actually changing during updates but this seemed like a quick easy way to get better performance from this endpoint. 
Screen Shot 2020-10-14 at 1 17 28 PM

I am also a fan of only updating sections bc I think it ensures that something doesn't get edited and forgotten about and then accidentally changed when other changes are meant to be saved. 

https://github.com/orgs/forem/projects/6#card-47372705

Test that making config changes in admin still works. 

## Added tests?

alt_text

### DIFF
--- a/app/views/admin/configs/_form_submission.html.erb
+++ b/app/views/admin/configs/_form_submission.html.erb
@@ -1,0 +1,8 @@
+<% if current_user.has_role?(:single_resource_admin, Config) %>
+  <div class="form-group">
+    <%= label_tag :confirmation_update %>
+    <%= text_field_tag :confirmation, nil, class: "form-control", placeholder: "Confirmation text", autocomplete: "off", required: true, pattern: @confirmation_text %>
+    <div class="alert alert-info">Type the sentence: <%= @confirmation_text %></div>
+    <%= f.submit "Update Site Configuration", class: "btn btn-danger btn-lg" %>
+  </div>
+<% end %>
--- a/app/views/admin/configs/show.html.erb
+++ b/app/views/admin/configs/show.html.erb
@@ -112,9 +112,12 @@
                                  placeholder: Constants::SiteConfig::DETAILS[:video_encoder_key][:placeholder] %>
                 <div class="alert alert-info"><%= Constants::SiteConfig::DETAILS[:video_encoder_key][:description] %></div>
               </div>
+              <%= render "form_submission", f: f %>
             </div>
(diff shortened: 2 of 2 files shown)


benhalpern (approved):
This works for me. Heads up on this change @pkfrank @jessleenyc 

I don't think this will be a _major_ UX change, but one of you should sign off on this because it will affect the experience for Forem creators right away.

benhalpern:
This will make each individual expansion area its own form.....

Which will probably be a good step towards possibly not _always_ require the same "are you sure" context and copy. Because if everything is dangerous, nothing is dangerous.

So I'd say this moves us in a good direction and we can ship it.

From a UX perspective I'd say this would be a good time to think about changing this to AJAX so one could submit one portion and then move on without a whole page refresh. But out of scope here.

Screen Shot 2020-10-14 at 4 11 14 PM

maestromac (approved):

vaidehijoshi (changes requested):
Hm, it looks like switching this into individual forms has borked the styles for the "required" bubble.

On `main`:
Screen Shot 2020-10-14 at 1 20 44 PM

On this branch:
Screen Shot 2020-10-14 at 1 21 10 PM

pkfrank (approved):
Looks good from a Forem Admin experience.  I don't think anyone will be thrown off too much.  Agreed this moves us into a better direction to update the "confirmation" text appropriately for more critical vs. less critical updates.

mstruve:
@vaidehijoshi  I am not sure why its like that but that is how it looks on production as well 
Screen Shot 2020-10-14 at 4 26 46 PM

vaidehijoshi (approved):

vaidehijoshi:
> @vaidehijoshi I am not sure why its like that but that is how it looks on production as well so I dont think this PR is the cause


@mstruve Hm, it looks like I wasn't on the _latest_ version of master and this was borked by some other (recent) commit. Am approving your PR and will open a separate issue for this UI regression 👍

PR #10865, opened 2020-10-15
Bust caches when navigation links are changed

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently if an admin changes the contents of their navigation links, the UI won't immediately reflect the change— causing confusion and difficulties for admins.

This moves and renames `bust_relevant_caches` to be usable across different admin paths— and fixes the issue where if an admin changes navigation links, they won't see the change until this fragment cache expires.

Closes https://github.com/forem/InternalProjectPlanning/issues/150


pkfrank (approved):
Awesome 👍  this will really streamline things for Forem Admins as they adjust the config.

app/controllers/admin/application_controller.rb
```diff
@@ -47,5 +47,14 @@ def authorization_resource
...
       authorize(authorization_resource, :access?, policy_class: InternalPolicy)
     end
+
+    def bust_content_change_caches
+      CacheBuster.bust("/tags/onboarding") # Needs to change when suggested_tags is edited.
+      CacheBuster.bust("/shell_top") # Cached at edge, sent to service worker.
+      CacheBuster.bust("/shell_bottom") # Cached at edge, sent to service worker.
+      CacheBuster.bust("/onboarding") # Page is cached at edge.
+      CacheBuster.bust("/") # Page is cached at edge.
+      Rails.cache.delete_matched("*-#{ApplicationConfig['RELEASE_FOOTPRINT']}") # Delete all caches tied to this key.
```
mstruve, line 57: Before we start using this method all over the place we need to figure out a different approach for deleting these `RELEASE_FOOTPRINT` keys. 

`delete_matched` uses Redis `KEYS` [which as noted in the docs should be avoided in production databases.](https://www.rubydoc.info/gems/redis-activesupport/4.1.1/ActiveSupport%2FCache%2FRedisStore:delete_matched)

> Performance note: this operation can be dangerous for large production databases, as it uses the Redis “KEYS” command, which is O(N) over the total number of keys in the database. Users of large Redis caches should avoid this method.

Once it fetches all the matching keys it then does a Redis scan going through ALL of the matching Redis keys 1000 at a time which turns out are a lot of keys AND I [just realized is actually the sole reason our SiteConfig updates take 20 seconds to complete.](https://ui.honeycomb.io/thepracticaldev/datasets/rails/result/9rATP15JT5U/trace/jYUWbVyZeNk) I assumed it was updating all the fields until today when I scrolled down far enough to realize it was due to this Redis request.

mstruve (changes requested):
We need to figure out something else besides `Rails.cache.delete_matched` for removing the release footprint keys. See my comment for more color.

app/controllers/admin/application_controller.rb
mstruve: It looks like the release footprint is not used in many places for our Redis keys. I found 3 when I searched. Given this, I wonder if we could prepend some value on to these that we could update when this bust cache happens to cause new keys to be set. 
```
/Users/mollystruve/forem/app/views/layouts/_styles.html.erb:
    1: <% cache "base_inline_styles_#{@story_show}_#{@article_index}_#{@home_page}_#{@article_show}_#{view_class}_#{@notifications_index}_#{@tags_index}_#{@reading_list_items_index}_#{@history_index}_#{ApplicationConfig['RELEASE_FOOTPRINT']}_#{user_signed_in?}_#{@shell}", expires_in: 8.hours do %>
    2    <% if @shell %>
    3      <style>

/Users/mollystruve/forem/app/views/liquid_embeds/show.html.erb:
    1  <base target="_parent">
    2: <% cache "liquid_tag_styles_#{ApplicationConfig['RELEASE_FOOTPRINT']}", expires_in: 8.hours do #TODO: Render specific ltag class instead of everything %>
    3    <style><%= Rails.application.assets["ltags/LiquidTags.css"].to_s.html_safe %></style>
    4  <% end %>

/Users/mollystruve/forem/app/views/shell/_bottom.html.erb:
    1  <!-- Start Bottom Shell -->
    2  <% unless internal_navigation? %>
    3:   <% cache("footer-and-signup-modal--#{user_signed_in?}--#{ApplicationConfig['RELEASE_FOOTPRINT']}", expires_in: 24.hours) do %>
    4      <%= render "layouts/footer" %>
    5      <%= render "layouts/signup_modal" unless user_signed_in? %>
```

benhalpern: If we included something like `SiteConfig.updated_at` which we could change when we update we could pass that in as a component of these keys, which would mean one extra Redis hit.... But we could assign a variable so that it's only one extra Redis look up _per request_.... Unless the Rails Settings Cached gem already does this (For some reason I thought this might be the functionality but I don't see where that might be written)

So anyway, this is a much better solution than `delete_cached` for this reason and others I think (just more overall flexibility with the caching if we introduce a pattern like this)....

Does that make sense?

mstruve: Works for me! 

We hit SiteConfig on a lot of pages and once you hit it once we have it cached locally so for many of these keys it won't even be an extra Redis hit.

benhalpern: Okay, added `admin_action_taken_at` to site config which will be set to current time when these admin actions are taken, busting those caches.

mstruve (approved):
🙌

citizen428 (approved):

mstruve (approved):
Got you ✅  for whenever you are ready!

PR #10866, opened 2020-10-15
Optimization/Flaky Fix:Combine Spec Assertions and Use Timecop for Datetime Assertion

mstruve:
- [x] Spec Optimization

## Description
Let me know what you think of using methods and assertions this way with aggregate failures. I noticed we are not doing anything in the individual assertions beyond just looking for different content on the page. To avoid reloading the entire page each time, I combined the assertions into a single one. This took the spec run time from around 6.5 seconds to around 3.5 seconds(excluding about 8 second load time for files). Not a giant win, but if people like it, I think this is a pattern we could repeat for other specs to speed them up. 

While running this spec I just found a flaky component to it where the date asserted and the date shown down match due to straddling a date. Adding Timecop fixes this issue. 
Fixes: 
```
  1) User index when user is unauthorized when 1 article shows all proper elements
     Failure/Error: expect(page).to have_selector(".comment-date", text: comment_date)
       expected to find visible css ".comment-date" with text "Oct 16" within #<Capybara::Node::Element tag="div" path="/HTML/BODY[1]/DIV[7]/DIV[2]/DIV[2]/DIV[2]/DIV[2]/DIV[3]/A[1]/DIV[1]"> but there were no matches. Also found "Oct 15", which matched the selector but not all filters. 

     # ./spec/system/user/view_user_index_spec.rb:53:in `block in shows_comments'
     # ./spec/system/user/view_user_index_spec.rb:50:in `shows_comments'
```

alt_text

### DIFF
--- a/spec/system/user/view_user_index_spec.rb
+++ b/spec/system/user/view_user_index_spec.rb
@@ -86,7 +86,7 @@ RSpec.describe "User index", type: :system, stub_elasticsearch: true do
       visit "/#{user.username}"
     end
 
-    it "shows all proper eleiments", :aggregate_failures, js: true do
+    it "shows all proper elements", :aggregate_failures, js: true do
       shows_header
       shows_articles
       shows_comments
--- a/spec/system/user/view_user_index_spec.rb
+++ b/spec/system/user/view_user_index_spec.rb
@@ -11,38 +11,42 @@ RSpec.describe "User index", type: :system, stub_elasticsearch: true do
     context "when 1 article" do
       before { visit "/#{user.username}" }
 
-      it "shows the header", js: true do
+      it "shows all proper elements", :aggregate_failures, js: true do
+        shows_header
+        shows_title
+        shows_articles
+        shows_comments
+        shows_comment_timestamp
+      end
+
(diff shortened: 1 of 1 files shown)


spec/system/user/view_user_index_spec.rb
```diff
@@ -78,21 +86,27 @@
       visit "/#{user.username}"
     end
 
-    it "shows the header", js: true do
+    it "shows all proper eleiments", :aggregate_failures, js: true do
```
Zhao-Andy, line 89: Small typo :)

```suggestion
    it "shows all proper elements", :aggregate_failures, js: true do
```

Zhao-Andy (approved):
Nice! I think it makes sense. There might be a more "RSpec" way of doing this, but I'm not 100% sure and can't remember any particular way of handling this.

spec/system/user/view_user_index_spec.rb
mstruve: lol damnit

maestromac (approved):
I like this way better. LGTM!

PR #10867, opened 2020-10-15
Store Article's main_image path unprocessed

maestromac:
- [x] Bug Fix
- [x] This kind feels like a refactor too

## Description
We are unnecessarily altering the image path on upload. Please see below for the explanation.

This also serves as a bugfix for forem-cloud app.
https://github.com/orgs/forem/projects/11#card-47453148
n/a

## Added tests?
- [x] yes

## Added to documentation?
- [x] no documentation needed

n/a

### DIFF
--- a/app/controllers/image_uploads_controller.rb
+++ b/app/controllers/image_uploads_controller.rb
@@ -38,11 +38,7 @@ class ImageUploadsController < ApplicationController
   end
 
   def cloudinary_link(uploaders)
-    links = if params[:wrap_cloudinary]
-              [ApplicationController.helpers.cloud_cover_url(uploaders[0].url)]
-            else
-              uploaders.map(&:url)
-            end
+    links = uploaders.map(&:url)
     respond_to do |format|
       format.json { render json: { links: links }, status: :ok }
     end
--- a/app/controllers/image_uploads_controller.rb
+++ b/app/controllers/image_uploads_controller.rb
@@ -34,10 +34,6 @@ class ImageUploadsController < ApplicationController
       return
     end
 
-    cloudinary_link(uploaders)
-  end
-
-  def cloudinary_link(uploaders)
     links = uploaders.map(&:url)
     respond_to do |format|
       format.json { render json: { links: links }, status: :ok }
--- a/lib/data_update_scripts/20201015190914_update_article_main_image_path.rb
+++ b/lib/data_update_scripts/20201015190914_update_article_main_image_path.rb
@@ -1,0 +1,15 @@
+module DataUpdateScripts
+  class UpdateArticleMainImagePath
+    def run
+      return unless ENV["FOREM_CONTEXT"] == "forem_cloud"
+
+      Article.find_each do |article|
+        next unless article.main_image&.starts_with? "https://res.cloudinary.com/"
+
+        index = article.main_image.index(URL.url)
+        article.main_image = article.main_image[index..].gsub("/images/", "/remoteimages/")
+        article.save
+      end
+    end
+  end
+end
--- a/lib/data_update_scripts/20201015190914_update_article_main_image_path.rb
+++ b/lib/data_update_scripts/20201015190914_update_article_main_image_path.rb
@@ -3,12 +3,12 @@ module DataUpdateScripts
     def run
       return unless ENV["FOREM_CONTEXT"] == "forem_cloud"
 
-      Article.find_each do |article|
+      Article.where.not(main_image: [nil, ""]).each do |article|
         next unless article.main_image&.starts_with? "https://res.cloudinary.com/"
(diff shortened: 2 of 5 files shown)


app/controllers/image_uploads_controller.rb
```diff
@@ -34,15 +34,7 @@ def create
...
 
-    cloudinary_link(uploaders)
-  end
-
-  def cloudinary_link(uploaders)
-    links = if params[:wrap_cloudinary]
-              [ApplicationController.helpers.cloud_cover_url(uploaders[0].url)]
-            else
-              uploaders.map(&:url)
-            end
```
maestromac, line 45: There's no need to pre-wrap these uploaded image URLs with Cloudinary/imgproxy. Instead, we should store it as a raw image URL which provides greater flexibility on how we want to optimize it later.  Plus, I searched through the codebase, and everywhere I looked, we are already trying to optimize it again, ie `cloud_cover_url(article.main_image)`.

lib/data_update_scripts/20201015190914_update_article_main_image_path.rb
```diff
@@ -0,0 +1,15 @@
+module DataUpdateScripts
+  class UpdateArticleMainImagePath
```
maestromac, line 2: This only updates `article.main_image` on Forem Cloud's apps. This updates the proxy path and removes the Cloudinary prefix.

jdoss (approved):

lib/data_update_scripts/20201015190914_update_article_main_image_path.rb
```diff
@@ -0,0 +1,15 @@
...
+    def run
+      return unless ENV["FOREM_CONTEXT"] == "forem_cloud"
+
+      Article.find_each do |article|
+        next unless article.main_image&.starts_with? "https://res.cloudinary.com/"
+
+        index = article.main_image.index(URL.url)
+        article.main_image = article.main_image[index..].gsub("/images/", "/remoteimages/")
+        article.save
```
benhalpern, line 11: Rather than `article.save` could we do this as `.update_column` so we don't run so many callbacks?

Possible also only loop through images `.where.not(main_image: [nil,""])` to avoid looping through so many?

benhalpern (changes requested):

lib/data_update_scripts/20201015190914_update_article_main_image_path.rb
maestromac: Yup let me update this. I didn't try to optimize it too much since this is for forem-cloud only.

lib/data_update_scripts/20201015190914_update_article_main_image_path.rb
```diff
@@ -0,0 +1,15 @@
...
+  class UpdateArticleMainImagePath
+    def run
+      return unless ENV["FOREM_CONTEXT"] == "forem_cloud"
+
+      Article.where.not(main_image: [nil, ""]).each do |article|
+        next unless article.main_image&.starts_with? "https://res.cloudinary.com/"
+
+        index = article.main_image.index(URL.url)
+        raw_image_path = article.main_image[index..].gsub("/images/", "/remoteimages/")
+        article.update_column(:main_image, raw_image_path)
```
mstruve, line 11: This column is stored in Elasticsearch so we need to reindex the article after we update this field.

mstruve (commented):
Were you able to test this locally at all with Cloudinary creds to ensure everything works still?

lib/data_update_scripts/20201015190914_update_article_main_image_path.rb
maestromac: 🤦 oh right. updated!

mstruve: Fwiw, you could still do the update column but you have to make sure to manually index the article when you are done by doing `article.index_to_elasticsearch` but going back to save works as well.

mstruve (approved):
I'm good with using `.save`

lib/data_update_scripts/20201015190914_update_article_main_image_path.rb
benhalpern: Ah....... Yeah I wasn't thinking about that. Yeah if this is Forem Cloud only then we don't need that optimization.

lib/data_update_scripts/20201015190914_update_article_main_image_path.rb
benhalpern: Yeah let's go with `save`.... I wasn't thinking about this being forem cloud only.... This works.

benhalpern (approved):

PR #10621, opened 2020-10-05
Improve small screen display of Mod Center

lisasy:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Improve small screen display of Mod Center by:
- Updating the center header 
- Using mobile-friendly navigation for the menu
- Hiding `Resources` link on screen screen
- Updating rows for small screen by collapsing content better
- Splitting huge `moderators.scss` into separate files for Mod Center and Mod Actions

## Related Tickets & Documents
Closes https://github.com/forem/InternalProjectPlanning/issues/55

#### Current:
Screen Shot 2020-10-05 at 2 36 17 PM

Screen Shot 2020-10-05 at 2 36 07 PM

#### Proposed in PR:
Screen Shot 2020-10-05 at 2 27 41 PM

Screen Shot 2020-10-05 at 2 35 49 PM

## Added tests?

- [ ] yes
- [x] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

PR #10868, opened 2020-10-15
Fix tag display bug and crayonsify area

benhalpern:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This pull request fixes a bug where inter-page navigation messes with the display of tags.

It also updates tags to crayons design and adds a quick link to adjusting tags. I made small design and copy adjustments to that page as well.

In my opinion navigating to tags from this area is still a valuable feature but is inhibited by poor design and lack of awareness around the customization.... And the bug that was in place.

(It was my old design so I think I'm allowed to be blunt about its poor quality 😄)

Beyond just the bug fixes I think this is an improvement we should lean in on.

A follow-up feature which I think could add more usability here could be an `indicator` mentioning the number of posts for that tag in the last 24 hours or since you last visited the page. Could encourage simpler and more intuitive navigation to that area.

Screen Shot 2020-10-15 at 1 48 02 PM
Screen Shot 2020-10-15 at 1 47 57 PM

### DIFF
--- a/app/views/articles/index.html.erb
+++ b/app/views/articles/index.html.erb
@@ -77,7 +77,7 @@
       </header>
 
       <% if user_signed_in? %>
-        <div id="homepage-feed"></div>
+        <div id="homepage-feed" style="min-height: 90vh"></div>
       <% else %>
         <%= render "stories/main_stories_feed" %>
       <% end %>
--- a/app/views/layouts/_user_config.html.erb
+++ b/app/views/layouts/_user_config.html.erb
@@ -1,6 +1,7 @@
 <script>
   try {
     const bodyClass = localStorage.getItem('config_body_class');
+    const userString = localStorage.getItem('current_user');
 
     if (bodyClass) {
       document.body.className = bodyClass;
@@ -29,6 +30,17 @@
     }
     if (window.frameElement) { // Hide top bar and footer when loaded within iframe
       document.body.classList.add("hidden-shell");
+    }
+    if (userString && userString.length > 0) {
+      const user = JSON.parse(userString)
+      const numTags = JSON.parse(user.followed_tags).length
+      if (numTags < 6) {
+        document.body.classList.add("user-tags-followed-"+numTags);
+      } else {
+        document.body.classList.add("user-tags-followed-max");
+      }
+      sidebarTags.style.height = "100vh"
+      sidebarTags.style.backgroundColor = "green"
     }
   } catch(e) {
       Honeybadger.notify(e);
--- a/app/assets/javascripts/initializers/initializeBaseUserData.js
+++ b/app/assets/javascripts/initializers/initializeBaseUserData.js
@@ -22,25 +22,6 @@ function initializeProfileImage(user) {
 function initializeUserSidebar(user) {
   if (!document.getElementById('sidebar-nav')) return;
   initializeUserProfileContent(user);
-
-  let followedTags = JSON.parse(user.followed_tags);
-  const tagSeparatorLabel =
-    followedTags.length === 0
-      ? 'FOLLOW TAGS TO IMPROVE YOUR FEED'
-      : 'OTHER POPULAR TAGS';
-
-  followedTags.forEach((tag) => {
-    const element = document.getElementById(
-      'default-sidebar-element-' + tag.name,
-    );
-
-    if (element) {
-      element.remove();
-    }
-  });
-
-  document.getElementById('tag-separator').innerHTML = tagSeparatorLabel;
-  document.getElementById('sidebar-nav-default-tags').classList.add('showing');
 }
 
 function addRelevantButtonsToArticle(user) {
--- a/app/assets/stylesheets/articles.scss
+++ b/app/assets/stylesheets/articles.scss
@@ -84,17 +84,6 @@
       display: block;
     }
 
-    .sidebar-nav {
-      .sidebar-nav-block {
-        .sidebar-nav-default-tags {
-          display: none;
-          &.showing {
-            display: block;
-          }
-        }
-      }
-    }
-
     .search-partner-mention {
       display: none;
       text-align: center;
--- a/app/assets/stylesheets/base/main.scss
+++ b/app/assets/stylesheets/base/main.scss
@@ -20,6 +20,22 @@ body {
     }
   }
 
+  &.user-tags-followed-0, &.user-tags-followed-1, &.user-tags-followed-2{
+    #sidebar-nav-followed-tags {
+      height: 90px;
+    }
+  }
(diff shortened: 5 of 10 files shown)


nickytonline (changes requested):
It looks fine in the logged out view, but in the logged on view, it's broken.

image

These two lines in the function `initializeUserSidebar` in intitializeBaseUserData.js appear to be part of the problem

```javascript
  document.getElementById('tag-separator').innerHTML = tagSeparatorLabel;
  document.getElementById('sidebar-nav-default-tags').classList.add('showing');
```

benhalpern:
@nickytonline whoops, I think I fixed that.

app/views/articles/index.html.erb
```diff
@@ -77,7 +77,7 @@
       </header>
 
       <% if user_signed_in? %>
-        <div id="homepage-feed"></div>
+        <div id="homepage-feed" style="min-height: 90vh"></div>
```
nickytonline, line 80: refactor (non-blocking): Consider moving this inline style to a CSS class.

nickytonline (approved):
LGTM 🚀

lisasy:
Screen Shot 2020-10-15 at 7 34 53 PM

I think any effort to deprecate older styles to our Crayons is a good one, but in this use case, it is more than just updating to Crayons: it's also shifting the information architecture when someone glances at their Home screen. I think that these updated styles draw attention to Tag section in a distracting and not very successful way.

> In my opinion navigating to tags from this area is still a valuable feature but is inhibited by poor design and lack of awareness around the customization.... And the bug that was in place.

I agree with you @benhalpern that our tag feature is valuable and that we don't do enough to encourage its utility because of gaps in our design. But I'm not sure that the solution here would be merely to update it to Crayons styling. I think this is one opportunity for us to align on the crisp problem and ideate on potential solutions to solve it. I think solving this problem is wider than the scope of merely updating the visual design in the left sidebar, and it is something that would deserve some dedicated product-thinking and exploration work.

ludwiczakpawel:
**+1 to what @lisasy said.** 

---

In addition to what Lisa also mentioned around information architecture, I have some feedback about visuals purely. The idea for the left sidebar was to be clean and readable, ideally just navigation without any boxes etc (I even wish we could drop the sponsors section as well as ads in here). Remember [that thing](https://www.figma.com/proto/RUnXeeDRux7AGP2i5Fz2DT/Home-v2?node-id=1%3A2&viewport=342%2C627%2C0.7002484798431396&scaling=min-zoom) I've shared while ago --> sticky left sidebar, only main nav, "customized" widgets (or in other words "tags you care about") in right sidebar, "footer" links at the bottom of the right sidebar...

I do agree that being able to jump to fav tags is quite useful, I have suggested some ideas to do so in the past [here](https://forem.dev/pp/declutter-the-sidebar-3fkk) (if you're in the mood of working on sidebar AND tags, this could be a good candidate for someone's (yours) backend skills :D).

Anyway, the version proposed here is quite heavy visually and steals a lot of attention, unnecessarily taking huge amount of space - I'd reconsider some design choices :)

benhalpern:
Cool, this all makes sense @lisasy @ludwiczakpawel...... I'll probably change this PR to include _the fix_ for now, but remove the distinct Crayons box so that we maintain the visual hierarchy (but fix the bug)...

benhalpern:
..... And then maybe I can take a look at some options for how we might accomplish @ludwiczakpawel's vision for the right sidebar. I definitely think that seems best.

ludwiczakpawel:
> And then maybe I can take a look at some options for how we might accomplish @ludwiczakpawel's vision for the right sidebar. I definitely think that seems best.

I really can't wait for this to happen so I'm happy to jam on it if/whenever you want :)

lisasy:
@ludwiczakpawel @benhalpern this would be a fun thing to do together in our 1on1on1on1 with @vaidehijoshi !

benhalpern:
Okay @ludwiczakpawel @lisasy I adjusted this to remove the weight of the crayons card, adjust the size of the header and to make the list of tag links closer similar to where it was— Along with the JS bus fixes this addresses.

So I think it's still an aesthetic improvement for now with minimal changes to the feel of the overall page until we close the loop on this whole area, eh?

Screen Shot 2020-10-16 at 2 44 54 PM

lisasy:
This LGTM.

@ludwiczakpawel This is an example of where having the `crayons-subtitle-3` use `fs-l` is a bit too large, and why `fs-base` could be more ideal. That's hopefully going to be resolved in this branch here: https://github.com/forem/forem/pull/10621

ludwiczakpawel:
@lisasy makes sense but I also think maybe we could have `crayons-subtitle-**4**` for this scenario? :) The sidebar on home page is narrower than the one in notifications so it naturally looks less balanced and more heavy. But I don't have strong opinion.

lisasy:
@ludwiczakpawel `crayons-subtitle-4` LGTM!

ludwiczakpawel (approved):

app/views/layouts/_user_config.html.erb
```diff
@@ -30,6 +31,17 @@
...
+    if (userString && userString.length > 0) {
+      const user = JSON.parse(userString)
+      const numTags = JSON.parse(user.followed_tags).length
+      if (numTags < 6) {
+        document.body.classList.add("user-tags-followed-"+numTags);
+      } else {
+        document.body.classList.add("user-tags-followed-max");
+      }
+      sidebarTags.style.height = "100vh"
+      sidebarTags.style.backgroundColor = "green"
```
ludwiczakpawel, line 43: I know it's merged already but what is this part about? @benhalpern

app/views/layouts/_user_config.html.erb
```diff
@@ -30,6 +31,17 @@
...
       document.body.classList.add("hidden-shell");
     }
+    if (userString && userString.length > 0) {
+      const user = JSON.parse(userString)
+      const numTags = JSON.parse(user.followed_tags).length
+      if (numTags < 6) {
+        document.body.classList.add("user-tags-followed-"+numTags);
+      } else {
+        document.body.classList.add("user-tags-followed-max");
+      }
```
ludwiczakpawel, line 41: @benhalpern again, I'm late to the party, but maybe instead of applying specific classes here (especially to `body` element) causing [a bit of mess in main.scss](https://github.com/forem/forem/pull/10868/files#diff-c5a53f74b905bb368fd774f676882b417a22c8d599a5c8097eee51ef8e5d3adfR23-R38), maybe we could just go with inline style here?

```
if (numTags < 6) {
  document.querySelector('#sidebar-nav-followed-tags ').style.height = "205px";
  ...
}
```

Even though it's inline style I think it's gonna be much cleaner this way.

PR #10870, opened 2020-10-15
Adds a new admin setting for experience levels.

### LINKED ISSUE #10845: Experience Level of Post range should be configurable

On a new cloud Forem my community is not related to software development. There are experience levels in my community though so when an author is managing a post it makes sense to adjust the experience level of post. Right now the range goes from 'Total Newbies' to 'Senior Devs'. 

The experience level range should be admin configurable. This config should probably live in the Community Content config section, although User Experience and Brand might make sense as well.

I don't think there are any possible alternatives

**Additional context**

Screen Shot 2020-10-14 at 10 48 17 AM

AfroDevGirl:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
These admin settings are used in the post management UI to determine the range of experience. These values fallback to the ones we currently see in DEV, "Total Newbies" and "Senior Devs". There is also a removal of the word coding when talking about experience level, since the context for each forem instance should be implied when referencing experience.

Closes #10845 

- Verify that the fallback values of 'Total Newbies' and 'Senior Devs' are present before testing the new functionality
- As an admin with a config update role navigate to /admin/config
- Toggle the Community Content section and observe there are two new fields at the bottom, Experience Low/High
- Update those values with any text and save the config
- Then, create a post and publish it
- In your personal dashboard, click manage to view a post's experience level and observe that the saved config values are there. There should be no changes in setting the experience level of a post.
- Navigate to /settings/ux and observe that under Content Customization the text now reads `What is your approximate experience level (1-5)?`

UX Settings:

Screen Shot 2020-10-15 at 3 26 01 PM

Admin Config:

Screen Shot 2020-10-15 at 3 26 32 PM

Post Management:

Screen Shot 2020-10-15 at 3 27 00 PM

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

_Edit: We need to add the values we want to display in DEV's admin dashboard to keep "Senior Devs" instead of the newly included defaults._

kim kardashian west saying "I wish everyone could experience this"

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -6,7 +6,6 @@ class OrganizationsController < ApplicationController
 
     @tab = "organization"
     @user = current_user
-    @tab_list = @user.settings_tab_list
 
     unless valid_image?
       render template: "users/edit"
@@ -29,7 +28,6 @@ class OrganizationsController < ApplicationController
   def update
     @user = current_user
     @tab = "organization"
-    @tab_list = @user.settings_tab_list
     set_organization
 
     unless valid_image?
@@ -108,7 +106,7 @@ class OrganizationsController < ApplicationController
   def organization_params
     params.require(:organization).permit(permitted_params)
       .transform_values do |value|
-        if value.class.name == "String"
+        if value.instance_of?("String")
           ActionController::Base.helpers.strip_tags(value)
         else
           value
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -106,7 +106,7 @@ class OrganizationsController < ApplicationController
   def organization_params
     params.require(:organization).permit(permitted_params)
       .transform_values do |value|
-        if value.instance_of?("String")
+        if value.instance_of?(String)
           ActionController::Base.helpers.strip_tags(value)
         else
           value
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -23,7 +23,7 @@ class SiteConfig < RailsSettings::Base
 
   # Authentication
   field :allow_email_password_registration, type: :boolean, default: false
-  field :allow_email_password_login, type: :boolean, default: false
+  field :allow_email_password_login, type: :boolean, default: true
   field :authentication_providers, type: :array, default: proc { Authentication::Providers.available }
   field :invite_only_mode, type: :boolean, default: false
   field :twitter_key, type: :string, default: ApplicationConfig["TWITTER_KEY"]
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -50,8 +50,8 @@ class SiteConfig < RailsSettings::Base
                                          default: ApplicationConfig["COMMUNITY_COPYRIGHT_START_YEAR"] ||
                                            Time.zone.today.year
   field :staff_user_id, type: :integer, default: 1
-  field :experience_low, type: :string, default: ApplicationConfig["EXPERIENCE_LOW"] || "Total Newbies"
-  field :experience_high, type: :string, default: ApplicationConfig["EXPERIENCE_HIGH"] || "Senior Devs"
+  field :experience_low, type: :string, default: "Total Newbies"
+  field :experience_high, type: :string, default: "Experienced Users"
 
   # Emails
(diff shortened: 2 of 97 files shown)


app/views/articles/manage.html.erb
```diff
@@ -51,23 +51,23 @@
       <%= f.hidden_field :article_id, value: @article.id %>
       <%= f.hidden_field :group, value: "experience_level" %>
       <span class="button-section button-section-newbie">
-        <button value="1" name="rating_vote[rating]" class="level-rating-button <%= "selected" if @rating_vote&.rating == 1.0 %>">1</button>
-        <button value="2" name="rating_vote[rating]" class="level-rating-button <%= "selected" if @rating_vote&.rating == 2.0 %>">2</button>
-        <button value="3" name="rating_vote[rating]" class="level-rating-button <%= "selected" if @rating_vote&.rating == 3.0 %>">3</button>
+        <button value="1" name="rating_vote[rating]" class="level-rating-button <%= "selected" if @rating_vote&.rating.to_d == 1.0.to_d %>">1</button>
```
AfroDevGirl, line 54: I was getting a rubocop issue on these. I'm happy to revert this particular change if y'all would prefer that.

app/models/site_config.rb
```diff
@@ -50,6 +50,8 @@ class SiteConfig < RailsSettings::Base
                                          default: ApplicationConfig["COMMUNITY_COPYRIGHT_START_YEAR"] ||
                                            Time.zone.today.year
   field :staff_user_id, type: :integer, default: 1
+  field :experience_low, type: :string, default: ApplicationConfig["EXPERIENCE_LOW"] || "Total Newbies"
+  field :experience_high, type: :string, default: ApplicationConfig["EXPERIENCE_HIGH"] || "Senior Devs"
```
fdoxyz, line 54: I think we don't need to add `ApplicationConfig["EXPERIENCE_LOW"]` AND `ApplicationConfig["EXPERIENCE_HIGH"]` here. I understand how it makes sense to add them at first glance since other entries have them, but those are in place because we were using ENV variables before and this avoided breaking some existing forems. Now we generally want to try to avoid adding new ENV variables in favor of purely `SiteConfig`.

Also, since this is a generalization effort I think most communities would benefit from a generalized default as well. So my suggestion would be leaving the defaults as "Total Newbies" and "Experienced Users"/"Advanced Users". There may be other better defaults so I'm open to hearing other's ideas around options for these defaults but anything along those lines should probably be fine.

fdoxyz (commented):
Thanks for taking a stab at this @AfroDevGirl!

I'm leaving a small implementation comment inline, which in general looks great to me. From the product context though @benhalpern may have some context or any other notes (if any)

app/models/site_config.rb
AfroDevGirl: Makes sense, I'll remove the ENV var defaults for now and wait for direction on what the `experience_high` default should be.

benhalpern (approved):
This looks good to me. Definitely helps this feature immediately become more relevant for different Forems.

I have a thought about generalizing this in the future moreso but I think this is a good stepping stone.

fdoxyz (changes requested):
Awesome, thanks for including the requested changes @AfroDevGirl.

The PR ran into some conflicts probably due to some recent edits to that file by other commits. Would you mind fixing those in order to merge this? Thanks!

app/controllers/organizations_controller.rb
```diff
@@ -106,7 +106,7 @@ def permitted_params
   def organization_params
     params.require(:organization).permit(permitted_params)
       .transform_values do |value|
-        if value.class.name == "String"
+        if value.instance_of?("String")
```
fdoxyz, line 109: Was this added because of Rubocop or maybe automatically during the commit (via Husky or similar)?

I think this is causing the CI to fail. If we're going to use `.instance_of?` it should be:

```suggestion
        if value.instance_of?(String)
```

AfroDevGirl: Yea I definitely didn't manually add this. I'll adjust.

fdoxyz (approved):
👍🏼

PR #10875, opened 2020-10-16
User color fields in admin

### LINKED ISSUE #10872: Add color picker in more admin areas
As a Forem Admin, I may be required to add color hexes in various configuration areas.

Ideally the same color picker that we use on `/settings` could be added in more areas.  For instance:

- [ ] `/admin/tags/{tag id}`
- [ ] `/admin/config` (in User Experience & Brand section)
- Likely other areas I'm overlooking right now

**Additional context**
We seem to have two different color pickers.  One visible on `/settings` (which I prefer) and a different one on `/t/{tag name}/edit` which is available to moderators of that tag.  It might be nice to standardize?

I also believe we do contrast / accessibility checks in the User Experience & Brand section, but I'm not sure if we do it everywhere.  But that's probably out of scope.

citizen428:

- [X] Refactor

## Description

This PR makes color pickers in the admin actual color fields instead of text inputs, similar to how we do it on the settings page.

Closes #10872 

Screen Shot 2020-10-16 at 10 49 40
Screen Shot 2020-10-16 at 10 50 17

This could certainly be prettier, but it's internal, so probably good enough for now. I noticed that there's currently some styling work going on in the admin area anyway, so someone more qualified than me can add the ✨

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/app/views/admin/configs/show.html.erb
+++ b/app/views/admin/configs/show.html.erb
@@ -1192,11 +1192,11 @@
               </div>
               <div class="form-group">
                 <%= f.label :primary_brand_color_hex %>
-                <%= f.text_field :primary_brand_color_hex,
-                                 class: "form-control",
-                                 value: SiteConfig.primary_brand_color_hex,
-                                 pattern: "^#+([a-fA-F0-9]{6})$",
-                                 placeholder: Constants::SiteConfig::DETAILS[:primary_brand_color_hex][:placeholder] %>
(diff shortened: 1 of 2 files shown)


ludwiczakpawel (approved):

benhalpern (approved):
Great

PR #10886, opened 2020-10-17
Fix weird highlighting bug

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Noticed this bug, where multiple links can be highlighted in the admin. This change fixes it and seems safe since we can make some assumptions about the request path.

image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed


benhalpern (approved):

fdoxyz:
Whoops missed to type in this comment in my review 😅

Tested this locally and works great.

__Non-blocking:__
Would it make sense to add a test for this? I'm not sure if we are planning on iterating on this menu layout, but as more menus can be added/updated in the near future it could be helpful to have a check to avoid regressions or odd cases.

I'm thinking something like rendering the existing dashboards and checking there is only one `crayons-link--current`

benhalpern:
Yeah, follow up test could make sense. Doesn't need to be a blocker due to all the request specs in place pretty much assuring no runtime errors.

PR #10888, opened 2020-10-17
Add loading=lazy to image tags

fredden:
- [x] Refactor
- [x] Feature
- [ ] ~Bug Fix~
- [x] Optimisation
- [ ] ~Documentation Update~

## Description
This pull request will add `loading="lazy"` to relevant image tags. This means that the user agent can choose to delay downloading an image if it determines that will produce a better user experience.

Resources:
- https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading

See #2776 where this work was started.

The website should behave mostly as it currently does, with smaller "footprint" on load, and likely better scores on performance analysis tools.

## Added tests?
- [ ] yes
- [ ] no, because I need help

## Added to documentation?
- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed


benhalpern:
Hi, sorry about not getting to this sooner. This is a great PR, I think our team was taking its time to look through all the files to make sure there are no small bugs, and in the process there were some conflicts created

Can yo fix the conflicts, and then we can circle back and get this reviewed in a timely manner?

rhymes:
It might be better to split this in smaller PR as going through 132 files is a bit daunting :)

There's also a bigger question: do we want each and every image on the website to be lazy loaded?

cc @ludwiczakpawel @lisasy

ludwiczakpawel:
Unfortunately I don't have enough experience with `loading="lazy"` so I'm not the best person to ask for a review :C Not sure what are the ups and downs of such a solution. Maybe @nickytonline ?

fredden:
I'll see if I can go through and work out where we'll get the best results from lazy-loading and reduce the delta here. I agree that lazily loading all images probably isn't the best idea. It'll probably take me a while to work out which templates run which pages etc.

citizen428:
Hi @fredden, just checking in to see if you plan on finishing this? Thanks for all your effort so far!

fredden:
Sorry about the lack of updates here. Yes, this is still on my list. I'll see if I can find some time to finish it off / fix it this month.

citizen428:
@fredden Great, thanks for getting back to me! No rush 😃

msarit:
Hey @fredden! Calling your attention to this PR; let us know if you need any help moving it forward! 🙏🏾

nickytonline:
This PR just came on my radar.

One thing to take into account for this PR @fredden, is that we do not necessarily want to lazy load all images. If images are in the first visible viewport, we want the image to load immediately.

For reference, see [Browser-level image lazy-loading for the web](https://web.dev/browser-level-image-lazy-loading/#avoid-lazy-loading-images-that-are-in-the-first-visible-viewport).

fredden:
I've been having trouble getting a working local environment set up to test this properly. Thanks for the link regarding first-visible content; that'll likely be tricky to work out in the templates as there is no "fold" online as there is in traditional print media. The plan is to only mark certain images as lazily-loaded, so hopefully I can find a good balance here.

citizen428:
Hi @fredden, just checking in, are you still planning to work on this or should we close the PR?

fredden:
Thanks for following up. I'm slowly going through the templates, working out where they're used and if it's sensible to lazily load the images therein. Some already have their images loading lazily.

It took me a while to get a working local development environment going again, and I'm not able to commit lots of time/brainpower to this right now. I'd like to complete this work I started. I hope to be finished in a few weeks.

nickytonline:
One other thing to consider is, the PR doesn't need to be all or nothing. We can add the loading attribute where it makes sense in a few templates or whatever seems a reasonable amount and submit that and have subsequent PRs as this PR is already 132 files which is quite large. It would also help with reviewing the PR. Less files means reviewers can evaluate a lot better if it makes sense where the loading attribute is being used.

fredden:
I've processed about half of the list so far. I Given there are already 34 changed files, let's call this "big enough" for now. I've done all the main pages/candidates. The remaining templates are harder to navigate to in development.

jacobherrington:
Congrats on making your first PR! Thanks for helping us make Forem (and DEV) better! :tada: 

This PR seems to have a merge conflict, it has been waiting on review for a while, so that makes sense!

@fredden are you able to handle the merge conflict with a rebase or merge commit, and then verify that things are still ready to review/working as expected!

Thanks for hanging with this change, it seems like it slipped through the cracks over the last couple of weeks.

fredden:
Thanks for flagging @jacobherrington. I've resolved the conflicts.

aitchiss (approved):
Looks good to me! Thanks for this 😄

rhymes (approved):
Thank you a lot @fredden! And thank you for sticking with this over all these months! :D

PR #10892, opened 2020-10-17
Add aliases for mod_roundrobin_notifications + email_follower_notifications to user model #6892

### LINKED ISSUE #6892: Rename mod_roundrobin_notifications + email_follower_notifications on user model
We have columns on our `users` table that are boolean values that we use to represent whether a user is subscribed to an email or a notification. However, these columns [don't have intuitive names](https://github.com/thepracticaldev/dev.to/pull/6852#discussion_r398669840). For example, `mod_roundrobin_notifications`, `email_follower_notifications`, and `welcome_notifications` indicate an ActiveRecord relation, not a boolean `true`/`false` value.

We should _carefully_ rename these columns to something that actually indicates the fact that these are boolean values. So, instead of `user.welcome_notifications`, the attribute could instead be: `user.subscribed_to_welcome_notifications`. 

Notably, in order to do this carefully, we'll want to:
1) add new columns with the correct names
2) update the new columns with the correct data
3) ensure that we replace the old column/attributes the new column/attributes in the correct places in the code base
4) finally then remove the old columns

matisnape:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Closes #6892 

n/a

## Added tests?

## Added to documentation?

- [x] no documentation needed

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -206,6 +206,8 @@ class User < ApplicationRecord
 
   alias_attribute :public_reactions_count, :reactions_count
   alias_attribute :subscribed_to_welcome_notifications?, :welcome_notifications
+  alias_attribute :subscribed_to_mod_roundrobin_notifications?, :mod_roundrobin_notifications
+  alias_attribute :subscribed_to_email_follower_notifications?, :email_follower_notifications
 
   scope :eager_load_serialized_data, -> { includes(:roles) }
   scope :registered, -> { where(registered: true) }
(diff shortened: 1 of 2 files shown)


matisnape:
`/spec/system/user/view_user_index_spec.rb:19` this spec is failing on master as well 🤔

rhymes (approved):
Looks great, thanks @matisnape

benhalpern (approved):

PR #10917, opened 2020-10-18
Add admin config to always show manually selected users on "Suggested people to follow" list

### LINKED ISSUE #9367: Allow a new Forem to toggle suggested users to follow on or off. When on, then it over rides the default algorithm.
This feature request is related to the amazing work completed in PR #7637
https://github.com/forem/forem/pull/7637

At the moment, the platform over rides the default user list when there has been some user activity, then switches over to the automated follower list, problem is that then that list can display just a single user.

I have suggested that for now there is a simple toggle switch that always overrides the system generated list, and it is up to the site admin to decide when the platform is busy enough to switch over.

sidemt:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

- Added a config to allow admins to choose to always show "Suggested users" instead of system-generated suggestions in "Suggested people to follow" screen.

I also made some related fixes:
- Fixed some missing whitespace in the description of "Suggested users" input field 
- Refactored some tests related to "Suggested users" and this change to resolve the linting errors I encountered

Closes #9367

http://localhost:3000/admin/config > Under "All Site Configuration" > Onboarding
- I added a checkbox
issue_9367

- I don't know if there is an easy way to see the onboarding process again but if you visit http://localhost:3000/users?state=follow_suggestions you can see the list of users in JSON format. You can validate that the list changes according to this config.

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed


benhalpern:
I like the naming "prefer x" is a good type of config language we can integrate more. 👍

app/controllers/users_controller.rb
```diff
@@ -280,7 +280,11 @@ def determine_follow_suggestions(current_user)
       attributes_to_select: INDEX_ATTRIBUTES_FOR_SERIALIZATION,
     ).suggest
 
-    recent_suggestions.presence || default_suggested_users
+    if SiteConfig.prefer_manual_suggested_users?
```
vaidehijoshi, line 283: Since we are allowing Forem creators to override follow suggestions, we can put this at the top of the method and short-circuit out if it is `true` (which would mean we wouldn't have to call out to the `Suggester::Users::Recent` if we don't need to 😄).

We also want to check that `default_suggested_users` is actually _set_ and has a value; otherwise, there could be a scenario where an admin checks the `prefer_manual_suggested_users` checkbox, but doesn't have any suggested users actually set!

```ruby
return default_suggested_users if SiteConfig.prefer_manual_suggested_users? && default_suggested_users

# recent_suggestions = ....
```

app/controllers/users_controller.rb
```diff
@@ -280,7 +280,11 @@ def determine_follow_suggestions(current_user)
...
+      default_suggested_users
+    else
+      recent_suggestions.presence || default_suggested_users
```
vaidehijoshi, line 286:
```suggestion
      recent_suggestions || default_suggested_users
```
Since `nil` is falsy in ruby, if `recent_suggestions` is `false`, ruby should return the value of `default_suggested_users` (so you shouldn't need `presence` I believe!)

app/lib/constants/site_config.rb
```diff
@@ -283,11 +283,15 @@ module SiteConfig
...
       suggested_users: {
-        description: "Determines which users are suggested to follow to new users during onboarding (comma" \
-        "separated, letters only). Please note that these users will be shown as a fallback if no" \
+        description: "Determines which users are suggested to follow to new users during onboarding (comma " \
+        "separated, letters only). Please note that these users will be shown as a fallback if no " \
         "recently-active commenters or producers can be suggested",
         placeholder: "List of valid usernames: comma separated, letters only e.g. ben,jess,peter,maestromac,andy,liana"
       },
+      prefer_manual_suggested_users: {
+        description: "Always show the users determined above as suggested people to follow even when " \
```
vaidehijoshi, line 292: I think we should reword this description somehow such that `suggested_users` doesn't _always_ have to be rendered _above_ the `prefer_manual_suggested_users` in the config view. If it ever gets moved around in `app/views/admin/configs/show.html.erb`, then we would have to remember to change this wording, too 😅 
```suggestion
        description: "Always show suggested users as suggested people to follow even when " \
```

app/views/admin/configs/show.html.erb
```diff
@@ -1043,6 +1043,12 @@
                 <div class="alert alert-info"><%= Constants::SiteConfig::DETAILS[:suggested_users][:description] %></div>
               </div>
 
+              <div class="form-group">
```
vaidehijoshi, line 1046: I think we should try to use the crayons classes that we use for checkboxes elsewhere in this view. For example, in the authentication section of this page, we have a `crayons-field--checkbox` class that renders nicely:

```ruby
<fieldset class="config-authentication-form">
  <div class="crayons-field--checkbox">
    <%= f.check_box :prefer_manual_suggested_users, checked: SiteConfig.prefer_manual_suggested_users, class: "crayons-checkbox" %>
    <div>
      <%= admin_config_label :prefer_manual_suggested_users %>
      <p class="crayons-field__description">
        <%= Constants::SiteConfig::DETAILS[:prefer_manual_suggested_users][:description] %>
       </p>
     </div>
   </div>
 </fieldset>
  <br>

  <%= render "form_submission", f: f %>
```

Which renders like this:

Screen Shot 2020-10-20 at 3 45 38 PM

Curious to hear what @lisasy thinks about using `config-authentication-form` here (maybe we ought to rename it if we use it in other places?)

spec/requests/user/user_suggestions_spec.rb
```diff
@@ -56,20 +59,31 @@
...
+        expect(response).to have_http_status(:ok)
+        expect(response.parsed_body.first).to include(
+          "id" => suggested_user.id,
+          "name" => suggested_user.name,
+          "username" => suggested_user.username,
+          "summary" => suggested_user.summary,
+          "profile_image_url" => Images::Profile.call(suggested_user.profile_image_url, length: 90),
+          "following" => false,
+        )
+      end
```
vaidehijoshi, line 86: Thanks for adding these tests! 👏

vaidehijoshi (commented):
Hi, @sidemt! Thanks so very much for taking this issue on and submitting this PR 😸

I left a couple of comments and suggestions for you, let me know what you think!

spec/requests/user/user_suggestions_spec.rb
juliannatetreault: I second @vaidehijoshi -- thank you for adding these thorough tests! 🎉

app/views/admin/configs/show.html.erb
lisasy: I've been using `config-authentication-form` squarely for the authentication section, so I wouldn't use it here. 

@sidemt can you merge master to this? We've made substantial updates to the styling and markup here to reflect more Crayons styling. Getting that merged would be a very cool thing to do.

vaidehijoshi: Sounds good; I defer fully to @lisasy on what crayons styles _should_ be used here :)

app/controllers/users_controller.rb
sidemt: Thanks for your suggestion! `presence` was in the original code before my change so I kept it. I'm not sure but I suppose we use `presence` here because `recent_suggestions` could be an empty array `[]` which is truthy?

app/controllers/users_controller.rb
sidemt: Updated. Thank you!

app/views/admin/configs/show.html.erb
sidemt: Updated. I referred to the User Experience and Brand > Public checkbox and used the same style.

sidemt:
@vaidehijoshi @lisasy Thanks for the review! I made changes according to your comments so could you please check again?

app/views/admin/configs/show.html.erb
lisasy: @sidemt the styling looks good to me 👍

lisasy (approved):
LGTM!

vaidehijoshi (approved):
Thanks for making the updates to this PR! I've restarted the build for you, and once it passes, we should be able to get this merged :)

app/controllers/users_controller.rb
vaidehijoshi: Ah, that makes much more sense! I think it's correct to leave it in, in that case 👍

PR #10754, opened 2020-10-09
Dynamic frontend for the sidebar navigation Items

Ridhwana:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In this PR I've created a dynamic frontend that uses the new NavigationLinks table. This means that the sidebar is no longer hardcoded but can be configured from http://localhost:3000/admin/navigation_links. 

In addition to this PR there is also related docs which you can find https://app.gitbook.com/@forem/s/forem-admin-guide/~/drafts/-MJN9RF0nZtuN9YWy-IM/admin/navigation-links/@drafts

## Related Tickets & Documents
Follow up PR of  https://github.com/forem/forem/pull/9444 and https://github.com/forem/forem/pull/10662
https://github.com/forem/forem/issues/4661
https://github.com/forem/forem/issues/5475

## QA Instructions, Screenshots, Recordings

1. Open up a rails console
2. Run `NavigationLink.all`, ensure that the count matches the sidebar item count on http://localhost:3000/
3. Add, edit and delete items and ensure that it shows the updated changes on the homepage.

A view of the backend:
Screenshot 2020-10-12 at 11 46 38

A view of the corresponding frontend:
Screenshot 2020-10-12 at 11 46 22

## Added tests?

- [x] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [x] docs.forem.com
- [ ] readme
- [ ] no documentation needed

## [optional] What gif best describes this PR or how it makes you feel?

alt_text

PR #10921, opened 2020-10-19
Make the footer links dynamic

Ridhwana:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
In [another PR](https://github.com/forem/forem/pull/10754) we added a NavigationLinks Table and made all the links on the sidebar dynamic, such that an admin can configure what links they want to show on http://localhost:3000/admin/navigation_links and it will appear on the sidebar. 

We now applying that same workflow to the footer and rendering the items from the NavigationLinks Table instead of hardcoding them. 

_Note: I've had to amend use grids more dynamically and make the template render in a dynamic way. This means that we were not able to group the links in the same manner_

https://github.com/forem/InternalProjectPlanning/projects/5#card-47436154

**Hardcoded Links:** 
Screenshot 2020-10-19 at 11 22 08

**Dynamic Links:**
Screenshot 2020-10-19 at 14 14 02

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

alt_text

### DIFF
--- a/app/assets/stylesheets/views/footer.scss
+++ b/app/assets/stylesheets/views/footer.scss
@@ -30,7 +30,6 @@
 
   &__nav {
     display: grid;
-    gap: var(--su-4);
     flex: 1 auto;
     grid-template-columns: repeat(2, 1fr);
 
--- a/app/assets/stylesheets/views/footer.scss
+++ b/app/assets/stylesheets/views/footer.scss
@@ -30,15 +30,18 @@
 
   &__nav {
     display: grid;
-    flex: 1 auto;
-    grid-template-columns: repeat(2, 1fr);
+    column-gap: var(--su-4);
+    grid-template-rows: repeat(9, 1fr);
+    grid-auto-flow: column;
 
     @media (min-width: $breakpoint-s) {
-      grid-template-columns: repeat(3, 1fr);
+      grid-template-rows: repeat(6, 1fr);
+      column-gap: var(--su-8);
     }
 
     @media (min-width: $breakpoint-l) {
-      grid-template-columns: repeat(3, 200px);
+      grid-template-rows: repeat(6, 1fr);
+      column-gap: var(--su-10);
     }
   }
 
--- a/app/views/articles/_sidebar_nav.html.erb
+++ b/app/views/articles/_sidebar_nav.html.erb
@@ -37,8 +37,8 @@
           <%= render "articles/sidebar_nav_link", link: link %>
         <% end %>
       </div>
+    <% end %>
   </div>
-  <% end %>
 </nav>
 
 <nav class="sidebar-tags-browser mb-6 overflow-y-auto scrolling-touch" aria-label="Secondary sidebar nav">
--- a/app/views/layouts/_footer.html.erb
+++ b/app/views/layouts/_footer.html.erb
@@ -11,15 +11,14 @@
     </div>
 
     <nav class="crayons-footer__nav">
-      <div>
         <a href="/" class="crayons-link crayons-link--block">Home</a>
-        <% if user_signed_in? %>
-          <a href="<%= readinglist_path %>" class="crayons-link crayons-link--block">Reading list</a>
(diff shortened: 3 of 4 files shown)


Ridhwana:
@ludwiczakpawel we're working on a bunch of priority issues for a launch, and making the footer dynamic is one of them. In order to make it dynamic, I had to amend the styling. Please can you review :)

ludwiczakpawel:
I do understand the reasoning but I don't think it's the right solution. 

image

Except the things that seem to be broken (see screenshot above) there are some design/product things I'm not sold on, for example - if we make this whole thing dynamic in that way, we will likely introduce some confusion around order of the links. Previously, links were more or less logically grouped together into columns... 

```
Link 1       Link 4       Link 7
Link 2       Link 5       Link 8
Link 3       Link 6     
```

And now it's no longer grouped in that way, instead we have this:

```
Link 1       Link 2       Link 3
Link 4       Link 5       Link 6
Link 7       Link 8     
```

The groups/columns I’ve mentioned were supposed to be:
1. Main information architecture: home, reading list, listings, podcasts, …
2. Less important pages, mostly content pages: CoC, FAQ, About, …
3. „Other stuff” like Sponsors, DEV Shop

With new solution it will be all kinda mixed together and not very logical therefore the columns-like UI here doesn’t make much sense.

Link2Twenty:
It feels like the backend needs a column field too.

link name
link location
link column
link position

Ridhwana:
Some discussions regarding this issue: 

> Previously, links were more or less logically grouped together into columns...
The Links should definitely run per column and not per row, that was my mistake and I'll fix that up now. 

>The groups/columns I’ve mentioned were supposed to be:
>Main information architecture: home, reading list, listings, podcasts, …
>Less important pages, mostly content pages: CoC, FAQ, About, …
>„Other stuff” like Sponsors, DEV Shop

These groupings will become difficult to form/maintain as the links are now dynamic and the user can create _any number of links and any type of links_, - these links may not confirm to these groupings. One "potential grouping" may also have many more links than another grouping which would have an impact on the UI. Hence I guess we could possibly maintain these groupings on the admin as well as per @Link2Twenty's suggestion of adding a `column name/grouping`, although I do feel like we should give this more thought before implementing - taking into consideration the dynamic nature of the links and how it will impact these groupings. Also maybe we should discuss do we limit the number of links that a user can create? 

However an interim solution I'd like to propose is that we match up the order of the sidebar and the footer. I think that if we want sponsors and DEV Shop to show up together on the footer, we should have them after each other on the sidebar as well. I can't think of a reason why the logical order on the sidebar and footer should be different. Having these reflect each other would at least provide a good order (but would not sort out the groupings issue). We could for now maintain 6 items in per column. 

It won't be an exact match to what we currently have on DEV and we'd need to figure out what the next step is to get this in a better shape. In the future these issues should be discussed with product/design beforehand.

Ridhwana:
Update: I've amended the flow to be column based instead of row based.

Ridhwana:
**_Updated Note:_**

screencapture-localhost-3000-admin-navigation-links-2020-10-19-14_20_06

If we amended our order of our navigation links on the sidebar to be logically inline with our groupings we could accomplish the following:

_Hardcoded Links:_
Screenshot 2020-10-19 at 14 20 38


_Dynamic Links:_
Screenshot 2020-10-19 at 14 20 23


However, keep in mind this will alter the order of the sidebar as well, which I don't think is a bad thing as it will be more logical:
Screenshot 2020-10-19 at 14 23 17

Thoughts?

pkfrank (approved):
This looks good to me from a functionality perspective.  That said, I'm not capable of "approving" the technical implementation so defer to others there.

benhalpern (approved):
This looks good to me right now. It's also a fairly small change we can build on later with further customization, but serves our needs very well now.

PR #10926, opened 2020-10-19
Adding admin link for admins

ludwiczakpawel:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR will add a link to admin section if user is an admin...

- Closes https://github.com/forem/InternalProjectPlanning/issues/173

image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/assets/javascripts/initializers/initializeBaseUserData.js
+++ b/app/assets/javascripts/initializers/initializeBaseUserData.js
@@ -16,7 +16,7 @@ function initializeUserProfileContent(user) {
 function initializeProfileImage(user) {
   if (!document.getElementById('comment-primary-user-profile--avatar')) return;
   document.getElementById('comment-primary-user-profile--avatar').src =
-  user.profile_image_90;
+    user.profile_image_90;
 }
 
 function initializeUserSidebar(user) {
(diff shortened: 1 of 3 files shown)


Zhao-Andy (approved):
I'm forgetting if `current_user` is able to be used here because of caching but _I think_ it should be okay since there's an `if user_signed_in?` above... cc @benhalpern

ludwiczakpawel:
Additional context from @atsmith813 https://github.com/forem/InternalProjectPlanning/issues/173#issuecomment-712192200

atsmith813:
This won't work as expected in production due to caching and service workers.

At this time, we can't modify the server-side views of the content cached by service workers to be user-specific. It's currently set up so that there's a "signed in" version and a "signed out" version. We can't introduce a user-specific specific version (i.e. users with role of admin) without changing a lot more.

An example of this is the profile image/avatar in the nav bar - a user-specific element. That is actually populated by some [async JS](https://github.com/forem/forem/blob/3ccecfbbb7fe2524f0c450274dbef9560d21d033/app/assets/javascripts/initializers/initializeBaseUserData.js#L112-L118). We can't render that server-side.

I suspect a "quick win" may be to do some similar JS to update this dropdown with an async script. I'm not sure where we could tell the client-side that the user is, in fact, an admin 🤔 . My knee-jerk reaction would be to add the info [to the  `user_data` used by many components on the front-end](https://github.com/forem/forem/blob/4e1e5a45624f9a9ade75f08b8605aa54ebf0a58d/app/controllers/async_info_controller.rb#L47-L72), but I'm hesitant to do that with roles/admin because that's sort of sensitive data.

benhalpern (changes requested):
This will actually need to be inserted dynamically because we may cache this page....

Probably makes sense in the `setCurrentUserToNavBar` function....

Will need to add `admin: @user.any_admin?` to `/async_info/base_data` in order to fetch this information.

Let me know if you need more help with any of that.

ludwiczakpawel:
Ok, I pushed an update. @benhalpern would you mind taking a look, I'm not sure if I'm approaching this correctly (this kind of work is pretty new for me).

benhalpern (approved):
Looks great to me

PR #10933, opened 2020-10-19
Fix Reading List Counter Without last_reacted_at Default

mstruve:
Same as #10763 but without the default for the column which locked up our database for 20+ min. `last_reacted_at` should be fine being nil.

alt_text


benhalpern (approved):
This looks good to me.

maestromac (approved):

PR #10934, opened 2020-10-19
Add invitations controller to Allow List

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When a forem is private it intercepts _most_ requests and delivers the auth page..... We have a list of controllers which should not have this functionality, and invitations should be one of them.

### DIFF
--- a/spec/requests/invitations_spec.rb
+++ b/spec/requests/invitations_spec.rb
@@ -1,0 +1,16 @@
+require "rails_helper"
+
+RSpec.describe "Invitations", type: :request do
+  let(:user) { create(:user) }
+
+  describe "Accept invitation" do
+    it "renders normal response even if site config is private" do
+      allow(SiteConfig).to receive(:public).and_return(false)
+      get "/users/invitation/accept?invitation_token=blahblahblahblah"
+      # This is a fake token, so the only thing we're testing for here is
+      # that we *do not* land on the "registrations" page which shouldn't
+      # interrupt the request, even for private forems. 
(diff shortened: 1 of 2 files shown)


spec/requests/invitations_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  let(:user) { create(:user) }
+
+  describe "Accept invitation" do
+    it "renders normal response even if site config is private" do
+      allow(SiteConfig).to receive(:public).and_return(false)
+      get "/users/invitation/accept?invitation_token=blahblahblahblah"
+      # This is a fake token, so the only thing we're testing for here is
+      # that we *do not* land on the "registrations" page which shouldn't
+      # interrupt the request, even for private forems. 
+      expect(response.body).not_to include("registration__actions")
```
Zhao-Andy, line 13: **(non-blocking)** I think doing a check like `expect(request.path).to eq("/users/invitations/accept"` is less brittle than checking for HTML on the page.

Zhao-Andy (approved):
One non blocking comment, otherwise looks good 👍

spec/requests/invitations_spec.rb
benhalpern: Yeah, but the path wouldn't redirect. The only thing we do is render a different view. There probably _are_ different tests we could do but nothing is immediately jumping out to me.

juliannatetreault (approved):
Yay, for the `invitations_spec`!

PR #10882, opened 2020-10-16
Part One: Fix Users Stuck in a "Created" but "Unregistered" State

juliannatetreault:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR is the first of likely a couple of tiny PRs aimed at fixing a user's registration status upon acceptance of an email invitation. This PR addresses the issue that, upon invite acceptance, some Forem users are still stuck in an "unregistered" state. To accomplish this, this PR does the following:
  

- Adds `#after_accept_path_for` to the `Application::Controller` that redirects new users who accepted a Forem invite via email to the onboarding process after setting a password for the new account

- Adds `update!` to the `Invitations::Controller` to update a user's `registered` status from `false` to `true` upon invitation acceptance and their `registered_at` timestamp to the time that the invite was accepted

- Adds a `respond_with` in place of a hard redirect to `Invitations::Controller` 

For further information on Devise's `devise_invitable` gem, please take a look at the GitHub repo: https://github.com/scambra/devise_invitable

## Related Tickets & Documents
Relates to [issue #133](https://github.com/forem/InternalProjectPlanning/issues/133)

## QA Instructions, Screenshots, Recordings
**To QA this PR, please make sure that you have `admin` privileges and then do the following**:

- _**Pro-Tip**_: Replace lines 15 to 22 the `Admin::Invitations::Controller` (right above the `redirect_to admin_invitations_path`) with the code below to make the invite acceptance link readily available in the view after sending an invite:
```
      user = User.invite!(email: email,
                   name: name,
                   username: username,
                   remote_profile_image_url: Users::ProfileImageGenerator.call,
                   saw_onboarding: false,
                   editor_version: :v2,
                   registered: false)
                   flash[:success] = accept_user_invitation_url(invitation_token: user.raw_invitation_token)
```

- Navigate to `/admin/invitations/new` and invite a new user to the site:
Screen Shot 2020-10-19 at 12 27 44 PM

- Using your console, check that the user, after sending the invite, is `registered: false` and has a `registered_at` timestamp of `nil`:
```
[2] pry(main)> user = User.last
[2] pry(main)> user.registered
=> false
[3] pry(main)> user.registered_at
=> nil
```

[...]

Issue #133, opened 2018-01-09
Edit profile: Github username has already been taken


Einenlum:
### Current Behavior

Editing my profile (whatever I edit) and submitting it gives me a Github error: "Github username has already been taken".

### Expected Behavior

Being able to edit my profile without error.

### Screenshots

![devto](https://user-images.githubusercontent.com/5675200/34707493-1ef4cfc6-f50e-11e7-8186-5d0ae56c6da0.png)

Einenlum:
Finally \o/ Wuhuuuu!

Thanks a lot for your help <3.

kylestratis:
@Zhao-Andy Is it possible to reopen this? I'm having the same error trying to edit my profile.

Zhao-Andy:
Hey @kylestratis, you should be good now. Let me know if otherwise.

Also, if anyone else comes across this; sorry the long standing bug. 🙈

kylestratis:
Thanks @Zhao-Andy - works perfectly now

guar47:
Hey @Zhao-Andy! Could you check my profile too, please?  I have the same problem.

Zhao-Andy:
Hey @guar47, fixed :)

guar47:
@Zhao-Andy Thank you, it's working!

PR #10939, opened 2020-10-19
Data update Script: Backfill Created, but Unregistered Users

juliannatetreault:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR adds a `data_update` script to backfill the `registered` and `registered_at` columns on users who are created, but not registered. Currently, there are some users who still are considered `registered: false` even though they've created an account via an invitation by email. By adding this `data_update` script, we'll "fix" the `registered` columns for those particular users, creating a clean slate for us to test whether or not the root issue of this problem has been fixed via pull request #10882 

Relates to PR #10882 and closes [Internal issue #133](https://github.com/forem/InternalProjectPlanning/issues/133).

Make sure that Travis and all other checks build green! ♻️ 

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

This _is_ a post-deployment task. 😎 

Miss Piggy Invited

### DIFF
--- a/lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
+++ b/lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
@@ -1,10 +1,8 @@
 module DataUpdateScripts
   class BackfillRegisteredForUsers
     def run
-      users = User.where(registered: false).or(User.where(registered_at: nil))
+      users = User.where.not(invitation_accepted_at: nil).where(registered: false).or(User.where(registered_at: nil))
       users.find_each do |user|
-        next if user.invitation_accepted_at.nil?
-
         user.update_columns(registered: true, registered_at: user.invitation_accepted_at)
       end
     end
--- a/lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
+++ b/lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
@@ -1,0 +1,12 @@
+module DataUpdateScripts
+  class BackfillRegisteredForUsers
+    def run
+      return if User.created_at.nil?
+
+      users = User.where(registered: false).or(User.where(registered_at: nil))
+      users.find_each do |user|
+        user.update_columns(registered: true, registered_at: user.created_at)
+      end
+    end
+  end
+end
--- a/lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
+++ b/lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
@@ -1,11 +1,11 @@
 module DataUpdateScripts
   class BackfillRegisteredForUsers
     def run
-      return if User.created_at.nil?
-
       users = User.where(registered: false).or(User.where(registered_at: nil))
       users.find_each do |user|
-        user.update_columns(registered: true, registered_at: user.created_at)
+        next if user.created_at.nil? || user.invitation_accepted_at.nil?
+
+        user.update_columns(registered: true, registered_at: user.invitation_accepted_at)
       end
     end
   end
--- a/lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
+++ b/lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
@@ -3,7 +3,7 @@ module DataUpdateScripts
     def run
       users = User.where(registered: false).or(User.where(registered_at: nil))
       users.find_each do |user|
-        next if user.created_at.nil? || user.invitation_accepted_at.nil?
+        next if user.invitation_accepted_at.nil?
 
         user.update_columns(registered: true, registered_at: user.invitation_accepted_at)
       end


benhalpern (commented):
This will mark every _invited_ user as registered.... I don't think that makes sense. For some users who were not able to successfully register, it makes sense, but I don't think we can blanket register all invited users.

lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
```diff
@@ -0,0 +1,12 @@
+module DataUpdateScripts
+  class BackfillRegisteredForUsers
+    def run
+      return if User.created_at.nil?
```
juliannatetreault, line 4: **Question for my reviewers**: if necessary, how should I differentiate between users who were invited but _have not yet accepted the invitation (are not registered yet) and users who were invited and _have accepted the invitation_ (are, or should be considered, registered)? I think this distinction is important, but I'm not sure how to proceed with it.

juliannatetreault:
> This will mark every _invited_ user as registered.... I don't think that makes sense. For some users who were not able to successfully register, it makes sense, but I don't think we can blanket register all invited users.

It looks like we commented at the same time @benhalpern . 😊  I actually added a question related to this on the script: https://github.com/forem/forem/pull/10939#discussion_r508048172

lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
benhalpern: One more thing: `User.created_at` won't work. `.created_at` needs to be invoked on the instance in the loop, not the class here.

But within the loop if we can determine which users _tried_ to accept an invitation that would definitely be the thing to check for.

If `invitation_accepted_at` is properly being set in that path, we should definitely check for users who _do_ have an `invitation_accepted_at` but do not have a `registered_at`.

benhalpern: .... And technically setting `user.registered_at = user.invitation_accepted_at` would be the most accurate date for that value.

lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
```diff
@@ -0,0 +1,12 @@
+module DataUpdateScripts
+  class BackfillRegisteredForUsers
+    def run
+      users = User.where(registered: false).or(User.where(registered_at: nil))
+      users.find_each do |user|
+        next if user.created_at.nil? || user.invitation_accepted_at.nil?
```
Zhao-Andy, line 6: I don't think `user.created_at.nil?` would ever be true right?

Zhao-Andy (commented):
One comment!

lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
juliannatetreault: I don't think so... I added it to be extra cautious!  I'm happy to take that check out, @Zhao-Andy 😄

Zhao-Andy: 👍 yeah I think taking it out makes more sense!

juliannatetreault: Done, @Zhao-Andy!

Zhao-Andy (approved):

lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
```diff
@@ -0,0 +1,12 @@
+module DataUpdateScripts
+  class BackfillRegisteredForUsers
+    def run
+      users = User.where(registered: false).or(User.where(registered_at: nil))
+      users.find_each do |user|
+        next if user.invitation_accepted_at.nil?
```
citizen428, line 6: Instead of using a guard clause I'd recommend not even fetching these users from the DB:

```ruby
users = User.where.not(invitation_accepted_at: nil).
  where(registered: false).or(User.where(registered_at: nil))

users.find_each do |user|
  user.update_columns(registered: true, registered_at: user.invitation_accepted_at)
end
```

mstruve: One step further since those `or`'s always throw me off
```ruby
users = User.where.not(invitation_accepted_at: nil).where(registered: [false, nil])
```

mstruve (approved):
suggestion for a small refactor otherwise LGTM!

lib/data_update_scripts/20201019192035_backfill_registered_for_users.rb
juliannatetreault: Great suggestions, @citizen428 and @mstruve! I made this change.

citizen428: @mstruve The `or` seemed necessary though, the condition checked two different columns, `registered` (a boolean) and `registered_at` (a timestamp).

juliannatetreault: You're correct @citizen428 -- I went with your implementation rather than using the `or` (even though it was a really nice way of setting it all up, had it worked) for that reason. 😊

PR #10942, opened 2020-10-19
Redirect to SiteConfig.app_domain

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I'm not totally sure what the best way to test this is, so if anyone wants to merge a test in here or suggest something, be my guest.

This functionality redirects from the originally set app domain to the current one. This should take care of some edge cases when domains change.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,6 +4,7 @@ class ApplicationController < ActionController::Base
   before_action :verify_private_forem
   protect_from_forgery with: :exception, prepend: true
   before_action :remember_cookie_sync
+  before_action :forward_to_app_config_domain
 
   include SessionCurrentUser
   include ValidRequest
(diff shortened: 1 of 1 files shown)


juliannatetreault (approved):
Thanks for adding all of the inline comments! Nothing stands out to me as wrong, so 👍 👍  from me.

app/controllers/application_controller.rb
```diff
@@ -163,6 +164,14 @@ def remember_cookie_sync
     end
   end
 
+  def forward_to_app_config_domain
+    return unless request.get? && # Let's only redirect get requests for this purpose.
+      request.host == ENV["APP_DOMAIN"] && # If the request equals the original set domain, e.g. forem-x.forem.cloud.
+      ENV["APP_DOMAIN"] != SiteConfig.app_domain # If the app domain config has now been set, let's go there instead.
+
+    redirect_to "#{request.protocol}#{SiteConfig.app_domain}#{request.fullpath}"
```
fdoxyz, line 172: **suggestion/question (non-blocking):** Does it make sense to use `URL.url(request.fullpath)` instead?

The only difference I see between this and `URL.url(request.fullpath)` is that if someone makes a GET request using `http://forem-x.forem.cloud` this will maintain the `http` and redirect to `http://real-domain.com`.

Using `URL.url(request.fullpath)` will redirect `http://forem-x.forem.cloud` to `https://real-domain.com` (or whatever protocol was specified for the deployment.

fdoxyz (approved):
LGTM

Not sure if it was intentional to build the redirect URL this way but left a non-blocking suggestion/question inline

app/controllers/application_controller.rb
benhalpern: Yeah this makes sense, thanks.

PR #10944, opened 2020-10-20
Fill organization userpic background fully to avoid empty corners

### LINKED ISSUE #10815: Corners of organization profile images not fully filling the space
**Describe the bug**

The very corners of profile images on **organizations** are being cut off.

**To Reproduce**

Visit an organization page, for example:
- https://dev.to/devteam
- https://dev.to/twilio

**Expected behavior**

The organization logo fills the entire profile image frame.

**Screenshots**

example of issue

query-string:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR makes organization user pic background to be fully covered which resolves empty corners issues described in #10815

Closes #10815

Screen Shot 2020-10-20 at 10 16 59

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/assets/stylesheets/views/profile.scss
+++ b/app/assets/stylesheets/views/profile.scss
@@ -101,10 +101,11 @@
   .crayons-logo {
     width: var(--su-9);
     height: var(--su-9);
-    border: var(--su-1) solid var(--profile-brand-color);
+    background-color: var(--profile-brand-color);
+    padding: var(--su-1);
 
     @media (min-width: $breakpoint-m) {
-      border-width: var(--su-2);
+      padding: var(--su-2);
       width: var(--su-10);
       height: var(--su-10);
     }
--- a/app/assets/stylesheets/views/profile.scss
+++ b/app/assets/stylesheets/views/profile.scss
@@ -104,6 +104,10 @@
     background-color: var(--profile-brand-color);
     padding: var(--su-1);
 
+    &::after {
+      content: none;
+    }
+
     @media (min-width: $breakpoint-m) {
       padding: var(--su-2);
       width: var(--su-10);


citizen428 (approved):

ludwiczakpawel (changes requested):
By default, `.crayons-logo`  element has also `::after` pseudo class to apply barely visible border inside. It can be more visible if you set your brand color to be much lighter.

With proposed change this border will be visible "outside" the logo, so I think we could overwrite it here and just drop it.

Here's a video for more details: https://d.pr/v/abGHIV

app/assets/stylesheets/views/profile.scss
```diff
@@ -101,10 +101,11 @@
   .crayons-logo {
     width: var(--su-9);
     height: var(--su-9);
-    border: var(--su-1) solid var(--profile-brand-color);
+    background-color: var(--profile-brand-color);
+    padding: var(--su-1);
 
     @media (min-width: $breakpoint-m) {
-      border-width: var(--su-2);
+      padding: var(--su-2);
```
ludwiczakpawel, line 108: ...orrrrr we could do this (instead of what I suggested in previous comment):

```
.crayons-logo {
  ...
  padding: var(--padding, var(--su-1));
  border-radius: calc(var(--radius) + var(--padding));
  @media (...) {
    --padding: var(--su-2);
  }
}
```

query-string:
Good catch @ludwiczakpawel thanks for suggestions! Resetting ::after pseudoclass seems to be the most ecological solution in this case (if we don't need this behavior – we are getting rid of it). 
Screen Shot 2020-10-21 at 09 03 50

ludwiczakpawel (approved):

PR #10945, opened 2020-10-20
Adding file extensions to gitattributes

Rafi993:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Adding file extensions to .gitattributes for better diff. It is mostly inspired from https://tekin.co.uk/2020/10/better-git-diff-output-for-ruby-python-elixir-and-more

Closes https://github.com/forem/forem/issues/10946

Try doing git diff using git cli before and after this commit on a ruby file

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed


citizen428 (approved):

.gitattributes
```diff
@@ -3,3 +3,7 @@
 # means that when checking out code, LF endings get converted
 # to CRLF.
 # *.svg text eol=lf
+*.rb   diff=ruby
+*.rake diff=ruby
+*.css  diff=css
+*.html diff=html
```
citizen428, line 9: Do you want to add HTML diffs for ERb files too?

citizen428 (approved):

.gitattributes
Rafi993: I thought so too but isn't erb little different (since it can have ruby code inside it)

citizen428: Yeah, but a lot of editors work quite ok if you set ERb's file type to HTML, so I guess it's worth at try.

Rafi993: I have added `*.erb  diff=html` now. Thank you @citizen428

.gitattributes
```diff
@@ -3,3 +3,8 @@
...
+*.erb  diff=html
```
rhymes, line 10: there's a missing newline at the end of the file

rhymes (commented):
Hi @Rafi993, thanks for the change.

What about frontend files? JS, JSX and so on? Should we add those as well

.gitattributes
Rafi993: > Hi @Rafi993, thanks for the change.
> 
> What about frontend files? JS, JSX and so on? Should we add those as well

They don't seem to have support for js or ts https://github.com/git/git/blob/master/userdiff.c#L78  as mentioned in this thread https://gist.github.com/tekin/12500956bd56784728e490d8cef9cb81

Rafi993: > there's a missing newline at the end of the file

I'll add the newline now.

rhymes (approved):

.gitattributes
rhymes: > They don't seem to have support for js or ts https://github.com/git/git/blob/master/userdiff.c#L78 as mentioned in this thread https://gist.github.com/tekin/12500956bd56784728e490d8cef9cb81

got it, thanks!

Rafi993:
Thank you @rhymes and @citizen428

PR #10950, opened 2020-10-20
Fix RSpec unit testing when using Docker dev setup

### LINKED ISSUE #10579: Attempting to run RSpec on Docker triggers Elastic error
**Describe the bug**

When I attempt to run a particular spec in Docker, there's a complaint about a missing VCR stub for elasticsearch.

The spec I'm running: `docker exec -it forem_rails bundle exec rspec spec/liquid_tags/vimeo_tag_spec.rb`
The error I get:
```
Tasks: TOP => db:test:load => db:test:purge => db:check_protected_environments
(See full trace by running task with --trace)
[Zonebie] Setting timezone: ZONEBIE_TZ="Prague"

An error occurred in a `before(:suite)` hook.
Failure/Error: next unless Search::Client.indices.exists(index: search_class::INDEX_NAME)

VCR::Errors::UnhandledHTTPRequestError:


  ================================================================================
  An HTTP request has been made that VCR does not know how to handle:
    HEAD http://elasticsearch:9200/chat_channel_memberships_test

  There is currently no cassette in use. There are a few ways
  you can configure VCR to handle this request:

    * If you're surprised VCR is raising this error
      and want insight about how VCR attempted to handle the request,
      you can use the debug_logger configuration option to log more details [1].
```

**To Reproduce**

1. Install Docker on WSL2 (Ubuntu) (Not sure if this is present in other environments, test and see?)
1. Run `bin/container-setup`
1. While docker is up, run `docker exec -it forem_rails bundle exec rspec spec/liquid_tags/vimeo_tag_spec.rb`

**Expected behavior**

All test cases pass, and RSpec reports the output.

**Screenshots**

image

- Windows 10 > (WSL2) Ubuntu > Docker

tudorpavel:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When using the Docker dev setup, Elasticsearch is available at `elasticsearch:9200` instead of `localhost:9200` so the VCR testing config was not ignoring the requests needed for RSpec setup when doing `Search::Cluster.recreate_indexes` in `spec/rails_helper.rb`. Also the WebMock setup allows `localhost` requests which should also include Elasticsearch requests.

This seems okay to me, but I'm not sure if there is a better solution. One issue I might have is that a detail related to the Docker dev setup is bleeding into common RSpec configs. What do you think, @citizen428?

Closes #10579 

With this change you can run unit tests when using the Docker dev setup, for example this spec was failing at `Search::Cluster.recreate_indexes` in `spec/rails_helper.rb` and is now passing.

```
bundle exec rspec spec/liquid_tags/vimeo_tag_spec.rb
```

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -48,7 +48,7 @@ allowed_sites = [
   "selenium-release.storage.googleapis.com",
   "developer.microsoft.com/en-us/microsoft-edge/tools/webdriver",
   "api.knapsackpro.com",
-  "elasticsearch:9200",
+  "elasticsearch",
 ]
 WebMock.disable_net_connect!(allow_localhost: true, allow: allowed_sites)
 
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -48,6 +48,7 @@ allowed_sites = [
   "selenium-release.storage.googleapis.com",
   "developer.microsoft.com/en-us/microsoft-edge/tools/webdriver",
   "api.knapsackpro.com",
+  "elasticsearch:9200",
 ]
 WebMock.disable_net_connect!(allow_localhost: true, allow: allowed_sites)
 
--- a/spec/support/initializers/vcr.rb
+++ b/spec/support/initializers/vcr.rb
@@ -13,7 +13,7 @@ VCR.configure do |config|
     "github.com/mozilla/geckodriver/releases",
     "selenium-release.storage.googleapis.com",
     "developer.microsoft.com/en-us/microsoft-edge/tools/webdriver",
-    "api.knapsackpro.com", "localhost", "127.0.0.1", "0.0.0.0"
+    "api.knapsackpro.com", "localhost", "127.0.0.1", "0.0.0.0", "elasticsearch"
   )
 
   # Removes all private data (Basic Auth, Set-Cookie headers...)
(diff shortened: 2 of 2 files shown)


spec/rails_helper.rb
```diff
@@ -48,6 +48,7 @@
   "selenium-release.storage.googleapis.com",
   "developer.microsoft.com/en-us/microsoft-edge/tools/webdriver",
   "api.knapsackpro.com",
+  "elasticsearch:9200",
```
mstruve, line 51: We are in the process of [setting up Docker for our CI](https://github.com/forem/forem/compare/mstruve/buildkite-ci#diff-053aa3ea46cb7c6a649e0d0fc592ce729b66e1d44be475516fd58857ba695862R51) and this is one of the things I also had to add so +1 for doing it here. I think you might also need to add redis here as well. Worth checking.

Also, I believe you only need the host, `elasticsearch` and not the port.

mstruve (approved):
Thanks for the fix! Since this does indeed solve the ticket I will approve, but I made a suggestion for also checking Redis if you have the time. 

Thanks for your contribution!

spec/rails_helper.rb
tudorpavel: > Also, I believe you only need the host, `elasticsearch` and not the port.

True, I added the port to be more specific but in this case it shouldn't be an issue.

> I think you might also need to add redis here as well. Worth checking.

It's possible I missed it because I get some other unrelated failures such as missing Chrome binary. However I ran the entire RSpec suite with my Docker dev setup and I didn't see a VCR or WebMock failure for Redis requests, so I'm not sure it's needed. At least not until someone explicitly makes some Redis requests from RSpec.

maestromac (approved):
LGTM!

PR #10954, opened 2020-10-20
Disable ability to upload videos on *some* forems Part 1

Ridhwana:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As a Forem admin, video upload does not work so I do not want videos highlighted to my community or have it appear as an option to my members.

**Ideal Solution**
One day, we'll figure out whether or not we can support native video for all Forems but until then, let's:

- remove the 'upload a video' button from /dashboard

**Additional Context:**
We'll want to continue supporting videos on DEV so want to note that the upload button should still exist on DEV.

In this PR I create a config variable to determine whether to enable videos for a Forem. I do not expose it through the config UI since admins don't need control over this at the moment. We simply want to disable it for all Forems except DEV.
 
Once we deploy this, I'll add an additional check:
```
  <% if policy(:video).new? && SiteConfig.enable_video_upload %>
    <a class="crayons-btn crayons-btn--secondary w-100 mt-4" href="<%= new_video_path %>" data-no-instant> Upload a video
    </a>
  <% end %>
```

## Related Tickets & Documents
https://github.com/forem/InternalProjectPlanning/issues/149

## QA Instructions, Screenshots, Recordings

1. Open the console.
2. Run SiteConfig.enable_video_upload

## Added tests?

- [ ] yes
- [x] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

## [optional] Are there any post deployment tasks we need to perform?

YES, please. Set the variable `SiteConfig.enable_video_upload = true` on DEV. Once this is done, I'll deploy a change where we only show the upload button when SiteConfig.enable_video_upload is true.

Please let me know when this is done.

## [optional] What gif best describes this PR or how it makes you feel?

alt_text

PR #10955, opened 2020-10-20
Disable ability to upload videos on *some* forems Part 2

Ridhwana:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As a Forem admin, video upload does not work so I do not want videos highlighted to my community or have it appear as an option to my members.

**Ideal Solution**
One day, we'll figure out whether or not we can support native video for all Forems but until then, let's:

- remove the 'upload a video' button from /dashboard

**Additional Context:**
We'll want to continue supporting videos on DEV so want to note that the upload button should still exist on DEV.

In [this PR](https://github.com/forem/forem/pull/10954) I created a config variable to determine whether to enable videos for a Forem. I do not expose it through the config UI since admins don't need control over this at the moment. We simply want to disable it for all Forems except DEV.
 
Here I've added the additional check to not show "Upload a video" unless the forem enables it through the Siteconfig.enable_video_upload config. 

```
  <% if policy(:video).new? && SiteConfig.enable_video_upload %>
    <a class="crayons-btn crayons-btn--secondary w-100 mt-4" href="<%= new_video_path %>" data-no-instant> Upload a video
    </a>
  <% end %>
```

https://github.com/forem/InternalProjectPlanning/issues/149
https://github.com/forem/forem/pull/10954

Go to http://localhost:3000/dashboard. You **should not see** the upload a video button. 

Now:
1. Open the console.
2. Run SiteConfig.enable_video_upload = true 
3. Make sure that your user is not a recently added one, so find your user and run the ffg:
u.created_at = 4.weeks.ago where u is your user. 
4. The Upload a video link should show up now :) 

## Added tests?

- [x] yes
- [] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Ensure that the ffg variable is set `SiteConfig.enable_video_upload = true` on DEV ebfore deploy.

alt_text

### DIFF
--- a/app/policies/video_policy.rb
+++ b/app/policies/video_policy.rb
@@ -1,15 +1,13 @@
 class VideoPolicy < ApplicationPolicy
+  def new?
+    user.created_at < 2.weeks.ago if user.created_at
+  end
+
   def create?
     user.created_at < 2.weeks.ago if user.created_at
   end
 
   def enabled?
-    new? && SiteConfig.enable_video_upload
-  end
-
-  private
-
-  def new?
-    user.created_at < 2.weeks.ago if user.created_at
+    SiteConfig.enable_video_upload
   end
 end
--- a/app/policies/video_policy.rb
+++ b/app/policies/video_policy.rb
@@ -1,9 +1,15 @@
 class VideoPolicy < ApplicationPolicy
-  def new?
+  def create?
     user.created_at < 2.weeks.ago if user.created_at
   end
 
-  def create?
+  def enabled?
+    new? && SiteConfig.enable_video_upload
+  end
+
+  private
+
+  def new?
     user.created_at < 2.weeks.ago if user.created_at
   end
 end
--- a/app/views/dashboards/_actions.html.erb
+++ b/app/views/dashboards/_actions.html.erb
@@ -51,7 +51,7 @@
     <% end %>
   <% end %>
 
-  <% if policy(:video).new? %>
+  <% if policy(:video).new? && SiteConfig.enable_video_upload %>
     <a class="crayons-btn crayons-btn--secondary w-100 mt-4" href="<%= new_video_path %>" data-no-instant> Upload a video
     </a>
   <% end %>
--- a/app/views/dashboards/_actions.html.erb
+++ b/app/views/dashboards/_actions.html.erb
@@ -51,7 +51,7 @@
     <% end %>
   <% end %>
 
-  <% if policy(:video).enabled? %>
+  <% if policy(:video).new? && policy(:video).enabled? %>
     <a class="crayons-btn crayons-btn--secondary w-100 mt-4" href="<%= new_video_path %>" data-no-instant> Upload a video
     </a>
   <% end %>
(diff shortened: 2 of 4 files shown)


app/views/dashboards/_actions.html.erb
```diff
@@ -51,7 +51,7 @@
     <% end %>
   <% end %>
 
-  <% if policy(:video).new? %>
+  <% if policy(:video).new? && SiteConfig.enable_video_upload %>
```
rhymes, line 54: what if we move this check in the policy or wrap them both in a helper?

spec/requests/dashboard_spec.rb
```diff
@@ -80,8 +80,16 @@
         expect(response.body).not_to include("Pro Analytics for #{organization.name}")
       end
 
-      it "does not render a link to upload a video" do
+      it "does not render a link to upload a video when enable_video_upload is false" do
         get dashboard_path
+        SiteConfig.enable_video_upload = false
```
rhymes, line 85: We should use `allow(SiteConfig).to receive(:enable_video_upload).and_return(false)` so this doesn't stay set for other tests

spec/requests/dashboard_spec.rb
```diff
@@ -80,8 +80,16 @@
...
+
+        expect(response.body).not_to include("Upload a video")
+      end
+
+      it "does not render a link to upload a video for a recent user" do
+        get dashboard_path
+        SiteConfig.enable_video_upload = true
```
rhymes, line 92: same as above

rhymes (commented):
Left a comment about the tests mostly :D

app/views/dashboards/_actions.html.erb
Ridhwana: ✅

spec/requests/dashboard_spec.rb
Ridhwana: ✅

spec/requests/dashboard_spec.rb
Ridhwana: ✅

app/policies/video_policy.rb
```diff
@@ -1,9 +1,15 @@
 class VideoPolicy < ApplicationPolicy
-  def new?
+  def create?
```
rhymes, line 2: Doesn't this change impact this controller action?

https://github.com/forem/forem/blob/2075a021fa15d3836943bbd281c819f5d4cbb393/app/controllers/videos_controller.rb#L6

Ridhwana: I'm not actually sure - I'll have to check. Thanks for picking that up, for now I'm going to push a method that encompasses 'enabled?' and leaves 'new?' as is. It will be more verbose which seems okay. Will re request a review in a bit.

rhymes (approved):
LGTM!

PR #10957, opened 2020-10-20
Add admin index

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently `/admin` redirects straight into `/admin/articles` which may or may not actually be the right place to go.

This PR introduces a simple index which allows users to pick from the right place to go.

This is an MVP, but IMO this dashboard should ultimately have descriptions, at-a-glance analytics, and generally be the most useful initial interface. For now I adopted the card view from tags and I think it's a decent start.

Screen Shot 2020-10-20 at 9 48 10 AM

### DIFF
--- a/app/controllers/admin/admin_portals_controller.rb
+++ b/app/controllers/admin/admin_portals_controller.rb
@@ -1,6 +1,6 @@
 module Admin
   class AdminPortalsController < Admin::ApplicationController
     layout "admin"
-    def index;end
+    def index; end
   end
 end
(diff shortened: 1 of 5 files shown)


ludwiczakpawel (approved):
Sweet!

pkfrank (approved):
Looks great, can't wait to see this evolve over time

PR #10958, opened 2020-10-20
Fix Basic Feed Issues

Zhao-Andy:
- [x] Bug Fix

## Description
This (hopefully) fixes issues new and small Forems are having with their regular "Feed" and timeframe feeds, as well as the feeds in tag index pages.

To summarize, some Forems were having issues where:
1. on the home page, they only see posts in `Latest` and no posts in `Feed`
2. on tag index pages, they see no posts at all.
3. on tag index pages, older posts in the `Feed` appear before some newer posts
4. also on tag index pages, the number of articles published was incorrect

To explain what I think is causing the problem:
1. In smaller Forems, the `score` for each article is significantly lower than on DEV. Articles on DEV get more traction sooner because we simply have more people engaging with them. Therefore, on smaller Forems **older** posts with **no engagement** (reactions, comments, etc) won't appear in the `Feed`, but will show up on `Latest` (an index less dependent on score).
2. When no posts appear on the tag index page, it's probably because the `scores` of the posts are too low, similar to point `#1`.
3. So IMO this is the algorithm working -- we shouldn't see a post that was just published higher than an older post that had a lot of engagement. This might just need some clarifying to Forem admins.
4. This number is cached for 2 hours, so newer posts don't clear that cache until at least two hours later.

I'll outline my proposed solutions in as PR review comments.

So this is hard to test locally, other than checking that my feed appears pretty much the same and that no logic is broken. I think code review will be really important here.

## Added tests?
- [x] no, because I need help

## Added to documentation?
- [x] no documentation needed

Yes, Forems may want to set these scores to what they were before or a different number. This change won't cause a major if this is not followed through on.

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -139,7 +139,7 @@ class StoriesController < ApplicationController
                                 cached_tagged_count
                               end
     @number_of_articles = user_signed_in? ? 5 : SIGNED_OUT_RECORD_COUNT
-    @stories = Articles::Feeds.const_get(SiteConfig.feed_strategy.classify)
+    @stories = Articles::Feeds::LargeForemExperimental
       .new(number_of_articles: @number_of_articles, tag: @tag, page: @page)
       .published_articles_by_tag
 
@@ -185,7 +185,7 @@ class StoriesController < ApplicationController
   end
 
   def featured_story
-    @featured_story ||= Articles::Feeds::LargeForemExperimental.find_featured_story(@stories) # TODO: maybe change this
+    @featured_story ||= Articles::Feeds::LargeForemExperimental.find_featured_story(@stories)
   end
 
   def handle_podcast_index
@@ -259,7 +259,7 @@ class StoriesController < ApplicationController
   end
 
   def assign_feed_stories
-    feed = Articles::Feeds.const_get(SiteConfig.feed_strategy.classify).new(page: @page, tag: params[:tag])
+    feed = Articles::Feeds::LargeForemExperimental.new(page: @page, tag: params[:tag])
     if params[:timeframe].in?(Timeframer::FILTER_TIMEFRAMES)
       @stories = feed.top_articles_by_timeframe(timeframe: params[:timeframe])
     elsif params[:timeframe] == Timeframer::LATEST_TIMEFRAME
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -337,7 +337,7 @@ class StoriesController < ApplicationController
     elsif params[:timeframe] == "latest"
       @stories.where("score > ?", -20).order(published_at: :desc)
     else
-      @stories.order(hotness_score: :desc).where("score > 2") # Andy: need to discuss this
+      @stories.order(hotness_score: :desc).where("score > #{SiteConfig.home_feed_minimum_score}")
     end
   end
 
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -136,7 +136,7 @@ class StoriesController < ApplicationController
                                 cached_tagged_count
                               end
     @number_of_articles = user_signed_in? ? 5 : SIGNED_OUT_RECORD_COUNT
-    @stories = Articles::Feeds::LargeForemExperimental
+    @stories = Articles::Feeds.const_get(SiteConfig.feed_strategy.classify)
       .new(number_of_articles: @number_of_articles, tag: @tag, page: @page)
       .published_articles_by_tag
 
@@ -182,7 +182,7 @@ class StoriesController < ApplicationController
   end
 
   def featured_story
-    @featured_story ||= Articles::Feeds::LargeForemExperimental.find_featured_story(@stories)
+    @featured_story ||= Articles::Feeds::LargeForemExperimental.find_featured_story(@stories) # TODO: maybe change this
   end
 
   def handle_podcast_index
@@ -256,7 +256,7 @@ class StoriesController < ApplicationController
   end
 
   def assign_feed_stories
-    feed = Articles::Feeds::LargeForemExperimental.new(page: @page, tag: params[:tag])
+    feed = Articles::Feeds.const_get(SiteConfig.feed_strategy.classify).new(page: @page, tag: params[:tag])
     if params[:timeframe].in?(Timeframer::FILTER_TIMEFRAMES)
       @stories = feed.top_articles_by_timeframe(timeframe: params[:timeframe])
     elsif params[:timeframe] == Timeframer::LATEST_TIMEFRAME
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -134,7 +134,7 @@ class StoriesController < ApplicationController
                                 Article.published.cached_tagged_by_approval_with(@tag).size
                               elsif SiteConfig.feed_strategy == "basic"
                                 Article.published.cached_tagged_with(@tag)
-                                  .where("score > #{SiteConfig.tag_feed_minimum_score}").size
+                                  .where("score > ?", SiteConfig.tag_feed_minimum_score).size
                               else
                                 cached_tagged_count
                               end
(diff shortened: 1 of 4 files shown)


app/models/site_config.rb
```diff
@@ -53,6 +53,8 @@ class SiteConfig < RailsSettings::Base
   field :staff_user_id, type: :integer, default: 1
   field :experience_low, type: :string, default: "Total Newbies"
   field :experience_high, type: :string, default: "Experienced Users"
+  field :tag_feed_minimum_score, type: :integer, default: 1
+  field :minimum_hotness_score, type: :integer, default: 7
```
Zhao-Andy, line 57: I think the best, possibly long term but definitely short term, solution is to make this hotness score configurable. I am _pretty sure_ that the issue in smaller Forems is that `7` is too high, and it needs to be adjusted to something lower for now.

Therefore, smaller Forems will need to set this in console to something lower:
```
SiteConfig.minimum_hotness_score = 5 # or 3?
```

app/models/site_config.rb
```diff
@@ -53,6 +53,8 @@ class SiteConfig < RailsSettings::Base
   field :staff_user_id, type: :integer, default: 1
   field :experience_low, type: :string, default: "Total Newbies"
   field :experience_high, type: :string, default: "Experienced Users"
+  field :tag_feed_minimum_score, type: :integer, default: 1
```
Zhao-Andy, line 56: If we want, DEV will need to set this in console original `2`:
```
SiteConfig.tag_feed_minimum_score = 2
```

app/controllers/stories_controller.rb
```diff
@@ -334,7 +337,7 @@ def stories_by_timeframe
     elsif params[:timeframe] == "latest"
       @stories.where("score > ?", -20).order(published_at: :desc)
     else
-      @stories.order(hotness_score: :desc).where("score > 2")
+      @stories.order(hotness_score: :desc).where("score > 2") # Andy: need to discuss this
```
Zhao-Andy, line 340: So here is another place where we have a score of `2` that we check for. I'm not sure if this is a place that we need to make configurable too, but wanted to highlight it in case we do.

app/controllers/stories_controller.rb
```diff
@@ -132,6 +132,9 @@ def handle_tag_index
 
     @num_published_articles = if @tag_model.requires_approval?
                                 Article.published.cached_tagged_by_approval_with(@tag).size
+                              elsif SiteConfig.feed_strategy == "basic"
+                                Article.published.cached_tagged_with(@tag)
+                                  .where("score > #{SiteConfig.tag_feed_minimum_score}").size
```
Zhao-Andy, line 137: Adding this in to not use the cached value **_at all_** and instead ping the database every time if the feed strategy is basic. I think it's okay for smaller Forems, but maybe we still want to use a cached value and instead change it from 2 hours to something shorter.

app/models/site_config.rb
```diff
@@ -53,6 +53,8 @@ class SiteConfig < RailsSettings::Base
   field :staff_user_id, type: :integer, default: 1
   field :experience_low, type: :string, default: "Total Newbies"
   field :experience_high, type: :string, default: "Experienced Users"
+  field :tag_feed_minimum_score, type: :integer, default: 0
+  field :home_feed_minimum_score, type: :integer, default: 0
```
Zhao-Andy, line 57: DEV can / will optionally change these values to their original values:
```
SiteConfig.tag_feed_minimum_score = 2
SiteConfig.home_feed_minimum_score = 2
```

Zhao-Andy:
I'm approving the CodeClimate issues of possible SQL injection:
Screen Shot 2020-10-20 at 11 37 46 AM

rhymes:
@Zhao-Andy what about using the `?` parameter, instead of interpolation?

Zhao-Andy:
Ah great point @rhymes ! Will update.

benhalpern:
Looks like this is in a pretty good place. I'd include @rhymes change and then merge if the tests are looking good to ensure tag pages especially start being functional on small forems.

app/controllers/stories_controller.rb
Zhao-Andy, line 137:
```suggestion
                                  .where("score > ?", SiteConfig.tag_feed_minimum_score).size
```

app/controllers/stories_controller.rb
```diff
@@ -334,7 +337,7 @@ def stories_by_timeframe
     elsif params[:timeframe] == "latest"
       @stories.where("score > ?", -20).order(published_at: :desc)
     else
-      @stories.order(hotness_score: :desc).where("score > 2")
+      @stories.order(hotness_score: :desc).where("score > #{SiteConfig.home_feed_minimum_score}")
```
Zhao-Andy, line 340:
```suggestion
      @stories.order(hotness_score: :desc).where("score > ?", SiteConfig.home_feed_minimum_score)
```

app/controllers/stories_controller.rb
```diff
@@ -487,7 +490,7 @@ def current_search_results_ordering
 
   def cached_tagged_count
     Rails.cache.fetch("article-cached-tagged-count-#{@tag}", expires_in: 2.hours) do
-      Article.published.cached_tagged_with(@tag).where("score > 2").size
+      Article.published.cached_tagged_with(@tag).where("score > #{SiteConfig.tag_feed_minimum_score}").size
```
Zhao-Andy, line 493:
```suggestion
      Article.published.cached_tagged_with(@tag).where("score > ?", SiteConfig.tag_feed_minimum_score).size
```

benhalpern (approved):

app/controllers/stories_controller.rb
```diff
@@ -487,7 +490,7 @@ def current_search_results_ordering
 
   def cached_tagged_count
     Rails.cache.fetch("article-cached-tagged-count-#{@tag}", expires_in: 2.hours) do
-      Article.published.cached_tagged_with(@tag).where("score > 2").size
+      Article.published.cached_tagged_with(@tag).where("score > ?", SiteConfig.tag_feed_minimum_score).size
```
benhalpern, line 493:
```suggestion
      Article.published.cached_tagged_with(@tag).where("score >= ?", SiteConfig.tag_feed_minimum_score).size
```

app/controllers/stories_controller.rb
```diff
@@ -334,7 +337,7 @@ def stories_by_timeframe
     elsif params[:timeframe] == "latest"
       @stories.where("score > ?", -20).order(published_at: :desc)
     else
-      @stories.order(hotness_score: :desc).where("score > 2")
+      @stories.order(hotness_score: :desc).where("score > ?", SiteConfig.home_feed_minimum_score)
```
benhalpern, line 340:
```suggestion
      @stories.order(hotness_score: :desc).where("score >= ?", SiteConfig.home_feed_minimum_score)
```

app/controllers/stories_controller.rb
```diff
@@ -132,6 +132,9 @@ def handle_tag_index
 
     @num_published_articles = if @tag_model.requires_approval?
                                 Article.published.cached_tagged_by_approval_with(@tag).size
+                              elsif SiteConfig.feed_strategy == "basic"
+                                Article.published.cached_tagged_with(@tag)
+                                  .where("score > ?", SiteConfig.tag_feed_minimum_score).size
```
benhalpern, line 137:
```suggestion
                                  .where("score >= ?", SiteConfig.tag_feed_minimum_score).size
```

benhalpern:
Made a couple small changes to use `>=` instead of `>`.... As Forems that have 1 score as the minimum should be inclusive of 1, and so forth.

juliannatetreault:
Is the best way to test this still to just make sure that the feed doesn't break/looks normal, to make sure that the Travis builds green, and that all other checks pass?

benhalpern:
@juliannatetreault Yeah. There is very little logically going on, so a quick manual QA should be enough to okay this provided the tests still pass.

I think this PR will need a follow up to add the interface functionality, but since we're working to ensure Forem partners are online ASAP I think this PR is in a good place assuming it did not break functionality.

juliannatetreault:
Awesome, I'll QA it now. Thanks for the clarification, @benhalpern!

juliannatetreault (approved):
My feed looked as expected and I didn't experience any of the issues that you laid out in your PR description, so 🚀 . Nice work!!

Zhao-Andy:
I'll merge once I'm back at my computer and ready to check the deploy. Feel free to merge for me and watch though if anyone would like!

PR #10962, opened 2020-10-20
Make 404 page generic

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix ?
- [ ] Optimization
- [ ] Documentation Update

## Description

This change makes the 404 page as generic as possible.

I think we should follow this up with a PR to make 404 pages dynamic (with Forem specifc copy) or even cuztomizable via `/admin/pages`. There are some technical considerations to that, but this is currently being prioritized as an urgent fix for some upcoming Forem launches, so I think we should start with this change.

Closes https://github.com/forem/InternalProjectPlanning/issues/178

After:
image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/public/404.html
+++ b/public/404.html
@@ -91,10 +91,9 @@
   <div class="not-found-page base-background-color">
     <div class="tv__outer">
         <div class="tv__middle">
-          <img class="tv__inner--special" src="https://i.imgur.com/Vyyjycv.gif" alt="404 not found image as TV static">
-          <img class="tv__inner" src="https://i.imgur.com/3v0LCiP.jpg" alt="404 not found DEV logo">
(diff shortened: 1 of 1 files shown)


pkfrank (approved):
Looks great, and agreed on the future direction (of giving Forem admins more customization options here)

juliannatetreault (approved):
🚀 !!

rhymes (approved):

benhalpern (approved):
This is great. It's a simple solution to fix the state of things now. I think it would be lovely to circle back on this with some new customization to allow communities to make this page really delightful— and I'm glad we didn't get hung up on that right now. 😄

PR #10977, opened 2020-10-21
Remove social icons from user liquid tag

benhalpern:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We had a quick conversation and realized the future of this feature's design does not need these icons. We were having some consistency issues, so this is the expedient solution.


jdoss (approved):

pkfrank (approved):

pkfrank:
@benhalpern I believe that this solution should be extended to the `comment` liquid tag as well.  Currently, it will try to grab the social icon but fail.  Example from a `forem.dev` comment preview below

example from forem.dev

PR #10982, opened 2020-10-21
Profile forms styling

ludwiczakpawel:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

#10733 introduced a bug causing styling issue on form fields in `/settings/profile`. This PR fixes it.

## Related Tickets & Documents

- #10733 

## QA Instructions, Screenshots, Recordings

**Before:**

image

**After:**

image

## Added tests?

- [ ] yes
- [x] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

PR #10984, opened 2020-10-21
Fix for color fields in settings

ludwiczakpawel:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In #10982 I've introduced styling fix for text fields. This PR takes it step further and fixes color type fields and makes them interactive as in the past. It also brings the `field.description` to the frontend.

- #10982 
- #10733 

**Before:**

image

**After:**

image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/views/users/_profile.html.erb
+++ b/app/views/users/_profile.html.erb
@@ -45,15 +45,21 @@
     <div class="crayons-card mb-6 grid grid-flow-row gap-6 p-6">
       <h2><%= group.name %></h2>
       <% if group.description.present? %>
-        <div class="color-base-40"><%= group.description %></div>
+        <div class="color-base-60"><%= group.description %></div>
       <% end %>
+
       <% group.profile_fields.each do |field| %>
-        <div class="crayons-field <%= "crayons-field--check_box" if field.input_type == "check_box" %>">
-          <div><%= field.label %></div>
+        <div class="crayons-field <%= "crayons-field--checkbox" if field.input_type == "check_box" %>">
           <% if field["input_type"] == "check_box" %>
-            <%= f.public_send(field["input_type"], "profile[#{field.attribute_name}]", checked: @user.profile.data[field.attribute_name]) %>
+            <%= f.public_send(field["input_type"], "profile[#{field.attribute_name}]", checked: @user.profile.data[field.attribute_name], class: "crayons-checkbox") %>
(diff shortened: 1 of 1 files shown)


benhalpern (approved):

app/views/users/_profile.html.erb
```diff
@@ -70,3 +81,23 @@
     <button type="submit" class="crayons-btn">Save Profile Information</button>
   </div>
 <% end %>
+
+<script>
```
rhymes, line 85: it'd be great if these instructions went in their own JS file so that they can be linted and compressed.

BTW I think we should start using event bubbling to have a single event handler on the immediate container instead of having N event handlers for each element but that's a larger conversation for another day and for another PR, outside the scope of this one. cc @nickytonline

benhalpern: We have a file called `initializeSettings` where these could go.

I feel like as a hot fix we can accept this though and then make another change after.

nickytonline: > it'd be great if these instructions went in their own JS file so that they can be linted and compressed.
> 
> 
> 
> BTW I think we should start using event bubbling to have a single event handler on the immediate container instead of having N event handlers for each element but that's a larger conversation for another day and for another PR, outside the scope of this one. cc @nickytonline 

@rhymes, definitely leverage event bubbling.

Link2Twenty:
I guess at some point it would make sense to make a `colorPicker` component so the colour could be just a shared state.

https://codesandbox.io/s/determined-moon-kn2qg

ludwiczakpawel:
@Link2Twenty yes, yes please!!! :)

nickytonline:
> I guess at some point it would make sense to make a `colorPicker` component so the colour could be just a shared state.
> 
> [codesandbox.io/s/determined-moon-kn2qg](https://codesandbox.io/s/determined-moon-kn2qg)

@Link2Twenty @ludwiczakpawel nice! I made one a while back too for a blog post. 

https://codepen.io/nickytonline/pen/ebxrpv

citizen428 (approved):

ludwiczakpawel:
@nickytonline nice, updating CSS variable is pretty smart!

PR #10985, opened 2020-10-21
Fix admin link

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The newly added admin link doesn't 100% work properly because the admin section is running a different "shell" than the user facing part so InstantClick kind of freaks out. If you click again it works.... So it could go unnoticed and thought of as an issue with a slow request, but it's actually broken if we don't skip instant click here.

### DIFF
--- a/app/views/layouts/_nav_menu.html.erb
+++ b/app/views/layouts/_nav_menu.html.erb
@@ -10,7 +10,7 @@
     </div>
 
     <div class="border-0 border-b-1 border-solid border-base-20 p-1">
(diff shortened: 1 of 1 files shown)


rhymes (approved):

Ridhwana (approved):

leewynne:
Could this be related to the same issue when hitting the primary logo to get back to the home feed?

Issue #172, opened 2018-03-04
Article URL is null


Nektro:
# _BUG_
### Current Behavior
When I try to read the below article in the provided screenshot, the URL is https://dev.to/null

### Expected Behavior
The URL would form correctly and I would be able to read the article

### Additional Info (steps to replicate, device/browser, helpful links)
Windows 10 1709 (Build 17074.1002)
Google Chrome 65.0.3325.106 (Official Build) beta (64-bit)
Incognito does not change the result
The article is from https://dev.to/dangolant

### Screenshots
![image](https://user-images.githubusercontent.com/5464072/36950509-cf68b990-1fc4-11e8-90a7-35487ccb11d8.png)

Zhao-Andy:
Thanks for the heads up. We'll fix this soon.

shobhitpuri:
+1. For any of the blogs on my profile (https://dev.to/shobhitpuri) as well, they are pointing to https://dev.to/null.

Zhao-Andy:
@maestromac fixed this :) Let us know if you see any others in case we missed them.

Nektro:
Looks good to me, I can see the article I mentioned and the ones from @shobhitpuri's profile

PR #10992, opened 2020-10-21
Standardize Image Placeholder Text in SiteConfig

juliannatetreault:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR standardizes the image placeholder text throughout the SiteConfig. Prior to this change, we used both `https://image.url/` and `https://image.url/image.png` -- this PR makes it so that we only use `https://image.url/image.png` as an image's placeholder text.

Closes [issue #172](https://github.com/forem/InternalProjectPlanning).

**To QA this change, please ensure that all of the checks pass and that all the SiteConfig only uses `https://image.url/image.png` as its image placeholder text and not `https://image.url/`.**

**Before**:
Screen Shot 2020-10-21 at 9 22 05 AM

**After**:
Screen Shot 2020-10-21 at 9 19 39 AM

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Nope!

Mindy Kaling High Standards

### DIFF
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -34,7 +34,7 @@ module Constants
         placeholder: "Campaign sidebar enabled or not"
       },
       campaign_sidebar_image: {
-        description: "https://image.url/image.png",
+        description: "https://url/image.png",
         placeholder: "Used at the top of the campaign sidebar"
       },
       campaign_url: {
@@ -107,7 +107,7 @@ module Constants
       },
       favicon_url: {
         description: "Used as the site favicon",
-        placeholder: "https://image.url/image.png"
+        placeholder: "https://url/image.png"
       },
       feed_strategy: {
         description: "Determines the main feed algorithm approach the app takes: basic or large_forem_experimental
@@ -156,7 +156,7 @@ module Constants
       },
       logo_png: {
         description: "Minimum 1024px, used for PWA etc.",
-        placeholder: "https://image.url/image.png"
+        placeholder: "https://url/image.png"
       },
       logo_svg: {
         description: "Used as the SVG logo of the community",
@@ -164,7 +164,7 @@ module Constants
       },
       main_social_image: {
         description: "Used as the main image in social networks and OpenGraph",
-        placeholder: "https://image.url/image.png"
+        placeholder: "https://url/image.png"
       },
       mailchimp_api_key: {
         description: "API key used to connect Mailchimp account. Found in Mailchimp backend",
@@ -188,7 +188,7 @@ module Constants
       },
       mascot_footer_image_url: {
         description: "Special cute mascot image used in the footer.",
-        placeholder: "https://image.url/image.png"
+        placeholder: "https://url/image.png"
       },
       mascot_footer_image_width: {
         description: "The footer mascot width will resized to this value, defaults to 52",
@@ -204,7 +204,7 @@ module Constants
       },
       mascot_image_url: {
         description: "Used as the mascot image.",
-        placeholder: "https://image.url/image.png"
+        placeholder: "https://url/image.png"
       },
       mascot_user_id: {
         description: "User ID of the Mascot account",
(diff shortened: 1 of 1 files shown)


app/lib/constants/site_config.rb
```diff
@@ -34,7 +34,7 @@ module SiteConfig
         placeholder: "Campaign sidebar enabled or not"
       },
       campaign_sidebar_image: {
-        description: "https://image.url",
+        description: "https://image.url/image.png",
```
Ridhwana, line 37: The way `https://image.url/image.png` reads feels strange to me. Should we make this https://url/image.png or https://image_url/image.png? This is completely personal preference with no rationale so feel free to ignore if you or others don't feel the same way..

Ridhwana (approved):

app/lib/constants/site_config.rb
vaidehijoshi, line 37: Rather than make these changes everywhere in the SiteConfig, we could assign `"https://image.url/image.png"` to a constant that we `freeze` in this file.

That way, if we decide to move from `image.png` to `image.jpg` down the road (for example), instead of changing every single one of these instances yet again, you need only change the constant. It also ensures that a contributor down the road can easily use the same constant again (instead of manually typing it out and then potentially making a typo or even accidentally specifying the wrong image type).

vaidehijoshi: @Ridhwana I agree with you, tbqh. When I read this, I'm not sure what `image.url/image.png` actually _means_ in terms for formatting 🤷

juliannatetreault: Thanks for the feedback, @Ridhwana! I've made your suggested change in this PR. 😊

juliannatetreault: I appreciate the suggestion! 

I attempted to add an `IMAGE_PLACEHOLDER` constant, but was unable to get it to work. After trying a couple of different approaches, I dug into it a little more and realized that the file itself is a giant constant, which was causing the constant that I tried to add, to not render properly in the SiteConfig. I'm guessing that this is also why we haven't assigned any of the placeholder text in this file to a constant, but 🤷‍♀️  @vaidehijoshi

Zhao-Andy (approved):

app/lib/constants/site_config.rb
vaidehijoshi: @juliannatetreault The file _itself_ isn't a constant, it's a module. However, there is a hash inside of this module that encapsulates the entire `DETAILS` of the SiteConfig (see [this line](https://github.com/forem/forem/blob/master/app/lib/constants/site_config.rb#L3)). I suspect you are probably assigning the object in the wrong place/incorrectly.

I created a `TEST` constant locally above the `DETAILS` object:
```ruby
TEST = "i love llamas".freeze
DETAILS = {
  # ...
}
```

And then used it inside of SiteConfig:
```ruby
campaign_sidebar_image: {
  description: TEST,
  placeholder: "Used at the top of the campaign sidebar"
},
```

and it does work:
Screen Shot 2020-10-21 at 9 32 55 AM

juliannatetreault: Oh, my apologiess -- that works! Why did you not put it in an array, though?

vaidehijoshi: @juliannatetreault If it's just a single string that we're going to reuse, it doesn't need to be in an array, right?  Constants can be anything -- strings, numbers, arrays, hashes/objects! In this case, we're using the constant to represent a single piece of placeholder text, so it doesn't need to be in array I think.

Just so I can understand your thinking and help clarify here, let me ask a question in reply to your question 😄 -- why did you think it should be in an array?

juliannatetreault: Interesting! I thought it needed to be in an array because I've only ever seen constants in an array... the more you know. 😅

vaidehijoshi: @juliannatetreault Heh, nope! In fact, the `DETAILS` constant isn't even an array -- it's a hash 😉 

Here's [a great resource](https://www.rubyguides.com/2017/07/ruby-constants/) on Ruby constants :)

app/lib/constants/site_config.rb
```diff
@@ -1,5 +1,7 @@
 module Constants
   module SiteConfig
+    IMAGE_PLACEHOLDER = "https://url/image.png".freeze
```
vaidehijoshi, line 3: I love this change! 😻 

Also love that this now sets the precedent for us to abstract _more_ of this file into constants, too!

vaidehijoshi (approved):

Issue #10996, opened 2020-10-21
Optimize the feed reader for websites with a large number of users


rhymes:
**Current situation**

Currently Forem's functionality to import feeds takes too long, especially for DEV which has 3420 feeds to go through.

Currently the import is sequential: the feed is downloaded, then parsed, then articles are built in memory and saved in the DB

Unfortunately, at least in the case of DEV, we don't really have solid metrics on what is actually slow in production, as it only tracks errors but nothing else. We should consider adding instrumentation before replacing it one way or another

**Variables to consider and things learned in benchmarking**:

- network latency is not a constant (fetching thousands of feeds can have different performance results depending on network conditions of the upstream servers)
- different feeds can have different lenghts thus users have a variable amount of articles to process at each run 
- currently we skip a random number of feeds at each run because feed fetching can be slow
- Nokogiri parsing occupies a lot of memory (there are literally millions of objects allocated by the nokogiri gem)

**Optimization ideas**

There are two main things we can optimize (my opinion is that we should find a combination of both that suits us):

1. make the actual fetching of feeds and parsing of them faster (it's all I/O, there's no reason for it to be sequential)
1. process multiple users in parallel (basically by doing things more or less sequentially but splitting the workload in separate workers, one per each user)

This why I think we should employ a combination of both:

- we process feeds sequentially, but downloading bytes from the web is inherently parallelizable, so we can download a bunch (in batches obviously) from the web and then start processing those
- we parse feeds sequentally, but those also can be parallelized
- both the above steps have an upper ceiling not just based on how many cores the machines will run on but also due to memory consumption (the most memory resource hungry of the two operations is parsing for obvious reasons)
- writing articles on the DB can be parallelized but that doesn't really need to be per user (we'd have 3420 jobs in the queue that could still be individually fast or slow)
- we can parallelize articles creation in batches by changing the logic a little bit: right now it's the single "future article" that's responsible to know if they actually exist. Each of them has a conditional check to see if it already exists. We could do this in one swoop for the entire batch and then remove from the batch of workers those that don't need to be processed at all
- we need to be careful at how many workers we add concurrently that write to the `articles` table as we could end up using too many ActiveRecord connections and exhaust the pools

**Plan of action**

The first step is to write a POC which parallelizes network fetching and parsing, this is part of a multi step plan (not necessarily in this order):

This PR is one step in a multi step plan (not necessarily in this order) which comprises:

- [ ] add monitoring to the existing `RssReader` to undestand what it's profile in production
[...]

PR #10998, opened 2020-10-21
Add Feeds::Import service class

rhymes:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR is one step in a multi step plan which aims to find the best strategy to ingest XML (RSS, ATOM and other) feeds, parse them, transform them into articles to facilitate cross posting. 

Please read https://github.com/forem/forem/issues/10996 for the details and the discussion

## TODO

- [ ] add error handling
- [ ] add metrics if possible so that we can keep track on how each segment takes
- [ ] inline todos might be tackled in this or separate PRs

1. delete all articles if possible
1. run `rails measure_feeds:setup` which adds a feed URL to each user you have locally
1. run `rails fetch_feeds_import`
1. check articles were imported

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed


app/services/feeds/import.rb
```diff
@@ -0,0 +1,152 @@
...
+      end
+
+      batch_of_users.update_all(feed_fetched_at: Time.current)
+
+      Hash[result.compact]
+    end
+
+    # TODO: put this in separate service object
+    def parse_feeds(feeds_per_user_id)
+      result = Parallel.map(feeds_per_user_id, in_threads: num_parsers) do |user_id, feed_xml|
```
mstruve, line 81: Let's keep a close eye on this, I don't want a process getting too large since we are already throwing multiple threads inside an already threaded process.

app/services/feeds/import.rb
```diff
@@ -0,0 +1,152 @@
...
+        [user_id, parsed_feed]
+      rescue Feedjira::NoParserAvailable
+        # TODO: add exception handling
+        nil
+      end
+
+      Hash[result.compact]
+    end
+
+    # TODO: currently this is exactly as in RSSReader, but we might find
```
mstruve, line 93: What if we made use of one of the [Sidekiq workflow plugins](https://github.com/mperham/sidekiq/wiki/Related-Projects#execution-ordering)? Here are a couple that I thought looked promising but probably worth checking out the rest of the list. 
[Sidekiq Superworker](https://github.com/socialpandas/sidekiq-superworker) (this one is not well maintained so maybe not 😬 )
or
[Gush](https://github.com/chaps-io/gush)

In the past, I have used Sidekiq Enterprise which came with the feature built-in. It was extremely handy and useful for complex workflows. I think using one of the OSS plugins might be a possible solution here.

mstruve (commented):
Couple of thoughts :) So pumped that you are working on this!!!!

lib/tasks/fetch.rake
```diff
@@ -6,6 +6,18 @@ task fetch_all_rss: :environment do
   RssReader.get_all_articles(force: false) # don't force fetch. Fetch "random" subset instead of all of them.
 end
 
+task fetch_rss_reader: :environment do
+  Rails.application.eager_load!
```
citizen428, line 10: Is it necessary to eager load all of Rails here or could you get away with `require_dependency`?

app/services/feeds/import.rb
rhymes: Yes, this is exactly the kind of thing we need to monitor :)

app/services/feeds/import.rb
rhymes: We can look into it, need to figure out if they are over-engineered or not. I'll write down some other notes in the issue https://github.com/forem/forem/issues/10996 that came up while talking with Ben on how we could optimize writing to the DB

app/services/feeds/import.rb
```diff
@@ -0,0 +1,152 @@
+# TODO: [rhymes]
+# => add Feeds::ImportUser to fetch a single user
+# => add Feeds::ValidateFeedUrl to validate a single feed URL
+module Feeds
+  class Import
+    def self.call(silence: true)
```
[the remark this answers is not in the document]
rhymes: Sure! That's definitely going to disappear before we merge though :D

app/services/feeds/import.rb
```diff
@@ -0,0 +1,152 @@
...
+      users.in_batches(of: users_batch_size) do |batch_of_users|
+        feeds_per_user_id = fetch_feeds(batch_of_users)
+        Rails.logger.info("feeds_per_user_id.length: #{feeds_per_user_id.length}")
+
+        feedjira_objects = parse_feeds(feeds_per_user_id)
+        Rails.logger.info("feedjira_objects.length: #{feedjira_objects.length}")
+
+        # NOTE: doing this sequentially to avoid locking problems with the DB
+        # and unnecessary conflicts
+        articles = feedjira_objects.map do |user_id, feed|
```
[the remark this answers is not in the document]
rhymes: Will do!

app/services/feeds/import.rb
```diff
@@ -0,0 +1,152 @@
...
+    def fetch_feeds(batch_of_users)
+      data = batch_of_users.pluck(:id, :feed_url)
+
+      result = Parallel.map(data, in_threads: num_fetchers) do |user_id, url|
+        response = HTTParty.get(url.strip, timeout: 10)
+
+        [user_id, response.body]
+      rescue StandardError
+        # TODO: add exception handling
+        # For example, we should stop pulling feeds that return 404 and disable them?
```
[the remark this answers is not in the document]
rhymes: Yeah, good point. Another thing Ben suggested could be to stop fetching articles from feeds if the user hasn't published in a while (we keep pulling in drafts even for non active users, I'll add a comment in the issue so it doesn't get lost here)

lib/tasks/fetch.rake
rhymes: Good question, don't know. I just copy and pasted the previous one. I'll try and let you know :)

app/services/feeds/import.rb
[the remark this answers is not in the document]
citizen428: Yes, that sounds like a sensible approach. I think the 404 is still valuable, we could e.g. count how many times they happen in a row and update a counter which we reset to 0 on successful import. X consecutive 404s on the feed and we automatically pause and send the user a notification to check their settings.

rhymes:
Setting this ready for review as I think we should merge it and then test it manually on dev using `Articles::DevFeedsImportWorker.perform_async`

lib/tasks/temporary/measure_feeds.rake
```diff
@@ -0,0 +1,96 @@
...
+      end
+
+      filepath = "/tmp/measure_feeds_#{importer}_#{num_users}_#{Time.current.iso8601}.txt"
+      puts "Saving memory profile to #{filepath}..."
+      puts memory_report.pretty_print(to_file: filepath)
+    end
+
+    def setup_feeds(num_users: nil)
+      # we start from no articles in the DB
+      Article.delete_all
```
mstruve, line 41: Can we remove this from this PR and keep it in a gist or something since it's temporary? I dont want code like this going out to production in the event for some reason it is accidentally triggered.

app/services/feeds/import.rb
```diff
@@ -0,0 +1,192 @@
...
+      @num_fetchers = 8
+      @num_parsers = 4
+    end
+
+    def call
+      total_articles_count = 0
+
+      users.in_batches(of: users_batch_size) do |batch_of_users|
+        feeds_per_user_id = fetch_feeds(batch_of_users)
+        Rails.logger.info("feeds_per_user_id.length: #{feeds_per_user_id.length}")
```
mstruve, line 33: Is this data we will want to gather from production? If so, our log level is ERROR so we will have to make these error logs if we want to see them, which I am fine with for a few test runs.

app/services/feeds/import.rb
```diff
@@ -0,0 +1,192 @@
...
+
+    def article_exists?(user, item)
+      title = item.title.strip.gsub('"', '\"')
+      feed_source_url = item.url.strip.split("?source=")[0]
+      relation = user.articles
+      relation.where(title: title).or(relation.where(feed_source_url: feed_source_url)).exists?
+    end
+
+    def report_error(error, metadata)
+      Honeybadger.context(metadata)
```
mstruve, line 180: We get a lot of Honeybadgers right now from the RSS reader but its relatively quiet bc they are all repeats now instead of new. This is likely going to be super noisy, what if we logged these instead so we dont sound any false alarms in HB?

mstruve (commented):
I good testing this in production! Made a couple of suggestions regarding logging and error messaging and the temporary rake task. Let me know what you think!

lib/tasks/temporary/measure_feeds.rake
rhymes: Makes sense! I'll move the script :)

app/services/feeds/import.rb
rhymes: Nope, these should go, forgot to remove them

app/services/feeds/import.rb
rhymes: As I can't see the logs in production I'm not sure what would be the best decision here. Let's keep in mind that this is mostly to be run 2 o 3 times to have a general idea of speed and memory occupation. I don't consider it the final version for all the conversations we had in the issue thread

rhymes: In a perfect situation we should only log errors that are actual errors. Like a feed being removed upstream shouldn't be an error, we should just stop fetching it eventually. Also: timeout errors are not super useful in Honeybadger as they might depend on the network conditions. In a sense, I need to know which errors can occurr in production, to know which ones to filter out :)

app/workers/articles/dev_feeds_import_worker.rb
```diff
@@ -0,0 +1,17 @@
...
+# This script will soon be removed. We need it to collect monitoring data on
+# Datadog about the production behavior of the `Feeds::Import` class
+module Articles
+  class DevFeedsImportWorker
+    include Sidekiq::Worker
+
+    sidekiq_options queue: :medium_priority, retry: 10
+
+    def perform
+      return unless SiteConfig.community_name == "DEV"
```
maestromac, line 11: nit: We have a handy `SiteConfig.dev.to?` that would it easier to grep in the future.

app/workers/articles/dev_feeds_import_worker.rb
```diff
@@ -0,0 +1,17 @@
...
+      return if Rails.cache.read("cancel_feeds_import").present?
```
maestromac, line 12: I may have missed it but where does this cache get written?

app/workers/articles/dev_feeds_import_worker.rb
rhymes: ahha right! this worker is temporary though :D

app/workers/articles/dev_feeds_import_worker.rb
rhymes: It's just an escape plan @mstruve thought about. Basically if the worker restarts and we don't want it to, by writing that key manually, it will stop

maestromac (approved):
This is existing!.LGTM :rocket: :rocket:

app/services/feeds/import.rb
mstruve: > As I can't see the logs in production I'm not sure what would be the best decision here. 

You should have access to all the [error logs in Datadog](https://app.datadoghq.com/logs?cols=host%2Cservice%2C%40syslog.procid&from_ts=1603992686819&index=&live=true&messageDisplay=inline&stream_sort=desc&to_ts=1603993586819) which is why I am suggesting using them. Let me know if you don't and I will investigate what the problem might be. 

> Let's keep in mind that this is mostly to be run 2 o 3 times to have a general idea of speed and memory occupation. 
Right, those few times will set off a lot of new HB which will be noisy

> In a perfect situation, we should only log errors that are actual errors.
100% agree, but if we log info statements and we want to see them then I have to turn on INFO logging which will create a bunch of logs we dont need. I am merely suggesting using the error logs as a temporary solution here to give us the insight we need into the process while also not interfering with our alerting system and causing alert fatigue.

rhymes: Ah got it, thanks. For some reason I still thought logs were only going to Heroku's drain. I'll switch them to error logs instead of HB reports

citizen428 (approved):
Based on the comments there are a few minor things you still may want to address, but I'll approve this for now so we can go ahed and collect more data soon. 🚀

mstruve (approved):
Sorry this took a bit for me to get to, I wanted to give the logic a good comb-over next to the original to make sure everything is in place. Lets see what we go!

PR #11007, opened 2020-10-21
Optimization:Remove 2 Feed Experiments

mstruve:
- [x] Optimization

## Description
We have been running these feed experiments for a while now and recently they came up in a PR. I took the initiative to start closing the loop on them. The reason I want to close the loop on this is that multiple code paths add complexity to our codebase and additional time to our specs when we test them. If we can narrow down to the ideal paths(doesn't even have to be 1!) then we can begin to optimize those paths for performance and other Forems. 

I recently looked [at the results of our A/B tests](https://dev.to/abtests) and found that because we have so many experiments running it's pretty hard to tell if any are the "true winner". In this case, I took a couple of experiments that seemed to have lower or very middle of the pack results and removed them. After allowing these tests to run with the new configurations we can check the results again in a week or two and continue to narrow down the scope of the experiment. 

https://github.com/orgs/forem/projects/6#card-47841833

## Added tests?

alt_text

### DIFF
--- a/app/services/articles/feeds/large_forem_experimental.rb
+++ b/app/services/articles/feeds/large_forem_experimental.rb
@@ -60,13 +60,6 @@ module Articles
         stories
       end
 
-      # Test variation: More random
-      def default_home_feed_with_more_randomness_experiment
-        @randomness = 7
-        _featured_story, stories = default_home_feed_and_featured_story(user_signed_in: true)
-        stories
-      end
-
       # Test variation: tags make bigger impact
       def more_tag_weight_experiment
         @tag_weight = 2
@@ -79,15 +72,6 @@ module Articles
         @randomness = 7
         _featured_story, stories = default_home_feed_and_featured_story(user_signed_in: true)
         stories
-      end
-
-      # Test variation: Base half the time, more random other half. Varies on impressions.
-      def mix_default_and_more_random_experiment
-        if rand(2) == 1
-          default_home_feed(user_signed_in: true)
-        else
-          default_home_feed_with_more_randomness_experiment
-        end
       end
 
       # Test variation: the more comments a post has, the higher it's rated!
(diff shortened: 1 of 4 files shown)


maestromac (approved):

app/services/articles/feeds/large_forem_experimental.rb
```diff
@@ -118,30 +102,26 @@ def more_experience_level_weight_experiment
...
           more_experience_level_weight_randomized_at_end_experiment
-        when 9
+        when 7
           more_comments_randomized_at_end_experiment
-        when 10
+        when 8
           more_comments_medium_weight_randomized_at_end_experiment
-        when 11
+        when 9
           more_comments_minimal_weight_randomized_at_end_experiment
```
vaidehijoshi, line 125: Hm, just seeing this again makes me 😩 I wonder how we can improve this.

vaidehijoshi (approved):
Thanks for this!!

app/services/articles/feeds/large_forem_experimental.rb
mstruve: You and me both which is why I decided to take ownership of it. We will get there!

citizen428: How about this?

Wrap `default_home_feed` in an `_experiment` method:

```ruby
def default_home_feed_experiment
  default_home_feed(user_signed_in: true)
end
```

Add this method to the `VARIANTS` constant and this becomes

```ruby
variant_method = VARIANTS.values.sample
public_send(variant_method)
```

If we don't want to add the wrapper, we can at least reduce it this to:

```ruby
if rand(10).zero?
  default_home_feed(user_signed_in: true)
else
  variant_method = VARIANTS.values.sample
  public_send(variant_method)
end
```

I prefer the first option with an explicit `default_home_feed_experiment`.

mstruve: @citizen428 my plan is to leave everything as is until we have "final paths" and then figure out the best way to handle those.

rhymes (approved):

PR #11008, opened 2020-10-21
Fix missing registration data change

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We're currently failing to mark folks as `registered_at`, which we use in certain places after the fact, on the email registration path..... This causes miscellaneous issues.

`registered_at` is different than `created_at` because we create the row in the database when we `invite` them, but they are not yet "registered" for the purpose of application functionality until they actually register.

Of course, this similar functionality exists in several places, and we should refactor this, but for now I think this is good enough with fixing the problem in the place and adding a test.

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -22,6 +22,8 @@ class RegistrationsController < Devise::RegistrationsController
     if recaptcha_disabled? || recaptcha_verified?
       build_resource(sign_up_params)
       resource.saw_onboarding = false
+      resource.registered = true
+      resource.registered_at = Time.current
       resource.editor_version = "v2"
       resource.save if resource.email.present?
       yield resource if block_given?
--- a/spec/requests/registrations_spec.rb
+++ b/spec/requests/registrations_spec.rb
@@ -179,6 +179,17 @@ RSpec.describe "Registrations", type: :request do
         expect(User.all.size).to be 1
       end
 
+      it "marks as registerd" do
+        post "/users", params:
+        { user: { name: "test #{rand(10)}",
+                  username: "haha_#{rand(10)}",
+                  email: "yoooo#{rand(100)}@yo.co",
+                  password: "PaSSw0rd_yo000",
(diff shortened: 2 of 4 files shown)


spec/requests/registrations_spec.rb
```diff
@@ -179,6 +179,17 @@ def mock_recaptcha_verification
         expect(User.all.size).to be 1
       end
 
+      it "marks as registerd" do
+        post "/users", params:
```
citizen428, line 183: **suggestion (non-blocking):** Rewrite this a change assertion so we really see that `registered_at` changes from `nil` to something:

```ruby
expect do
  post "/users", params: ...
end.to change { User.last.registered_at }
expect(User.last.registered).to be true
```

citizen428 (approved):

app/controllers/registrations_controller.rb
```diff
@@ -22,6 +22,8 @@ def create
     if recaptcha_disabled? || recaptcha_verified?
       build_resource(sign_up_params)
       resource.saw_onboarding = false
+      resource.registered = true
+      resource.registered_at = Time.current
```
rhymes, line 26: Do we need a backfill script?

benhalpern: This is something that can be updated manually through admin, and I think it might be hard to identify which users would have gone through this path erroneously. It _may_ be technically possible to remedy this, but I think for this particular issue it's best to allow admins to deal with this manually if they recognize an issue.

benhalpern: .........Actually scratch that. I think it does make sense. I realized it's logically easy enough to determine that if someone was not invited, they should by definition be registered. So I added a script. 👍

mstruve (commented):
DUS needs to be updated in order to work properly otherwise this looks good!

juliannatetreault (approved):
👍 👍

PR #11022, opened 2020-10-22
Add Explore Tags for Signed-In Users

juliannatetreault:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR adds an "Explore Tags" section to the sidebar nav for signed-in users who _do not_ follow any tags. Rather than linking to `/dashboard/tags` like the `cog.svg` does for signed-in users who _follow_ tags, the `cog.svg` links to the tags overview, `/tags`. Additionally, below the "Explore Tags" header, the top five tags for the site are displayed with the option to follow them, making it easy for users to begin following loved tags. This PR allows adds a test to ensure that "Explore Tags" works properly.

Closes https://github.com/forem/InternalProjectPlanning/issues/199

**To QA this PR, please do the following:**

- Navigate to `localhost:3000` as a signed-in user. If you are following tags you should see the header, "My Tags" displayed with the tags that you're following below the header. Additionally, if you click on the cog to the right of the header, you should be brought to the `/dashboard/tags` page:
Screen Shot 2020-10-22 at 11 01 43 AM
Screen Shot 2020-10-22 at 11 02 05 AM

- Navigate to `localhost:3000` as a signed-in user. If you are _not_ following any tags, you should see the header "Explore" displayed with the top 5 tags for the site below the header. Upon hover, you should be shown a "Follow" button, giving you the option to follow the tags right from the navbar. Additionally, if you click on the link to the right of the "Explore" header, "All tags", you should be brought to the `/tags` page where you should see a list of the  "Top Tags" for the site:
Screen Shot 2020-10-23 at 8 18 06 AM
Screen Shot 2020-10-22 at 11 00 54 AM

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Nope!

Astronaut Holding Exploration Sign

### DIFF
--- a/app/views/articles/_sidebar_nav.html.erb
+++ b/app/views/articles/_sidebar_nav.html.erb
@@ -56,9 +56,8 @@
     <div id="sidebar-nav-followed-tags" class="overflow-y-auto mb-2" style="max-height: 42vh;"></div>
     <% elsif user_signed_in? %>
       <header class="p-2 pr-0 flex items-center justify-between">
-        <h3 class="crayons-subtitle-3">Explore Tags</h3>
         <a href="/tags" class="crayons-btn crayons-btn--icon crayons-btn--ghost-dimmed" aria-label="Tags Overview" title="Tags Overview">
-          <%= inline_svg_tag("cog.svg", aria: true, class: "crayons-icon") %>
+          <h3 class="crayons-subtitle-3">Explore Tags</h3>
         </a>
       </header>
       <div id="sidebar-nav-explore-tags" class="overflow-y-auto" style="max-height: 42vh">
--- a/app/views/articles/_sidebar_nav.html.erb
+++ b/app/views/articles/_sidebar_nav.html.erb
@@ -46,7 +46,7 @@
 
 <nav class="mb-6" aria-label="Secondary sidebar nav">
 <nav class="mb-6" aria-label="Secondary sidebar nav">
-  <% if user_signed_in? && current_user.following_tags_count >= 1 %>
+  <% if user_signed_in? && current_user.following_tags_count.positive? %>
     <header class="p-2 pr-0 flex items-center justify-between">
       <h3 class="crayons-subtitle-3">My Tags</h3>
       <a href="/dashboard/following_tags" class="crayons-btn crayons-btn--icon crayons-btn--ghost-dimmed" aria-label="Customize tag priority" title="Customize tag priority">
@@ -55,9 +55,10 @@
     </header>
     <div id="sidebar-nav-followed-tags" class="overflow-y-auto mb-2" style="max-height: 42vh;"></div>
     <% elsif user_signed_in? %>
-      <header class="p-2 pr-0 flex items-center justify-between">
-        <a href="/tags" class="crayons-btn crayons-btn--icon crayons-btn--ghost-dimmed" aria-label="Tags Overview" title="Tags Overview">
-          <h3 class="crayons-subtitle-3">Explore Tags</h3>
+     <header class="p-2 pr-0 flex items-center justify-between">
+       <h3 class="crayons-subtitle-3">Explore</h3>
+       <a href="<%= tags_path %>" class="fs-s crayons-link crayons-link--brand pr-2" title="Browse all tags">
+        All tags
         </a>
       </header>
       <div id="sidebar-nav-explore-tags" class="overflow-y-auto" style="max-height: 42vh">
--- a/app/views/articles/_sidebar_nav.html.erb
+++ b/app/views/articles/_sidebar_nav.html.erb
@@ -53,7 +53,7 @@
         <%= inline_svg_tag("cog.svg", aria: true, class: "crayons-icon") %>
       </a>
     </header>
-    <div id="sidebar-nav-explore-tags" class="overflow-y-auto mb-2" style="max-height: 42vh;"></div>
+    <div id="sidebar-nav-followed-tags" class="overflow-y-auto mb-2" style="max-height: 42vh;"></div>
     <% elsif user_signed_in? %>
       <header class="p-2 pr-0 flex items-center justify-between">
         <h3 class="crayons-subtitle-3">Explore Tags</h3>
--- a/app/views/articles/_sidebar_nav.html.erb
+++ b/app/views/articles/_sidebar_nav.html.erb
@@ -45,28 +45,50 @@
 </nav>
 
 <nav class="mb-6" aria-label="Secondary sidebar nav">
-  <% if user_signed_in? %>
+<nav class="mb-6" aria-label="Secondary sidebar nav">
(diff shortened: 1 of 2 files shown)


app/views/articles/_sidebar_nav.html.erb
```diff
@@ -45,29 +45,51 @@
...
-          <a class="follow-action-button sidebar-nav-link-follow crayons-btn crayons-btn--s"
-            href="#" id="sidebar-nav-link-follow-<%= tag_array.second %>"
-            data-info='{"id":<%= tag_array.first %>,"className":"Tag"}'>
-            Follow
-          </a>
+    <% elsif user_signed_in? %>
+      <header class="p-2 pr-0 flex items-center justify-between">
+        <h3 class="crayons-subtitle-3">Explore Tags</h3>
+        <a href="/tags" class="crayons-btn crayons-btn--icon crayons-btn--ghost-dimmed" aria-label="Tags Overview" title="Tags Overview">
+          <%= inline_svg_tag("cog.svg", aria: true, class: "crayons-icon") %>
```
jacobherrington, line 61: The cog icon strikes me as a bit confusing in this context, but I think a different icon something we can follow up with at a later point, personally.

jacobherrington (approved):
This looks good to me! It might be good to have some design followup behind this change, but I feel like it is in a reasonable place to merge to solve an immediate issue for one of our early adopters.

app/views/articles/_sidebar_nav.html.erb
juliannatetreault: @jacobherrington: what are your thoughts on linking "Explore Tags" to the `/tags` page and just taking the `cog_svg` out entirely? It would look like **this when not hovered over**:
Screen Shot 2020-10-22 at 12 21 52 PM
**And this when hovered over**:
Screen Shot 2020-10-22 at 12 22 15 PM

jacobherrington: I don't have a strong opinion, but this does feel a little more natural to me!

juliannatetreault: Changed @jacobherrington!

app/views/articles/_sidebar_nav.html.erb
```diff
@@ -45,29 +45,50 @@
 </nav>
 
 <nav class="mb-6" aria-label="Secondary sidebar nav">
-  <% if user_signed_in? %>
+<nav class="mb-6" aria-label="Secondary sidebar nav">
+  <% if user_signed_in? && current_user.following_tags_count >= 1 %>
```
rhymes, line 49: non blocking: `.following_tags_count.positive?`

rhymes (approved):
Works great!

app/views/articles/_sidebar_nav.html.erb
```diff
@@ -45,29 +45,50 @@
...
-            #<%= tag_array.second %>
-          </a>
-          <a class="follow-action-button sidebar-nav-link-follow crayons-btn crayons-btn--s"
-            href="#" id="sidebar-nav-link-follow-<%= tag_array.second %>"
-            data-info='{"id":<%= tag_array.first %>,"className":"Tag"}'>
-            Follow
-          </a>
+    <% elsif user_signed_in? %>
+      <header class="p-2 pr-0 flex items-center justify-between">
+        <a href="/tags" class="crayons-btn crayons-btn--icon crayons-btn--ghost-dimmed" aria-label="Tags Overview" title="Tags Overview">
```
ludwiczakpawel, line 59: Lets drop these classes here, since they add some unnecessary padding and make things misaligned.
Also, shouldn't we use `<%= tags_path %>` instead of `/tags` ?
 
```suggestion
        <a href="<%= tags_path %>" aria-label="Tags Overview" title="Tags Overview">
```

app/views/articles/_sidebar_nav.html.erb
```diff
@@ -45,29 +45,50 @@
...
+          <h3 class="crayons-subtitle-3">Explore Tags</h3>
+        </a>
```
ludwiczakpawel, line 61: Actually... How about we drop that entire link and add one on the right (similar to what we do with cog icon), and also update the copy a little. For example:

```
<header class="p-2 pr-0 flex items-center justify-between">
  <h3 class="crayons-subtitle-3">Explore</h3>
  <a href="<%= tags_path %>" class="fs-s crayons-link crayons-link--brand pr-2" title="Browse all tags">
    All tags &raquo;
  </a>
</header>
```

ludwiczakpawel (changes requested):
Little changes requested. Otherwise looking good!

app/views/articles/_sidebar_nav.html.erb
juliannatetreault: Good call! 😄

app/views/articles/_sidebar_nav.html.erb
juliannatetreault: I spoke too soon... better call! 😂

ludwiczakpawel (approved):
⛴ Ship it! ⛴

krzysiek1507:
This PR introduced an issue.

The cache key [here](https://github.com/forem/forem/blob/45fb6504dc9e6f51d5bdd6e7edd1498720a4d71a/app/views/articles/_sidebar.html.erb#L4) doesn't know use `current_user` and it shares the same view for all guest or signed in users. The problem is that all of the signed in users can see `Explore` or `My tags` based on the first user who populated the cache. The fix would be to extend the cache with `current_user&.id` and `current_user&.following_tags_count`. The downside is that the amount of caches increases by the number of signed in users.

krzysiek1507:
Well, in fact only `current_user&.following_tags_count&.positive?` would be sufficient.

krzysiek1507:
It would have to be changed also here: https://github.com/forem/forem/blob/467e99556cf1fc088bbd91ec0612d4a64ff91ffd/app/views/articles/index.html.erb#L29

I'll add an issue.

juliannatetreault:
> It would have to be changed also here:
> 
> https://github.com/forem/forem/blob/467e99556cf1fc088bbd91ec0612d4a64ff91ffd/app/views/articles/index.html.erb#L29
> 
> I'll add an issue.

Thanks so much for pointing this out @krzysiek1507! I am addressing the issues now.

krzysiek1507:
You are welcome @juliannatetreault. Ping me please in the PR!

juliannatetreault:
I was unable to tag you on the PR itself, both as a reviewer and in the comments, so here is a link to the PR that fixes the bug: https://github.com/forem/forem/pull/11055

cc @krzysiek1507 😄

PR #11023, opened 2020-10-22
General alert modal for rate limits

joshpuetz:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

**Concept**: we want to create reusable messaging to alert users about exceeding rate limits or user interaction limits. Right now these limits are indicated by backend throwing HTTP status code `429 RATE LIMIT EXCEEDED`, but it's up to the frontend to do something with it. Some UIs catch this error and show it in an `alert`, others just throw it away and don't inform the user. Ideally the frontend would have a reusable modal to display (design listed here: https://www.figma.com/file/W0NVUGtoMdQwRc9pH4M5dd/Interaction-limits?node-id=16%3A1)

**Help wanted**: I need help understanding where best to place code. Here's the approach I've taken
- I'm using the crayons `<Modal>` preact component, but would plan to wrap it in a `<RateLimitModal>` component that could dynamically set the explanation text depending on which rate limit was exceeded (individual views would pass this information in when the tried to display the model).
- To be able to use this modal on any view in the application, I created a Javascript pack that's loaded with `javascript_packs_with_chunks_tag` and included in the shell's bottom section.
- I added a utility in `app/assets/javascripts/utilities/showModal.js` to allow Javascript running anywhere in the application the ability to show the rate limit modal
- As a proof of concept, I've hooked up the rate limit modal on the comment posting page (this view is using Javascript, but not Preact components).

There's an alternate path I could see taking:
- Don't use the `<Modal>` preact component, and instead use standard HTML markup
- Expand the plain old Javascript code in `app/assets/javascripts/utilities/showModal.js` (or something in `app/assets/javascripts/initializers`) to handle showing, modifying, and hiding the modal

Additionally:
- Is trying to make this globally available a good idea?

**Please let me know what you think!!**

closes https://github.com/forem/InternalProjectPlanning/issues/192

A_Catskill_Eagle_Similique_commodi_-_Forem_Community_👩💻👨💻

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [ ] no documentation needed

halp!

### DIFF
--- a/app/assets/javascripts/utilities/showModal.js
+++ b/app/assets/javascripts/utilities/showModal.js
@@ -12,3 +12,10 @@ function showModal(context) {
   document.body.classList.add('modal-open');
   initSignupModal();
 }
+
+function showRateLimitModal() {
+    if(document.getElementsByClassName('crayons-modal')[0]) {
+      document.getElementsByClassName('crayons-modal')[0].classList.remove('hidden');
+    } 
+}
+
--- a/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
+++ b/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
@@ -247,7 +247,7 @@ function handleCommentSubmit(event) {
       } else {
         response.json().then(function parseError(errorReponse) {
           form.classList.remove('submitting');
-           showRateLimitModal();
+          showRateLimitModal();
           return false;
         });
       }
--- a/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
+++ b/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
@@ -247,7 +247,7 @@ function handleCommentSubmit(event) {
       } else {
         response.json().then(function parseError(errorReponse) {
           form.classList.remove('submitting');
-          window.alert(errorReponse.error); // eslint-disable-line no-alert
+           showRateLimitModal();
           return false;
         });
       }
--- a/app/javascript/packs/rateLimitModal.jsx
+++ b/app/javascript/packs/rateLimitModal.jsx
@@ -1,0 +1,23 @@
+import { h, render } from 'preact';
+import { Modal } from '../crayons/Modal'; 
+
+function closeRateLimitModal() {
+  if(document.getElementsByClassName('crayons-modal')[0]) {
+    document.getElementsByClassName('crayons-modal')[0].classList.add('hidden');
+  } 
+}
+
+document.addEventListener('DOMContentLoaded', () => {
+  const root = document.getElementById('rate-limit-modal');
+
+  render(
+    <Modal 
+      title="This is a modal title"
+      className='hidden'
+      onClose={closeRateLimitModal}
+    >
+      This is the modal body content
+    </Modal>
+  , root);
+});
+
--- a/app/javascript/packs/rateLimitModal.jsx
+++ b/app/javascript/packs/rateLimitModal.jsx
@@ -1,23 +1,16 @@
 import { h, render } from 'preact';
-import { Modal } from '../crayons/Modal'; 
-
-function closeRateLimitModal() {
-  if(document.getElementsByClassName('crayons-modal')[0]) {
-    document.getElementsByClassName('crayons-modal')[0].classList.add('hidden');
-  } 
-}
+import { RateLimitModal } from '../rateLimitModal/rateLimitModal'; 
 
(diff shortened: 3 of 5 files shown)


vaidehijoshi:
> Don't use the <Modal> preact component, and instead use standard HTML markup

@joshpuetz Would this be with flash messages or something else?

It feels a bit like an antipattern to use a modal here just because you can imagine a user:
1. clicks a button to "publish" on an article or "save" on a comment box
2. the request goes from client to server, server responds with a 429
3. the client is awaiting that response on the frontend, and if the response comes back unsuccessful/the promise rejects, the client opens up a new modal (_through no action of the user, from the user's perspective_) telling them they have to wait.

My initial gut reaction is that the modal is much more heavyweight than we need it to be for what we're trying to do here...but it's also possible that I'm missing some context. Hence, I'm curious to know what the "standard HTML markup" version of this would look like :)

joshpuetz:
> > Don't use the  preact component, and instead use standard HTML markup
> 
> @joshpuetz Would this be with flash messages or something else?
> 
> It feels a bit like an antipattern to use a modal here just because you can imagine a user:
> 
> 1. clicks a button to "publish" on an article or "save" on a comment box
> 2. the request goes from client to server, server responds with a 429
> 3. the client is awaiting that response on the frontend, and if the response comes back unsuccessful/the promise rejects, the client opens up a new modal (_through no action of the user, from the user's perspective_) telling them they have to wait.
> 
> My initial gut reaction is that the modal is much more heavyweight than we need it to be for what we're trying to do here...but it's also possible that I'm missing some context. Hence, I'm curious to know what the "standard HTML markup" version of this would look like :)

Thanks for taking a look @vaidehijoshi! 

These are modals for using interaction and rate limits (for example: a user is creating hundreds of comments a minute). Right now the user is getting a javascript `alert` box: 

localhost_3000_says_and_The_Far-Distant_Oxus_Aut_consequuntur_-_Forem_Community_👩💻👨💻

We want to replace that with something more friendly (but it's still meant to interrupt the user). The design of the modal would be the same if I use Preact or plain old HTML. Here's the way it looks in @lisasy's designs:

Interaction_limits_–_Figma

This will look the same if I use preact versus inject HTML into the Dom manually: it's the frontend implementation I'm looking for feedback on, not the design of the modal.

vaidehijoshi:
> These are modals for using interaction and rate limits (for example: a user is creating hundreds of comments a minute). Right now the user is getting a javascript `alert` box:
> 
> localhost_3000_says_and_The_Far-Distant_Oxus_Aut_consequuntur_-_Forem_Community_👩💻👨💻
> 
> We want to replace that with something more friendly (but it's still meant to interrupt the user). The design of the modal would be the same if I use Preact or plain old HTML. 
> This will look the same if I use preact versus inject HTML into the Dom manually: it's the frontend implementation I'm looking for feedback on, not the design of the modal.

Ah, I didn't fully understand the first time I read this that you meant that it was literally using JS `alert`! I haven't run into that before -- I have only ever seen `429` rate limits as those red flash messages across the app so that's part of the reason I was confused. Thanks for clarifying that.

@joshpuetz I think that using the already-available `<Modal>` preact component is the right approach, IMHO. I like the idea of wrapping it in a reusable component that is responsible for figuring out what message to display based on the response/type of rate limit.

app/assets/javascripts/utilities/showModal.js
```diff
@@ -12,3 +12,10 @@ function showModal(context) {
   document.body.classList.add('modal-open');
   initSignupModal();
 }
+
+function showRateLimitModal() {
+    if(document.getElementsByClassName('crayons-modal')[0]) {
```
lisasy, line 17: Is it good practice for us to call the class name by its Crayons class name (which is descriptive of styles)? An alternative could be specifying a `rate-limit-modal` class and calling it that way.

@ludwiczakpawel @nickytonline

ludwiczakpawel: [I'm a big fan of using `js-something` class names for purpose of calling an element in JS](https://dev-internal.slack.com/archives/C7GE84DEJ/p1579172524145300). That's also helpful with indicating whether or not an element has some JS associated.

joshpuetz:
Closing this in favor of https://github.com/forem/forem/pull/11102

PR #10517, opened 2020-10-01
Added form fields for the admin podcasts dashboard.

wotta:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adding more edit capabilities in the admin dashboard for podcasts.

## Related Tickets & Documents
Closes https://github.com/forem/forem/issues/10482

## QA Instructions, Screenshots, Recordings
When this issue is being pulled for a functional review one should have access to an admin 

## Added tests?

- [ ] yes
- [ ] no, because they aren't needed
- [x] no, because I need help _at least I think_

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

## [optional] Are there any post deployment tasks we need to perform?
Not that I know of 

## [optional] What gif best describes this PR or how it makes you feel?

hoping this is what they wanted :D

PR #11024, opened 2020-10-22
Ability to Mark Podcasts as Published in Admin Podcasts

juliannatetreault:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR gives Admins the ability to publish and unpublish podcasts from `/admin/podcasts/:id/edit` via a "Published" checkbox. This change makes it so that Admins no longer have to navigate to `/resource_admin/podcasts` to publish and unpublish podcasts. Additionally, this PR updates the existing `podcast_spec.rb` to test that the checkbox properly updates the podcast.

Closes https://github.com/forem/InternalProjectPlanning/issues/206

**To QA this PR, first ensure that you are and `admin` and then please do the following**:

- Navigate to `/admin/podcasts/:id/edit` and click the "Published" checkbox and then "Update" to publish or unpublish a podcast:
Screen Shot 2020-10-22 at 12 16 26 PM

- After updating the podcast, you should be redirected back to `/admin/podcasts` where you will see the status (published or unpublished) of the podcast reflected in the podcast list, as well as a flash message, indicating that the update was successful:
Screen Shot 2020-10-22 at 12 16 56 PM

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

N/A

Rainbow Microphone

### DIFF
--- a/spec/requests/admin/podcasts_spec.rb
+++ b/spec/requests/admin/podcasts_spec.rb
@@ -64,9 +64,10 @@ RSpec.describe "/admin/podcasts", type: :request do
       put admin_podcast_path(podcast), params: { podcast: { title: "hello",
                                                             feed_url: "https://pod.example.com/rss.rss",
                                                             published: true } }
-      expect(podcast.reload.title).to eq("hello")
-      expect(podcast.reload.feed_url).to eq("https://pod.example.com/rss.rss")
-      expect(podcast.reload.published).to eq(true)
+      podcast.reload
+      expect(podcast.title).to eq("hello")
+      expect(podcast.feed_url).to eq("https://pod.example.com/rss.rss")
+      expect(podcast.published).to eq(true)
     end
 
     it "redirects after update" do
--- a/spec/requests/admin/podcasts_spec.rb
+++ b/spec/requests/admin/podcasts_spec.rb
@@ -2,7 +2,7 @@ require "rails_helper"
 
 RSpec.describe "/admin/podcasts", type: :request do
   let(:admin) { create(:user, :super_admin) }
-  let(:podcast) { create(:podcast) }
+  let(:podcast) { create(:podcast, published: false) }
   let(:user) { create(:user) }
 
   before do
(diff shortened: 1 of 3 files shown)


spec/requests/admin/podcasts_spec.rb
```diff
@@ -60,12 +60,13 @@
...
-    it "updates" do
+    it "updates the podcast" do
       put admin_podcast_path(podcast), params: { podcast: { title: "hello",
-                                                            feed_url: "https://pod.example.com/rss.rss" } }
-      podcast.reload
-      expect(podcast.title).to eq("hello")
-      expect(podcast.feed_url).to eq("https://pod.example.com/rss.rss")
+                                                            feed_url: "https://pod.example.com/rss.rss",
+                                                            published: true } }
+      expect(podcast.reload.title).to eq("hello")
```
jacobherrington, line 67: Just curious can we call `podcast.reload` once here like it was before, or is it necessary to call it several times?

juliannatetreault: @jacobherrington we can call it once! I like how it looks this way, but I'm happy to revert the change, as I have no strong preference. 😊

juliannatetreault: Changed @jacobherrington !

fdoxyz (approved):
There's a bit of a larger refactor for this page in progress (#10517) to get the full features available in `/admin/podcasts`. 

That PR has some changes requested and this won't have a big effect on the progress done there, so in favor of shipping this fix it LGTM. Thanks @juliannatetreault! 👍

Zhao-Andy (approved):
LGTM!

spec/requests/admin/podcasts_spec.rb
jacobherrington: Fewer calls is probably faster, but probably not **that** much faster.

PR #11025, opened 2020-10-22
Spec Fix:Set Javascript Locale to Match Time.zone

mstruve:
- [x] Spec Fix(I think)

## Description
I got some information from my logging that was showing that the date we were displaying from the local_date helper did not match UTC or the Timezone set by zonebie.
Broken on Travis
```ruby
# Readable Compare date 2020-10-23 00:30:08 +0800 (timezone)
# 2020-10-22T16:30:08Z (utc)
# Thursday, October 22, 2020, 8:30:08 PM (??? some other locale time)
```
When I went back to the `local_date()` helper method I tracked it back to the PR that created it. It was here I realized that the helper method was not the final say in the date. Instead, it [actually gets set by a javascript date helper](https://github.com/forem/forem/blob/master/app/assets/javascripts/initializers/initializeArticleDate.js#L6) to ensure it matches the user's locale time. This made me realize we needed to go a step further and not only did the backend of the browser need to match our Time.zone set by Zonebie, but our javascript needed to as well. After a lot of googling I came across [this old post](https://makandracards.com/makandra/10753-mock-the-browser-time-or-time-zone-in-selenium-features) about [timemachine.js](https://github.com/schickling/timemachine) that you can use to set the time in your javascript. It's a little older but since it's only for our tests I figured it would be fine. 

@nickytonline would love your thoughts on this!

https://github.com/orgs/forem/projects/6#card-47746166

alt_text

### DIFF
--- a/app/assets/javascripts/base.js.erb
+++ b/app/assets/javascripts/base.js.erb
@@ -6,6 +6,16 @@
 //= require honeybadger-js/dist/honeybadger.js
 //= require ahoy.js/dist/ahoy.js
 //= require serviceworker-companion
+//= require timemachine.js
+
+<% if Rails.env.test? %>
+  // Ensure that locale time in javascript matches the Timezone on
+  // our server during testing
+  timemachine.config({
+    dateString: '<%= Time.zone.now.iso8601 %>',
+    tick: true
+  });
+<% end %>
 
 var instantClick
   , InstantClick = instantClick = function(document, location, $userAgent) {
--- a/app/assets/javascripts/base.js.erb
+++ b/app/assets/javascripts/base.js.erb
@@ -6,16 +6,6 @@
 //= require honeybadger-js/dist/honeybadger.js
 //= require ahoy.js/dist/ahoy.js
 //= require serviceworker-companion
-//= require timemachine.js
-
-<% if Rails.env.test? %>
-  // Ensure that locale time in javascript matches the Timezone on
-  // our server during testing
-  timemachine.config({
-    dateString: '<%= Time.zone.now.iso8601 %>',
-    tick: true
-  });
-<% end %>
 
 var instantClick
   , InstantClick = instantClick = function(document, location, $userAgent) {
--- a/app/assets/javascripts/initializers/initializeTimemachine.js.erb
+++ b/app/assets/javascripts/initializers/initializeTimemachine.js.erb
@@ -1,0 +1,10 @@
+//= require timemachine.js
+
+<% if Rails.env.test? %>
+  // Ensure that locale time in javascript matches the Timezone on
+  // our server during testing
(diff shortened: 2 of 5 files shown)


app/assets/javascripts/base.js.erb
```diff
@@ -6,6 +6,16 @@
 //= require honeybadger-js/dist/honeybadger.js
 //= require ahoy.js/dist/ahoy.js
 //= require serviceworker-companion
+//= require timemachine.js
+
+<% if Rails.env.test? %>
```
mstruve, line 11: I'm sure there is a much better place to put this. @vaidehijoshi or @nickytonline please feel free to jump in, I am not very familiar with our js organization.

rhymes (changes requested):
Awesome! Out loud makes perfect sense that the frontend and the backend are aligned in tests, on the other end though I wonder if this it's going to hide future issues. After all in production Forem servers run on UTC, clients don't. Do you think that could be an issue. We could probably catch those in a future CI pipeline where we run some builds in UTC and some in the random zonebie provided time zone. This also might not be needed if we eventually move away from Rails system tests and use end to end testing where I'm sure we can simulate whatever time zone we want.

Anyhow:

I'm not sure what's the best place either for the time machine initializer but it seems a good one! :D 

Great find for https://makandracards.com/makandra/10753-mock-the-browser-time-or-time-zone-in-selenium-features - we should add it to our knowledge base ehheh.

I requested changes because the library shoulnd't be packaged in production

package.json
```diff
@@ -164,6 +164,7 @@
     "pusher-js": "^7.0.0",
     "rails-erb-loader": "^5.5.2",
     "stimulus": "^1.1.1",
+    "timemachine": "^0.3.2",
```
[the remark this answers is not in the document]
mstruve: Ah, thank you! I didnt realize you could do that like gems!

app/assets/javascripts/initializers/initializeTimemachine.js.erb
```diff
@@ -0,0 +1,10 @@
+//= require timemachine.js
+
+<% if Rails.env.test? %>
+  // Ensure that locale time in javascript matches the Timezone on
```
vaidehijoshi, line 4:
```suggestion
  // Ensure that locale time in JavaScript matches the Timezone on
```
😉

vaidehijoshi (approved):
This is great! Agree with @rhymes on where to include in dependencies 👍

app/assets/javascripts/initializers/initializeTimemachine.js.erb
mstruve: You are like me and Elasticsearch 😅

rhymes (approved):

PR #11027, opened 2020-10-22
Experiment with the removal of inline critical path css

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Since early in DEV, we have _inlined_ critical path css on most pages. This means that visitors can view fully rendered CSS _as soon as the page starts rendering_. We do not make external CSS calls.

But it means that users cannot make as good use of browser caching, and the pages being downloaded are usually _much_ bigger.

The _ideal_ scenario is that we smartly pluck _only_ the CSS we need, but we have strayed _way_ too far from that (our file called "minimal" is actually quite big now). It would be great if we could come up with a cool solution to detect which css classes the page will use and only use that subset from our awesome crayons and utility classes. But that strikes me as a bigger project.

Largely I suspect this won't have a massive impact, but it could open up more smart caching opportunities as we grow the Forem network and we have more overall traffic coming from _return_ visitors than search/social/etc. In the big picture this should save bytes for our users, while possibly slowing down _the first time_ someone visits a Forem (but not in an unusual way, we're still very lightweight compared to most of the web). We can also make better use of new HTTP standards which help parallelize things, and overall become a bit more "conventional"

In terms of smart caching opportunities, a big one is sharing assets across the Forem Cloud ecosystem. We already have the healthy constraint that we cannot use Forem-specific ENV vars in assets because most customization now happens at the DB layer and we compile the assets outside of the app's context. So all Forems will essentially share the same CSS and JS.

I think of this as a _test_ because there are implications we can pay attention to. It's really hard to genuinely determine if this will affect user behavior, but this may affect SEO (hopefully in a good way because it will simplify the pages and make things more typical for Google bots).

If nobody disagrees with this, I say we do it. One reason it makes sense to mess with this _now_ is that we're told Google's not touching search results as to not give the impression of making a change which would politically bias results leading up to the election. It's a good time to make changes with fewer variables at play.

If there are bad outcomes from this change, we come back to this PR and copy the old stuff back in. 😄 

## Technical review:

For this to _technically_ work, it shouldn't really change anything in the UX.


vaidehijoshi (approved):

atsmith813 (approved):
send_it_gif

app/assets/javascripts/initializePage.js
```diff
@@ -46,7 +46,6 @@ function callInitializers() {
   initializeArticleDate();
   initializeArticleReactions();
   initNotifications();
-  initializeStylesheetAppend();
```
vaidehijoshi, line 49: I think we want to remove the reference to this (even though it's just a comment) in line 2 of this file

vaidehijoshi (approved):
Pending CI going greeN :)

app/assets/javascripts/initializePage.js
benhalpern: Good call.

docs/technical-overview/architecture.md
```diff
@@ -22,14 +22,16 @@ We also use server-side caching: [Rails caching][rails_caching]. If you see
 `Rails.cache` or `<%= cache ... %>`, this is code affected in production by
 caching.
 
-## We use inline CSS and deferred scripts for usage performance improvements
+## We Mostle defer scripts for usage performance improvements
```
Zhao-Andy, line 25: Small typo: `Mostle` should be `mostly`

jacobherrington (approved):

Issue #7086, opened 2020-04-04
Article show page: broken timestamp


mzakzook:
**Describe the bug**

The date printed to my post is one day ahead. Hover-over shows correct date.

<img width="1094" alt="Screen Shot 2020-03-28 at 10 06 09 AM" src="https://user-images.githubusercontent.com/19253610/78457486-5b107f80-765f-11ea-9287-e888552291dd.png">

**To Reproduce**

**Expected behavior**

**Screenshots**

<img width="1094" alt="Screen Shot 2020-03-28 at 10 06 09 AM" src="https://user-images.githubusercontent.com/19253610/78457497-711e4000-765f-11ea-95c5-55f60845d056.png">

**Desktop (please complete the following information):**

- OS:
- Browser:
- Version:

**Smartphone (please complete the following information):**

- Device:
- OS:
- Browser:
- Version:

**Additional context**

My timezone is Pacific Time.

citizen428:
Thanks for the issue @mzakzook! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

citizen428:
For anyone investigating this: if I had to venture a guess, I'd say we store this in UTC without time zone info, so when a post is written at specific times of the day (anything from midnight - UTC offset to midnight) this happens. This may be wrong though, as it's just a gut feeling based on previous experiences.

rhymes:
This happens because the datetime in the HTML `title` attribute is correctly rendered using the client's computer timezone, while `Mar 23`, the one shown in the text, is computed server side and the server is always in UTC. 

Great catch @mzakzook and good instincts @citizen428 :D

rhymes:
I played with it a bit and it's definitely doable, we already have the function to do it. There are a few decisions to be made:

- do we want to localize the visible string? Right now we're only localizing `title` attributes, both for articles and for comments:

[...]

PR #11028, opened 2020-10-22
Bug Fix:Use local_date Helper for All Published Timestamps

mstruve:
- [x] Bug Fix

## Description
In https://github.com/forem/forem/pull/9285 we added a `local_date` helper to ensure that article published timestamps were shown in locale time and not server time. However, [this was not applied to all of the published timestamps](https://github.com/forem/forem/issues/7086) leading to continued bugs. This adds the helper to two other places I found that we display article published timestamps

Closes https://github.com/forem/forem/issues/7086

## Added tests?
The article show page is already tested and is very flaky which is why I have issued https://github.com/forem/forem/pull/11025 which will set up a testing framework that ensures the locale time on our frontend always matches the server testing time. Once that is merged we can go back and add Dashboard specs if we want to. 

alt_text

### DIFF
--- a/app/views/articles/_about_author.html.erb
+++ b/app/views/articles/_about_author.html.erb
@@ -1,6 +1,6 @@
 <div>
   <p class="mb-4 color-base-60 fs-s">
-    Posted on <time datetime="<%= @article.published_timestamp %>"><%= @article.readable_publish_date %></time> by:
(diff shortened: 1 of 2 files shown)


rhymes (approved):
Cool! Right now we don't really have a policy about which dates should be localized and which don't but I think these definitely should and we should eventually move to all :D 

Thanks @mstruve!

joshpuetz (approved):

PR #11029, opened 2020-10-22
Use APP_NAME rather than APP_DOMAIN in chat channels

vaidehijoshi:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
`APP_NAME` may not always be set (it is not "required" per se), which can cause issues while deriving the `pusher_channels`. 

We should use `APP_DOMAIN`, which is more common and always has a default value.

https://github.com/forem/InternalProjectPlanning/issues/212

Ensure that chat works correctly.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

### DIFF
--- a/.env_sample
+++ b/.env_sample
@@ -4,7 +4,6 @@ export SKIP_SERVICEWORKERS="true"
 # App core values
 export APP_DOMAIN="localhost:3000"
 export APP_PROTOCOL="http://"
-export APP_NAME="forem_local"
 export FOREM_OWNER_SECRET="secret"
 
 # Openresty domain + Protocol setting for development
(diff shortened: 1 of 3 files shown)


benhalpern (approved):

mstruve (approved):
Can we close the loop and make sure when @jdoss returns he knows he can remove this from his list of env variables?

vaidehijoshi:
> Can we close the loop and make sure when @jdoss returns he knows he can remove this from his list of env variables?

@mstruve great call! set up a reminder in the #systems channel so that he'll be in the loop and the robots will remember in case i forget 👍 🤖

PR #11030, opened 2020-10-23
Fix honeybadger console error

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This was being caused by two reasons: 

1. There was a caught runtime error throwing to Honeybadger (this is what it should do, but I fixed the error so it won't be caught)
2. Honeybadger is not yet defined at this point because the error is happening in the inline portion of the script we run right in the html to help set the theme, etc.

Since we define Honeybadger in an async script, we can't technically call it until later.

This is a hot fix which introduces a race condition, but we should reorganize the code later for a permanent fix.

Screen Shot 2020-10-22 at 8 26 35 PM

### DIFF
--- a/app/views/layouts/_user_config.html.erb
+++ b/app/views/layouts/_user_config.html.erb
@@ -39,10 +39,10 @@
       } else {
         document.body.classList.add("user-tags-followed-max");
       }
-      sidebarTags.style.height = "100vh"
-      sidebarTags.style.backgroundColor = "green"
     }
   } catch(e) {
+    setTimeout(function(){
       Honeybadger.notify(e);
+    },1000)
   }
 </script>


nickytonline (approved):

app/views/layouts/_user_config.html.erb
```diff
@@ -39,10 +39,10 @@
       } else {
         document.body.classList.add("user-tags-followed-max");
       }
-      sidebarTags.style.height = "100vh"
-      sidebarTags.style.backgroundColor = "green"
     }
   } catch(e) {
+    setTimeout(function(){
```
nickytonline, line 44: This is definitely a workaround, so I'll open up an issue about sorting this out.

citizen428 (approved):

PR #11037, opened 2020-10-23
Add temporary RSS reader worker to gather monitoring data

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

In the quest to write a new module to ingest feeds faster, see https://github.com/forem/forem/issues/10996, we need to know how the current one behaves in production on dev.to which is currently the biggest collection of feeds the Forem ecosystem has. We know it's slow, we have a rough idea about why it's slow but as for everything, navigating with real world data would be better. 

I'm adding a Sidekiq worker that we can call manually.

**question**: should we replace the call in 

https://github.com/forem/forem/blob/5b1f05876402fa6ee814c42f9e495ad6c983ff8c/lib/tasks/fetch.rake#L3-L7

with the worker called inline? `Articles::DevRssReaderWorker.new.perform` ?

https://github.com/forem/forem/issues/10996

1. change your community name to DEV
1. run the worker

## Added tests?

- [x] yes
- [ ] no, because I need help

### DIFF
--- a/app/workers/articles/dev_rss_reader_worker.rb
+++ b/app/workers/articles/dev_rss_reader_worker.rb
@@ -1,0 +1,17 @@
+# NOTE: [rhymes]
+# This script will soon be removed. We need it to collect monitoring data on
+# Datadog about the production behavior of the `RssReader` class
+module Articles
+  class DevRssReaderWorker
+    include Sidekiq::Worker
+
+    sidekiq_options queue: :medium_priority, retry: 10
+
+    def perform
+      return unless SiteConfig.community_name == "DEV"
+
+      # we force fetch to have realistic data
+      RssReader.get_all_articles(force: true)
+    end
+  end
+end
--- a/app/workers/articles/dev_rss_reader_worker.rb
+++ b/app/workers/articles/dev_rss_reader_worker.rb
@@ -9,6 +9,7 @@ module Articles
 
     def perform
       return unless SiteConfig.community_name == "DEV"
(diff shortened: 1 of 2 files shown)


app/workers/articles/dev_rss_reader_worker.rb
```diff
@@ -0,0 +1,17 @@
...
+  class DevRssReaderWorker
+    include Sidekiq::Worker
+
+    sidekiq_options queue: :medium_priority, retry: 10
+
+    def perform
+      return unless SiteConfig.community_name == "DEV"
+
+      # we force fetch to have realistic data
+      RssReader.get_all_articles(force: true)
```
mstruve, line 14: Since we dont want to end up in an infinite retry loop if for some reason this keeps restarting(Once Sidekiq jobs are running in our system there is no way to cancel them) due to deploys, let's add a little "quick exit" clause here that we could trigger if we want the worker to stop being enqueued. 
```ruby
return if Rails.cache.read("cancel_rss_job").present?
```

mstruve (approved):

app/workers/articles/dev_rss_reader_worker.rb
rhymes: Love the escape hatch! I'll add it

maestromac (approved):
LGTM!

benhalpern (approved):
Looks good to me. Someone from SRE should merge this to start observing. 👍

PR #11039, opened 2020-10-23
Confirmations that users GDPR data was deleted

lightalloy:
- [x] Feature

## Description
- list of internal notifications for admins to destroy users gdpr data that should replace slack notifications
- admins are able to confirm that the data was deleted, then the notification is destroyed

Removing slack notifications is the next step.

https://github.com/forem/forem/projects/31#card-45606967

List of the users whose gdpr data should be deleted:
изображение
A new record in this list should appear after a user profile is destroyed by themselves or by an admin.

Notice after user delete (by admin):
изображение

## Added tests?
- [x] yes


app/controllers/admin/application_controller.rb
```diff
@@ -41,7 +41,7 @@ class ApplicationController < ApplicationController
     private
 
     def authorization_resource
-      self.class.name.demodulize.sub("Controller", "").singularize.constantize
+      self.class.name.sub("Admin::", "").sub("Controller", "").singularize.constantize
```
lightalloy, line 44: Demodulize won't work in case the controller has one more namespace apart from `Admin::`. Are there better ideas for this?

app/controllers/admin/invitations_controller.rb
```diff
@@ -15,7 +15,7 @@ def create
       User.invite!(email: email,
                    name: name,
                    username: username,
-                   remote_profile_image_url: Users::ProfileImageGenerator.call,
+                   remote_profile_image_url: ::Users::ProfileImageGenerator.call,
```
lightalloy, line 18: The tests were complaining about uninitialized `Admin::Users::ProfileImageGenerator`, so I had to specify this.

app/controllers/admin/application_controller.rb
citizen428: `demodulize works fine for nested modules too:

```ruby
Admin::Something::TestsController.name.demodulize.chomp("Controller")
#=> "Tests"
```

app/controllers/admin/users/gdpr_delete_requests_controller.rb
```diff
@@ -0,0 +1,17 @@
+module Admin
+  module Users
+    class GdprDeleteRequestsController < Admin::ApplicationController
+      layout "admin"
+
+      def index
+        @gdpr_delete_requests = ::Users::GdprDeleteRequest.order("created_at desc").page(params[:page]).per(50)
```
citizen428, line 7: We moved away from strings for AR as much as possible:

```suggestion
        @gdpr_delete_requests = ::Users::GdprDeleteRequest.order(created_at: :desc).page(params[:page]).per(50)
```

app/models/users/gdpr_delete_request.rb
```diff
@@ -0,0 +1,7 @@
+module Users
```
citizen428, line 1: I have no problem with the namespace, but is it neccessary? Will there be other types of `GdprDeleteRequest` and we want to avoid a name clash?

app/controllers/admin/users/gdpr_delete_requests_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+      end
+
+      def destroy
+        @gdpr_delete_request = ::Users::GdprDeleteRequest.find(params[:id])
```
citizen428, line 11: Since the deletion is the confirmation by the admin that they removed the data, I think we should log this. Something like:

```ruby
AUDIT_LOG_CATEGORY = "admin.gdpr_delete.confirm".freeze

def destroy
  ...
  AuditLog.create(
    user: current_user, 
    category: AUDIT_LOG_CATEGORY,
    roles: current_user.roles_name,
    ...
)
```

app/views/admin/users/gdpr_delete_requests/index.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      <tr>
+        <td><%= r.created_at.strftime("%m-%d-%Y %H:%m") %></td>
+        <td><%= r.user_id %></td>
+        <td><%= r.username %></td>
+        <td><%= r.email %></td>
+        <td>
+          <%= link_to "Done!",
+                      url_for(action: :destroy, id: r.id),
+                      method: :delete,
+                      data: { confirm: "Have you destroyed data from external services? (GA, Maichimp)" },
```
citizen428, line 26: Typo:

```suggestion
                      data: { confirm: "Have you destroyed data from external services? (GA, Mailchimp)" },
```

db/migrate/20201012072557_create_users_gdpr_delete_requests.rb
```diff
@@ -0,0 +1,11 @@
+class CreateUsersGdprDeleteRequests < ActiveRecord::Migration[6.0]
+  def change
+    create_table :users_gdpr_delete_requests do |t|
+      t.integer :user_id
+      t.string :email
```
citizen428, line 5: Should the email and/or username be mandatory, i.e. `null: false`? 

Also: should they have a simple boolean status column, e.g. `done`? The reason I'm asking is that we may want a uniquess validation across all these attributes so that a user can only have one active open request, but theoretically more in the future (sign up, trigger GDPR delete, sign up again later, trigger GDPR delete again).

mstruve: Is it an issue that we are now storing the user's email on this GdprDeleteRequest when the point is to delete all of the user's GDPR data which I assume includes their email?

lightalloy: Currently, the record is deleted when an admin confirms that the data from 3-rd party services was deleted. I thought that we shouldn't keep emails as the data should be deleted :thinking:

mstruve: I think I might be confused as to the flow here. So is this what the flow looks like?
1. User deletes themselves 
2. User delete triggers `GdprDeleteRequest` to be created, is that right???
3. GDPR data is deleted by admin
4. Admin deletes `GdprDeleteRequest`

citizen428: Yes, that's how I understand the flow too @mstruve. The confirm button actually deletes the entry in `GdprDeleteRequest` via a `DELETE` request.

app/controllers/admin/application_controller.rb
lightalloy: I meant that in this case I wanted to keep the namespace as a part of the resource name. In this case, `Users::GdprDeleteRequest`model and `Admin::Users::GdprDeleteRequestsController`.

app/models/users/gdpr_delete_request.rb
lightalloy: Maybe, in this case, namespace adds more problems than it solves :thinking: 
I often tend to "namespace all the things" :)

db/migrate/20201012072557_create_users_gdpr_delete_requests.rb
lightalloy: @mstruve That's right, after admin confirms, the record is destroyed, so we won't be storing email anymore.

app/controllers/admin/application_controller.rb
citizen428: Oh, now I get it, that makes sense. 👍

app/models/users/gdpr_delete_request.rb
citizen428: Yes, I like namespaces too, just this specific one didn't seem to add too much. But feel free to leave it as is.

mstruve (commented):
The idea of adding an entirely new model and table to our app only for keeping track of GDPR delete requests which are all likely going to have very short lifespans feels like we are adding a lot of complexity for not a lot of functionality. 

I have been thinking about this PR for the past couple of days and I am wondering, is this something we could accomplish with say AuditLog? Or some other generic model that could short term track these requests for us? 

Did we consider "soft-delete"ing the user and giving it a status like "gdpr_data_pending" or something? Then we could use that status to track those users that need GDPR data deleted and then once it is done finish the user deletion. 

Thoughts?

package.json
```diff
@@ -73,6 +73,11 @@
       "svgo --disable=removeViewBox --pretty"
     ]
   },
+  "prettier": {
```
citizen428, line 76: I'm not sure we should add new Prettier options as part of a PR. Probably best to do this separately and get the team to agree.

citizen428:
I think the main idea here was that we delete the user ASAP since that's something that's fully under control and doesn't require manual intervention. I guess this is why @lightalloy decided to use an extra model, since we need to temporarily store the email address that needs to be removed somewhere. While this is nice and respectful towards the user, the GDPR grants us up to 30 days for fullfilling a deletion request, so we could theoretically get away with a column on the users table. We just would need to make sure that we stay within the 30 days, so maybe it should be a timestamp instead of a boolean, e.g. `gdpr_deletion_requested_at`. This would also give us a natural sort order for the admin area and we could even apply special styles when we get too close to 30 days.

package.json
lightalloy: An unintentional change :grimacing:

mstruve:
> We just would need to make sure that we stay within the 30 days, so maybe it should be a timestamp instead of a boolean, e.g. gdpr_deletion_requested_at. This would also give us a natural sort order for the admin area and we could even apply special styles when we get too close to 30 days. Whether we want to add yet another column to the users table is a different story, though in this specific case I'd probably be ok with it.

I am with you that adding a column to the user's table always gives me pause but I think in this case I would also be OK with it. (It also helps that you are removing a bunch of them 😉 ) This keeps all the data close together in a single model which I think really drives down the chance of something breaking. For example like a user is deleted and their GDPR request errors out then you are left with nothing kinda scenario.

@lightalloy what are your thoughts? I know it's frustrating when we throw out completely new ideas and architecture to solve a problem but I am really trying to be thoughtful as we grow this app to make it the more resilient and maintainable it can be.

lightalloy:
I will think about other approaches but for now, adding a column to `users` table and changing the deletion logic seems too much for this task :shrug: : this way we'll store both users and "deleted users" in the table which may also require changing other logic related to the users which I wouldn't want to touch :thinking:

lightalloy:
I explored other possible solutions for this task and decided to reopen this or for the discussion:

As I mentioned before, I don't think that adding a column to `users` table and soft-deleting is a good decision as it can affect other logic related to users and deleting users. Deleting is a complicated process already so I would like to avoid modifying the logic while working on the task.
So I explored using `AuditLog` for this feature. It's possible but I have concerns about it: E.g. we'll have to store email in `AuditLog`s to be able to pass it to the admin (so that they would be able to destroy the related gdpr data) and later delete it. Modifying log records seem like a hack. It's possible to create a special `AuditLog` record for gdpr delete requests and then delete it but this doesn't seem to be a proper use of logs either.
So, for me, creating a separate entity for gdpr delete requests still seem like a better solution than interfering with the app logic.

citizen428:
Curious to hear what others think about this, e.g. @rhymes or @atsmith813.

app/controllers/admin/users/gdpr_delete_requests_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      def index
+        @gdpr_delete_requests = ::Users::GdprDeleteRequest.order(created_at: :desc).page(params[:page]).per(50)
+      end
+
+      def destroy
+        @gdpr_delete_request = ::Users::GdprDeleteRequest.find(params[:id])
+        @gdpr_delete_request.destroy
+
+        AuditLog.create(
```
rhymes, line 14: ❤️

app/controllers/admin/users_controller.rb
```diff
@@ -76,9 +76,13 @@ def full_delete
...
       begin
         Moderator::DeleteUser.call(user: @user)
+        link = helpers.link_to("the page", admin_users_gdpr_delete_requests_path, "data-no-instant" => true)
+        # rubocop:disable Rails/OutputSafety
         message = "@#{@user.username} (email: #{@user.email.presence || 'no email'}, user_id: #{@user.id}) " \
           "has been fully deleted." \
-          "If this is a GDPR delete, delete them from Mailchimp & Google Analytics."
+          "If this is a GDPR delete, delete them from Mailchimp & Google Analytics " \
+          " and confirm on #{link}.".html_safe
+        # rubocop:enable Rails/OutputSafety
```
rhymes, line 85: You can probably avoid `Rails/OutputSafety` and calling `helpers.link_to` by using `content_tag` here with returns a `SafeBuffer`, see https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsoutputsafety

app/models/users/gdpr_delete_request.rb
rhymes: we're a bit inconsistent with namespaces for models to be fair. I like the idea of having separate tables for separate types of GDPR requests so I think we can keep this

app/views/admin/users/_menu.html.erb
```diff
@@ -0,0 +1,48 @@
...
+      <%= link_to "Invitations", admin_invitations_path, class: "nav-link #{'active' if params[:controller] == 'admin/invitations'}" %>
+    </li>
+    <% if params[:controller] == "admin/invitations" %>
+      <li class="nav-item">
+        <%= link_to "New", new_admin_invitation_path, class: "nav-link" %>
+      </li>
+    <% end %>
+    <li class="nav-item" style=";">
+      <%= link_to admin_users_gdpr_delete_requests_path, class: "nav-link #{'active' if params[:controller] == 'admin/users/gdpr_delete_requests'}" do %>
+        Gdpr Delete Requests
```
rhymes, line 41: small non blocking nitpick: `GDPR` all caps

app/controllers/admin/users_controller.rb
```diff
@@ -76,9 +76,13 @@ def full_delete
       @user = User.find(params[:id])
       begin
         Moderator::DeleteUser.call(user: @user)
+        link = helpers.link_to("the page", admin_users_gdpr_delete_requests_path, "data-no-instant" => true)
+        # rubocop:disable Rails/OutputSafety
         message = "@#{@user.username} (email: #{@user.email.presence || 'no email'}, user_id: #{@user.id}) " \
           "has been fully deleted." \
-          "If this is a GDPR delete, delete them from Mailchimp & Google Analytics."
+          "If this is a GDPR delete, delete them from Mailchimp & Google Analytics " \
```
rhymes, line 83: missing space between the dot and `If`

rhymes (approved):
Really cool work @lightalloy! I left a few non blocking notes!

app/controllers/admin/users/gdpr_delete_requests_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+          user: current_user,
+          category: "admin.gdpr_delete.confirm".freeze,
```
mstruve, line 16: I dont think we need `.freeze` here.

mstruve (approved):
Thanks for considering my feedback and exploring other avenues @lightalloy!

Left a small non-blocking comment otherwise LGTM!

app/models/users/gdpr_delete_request.rb
citizen428: I don't think there are separate types of GDPR requests though. The "right to be forgotten" generally means the user and all their associated data, I don't think we need to be more granular than this.

citizen428 (approved):

PR #11041, opened 2020-10-23
Don't use instantclick for registration links

Zhao-Andy:
- [x] Bug Fix

## Description
This fixes an issue where navigating to the registration page `/users/signup` via InstantClick (clicking the link itself) passes along the InstantClick param `?i=i`, which in turn breaks CSS for the email signup link.

**So the biggest thing to consider is whether we want that InstantClick navigation feeling still when clicking from the homepage to the "Create New Account" registration page.** I think it's a good touch but it is also prone to causing problems with the rest of our sign up links.

1. Add the two changed lines to your respective files, probably easier than pulling this PR and checking out via Git.
2. Have email registration enabled via /admin/config, Authentication section
3. Log out or open an incognito/private window
4. Go to the homepage
5. Click the Create New Account link
6. Click the Sign up with email link
7. See that the CSS renders properly

## Added tests?

## Added to documentation?
- [x] no documentation needed

No

A dude weighs his options in his mind, using his hands to gesture like he's considering two options.

### DIFF
--- a/app/views/shared/_authentication_actions.html.erb
+++ b/app/views/shared/_authentication_actions.html.erb
@@ -1,4 +1,4 @@
-<a href="<%= sign_up_path(state: "new-user") %>" class="crayons-btn" aria-label="Create new account">
+<a href="<%= sign_up_path(state: "new-user") %>" class="crayons-btn" aria-label="Create new account" data-no-instant>
   Create new account
 </a>
 <a href="<%= sign_up_path %>" class="crayons-btn crayons-btn--ghost-brand" aria-label="Log in">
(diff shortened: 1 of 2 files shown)


app/views/shared/_authentication_actions.html.erb
```diff
@@ -1,4 +1,4 @@
-<a href="<%= sign_up_path(state: "new-user") %>" class="crayons-btn" aria-label="Create new account">
+<a href="<%= sign_up_path(state: "new-user") %>" class="crayons-btn" aria-label="Create new account" data-no-instant>
```
Zhao-Andy, line 1: This is the point of contention, described in a video: https://imgur.com/a/diTOh8g

vaidehijoshi (approved):
I personally think that removing instant click params is okay to do here (and it works as expected for me locally!) -- but I _also_ think that we should figure out why the instant click `i=i` query params are affecting CSS. Seems possible that, down the road, the CSS could break because of a similar scenario and then we'd have to start checking for params/stripping them out one by one, which won't be scalable.

jacobherrington (approved):
I think it is okay to remove for the sake of fixing an immediate bug, but I agree with @vaidehijoshi, we need to know why this is an issue.

If we figure that out, we could potentially find a solution that preserves the benefits of InstantClick.

PR #11045, opened 2020-10-23
Copy username to search_fields

### LINKED ISSUE #10967: Search doesn't find person with exact username
**Describe the bug**
Searching with exact username does not find the person.

**To Reproduce**
1. Go to dev.to
2. Search for "davidkpiano"
3. Switch to "people" to enable the filter (url: https://dev.to/search?q=davidkpiano&filters=class_name:User)
4. "No results match that query"

**Expected behavior**
Show link to https://dev.to/davidkpiano

**Screenshots**
image

- OS, version: Ubuntu 20.04
- Browser, version: Firefox 81.0.2

krzysiek1507:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR enabled searching a user by username with wildcard. E.g. when a user is `krzysiek1507`, it finds by the whole username but also by a part of it: `krzysiek`, `krzysiek15`. Since the pattern is: `word*`, it won't find by a substring from the middle or the end of the username.

Closes #10967

1. Try to find by a username (NOTE: to fail it, the username cannot be equal or be part of first or last name)
2. Run:
```ruby
Search::User.update_mappings
User.all.each(&:touch) # if your db is small
```
3. Search again for the username

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

1. Mappings for User have to be updated.
2. Users have to be reindexed.

### DIFF
--- a/lib/data_update_scripts/20201030134117_reindex_users_for_username_search.rb
+++ b/lib/data_update_scripts/20201030134117_reindex_users_for_username_search.rb
@@ -1,0 +1,9 @@
+module DataUpdateScripts
+  class ReindexUsersForUsernameSearch
+    def run
+      User.select(:id).in_batches(of: 200) do |batch|
+        Search::BulkIndexWorker.set(queue: :default).perform_async("User", batch.ids)
+      end
+    end
+  end
+end
--- a/config/elasticsearch/mappings/users.json
+++ b/config/elasticsearch/mappings/users.json
@@ -58,7 +58,8 @@
       "type": "text"
     },
     "username": {
-      "type": "keyword"
+      "type": "keyword",
+      "copy_to": "search_fields"
     },
     "profile_fields": {
       "type": "nested",
--- a/lib/data_update_scripts/20200914042434_reindex_users_for_profiles.rb
+++ b/lib/data_update_scripts/20200914042434_reindex_users_for_profiles.rb
@@ -1,9 +1,9 @@
 module DataUpdateScripts
   class ReindexUsersForProfiles
     def run
-      User.select(:id).in_batches(of: 200) do |batch|
-        Search::BulkIndexWorker.set(queue: :default).perform_async("User", batch.ids)
-      end
+      # User.select(:id).in_batches(of: 200) do |batch|
+      #   Search::BulkIndexWorker.set(queue: :default).perform_async("User", batch.ids)
+      # end
     end
   end
 end
--- a/spec/lib/data_update_scripts/reindex_users_for_username_search_spec.rb
+++ b/spec/lib/data_update_scripts/reindex_users_for_username_search_spec.rb
@@ -1,0 +1,12 @@
+require "rails_helper"
+require Rails.root.join("lib/data_update_scripts/20201030134117_reindex_users_for_username_search.rb")
+
+describe DataUpdateScripts::ReindexUsersForUsernameSearch do
+  let(:user) { create :user }
+
+  it "reindexes users" do
+    sidekiq_assert_enqueued_with(job: Search::BulkIndexWorker, args: ["User", [user.id]], queue: "default") do
+      described_class.new.run
+    end
+  end
+end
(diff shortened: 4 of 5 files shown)


mstruve (commented):
Thanks for the P!

In order for this to work, we will need to reindex all of our Users via a DataUpdateScript. [This PR has an example of what you want to do](https://github.com/forem/forem/pull/10313/files#diff-67c8ba8c8c362bb2488ed4b089fcfbf4a47a1090c6ab173f5b1eb4de5f729895R1).

As for a spec, I would [add a spec here that searches for a username and asserts that the correct document is returned](https://github.com/forem/forem/blob/master/spec/services/search/user_spec.rb#L22).

krzysiek1507:
Sure! I'll add the later!

Should I do anything with running the update script or you will handle after the deploy?

mstruve:
> Should I do anything with running the update script or you will handle after the deploy?

It runs automatically! [Here are the docs on them which explain how they work](https://docs.forem.com/backend/data-update-scripts/).

krzysiek1507:
Test added!

mstruve:
Great! As far as the data update script you need to add that as well. I wasn't quite clear about that in my first comment. Sorry about that. Check out the docs I linked above, those will guide you through how to do it and [this PR has an example which you could probably just use here](https://github.com/forem/forem/pull/10313/files#diff-67c8ba8c8c362bb2488ed4b089fcfbf4a47a1090c6ab173f5b1eb4de5f729895R1)

krzysiek1507:
Hi @mstruve. I added the update script but I have 2 questions:
1. Should I add a test for it?
2. Do I need to run `Search::User.update_mappings` in this script? https://github.com/forem/forem/pull/11045/files#diff-2042cc5083b623df97f86d8c60904b49ca4647d58c2e5391782ae7740c50c9f2R4

lib/data_update_scripts/20201030134117_reindex_users_for_username_search.rb
```diff
@@ -0,0 +1,11 @@
+module DataUpdateScripts
+  class ReindexUsersForUsernameSearch
+    def run
+      Search::User.update_mappings
```
mstruve, line 4: No need to run this, this is run by our deploy processes so it is done well before any of the code hits production servers.

mstruve (commented):
1. Yeah, never hurts to add a little test. Checkout the other tests for the data update scripts to get an idea of how to put it to gether. 
2. No need to run update_mappings! Left a comment on that. 

Since we are reindexing all users can you go into the `DataUpdateScripts::ReindexUsersForProfiles` which does the same thing and comment out the run body like this? The reason we do this is bc there is not need for someone who is behind on their data update scripts to run the same script twice. However, we like to keep the script class bc it allows us to see the history of the scripts and match the database entry to an actual class. 
```
module DataUpdateScripts
  class ReindexUsersForProfiles
    def run
      # User.select(:id).in_batches(of: 200) do |batch|
      #   Search::BulkIndexWorker.set(queue: :default).perform_async("User", batch.ids)
      # end
    end
  end
end
```

lib/data_update_scripts/20201030134117_reindex_users_for_username_search.rb
krzysiek1507: Removed!

krzysiek1507:
Hi @mstruve. Updated! I amended the second to last commit automatically. Sorry!

rhymes:
Hi @krzysiek1507 I merged master in that contains a fix that should deal with failing specs

krzysiek1507:
Thanks!

mstruve (approved):
Fantastic!!!! 🚀

atsmith813 (approved):

krzysiek1507:
Thanks!

PR #11047, opened 2020-10-23
Fix NoMethodError for tag moderators channel

jacobherrington:
Secondarily, it adds this channel to the seeds file, since we pretty
much expect it to exist.

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In development, it is possible to hit a NoMethodError when adding a tag
moderator because the "tag-moderator" slug is hard-coded.

This change helps mitigate that by adding a safe navigation operator. We
can always remove it if we decide to get rid of the hard-coded value.

Another solution is creating the channel when it doesn't exist, but I'm not sure we should force that choice on a Forem. If anyone thinks that's fine, we can swap to that approach.

I added a regression test to detect if this bug reappears, but we can cut it out if anyone thinks that is overkill.

1. Navigate to `/admin/tags`
2. Edit a tag > add a tag moderator
3. If the tag moderators channel doesn't exist you'll get a NoMethodError

## Added tests?

- [x] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/labor/assign_tag_moderator.rb
+++ b/app/labor/assign_tag_moderator.rb
@@ -25,12 +25,14 @@ module AssignTagModerator
   end
 
   def self.add_to_chat_channels(user, tag)
-    channels = user.chat_channels
+    user_channels = user.chat_channels
 
-    ChatChannel.find_by(slug: "tag-moderators").add_users(user) unless channels.exists?(slug: "tag-moderators")
+    ChatChannel.find_by(slug: "tag-moderators")&.add_users(user) unless
+      user_channels.exists?(slug: "tag-moderators")
 
     if tag.mod_chat_channel_id
-      ChatChannel.find(tag.mod_chat_channel_id).add_users(user) unless channels.exists?(id: tag.mod_chat_channel_id)
+      ChatChannel.find(tag.mod_chat_channel_id).add_users(user) unless
+        user_channels.exists?(id: tag.mod_chat_channel_id)
     else
       channel = ChatChannels::CreateWithUsers.call(
         users: ([user] + User.with_role(:mod_relations_admin)).flatten.uniq,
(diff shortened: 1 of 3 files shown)


spec/labor/assign_tag_moderator_spec.rb
```diff
@@ -6,40 +6,64 @@
   let(:mod_relator) { create(:user) }
   let(:tag_one) { create(:tag) }
   let(:tag_two) { create(:tag) }
+  let!(:channel) do
+    ChatChannel.create(slug: "tag-moderators",
```
mstruve, line 10: Can we use a factory object here instead?

app/labor/assign_tag_moderator.rb
```diff
@@ -25,12 +25,14 @@ def self.add_tag_moderators(user_ids, tag_ids)
...
 
   def self.add_to_chat_channels(user, tag)
-    channels = user.chat_channels
+    user_channels = user.chat_channels
 
-    ChatChannel.find_by(slug: "tag-moderators").add_users(user) unless channels.exists?(slug: "tag-moderators")
+    ChatChannel.find_by(slug: "tag-moderators")&.add_users(user) unless
+      user_channels.exists?(slug: "tag-moderators")
 
     if tag.mod_chat_channel_id
```
mstruve, line 33: We could add the unless to this if to keep everything in one place. 
```ruby
if tag.mod_chat_channel_id && !user_channels.exists?(id: tag.mod_chat_channel_id)
```

mstruve (approved):
One refactor suggestion and one spec suggestion otherwise LGTM!

Zhao-Andy (approved):
Good catch! Agree with Molly's suggestions.

jacobherrington:
Oops :sweat_smile:

PR #11049, opened 2020-10-23
List single_resource_admin resources in the UI

jacobherrington:

- [ ] Refactor
- [x] Feature ?
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This seems like important information to be showing admins.

Before:
image

After:
image

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -33,6 +33,9 @@ class Role < ApplicationRecord
 
   scopify
 
+  # Returns a somewhat friendly name for the resource related to a given role.
+  # In the case of Tag Moderators, a resource_type is not present, so we use the
+  # resource_id to grab the specific Tag related to that moderator's role.
   def resource_name
     return resource_type unless resource_id
 
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -32,4 +32,10 @@ class Role < ApplicationRecord
             inclusion: { in: ROLES }
 
   scopify
+
+  def resource_name
+    return resource_type unless resource_id
+
+    Tag.find(resource_id).name
+  end
 end
--- a/app/views/admin/users/_activity.html.erb
+++ b/app/views/admin/users/_activity.html.erb
@@ -15,7 +15,7 @@
     <p>(view full history in notes below)</p>
     <ul>
       <% @user.roles.each do |role| %>
-        <li><%= role.name %> <em><%= "- #{Tag.find(role.resource_id).name}" if role.resource_id.present? %></em></li>
+        <li><%= role.name %> <em><%= role.resource_name.to_s %></em></li>
       <% end %>
     </ul>
   </div>


app/views/admin/users/_activity.html.erb
```diff
@@ -15,7 +15,7 @@
     <p>(view full history in notes below)</p>
     <ul>
       <% @user.roles.each do |role| %>
-        <li><%= role.name %> <em><%= "- #{Tag.find(role.resource_id).name}" if role.resource_id.present? %></em></li>
+        <li><%= role.name %> <em><%= role.resource_name.to_s %></em></li>
```
juliannatetreault, line 18: _**Quick, non-blocking thought**_: I could foresee a scenario in which someone is assigned to be a tag mod for a handful of tags (10+) in addition to any roles that they currently have, making this roles section a bit long. My local user has 3 admin-related roles and is a tag mod for 5 different tags and the list seems not too far from the "bit too long" scenario. This being said, it might be nice to paginate this section (or have a roles dropdown -- something that doesn't hide the role history but doesn't make the page super long) at some point.

juliannatetreault (approved):
I left a quick, non-blocking thought, but that aside, this works for me and is a nice touch to the "Current Roles" section! 🚀

app/views/admin/users/_activity.html.erb
jacobherrington: For sure, I'll wait for the designers to get to this page before I attempt anything though :laughing:

app/models/role.rb
```diff
@@ -32,4 +32,10 @@ class Role < ApplicationRecord
             inclusion: { in: ROLES }
 
   scopify
+
+  def resource_name
+    return resource_type unless resource_id
+
+    Tag.find(resource_id).name
```
rhymes, line 39: Can you explain this method? I don't get it 😭 
We return the resource type if there is no resource id but if there is a resource id we return the tag name? But `resource_type` can be:

```ruby
[1] pry(main)> Rolify.resource_types
=> ["LiquidTags::UserSubscriptionTag", "User"]
```

I tried to read this a couple of times out loud but it's still not clear to me, why do we load tags if it can be anything else?

jacobherrington: It's been a couple of days since I wrote this, so I have to think! :laughing: 

My thought was that many Roles have a `resource_type` (specifically, I was thinking of `single_resource_admin`s), and that type constitutes a resource "name": 

```ruby
# In the case of a single resource admin, where resource_id is always nil
# but resource_type indicates which model is associated with this admin
User.last.add_role(:single_resource_admin, Article)
User.last.roles.first.resource_name #=> "Article"

# In the case of a tag moderator (using one I made via the UI)
u.roles.last.resource_name #=> "python"
```

It seems like I may have missed a use case? Or does this make sense to you?

rhymes: Got it! My suggestion then is to add it this explainer as a comment if possible :)

jacobherrington: Comment added!

rhymes (approved):
LGTM

PR #11055, opened 2020-10-24
Add following_tags_count to release_adjusted_cache_key

juliannatetreault:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR fixes a caching bug that was introduced in PR #11022, as pointed out by a community member (thank you so much!). Please see this comment for further information: https://github.com/forem/forem/pull/11022#issuecomment-716046885

Relates to PR #11022 

To QA this change, please ensure that "Explore" and "My Tags" properly render in the side navbar, that all checks pass, and that Travis builds green. I'm open (as always 😄 ) to any ideas for how to better test these changes!

## Added tests?

- [ ] yes
- [x] no, because they aren't needed -- there are already tests for this part of the code
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

N/A

alt_text

### DIFF
--- a/app/views/articles/_sidebar.html.erb
+++ b/app/views/articles/_sidebar.html.erb
@@ -1,7 +1,7 @@
 <div id="sidebar-wrapper-left" class="sidebar-wrapper sidebar-wrapper-left crayons-layout__sidebar-left">
   <div class="sidebar-bg" id="sidebar-bg-left"></div>
   <aside class="side-bar" aria-label="Primary sidebar">
-    <% cache(release_adjusted_cache_key("main-sidebar-nav--#{user_signed_in?}"), expires_in: 15.minutes) do %>
+    <% cache(release_adjusted_cache_key("main-sidebar-nav--#{user_signed_in?}--#{current_user&.following_tags_count&.positive?}"), expires_in: 15.minutes) do %>
       <%= render "articles/sidebar_nav" %>
       <% @sponsorships = Sponsorship.gold.live.includes(:organization).order(featured_number: :asc) %>
       <%# the pattern .present?/.each has the advantage of issuing only a single SQL query to load objects in memory %>
--- a/app/views/articles/index.html.erb
+++ b/app/views/articles/index.html.erb
@@ -26,7 +26,7 @@
 
 <%= javascript_packs_with_chunks_tag "homePage", defer: true %>
 
-<% cache(release_adjusted_cache_key("main-stories-index-#{params}-#{user_signed_in?}"), expires_in: 90.seconds) do %>
+<% cache(release_adjusted_cache_key("main-stories-index-#{params}-#{user_signed_in?}-#{current_user&.following_tags_count&.positive?}"), expires_in: 90.seconds) do %>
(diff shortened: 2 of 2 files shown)


mstruve (approved):
LGTM!

app/views/articles/_sidebar.html.erb
```diff
@@ -1,7 +1,7 @@
 <div id="sidebar-wrapper-left" class="sidebar-wrapper sidebar-wrapper-left crayons-layout__sidebar-left">
   <div class="sidebar-bg" id="sidebar-bg-left"></div>
   <aside class="side-bar" aria-label="Primary sidebar">
-    <% cache(release_adjusted_cache_key("main-sidebar-nav--#{user_signed_in?}"), expires_in: 15.minutes) do %>
+    <% cache(release_adjusted_cache_key("main-sidebar-nav--#{user_signed_in?}--#{current_user&.following_tags_count&.positive?}"), expires_in: 15.minutes) do %>
```
juliannatetreault, line 4: I actually had this in my initial implementation of this, but after speaking with another teammate, decided that it was probably overkill... the more you know! 😄

jacobherrington (approved):
:+1:

vaidehijoshi:
> no, because they aren't needed -- there are already tests for this part of the code

I'm curious whether we know why our pre-existing tests didn't catch this? It seems like if we have tests for this, they may not have quite enough surface area/coverage since they all passed in the initial PR.

juliannatetreault:
> > no, because they aren't needed -- there are already tests for this part of the code
> 
> I'm curious whether we know why our pre-existing tests didn't catch this? It seems like if we have tests for this, they may not have quite enough surface area/coverage since they all passed in the initial PR.

@vaidehijoshi It should really read: "...there are already tests for this feature." The preexisting tests didn't catch this because this was a caching issue that occurred on production and not on `localhost`--the feature wasn't caching in the same way locally. It looks like we have tests that test that the `#release_adjusted_cache_key` works properly, which I believe should cover the changes I made. If you feel like there should be additional test coverage for this, then I'm more than open to ideas for how to better test the caching and more than happy to write them. 😊

vaidehijoshi:
> The preexisting tests didn't catch this because the tests were written for the feature and not the possibility of caching

I took a deeper look at those `release_adjusted_cache_key` specs and figured out that we _do_ have specs to capture the caching part of this..._however_ [they don't test the `path` part of the cache key's value](https://github.com/forem/forem/blob/5457f0c0b702af71cc864a0fbdd065238760b7f5/spec/helpers/application_helper_spec.rb#L38-L61); instead, they just test the params, release footprint, and site config part of the cache key.

All of this is to say, it makes much more sense to me now how we didn't catch this -- because the `path` part of the cache key wasn't ever really tested before (I don't think we have great test infrastructure for actually testing any of our cache key _values_ unfortunately. 😅)

vaidehijoshi (approved):

PR #11056, opened 2020-10-24
Extract subtitles to a constant to reduce allocations

krzysiek1507:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Subtitles don't change so we can move them to a constant and save some memory (if you need stats I can provide them) because this hash and strings are created every time `title_with_timeframe` is called (which is 3 times on https://github.com/forem/forem/blob/467e99556cf1fc088bbd91ec0612d4a64ff91ffd/app/views/articles/index.html.erb).

This PR reduces the number of creations of this hash from 3 * number of calls of `render "articles/index"` to 1. (Similar on `tags/_meta`).

Check if the title contains values from the hash.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -13,6 +13,14 @@ module ApplicationHelper
 
   LARGE_USERBASE_THRESHOLD = 1000
 
+  SUBTITLES = {
+    "week" => "Top posts this week",
+    "month" => "Top posts this month",
+    "year" => "Top posts this year",
+    "infinity" => "All posts",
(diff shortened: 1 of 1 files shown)


citizen428 (approved):
Nice first contribution, thanks @krzysiek1507 👏

rhymes (approved):
Love it, thanks!

PR #11080, opened 2020-10-26
Replacing old button with crayon buttons to make it more visible

Rafi993:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Replaces old button with crayon button it also improves button visibility

Closes https://github.com/forem/forem/issues/11081

### Before
old_button
### After
visible_button

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

### DIFF
--- a/app/views/admin/mods/index.html.erb
+++ b/app/views/admin/mods/index.html.erb
@@ -48,7 +48,7 @@
           <% if params[:state] == "potential" %>
             <td>
               <%= form_with model: [:admin, mod], url: admin_mod_path(mod.id), method: :patch, local: true do |f| %>
-                <%= f.submit "Make Trusted", class: "btn btn-light js-add-to-mod-channel" %>
+                <%= f.submit "Make Trusted", class: "crayons-btn js-add-to-mod-channel" %>
               <% end %>
             </td>
           <% end %>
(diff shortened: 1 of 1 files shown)


app/views/admin/mods/index.html.erb
```diff
@@ -48,7 +48,7 @@
           <% if params[:state] == "potential" %>
             <td>
               <%= form_with model: [:admin, mod], url: admin_mod_path(mod.id), method: :patch, local: true do |f| %>
-                <%= f.submit "Make Trusted", class: "btn btn-light js-add-to-mod-channel" %>
+                <%= f.submit "Make Trusted", class: "crayons-btn js-add-to-mod-channel" %>
```
ludwiczakpawel, line 51:
```suggestion
                <%= f.submit "Make Trusted", class: "crayons-btn crayons-btn--secondary js-add-to-mod-channel" %>
```

I think this button shouldn't be "prmiary" one (purple) but secondary. Primary button should be "reserved" for the most important action on a page (or within certain area). 

I do agree that the button you have fixed originally could be a bit more visible and I think `crayons-btn--secondary` will be a better replacement in this case.

ludwiczakpawel (changes requested):

app/views/admin/mods/index.html.erb
Rafi993: secondary_button

This is how it looks after the above change

ludwiczakpawel (approved):
🙌

rhymes (approved):

PR #11087, opened 2020-10-26
Flaky Spec Fix:Remove js:true From Local Timestamp Checks and js Fix attempt code

mstruve:
- [x] Spec Fix

## Description
Ok, these 2 specs have really been a trip for me. I have tried for days to fix them and get them to behave but they are still flaky. Rather than sink more time into it I have decided to remove the `js:true` component and take out the locale timestamp formatting from the spec. I do not know what the best way to test that is but I hope we get a QA person soon that does which is why I put in the TODO for both specs. 

I choose to only remove the `js:true` part rather than the entire specs bc I do think its valuable to test that the published date is at least present on the page even if we cant test its time zone accurately. If other's feel like the specs should be commented out altogether I am open to that as well. Let me know what you all think!

## Added tests?
- [x] updated

alt_text


rhymes (approved):
Sounds like a plan, thanks for investigating!

vaidehijoshi (approved):
Thanks for leaving these `TODOs` with super clear explanations! Also sorry that this still ended up being so flaky 😓

PR #11090, opened 2020-10-26
[Team Email Login] Refactor the Email Authentication flow

msarit:

- [X] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Per [design specs](https://user-images.githubusercontent.com/816402/96673429-58375c80-131b-11eb-848a-0ccacec0e0cf.jpg), this PR refactors the process of enabling/disabling Email-Password Authentication for Forems.

## Related Issues
- https://github.com/forem/InternalProjectPlanning/issues/166
- https://github.com/forem/InternalProjectPlanning/issues/169

🎥  **A video demo appears below these instructions**

1. Navigate to `localhost:xxxx/admin/config`. Ensure that your user account has the `single_resource_admin` role. To add this role, run `User.find(<your-user-id>).add_role("single_resource_admin").save!` in your rails console.
2. Toggle open the `Authentication` section. Click the `Enable` or `Edit` button to the right of the `Email address` label (the button's text depends on whether email auth has been enabled for the forem). The button should disappear, and a section containing the reCAPTCHA fields, a `Disable` button and a `Cancel` button should appear below.
3. Run through the various flows for enabling/disabling email authentication. A few things to verify:
- That disabling email authentication also turns Invite-Only Mode off (its checkbox should be unchecked).
- That disabling email authentication removes the option of email sign in at `localhost:xxxx/enter` and removes the option of email registration at `localhost:xxxx/enter?state=new-user`.
- That the Invite-Only Mode checkbox is disabled when Email Auth is turned off.
- That the option to turn Email Auth off is absent when Invite-Only Mode is enabled.
---

https://www.loom.com/share/666d3ca2832045d5a3b409d57e2a7235

## Added tests?

- [X] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [X] no documentation needed

No

Dancing Pumpkin


app/controllers/admin/configs_controller.rb
```diff
@@ -115,8 +115,7 @@ class ConfigsController < Admin::ApplicationController
         facebook_key
         facebook_secret
         invite_only_mode
-        allow_email_password_registration
-        allow_email_password_login
+        allow_both_email_signup_and_login
```
msarit, line 118: Seeing as we are no longer passing the `allow_email_password_registration` and `allow_email_password_login` through params, I deleted them from ALLOWED_PARAMS

app/controllers/admin/configs_controller.rb
```diff
@@ -208,6 +209,17 @@ def clean_up_params
       config[:credit_prices_in_cents]&.transform_values!(&:to_i)
     end
 
+    def toggle_email_password_authentication
```
msarit, line 212: This method updates the email-registration and email-login fields; I also update the invite-only field according to the [design specs](https://user-images.githubusercontent.com/816402/96673429-58375c80-131b-11eb-848a-0ccacec0e0cf.jpg), which indicate that turning email-auth off should also disable `invite-only` mode.

app/javascript/admin/controllers/config_controller.js
```diff
@@ -1,9 +1,64 @@
...
+  '#email-auth-settings-section',
+);
+const modalAnchor = document.querySelector('.admin-config-modal-anchor');
+const emailAuthModalTitle = 'Disable email address registration';
+const emailAuthModalBody =
+  '<p>If you disable email address as a registration option, people cannot create an account with their email address.</p><br /><p>However, people who have already created an account using their email address can continue to login.</p>';
+const emailAuthModalBodyInviteOnly =
+  '<p>If you disable email address as a registration option, people cannot create an account with their email address. This will also disable <strong>invite-only mode</strong>.</p><br /><p>However, people who have already created an account using their email address can continue to login.</p>';
+
+const adminConfigModal = (
```
msarit, line 17: So I am not sure of my implementation of the modal will pass QA 🙈  but let me explain my thinking:

- I thought to make the modal implementation as general as possible, so that if other refactors of the Admin/Config page require a modal, we can use this one. I basically make all variable modal content passable as parameters.
- Regarding the text content passed into the modal, I thought to use the `Constants::SiteConfig::DETAILS` file that we have on the Rails side (and define the text content in that file), but couldn't figure out how to read the file.

app/lib/constants/site_config.rb
```diff
@@ -4,14 +4,6 @@ module SiteConfig
     SVG_PLACEHOLDER = "<svg ...></svg>".freeze
 
     DETAILS = {
-      allow_email_password_registration: {
```
msarit, line 7: Deleted because we no longer have these fields in the `config` show page

app/views/admin/configs/show.html.erb
```diff
@@ -175,30 +178,27 @@
...
+                    <div style="display: flex; align-content: center; justify-content: space-between; align-items: center;">
+                      <div style="display: flex; align-content: center; justify-content: space-between; align-items: center;">
+                        <p class="config-authentication__item--label fw-medium">
+                          Email address
+                        </p>
+                      </div>
+                      <button
+                        class="crayons-btn crayons-btn--secondary"
+                        data-button-text="<%= SiteConfig.allow_both_email_signup_and_login ? "edit" : "enable" %>"
+                        data-target="config.emailAuthSettingsBtn"
```
msarit, line 190: In order to fulfill the design specs requirement of a button for enabling/disabling Email Auth, I employ the hidden `allow_both_email_signup_and_login` field to pass as a param, which then updates the fields actually relevant to email auth in the controller.

I did consider sticking with the already-existing checkboxes for `allow_email_password_registration` and `allow_email_password_login`, and simply hiding those, but I decided my implementation was cleaner.

app/views/admin/configs/show.html.erb
```diff
@@ -1315,4 +1327,6 @@
       </div>
     </div>
   </div>
+  <%#= render partial: "modal" %>
+  <div class="admin-config-modal-anchor"></div>
```
msarit, line 1331: This is where the `adminConfigModal` is anchored when activated

app/views/shared/authentication/_providers_registration_form.html.erb
```diff
@@ -8,7 +8,7 @@
       <%= params[:state] == "new-user" ? "Sign up" : "Continue" %> with <%= provider.official_name %>
     </a>
   <% end %>
-  <% if params[:state] == "new-user" && SiteConfig.allow_email_password_registration %>
+  <% if params[:state] == "new-user" && SiteConfig.allow_email_password_registration && !SiteConfig.invite_only_mode %>
```
msarit, line 11: While QAing my work, I noticed that when `invite-only mode` was turned on, the Sign Up page still had the Sign up with Email option. So I updated this if-clause to correct this.

vaidehijoshi (changes requested):
Hi @msarit! I'm trying to work through your QA instructions but I'm running into trouble. Let me know if I am missing something. I wanted to test the button text as well as this:
> That disabling email authentication also turns Invite-Only Mode off (its checkbox should be unchecked).

In order to do that, it seems like I need to turn _on_ invite-only mode. But I can't seem to do that through the UI anymore so I can't continue through your QA instructions yet. 😅

Here's what I did:
Screen_Shot_2020-10-26_at_10_45_26_AM
Screen_Shot_2020-10-26_at_10_45_40_AM
Screen_Shot_2020-10-26_at_10_45_46_AM

I have confirmed that this _does_ work on `master`, so I _think_ that something about these changes might have impacted this? Let me know if you are experiencing the same.

msarit:
@vaidehijoshi Great catch! Let me investigate

msarit:
I caught the reason for the bug @vaidehijoshi reported above. I'm changing this PR's status to `draft` while I work that bug out. Thanks! 🙏

lisasy:
Hello @msarit , fantastic job so far!

I have a question with the "Disable" button that shows up when someone first presses "Enable" for the first time. At this stage, `Email address` won't actually be enabled until they "Update Site Configuration" below, correct?

If this is the case, then I think it's misleading to have a button with the text label "Disable", as well as its accompanying interstitial. However, those 2 elements are apart of the spec for when someone *already has Email address enabled* and is attempting to disable it.

My recommendations here would be:
1. Change "Disable" label to "Undo"
2. Don't show interstitial dialog
3. Make it so that only what you have *right now* happens when the enabled Email address is set in the background.

Is this clear? This flow is basically the top row here: https://user-images.githubusercontent.com/816402/96673429-58375c80-131b-11eb-848a-0ccacec0e0cf.jpg

As you can see, there is no dialog.

app/assets/stylesheets/admin.scss
```diff
@@ -125,3 +125,21 @@ label {
   flex-direction: column;
   gap: var(--su-4);
 }
+
+.admin-config-checkmark {
```
msarit, line 129: I added these classes to reproduce the green `Enabled` checkmark, and to render the `not-allowed` cursor when the user hovers over the disabled `Invite Only Mode` checkbox.

app/views/admin/configs/show.html.erb
```diff
@@ -175,30 +178,33 @@
                     <%= inline_svg_tag("email.svg", class: "crayons-icon", aria: true, title: "Twitter logo") %>
                   </figure>
                   <div class="config-authentication__item--container">
-                    <div class="config-authentication__item--label">
-                      <p class="fw-medium">
-                        Email address and password
-                      </p>
+                    <div style="display: flex; justify-content: space-between; align-items: baseline;">
```
msarit, line 181: @lisasy I need help locating a class that covers the CSS properties in this `<style>` tag here 🙏

joshpuetz:
I have a bunch of issues around the 'undo button. It's quite confusing to me: when email address auth is enabled, the 'undo' button disables it: the label should be 'disable' or something like that.

The positioning of the button is odd to me as well: it's expect it to be up near where the 'enabled' button is located.

Finally: there's no indication that I need to submit the form after pressing 'undo': I can press it multiple times, and I'm not sure what will happen unless I scroll down and submit the form.

Thoughts @lisasy ?

app/views/admin/configs/show.html.erb
```diff
@@ -234,11 +224,59 @@
...
                           </div>
                         </div>
+                        <div class="mt-6">
+                          <% if SiteConfig.allow_both_email_signup_and_login %>
+                            <button
+                              class="crayons-btn crayons-btn--danger"
+                              data-action="click->config#activateEmailAuthModal"
+                              <%= SiteConfig.invite_only_mode ? 'disabled' : ''%>>
+                              <% if SiteConfig.invite_only_mode %>
+                                <span
```
msarit, line 234: @lisasy @ludwiczakpawel This spanned `tooltip` is NOT working on the frontend; I believe I have implemented it [as instructed in this post](https://forem.team/picasso/crayons-update-episode-1-39ah). Please help me debug 🙏

app/views/admin/configs/show.html.erb
```diff
@@ -234,11 +224,59 @@
...
+                              Undo
+                            </button>
+                          <% end %>
+                        </div>
                       </fieldset>
                     </div>
                   </div>
                 </section>
 
+                <div class="form-group">
```
msarit, line 263: I moved the 3rd-party Authentication Dropdown below the Email Authentication block, to place it closer to the associated key fields @joshpuetz

app/javascript/admin/controllers/config_controller.js
joshpuetz: Ah, another modal dialog box! I wonder if there's a way we could use the non-admin modal for this. In any case: looks good!

app/views/admin/configs/show.html.erb
joshpuetz: I agree with this approach: it's clear in the controller that it's translating this value into database values.

app/views/admin/configs/show.html.erb
```diff
@@ -1328,4 +1366,6 @@
       </div>
     </div>
   </div>
+  <%#= render partial: "modal" %>
```
joshpuetz, line 1369: Suggestion: remove this comment. There's not actually a partial being rendered here: a dom node is being inserted by javascript!

joshpuetz (changes requested):

app/views/admin/configs/show.html.erb
msarit: Oops! Will do!

msarit: @joshpuetz done!

joshpuetz (approved):
Looking good to me!

app/assets/stylesheets/admin.scss
```diff
@@ -118,10 +118,34 @@ label {
     padding-top: var(--su-4);
     padding-bottom: var(--su-4);
   }
+
+  &--align {
+    display: flex;
+    justify-content: space-between;
+    align-items: baseline;
```
vaidehijoshi, line 125: ICYMI, you can also use the utility classes here as well! ([here's what you could use](https://storybook.forem.com/?path=/story/5-css-utility-classes-align-items--items-baseline) instead of `align-items: baseline;`

app/assets/stylesheets/admin.scss
```diff
@@ -118,10 +118,34 @@ label {
...
 }
 
 .config-authentication-form {
   display: flex;
   flex-direction: column;
   gap: var(--su-4);
 }
+
+.admin-config-checkmark {
+  opacity: 100;
```
vaidehijoshi, line 136:
```suggestion

```
I don't think you need this style

app/assets/stylesheets/admin.scss
```diff
@@ -118,10 +118,34 @@ label {
...
+  width: var(--su-5);
+  height: var(--su-5);
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
```
vaidehijoshi, line 141:
```suggestion

```
I don't think you need these, either

app/assets/stylesheets/admin.scss
```diff
@@ -118,10 +118,34 @@ label {
...
   display: flex;
   flex-direction: column;
   gap: var(--su-4);
 }
+
+.admin-config-checkmark {
+  opacity: 100;
+  width: var(--su-5);
+  height: var(--su-5);
+  display: inline-flex;
```
vaidehijoshi, line 139:
```suggestion

```
I don't think you need this one

app/assets/stylesheets/admin.scss
```diff
@@ -118,10 +118,34 @@ label {
...
+  align-items: center;
+  justify-content: center;
+  background-color: var(--accent-success);
+  border-radius: 100%;
+  border: 2px solid var(--base-inverted);
+  padding: 2px;
```
vaidehijoshi, line 145: Is it possible to use [a crayons variable](https://storybook.forem.com/?path=/story/5-css-utility-classes-padding--p-1) here instead of `2px`? I know we are trying to use variables rather than hardcoding values/magic numbers unless we have a good reason not to.

Maybe @lisasy knows which variable (`var(--su-1)`) would work best? (Although, personally, I think the icon is easier to read without _any_ padding 😛)

app/views/admin/configs/show.html.erb
```diff
@@ -234,11 +224,59 @@
...
+                        <div class="mt-6">
+                          <% if SiteConfig.allow_both_email_signup_and_login %>
+                            <button
+                              class="crayons-btn crayons-btn--danger"
+                              data-action="click->config#activateEmailAuthModal"
+                              <%= SiteConfig.invite_only_mode ? 'disabled' : ''%>>
+                              <% if SiteConfig.invite_only_mode || authentication_enabled_providers.none? %>
+                                <span
+                                  class="crayons"
+                                  data-tooltip="<%= Constants::SiteConfig::DETAILS[:invite_only_mode][:cannot_disable_email] %>">
```
vaidehijoshi, line 236: Does this need to actually live _inside_ of SiteConfig constant? I think we can just inline it here. I'm suggesting this because I think it adds more complexity -- for example, I had to figure out what "this" mean in the "You cannot do this" text. Then I scrolled to this point in the file and figured out that the text was describing whether or not someone could click on the "disable" button, which was not clear to me. I'm also suggesting it because the rest of the SiteConfig values only include the `description` and `placeholder`, so adding a `cannot_disable_email` seems a bit odd, IMO. And because this is used by the `data-tooltip`, I think it it's okay to put it inline.
```suggestion
                                  data-tooltip="You cannot do this until you enable at least one other registration option.">
```

app/views/admin/configs/show.html.erb
```diff
@@ -234,11 +224,59 @@
...
                         </div>
+                        <div class="mt-6">
+                          <% if SiteConfig.allow_both_email_signup_and_login %>
+                            <button
+                              class="crayons-btn crayons-btn--danger"
+                              data-action="click->config#activateEmailAuthModal"
+                              <%= SiteConfig.invite_only_mode ? 'disabled' : ''%>>
+                              <% if SiteConfig.invite_only_mode || authentication_enabled_providers.none? %>
+                                <span
+                                  class="crayons"
```
vaidehijoshi, line 235:
```suggestion
                                  class="crayons-tooltip"
```
I think the class you specified here is incorrect, which is why it wasn't rendering :)

Here's what it looks like once you have `crayons-tooltip` as the class:
Screen Shot 2020-10-27 at 4 36 47 PM
As you can see, the styling is pretty wonky -- I think you'll need to adjust that so that it renders properly 😅

app/views/admin/configs/show.html.erb
vaidehijoshi: I think that post has the wrong class in the example, which is why it wasn't working for you! I did some digging and found [this example](https://github.com/forem/forem/blob/125584d3541c627683e84fdad5cb4112d645d33b/app/views/pages/crayons.html.erb#L72) from our codebase, which is the right class :)

vaidehijoshi:
I realize that this is following the design mocks, but I wanted to flag something that seemed a bit odd to me about this flow. Maybe it's just me, but the red `Confirm` button felt "destructive" enough to me that, the first time I went through this flow, I assumed that I had actually disabled the feature when I clicked the button. I suspect a Forem creator might make the same mistake/assumption?  (cc @lisasy)
Screen_Shot_2020-10-27_at_4_41_39_PM

app/javascript/admin/controllers/config_controller.js
vaidehijoshi: I didn't investigate this but I'm curious if the regular crayons modal (`app/javascript/crayons/Modal/Modal.jsx`) was usable at all here? It would be ideal to reuse it, like @joshpuetz mentioned.

If we decide that it is not workable to reuse that modal, then I would suggest that ultimately we move the `adminConfigModal` out of this file (and then import it here) so that we can re-use it across the admin sections of the app. I haven't used stimulus js before, but I'd hope that this would Just Work™️. Right now, the `adminConfigModal` is coupled to this specific `config` controller, and if we decide that we can't use the crayons modal, it would be ideal to make this modal as reusable as possible for the admin section of the app.

But I think that could be a separate PR if need be.

vaidehijoshi (commented):
Left a few suggestions and thoughts around cleaning up CSS + moving some things around! I had one larger design question (maybe @lisasy has some thoughts there), and I also figured out why that tooltip wasn't working for ya 😉

app/views/admin/configs/show.html.erb
ludwiczakpawel: @msarit hah, yes.. the class name I entered was wrong 🤦‍♂️🤦‍♂️🤦‍♂️ it should be `crayons-tooltip` instead of just `crayons`... Sorry for that mistake!

Thanks @vaidehijoshi for digging this!

app/views/admin/configs/show.html.erb
ludwiczakpawel: Two options to make it look better. Use whatever you prefer:

1. instead of adding this `span` **inside** button, try to **wrap it** like this:
  ```
  <span class="crayons-tooltip inline-block" data-tooltip="...">
    <button ...>...</button>
  </span>
  ```
2. OR drop the `span` at all, and add tooltip to the `button` like this:
  ```
  <button class="crayons-btn ... crayons-tooltip" data-tooltip="...">...</button>
  ```

Usually I'd go with option 2 and it would work just fine. But because your button is disabled, its opacity is decreased a little (that's the style we apply for disabled buttons). Therefore the tooltip opacity will be decreased as well (since it's attached to that disabled button). And if that bothers you (I personally don't have a problem with that), you can go with option 1 :).

app/javascript/admin/controllers/config_controller.js
msarit: @vaidehijoshi These are awesome points; I've created a note in Team Email Login to address this as a separate PR 👍

app/assets/stylesheets/admin.scss
msarit: @vaidehijoshi I think you'll like it now 😄

app/views/admin/configs/show.html.erb
```diff
@@ -234,11 +228,52 @@
                                              placeholder: Constants::SiteConfig::DETAILS[:recaptcha_secret_key][:placeholder] %>
                           </div>
                         </div>
+                        <div class="mt-6">
+                          <% if SiteConfig.allow_both_email_signup_and_login %>
+                            <button
+                              class="crayons-btn crayons-btn--danger <%= cannot_disable_email_auth ? "crayons-tooltip" : "" %>"
```
vaidehijoshi, line 234: I think this could be a helper in and of itself:
```ruby
def disable_email_tooltip_class
    (SiteConfig.invite_only_mode || authentication_enabled_providers.none?) ? "crayons-tooltip" : ""
end
```

And then you could could just call the helper here and not even have to do any conditional inline checks
```suggestion
                              class="crayons-btn crayons-btn--danger <%= disable_email_tooltip_class %>"
```

Since you're going to the trouble of making a helper, may as well make it do more work for you 😉

app/views/admin/configs/show.html.erb
```diff
@@ -234,11 +228,52 @@
...
+                              data-action="click->config#activateEmailAuthModal"
+                              data-tooltip="<%= cannot_disable_email_auth ? disable_email_auth_tooltip_text : "" %>"
+                              <%= cannot_disable_email_auth ? "disabled" : "" %>>
```
vaidehijoshi, line 237: Could do something similar here, too:
```ruby
def disable_button_class
  (SiteConfig.invite_only_mode || authentication_enabled_providers.none?) ? "disabled" : ""
end
```

```suggestion
                              <%= disabled_button_class %>>
```

If chose to do this, then ultimately, you wouldn't need the `cannot_disable_email_auth` helper, because you'd have two new, more specific helpers determining the class for you.

app/models/site_config.rb
```diff
@@ -26,7 +26,8 @@ class SiteConfig < RailsSettings::Base
 
   # Authentication
   field :allow_email_password_registration, type: :boolean, default: false
-  field :allow_email_password_login, type: :boolean, default: true
+  field :allow_email_password_login, type: :boolean, default: false
```
vaidehijoshi, line 29: I didn't catch this on the first review, but I just noticed this now -- I'm curious [why this is changing](https://github.com/forem/forem/pull/10838/files#diff-68da686a0f7348bf9a2c59e669043f21019ab8dfb6363cea6daa534c5c8296d7R26) to `false`? @forem/creator-onboarding intentionally will need this to be `true` for in order to [ensure that email login defaults to being "on"](https://github.com/forem/InternalProjectPlanning/issues/91) for newly-created forems.

msarit: No worries, I will change it back to `true`. I had it that way to test the feature I was building 😄

vaidehijoshi: Thanks, @msarit! Just for more context for you + @joshpuetz + @lisasy...

@forem/creator-onboarding is making changes to SiteConfig this cycle in order to make the forem creator onboarding process as smooth as possible. So, if you end up needing to make any changes to that file, please do drop us a line 🙏 We're trying our best to keep tabs on all the SiteConfig-related PRs, but there have been a lot this cycle! So if you end up changing any "defaults" in SiteConfig there is a good chance that it will affect our cycle work, and pinging us directly would be super helpful :)

lisasy:
> I realize that this is following the design mocks, but I wanted to flag something that seemed a bit odd to me about this flow. Maybe it's just me, but the red `Confirm` button felt "destructive" enough to me that, the first time I went through this flow, I assumed that I had actually disabled the feature when I clicked the button. I suspect a Forem creator might make the same mistake/assumption? (cc @lisasy)
> Screen_Shot_2020-10-27_at_4_41_39_PM

Sigh. The unfortunate thing about this whole procedure is that there is so much density and information here that it isn't clear to people that they have to take that final action. This PR is one small step in a set of so many others in order to make refinements overall to the UX. A longer term approach is here:

Screen Shot 2020-10-28 at 9 53 42 AM

Something we could do in the shorter-term is add a bolded sentence in that dialog that says something like "**_You must update site config to save this action_**", with the intent to remove it once we implement better UX/UI overall.

vaidehijoshi:
> Something we could do in the shorter-term is add a bolded sentence in that dialog that says something like "You must update site config to save this action", with the intent to remove it once we implement better UX/UI overall.

@lisasy I like this solution as a stop-gap, perhaps with a `TODO` comment as well that we intend to remove it for something more robust.

vaidehijoshi (approved):
Great job working through the comments and addressing feedback, @msarit! 👏

Issue #3613, opened 2019-08-03
When we change theme, we have to refresh the page to update


mbackermann:
**Describe the bug**

When we go to settings and change the theme options and save it, we have to refresh the page so the changes can take place.

**To Reproduce**

Go to settings -> change the theme and save

**Expected behavior**

When we save and the page get refreshed, it should already update your preferences

**Desktop (please complete the following information):**

- OS: Mac OS
- Browser: Chrome 
- Version: 75.0

Link2Twenty:
This will be a lot easier to fix when #2154 is done.

It should just be a bit of JS changing all the css variables.
https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#Values_in_JavaScript

shilpi230:
@rhymes I would like to take this up. Thanks!

rhymes:
@shilpi230 yes, thank you!

citizen428:
@shilpi230 When you fix this, please also take a look at `UsersController#update`, where we specifically set a flash message telling the user to reload the page after the updated their theme:

```ruby
      if config_changed?
        notice = "Your config has been updated. Refresh to see all changes."
      end
```

The `config_changed?` method doesn't have the best name and can be found in the same file:

```ruby
  def config_changed?
    params[:user].include?(:config_theme)
  end
```

If manual refreshes after theme changes are no longer necessary, both the check and the predicate method can be removed.

shilpi230:
Sure @citizen428 Thanks!

shilpi230:
@citizen428 @rhymes The issue over here right now is, to load the user's selected theme we are getting it from localStorage. Which resets when we refresh the page. To solve this we will need to reset the localStorage explicitly.

cmgorton:
Thank you for opening this request.
Since this has been open for awhile without much progress lately I am going to close it here. 
The Forem core team now uses an internal RFC ("request for comments") process to assess and prioritize new features. This process is intended to provide a consistent and standardized path for new changes to enter the Forem ecosystem. If you'd like to propose a new feature, please visit [forem.dev](https://forem.dev/) to start a discussion around a new feature! 🙏

PR #11091, opened 2020-10-26
Change theme after update without manual refreshing

shilpi230:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

https://github.com/forem/forem/issues/3613

- Go to User's Settings page. 
- Navigate to UX tab, select some theme.
- Theme should get updated website wide.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed


Link2Twenty:
I wrote a quick PoC for this a couple of months ago

```js
async function updateStyles(theme) {
  const css = await fetch(`https://dev.to/assets/themes/${theme}.css`);

  let string = await css.text();
  string = string.substring(6, string.length - 2);

  const styles = string.split(";");
  const root = document.documentElement;

  for (let style of styles) {
    const [prop, setting] = style.split(": ");
    root.style.setProperty(prop, setting);
  }
}
```

This function can take a theme name (like pink) and it will update the theme locally.

```js
updateStyles('pink');
```

_It'll work if you paste those into your console_

I suppose it would be possible to, on this [page](https://dev.to/settings/ux), update the theme dynamically as the user press on the available options as a preview. If they don't press save we'd need to set it back to their current theme though.

shilpi230:
Hey @Link2Twenty Thanks! I will have a look into it. I deleted this branch, because I did some mistake while pulling the code from master.

PR #11092, opened 2020-10-26
Add admin shortcut to tag page

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The problem that this is solving is that setting any admin-only functionality lacks intuitive ergonomics for admins. If they recognize that a tag needs to be adjusted or elevated, it's a pain to get to the proper admin dashboard. This adds an admin button alongside mod buttons. Also made a style adjustment.

Old:

Screen Shot 2020-10-26 at 11 59 40 AM

New:

Screen Shot 2020-10-26 at 11 53 18 AM

This creates a button a shortcut to admin, alongside moderation to allow admins to more easily modify admin-level tag decisions.

### DIFF
--- a/app/assets/javascripts/initializers/initializeAllTagEditButtons.js
+++ b/app/assets/javascripts/initializers/initializeAllTagEditButtons.js
@@ -4,7 +4,7 @@ function initializeAllTagEditButtons() {
   var tagEditButton = document.getElementById('tag-edit-button');
   var tagAdminButton = document.getElementById('tag-admin-button');
   var user = userData();
-  if (user.admin) {
+  if (user.admin && tagAdminButton) {
     tagAdminButton.style.display = 'inline-block';
     document.getElementById('tag-admin-button').style.display = 'inline-block';
   }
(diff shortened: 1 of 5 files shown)


lisasy:
I like this and this LGTM. Curious @ludwiczakpawel's thoughts around using `ff-monospace` as a variant of the button text style.

benhalpern:
Yeah, @lisasy I think the monospace is just inherited from the container, which is legacy style.

I'd be down to give this a bit of a more thorough quick crayons pass similar to the tags page, or just ship this as a minor improvement and circle back next time for more.

ludwiczakpawel:
Yeap, since the monospace is inherited, lets ship it as is and circle back later on.

ludwiczakpawel (approved):

Zhao-Andy (approved):
LGTM. One thing I noticed is that if the user is not an admin or mod, the last button won't be rounded:
Screen Shot 2020-10-27 at 9 43 39 AM

Pretty sure this is because the admin button exists in the HTML, but is hidden visually. This can probably be adjusted via CSS and checking for `display:none` or something, but probably not a huge deal.

PR #11095, opened 2020-10-26
Optimization:Limit Tags Searched For Follow Suggestions

mstruve:
- [x] Optimization

## Description
Was recently looking at some slow traces on Honeycomb when I came across one that had a 1 second SQL query in it. Upon further investigation, I noticed that the query was trying to suggest users to follow and was using the current user's followed tags. This particular user had 20 tags they were following which made the query pretty intense with 20 `OR` tagging statements. To save us some process time and power I decided a good solution would be to limit the number of tags when suggesting users to follow. 

## Added tests?
- [x] No, overall behavior has remained the same so current tests should be sufficient 

## Added to documentation?
- [x] No documentation needed

alt_text

### DIFF
--- a/app/services/suggester/users/recent.rb
+++ b/app/services/suggester/users/recent.rb
@@ -22,7 +22,7 @@ module Suggester
 
       def tagged_article_user_ids(num_weeks = 1)
         Article.published
-          .tagged_with(user.decorate.cached_followed_tag_names, any: true)
+          .tagged_with(user.decorate.cached_followed_tag_names.sample(5), any: true)
           .where("score > ? AND published_at > ?", article_reaction_count, num_weeks.weeks.ago)
           .pluck(:user_id)
           .each_with_object(Hash.new(0)) { |value, counts| counts[value] += 1 }


maestromac (approved):

juliannatetreault (approved):
Great catch! 👍 👍

benhalpern (approved):

rhymes (commented):
+1 for the idea, is there a chance that sampling will change behavior in a significant way for smaller communities? I'm not 100% sure I follow the "why" behind the logic of `Suggester::Users::Recent` :blush:

mstruve:
>  is there a chance that sampling will change behavior in a significant way for smaller communities (like a few tags and a few users)

If a community has few set suggested tags then this would not affect them at all bc we would likely sample all the tags if the user chose them. The only disadvantage here would be a Forem with a large number of tags and a low number of articles and you happened to select the tags that have no articles. This could happen with the sample or could just happen if the person choose only a couple non popular tags. 

TL;DR I think this change will likely have a very minimal negative effect and preventing large OR queries out weighs that.

PR #11096, opened 2020-10-26
Refactor:Replace Production Check with Average Metrics

mstruve:
- [x] Refactor

## Description
Ideally, we would like to minimize the places we are calling `Rails.env.production?` in our codebase because the more we do that, the larger the difference between production and our development environment becomes. The more differences we have, the harder it can be to catch bugs bc we have more code paths that are not being executed when we test locally or in our test suite. This removes a few of those production check statements and replaces them with metrics that gather averages for different counts and scores. 

I changed the SQL statements to be `>=` which then easily handled testing situations where we create only one article, that article is then returned. 

## QA Instructions
Code Path 1
1. Remove your user in development 
2. Set `SiteConfig.suggested_tags = ["beginners", "java", "python"]`
3. Sign up 
4. During onboarding select the 3 tags and click through to the follow suggestion page
5. You should still see user follow suggestions 
Code Path 2
1. Remove your user in development 
2. Sign up 
3. During onboarding DO NOT select any tags and click through to the follow suggestion page
4. You should still see user follow suggestions 

## Added tests?
- [x] Yes/Updated them to be better 

## Added to documentation?
- [x] No documentation needed

alt_text


app/services/suggester/users/recent.rb
```diff
@@ -23,7 +23,7 @@ def suggest
       def tagged_article_user_ids(num_weeks = 1)
         Article.published
           .tagged_with(user.decorate.cached_followed_tag_names, any: true)
-          .where("score > ? AND published_at > ?", article_reaction_count, num_weeks.weeks.ago)
+          .where("score >= ? AND published_at > ?", article_score_average, num_weeks.weeks.ago)
```
mstruve, line 26: I am not sure why this method was `article_reaction_count` since we are using it to filter by score so I updated it. Probably leftover from a previous code change.

app/services/suggester/users/recent.rb
```diff
@@ -55,15 +55,21 @@ def recent_commenters(num_comments = 2, limit = 8)
       end
 
       def established_user_article_count
-        Rails.env.production? ? 4 : -1
+        Rails.cache.fetch("established_user_article_count", expires_in: 1.day) do
```
mstruve, line 58: Not married to the 1 day expiration, it could probably be longer. If anyone has an opinion, let me know!

app/services/suggester/users/recent.rb
```diff
@@ -55,15 +55,21 @@ def recent_commenters(num_comments = 2, limit = 8)
...
+          User.average(:articles_count)
+        end
       end
 
       def established_user_comment_count
-        Rails.env.production? ? 4 : -1
+        Rails.cache.fetch("established_user_comment_count", expires_in: 1.day) do
+          User.average(:comments_count)
```
rhymes, line 65: this is going to change significantly, as we go from 4 to 1.136, at least for DEV

app/services/suggester/users/recent.rb
```diff
@@ -55,15 +55,21 @@ def recent_commenters(num_comments = 2, limit = 8)
       end
 
       def established_user_article_count
-        Rails.env.production? ? 4 : -1
+        Rails.cache.fetch("established_user_article_count", expires_in: 1.day) do
+          User.average(:articles_count)
```
rhymes, line 59: this is going to change significantly, as we go from 4 to 0.87, at least for DEV

rhymes (commented):
Left a note on the magic numbers, as mentioned in https://github.com/forem/forem/pull/11095 I find `Suggester::Users::Recent` a bit opaque in explaining the "why" of its logic but if these new numbers make sense, LGTM! :=)

app/services/suggester/users/recent.rb
mstruve: Given this is only used in the onboarding flow I am not very concerned about this value changing from 4 to essentially 2 since its greater than or equal to and counts are whole numbers

app/services/suggester/users/recent.rb
mstruve: This is a bit of a bigger change from 4 to 1, what if we played with the ordering to ensure the ones with higher counts were near the top in line 47 and 54?

rhymes: not sure but yeah, we can play with the numbers until we get it right

cc @benhalpern @vaidehijoshi

mstruve: Ok, I let this sit for a bit and came up with this idea. What do you think of limiting those we take the average from to the ones with counts greater than 0 and then article scores greater than or equal to 0?
```ruby
irb(main):005:0> User.where("articles_count > 0").average(:articles_count).to_i
=> 7
irb(main):007:0> User.where("comments_count > 0").average(:comments_count).to_i
=> 5
irb(main):009:0> Article.where("score >= 0").average(:score).to_i
=> 16
```

vaidehijoshi: > limiting those we take the average from to the ones with counts greater than 0 and then article scores greater than or equal to 0?

@mstruve I like this idea, it also feels more in line with what a Forem creator would want as well

mstruve: Updated and ready for rereview! Thanks for all the feedback!

rhymes (approved):

Ridhwana (approved):
This looks good @mstruve

PR #11102, opened 2020-10-26
Rate Limit Modal - Comments

joshpuetz:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Here's a non-preact implementation of the same modal I sketched out in https://github.com/forem/forem/pull/11023. After trying (and mostly failing) to integrate the Preact component with arguments on non-Preact pages, I went back to the drawing board and came up with a plain on Javascript implementation. I've hooked it up to the comments page: making too many comments in a short amount of time now triggers this modal.

Concept: we want to create reusable messaging to alert users about exceeding rate limits or user interaction limits. Right now these limits are indicated by backend throwing HTTP status code 429 RATE LIMIT EXCEEDED, but it's up to the frontend to do something with it. Some UIs catch this error and show it in an alert, others just throw it away and don't inform the user. Ideally the frontend would have a reusable modal to display (design listed here: https://www.figma.com/file/W0NVUGtoMdQwRc9pH4M5dd/Interaction-limits?node-id=16%3A1)

closes forem/InternalProjectPlanning#192

To test:
- navigate to `admin/config`. Under the "Rate limits and anti-spam" section set "Limit number of comments created" to a low number (like 1)
- navigate to a post and start making comments. 
- you should receive an alert:
Little_Hands_Clapping_Exercitationem_voluptatum_-_Forem_Community_👩💻👨💻

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [ ] No documentation needed
- [x] Inline/JS docs

giphy

### DIFF
--- a/app/assets/javascripts/utilities/showUserAlertModal.js
+++ b/app/assets/javascripts/utilities/showUserAlertModal.js
@@ -1,0 +1,50 @@
+const modalId = 'user-alert-modal';
+
+function showUserAlertModal() {
+  let modalDiv = buildModalDiv();
+  toggleUserAlertModal();
+}
+
+function toggleUserAlertModal() {
+  let modalDiv = document.getElementById(modalId);
+  if (modalId) {
+    modalDiv.classList.toggle('hidden');
+  }
+}
+
+function buildModalDiv() {
+  let modalDiv = document.getElementById(modalId);
+  if (!modalDiv) {
+    modalDiv = buildModalHTML("Sample Modal", "This is my sample modal dialog box");
+    document.body.appendChild(modalDiv);
+  }
+  return modalDiv;
+}
+
+
+function buildModalHTML(title, text) {
+  let wrapper = document.createElement('div');
+  wrapper.innerHTML= `<div id="${modalId}" data-testid="modal-container" class="crayons-modal hidden">
+      <div role="dialog" aria-modal="true" class="crayons-modal__box">
+        <div class="crayons-modal__box__header">
+          <h2>${title}</h2>
+            <button class="crayons-btn crayons-btn--ghost crayons-btn--icon" type="button" 
+                onClick="toggleUserAlertModal();" aria-label="Close">
+              <svg width="24" height="24" viewBox="0 0 24 24" class="crayons-icon"
+                xmlns="http://www.w3.org/2000/svg" role="img" aria-labelledby="714d29e78a3867c79b07f310e075e824">
+                <title id="714d29e78a3867c79b07f310e075e824">Close</title>
+                <path
+                  d="M12 10.586l4.95-4.95 1.414 1.414-4.95 4.95 4.95 4.95-1.414 1.414-4.95-4.95-4.95 4.95-1.414-1.414 4.95-4.95-4.95-4.95L7.05 5.636l4.95 4.95z">
+                </path>
+              </svg>
+            </button>
+        </div>
+        <div class="crayons-modal__box__body">
+          <p>${text}</p>
+        </div>
+      </div>
+      <div data-testid="modal-overlay" class="crayons-modal__overlay"></div>
+    </div>
+  `;
+  return wrapper.firstChild;
+}
--- a/app/assets/javascripts/utilities/showUserAlertModal.js
+++ b/app/assets/javascripts/utilities/showUserAlertModal.js
@@ -3,8 +3,8 @@ function showUserAlertModal(title, text, confirm_text, link, link_text) {
   toggleUserAlertModal();
 }
 
-function showRateLimitModal(action_text, wait_time_text, next_action_text) {
-  let rateLimitText = buildRateLimitText(action_text, wait_time_text, next_action_text);
+function showRateLimitModal(action_text, next_action_text) {
+  let rateLimitText = buildRateLimitText(action_text, next_action_text);
   let rateLimitLink = "/faq";
   showUserAlertModal('Wait a Moment...', rateLimitText, 'Got it', rateLimitLink, "Why do I have to wait?")
 }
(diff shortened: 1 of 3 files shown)


ludwiczakpawel:
@joshpuetz I think that makes sense overall.

I wonder if Modals in general could be built utilizing power of [`<template>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template) tag. So we keep the general scaffold somewhere in HTML, instead of JS so if we ever want to adjust modals we'd do that in one place instead everywhere... Plus it would then be just much easier to use modals for other scenarios...

So here, we would have a `<tamplate>` with empty modal somewhere in the markup and you could access it, activate it and fill the data (content) through JS. But without actually building the markup in JS.

@nickytonline I'm keen to know your thoughts on this.

I guess that's a bigger discussion for other components we have like snackbars, etc and I don't want to cannibalize this PR to have that convo. But wanted to plant a seed in your minds :D

joshpuetz:
> @joshpuetz I think that makes sense overall.
> 
> I wonder if Modals in general could be built utilizing power of [`<template>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template) tag. So we keep the general scaffold somewhere in HTML, instead of JS so if we ever want to adjust modals we'd do that in one place instead everywhere... Plus it would then be just much easier to use modals for other scenarios...
> 
> So here, we would have a `<tamplate>` with empty modal somewhere in the markup and you could access it, activate it and fill the data (content) through JS. But without actually building the markup in JS.
> 
> @nickytonline I'm keen to know your thoughts on this.
> 
> I guess that's a bigger discussion for other components we have like snackbars, etc and I don't want to cannibalize this PR to have that convo. But wanted to plant a seed in your minds :D

@ludwiczakpawel  `<template>` is awesome! I used it during the work on the Codeland page, and it was great! I have considered using it for this implementation, but I'm wary about stashing it somewhere on the page and having it be external to this script versus living alongside this script: I'd have a file (or the template tag code) sitting...somewhere, far from this file. 

As a compromise I'm going to clean up the html template/fragment in this file, and make it as separate from the rest of the code as I can.

joshpuetz:
CodeClimate HATES my functions with five arguments (I don't love them either). What's a good refactor? Just stuffing them in an array feels like cheating.

General_alert_modal_for_rate_limits__non-preact_version__-_forem_forem_-_Code_Climate

nickytonline:
You could use a `<template />` tag and clone it's content when you need it, but you need to do this via JS, so I would probably consider creating it in Preact. Sure the markup is rendered by JSX which turns into JS instead of pulling some markup from the DOM, but I think in terms of a reusable component, it makes sense to keep it in Preact land. Having said that, admin pages have been predominantly using Stimulus. As far as I know, none of the admin pages use Preact at the moment.

And this brings up a whole discussion about what should our components be built with. Currently we have Stimulus, HTML/CSS + Vanilla JS, and Preact. We're building out a design system and some core components such as this modal, so we need to define when to use one over the other or potentially go all in on one.

I can't remember when Stimulus was introduced and why it was preferred over Preact, so it would be good to have that discussion as well.

joshpuetz:
> You could use a `<template />` tag and clone it's content when you need it, but you need to do this via JS, so I would probably consider creating it in Preact. Sure the markup is rendered by JSX which turns into JS instead of pulling some markup from the DOM, but I think in terms of a reusable component, it makes sense to keep it in Preact land. Having said that, admin pages have been predominantly using Stimulus. As far as I know, none of the admin pages use Preact at the moment.
> 
> 
> 
> And this brings up a whole discussion about what should our components be built with. Currently we have Stimulus, HTML/CSS + Vanilla JS, and Preact. We're building out a design system and some core components such as this modal, so we need to define when to use one over the other or potentially go all in on one.
> 
> 
> 
> I can't remember when Stimulus was introduced and why it was preferred over Preact, so it would be good to have that discussion as well.

Hey @nickytonline: I started making a Preact version of this in https://github.com/forem/forem/pull/11023, but abandoned it due to difficulty calling it with arguments from non-Preact pages.

Regarding the template tag idea: it's a good one, but this modal will be shown on multiple pages on the site and I'm reluctant to include this markup on _every_ page!

nickytonline:
Let me take a peek at the Preact one you did @joshpuetz. 👀

joshpuetz:
Tested added and ready for review!

app/assets/javascripts/utilities/showUserAlertModal.js
```diff
@@ -0,0 +1,145 @@
...
+ * @private
+ * @function modalHTML
+ * 
+ * @param {string} title The title/heading text to be displayed
+ * @param {string} text The body text to be displayed
+ * @param {string} confirm_text Text of the confirmation button
+ * 
+ * @returns {string} HTML for the modal
+ */
+const modalHTML = (title, text, confirm_text) => `<div id="${modalId}" data-testid="modal-container" class="crayons-modal hidden">
```
nickytonline, line 54: nitpick (non-blocking): Consider naming this `getModalHtml` or `buildModalHtml`.

app/assets/javascripts/utilities/showUserAlertModal.js
```diff
@@ -0,0 +1,145 @@
...
+ * @private
+ * @function buildModalHTML
+ * 
+ * @param {string} title The title/heading text to be displayed
+ * @param {string} text The body text to be displayed
+ * @param {string} confirm_text Text of the confirmation button
+ * 
+ * @returns {Element} DOM node of alert modal with formatted text
+ */
+function buildModalHTML(title, text, confirm_text) {
```
nickytonline, line 141: nitpick (non-blocking): Consider calling this `getModal()` as it's returning an element not HTML.

app/assets/javascripts/utilities/showUserAlertModal.js
```diff
@@ -0,0 +1,145 @@
+/**
```
nickytonline, line 1: More of a side note, but I see why you had difficulties in putting this in Preact land/modern JS code base (ES modules) as it needs to be called from the legacy JS scripts. It would require a `setTimeout` or `setInterval` to determine when the modal component was loaded to be able to use it which is not that ideal.

spec/system/comments/user_fills_out_comment_spec.rb
```diff
@@ -24,6 +25,20 @@
     expect(page).to have_text(raw_comment)
   end
 
+  it "User makes too many comments" do
```
nickytonline, line 28: This is pretty much the way to test non-Preact/ES module code for now. Is the submit button being tested here, the one in the modal? Maybe check for some other things in the modal as well.

msarit (approved):
After pending issues have been addressed, LGTM! 🚀

nickytonline (approved):

PR #11104, opened 2020-10-26
Add Imgproxy docs

maestromac:
- [x] Documentation Update

## Description
This addes docs and simplified some specs. I've decided not to add this as part of the Procfile and make it a requirement because the app can operate without it.

Resolves https://github.com/orgs/forem/projects/11#card-47096747
Kindly follow the instruction on `imgproxy.md`

## Added tests?
- [x] Yes
## Added to documentation?
- [x] Docs.forem.com
nope


app/services/images/optimizer.rb
```diff
@@ -53,7 +53,14 @@ def self.translate_cloudinary_options(options)
     end
 
     def self.imgproxy_enabled?
-      Imgproxy.config.key.present? && Imgproxy.config.salt.present?
+      if Rails.env.production?
```
rhymes, line 56: as we're trying to remove `if Rails.env.production?` around the code as a general principle, see https://github.com/forem/forem/pull/11096, is there a way for us to not add this check? simply check `(key && salt) || env_var`, would that suffice?

docs/installation/imgproxy.md
```diff
@@ -0,0 +1,35 @@
+---
+title: Imgproxy
+---
+
+Imgproxy is a standalone server for resizing images. It is optional and you do
+not need it to start Forem. We are currently only it using Forem-cloud's
+application.
```
rhymes, line 7: `to start Forem locally`

`We are currently only using it on Forem Cloud`

docs/installation/imgproxy.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Installation
+
+On macOS, use homebrew with `brew install imgproxy`. For others, please see the
+[official documentation](https://docs.imgproxy.net/#/installation)
```
rhymes, line 12: `For other operating systems`

I wonder if this page should be linked in the various installation guides, otherwise when is the user going to hit the "imgproxy" page in its installation workflow?

docs/installation/imgproxy.md
```diff
@@ -0,0 +1,35 @@
...
+1. Startup Imgproxy in a terminal. If you installed via homebrew, it's
+   `imgproxy`.
+1. Startup the Forem app.
+1. That's it :)
+
+You should verify it's working by checking a user's profile image. It should
+have a structure similar to
+`http://localhost:8080/unsafe/rs:fill:320:320/aHR0cDovL2xvY2Fs/aG9zdD.......`.
+You should also see activities from Imgproxy terminal and all the images should
+load.
```
rhymes, line 27: Is there a `curl` command we can use to check? Like fetching one of the builtin images

docs/installation/imgproxy.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Sidenote.
+
+- Because Imgproxy is a standalone server of its own, all images URL given to it
+  needs to be an absolute URL.
```
rhymes, line 32: `need to be absolute URLs`

docs/installation/imgproxy.md
```diff
@@ -0,0 +1,35 @@
...
+- When working with docker/podman on Linux, provide the host network option, ie
+  `docker run -p 8080:8080 --network="host" -it darthsim/imgproxy` so Imgproxy
+  can properly access the localhost.
```
rhymes, line 35: as you mention podman, is this command going to work with it as well?

docs/technical-overview/stack.md
```diff
@@ -11,7 +11,9 @@ For the Forem tech stack we use:
 - [_Redis_](https://redis.io/) to store cached data
 - [_Fastly_](https://www.fastly.com/) for
   [edge caching](https://dev.to/ben/making-devto-insanely-fast)
-- [_Cloudinary_](https://cloudinary.com/) for image manipulation/serving
+- [_Cloudinary_](https://cloudinary.com/) and/or
+  [_Imgproxy_](https://github.com/imgproxy/imgproxy) for image
+  manipulation/serving
```
rhymes, line 16: one day there'll only be one of the two here :P

rhymes (commented):
Left some notes, thanks @maestromac!

docs/technical-overview/stack.md
maestromac: one day

docs/installation/imgproxy.md
maestromac: I can clarify this further.

docs/installation/imgproxy.md
maestromac: yup, let me update this too.

app/services/images/optimizer.rb
maestromac: Yeah I was reluctant to have this weird requirement either but your suggestion is a good middle ground.

docs/installation/imgproxy.md
maestromac: I just added a curl command but just realized I'm making the assuming that our users would use `localhost:3000`. Since the url is encoded it could potentiall be confusing. What do you think?

rhymes: maybe we can replace that with https://docs.imgproxy.net/#/healthcheck?id=imgproxy-health but it might be useless as we're testing the server, not the proxy to our app :D Nevermind

maestromac: Haha, well I think I have there is better than what I had before so I'm going to leave it as is for the time being.

app/services/images/optimizer.rb
maestromac: Oh second thought, I want to be able to offer a simple setup in development while also prevent accidental release without key and salt on Production. Unlike other env var that accept literally any string, these need to hex-encoded or Imgproxy won't start. Checking only `(key && salt) || env_var` isn't strict enough. If it alright, I want to allow this area to use `Rails.env.production?`

mstruve: > these need to hex-encoded or Imgproxy won't start.

What happens if the key and salt are not properly encoded? Does the app fail to boot? Does it crash when you try to upload a picture?

maestromac: The rail's side would be fine and boot normally. The imgproxy server has to have hex key or it refuses to boot with
```
FATAL   [2020-10-27T19:13:25Z] IMGPROXY_KEY expected to be hex-encoded strings. Invalid: optional
```
If imgproxy server didn't start or wrong/mismatch keys were used in either place, the web app would not be able to render pictures.

mstruve: > Checking only (key && salt) || env_var isn't strict enough :(

The key and salt are going to be configured variables which means they are not going to change often so once they are set the app should be in a pretty good spot. If for some reason they get changed incorrectly and the sign is broken images, I think that is something we can live with. 

Having a production only code path really limits us in our ability to test and run things locally. I think having a production only code path poses a bigger risk for problems down the road with production than having a loose requirement of key/secret presence.

maestromac: Okay, I am undoing my change and keeping it only 1 code path, requiring key/salt in all environment.

docs/installation/imgproxy.md
```diff
@@ -0,0 +1,74 @@
+---
+title: Imgproxy
+---
+
+# Imgproxy
+
+Imgproxy is a standalone server for resizing images. It is optional and you do
+not need it to start Forem locally. We are currently only it using Forem Cloud.
```
mstruve, line 8:
```suggestion
not need it to start Forem locally. We are currently only using it in Forem Cloud.
```

docs/installation/imgproxy.md
```diff
@@ -0,0 +1,74 @@
...
+   > docker run -p 8080:8080 \
+      -e IMGPROXY_KEY="your key" \
+      -e IMGPROXY_SALT="your salt" \
+      -it darthsim/imgproxy
+   ```
+
+1. That's it :)
+
+You should verify it's working by starting the forem app locally and see that
+each image are loaded properly, or run the following command while the forem app
```
mstruve, line 57:
```suggestion
each image is loaded properly, or run the following command while the forem app
```

mstruve (approved):
Couple of small suggestions for typos otherwise LGTM!

docs/installation/imgproxy.md
```diff
@@ -0,0 +1,74 @@
...
+---
+
+# Imgproxy
+
+Imgproxy is a standalone server for resizing images. It is optional and you do
+not need it to start Forem locally. We are currently only it using Forem Cloud.
+
+## Installation
+
+- MacOS: use homebrew with `brew install imgproxy`.
```
vaidehijoshi, line 12:
```suggestion
- MacOS: install via homebrew with `brew install imgproxy`.
```

docs/installation/imgproxy.md
```diff
@@ -0,0 +1,74 @@
...
+## Usage
+
+1. Generate a key/salt pair by running the following in your terminal twice.
+   Copy those values to your .env in the next step
+
+   ```
+   echo $(xxd -g 2 -l 64 -p /dev/random | tr -d '\n')
+   ```
+
+1. In your `.env`, add the followings.
```
vaidehijoshi, line 31:
```suggestion
1. In your `.env`, add the following.
```

docs/installation/imgproxy.md
```diff
@@ -0,0 +1,74 @@
...
+  [docker installation](https://docs.imgproxy.net/#/installation?id=docker) or
+  [build from source](https://docs.imgproxy.net/#/installation?id=from-the-source).
+
+For more options not covered here, please take a look at the
+[official installation documentation](https://docs.imgproxy.net/#/installation).
+
+## Usage
+
+1. Generate a key/salt pair by running the following in your terminal twice.
+   Copy those values to your .env in the next step
```
vaidehijoshi, line 25:
```suggestion
   Copy those values to your `.env` in the next step
```

docs/installation/imgproxy.md
```diff
@@ -0,0 +1,74 @@
...
+   # If you are using Docker or Podman. The commands are identical for both
+   > docker run -p 8080:8080 \
+      -e IMGPROXY_KEY="your key" \
+      -e IMGPROXY_SALT="your salt" \
+      -it darthsim/imgproxy
+   ```
+
+1. That's it :)
+
+You should verify it's working by starting the forem app locally and see that
```
vaidehijoshi, line 56:
```suggestion
You should verify it's working by starting the Forem app locally and see that
```

docs/installation/imgproxy.md
vaidehijoshi, line 57:
```suggestion
each image are loaded properly, or run the following command while the Forem app
```

vaidehijoshi (approved):
Minor comments only from me! 😉

docs/installation/imgproxy.md
maestromac: Manually updated this. Thank you!!

rhymes (approved):

PR #11107, opened 2020-10-27
Add config option for Forem collective noun

jacobherrington:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Not everyone will want to use "Community" to describe their Forem, so
now they have some flexibility there.

Closes: https://github.com/forem/InternalProjectPlanning/issues/221

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed


rhymes (approved):
Cool!

benhalpern (approved):

vaidehijoshi:
This was merged before @forem/creator-onboarding got a look at it, but I wonder...could we use `community_name` here instead of adding another override-able config? (the `community_name` already exists and is a required field)

jacobherrington:
@vaidehijoshi that's probably a good way to refactor all of this - the only issue would be handling instances in which folks didn't want to include the "collective noun" value. For example, when we say "Welcome to DEV" versus "Welcome to the DEV Community"

We should definitely consider revisiting this!

PR #11113, opened 2020-10-27
Add chaining to shortcuts

### LINKED ISSUE #11094: Add chaining to shortcuts
Currently shortcuts are a single key + any modifiers. It's quite common, on github and on twitter for instance, to be able to chain key presses to achieve something for instance `g` then `h` to go home.

I think the new keyboard shortcut hook, #10713, should remember key presses for half a second and we should add a then syntax. Pressing a modifier key or nothing for half a second should clear the chain.

**Example:**

```js
useKeyboardShortcuts({
  "KeyG~KeyH": (e) => {
    e.preventDefault();
    alert("G has been pressed quickly followed by H");
});
```

We don't support chaining of keys in shortcuts.

**Additional context**

I noticed `G` for 'go to' is a common practice so perhaps something we should support.

Link2Twenty:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Add the ability to have chain shortcuts (pressing keys in a certain order) such as G then H

### Syntax

Key's joined by `~` are using chaining, keys with a `+` are using a key modifier.

```js
useKeyboardShortcuts({
  "KeyY": () => {
    console.log("Just Y pressed alone");
  },  
  "ctrl+KeyY": () => {
    console.log("Control/Command and Y pressed");
  },
  "KeyG~KeyH": () => {
    console.log("G was press followed by H");
  }
})
```

closes #11094

Here's an [example](https://codesandbox.io/s/example-of-key-chains-in-shortcut-hook-t64iv);

## Added tests?

- [x] Yes
- [ ] No, and this is why: not needed

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

N/A

fine tuning


app/javascript/shared/components/useKeyboardShortcuts.jsx
```diff
@@ -44,6 +44,44 @@ function isFormField(element) {
...
+
+  // Set up key chains
+  useEffect(() => {
+    if (!keyChainQueue && keyChain.length === 0) return;
+    let timeout;
+
+    timeout = window.setTimeout(() => {
+      clearTimeout(timeout);
+      setKeyChain([]);
+    }, 500);
```
rhymes, line 74: would it make sense to have this configurable? If I'm not mistaken in reading the code, GitHub uses 1.5 seconds, I did a non scientific experiment and indeed hotkeys are rvalid for a while :D

https://github.com/github/hotkey/blob/0691586c0eca5d198b5619b7321594706f5cc194/src/index.ts#L22

rhymes (commented):
Thanks @Link2Twenty!

app/javascript/shared/components/useKeyboardShortcuts.jsx
Link2Twenty: As in make is variable set in the backend? Or are you thinking we pull it from the env file?

reobin:
Nice @Link2Twenty !

I think it would be possible for tests to be added for this unless I'm mistaken. Something like:

1. Register a chained shortcut
2. Press 1st key
3. wait an arbitrary amount of time
4. Press the 2nd key
5. Verify that the shortcut callback has been called

Link2Twenty:
> Nice @Link2Twenty !
> 
> I think it would be possible for tests to be added for this unless I'm mistaken. Something like:
> 
> 1. Register a chained shortcut
> 2. Press 1st key
> 3. wait an arbitrary amount of time
> 4. Press the 2nd key
> 5. Verify that the shortcut callback has been called

Good point I've added a test for getting the chain in time and also missing it.

reobin:
This keyboard shortcut hook is becoming quite something! It is definitely going into my side projects from now on 😅

One thing that would be nice to have is the ability to add more than one shortcut for the same callback. It's possible right now, but you have to define the callback twice, and the event listener is registered twice as well:

```js
useKeyboardShorcuts({
  a: () => something(),
  b: () => something(),
});
```

With a separator, we could improve that into:

```js
useKeyboardShorcuts({
  "a,b": () => something(),
});
```

It could, of course, be something else than `,`. We would not want to create any confusion with shortcuts like `ctrl+,`.

Just throwing an idea out there, we could create an issue for that if it's something we would want to do.

Link2Twenty:
@rhymes I've made it so you can pass in a options object. and `{timeout: 1500}` will set the key timeout to 1.5 seconds. If there is no timeout in the options object it will default to 500ms as before.

Link2Twenty:
> and the event listener is registered twice as well

The event listener should only be registered once, but you're right being able to have multiple keys be able to do the same thing is a good idea.

app/javascript/shared/components/useKeyboardShortcuts.jsx
```diff
@@ -31,19 +31,24 @@ function isFormField(element) {
...
- * useKeyboardShortcuts(shortcuts, someElementOrWindowObject);
+ * useKeyboardShortcuts(shortcuts, someElementOrWindowObject, {timeout: 1500});
  *
  * @param {object} shortcuts List of keyboard shortcuts/event
  * @param {EventTarget} [eventTarget=window] An event target.
+ * @param {object} [options = {}] An object for extra options
  *
  */
-export function useKeyboardShortcuts(shortcuts, eventTarget = window) {
+export function useKeyboardShortcuts(shortcuts, eventTarget = window, options = {}) {
```
rhymes, line 51: IIRC you can declare `options = { timeout: 500 }` in the params in JavaScript, right? This way the props declarations can also check they are the right params

Link2Twenty: You're right, I've now changed that an made it cope with custom options a little better.

app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx
```diff
@@ -18,6 +18,57 @@ describe('Keyboard shortcuts for components', () => {
...
+      };
+
+      renderHook(() =>
+        useKeyboardShortcuts(shortcut, document, { timeout: 1500 }),
+      );
+      fireEvent.keyDown(document, { code: "KeyA" });
+      await new Promise(resolve => setTimeout(() => {
+        fireEvent.keyDown(document, { code: "KeyB" });
+        resolve();
+      },600));
```
nickytonline, line 33: Refactor (blocking): We want to avoid timeouts in tests as it will increase build times.

```suggestion
      await fireEvent.keyDown(document, { code: "KeyB" });
```

app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx
```diff
@@ -18,6 +18,57 @@ describe('Keyboard shortcuts for components', () => {
...
+      };
+
+      renderHook(() =>
+        useKeyboardShortcuts(shortcut, document),
+      );
+      fireEvent.keyDown(document, { code: "KeyA" });
+      await new Promise(resolve => setTimeout(() => {
+        fireEvent.keyDown(document, { code: "KeyB" });
+        resolve();
+      },200));
```
nickytonline, line 50: Refactor (blocking): We want to avoid timeouts in tests as it will increase build times.

```suggestion
      await fireEvent.keyDown(document, { code: "KeyB" });
```

app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx
```diff
@@ -18,6 +18,57 @@ describe('Keyboard shortcuts for components', () => {
...
+    it('should not fire a function when chained keydown is missed', async () => {
+      const shortcut = {
+        "KeyA~KeyB": jest.fn()
+      };
+
+      renderHook(() =>
+        useKeyboardShortcuts(shortcut, document),
+      );
+      fireEvent.keyDown(document, { code: "KeyA" });
+      await new Promise(resolve => setTimeout(() => {
```
nickytonline, line 64: This setTimeout appears to be necessary, but I'm looking into this still.

nickytonline (approved):
Looking good @Link2Twenty. Just a couple change requests.

nickytonline (changes requested):
Ooops, I meant to click Request changes. 🙃

nickytonline:
One thing that would be great in a separate PR @Link2Twenty would be to document this functionality in the frontend documentation on docs.forem.com

app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx
Link2Twenty: This it to test updating the timeout option works. If you feel we don't need to test this functionality we can take it out?

nickytonline: If it's to test that, consider adding a very low number for the timeout option since these tests add almost 2 seconds to the test run.

app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx
```diff
@@ -18,6 +18,54 @@ describe('Keyboard shortcuts for components', () => {
...
+      expect(shortcut["KeyA~KeyB"]).toHaveBeenCalledTimes(1);
+    });
+    
+    it('should not fire a function when chained keydown is missed', async () => {
+      const shortcut = {
+        "KeyA~KeyB": jest.fn()
+      };
+
+      renderHook(() =>
+        useKeyboardShortcuts(shortcut, document),
```
nickytonline, line 58: We can pass in a timeout to the options with the value of zero and then have the test  run with a timeout of the 25ms.

```suggestion
		const timeout = 0;
        useKeyboardShortcuts(shortcut, document, { timeout }),
```

app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx
```diff
@@ -18,6 +18,54 @@ describe('Keyboard shortcuts for components', () => {
...
+      );
+      fireEvent.keyDown(document, { code: "KeyA" });
+      await new Promise(resolve => setTimeout(() => {
+        fireEvent.keyDown(document, { code: "KeyB" });
+        resolve();
+      },600));
```
nickytonline, line 64: We can use the `timeout` value above and add a minimal amount of time to get this test to pass.

```suggestion
      },timeout + 25));
```

app/javascript/shared/components/useKeyboardShortcuts.jsx
```diff
@@ -31,19 +36,68 @@ function isFormField(element) {
...
- * useKeyboardShortcuts(shortcuts, someElementOrWindowObject);
+ * useKeyboardShortcuts(shortcuts, someElementOrWindowObject, {timeout: 1500});
  *
  * @param {object} shortcuts List of keyboard shortcuts/event
  * @param {EventTarget} [eventTarget=window] An event target.
+ * @param {object} [options = {}] An object for extra options
  *
  */
-export function useKeyboardShortcuts(shortcuts, eventTarget = window) {
+export function useKeyboardShortcuts(shortcuts, eventTarget = window, options = { timeout: 500 }) {
```
nickytonline, line 56: Refactor (blocking): The `timeout` property set to `500` is not necessary as the spread `defaultOptions` for props already handles that case. I always have maintenance on the brain. 😄

```suggestion  
export function useKeyboardShortcuts(shortcuts, eventTarget = window, options = {}) {
```

app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx
```diff
@@ -18,6 +18,54 @@ describe('Keyboard shortcuts for components', () => {
       expect(shortcut.KeyK).toHaveBeenCalledTimes(1);
     });
     
+    it('should fire a function when longer chained keydown is detected', async () => {
+      const shortcut = {
+        "KeyA~KeyB": jest.fn()
+      };
+
+      renderHook(() =>
+        useKeyboardShortcuts(shortcut, document, { timeout: 1500 }),
```
nickytonline, line 27: Refactor (blocking): This is adding 1.5 seconds to our tests. Consider passing in a super low value, even `0`, then test with a `setTimeout` time delay of the actual timeout plus a small number, e.g. 25ms.

nickytonline (changes requested):
Some more request changes. The idea of this PR is great, just thinking about test run times is all.

app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx
Link2Twenty: The timeout inside the function isn't on an await is it still waited for?

Link2Twenty:
Is the build erroring me?

rhymes:
@Link2Twenty looks lime it timed out, I restarted it

rhymes (approved):
LGTM! :) Thanks @Link2Twenty

Link2Twenty:
@nickytonline test time delay is down to 5ms

app/javascript/shared/components/useKeyboardShortcuts.jsx
```diff
@@ -1,4 +1,4 @@
-import { useEffect } from "preact/hooks";
+import { useState, useEffect, useCallback } from "preact/hooks";
 import PropTypes from 'prop-types';
 import { h } from 'preact';
```
nickytonline, line 3: refactor (non-blocking): There is no JSX in this file so it can be renamed to `useKeyboardShortcuts.js`.

refactor (non-blocking): I think this is remnants of a previous PR when there was JSX in this file. It'd be good to get it cleaned up since these hooks have been reworked.
 
```suggestion
```

nickytonline (changes requested):
Looking good @Link2Twenty, some comments and a small change request.

app/javascript/shared/components/useKeyboardShortcuts.jsx
```diff
@@ -31,19 +36,68 @@ function isFormField(element) {
...
  *
  * @param {object} shortcuts List of keyboard shortcuts/event
  * @param {EventTarget} [eventTarget=window] An event target.
+ * @param {object} [options = {}] An object for extra options
  *
  */
-export function useKeyboardShortcuts(shortcuts, eventTarget = window) {
+export function useKeyboardShortcuts(shortcuts, eventTarget = window, options = {}) {
+  const [keyChain, setKeyChain] = useState([]);
+  const [keyChainQueue, setKeyChainQueue] = useState(null);
```
nickytonline, line 58: refactor (question): Just wondering if `keyChainQueue` is necessary. Why not just append to `keyChain` via `setKeyChain`?

Link2Twenty: I was avoiding having `keyChain` in the useEffect that set and unset the eventListener to prevent it being added and removed each key press.

app/javascript/shared/components/useKeyboardShortcuts.js
```diff
@@ -75,8 +75,8 @@ export function useKeyboardShortcuts(shortcuts, eventTarget = window, options =
   
   // update mergedOptions if options prop changes
   useEffect(() => {
-    setMergedOptions({...defaultOptions, ...options})
-  },[options])
+    setMergedOptions({...defaultOptions, ...{timeout: options.timeout} });
+  }, [options.timeout]);
```
nickytonline, line 79: Good catch. Was the linter complaining about this?

nickytonline:
One of the build nodes was dead. I've restarted it.

nickytonline (approved):
Thanks so much for the PR @Link2Twenty and thanks for being patient during the review. 🚀

nickytonline:
Restarting the build node that failed due to flaky tests and then this is good to merge if someone gets to it before me.

nickytonline:
Restarting the build node that failed. 🤞🏻

nickytonline:
Looks like there is a valid failure here @Link2Twenty. Mind taking a look at it?

```bash
FAIL  app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx
  ● Keyboard shortcuts for components › useKeyboardShortcuts › should not fire a function when chained keydown is missed, timeout should only be 0ms
    expect(jest.fn()).not.toHaveBeenCalled()
    Expected number of calls: 0
    Received number of calls: 1
    1: {"isTrusted": false}
      50 |       }, 5));
      51 | 
    > 52 |       expect(shortcut["KeyA~KeyB"]).not.toHaveBeenCalled();
         |                                         ^
      53 |     });
      54 |     
      55 |     it('should not add event listener if shortcut object is empty', () => {
      at Object.<anonymous> (app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx:52:41)
      at asyncGeneratorStep (app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx:11:103)
      at _next (app/javascript/shared/components/__tests__/useKeyboardShortcuts.test.jsx:13:194)
(node:8362) UnhandledPromiseRejectionWarning: FetchError: invalid json response body at  reason: Unexpected end of JSON input
(node:8362) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 6)
```

nickytonline:
Thanks so much for the PR @Link2Twenty! 🥳

PR #11115, opened 2020-10-27
Autoresize on comment textarea + comment submit on `CMD/CTRL+Return`

### LINKED ISSUE #10829: Comments box should grow as you type
**Describe the bug**

The comment box should grow as you type so you can see your whole comment the entire time.

**To Reproduce**

Type a long comment and you will see the comment box does not grow.

**Expected behavior**

A growing comment box.

### LINKED ISSUE #11086: cmd/ctrl+return to submit comment
This functionality used to exist and then went missing without a regression test. We now have [more sophisticated approaches for doing keyboard shortcuts](https://forem.dev/link2twenty/keyboard-shortcuts-5h66). We should re-enable this.

ludwiczakpawel:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR does two things:

1. Improves autoresizing when writing a comment (not sure why it doesn't work on replying comment form but it also doesn't work currently so at least I'm not breaking it :D...) 
2. Brings back possibility to to hit CMD+Return to submit a comment.

- Closes #10829 
- Closes #11086 

## Added tests?

- [ ] Yes
- [x] Not needed.

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
+++ b/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
@@ -179,6 +179,11 @@ function handleCommentSubmit(event) {
   event.preventDefault();
   var form = event.target;
   form.classList.add('submitting');
+  var textarea = form.getElementsByClassName('comment-textarea')[0];
+  if (textarea) {
+    textarea.style.height = null;
+    textarea.blur();
+  }
   var parentComment = document.getElementById("comment-node-" + event.target.dataset.commentId);
 
   var body = JSON.stringify({
@@ -392,7 +397,8 @@ function handleFormClose(event) {
 
 function handleSizeChange(event) {
   var textarea = event.target;
-  textarea.style.height = textarea.scrollHeight + "px";
+  var oldHeight = parseInt(textarea.style.height.replace('px',''));
+  textarea.style.height = textarea.scrollHeight + (textarea.scrollHeight > oldHeight ? 15 : 0) + "px";
 }
 
 function handleButtonsActivation(event) {
--- a/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
+++ b/app/assets/javascripts/initializers/initializeCommentsPage.js.erb
@@ -341,7 +341,7 @@ function replaceSelectedText(textArea, text) {
 
     // IE 8-10
     else if(document.selection) {
-      t ieRange = document.selection.createRange();
+      const ieRange = document.selection.createRange();
       ieRange.text = text;
 
       // Move cursor after the inserted text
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/initializers/initializeCommentsPage.js.erb
```diff
@@ -353,7 +341,7 @@ function replaceSelectedText(textArea, text) {
 
     // IE 8-10
     else if(document.selection) {
-      const ieRange = document.selection.createRange();
+      t ieRange = document.selection.createRange();
```
rhymes, line 344: is this a typo?

ludwiczakpawel: ummm... definitely. good catch. reverting this guy.

rhymes (commented):
I tested it, the keyboard shortcut works but the resizing mechanism is weird, it's like there's a delayed jump in size when you type enough data, also, it doesn't shrink back if you delete it. Maybe a timed enlarge/shrink? Not sure it can be done

What if we don't do auto resize at all and rely on the browser's default text area mechanism? While I'm typing this in this box in GitHub it doesn't auto resize but I can just resize it with the handle, the same I can do in the comment box.

ludwiczakpawel:
@rhymes github does some autoresizing though 🤷  

I'll look at the glitch/delay thingy. i think i know where it might be coming from but need to double check.

Also, should it shrink when deleting text though? I'm not sure if it's necessary, but should be fairly easy to add it (I don't have strong opinion tbh).

> rely on the browser's default text area mechanism
Browser default mechanism is actually no mechanism, right? You have to manually resize it, correct?

rhymes:
ah yeah, GitHub auto resizes by scrolling down and enlarging the box

app/assets/javascripts/initializers/initializeCommentsPage.js.erb
```diff
@@ -403,19 +391,8 @@ function handleFormClose(event) {
...
-    textarea.classList.add('embiggened-max');
-    textarea.classList.remove('embiggened-more', 'embiggened');
-  } else if (lineCount > 3) {
-    textarea.classList.add('embiggened-more');
-    textarea.classList.remove('embiggened-max', 'embiggened');
-  } else {
-    textarea.classList.add('embiggened');
-    textarea.classList.remove('embiggened-more', 'embiggened-max');
-  }
+  textarea.style.height = textarea.scrollHeight + "px";
```
benhalpern, line 395:
```suggestion
  var oldHeight = parseInt(textarea.style.height.replace('px',''));
  textarea.style.height = textarea.scrollHeight + (textarea.scrollHeight > oldHeight ? 15 : 0) + "px";
```

benhalpern:
@ludwiczakpawel @rhymes I submitted a couple small changes directly which I think handle the glitchiness better, and create proper blurring affects on submit.

benhalpern (approved):
All in all looks great. Functionality is good, and I made some tiny changes to fix the glitchiness. I think this is a definite improvement over what's currently in prod.

rhymes (approved):
LGTM! We can keep refining it to find the perfect recipe 👓

PR #11116, opened 2020-10-27
A11y: Add shortcut to bookmark an article

### LINKED ISSUE #11018: feat: Add a keyboard shortcut to save an article in the feed
This feature will complement #10468 very well.

Basically, something like `s` or `b` would save/unsave an article when you are navigating the feed with the keyboard as long as the focus is inside an article item.

Using the `useKeyboardShortcuts` hook, we can listen for the chosen key.

On press:
* fetch the parent article of the currently focused element using `element.closest`
* get the id of the article
* bookmark the article using the id

1. It could also click on the `save` button, which already has an `onClick` event listener that bookmarks, but that's hardly maintainable. With a refactor to that event listener, both shortcut and `onClick` could use the same `bookmark` function, which is a better option IMO.

2. Another option is to add an `onKeyDown` event listener on the article itself, which would make the process easier but would not use the `useKeyboardShorcuts` hook. Using the hook will be necessary for the eventual keyboard shortcuts modal.

3. Another option is to add the `useKeyboardShortcuts` hook on all the articles, which would have the benefits of the second option but would add potentially hundreds of `keydown` event listeners.

**Additional context**
#10713 
#10879

reobin:
##  What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds a shortcut on `b` (for now?) to save an article as long as the user is focused somewhere inside the article.

Also adds a shortcut on the signup modal to close it on `Escape` keypress. It helps to keep the navigation mouseless after pressing `b`.

This will be more useful once #10468 is merged, but it can live on its own.

Also, `b` is to discuss here. Other options have been discussed such as `ctrl+s`.

Closes #11018 

1. Focus on article (anywhere inside)
2. Press `b`
3. Article is saved/unsaved
    - Logged in: actually saves it
    - Logged out: Opens the signup modal

## Added tests?

- [ ] Yes
- [x] No, and this is why: not needed

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

alt_text

### DIFF
--- a/app/assets/javascripts/utilities/buildArticleHTML.js
+++ b/app/assets/javascripts/utilities/buildArticleHTML.js
@@ -237,7 +237,9 @@ function buildArticleHTML(article) {
     var saveButton = '';
     if (article.class_name === 'Article') {
       saveButton =
-        '<button type="button" class="crayons-btn crayons-btn--secondary crayons-btn--s bookmark-button" data-reactable-id="' +
+        '<button type="button" id="article-save-button-' +
+        article.id +
+        '" class="crayons-btn crayons-btn--secondary crayons-btn--s bookmark-button" data-reactable-id="' +
         article.id +
         '">\
                       <span class="bm-initial">Save</span>\
--- a/app/javascript/articles/Feed.jsx
+++ b/app/javascript/articles/Feed.jsx
@@ -1,47 +1,49 @@
-import { h, Component } from 'preact';
+import { h } from 'preact';
+import { useEffect, useState } from 'preact/hooks';
 import PropTypes from 'prop-types';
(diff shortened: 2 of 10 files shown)


Link2Twenty:
Looks like an unrelated failure if someone could run Travis again?

rhymes (approved):
It seems to be working @reobin!

Just a note, please remember to prefix your username to the name of PRs, see the contributing guide for details https://docs.forem.com/contributing/forem/ - thank you!

benhalpern (approved):

PR #11120, opened 2020-10-27
Updating theme dynamically without manual refresh

### LINKED ISSUE #3613: When we change theme, we have to refresh the page to update

shilpi230:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Instead of manually refreshing the page after saving the theme. This ticket fixes it and changes the theme dynamically on save. 

Closes #3613

- Go to User's Settings page.
- Navigate to UX tab, select some theme.
- Theme should get updated website wide.
- [Loom Recording](https://www.loom.com/share/2503b9e5c9d5457692bfbfee861e393c)
## Added tests?

- [ ] Yes
- [x] Its a bug fix which doesn't need test case.

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/assets/javascripts/base.js.erb
+++ b/app/assets/javascripts/base.js.erb
@@ -1,3 +1,5 @@
+//= require jquery
+//= require jquery_ujs
 //= require_tree ./initializers
 //= require_tree ./utilities
 //= require lib/xss
--- a/app/assets/javascripts/utilities/updateStyles.js
+++ b/app/assets/javascripts/utilities/updateStyles.js
@@ -1,0 +1,6 @@
+async function updateStyles(theme) {
+  const css = await fetch(`https://dev.to/assets/themes/${theme}.css`);
+
+  let string = await css.text();
+  document.getElementById('body-styles').innerHTML = `<style>${string}</style>`;
+}
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -42,9 +42,6 @@ class UsersController < ApplicationController
     if @user.update(permitted_attributes(@user))
       RssReaderFetchUserWorker.perform_async(@user.id) if @user.feed_url.present?
       notice = "Your profile was successfully updated."
-      if config_changed?
-        notice = "Your config has been updated. Refresh to see all changes."
-      end
       if @user.export_requested?
         notice += " The export will be emailed to you shortly."
         ExportContentWorker.perform_async(@user.id, @user.email)
@@ -52,7 +49,10 @@ class UsersController < ApplicationController
       cookies.permanent[:user_experience_level] = @user.experience_level.to_s if @user.experience_level.present?
       flash[:settings_notice] = notice
       @user.touch(:profile_updated_at)
-      redirect_to "/settings/#{@tab}"
+      respond_to do |format|
+        format.html { redirect_to "/settings/#{@tab}" }
+        format.js
+      end
     else
       Honeycomb.add_field("error", @user.errors.messages.reject { |_, v| v.empty? })
       Honeycomb.add_field("errored", true)
@@ -318,8 +318,20 @@ class UsersController < ApplicationController
     @tab = current_tab
   end
 
-  def config_changed?
-    params[:user].include?(:config_theme)
+  def less_than_one_day_old?(user)
+    # we check all the `_created_at` fields for all available providers
+    # we use `.available` and not `.enabled` to avoid a situation in which
+    # an admin disables an authentication method after users have already
+    # registered, risking that they would be flagged as new
+    # the last one is a fallback in case all created_at fields are nil
+    user_identity_age = Authentication::Providers.available.map do |provider|
+      user.public_send("#{provider}_created_at")
+    end.detect(&:present?)
+
+    user_identity_age = user_identity_age.presence || 8.days.ago
+
+    range = 1.day.ago.beginning_of_day..Time.current
+    range.cover?(user_identity_age)
   end
 
   def destroy_request_in_progress?
--- a/app/views/users/_ux.html.erb
+++ b/app/views/users/_ux.html.erb
@@ -62,3 +62,19 @@
     <button class="crayons-btn" type="submit">Save</button>
   </div>
 <% end %>
+<script>
+  document.getElementById("user_config_theme_default").onclick = function() {
+    updateStyles("minimal");
+  };
+  var config_themes = ["night", "minimal_light", "pink", "ten_x_hacker"];
+  for( let theme of config_themes){
+    document.getElementById(`user_config_theme_${theme}_theme`).onclick = function() {
+      if (theme === "ten_x_hacker"){
+        theme = "hacker";
+      } else if( theme === "minimal_light" ) {
+        theme = "minimal";
+      }
+      updateStyles(theme);
+    };
+  }
+</script>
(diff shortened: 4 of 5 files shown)


shilpi230:
Can somebody please help me in fixing the test failure. It seems some setup configuration has been missing. Netlify deployment failed.

app/assets/javascripts/base.js.erb
```diff
@@ -1,3 +1,5 @@
+//= require jquery
+//= require jquery_ujs
```
rhymes, line 2: Is there a way to do this without UJS? We purposefully don't include it as we don't use it around the website, neither jQuery

app/assets/javascripts/utilities/updateStyles.js
```diff
@@ -0,0 +1,6 @@
+async function updateStyles(theme) {
+  const css = await fetch(`https://dev.to/assets/themes/${theme}.css`);
```
rhymes, line 2: this should take into account that this is not going to run on DEV only, so you'll have to fetch the path without the URL part: `/assets/themes`.

Also, maybe we can call this `updateTheme` so it's a bit easier to find in the future

app/controllers/users_controller.rb
```diff
@@ -318,8 +318,20 @@ def set_current_tab(current_tab = "profile")
     @tab = current_tab
   end
 
-  def config_changed?
-    params[:user].include?(:config_theme)
+  def less_than_one_day_old?(user)
```
rhymes, line 321: is this a leftover?

app/views/users/_ux.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= javascript_packs_with_chunks_tag "stickySaveFooter", defer: true %>
 
-<%= form_for @user, html: { id: "ux-customization-form", class: "sticky-footer-form grid gap-6" } do |f| %>
+<%= form_for @user, remote: true, authenticity_token: true, html: { id: "ux-customization-form", class: "sticky-footer-form grid gap-6" } do |f| %>
```
rhymes, line 3: I think you can replace by manually using `sendFetch` on save https://github.com/forem/forem/blob/master/app/assets/javascripts/utilities/sendFetch.js

app/controllers/users_controller.rb
```diff
@@ -42,17 +42,17 @@ def update
     if @user.update(permitted_attributes(@user))
       RssReaderFetchUserWorker.perform_async(@user.id) if @user.feed_url.present?
       notice = "Your profile was successfully updated."
-      if config_changed?
-        notice = "Your config has been updated. Refresh to see all changes."
-      end
```
rhymes, line 47: this should probably stay

rhymes (changes requested):
Hi @shilpi230, thanks for taking this on. I played a bit with it and it's super cool!

Aside from the inline notes I noticed that it's triggered on radio select, which makes the "save" button a little bit of a UX conundrum, maybe @ludwiczakpawel can weigh in here

The failing test is because you have removed dire noticed in the redirect that should probably stay there as it's referred to any config save

ludwiczakpawel (changes requested):
For some reason, when I switch to _default_ theme, the _minimal light_ is actually being loaded.

---

Also, agree with @rhymes there's something off with UX. When you switch the theme, then go to another page without saving  you can still see that new theme. And that's a little suboptimal if you haven't saved it.

app/assets/javascripts/utilities/updateStyles.js
shilpi230: sure will take this in account and makes changes! Thanks!

app/controllers/users_controller.rb
shilpi230: no.

app/controllers/users_controller.rb
shilpi230: @citizen428 asked me to remove this piece of code. Once manual refresh is fixed.

shilpi230:
@ludwiczakpawel _"For some reason, when I switch to default theme, the minimal light is actually being loaded."_ i.e because I thought for default we are using minimal light theme. Which I am assuming is not correct from your statement. I will fix that. Thanks @rhymes @ludwiczakpawel for your comments. Will update those changes. Thanks!

shilpi230:
@rhymes test case is failing in this file `spec/system/user_selects_looking_for_work_spec.rb` Which is expecting this `Your profile has been updated`  but getting something like this `DEV(local)\nWrite a post\nSetup not completed yet, missing suggested tags and suggested users. Please visit the configuration page.` Which is somehow telling that the some configuration is missing.

app/controllers/users_controller.rb
rhymes: so what is it used for? I can't find it used in this PR and it's not present in master 😕

app/assets/javascripts/utilities/updateStyles.js
nickytonline: As @rhymes suggested, so this should be good.

```suggestion
async function updateStyles(theme) {
  const css = await fetch(`/assets/themes/${theme}.css`);
```

app/views/users/_ux.html.erb
```diff
@@ -62,3 +62,19 @@
     <button class="crayons-btn" type="submit">Save</button>
   </div>
 <% end %>
+<script>
+  document.getElementById("user_config_theme_default").onclick = function() {
+    updateStyles("minimal");
+  };
+  var config_themes = ["night", "minimal_light", "pink", "ten_x_hacker"];
+  for( let theme of config_themes){
```
nickytonline, line 70: refactor (blocking): This means we're adding `n` number of onclicks for changing themes. We can do this with one `click` handler via `addEventListener` and in the event handler, we can detect which theme was clicked.

Here's some sample code.

```javascript
document.querySelector('.theme-selector-field').addEventListener('click', e => {
 const { target } = e;

 if (target.getAttribute('name') === 'user[config_theme]') {
    alert(`You clicked on the ${target.value} theme`);
 }
});
```

Also, the radio button group should be wrapped in a `<fieldset />` instead of a `<div />`, but that issue is not related to this PR.

nickytonline (changes requested):
@shilpi230, thanks for this. Just some request changes.

app/controllers/users_controller.rb
shilpi230: Ohh really sorry for this,  I think its leftover from conflict that I resolved. Will remove it if its not in master. Really sorry for this. Thanks!

app/views/users/_ux.html.erb
nickytonline: I created #11243 for the fieldset issue mentioned.

app/controllers/users_controller.rb
rhymes: ahah no worries :) thanks for taking the time to look into this!

cmgorton:
Hi @shilpi230 !

PR #11121, opened 2020-10-27
Enable Puma's nakayoshi_fork feature

benhalpern:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

gc.compact

Here are a few posts about `GC.compact` in Rails....

- https://dev.to/palkan_tula/cables-profiling-returns-gc-compact-jemalloc-33c9
- https://engineering.appfolio.com/appfolio-engineering/2019/3/22/ruby-27-and-the-compacting-garbage-collector
- https://www.infoworld.com/article/3404481/ruby-27-improves-garbage-collection-pattern-matching.html

My thoughts here are that if tests pass, this is something we could deploy and observe. There may be more ways we could use this in specific cases.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -26,10 +26,9 @@ pidfile ENV.fetch("PIDFILE") { "tmp/pids/server.pid" }
 #
 workers ENV.fetch("WEB_CONCURRENCY", 2)
 
-# Code that runs before the app begins forking
-before_fork do
-  GC.compact
-end
+# calls GC a handful of times and compacts the heap on Ruby 2.7+ before forking.
+# This may reduce memory usage of Puma on MRI.
+nakayoshi_fork
 
 # Use the `preload_app!` method when specifying a `workers` number.
 # This directive tells Puma to first boot the application and load code
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -26,6 +26,11 @@ pidfile ENV.fetch("PIDFILE") { "tmp/pids/server.pid" }
 #
 workers ENV.fetch("WEB_CONCURRENCY", 2)
 
+# Code that runs before the app begins forking
+before_fork do
+  GC.compact
+end
+
 # Use the `preload_app!` method when specifying a `workers` number.
 # This directive tells Puma to first boot the application and load code
 # before forking the application. This takes advantage of Copy On Write


krzysiek1507:
Hi! If you deploy this, could you please share the stats?

benhalpern:
@krzysiek1507 absolutely. If folks feel like this is worth attempting a deploy, I think the immediate change in memory would 10000% be worth publishing. 😄

krzysiek1507:
That'd be great! :D 

Have you considered https://fullstaqruby.org/?

benhalpern:
@krzysiek1507 that is very interesting.... And we'll definitely be looking more and more at this sort of stuff, as we want to ensure the cost-effectiveness of the Forem network. Most of our choices have been DEV-centric, and while we certainly cared about efficiency, it will be exciting to explore solutions now that resource management is much more central to our concerns.

.... So no, haven't really explored fullstaq in detail, but I think we'll be paying much more attention to innovations in this space going forward. I'd think @citizen428 has been paying closest attention to this stuff from our team.

mstruve (approved):
Tested locally and works well, why not give it a shot! 🚀

rhymes:
Seems like Puma 5 had added a similar experimental feature in its release in cluster mode, they are also inviting people to contribute data:

> If you upgrade and try any of the 3 new features, please post before and after results or screenshots to this Github issue. “It didn’t do anything” is still a useful report in this case. Posting ~24 hours of “before” and ~24 hours of “after” data would be most helpful.

from https://www.speedshop.co/2020/09/17/we-made-puma-faster-with-sleep-sort.html

Maybe it's worth testing `nakayoshi_fork` ?

benhalpern:
@rhymes Good call. I've experimented with the old gem, but this being the more straightforward use of `GC.compact` seems ideal.

I took some "before" screenshots.

citizen428:
@benhalpern Hm, I could have sworn I added this/recommend adding this way before I even joined the team.. 😕 Anyway, 👍 from me and 

@krzysiek1507 I've been following Fullstaq Ruby a bit. To me this seems like REE (Ruby Enterprise Edition) revamped, which I used way back in the day before that went EOL. It's certainly interesting, but we already use jemalloc, which means the only other things this offers is the `malloc_trim` patch. While that could certainly be useful, I'd hope that it or something like it finds its way into mainline Ruby if proven useful in the wild. Personally I'm much more interested in TruffleRuby, especially on Substrate VM.

citizen428 (approved):

benhalpern:
> Hm, I could have sworn I added this/recommend adding this way before I even joined the team... 😕 Anyway, 👍 from me.

@citizen428 We did go on this recommendation, but it didn't change much so we pulled it out to follow up with other experiments in the future. However, this was pre Ruby 2.7 when it was a gem and not something built into Puma with a couple lines of code, and a very different time in the project in terms of what the app does and how we configure it, so I suspect a better analysis.

So yeah, we followed the recommendation, but didn't really see much of a difference, so went with the the choice of being closer to defaults/conventions (all else equal).... Here's hoping this time we see more of an impact.

nateberkopec:
Watching this to hear any impact you see from `nakayoshi_fork`.

> which means the only other things this offers is the malloc_trim patch

as far as I remember, malloc_trim performs about as well as jemalloc.

citizen428:
Thanks @benhalpern, that clears it up, let's hope we see bigger improvements this time then.

benhalpern:
https://forem.dev/foremteam/results-of-implementing-puma-s-nakayoshifork-on-forem-3pdl

PR #11005, opened 2020-10-21
Placeholder assets

ludwiczakpawel:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR introduces placeholder images for:

- mascot user,
- social media image
- icon (PWA etc)

## Related Tickets & Documents

- https://github.com/forem/InternalProjectPlanning/issues/94
- https://github.com/forem/InternalProjectPlanning/issues/92
- https://github.com/forem/InternalProjectPlanning/issues/93

## QA Instructions, Screenshots, Recordings

**Mascot user:**

mascot

**Social media image:**

social-media-cover

**Icon:**

icon

## Added tests?

- [ ] yes
- [x]  no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

PR #9747, opened 2020-08-11
Fall back to configured social image instead of hardcoded

benhalpern:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Pretty straightforward, no need to fall back to hardcoded image.

Addresses: https://github.com/forem/InternalProjectPlanning/issues/93

PR #11126, opened 2020-10-27
Add default for main_social_image

atsmith813:
- [x] Refactor

## Description
This PR makes setting the `main_social_image` optional when setting up a Forem by setting a default value in the `SiteConfig`. This is part of @forem/creator-onboarding's efforts to reduce the amount of required configs as much as impossible and follows up on https://github.com/forem/forem/pull/11005 and https://github.com/forem/forem/pull/9747

Closes https://github.com/forem/InternalProjectPlanning/issues/93

1. Fire up the app locally and navigate to [http://localhost:3000/admin/config](http://localhost:3000/admin/config).
2. Verify that under **Get Started** you no longer see a **Main social image** field.
3. Verify that under **Images** you see the default value set for **Main social image**
Screen Shot 2020-10-27 at 4 37 07 PM

## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

unsure_gif

### DIFF
--- a/app/lib/url.rb
+++ b/app/lib/url.rb
@@ -60,11 +60,12 @@ module URL
     nil
   end
 
-  # Creatres an Image URL - a shortcut for the .image_url helper
+  # Creates an Image URL - a shortcut for the .image_url helper
   #
   # @param image_name [String] the image file name
-  def self.local_image(image_name)
-    host = ActionController::Base.asset_host ||= url(nil)
+  # @param host [String] (optional) the host for the image URL you'd like to use
+  def self.local_image(image_name, host: nil)
+    host ||= ActionController::Base.asset_host || url(nil)
     ActionController::Base.helpers.image_url(image_name, host: host)
   end
 
--- a/app/lib/url.rb
+++ b/app/lib/url.rb
@@ -7,7 +7,11 @@ module URL
   end
 
   def self.domain
-    SiteConfig.respond_to?(:app_domain) ? SiteConfig.app_domain : ApplicationConfig["APP_DOMAIN"]
+    if Rails.application&.initialized? && SiteConfig.respond_to?(:app_domain)
+      SiteConfig.app_domain
+    else
+      ApplicationConfig["APP_DOMAIN"]
+    end
   end
 
   def self.url(uri = nil)
--- a/app/lib/url.rb
+++ b/app/lib/url.rb
@@ -7,7 +7,7 @@ module URL
   end
 
   def self.domain
-    Rails.application&.initialized? ? SiteConfig.app_domain : ApplicationConfig["APP_DOMAIN"]
+    SiteConfig.respond_to?(:app_domain) ? SiteConfig.app_domain : ApplicationConfig["APP_DOMAIN"]
   end
 
   def self.url(uri = nil)
@@ -60,7 +60,8 @@ module URL
   #
   # @param image_name [String] the image file name
   def self.local_image(image_name)
-    ActionController::Base.helpers.image_url(image_name)
+    host = ActionController::Base.asset_host ||= url(nil)
+    ActionController::Base.helpers.image_url(image_name, host: host)
   end
 
   def self.organization(organization)
--- a/app/lib/url.rb
+++ b/app/lib/url.rb
@@ -7,7 +7,7 @@ module URL
   end
 
   def self.domain
-    SiteConfig.respond_to?(:app_domain) ? SiteConfig.app_domain : ApplicationConfig["APP_DOMAIN"]
+    Rails.application&.initialized? ? SiteConfig.app_domain : ApplicationConfig["APP_DOMAIN"]
   end
 
   def self.url(uri = nil)
@@ -56,11 +56,11 @@ module URL
     nil
   end
 
-  # Creates an Image URL - a shortcut for the .image_url helper
+  # Creatres an Image URL - a shortcut for the .image_url helper
   #
   # @param image_name [String] the image file name
   def self.local_image(image_name)
-    host = ActionController::Base.asset_host ||= url(nil)
+    host = ActionController::Base.asset_host ||= "#{protocol}#{ENV['APP_DOMAIN']}"
     ActionController::Base.helpers.image_url(image_name, host: host)
   end
 
--- a/app/lib/url.rb
+++ b/app/lib/url.rb
@@ -60,7 +60,7 @@ module URL
   #
   # @param image_name [String] the image file name
   def self.local_image(image_name)
-    host = ActionController::Base.asset_host ||= "#{protocol}#{ENV['APP_DOMAIN']}"
+    host = ActionController::Base.asset_host ||= "#{protocol}#{ApplicationConfig['APP_DOMAIN']}"
     ActionController::Base.helpers.image_url(image_name, host: host)
   end
 
--- a/app/lib/url.rb
+++ b/app/lib/url.rb
@@ -56,6 +56,13 @@ module URL
     nil
   end
 
+  # Creatres an Image URL - a shortcut for the .image_url helper
+  #
+  # @param image_name [String] the image file name
+  def self.local_image(image_name)
+    ActionController::Base.helpers.image_url(image_name)
+  end
+
   def self.organization(organization)
     url(organization.slug)
   end
--- a/app/lib/url.rb
+++ b/app/lib/url.rb
@@ -7,7 +7,7 @@ module URL
   end
 
   def self.domain
-    Rails.application&.initialized? ? SiteConfig.app_domain : ApplicationConfig["APP_DOMAIN"]
+    SiteConfig.respond_to?(:app_domain) ? SiteConfig.app_domain : ApplicationConfig["APP_DOMAIN"]
   end
 
   def self.url(uri = nil)
@@ -60,7 +60,7 @@ module URL
   #
   # @param image_name [String] the image file name
   def self.local_image(image_name)
-    host = ActionController::Base.asset_host ||= "#{protocol}#{ApplicationConfig['APP_DOMAIN']}"
+    host = ActionController::Base.asset_host ||= url(nil)
     ActionController::Base.helpers.image_url(image_name, host: host)
   end
 
--- a/app/lib/url.rb
+++ b/app/lib/url.rb
@@ -56,7 +56,7 @@ module URL
     nil
   end
 
-  # Creatres an Image URL - a shortcut for the .image_url helper
+  # Creates an Image URL - a shortcut for the .image_url helper
   #
   # @param image_name [String] the image file name
   def self.local_image(image_name)
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -12,6 +12,7 @@ class SiteConfig < RailsSettings::Base
 
   STACK_ICON = File.read(Rails.root.join("app/assets/images/stack.svg")).freeze
   LIGHTNING_ICON = File.read(Rails.root.join("app/assets/images/lightning.svg")).freeze
+  MAIN_SOCIAL_IMAGE = File.read(Rails.root.join("app/assets/images/social-media-cover.png")).freeze
 
   # Meta
   field :admin_action_taken_at, type: :datetime, default: Time.current
@@ -84,7 +85,7 @@ class SiteConfig < RailsSettings::Base
   field :recaptcha_secret_key, type: :string, default: ApplicationConfig["RECAPTCHA_SECRET"]
 
   # Images
-  field :main_social_image, type: :string
+  field :main_social_image, type: :string, default: MAIN_SOCIAL_IMAGE
   field :favicon_url, type: :string, default: "favicon.ico"
   field :logo_png, type: :string
   field :logo_svg, type: :string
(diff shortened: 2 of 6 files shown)


app/models/site_config.rb
```diff
@@ -12,6 +12,7 @@ class SiteConfig < RailsSettings::Base
 
   STACK_ICON = File.read(Rails.root.join("app/assets/images/stack.svg")).freeze
   LIGHTNING_ICON = File.read(Rails.root.join("app/assets/images/lightning.svg")).freeze
+  MAIN_SOCIAL_IMAGE = ActionController::Base.helpers.image_url("social-media-cover.png").freeze
```
atsmith813, line 15: Any suggestions for a better way to do this? This feels a little funky to me 🤔

benhalpern: I think when including helpers this is kind of the appropriate way to do it.....

We do have these littered in different parts of the codebase, and it might make sense to create an object that wraps this class just for a cleaner API? 🤷‍♂️

vaidehijoshi (commented):
This works well, but as I was playing with it locally, I had some thoughts about how the "default image" feature will be experienced by Forem creators. Curious to hear @atsmith813's and others' thoughts here.

Screen_Shot_2020-10-27_at_3_01_23_PM

atsmith813:
@vaidehijoshi agreed - that part is clunky 🤔  in production this value will be full URL. We don't set an `asset_host` in development so the host is `nil` making it look like it's just the path.

app/models/site_config.rb
citizen428: Theoretically, we could lift it into our `URL` module: `URL.image`, which just wraps the `ActionController` part. Feels less clunky and if Rails ever adds a different/better way to do this we only have to update it in one place.

citizen428 (approved):

ludwiczakpawel:
@vaidehijoshi re what you said about removing social media cover image value and not being able to get it back again:
my thinking is: if you don't set a value for social media image (means: not provide your own OR removing the value form field), your forem will use default placeholder and that's it. no need for checkboxes or toggles.

app/models/site_config.rb
atsmith813: Updated!

atsmith813:
@vaidehijoshi I looked into this more 🤔 . I think what you're describing is if a user clears the value in `main_social_image` and submits the form, it should revert back to the default value instead of leaving the value blank. Is that correct?

If so, I took a glance into [the `rails-settings-cached` gem powering the SiteConfig](https://github.com/huacnlee/rails-settings-cached) and there doesn't seem to be an easy way to get to the `default` value short of looping through the configs and then setting it to the default value if the current value is blank in the controller. My concern with that approach besides performance over time is that there may be cases where leaving a field blank is intentionally unsettling something 🤔 .

And that sounds like it goes back to your suggestion of a checkbox to "use default". I can get behind that idea so that if you clear the value of the box and _**don't**_ select the "use default" checkbox, the value is, in fact, erased.

That logic would really need to go on every field and feels like a bigger SiteConfig/admin change that may be better for a separate PR. Maybe @forem/smash has thoughts or whats to take that one on?

cc: @ludwiczakpawel 

Also, wrt to the URL, It looks like this won't be a full URL in production after all. We only set the `asset_host` on the DEV instance, not Forem Cloud instances in production 😢

atsmith813:
WRT to the value being a URL in production, I'm not entirely sure now. We only set `asset_host` in the DEV instance 😢 . I thought it would look at the app domain by default, but that doesn't seem to be true in development. I'll keep exploring it.

vaidehijoshi:
> @vaidehijoshi re what you said about removing social media cover image value and not being able to get it back again:
> my thinking is: if you don't set a value for social media image (means: not provide your own OR removing the value form field), your forem will use default placeholder and that's it. no need for checkboxes or toggles.

> @vaidehijoshi I looked into this more 🤔 . I think what you're describing is if a user clears the value in main_social_image and submits the form, it should revert back to the default value instead of leaving the value blank. Is that correct?

@atsmith813 + @ludwiczakpawel The situation I was thinking of is this:
1) I'm a Forem creator and I'm setting up my new Forem. I see that my Forem instance comes with all these nice default images, like a logo and a social image -- great!
2) I delete the main social image url and I use my own image. I preview my image and I decide that I actually don't like it, and I want to use what Forem provided me to start.
3) I go back to `/admin/config` but then I realize that I can't get that "default provided image from Forem" back again.

I realize this is a small thing, but I just worry that it might happen. Sure, as a Forem creator, I can upload some _other_ image, but maybe I'm not technically savvy or maybe I realize that I need to crop and resize the image I want to use, or some other reason. This could be solved by us just adding some text (`This image is a default - if you choose to remove it, you'll need to provide your own`), or by allowing the user to _always_ be able to use the default image so that even if they upload their own and decide to back out of their decision, they have our default to fall back on.

I don't think that this needs to block this PR, but I do think it's something we should think about as we start providing defaults to Forem creators, because those creators might approach the configuration process with very different mindsets.

app/lib/url.rb
```diff
@@ -56,6 +56,13 @@ def self.user(user)
     nil
   end
 
+  # Creatres an Image URL - a shortcut for the .image_url helper
```
vaidehijoshi, line 59:
```suggestion
  # Creates an Image URL - a shortcut for the .image_url helper
```

app/models/site_config.rb
```diff
@@ -12,6 +12,7 @@ class SiteConfig < RailsSettings::Base
 
   STACK_ICON = File.read(Rails.root.join("app/assets/images/stack.svg")).freeze
   LIGHTNING_ICON = File.read(Rails.root.join("app/assets/images/lightning.svg")).freeze
```
vaidehijoshi, line 14: non-blocker: oooh, it would be great if we could create a helper in `URL` for these, too. i'm not a big fan of us doing `File.read` directly within the `SiteConfig`

app/lib/url.rb
```diff
@@ -56,6 +56,13 @@ def self.user(user)
...
+  #
+  # @param image_name [String] the image file name
+  def self.local_image(image_name)
+    ActionController::Base.helpers.image_url(image_name)
+  end
```
vaidehijoshi, line 64: This is great!!

vaidehijoshi (approved):

app/models/site_config.rb
atsmith813: These are actually SVGs - the file contents itself go into the field so we need `File.read`
Screen Shot 2020-10-28 at 11 46 26 AM

app/lib/url.rb
benhalpern: Yeah, good improvement.

atsmith813:
@vaidehijoshi @citizen428 I'm going to bring this back to draft because I want to explore the whole "value should be a URL" thing to make it more intuitive. I've got some ideas that I think will clear it up.

app/lib/url.rb
```diff
@@ -7,7 +7,7 @@ def self.protocol
   end
 
   def self.domain
-    Rails.application&.initialized? ? SiteConfig.app_domain : ApplicationConfig["APP_DOMAIN"]
+    SiteConfig.respond_to?(:app_domain) ? SiteConfig.app_domain : ApplicationConfig["APP_DOMAIN"]
```
atsmith813, line 10: This didn't work as we expected. Because we ended up calling this method from within `SiteConfig` itself, we hit a weird edge case where the original `Rails.application&.initialized?` logic was `true`, but `app_domain` hadn't been defined yet on `SiteConfig` so we ended up with an `undefined method` error when starting up the app.

app/lib/url.rb
```diff
@@ -56,6 +56,14 @@ def self.user(user)
     nil
   end
 
+  # Creates an Image URL - a shortcut for the .image_url helper
+  #
+  # @param image_name [String] the image file name
+  def self.local_image(image_name)
+    host = ActionController::Base.asset_host ||= url(nil)
+    ActionController::Base.helpers.image_url(image_name, host: host)
+  end
```
atsmith813, line 65: I did some more digging on this and it turns out unless you explicitly define `asset_host`, `image_url` will basically return a relative path.

In the DEV instance, we set the `asset_host` https://github.com/forem/forem/blob/034b049e3d69ce2417440decd4c09bcf91f623f4/config/environments/production.rb#L54

In other instances, we do _not_ define `FASTLY_CDN_URL` - making it `nil` and returning a `nil` host. When you provide a `host` to `image_url`, it overrides whatever is set as `assset_host`. So here, we basically reassign the `host` explicitly if one is defined (i.e. on the DEV instance where we use the Fastly CDN to serve assets) or we set it to the app protocol/domain (i.e. every other instance of Forem right now) to return a pretty URL.

Screen Shot 2020-10-28 at 3 30 07 PM

Another option is to change the logic in `production.rb` so that it sets the `asset_host` based on this same criteria. But I figured we can re-visit that for when we move other instances of Forem to a CDN and we have to set that value at that point anyways. What do you think?

cc: @benhalpern @vaidehijoshi @citizen428 @jdoss

app/lib/url.rb
```diff
@@ -7,7 +7,11 @@ def self.protocol
   end
 
   def self.domain
-    Rails.application&.initialized? ? SiteConfig.app_domain : ApplicationConfig["APP_DOMAIN"]
+    if Rails.application&.initialized? && SiteConfig.respond_to?(:app_domain)
+      SiteConfig.app_domain
+    else
+      ApplicationConfig["APP_DOMAIN"]
+    end
```
atsmith813, line 14: This didn't work as we expected. Because we ended up calling this method from within `SiteConfig` itself, we hit a weird edge case where the original `Rails.application&.initialized?` logic was `true`, but `app_domain` hadn't been defined yet on `SiteConfig` so we ended up with an `undefined method` error when starting up the app.

cc: @benhalpern @vaidehijoshi @citizen428 @jdoss

app/lib/url.rb
```diff
@@ -56,6 +60,15 @@ def self.user(user)
...
 
+  # Creates an Image URL - a shortcut for the .image_url helper
+  #
+  # @param image_name [String] the image file name
+  # @param host [String] (optional) the host for the image URL you'd like to use
+  def self.local_image(image_name, host: nil)
+    host ||= ActionController::Base.asset_host || url(nil)
+    ActionController::Base.helpers.image_url(image_name, host: host)
+  end
+
```
atsmith813, line 71: I did some more digging on this and it turns out unless you explicitly define `asset_host`, `image_url` will basically return a relative path.

In the DEV instance, we set the `asset_host` https://github.com/forem/forem/blob/034b049e3d69ce2417440decd4c09bcf91f623f4/config/environments/production.rb#L54

In other instances, we do _not_ define `FASTLY_CDN_URL` - making it `nil` and returning a `nil` host. When you provide a `host` to `image_url`, it overrides whatever is set as `assset_host`. So here, we basically reassign the `host` explicitly if one is defined (i.e. on the DEV instance where we use the Fastly CDN to serve assets) or we set it to the app protocol/domain (i.e. every other instance of Forem right now) to return a pretty URL.

Screen Shot 2020-10-28 at 3 30 07 PM

Another option is to change the logic in `production.rb` so that it sets the `asset_host` based on this same criteria. But I figured we can re-visit that for when we move other instances of Forem to a CDN and we have to set that value at that point anyway. What do you think?

cc: @benhalpern @vaidehijoshi @citizen428 @jdoss

atsmith813:
I made some notable changes so I'm re-requesting review @vaidehijoshi @citizen428

app/lib/url.rb
vaidehijoshi: It might be nice to do this in `production.rb` _preemptively_, just because I think the fact that `asset_host` being tied closely to `FASTLY_CDN_URL` is going to trip us up at some point in the future, too. It would be nice to have a `asset_host` fallback in production environments so that it's never `nil`.

But I think this solution works well, too. Mostly I'm trying to future-proof this for us down the road.

vaidehijoshi (approved):
Great work here, @atsmith813 👏

citizen428 (approved):
Very nice, thanks @atsmith813! 🚀

rhymes (approved):
Great work!

ludwiczakpawel:
@vaidehijoshi 

> _I delete the main social image url_ and I use my own image. I preview my image and I decide that I actually don't like it, and I want to use what Forem provided me to start.

I guess I'm thinking slightly different here: I don't have to delete anything: I just provide my image and call it a day (ok, replacing URL, kinda indicates i "delete" the original one :P). If later on I decide it's wrong, I remove my image (clears the value of that URL field) and hit "Submit" - and from now on the default Forem social media image is going to be used instead. No action required.
This also means: admin can not remove social media image entirely. It's either their custom one or our default one (which I think is fair constraint).

vaidehijoshi:
> If later on I decide it's wrong, I remove my image (clears the value of that URL field) and hit "Submit" - and from now on the default Forem social media image is going to be used instead. No action required.
> 
> This also means: admin can not remove social media image entirely. It's either their custom one or our default one (which I think is fair constraint).

@ludwiczakpawel  I think this would work and solve the issue that I'm worried about! The only problem is that this is not currently how the default images work on SiteConfig, so we would need to build for this (if you check out this branch and try it out for example, it won't do that, and if you check out other default images on SiteConfig on the main branch, it also won't do that). And as @atsmith813 mentioned, this will involve a larger overhaul of how we store defaults in SiteConfig...but I agree with you that it should work this way and an an admin should never be able to fully remove these values that explicitly have been given defaults.

ludwiczakpawel:
Gotcha, understood! :)

PR #11127, opened 2020-10-27
Grab email for existing Twitter users if previous email was blank

Zhao-Andy:
- [x] Bug Fix

## Description
So there's an edge case where Forems that did not enable the correct permissions for getting the user's email address through the Twitter OAuth process were not able to get the user's email.

This PR fixes it, but it may not be the _best_ way to do so. First, it's a monkey patch specific for Twitter as opposed to other providers, and I don't think it's good practice to have code or solutions like this. For this to resolve the problem, users will need to sign in through Twitter again, and the email should pull in properly now. Unfortunately, we can't do anything until they've accepted the new forem-specific Twitter OAuth app's permissions.

## Added tests?
- [x] No, and this is why: it's a monkey patch and sort of a PoC

## Added to documentation?
- [x] No documentation needed

### DIFF
--- a/app/services/authentication/authenticator.rb
+++ b/app/services/authentication/authenticator.rb
@@ -130,6 +130,9 @@ module Authentication
       user.tap do |model|
         user.unlock_access! if user.access_locked?
         user.assign_attributes(provider.existing_user_data)
+        if user.email.blank? && provider.name == "twitter"
+          user.update_column(:email, provider.get_email)
+        end
 
         update_profile_updated_at(model)
 
--- a/app/services/authentication/providers/twitter.rb
+++ b/app/services/authentication/providers/twitter.rb
@@ -28,6 +28,10 @@ module Authentication
         }
       end
 
+      def get_email
+        info.email.to_s
+      end
+
       def self.settings_url
         SETTINGS_URL
       end


app/services/authentication/authenticator.rb
```diff
@@ -130,6 +130,9 @@ def update_user(user)
       user.tap do |model|
         user.unlock_access! if user.access_locked?
         user.assign_attributes(provider.existing_user_data)
+        if user.email.blank? && provider.name == "twitter"
+          user.update_column(:email, provider.get_email)
```
Zhao-Andy, line 134: I'm using `update_column` here because it won't trigger a confirmation email. I'm assuming the email they have on Twitter is valid already, like we do with the sign up process.

app/services/authentication/providers/twitter.rb
```diff
@@ -28,6 +28,10 @@ def existing_user_data
         }
       end
 
+      def get_email
+        info.email.to_s
+      end
```
Zhao-Andy, line 33: This method has to be in this class because `info` is `protected`, and can't be in called in the `Authenticator` class.

app/services/authentication/authenticator.rb
rhymes: Can you switch to `update_columns(email: email)` which is a bit clearer? I wonder if there's a way to test this

app/services/authentication/providers/twitter.rb
rhymes: I think we should promote this to the base class, as it'd be weird to have a method that's only available on one provider, we coul also just call it `user_email` similar to:

```ruby
      def user_nickname
        info.nickname.to_s
      end
```

you can find in the `Provider` base class

rhymes (commented):
I agree with you, it feels like a patch for something that shouldn't happen in the first place. What if we find a way to avoid this situation and then run a data update script to fix the email of existing users? Identities have the raw payload so we can get the Twitter email from there, without requiring the user to sign up again. Would that work?

Zhao-Andy:
@rhymes thanks for the review! Agree that a data update script would work, but I wouldn't want to run that same script for all the other Forems. I'm starting to think that running a script in console is the best way now, since this is really a one time issue for a single Forem.

rhymes:
@Zhao-Andy but then, if it's a one time issue, do we need the PR at all? Can't we just use a script?

Zhao-Andy:
I'm closing this in favor of running a script in Rails console. All feedback gathered is at least trending toward _not_ having a solution in code for this, and I think doing this in Rails console is better than a data update script.

Zhao-Andy:
Oops @rhymes  seeing your message now. Yeah we can just use a script 😅

PR #11130, opened 2020-10-28
[Crayons] Update tag page header

lisasy:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
Update tag page header to Crayons styling.

Closes https://github.com/orgs/forem/projects/8#card-46404965

Proposed design fixes:
Screen Shot 2020-10-27 at 5 55 26 PM

Screen Shot 2020-10-27 at 5 55 37 PM

## Added tests?

- [ ] Yes
- [x] No, and this is why: these style changes don't require tests.

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/views/articles/tag_index.html.erb
+++ b/app/views/articles/tag_index.html.erb
@@ -4,31 +4,34 @@
 <% expiry_minutes = params[:timeframe].blank? || params[:timeframe] == "latest" ? 4 : 20 %>
 <% flag = true %>
 <% cache("tag-stories-index-#{params}-#{flag}-#{@tag_model.updated_at}-#{user_signed_in?}", expires_in: expiry_minutes.minutes) do %>
-  <style>
-    .tag-header {
-      border: 2px solid<%= HexComparer.new([@tag_model.bg_color_hex || "#0000000", @tag_model.text_color_hex || "#ffffff"]).brightness(0.88) %>;
-      box-shadow: 5px 6px 0px<%= HexComparer.new([@tag_model.bg_color_hex || "#0000000", @tag_model.text_color_hex || "#ffffff"]).brightness(0.88) %>;
-    }
-  </style>
-  <div class="user-profile-header tag-header">
-    <div class="tag-or-query-header-container">
-      <h1>
-        <% if @tag_model.badge_id && @tag_model.badge %>
-            <img src="<%= optimized_image_url(@tag_model.badge.badge_image_url, width: 180) %>" style="transform: rotate(<%= -25 + (@tag_model.id.digits.first * 5) %>deg);" />
-        <% end %>
-        <% if @tag_model && @tag_model.pretty_name.present? %>
-          <%= @tag_model.pretty_name %>
-        <% else %>
-          <%= @tag %>
-        <% end %>
+  <div class="crayons-layout">
+    <header class="tag-header crayons-card branded-4 relative" style="border-top-color: <%= HexComparer.new([@tag_model.bg_color_hex || "#0000000", @tag_model.text_color_hex || "#ffffff"]).brightness(0.88) %>">
+      <div class="tag-header__title flex justify-between items-center mb-2 s:mb-4">
+        <div class="inline-flex items-center">
+          <h1 class="crayons-title">
+            <% if @tag_model && @tag_model.pretty_name.present? %>
+              <%= @tag_model.pretty_name %>
+            <% else %>
+              <span class="crayons-tag__prefix">#</span>
+              <%= @tag %>
+            <% end %>
+          </h1>
+          <% if @tag_model.badge_id && @tag_model.badge %>
+              <img src="<%= optimized_image_url(@tag_model.badge.badge_image_url, width: 64) %>" class="" style="transform: rotate(<%= -25 + (@tag_model.id.digits.first * 5) %>deg); width: 64px; height: 64px;" />
+          <% end %>
+        </div>
         <% if @tag_model %>
-            <span class="user-profile-follow-button-wrapper">
-              <button id="user-follow-butt" class="crayons-btn follow-action-button user-profile-follow-button" style="color:<%= @tag_model.text_color_hex %>;background-color:<%= @tag_model.bg_color_hex %>" data-info='{"id":<%= @tag_model.id %>,"className":"Tag"}'>&nbsp;</button>
-            </span>
+          <button id="user-follow-butt" class="crayons-btn crayons-btn--secondary follow-action-button" data-info='{"id":<%= @tag_model.id %>,"className":"Tag"}'>&nbsp;</button>
         <% end %>
-      </h1>
-    </div>
+      </div>
+      <% if @tag_model && @tag_model.short_summary.present? %>
+        <p>
+          <%= sanitize @tag_model.short_summary %>
+        </p>
+      <% end %>
+    </header>
   </div>
+
   <div class="home sub-home" id="index-container"
       data-params="<%= params.merge(sort_by: "hotness_score", sort_direction: "desc").to_json(only: %i[tag username sort_by sort_direction]) %>" data-which=""
       data-tag="<%= @tag %>"
--- a/app/views/articles/tags/_sidebar.html.erb
+++ b/app/views/articles/tags/_sidebar.html.erb
@@ -2,8 +2,9 @@
   <div id="sidebar-wrapper-left" class="sidebar-wrapper sidebar-wrapper-left">
     <div class="sidebar-bg" id="sidebar-bg-left"></div>
     <div class="side-bar">
+
       <% if @tag_model && @tag_model.short_summary.present? %>
-        <div class="widget">
+        <div class="widget block s:hidden">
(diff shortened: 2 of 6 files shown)


lisasy:
@ludwiczakpawel I went a little bit off of the original design spec when it came to the tag brand image; originally, I had it pinned to the bottom right with absolute positioning, but having an absolutely positioned image was tricky with the body paragraph that could either not exist at all, or be really long.

It seemed like appending the image next to the tag title would be the *easiest* to adapt to the different scenarios. What do you think?

👇 Here is a random attempt at using flex-box instead of absolute positioning. As you see, the wide container width of the body paragraph seems tricky to work with.

Screen Shot 2020-10-27 at 5 51 08 PM

app/views/articles/tags/_sidebar.html.erb
```diff
@@ -2,8 +2,9 @@
   <div id="sidebar-wrapper-left" class="sidebar-wrapper sidebar-wrapper-left">
     <div class="sidebar-bg" id="sidebar-bg-left"></div>
     <div class="side-bar">
+
       <% if @tag_model && @tag_model.short_summary.present? %>
-        <div class="widget">
+        <div class="widget block s:hidden">
```
lisasy, line 7: @lisasy remove this change.

app/views/articles/tags/_sidebar.html.erb
```diff
@@ -12,6 +13,7 @@
           </div>
         </div>
       <% end %>
+
```
lisasy, line 16: @lisasy remove new line

app/views/articles/tag_index.html.erb
```diff
@@ -4,31 +4,34 @@
...
-            <img src="<%= optimized_image_url(@tag_model.badge.badge_image_url, width: 180) %>" style="transform: rotate(<%= -25 + (@tag_model.id.digits.first * 5) %>deg);" />
-        <% end %>
-        <% if @tag_model && @tag_model.pretty_name.present? %>
-          <%= @tag_model.pretty_name %>
-        <% else %>
-          <%= @tag %>
-        <% end %>
+  <div class="crayons-layout">
+    <header class="tag-header crayons-card branded-4 relative" style="border-top-color: <%= HexComparer.new([@tag_model.bg_color_hex || "#0000000", @tag_model.text_color_hex || "#ffffff"]).brightness(0.88) %>">
+      <div class="tag-header__title flex justify-between items-center mb-2 s:mb-4">
```
ludwiczakpawel, line 9: Instead of having `mb-2 s:mb-4` here, I'd move it to the paragraph on line 28 as a `pt-2 s:pt-4` - therefore if there's no `short_summary`, it will be better aligned:

**With `mb...`** 

image

**With `pt...`**

image

ludwiczakpawel: You don't seem to be referencing `tag-header__title` anywhere in CSS or JS so I think we can drop it, right?

app/views/articles/tag_index.html.erb
```diff
@@ -4,31 +4,34 @@
...
+          </h1>
+          <% if @tag_model.badge_id && @tag_model.badge %>
+              <img src="<%= optimized_image_url(@tag_model.badge.badge_image_url, width: 64) %>" class="" style="transform: rotate(<%= -25 + (@tag_model.id.digits.first * 5) %>deg); width: 64px; height: 64px;" />
+          <% end %>
+        </div>
         <% if @tag_model %>
-            <span class="user-profile-follow-button-wrapper">
-              <button id="user-follow-butt" class="crayons-btn follow-action-button user-profile-follow-button" style="color:<%= @tag_model.text_color_hex %>;background-color:<%= @tag_model.bg_color_hex %>" data-info='{"id":<%= @tag_model.id %>,"className":"Tag"}'>&nbsp;</button>
-            </span>
+          <button id="user-follow-butt" class="crayons-btn crayons-btn--secondary follow-action-button" data-info='{"id":<%= @tag_model.id %>,"className":"Tag"}'>&nbsp;</button>
```
ludwiczakpawel, line 24:
```suggestion
          <button id="user-follow-butt" class="crayons-btn follow-action-button" data-info='{"id":<%= @tag_model.id %>,"className":"Tag"}'>Follow</button>
```

This will make "the jump" (between _loading page_ state and _once its loaded_ state) more smooth.

app/views/articles/tag_index.html.erb
```diff
@@ -4,31 +4,34 @@
...
         <% end %>
-      </h1>
-    </div>
+      </div>
+      <% if @tag_model && @tag_model.short_summary.present? %>
+        <p>
```
ludwiczakpawel, line 28: Just a random idea... As an addition to what I suggested with padding...

What if we make this paragraph width 75%? Since our page is pretty wide, for long texts the lines will be quite long.. If we shorten this, it's just more eye-friendly and readable.

```suggestion
        <div class="max-w-100 m:max-w-75 pt-2 s:pt-4">
```

^ just try it out...

app/views/articles/tag_index.html.erb
```diff
@@ -4,31 +4,34 @@
...
+          <%= sanitize @tag_model.short_summary %>
```
ludwiczakpawel, line 29: It seems like the value of `@tag_model.short_summary ` can be any HTML, not just text, so instead of wrapping it with `<p>` I'd wrap it with `<div>`. 

**BUT...** I don't think it makes sense to let anyone provide any HTML here - I think it should be just a text with nothing else, therefore I'd suggest using this method:

```suggestion
          <%= strip_tags(@tag_model.short_summary) %>
```

^ it will strip all HTML tags from value of `@tag_model.short_summary`

app/views/articles/tags/_sidebar.html.erb
ludwiczakpawel: Can we just drop this widget enitrely?

ludwiczakpawel (changes requested):
Minor changes, otherwise looking good.

ludwiczakpawel:
>  I went a little bit off of the original design spec when it came to the tag brand image; originally, I had it pinned to the bottom right with absolute positioning, but having an absolutely positioned image was tricky with the body paragraph that could either not exist at all, or be really long.

Oh yea, I can see that being an issue... I think your solution makes sense. Also, a random idea for consideration:

image

lisasy:
@ludwiczakpawel I updated this PR with layout changes:

### Badge image exists
Screen Shot 2020-10-28 at 1 36 29 PM

### Badge image does not exist
Screen Shot 2020-10-28 at 1 36 47 PM

### No summary
Screen Shot 2020-10-28 at 1 45 35 PM

app/views/articles/tag_index.html.erb
```diff
@@ -4,31 +4,34 @@
...
-        <% end %>
+  <div class="crayons-layout">
+    <header class="tag-header crayons-card branded-4 relative" style="border-top-color: <%= HexComparer.new([@tag_model.bg_color_hex || "#0000000", @tag_model.text_color_hex || "#ffffff"]).brightness(0.88) %>">
+      <div class="flex justify-between items-center">
+        <div class="inline-flex items-center">
+          <h1 class="crayons-title">
+            <% if @tag_model && @tag_model.pretty_name.present? %>
+              <%= @tag_model.pretty_name %>
+            <% else %>
+              <span class="crayons-tag__prefix">#</span>
```
ludwiczakpawel, line 15: Since this is not really part of an actual `crayons-tag` component here, I'd just go with this instead:

```suggestion
              <span class="opacity-50">#</span>
```

app/views/articles/tag_index.html.erb
```diff
@@ -4,31 +4,34 @@
...
-            </span>
+          <button id="user-follow-butt" class="crayons-btn follow-action-button" data-info='{"id":<%= @tag_model.id %>,"className":"Tag"}'>Follow</button>
         <% end %>
-      </h1>
-    </div>
+      </div>
+      <% if @tag_model && @tag_model.short_summary.present? %>
+        <div class="max-w-100 m:max-w-75 pt-2 s:pt-4">
+          <%= strip_tags(@tag_model.short_summary) %>
+        </div>
```
ludwiczakpawel, line 30: Oh, if you decided to go with `strip_tags`, then I think we can keep the `p` :) Sorry for the confusion!

```suggestion
        <p class="max-w-100 m:max-w-75 pt-2 s:pt-4">
          <%= strip_tags(@tag_model.short_summary) %>
        </p>
```

ludwiczakpawel (approved):
Nice! I've pushed two tiny fixes, and it's good to go :)

Zhao-Andy (approved):
LGTM! Liking the new look!

Zhao-Andy:
Looks like a test is failing, I'll push a fix.

lisasy:
@Zhao-Andy I'm going to work on resolving some of the tests to get Travis CI to pass.

PR #11132, opened 2020-10-28
Tooltip fix.

ludwiczakpawel:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR fixes some edge cases for recently introduced `crayons-tooltip`.

Nah, just approve it :D 

## Added tests?

- [ ] Yes
- [x] No, not needed.

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/assets/stylesheets/components/tooltips.scss
+++ b/app/assets/stylesheets/components/tooltips.scss
@@ -2,7 +2,6 @@
   position: relative;
 
   &:after {
-    all: unset;
     content: attr(data-tooltip);
     position: absolute;
     bottom: 100%;
@@ -21,11 +20,9 @@
     opacity: 0;
     transition: all var(--transition-props);
     z-index: 1000;
-    text-align: left;
   }
 
   &:hover {
-    z-index: 1000;
     &:after {
       opacity: 1;
       transform: translateY(0);
--- a/app/assets/stylesheets/components/tooltips.scss
+++ b/app/assets/stylesheets/components/tooltips.scss
@@ -22,13 +22,15 @@
     transition: all var(--transition-props);
     z-index: 1000;
     text-align: left;
+    transition-delay: 250ms;
   }
 
   &:hover {
     z-index: 1000;
     &:after {
       opacity: 1;
-      transform: translateY(0);
+      transform: translateY(calc(-1 * var(--su-1)));
+      transition-delay: 0s;
     }
   }
 }
--- a/app/assets/stylesheets/components/tooltips.scss
+++ b/app/assets/stylesheets/components/tooltips.scss
@@ -2,6 +2,7 @@
   position: relative;
 
   &:after {
+    all: unset;
     content: attr(data-tooltip);
     position: absolute;
     bottom: 100%;
@@ -20,9 +21,11 @@
     opacity: 0;
     transition: all var(--transition-props);
     z-index: 1000;
+    text-align: left;
   }
 
   &:hover {
+    z-index: 1000;
     &:after {
       opacity: 1;
       transform: translateY(0);


Link2Twenty:
Would now be a good time to add a z-index property?

https://github.com/forem/forem/blob/master/app/assets/stylesheets/config/_variables.scss

```scss
////////////////////////////////////////////////////////////////////////////////////
// z index ////////////////////////////////////////////////////////////////////////
///////////////////////////////////////////////////////////////////////////////////

--zi-app-bar: 100;
--zi-drawer: 300;
--zi-modal: 500;
--zi-snackbar: 700;
--zi-tooltip: 1000;
```

app/assets/stylesheets/components/tooltips.scss
```diff
@@ -20,12 +21,16 @@
     opacity: 0;
     transition: all var(--transition-props);
     z-index: 1000;
+    text-align: left;
+    transition-delay: 250ms;
   }
 
   &:hover {
+    z-index: 1000;
```
vaidehijoshi, line 29: why `1000` here? does it need to be higher than some _other_ z-index here? if so, can we include a note (kind of like [this one](https://github.com/forem/forem/blob/dd6e2246ab9bc092d8cf95684842f4f9055e68a1/app/assets/stylesheets/scaffolds.scss#L74)? whenever i see a number like this one it feels a bit "magic number-y" to me, so i'm curious to know why the number has to be what it is (although if we use a variable here, i think it would become more obvious)

benhalpern:
> Would now be a good time to add a z-index property?

Good call!

lisasy:
@ludwiczakpawel what would be a good way for me to "repro" this component here? I miss our "page/crayons" page 😭

ludwiczakpawel:
@lisasy it should be there actually :) 

image

app/assets/stylesheets/components/tooltips.scss
ludwiczakpawel: It's semi-random number :D 

z-indexes are hard. And they are especially hard in codebase like ours where we try to maintain "properly written" code like Crayons as well as some legacy code (pre-Crayons era). Reason is: before Crayons all z-indexes where pretty random... When someone was working on a piece of UI and found it being "behind" another element, the easiest solution was to bump the z-index until it's no longer behind :D. That's how we ended up with many different z-indexes trying to overwrite each other :).

We had this convo about refactoring z-indexes for Crayons but at the time it felt like too much work so we kinda continue this nasty practice of using semi-random values.

But yea, I guess we're in a place now where we should take a step back and refactor finally all of these z-indexes.. Let's merge this as is, and I'll kick-off work cleaning up all our z-indexes.

cc @benhalpern @Link2Twenty

Link2Twenty: Is it intentional that you're raising the tooltip instigator rather than the tooltip?
___
For the labelled comment Vaidehi's example could be changed to this, down the road, which will be even clearer hopefully

```css
z-index: calc(--zi-app-bar + 10); // Must be higher than the z-index of the sticky-nav.
```

I think 1000 is a good number for tooltips because generally you want a large stepping range between z-indexes, I usually go for 100, meaning you can slot things in between them without needing an entire refactor.

ludwiczakpawel: @Link2Twenty yes, it's intentional and the reason was that one of our devs used this tooltip component on a button element which was placed pretty unfortunate, next to another element being elevated with z-index... TL;DR this felt to be a bit more bulletproof, since having lower z-index on :hover element is very unlikely. 

Btw. https://github.com/forem/forem/pull/11157

Zhao-Andy (approved):
> Just approve it™️

PR #11136, opened 2020-10-28
Encode domain name in pusher channels

maestromac:
- [x] Feature
- [x] (development) Bug Fix

## Description
https://github.com/forem/forem/pull/11029 fixed a production issue but broke development.`localhost:3000` is not a valid pusher channel name because Pusher [channel specification](https://github.com/pusher/pusher-http-ruby/blob/master/lib/pusher/channel.rb#L11) doesn't allow colon. I was receiving the following error in development and test:

```
Pusher::Error:
       Illegal channel name 'private-message-notifications--localhost:3000-25'
```

This feature encodes the domain name, which would fix this issue but also provide the added benefit of obfuscating channel names.

This is an additional update to https://github.com/forem/forem/pull/11029

Prior to this fix you probably can't _send a message and see it appear immediately_ in development.

1. Bootup the app locally and make sure you are logged in.
1. Open up the Rails console. Find the first user and take note of their username ie `User.first.username`.
1. Change their inbox type to open `User.first.update(inbox_type: "open")`
1. Navigate to their profile page (ie `localhost:3000/{User.first.username}`) and click `chat` on the right side of their profile banner.
1. You should be able to see and send messages without issue.

## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

nope

### DIFF
--- a/app/javascript/utilities/connect/getUnopenedChannels.jsx
+++ b/app/javascript/utilities/connect/getUnopenedChannels.jsx
@@ -23,9 +23,9 @@ class UnopenedChannelNotice extends Component {
 
   componentDidMount() {
     const { pusherKey } = this.props;
-    const appName = document.body.dataset.appName;
+    const appDomain = document.body.dataset.appDomain;
     setupPusher(pusherKey, {
-      channelId: `private-message-notifications--${appName}-${window.currentUser.id}`,
+      channelId: `private-message-notifications--${appDomain}-${window.currentUser.id}`,
       messageCreated: this.receiveNewMessage,
       messageDeleted: this.removeMessage,
       messageEdited: this.updateMessage,
--- a/app/javascript/utilities/connect/getUnopenedChannels.jsx
+++ b/app/javascript/utilities/connect/getUnopenedChannels.jsx
@@ -23,7 +23,7 @@ class UnopenedChannelNotice extends Component {
 
   componentDidMount() {
     const { pusherKey } = this.props;
-    const appDomain = document.body.dataset.appDomain;
+    const { appDomain } = document.body.dataset;
     setupPusher(pusherKey, {
       channelId: `private-message-notifications--${appDomain}-${window.currentUser.id}`,
       messageCreated: this.receiveNewMessage,
--- a/app/models/chat_channel.rb
+++ b/app/models/chat_channel.rb
@@ -114,9 +114,9 @@ class ChatChannel < ApplicationRecord
   def pusher_channels
     # TODO: use something more unique here (uuid?) rather than just id.
     if invite_only?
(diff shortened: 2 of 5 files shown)


app/models/chat_channel.rb
```diff
@@ -166,7 +166,11 @@ def pending_users_select_fields
   end
 
   def self.pm_notifications_channel(user_id)
-    "private-message-notifications--#{ApplicationConfig['APP_DOMAIN']}-#{user_id}"
+    "private-message-notifications--#{urlsafe_encoded_app_domain}-#{user_id}"
+  end
+
+  def self.urlsafe_encoded_app_domain
+    Base64.urlsafe_encode64(ApplicationConfig["APP_DOMAIN"])
```
maestromac, line 173: `urlsafe_encode64` is the best option because it uses ‘-’ instead of ‘+’ and ‘_’ instead of ‘/’. Pusher doesn't allow `+`, which would've been included if we use the simple `encode64.

ref https://apidock.com/ruby/Base64/urlsafe_encode64

app/javascript/utilities/connect/getUnopenedChannels.jsx
```diff
@@ -23,9 +23,9 @@ class UnopenedChannelNotice extends Component {
 
   componentDidMount() {
     const { pusherKey } = this.props;
-    const appName = document.body.dataset.appName;
+    const appDomain = document.body.dataset.appDomain;
```
nickytonline, line 26: refactor (non-blocking): Prefer destructuring.

```suggestion
    const { appDomain } = document.body.dataset;
```

nickytonline (approved):
LGTM. Left a suggestion but it's non-blocking. Also, I assume that there's no need to change tests for this change as it's internal component state changes?

maestromac:
@nickytonline yup no need to change JS test

app/javascript/utilities/connect/getUnopenedChannels.jsx
```diff
@@ -23,9 +23,9 @@ class UnopenedChannelNotice extends Component {
 
   componentDidMount() {
     const { pusherKey } = this.props;
-    const appName = document.body.dataset.appName;
+    const { appDomain } = document.body.dataset;
```
vaidehijoshi, line 26: oh man, I should have changed this in the original PR -- sorry about that 😞  and thank you for fixing this!

vaidehijoshi (approved):

app/javascript/utilities/connect/getUnopenedChannels.jsx
maestromac: that alright!, it was doing no harm.

app/models/chat_channel.rb
rhymes: <3

PR #11141, opened 2020-10-28
Improve error messages and validation for admin pages

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We should be rending error messages here and this form needs some client-side validation.

Closes https://github.com/forem/InternalProjectPlanning/issues/232

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/controllers/admin/pages_controller.rb
+++ b/app/controllers/admin/pages_controller.rb
@@ -23,14 +23,25 @@ module Admin
 
     def update
       @page = Page.find(params[:id])
-      @page.update!(page_params)
-      redirect_to "/admin/pages"
+      @page.assign_attributes(page_params)
+      if @page.valid?
+        @page.update!(page_params)
+        redirect_to admin_pages_path
+      else
+        flash.now[:error] = @page.errors.full_messages.join(", ")
+        render :edit
+      end
     end
 
     def create
       @page = Page.new(page_params)
-      @page.save!
-      redirect_to "/admin/pages"
+      if @page.valid?
+        @page.save!
+        redirect_to admin_pages_path
+      else
+        flash.now[:error] = @page.errors.full_messages.join(", ")
+        render :new
+      end
     end
 
     def destroy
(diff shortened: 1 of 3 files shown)


spec/system/admin/admin_manages_pages_spec.rb
```diff
@@ -54,6 +54,18 @@
       end
     end
 
+    it "has client-side validation" do
```
jacobherrington, line 57: This is not the sturdiest of all specs, but testing for HTML validation is a bit tricky, and if this fails we know something changed, which will help find any bugs.

app/controllers/admin/pages_controller.rb
```diff
@@ -23,14 +23,25 @@ def edit
...
     def update
       @page = Page.find(params[:id])
-      @page.update!(page_params)
-      redirect_to "/admin/pages"
+      @page.assign_attributes(page_params)
+      if @page.valid?
+        @page.update!(page_params)
+        redirect_to admin_pages_path
+      else
+        flash.now[:error] = @page.errors.full_messages.join(", ")
```
Zhao-Andy, line 31: I like using `model.errors_as_sentence` :)

https://github.com/forem/forem/blob/531f26bad69e6cb03468d1f84d2e4a64c3dea757/app/models/application_record.rb#L56-L58

Zhao-Andy: Also curious, why `flash.now` over regular `flash`? So that it doesn't have a chance to persist?

Zhao-Andy (approved):
LGTM! Couple of comments, but non blocking.

juliannatetreault (approved):
Nice improvements! 👍

app/controllers/admin/pages_controller.rb
jacobherrington: Cool, I'll do that! Thanks!

`flash` uses the flash message on the next request (so would be ideal before `redirect_to`), whereas `flash.now` will use the flash message on the current request (so generally, called before `render`).

PR #11142, opened 2020-10-28
Mark Posts as Published in Admin Articles

juliannatetreault:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
**Issue**:
> We seem to be unable to mark a post as approved via /admin. Instead, we're forced to use /resource_admin.

**Solution**:
This PR gives Admins the ability to mark posts as `published` from `/admin/articles`. To make this work as expected, I refactored the `#update` method in the `Admin::ArticlesController`. Additionally, I added additional tests to the `articles_spec.rb` and refactored the existing tests a little.

Closes https://github.com/forem/InternalProjectPlanning/issues/231

**To QA these changes, first ensure that you have an `admin` role, and then please do the following**:

- Navigate to `/admin/articles`:
Screen Shot 2020-10-28 at 11 05 10 AM

- Mark an article of your choice as `approved` and ensure that it updates properly:
Screen Shot 2020-10-28 at 11 05 51 AM

- Update other fields on the article to ensure that everything works as expected:
Screen Shot 2020-10-28 at 11 06 49 AM

- Go to the article's show page to double-check that everything looks as expected:
Screen Shot 2020-10-28 at 11 09 23 AM

- Try to break things! 🔨 

## Added tests?

- [x] Yes and refactored existing specs in `articles_spec.rb`
      have not been included_

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

N/A

Seal of Approval

### DIFF
--- a/app/controllers/admin/articles_controller.rb
+++ b/app/controllers/admin/articles_controller.rb
@@ -44,7 +44,7 @@ module Admin
       article.email_digest_eligible = article_params[:email_digest_eligible].to_s == "true"
       article.boosted_additional_articles = article_params[:boosted_additional_articles].to_s == "true"
       article.boosted_dev_digest_email = article_params[:boosted_dev_digest_email].to_s == "true"
-      article.user_id = article_params[:user_id].to_i
+      article.user_id = article_params[:user_id].to_i if article_params[:user_id].present?
       article.update(article_params)
       article.co_author_ids = article_params[:co_author_ids]&.split(",")&.map(&:strip)
       if article.save
--- a/app/controllers/admin/articles_controller.rb
+++ b/app/controllers/admin/articles_controller.rb
@@ -45,7 +45,8 @@ module Admin
       article.boosted_additional_articles = article_params[:boosted_additional_articles].to_s == "true"
       article.boosted_dev_digest_email = article_params[:boosted_dev_digest_email].to_s == "true"
       article.user_id = article_params[:user_id].to_i
-      article.co_author_ids = article_params[:co_author_ids].split(",").map(&:strip)
+      article.update(article_params)
+      article.co_author_ids = article_params[:co_author_ids]&.split(",")&.map(&:strip)
       if article.save
         flash[:success] = "Article saved!"
       else
--- a/spec/requests/admin/articles_spec.rb
+++ b/spec/requests/admin/articles_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe "/admin/articles", type: :request do
 
   context "when updating an Article via /admin/articles" do
     let(:super_admin) { create(:user, :super_admin) }
-    let!(:article) { create(:article) }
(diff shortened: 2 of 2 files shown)


fdoxyz (commented):
Tested this locally and works well thanks @juliannatetreault! I have a question regarding the tests though:

In the specs I noticed we're using `article.update_columns(approved: true)` and if we introduce future changes to the HTML/routes/controller we won't be able to catch regressions because of this. What I mean is, the update of columns is performed directly via ActiveRecord (which won't go through the stack). When admins update something they will perform a PUT/PATCH request that will actually go through the controllers/models/etc.

In this case, maybe it would be good to replace the `.update_columns(...)` calls for something along the lines of:

```ruby
put admin_article_path(podcast), params: { article: { id: article.id, approved: true } }
```

juliannatetreault:
> Tested this locally and works well thanks @juliannatetreault! I have a question regarding the tests though:
> 
> In the specs I noticed we're using `article.update_columns(approved: true)` and if we introduce future changes to the HTML/routes/controller we won't be able to catch regressions because of this. What I mean is, the update of columns is performed directly via ActiveRecord (which won't go through the stack). When admins update something they will perform a PUT/PATCH request that will actually go through the controllers/models/etc.
> 
> In this case, maybe it would be good to replace the `.update_columns(...)` calls for something along the lines of:
> 
> ```ruby
> put admin_article_path(article), params: { article: { approved: true } }
> ```
> 
> Let me know if that makes sense :)

That absolutely makes sense and I'm happy to make those changes! I'll work on doing that now @fdoxyz . 😊

app/controllers/admin/articles_controller.rb
```diff
@@ -44,8 +44,9 @@ def update
       article.email_digest_eligible = article_params[:email_digest_eligible].to_s == "true"
       article.boosted_additional_articles = article_params[:boosted_additional_articles].to_s == "true"
       article.boosted_dev_digest_email = article_params[:boosted_dev_digest_email].to_s == "true"
-      article.user_id = article_params[:user_id].to_i
-      article.co_author_ids = article_params[:co_author_ids].split(",").map(&:strip)
+      article.user_id = article_params[:user_id].to_i if article_params[:user_id].present?
```
juliannatetreault, line 47: I added this guard clause here because while pairing with @atsmith813, he noticed that without the guard clause, this line was overwriting the user and `user_id`. 🔎

fdoxyz (approved):
Gave it another spin and works as expected 😃 👍

juliannatetreault:
> Gave it another spin and works as expected 😃 👍

🎉 Thanks for the feedback and for the review @fdoxyz! 😄

benhalpern (approved):
Nice stuff. I'm going to merge it now.

PR #11149, opened 2020-10-29
Add feature flag for profile admin section

citizen428:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds a feature flag `:profile_admin` which removes the menu item from the admin index and sidebar and also disables the routes. 

Anyone should feel free to merge this, it has tests for all scenarios.

https://github.com/forem/InternalProjectPlanning/issues/147

### Admin index:

0. Ensure the feature flag is disabled.
1. Go to `/admin`
2. Ensure that you don't see the profile admin link (it would be right after "Config")
    Screen Shot 2020-10-29 at 10 15 24
3. Enable the feature flag in the Rails console:
    ```ruby
    Flipper.enable(:profile_admin)
    ```
4. Reload `/admin`, you should now see a "Config: Profile Setup" option:
    Admin_Portals

### Admin sidebar menu:

0. Ensure the feature flag is disabled.
1. Go an admin page, e.g. `/admin/articles`
2. Ensure that you don't see the profile admin link (it would be right under "Config")
    Screen Shot 2020-10-30 at 11 31 52
3. Enable the feature flag in the Rails console:
    ```ruby
    Flipper.enable(:profile_admin)
    ```
4. Reload the page, you should now see a "Config: Profile Setup" option:
    Articles

### Routing constraint

0. Ensure the feature flag is disabled.
1. Go to `/admin/profile_fields`
2. You should be redirected to our root route.
3. Enable the feature flag in the Rails console:
    ```ruby
    Flipper.enable(:profile_admin)
    ```
4. Going to `/admin/profile_fields` should now render the profile admin page.

## Added tests?

- [X] Yes
      have not been included_

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [X] No documentation needed

### DIFF
--- a/app/views/admin/admin_portals/index.html.erb
+++ b/app/views/admin/admin_portals/index.html.erb
@@ -1,5 +1,12 @@
   <div class="grid gap-2 m:gap-4 l:gap-6 m:grid-cols-2 l:grid-cols-3 px-2 m:px-0">
-    <% Admin::ApplicationController::MENU_ITEMS.each do |menu_item| %>
+    <%# Temporary code while we use a feature flag for the profile admin. %>
+    <% menu_items = Admin::ApplicationController::MENU_ITEMS.dup %>
+    <% unless Flipper.enabled?(:profile_admin) %>
+      <% index = menu_items.find_index { |item| item[:name] == "config: profile setup" } %>
+      <% menu_items.delete_at(index) %>
+    <% end %>
+
+    <% menu_items.each do |menu_item| %>
       <div class="tag-card crayons-card branded-4 p-4 m:p-6 m:pt-4 flex flex-col relative">
         <h3 class="crayons-tag crayons-tag--l mb-2">
           <a href="/admin/<%= menu_item[:controller] %>" class="crayons-link">
--- a/app/views/admin/admin_portals/index.html.erb
+++ b/app/views/admin/admin_portals/index.html.erb
@@ -1,12 +1,5 @@
   <div class="grid gap-2 m:gap-4 l:gap-6 m:grid-cols-2 l:grid-cols-3 px-2 m:px-0">
-    <%# Temporary code while we use a feature flag for the profile admin. %>
-    <% menu_items = Admin::ApplicationController::MENU_ITEMS.dup %>
-    <% unless Flipper.enabled?(:profile_admin) %>
-      <% index = menu_items.find_index { |item| item[:name] == "config: profile setup" } %>
-      <% menu_items.delete_at(index) %>
-    <% end %>
-
-    <% menu_items.each do |menu_item| %>
+    <% admin_menu_items.each do |menu_item| %>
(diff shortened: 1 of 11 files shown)


app/views/admin/admin_portals/index.html.erb
```diff
@@ -1,5 +1,12 @@
   <div class="grid gap-2 m:gap-4 l:gap-6 m:grid-cols-2 l:grid-cols-3 px-2 m:px-0">
-    <% Admin::ApplicationController::MENU_ITEMS.each do |menu_item| %>
+    <%# Temporary code while we use a feature flag for the profile admin. %>
```
citizen428, line 2: This is not pretty. BUT: dealing with frozen constants is a bit painful and I didn't feel like adding a whole new helper for something that's hopefully quite temporary. If people prefer that I can extract it though.

jacobherrington (approved):
Works for me :+1:

vaidehijoshi (changes requested):
Yay, feature flags!! 👏 

Can we remove it from the sidebar when the feature is disabled as well? I wonder if we could just disable the entire `/admin/profile_fields` route and hide _that_ behind a feature flag (in my opinion, that would be the ideal solution here so that we don't need to worry about every place that we link to it), but I think we should definitely remove it [from the sidebar](https://github.com/forem/forem/blob/5457f0c0b702af71cc864a0fbdd065238760b7f5/app/controllers/admin/application_controller.rb#L29) since that will show up on every admin view :) 

Screen_Shot_2020-10-29_at_9_27_58_AM

citizen428:
@vaidehijoshi 

> Can we remove it from the sidebar when the feature is disabled as well? 

Definitely, my oversight, thanks for catching that.

>  wonder if we could just disable the entire `/admin/profile_fields` route and hide that behind a feature flag (in my opinion, that would be the ideal solution here so that we don't need to worry about every place that we link to it)

Sure, we can disable the route, but we'll still have to remove the links too, otherwise they lead to a 404.

citizen428:
@vaidehijoshi I now did the following:

1. Add a new `AdminHelper` module with an `admin_menu_items` method that is used for both the sidebar partial and admin index view with keeps all the logic in one place.
2. Added a spec for the sidebar too.
3. Disabled all admin profile routes via a constraint and added a routing spec for that. The spec only tests for the `GET` route, but IMHO that's good enough since it's all done via a `resources` block, so if one is disabled, all are.

Re-requesting review now.

vaidehijoshi (commented):
This is great, @citizen428! Thanks for taking my feedback into account :)

I think there are just a few specs (in `profile_field_groups_spec.rb`) that need the feature flag to mocked to be on, and then this is should be good to go !

citizen428:
Since it took less than a minute I fixed it now.

citizen428:
Thanks @vaidehijoshi 😃 It was quick enough so I fixed it right away.

vaidehijoshi (approved):
🎉🎉🎉

PR #11154, opened 2020-10-29
Drop custom color pickers

ludwiczakpawel:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

We used to use custom JS color pickers in two places: tag editing for mods as well as tag editing for admins. Actually only in the view for mods because the color picker for admin has already been replaced with browser native solution.

This PR drops custom color picker entirely and replaces it with browser native solution.

1. Be a mod.
2. Go to tag edit view.
3. Change tag colors.

## Added tests?

- [ ] Yes
- [x] No, not needed

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed


citizen428 (approved):

rhymes (approved):
I can't see the difference betwen before and after so I guess it's working :D

PR #11164, opened 2020-10-29
Refactor:Remove Production Conditional for Page View Updates

mstruve:
- [x] Refactor

## Description
Just like #11144 this PR removes a couple more `Rails.env.production?` checks. There is really no reason why we can't run these locally so why not. As stated in the other PR, the more we can make development like production the better. 

## Added tests?
- [x] No, be there is no new functionality

alt_text

### DIFF
--- a/app/controllers/page_views_controller.rb
+++ b/app/controllers/page_views_controller.rb
@@ -32,7 +32,7 @@ class PageViewsController < ApplicationMetalController
   private
 
   def update_article_page_views
-    return if Rails.env.production? && rand(15) != 1 # We don't need to update the article page views every time.
+    return if skip_page_view_update?
 
     @article = Article.find(page_view_params[:article_id])
     new_page_views_count = @article.page_views.sum(:counts_for_number_of_views)
@@ -46,7 +46,7 @@ class PageViewsController < ApplicationMetalController
   end
 
   def update_organic_page_views
-    return if Rails.env.production? && rand(100) != 1 # We need to do this operation only once in a while.
+    return if skip_organic_page_view_update?
 
     page_views_from_google_com = @article.page_views.where(referrer: "https://www.google.com/")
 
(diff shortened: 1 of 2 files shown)


app/controllers/page_views_controller.rb
```diff
@@ -64,4 +64,14 @@ def update_organic_page_views
       .where("created_at > ?", 1.month.ago).sum(:counts_for_number_of_views)
     @article.update_column(:organic_page_views_past_month_count, organic_count_past_month_count)
   end
+
+  def skip_page_view_update?
+    # We don't need to update the article page views every time.
+    rand(15) != 1
```
mstruve, line 70: I think we can do better here than just random updates but I don't have a solution in mind at the moment. Just wanted to put it out there though this is on my radar as something we could improve.

spec/requests/page_views_spec.rb
```diff
@@ -4,6 +4,13 @@
   let(:user) { create(:user, :trusted) }
   let(:article) { create(:article) }
 
+  before do
+    # rubocop:disable RSpec/AnyInstance
+    allow_any_instance_of(PageViewsController).to receive(:skip_page_view_update?).and_return(false)
```
mstruve, line 9: If there is a better way to do this without using any_instance I am all ears but this is the best way I know how and in this case I dont see a lot of harm in using any_instance

atsmith813: I oddly feel like this works well for this given scenario 🙈

atsmith813 (approved):

spec/requests/page_views_spec.rb
maestromac: This might be overkill but you could extract this `rand()` usage into a module and call that. Then you can just stub the module without using `allow_any_instance_of`.

maestromac (approved):

PR #11084, opened 2020-10-26
Notifications redesign

ludwiczakpawel:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR applies a bit of styling to our Notifications view.

It's a first take on this and there are clearly things we could improve (especially reaction buttons and icons and their states) but probably later in future. 

## QA Instructions, Screenshots, Recordings

If you want to populate your notifications locally, go [here](https://github.com/forem/forem/pull/10828).

image

image

## Added tests?

- [ ] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

PR #11168, opened 2020-10-29
Schema Fix:Removed accidentally added changes

mstruve:
- [x] Bug Fix

## Description
In #11084 there were some things changed in the schema that should not have been. This fixes those. 

alt_text

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -485,6 +485,7 @@ ActiveRecord::Schema.define(version: 2020_10_19_012200) do
     t.datetime "updated_at", null: false
     t.bigint "user_id"
     t.datetime "verified_at"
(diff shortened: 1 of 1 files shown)


ludwiczakpawel (approved):

vaidehijoshi (approved):

vaidehijoshi:
I wonder if we can make it easier to see when this file has changed in version control. Local issues with databases sometimes do happen, but it would be ideal if we could catch those or guard against them accidentally being committed somehow 🤔

ludwiczakpawel:
@vaidehijoshi how about a github action specifically triggered only when my PRs contain schema.rb change? :)

PR #11169, opened 2020-10-29
[Front-End Q/A] Config Authentication view

lisasy:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

ToDos
- [x] Implement tooltip on invite-only mode
- [x] Fix padding/spacing with checkboxes, row expanded + collapsed
- [x] Fix jumping modal interaction
- [x] Update copy in interstitial dialog
- [x] When modal closes after disabling, collapsed the expanded row and change state.

## Related Tickets & Documents
Closes https://github.com/forem/InternalProjectPlanning/issues/246

## QA Instructions, Screenshots, Recordings

_Please replace this line with instructions on how to test your changes, as well
as any relevant images for UI changes._

## Added tests?

- [ ] Yes
- [ ] No, and this is why: _please replace this line with details on why tests
      have not been included_
- [ ] I need help with writing tests

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [ ] No documentation needed

## [optional] What gif best describes this PR or how it makes you feel?

alt_text

PR #11185, opened 2020-10-30
[Team Email Login] Refactor Authentication Provider Enable/Disable

msarit:

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## IMPORTANT!
Re: the discussion [here](https://dev-internal.slack.com/archives/C7GE84DEJ/p1604506132027200), Team Email Login will refactor the `SiteConfig.allow_both_email_and_password` functionality to ensure that Email Login is on all the time. Those changes are NOT a part of this PR, and will be made after this PR is merged (for clarity's sake).

## Description
This PR refactors the enabling and disabling of the Authentication Providers, according to this [design mockup](https://user-images.githubusercontent.com/816402/96671965-f5909180-1317-11eb-990b-9bf4cabfa5cd.jpg).
I have left pertinent notes throughout my code changes, explaining my thought process.
**Allow me to say I'm quite proud of this refactor 😄  I learned a lot, and looking forward to learning even more from the QA process 🙏** 

closes https://github.com/forem/InternalProjectPlanning/issues/167
- ~I'm going to wait on the merge of Lisa's [Email Auth UX Fixes PR](https://github.com/forem/forem/pull/11169) before finalizing work on this PR; I will need to incorporate several of her updates~

1. Navigate to `localhost:xxxx/admin/config` as a user with `:admin` (or `:super_admin`) and `:single_resource_admin` roles.
2. Go through the process of enabling and disabling auth providers. Here are the configurations I could think of:
- Can enable all or any combo of available Auth Providers, unless Invite-Only Mode is on.
- Can disable all available Auth Providers when Email Auth is on _(this is currently not possible; one Auth Provider must always be on)_
- Cannot disable an Auth Provider **if it is the sole enabled provider AND Email Auth is off**
- When Invite-Only Mode is enabled, all Auth Providers are disabled (and their buttons are disabled too)
3. Ensure that, based on Auth Provider setting changes, the expected auth options display on the login/create-account frontend pages

## Added tests?

- [X] Yes
- [ ] No, and this is why:

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [X] No documentation needed

No


app/controllers/admin/configs_controller.rb
```diff
@@ -223,6 +221,23 @@ def toggle_email_password_authentication
       end
     end
 
+    def update_enabled_auth_providers(value)
```
msarit, line 224: These private functions exist because the text-field `value` of `auth_providers` received from the front end is a string.

These functions convert this string to an array, checks the validity of each array item, then builds the final array with which `SiteConfig.authentication_providers` is updated.

app/controllers/admin/configs_controller.rb
```diff
@@ -223,6 +221,23 @@ def toggle_email_password_authentication
       end
     end
```
msarit, line 223: The following 3 private functions exist because the text-field `value` of `auth_providers` received from the front end is a string.

These functions convert this string to an array, checks the validity of each array item, then builds the final array with which `SiteConfig.authentication_providers` is updated.

app/views/admin/configs/show.html.erb
```diff
@@ -161,6 +161,59 @@
                   How can people create an account on your community?
                 </p>
 
+                <section>
+                  <div class="form-group">
```
msarit, line 165: In order to implement the design team's [mockup](https://user-images.githubusercontent.com/816402/96671965-f5909180-1317-11eb-990b-9bf4cabfa5cd.jpg) for Enabling/Disabling Authentication Providers, I employed the use of a hidden text field.

Whenever the auth_providers' `Enable`, `Undo` or `Disable` buttons are clicked, handlers in `config_controller.js` fire off such that this hidden text field populates with a string of auth_providers that should remain enabled.

This string is passed as `params[:auth_providers_to_be_enabled]` to the `configs_controller.rb`, where it is processed appropriately, before updating the `SiteConfig.authentication_providers` value.

It is possible for ALL `auth_providers` to be disabled (this was NOT the case when we had them as a dropdown list). However, I have built 2 important guards in:
- If there is just one `auth_provider` enabled, it cannot be disabled unless Email Authentication is on.
- In the case when more than 1 `auth_provider` is currently enabled, if Email Auth is off and the user disables all `auth_providers` at once, this disabling IS NOT processed in the controller (because it would effectively lock everyone out of the forem).

Finally, when `invite_only_mode` is enabled, this action disables the

vaidehijoshi (commented):
Hi, @msarit! I'm so excited about this new design for authentication on SiteConfig! Great work taking on this refactor :) I haven't done an in-depth review of the _code_ just yet, and focused on just QAing these changes to start. There a couple things I noticed, please feel free to let me know what you think!

1. > Can disable all available Auth Providers when Email Auth is on (_this is currently not possible; one Auth Provider must always be on_)

I went to QA this, and noticed that it actually is possible for all auth providers to be turned off and for email auth to be the _only_ thing that is enabled. Is this intentional, or do we _not_ want this to happen yet? I just wasn't sure from your comment in the PR description, so I wanted to flag that here since I noticed it in QA:
Screen_Shot_2020-11-01_at_5_09_28_PM

2. > When Invite-Only Mode is enabled, all Auth Providers are disabled (and their buttons are disabled too)

I ended up in a state where I enabled all my auth providers, and then turned on invite-only mode. I noticed that I cannot actually _use_ any of my auth providers (I'm assuming it has been "disabled" on the backend), but the frontend/UI doesn't reflect that. I think we should indicate to the user somehow that they _cannot_ use any of those providers. Right now, I can click on the drop down and keep trying to select a provider, and update the SiteConfig, but my changes don't actually take effect -- I also don't get an error message, which will be a confusing experience for the user, I think! We probably should just visually "disable" all the auth fields once invite-only mode has been switched on.
Screen_Shot_2020-11-01_at_5_15_06_PM

3. Something seems to be up with the `cancel` button text in the admin modal -- I think it should read `Cancel` (it does on `master` currently), but I ended up in a state where the button text is `undefined`.
Screen Shot 2020-11-01 at 5 08 06 PM

4. A smaller observation: is it possible to change the animation time of how this tooltip renders? Right now, because we are adding a tooltip to _each_ of these disabled buttons, the tooltip text overlays and it is possible to end up with a 2 tooltips on top of one another 😃 Another option would be to add the tooltip to the larger containing element, but I think the "right" way of doing this is to adjust the tooltip animation. 
tooltip_lag

vaidehijoshi (changes requested):
requesting changes for the admin modal button bug noted in my previous comment 😅

msarit:
@vaidehijoshi Thanks so much for jumping into this PR 😄 

1. I've fixed the bug related to the Modal buttons and updated the docs accordingly
2. Regarding the tooltip behavior, I agree 😅  but wasn't sure what to do about it. I implemented your suggestion about placing the tooltip on the buttons' parent element, and here are the video demos of that (desktop and mobile):

demo-tooltip-on-parent-element

demo-tooltip-on-parent-element-mobile

joshpuetz:
There's been an important change to the behavior of this section that I think may need some input from @benhalpern: Previously the options for account registration and user login were separate per third party authentication provider, and now they are combined. 

What this means in practice is that prior to this PR one could have invite-only site where users could log in with Facebook, GitHub, etc. Now they can't: invite only means only the Email login option is enabled.

How do we feel about this? I don't have strong opinions one way or another.

app/javascript/admin/controllers/config_controller.js
```diff
@@ -75,28 +95,107 @@ export default class ConfigController extends Controller {
...
+    this.listAuthToBeEnabled();
+    this.hideAuthProviderSettings();
+  }
+
+  authProviderModalTitle(provider) {
+    return `Disable ${provider} registration`;
+  }
+
+  authProviderModalBody(provider) {
+    return `<p>If you disable ${provider} as a registration option, people cannot create an account with ${provider}.</p><br /><p>However, people who have already created an account with ${provider} can continue to login.</p><br /><p><strong>You must update site config to save this action!</strong></p>`;
```
joshpuetz, line 145: This isn't true: disabling a third party auth provider **also** turns off the ability to log in to a forem with that provider.

app/views/admin/configs/show.html.erb
```diff
@@ -256,136 +254,60 @@
...
+                          </button>
                         </div>
-                      </fieldset>
-                    </div>
-                  </div>
+                        <div id="<%= provider.provider_name %>-auth-settings" class="config-authentication__item--body hidden">
+                          <fieldset class="config-authentication-form">
+                            <%= render partial: "auth_provider_settings",
+                                       locals: {
+                                         f: f,
```
joshpuetz, line 302: f?!?!

joshpuetz (changes requested):
Tremendous job all around, this is **so much nicer** than the current interface!

I have a blocking questions about the wording on one of the modals that isn't matching the behavior I'm seeing.

app/javascript/admin/controllers/config_controller.js
msarit: This is true! Hmmm ... I concur with both remedial suggestions but I don't think I can take the decision without @ben and/or @lisasy

app/views/admin/configs/show.html.erb
msarit: @joshpuetz You mean rename `auth_provider_settings` file?

joshpuetz: No, I mean the local variable `f`: what is it?

msarit: Oh! 😆  It's the form! 🤣

joshpuetz: Ah, true, f is used for 'form' a lot in Rails. I still feel like naming it could be helpful, curious what others think. Definitely not something that would block my approval!

app/javascript/admin/controllers/config_controller.js
```diff
@@ -78,30 +98,156 @@ export default class ConfigController extends Controller {
...
+
+  showHideAuthEnabledIndicator(indicatorDiv, action) {
+    if (action === 'show') {
+      indicatorDiv.classList.add('flex', 'items-center');
+      indicatorDiv.classList.remove('hidden');
+    } else if (action === 'hide') {
+      indicatorDiv.classList.remove('flex', 'items-center');
+      indicatorDiv.classList.add('hidden');
+    }
+  }
```
vaidehijoshi, line 250: It would be ideal if we didn't have to pass in `'show'` and `'hide'` here to figure out what class should be rendered on the div. I haven't tried it locally yet, but I _think_ it should be possible to remove the conditional entirely and toggle the class, instead. As long as the div starts out with a default class, it _should_ work...I'm imagining something like this:
```javascript
toggleAuthEnabledIndicator(indicatorDiv) {
  indicatorDiv.classList.toggle('visible');
}
```

I think you can port more of this over to CSS and slim down JavaScript. This means that you can't use the generated utility classes, but IMO, that's okay here.

We could do this with CSS in `admin.scss` by adding a class to the `indicatorDiv` element(s):
```
.your-class-here {
  display: none; // The default visibility is hidden
}

.visible { // You can toggle this class and it should theoretically override the default styles
  display: flex;
  align-items: center;
}
```

Ideally, you wouldn't to check if we're trying to hide/show, you'd just toggle visibility on/off by calling the function at the right time. [Here's a short tutorial](https://www.digitalocean.com/community/tutorials/js-classlist) that might be helpful if you want to investigate it further!

Let me know what you think :)

vaidehijoshi:
Great work here, @msarit! I've tested this out locally and the logic works as expected 👍 

I left a few comments on naming/cleanup throughout -- let me know what you think!

I also noticed that the icons in the auth section are a bit misaligned:
Screen Shot 2020-11-05 at 11 14 37 AM

I think you can fix that by removing [this](https://github.com/forem/forem/blob/8d08191e9a455225438869e027f989cbd07e1e5c/app/assets/stylesheets/admin.scss#L128) class though!
Screen_Shot_2020-11-05_at_10_09_02_AM

app/assets/javascripts/utilities/showUserAlertModal.js
```diff
@@ -128,18 +138,18 @@ function buildModalDiv(title, text, confirm_text) {
...
  * @param {string} text The body text to be displayed
  * @param {string} confirm_text Text of the confirmation button
- * 
+ *
  * @returns {Element} DOM node of alert modal with formatted text
  */
 function getModal(title, text, confirm_text) {
   let wrapper = document.createElement('div');
-  wrapper.innerHTML= getModalHtml(title, text, confirm_text);
+  wrapper.innerHTML = getModalHtml(title, text, confirm_text);
```
joshpuetz, line 153: Oh nice, what linter are you using for Javascript?

app/helpers/authentication_helper.rb
```diff
@@ -33,23 +37,38 @@ def waiting_on_first_user?
...
     SiteConfig.invite_only_mode || authentication_enabled_providers.none? ? "disabled" : ""
   end
 
-  def disable_email_auth_tooltip_text
+  def auth_provider_enable_btn_disable_class
+    SiteConfig.invite_only_mode ? "disabled" : ""
+  end
+
+  def auth_provider_disable_btn_disable_class
+    !SiteConfig.allow_email_password_login && authentication_enabled_providers.count == 1 ? "disabled" : ""
```
joshpuetz, line 61: **Suggestion (nitpick)** This is the logic check as line 49 and line 68, consider extracting it to a method

joshpuetz (approved):
LGTM pending fixes from other reviewers!

app/javascript/admin/controllers/config_controller.js
msarit: This has been addressed @joshpuetz 👍

spec/requests/admin/podcasts_spec.rb
```diff
@@ -97,6 +98,7 @@
       expect(podcast.reachable).to eq(true)
       expect(podcast.published).to eq(true)
     end
+    # rubocop:enable RSpec/MultipleExpectations
```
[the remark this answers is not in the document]
msarit: Yes this is intentional

app/helpers/authentication_helper.rb
```diff
@@ -33,23 +37,38 @@ def waiting_on_first_user?
     SiteConfig.waiting_on_first_user
   end
 
-  def disable_email_tooltip_class
+  def email_auth_disable_btn_tooltip_class
```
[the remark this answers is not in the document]
msarit: @vaidehijoshi I included 'disable_btn' in the helper name to clue me in that this helper should be called on `Disable` buttons only 😄  But I will think of a better name ❤️

app/assets/javascripts/utilities/showUserAlertModal.js
msarit: 🤦‍♀️  I will fix. And I'm taking JS linting suggestions ...

app/javascript/admin/controllers/config_controller.js
lisasy: @nickytonline and @ludwiczakpawel : What's your point-of-view here on whether we accomplish a toggle using Javascript OR CSS? I don't think we have this "best practice" defined in our docs yet, and would love to hear your thoughts.

vaidehijoshi: @lisasy Something I want to add here is that sometimes it's not always possible to remove _all_ the JS, but the more that we can lean into CSS, the better.

CSS renders/parses faster than JS. CSS fails much more elegantly than JavaScript (CSS is more forgiving, whereas JS is more error prone and you have to assume or _check_ that an element exists in the DOM before you can do anything to it). Also, when used correctly, CSS is easier to maintain and debug (I can't say the same about JS 😅).

I can't remember where I originally learned about this years ago, but [here](https://gomakethings.com/when-to-use-css-vs.-javascript) is a relevant blog post.

msarit:
@vaidehijoshi re: your [styling observations here](https://github.com/forem/forem/pull/11185#issuecomment-722584966), yes @lisasy is going to fix those quickly once this PR is merged 👍

app/helpers/authentication_helper.rb
msarit: Done!

app/assets/javascripts/utilities/showUserAlertModal.js
msarit: Done!

app/javascript/admin/controllers/config_controller.js
msarit: Done! @vaidehijoshi

app/helpers/authentication_helper.rb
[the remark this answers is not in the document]
msarit: @vaidehijoshi Done!

app/views/admin/configs/show.html.erb
msarit: I'm keeping it as `f` 😄  @vaidehijoshi

app/helpers/authentication_helper.rb
```diff
@@ -33,23 +37,45 @@ def waiting_on_first_user?
...
+    email_login_disabled_and_one_auth_provider_enabled ? "crayons-tooltip" : ""
+  end
+
+  def disabled_attr_on_email_auth_disablebtn
+    invite_only_mode_or_no_enabled_providers ? "disabled" : ""
   end
 
-  def disable_email_tooltip_content
-    SiteConfig.invite_only_mode || authentication_enabled_providers.none? ? disable_email_auth_tooltip_text : ""
+  def disabled_attr_on_auth_rpovider_enablebtn
```
vaidehijoshi, line 64:
```suggestion
  def disabled_attr_on_auth_provider_enablebtn
```

app/views/admin/configs/show.html.erb
```diff
@@ -252,136 +253,60 @@
...
+                            </div>
+                          </div>
+                          <button
+                            class="crayons-btn crayons-btn--secondary"
+                            id="<%= provider.provider_name %>-auth-btn"
+                            data-button-text="<%= authentication_enabled_providers.include?(provider) ? "edit" : "enable" %>"
+                            data-action="click->config#enableOrEditAuthProvider"
+                            data-auth-provider-enable="<%= provider.provider_name %>"
+                            data-enable-auth="<%= authentication_enabled_providers.include?(provider) ? "true" : "false" %>"
+                            <%= disabled_attr_on_auth_rpovider_enablebtn %>>
```
vaidehijoshi, line 293:
```suggestion
                            <%= disabled_attr_on_auth_provider_enablebtn %>>
```

spec/helpers/authentication_helper_spec.rb
```diff
@@ -54,4 +64,55 @@
...
+        allow(SiteConfig).to receive(:authentication_providers).and_return([])
+      end
+
+      it "returns 'crayons-tooltip' class for relevant helpers" do
+        expect(tooltip_class_on_auth_provider_enablebtn).to eq("crayons-tooltip")
+        expect(tooltip_class_on_email_auth_disablebtn).to eq("crayons-tooltip")
+      end
+
+      it "returns 'disabled' attribute for relevant helper" do
+        expect(disabled_attr_on_auth_rpovider_enablebtn).to eq("disabled")
```
vaidehijoshi, line 81:
```suggestion
        expect(disabled_attr_on_auth_provider_enablebtn).to eq("disabled")
```

app/assets/stylesheets/admin.scss
```diff
@@ -128,6 +128,17 @@ label {
...
+  &--label {
+    .enabled-indicator-hidden {
+      display: none;
+    }
+
+    .enabled-indicator-visible {
+      display: flex;
+      align-items: center;
+    }
+  }
```
vaidehijoshi, line 141: any reason not to make these classes just `hidden` and `visible`? 

**also, this is not a blocker, but...** instead of making `display: none;` a class, you could just make that the _default_, and then only toggle the visible class, which would allow you to leverage CSS here _even further_! 

```css
&--label {
  display: none;
  
  .visible {
    display: flex;
    align-items: center;
  }
}
```

I'm not 100% sure that _exact_ CSS is correct (I haven't tried it locally), but something like that _should_ work.

Then, your JS would just only ever have to do this:
```javascript
this.enabledIndicatorTarget.classList.toggle('visible');
```

since `display: none` would be the _default_ you wouldn't have to go through the effort of adding `enabled-indicator-hidden` to the `<div>` like you do [here](https://github.com/forem/forem/pull/11185/commits/f4763660d3f1e9845deecad3e2662849425118ad#diff-892ec00ceea1d889fc3c52af2033fa1932e7c61d267d7acfc1027ae5b5029e0cR177) and [here](https://github.com/forem/forem/pull/11185/commits/f4763660d3f1e9845deecad3e2662849425118ad#diff-892ec00ceea1d889fc3c52af2033fa1932e7c61d267d7acfc1027ae5b5029e0cR281).

Instead, it would look like this:
```html
<div class="<%= SiteConfig.allow_both_email_signup_and_login ? "visible" : "" %>"
```
The beauty of this is that you're leaning on CSS's "cascade" (_cascading style sheets_) and defining the _default_ style _within the `&--label` class itself_, rather than having to define a `hidden` class and then applying it as the default :)

I'm happy to pair on this with you in a separate PR if you're curious and would like to figure it out together!

app/javascript/admin/controllers/config_controller.js
```diff
@@ -238,16 +236,5 @@ export default class ConfigController extends Controller {
...
-
-  showHideAuthEnabledIndicator(indicatorDiv, action) {
-    if (action === 'show') {
-      indicatorDiv.classList.add('flex', 'items-center');
-      indicatorDiv.classList.remove('hidden');
-    } else if (action === 'hide') {
-      indicatorDiv.classList.remove('flex', 'items-center');
-      indicatorDiv.classList.add('hidden');
-    }
-  }
```
vaidehijoshi: YAY! Love your improvements (and that we can get rid of this! 😄)

vaidehijoshi (approved):
Great work here, @msarit! This PR has come so far since it's original iteration 🤩 

I left a few minor typo fix suggestions, and a comment about the hidden/visible class (like I mentioned in my comment, I'm happy to pair on that if you'd like, as part of a separate PR!)

msarit:
> Great work here, @msarit! This PR has come so far since it's original iteration 🤩
> 
> I left a few minor typo fix suggestions, and a comment about the hidden/visible class (like I mentioned in my comment, I'm happy to pair on that if you'd like, as part of a separate PR!)

Thanks so much @vaidehijoshi ❤️ 

I went ahead and implemented the proper CSS cascade 😅 

However, I either `add`ed or `remove`d the `.visible` class, depending on the purpose of the button initiating the event. I think that makes the application of the `visible` class more deliberate, than merely reversing its presence or absence with `toggle`.

PR #11187, opened 2020-10-30
Filter by a few other roles when adding potential mods

Zhao-Andy:
- [x] Refactor
- [x] Bug Fix

## Description
This updates the potential tag mod page `/admin/mods?state=potential` (used for finding and adding new mods) to not show `warned` and `comment_banned` users. Also refactors it a bit.

https://github.com/forem/InternalProjectPlanning/issues/248

1. In Rails console, ban a user `User.first.add_role :banned`
2. Visit `/admin/mods?state=potential`
3. Ensure that user with ID 1 does not appear

## Added tests?
- [x] No, and this is why: _there are already tests for this_

## Added to documentation?
- [x] No documentation needed

A detective from Scooby Doo uses a magnifying lens, and covers his whole face with it so you only see a blinking eye.

### DIFF
--- a/app/queries/admin/moderators_query.rb
+++ b/app/queries/admin/moderators_query.rb
@@ -4,7 +4,7 @@ module Admin
       state: :trusted
     }.with_indifferent_access.freeze
 
-    NAMES_FOR_POTENTIAL = %i[banned warned trusted comment_banned].freeze
+    VALID_ROLES = %i[banned warned trusted comment_banned].freeze
 
     def self.call(relation: User.all, options: {})
       options = DEFAULT_OPTIONS.merge(options)
--- a/app/queries/admin/moderators_query.rb
+++ b/app/queries/admin/moderators_query.rb
@@ -4,20 +4,20 @@ module Admin
       state: :trusted
     }.with_indifferent_access.freeze
 
+    NAMES_FOR_POTENTIAL = %i[banned warned trusted comment_banned].freeze
+
     def self.call(relation: User.all, options: {})
       options = DEFAULT_OPTIONS.merge(options)
       state, search = options.values_at(:state, :search)
 
       relation = if state.to_s == "potential"
                    relation.where(
-                     "id NOT IN (SELECT user_id FROM users_roles WHERE role_id = ?)
-                     AND id NOT IN (SELECT user_id FROM users_roles WHERE role_id = ?)",
-                     role_id_for(:trusted),
-                     role_id_for(:banned),
+                     "id NOT IN (SELECT user_id FROM users_roles WHERE role_id IN (?))",
+                     potential_role_ids,
                    ).order("users.comments_count" => :desc)
                  else
                    relation.joins(:roles)
-                     .where(users_roles: { role_id: role_id_for(state) })
+                     .where(users_roles: { role_id: Role.find_by(name: state)&.id })
                  end
 
       relation = search_relation(relation, search) if search.presence
@@ -25,8 +25,8 @@ module Admin
       relation
     end
 
-    def self.role_id_for(role)
-      Role.find_by(name: role)&.id
+    def self.potential_role_ids
+      Role.where(name: NAMES_FOR_POTENTIAL).select(:id)
     end
 
     def self.search_relation(relation, search)
--- a/app/queries/admin/moderators_query.rb
+++ b/app/queries/admin/moderators_query.rb
@@ -26,7 +26,7 @@ module Admin
     end
 
     def self.potential_role_ids
-      Role.where(name: NAMES_FOR_POTENTIAL).select(:id)
+      @potential_role_ids ||= Role.where(name: VALID_ROLES).select(:id)
     end
 
     def self.search_relation(relation, search)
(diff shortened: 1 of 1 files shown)


app/queries/admin/moderators_query.rb
```diff
@@ -4,29 +4,29 @@ class ModeratorsQuery
...
 
       relation = search_relation(relation, search) if search.presence
 
       relation
     end
 
-    def self.role_id_for(role)
-      Role.find_by(name: role)&.id
+    def self.potential_role_ids
+      Role.where(name: NAMES_FOR_POTENTIAL).select(:id)
```
Zhao-Andy, line 29: Feel like this could be refactored, cached, or frozen somehow...

app/queries/admin/moderators_query.rb
```diff
@@ -4,29 +4,29 @@ class ModeratorsQuery
       state: :trusted
     }.with_indifferent_access.freeze
 
+    NAMES_FOR_POTENTIAL = %i[banned warned trusted comment_banned].freeze
```
rhymes, line 7: names for potential roles?

app/queries/admin/moderators_query.rb
```diff
@@ -4,29 +4,29 @@ class ModeratorsQuery
...
-                     AND id NOT IN (SELECT user_id FROM users_roles WHERE role_id = ?)",
-                     role_id_for(:trusted),
-                     role_id_for(:banned),
+                     "id NOT IN (SELECT user_id FROM users_roles WHERE role_id IN (?))",
+                     potential_role_ids,
                    ).order("users.comments_count" => :desc)
                  else
                    relation.joins(:roles)
-                     .where(users_roles: { role_id: role_id_for(state) })
+                     .where(users_roles: { role_id: Role.find_by(name: state)&.id })
```
rhymes, line 20: question: if the role is not found, is there a point in appending this all clause?

rhymes (commented):
Looks fine! Left a couple of notes

app/queries/admin/moderators_query.rb
Zhao-Andy: Hmm good question... I'll revisit this.

rhymes:
@vaidehijoshi there's also another aspect of using `.select(:id)` instead of pluck, Rails will build a subquery, instead of loading ids in memory: 

```sql
SELECT "users".* FROM "users" WHERE (id NOT IN (SELECT user_id FROM users_roles WHERE role_id IN (SELECT "roles"."id" FROM "roles" WHERE "roles"."name" IN ('banned', 'warned', 'trusted', 'comment_banned'))))
```

instead of

```sql
SELECT "users".* FROM "users" WHERE (id NOT IN (SELECT user_id FROM users_roles WHERE role_id IN (1)))
```

It can be more efficient than loading ids in memory, probably the same in this case as we load very few ids

app/queries/admin/moderators_query.rb
```diff
@@ -4,29 +4,29 @@ class ModeratorsQuery
       state: :trusted
     }.with_indifferent_access.freeze
 
+    VALID_ROLES = %i[banned warned trusted comment_banned].freeze
```
citizen428, line 7: This constant doesn't seem to be used anywhere. Should this be `NAMES_FOR_POTENTIAL`, the name you use inside `potential_role_ids`? It's also a somehat less confusing name.

app/queries/admin/moderators_query.rb
citizen428: Rails has a built-in `ids` method for this purpose, which is even simpler than `pluck(:id)` (though it creates the exact same query):

```suggestion
    def self.potential_role_ids
      Role.where(name: NAMES_FOR_POTENTIAL).ids
```

citizen428 (dismissed):
This defines one constant (`VALID_ROLES`) which seems to be never used and uses one constant (`NAMES_FOR_POTENTIAL`) which doesn't seem to be defined.

app/queries/admin/moderators_query.rb
Zhao-Andy: Yeah, sorry I committed half of it and didn't commit the rest 🤦

app/queries/admin/moderators_query.rb
```diff
@@ -4,29 +4,32 @@ class ModeratorsQuery
...
+                     "id NOT IN (SELECT user_id FROM users_roles WHERE role_id IN (?))",
+                     potential_role_ids,
                    ).order("users.comments_count" => :desc)
-                 else
+                 elsif role_id.present?
                    relation.joins(:roles)
-                     .where(users_roles: { role_id: role_id_for(state) })
+                     .where(users_roles: { role_id: role_id })
+                 else
+                   User.none
```
Zhao-Andy, line 23: This seems like an over-optimization to me but 🤷‍♂️

rhymes (approved):
LGTM!

juliannatetreault (approved):
Works as expected! 👍 👍

app/queries/admin/moderators_query.rb
citizen428: IMHO this is great 👍 Consistently returning AR collections is a good practice that we should follow more.

PR #11190, opened 2020-10-30
Feature: Add delete button to following tag

### LINKED ISSUE #9248: User have ability to remove following tag in the dashboard

When I do working on issue #9211, I realize that we can't remove the following tag from the dashboard

Add the feature that we can click a delete button to remove the following tag that we don't want to follow

**Additional context**

image

FreakLevel:

- [x] Feature

## Description

Now the cards of following tags have a delete button.

Close #9248

Screen Shot 2020-10-30 at 13 38 38

### DIFF
--- a/app/controllers/follows_controller.rb
+++ b/app/controllers/follows_controller.rb
@@ -85,6 +85,12 @@ class FollowsController < ApplicationController
     redirect_to "/dashboard/following" if @follow.update(follow_params)
   end
 
+  def destroy
+    @follow = Follow.find(params[:id])
+    authorize @follow
+    redirect_to "/dashboard/following" if @follow.destroy
+  end
+
   private
 
   def follow_params
--- a/app/policies/follow_policy.rb
+++ b/app/policies/follow_policy.rb
@@ -7,6 +7,10 @@ class FollowPolicy < ApplicationPolicy
     user_is_follower?
   end
 
+  def destroy?
+    !user_is_banned?
+  end
+
   def permitted_attributes
     %i[points]
   end
--- a/app/views/dashboards/following_tags.html.erb
+++ b/app/views/dashboards/following_tags.html.erb
@@ -26,14 +26,21 @@
             <% if tag %>
               <% color = HexComparer.new([tag.bg_color_hex || "#0000000", tag.text_color_hex || "#ffffff"]).brightness(0.8) %>
               <div class="crayons-card branded-2 p-4 m:p-6 m:pt-4 flex flex-col single-article break-word content-center <% if follow.points < 0 %>opacity-75<% end %>" style="border-top-color: <%= color %>;" id="follows-<%= follow.id %>">
-                <h3 class="s:mb-1 -ml-1 p-0 fw-medium">
-                  <a href="/t/<%= tag.name %>" class="crayons-tag crayons-tag--l">
-                    <span class="crayons-tag__prefix">#</span><%= tag.name %>
-                  </a>
-                  <% if follow.points < 0 %>
-                    <span class="crayons-indicator crayons-indicator--critical crayons-indicator--outlined" title="This tag has negative follow weight">Anti-follow</span>
+                <div class="crayons-card__follow">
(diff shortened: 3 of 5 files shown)


app/controllers/follows_controller.rb
```diff
@@ -85,6 +85,12 @@ def update
     redirect_to "/dashboard/following" if @follow.update(follow_params)
   end
 
+  def destroy
+    @follow = Follow.find(params[:id])
+    authorize @follow
+    redirect_to "/dashboard/following" if @follow.destroy
```
rhymes, line 91: what happens if the destroy doesn't work?

ps. you can replace the hardcoded path with `dashboard_following_path`

app/policies/follow_policy.rb
```diff
@@ -7,6 +7,10 @@ def update?
     user_is_follower?
   end
 
+  def destroy?
+    !user_is_banned?
```
rhymes, line 11: Should we also check that the requesting user is the right one? See `user_is_follower?`

rhymes (changes requested):
Hi @FreakLevel! 

Thanks for the PR!

A couple of general things in addition to the inline notes:

- the UI for the unfollow on hover is a bit off:

Screenshot 2020-10-30 at 5 45 30 PM

as you can see the highlighted box is not correctly padded on the bottom. 

But in general I'm not sure a giant `X` is what we'd like for the unfollow functionality (personally I would also ask for confirmation but it might be superfluous). cc @ludwiczakpawel  here

- the PR is missing tests :D Let me know if you need help with those, probably should be added in somewhere in `spec/system/dashboards/user_followers_display_spec.rb`

ludwiczakpawel:
I have some thoughts/concerns about UI implementation but let me circle back to this on Monday if I'm not able to look into this and test properly today.

app/views/dashboards/following_tags.html.erb
```diff
@@ -26,14 +26,21 @@
...
+                      <span class="crayons-tag__prefix">#</span><%= tag.name %>
+                    </a>
+                    <% if follow.points < 0 %>
+                      <span class="crayons-indicator crayons-indicator--critical crayons-indicator--outlined" title="This tag has negative follow weight">Anti-follow</span>
+                    <% end %>
+                  </h3>
+                  <%= form_for(follow, html: { method: :delete }) do %>
+                    <button type="submit" class="crayons-btn crayons-btn--ghost crayons-btn--s">
+                      <%= inline_svg_tag("cancel.svg", class: "crayons-icon", aria: true, title: "Unfollow", height: 16, width: 16) %>
+                    </button>
```
ludwiczakpawel, line 41: I'd recommend using x.svg instead of cancel.svg - the SVG code is just a bit cleaner.

As for button, it must include some extra class indicating it's an icon only type of button.

```suggestion
                    <button type="submit" class="crayons-btn crayons-btn--ghost crayons-btn--s crayons-btn--icon">
                      <%= inline_svg_tag("x.svg", class: "crayons-icon", aria: true, title: "Unfollow", height: 24, width: 24) %>
                    </button>
```

^ that should look slightly better.

ludwiczakpawel (changes requested):
Regardless of the changes I've suggested in the code I do have some issues about the UI/UX flow..

So we used to have a "Follow" / "Unfollow" button for tags in dashboard but some went off with some updates couple months ago and it stopped working. I'd be much more in favor of bringing the previous UX rather than adding an "x" button.

Here's the flow we'd like to have eventually: 

image

BUT! Given the scope of this flow is much more bigger than the scope of this PR, I think the good "first iteration"  would be at least replacing your "X" button with the "Unfollow" button at the bottom of this card. The rest of the behavior could easily wait or be tackled in a separate PR.

cmgorton:
Hey @FreakLevel 
Thank you so much for your contribution! Since this PR has not been active for a while I am going to go ahead and close it. 
We couldn't do what we do without awesome contributors like yourself, thank you so much for being a part of our awesome Forem community 🤗

PR #11197, opened 2020-10-30
Refactor:Show Same Number of Articles for all Envs

mstruve:
- [x] Refactor

## Description
There is no reason we can't load the same amount of articles for the feed in development and test as we do in production. 

## Added tests?
- [x] No, minor implementation detail change but I did update some and fixed some descriptions!

alt_text

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -12,7 +12,7 @@ class StoriesController < ApplicationController
     ]
   }.freeze
 
-  SIGNED_OUT_RECORD_COUNT = (Rails.env.production? ? 60 : 10).freeze
+  SIGNED_OUT_RECORD_COUNT = 60
 
   before_action :authenticate_user!, except: %i[index search show]
   before_action :set_cache_control_headers, only: %i[index search show]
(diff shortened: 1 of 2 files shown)


vaidehijoshi (approved):

rhymes (commented):
Cool! One note:

`SIGNED_OUT_RECORD_COUNT` is used twice in a conditional in `app/controllers/stories_controller.rb` based on the sign in status of the user.

What if we also surface those two numbers, `5` and `2` as constants that are dispersed around the code?

mstruve:
@rhymes I think that would be a good refactor in the future but is out of the scope of this particular PR which I just want to focus on removing the Rails.env conditional

vaidehijoshi:
> What if we also surface those two numbers, `5` and `2` as constants that are dispersed around the code?

I like @rhymes suggestion here :) FWIW, I've started making issues for these kinds of small refactors in the internal project planning repo and linking back to the original PR where it was discovered/suggested. That way, they aren't forgotten and are documented somewhere, but don't have to block other changes

rhymes (approved):

PR #11198, opened 2020-10-30
Refactor:Always Load Last 24 Hour Podcasts

mstruve:
- [x] Refactor

## Description
There is not really a good reason to treat production any different when showing recent podcasts. If devs want to mimic this behavior locally they can always go in and update podcast published dates to be within a day. 

## Added tests?
- [x] No, minor implementation change

alt_text


Zhao-Andy (approved):
Makes sense to me!

rhymes (approved):
Agreed!

PR #11200, opened 2020-10-31
Optimization:Speed Up Notifications Spec By Combining Assertions

mstruve:
- [x] Optimization

## Description
`spec/requests/notifications_spec.rb` is by far [our longest-running spec](https://knapsackpro.com/dashboard/organizations/1142/projects/1022/test_suites/1434/builds/3dc22658-c4f4-4f22-90a8-6a067bfcde92) with an average run time around 75-85 seconds(over a min). The reason is that setting up each of the specs takes time since we have to create all the models and notifications, then we have to send the notifications. Most of these tests we run the set up 4 or 5 times to check different attributes on the page. This change runs the setup once and then using `:aggregate_failures` it checks the page for all the necessary attributes for each spec.

With this change locally the entire file was taking around 45 seconds to run. In [the most recent knapsack run](https://knapsackpro.com/dashboard/organizations/1142/projects/1022/test_suites/1434/builds/864ee901-74bb-4dfc-9513-68427330da9c) it took under 40 seconds 
Screen Shot 2020-10-30 at 8 16 40 PM

This can be merged by any core member

## Added tests?
- [x] Updated

alt_text


vaidehijoshi (approved):
Nice little win here!!

rhymes (approved):
Thank you! It would be great if we could add your explanation at the beginning of the file as a comment, this way in 6 months time we'll still know why certain tricks were applied :) As this is definitely an optimization that might not be self evident for most

mstruve:
Rather than adding comments to the individual spec file which isn't super useful unless you are in that file, I added them to the acceptance test docs where they are more visible

rhymes:
@mstruve great call!

PR #11225, opened 2020-11-01
Spec Fix:Remove Extra Space From Single Number Dates in Spec

mstruve:
- [x] Spec Fix

## Description
Remove the extra white space that occurs when a published_at day is a single number in the specs

This PR can be merged by any core team member. 

alt_text


vaidehijoshi (approved):
Huh, it's interesting that this started failing _now_ and not before?

juliannatetreault (approved):
So strange, but great catch!

mstruve:
> Huh, it's interesting that this started failing now and not before?

@vaidehijoshi I am going to guess that our javascript formatting handled that before, since I removed that javascript from the spec bc it made it flaky, then we started seeing this issue.

Issue #9658, opened 2020-08-06
The Series list is showing empty series (with 0 articles)


beeman:
Today I saw there is a page that shows the users' series. While this is awesome, I think it would be even more awesome if it would not show the empty articles.

In my case, as seen in the screenshot, I have a few series that are there as work-in-progress, I'd love those not to be visible until they have at least one post published. In another case, a series appears double. Probably because I recreated some of the posts.

**To Reproduce**

- Visit https://dev.to/beeman/series
- Admire my series
- Be surprised there are some empty and double series

**Expected behavior**

I would love it if the empty series (0 articles) would not appear, and add an option for authors to delete series.

**Screenshots**

![image](https://user-images.githubusercontent.com/36491/89588174-be2a4280-d808-11ea-99aa-280509f3990b.png)

**Desktop (please complete the following information):**

- OS: macOS Catalina 10.15.5 (19F101)
- Browser: Brave
- Version: Version 1.11.104 Chromium: 84.0.4147.105 (Official Build) (64-bit)

**Smartphone (please complete the following information):**

- Device:
- OS:
- Browser:
- Version:

**Additional context**

I'm happy to give a go at implementing this, it would be great if someone could give me some pointers on what changes are expected :-)

rhymes:
@beeman go ahead! I think we can start by only showing series that have at least one article in the public series page. 

You can find the code that selects series here: https://github.com/forem/forem/blob/master/app/controllers/collections_controller.rb and its tests here https://github.com/forem/forem/blob/master/spec/requests/collections_spec.rb (maybe you can add a test to make sure that series without articles don't appear in the result).

I would deal with "deleting empty series" separately to keep the PR small

rhymes:
@beeman are you still working on this? Let us know, thanks!

beeman:
@rhymes I started on it but got distracted. Happy for someone else to pick it up.

If I decide to look at it again, I'll post a message here first.

citizen428:
I unassigned you now @beeman, thanks for giving it a shot 😃

cwray-tech:
Hey, @citizen428  I would really like this issue if possible. Also, I would like to add edit and delete functionality to the series. Should I do this all in the same pr, or is there another issue I should tackle as well? Thanks!

rhymes:
@cwray-tech I think they should be in different PRs are they are related to different functionalities. I'm assigning you this issue

cwray-tech:
[...]

PR #11227, opened 2020-11-02
Fix Bug: Series index page showing series with 0 articles -#9658

cwray-tech:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
I added a join and distict method to the collections controller index method to only get collections that contain articles. 

Closes https://github.com/forem/forem/issues/9658

To test this functionality, you can create an article, add a series to that article. Save the Series/article, and create another article and a new series. 

After saving the new series/ article, replace the second article series with the first series created.

There should now be one series without any articles. 

Go to the series index view, by clicking on the series of the article, then clicking the "back to all series" link. Here, you should only see the first series created with the two articles.

## Added tests?

- [ ] Yes
      have not been included_
- [x] I need help with writing tests. I am not sure if a test is needed as there is already a test written for the index method. 

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

No

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -1,7 +1,7 @@
 class CollectionsController < ApplicationController
   def index
     @user = User.find_by!(username: params[:username])
-    @collections = @user.collections.order(created_at: :desc)
+    @collections = @user.collections.joins(:articles).distinct.order(created_at: :desc)
   end
 
   def show
--- a/spec/system/collections/user_views_collections_spec.rb
+++ b/spec/system/collections/user_views_collections_spec.rb
@@ -2,17 +2,26 @@ require "rails_helper"
 
 RSpec.describe "Visiting collections", type: :system do
   let(:user) { create(:user) }
-  let!(:collection1) { create(:collection, :with_articles, user: user) }
-  let!(:collection2) { create(:collection, user: user) }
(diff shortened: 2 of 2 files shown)


cwray-tech:
@citizen428 Thank you SO much for your help with this. All credit goes your way. (:

rhymes (changes requested):
Hi @cwray-tech - thanks for tackling this! I think we should add a test to make sure this doesn't "regress" in the future.

I think the simplest would be to alter https://github.com/forem/forem/blob/master/spec/system/collections/user_views_collections_spec.rb by creating a collection with no articles and making sure it's not visible.

Let us know if you need further help

cwray-tech:
@rhymes thank you, yes will do! This is so good for me. Learning a lot. (: Sorry I was not able to run tests locally, so I didn't catch that. I think I am able to run tests now so will try to ensure they all pass before adding another commit.

citizen428:
@cwray-tech No, thank you! This is not about credit but about building Forem together with our community, so thanks a lot for your contribution and tackling this issue, even if it was maybe a bit outside your comfort zone 💯

citizen428 (approved):

rhymes (approved):
Looks awesome! Thanks @cwray-tech

PR #11233, opened 2020-11-02
Refactor:Implement Carrierwave Monkeypatch in all Envs

mstruve:
- [x] Refactor

## Description
Having patches like this in the code that are only applied in Production can be risky. The reason it is risky is bc if a problem comes up with the patch say because of a gem upgrade we won't have any indication of it(at this point down the road with some more intricate testing strategies we likely will) until the code is in use by users. With the guard removed, there is a chance we can catch issues before then. 

I also added a test for this patch so if the patch is removed accidentally, the test will break as a warning. 

I tested this locally and everything works as expected and with the tests passing **I dont think** there is any reason why we need this guard clause. 
 
## Added tests?
- [x] YES

alt_text

### DIFF
--- a/config/initializers/carrierwave_monkeypatch.rb
+++ b/config/initializers/carrierwave_monkeypatch.rb
@@ -3,15 +3,13 @@
 # We also force this to use the SiteConfig instead of APP_DOMAIN because the value
 # could change after initial boot.
 
-if Rails.env.production?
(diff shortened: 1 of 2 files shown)


maestromac (approved):
Good thinking :+1:

rhymes (approved):

PR #11234, opened 2020-11-02
Add users param to Feeds::Import

rhymes:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This will allows us to pass a subset of users to `Feeds::Import`, directly or via the temporary `Articles::DevFeedsImportWorker` job

#10998 

1. Add a feed to a user
1. Load the console
1. Run `Feeds::Import.call(users: User.where(id: ID_OF_THE_USER))`
1. Set `SiteConfig.community_name` to `DEV`
1. Run `Articles::DevFeedsImportWorker.new.perform([ID_OF_THE_USER])`

## Added tests?

- [x] Yes
      have not been included_


spec/services/feeds/import_spec.rb
```diff
@@ -63,28 +63,54 @@
...
+
+        expect(Rails.logger).to have_received(:error).at_least(:once)
+      end
+    end
+
+    context "with an explicit set of users", vcr: { cassette_name: "feeds_import" } do
+      it "accepts a subset of users" do
+        num_articles = described_class.call(users: User.with_feed.limit(1))
+
+        verify(format: :txt) { num_articles }
```
mstruve, line 105: Totally random question here, why do you choose to use a `verify` here rather than `expect(num_articles).to eq(10)` ?

app/models/user.rb
```diff
@@ -211,6 +211,7 @@ def sync_profile
 
   scope :eager_load_serialized_data, -> { includes(:roles) }
   scope :registered, -> { where(registered: true) }
+  scope :with_feed, -> { where.not(feed_url: [nil, ""]) }
```
mstruve, line 214: ❤️

mstruve (approved):
Nice!

app/models/user.rb
citizen428: Given that we have quite a few users, do you think it could make sense to add a [partial index](https://use-the-index-luke.com/sql/where-clause/partial-and-filtered-indexes) for this? 

```sql
CREATE INDEX feed_url_present_id
ON users (feed_url)
WHERE coalesce(feed_url, '') <> '';
```

app/services/feeds/import.rb
```diff
@@ -3,13 +3,13 @@
...
+    def self.call(users: nil)
+      new(users: users).call
     end
 
-    # TODO: add `users` param
-    def initialize
-      @users = User.where.not(feed_url: [nil, ""])
+    def initialize(users: nil)
+      # using nil here to avoid an unnecessary table count to check presence
+      @users = users.nil? ? User.with_feed : users
```
citizen428, line 12: Assuming `users` won't be a literal `false`, this could change to the following which feels slightly more idiomatic in Ruby. 

```suggestion
      @users = users || User.with_feed
```

spec/services/feeds/import_spec.rb
```diff
@@ -63,28 +63,54 @@
       end
     end
 
-    # it "reports an article creation error" do
-    #   allow(described_classs).to receive(:create_articles_from_user_feed).and_raise(StandardError)
-    #   allow(Honeybadger).to receive(:notify)
+    it "queues as many slack messages as there are articles", vcr: { cassette_name: "feeds_import" } do
```
citizen428, line 66: We do have `configure_rspec_metadata!` enabled, so it would infer the cassette name from the spec description:

```suggestion
    it "queues as many slack messages as there are articles", :vcr do
```

citizen428 (approved):
Just left some comments, will leave it up to your judgment which ones (if any) you want to implement.

spec/services/feeds/import_spec.rb
rhymes: I want to avoid that, I'd prefer to recycle the same spec instead of having multiple VCR recordings with the same data

app/services/feeds/import.rb
rhymes: Ah ah the `.nil?` is a remnant of me using `users: Users.none` as the default but yeah, we can change that

rhymes:
@citizen428 the index is definitely on my TODO list. not sure if total or partial but it's definitely there, thanks for the reminder :-)

citizen428:
> not sure if total or partial 

I just did a Blazer query and only 3,492 users have a `feed_url` that's not `null` or an empty string. That's such a small percentage that I think partial makes more sense.

rhymes:
@citizen428 agreed!

PR #11235, opened 2020-11-02
Feature:Default Rails Consoles to Readonly

mstruve:
- [x] Feature

## Description
Opening a production console is always a bit scary because you want to be extra careful you don't change anything you shouldn't. With this PR, if we have a BLAZER_DATABASE_URL present such as on Heroku, the console will connect to the database with that URL instead. The advantage here is that the BLAZER_DATABASE_URL is read-only and shields user sensitive data. 

## Added tests?
- [x] No, bc this is a production only feature. 

alt_text


mstruve:
I think @benhalpern and I came up with some better ideas and solutions for some of our access woes so I am going to close this while we explore those.

PR #11237, opened 2020-11-02
Developer experience/security: Make dev/test blow up if current_user is used erroneously

benhalpern:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We have paths in the app where the whole response is cached _at the edge_. This is outlined here:

https://docs.forem.com/technical-overview/architecture/#we-cache-many-content-pages-on-the-edge

But we have _consistently_ had issues where someone forgets this constraint and we push code that is problematic.

We have this issue right now. Thankfully it did not leak private data, as could be possible with this issue.

This PR redefines `current_user` in the event of cache headers being in place that only get set if we have a code path that should be edge-cached.

This PR includes removal of some code that was put in place that violates the guidelines.

Having code that is automatically defensive against some of our issues instead of having to catch them in review seems like a step in the right direction.

I'm sure there are cleaner ways to deal with this if anyone has any suggestions.


app/controllers/application_controller.rb
```diff
@@ -172,9 +172,27 @@ def forward_to_app_config_domain
...
   protected
 
   def configure_permitted_parameters
     devise_parameter_sanitizer.permit(:sign_up, keys: %i[username name profile_image profile_image_url])
   end
+
+  def edge_cached_action?
+    # X-Accel-Expires will only be present if we are setting cache headers
+    # And let's only run this code in development and test
+    !Rails.env.production? && response.headers["X-Accel-Expires"].present?
```
atsmith813, line 196: **question**: Is there a way we can make this a little more future proof? I suspect how we define what's "cached" will change in the future. For example, this header is only for Nginx though it is currently always set on cached responses. If we change how we're using Nginx caching, this might break.

**suggestion**: I know this a protected method, so maybe some sort of spec to test the behavior higher up would help account for the future updates?

atsmith813 (commented):
**question (blocking)**: Can we add some specs around this new logic?

mstruve (commented):
I really like this idea. What do you think about putting this in a separate file and then only including that file in nonproduction environments like we did for [include DevelopmentDependencyChecks](https://github.com/forem/forem/blob/master/app/controllers/application_controller.rb#L15)

I think this would be a bit cleaner and I like having the different env code in a separate file.

benhalpern:
> What do you think about putting this in a separate file and then only including that file in nonproduction environments like we did for include DevelopmentDependencyChecks

@mstruve good call

app/controllers/application_controller.rb
```diff
@@ -172,9 +172,27 @@ def forward_to_app_config_domain
     redirect_to URL.url(request.fullpath)
   end
 
+  def current_user
+    return super unless edge_cached_action?
+
+    return nil unless session_current_user_id.present?
+      message = "😱😱😱😱😱 You cannot use current_user in this action\
```
citizen428, line 179: Can we move this message out into a constant? We don't need to recreate this string on every method invocation.

app/controllers/application_controller.rb
```diff
@@ -172,9 +172,27 @@ def forward_to_app_config_domain
...
+  because this code path will be cached at the edge in production.\
+  Check out this section in the docs: https://docs.forem.com/technical-overview/architecture/#we-cache-many-content-pages-on-the-edge"
+      p message
```
citizen428, line 182: A debugging message slipped through.

app/controllers/application_controller.rb
```diff
@@ -172,9 +172,27 @@ def forward_to_app_config_domain
...
+      Object.const_set("YouCannotUseCurrentUserHereCheckTheTerminalOutput", Class.new)
```
citizen428, line 183: 1. What is the purpose of this?
2. Classes really shouldn't be defined inside methods. This will generate two warnings on every call to `current_user` after the first:

```ruby
(irb):2: warning: already initialized constant YouCannotUseCurrentUserHereCheckTheTerminalOutput
(irb):2: warning: previous definition of YouCannotUseCurrentUserHereCheckTheTerminalOutput was here
```

app/controllers/application_controller.rb
citizen428, line 179: Nitpick: it seems the indentation is off here. Also, Rubocop normally requires a blank line after guard clauses.

citizen428 (dismissed):
Requesting changes as there's still some debugging code in there and some things that I really don't understand (repeatedly trying to set a constant inside `current_user` for example).

app/controllers/concerns/caching_headers.rb
```diff
@@ -21,6 +21,8 @@ def set_cache_control_headers(
 
     request.session_options[:skip] = true # no cookies
 
+    @edge_caching_in_place = true # Used primarily for downstream meta information about whether a path is cached.
```
mstruve, line 24: Is this Thread-safe? I think it is because the modules are included and used in instances of the ApplicationController so this would be tied to a particular instance...can someone double check my logic there? Is that correct? I want to make sure we don't start leaking variables.

app/controllers/concerns/edge_cache_safety_check.rb
```diff
@@ -0,0 +1,17 @@
+module EdgeCacheSafetyCheck
+  extend ActiveSupport::Concern
+
+  CANNOT_USE_CURRENT_USER = "You may not use current_user in this cached code path.".freeze
```
mstruve, line 4: 🔥

app/controllers/application_controller.rb
```diff
@@ -13,6 +13,7 @@ class ApplicationController < ActionController::Base
   include ImageUploads
   include VerifySetupCompleted
   include DevelopmentDependencyChecks if Rails.env.development?
+  include EdgeCacheSafetyCheck unless Rails.env.production?
```
mstruve, line 16: Nice!

mstruve (commented):
One small detail I want to double-check otherwise this looks good!!!

app/controllers/concerns/caching_headers.rb
benhalpern: I think we can set instance variables here like we could with any method that sits within a controller. But yeah, maybe it's different with included modules?

... For this particular functionality it shouldn't matter since we only care about this in dev environment, but of course better do it the right way so we don't wrongly rely on this if it is not thread safe for future functionality.

atsmith813 (approved):

app/controllers/concerns/edge_cache_safety_check.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def current_user
+    # In production, using current_user in a view which will be cached will result in a cache leak.
+    # https://docs.forem.com/technical-overview/architecture/#we-cache-many-content-pages-on-the-edge
+    return super unless @edge_caching_in_place
```
citizen428, line 10: This relies on an instance variable set by another concern. To me, this indicates that the two aren't really independent because one has knowledge about a very specific implementation detail of the other. Nothing good ever comes from this sort of implicit dependency.

I see several options:

1. We use the guaranteed thread-safe [request_store](https://github.com/steveklabnik/request_store) gem by Steve Klabnik:
    ```ruby
    #  app/controllers/concerns/caching_headers.rb
    RequestStore.store[:edge_caching_in_place] = true 

    # app/controllers/concerns/edge_cache_safety_check.rb
    return super unless RequestStore.store[:edge_caching_in_place]
2. You move this `current_user` implementation into the `CachingHeaders` concern but wrap it in a conditional (Ruby allows arbitrary code in class/module bodies):
    ```ruby
    unless Rails.env.production?
      def current_user
        ...
      end
    end
    ```

1. has the advantage of being future proof if you ever need to access the value of `RequestStore.store[:edge_caching_in_place]` anywhere else in the app. 2. is nice because everything caching related is stored in the same concern, without hidden dependencies. One could even consider combining both approaches once we actually do need to access this value in other parts of the code.

app/controllers/concerns/edge_cache_safety_check.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    return if session_current_user_id.blank?
+
+    @current_user_called_erroneously = true
```
citizen428, line 14: Is this used anywhere?

app/controllers/concerns/edge_cache_safety_check.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    CANNOT_USE_CURRENT_USER
```
citizen428, line 16: Shouldn't this raise, not just return a string instead of a user object?

benhalpern: The issue is that `user_signed_in?` still uses `current_user` for truthiness. So returning the string means that it will show the error if someone tries to call a method on `current_user` but `user_signed_in?` and anything else that relies on `current_user` for boolean presence will still work.

app/controllers/concerns/edge_cache_safety_check.rb
benhalpern: Good call. This was left over of another implementation idea.

app/controllers/concerns/edge_cache_safety_check.rb
benhalpern: I went with #1.

app/controllers/concerns/caching_headers.rb
```diff
@@ -21,6 +21,8 @@ def set_cache_control_headers(
 
     request.session_options[:skip] = true # no cookies
 
+    RequestStore.store[:edge_caching_in_place] = true # Used primarily for downstream meta information about whether a path is cached.
```
benhalpern, line 24:
```suggestion
    RequestStore.store[:edge_caching_in_place] = true # To be observed downstream.
```

app/controllers/concerns/caching_headers.rb
```diff
@@ -21,6 +21,8 @@ def set_cache_control_headers(
 
     request.session_options[:skip] = true # no cookies
 
+    RequestStore.store[:edge_caching_in_place] = true # To be observed downstream.
```
mstruve, line 24: 🔥

mstruve (approved):
Safety first!!!

app/controllers/concerns/caching_headers.rb
rhymes: I think we should surface this gem in the `Gemfile`, right now it's a transitive dependency of `rails-settings-cached` but as any of the other gems that we use directly, it should be listed in the `Gemfile`. Something like:

```ruby
gem 'request_store', '~> 1.5' # RequestStore gives you per-request global storage
```

benhalpern: @rhymes 👍

PR #11238, opened 2020-11-02
Replace SimpleIDN usage with Addressable 

maestromac:
- [x] Bug Fix

## Description
The one problem with SimpleIDN gem is that it downcase the URL completely. Addressable, on the other hand, provides a normalized method that would only convert non-Unicode string to ASCII.

Resolves https://github.com/forem/InternalProjectPlanning/issues/239

The test kind of already covers this pretty well, no QA needed.

## Added tests?
- [x] No, and this is why: updated existing spec

## Added to documentation?
- [x] No documentation needed

n/a


mstruve (approved):
Nice find and +1 for removing a gem! 🚀

rhymes (approved):
:fire:

Ridhwana (approved):
Nice One @maestromac! Thanks 😊

PR #11242, opened 2020-11-03
Add trusted role to new tag moderators

Zhao-Andy:
- [x] Refactor
- [x] Feature

## Description
This adds the `trusted` role to new tag moderators. For whatever reason before, we didn't add this to new tag mods -- maybe because we assumed they were already trusted.

Also adds a bunch of little refactoring and quality of life changes:
- make add/remove tag moderator into separate routes
- adds flash messages on submit
- adds error handling
- Updates design to match Crayons
- adds explanation for what `tag.supported` means

Screenshot_2020-11-02 Tags

### Before:
Screenshot_2020-11-02 Tags(1)

## Related Tickets
Closes https://github.com/forem/InternalProjectPlanning/issues/234

## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

Nope

A broom goes sweep sweep sweep


Zhao-Andy:
QA, I would try:
- adding a tag moderator
- removing a tag moderator
- updating the colors, both the text input and the color field
- updating the wiki markdown (this wasn't working before)

app/labor/assign_tag_moderator.rb
```diff
@@ -1,63 +1,73 @@
...
-    return if user.has_role?(:trusted)
-    return if user.has_role?(:banned)
-
-    user.add_role :trusted
-    user.update(email_community_mod_newsletter: true)
-    MailchimpBot.new(user).manage_community_moderator_list
-    Rails.cache.delete("user-#{user.id}/has_trusted_role")
-    NotifyMailer.with(user: user).trusted_role_email.deliver_now
-  end
+  class << self
```
citizen428, line 2: Personally, not a big fan of this form. It's more obscure to people new to Ruby and adds an extra level of indentation. I prefer the most explicit form, which is the one this PR changes. If we don't want that for whatever reason, `extend self` at least doesn't add extra identation and requires roughly the same level of Ruby knowledge as `class << self`:

```ruby
module Test
  extend self

  def test
    "Hi, Andy! 👋"
  end
end

Test.test
#=> "Hi, Andy! 👋"

Test.methods - Object.methods
#=> [:test]
```

app/labor/assign_tag_moderator.rb
```diff
@@ -1,63 +1,73 @@
...
+    def add_trusted_role(user)
+      return if user.has_role?(:trusted)
+      return if user.has_role?(:banned)
+
+      user.add_role :trusted
```
citizen428, line 7: **nitpick (non-blocking):** Can we add the parenthesis here? All the surrounding code has them.

app/views/admin/tags/show.html.erb
```diff
@@ -11,84 +11,35 @@
...
-    <% end %>
+    <div class="crayons-card__header">
+      <h4 class="crayons-card__headline">Moderators</h4>
+    </div>
+    <div class="crayons-card__body flex flex-col gap-2">
+      <% if @tag.tag_moderator_ids.any? %>
+        <ul class="list-none">
+          <% @tag.tag_moderator_ids.each do |id| %>
+              <li class="flex justify-between">
+                <a href="/<%= User.find_by(id: id)&.username %>">@<%= User.find_by(id: id)&.username %></a>
```
citizen428, line 22: I think this is a bit too much AR for a view. Also this should probably use `URL.user` instead of manually constructing the link.

config/routes.rb
```diff
@@ -103,7 +103,12 @@
           post "save_status"
         end
       end
-      resources :tags, only: %i[index update show]
+      resources :tags, only: %i[index update show] do
+        member do
```
citizen428, line 107: I know we don't do this elsewhere, but I really prefer routing like this:

```ruby
  resource :tag_admin, only: %i[create destroy] # or update if you really want patch
```

Then you make a separate nested controller with two actions and everything is cleanly separated and the routes are actually ResSTful.

http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/

app/labor/assign_tag_moderator.rb
Zhao-Andy: Ah gotcha. I'll revert it back, good point about `self.method` probably being easier to understand for newcomers to Ruby.

app/models/tag.rb
```diff
@@ -89,6 +89,10 @@ def validate_name
     errors.add(:name, "contains non-ASCII characters") unless name.match?(/\A[[a-z0-9]]+\z/i)
   end
 
+  def errors_as_sentence
+    errors.full_messages.to_sentence
+  end
```
Zhao-Andy, line 94: This needs to be added here since the `Tag` model doesn't inherit from `ApplicationRecord`.

config/routes.rb
Zhao-Andy: Hmm so I tried:
```ruby
namespace: admin do
  resources :tags, only: %i[index update show] do
    resource :tag_moderator, only: %i[create destroy], module: "tags"
  end
end
```

The path is correct, but the controller it hits isn't:
```
# show-routes in console

POST (/locale/:locale)/admin/tags/:tag_id/tag_moderator(.:format)
admin/tags/tag_moderators#create {:locale=>nil}
DELETE (/locale/:locale)/admin/tags/:tag_id/tag_moderator(.:format)
admin/tags/tag_moderators#destroy {:locale=>nil}
```

My controller is named like the example:
```
module Admin
  class Tags::TagModeratorsController < ApplicationController
  end
end
```

Any ideas?

maestromac: @Zhao-Andy it looks like the right controller. That's where you placed your controller correct? also, since it's already in a module I would think `Tags::ModeratorsController` is better, you can ommit the 2nd `Tag`.

Zhao-Andy: Yeah, agree that moderators alone would look better. The above code doesn't work though, I keep getting `Uninitialized constant Admin::Tags`.

citizen428: @Zhao-Andy The route is correct, you just don't have that controller. What I meant (taking @maestromac's sensible suggestion re naming into account):
`config/routes.rb`
```ruby
namespace: admin do
  resources :tags, only: %i[index update show] do
    resource :moderator, only: %i[create destroy], module: "tags"
  end
end
```
`app/controllers/admin/tags/moderator_controller.rb`
```ruby
module Admin
  module Tags
    module ModeratorsController < ApplicationController
    
      def create
         ...
      end

      def destroy
        ...
      end
    end
  end
end
```

Zhao-Andy: Ahh thanks! My file path was the wrong part -- I had it only under `/admin`.

app/controllers/admin/tags/moderators_controller.rb
```diff
@@ -0,0 +1,44 @@
+module Admin
+  module Tags
+    class ModeratorsController < Admin::ApplicationController
+      def authorization_resource
+        Tag
+      end
```
Zhao-Andy, line 6: This is to monkey patch `ApplicationController.authorization_resource` for the related `InternalPolicy` (which should be renamed to `AdminPolicy` in a future PR). Without it, `authorization_resource` returns `Moderator`.

citizen428 (approved):

benhalpern (approved):
Looks like a great PR

PR #11245, opened 2020-11-03
Add --spec option for data_update generator

citizen428:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [X] Documentation Update

## Description

During my recent profile generalization work, I had to write quite a few data update scripts and got tired of adding the corresponding specs manually. I decided to add `--spec` / `--no-spec`options to the generator, with `--spec` being the deafault:

```
❯ rails generate data_update -h
Usage:
  rails generate data_update NAME [options]

Options:
  [--skip-namespace], [--no-skip-namespace]  # Skip namespace (affects only isolated applications)
  [--spec], [--no-spec]                      # Indicates when to generate spec
                                             # Default: true
[output snipped]
```

The generated specs are simple and follow our current conventions. The included message is borrowed from Rails' default generators:

```ruby
require "rails_helper"
require Rails.root.join(
  "20201103044008_do_the_thing.rb",
)

describe DataUpdateScripts::DoTheThing do
  pending "add some examples to (or delete) #{__FILE__}"
end
```

I also noticed that our generator doesn't play nicely with `rails destroy data_update`: since we prepend the timestamp to the filename during runtime it will yield a different value on `destroy` vs `generate` and the file won't actually get removed. It works for the newly added specs though, as those use stable filenames. I made a mental note of looking into this during the next downtime, the Rails migration generator should have clues on how to handle this case properly.

n/a

1. Run `rails generate data_update DoSomething` and verify that the script and spec were created and have the right content.
2. Run `rails generate data_update DoSomethingElse --no-spec` and verify that the script gets created as expected but the spec doesn't.

## Added tests?

- [ ] Yes
- [X] No, and this is why: we didn't have any tests for this yet and I want to go back to writing the actual data update script that prompted this. I did make a mental note to add specs for the whole generator during the next cooldown period.

## Added to documentation?

- [X] Docs.forem.com
- [ ] README
- [ ] No documentation needed

### DIFF
--- a/docs/backend/data-update-scripts.md
+++ b/docs/backend/data-update-scripts.md
@@ -40,6 +40,26 @@ module DataUpdateScripts
 end
 ```
 
+The generator will also automatically create the corresponding spec file.
+
+```ruby
+require "rails_helper"
+require Rails.root.join(
+  "20201103042915_backfill_column_for_articles.rb",
+)
+
+describe DataUpdateScripts::BackfillColumnForArticles do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
+```
+
+While we encourage adding tests for data update scripts, you can skip spec
+creation by adding the `--no-spec` option to the `rails generate` command:
+
+```
+rails generate data_update BackfillColumnForArticles --no-spec
+```
+
 Once your script is in place then you can either run `rails data_updates:run`
 manually or you can let our setup script handle it. In our local
 [bin/setup](https://github.com/forem/forem/blob/master/bin/setup) script you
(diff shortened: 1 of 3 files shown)


rhymes (approved):
Works well, cool!

mstruve (approved):
SO AWESOME!!! Thanks @citizen428!

maestromac (approved):
Love it!

vaidehijoshi:
This is rad! Thanks for taking the initiative and building this out for all of us to use, @citizen428! ❤️

PR #11246, opened 2020-11-03
Prepare for profile column drop

citizen428:

- [X] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR came out of a discussion (see [conclusion](https://github.com/forem/forem/pull/10707#issuecomment-719141336) in #10707 and some work done in the closed #11110 branch.

1. Adds a new `:settings_only` option to the `display_area` enum.
2. Updates the CSV importer and the DEV profile fields CSV file to add `display_area` .
3. Adds a data update script for ensuring that all DEV fields are current and profiles are really up to date before starting to ignore the columns in the next PR.

#6994 

## Added tests?

- [X] Yes
      have not been included_

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [X] No documentation needed

### DIFF
--- a/app/models/profile_field.rb
+++ b/app/models/profile_field.rb
@@ -11,7 +11,8 @@ class ProfileField < ApplicationRecord
 
   enum display_area: {
     header: 0,
-    left_sidebar: 1
+    left_sidebar: 1,
+    settings_only: 2
   }
 
   belongs_to :profile_field_group, optional: true
--- a/app/services/profile_fields/field_definition.rb
+++ b/app/services/profile_fields/field_definition.rb
@@ -32,7 +32,7 @@ module ProfileFields
     end
 
     def add_fields
-      self.class.fields.each { |field| ProfileField.create(field) }
+      self.class.fields.each { |field| ProfileField.find_or_create_by(field) }
     end
   end
 end
--- a/app/services/profile_fields/import_from_csv.rb
+++ b/app/services/profile_fields/import_from_csv.rb
@@ -1,6 +1,6 @@
 module ProfileFields
   class ImportFromCsv
-    HEADERS = %i[label input_type placeholder_text description group].freeze
+    HEADERS = %i[label input_type placeholder_text description group display_area].freeze
 
     def self.call(file)
       CSV.foreach(file, headers: HEADERS, skip_blanks: true) do |row|
--- a/lib/data/dev_profile_fields.csv
+++ b/lib/data/dev_profile_fields.csv
@@ -1,29 +1,29 @@
-Display email on profile,check_box,,,Basic
-Name,text_field,John Doe,,Basic
-Website URL,text_field,https://yoursite.com,,Basic
-Summary,text_area,A short bio...,,Basic
-Location,text_field,"Halifax, Nova Scotia",,Basic
-Facebook URL,text_field,https://facebook.com/...,,Links
-Youtube URL,text_field,https://www.youtube.com/channel/...,,Links
-StackOverflow URL,text_field,https://stackoverflow.com/users/...,,Links
-LinkedIn URL,text_field,https://www.linkedin.com/in/...,,Links
-Behance URL,text_field,https://www.behance.net/...,,Links
-Dribbble URL,text_field,https://dribbble.com/...,,Links
-Medium URL,text_field,https://medium.com/@...,,Links
-GitLab URL,text_field,https://gitlab.com/...,,Links
-Instagram URL,text_field,https://www.instagram.com/...,,Links
-Mastodon URL,text_field,https://...,,Links
-Twitch URL,text_field,https://www.twitch.tv/...,,Links
-Education,text_field,,,Work
-Employer name,text_field,Acme Inc.,,Work
-Employer URL,text_field,https://dev.com,,Work
-Employment title,text_field,Junior Frontend Engineer,,Work
-Looking for work,check_box,,,Work
-"Display ""looking for work"" on profile",check_box,,,Work
-Recruiters can contact me about job opportunities,check_box,,,Work
-Skills/Languages,text_area,,What tools and languages are you most experienced with? Are you specialized or more of a generalist?,Coding
-Currently learning,text_area,,"What are you learning right now? What are the new tools and languages you're picking up right now?",Coding
-Currently hacking on,text_area,,What projects are currently occupying most of your time?,Coding
-Available for,text_area,,"What kinds of collaborations or discussions are you available for? What's a good reason to say Hey! to you these days?",Coding
-Brand color 1,color_field,#000000,"Used for backgrounds, borders etc.",Branding
-Brand color 2,color_field,#000000,Used for texts (usually put on Brand color 1).,Branding
+Display email on profile,check_box,,,Basic,settings_only
+Name,text_field,John Doe,,Basic,header
+Website URL,text_field,https://yoursite.com,,Basic,header
+Summary,text_area,A short bio...,,Basic,header
+Location,text_field,"Halifax, Nova Scotia",,Basic,header
+Facebook URL,text_field,https://facebook.com/...,,Links,header
+Youtube URL,text_field,https://www.youtube.com/channel/...,,Links,header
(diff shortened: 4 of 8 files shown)


fdoxyz (commented):
I tried to play around with the branch locally and everything seems to be working well after running `bin/setup`, which [AFAIK](https://docs.forem.com/backend/data-update-scripts) runs the `rails data_updates:run`.

**Side note:** I took a look at the failing specs and adding the `display_area` on `spec/fixtures/files/profile_fields.csv` seems to fix the failing specs. Just sharing it as diving into that helped me understand a bit more the way the import is done.

Now my doubts are related to what would be the expected outcome from the preparation process in the DB. After the setup I took a look in the console (my user's Profile data) and saw this:

Screen Shot 2020-11-03 at 14 04 57

So it seems like the fields were added correctly to the Profile's data `jsonb`. But after I edit some of my Profile's settings (only what's visible in `/settings`) I reload the Profile and see this:

Screen Shot 2020-11-03 at 14 07 03

My questions:
1. Is the expected outcome to preload all these fields into the Profile data field? If the user had empty fields it's okay that they come up empty strings (first screenshot), right?
1. If I go in `/admin/profile_fields` I don't see for example any of the Coding Profile fields. Are they supposed to be there or is this only meant to setup the `data` jsonb values?
   - This may as well be something I'm missing or overlooking regarding the data update script but I'm not too sure
1. I guess that since the Profile fields are not there that's why the fields in the `data` jsonb are being wiped out (second screenshot). Is this intentional?
   - Meaning that for example, if an Admin removed a Profile field then that data will be wiped out from the DB right (from jsonb hash)? I think this is good since I think it will free up memory, but wanted to ask to understand the expectation

citizen428:
Thanks for the detailed review/questions @fdoxyz!

>  everything seems to be working well

Good news, thanks for confirming!

> Side note: I took a look at the failing specs 

Thanks for this, you are right. As mentioned in the description, this PR was still quite rough, I just wanted feedback as early as possible. I pushed an update.

> After the setup I took a look in the console (my user's Profile data) and saw this [...]  But after I edit some of my Profile's settings

This is correct. The `ExtractData` service just takes all the user data and dumps it into the `data` JSONB column of the profile. Updates on the other hand go through the service object, which cleans out keys with `nil` values and keys that no longer exist as profile fields to save space. I added a `compact` call in the data update script now, which will remove keys with `nil` values to be more consistent.

---
To answer your questions:

> Is the expected outcome to preload all these fields into the Profile data field? If the user had empty fields it's okay that they come up empty strings (first screenshot), right?

It was expected but should no longer be the case, as the expectation was a bit unreasonable/inconsistent with the `Profiles::Update` service (which didn't exist when `Profiles::ExtractData` was first written).

> If I go in /admin/profile_fields I don't see for example any of the Coding Profile fields.  

Strange: All 29 profile fields defined in `lib/data/dev_profile_fields.csv` should get generated via CSV import. I just tried again locally, by removing all of them and running the script and they all came back.  I also can see all the coding fields (just to double check, you did enable the `profile_admin` featur flag, right?):

Screen Shot 2020-11-04 at 10 24 28

Could you try again please?

> I guess that since the Profile fields are not there that's why the fields in the data jsonb are being wiped out (second screenshot). Is this intentional? 

Yes, as explained above this behavior is intentional and handled via `Profiles::Update`:

```ruby
      # We don't update `data` directly. This uses the defined store_attributes
      # so we can make use of their typecasting.
      @profile.assign_attributes(@updated_profile_attributes)

      # Before saving, filter out obsolete profile fields
      @profile.data.slice!(*Profile.attributes)
```

Thanks again!

mstruve:
> If I go in /admin/profile_fields I don't see for example any of the Coding Profile fields.

I ran into the same issue but I didn't have the profile_admin feature flag enabled. I went to the admin UI, enabled `profile_admin`. 
Screen Shot 2020-11-04 at 5 41 07 PM

then I ran the script again and still not seeing those fields in admin
Screen Shot 2020-11-04 at 5 42 12 PM

Am I missing a step?

fdoxyz (approved):
Awesome thanks for the clarification @citizen428!

@mstruve I tried this again too and had the same problem. I think it has to do with the way the data update scripts are stored in the DB when they run. Even if you switch to another branch, when you come back to this branch and try to re-run the script it won't do anything (because the record in the DB will still be there)

Screen Shot 2020-11-04 at 16 31 15

So, the first time I reviewed the PR the update script (yesterday Nov 3rd) and no fields were added

Screen Shot 2020-11-04 at 16 25 15

Now after removing the data update script record manually (`DataUpdateScript.last.destroy`) and re-running the script I did get the expected result:

Screen Shot 2020-11-04 at 16 42 25

So, maybe when reviewing this one or any other of the PRs that used this script with the same name then it was actually not running.

This leads me to believe this could be a possible enhancement/feature request for the data update scripts? So they also store the environment/branch where they ran. Or maybe the data_updates.rake task could also remove all record of previously ran tasks that are not currently available in `lib/data_update_scripts/...` so if you switch branches then next time you're back the scripts will actually run 🤔

Anyways, from what I was able to test around the app after the script ran properly this LGTM. Maybe it would be good for someone with console access to double check that the script was actually ran properly when this is deployed.

mstruve:
I also deleted the DataUpdateScript record between `bin/setup` but still didn't see the fields. I just ran the import manually in a console and boom I see the new profile fields. I then deleted the script db record, added a puts statement to the script class, and ran `bin/setup` again and sure enough the puts statement printed so it's being run so everything appears to be working as it should minus the fact that the first time the script didn't populate the fields. idk, computers are weird sometimes. 

With that said, is `profile_admin` needed to ensure this runs properly on all envs?

maestromac (commented):
I ran `bin/setup` and couldn't see it via the UI either. 
```
[1] pry(main)> ProfileField.count
   (1.1ms)  SELECT COUNT(*) FROM "profile_fields"
=> 5
```

This could be unrelated: When I change back to master branch and did a `bin/rails db:reset` without running `bin/setup`, I would see the ProfileField generated by seed.rb appear on /admin/profield_field. If I were to run `bin/setup` it would erase those fields down to 5 fields. Is that expected?

citizen428:
Thanks @mac, I think this was an important clue, I have an updated to the seeds file in another branch that I should probably move over here.

@mstruve 

> With that said, is profile_admin needed to ensure this runs properly on all envs?

No. The feature flag enables/disables the routes for profile admin controllers and the corresponding views/links. It does not affect profile field creation through scripts, etc. and is not necessary for the site to use profile related features.

citizen428:
@maestromac 

> This could be unrelated: When I change back to master branch and did a bin/rails db:reset without running bin/setup, I would see a lot of the ProfileField generated by seed.rb appear on /admin/profield_field. If I were to run bin/setup it would erase those fields down to 5 fields. Is that expected?

Haha, yes, that's it I think:

From `bin/setup`:
```ruby
  puts "\n== Initializing the application =="
  system! "bin/rails app_initializer:setup forem:setup"
```

The `app_initializer:setup` task runs the data updates, but then `forem:setup` does the following:

```ruby
    # TODO: [@forem/oss] Remove the destroy_all call
    ProfileField.destroy_all
    ProfileFields::AddBaseFields.call
```

I guess it's time I resolve this TODO 😉

citizen428:
Can y'all give this another spin? To make sure the data update script will run, issue the following in a console first, just in case:

```ruby
DataUpdateScript.delete_by(file_name: "20201103050112_prepare_for_profile_column_drop"
```

mstruve:
Did you mean to push new changes to that `forem:setup` before we tried? Right now code is still the same so the rake task doesnt work as it should

citizen428:
@mstruve Yes, I did, clearly that migraine yesterday did not help 🤦 Sorry for the delay here...

maestromac (approved):
It's working now :+1:

mstruve (approved):
Worked like a charm for me!

citizen428:
Thanks for feeback everyone!

image

PR #11249, opened 2020-11-03
Comments inside article

ludwiczakpawel:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR changes a little the way we display articles:

### Comments section is now INSIDE article section

This will prepare the article view for some of the design updates that we're workin on for Comments section.

### Author section at the bottom of the article is now dropped

Same as above but it also not really needed since we display almost the same info in the right sidebar which is sticky.

image

## Added tests?

- [ ] Yes
- [x] No, not needed

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed


lisasy (approved):
LGTM!

citizen428 (approved):
I'm gonna trust you on this one.

ludwiczakpawel:
> I'm gonna trust you on this one.

@citizen428 I can smell the scepticism in your voice :D So here's a bit more explanation and reasoning behind this change:

1. We want _Discussion_ to be more seamless and integrated part of a post. Right now posts are mostly long format (esseys) but in future we gonna have more post formats like 
  - photo
  - gallery
  - short status (twitter-like)
  - code snippet
  - ... and much more
  Some of these will be pretty short and without any real text-heavy content like you can see currently. Therefore, to drive an engagement  we need to encourage people to start discussion (leave comments), react etc.. That's why we want "Discussion" section to feel more like a part of the post itself rather than a standalone section. The post itself is only meant to start a conversation and the conversation will hopefully happen in our "Discussion" section. And this is why this part needs to be integrated into post as "smoothly" and seamless as possible. When you visit a post you should feel almost like post author is inviting you to participate in conversation.
2. "Discucssion" section is the only part of the article view that is still a little bit "outdated" in terms of UI. Since the "Discucssion" section is a pretty huge beast in terms of code, I've decided to work on that part a bit later and ship it gradually (that's why for example couple weeks ago I've shipped the update for Comment form UI). Now, we still haven't decided on the direction re how comments will look like but all my explorations basically benefit (visually) from being in the same box as the rest of the post. Comments, by its nature, have a bit "boxy" design (you know - each comment is within its own box visually) so if we can reduce "boxy" design elsewhere it would definitely feel less overwhelming generally (more boxes == heavier design == more overwhelming).

^ are these two points with explanation make you feel a bit more comfortable with approving the PR? :)

Link2Twenty:
We do lose the post author reminder at the bottom of the post, but I guess all that info is in the side bar anyway.

ludwiczakpawel:
@Link2Twenty that is correct, we basically have almost the same information in the **sticky** sidebar.

benhalpern:
@ludwiczakpawel let's hold off on deploying this until the USA morning time. This should be safe in terms of caching issues, but I want to be around to monitor for any issues.

citizen428:
> @citizen428 I can smell the scepticism in your voice :D

Not at all, if you did it must have been poor phrasing on my part. Sorry for that. What I meant is: I don't have enough insight into your long term plans here (my fault for never attending a design crit), so I'll trust you know what you're doing here 😄 Thanks for the extra context though!

ludwiczakpawel:
> you know what you're doing here

well... I didn't say THAT 😂

PR #11252, opened 2020-11-03
 Replace old assets with new ones if cached page mismatch

benhalpern:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

If users access a page with an out-of-date service worker shell, or otherwise have "old" assets loaded, those assets may have been purged or not match up with new HTML. This attempts to reconcile these scenarios.

## Added tests?

- [ ] Yes
- [x] No, and this is why: _please replace this line with details on why tests
      have not been included_

This is a localized patch and somewhat experimental. I'm not sure how testable it is.

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

Documentation should be added once we confirm this works as expected.

alt_text


app/views/layouts/application.html.erb
```diff
@@ -21,6 +21,47 @@
...
+      var pageCacheCheck = document.querySelector("meta[name='page-cached-at']");
+      var headCrayonsPath = document.getElementById("main-crayons-stylesheet")
+      var headViewsPath = document.getElementById("main-views-stylesheet")
+      var headMinimalPath = document.getElementById("main-minimal-stylesheet")
+      var pageCrayonsPath = document.querySelector("meta[name='proper-stylesheet-crayons']");
+      var pageViewsPath = document.querySelector("meta[name='proper-stylesheet-views']");
+      var pageMinimalPath = document.querySelector("meta[name='proper-stylesheet-minimal']");
+
+      var headBaseJSPath = document.getElementById("main-base-javascript")
+      var pageBaseJSPath = document.querySelector("proper-js-base")
```
vaidehijoshi, line 49: I'm a little worried about hardcoding all of this (and then having to remember to change it whenever crayons change). Also thinking about the eventual "extract out crayons" path here, too. 🤔

vaidehijoshi (commented):
>  No, and this is why: _please replace this line with details on why tests
have not been included_

Can we add tests for this? I'm not sure the best way to QA it locally (please let me know how!)...but tests would also help us get a sense of how this is meant to work. I especially think they're important if we're going to hard code the query selectors, because those could change and we wouldn't have any way of catching it aside from paying attention during PR review.

app/views/layouts/application.html.erb
benhalpern: Because this is dealing with changing conditions based on caching, I'm not sure what a good way to test this might be. I'm all for answers here!

benhalpern: @vaidehijoshi I added a regression test, as well as moved this code to a partial to clean up the main view.

It still doesn't test all the logic, but ensures if we change something that shouldn't be changed, it will be caught in the test.

spec/requests/stories_index_spec.rb
```diff
@@ -487,6 +487,18 @@ def create_live_sponsor(org, tag)
         get "/t/#{tag.name}"
         expect(response.cookies["remember_user_token"]).to be nil
       end
+
+      it "has necessary asset reconciliation code" do
+        # Ensure code elements are available for fixing assets if necessary.
+        # app/views/layouts/_asset_reconciliation.html.erb
+        # Basic regression test to ensure we don't accidentally remove something we should not.
+
+        expect(response.body).to include('<meta name="head-cached-at"')
```
vaidehijoshi, line 496: It looks like the tests are failing because `response` is not defined; I think you're missing the actual `get` request here, which is why `response` can't be found :)

app/views/layouts/_asset_reconciliation.html.erb
```diff
@@ -0,0 +1,65 @@
...
+        link.rel = "stylesheet";
+        link.href = pageViewsPath.content;
+        head.appendChild(link);
+        var link = document.createElement("link");
+        link.type = "text/css";
+        link.rel = "stylesheet";
+        link.href = pageMinimalPath.content;
+        head.appendChild(link);
+
+        // Take a breath and then remove the old ones.
```
vaidehijoshi, line 55: Thanks for adding these comments!! I think they're super useful. What's the reason for "taking the breath" with `setTimeout` here? It's not immediately obvious to me, so it might be nice to include that in the comment since you've taken the time to document it all!

app/views/layouts/application.html.erb
```diff
@@ -21,6 +21,8 @@
         <%= @hero_html.html_safe %>
       </div>
     <% end %>
+    
+    <%= render "layouts/asset_reconciliation" %>
```
vaidehijoshi, line 25: 👍  I'm glad we're pulling this out of the main layout into a specific partial, a great change

app/views/layouts/_styles.html.erb
```diff
@@ -4,7 +4,8 @@
...
 
-
-<%= stylesheet_link_tag "crayons", media: "all" %>
-<%= stylesheet_link_tag "views", media: "all" %>
-<%= stylesheet_link_tag "minimal", media: "all" %>
+<div id="head-stylesheets">
+  <%= stylesheet_link_tag "crayons", media: "all", id: "main-crayons-stylesheet" %>
+  <%= stylesheet_link_tag "views", media: "all", id: "main-views-stylesheet" %>
+  <%= stylesheet_link_tag "minimal", media: "all", id: "main-minimal-stylesheet" %>
+</div>
```
benhalpern, line 11:
```suggestion
</div>

```

app/views/layouts/_asset_reconciliation.html.erb
```diff
@@ -0,0 +1,60 @@
+<meta name="page-cached-at" content="<%= Time.current.to_i %>">
+
+<meta name="proper-stylesheet-crayons" content="<%= stylesheet_url "crayons" %>">
+<meta name="proper-stylesheet-views" content="<%= stylesheet_url "views" %>">
+<meta name="proper-stylesheet-minimal" content="<%= stylesheet_url "minimal" %>">
```
benhalpern, line 5: These are meta tags which _store_ the location of the proper stylesheet URLs, but don't do anything unless we insert them with JS below.

app/views/layouts/_asset_reconciliation.html.erb
```diff
@@ -0,0 +1,60 @@
...
+  var headCacheCheck = document.querySelector("meta[name='head-cached-at']");
+  var pageCacheCheck = document.querySelector("meta[name='page-cached-at']");
+  var headCrayons = document.getElementById("main-crayons-stylesheet")
+  var headViews = document.getElementById("main-views-stylesheet")
+  var headMinimal = document.getElementById("main-minimal-stylesheet")
+  var pageCrayons = document.querySelector("meta[name='proper-stylesheet-crayons']");
+  var pageViews = document.querySelector("meta[name='proper-stylesheet-views']");
+  var pageMinimal = document.querySelector("meta[name='proper-stylesheet-minimal']");
+
+  if (headCacheCheck && headCrayons && // If these elements exist, we can proceed.
```
benhalpern, line 24: If the page is more current than the head _and_ we have any assets which don't match up with the ones in the head, we add new stylesheets and get rid of the old ones.

app/views/layouts/_styles.html.erb
```diff
@@ -4,7 +4,6 @@
 <%# But if we in-line everything we lose the ability to cache assets for repeat visitors %>
 <%# This is a production test to see how this might affect UX and/or SEO %>
 
-
-<%= stylesheet_link_tag "crayons", media: "all" %>
-<%= stylesheet_link_tag "views", media: "all" %>
-<%= stylesheet_link_tag "minimal", media: "all" %>
+<%= stylesheet_link_tag "crayons", media: "all", id: "main-crayons-stylesheet" %>
```
benhalpern, line 7: Just adding IDs here

benhalpern:
Needing to merge this to deal with production issues.

PR #11253, opened 2020-11-03
Improve Containerfile CMD readability

PeterDaveHello:
### What type of PR is this? (check all applicable)

- [x] Refactor

## Description

The Containerfile CMD instruction needs some spaces ;)
(Or maybe it doesn't! ... but it has some spaces at the beginning! Consistency is important!)

## Added tests?

- [x] No, and this is why: Unless I need to add Docker build to CI system, I don't think it's something need to be revised with test cases.

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -59,4 +59,4 @@ VOLUME "${APP_HOME}"/public/
 
 ENTRYPOINT ["./scripts/entrypoint.sh"]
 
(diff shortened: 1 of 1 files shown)


rhymes (approved):
Thanks @PeterDaveHello!

PR #11256, opened 2020-11-03
Show canonical URL on article even after it has been updated

### LINKED ISSUE #10028: Canonical URL Missing Originally Publish At X
When I change the canonical url on an existing article. I cannot see the Originally published at `xxx`. I added the `canonical_url` option in the frontmatter. 

[Here](https://dev.to/hmajid2301/how-to-add-offline-search-to-a-gatsby-blog-4bnm
) is an article which was originally uploaded with a `canonical_url` option in the frontmatter.

[Here](https://dev.to/hmajid2301/react-native-expo-with-virtualbox-and-genymotion-3d7c) is an example of an article where the `canonical_url` has been added to the frontmatter. However it is set correctly in the `<head>` i.e.
`<link rel="canonical" href="https://haseebmajid.dev/blog/react-native-with-virtualbox">`

**To Reproduce**

1. Go to an article without a canonical_url in the frontmatter
2. Press `Edit` button
3. Add `canonical_url` to frontmatter
4. Press `save changes` button

**Expected behavior**

I would expect the article to update with a ` Originally published at <sitename.com>`

**Screenshots**

- OS, version: Linux POP_!OS 20.04
- Browser, version: Firefox, 79

meisekimiu:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
The canonical URL link on articles doesn't show up if the article has been updated, with the "Originally published at" text being replaced with the "Updated on" text instead. I'm guessing this behavior wasn't intended since the related issue is marked as a bug.

This fix shows both canonical url and update date at the same time.

Fixes #10028

To test, either add a canonical URL to a preexisting article that is at least one day old, or update an article that is at least one day old that already has a canonical URL.

## Added tests?

- [x] Yes
- [ ] No, and this is why: 

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

Non-canonical!

### DIFF
--- a/app/views/articles/show.html.erb
+++ b/app/views/articles/show.html.erb
@@ -159,7 +159,7 @@
                   <em>with <%= @article.co_author_name_and_path.html_safe %></em>
                 <% end %>
 
-    <% if should_show_crossposted_on?(@article) %>
+                <% if should_show_crossposted_on?(@article) %>
                   <em>
                     Originally published at
                     <a href="<%= @article.canonical_url || @article.feed_source_url %>" style="color:#1395b8"><%= get_host_without_www(@article.canonical_url || @article.feed_source_url) %></a>
@@ -167,11 +167,9 @@
                       on
                       <time datetime="<%= (@article.originally_published_at || @article.published_at)&.utc&.iso8601 %>"><%= (@article.originally_published_at || @article.published_at)&.strftime("%b %d, %Y") %></time>
                     <% end %>
-                    <% if should_show_updated_on?(@article) %>
-                      <em>(Updated on <time datetime="<%= @article.edited_at&.utc&.iso8601 %>"><%= @article.edited_at&.strftime("%b %d, %Y") %></time>)</em>
-                    <% end %>
                   </em>
-                <% elsif should_show_updated_on?(@article) %>
+                <% end %>
+                <% if should_show_updated_on?(@article) %>
                   <em>Updated on <time datetime="<%= @article.edited_at&.utc&.iso8601 %>"><%= @article.edited_at&.strftime("%b %d, %Y") %></time></em>
                 <% end %>
 
--- a/app/views/articles/show.html.erb
+++ b/app/views/articles/show.html.erb
@@ -159,9 +159,7 @@
                   <em>with <%= @article.co_author_name_and_path.html_safe %></em>
                 <% end %>
 
-                <% if should_show_updated_on?(@article) %>
-                  <em>Updated on <time datetime="<%= @article.edited_at&.utc&.iso8601 %>"><%= @article.edited_at&.strftime("%b %d, %Y") %></time></em>
-                <% elsif should_show_crossposted_on?(@article) %>
+    <% if should_show_crossposted_on?(@article) %>
                   <em>
                     Originally published at
                     <a href="<%= @article.canonical_url || @article.feed_source_url %>" style="color:#1395b8"><%= get_host_without_www(@article.canonical_url || @article.feed_source_url) %></a>
@@ -169,7 +167,12 @@
                       on
                       <time datetime="<%= (@article.originally_published_at || @article.published_at)&.utc&.iso8601 %>"><%= (@article.originally_published_at || @article.published_at)&.strftime("%b %d, %Y") %></time>
                     <% end %>
+                    <% if should_show_updated_on?(@article) %>
+                      <em>(Updated on <time datetime="<%= @article.edited_at&.utc&.iso8601 %>"><%= @article.edited_at&.strftime("%b %d, %Y") %></time>)</em>
+                    <% end %>
                   </em>
+                <% elsif should_show_updated_on?(@article) %>
+                  <em>Updated on <time datetime="<%= @article.edited_at&.utc&.iso8601 %>"><%= @article.edited_at&.strftime("%b %d, %Y") %></time></em>
                 <% end %>
 
                 <span class="mr-4">・<%= @article.reading_time < 1 ? 1 : @article.reading_time %> min read</span>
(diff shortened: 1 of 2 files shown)


app/views/articles/show.html.erb
```diff
@@ -159,17 +159,20 @@
...
-                <% elsif should_show_crossposted_on?(@article) %>
+    <% if should_show_crossposted_on?(@article) %>
                   <em>
                     Originally published at
                     <a href="<%= @article.canonical_url || @article.feed_source_url %>" style="color:#1395b8"><%= get_host_without_www(@article.canonical_url || @article.feed_source_url) %></a>
                     <% if @article.crossposted_at %>
                       on
                       <time datetime="<%= (@article.originally_published_at || @article.published_at)&.utc&.iso8601 %>"><%= (@article.originally_published_at || @article.published_at)&.strftime("%b %d, %Y") %></time>
                     <% end %>
+                    <% if should_show_updated_on?(@article) %>
```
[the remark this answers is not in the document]
meisekimiu: That's a good point! I actually noticed that right as I was going to commit the initial changes, thinking at the very least the "Updated At" text should be extracted out instead of duplicating the code, but I got distracted by some issues I had with the pre-commit hooks (which I also think is why the formatting got all messed up on that one line... I'm not sure how that happened otherwise since the indentation looked fine in my editor at the time!)

One reason why I didn't go in this direction from the beginning is because I think it looks a bit weird to *just* show both crosspost and update text without any kind of separator between them, and in the original commit I chose parenthesis to separate them. For now to keep things simple I removed the parenthesis in my newest commit:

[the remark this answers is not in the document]
citizen428: You could extract it into a helper that does something like the following:

```ruby
originally_published_at = ... if ...
updated_at = ...if...

[originally_published_at, updated_at].compact.join(" | ")
```

This only adds the separator when both are non-nil:

```ruby
["something", nil].compact.join(" | ")
#=> "something"
[nil, "somethingelse"].compact.join(" | ")
#=> "somethingelse"
["something", "somethingelse"].compact.join(" | ")
#=> "something | somethingelse
```

spec/views/articles_spec.rb
```diff
@@ -54,4 +54,14 @@ def create_comment(parent_id = nil)
     expect(rendered).to have_text("Originally published at")
     expect(rendered).to have_text("example.com")
   end
+
+  it "shows a note about the canonical URL after edit" do
+    allow(article1).to receive(:canonical_url).and_return("https://example.com/lamas")
+    allow(article1).to receive(:published_at).and_return(DateTime.strptime("1", "%s"))
```
rhymes, line 60: I think you can use `Time.zone.at(0)` here if you want to use the "beginning of time"

spec/views/articles_spec.rb
```diff
@@ -54,4 +54,14 @@ def create_comment(parent_id = nil)
...
+    allow(article1).to receive(:edited_at).and_return(DateTime.now)
```
rhymes, line 61: You should use `Time.current` instead of `DateTime.now`, check [this guide](http://danilenko.org/2012/7/6/rails_timezones/) on Rails and time

app/views/articles/show.html.erb
```diff
@@ -171,6 +169,9 @@
                     <% end %>
                   </em>
                 <% end %>
+                <% if should_show_updated_on?(@article) %>
+                  <em>Updated on <time datetime="<%= @article.edited_at&.utc&.iso8601 %>"><%= @article.edited_at&.strftime("%b %d, %Y") %></time></em>
```
rhymes, line 173: Maybe we can replace this with the `local_date` helper: https://github.com/forem/forem/blob/master/app/helpers/date_helper.rb

rhymes (changes requested):
Thanks @meisekimiu, aside from the notes inline I think the UI needs some of those `・` chars to separate the pieces, this is how I see it right now:

Screenshot_2020-11-06  A Many-Splendoured Thing Soluta necessitatibus

All of those sections should be separated? cc @ludwiczakpawel

meisekimiu:
I added the `・` character as a divider and made the requested changes. I didn't extract out any of the view code into a helper function since I thought it'd make the most sense to keep the view code actually inside the template instead of extracting it out into a helper, especially all for just one extra divider symbol.

app/views/articles/show.html.erb
```diff
@@ -159,17 +159,21 @@
...
                   <em>
                     Originally published at
                     <a href="<%= @article.canonical_url || @article.feed_source_url %>" style="color:#1395b8"><%= get_host_without_www(@article.canonical_url || @article.feed_source_url) %></a>
                     <% if @article.crossposted_at %>
                       on
-                      <time datetime="<%= (@article.originally_published_at || @article.published_at)&.utc&.iso8601 %>"><%= (@article.originally_published_at || @article.published_at)&.strftime("%b %d, %Y") %></time>
+                      <%= local_date(@article.originally_published_at || @article.published_at, show_year: (@article.originally_published_at || @article.published_at).year != Time.current.year) %>
                     <% end %>
                   </em>
+                  <% if should_show_updated_on?(@article) %>
```
citizen428, line 171: Could you just move this into the `if` a couple of lines below?

```ruby
<% if should_show_updated_on?(@article) %>
  ・<em>Updated on <%= local_date(@article.edited_at, show_year: @article.edited_at.year != Time.current.year) %></em>
<% end %>
```

citizen428 (approved):

rhymes (approved):
Looks good! Thanks!

PR #11257, opened 2020-11-03
Update organization <select> to use Crayons

Zhao-Andy:
- [x] Bug Fix

## Description
This updates the organization settings dropdown to something much nicer. I'll let @forem/design have the final say but figured this would be a quick and easy update. :)

 
Screen Shot 2020-11-03 at 3 34 18 PM

### DIFF
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -22,7 +22,7 @@
 <header class="crayons-layout crayons-layout--limited-l crayons-layout--1-col p-4">
   <% if @tab == "organization" && @organizations.present? %>
     <h1 class="crayons-title">Settings for
-      <select class="crayons-select lh-tight w-25 align-text-top" onchange="window.location.href=this.value">
+      <select class="crayons-select lh-tight w-25 align-text-top w-max" onchange="window.location.href=this.value">
         <% options = @organizations.pluck(:slug, :id).each do |arr| %>
           <% arr[0] = "@#{arr[0]}" %>
           <% arr[1] = "/settings/organization/#{arr[1]}" %>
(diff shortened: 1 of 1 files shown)


app/views/users/edit.html.erb
```diff
@@ -22,18 +22,16 @@
...
-            <% arr[0] = "@#{arr[0]}" %>
-            <% arr[1] = "/settings/organization/#{arr[1]}" %>
-          <% end.to_h.merge("New Organization" => "/settings/organization/new") %>
-          <%= options_for_select(options, "/settings/organization/#{params[:org_id]}") %>
-          <%# For example: %>
-          <%# <option value="/settings/organization/1>@devteam</option> %>
-          <%# <option value="/settings/organization/new" selected>New Organization</option> %>
-        </select>
-      </span>
+      <select class="crayons-select lh-tight w-25 align-text-top" onchange="window.location.href=this.value">
```
Zhao-Andy, line 25: TLDR; I removed a `<span>` that wrapped the `<select>` and added new utility classes.

lisasy, line 25:
```suggestion
      <select class="crayons-select lh-tight w-25 align-text-top w-max" onchange="window.location.href=this.value">
```

Adding `w-max` to the `select` element will show the entire string, especially on a smaller screen.

### Before (without `w-max`)
Screen Shot 2020-11-03 at 1 42 02 PM

### After (with `w-max`)
Screen Shot 2020-11-03 at 1 41 50 PM

lisasy (approved):
If you add `w-max` to the select, it will render better, especially on small screens. Otherwise this looks great.

rhymes (approved):

PR #11260, opened 2020-11-04
Fixing date in listing modal

Rafi993:
##  What type of PR is this? 

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When listing modal is refreshed the date does not appear this PR fixes that.

Closes https://github.com/forem/forem/issues/11261

1. Go to /listings
2. Click on a listing with date. It will open up a modal and it will show the date
3. Now refresh the page
4. The modal will show up but with date

## Added tests?

- [ ] Yes
      have not been included_

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -3,7 +3,8 @@ class ListingsController < ApplicationController
 
   INDEX_JSON_OPTIONS = {
     only: %i[
-      title processed_html tag_list category id user_id slug contact_via_connect location
+      title processed_html tag_list category id user_id slug contact_via_connect location bumped_at
+      originally_published_at
     ],
     include: {
       author: { only: %i[username name], methods: %i[username profile_image_90] },
--- a/spec/requests/listings_spec.rb
+++ b/spec/requests/listings_spec.rb
@@ -46,7 +46,7 @@ RSpec.describe "/listings", type: :request do
       expect(response).to have_http_status(:ok)
     end
 
-    it "has correct keys in the response" do
+    it "listings have correct keys" do
       get "/listings"
 
       parsed_response = Nokogiri.HTML(response.body)
--- a/spec/requests/listings_spec.rb
+++ b/spec/requests/listings_spec.rb
@@ -1,4 +1,5 @@
 require "rails_helper"
+require "nokogiri"
 
 RSpec.describe "/listings", type: :request do
   let(:user) { create(:user) }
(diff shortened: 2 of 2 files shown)


citizen428 (approved):

rhymes (approved):
Thanks a lot @Rafi993!

rhymes:
@Rafi993 I wonder if we can seal this with a test to avoid regression

Rafi993:
> @Rafi993 I wonder if we can seal this with a test to avoid regression

Yes it would be awesome too. Is there any similar test that I can use for reference ?

rhymes:
@Rafi993 I think we can add a test entry to https://github.com/forem/forem/blob/8d08191e9a455225438869e027f989cbd07e1e5c/spec/requests/listings_spec.rb#L31 similar to https://github.com/forem/forem/blob/8d08191e9a455225438869e027f989cbd07e1e5c/spec/requests/api/v0/articles_spec.rb#L24-L36

does it make sense?

Rafi993:
> @Rafi993 I think we can add a test entry to
> 
> https://github.com/forem/forem/blob/8d08191e9a455225438869e027f989cbd07e1e5c/spec/requests/listings_spec.rb#L31
> 
> similar to
> https://github.com/forem/forem/blob/8d08191e9a455225438869e027f989cbd07e1e5c/spec/requests/api/v0/articles_spec.rb#L24-L36
> 
> does it make sense?

There is one issue though. Unlike the above sample it does not return json but rather returns rendered response as html (after passing it to views). Should I check the selectors in the rendered html is that ok?

Rafi993:
> @Rafi993 I think we can add a test entry to
> 
> https://github.com/forem/forem/blob/8d08191e9a455225438869e027f989cbd07e1e5c/spec/requests/listings_spec.rb#L31
> 
> similar to
> https://github.com/forem/forem/blob/8d08191e9a455225438869e027f989cbd07e1e5c/spec/requests/api/v0/articles_spec.rb#L24-L36
> 
> does it make sense?

There is one issue here. **/listings** returns html (unlike json in the sample).  But luckily the data was stored in **data-listings** attribute.  so I tried the following

```ryby
    it "has correct keys in the response" do
      get "/listings"

      parsed_response = Nokogiri.HTML(response.body)
      listings = parsed_response.xpath("//*[@id='listings-index-container']")["data-displayedlisting"]

      index_keys = %w[
        title processed_html tag_list category id user_id slug contact_via_connect location bumped_at
        originally_published_at
      ]

      expect(listings.first.keys).to match_array index_keys
    end
```

**data-displayedlisting** is empty in the test while it has value when viewing the actual dom let me fix that too so that the test can pass.

rhymes (approved):
Thanks for adding the test! :-)

PR #11263, opened 2020-11-04
Refactor DataUpdateScript to make its code simpler

turara:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Remove unnecessary where method and Pathname initialization.

About where method, it was originally introduced to select ids of enqueued scripts, so it's no longer needed with enqueued method.

## Added tests?

- [ ] Yes
- [x] No, and this is why: because methods changed already have tests or are private.

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/models/data_update_script.rb
+++ b/app/models/data_update_script.rb
@@ -29,7 +29,7 @@ class DataUpdateScript < ApplicationRecord
 
     def filenames
       Dir.glob("*.rb", base: DIRECTORY).map do |f|
-        File.basename(".rb")
+        File.basename(f, ".rb")
       end
     end
 
--- a/app/models/data_update_script.rb
+++ b/app/models/data_update_script.rb
@@ -12,7 +12,7 @@ class DataUpdateScript < ApplicationRecord
     def scripts_to_run
       insert_new_scripts
 
-      enqueued.where(id: ids).order(file_name: :asc)
+      enqueued.order(file_name: :asc)
     end
 
     # true if there are more files on disk or any scripts to run, false otherwise
@@ -29,7 +29,7 @@ class DataUpdateScript < ApplicationRecord
 
     def filenames
       Dir.glob("*.rb", base: DIRECTORY).map do |f|
-        Pathname.new(f).basename(".rb").to_s
+        File.basename(".rb")
       end
     end
 


rhymes:
Hi @turara, would you be willing to explain the PR? It's unclear to me what problem you're trying to tackle. Thank you!

turara:
@rhymes Thank you for your comment!

I was trying to improve inefficient code and make them simpler. Is it not recommended to make a PR for that? If so, it's my bad.

app/models/data_update_script.rb
```diff
@@ -12,7 +12,7 @@ class << self
     def scripts_to_run
       insert_new_scripts
 
-      enqueued.where(id: ids).order(file_name: :asc)
```
mstruve, line 15: Ummm, I have no idea why we did it this way but great catch!!!

app/models/data_update_script.rb
```diff
@@ -29,7 +29,7 @@ def scripts_to_run?
 
     def filenames
       Dir.glob("*.rb", base: DIRECTORY).map do |f|
-        Pathname.new(f).basename(".rb").to_s
+        File.basename(f, ".rb")
```
mstruve, line 32: According to my testing, this looks to return the same thing but is a bit simpler to read.

mstruve (approved):
Welcome to the Forem project @turara! 

Not sure how we let that `ids` issue slip through the cracks but really nice catch on that one. Good work here! Thanks for the PR! 

One small bit of feedback I have is you might want to try making the title a little more descriptive next time so it gives a hint as to what the PR is trying to accomplish. `Update data_update_script.rb` is just a tad vague 😊

app/models/data_update_script.rb
turara: Git blame told me that `ids` was originally meant to be load_script_ids below :)

```ruby
    def load_script_ids
      filenames.
        map { |file_name| find_or_create_by(file_name: file_name) }.
        select(&:enqueued?).
        map(&:id)
    end
```

turara:
@mstruve Thank you for your review and feedback! I changed the title a bit.

rhymes (approved):
Thank you @turara!

PR #11264, opened 2020-11-04
Container: Update to Ruby 2.7.2

jdoss:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Updates the container to Fedora 33 and Ruby 2.7.2

I have built it locally and it works.
We should let this container build once it is merged to master and then deploy it to forem.dev for testing.

https://twitter.com/i/status/1323428115943620608


rhymes (changes requested):
Changing `.ruby-version` won't just update the container, it will also cause a mismatch in the non containerized version AFAIK.

We need to upgrade Ruby all at once:

- `.github/workflows/dependabot.yml`
- `.travis.yml`
- `docs/.ruby-version`

Also need to regenerate the `Gemfile.lock` by running `bundle install`

citizen428:
@rhymes I have a branch for the actual update. @jdoss Maybe reduce this one to only the `Containerfile`, then we merge and I rebase my update branch.

citizen428:
Actually, I just double checked, I already have that change in my branch, because I was waiting for it. So if you don't mind I just close this one @jdoss!

jdoss:
Just grab my change to the container file @citizen428 and I'll close this.

PR #11268, opened 2020-11-04
Exclude code highlight panel markup from RSS description 

### LINKED ISSUE #10927: Don't include "Enter fullscreen mode Exit Fullscreen mode" for code blocks in RSS feed
**Describe the bug**

RSS feed versions of dev.to articles that include syntax-highlighted code (e.g.,  `<pre><code>...</code></pre>`) also include the following text after the code:

```

    Enter fullscreen mode
    


    Exit fullscreen mode
    


```

**To Reproduce**

1. Consider a post like https://dev.to/phast184/remotes-commits-and-merges-is-that-a-long-distance-relationship-bfo
2. The RSS feed at https://dev.to/feed/phast184 includes the extra UI element `<div class="highlight__panel-action js-fullscreen-code-action">`

**Expected behavior**

Before publishing the feed, it would be great if this UI `<div>` could be stripped out.  It doesn't make sense in the syndicated article.

query-string:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Based on #10927 requirements this PR introduces a new Article method that sanitizes `processed_html` output with Nokogiri

Closes #10927 

Open http://localhost:3000/feed/username, resulting XML markup doesn't have `highlight__panel` tags anymore

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -386,6 +386,11 @@ class Article < ApplicationRecord
     self.co_author_ids = list_of_co_author_ids.split(",").map(&:strip)
   end
 
+  def plain_html
+    doc = Nokogiri::HTML.fragment(processed_html)
+    doc.search("highlight__panel").each(&:remove).to_html
+  end
+
   private
 
   def search_score
--- a/app/views/articles/feed.rss.builder
+++ b/app/views/articles/feed.rss.builder
@@ -24,7 +24,7 @@ xml.rss version: "2.0" do
         xml.pubDate article.published_at.to_s(:rfc822) if article.published_at
         xml.link app_url(article.path)
         xml.guid app_url(article.path)
-        xml.description sanitize(article.processed_html, tags: allowed_tags, attributes: allowed_attributes)
+        xml.description sanitize(article.plain_html, tags: allowed_tags, attributes: allowed_attributes)
         article.tag_list.each do |tag_name|
           xml.category tag_name
         end
--- a/spec/models/article_spec.rb
+++ b/spec/models/article_spec.rb
@@ -980,4 +980,10 @@ RSpec.describe Article, type: :model do
       expect(article.co_author_ids).to match_array([co_author1.id, co_author2.id])
     end
   end
+
+  describe "#plain_html" do
+    it "doesn't include js panel markup" do
+      expect(article.plain_html).not_to include("Enter fullscreen mode")
+    end
+  end
 end


msarit:
Hey @query-string! 👋🏾 

Thanks for this PR; could you please resolve the conflicts? I will tag our devs to review it soon. 🙏🏾

msarit:
Hey @query-string! 👋🏾
Just checking to see if you're still working on this PR? Let us know if you need any help or have questions. 🙏🏾

query-string:
Hey @msarit, sorry for holding on for so long, will take a look ASAP

citizen428:
Hi @query-string, did you get a chance to look at this in the meantime? 😃

query-string:
@msarit @citizen428 ✔️

citizen428 (approved):

rhymes (commented):
Hi @query-string, thanks for the change!

It feels a bit odd to have `plain_html` as a model method, after all it's very specific operation we're performing here. Even the test below described as "doesn't include js panel markup" is an indication this logic should be sitting elsewhere. What if we keep it closer to the RSS/feed machinery? 

I propose two solutions (in order of complexity):

- moving `plain_html` to `app/helpers/articles_helper.rb` (plus the test as well)
- creating a `rss_feed_helper` containing a `description` method which calls somethingl ike`sanitize(plain_html(article), tags: allowed_tags, attributes: allowed_attributes)` (after all, what we're populating is the `description` field of the RSS item)

What do you think?

citizen428:
@rhymes Interesting point. I actually quite like the symmetry of `processed_html` and `plain_html` living right next to each other, but I can relate to your reasoning too. 🤔

rhymes:
@citizen428 `.processed_html` is a byproduct of `Article.evaluate_markdown` which IMHO should live elsewhere but that's part of a larger conversation around services vs model methods.

I'm okay with merging this as there's nothing inherently wrong with the feature, maybe we can revisit this conversation within the context of a larger refactory?

citizen428:
@rhymes 

> `.processed_html` is a byproduct of `Article.evaluate_markdown`

I understand that. I think unless we move both as part of a larger refactor, it's better to keep these related things together. `User` and `Article` are the two models most desperately in need of being broken down into smaller objects, but until that happens I prefer keeping related things close to each other. 

One of my favorite ways of seeing potential extractions is using inline concerns, e.g. 

```ruby
concerning :MarkdownParsing do
  class_methods do
    def evaluate_markdown 
      ...
    end
  end

  def processed_html
    ...
  end
end
```

rhymes (approved):
LGTM!

Thanks @query-string!

PR #11277, opened 2020-11-04
Comment out Flaky Onboarding spec

mstruve:
This Spec appears to be failing [on the master branch](https://travis-ci.com/github/forem/forem/jobs/428276212) and locally for me. It appears that [this spec was just added](https://github.com/forem/forem/pull/11259) but the initial build was green. I am not sure why it is causing issues now but this will comment it out until we get a chance to fix it.

PR #11259, opened 2020-11-03
Conditionally hide onboarding task card

vaidehijoshi:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As a part of the ✨ NEW CREATOR ONBOARDING FLOW ✨, we need to change how we deal with the onboarding code. There is a lot to untangle, since we currently force _all_ users to go through that flow.

**There will be more PRs that touch onboarding-related code, but in an attempt to make these PRs easy to review + QA, I am attempting to keep them small and encapsulated around incremental changes.**

This PR conditionally renders the onboarding task card by checking that a user `saw_onboarding` before rendering the card. On a practical level, this check doesn't do anything for _current Forems_, because every single user is going to be _forced_ through onboarding. So really, this check is redundant, currently.

HOWEVER! I have more code in the pipeline that assumes that Forem creators will _not_ be going through the "usual onboarding flow", located at `/onboarding`. To that end, we need to change the logic where we assume that every user will have seen onboarding -- for Forem creators, `saw_onboarding` will be `false`, as they will have their own concept of "creator onboarding".

Specifically, this PR:

- [x] Checks that the `current_user.saw_onboarding` before rendering the task card.
- [x] Improves upon and adds new tests around our onboarding flow (which I will use in future PRs when I add specs around the creator onboarding flow).
- [x] Removes a bunch of redundant `saw_onboarding: true` checks from test `user`s in our `/spec` files.
- [x] Fixes some typos + capitalization issues that I found in related files.

## Related Tickets & Documents

A step towards completing https://github.com/forem/InternalProjectPlanning/issues/81.

## QA Instructions, Screenshots, Recordings

**To QA this, please ensure that the onboarding flow continues to work as it _currently does_.**

Here's what you'll want to test:
1. Go through the sign up + sign in process for your local Forem instance. Ensure that you see the onboarding task card, and that you can close it:
Screen_Shot_2020-11-03_at_3_31_14_PM
2. When you close the task card, it should not appear again:
Screen Shot 2020-11-03 at 3 20 05 PM
3. Delete the `task-card-closed` key from your local storage: 
Screen_Shot_2020-11-03_at_3_20_34_PM
4. Ensure that the task card renders again 😸 
Screen Shot 2020-11-03 at 3 19 55 PM

## Added tests?

[...]

PR #11279, opened 2020-11-05
Refactor Completing Onboarding spec for flakiness

vaidehijoshi:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In https://github.com/forem/forem/pull/11259, I added some specs that later on, caused flakiness in some CI builds.

@mstruve noticed these and was able to repro them locally (which, confusingly, never was an issue for me on the original PR 🤔), and went ahead and commented them out temporarily in https://github.com/forem/forem/pull/11277.

I did some investigating, and wasn't actually able to find a specific _line_ or issue that was causing these flaky specs. However, noticed some things that I missed in the spec the first time around that could be refactored. Notably, I also realized that I never mocked the `SiteConfig` variables, `allow_email_password_registration` + `allow_email_password_login`. I'm not _entirely_ sure why the spec would have failed in the way it did in when it became flaky, but it seems like adding these mocks in and refactoring the file has resulted in the test file passing consistently for me when running it locally 🤷 

Fixes commented out flaky specs in https://github.com/forem/forem/pull/11277.

Check out this branch, and run `rspec ./spec/system/onboardings/user_completes_onboarding_spec.rb` locally. You should be able to re-run it a few times and it should continue to pass.

_If it doesn't pass for you, please let me know! It's hard to QA and fix flaky specs as they are not consistently reproducible. 😕_

## Added tests?

- [ ] Yes
- [x] No, and this is why: _this refactors a preexisting spec file, and doesn't add any other, new code_

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

### DIFF
--- a/spec/system/onboardings/user_completes_onboarding_spec.rb
+++ b/spec/system/onboardings/user_completes_onboarding_spec.rb
@@ -2,19 +2,21 @@ require "rails_helper"
 
 RSpec.describe "Completing Onboarding", type: :system, js: true do
   let(:password) { Faker::Internet.password(min_length: 8) }
-  let(:user) { create(:user, password: password, password_confirmation: password, saw_onboarding: false) }
 
-  after do
-    sign_out user
+  before do
+    allow(SiteConfig).to receive(:allow_email_password_registration).and_return(true)
+    allow(SiteConfig).to receive(:allow_email_password_login).and_return(true)
   end
 
   context "when the user hasn't seen onboarding" do
+    let(:user) { create(:user, password: password, password_confirmation: password, saw_onboarding: false) }
+
     before do
       visit sign_up_path
       log_in_user(user)
     end
 
-    xit "logs in and redirects to onboarding if it hasn't been seen" do
+    it "logs in and redirects to onboarding if it hasn't been seen" do
       expect(page).to have_current_path("/onboarding", ignore_query: true)
       expect(page.html).to include("onboarding-container")
     end
@@ -28,26 +30,19 @@ RSpec.describe "Completing Onboarding", type: :system, js: true do
   end
 
   context "when the user has seen onboarding" do
-    before do
-      user.update(saw_onboarding: true)
+    let(:user) { create(:user, password: password, password_confirmation: password) }
 
+    before do
       visit sign_up_path
       log_in_user(user)
     end
 
-    xit "logs in and renders the feed" do
+    it "logs in and renders the feed" do
       expect(page).to have_current_path("/?signin=true")
       expect(page.html).not_to include("onboarding-container")
     end
 
-    it "renders the feed and onboarding task card" do
-      visit "/"
-
-      wait_for_javascript
-      expect(page).to have_css(".onboarding-task-card")
-    end
-
-    it "can dismiss the onboarding task card" do
+    it "renders and can dismiss the onboarding task card" do
       visit "/"
 
       wait_for_javascript
--- a/spec/system/onboardings/user_completes_onboarding_spec.rb
+++ b/spec/system/onboardings/user_completes_onboarding_spec.rb
@@ -30,7 +30,7 @@ RSpec.describe "Completing Onboarding", type: :system, js: true do
   end
 
   context "when the user has seen onboarding" do
-    let(:user) { create(:user, password: password, password_confirmation: password) }
+    let(:user) { create(:user, password: password, password_confirmation: password, saw_onboarding: true) }
 
     before do
       visit sign_up_path


spec/system/onboardings/user_completes_onboarding_spec.rb
```diff
@@ -2,19 +2,21 @@
 
 RSpec.describe "Completing Onboarding", type: :system, js: true do
   let(:password) { Faker::Internet.password(min_length: 8) }
-  let(:user) { create(:user, password: password, password_confirmation: password, saw_onboarding: false) }
 
-  after do
-    sign_out user
+  before do
+    allow(SiteConfig).to receive(:allow_email_password_registration).and_return(true)
+    allow(SiteConfig).to receive(:allow_email_password_login).and_return(true)
```
vaidehijoshi, line 8: Since this spec only uses email/password for logging in, we need to mock these. I suspect this was causing the flakiness but I really don't know for sure.

spec/system/onboardings/user_completes_onboarding_spec.rb
```diff
@@ -2,19 +2,21 @@
 
 RSpec.describe "Completing Onboarding", type: :system, js: true do
   let(:password) { Faker::Internet.password(min_length: 8) }
-  let(:user) { create(:user, password: password, password_confirmation: password, saw_onboarding: false) }
 
-  after do
-    sign_out user
```
vaidehijoshi, line 8: I realized this isn't actually necessary for this spec's purposes (and wasn't sure if the sign in/out process was causing the pundit error), so I went ahead and just removed it.

spec/system/onboardings/user_completes_onboarding_spec.rb
```diff
@@ -28,26 +30,19 @@
   end
 
   context "when the user has seen onboarding" do
-    before do
-      user.update(saw_onboarding: true)
```
vaidehijoshi, line 32: Rather than updating the user's `saw_onboarding` attribute, I just created a new one for this context.

spec/system/onboardings/user_completes_onboarding_spec.rb
```diff
@@ -28,26 +30,19 @@
...
+    let(:user) { create(:user, password: password, password_confirmation: password) }
```
mstruve, line 33: I think it might be a good idea to include `saw_onboarding: true` here instead of relying on the default. That way if the default changes this spec doesn't surprisingly fail.

mstruve (commented):
I noticed that for this spec you are using `log_in_user` when I searched the codebase for that I saw that it's only used in one other spec `spec/system/authentication/user_logs_in_with_email_spec.rb`. Everywhere else we use `sign_in(user)` and the `sign_in` [method we actually had to patch in order to ensure that a user would authenticate properly before the next request was sent](https://github.com/forem/forem/blob/master/spec/support/initializers/warden.rb) bc I [found a race condition which was causing all of our flaky Pundit errors before](https://github.com/forem/forem/pull/8379). 

Any reason not to switch to using the warden helper that we know is reliable when it comes to signing in?

spec/system/onboardings/user_completes_onboarding_spec.rb
vaidehijoshi: @mstruve Hm, I actually think we should explicitly _not_ have to include `saw_onboarding: true` here. In fact, in https://github.com/forem/forem/pull/11259, I explicitly removed all the instances in our specs where we were [explicitly setting `saw_onboarding` to `true`](https://github.com/forem/forem/pull/11259#discussion_r517022340) 

The reason behind this is that I think that if someone goes into the user factory and _does_ change the default from `true` to `false`, I would actually expect a bunch of specs to start failing. I don't think it would be a surprise because if you change a value in the factory and it is implicitly used by every factory instance, then you'd _want_ the specs to tell you (by failing) about all the places where you were relying on that (plus, a ton of other tests would fail too, I think).

IMO, we should be relying on our tests to tell us when these things change, rather than having to explicitly override a factory attribute for clarity's sake. I personally think it's more confusing when factory instances override a default value with the same value as the default _itself_. If we're overriding a value, I would expect it to be different from the default (otherwise, I kind of feel like the default isn't doing much at that point).

Let me know if that ^^ makes sense or if you totally disagree :)

vaidehijoshi:
> I noticed that for this spec you are using `log_in_user` when I searched the codebase for that I saw that it's only used in one other spec `spec/system/authentication/user_logs_in_with_email_spec.rb`. Everywhere else we use `sign_in(user)` and the `sign_in` [method we actually had to patch in order to ensure that a user would authenticate properly before the next request was sent](https://github.com/forem/forem/blob/master/spec/support/initializers/warden.rb) bc I [found a race condition which was causing all of our flaky Pundit errors before](https://github.com/forem/forem/pull/8379).
> 
> Any reason not to switch to using the warden helper that we know is reliable when it comes to signing in?

I followed the setup from `user_logs_in_with_email_spec` for two reasons:
1) I needed _some_ authentication method to "mock" the sign up process to test the onboarding flow
2) I picked email/password login because I know we'll be explicitly using _only_ that auth method for Forem creator onboarding (and I expect those tests to live in this file as well, since they'll test the creator onboarding flow).

I couldn't find where `sign_in` is defined (maybe you know where I should be looking for that?), but the reason I went with `log_in_user` is because it's what we were using in the email/password login spec. FWIW, it just fills in the email/password form and clicks the "continue" button:
```ruby
def log_in_user(user)
    fill_in("user_email", with: user.email)
    fill_in("user_password", with: user.password)
    click_button("Continue", match: :first)
  end
```

It's definitely possible that the `log_in_user` method itself is flaky, but it _is_ what we're using in the email/password login spec, and that one doesn't _seem_ to be flaky, from what I have noticed at least 🙃

spec/system/onboardings/user_completes_onboarding_spec.rb
mstruve: > The reason behind this is that I think that if someone goes into the user factory and does change the default from true to false, I would actually expect a bunch of specs to start failing. 

Ideally, I think specs should only fail bc code is broken. The point of a test, as I see it, is to test the code and alert you to when the code breaks and no longer performs the expected behavior. In this case, the code is not broken we only changed the way we set up the test. The fact that a change in a factory setting breaks a bunch of tests simply bc the test set up changed seems less than ideal and makes it a pain if we want to tweak our factory or other general test settings

mstruve:
The [sign_in method is provided to us from Devise::Test::IntegrationHelpers](https://github.com/heartcombo/devise/blob/8664bac682c519151fcd94b61dc37132c225915e/lib/devise/test/integration_helpers.rb#L37) which ensures the session and everything is set correctly for the test. I noticed that the `log_in_user` method in the other spec is being used without `js: true`. Having javascript enabled and see this be flaky makes me wonder if we need the more robust warden helper to ensure the session backend and frontend are all in sync with the user session.

spec/system/onboardings/user_completes_onboarding_spec.rb
vaidehijoshi: Chatted with @mstruve about this a bit synchronous, and I think I have a better sense of the issue that she's trying to avoid here.

To close the loop for anyone else who stumbles upon this...we think that [these changes](https://github.com/forem/forem/pull/11259/commits/6b3bef932df71de73fcf7c39491131248239e8ad) should remain, because the value of that attribute isn't pertinent/relevant to those tests. But in this case, it's the fact that I am testing something that _relies_ on the default that could lead to changing the default value of `saw_onbaording` in the user factory, but then being forced to sift through and debug whether the test is failing because of the default value or because of something else. So I think it's valid for us to be explicit here 👍

spec/system/onboardings/user_completes_onboarding_spec.rb
msarit: Hey @vaidehi, I actually support your approach to fixing the flakiness; in my previous work on `/admin/config` my specs would keep failing intermittently  if I hadn't defined some Config settings like `mascot image` in my spec file

vaidehijoshi:
Going to close this since I believe some of this has been addressed in other PRs. If the specs continue to be flaky, we can open a new PR to fix them.

PR #11283, opened 2020-11-05
Comments UI update

ludwiczakpawel:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR cleans up (a bit) our frontend code responsible for comments AND it introduces slightly updated UI as well.

Unfortunately it would be pretty hard to split it into several PRs because all "important" parts of code are related and dependent on each other.

Click around and break things.

image

image

## Added tests?

- [x] No, not yet. I'm about too but I may need some help with this because I'm pretty unexperienced with writing tests...

## Added to documentation?

- [x] No documentation needed

There's always something related with cache...


app/views/comments/_comment.html.erb
```diff
@@ -1,7 +1,21 @@
 <% if comment && comment.user %>
-  <% if comment.depth < 3 && !should_be_hidden?(comment, @root_comment) %>
-    <details open>
-      <summary><span>&nbsp;</span></summary>
+  <% if comment.depth < 4 && !should_be_hidden?(comment, @root_comment) %>
```
rhymes, line 2: did we add a new level of nesting?

rhymes (commented):
Looks cool, love the index and the delete page!!!! A few things:

- I'm not sure the collapse icon is very visible, I barely noticed it on the light theme
- this is what happens when you delete a comment, it looks like the icon gets botched, or maybe it's me having issues with the Sloan icon?:

Screenshot_2020-11-05  Fear and Trembling Et unde

ludwiczakpawel:
@rhymes 

> this is what happens when you delete a comment, it looks like the icon gets botched, or maybe it's me having issues with the Sloan icon?:

yea, there are issues with Sloan avatar **locally**... And this text is `alt` value... Not sure if we should fix it anyhow..? I mean, i could do it quickly there...

> I'm not sure the collapse icon is very visible, I barely noticed it on the light theme

Hm, ok, I'll try to make them more "pop".

app/views/comments/_comment.html.erb
ludwiczakpawel: No, we are not. But I slightly changed the logic of displaying  and applying appropriate classes for child nodes (it used to be applied elsewhere, inside `details` element. So this should not really change the logic for how the comments are nested, just some mechanisms for how we display it..

I've tried to test it through and it worked okey on my local env but if you can think of a testing scenario for this and make sure it doesn't really change anything significant, I'd really appreciate your help.

rhymes: I just added a bunch of comments to the same thread and saw they don't get nested too much so I think we're okay :D

citizen428, line 2: I wonder if we should move this hardcoded 4 into a constant, e.g. `MAX_COMMENT_DEPTH`. Then we don't need to worry about the value here and a few lines below getting out of sync.

app/views/comments/_comment_header.html.erb
```diff
@@ -0,0 +1,34 @@
+<header class="comment__header">
+  <a href="/<%= comment.user.username %>" class="crayons-link crayons-link--secondary flex items-center fw-medium">
```
citizen428, line 2: I recommend using our `URL` helper module as much as possible instead of manually building links:

```suggestion
  <a href="<%= URL.user(comment.user) %>" class="crayons-link crayons-link--secondary flex items-center fw-medium">
```

app/views/comments/_comment_header.html.erb
```diff
@@ -0,0 +1,34 @@
...
+
+  <%= render "comments/comment_date", decorated_comment: decorated_comment %>
+
+  <div class="comment__dropdown">
+    <button class="dropbtn crayons-btn crayons-btn--s crayons-btn--ghost-dimmed crayons-btn--icon" aria-label="Toggle dropdown menu">
+      <%= inline_svg_tag("overflow-horizontal.svg", aria: true, class: "crayons-icon pointer-events-none", title: "Dropdown menu") %>
+    </button>
+    <div class="dropdown">
+      <div class="crayons-dropdown p-1 right-1 s:right-0 s:left-auto fs-base">
+        <a href="<%= comment.path %>" class="crayons-link crayons-link--block">Permalink</a>
```
citizen428, line 19: This is another case where our `URL` module could help:

```suggestion
        <a href="<%= URL.comment(comment) %>" class="crayons-link crayons-link--block">Permalink</a>
```

Yes, this is slightly longer, but at least all URLs are constructed in consistent ways, so if there's ever a need for change we can do it in a single place.

app/views/comments/_comment_header.html.erb
```diff
@@ -0,0 +1,34 @@
...
+        <span class="comment-actions hidden" data-user-id="<%= comment.user_id %>" data-action="settings-button" data-path="<%= comment.path %>/settings">
+        </span>
+        <% action = comment.hidden_by_commentable_user ? "Unhide" : "Hide" %>
```
citizen428, line 22: I think you only use this in one place, so no need for a variable.

app/views/comments/index.html.erb
```diff
@@ -37,96 +37,113 @@
...
-        <%= @commentable.is_a?(PodcastEpisode) ? @commentable.podcast.name : @commentable.user.name %>
-        <span class="published-at"><%= "on #{@commentable.published_at.strftime('%B %d, %Y')}" if @commentable.published_at %></span>
-      </h4>
-    </a>
-    <% if @commentable.processed_html.present? %>
-      <% if @commentable.processed_html.size < 350 %>
-        <div class="body">
-          <%= sanitize_rendered_markdown(@commentable.processed_html) %>
+<div class="crayons-layout crayons-layout--limited-l gap-0">
+  <% if @commentable.class.name == "Article" %>
```
citizen428, line 41: I wonder if this `if` can move into the payment pointer partial  so the logic on whether or not it renders something is self-contained. Then instead of passing in the user, we'd pass in the whole `commentable`. But I don't know all use cases of this, so maybe that doesn't make sense.

app/views/comments/index.html.erb
```diff
@@ -37,96 +37,113 @@
...
+    <%= render "shared/payment_pointer", user: @commentable.user %>
+    <span id="comment-article-indicator" data-article-id="<%= @commentable.id %>"></span>
+  <% end %>
+
+  <% if !@root_comment %>
```
citizen428, line 46: Ruby has a special keyword for `if !`, called `unless`:

```suggestion
  <% unless @root_comment %>
```

app/views/comments/index.html.erb
```diff
@@ -37,96 +37,113 @@
...
+    <article class="crayons-article crayons-card crayons-card--secondary s:mx-2 m:mx-4 -mb-1 z-0">
+      <header class="crayons-article__header">
+        <% if @commentable.respond_to?(:main_image) && @commentable.main_image.present? %>
```
citizen428, line 49: This works for now, but in the long run we may want a better solution leveraging OOP (e.g. `main_image` always returning `nil` for classes that don't have one, so can skip the first check.

citizen428 (approved):
I left some comments, but a lot of them aren't even specific to this PR, but about how we write view code in general. I'd really like to see less conditionals and more encapsulated logic/use of OOP. Something like Github's [ViewComponent](https://viewcomponent.org/) seems like a good approach.

app/views/comments/_comment.html.erb
ludwiczakpawel: I think that makes a lot of sense. But not sure if I should be the person doing this - not because it's not part of my role but because I just don't know how and where to place it to make it the most efficient and configurable.. Because I also imagine this COULD be something we will eventually add to admin config...

app/views/comments/_comment_header.html.erb
ludwiczakpawel: Oh nice, I'll replace it wherever I can! :)

app/views/comments/_comment_header.html.erb
ludwiczakpawel: good point!

app/views/comments/index.html.erb
ludwiczakpawel: Yea, since this partial is being used in other places too, I think I'll just pass on it for now. Someone more experienced than myself should chime-in (in a separate PR) and optimize it if it makes sense :)

app/views/comments/index.html.erb
ludwiczakpawel: I think that's a good idea but obviously, not something I'm capable of doing probably.. :C

app/views/comments/_comment.html.erb
ludwiczakpawel: FYI I reverted that logic to not make things too confusing :)

rhymes (commented):
There are two failing tests because the "Edit" button has been moved into a dropdown: https://travis-ci.com/github/forem/forem/jobs/429675929

Also, I'm not sure `Edit` should be at the bottom, shouldn't edit come before delete? 

I also noticed that there's an a11y issue regarding contrast in the default theme and the night theme (I haven't checked the others): the date color is really light. If you install [the aXe plugin](https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd) you can see it flagged

ludwiczakpawel:
@rhymes I've bumped up the size and contrast of the timestamp and I swapped the order of buttons.. But I have no idea how to fix specs.. Is this something you could help me out with maybe? 🙏

rhymes:
@ludwiczakpawel i'll take a look

rhymes:
@ludwiczakpawel I noticed something else that is odd: the dropdown menu appears only if you know where to look for it and disappears:

Screen-Recording-2020-11-06-at-3

ludwiczakpawel:
@rhymes JS for our dropdowns in general is a mess.. i've pushed a fix for this one and it should be fine but we should circle back to dropdowns at some point and refactor them 🤷

rhymes:
@ludwiczakpawel I don't think we should ship broken code in a refactor though. About the rest: is the appeareance of the dropdown a thing that's only on hover or should it be visible at all time?

ludwiczakpawel:
@rhymes not sure what do you mean? the issue you reported has been fixed in recent commit.. what i meant earlier was: when I looked at our code for dropdowns (not only in comments but in general) I've noticed its pretty messy and spaghetti-like making it pretty hard to work with. so we do not ship anything broken right now (i hope :D) - just to be clear:)

benhalpern:
This is looking gorgeous overall.

There are definitely going to be some caching concerns for this deployment, so let's make sure we take our time to think this out a little bit.

In terms of the design I personally think the elements which show up on hover are a little jarring, and it's a little odd that I don't know where to aim until I'm hovering on a comment. The narrow view where they are always present seems more pleasant to me. I think Twitter, etc. normalize the ever-presence of these kinds of menus.

The stacked comments on the show page is really inspired, I love it!

One last things: The header seems _a little more indented_ than I'm expecting. The menu button seems more inset than seems natural. I'd push for _slightly_ less top padding in these areas.

Screen Shot 2020-11-06 at 1 18 17 PM

ludwiczakpawel:
> There are definitely going to be some caching concerns for this deployment, so let's make sure we take our time to think this out a little bit.

Oh definitely!

> the elements which show up on hover are a little jarring, and it's a little odd that I don't know where to aim until I'm hovering on a comment

I've made little changes:

- "collapse" icon is now visible by default but has reduced opacity until you hover.
- "3 dots" icon is now a bit smaller (click target is the same) but remains visible only on hover. The reason for not making it visible persistent it it feels a bit misbalanced when there's lots of comments one under another and that icon is the only one being visible on the right of entire comments section. 

^ I think this feels a bit less jarring yet still pretty well-balanced. But I'm keen to know your thoughts. 

> One last things: The header seems a little more indented than I'm expecting. The menu button seems more inset than seems natural. I'd push for slightly less top padding in these areas.

I'm not sure if I know what do you mean here or if I can see it the same way you do..?

rhymes:
I wonder if the reason why we can't get the tests to pass it's due to the fact that the collapse icon is only available on hover, it's strange though.

ludwiczakpawel:
@rhymes no idea.. i'm a such a noob when it comes to tests/specs... :c

benhalpern:
> "3 dots" icon is now a bit smaller (click target is the same) but remains visible only on hover. The reason for not making it visible persistent it it feels a bit misbalanced when there's lots of comments one under another and that icon is the only one being visible on the right of entire comments section.

I'll defer to @lisasy for final thoughts here. I still like it being present the way I expect on Twitter etc. but I'm also cool if the consensus is that we like it on hover.

What I'm saying about the inset distance for the header (user title and menu button) is that I'm just used to it being a little closer to the edge.

This change: 
Screen Shot 2020-11-16 at 6 59 25 PM
Twitter:
Screen Shot 2020-11-16 at 6 59 32 PM

It makes more sense in the _non-hover_ state, IMO.

Anyway, I'm cool with whatever we do here, just wanted to raise the discussion.

Once we have this all settled 100% we'll touch base on deployment strategy and caching expectations.

ludwiczakpawel:
> I'll defer to @lisasy for final thoughts here. I still like it being present the way I expect on Twitter etc. but I'm also cool if the consensus is that we like it on hover.

If you feel strongly about it, I'll make it visible by default and just reduce the opacity a little when it's not on hover. It's closer to the original (current) version of comments and we can take it step further (in one way or the other) with next iteration of this UI (aka "timeline" https://forem.team/pp/comment/ll )

> What I'm saying about the inset distance for the header (user title and menu button) is that I'm just used to it being a little closer to the edge.

Oh, I see. Gotcha, I think it looks good no matter how close to the edge so I'm happy with closer variant as well.

> Once we have this all settled 100% we'll touch base on deployment strategy and caching expectations.

 Awesome <3 I still need a bit of help with failing test. It happens because we moved "Edit" and "Delete" buttons to the dropdown and they are now not visible by default. I'm not experienced enough with Capybara to figure this out :(

rhymes:
@ludwiczakpawel thanks for leaving the three dots visibile 😭

spec/system/comments/user_edits_a_comment_spec.rb
```diff
@@ -28,6 +28,7 @@ def assert_updated
       visit article.path.to_s
       wait_for_javascript
 
+      click_on(class: 'comment__dropdown-trigger')
```
rhymes, line 31: THE HOLY GRAIL!

lisasy:
> > "3 dots" icon is now a bit smaller (click target is the same) but remains visible only on hover. The reason for not making it visible persistent it it feels a bit misbalanced when there's lots of comments one under another and that icon is the only one being visible on the right of entire comments section.
> 
> I'll defer to @lisasy for final thoughts here. I still like it being present the way I expect on Twitter etc. but I'm also cool if the consensus is that we like it on hover.
> 
> What I'm saying about the inset distance for the header (user title and menu button) is that I'm just used to it being a little closer to the edge.
> 
> This change:
> Screen Shot 2020-11-16 at 6 59 25 PM
> Twitter:
> Screen Shot 2020-11-16 at 6 59 32 PM
> 
> It makes more sense in the _non-hover_ state, IMO.
> 
> Anyway, I'm cool with whatever we do here, just wanted to raise the discussion.
> 
> Once we have this all settled 100% we'll touch base on deployment strategy and caching expectations.

Based on the current state of the PR, I'm seeing that the "..." button is persistently available. I think that making this persistently available is the right call for UX so that people can easily discover that they can take actions here — this makes it all the more extra accessible, and there is no need to hide this.

lisasy:
@ludwiczakpawel I think this is a pretty awesome PR and a great first step towards updating the general styles before we iterate on other chunks of work, such as the nesting interaction, for example. Given this scope, my feedback here is mainly visual:

### In this PR
Screen Shot 2020-11-17 at 10 02 12 AM

I noticed that with the way that nesting works right now, it doesn't feel like there is enough contrast and hierarchy to help the audience distinguish between a top-level comment and its replies. I think there are two things we can do here:
1. Increase margin on left to create a clearer "line" between a top-level comment and reply
2. Unify the avatar sizes for both top-level comment and reply so that people don't pay attention to the size different and focus instead on the margin.


### Top-level comment vs. reply
Here it was this change could look like:
Screen Shot 2020-11-17 at 9 57 28 AM


### Additional things we can do:
There were two others things we can do to create more density amidst the varying information here.
1. **Tighten the comment body by making the first `p` have a `margin-top: 0`.** Since all `p` tags automatically have their own top-margin, I think that eliminating this for the first `p` in a comment body will get read of unnecessary white-space.
2. **On breakpoint-xl, keep the padding area for both the comments and post body as it is on breakpoint-l.** Since the width of the card itself is relatively constrained on breakpoint-xl, the `--su-8` padding still looks pretty good here. Maintaining information density would be good for the eye would help with readability here.

In the screenshot below, I am illustrating what a `--su-8` padding for all sides could look like, and it didn't look so bad! What do you think?

Screen Shot 2020-11-17 at 9 54 52 AM

lisasy:
I don't know if this suggestion needs to happen in *this* PR, but I have some ideas around how we can improve the comments dropdown:

1. Change label from "Copy Permalink" to "Copy link"
2. Eliminate the "Permalink" link and make the comment box itself clickable.

Screen Shot 2020-11-17 at 10 18 44 AM

Since we now have the "Copy Permalink" feature, I'm not sure what motivation someone has to just go to the comment itself, but if they do, why not make that whole thing clickable?

@ludwiczakpawel @benhalpern

marcysutton (changes requested):
Cool update! I found some accessibility things that would be good to address, either here or in future PRs if they're out of scope.

One thing I found on the rendered flow that didn't seem to show up in the changed files was a missing label for the Reply textarea. The "Add to the Discussion" textareas have an `aria-label`, but a field with a "Reply..." placeholder does not. Ideally these attributes would have different values so they aren't redundant. Here's a link to more info: https://www.w3.org/WAI/tutorials/forms/instructions/#placeholder-text

Let me know if you have any accessibility questions!

ludwiczakpawel:
Thanks @marcysutton - that is extremely helpful. I'm a complete newbie in terms of accessibility etc so I really appreciate your feedback here. I'll try to implement as much as possible today.

@lisasy, I'll give it a look with alignment.

> I don't know if this suggestion needs to happen in this PR, but I have some ideas around how we can improve the comments dropdown: Change label from "Copy Permalink" to "Copy link", Eliminate the "Permalink" link and make the comment box itself clickable. Since we now have the "Copy Permalink" feature, I'm not sure what motivation someone has to just go to the comment itself, but if they do, why not make that whole thing clickable?

- I'm all in for renaming "Copy permalink" to "Copy link".
- I definitely don't want to make the whole comment box clickable - it's a bit shady and sometimes tricky to solve since there are other clickable elements inside the comment (author, avatar, buttons, and even stuff in comment content like links, liquid tags). It will be also pretty hard to select text if you ever want to. Another thing is when you misclick the link in the comment content and you will instead land on comment view. 
- But I do think we can drop the "Permalink" button entirely since every comment has a timestamp (next to user's name) which and it's a pretty common pattern to make it clickable.

> Additional things we can do: There were two others things we can do to create more density amidst the varying information here...

Let's handle these separately - all the margins and spacing between elements in content (as well as padding in content area) is part of a bigger discussion and also code change. I'm not sure if I'm sold on everything at the moment tbh :)

app/views/comments/_comment_header.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      <span class="crayons-tooltip inline-block spec-op-author" data-tooltip="<%= get_ama_or_op_banner(commentable) %>">
+        <%= inline_svg_tag("small-medal.svg", aria: true, class: "crayons-icon", title: get_ama_or_op_banner(commentable)) %>
+      </span>
+    <% end %>
+  </a>
+
+  <%= render "comments/comment_date", decorated_comment: decorated_comment %>
+
+  <div class="comment__dropdown">
+    <button class="dropbtn comment__dropdown-trigger crayons-btn crayons-btn--s crayons-btn--ghost crayons-btn--icon" aria-label="Toggle dropdown menu">
```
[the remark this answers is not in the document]
ludwiczakpawel: Unfortunately we do not have that any JS to indicate dropdown state like that. BUT dropdowns overall are pretty outdated and need a bit of refactor so I'm sure we could add something in later on.

As for the `aria-haspopup` - no problems!

app/views/comments/_comment_header.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      <%= inline_svg_tag("small-overflow-horizontal.svg", aria: true, class: "crayons-icon pointer-events-none", title: "Dropdown menu") %>
+    </button>
+    <div class="dropdown">
+      <div class="crayons-dropdown p-1 right-1 s:right-0 s:left-auto fs-base">
+        <a href="<%= URL.comment(comment) %>" class="crayons-link crayons-link--block">Permalink</a>
```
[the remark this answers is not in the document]
ludwiczakpawel: Isn't adding a timestamp a bit too granular here? Would this be enough?

```
<a href="..." ... aria-label="Copy link to <%= comment.user.name %>'s comment">Copy link</a>
```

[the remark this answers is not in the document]
ludwiczakpawel: Actually.. Would it make sense to make context a bit more generic? My point is, for every of these links we want to give information about comment's author (and timestamp) which seems to be pretty redundant and sometimes weird (when it's about your own comment).. What if we do something like this instead:

```
<ul class="dropdown ..." aria-label="${authorName}'s comment options">
    <li><a ... aria-label="Copy link to this comment">Copy link</a></li>
    <li><a ... aria-label="Go to comment's settings">Settings</a></li>
    <li><a ... aria-label="Hide this comment">Hide</a></li>
    <li><a ... aria-label="Moderate this comment">Moderate</a></li>
    <li><a ... aria-label="Report this comment as abusive">Report abuse</a></li>
    <li><a ... aria-label="Edit this comment">Edit</a></li>
    <li><a ... aria-label="Delete this comment">Delete</a></li>
</ul>
```

Would that work? Would that be more "human"? 
(or even `<ul class="dropdown ..." aria-label="Comment's options">` <-- without ${authorName} ?)

app/views/articles/_conduct_and_abuse_actions.html.erb
```diff
@@ -1,3 +1,5 @@
-<a href="/code-of-conduct" class="crayons-link crayons-link--brand">Code of Conduct</a>
-<span class="opacity-25 px-2">&bull;</span>
-<a href="/report-abuse" class="crayons-link crayons-link--brand">Report abuse</a>
+<p class="fs-s align-center">
```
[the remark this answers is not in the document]
ludwiczakpawel: I went with `nav` element.

app/views/comments/_comment_header.html.erb
```diff
@@ -0,0 +1,34 @@
...
+    <% if commentable_author_is_op?(commentable, comment) %>
+      <span class="crayons-tooltip inline-block spec-op-author" data-tooltip="<%= get_ama_or_op_banner(commentable) %>">
+        <%= inline_svg_tag("small-medal.svg", aria: true, class: "crayons-icon", title: get_ama_or_op_banner(commentable)) %>
+      </span>
+    <% end %>
+  </a>
+
+  <%= render "comments/comment_date", decorated_comment: decorated_comment %>
+
+  <div class="comment__dropdown">
```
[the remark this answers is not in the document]
ludwiczakpawel: Yea, as mentioned elsewhere, dropdowns JS need a bit more love so that's definitely out of scope for now. But worth keeping it in mind 👍

app/views/comments/_comment_header.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      <%= inline_svg_tag("small-overflow-horizontal.svg", aria: true, class: "crayons-icon pointer-events-none", title: "Dropdown menu") %>
+    </button>
+    <div class="dropdown">
+      <div class="crayons-dropdown p-1 right-1 s:right-0 s:left-auto fs-base">
+        <a href="<%= URL.comment(comment) %>" class="crayons-link crayons-link--block">Permalink</a>
+        <a href="<%= URL.comment(comment) %>" class="crayons-link crayons-link--block permalink-copybtn" data-no-instant>Copy Permalink</a>
+        <span class="comment-actions hidden" data-user-id="<%= comment.user_id %>" data-action="settings-button" data-path="<%= URL.comment(comment) %>/settings"></span>
+        <% action = comment.hidden_by_commentable_user ? "Unhide" : "Hide" %>
+        <span class="comment-actions hidden" data-action="hide-button" data-commentable-user-id="<%= commentable.user_id %>" data-user-id="<%= comment.user_id %>">
+          <a href="#" class="crayons-link crayons-link--block hide-comment" data-hide-type="<%= action.downcase %>" data-comment-id="<%= comment.id %>">
```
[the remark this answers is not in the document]
ludwiczakpawel: > Not sure if this should be part of this PR or not but when the "Hide comment" alert dialog that pops up and I try to cancel it with the keyboard, it hides the comment anyway.

Hmm that doesn't happen to me either locally or on production. Not sure what's going on there for you.. Can you give it another try? And also on prod?

Re kicking back focus to the right place - definitely room for improvement here, but probably not this PR - it sounds like a bigger refactor too.

msarit:
> Since we now have the "Copy Permalink" feature, I'm not sure what motivation someone has to just go to the comment itself, but if they do, why not make that whole thing clickable? (Lisa feedback)
---
> I definitely don't want to make the whole comment box clickable - it's a bit shady and sometimes tricky to solve since there are other clickable elements inside the comment (author, avatar, buttons, and even stuff in comment content like links, liquid tags). It will be also pretty hard to select text if you ever want to. Another thing is when you misclick the link in the comment content and you will instead land on comment view. (Pawel response)

---

I'm with @ludwiczakpawel on this one 👍

benhalpern (approved):
This looks good to me.

I think we will want to merge this and then merge https://github.com/forem/forem/pull/11472 right after (If folks are okay with that approach)

cc: @mstruve

ludwiczakpawel:
i've added "merge by any core" tag because I don't really know how do we want to sync up all of that ^. cc @benhalpern @mstruve

PR #11291, opened 2020-11-05
Refactor comment dropdown menu functionality

benhalpern:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This is a refactor which should not affect user functionality, but it does have nice positive impacts.

- IMO it cleans up the code a bit.
- It saves a good deal of bandwidth. Every comment dropdown at the moment is about 800 characters. I'm not sure how it shakes up after gzipping etc, but this is definitely an improvement for overall file size. A page with 50 comments currently uses 4,000 characters just to render these dropdown menus. 😱
- SEO will enjoy slightly smaller pages. It is also good SEO practice to strip away sections that do not serve the page's content and important navigation, where appropriate, so this should be a solid benefit in that regard. 

I also removed a bit of dead code along the way.

### DIFF
--- a/app/assets/javascripts/initializers/initializeBaseUserData.js
+++ b/app/assets/javascripts/initializers/initializeBaseUserData.js
@@ -46,50 +46,6 @@ function addRelevantButtonsToArticle(user) {
   }
 }
 
-function addRelevantButtonsToComments(user) {
-  if (document.getElementById('comments-container')) {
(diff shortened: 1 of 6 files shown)


benhalpern:
Closing this because it conflicts too much since we changed up the HTML in another PR. Will approach this again later.

PR #11297, opened 2020-11-06
Add custom instrumentation to Feeds::Import

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This should give us a bit more visibility and answer the question about how fast each section roughly is

#10998
#11234 

## Added tests?

- [ ] Yes
- [x] No, and this is why: this is likely temporary and also I'm trying to understand if this is the best approach

### DIFF
--- a/app/services/feeds/import.rb
+++ b/app/services/feeds/import.rb
@@ -24,9 +24,11 @@ module Feeds
       users.in_batches(of: users_batch_size) do |batch_of_users|
         feeds_per_user_id = fetch_feeds(batch_of_users)
         Rails.logger.error("feeds::import::feeds_per_user_id.length: #{feeds_per_user_id.length}")
+        DatadogStatsClient.count("feeds::import::fetch_feeds.count", feeds_per_user_id.length)
 
         feedjira_objects = parse_feeds(feeds_per_user_id)
         Rails.logger.error("feeds::import::feedjira_objects.length: #{feedjira_objects.length}")
+        DatadogStatsClient.count("feeds::import::parse_feeds.count", feedjira_objects.length)
 
         # NOTE: doing this sequentially to avoid locking problems with the DB
         # and unnecessary conflicts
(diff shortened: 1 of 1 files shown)


app/services/feeds/import.rb
```diff
@@ -24,26 +24,32 @@ def call
       users.in_batches(of: users_batch_size) do |batch_of_users|
         feeds_per_user_id = fetch_feeds(batch_of_users)
         Rails.logger.error("feeds::import::feeds_per_user_id.length: #{feeds_per_user_id.length}")
+        DatadogStatsClient.count("feeds::import::fetch_feeds.count", feeds_per_user_id.length)
 
         feedjira_objects = parse_feeds(feeds_per_user_id)
         Rails.logger.error("feeds::import::feedjira_objects.length: #{feedjira_objects.length}")
+        DatadogStatsClient.count("feeds::import::parse_feeds.count", feedjira_objects.length)
```
maestromac, line 31: Just a though, do you want to keep the name consistent with the Rails.logger?
```suggestion
        DatadogStatsClient.count("feeds::import::feedjira_objects.length", feedjira_objects.length)
```

rhymes: I actually wanted to change the logger, as "feedjra objects" doesn't mean much long term in the logs. I prefer counting the times the actions was invoked. I'll unify them but the other way :D

app/services/feeds/import.rb
```diff
@@ -23,27 +23,30 @@ def call
...
 
         # NOTE: doing this sequentially to avoid locking problems with the DB
         # and unnecessary conflicts
         articles = feedjira_objects.flat_map do |user_id, feed|
           # TODO: replace `feed` with `feed.url` as `RssReader::Assembler`
           # only actually needs feed.url
           user = batch_of_users.detect { |u| u.id == user_id }
-          create_articles_from_user_feed(user, feed)
+
+          DatadogStatsClient.time("feeds::import::create_articles_from_user_feed", tags: ["user_id:#{user_id}"]) do
```
mstruve, line 38: I have never used the `time` stat here, interested to see how it works! If we find the Datadog metrics not giving us enough insight or cumbersome to work with I would suggest changing this to a Honeycomb span.

mstruve (approved):
Let's give it a go!

maestromac (approved):

benhalpern (approved):

PR #11299, opened 2020-11-06
Validate Image URLs in SiteConfig

juliannatetreault:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
**The issue**:
> As a Forem Admin, I might make a mistake and add erroneous info into the image URL field. This shouldn't break the site.

**The solution**:
This PR adds validations to the `Admin::ConfigsController` that check an image's URL against some regexp to ensure that the image URL is valid before saving the changes to the SiteConfig.

Closes https://github.com/forem/InternalProjectPlanning/issues/249

**To QA this PR, first ensure that you have an admin role, and then please do the following**:

- Navigate to `/admin/config`:
Screen Shot 2020-11-06 at 3 02 09 PM

- Navigate to the "Images" section and attempt to update an image with an invalid image URL, such as `![low_res]https://fjhkfhwk.png`. You should be shown an error message: "😭  Image URL must be a valid URL":
Screen Shot 2020-11-06 at 3 03 30 PM
Screen Shot 2020-11-06 at 3 04 04 PM

- Navigate back to the "Images" section and update the form with a valid image URL--you should be shown a success message upon update:
Screen Shot 2020-11-10 at 3 44 08 PM
Screen Shot 2020-11-10 at 3 43 32 PM

- Attempt to update the SiteConfig with invalid image URLs in other sections where images can be uploaded (Mascot, Onboarding, etc.) and ensure that it fails with the error message and redirect _back_ to the `admin_config_path`:
Screen Shot 2020-11-06 at 3 06 10 PM
Screen Shot 2020-11-06 at 3 04 04 PM

- Try to break things! 🔨 

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

Nope!

Upload Button

### DIFF
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -142,7 +142,7 @@ module Admin
         onboarding_taskcard_image
       ].freeze
 
-    VALID_URL = %r{\A(http|https)://([/|.|\w|\s|-])*.[a-z]{2,5}(:[0-9]{1,5})?(/.*)?$}.freeze
+    VALID_URL = %r{\A(http|https)://([/|.|\w|\s|-])*.[a-z]{2,5}(:[0-9]{1,5})?(/.*)\z}.freeze
 
     layout "admin"
 
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -133,6 +133,7 @@ module Admin
 
     before_action :extra_authorization_and_confirmation, only: [:create]
     before_action :validate_inputs, only: [:create]
+    before_action :validate_image_urls, only: [:create]
     after_action :bust_content_change_caches, only: [:create]
 
     def show
@@ -202,6 +203,12 @@ module Admin
       redirect_to admin_config_path, alert: "😭 #{errors.join(',')}" if errors.any?
     end
 
+    def validate_image_urls
+      errors = []
+      errors << "Image URL must be a valid URL" unless valid_image_url
+      redirect_to admin_config_path, alert: "😭 #{errors.join(',')}" if errors.any?
+    end
+
     def clean_up_params
       config = params[:site_config]
       return unless config
@@ -232,6 +239,14 @@ module Admin
     def brand_color_not_hex
       hex = params.dig(:site_config, :primary_brand_color_hex)
       hex.present? && !hex.match?(/\A#(\h{6}|\h{3})\z/)
+    end
+
+    def valid_image_url
+      image_url = params.dig(:site_config, :main_social_image) ||
+        params.dig(:site_config, :logo_png) ||
+        params.dig(:site_config, :secondary_logo_url)
+      valid_url = %r{\A(https:)//([/|.|\w|\s|-])*\.(?:jpg|gif|png)/}
+      image_url.present? && image_url.match?(valid_url)
     end
   end
 end
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -244,7 +244,13 @@ module Admin
     def valid_image_url
       image_url = params.dig(:site_config, :main_social_image) ||
         params.dig(:site_config, :logo_png) ||
-        params.dig(:site_config, :secondary_logo_url)
+        params.dig(:site_config, :secondary_logo_url) ||
+        params.dig(:site_config, :campaign_sidebar_image) ||
+        params.dig(:site_config, :mascot_image_url) ||
+        params.dig(:site_config, :mascot_footer_image_url) ||
+        params.dig(:site_config, :onboarding_logo_image) ||
+        params.dig(:site_config, :onboarding_background_image) ||
+        params.dig(:site_config, :onboarding_taskcard_image)
       valid_url = %r{\A(https:)//([/|.|\w|\s|-])*\.(?:jpg|gif|png)/}
       image_url.present? && image_url.match?(valid_url)
     end
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -129,7 +129,7 @@ module Admin
         home_feed_minimum_score
       ].freeze
 
-    VALID_IMAGE_URLS =
+    IMAGE_FIELDS =
       %i[
         logo_png
         secondary_logo_url
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -129,28 +129,23 @@ module Admin
         home_feed_minimum_score
       ].freeze
 
-    # VALID_IMAGE_URLS =
-    #   %i[
-    #     logo_png
-    #     secondary_logo_url
-    #     mascot_image_url
-    #     mascot_footer_image_url
-    #     campaign_sidebar_image
-    #     onboarding_logo_image
-    #     onboarding_background_image
-    #     onboarding_taskcard_image
-    #   ].freeze
+    VALID_IMAGE_URLS =
+      %i[
+        logo_png
+        secondary_logo_url
+        mascot_image_url
+        mascot_footer_image_url
+        campaign_sidebar_image
+        onboarding_logo_image
+        onboarding_background_image
+        onboarding_taskcard_image
+      ].freeze
 
     layout "admin"
 
     before_action :extra_authorization_and_confirmation, only: [:create]
     before_action :validate_inputs, only: [:create]
-    after_action :validate_image_urls, only: [:create]
-    # before_action :validate_image_urls, only: [:valid_image_url]
-    # before_action :validate_image_urls, if: :valid_image_url, only: [:create]
-    # before_action :validate_image_urls, only: [:valid_image_url], if: -> { :VALID_IMAGE_URLS }
-    # before_save :validate_image_urls
-    # before_action :validate_image_urls, only: [:valid_image_url]
+    before_action :validate_image_urls, only: [:create], if: -> { VALID_IMAGE_URLS.present? }
     after_action :bust_content_change_caches, only: [:create]
 
     def show
@@ -159,7 +154,6 @@ module Admin
 
     def create
       clean_up_params
-      # validate_image_urls
 
       config_params.each do |key, value|
         if value.is_a?(Array)
@@ -271,7 +265,6 @@ module Admin
         params.dig(:site_config, :onboarding_taskcard_image)
       valid_url = %r{\A(https:)//([/|.|\w|\s|-])*\.(?:jpg|gif|png)/}
       image_url.present? && image_url.match?(valid_url)
-      # VALID_IMAGE_URLS.present? && VALID_IMAGE_URLS.match?(valid_url)
     end
   end
 end
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -129,11 +129,28 @@ module Admin
         home_feed_minimum_score
       ].freeze
 
+    VALID_IMAGE_URLS =
+      %i[
+        logo_png
+        secondary_logo_url
+        mascot_image_url
+        mascot_footer_image_url
+        campaign_sidebar_image
+        onboarding_logo_image
+        onboarding_background_image
+        onboarding_taskcard_image
+      ].freeze
+
     layout "admin"
 
     before_action :extra_authorization_and_confirmation, only: [:create]
     before_action :validate_inputs, only: [:create]
-    before_action :validate_image_urls, only: [:create]
+    after_action :validate_image_urls, only: [:create]
+    # before_action :validate_image_urls, only: [:valid_image_url]
+    # before_action :validate_image_urls, if: :valid_image_url, only: [:create]
+    # before_action :validate_image_urls, only: [:valid_image_url], if: -> { :VALID_IMAGE_URLS }
+    # before_save :validate_image_urls
+    # before_action :validate_image_urls, only: [:valid_image_url]
     after_action :bust_content_change_caches, only: [:create]
 
     def show
@@ -142,6 +159,7 @@ module Admin
 
     def create
       clean_up_params
+      # validate_image_urls
 
       config_params.each do |key, value|
         if value.is_a?(Array)
@@ -253,6 +271,7 @@ module Admin
         params.dig(:site_config, :onboarding_taskcard_image)
       valid_url = %r{\A(https:)//([/|.|\w|\s|-])*\.(?:jpg|gif|png)/}
       image_url.present? && image_url.match?(valid_url)
+      # VALID_IMAGE_URLS.present? && VALID_IMAGE_URLS.match?(valid_url)
     end
   end
 end
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -130,7 +130,7 @@ module Admin
       ].freeze
 
     IMAGE_FIELDS =
-      %i[
+      %w[
         main_social_image
         logo_png
         secondary_logo_url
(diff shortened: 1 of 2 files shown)


rhymes:
Hi @juliannatetreault I glanced quickly at the code and I have a consideration:

validations in general should be done at the model level if possible, this way they get validated also if they are manually changed in the console, same as any other model. Unfortunately in the case of `SiteConfig` there's no super easy way as they are not regular columns with a name and a simple value, but this thread has a couple of possible solutions: https://github.com/huacnlee/rails-settings-cached/issues/153 (`SiteConfig` is based on that gem). I haven't tried to be frank.

I guess that inside the body of one of the solution you can invoke `ActiveModel::Validations::UrlValidator` which is the class we use internally to validate URLs in the user's profile and other places

But don't consider it a blocker, if all you need is some simple escape route to clean up user data your approach should be enough :)

juliannatetreault:
> huacnlee/rails-settings-cached#153

Hey, @rhymes!

Thanks for the feedback -- I actually read through that thread yesterday! The reason I'm doing the validations in the `Admin::ConfigsController` rather than via the traditional model validations is to better follow our code's convention -- our validations seem to live in the `Admin::ConfigsController` and not the model itself and I would guess it has something to do with the fact that the model isn't a traditional model and inherits from `RailsSettings::Base`. If you and others on the team think I should move this work into the `SiteConfig` model once I finish my work and open the PR up for review, then I'll try approaching it in the way that you suggested. 😊

rhymes:
@juliannatetreault I think we can do that as a separate refactor once this "shield" is in production for all forems! Glad we ended up reading the same thing hahaha

app/controllers/admin/configs_controller.rb
```diff
@@ -129,10 +129,23 @@ class ConfigsController < Admin::ApplicationController
...
+        onboarding_logo_image
+        onboarding_background_image
+        onboarding_taskcard_image
+      ].freeze
+
     layout "admin"
 
     before_action :extra_authorization_and_confirmation, only: [:create]
     before_action :validate_inputs, only: [:create]
+    before_action :validate_image_urls, only: [:create], if: -> { VALID_IMAGE_URLS.present? }
```
juliannatetreault, line 148: This `before_action` is still too aggressive and validates _every_ field regardless of whether or not it is an image on `create` -- any ideas for how to adjust this to only validate images on `create`? 🤔  I've tried a ton of different approaches at this point and still cannot get this to work only for images.

spec/requests/admin/configs_spec.rb
```diff
@@ -237,13 +237,31 @@
           expect(SiteConfig.main_social_image).to eq(expected_image_url)
         end
 
+        it "only updates the main_social_image if given a valid image URL" do
+          expected_image_url = "![logo_lowres]https://dummyimage.com/300x300"
+          expect do
+            post "/admin/config", params: { site_config: { main_social_image: expected_image_url },
+                                            confirmation: confirmation_message }
+            # expect(SiteConfig.main_social_image).not_to eq(expected_image_url)
```
juliannatetreault, line 245: I'll remove this comment in addition to the comments on lines  `261`, `279`, and `302` once I have the validation in the controller working!

app/controllers/admin/configs_controller.rb
```diff
@@ -129,10 +129,23 @@ class ConfigsController < Admin::ApplicationController
         home_feed_minimum_score
       ].freeze
 
+    VALID_IMAGE_URLS =
```
Zhao-Andy, line 132: I feel like `VALID_IMAGE_URLS` is not quite describing what you want here. Maybe `VALID_IMAGE_FIELDS` or just `IMAGE_FIELDS`?

app/controllers/admin/configs_controller.rb
Zhao-Andy: So `VALID_IMAGE_URLS.present?` always returns true because the constant is always there. I'm guessing you're trying to do something like:
```
validate_image_urls, only: [:create], if -> { site config params have image fields in them }
```

Since you have `VALID_IMAGE_URLS` already, you want to compare that to the site config params' keys and see if the two arrays have intersecting elements.

juliannatetreault: Great suggestions! I've changed the constant name and will try comparing the constant against the params like you mentioned. 😊 

Thank you so much for the extra set of eyes, @Zhao-Andy!

app/controllers/admin/configs_controller.rb
```diff
@@ -233,5 +256,10 @@ def brand_color_not_hex
       hex = params.dig(:site_config, :primary_brand_color_hex)
       hex.present? && !hex.match?(/\A#(\h{6}|\h{3})\z/)
     end
+
+    def valid_image_url(url)
+      valid_url = %r{\A(http|https)://([/|.|\w|\s|-])*.(?:jpg|gif|png)}
```
citizen428, line 261: **suggestion (non-blocking):** It would be great to move the regular expression into a constant so it doesn't get recompiled on every method invocation.

spec/requests/admin/configs_spec.rb
```diff
@@ -237,13 +237,29 @@
           expect(SiteConfig.main_social_image).to eq(expected_image_url)
         end
 
+        it "only updates the main_social_image if given a valid image URL" do
+          expected_image_url = "![logo_lowres]https://dummyimage.com/300x300"
```
citizen428, line 241: **nitpick (non-blocking):** Total nitpick, but this isn't really expected, since the spec uses a `not_to change` assertion. I'd maybe just call it `invalid_image_url`.

app/controllers/admin/configs_controller.rb
juliannatetreault: TIL! Great call -- I've made this change.

spec/requests/admin/configs_spec.rb
juliannatetreault: This is not a nitpick at all, this is great! I've made this change.

juliannatetreault:
I'm a bit perplexed as to why the specs are failing... it seems like the `before_action` is likely the cause because when I alter it to something like the code shown below, 4 of the 8 failing specs pass (this approach doesn't _actually_ work as far as validating the image URLs goes, though):
```
  before_action :validate_image_urls, only: [:create], if: -> { params[:site_config].key?(IMAGE_FIELDS) }
```

When I drop into the console via a `binding.pry` in `configs_controller.rb`, I am able to update and see the expected params without any issues and everything seems to work as expected. When I drop into the console via a `binding.pry` in the `configs_spec.rb` though, I am unable to update the image URLs for 4 out of the 8 failing specs (the specs that do not try to update to an `invalid_image_url`, but _do_ try to update the image's URL from the default to something new) and for the other 4 failing specs, I consistently get a `nil` value.

Does anyone have any idea why this might be happening? 🤔

vaidehijoshi:
> Does anyone have any idea why this might be happening and/or what I'm missing or not considering here? 🤔

👋 Hi, @juliannatetreault! I know you didn't request my review so I hope you don't mind me chiming in here...

I actually don't think it's the `before_action` that's the culprit here -- what you have in the actual code seems totally okay and worked well when I tried it locally! I actually think the tests are the issue, specifically, the url that you are passing in as the image url.

I took a look at the spec file, and it looks like the tests that are failing for you are the ones where you're passing in `expected_image_url`, and `expected_image_url` is set to `"https://dummyimage.com/300x300"` (notice that there is no file type extension at the end of that!).

I started looking into one of the failing specs (the test on line `230`) and checked what the regex itself was doing. Here's what I saw:
```ruby
➜  forem git:(juliannatetreault/validate-image-urls-in-site-config-249) ✗ rspec spec/requests/admin/configs_spec.rb:230
From: /Users/vaidehijoshi/Code/vaidehijoshi/forem/app/controllers/admin/configs_controller.rb:263 Admin::ConfigsController#valid_image_url:

    261: def valid_image_url(url)
    262:   binding.pry
 => 263:   url.match?(VALID_URL)
    264: end

[1] pry(#<Admin::ConfigsController>)> url
=> "https://dummyimage.com/300x300"
[2] pry(#<Admin::ConfigsController>)> VALID_URL
=> /\A(http|https):\/\/([\/|.|\w|\s|-])*.(?:jpg|gif|png)/
[3] pry(#<Admin::ConfigsController>)> url.match?(VALID_URL)
=> false
```

Looking at the above breakpoint, we can see that the regex is actually _not_ matching the `url` that is `"https://dummyimage.com/300x300"`. And that's because the URL from the test _itself_ isn't actually valid -- it doesn't end in `jpg`, `gif` or `png`, which is what the regex is `match`ing against.

When I saw that, I wanted to confirm my assumption. So, in the same breakpoint, I created a new dummy url that ended in `jpg`, and tried that against the regex:
```ruby
[4] pry(#<Admin::ConfigsController>)> this_should_be_valid_i_hope = "https://dummyimage.com/300x300.jpg"
=> "https://dummyimage.com/300x300.jpg"
[5] pry(#<Admin::ConfigsController>)> this_should_be_valid_i_hope.match?(VALID_URL)
=> true
```
So the regex is doing exactly what you want it to: it's confirming that the url does not end with the correct extension, so `valid_image_url` returns `false`.

If you change the tests from `expected_image_url = "https://dummyimage.com/300x300"` to `expected_image_url = "https://dummyimage.com/300x300.jpg"` (or `png` or `gif`), this test (and all the others!) will pass:
```ruby
➜  forem git:(juliannatetreault/validate-image-urls-in-site-config-249) ✗ rspec spec/requests/admin/configs_spec.rb:230
From: /Users/vaidehijoshi/Code/vaidehijoshi/forem/app/controllers/admin/configs_controller.rb:263 Admin::ConfigsController#valid_image_url:

    261: def valid_image_url(url)
    262:   binding.pry
 => 263:   url.match?(VALID_URL)
    264: end

[1] pry(#<Admin::ConfigsController>)> url
=> "https://dummyimage.com/300x300.jpg"
[2] pry(#<Admin::ConfigsController>)> VALID_URL
=> /\A(http|https):\/\/([\/|.|\w|\s|-])*.(?:jpg|gif|png)/
[3] pry(#<Admin::ConfigsController>)> url.match?(VALID_URL)
=> true
[4] pry(#<Admin::ConfigsController>)> continue
.

Finished in 19.08 seconds (files took 12.2 seconds to load)
1 example, 0 failures
```

TL;DR, the url you are testing against, `expected_image_url = "https://dummyimage.com/300x300"`, actually _is invalid_, which is why it never updates (and your regex is doing exactly what you want it to 😉). If you update those urls to have a valid extension at the end, all your tests will pass locally.

If it were me, I would probably just pull out the valid image url into a variable:
```ruby
let(:valid_image_url) { "https://dummyimage.com/300x300.jpg" }
```
and then just reference that every single place that we're currently using the same string over and over:

```ruby
post "/admin/config", params: { site_config: { main_social_image: valid_image_url },
expect(SiteConfig.main_social_image).to eq(valid_image_url)
```
And you wanted, you could do the same thing with the invalid image urls that you added in this PR too, for clarity/consistency.

app/controllers/admin/configs_controller.rb
```diff
@@ -129,10 +129,26 @@ class ConfigsController < Admin::ApplicationController
...
+        secondary_logo_url
+        campaign_sidebar_image
+        mascot_image_url
+        mascot_footer_image_url
+        onboarding_logo_image
+        onboarding_background_image
+        onboarding_taskcard_image
+      ].freeze
+
+    VALID_URL = %r{\A(http|https)://([/|.|\w|\s|-])*.(?:jpg|gif|png)}.freeze
```
Ridhwana, line 145: @juliannatetreault quick one, do we not accept jpeg?

rhymes:
Great review @vaidehijoshi - the URL https://dummyimage.com/300x300 is actually a valid image though, as can be other extension less images on the web: we can't really check if a URL is an image without downloading the image, so I wonder if we should just validate that it's a URL, not an "image URL"

Ridhwana:
Oops @juliannatetreault, sorry I read @vaidehijoshi's review only once I pushed the spec changes that you asked me to have a glance at. We basically came to the same conclusion though and I fixed them by adding the .png to the URL's 😊

Regarding @rhymes comment: 
>the URL https://dummyimage.com/300x300 is actually a valid image though, as can be other extension less images on the web: we can't really check if a URL is an image without downloading the image, so I wonder if we should just validate that it's a URL, not an "image URL"

> I don't think we should enforce a specific type of URL for images. If we really want to make sure that something at the other end of a URL is an image, we can only download it and check its content type to make sure it's image/*

We can revert or keep the test changes should we decide to go with validating just the URL and not the image url 🙂

app/controllers/admin/configs_controller.rb
juliannatetreault: @Ridhwana I'm going to remove the part of the validation that checks explicitly for the image URL and leave the validation to just validate the URL per yours and @rhymes' suggestion, so this won't matter in a couple of minutes. 😊

app/controllers/admin/configs_controller.rb
```diff
@@ -129,10 +129,26 @@ class ConfigsController < Admin::ApplicationController
...
+        secondary_logo_url
+        campaign_sidebar_image
+        mascot_image_url
+        mascot_footer_image_url
+        onboarding_logo_image
+        onboarding_background_image
+        onboarding_taskcard_image
+      ].freeze
+
+    VALID_URL = %r{\A(http|https)://([/|.|\w|\s|-])*.[a-z]{2,5}(:[0-9]{1,5})?(/.*)?$}.freeze
```
juliannatetreault, line 145: This should check that the URL is valid without strictly checking that the URL is a valid _image URL_ like I was previously doing -- I have removed `.(?:jpg|gif|png)` and have added a check to ensure that the last part of the URL is valid instead. How do we feel about this?

cc @rhymes and @Ridhwana

spec/requests/admin/configs_spec.rb
```diff
@@ -231,42 +231,87 @@
...
                                           confirmation: confirmation_message }
           expect(SiteConfig.main_social_image).to eq(expected_image_url)
         end
 
+        it "updates main_social_image with a valid image" do
+          expected_image = "https://dummyimage.com/300x300"
+          post "/admin/config", params: { site_config: { main_social_image: expected_image },
+                                          confirmation: confirmation_message }
+          expect(SiteConfig.main_social_image).to eq(expected_image)
+        end
```
juliannatetreault, line 245: I added tests like this one on line `240` to ensure that an image is validated and can be uploaded even if it does _not_ contain a specific image URL (a URL ending in .`png`, for example).

cc @rhymes and @Ridhwana 😄

app/controllers/admin/configs_controller.rb
Ridhwana: @rhymes I remember you telling me in one of my PR's that we could parse the url and use `url.is_a?(URI::HTTP)`, like this https://github.com/forem/forem/blob/master/app/liquid_tags/reddit_tag.rb#L69-L73 instead of using a REGEX?

spec/requests/admin/configs_spec.rb
Ridhwana: This looks good! Thanks @juliannatetreault

app/controllers/admin/configs_controller.rb
rhymes: @juliannatetreault I think it's fine, unfortunately there's no builtin way to validate URLs and `URI.parse()` is not super reliable by itself withouth checking at least the scheme. 

One final note: the regexp should end with `\z`, not `$`, the reason is because `$` matches end of line, not end of string. See https://rubystyle.guide/#caret-and-dollar-regexp

rhymes (commented):
Thanks for the change @juliannatetreault!

Ridhwana (approved):

app/controllers/admin/configs_controller.rb
juliannatetreault: Thank you for the feedback! I've made this change, so it should be good-to-go pending your review @rhymes 😄

rhymes (approved):
LGTM

benhalpern (approved):

PR #11308, opened 2020-11-07
Database config #11273

### LINKED ISSUE #11273: Removal of env var value PRACTICALDEVELOPER_DATABASE_PASSWORD
Postgres production values don't conform to rails convention. It's also aiding malicious users in narrowing down the attack surface giving them DB name and user name. This current config makes me inclined to create a DB on my target production with the same name and user so I don't have to deviate the config. 

```
production:
  <<: *default
  database: PracticalDeveloper_production
  pool: <%= ENV['DATABASE_POOL_SIZE'] %>
  username: PracticalDeveloper
  password: <%= ENV['PRACTICALDEVELOPER_DATABASE_PASSWORD'] %>
```

I'd like to see production config conform to something like the development config. Making use of `DATABASE_URL`. 
And either making a new env variable for database name or consolidating that into the prostgres url.

```
development:
  <<: *default
  url: <%= ENV.fetch('DATABASE_URL', '') %>
  database: PracticalDeveloper_development
  variables:
      statement_timeout: <%= ENV.fetch('STATEMENT_TIMEOUT', 10_000).to_i %>
```

Or alternativly move away from DATABASE_URL and expand all the DB config into independant vars.

I'd be happy to work on this as a first issue (just set the system up locally yesterday)

ofhope:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

Proposed config for db. 

- Removing db name references
- Removing db user name references 
- Removing test db fallback to `DATABASE_URL` (this is risky if someone has their prod or qa/staging env set in a CI/CD pipeline while tests are run it could wipe a production db)
- Documenting env config in .env_sample

Name, db name, password, port, host can all be controlled via the database url.
`postgresql://[user[:password]@][netloc][:port][,...][/dbname][?param1=value1&...]`
https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING

Fixes #11273 

## Added tests?

- [ ] Yes
- [x] No, and this is why: Its an application environment configuration change.

## Added to documentation?

This is an inital Draft PR, documentation on configuring a DATABASE_URL will need to be updated before merging.

- [ ] Docs.forem.com
- [ ] README
- [ ] No documentation needed

##  Are there any post deployment tasks we need to perform?
Yes this is a **breaking change** for anyone running the previous config in production. The migration path is fiarly straight forward. A new `DATABASE_URL` env will need to be included in the production environement before deploying and optionally cleaning up `PRACTICALDEVELOPER_DATABASE_PASSWORD` after deployment.


.env_sample
```diff
@@ -40,6 +40,15 @@ export ELASTICSEARCH_URL="http://localhost:9200"
 # SSL config
 export FORCE_SSL_IN_RAILS="not applicable in dev"
 
+# Postgres DB
+# export DATABASE_URL="not applicable in dev. Set by docker-compose."
```
citizen428, line 44: This wording is maybe not ideal, not too many people actually use the docker-compose workflow, even on our team.

I'd probably also use actual comments for the explanations and examples for the values, e.g.

```
# Not applicable when using docker-compose
# See: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING
# export DATABASE_URL="postgresql://[user[:password]@][netloc][:port][,...][/dbname][?param1=value1&...]"
```

Maybe we could even set a default that will work for quite a few local installs, e.g.

```
export DATABASE_URL="psql postgres:///Forem_development"
```

config/database.yml
```diff
@@ -19,7 +19,7 @@ default: &default
   encoding: unicode
   # For details on connection pooling, see Rails configuration guide
   # http://guides.rubyonrails.org/configuring.html#database-pooling
-  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  pool: <%= ENV['DATABASE_POOL_SIZE'] || ENV.fetch('RAILS_MAX_THREADS', 5).to_i %>
```
citizen428, line 22: Maybe consistently use `.fetch` instead?

```suggestion
  pool: <%= ENV.fetch("DATABASE_POOL_SIZE") { ENV.fetch("RAILS_MAX_THREADS", 5).to_i } %>
```

(You'll notice our Rubocop config prefers double quotes, so we should be consistent in this regard too)

.env_sample
```diff
@@ -40,6 +40,15 @@ export ELASTICSEARCH_URL="http://localhost:9200"
...
+# export RAILS_MAX_THREADS="not applicable in dev. Set by docker-compose."
+# export DATABASE_URL_TEST="Differentiate test db from dev db. Test DB is wiped on each test run."
+ export DATABASE_NAME_TEST=PracticalDeveloper_test
```
citizen428, line 47: I newly added things we should use "Forem" instead of "PracticalDeveloper".

.env_sample
ofhope: Yeah good point. These aren't strictly necessary but thought it could be useful to include available config variables.

config/database.yml
ofhope: Whats the severity of deprecating one of these variables? I'm getting an error when using fetch on DATABASE_POOL_SIZE without a default. We can emulate existing behaviour by fetching with a default of nill 

`pool: <%= ENV.fetch("DATABASE_POOL_SIZE", nil) || ENV.fetch("RAILS_MAX_THREADS", 5).to_i %>`

Although having two config variables is confusing. Ideally this could be consolidated to

`pool: <%= ENV.fetch("DATABASE_POOL_SIZE", 5).to_i`

ofhope: Oh! never mind I just noticed the suggestion has a return block to handle this.

citizen428 (approved):
Thanks for all the changes @ofhope 👍

config/database.yml
```diff
@@ -64,10 +64,10 @@ development:
 # Do not set this db to the same as development or production.
 test:
   <<: *default
-  url: <%= ENV['DATABASE_URL_TEST'] || ENV['DATABASE_URL'] || ''%>
-  database: PracticalDeveloper_test<%= ENV['TEST_ENV_NUMBER'] %>
+  url: <%= ENV.fetch("DATABASE_URL_TEST", "postgres:///Forem_test") %>
```
mstruve, line 67: We need to update our `travis.yml` file with these updated variables in order for the CI to pass.

config/database.yml
```diff
@@ -86,11 +86,8 @@ test:
 # You can use this database configuration with:
 #
 #   production:
-#     url: <%= ENV['DATABASE_URL'] %>
+#     url: <%= ENV.fetch("DATABASE_URL") %>
 #
 production:
   <<: *default
-  database: PracticalDeveloper_production
-  pool: <%= ENV['DATABASE_POOL_SIZE'] %>
```
mstruve, line 94: Why did you choose to get rid of the pool size here for production?

mstruve (commented):
Thanks for the PR @ofhope! We need to update our travis.yml file to ensure the specs run for this branch. I also had one question about the choice to remove pool from the production options. Beyond that, this looks like its nearly ready to go!!!

config/database.yml
ofhope: Awesome thanks for the info, I hadn't looked into why the build was failing.

config/database.yml
ofhope: Production extends the default config in database.yml which has the pool configured or deafults to 5.
Previously (current master) has two config variables to effect pool size.
The default config `pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>`
and production config `pool: <%= ENV['DATABASE_POOL_SIZE'] %>` which if null would default to RAILS_MAX_THREADS or 5.

The new config in default inherited by production is trying to maintain the same behaviour with 
`pool: <%= ENV.fetch("DATABASE_POOL_SIZE") { ENV.fetch("RAILS_MAX_THREADS", 5).to_i `

ofhope:
Sorry for the delay on this PR change. I got distracted at work.

I ended up including `database: <%= ENV.fetch("DATABASE_NAME", "Forem_production") %>` which is used for the rails task `rails db:create`. I didn't realise test-console-check uses production config.

mstruve:
@maestromac can I hand this off to you to make sure it gets the attention it needs to get the final mile?

.env_sample
```diff
@@ -39,6 +39,19 @@ export OPENRESTY_URL=""
 # SSL config
 export FORCE_SSL_IN_RAILS="not applicable in dev"
 
+# Postgres DB
+export DATABASE_NAME="Forem_development" # Used when creating a db instance `rails db:create`
+export DATABASE_URL="postgresql://forem:forem@localhost:5432/$DATABASE_NAME"
```
rhymes, line 44: These require the `forem` role to be created, which is not true by default, `rails db:create` doesn't create roles.

and by default you get:

```shell
psql: error: FATAL:  role "forem" does not exist
```

I think we should remove the `username:password` combo

I think we should also make it explicit that `DATABASE_URL` will always take precedence on `DATABASE_NAME` if they are mismatched:

```shell
➜  forem git:(master) DATABASE_URL=postgres://localhost/PracticalDeveloper_test rails c -e development
[1] pry(main)> ActiveRecord::Base.connection.current_database
=> "PracticalDeveloper_test"
[2] pry(main)> Rails.configuration.database_configuration["development"]["database"]
=> "PracticalDeveloper_development"
```

as you can see I should be connected to `PracticalDeveloper_development` becase that's what `database:` in the file `database.yml` file says, but instead I'm connected to `PracticalDeveloper_test` because that's what the environment variable says

.env_sample
```diff
@@ -39,6 +39,19 @@ export OPENRESTY_URL=""
...
+export RAILS_MAX_THREADS=5
+export DATABASE_NAME_TEST=Forem_test
+export DATABASE_URL_TEST="postgresql://forem:forem@localhost:5432/$DATABASE_NAME_TEST" # Differentiate test db from dev db. Test DB is wiped on each test run.
```
rhymes, line 47: same issue about requiring a usernme and password

config/database.yml
```diff
@@ -19,18 +19,18 @@ default: &default
   encoding: unicode
   # For details on connection pooling, see Rails configuration guide
   # http://guides.rubyonrails.org/configuring.html#database-pooling
-  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  pool: <%= ENV.fetch("DATABASE_POOL_SIZE") { ENV.fetch("RAILS_MAX_THREADS", 5).to_i } %>
```
rhymes, line 22: shouldn't `DATABASE_POOL_SIZE` be present in the `.env_sample` as well? If we decide to keep the separate env var we might also need to explain that it should generally have the same value as `RAILS_MAX_THREADS` (which is why Rails by default doesn't introduce a separate env var). See https://github.com/rails/rails/pull/23528 as to why.

@mstruve @maestromac are we using at all `DATABASE_POOL_SIZE`? Note that's not present in the default Rails DB configuration https://github.com/rails/rails/blob/6-1-stable/railties/lib/rails/generators/rails/app/templates/config/databases/postgresql.yml.tt

config/database.yml
```diff
@@ -64,10 +64,10 @@ development:
 # Do not set this db to the same as development or production.
 test:
   <<: *default
-  url: <%= ENV['DATABASE_URL_TEST'] || ENV['DATABASE_URL'] || ''%>
-  database: PracticalDeveloper_test<%= ENV['TEST_ENV_NUMBER'] %>
+  url: <%= ENV.fetch("DATABASE_URL_TEST", "postgres:///Forem_test") %>
+  database: <%= ENV.fetch("DATABASE_NAME_TEST", "Forem_test") %><%= ENV.fetch("TEST_ENV_NUMBER", "") %>
```
rhymes, line 68: no need to use `.fetch` for `TEST_ENV_NUMBER` if the default is an empty string, `nil` will be converted to empty string by `<%=`

config/database.yml
```diff
@@ -86,11 +86,9 @@ test:
...
-#     url: <%= ENV['DATABASE_URL'] %>
+#     url: <%= ENV.fetch("DATABASE_URL") %>
 #
 production:
   <<: *default
-  database: PracticalDeveloper_production
-  pool: <%= ENV['DATABASE_POOL_SIZE'] %>
-  username: PracticalDeveloper
-  password: <%= ENV['PRACTICALDEVELOPER_DATABASE_PASSWORD'] %>
+  url: <%= ENV.fetch("DATABASE_URL", "") %>
```
rhymes, line 93: same, `.fetch` is convenient if there's a valued default or if we want to raise the exception if the value is missing, but if we default to an empty string, then there's no need for `.fetch`

rhymes (changes requested):
Hi @ofhope, thank you! Sorry I circled back on this PR this late. I left some notes as I think we might want to simplify the configuration a bit.

What if we just remove the fallback for test and encourage people to use the URL with different samples for dev, test and production?

Also we need to be sure that production forems aren't just using `DATABASE_URL` in production as I suspect, which would probably should be a thing we encourage instead of keeping the separate values

I wonder if the `database.yml` file shouldn't eventually just include the variables (pool size and statement timeout) and that's it. All configurations be something like:

```yaml
test:
  <<: *default
  url: ENV.fetch("TEST_DATABASE_URL")
  variables:
      statement_timeout: <%= ENV.fetch('STATEMENT_TIMEOUT', 10_000).to_i %>
```

using the URL, what do you think?

config/database.yml
ofhope: It's a great point. If possible I think it'd be simpler to drop DATABASE_POOL_SIZE from the .env_sample. 
Looking through the Rails source for `6.2.0` DATABASE_POOL_SIZE isn't used. 

I'll add `WEB_CONCURRENCY` to the env_sample as well.

ofhope: It could be worth noting the documention from Heroku too

> As you add more concurrency to your application, it will need more connections to your database. A good formula for determining the number of connections each application will require is to multiply the RAILS_MAX_THREADS by the WEB_CONCURRENCY. This combination will determine the number of connections each dyno will consume. 

https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#database-connections

.env_sample
ofhope: @rhymes I removed the username:password combo. And tried to clear in the .env_sample in one line summarisation.

Should I add more detailed documentation as apart of this PR? Maybe in config-env.md

config/database.yml
rhymes: I agree that we should drop `DATABASE_POOL_SIZE`. Let's stay as close as possible to Rails and Puma defaults.

BTW it's also possible to change the pool size by changing the URL:

```shell
> DATABASE_URL=postgres://localhost/PracticalDeveloper_development?pool=10  rails c
[1] pry(main)> ActiveRecord::Base.connection_pool.stat
=> {:size=>10,
 :connections=>0,
 :busy=>0,
 :dead=>0,
 :idle=>0,
 :waiting=>0,
 :checkout_timeout=>2.0}
```

:)

maestromac: @rhymes sorry I missed this. yes we are using `DATABASE_POOL_SIZE` on production.

rhymes: ok cool, let's keep it then :)

jdoss (approved):
LGTM from @forem/systems end.

jdoss:
@maestromac I am not sure if this will impact how we do things on Heroku or not. PTAL.

rhymes (approved):
let's do this, bar @forem/sre having the final say

config/database.yml
```diff
@@ -19,18 +19,18 @@ default: &default
...
   checkout_timeout: 2
   variables:
-      statement_timeout: <%= ENV.fetch('STATEMENT_TIMEOUT', 2500).to_i %>
+      statement_timeout: <%= ENV.fetch("STATEMENT_TIMEOUT", 2500).to_i %>
 
 development:
   <<: *default
-  url: <%= ENV.fetch('DATABASE_URL', '') %>
-  database: PracticalDeveloper_development
+  url: <%= ENV.fetch("DATABASE_URL", "postgres:///Forem_development") %>
```
maestromac, line 30: quick question: how come the backup to the ENV fetch here doesn't have localhost and port, as supposed to `export DATABASE_URL="postgresql://localhost:5432/$DATABASE_NAME`?

maestromac (approved):
LGTM. I'm going to get a final approval from @benhalpern

config/database.yml
rhymes: I think because `localhost:5432` is the default:

```shell
$ psql postgres:///PracticalDeveloper_development
psql (13.1)
Type "help" for help.

PracticalDeveloper_development=#
```

msarit:
Hi @ofhope 👋🏾  Thanks for your work on this PR! Please resolve the conflicts 🙏🏾 

@benhalpern Looks like we need your final weigh-in on this PR 🙏🏾

aitchiss:
> LGTM. I'm going to get a final approval from @benhalpern

@benhalpern - any chance you could take a look at this please and let us know if we're good to go?

maestromac:
@aitchiss I will handle this

PR #11343, opened 2020-11-09
Update shell version when admin action taken

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When an admin makes a change, it will be immediately reflected across the site in general, but for existing users with a cached shell top (cached by service workers) will not immediately get the update.

This makes appropriate changes such that a config change will ensure users fetch the new shell. This will help admin UX because they will see the new stuff right away and not think of this as a bug, as it would now with their own page not clearing the cache right away.

Unrelated line I fixed: Removed an old `fb:pages` hardcoded meta tag we no longer user or need. (If for some reason we still need this, we can add it back in as _generalized_ via config, rather than the hardcoded DEV version).

### DIFF
--- a/app/controllers/admin/application_controller.rb
+++ b/app/controllers/admin/application_controller.rb
@@ -17,6 +17,7 @@ module Admin
       CacheBuster.bust("/tags/onboarding") # Needs to change when suggested_tags is edited.
       CacheBuster.bust("/shell_top") # Cached at edge, sent to service worker.
       CacheBuster.bust("/shell_bottom") # Cached at edge, sent to service worker.
+      CacheBuster.bust("/async_info/shell_version") # Checks if current users should be busted..
       CacheBuster.bust("/onboarding") # Page is cached at edge.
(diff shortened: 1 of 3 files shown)


mstruve (approved):

app/controllers/admin/application_controller.rb
```diff
@@ -17,6 +17,7 @@ def bust_content_change_caches
       CacheBuster.bust("/tags/onboarding") # Needs to change when suggested_tags is edited.
       CacheBuster.bust("/shell_top") # Cached at edge, sent to service worker.
       CacheBuster.bust("/shell_bottom") # Cached at edge, sent to service worker.
+      CacheBuster.bust("/async_info/shell_version") # Checks if current users should be busted..
```
atsmith813, line 20:
```suggestion
      CacheBuster.bust("/async_info/shell_version") # Checks if current users should be busted.
```

atsmith813 (approved):

PR #11347, opened 2020-11-09
Improve information architecture of Settings

lisasy:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
Improve the information architecture of Settings by:
- Renaming categories
- Moving sections around so it better fits user expectations.
[Design](https://www.figma.com/file/3fYayQ93vwmueffgDzPeVN/Settings-IA-update-settings-ia?node-id=1%3A123)

Note: There are small tweaks that could be made, but I'm holding them off in this PR. That includes: adding icons to the navigation links, updating the content so that it's clearer, updating `Organization` view, and improving UX for some of the content cards.

After this PR, some of the possible follow-up optimizations we could make might be:
- Improve the overall UX of cards for features that people need to "turn on" so we don't show everything up front (e.g. Comments template)
- Improve the Organizations section & UX

Screen Shot 2020-11-11 at 3 34 36 PM
Screen Shot 2020-11-11 at 6 02 30 PM

## Added tests?

- [ ] Yes
- [x] No, and this is why: I didn't add anything new, but I may need to update tests based on the changes that I made.

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed


app/views/users/_extensions.html.erb
```diff
@@ -0,0 +1,53 @@
...
+      <%= f.label :payment_pointer, "Payment Pointer", class: "crayons-field__label" %>
+      <%= f.text_field :payment_pointer, class: "crayons-textfield", placeholder: "$pay.somethinglikethis.co/value" %>
+    </div>
+    <%= f.hidden_field :tab, value: @tab, id: nil %>
+    <div><button type="submit" class="crayons-btn">Save Web Monetization Settings</button></div>
+  <% end %>
+</div>
+
+<%= render partial: "integrations" %>
+<%= render partial: "publishing_from_rss" %>
```
lisasy, line 53: I'm adding partials into the views that used to be their own views, but there are 2 things that aren't rendering:

1. Within the existing`integrations` view, the `integrations_github_repositories` isn't rendering.
2. When I try to add `<%= render partial: "response_templates" %>` here, the entire view display an error page.

I'm curious if there is something within those 2 views that need to be specified if I render them as partials. Can you help me here? Thank you!

app/lib/constants/settings.rb
```diff
@@ -2,15 +2,14 @@ module Constants
...
-      Publishing\ from\ RSS
+      Account
+      Billing
+      Integrations
       Organization
       Response\ Templates
-      Billing
-      Account
-      Misc
+      Extensions
```
rhymes, line 12: ❤️ love this new category

app/lib/constants/settings.rb
```diff
@@ -2,15 +2,14 @@ module Constants
...
       Profile
-      UX
-      Integrations
+      Customization
       Notifications
-      Publishing\ from\ RSS
+      Account
+      Billing
+      Integrations
       Organization
```
rhymes, line 10: As we can join multiple organizations, should this be finally be plural :D ?

app/views/users/_extensions.html.erb
rhymes: Yes, that's because the logic to make them work is inside the controller. There's a 1:1 correspondence between the name of the section, so for example `extensions` in this case and the logic in the controller:

https://github.com/forem/forem/blob/824e9a13cb1c26d83e23a33f368da5579529e968/app/controllers/users_controller.rb#L231-L246

All the tabs that need special data to be loaded in advance are handled here. I think you can just adapt the `case...when` statement that calls the various functions. Let me know if you need help with that!

rhymes (commented):
Looks cool @lisasy! Left some feedback inline

lisasy:
@ludwiczakpawel It looks a bit complicated to me to add the Twemojis because it might requiring touching some files that I'm not comfortable with. It's not a high-priority at all, but if it's a straightforward task for you, would you be interested in trying to do this?

rhymes (commented):
Thanks @lisasy for the changes!

Now that we reorganized the IA of this, I think http://localhost:3000/settings/integrations should be removed as it's not accessible from the sidebar menu anymore, same for the 

```ruby
    when "integrations"
      handle_integrations_tab
```

part in app/controllers/users_controller.rb on line 239 and the `app/views/users/_integrations.html.erb` file. The Stackbit integration is present in http://localhost:3000/settings/extensions

One question related to the re-architecture:

Screenshot_2020-11-12 Settings - DEV Community 👩‍💻👨‍💻

As both monetization's and stackbit's sections are marked as beta, I was wonder if we could move the RSS part be moved before that in the sequence. Does it make sense?

ludwiczakpawel:
> It looks a bit complicated to me to add the Twemojis because it might requiring touching some files that I'm not comfortable with.

no problem, i can take care of it. have you picked right twemojis for these pages?

lisasy:
@ludwiczakpawel is this table below helpful for you? I don't know if you want me to find the specific emoji names, or if referencing them here as emojis works. 

|Tab|Emoji|
|---|---|
|Profile|😀
|Customization|⚙️
|Notification|📭
|Account| 🌱
|Billing|💳
|Organization| 🏢
|Extensions|⚡️

ludwiczakpawel:
I've pushed the icons. 

image


It's not an ideal way of doing so tbh because if you change name of the tab, you will also have to change name of particular twemoji :C But I have no better idea to do it, maybe some dev could chime in if it's worth it.

app/views/users/_profile.html.erb
```diff
@@ -77,6 +77,8 @@
     </div>
   <% end %>
 
+  <%= render partial: "users/integrations_github_repositories", locals: { show_integration: @github_repositories_show } %>
```
ludwiczakpawel, line 80: @lisasy  It's quite weird we show github repos here - shouldn't that be part of "Extensions"?

lisasy:
> I've pushed the icons.
> 
> image
> 
> It's not an ideal way of doing so tbh because if you change name of the tab, you will also have to change name of particular twemoji :C But I have no better idea to do it, maybe some dev could chime in if it's worth it.
> 
> Also, is this ready to be reviewed or should this still be a draft?

Thank you! 😇

app/views/users/_profile.html.erb
lisasy: Initially, this has always been apart of the "Integrations" tab, but hear me out: as I was editing the title + subtitle of the this card, I realized that the goal of this setting is so that users who have linked GitHub to their Forem account can _pin their repositories to their profile page_. That's why I wanted to put it in "Profile" because that's what this setting impacts.

This is at the same time a type of "integration", yes, but the value it provides to people might be more relevant in Profile. I could imagine in the future, there might be similar "integrations" like this... What do you think?

ludwiczakpawel: I'm not sure TBH, mostly because as you know there's a plan to split profile into its own sections (tabs) - remember that design where we had "Overview" as a profile view - it would include all the profile details as well as "entry points" for all kinds of subpages for profile - like `https://dev.to/lisasy/posts` or `https://dev.to/lisasy/comments` or `https://dev.to/lisasy/gallery` (if we have extension for galleries one day) or `https://dev.to/lisasy/repos` and so on. I imagine "GitHub Repos" will definitely  be displayed on the profile overview but only as an entry point to `https://dev.to/lisasy/gtihub-repos` subpage if you know what I mean. Therefore I think it's "settings" should be placed within "Extensions" tab, not the "Profile". 

I do understand the idea of _extensions extending the profile view_, but I still feel like they should be contained within "Extensions" page at least until we have better understanding and guidelines around WHAT can be placed WHERE in the overall IA. 

Would you agree with this sentiment? Does that make sense?

lisasy: @ludwiczakpawel That totally makes sense -- thanks for reminding me of that approach we want to move towards.

I'll move the Github repos back to extensions and make a commit!

rhymes:
@lisasy I sent a commit to fix the failing specs, can't wait to see this restructuring live!

lisasy:
@rhymes thank you so much!

rhymes (approved):
LGTM!

msarit (approved):
Everything looks good! But, re: [Rhymes' comment](https://github.com/forem/forem/pull/11347/files#r520514079), did we change "Organization" to "**Organizations**" in `app/lib/constants/settings.rb:10`?

lisasy:
> Everything looks good! But, re: [Rhymes' comment](https://github.com/forem/forem/pull/11347/files#r520514079), did we change "Organization" to "**Organizations**" in `app/lib/constants/settings.rb:10`?

No, I didn't change it here. I think this is a detail we don't need to include right now, and there are additional incremental updates we want to make in the `Organization` page overall. 

Thanks for the review @msarit :)

PR #11350, opened 2020-11-10
Update data update script

citizen428:

- [X] Refactor
- [X] Bug Fix

## Description

In #11246 I added a data update script to do two things:

1. Ensure all current DEV profile fields are present so we can drop the columns from `User` and rely on delegation.
2. Migrate the data one more time just in case and compact it (remove `nil` values) while doing so.

The second point wasn't strictly necessary, but it felt better to do this before we drop columns and data is gone for good. Alas the process errored out without any exception **and** the call to `compact` was in the wrong place, so it isn't even easy to see which records were process already. Yeah, me! 👏 🙄 

This PR does the following:

1. Puts the call to `compact` in the correct place. 
2. Explicitly logs errors to Honeycomb and continues after an error.
3. Stores the last processed user id in the Rails cache so if things terrminate again we know how far we got.

#6994 

## Added tests?

- [X] Yes, updating the existing one. Looking back I should have noticed that the `compact` was in the wrong place.

## Added to documentation?

- [X] No documentation needed

No, but there is a pre-deployment task. To ensure the data update script can run again, remove its entry from any Forem instance where you want to run the script again:

```ruby
DataUpdateScript.delete_by(file_name: "20201103050112_prepare_for_profile_column_drop")
```

image

### DIFF
--- a/lib/data_update_scripts/20201103050112_prepare_for_profile_column_drop.rb
+++ b/lib/data_update_scripts/20201103050112_prepare_for_profile_column_drop.rb
@@ -1,5 +1,7 @@
 module DataUpdateScripts
   class PrepareForProfileColumnDrop
+    HONEYCOMB_PREFIX = "data_update_20201103050112".freeze
+
     def run
       # Make sure all current DEV profile fields exist. The import task is
       # idempotent so we don't need further checkes here.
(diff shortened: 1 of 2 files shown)


Ridhwana (approved):
This looks good. 

@citizen428 I assume that this will only affect new communities who will still run this script for the first time, but existing communities will not receive this change as the data script will not be run again. Is this correct?

mstruve (approved):

citizen428:
@Ridhwana As I'm updating the existing script, this shouldn't run again on any communities where it already ran, because it's in the `DataUpdateScript` table with a status of either `succeeded` or `failed`.

citizen428:
@mstruve I know you finished running this on DEV yesterday. How about the other communities, did it finish everywhere? Would you mind merging this, so you can ensure that it re-runs on communities that may need it (if any)?

mstruve:
> How about the other communities, did it finish everywhere?

Before merging this I think I am going to try and put in place an alert system for Failed DUS scripts bc having to check every environment is not going to scale. We already have more than 10 and I am dreading it. https://github.com/orgs/forem/projects/6#card-49151524

benhalpern:
@mstruve looks like the other PR for the observability has reviews, so you can merge this one once you're in a good place to merge the other, eh?

mstruve:
👍  that's the plan, just merged the other one. Will wait and see what it returns before proceeding with this

mstruve:
This ran successfully on all of the Forems. I'll merge Monday morning and that's all we have to do :)

Issue #82, opened 2017-08-28
Header image higher on main page


kay-is:
The header image seems to be a tiny bit higher on the main page than on the actual post page it belongs to, at least on mobile.

So when you are opening the first post, the title moves up a bit.

benhalpern:
Hey I just realized this was fixed with the new design (about a month back). 🎉

PR #11354, opened 2020-11-10
Add podcast_appearances joined table (forem#82)

NenadPantelic:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Created m2m table `podcast_appearances` which connects users and podcast episodes. Added corresponding `has_many` associations. Added podcast appearance pundit policy and controller with CRUD options.
These changes are part of the MLH issue [#82](https://github.com/MLH-Fellowship/dev.to/issues/82)

_These changes should be tested strictly through Rails console._

## Added tests?

- [x ] Yes
- [ ] No 

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

No


benhalpern:
lgtm

NenadPantelic:
@benhalpern, @mkcode I committed other functionalities (read, update and delete). I suppose the only thing left is to add tests? What do you think about this code?

app/controllers/podcast_appearances_controller.rb
```diff
@@ -0,0 +1,123 @@
+class PodcastAppearancesController < ApplicationController
```
rhymes, line 1: Hi @NenadPantelic, thanks for the changes. I recommend, see also the title of this PR, to separate the the table code from the controller code (policy included) and submit the latter separately once the table is merged. Also because this controller doesn't have neither tests nor templates.

app/models/podcast_appearance.rb
```diff
@@ -0,0 +1,11 @@
+class PodcastAppearance < ApplicationRecord
+  belongs_to :user
+  belongs_to :podcast_episode
+  validates :podcast_episode_id, uniqueness: { scope: :user_id }
+  validates :podcast_episode_id, :user_id, :role, presence: true
+  validates :role, inclusion: { in: %w[owner host guest], message: "%<value> is not a valid role" }
```
rhymes, line 6: is `owner` necessary if https://github.com/MLH-Fellowship/dev.to/issues/81 and https://github.com/forem/forem/pull/11410 are to be merged?

I guess we can find the list of a podcast owners through that if I'm not mistaken about the intent here

app/models/podcast_appearance.rb
```diff
@@ -0,0 +1,11 @@
...
+
+  def podcast_episode_creator_id
+    podcast_episode.podcast.creator_id
+  end
```
rhymes, line 10: as this seems to be only necessary in the policy, I wouldn't add it as a method here

config/routes.rb
```diff
@@ -251,6 +251,8 @@
   resources :podcasts, only: %i[new create]
   resources :article_approvals, only: %i[create]
   resources :video_chats, only: %i[show]
+  # TODO: refactor this so that URL is nested inside that podcast_episode resource
+  resources :podcast_appearances, only: %i[index show new create edit update destroy]
```
rhymes, line 255: this also should be deferred to the same separate PR

rhymes (changes requested):
Hi @NenadPantelic, I left some notes about splitting this in two PRs.

I would also suggest add some basic testing to `podcast_appearance` and checking the new relations added to `podcast_episode` in its tests.

You can take those as reference: 

https://github.com/forem/forem/blob/910f79e04e402214f33600cddb44367e9c6e5d45/spec/models/podcast_episode_spec.rb#L7-L19

NenadPantelic:
Hi @rhymes. I'll write here just to summarize and have it in one place. 

1. About PR splitting:  Yeah, I suppose I should separate commits into db related and controller related. Another way would be to rename the PR title, but I'd say that for atomicity it's better to split PRs into two :smiley: 
2. Testing: I will definitely add some (model) tests, that's the only thing left here
3. > is `owner` necessary if [MLH-Fellowship#81](https://github.com/MLH-Fellowship/dev.to/issues/81) and #11410 are to be merged? I guess we can find the list of podcast owners through that if I'm not mistaken about the intent here

I discussed this with @benhalpern and @mkcode. We defined those roles initially, but of course, after podcast ownership is defined, it will be completely redundant to have the role `owner`.
4.  
> as this seems to be only necessary in the policy, I wouldn't add it as a method here

You're right. At that moment I was not sure If I'll need that later, so I created it as model property, but I will move it to the policy class.
5. I'll first deal with that model/db things -> merge conflicts and add tests. 
6. For the controller part, I created some dummy views to test the complete logic manually, but I haven't added it - that's the part of this [issue](https://github.com/MLH-Fellowship/dev.to/issues/86)
Thanks for the review :blush:

rhymes:
@NenadPantelic thanks for the explanation! :-)

NenadPantelic:
@rhymes @benhalpern @mkcode I've separated files  (I'll have two PRs for this issue - model/db and controller/pundit). Can you please check my commits?

app/models/podcast_appearance.rb
```diff
@@ -0,0 +1,11 @@
+class PodcastAppearance < ApplicationRecord
```
rhymes, line 1: should this be renamed to `PodcastEpisodeAppearance`? After all it's the join of users and podcast episodes, not podcasts

app/models/podcast_appearance.rb
rhymes: as mentioned, this should be removed as it was only used for the policy

db/migrate/20201107111600_create_podcast_appearances.rb
```diff
@@ -0,0 +1,13 @@
+class CreatePodcastAppearances < ActiveRecord::Migration[6.0]
+  def change
+    create_table :podcast_appearances do |t|
+      t.references :user, null: false, foreign_key: true
+      t.references :podcast_episode, null: false, foreign_key: true
```
rhymes, line 5: this should both have `index: false` so not to create an additional index, since they are already indexed by the unique index on line 11

rhymes (changes requested):
Hi @NenadPantelic, thanks a lot for the changes! It occurred to me that the model should be called `PodcastEpisodeAppearance` instead of `PodcastAppearance`, because we're actually tracking the former. What do you think?

NenadPantelic:
@rhymes I have a question: my rspec model fails because in `PodcastEpisodeAppearances` I have association with `User` model. This [migration](https://github.com/forem/forem/blob/master/db/migrate/20201009040438_remove_twitch_columns_from_user.rb) removes the `twitch_username` column which is still used in the model code (being validated). I suggest commenting out those portions of the `User` model code in order to commit my changes. I'll leave notes that that should be removed/refactored.

rhymes:
@NenadPantelic that's odd, `twitch_username` is definitely unused anymore. I just grepped the latest master and this is the result:

```ruby
db/migrate/20190420000607_add_twitch_columns_to_user.rb
4:      t.string :twitch_username, index: true, unique: true

db/migrate/20201009040438_remove_twitch_columns_from_user.rb
5:      remove_column :users, :twitch_username
```

It's also odd becaues that column was removed in October, so unless your `master` copy hasn't been updated since then you shouldn't have this issue.

NenadPantelic:
@rhymes I solved the problem with the master sync and updated the things you suggested. I think this is ready for review.

NenadPantelic:
I added a small change - renamed composite index on a podcast episode and user to follow the default naming pattern Rails provides (I'm a fan of proper naming and code style). :smile:  I think this PR is done. @benhalpern @rhymes what do you think?

db/migrate/20201107111600_create_podcast_episode_appearances.rb
```diff
@@ -0,0 +1,18 @@
+class CreatePodcastEpisodeAppearances < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
```
rhymes, line 2: we shouldn't disable transactions during creation of new tables, if anything goes wrong, the migration could lead the DB in a corrupt state. I believe you added this so to be able to add `algorithm: concurrently` to `add_index` below, but there's no necessity to add an index on an empty table concurrently :-) `algorithm: concurrently` main purpose is not to slow down table access while the index is being created, but the table is new and empty by the time the index is added

db/migrate/20201107111600_create_podcast_episode_appearances.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.references :user, null: false, foreign_key: true, index: false
+      t.references :podcast_episode, null: false, foreign_key: true, index: false
+      t.string :role, null: false, default: "guest"
+      t.boolean :approved, null: false, default: false
+      t.boolean :featured_on_user_profile, null: false, default: false
+      t.timestamps
+    end
+    add_index :podcast_episode_appearances,
+              %i[podcast_episode_id user_id],
+              unique: true, algorithm: :concurrently,
```
rhymes, line 15: remember to remove `algorithm: concurrently`

spec/models/podcast_episode_spec.rb
```diff
@@ -9,6 +9,7 @@
 
       it { is_expected.to belong_to(:podcast) }
       it { is_expected.to have_many(:comments).inverse_of(:commentable).dependent(:nullify) }
+      it { is_expected.to have_many(:podcast_episode_appearances).dependent(:destroy) }
```
rhymes, line 12: this is missing the check for `  has_many :users, through: :podcast_episode_appearances` I think

spec/models/podcast_episode_appearance_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    it do
+      expect(podcast_episode_appearance).to belong_to(:user).class_name("User").inverse_of(:podcast_episode_appearances)
+        .inverse_of(:podcast_episode_appearances)
+    end
+
+    it { is_expected.to belong_to(:podcast_episode) }
+
+    it { is_expected.to validate_presence_of(:podcast_episode_id) }
+    it { is_expected.to validate_presence_of(:user_id) }
```
rhymes, line 17: missing the presence check and inclusion validation for `role`

rhymes (changes requested):
Thanks for the changes! The major thing we need to change is the note I left in the migration. I left feedback in the review for that and the rest.

db/migrate/20201107111600_create_podcast_episode_appearances.rb
NenadPantelic: Yeah, I totally forgot that I dumped some data to test some things. That's the reason I added concurrently. Sorry :sweat_smile:

spec/models/podcast_episode_appearance_spec.rb
NenadPantelic: @rhymes Can you give me a hand? I'm constantly getting:
`# PodcastEpisodeAppearance validations is expected to validate that :role cannot be empty/falsy`, even though I've added ` role { "guest" }` into factory bot class. It returns:
```ArgumentError: invalid value for Integer(): "". ```
 I've also tried `  let(:podcast_episode_appearance) { create(:podcast_episode_appearance, role: "guest") }` but it gives the same result. I've checked - factory bot gives me a guest value for the role of instance it created.

Comment (rhymes) [reply spec/models/podcast_episode_appearance_spec.rb:17]: I found the solution, it was a weird error message indeed. It doesn't have to do with `shoulda-matchers` but I was able to reproduce it locally in the console:

```ruby
> pea = PodcastEpisodeAppearance.new
> pea.role = "foobar"
> pea.validate!
ArgumentError: invalid value for Integer(): "foobar"
from /forem/vendor/bundle/ruby/2.7.0/gems/i18n-1.8.5/lib/i18n/interpolate/ruby.rb:34:in `sprintf'
```

Basically the issue is the error message that's lacking a `s` for the string interpolation.

It should be:

```ruby
validates :role, inclusion: { in: %w[host guest], message: "%<value>s is not a valid role" }
```

and not

```ruby
validates :role, inclusion: { in: %w[host guest], message: "%<value> is not a valid role" }
```

NenadPantelic: Really weird. I google so many things about `shoulda-matchers` and inclusion validation. I've even tested many things about FactoryBot instance creation, and at the end, the error is so silly. It's not verbose and descriptive at all. Thanks for your help :clap:

NenadPantelic:
I've changed things @rhymes suggested. Can you take a look at this now? @rhymes @benhalpern

spec/models/podcast_episode_spec.rb
rhymes: can you address this as well @NenadPantelic?

NenadPantelic: @rhymes I've done that

rhymes: sorry, didn't see it! Thanks!

rhymes (approved):
LGTM!

mstruve (approved):
Thanks for breaking down this larger feature into smaller PRs!!! It makes our lives as reviewers SO MUCH easier! 

This new table LGTM!

NenadPantelic:
@rhymes @mstruve Thanks :smile:

NenadPantelic:
Note: sorry for the inconvenience - this PR is related to MLH's [#82](https://github.com/MLH-Fellowship/dev.to/issues/82) not the domestic #82

PR #11357, opened 2020-11-10
Profile UI update

ludwiczakpawel:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR introduces a slightly updated layout and UI for user's profile. The idea is to consolidate left and right sidebars into one basically. This is a temporary (or not?) solution until we're done with rethinking IA for entire user view.

Just click around.

image

## Added tests?

- [ ] Yes
- [x] No, not yet because I'm not sure if any are needed.

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed


rhymes:
I set it as draft as apparently you're still working on it :)

ludwiczakpawel:
@rhymes hah yea, i was pushing some "fixes" i've noticed.. it should now be good for review :)

lisasy:
Will you be putting this section into a `crayons-card` as well?

    5 posts published
    21 comments written
    3 tags followed

lisasy:
Screen Shot 2020-11-10 at 10 09 55 AM

I had an idea for the placement of the "More info" button on small screen when we collapse the left sidebar. What if we placed it into the top header section? I think from a visual/info arch perspective, this might create a cleaner divide between information about the user vs. the user's posts/comments.

We'd still have the same interaction with the cards displaying below, so there would need to be some tweaking in CSS/JS to make this work. What do you think?

lisasy (approved):
Overall this looks good to me. I left two comments.

fdoxyz (commented):
I took this for a spin locally and it's looking great! ✨

The only detail I noticed was the same as Lisa's suggestion from above:

> Will you be putting this section into a crayons-card as well?

```
5 posts published
21 comments written
3 tags followed
```

Maybe it's my personal preference but I think I'd like it a little better even just without the `<ul>` decorator (bullet points), similar to how it currently looks.

Side note: I'm restarting the build because I'm not sure of what these errors mean. There was a recent NodeJS update so after Travis cache was cleared we may get them to pass, otherwise we may need to merge master here to get the tests back to green

app/assets/javascripts/initializers/initializeUserProfilePage.js
```diff
@@ -33,3 +33,18 @@ function initializeUserProfilePage() {
     }
   }
 }
+
+function initializeProfileInfoToggle() {
```
rhymes, line 37: I don't understand how to trigger this :|

app/views/users/_github_repositories_area.html.erb
```diff
@@ -1,21 +1,37 @@
-<% repositories.each do |repo| %>
-  <a class="widget" href="<%= repo.url %>" target="_blank" rel="noopener">
-    <header>
-      <h4><span class="emoji"><img src="<%= asset_path("github-logo.svg") %>" alt="github logo"></span><%= repo.name %></h4>
+<% unless repositories.empty? %>
```
rhymes, line 1: non blocking: `if repositories.present?` is clearer :D

app/views/users/_organizations_area.html.erb
```diff
@@ -1,18 +0,0 @@
-<%# it's highly unlikely a single use will belong to so many organizations to slow down
-    this fetch, thus we can safely use the .present?/.each pattern to load them %>
```
rhymes, line 2: can we keep this comment? I saw it has been removed when you moved over the code

app/views/users/show.html.erb
```diff
@@ -177,23 +177,19 @@
...
-        </div>
-      </button>
-      <button class="on-page-nav-butt on-page-nav-butt-right" style="margin-top:-1px;" id="on-page-nav-butt-right" aria-label="nav-button-right">
-        <img style="height:28px;width:28px;" src="<%= asset_path "ribbon.svg" %>" alt="right-sidebar-nav">
-      </button>
-    </div>
+<div
+  class="crayons-layout crayons-layout--limited-l crayons-layout--2-cols crayons-layout--2-cols--1-2 pt-4 m:pt-0"
+  id="index-container"
+  data-params="<%= params.merge(user_id: @user.id, class_name: "Article", sort_by: "published_at", sort_direction: "desc").to_json(only: %i[tag user_id q class_name sort_by sort_direction]) %>" data-which="<%= @list_of %>"
```
rhymes, line 183: wonder if we can move the value of `data-params` up to the controller 👁️

rhymes (commented):
Looks cool, left a few notes

app/assets/javascripts/initializers/initializeUserProfilePage.js
ludwiczakpawel: It's being added here: https://github.com/forem/forem/pull/11357/files#diff-16747096d6c3ff6309080c7463dca64c498ec9c117509745d9999e0f85119b86R60

I don't think this is the ideal way of running/triggering JS but it seems like it's a commonly used method on DEV so I just sticked to it.

app/views/users/show.html.erb
ludwiczakpawel: Since I don't know the answer, it's probably not something you want me to do :D

ludwiczakpawel:
@rhymes, I fixed the things you've mentioned.

@lisasy, good idea about placing button inside header, should be done now! Also, I intentionally put the stats outside the crayons-card box. not sure why - subconsciously it looked better in my mind and I still like it :D But I'm happy to wrap it in box since sometimes profiles will be pretty empty and the stats box is the only one that will be displayed soo.. whatever.. i'll bring it back and I've also added few icons there to make it less boring. cc @fdoxyz 

@fdoxyz thanks for restarting the build!

app/assets/javascripts/initializers/initializeUserProfilePage.js
rhymes: No I mean, as a user, how to I trigger that? I tried clicking on it but nothing happens

app/views/users/_sidebar.html.erb
```diff
@@ -75,16 +75,19 @@
...
+      <div class="flex items-center mb-4">
+        <%= inline_svg_tag("post.svg", aria: true, class: "crayons-icon mr-3 color-base-50", title: "Post") %>
+        <%= pluralize @user.articles.published.size, "post" %> published
+      </div>
+      <div class="flex items-center mb-4">
+        <%= inline_svg_tag("comment.svg", aria: true, class: "crayons-icon mr-3 color-base-50", title: "Comment") %>
+        <%= pluralize @user.comments.where(deleted: false).size, "comment" %> written
+      </div>
+      <div class="flex items-center">
+        <%= inline_svg_tag("hashtag.svg", aria: true, class: "crayons-icon mr-3 color-base-50", title: "Tags") %>
```
rhymes, line 88: the title is plural but it should describe the image, so should it be "Hashtag" ?

ludwiczakpawel: Probably "Tag" instead of "Hashtag" since we don't use the latter in other places.. I'll rename the file too.

benhalpern (approved):
This looks great.

PR #11358, opened 2020-11-10
Add configs/show api for admins

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This is an initial admin API endpoint which could help third party tools (or forem.com) manage forem config more easily. This is the initial MVP of admin API.

The next step would be a write endpoint.

The configs endpoint can be accessed only by super admins.

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [x] Docs.forem.com
- [ ] README
- [ ] No documentation needed


app/controllers/api/v0/admin/configs_controller.rb
```diff
@@ -0,0 +1,26 @@
+module Api
+  module V0
+    module Admin
+      class ConfigsController < ApiController
+        before_action :authorize_admin
```
citizen428, line 5: Doesn't this rely on having a logged-in user? I'm probably just misunderstanding something, but I expected a call to `authenticate!` somewhere.

app/controllers/api/v0/admin/configs_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+        after_action :verify_authorized
+
+        respond_to :json
```
citizen428, line 8: You can move this into `routes.rb` which would be more consistent with how  we do it elsewhere:

```ruby
resources :configs, only: %i[show create], defaults: { format: :json }
```

docs/api_v0.yml
```diff
@@ -1760,6 +1761,29 @@ tags:
     description: User or organization profile images
 
 paths:
+  /admin/configs/all:
```
benhalpern, line 1764: @rhymes I'd appreciate your help in constructing the spec properly, and what I should define.

I think in general we shouldn't make promises about the individual attributes of `SiteConfig` because we should maintain the freedom to change this a little more liberally right now.

config/routes.rb
```diff
@@ -208,6 +208,10 @@
             get "/:org_username", to: "organizations#show"
           end
         end
+
+        namespace :admin do
+          resources :configs, only: %i[show create], defaults: { format: :json }
```
rhymes, line 213: Do we plan to access single keys separately? Otherwise I'd suggest using the [singular resources routes](https://guides.rubyonrails.org/routing.html#singular-resources)

Something like:

```ruby
        namespace :admin do
          resource :configuration, only: %i[show]
        end
```

This will generate the following route:

```
                           api_admin_configuration GET      (/locale/:locale)/api/admin/configuration(.:format)                                    api/v0/admin/configurations#show {:locale=>nil, :format=>"json"}
```

and you can access it with: `curl localhost:3000/api/admin/configuration` (no need for `all` or anything else)

docs/api_v0.yml
```diff
@@ -1732,6 +1731,8 @@ components:
         profile_image: https://res.cloudinary.com/...jpeg
 
 tags:
+  - name: adminconfigs
+    description: Configs is the site-wide configuration set by admins (requires authorization)
```
rhymes, line 1735:
```suggestion
  - name: admin-configuration
    description: Site-wide configuration set by admins (requires super admin authorization)
```

docs/api_v0.yml
rhymes: yeah, I think this should definitely treat configuration as a unified object, not treat single keys individually, see my comment in the route file

app/controllers/api/v0/admin/configs_controller.rb
```diff
@@ -0,0 +1,14 @@
+module Api
+  module V0
+    module Admin
+      class ConfigsController < ApiController
+        before_action :authenticate!
```
rhymes, line 5: I'm not sure we should enable OAuth2 here for now, maybe we can use `authenticate_with_api_key_or_current_user!` ?

spec/requests/api/v0/admin/configs_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        expect(response.parsed_body["community_name"]).to eq SiteConfig.community_name
+      end
+    end
+
+    context "when user is not super admin" do
+      it "renders unauthorized json" do
+        headers = { "api-key" => api_secret.secret, "content-type" => "application/json" }
+        get "/api/admin/configs/all", headers: headers
+
+        expect(response.status).to eq 401
```
rhymes, line 37: instead of using the error code you can use `expect(response).to have_http_status(:unauthorized)`

rhymes (commented):
Left some notes

config/routes.rb
benhalpern: Changed to `resource :config` ✅

app/controllers/api/v0/admin/configs_controller.rb
benhalpern: Good call.

spec/requests/api/v0/admin/configs_spec.rb
benhalpern: ✅

spec/requests/api/v0/admin/configs_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "rails_helper"
+
+RSpec.describe "Api::V0::Admin::Configs", type: :request do
+  let(:api_secret) { create(:api_secret) }
+  let(:user) { api_secret.user }
+
+  describe "GET /api/admin/configs/all" do
```
rhymes, line 7: this should reflect the new URL `/api/admin/config`

docs/api_v0.yml
```diff
@@ -1760,6 +1761,29 @@ tags:
...
 
 paths:
+  /admin/config:
+    get:
+      operationId: getConfig
+      summary: Admin config
+      description: |
+        This endpoint returns the "site config" as set by admin.
+      tags:
+        - adminconfig
```
rhymes, line 1771: `admin-configuration`

rhymes (approved):
A couple of small details to address. Approving ahead of time

spec/requests/api/v0/admin/configs_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "rails_helper"
+
+RSpec.describe "Api::V0::Admin::Configs", type: :request do
+  let(:api_secret) { create(:api_secret) }
+  let(:user) { api_secret.user }
+
+  describe "GET /api/admin/config" do
+    before do
+      SiteConfig.community_name = "ANYTHING"
```
citizen428, line 9: **suggestion (non-blocking):** I'd feel better with this:

```suggestion
      expect(SiteConfig).to receive(:community_name).and_return("ANYTHING")
```

citizen428 (approved):

PR #11362, opened 2020-11-10
Simplify admin config permissions and add links to help pages

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR simplifies the configs page by requiring _super admin_ priviliges but not the special "single resource admin" ones, which is really unintuitive.

It also adds links to help pages for controllers where that exists. It is available in the bottom right if there is a help page for the view. It is designed to be flexible so if we move to another place for the help guides, all the updates are in one place.

Screen Shot 2020-11-10 at 4 11 13 PM

## Added tests?

- [x] Yes
      have not been included_

Adjusted existing tests

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

alt_text


jacobherrington (approved):
Works for me! I think this is a lot less confusing regarding permissions.

I don't think this will break anything on existing Forems, but we might want to make sure we communicate this change to anyone that gets it soon.

app/views/layouts/admin.html.erb
```diff
@@ -71,6 +71,14 @@
         </div>
       <% end %>
       <%= yield %>
+      <% if @help_url %>
+        <a class="admin-help-button crayons-btn crayons-btn--icon-rounded crayons-btn--s" href="<%= @help_url %>" target="_blank">
```
rhymes, line 75: we should add the usual `rel="noopener noreferer"` so we avoid leaking IPs

app/views/layouts/admin.html.erb
```diff
@@ -71,6 +71,14 @@
...
+          <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="54px" height="54px">
+            <path fill="none" d="M0 0h24v24H0z"/>
+            <path fill="#ffffff" d="M12 22C6.477 22 2 17.523 2 12S6.477 2 12 2s10 4.477 10 10-4.477 10-10 10zm0-2a8 8 0 1 0 0-16 8 8 0 0 0 0 16zm-1-5h2v2h-2v-2zm2-1.645V14h-2v-1.5a1 1 0 0 1 1-1 1.5 1.5 0 1 0-1.471-1.794l-1.962-.393A3.501 3.501 0 1 1 13 13.355z"/>
+          </svg>
```
rhymes, line 79: can this be saved as a file? If we use one the `inline_svg_tag` we can also add the a11y label as it's missing here

app/controllers/admin/configs_controller.rb
```diff
@@ -189,7 +189,7 @@ def raise_confirmation_mismatch_error
     end
 
     def extra_authorization_and_confirmation
-      not_authorized unless current_user.has_role?(:single_resource_admin, Config) # Special additional permission
+      not_authorized unless current_user.has_role?(:super_admin) # Special additional permission
```
rhymes, line 192: I think the inline comment is misleading now that we reverted to a regular role

rhymes (commented):
Cool! Left a couple of notes! :) The inline help link is extremely useful!

app/controllers/admin/application_controller.rb
```diff
@@ -1,8 +1,24 @@
 module Admin
   class ApplicationController < ApplicationController
     before_action :authorize_admin
+    before_action :assign_help_url
     after_action :verify_authorized
 
+    HELP_URLS = {
```
Ridhwana, line 7: **(comment, non-blocking):~** This seems like it would be beneficial to put into a constants file

app/views/layouts/admin.html.erb
benhalpern, line 75:
```suggestion
        <a class="admin-help-button crayons-btn crayons-btn--icon-rounded crayons-btn--s" href="<%= @help_url %>" target="_blank" rel="noopener noreferer" >
```

app/controllers/admin/configs_controller.rb
benhalpern, line 192:
```suggestion
      not_authorized unless current_user.has_role?(:super_admin)
```

app/views/layouts/admin.html.erb
benhalpern: Cool, added this.

app/controllers/admin/application_controller.rb
benhalpern: My feeling that it was simplest to have it near where we need it for now and as the feature expands we could move it.

app/views/layouts/admin.html.erb
benhalpern: Made this change.

rhymes (approved):
Great! 

Just came to me that we should probably also update https://docs.forem.com/admin/ as it mentions the need for `:single_resource_admin` to access

benhalpern:
@rhymes single resource admin is still a thing.... This is the removal of this one feature which was sort of the wrong use of it anyway.

Single resource admin is for giving _less permission_ to folks who might only need access to a single interface. So the current docs are _more_ accurate because we no longer have this one additional exception.

Issue #10878, opened 2020-10-16
Update existing keyboard shortcuts


Link2Twenty:
Currently there are a few different keyboard shortcuts in Forem and they've been built a few different ways. It would be good to now unify these under the new approach (#10713)

**Describe the solution you'd like**

Code should be changed to the new solution

**Describe alternatives you've considered**

We leave the old code as legacy until it breaks

**Additional context**

#10713 is waiting to be merged this issue is just a place to store code that needs updating

- [x] [Article form](https://github.com/forem/forem/blob/ebe398ba23f0c92adfa41274759c5e25e899a57a/app/javascript/article-form/articleForm.jsx)
- [x] [Article form component](https://github.com/forem/forem/blob/04f1260bf5579bb2f978247d7c56391b68e9c577/app/javascript/article-form/components/keyboardShortcutsHandler.jsx#L6)
- [x] [Search](https://github.com/forem/forem/blob/653ec123006b8bf7f93b9f32ca13a9a0c959200a/app/javascript/Search/Search.jsx)
- [x] [Listings modal has a listener on 'Escape' to close it](https://github.com/forem/forem/blob/04f1260bf5579bb2f978247d7c56391b68e9c577/app/javascript/listings/listings.jsx#L91)

reobin:
Search and toggle Zen mode (`0`) are done in #10468 🌞

Let's make a list of the missing ones. They should be fun to update.

reobin:
2 more:
- [Article form has a listener on 'P](https://github.com/forem/forem/blob/04f1260bf5579bb2f978247d7c56391b68e9c577/app/javascript/article-form/components/keyboardShortcutsHandler.jsx#L6) (The whole component is probably obsolete now)
- [Listings modal has a listener on 'Escape' to close it](https://github.com/forem/forem/blob/04f1260bf5579bb2f978247d7c56391b68e9c577/app/javascript/listings/listings.jsx#L91)

Link2Twenty:
@reobin do you wanna make an issue for a shortcut 'store' that can compile shortcuts per component for documentation?

reobin:
@Link2Twenty Sure! At least, to start the discussion on how to do it best.

citizen428:
@Link2Twenty I assigned you this issue now since you seem to be coordinating this effort (thanks for that btw).

reobin:
Seems like it can be closed already @Link2Twenty good work

Link2Twenty:
We shared the work 🙂

PR #11363, opened 2020-11-10
Set default timeout to 0 ms for keyboard shortcuts.

nickytonline:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I missed this when merging #11113. It adds a timeout with a default value of 500ms, but this is only supposed to be for chained keyboard strokes. This PR sets the timeout to 0ms by default. We'll look into fixing the actual issue in another PR (see https://github.com/forem/forem/pull/11363#issuecomment-725010112)

Fixes a regression in:
- #10468
- #10878 
- #11116

1. For any of the keystrokes supported in the above mentioned tickets, there should no longer be a 500ms delay.

## Added tests?

- [ ] Yes
- [x] No, and this is why: There are tests that already cover this.

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

A movie character saying "Whoopsies"

### DIFF
--- a/app/javascript/shared/components/useKeyboardShortcuts.js
+++ b/app/javascript/shared/components/useKeyboardShortcuts.js
@@ -27,7 +27,7 @@ function isFormField(element) {
 
 // Default options to be used if null
 const defaultOptions = {
-  timeout: 0,
+  timeout: 0, // The default is zero as we want no delays between keystrokes by default.
 };
 
 /**
--- a/app/javascript/shared/components/useKeyboardShortcuts.js
+++ b/app/javascript/shared/components/useKeyboardShortcuts.js
@@ -1,5 +1,5 @@
-import { useState, useEffect, useCallback } from "preact/hooks";
-import PropTypes from "prop-types";
+import { useState, useEffect, useCallback } from 'preact/hooks';
+import PropTypes from 'prop-types';
 
 /**
  * Checker that return true if element is a form element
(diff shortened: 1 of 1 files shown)


app/javascript/shared/components/useKeyboardShortcuts.js
```diff
@@ -12,18 +12,22 @@ function isFormField(element) {
...
+      type !== 'checkbox' &&
+      type !== 'radio') ||
     element.isContentEditable
   );
 }
 
 // Default options to be used if null
 const defaultOptions = {
-  timeout: 500
+  timeout: 0,
```
nickytonline, line 30: This is the actual fix. The rest is prettier in action.

nickytonline:
@reobin and @Link2Twenty, I think we can rework the keyboard usage to only apply a timeout to combined keystrokes. That'll come in another PR though. I can look at that, or if one of you want to look into that, go for it. Thanks again to both of you for all your contributions. 🔥

maestromac (approved):

app/javascript/shared/components/useKeyboardShortcuts.js
vaidehijoshi: Should we improve the documentation here to include a reason why we are explicitly having the default timeout be `0` here? Just thinking that we might want to document that for our future selves since it is intentional. But not a blocker; let me know what you think

vaidehijoshi (approved):

mstruve (approved):

nickytonline:
@vaidehijoshi, I added a comment about the default 0ms timeout value.

PR #11364, opened 2020-11-10
Convert inline JS to Stimulus

jacobherrington:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Nothing special, just refactored some really redundant JS in these partials to a Stimulus controller.

You'll need to run `Flipper.enable(:profile_admin)` in your console to be able to see this UI.

A tiny UX tweak: Converted these to buttons instead of divs.

Before:
image

After:
image

## Added tests?

- [ ] Yes
- [x] No, and this is why: Tests for this functionality didn't exist previously, and testing this code would mostly just be testing that the browser API works as expected because we are just toggling a CSS class.

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/javascript/admin/controllers/modal_controller.js
+++ b/app/javascript/admin/controllers/modal_controller.js
@@ -1,0 +1,10 @@
+import { Controller } from 'stimulus';
+
+export default class ModalController extends Controller {
+  toggleModal() {
+    let modal = document.getElementById(this.data.get('target-id'));
+    if (modal) {
+      modal.classList.toggle('hidden');
+    }
+  }
+}
--- a/app/javascript/admin/controllers/modal_controller.js
+++ b/app/javascript/admin/controllers/modal_controller.js
@@ -1,10 +1,11 @@
 import { Controller } from 'stimulus';
 
 export default class ModalController extends Controller {
+  static targets = ['toggle'];
+
   toggleModal() {
-    let modal = document.getElementById(this.data.get('target-id'));
-    if (modal) {
-      modal.classList.toggle('hidden');
+    if (this.toggleTarget) {
+      this.toggleTarget.classList.toggle('hidden');
     }
   }
 }
--- a/app/views/admin/profile_fields/_add_group_modal.html.erb
+++ b/app/views/admin/profile_fields/_add_group_modal.html.erb
@@ -1,4 +1,4 @@
-<button class="btn btn-primary" type="button" onclick="showAddGroupModal()">
+<button class="crayons-btn" type="button" data-controller="modal" data-action="modal#toggleModal" data-modal-target-id="add-group-modal">
   Add group
 </button>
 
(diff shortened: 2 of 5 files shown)


app/javascript/admin/controllers/modal_controller.js
```diff
@@ -0,0 +1,10 @@
+import { Controller } from 'stimulus';
+
+export default class ModalController extends Controller {
+  toggleModal() {
+    let modal = document.getElementById(this.data.get('target-id'));
```
nickytonline, line 5: refactor (non-blocking): Just wondering why we're not using Stimulus' [targets](https://stimulusjs.org/reference/targets) via the `data-target` attribute.

Also, is `target-id` being used instead of `modal-target-id` because there is some naming convention wizardry going on?

citizen428:
This works fine, but I agree with @nickytonline that targets seem to be more idiomatic for Stimulus. I applied the following change locally and everything still seems to work:

```diff
diff --git a/app/javascript/admin/controllers/modal_controller.js b/app/javascript/admin/controllers/modal_controller.js
index 3b63fda14..e108263f7 100644
--- a/app/javascript/admin/controllers/modal_controller.js
+++ b/app/javascript/admin/controllers/modal_controller.js
@@ -1,10 +1,11 @@
 import { Controller } from 'stimulus';

 export default class ModalController extends Controller {
+  static targets = [ "toggle" ]
+
   toggleModal() {
-    let modal = document.getElementById(this.data.get('target-id'));
-    if (modal) {
-      modal.classList.toggle('hidden');
+    if (this.toggleTarget) {
+      this.toggleTarget.classList.toggle('hidden');
     }
   }
 }
diff --git a/app/views/admin/profile_fields/_add_group_modal.html.erb b/app/views/admin/profile_fields/_add_group_modal.html.erb
index 29fae0ecc..494e4ade0 100644
--- a/app/views/admin/profile_fields/_add_group_modal.html.erb
+++ b/app/views/admin/profile_fields/_add_group_modal.html.erb
@@ -7,7 +7,7 @@
     <div class="crayons-modal__box">
       <header class="crayons-modal__box__header">
         <h2>Add Group</h2>
-        <button type="button" data-controller="modal" data-action="modal#toggleModal" data-modal-target-id="add-group-modal" class="crayons-btn crayons-btn--icon crayons-btn--ghost">
+        <button type="button" data-controller="modal" data-action="modal#toggleModal" data-target="modal.toggle" class="crayons-btn crayons-btn--icon crayons-btn--ghost">
           <svg width="24" height="24" viewBox="0 0 24 24" class="crayons-icon" xmlns="http://www.w3.org/2000/svg">
             <path d="M12 10.586l4.95-4.95 1.414 1.414-4.95 4.95 4.95 4.95-1.414 1.414-4.95-4.95-4.95 4.95-1.414-1.414 4.95-4.95-4.95-4.95L7.05 5.636l4.95 4.95z" />
           </svg>
```

rhymes (commented):
Cool refactoring! The option to enable is `Flipper.enabled?(:profile_admin)` not `:profile_fields` :D 

https://github.com/forem/forem/blob/27a3df7d73ae6e99cd740125d42a27bdc8563360/app/helpers/admin_helper.rb#L42

*Non blocking* aside from what Nick and Michael highlighted, there's one thing odd I noticed in the behavior:

profile

the "add field" button moves when the modal opens, but it also happens on master :D I wonder if that was intentional or just a bug

app/javascript/admin/controllers/modal_controller.js
jacobherrington: Yeah, I had some trouble using targets for this the first time through, but looking at it with fresh eyes, that is the right call. I just needed to add some kind of container element so that the same target could be shared by both buttons.

jacobherrington: To the second question, yes! :+1:

jacobherrington:
@nickytonline @citizen428 Updated to use targets instead of the IDs, good call.

jacobherrington:
@rhymes That is a CSS bug that existed before I did this... and I can't tell you why, but for some reason the changes I just pushed fixed it. :joy: I didn't want to fix it because I didn't want to change CSS be used elsewhere in this PR (I think that should be a different PR).

Again, it's fixed now, I just don't know why... :upside_down_face:

rhymes (approved):
LOL to the accidental bug fix😂

nickytonline (approved):
Looks great. Thanks for making the changes @jacobherrington! 🚀

PR #11377, opened 2020-11-11
feat: Preload links on focus

### LINKED ISSUE #11372: Keyboard navigation on feed requires hitting "return" twice to access an article
**Describe the bug**

When navigation through feed with keyboard (`j` & `k`) I have to press `return` twice to access selected article... However the first press activates the progress bar on top of the header.

**To Reproduce**

1. Go to home page
2. Press `j` and/or `k` to select random article on feed.
3. Press `return` - it will activate the progress bar on top of the page but it will not take you to the article itself.
4. Press `return` again - it will now take you to the selected article.

**Expected behavior**

Take me to the article on first `return` press.

- OS, version: macOS X 10.15.7
- Browser, version: Chrome 86.0.42x

reobin:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Links are currently preloaded on mouse hover.

This adds the same functionality but for keyboard focus.

Fixes some issues with links needing 2 hits on return in order to work.

Closes #11372 

1. Navigate any link using the keyboard
2. Press return
3. Link is followed without delay or the need to press return again

## Added tests?

- [ ] Yes
- [x] No, and this is why: not needed

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/assets/javascripts/base.js.erb
+++ b/app/assets/javascripts/base.js.erb
@@ -245,6 +245,25 @@ var instantClick
     getImageForLink(a);
   }
 
+  function focusListener(e) {
+    var a = getLinkTarget(e.target)
+
+    if (!a || !isPreloadable(a)) {
+      return
+    }
+
+    a.removeEventListener('focus', focusListener)
+
+    if (!$delayBeforePreload) {
+      preload(a.href)
+    }
+    else {
+      $urlToPreload = a.href
+      $preloadTimer = setTimeout(preload, $delayBeforePreload)
+    }
+    getImageForLink(a);
+  }
+
   function clickListener(e) {
     try{
       var a = getLinkTarget(e.target)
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/base.js.erb
```diff
@@ -245,6 +245,25 @@ var instantClick
     getImageForLink(a);
   }
 
+  function focusListener(e) {
```
rhymes, line 248: It would be great to explain what you mentioned in the PR as an inline comment to the function to answer the why we need this :)

rhymes (commented):
Tested locally and it requires only one instance of the enter key instead of two :)

app/assets/javascripts/base.js.erb
```diff
@@ -245,6 +245,28 @@ var instantClick
...
+  // It also covers the issue where a user needs to press <return>
+  // twice in order to follow a focused link.
+  function focusListener(e) {
+    var a = getLinkTarget(e.target)
+
+    if (!a || !isPreloadable(a)) {
+      return
+    }
+
+    a.removeEventListener('focus', focusListener)
```
nickytonline, line 258: refactor (blocking): There is no need to remove the event listener here because the `focusListener` is on the `document.body`. This line won't fail, but it won't change anything.

```suggestion
```

nickytonline (changes requested):
Thanks for the PR @reobin. One small change request and this is good to go.

ludwiczakpawel (approved):
It works locally 🙌

nickytonline (approved):
Thanks so much for the PR! 😎🚀

PR #11380, opened 2020-11-11
Auto generate mascot accounts

atsmith813:
- [x] Refactor

## Description
This PR removes the `SiteConfig` requirement of `mascot_user_id` by auto-generating a mascot account when setting up a Forem.

Eventually, we'll want to 1) move some more mascot fields to `SiteConfig` and 2) refactor Mascot altogether into its own domain.

closes https://github.com/forem/InternalProjectPlanning/issues/94

1. Set `ENV["MODE"]` in your environment to `"STARTER"`.
2. Reset your DB (`rails db:reset`) - **warning** this will erase your local database.
3. Fire up the app locally.
4. Go through onboarding, create an account.
5. Fire up a console and verify that `User.mascot_account` returns a mascot user.

## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

blank_face_with_smoke_gif


mstruve (approved):
First time using starter mode, very cool! Worked like a charm! 🚀

app/services/users/create_mascot_account.rb
```diff
@@ -0,0 +1,29 @@
...
+      registered_at: Time.current,
+      password: SecureRandom.hex
+    }.freeze
+
+    def self.call
+      new.call
+    end
+
+    def call
+      return if SiteConfig.mascot_user_id
```
vaidehijoshi, line 18: It might be useful to know if somehow this service was called and the `mascot_user_id` for some strange reason was _already_ set...maybe we could log it with a warning? Right now will short-circuit silently, but I wonder if there could be a scenario where we'd actually want to know that this service was being called again and again and from the `registrations_controller`... 🤔 

What do you think?

app/services/users/create_mascot_account.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      mascot = User.create(mascot_params)
```
vaidehijoshi, line 20:
```suggestion
      mascot = User.create!(mascot_params)
```
Any reason not to fail hard if for some reason the mascot user isn't created?

vaidehijoshi (commented):
I _think_ that a side-effect of us generating the mascot user as part of the "sign up as first user" flow is that we will be suggesting the mascot as one of the suggested users in the onboarding flow (I think that `Suggester::Users::Recent` is responsible for this).

Screen Shot 2020-11-11 at 2 52 14 PM

Just wanted to highlight that here since it will impact _new_ Forems that are set up with this change deployed (it _won't_ affect DEV though) and wanted to know whether we think we want to be suggesting the mascot by default or not since that is a significant enough change from what we _currently_ do in the onboarding flow.

What do we think -- do we want to allow this behavior or should we skip the `mascot_user_id` from the suggested users that are used in the follow users slide of onboarding?

benhalpern (approved):
Looks great. Big improvement here removing the need to jump through the hoops to get this set up.

mstruve:
> do we want to allow this behavior or should we skip the mascot_user_id from the suggested users that are used in the follow users slide of onboarding?

I think skipping it would be a bit cleaner from an experience perspective and since we are going to be heading towards removing it as a user the minor code change to make that happen will be short-lived.

app/services/users/create_mascot_account.rb
atsmith813: I updated it to raise an error.

atsmith813:
@vaidehijoshi I think this only happens in development because of our environment guards

https://github.com/forem/forem/blob/1a3f2739f8ba85fcec637a77eedd1fb2795c4c27/app/services/suggester/users/recent.rb#L57-L67

The count is `0` for the mascot account, but in environments outside of Production, we set the limit to `-1`, therefore picking up the mascot account in development.

vaidehijoshi:
@atsmith813 nooooooooo not another `Rails.env.production?` check!!! 🙈 

good digging here -- i missed that `-1` when i looked into it last night.

vaidehijoshi (approved):

atsmith813:
@mstruve @benhalpern made some code updates so I'm re-requesting review!

app/services/users/create_mascot_account.rb
```diff
@@ -0,0 +1,29 @@
+module Users
+  class CreateMascotAccount
+    MASCOT_PARAMS = {
+      email: "mascot@forem.com",
+      username: "mascot",
+      name: "Mascot",
+      profile_image: SiteConfig.mascot_image_url,
```
benhalpern, line 7: Hmmm if this is getting created before the user has configured the site, maybe this should use one of our auto-generated images instead?

`Users::ProfileImageGenerator.call`

atsmith813: We use a default image for `mascot` https://github.com/forem/forem/blob/2225fe99990fddfc529fac8017950412f63cffd2/app/models/site_config.rb#L101

Are you saying a profile image would be a better fit than the mascot image? Or that the mascot image won't work?

benhalpern: Ah, gotcha. I wasn't putting it all together. Nevermind this all looks good.

benhalpern (approved):

PR #11384, opened 2020-11-12
Badges area on profile

ludwiczakpawel:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

With recent profile UI update it turned out that Badges section can be pretty big for users with lots of badges.. This PR improves sizing for badges and it also limits number of badges displayed by default: so it will only display 6 badges, and if user has more, it will show "Show more" button that will reveal all badges.

| Case | Screenshot |
|------|------|
| <= 6 badges | image |
 |> 6 badges | image |
| > 6 badges, revealed | image |

## Added tests?

- [ ] Yes
- [x] No, not needed.

## Added to documentation?

- [ ] README
- [x] No documentation needed


app/views/articles/_badges_area.html.erb
```diff
@@ -1,13 +1,20 @@
+<% badges_limit = 5 %>
```
rhymes, line 1: we should move this up as a `@badges_limit` variable in the controller. Specifically around this line

https://github.com/forem/forem/blob/de5a22ecf7d93d8b5bd692481dbd8618201c7ab0/app/controllers/stories_controller.rb#L225

rhymes (commented):
Cool! Left a note! I would even go down to the top 4 by default :D

app/views/articles/_badges_area.html.erb
ludwiczakpawel: Cool, I moved it to the controller. I'd appreciate another look, since you know - it's me :D 

Re the number limit, let's keep it 6 for now because on mobile it will actually show 2 rows of badges (grid has 3 cols on mobile and 2 cols on bigger breakpoint) - and if we limit it to 4 badges only it will look off on  mobile (1st row: 3 badges, 2nd row: 1 badge) where with 6 badges limit it should look fine on both mobile and desktop.

rhymes (approved):

fdoxyz (approved):
LGTM 👍

PR #11386, opened 2020-11-12
Replace calls to Flipper with FeatureFlag

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

As we introduced the `FeatureFlag` wrapper in https://github.com/forem/forem/pull/8149 we should try to use that around the code instead of directly accessing Flipper. This will help us if we ever have to switch feature flag framework, would also enable us to add metrics on feature flag calls (eg. "how many times do we access flagged features?") and so on.

Note: the API is the same, just use `FeatureFlag` instead of `Flipper` and if anything missing is needed, add it to the wrapper module.

#8149

1. Load the console, try `FeatureFlag.enable(:foobar)` and check with `Flipper.enabled?(:foobar)` that is indeed enabled
1. Do nothing as this is already tested automatically

## Added tests?

- [x] Yes
      have not been included_

### DIFF
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -39,7 +39,7 @@ module AdminHelper
   PROFILE_ADMIN = { name: "config: profile setup", controller: "profile_fields" }.freeze
 
   def admin_menu_items
-    return MENU_ITEMS unless Flipper.enabled?(:profile_admin)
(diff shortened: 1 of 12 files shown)


Ridhwana (approved):
Nice refactor!

mstruve (approved):
NICE!

PR #11387, opened 2020-11-12
Fixed the scroll bar problem in Safari

damla:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As you guys know, while editing a new post, there is a modal box used for demonstrating markdown formats and liquid tags. In Safari, I realized that the scroll bar is overflowing at the bottom of the modal as shown in Screenshot no.1. So, adding "overflow: hidden;" to the class of that modal, which is "crayons-modal__box", fixed the problem. 

I've also checked it on Firefox and Google Chrome and they do not have any bug like this.(Screenshot no.2 and Screenshot no.3)

**Folder location:** `forem/app/assets/stylesheets/components/modal.scss`

  safari
<p align="center">Screenshot no.1 - Problem in Safari</p>

firefox
<p align="center">Screenshot no.2 - Firefox</p>

google_chrome
<p align="center">Screenshot no.3 - Google Chrome</p>

_You may test the changes by entering to "write a new post page" and clicking on Markdown or Liquid Tags link found in the right section._

## Added tests?

- [ ] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

like a cat

### DIFF
--- a/app/assets/stylesheets/components/modals.scss
+++ b/app/assets/stylesheets/components/modals.scss
@@ -56,7 +56,6 @@
     width: 100%;
     height: 100%;
     pointer-events: auto;
-    overflow: hidden;
 
     @media (min-width: $breakpoint-s) {
       @include generate-box(
@@ -76,7 +75,8 @@
       align-items: center;
       justify-content: space-between;
       padding: var(--modal-header-padding);
-
+      flex-shrink: 0;
+      
       &__title {
         font-size: var(--modal-header-font-size);
       }
--- a/app/assets/stylesheets/components/modals.scss
+++ b/app/assets/stylesheets/components/modals.scss
@@ -56,6 +56,7 @@
     width: 100%;
     height: 100%;
     pointer-events: auto;
+    overflow: hidden;
 
     @media (min-width: $breakpoint-s) {
       @include generate-box(


rhymes:
Hi @damla, I wonder if this only applies to some versions of Safari because I'm not experiencing the issue. I tried with `Version 14.0 (15610.1.28.1.9, 15610)`

damla:
Hi @rhymes, I checked my version and it is `Version 14.0 (15610.1.28.1.9, 15610)` too. Then, I tried the same scenario with a long post (scrollable) and empty post. It seems like when the post is scrollable, modal with markdown does not having this problem but liquid tag does. I screen recorded it to demonstrate it better.
test1
I converted it to a gif so, it is a little bit slow 😕

rhymes:
Thanks! I wonder if it's a question of resolution or I haven't fully understood where the defect lies :D 

This is from my Safari:

modal

damla:
I updated my macOS version to` 10.15.7` to be sure, but it still has the same problem. My screen size is 13 inch btw, maybe it can be helpful.

damla:
Hi @rhymes, I just realized when I disconnect my mouse and just use the trackpad, the scroll bar become hidden as default. Here is a screenshot of it:
image

But when I plugged in the dongle, scroll bar is always there like shown in this screenshot:
image

After I added "overflow:hidden;" from dev tools in Safari, it becomes like this:
image

ludwiczakpawel:
I believe the issue that @damla is talking about is the scenario when you can notice ~1px of the scrollbar being on top of the that black shadow, is that correct?

If yes, I think I know what the cause of this issue, and it's actually pretty funny bug (well, in a way :))

On Safari, when you have smaller window resolution (height in particular), the modal header gets shrinked (see how squeezed it is on screenshot? This behavior is kinda correct and it's quite funny how browser handle shrinking elements like that ("like that" = part of flexbox). So browser doesn't shrink this element with whole pixels (50px, 49px, 48px...) but with decimal fractions which could be anything like (50px, 49.999px, 49.998px, ...). The actual values for fractions are dependent on window size and the layout of particular elements (in our case: modal header + modal body). 

Now the fun part is that even if one element (modal header) has height with values with decimal fractions, the other element (modal body) doesn't really care and tries to use rounded numbers.

That's why you can *sometimes* spot that scrollbar being on top of the shadow - and it's not really `1px` - it can be anything between `0px` and `1px` :).

And fix for this is actually quite simple: we have to add `flex-shrink: 0;` to [`.modal__header` element](https://github.com/forem/forem/blob/master/app/assets/stylesheets/components/modals.scss#L72) - that should do the trick :)

---

PS. It was really fun to debug this :)

ludwiczakpawel (changes requested):
^

ludwiczakpawel (approved):
⛴ Ship it! ⛴

damla:
Hi @ludwiczakpawel, thanks for explaining the bug in detail, I've learned a lot from it 🙃

rhymes (approved):

PR #11388, opened 2020-11-12
Hotfix: Remove mascot from JS template for modals

rhymes:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Right now the app is not booting locally because `showUserAlertModal.js.erb` requires `SiteConfig` which is not accessible when the app compiles that template on boot. 

This is just an interim solution, we can find a better way to show the mascot in the modal in a separate PR

cc @lisasy @maestromac @nickytonline 

## Related Tickets & Documents

#11344

PR #11390, opened 2020-11-12
Revert "Remove mascot from JS template for modals (#11388)"

rhymes:
This reverts commit f8f3795b88436f67ca714d5088aee3a6969dd091.

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Me and @maestromac confirmed it was something "only on my machine" and we couldn't reproduce it, so this should go back to normal

### DIFF
--- a/app/assets/javascripts/utilities/showUserAlertModal.js.erb
+++ b/app/assets/javascripts/utilities/showUserAlertModal.js.erb
@@ -64,7 +64,7 @@ const getModalHtml = (
 ) => `<div id="${modalId}" data-testid="modal-container" class="crayons-modal crayons-modal--m hidden">
     <div role="dialog" aria-modal="true" class="crayons-modal__box">
       <div class="crayons-modal__box__header border-b-0 justify-end">
-          <button class="crayons-btn crayons-btn--ghost crayons-btn--icon" type="button"
+          <button class="crayons-btn crayons-btn--ghost crayons-btn--icon" type="button" 
               onClick="toggleUserAlertModal();" aria-label="Close">
             <svg width="24" height="24" viewBox="0 0 24 24" class="crayons-icon"
               xmlns="http://www.w3.org/2000/svg" role="img" aria-labelledby="714d29e78a3867c79b07f310e075e824">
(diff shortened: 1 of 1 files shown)


maestromac (approved):
LGTM!

We would like more people to try to boot this up locally to verify that it is working without throwing any site_config related warnings.

joshpuetz:
Works/boots just fine locally for me!

joshpuetz (approved):

juliannatetreault (approved):
Tested this out and the images loaded as expected -- works for me! 🚀

atsmith813:
`bin/startup` doesn't work for me, but `rails s` works

rhymes:
@atsmith813 what error do you get ? I tried with `/bin/startup` and it worked. Then I had an idea: calling clear cache on site config and booting it up again to see if maybe it was that's the reason it was working but still no issues anymore. Mmm

joshpuetz:
I should clarify I ran locally with `overmind start -f Procfile.dev`

atsmith813:
Screen Shot 2020-11-12 at 11 05 59 AM

cc: @rhymes

maestromac:
I also used hivemind but I double checked with `bin/rails server`

rhymes:
@atsmith813 mmm weird, can you run bin/setup and then bin/startup?

benhalpern (changes requested):
This will, indeed, not work because for Forem Cloud, assets are compiled outside of the container so we can't use anything that relies on the database or even environment variables.

vaidehijoshi:
> This will, indeed, not work because for Forem Cloud, assets are compiled outside of the container so we can't use anything that relies on the database or even environment variables.

Hm, if this is the case, then we need to rethink all of the "default" images that we are adding to `SiteConfig` then, potentially? If we access them through `SiteConfig`, then we're accessing the DB, and if we can't access them from SiteConfig, maybe it doesn't make sense to define the defaults in there.

benhalpern:
@vaidehijoshi do any of those images need to be referenced within JavaScript files in a way that where they can't be passed in HTML via `data` params?

vaidehijoshi:
> @vaidehijoshi do any of those images need to be referenced within JavaScript files in a way that where they can't be passed in HTML via `data` params?

Hm, I'm not immediately sure! It just seems annoying/a potential pain point down the road that we won't be able to use these image files that exist in our DB in any other way, so I just wanted to flag it as a downside of our "put default images in the SiteConfig" approach. I was just pondering if there was a better/another place to put those default assets......but I suppose fetching it from JS is ok given the constraint of where the assets are compiled 🤷

rhymes:
I'm going to close this, I kept a record of what we could do and opened an issue in https://github.com/forem/forem/issues/11492

PR #11394, opened 2020-11-12
Filter by tags of reading list items not tags you follow

Zhao-Andy:
- [x] Refactor
- [x] Feature
- [x] Optimization

## Description
This updates the reading list to get the tags of the articles in your reading list as opposed to filtering by the tags you follow. General consensus among the team was that we should make this change, since the current version (tags you follow) was not really intuitive.

Closes https://github.com/forem/InternalProjectPlanning/issues/240

1. Save some articles from the home page: localhost:3000
2. Go to your reading list: localhost:3000/readinglist
3. See that the proper tags are displayed, and that there are no duplicates

## Added tests?
- [x] No, and this is why: We don't test for the `state` of the `readingList` component at all, so I didn't feel the need to add one. Happy to though, if I should!

## Added to documentation?
- [x] No documentation needed

Lisa Simpson says, "That was too easy," and has a skeptical look on her face.

### DIFF
--- a/app/javascript/searchableItemList/searchableItemList.js
+++ b/app/javascript/searchableItemList/searchableItemList.js
@@ -77,7 +77,7 @@ export function performInitialSearch({ searchOptions = {} }) {
     const reactions = response.result;
     const availableTags = [
       ...new Set(reactions.flatMap((rxn) => rxn.reactable.tag_list)),
-    ];
+    ].sort();
     component.setState({
       page: 0,
       items: reactions,
--- a/app/javascript/searchableItemList/searchableItemList.js
+++ b/app/javascript/searchableItemList/searchableItemList.js
@@ -75,12 +75,16 @@ export function performInitialSearch({ searchOptions = {} }) {
   const responsePromise = fetchSearch('reactions', dataHash);
   return responsePromise.then((response) => {
     const reactions = response.result;
+    const availableTags = [
+      ...new Set(reactions.flatMap((rxn) => rxn.reactable.tag_list)),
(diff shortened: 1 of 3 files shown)


rhymes (approved):
This makes me so happy! It's a usability boost to the reading list ❤️ 

Thanks @Zhao-Andy!

app/javascript/searchableItemList/searchableItemList.js
```diff
@@ -75,12 +75,16 @@ export function performInitialSearch({ searchOptions = {} }) {
   const responsePromise = fetchSearch('reactions', dataHash);
   return responsePromise.then((response) => {
     const reactions = response.result;
+    const availableTags = [
+      ...new Set(reactions.flatMap((rxn) => rxn.reactable.tag_list)),
+    ];
```
rhymes, line 80: I'm requesting this change after I approved, sorry Andy: I think we should call `.sort()` here so that tags will always be sorted alphabetically and not depend on the order of the array of reading list items

rhymes (changes requested):

ludwiczakpawel (commented):
I love that change <3 :)

rhymes (approved):
LGTM!

Zhao-Andy:
In case anyone was curious, sorting after the array has been de-duplicated with `new Set` is a bit faster than sorting it before de-duping:

https://jsbench.github.io/#250ce4b68ab1459c12a0547159ea5813

joshpuetz (approved):

PR #11398, opened 2020-11-13
Add system spec for profile fields

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Just adding a missing spec for this part of the admin UI.

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

### DIFF
--- a/spec/system/admin/admin_manages_profile_fields_spec.rb
+++ b/spec/system/admin/admin_manages_profile_fields_spec.rb
@@ -6,13 +6,13 @@ RSpec.describe "Admin manages profile fields", type: :system do
   let!(:profile_field) { create(:profile_field, profile_field_group: profile_field_group, label: "Delete Me Too!") }
 
   before do
-    Flipper.enable(:profile_admin)
+    FeatureFlag.enable(:profile_admin)
     sign_in admin
     visit admin_profile_fields_path
   end
 
   after do
-    Flipper.disable(:profile_admin)
+    FeatureFlag.disable(:profile_admin)
   end
 
   it "adds a profile group" do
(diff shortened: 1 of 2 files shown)


rhymes (approved):

spec/system/admin/admin_manages_profile_fields_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "rails_helper"
+
+RSpec.describe "Admin manages profile fields", type: :system do
+  let(:admin) { create(:user, :super_admin) }
+  let!(:profile_field_group) { create(:profile_field_group, name: "Delete Me!") }
+  let!(:profile_field) { create(:profile_field, profile_field_group: profile_field_group, label: "Delete Me Too!") }
+
+  before do
+    Flipper.enable(:profile_admin)
```
rhymes, line 9: sorry, can we replace this with `FeatureFlag.enable` ? Even better, use `allow(FeatureFlag).to receive(:enable).and_return(true)` so that it won't even write to the DB and be a little faster?

See #11386

rhymes (changes requested):
Requested changes because of `Flipper`usage

spec/system/admin/admin_manages_profile_fields_spec.rb
jacobherrington: Great point! I tried using `allow()`initially, but it doesn't seem to work. I'll definitely use the wrapper though!

rhymes (approved):
LGTM!

fdoxyz (approved):
Tests, nice!

PR #11400, opened 2020-11-13
Move badges_area template in proper place and refactor the partial

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Right now the `badges_area` template is in the `articles` folder, but it's only rendered in the `users/*` templates. 
In addition to it being moved I also refactored the partial a bit, used local variables and fixed an unlikely bug in which if the limit is set to 1 for users with 1 badge, the "show all" would still manifest itself

1. assign yourself some badges, something like: `user.badge_achievements.create(badge: Badge.last)`
1. got to your profile page

## Added tests?

- [ ] Yes
- [x] No, and this is why: there are not tests for this part and I second @ludwiczakpawel choice in #11357 to not add them right now


ludwiczakpawel (approved):
>  I second @ludwiczakpawel choice in #11357 to not add them

It's not like it was my "choice" :D I just can't (don't know how to!) write tests :D

mstruve (approved):
Nice cleanup!

Issue #81, opened 2017-08-25
Support POSSE on dev.to


bigblind:
I like the idea of [POSSE](https://indieweb.org/POSSE), which stats for **P**ost on your **O**wn **S**ite, **S**yndicate **E**lsewhere.  It promotes the idea that you should own your own content. 

I'd love it if there was some way to automatically post to dev.to when I post to my own site. Is there a possibility that you'd create an API endpoint for posting?

I understand that this isn't something you'd consider easily, as it'd open the door for automatically posting all kinds of low-quality content, and I'm not sure how to solve that problem. Still, it'd be cool.

Zhao-Andy:
API endpoints are definitely something we're considering. Good to hear that people are still interested.

We currently have our beta feature that pulls RSS feeds and creates articles for your account; not sure if that's something you've tried yet. You can find it here: https://dev.to/settings/publishing-from-rss

bigblind:
Oh, that's amazing!

darol100:
I am highly interested in this feature as well. I think Medium has done a really good job of documenting their API. What I think is worth looking at what they have done https://github.com/Medium/medium-api-docs.

If you implement APIs, I will be more than welcome to help you writing SDK's in couple different languages.

darol100:
Hey Stale (bot),

As far I know, this feature has not been implemented yet.

darol100:
Good bot!!

jessleenyc:
Hey @darol100 and @bigblind -- sorry for the silence here. We're planning to create an endpoint for posting by the end of this quarter. Was that the request? Not sure if POSSE implies any other needs.

Mayeu:
@jessleenyc If I may, to be able to fully POSSE one need to:
1. Being able to post new articles via the API with support for the canonical origin
2. Being able to collect feedback (comments, reaction, etc.) ideally via something like [webmention](https://indieweb.org/Webmention)
3. Being able to comment via the API

Having (1) will already be awesome, the rest can be split out in other tasks for after :)

philnash:
I believe that the ability to ingest feeds via RSS solves 1. It is a pain that Medium requires you to use a proprietary API in order to post and I think it fits much better with the ethos of the IndieWeb to use easily implementable standards like RSS to enable this and not another new API.

At least, if you do implement an API for posting, please don't remove the RSS ingesting feature.

Mayeu:
[...]

PR #11410, opened 2020-11-14
Podcast ownership join table (MLH Fellowship #81)

Manasa2850:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

https://github.com/MLH-Fellowship/dev.to/issues/81

## Added tests?

- [x] Yes
- [ ] No, and this is why: the code I've added doesn't require tests as of now

## Added to documentation?

- [ ] README
- [x] No documentation needed

No

alt_text


Manasa2850:
@benhalpern I've made all the changes suggested by Chris. 
Could you please review it and let me know if I need to make some changes?

app/controllers/podcast_ownerships_controller.rb
```diff
@@ -0,0 +1,42 @@
+class PodcastOwnershipsController < ApplicationController
+  before_action :authenticate_user!
+
+  def new
+    @podcast_ownership = PodcastOwnership.new
+  end
+
+  def create
+    @podcast_ownership = PodcastOwnership.new(podcast_ownership_params)
```
rhymes, line 9: where is `podcast_ownership_params` defined?

app/controllers/podcast_ownerships_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def update
+    @podcast_ownership = PodcastOwnership.find(params[:id])
+
+    if @podcast_ownership.update(podcast_ownership_params)
+      @podcast_ownership.save
+      flash[:success] = "Updated successfully"
+      redirect_to pod_path
+    else
+      flash[:error] = "Error updating!"
```
rhymes, line 32: It's probably better to add `@podcast_ownership.errors_as_sentence` to the error message, this way you can tell the client which errors are present:

https://github.com/forem/forem/blob/910f79e04e402214f33600cddb44367e9c6e5d45/app/models/application_record.rb#L56-L58

app/controllers/podcast_ownerships_controller.rb
```diff
@@ -0,0 +1,42 @@
+class PodcastOwnershipsController < ApplicationController
```
rhymes, line 1: This controller doesn't have neither tests nor templates, I wonder if it should be added in its own separate PR, after we merge the table (it seems also off topic, as mentioned by the title)

app/models/podcast.rb
```diff
@@ -2,6 +2,7 @@ class Podcast < ApplicationRecord
   resourcify
 
   has_many :podcast_episodes
+  has_many :users, through: :podcast_ownerships
```
rhymes, line 5: is the goal to have multiple owners on a podcast? It might be beneficial to call this `.owners` instead of `.users`, similar to how we call `.creator` instead of `.user` the creator of the podcast

app/models/podcast_ownership.rb
```diff
@@ -0,0 +1,4 @@
+class PodcastOwnership < ApplicationRecord
+  belongs_to :user
+  belongs_to :podcast
+end
```
rhymes, line 4: Same here, you could do something like:

```ruby
belongs_to :owner, class_name: "User", inverse_of: :owners
belongs_to :podcast, inverse_of: :podcasts
```

app/policies/podcast_ownership_policy.rb
```diff
@@ -0,0 +1,27 @@
+class PodcastOwnershipPolicy < ApplicationPolicy
+  def update?
+    user_is_owner? || user_admin?
+  end
+
+  def new?
+    true
```
rhymes, line 7: probably should be the same as `create?`, forbidding this action to banned users

app/policies/podcast_ownership_policy.rb
```diff
@@ -0,0 +1,27 @@
+class PodcastOwnershipPolicy < ApplicationPolicy
+  def update?
+    user_is_owner? || user_admin?
```
rhymes, line 3: should we check `user_is_banned?` here as well?

app/policies/podcast_ownership_policy.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def destroy?
+    update?
+  end
+
+  private
+
+  def user_is_owner?
+    if record.instance_of?(PodcastOwnership)
```
rhymes, line 21: what's the scenario for it to be something other than a `PodcastOwnership` ?

db/migrate/20201114130315_create_podcast_ownerships.rb
```diff
@@ -0,0 +1,9 @@
+class CreatePodcastOwnerships < ActiveRecord::Migration[6.0]
+  def change
+    create_table :podcast_ownerships do |t|
+      t.references :podcast, index: true, foreign_key: true
+      t.references :user, index: true, foreign_key: true
```
rhymes, line 5: can you add `null: false` here? Unfortunately by default [add_reference](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_reference) defaults to `null: true`

rhymes (changes requested):
Hi @Manasa2850, welcome to the Forem repo! Thank you for the contribution.

I'm unfamiliar with this feature at the moment, would you mind filling in the rest of the pull request template? This way I can help review it better. Right now I just checked the code as is and left some notes

rhymes:
@Manasa2850 given the conflicts I suspect you haven't updated your local branch before submitting this, can you also please take a look at that, I suggest solving the conflict before addressing any other change

Manasa2850:
Hi @rhymes. Thanks for the review!
This is the issue I'm working on https://github.com/MLH-Fellowship/dev.to/issues/81

app/controllers/podcast_ownerships_controller.rb
Manasa2850: PodcastOwnership is the join table between users and podcasts so we get the params from there

app/controllers/podcast_ownerships_controller.rb
Manasa2850: Hmm yeah.. Chris suggested that I add a Pundit policy and a controller too...hence I added it here...we can have it in a separate PR if required too!

rhymes:
@Manasa2850 thanks for the details!

Manasa2850:
@rhymes I'll resolve the merge conflicts and make the changes you've suggested!

app/policies/podcast_ownership_policy.rb
Manasa2850: The user who tries to edit the podcast need not be the owner. So if it's not an instance of PodcastOwnership (i.e the user is not the owner of the podcast), they will not be permitted to modify it.

app/controllers/podcast_ownerships_controller.rb
Manasa2850: @rhymes I'll also be working on a follow up issue https://github.com/MLH-Fellowship/dev.to/issues/85
So I can remove the controller and the associated pundit policy from this PR if you wish

rhymes: Yeah, I think it'd be simpler to merge one bit at the time, this way we don't merge broken code (if anyone accesses the controller URL they will get a 400 error because the templates aren't there)

app/controllers/podcast_ownerships_controller.rb
rhymes: Yeah but the method `podcast_ownership_params` doesn't exist :) I guess you'll have to define it in a similar fashion to https://github.com/forem/forem/blob/910f79e04e402214f33600cddb44367e9c6e5d45/app/controllers/admin/users_controller.rb#L167-L175

but as suggested, I think this should all split in a separate PR :)

app/policies/podcast_ownership_policy.rb
rhymes: I'm not sure I understand fully that it corresponds to the logic you wrote but once you'll write a test for the policy you can test its feasibility. See https://github.com/forem/forem/tree/master/spec/policies for policy specs

rhymes:
@Manasa2850 I think you have quite a few issues derived by conflict resolution. 
For example https://github.com/forem/forem/pull/11410/files#diff-9802ca3c9c4cf89904fd44bc114e35ebdf2c5dd3d5b645491e2b253e1afef29bR119 contains something that shouldn't be there: `has_many :page_views, dependent: :destroy` is already present a few line below as `has_many :page_views, dependent: :nullify` which is the correct statement

The `schema.rb` also is completely off :-( check the diff here https://github.com/forem/forem/pull/11410/files#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3 

that's why Travis isn't working

app/controllers/podcast_ownerships_controller.rb
Manasa2850: Okay @rhymes I'll remove the controller and the pundit policy files from this PR. We can keep this PR just for the join table.
Thanks for suggesting that!

Manasa2850:
@rhymes I would like to create a new PR by rebasing and removing the controller and pundit policy. I'll be closing this one.
Hope that's fine :)

Manasa2850:
@rhymes I've resolved the merge conflicts!
I've also removed the controller and pundit policy files from this PR. I'll add it in another PR
Please let me know if something more needs to be done

Manasa2850:
I'm not sure why the `<<<<<<< HEAD` line is coming. Sorry I'm a little new to resolving merge conflicts through the command line

app/models/podcast.rb
```diff
@@ -1,6 +1,8 @@
 class Podcast < ApplicationRecord
   resourcify
 
+  has_many :podcast_episodes
```
rhymes, line 4: this is already present on line 7

app/models/podcast.rb
rhymes: see my note above about explicitly naming this `.owners`

app/models/podcast_ownership.rb
rhymes: see my note above about renaming this to `owner` if possible

app/models/user.rb
```diff
@@ -120,6 +120,7 @@ def sync_profile
   # we keep page views as they belong to the article, not to the user who viewed it
   has_many :page_views, dependent: :nullify
 
+  has_many :podcasts, through: :podcast_ownerships
```
rhymes, line 123: as https://github.com/MLH-Fellowship/dev.to/issues/82 is coming down the pipeline, the association here should be named `:owned_podcasts`, so that a user can be the owner of one or more podcasts and also appear in one or more podcasts

db/migrate/20201114130315_create_podcast_ownerships.rb
rhymes: don't forget `null: false`

rhymes (changes requested):
Hi @Manasa2850, there a few details to adjust but we're almost there :) See my notes below, thank you!

app/models/user.rb
Manasa2850: @rhymes can we check this once with @benhalpern before changing it? Because in my previous meeting with Ben and Chris, we had decided on `podcast_ownerships`

Manasa2850:
@rhymes I'm getting this error in the Travis CI test `ActiveRecord::PendingMigrationError:` 
Could you please help me resolve this?

Manasa2850:
@rhymes I ran the command `rails db:migrate RAILS_ENV=test` like mentioned in the Travis CI log. But this changes my `db/schema.rb` completely which then leads to merge conflicts :(

app/models/user.rb
rhymes: @Manasa2850 what I meant is changing `user.podcasts` to `user.owned_podcasts` or `user.podcasts_owned`

that's because https://github.com/forem/forem/pull/11354 will add the concept of "podcasts that I have been a guest on" which is different from "podcasts that I owned and manage". They can't be both called `podcasts`. The name `through :podcast_ownerships` is correct :)

app/models/podcast_ownership.rb
```diff
@@ -0,0 +1,6 @@
+class PodcastOwnership < ApplicationRecord
+  belongs_to :owner, class_name: "User"
+  belongs_to :podcast
+  validates :podcast_id, uniqueness: { scope: :owner_id }
+  validates :podcast_id, :owner_id, :role, presence: true
```
rhymes, line 5: I believe this has been mutuated from https://github.com/forem/forem/pull/11354/files#diff-80ed95ceccb910e831a5c371a28a8b35b37729d2d4ad1d063702c9a084086b1fR4-R5 but this model doesn't have the `role` column, and it shouldn't have.

The uniqueness validation if needed needs to be backed by a DB index, see https://github.com/forem/forem/pull/11354/files#diff-1953bb445767352adbf509787151f41c6f293305f5b4d65189ccb33a15a261c3R11

ps. the name of the column is `user_id`, `:owner` it's just an alias to help us humans in the application logic :), if you check [belongs_to](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-belongs_to) options section you can find how `belongs_to :owner, class_name: "User"` works

rhymes (changes requested):
@Manasa2850 AFAIK the `schema.rb` in this PR doesn't correspond to what's written in `db/migrate/20201114130315_create_podcast_ownerships.rb`:

```ruby
    t.bigint "podcast_id"
    t.bigint "user_id"
```

are missing `null: false`

and the version of the schema.rb file is wrong. In the diff is `2020_11_10_180614` but the migration file you added is `db/migrate/20201114130315_create_podcast_ownerships.rb` so the version should be `2020_11_14_130315`, which is why Travis is saying that a migration is pending, because the schema is one migration behind and thus it thinks there's still one to be run

ps. don't forget the model tests (also run them locally if possible so you can anticipate Travis errors) as explained in my previous review here https://github.com/forem/forem/pull/11410#pullrequestreview-530771966

Manasa2850:
@rhymes I get this error in Travis CI `Factory not registered: "podcast_ownership"` in the spec file. I'm a little new to writing tests. Could you please help me out with this?

rhymes:
Sure @Manasa2850, as `podcast_ownership` is a new class it needs a factory for tests. A factory is a set of instructions that builds a real model object, through a shortcut. If you look into https://github.com/forem/forem/tree/master/spec/factories you'll find one per each model. What Travis is telling you is that `FactoryBot` (the gem we use to build objects in tests) is not aware of `podcast_ownership`, thus you just need a new `spec/factories/podcast_ownership.rb` file (no need to explicitly register anything) which contains something similar to:

https://github.com/forem/forem/blob/994fbf99ad5f60f6a1ff7db2f7f874aa4b186340/spec/factories/rating_votes.rb#L1-L3

For example, I'd imagine, since podcast ownership needs a user and a podcast, you could do:

```ruby
FactoryBot.define do
  factory :podcast_ownership do
    owner { user }
    podcast
  end
end
```

(I haven't tested it, check [FactoryBot's documentation](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#specifying-the-factory))

`FactoryBot` when inside a test you then call `create(:podcast_ownership)` will load or create a user object and a podcast object automatically.

Anyway, that won't work until the definition of the various associations is fixed as some are incomplete. If you load the console and try to create a `PodcastOwnership` object you will see:

```ruby
[4] pry(main)> po = PodcastOwnership.create!(owner: User.last, podcast: Podcast.last)
  User Load (0.9ms)  SELECT "users".* FROM "users" ORDER BY "users"."id" DESC LIMIT $1  [["LIMIT", 1]]
  Podcast Load (0.3ms)  SELECT "podcasts".* FROM "podcasts" ORDER BY "podcasts"."id" DESC LIMIT $1  [["LIMIT", 1]]
ActiveModel::MissingAttributeError: can't write unknown attribute `owner_id`
from /Users/rhymes/.gem/ruby/2.7.2/gems/activemodel-6.0.3.4/lib/active_model/attribute.rb:206:in `with_value_from_database'
```

see the `:foreign_key` section in Rails's [belongs_to](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-belongs_to) documentation.

I suggest you load the console and play with creating `PodcastOwnership` objects manually, see if then you can retrieve them from `user.owned_podcasts` and `podcast.owners`

Good job so far!

Manasa2850:
@rhymes I added the factory. I also tested it out on rails console!
There's a merge conflict regarding the `schema.rb` version. If I rebase that, I'll get the `ActiveRecord::PendingMigrationError:`
Could you let me know what I need to do?

Manasa2850:
@rhymes could you help me with the Travis CI error? `Expected PodcastOwnership to have a belongs_to association called owner (PodcastOwnership does not have a owner_id foreign key.)`
When I had written `it { is_expected.to belong_to(:owner) }` in the spec file, I got the error `Expected PodcastOwnership to have a belongs_to association called user (no association called user)`

rhymes:
Hi @Manasa2850, I'll take a look, would it be okay with you if I went ahead and adjusted the relations with a commit?

Manasa2850:
@rhymes. Sure you can do that. Thank you

rhymes:
@Manasa2850 I can't push to your branch, but I exported the patch file in a gist: https://gist.githubusercontent.com/rhymes/98c74c5c190508105ca717f0406719b9/raw/78820d192c0394325c56ea89d6c8fcd866bdbf36/0001-Adjust-various-many-to-many-relations.patch

If you download it you can apply it locally with:

```shell
$ git am 0001-Adjust-various-many-to-many-relations.patch
$ rm 0001-Adjust-various-many-to-many-relations.patch
```

and then you can push it here with `git push`

db/schema.rb
```diff
@@ -1263,7 +1272,7 @@
     t.datetime "last_article_at", default: "2017-01-01 05:00:00"
     t.datetime "last_comment_at", default: "2017-01-01 05:00:00"
     t.datetime "last_followed_at"
-    t.datetime "last_moderation_notification", default: "2017-01-01 05:00:00"
+    t.datetime "last_moderation_notification", default: "2016-12-31 18:30:00"
```
rhymes, line 1275: this needs to be manually reverted to the value it has on the left

rhymes (commented):
Aside from the patch and the `schema.rb` fix there's one last thing that we need to take care of before we merge this in: uniqueness.

Right now a user can be the owner of a podcast multiple times, but I'm sure we want one user to appear as the owner of a podcast only once. How can we fix that? By adding the combination of a unique index on both `podcast_id` and `user_id` in the `podcast_ownerships` table and a validation for such uniqueness in the model file.

You can take inspiration from https://github.com/forem/forem/blob/cf619caf70f5ed129303b5b701b6e438913ed350/app/models/article.rb#L66 and https://github.com/forem/forem/pull/11354/files#diff-1953bb445767352adbf509787151f41c6f293305f5b4d65189ccb33a15a261c3R11 for example



Sorry that wasn't made clear before hand.

Manasa2850:
@rhymes I've added everything you've suggested. I guess I need to make some changes to `spec/services/users/delete_spec.rb` too because that's causing the Travis CI build to fail. Could you let me know what changes I need to make there? Thanks!

rhymes:
@Manasa2850 yeah, I can look into that. Thanks for adding the index in the migration, there's one last step: https://github.com/forem/forem/pull/11410/files#diff-82c6591538e42fcabb8338269551c7fee28225c07759c053023d25844609194dR8 is missing the corresponding index in the schema (you need to rollback the migration and re-run it so that it picks up the new index)

rhymes:
@Manasa2850 the test should be fixed by changing this in the `app/models/user.rb` on line 121:

```ruby
  has_many :podcast_ownerships, dependent: :destroy, inverse_of: :owner
```

what's missing is the indication of the inverse of that relation

rhymes:
Hi @Manasa2850, to solve the following:

```text
  1) PodcastOwnership validations is expected to validate that :podcast_id is case-sensitively unique within the scope of :user_id

     Failure/Error: it { is_expected.to validate_uniqueness_of(:podcast_id).scoped_to(:user_id) }

     

     ActiveRecord::InvalidForeignKey:

       PG::ForeignKeyViolation: ERROR:  insert or update on table "podcast_ownerships" violates foreign key constraint "fk_rails_3710d65292"

       DETAIL:  Key (podcast_id)=(0) is not present in table "podcasts".
```

you can add this to the test:

```ruby
  describe "validations" do
    subject { podcast_ownership }
```

By default rspec tries to create an empty object, but that's not going to work because of the foreign key constraint as mentioned in the error, but if we tell it to recycle an already valid object, then the error goes away.

Manasa2850:
@rhymes thank you! I'll do that!

db/schema.rb
```diff
@@ -878,6 +878,15 @@
     t.index ["website_url"], name: "index_podcast_episodes_on_website_url"
   end
 
+  create_table "podcast_ownerships", force: :cascade do |t|
+    t.datetime "created_at", precision: 6, null: false
+    t.bigint "podcast_id"
+    t.datetime "updated_at", precision: 6, null: false
+    t.bigint "user_id"
+    t.index ["podcast_id"], name: "index_podcast_ownerships_on_podcast_id", unique: true
+    t.index ["user_id"], name: "index_podcast_ownerships_on_user_id", unique: true
```
rhymes, line 887: I don't this is correct @Manasa2850, it doesn't reflect what's written in `db/migrate/20201114130315_create_podcast_ownerships.rb`

If you run the revert the migration and re-run it you'll notice the following changes from this version:

```diff
diff --git a/db/schema.rb b/db/schema.rb
index 526190a79..00ddfdfa7 100644
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -880,11 +880,12 @@
 
   create_table "podcast_ownerships", force: :cascade do |t|
     t.datetime "created_at", precision: 6, null: false
-    t.bigint "podcast_id"
+    t.bigint "podcast_id", null: false
     t.datetime "updated_at", precision: 6, null: false
-    t.bigint "user_id"
-    t.index ["podcast_id"], name: "index_podcast_ownerships_on_podcast_id", unique: true
-    t.index ["user_id"], name: "index_podcast_ownerships_on_user_id", unique: true
+    t.bigint "user_id", null: false
+    t.index ["podcast_id", "user_id"], name: "index_podcast_ownerships_on_podcast_id_and_user_id", unique: true
+    t.index ["podcast_id"], name: "index_podcast_ownerships_on_podcast_id"
+    t.index ["user_id"], name: "index_podcast_ownerships_on_user_id"
   end
```

A tip: now that we have a unique index on both columns, we don't need the individual indexes on each column, see `index: true` in the migation file. 

Could you address that?

Basically you need to run:

- `bundle exec rails db:rollback` or `rails db:migrate:down VERSION=20201114130315`
- remove the `index: true` from both columns in the migration file
- re-run `bundle exec rails db:migrate`
- check the changes to the schema and commit those that pertain to `podcast_ownerships`
- push the code

Let me know if anything isn't clear

rhymes (changes requested):

db/schema.rb
Manasa2850: @rhymes I ran the migration again!

rhymes: > A tip: now that we have a unique index on both columns, we don't need the individual indexes on each column, see index: true in the migation file.

Can you also remove the `index: true` from the singular fields? See my instructions above

rhymes: @Manasa2850 don't forget to re-run the migration after having removed `index: true`

Manasa2850: @rhymes yup I've done that!

rhymes: These two lines tell me that the migration still contain separate indexes for those columns:

```ruby
    t.index ["podcast_id"], name: "index_podcast_ownerships_on_podcast_id"
    t.index ["user_id"], name: "index_podcast_ownerships_on_user_id"
```

If you check the documentation for [references/add_reference](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_reference) you'll notice how `true` is the default for the `index` param. I guess it needs to be set as `false`. Could you address that?

Manasa2850: @rhymes running `rails db:migrate` was still not reflecting the change :(
So I had to remove those 2 lines manually

rhymes: @Manasa2850 that's because you need to set explicitly `index: false` in the migration, see my comment above. The migration file and the schema need to reflect each other, they can't be misaligned :)

rhymes (approved):
Thanks for the changes @Manasa2850! 

LGTM

Manasa2850:
@rhymes thanks a ton for all the help!

mstruve (approved):
Welcome to Forem @Manasa2850! I love how clean and concise this code is. I always cringe when PRs try to do too much at once but you really kept this nice and contained which made reviewing it a piece of cake. 

Hope to see you tackling other issues in the future!

Ridhwana (approved):

PR #11411, opened 2020-11-14
Add partial index on users.feed_url

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Rows with `feed_url` populated are traversed each time the rss/feed reader/import runs, but there's no index on those. We add a partial index as rows with `feed_url` are a tiny percentage of the total (less than 0.7% on DEV for example).

Kudos to @citizen428 for the initial suggestion https://github.com/forem/forem/pull/11234#discussion_r516391885

https://github.com/forem/forem/pull/10998

## Added tests?

- [ ] Yes
- [x] No, and this is why: we don't explicitly test for the existence of DB index, that might change in the future, that future isn't now though :D

## Added to documentation?

- [ ] README
- [x] No documentation needed

### DIFF
--- a/db/migrate/20201114151157_add_partial_index_to_users_feed_url.rb
+++ b/db/migrate/20201114151157_add_partial_index_to_users_feed_url.rb
@@ -1,0 +1,9 @@
+class AddPartialIndexToUsersFeedUrl < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def change
+    # adds an index only on users with a feed URL, as they are a tiny percentage of the total
+    # number of users, there is no need to add a full index as most entries will be empty
+    add_index :users, :feed_url, where: "COALESCE(feed_url, '') <> ''", algorithm: :concurrently
+  end
+end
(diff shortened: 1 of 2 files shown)


atsmith813 (approved):

db/migrate/20201114151157_add_partial_index_to_users_feed_url.rb
```diff
@@ -0,0 +1,9 @@
+class AddPartialIndexToUsersFeedUrl < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def change
+    # adds an index only on users with a feed URL, as they are a tiny percentage of the total
+    # number of users, there is no need to add a full index as most entries will be empty
+    add_index :users, :feed_url, where: "COALESCE(feed_url, '') <> ''", algorithm: :concurrently
```
mstruve, line 7: TIL about `COALESCE` and `<>` in postgres!

mstruve (approved):

PR #11412, opened 2020-11-14
Refactor RssReader and Feeds::Import common parts: take 1

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This is an attempt of extracting auxiliary services that are common between the two, they also shorten the length of the classes, making it easier to understand the content without getting lost in the minutiae

#10998 

## Added tests?

- [ ] Yes
- [x] No, and this is why: the point of the refactor is to have the functional tests on both `RssReader` and `Feeds::Import` pass as they are without changes, regression testing. Testing the singular auxilary services is not worth it as they aren't used outside the context of the feeds importers

## Added to documentation?

- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/models/tweet.rb
+++ b/app/models/tweet.rb
@@ -22,7 +22,7 @@ class Tweet < ApplicationRecord
       text.gsub!("@#{uname}", "<a href='https://twitter.com/#{uname}'>#{"@#{uname}"}</a>")
     end
     hashtags_serialized.each do |tag|
-      tag_text = tag[:text]
+      tag_text = tag.text
       text.gsub!("##{tag_text}",
                  "<a href='https://twitter.com/hashtag/#{tag_text}'>#{"##{tag_text}"}</a>")
     end
--- a/app/services/feeds/check_item_medium_reply.rb
+++ b/app/services/feeds/check_item_medium_reply.rb
@@ -1,4 +1,4 @@
-# Check if a Feedjira item represents a Medium reply or not
+# Checks if a Feedjira item represents a Medium reply or not
 module Feeds
   class CheckItemMediumReply
     MEDIUM_DOMAIN = "medium.com".freeze
--- a/app/services/feeds/check_item_medium_reply.rb
+++ b/app/services/feeds/check_item_medium_reply.rb
@@ -1,4 +1,5 @@
-# Checks if a Feedjira item represents a Medium reply or not
+# Checks if a Feedjira item represents a user's comment on a Medium post as
+# unfortunately they include those as items in their feeds
 module Feeds
   class CheckItemMediumReply
     MEDIUM_DOMAIN = "medium.com".freeze
--- a/app/services/feeds/check_item_medium_reply.rb
+++ b/app/services/feeds/check_item_medium_reply.rb
@@ -1,0 +1,37 @@
+# Check if a Feedjira item represents a Medium reply or not
+module Feeds
+  class CheckItemMediumReply
+    MEDIUM_DOMAIN = "medium.com".freeze
+
+    def self.call(item)
+      new(item).call
+    end
+
+    def initialize(item)
+      @item = item
+    end
+
+    def call
+      get_host_without_www(item.url.strip) == MEDIUM_DOMAIN &&
+        !item[:categories] &&
+        content_is_not_the_title?(item)
+    end
+
+    private
+
+    attr_reader :item
+
+    def get_host_without_www(url)
+      url = "http://#{url}" if URI.parse(url).scheme.nil?
+      host = URI.parse(url).host.downcase
+      host.start_with?("www.") ? host[4..] : host
+    end
+
+    def content_is_not_the_title?(item)
+      # [[:space:]] removes all whitespace, including unicode ones.
+      content = item.content.gsub(/[[:space:]]/, " ")
+      title = item.title.delete("…")
+      content.include?(title)
+    end
+  end
+end
--- a/app/services/feeds/check_item_previously_imported.rb
+++ b/app/services/feeds/check_item_previously_imported.rb
@@ -1,0 +1,25 @@
+# Checks if an item has been previously imported for the given user
+# Item transformed as Article objects are unique per user, not globally
+module Feeds
+  class CheckItemPreviouslyImported
+    def self.call(item, user)
+      new(item, user).call
+    end
+
+    def initialize(item, user)
+      @item = item
+      @user = user
+    end
+
+    def call
+      title = item.title.strip.gsub('"', '\"')
+      feed_source_url = item.url.strip.split("?source=")[0]
+      relation = user.articles
+      relation.where(title: title).or(relation.where(feed_source_url: feed_source_url)).exists?
+    end
+
+    private
+
+    attr_reader :item, :user
+  end
+end
--- a/app/services/feeds/import.rb
+++ b/app/services/feeds/import.rb
@@ -142,7 +142,7 @@ module Feeds
           feeds_import_info: {
             username: user.username,
             feed_url: user.feed_url,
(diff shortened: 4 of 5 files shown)


app/models/tweet.rb
```diff
@@ -22,7 +22,7 @@ def processed_text
       text.gsub!("@#{uname}", "<a href='https://twitter.com/#{uname}'>#{"@#{uname}"}</a>")
     end
     hashtags_serialized.each do |tag|
-      tag_text = tag[:text]
+      tag_text = tag.text
```
rhymes, line 25: I was getting this deprecation error while running the tests:

> /forem/app/models/tweet.rb:25:in `block in processed_text': [DEPRECATION] #[:text] is deprecated. Use #text to fetch the value.

app/services/feeds/check_item_medium_reply.rb
```diff
@@ -0,0 +1,37 @@
+# Checks if a Feedjira item represents a Medium reply or not
+module Feeds
+  class CheckItemMediumReply
+    MEDIUM_DOMAIN = "medium.com".freeze
```
mstruve, line 4: My only thought here would be to remove the specific domain from the class name and make the class accept the domain as an argument. That way if we had to do this for another domain we could use the same class. Not a blocker but a thought I had to make this class more vendor agnostic.

mstruve (approved):
🔥

app/services/feeds/check_item_medium_reply.rb
rhymes: Good idea! I initially wondered that there might have been a reason why the method was called `medium_reply?` and not just "reply?("medium.com")". I actually wanted to ping @maestromac about this but then I forgot. It seems due to some inconstencies in how Medium formats RSS feeds. Do you know?

mstruve: I do not know, @maestromac would be the one to ask or maybe @Zhao-Andy would know. I'm not very familiar with the RSS import stuff.

maestromac: Medium seems to be the only one that includes the user's comment in their RSS feed, for some reason.

app/services/feeds/import.rb
```diff
@@ -122,7 +122,7 @@ def create_articles_from_user_feed(user, feed)
       articles = []
 
       feed.entries.reverse_each do |item|
-        next if medium_reply?(item) || article_exists?(user, item)
+        next if Feeds::CheckItemMediumReply.call(item) || Feeds::CheckItemPreviouslyImported.call(item, user)
```
citizen428, line 125: These two services only seem to be used together, in a guard clause. I'm starting to wonder if we should tie them together with another service, e.g. `Feeds::ShouldProcessItem` or something? It could be really simple for now:

```ruby
module Feeds
  class ShouldProcessItem
    def self.call(item, user)
      CheckItemMediumReply.call(item) && CheckItemPreviouslyImported.call(item, user)
    end
  end
end
```

Then the guard clauses need to switch:

```ruby
next unless Feed::ShouldProcessItem.call(item, user)
return unless Feed::ShouldProcessItem.call(item, user)
```

If we ever need more checks they wouldn't need to be added in two different places.

app/services/rss_reader.rb
```diff
@@ -95,34 +87,16 @@ def make_from_rss_item(item, user, feed)
...
-    relation = user.articles
-    relation.where(title: title).or(relation.where(feed_source_url: feed_source_url)).exists?
-  end
-
   def report_error(error, metadata)
     Honeybadger.context(metadata)
     Honeybadger.notify(error)
   end
+
+  def get_item_count_error(feed)
```
citizen428, line 95: I know you just moved this, but maybe it would be a good opportunity to make this a bit more idiomatic Ruby:

1. Remove `get_` prefix.
2. Use guard clause.

```ruby
def item_count_error(feed)
  return "NIL FEED, INVALID URL" unless feed

  feed.entries ? feed.entries.length : "no count"
end
```

As a bonus, the strings could maybe become constants.

app/services/rss_reader.rb
```diff
@@ -95,34 +87,16 @@ def make_from_rss_item(item, user, feed)
...
+    if feed
+      feed.entries ? feed.entries.length : "no count"
```
citizen428, line 97: I'm wondering about the conditional here. I thought `entries` always returns an array, will it really return `nil` if there are items?

app/services/feeds/check_item_previously_imported.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    def initialize(item, user)
+      @item = item
+      @user = user
+    end
+
+    def call
+      title = item.title.strip.gsub('"', '\"')
+      feed_source_url = item.url.strip.split("?source=")[0]
```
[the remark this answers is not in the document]
rhymes: Are you sure? If the URL doesn't contains `?source` split will be the string:

```ruby
> "foobar".split("?source=")[0]
=> "foobar"
```

If the URL is blank yes this will be nil, but I think that's why it also looks for the title. We could make the query smarter to don't check both if one is missing but I think these optimizations can wait for later. I'll make a not for it!

BTW I just copied and pasted it from the existing code:

https://github.com/forem/forem/blob/cf619caf70f5ed129303b5b701b6e438913ed350/app/services/rss_reader.rb#L117-L122

app/services/feeds/import.rb
rhymes: Good call, though I'm not 100% convinced on adding a third service that's just a wrapper for two. It'd be nice to use dependency injection, so something like:

```
def call(item, user, checkers: [array of services to send .call on])
  # something like checkers.all? { |c| c.call?(item, user)
end
```

but they both feel a bit over-engineered for now

app/services/rss_reader.rb
rhymes: Added 1 and 2, I haven't added the constants because I definitely want to change how we signal "invalid url" using symbols, not generic strings. Something like `:invalid_url` and `:zero_count`. Made a note of this as well

app/services/rss_reader.rb
rhymes: 🤷 it might not be true anymore, hard to verify with a regression test. I don't want to poke the bear :D

app/services/feeds/check_item_previously_imported.rb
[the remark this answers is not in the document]
citizen428: Yes, my mistake, I had only tested the blank case, not URLs which don't contain "?source=". I agree it can wait though.

app/services/feeds/import.rb
citizen428: Yes, fair point, I'm not saying we should do either. I guess something here just rubs me the wrong way, e.g. if they are only used together, maybe this split was premature and both of them together should have moved into a single validity checker service.

app/services/rss_reader.rb
citizen428: Hm, the difference between symbols and frozen strings are minimal in this specific case (MRI implementation details aside): neither is resistant to typos while checking nor is there a significant difference in expressiveness. The name of the method makes me wonder if these should be actual errors, e.g. exception classes. "Number of entries or error message/symbol" is an unwieldy return type.

citizen428 (approved):
@rhymes I left a few more comments/some food for thought, but nothing to hold this PR back.

app/services/rss_reader.rb
rhymes: ahhaha i agree, that method is a bit odd

PR #11413, opened 2020-11-15
Allow iframe inside details liquid tag

### LINKED ISSUE #11060: Article rendering: YouTube embed in a collapsible section
YouTube embed (`{% youtube <id> %}`) is not being rendered if it appears inside the collapsible details, like this:

```
{% details <summary> %}
    {% youtube <id> %}
{% enddetails %}
```

However, some other liquid tags, e.g. Github embed, do work inside. Is it intentional behavior or a bug?

akashdotsrivastava:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Added iframe and its missing attributes to the list of allowed tags and attributes for `sanitize` Rails helper, which allows `iframe` elements in details tag to be parsed correctly when sanitized.

**Additional Changes**
The issue was with `sanitize` used in https://github.com/forem/forem/blob/910f79e04e402214f33600cddb44367e9c6e5d45/app/liquid_tags/details_tag.rb#L13 It by default, does not allow `iframe` as a tag while sanitizing, hence the youtube tag render, an iframe equivalent, though present in the xpath available, was not passing `sanitize`.

Unfortunately, there is no way to `add`/`insert` a new tag to the list of default allowed tags for `sanitize`. It either takes the default list (with no `tags` and `attributes` option), or only takes a list of tags and attributes to be allowed while sanitization, through array options `tags` and `attributes`. https://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html#method-i-sanitize

To allow `iframe` along with all the existing tags and attributes by default, I picked the default lists of tags and attributes from https://github.com/rails/rails-html-sanitizer/blob/51dc564c6509201070f72456bb2c13f87bb373d6/lib/rails/html/sanitizer.rb#L108..L111. I added the required tag for youtube video `iframe` and a couple of missing attributes for the same in the attributes list.

Closes #11060 

1. Create a new post with this content (a details tag with some text content and a youtube video tag)
```
{% details Details and Video %}
    This is a youtube video
    {% youtube 3xTiHxHDb4U %}
{% enddetails %}
```
2. Look at preview and observe both text and video rendering correctly when details arrow is clicked

Screenshot 2020-11-15 at 11 07 33 AM
Screenshot 2020-11-15 at 11 07 51 AM

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

alt_text

### DIFF
--- a/app/liquid_tags/details_tag.rb
+++ b/app/liquid_tags/details_tag.rb
@@ -3,6 +3,12 @@ class DetailsTag < Liquid::Block
 
   PARTIAL = "liquids/details".freeze
 
+  SANITIZATION_ALLOWED_TAGS = %w[strong em b i p code pre tt samp kbd var sub sup dfn cite big
+                                 small address hr br div span h1 h2 h3 h4 h5 h6 ul ol li dl dt
+                                 dd abbr acronym a img blockquote del ins iframe].freeze
+  SANITIZATION_ALLOWED_ATTRIBUTES = %w[href src width height alt cite datetime title class name
+                                       xml:lang abbr allowfullscreen loading].freeze
+
   def initialize(_tag_name, summary, _parse_context)
     super
     @summary = sanitize(summary.strip)
@@ -10,7 +16,8 @@ class DetailsTag < Liquid::Block
 
   def render(_context)
     content = Nokogiri::HTML.parse(super)
-    parsed_content = sanitize(content.xpath("//html/body").inner_html)
+    parsed_content = sanitize(content.xpath("//html/body").inner_html, tags: SANITIZATION_ALLOWED_TAGS,
+                                                                       attributes: SANITIZATION_ALLOWED_ATTRIBUTES)
 
     ApplicationController.render(
       partial: PARTIAL,
--- a/app/liquid_tags/details_tag.rb
+++ b/app/liquid_tags/details_tag.rb
@@ -3,11 +3,8 @@ class DetailsTag < Liquid::Block
 
   PARTIAL = "liquids/details".freeze
 
-  SANITIZATION_ALLOWED_TAGS = %w[strong em b i p code pre tt samp kbd var sub sup dfn cite big
-                                 small address hr br div span h1 h2 h3 h4 h5 h6 ul ol li dl dt
-                                 dd abbr acronym a img blockquote del ins iframe].freeze
-  SANITIZATION_ALLOWED_ATTRIBUTES = %w[href src width height alt cite datetime title class name
-                                       xml:lang abbr allowfullscreen loading].freeze
+  SANITIZATION_ALLOWED_TAGS = Rails::Html::SafeListSanitizer.allowed_tags + %w[iframe]
+  SANITIZATION_ALLOWED_ATTRIBUTES = Rails::Html::SafeListSanitizer.allowed_attributes + %w[allowfullscreen loading]
 
   def initialize(_tag_name, summary, _parse_context)
     super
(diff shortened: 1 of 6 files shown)


rhymes (changes requested):
Hi @akashdotsrivastava, thanks for the changes!

A couple of things: I think we can use 

```ruby
Rails::Html::SafeListSanitizer.allowed_tags + ["iframe"]
```

same for the attributes. This way we don't have to duplicate the default allowed tags list.

Regarding allowing the iframe, could you add a few test cases? I suspect this change will work not only with YouTube but any of the other liquid tags that renders as an `<iframe>`

I tried with `codesandbox` and it worked:

Screenshot_2020-11-15 details + codesandbox


1. it'd be nice to preserve the previous simple case for `<details>`
1. it'd be nice to add this case with the iframe separately
1. finally it'd be nice if we had a separate test case, maybe beside this https://github.com/forem/forem/blob/910f79e04e402214f33600cddb44367e9c6e5d45/spec/system/articles/user_creates_an_article_spec.rb#L16 that checks that one of the other combo of `<details>` + embedded liquid tag works (one is sufficient, it's more of a regression test). 

What do you think?

akashdotsrivastava:
> What do you think?

@rhymes Thanks for the feedback. All of it makes sense. Changes done.

rhymes (approved):
Awesome, thanks @akshayymahajan!

akshayymahajan:
@rhymes, I think you wanted to tag @akashdotsrivastava 😃

rhymes:
@akashdotsrivastava @akshayymahajan sorry :-(

msarit (approved):
Great work, @akashdotsrivastava! Thanks for contributing!

rhymes:
Setting this back to draft temporarily to avoid accendital merges as we want to review it from a security perspective, thank you for the patience @akashdotsrivastava

msarit:
@rhymes Hey Rhymes, pinging you re: ensuring security on this PR 🙏🏾

citizen428:
@rhymes What are our next steps on this PR?

rhymes:
@citizen428 I think we should surface it to the team, as I'm not sure if it's safe to merge and I'd like feedback on the implementation and possible security considerations

msarit:
@rhymes @citizen428 Hey people! This PR needs to be discussed re: security concerns. 🙏🏾

msarit:
I pinged relevant players for this PR in [this issue](https://github.com/forem/security/issues/374)

maestromac:
Hey @akashdotsrivastava , thank you again for your time and contribution. Due to a combination of reasons, we can not proceed with this solution.

The problem is with our sequential parsing; We would parse markdown first then the liquid tag. If the content of this details liquid tag contains markdown, those will not be processed and remain as raw markdown. Manually allowing the iframe tag to get through would open up the possibility of XSS.

The proper path forward would be to add details and a summary tag to our `RenderedMarkdownScrubber` and eventually remove the details liquid tag. Using raw details and summary tags would eliminate the problem we're trying to solve in this PR and allow full markdown parsing within the details block.

You are more than welcome to open up another PR and address those concerns. Let me know and I am more than happy to provide you directions.

akashdotsrivastava:
@maestromac Thanks for the update. 
I'll have a look and try to come up with a different PR using the route above in some time. I will update if I need help. Thanks again ♥️👍

maestromac:
@akashdotsrivastava . It's really a 1 line change + specs. You do not have to worry about removing Details liquid tag. I will do that since it requires a bit of planning and migration.

Issue #11449, opened 2020-11-16
Give Forem Admin Ability to Customize Emojis in Dropdown Display URLs


michael-tharrington:
**Is your feature request related to a problem? Please describe.**

As a Forem Admin, I'm unable to change the emojis that appear in the display URLs for key pages of the site  — i.e. these emojis: "👨‍💻👩‍💻 " that are visible beside the page name of the URL for many DEV pages. (See image under "Additional Context" for more clarification.)

**Describe the solution you'd like**

As a Forem Admin, I'd like to be able to choose the emojis to be placed in these URLs — I think by default these emojis should be turned off. This option should most likely live in `admin/config` > Community Content.

IF we could allow Forem Admin the ability to provide us with multiple inputs here that would alternate, I think that'd be an excellent bonus.

**Describe alternatives you've considered**

Perhaps we'd rather give Forem Admins more granular controls over each individual page's dropdown/display URL appearance. This would allow the Forem Admin to set separate emojis for each page if they'd like. If we went this route, I think these granular options would live in `admin/pages` and admin would need to click into individual pages to modify the emojis that appear with each page.

**Additional context**

This search should help to show where these emojis: "👨‍💻👩‍💻 " appear - https://github.com/forem/forem/search?q=%F0%9F%91%A9%E2%80%8D%F0%9F%92%BB%F0%9F%91%A8%E2%80%8D%F0%9F%92%BB

I struggled to describe this, but these are where the emojis are placed in the dropdown/display URLs:
![Image 2020-11-16 at 5 09 58 PM](https://user-images.githubusercontent.com/16007075/99314145-a3fff780-282e-11eb-9357-45e36dc6c900.png)

vaidehijoshi:
This is something that we should definitely remove from all the hard-coded instances that @michael-tharrington linked to; it is far too DEV-specific.

I see two paths forward here:
1) Remove the hardcoded emojis entirely (which removes it from DEV, too), and allow the Forem creator to add an emoji into the SiteConfig's community name like this:

<img width="1182" alt="Screen Shot 2020-11-16 at 2 33 05 PM" src="https://user-images.githubusercontent.com/6921610/99316176-bec77180-2818-11eb-8614-1a24f6cbfda5.png">

2) Add a `SiteConfig.community_emoji`, which should be _optional_ and not required, which we substitute for the places where the emojis are currently hardcoded.

[...]

PR #6887, opened 2020-03-26
Generalize meta content and copyright year

Ridhwana:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I've updated the meta keywords to be as general as possible. The ffg has been done thus far:

- [x] Change all the meta tags with og:site_name from `The Practical DEV` to `<COMMUNITY_NAME> Community`.
- [x] Replace  <meta name="twitter:image:src" content="http://i.imgur.com/B4JNl1w.png"> to `<meta name="twitter:image:src" content="<%= SiteConfig.main_social_image %>">` consistently.
- [x] Replace  `<meta property="og:image" content="http://i.imgur.com/B4JNl1w.png" />` with `  <meta property="og:image" content="<%= SiteConfig.main_social_image %>" />`
- [x] Make sure that the url in any of the meta keys use `<%= app_url %>` as a prepend instead of `https://dev.to`
- [x] Copyright year in the footer is now dynamic using an env
- [x] Change all meta keys to be dynamic

What I have not done in this PR:

1. I have not tackled converting all assets in app/views to use env variables for meta content, since it seems to be mostly tackled in [PR 5829](https://github.com/thepracticaldev/dev.to/pull/5829/files) and I want to avoid any merge conflicts. 
2. I have not amended the meta tag with keywords. I'll tackle this in a subsequent PR as this PR has a lot of repetitive changes already. 
3. https://dev-internal.slack.com/archives/CSYHV3NJ2/p1585567243013800 separate out the social media site config.
4. There are some meta tags that I will still tackle in the subsequent PR. 

**Notify the other communities to set this variable when they deploy**

## Related Tickets & Documents
[#5178](https://github.com/thepracticaldev/dev.to/issues/5178)

## Added tests?

- [ ] yes
- [x] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.dev.to
- [ ] readme
- [x] no documentation needed

## [optional] What gif best describes this PR or how it makes you feel?

alt_text

PR #11450, opened 2020-11-16
Add community_emoji to SiteConfig

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As mentioned in https://github.com/forem/forem/issues/11449, we currently hardcode some emojis in the `<title>` tags of our site, but that locks in other Forem creators to the hardcoded emojis that we chose 😕 

Instead, we want to allow Forem creators to choose _their own_ emojis to associate with the site (or to not have an emoji at all!). This PR removes the hardcoded emojis and add an optional `community_emoji` field to `SiteConfig`, which can be blank or which a Forem creator can override with their own choice of emoji.

It also removes a hardcoded emoji on the `/connect` page, since associating an emoji with a specific _page_ is a much larger feature and seems like it deserves a product discussion.

Closes https://github.com/forem/forem/issues/11449.

Starting with a fresh Forem/SiteConfig:
1. Head to `/admin/config`. You should see the default emoji in the `community_emoji` field within `Community Content`:
one
2. Try clearing out the emoji entirely -- you should be able to do it with no problems since it is _not_ a required field:
two
3. Try adding some 🥐s for emojis. You should be able to update the SiteConfig without any issues:
three
4. Check the `<title>` tag on the articles index page. You should see your community emojis reflected in the DOM!

## Added tests?

- [x] Yes (added a new test and updated preexisting tests!)
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

**We will need to add `👩‍💻👨‍💻` as the `community_emoji` on DEV. We should also add them to forem.team and forem.dev if we want something _other than_ the default `🌱`.**

## What ~gif~ EMOJI best describes this PR or how it makes you feel?
🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐🥐

### DIFF
--- a/app/views/chat_channels/index.html.erb
+++ b/app/views/chat_channels/index.html.erb
@@ -1,4 +1,4 @@
-<% title("#{community_name} Connect 👩‍💻💬👨‍💻") %>
+<% title("#{community_name} Connect") %>
 <%= content_for :page_meta do %>
   <link rel="canonical" href="<%= app_url(connect_path) %>" />
   <meta name="description" content="<%= community_name %> Connect">
--- a/spec/requests/admin/configs_spec.rb
+++ b/spec/requests/admin/configs_spec.rb
@@ -124,6 +124,14 @@ RSpec.describe "/admin/config", type: :request do
           expect(SiteConfig.community_description).to eq(description)
         end
 
+        it "updates the community_emoji" do
+          allow(SiteConfig).to receive(:community_emoji).and_call_original
+          emoji = "🥐"
+          post "/admin/config", params: { site_config: { community_emoji: emoji },
+                                          confirmation: confirmation_message }
+          expect(SiteConfig.community_emoji).to eq(emoji)
+        end
+
         it "updates the community_name" do
           name_magoo = "Hey hey #{rand(100)}"
           post "/admin/config", params: { site_config: { community_name: name_magoo },
--- a/spec/requests/admin/configs_spec.rb
+++ b/spec/requests/admin/configs_spec.rb
@@ -124,12 +124,21 @@ RSpec.describe "/admin/config", type: :request do
           expect(SiteConfig.community_description).to eq(description)
         end
 
-        it "updates the community_emoji" do
+        it "updates the community_emoji if valid" do
           allow(SiteConfig).to receive(:community_emoji).and_call_original
           emoji = "🥐"
           post "/admin/config", params: { site_config: { community_emoji: emoji },
                                           confirmation: confirmation_message }
           expect(SiteConfig.community_emoji).to eq(emoji)
+        end
+
+        it "does not update the community_emoji if invalid" do
+          allow(SiteConfig).to receive(:community_emoji).and_call_original
+          not_an_emoji = "i love croissants"
+          expect do
+            post "/admin/config", params: { site_config: { community_emoji: not_an_emoji },
+                                            confirmation: confirmation_message }
+          end.not_to change(SiteConfig, :community_emoji)
         end
 
         it "updates the community_name" do
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -14,6 +14,7 @@ module Admin
     COMMUNITY_PARAMS =
       %i[
         community_name
+        community_emoji
         collective_noun
         collective_noun_disabled
         community_description
--- a/app/controllers/admin/configs_controller.rb
+++ b/app/controllers/admin/configs_controller.rb
@@ -126,6 +126,7 @@ module Admin
         home_feed_minimum_score
       ].freeze
 
+    EMOJI_ONLY_FIELDS = %w[community_emoji].freeze
     IMAGE_FIELDS =
       %w[
         main_social_image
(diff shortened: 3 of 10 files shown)


app/views/chat_channels/index.html.erb
```diff
@@ -1,4 +1,4 @@
-<% title("#{community_name} Connect 👩‍💻💬👨‍💻") %>
```
vaidehijoshi, line 1: This is the hardcoded emoji I removed, as mentioned in the PR description.

pkfrank:
Do we think there are any SEO implications that we should be aware of on DEV?  I know that we intend to reinstate the emojis  on `dev.to` once this is merged in but just want to make sure we've given that a quick thought.

mstruve (approved):
LOVE IT! I think the CodeClimate about instance variables in this case since it is site wide and not an issue with threading is safe to ignore.

vaidehijoshi:
> Do we think there are any SEO implications that we should be aware of on DEV? I know that we intend to reinstate the emojis on `dev.to` once this is merged in but just want to make sure we've given that a quick thought.

I _do_ think it will affect SEO (at least until we add the 2 emojis we have hardcoded back into our own SiteConfig) but I am not sure what our general "best practices" around this have been. I took a look at [a previous PR](https://github.com/forem/forem/pull/6887) where we made these kinds of "generalization" changes and I didn't see any flags that made it evident that we can't change any `<title>` tags...but I agree with you that we should double check that.

I wonder if @benhalpern has insights here about whether it's safe to proceed/what we should do if it _isn't_.

Zhao-Andy (approved):
Nice!! Yay to emojis!

Re: SEO. If we think it's a big enough concern then we could split the SiteConfig addition with the title change addition. Personally think it's more trouble than it's worth, since we would probably have only a minute or two where the emojis were missing from the dev.to `<title>`.

pkfrank:
Can we consider this "do not merge" until @benhalpern weighs in?

It's probably fine, but minimal reason to _not_ treat this with some extra care as these SEO things can be fickle.

vaidehijoshi:
@pkfrank sure thing—have converted this to a draft until we get @benhalpern’s weigh in here 👍🏽

vaidehijoshi:
I slept on this and had some new thoughts on how to proceed here:

Maybe the right thing to do is just to remove the hardcoded emojis completely and not even add a `community_emoji` field at all. If we add it, we're now basically affording a Forem admin a way of changing this emoji an indefinite amount of times, which will impact the `<title>` for that Forem instance. The side effect of that is that we're allowing them to mess around with their SEO, potentially, which we might not want to encourage by adding the emoji as a configurable field to `SiteConfig`.

The other issue here is that I'm not currently validating that the field itself is an emoji -- I could add this, but validating emojis in a regex is not necessarily simple from what I can tell, and I would like to avoid [adding any dependencies](https://github.com/janlelis/unicode-emoji) for this.

These ^^ reasons, alongside the general weariness of [adding yet another value to SiteConfig](https://github.com/forem/forem/issues/11449#issuecomment-728814375) makes me think that this "feature" is actually maybe not worth it. I kind of feel like we should remove the hardcoded emojis from the codebase and call it a day.

Curious what folks + @benhalpern think about this (also curious re: his understanding of SEO consequences)

benhalpern:
I like this. We already don't use emojis in key leaf pages for the SEO paranoia factor, so unless this changes that behavior (and at a glance I don't see that it does) this seems cool.

If we ever move away from having emojis as part of the title, we can ditch this feature and alert admins, but otherwise there might be new places we could include these (including places like forem.com where it might be fun to reference communities by their emojis)

TLDR... Yay! (I also marked the codeclimate as wontfix)

vaidehijoshi:
UPDATE: I've confirmed with @benhalpern that this approach is okay for SEO, which also makes me comfortable with rolling this out to Forem creators/admins as well, since I know it's not going to destroy their SEO 😅 

However, I _have_ added some validations to ensure that non-emoji characters can't be used in this field. Apparently we already had the `emoji_regex` gem as a dependency, so I was able to use it here!! 🙌 

**TO QA MY RECENT UPDATES, PLEASE DO THE FOLLOWING:**
1) Try updating the `community_emoji` field with a non-emoji:
Screen_Shot_2020-11-17_at_11_53_02_AM
...it should not update and you should get an error message:
Screen Shot 2020-11-17 at 11 53 28 AM

2) Try updating the field with any valid emoji -- you shouldn't see any errors, and the field should update:
Screen_Shot_2020-11-17_at_11_54_21_AM

3) Try setting the `community_emoji` to blank -- you shouldn't see any errors, and the field should update:
Screen_Shot_2020-11-17_at_11_55_12_AM

Am re-requesting @mstruve + @Zhao-Andy's reviews here as I've added some changes since their approval(s)

Zhao-Andy (approved):
👍 👍 👍

spec/requests/admin/configs_spec.rb
```diff
@@ -124,6 +124,23 @@
...
+          allow(SiteConfig).to receive(:community_emoji).and_call_original
+          emoji = "🥐"
+          post "/admin/config", params: { site_config: { community_emoji: emoji },
+                                          confirmation: confirmation_message }
+          expect(SiteConfig.community_emoji).to eq(emoji)
+        end
+
+        it "does not update the community_emoji if invalid" do
+          allow(SiteConfig).to receive(:community_emoji).and_call_original
+          not_an_emoji = "i love croissants"
```
mstruve, line 137: +1

mstruve (approved):
TIL alongside is one word. Also, the code looks pretty sound as well 😉  🚀

Issue #1346, opened 2018-12-17
Should be able to update all followed tags with one click


2lach:
**Problem/ feature request:**
On the [Followed tags page](https://dev.to/dashboard/following) the user can only submit a single new value for follow weight per submit.

Current page:
![following page current](https://user-images.githubusercontent.com/11457600/50086369-1ad48600-01fd-11e9-85cc-3088c65127da.jpg)

**Proposed solution**
Since i suppose most users of dev.to (myself included) update several topics at the same time, I suggest a `update all` button. 
This way there is no need for several clicks when updating follow weights = () => which makes happier users and also less data traffic with fewer requests.

Solution illustrated:
![following page proposal](https://user-images.githubusercontent.com/11457600/50086740-3ee49700-01fe-11e9-8f99-3e4336d0109f.jpg)

**Alternatives**
Keep it as is or maybe just save all edited values in localStorage so the input doesn't loose it on Submit. 
Not as awesome as with an update all button. but still the follow weight is a nice feature in an already awesome community. 

Have an awesome day!

Zhao-Andy:
Hey @2lach, thanks for the feature request! This definitely can be improved, and I think we'll want to discuss which option is best to resolve the UX here.

2lach:
Glad to it @Zhao-Andy 👍 

IMHO there are two main ways to go about the UX:

**Option 1**: Would be to do as in the image above  in the feature  request -> `Solution illustrated` which is basically adds a new button and keeps the existing submit buttons.
The new button can be added at the top or bottom of the page. The bottom of the page feels more natural than placing it at the top as in my illustration.

**Options 2**:
Remove all existing submit buttons and only keep the  `Update all` this would give the user less options but also make the UI clearer and cleaner.

Link2Twenty:
While we're looking at this they should be number inputs really.

```html
<input type="number" step="0.1" />
```

**Left field:**
Thought I'd throw in a totally random idea, how about importance sliders, 0 to 10, that update on change?

jessleenyc:
For whoever wants to grab this, I think option 2 makes a lot of sense here.

2lach:
Hey @jessleenyc, hope all is well! 
I totally forgot about this 😋 cool that you brought it up again 👍 I could take some time and fix the UI/UX part.

However, being honest i don't see that i will have the time to get into and fix the backend parts (been a couple of years since i wrote a line of ruby).

So if thats cool i can fix the front-end.

Have a good one 😎

[...]

PR #11454, opened 2020-11-17
Update all followed tag with one click

kuei0221:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Hi, this is my first PR for forem, so let me know if I miss anything 😄
This PR is for https://github.com/forem/forem/issues/1346, adding a new button to update all followed tags.
I follow an idea discussed in the issue:  remove all separated save button, and combine them into one.

By clicking the `Update Weight` button, an array of hash with id and weight point will be pass `[{ id:1, points: 3.0 }, { id: 2, points: 5.0 }]`, then the backend will handle the work. The form will only pass the changed weight point. If nothing changed yet, nothing pass.

Change in this PR:
1. Replace follows#update with follows#bulk_update in both route and controller
2. Replace FollowPolicy#update? with #bulk_update?
3. Combine all save button in following_tags.html.erb into one.
4. Mark changed weight points, and disable unchanged when submitting

https://github.com/forem/forem/issues/1346
截圖 2020-11-26 上午2 17 19

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed


mstruve (approved):
@kuei0221 welcome to the Forem community!!!! Thank you for taking the plunge and contributing to our project. We couldn't do what we do without awesome contributors like yourself. 

I did some manual testing and this works great! Backend looks solid to me, however, the js script code might be better off somewhere else. I tagged a couple of coworkers who have more experience with the frontend than myself to helps us figure that out 😉

app/views/dashboards/following_tags.html.erb
```diff
@@ -54,3 +60,36 @@
     </div>
   </div>
 </main>
+
+<script>
+  document.getElementById('follows_update_form').addEventListener('submit', checkChanged)
```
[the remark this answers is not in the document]
Zhao-Andy: I think this script should live in our Webpack bundle, via `app/javascript/packs`. @kuei0221 I would put the JS code you have here in a regular JS file in that directory, and then you can load it on this `.erb` file like `
<%= javascript_packs_with_chunks_tag "newFileName", defer: true %>`.

Here's a bit more info on that process: https://docs.dev.to/frontend/webpacker/

app/views/dashboards/following_tags.html.erb
```diff
@@ -54,3 +60,36 @@
...
+  };
+
+  function checkChanged(event) {
+    if (document.querySelector('input[changed]')) {
+      disableAllUnchanged();
+    }
+    else {
+      event.preventDefault();
+    }
+  };
```
Zhao-Andy, line 78: **non-blocking, nitpick**

I think this is a neat idea! I personally like having the button be `disabled` too if nothing is changed. Maybe you can add `button.disabled = true` in the `else`?

app/views/dashboards/following_tags.html.erb
```diff
@@ -54,3 +60,36 @@
...
+
+  function disableAllUnchanged() {
+    document.querySelectorAll('div[id^="follows"]').forEach(disableUnchanged)
+  };
+
+  function disableUnchanged(item) {
+    item = item.getElementsByTagName('input')
+    var id = item[0]
+    var point = item[1]
```
Zhao-Andy, line 87: Ideally we use ES6 here and not use `var`. I think re-defining `item` to something more descriptive and not erasing the argument is best practice. :)

Zhao-Andy (commented):
Thanks for this PR! I think it's a great implementation so far :) Left some comments about a few details that I think should be cleaned up. Otherwise, it works well!

app/assets/javascripts/initializers/initScrolling.js
```diff
@@ -121,22 +121,13 @@ function buildTagsHTML(tag) {
     antifollow +
     '</h3>' +
     '<p class="grid-cell__summary truncate-at-3"></p>' +
-    '<form class="edit_follow flex items-center flex-nowrap mb-4" id="edit_follow_' +
+    '<input name="follows[][id]" id="follow_id" type="hidden" form="follows_update_form" value="' +
```
Zhao-Andy, line 124: I think `follows[id]` should work, don't need the extra empty bracket `[]`.

app/assets/javascripts/initializers/initScrolling.js
```diff
@@ -121,22 +121,13 @@ function buildTagsHTML(tag) {
...
     '">' +
-    '<label for="follow_points" class="fs-s flex-1 pr-2 color-base-60 align-right whitespace-nowrap">Follow weight:</label>' +
-    '<input step="any" class="crayons-textfield flex-1 fs-s" required="required" type="number" style="max-width:90px" value="' +
+    '<label for="follow_points" class="fs-s flex-1 pr-2 color-base-60 align-left whitespace-nowrap">Follow weight:</label>' +
+    '<input step="any" class="crayons-textfield flex-1 fs-s" required="required" type="number" form="follows_update_form" value="' +
     tag.points +
-    '" name="follow[points]" id="follow_points">' +
-    '<button type="submit" class="crayons-btn crayons-btn--ghost crayons-btn--s" name="commit">Save</button>' +
-    '</form>' +
+    '" name="follow[][points]" id="follow_points" onchange="addChanged(this)">' +
```
Zhao-Andy, line 130: Also here -- this seems to work as `follow` or `follows` but let's be consistent and keep it as one or the other. :)

app/controllers/follows_controller.rb
```diff
@@ -79,16 +79,22 @@ def create
...
   end
 
-  def update
-    @follow = Follow.find(params[:id])
-    authorize @follow
-    redirect_to "/dashboard/following" if @follow.update(follow_params)
+  def bulk_update
+    @follows = Follow.where(id: params_for_update.keys).includes(:follower, :followable)
+    authorize @follows
+    redirect_to "/dashboard/following" if @follows.all? { |follow| follow.update(params_for_update[follow.id.to_s]) }
```
rhymes, line 85: I wonder what would happen if any of this updates would fail. Should we separate the update step from the `if` and the redirection and do them one at the time?

ps. you can replace `/dashboard/following` with `dashboard_following_path`

app/views/dashboards/following_tags.html.erb
```diff
@@ -19,6 +19,11 @@
             Adjust tag weight to modify your home feed. Higher values mean more appearances for that tag.
             <span class="crayons-indicator crayons-indicator--outlined crayons-indicator--accent ml-3">Default 1.0</span>
         </div>
+
+        <%= form_with url: bulk_update_follows_path, method: :patch, local: true, id: 'follows_update_form' do |f| %>
+          <button type="submit" class="crayons-btn crayons-btn--secondary crayons-btn--m mb-3" name="commit">Update Weight</button>
```
rhymes, line 24: should this be plural? Now that we update multiple weights

rhymes (commented):
Thanks @kuei0221, left a couple of notes. I'm also not convinced about the repeated `Follow weight:` labels, can we remove them? The page already has a help section explaining what the numbers are.

I would also change the "update weight" button to the main button, the blue one so it's more visible and it indicates the primary action

app/views/dashboards/following_tags.html.erb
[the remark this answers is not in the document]
kuei0221: Thanks! I will follow your guide 😄

app/views/dashboards/following_tags.html.erb
kuei0221: Thanks! But if we disable the button in the `else`, because the form submission is stopped by `preventDefault()`, all the buttons will be disabled and the user won't have access to change them until refreshing the page.

app/assets/javascripts/initializers/initScrolling.js
kuei0221: Rack will parse the empty bracket to an array (by [normalize_params](https://github.com/rack/rack/blob/1d70c73043e8fd41bf204ff794730aed2620d7a4/lib/rack/query_parser.rb#L102)), and append the following value to it.
So with `follows[][id]`, the params passed to the server will be `follows: [{id: 1}, {id: ...}, {...}]`
By doing this, the form can pass an array of hashes, which is easier for rails to handle the parameter restriction.

Here is a[ blog](http://codefol.io/posts/How-Does-Rack-Parse-Query-Params-With-parse-nested-query/) that describes this feature. Although this was post several years ago, rack remains the same mechanism.

I have fought on this trick for several hours, hope it makes sense 😅

app/controllers/follows_controller.rb
kuei0221: Oops, thanks for reminding this. I think it would better to add a transaction for this and replace `update` with `update!`, so it will abort the entire update if any update step fails, so the process will be easy to trace. Thanks!

app/javascript/packs/disableUnchangedButton.js
```diff
@@ -0,0 +1,35 @@
+document.getElementById('follows_update_form').addEventListener('submit', checkChanged);
```
rhymes, line 1: could you rename this file in something self explanatory?

I worry that `disableUnchangedButton` doesn't really communicate that it's related to the dashboard tags. Maybe something like `dashboardTagsDisableUnchangedButtons.js` ?

app/javascript/packs/disableUnchangedButton.js
```diff
@@ -0,0 +1,35 @@
...
+function disableUnchanged(item) {
+  let inputs = item.getElementsByTagName('input')
+  let id = inputs[0]
+  let point = inputs[1]
+
+  if (!point.hasAttribute('changed')) {
+    point.setAttribute('disabled', true)
+    id.setAttribute('disabled', true)
+  }
+};
```
rhymes, line 35: missing new line at the end of the file

app/views/dashboards/following_tags.html.erb
```diff
@@ -59,37 +60,4 @@
...
-    var point = item[1]
-
-    if (!point.hasAttribute('changed')) {
-      point.setAttribute('disabled', true)
-      id.setAttribute('disabled', true)
-    }
-  };
-
-</script>
+</main>
```
rhymes, line 63: missing new line at the end of the file

app/controllers/follows_controller.rb
```diff
@@ -79,16 +79,25 @@ def create
...
   end
 
-  def update
-    @follow = Follow.find(params[:id])
-    authorize @follow
-    redirect_to "/dashboard/following" if @follow.update(follow_params)
+  def bulk_update
+    @follows = Follow.where(id: params_for_update.keys).includes(:follower, :followable)
+    authorize @follows
+    Follow.transaction do
```
rhymes, line 85: should we write a test for this? Maybe by submitting a request with non decimal points? Something to make this explode

kuei0221: It is a great idea to add a test here, but I cannot figure out a request to make it explode.
ActiveRecord will change the type of data when saving (here points is float so is `to_f`), so nondecimal points or even negative integers will pass. Besides, the array or hash will not pass the policy before the transaction, and I didn't' find any validation for the weight point.

Do you have any idea? 🙏

rhymes: That was the only one I had in mind. I don't think it's worth it to go down a rabbit hole of deep mocks. Can `follow.id` be invalid ?

kuei0221: I cannot find either. I have tried none exist id or updating the same id twice, but none works. None exist id will be ignored when `Follow.where`, duplicated id will only cut into one when `params_for_update`.

rhymes: Ok, let's leave it like this in the meantime, I don't think it's super critical to have a test for the transaction that we know it's going to be rolled back. If you could address the other notes we can merge this afterwards. Thanks for bouncing ideas!

kuei0221:
I have updated the code according to the new merged PR #11513, renamed the point which can be updated by user to `explicit_points`.
Now the tag insert by `initScrolling.js` will still show calculated points to the user (but update explicit_points), because the inserted tag is generated by another controller action `followings#tags` and it needs an update. I think it a new issue and needs another PR for this.
And maybe it will be nice if we have a test to check what generated by `initScrolling` is equal to the origin. 😉

rhymes:
@benhalpern should this functionality use `explicit_points` directly ?

Zhao-Andy (approved):
Looks great to me! Thanks so much for the PR! Excited for people to use this feature :)

Zhao-Andy:
Sorry, I dismissed my approval because I think we should use `explicit_points` based on the description from #11513. Will wait for @benhalpern to confirm.

app/controllers/follows_controller.rb
```diff
@@ -79,17 +79,25 @@ def create
     render json: { outcome: @result }
   end
 
-  def update
-    @follow = Follow.find(params[:id])
-    authorize @follow
-    @follow.explicit_points = follow_params[:explicit_points]
-    redirect_to dashboard_following_path if @follow.save
+  def bulk_update
```
citizen428, line 82: **suggestion (non-blocking):** This is due to my personal dislike for non-REsTful actions in Rails controllers, but I'd probably do a nested `BulkUpdatesController` with a resourceful route declaration in `routes.rb`, see http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/

nickytonline:
From a UX standpoint, it would probably make sense to make the *Update Weights* button sticky. Otherwise, so save, the user would always have to scroll back to the top. cc: @ludwiczakpawel

ludwiczakpawel:
> From a UX standpoint, it would probably make sense to make the _Update Weights_ button sticky. Otherwise, so save, the user would always have to scroll back to the top. cc: @ludwiczakpawel

Yes, definitely.

In fact, we should probably put that button in the bottom, below the grid. And once any value changes, make it sticky <-- we do exactly that when you try to update something in https://dev.to/settings - the "Save" button sticks to the bottom edge of the screen.

Would be great to ship it that way, but if this is too much of a hassle we can probably do it in a separate PR.

benhalpern:
Sorry for the delay, yes @Zhao-Andy @rhymes we should use `explicit_points` here.

`explicit_points` are the ones the user interacts with directly, `implicit_points` are the ones that the system interacts with to nudge based on user activity, and `points` is the sum of these values which are ultimately used for sorting in general.

rhymes (approved):
Ben confirmed, re-tested this manually, LGTM!

Thanks @kuei0221!

@nickytonline @ludwiczakpawel I think we can work on improving the UX in a separate PR as this has been open for quite a while

nickytonline:
> Ben confirmed, re-tested this manually, LGTM!
> 
> Thanks @kuei0221!
> 
> @nickytonline @ludwiczakpawel I think we can work on improving the UX in a separate PR as this has been open for quite a while

Done. I created a new issue.

PR #11463, opened 2020-11-17
Fixed Preact components not unmounting when InstantClick page lifecycle events are active

### LINKED ISSUE #11371: Jumping over 2 feed items with keyboard navigation
**Describe the bug**

We've recently introduced keyboard navigation for jumping between feed items with `j` and `k`. But in SOME cases it jumps over 2 cards at once. It's pretty hard for me to determine when exactly that happens and what causes it...

**To Reproduce**

As mentioned above: not sure :D In the video below it started happening after returning back to home page from an article view.

**Screenshots**

Video:  https://d.pr/v/RimTip

- OS, version: macOS X 10.15.7
- Browser, version: Chrome 86.0.42x

### LINKED ISSUE #11458: Preact Components not Unmounting due to InstantClick page lifecycle events.
**Describe the bug**

I know how to fix it @reobin. I will create an issue and put a PR up for it. We need to explicitly call `unmountComponentAtNode` in an `InstantClick.on('change', () => {})` event. This is something that normally we wouldn't have to do, but because we use InstantClick, it's necessary. It also explains why I saw components repeated in the Preact dev tools.

I need to do it for pretty much all out components that get mounted via `render`. e.g.

```jsx
import { h, render } from 'preact';
import { unmountComponentAtNode } from 'preact/compat';
import { Search } from '../Search';
import 'focus-visible';

document.addEventListener('DOMContentLoaded', () => {
  const root = document.getElementById('header-search');

  render(<Search />, root);

  InstantClick.on('change', () => {
    unmountComponentAtNode(root);
  });
});
```

**To Reproduce**

You will not see this behaviour as it's something under the hood. It's come to light due to issue #11371

**Expected behavior**

Preact components should be unmounted when an InstantClick even changes the page.

**Screenshots**

- OS, version:
- Browser, version:

- Device:
- OS, version:
- Browser, version:

**Additional context**

nickytonline:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Closes #11458
Closes #11371

1. There are no explicit instructions aside from giving the entire site a test drive to ensure nothing is broken, but definitely test #11371. In theory though, everything should behave as it was before.

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

Estelle from Golden Girls saying "I'm as giddy as a 60 year old


app/javascript/packs/Chat.jsx
```diff
@@ -1,7 +1,7 @@
-
-import { h, render } from 'preact';
+import { h } from 'preact';
 import Chat from '../chat/chat';
 import { Snackbar } from '../Snackbar/Snackbar';
+import { render } from '@utilities/preact';
```
nickytonline, line 4: All the files have this render function imported now.

app/javascript/utilities/preact/index.js
```diff
@@ -0,0 +1,20 @@
...
+ * Renders a Preact component into the given parent DOM element.
+ *
+ * Note: This function is calling Preact's render under the hood along with
+ * a call to InstantClick's change event that explicitly unmounts the component rendered.
+ *
+ * @param {ComponentChild} vnode
+ * @param {Element | Document | ShadowRoot | DocumentFragment} parent
+ * @param {Element | Text} replaceNode
+ */
+export function render(vnode, parent, replaceNode) {
```
nickytonline, line 14: This is the magic sauce.

nickytonline:
I still need to add some tests.

marcysutton:
Seems like it fixed the skipping keyboard navigation issue, as I can't repro that locally with this branch. I did find it stops short of the last post in the list though, like it's off by one. On the live site, the `j` keyboard shortcut stops working after a while too.

reobin:
You are right @marcysutton ! I just checked and it's a regression bug introduced with [this commit](https://github.com/forem/forem/commit/449cf89737).

It basically removed the following code from the article HTML (`app/assets/javascripts/utilities/buildArticleHTML.js`):
```js
    // this variable was removed
    var navigationLink = `
      <a
        href="${article.path}"
        aria-labelledby="article-link-${article.id}"
        class="crayons-story__hidden-navigation-link"
      >
        ${article.title}
      </a>
    `;

    var articleElement =
      '<article class="crayons-story" ' +
      'data-article-path="' +
      article.path +
      'id="article-' +
      article.id +
      '" data-content-user-id="' +
      article.user_id +
      '">' +
      navigationLink + // this line was removed
     ...
```

@nickytonline We could add it back in this PR if you don't mind, if not I'll create a separate PR

Note: Awesome stuff in this PR 🎉

nickytonline:
> You are right @marcysutton ! I just checked and it's a regression bug introduced with [this commit](https://github.com/forem/forem/commit/449cf89737).
> 
> It basically removed the hidden navigation link from the article HTML in `app/assets/javascripts/utilities/buildArticleHTML.js`:
> 
> ```js
>     // this variable was removed
>     var navigationLink = `
>       <a
>         href="${article.path}"
>         aria-labelledby="article-link-${article.id}"
>         class="crayons-story__hidden-navigation-link"
>       >
>         ${article.title}
>       </a>
>     `;
> 
>     var articleElement =
>       '<article class="crayons-story" ' +
>       'data-article-path="' +
>       article.path +
>       'id="article-' +
>       article.id +
>       '" data-content-user-id="' +
>       article.user_id +
>       '">' +
>       navigationLink + // this line was removed
>      ...
> ```
> 
> @nickytonline ~We could add it back in this PR if you don't mind, if not I'll create a separate PR~
> 
> Note: Awesome stuff in this PR 🎉
> 
> EDIT: could not resist doing the PR :)

@reobin, for the regression you mentioned, I would create a separate PR that can be linked to the keyboard issue. This PR is addressing a general issue in the application in regards to components not unmounting due to InstantClick.

app/javascript/Search/Search.jsx
```diff
@@ -109,7 +109,8 @@ export class Search extends Component {
   componentWillUnmount() {
     document.removeEventListener('keydown', this.globalKeysListener);
     window.removeEventListener('popstate', this.syncSearchUrlWithInput);
-    InstantClick.off('change', this.enableSearchPageListener);
+    InstantClick.off &&
```
nickytonline, line 112: This addition existence check is required due to the component being explicitly unmounted outside the normal Preact component lifecycle.

app/javascript/utilities/preact/index.js
```diff
@@ -0,0 +1,23 @@
+import { render as preactRender } from 'preact';
+import { unmountComponentAtNode } from 'preact/compat';
+
+/**
+ * Renders a Preact component into the given parent DOM element.
+ *
+ * Note: This function is calling Preact's render under the hood along with
+ * a call to InstantClick's change event that explicitly unmounts the component rendered.
+ *
+ * @param {object} vnode The virtual DOM node representing a component.
```
nickytonline, line 10: I used the same parameter names as they use in the preact render method.

app/javascript/Search/Search.jsx
citizen428: Would this comment make sense as a source code comment?

app/javascript/utilities/preact/index.js
citizen428: > My only reservation about having this function is it hides what's really going on.

💯 Not a big fan tbh, but I also don't have a better suggestion. Except maybe for making the intention of this a bit more explicit where it's used, at the cost of slightly longer imports:

```js
import { instantClickRender as render } from '@utilities/preact';
```

rhymes:
@nickytonline I noticed one thing. When I test the `J`-`K` navigation keys the first time they work, if I then scroll up, click the logo and try them again, they stop working. I tried logged out, don't know if that's relevant

app/javascript/utilities/preact/index.js
nickytonline: I think explicitly calling it `instantClickRender` makes sense. I'll think of some other possible ways to approach it his as well because of my reservations and your "💯 not a fan" 😎

rhymes (approved):
I haven't completely understood why the remount is needed but I agree renaming the function. Also: the functionality seems to be working!

Good catch @nickytonline

reobin:
Tried it for curiosity, and I think the issue @rhymes mentioned when the user is logged out is independent from #11465

It seems like the shortcuts are unregistered, but not registered again after the user clicks on the logo. Or the `week`, `month`, etc. links.

It does only happen when the user is logged out, so it might be a problem with `app/javascript/packs/homePageFeedShortcuts.jsx`

nickytonline:
> I haven't completely understood why the remount is needed but I agree renaming the function. Also: the functionality seems to be working!
> 
> 
> 
> Good catch @nickytonline 

@rhymes, it's unmount not remount. It's necessary because one it will keep all old components in memory but also, any event handlers attached are never removed from Preact components when InstantClick overwrites the DOM.

app/javascript/Search/Search.jsx
nickytonline: Done.

app/javascript/utilities/preact/index.js
nickytonline: I went with the rename to `instantClickRender`.

nickytonline:
Just need to double check something. In some cases I see the search bar disappearing. Investigating...

citizen428 (approved):
Thanks for renaming this @nickytonline. While it's a bit verbose at least it doesn't obscure the underlying complexity, the previous version made it easy to miss that we're dealing with a non-standard `render` method.

nickytonline:
Thanks for the reviews everyone. I'm digging into this a bit deeper and this PR is a good first attempt, but there are other issues that I see coming up as well in regards to this general fix in this PR. In some cases, the blog post form never leaves memory because it's not properly disposed of when it is closed. With React, it's usually an all in approach, so the root component of your application is typically not destroyed unless you do a full page refresh or you explicitly unmount it for some reason.

[Reconcilliation](https://reactjs.org/docs/reconciliation.html) is usually what handles unmounting of components.

In our case, we sprinkle Preact where it makes sense to provide good user interactions in the application. Because of that, we have several components dispersed through the site that all have their own root components, typically the component itself, e.g. article comment subscription.

In some cases, the fix in this PR addresses that, but in other cases it does not. For example, the search bar at the top is long lived throughout the application, so it makes no sense to destroy it when pages change via InstantClick.

In the case of the new blog/article post form, when we close the form via the X in the top right corner of the application, the form disappears, but the component does not. In a case like this, we need to explicitly unmount the component when that action occurs.

I'm still digging some more, but this PR is providing some interesting discoveries for me as I've typically never had to explicitly destroy Preact/React components.

For the time being, I'm going to close this and keep thinking about this issue. I may reopen this, but I might also open another PR instead depending on how things go.

nickytonline:
Also, based on this comment, https://github.com/preactjs/preact/issues/53#issuecomment-522325210, I'm going to opt to use `render(null, container)` instead of importing `unmountComponentAtNode` from `preact/compat`.

nickytonline:
I looked at `data-instant-track`, that @developit suggested (see https://twitter.com/_developit/status/1329498741318053888), but it appears to cause **full page refreshes** if changes are detected in src/href attributes or innerHTML. We’re on an older version of InstantClick that has been tweaked for our usage, e.g. #5138, and the way it currently operates is to only monitor the `<head />` tag of the document for `data-instant-track`.

Even with the new version though, `var trackedElements = document.querySelectorAll('[data-instant-track]')` (see https://github.com/dieulot/instantclick/blob/aea04389223a8deb44cf7c997a88dcee7bcadd5a/src/instantclick.js#L335) vs.  `var elems = doc.head.children` checking for `data-instant-track` (see https://github.com/forem/forem/blob/master/app/assets/javascripts/base.js.erb#L333), it still causes a full page refresh which is not what we want, so I’m not sure how I can unmount components by leveraging the `data-instant-track` attribute. 

Having said that though, it gives me an idea to maybe use another custom data attribute to keep track of components to unmount, or maybe pop all the components to unmount in a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap). Maybe that's overkill though. And I still need to unmount before the changePage replaces with the new DOM nodes (see https://github.com/forem/forem/blob/master/app/assets/javascripts/base.js.erb#L120). I can definitely hack on the existing code we've already modified and unmount all the components that need to unmount before that happens, but I'm trying to still see if there is a better way to handle this.

Gonna think on this some more still.

citizen428:
Thanks for this writeup @nickytonline, super interesting! ❤️

PR #11464, opened 2020-11-18
Temp Feature: Award Hacktoberfest Badges Rake Task

mstruve:
- [x] Temp Feature

## Description
Award Hacktoberfest badges to subscribed users. This rake task will be run every 30 min so the 2 hour buffer should be more than enough to make sure we dont miss any. No tests bc I am testing in production right now 😅 

https://github.com/forem/InternalProjectPlanning/issues/334

### DIFF
--- a/lib/tasks/temporary/hacktoberfest_badges.rake
+++ b/lib/tasks/temporary/hacktoberfest_badges.rake
@@ -1,0 +1,23 @@
+# TODO: Will be removed my @mstruve once badges are rewarded for 2020
+task award_hacktoberfest_badges: :environment do
+  badge_id = Badge.find_by(slug: "hacktoberfest-2020")&.id
+
+  User.where.not(github_username: nil).includes(:identities).find_each do |user|
+    identity = user.identities.github.first
+    next unless identity
+
+    github_user_id = identity.uid
+    response = HTTParty.get(
+      "https://hacktoberfest.digitalocean.com/api/state/#{github_user_id}",
+      headers: { "Authorization" => ENV["HACKTOBERFEST_AUTH_KEY"] },
+    )
+    if response # contains completed Hacktoberfest indicator then award the badge
+      BadgeAchievement.create(
+        user_id: user.id,
+        badge_id: badge_id,
+        rewarding_context_message_markdown: "Congrats!",
+      )
+      user.save
+    end
+  end
+end
--- a/lib/tasks/temporary/hacktoberfest_badges.rake
+++ b/lib/tasks/temporary/hacktoberfest_badges.rake
@@ -1,8 +1,14 @@
 # TODO: Will be removed my @mstruve once badges are rewarded for 2020
 task award_hacktoberfest_badges: :environment do
   badge_id = Badge.find_by(slug: "hacktoberfest-2020")&.id
+  user_ids = UserSubscription.where(
+    user_subscription_sourceable_id: 529_344,
+    created_at: 3.hours.ago,
+  ).pluck(:subscriber_id)
 
-  User.where.not(github_username: nil).includes(:identities).find_each do |user|
+  User.where(id: user_ids).includes(:identities).find_each do |user|
+    next if u.badges.exists?(id: badge_id)
+
     identity = user.identities.github.first
     next unless identity
 
--- a/lib/tasks/temporary/hacktoberfest_badges.rake
+++ b/lib/tasks/temporary/hacktoberfest_badges.rake
@@ -1,13 +1,14 @@
 # TODO: Will be removed my @mstruve once badges are rewarded for 2020
+# rubocop:disable  Metrics/BlockLength
 task award_hacktoberfest_badges: :environment do
   badge_id = Badge.find_by(slug: "hacktoberfest-2020")&.id
   user_ids = UserSubscription.where(
     user_subscription_sourceable_id: 529_344,
-    created_at: 3.hours.ago,
+    created_at: 2.hours.ago..Time.current,
   ).pluck(:subscriber_id)
 
   User.where(id: user_ids).includes(:identities).find_each do |user|
-    next if u.badges.exists?(id: badge_id)
+    next if user.badges.exists?(id: badge_id)
 
     identity = user.identities.github.first
     next unless identity
@@ -17,7 +18,7 @@ task award_hacktoberfest_badges: :environment do
       "https://hacktoberfest.digitalocean.com/api/state/#{github_user_id}",
       headers: { "Authorization" => ENV["HACKTOBERFEST_AUTH_KEY"] },
     )
-    if response # contains completed Hacktoberfest indicator then award the badge
+    if response.parsed_response == "completed"
       BadgeAchievement.create(
         user_id: user.id,
         badge_id: badge_id,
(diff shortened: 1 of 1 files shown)


lib/tasks/temporary/hacktoberfest_badges.rake
```diff
@@ -0,0 +1,23 @@
...
+  User.where.not(github_username: nil).includes(:identities).find_each do |user|
+    identity = user.identities.github.first
+    next unless identity
+
+    github_user_id = identity.uid
+    response = HTTParty.get(
+      "https://hacktoberfest.digitalocean.com/api/state/#{github_user_id}",
+      headers: { "Authorization" => ENV["HACKTOBERFEST_AUTH_KEY"] },
+    )
+    if response # contains completed Hacktoberfest indicator then award the badge
```
citizen428, line 14: I can't seem to find any information about this API and also don't have a key 😢

lib/tasks/temporary/hacktoberfest_badges.rake
```diff
@@ -0,0 +1,23 @@
+# TODO: Will be removed my @mstruve once badges are rewarded for 2020
+task award_hacktoberfest_badges: :environment do
+  badge_id = Badge.find_by(slug: "hacktoberfest-2020")&.id
+
+  User.where.not(github_username: nil).includes(:identities).find_each do |user|
```
citizen428, line 5: How do you feel about this instead?

```ruby
Identity.github.joins(:user).pluck(:uid, :user_id).find_each do |user_id, github_user_id|
  response = ...
  BadgeAchievement.create(...) # or find_or_create if their could be duplicates
end
```

This way we don't need to instantiate any objects but have the ids for the API call and badge creation. 

The only thing this wouldn't allow us to do is the `user.save` call at the end. I think this was done only to update the `updated_at` timestamp, in which case we could make the following change to `BadgeAchievement`:

```diff
- belongs_to :user
+ belongs_to :user, touch: true
```

https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-touch

BUT: since we don't instantiate full objects anymore, we probably can get ways with materializing a hash in the beginning. Then the following would work and doesn't need any update to `BadgeAchievement`:

```ruby
users = Identity.github.joins(:user).pluck(:uid, :user_id).to_h
users.each do |user_id, github_user_id|
  ...
end
User.where(id: users.keys).update_all(updated_at: Time.current)
```

lib/tasks/temporary/hacktoberfest_badges.rake
rhymes, line 14: This if will always be trueish. 

```ruby
[1] pry(main)> HTTParty.get("https://hacktoberfest.digitalocean.com/api/state/foobar")
=> "not-authorized"
```

it's better to check the return code:

```ruby
[2] pry(main)> response = HTTParty.get("https://hacktoberfest.digitalocean.com/api/state/foobar")
=> "not-authorized"
[3] pry(main)> response.code
=> 401
[4] pry(main)> response.ok?
=> false
```

lib/tasks/temporary/hacktoberfest_badges.rake
mstruve: LOL great minds think alike! I originally did this same thing creating 2 hashes but now I realize the second one is not needed bc the username was used to get the user id from BadgeRewarder. I originally changed it bc I was not sure about why we were saving user. I think you are right that its likely to update `updated_at` but it also might be to bust certain caches. I will give this a test once I get the API information bc it would be much faster. I also tried the hash idea in a console and the console "crashed" on me. 400k key hash I dont think should do that but I guess we can try it in a rake task and see. 

```ruby
# just under 400k of these
identities = Identity.github.joins(:user).pluck(:uid, :user_id).to_h]

users.each do |user_id, username|
...
end
```

lib/tasks/temporary/hacktoberfest_badges.rake
mstruve: We are getting this info direct from Github

mstruve: This is not done, we will have to check the response for an indicator that Hacktoberfest was completed, not sure yet if that will just be a status code or a value in the response hash I need to check

mstruve:
Thank you @rhymes and @citizen428 for the eyes!!!!

lib/tasks/temporary/hacktoberfest_badges.rake
```diff
@@ -0,0 +1,34 @@
...
+
+    identity = user.identities.github.first
+    next unless identity
+
+    github_user_id = identity.uid
+    response = HTTParty.get(
+      "https://hacktoberfest.digitalocean.com/api/state/#{github_user_id}",
+      headers: { "Authorization" => ENV["HACKTOBERFEST_AUTH_KEY"] },
+    )
+    if response.parsed_response == "completed"
```
mstruve, line 21: There are two response options that I have seen in the manual run I am doing, "registered" which we do not award for, and "completed" which we do award the badge for. 

I am running the first 6k manually so I would expect any possible rate-limiting will show up if there is an issue on that front.

lib/tasks/temporary/hacktoberfest_badges.rake
```diff
@@ -0,0 +1,34 @@
+# TODO: Will be removed my @mstruve once badges are rewarded for 2020
+# rubocop:disable  Metrics/BlockLength
+task award_hacktoberfest_badges: :environment do
+  badge_id = Badge.find_by(slug: "hacktoberfest-2020")&.id
+  user_ids = UserSubscription.where(
+    user_subscription_sourceable_id: 529_344,
```
vaidehijoshi, line 6: is this this the id for `https://dev.to/devteam/how-to-claim-your-hacktoberfest-completion-badge-3l2h`?

lib/tasks/temporary/hacktoberfest_badges.rake
```diff
@@ -0,0 +1,34 @@
...
+    github_user_id = identity.uid
+    response = HTTParty.get(
+      "https://hacktoberfest.digitalocean.com/api/state/#{github_user_id}",
+      headers: { "Authorization" => ENV["HACKTOBERFEST_AUTH_KEY"] },
+    )
+    if response.parsed_response == "completed"
+      BadgeAchievement.create(
+        user_id: user.id,
+        badge_id: badge_id,
+        rewarding_context_message_markdown: "Congrats!",
```
vaidehijoshi, line 25: The only question I have about the markdown is whether we need to say something about getting this badge from hacktoberfest? Just thinking about the email I got that told me I got the badge and I wonder if we should be trying to maintain "copy clarity"?

vaidehijoshi (approved):

lib/tasks/temporary/hacktoberfest_badges.rake
mstruve: yes

lib/tasks/temporary/hacktoberfest_badges.rake
mstruve: This is the default message but the email and how it appears in your profile comes out with Hacktoberfest. 
Email:
Screen Shot 2020-12-03 at 3 10 34 PM
Notifications:
Screen Shot 2020-12-03 at 3 11 15 PM

vaidehijoshi: Ah, gotcha 👍

maestromac (approved):
LGTM!

PR #11472, opened 2020-11-18
Add article cachebust backoff update script

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This is a script proposed to be paired with https://github.com/forem/forem/pull/11283

That PR makes pretty significant changes which could cause inconsistencies in navigation if we do not pair it with a cache-busting approach.

We cannot bust all the site at once, but when we go to merge that I think we need some kind of gradual cachebust.

This would create lots of short running jobs that would gradually run gradually more spaced out over time and be done in a few hours.

This script should be abundantly safe with small communities, but we need something or else there could be stale cache issues. With DEV I think it's a script similar to the manual steps we've taken in the past in similar situations.

### DIFF
--- a/lib/data_update_scripts/20201118141822_gradual_article_cache_bust.rb
+++ b/lib/data_update_scripts/20201118141822_gradual_article_cache_bust.rb
@@ -8,13 +8,16 @@ module DataUpdateScripts
       # This uses exponential backoff in order to bust the caches of more recent articles first, but not
       # bust everything in too short a period, so that the edge cache can gradually re-heat without everything
       # going cold at once. Articles with high "hotness score" (e.g. more recent) are the ones most in need of busting.
+
+      # It only busts the "?i=i" variant of the path which is the "internal nav" version, aka the page when swapped
+      # with internal navigation, not the one landed on directly (which wouldn't have cache mismatches)
       Article.published.order("hotness_score DESC").limit(2000).each_with_index do |article, index|
         n = index + 300 # + 300 gives the server time to boot up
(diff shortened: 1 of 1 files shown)


jacobherrington (approved):
The comment gives great context. This makes sense to me :+1:

lib/data_update_scripts/20201118141822_gradual_article_cache_bust.rb
```diff
@@ -0,0 +1,24 @@
...
+      # CSS will not create meaningful visual regressions for most users.
+      # All articles have their caches naturally recycle once per day to finalize the transition.
+
+      # This uses exponential backoff in order to bust the caches of more recent articles first, but not
+      # bust everything in too short a period, so that the edge cache can gradually re-heat without everything
+      # going cold at once. Articles with high "hotness score" (e.g. more recent) are the ones most in need of busting.
+
+      # It only busts the "?i=i" variant of the path which is the "internal nav" version, aka the page when swapped
+      # with internal navigation, not the one landed on directly (which wouldn't have cache mismatches)
+      Article.published.order("hotness_score DESC").limit(1500).each_with_index do |article, index|
```
mstruve, line 14: Could optimize this and the loop below by only selecting the path
```ruby
Article.published.order("hotness_score DESC").select(:path)
```

mstruve (approved):
Seems fine to me, whats the rollout plan? merge immediately after the UI changes?

PR #11475, opened 2020-11-18
Ludwiczakpawel ludwiczakpawel/comments ui update

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We use the parameter `?i=i` to trigger internal navigation which only returns core content and not the head/footer. This is secretly triggered in service worker contexts.

But when we _redirect_ we currently lose the `i=i` param and therefore wind up having two headers in some scenarios.

This PR may not take care of all cases, but it fixes the bug in many cases.

## Added tests?

- [x] Yes
      have not been included_

satellite


benhalpern:
Whoops, included a bunch of stuff from a wrong branch. Closing.

PR #11476, opened 2020-11-18
Preserve internal navigation indicator on redirect

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This fixes the "double header" for stories path redirects after a user updates their username, etc. This is the same issue as here:

https://github.com/forem/forem/issues/11488

But a different variation of the same bug. This is a specific solution for the stories controller issues. I believe it's localized enough to be worth the quick fix. If we see the same `?i=i` problem elsewhere on the site we could reach for a more general solution.

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

window smush


app/controllers/stories_controller.rb
```diff
@@ -369,7 +376,7 @@ def assign_listings
   def redirect_to_lowercase_username
     return unless params[:username] && params[:username]&.match?(/[[:upper:]]/)
 
-    redirect_to "/#{params[:username].downcase}", status: :moved_permanently
+    redirect_to "/#{params[:username].downcase}" + internal_nav_param, status: :moved_permanently
```
vaidehijoshi, line 379: could we extract this into a helper? or come up with a way to make it easier for a developer to know when they should be adding the `internal_nav_param` or not? right now, from looking at this, i have no idea when i should append this to a path and when i don't need to, so i have a feeling that it wouldn't be immediately clear when you'd need to include it in a redirected path or not

vaidehijoshi (commented):
this seems to work for the _one_ flow that i QAed (change user's username, navigate to old users' old username in the route, expect a redirect). is there anything else i should QA for?

left one question about how we can make it clearer/easier for future devs to know when to include the internal param or not!

app/controllers/stories_controller.rb
rhymes: Agreed, maybe the helper could have a comment explaining what `i=i` is and when to use it. Thanks!

ps. `"#{url}?i=i"` is more idiomatic and faster than `"#{url}" + "?i=i"

rhymes: Agreed, maybe the helper could have a comment explaining what `i=i` is and when to use it. Thanks!

ps. `"#{url}?i=i"` is more idiomatic and faster than `"#{url}" + "?i=i"

rhymes: Agreed, maybe the helper could have a comment explaining what `i=i` is and when to use it. Thanks!

ps. `"#{url}?i=i"` is more idiomatic and faster than `"#{url}" + "?i=i"

rhymes, line 379: Agreed, maybe the helper could have a comment explaining what `i=i` is and when to use it. Thanks!

ps. `"#{url}?i=i"` is more idiomatic and faster than `"#{url}" + "?i=i"

rhymes (commented):
+1 on the helper suggestion

benhalpern:
@vaidehijoshi @rhymes I extracted this into something we can re-use more often that encapsulates everything we need from it, and added some documentation to explain more about this overall situation and where URL redirects might originate in the app most relevantly.

rhymes (approved):
Thanks @benhalpern!

LGTM

atsmith813 (approved):

PR #11478, opened 2020-11-18
Fix Facebook Login profile image download failure

Zhao-Andy:
- [x] Bug Fix

## Description
This fixes a bug with Facebook Login where you can't create an account because the profile image download process fails. The workaround is:
1. Grab the user's Facebook UID and OAuth access token
2. Request the photo directly via a separate call
3. From that JSON response, grab the direct link to the requested profile image
4. Pass along to Carrierwave to download as normal

Closes https://github.com/forem/InternalProjectPlanning/issues/238

1. Make a Facebook OAuth app https://docs.forem.com/backend/auth-facebook/
2. Add your Facebook keys to your config
3. Try to login via Facebook
4. See that your user creates successfully

Nope!

## Added tests?
- [x] No, and this is why: it's hard to stub this, and I think this is more of an implementation detail

## Added to documentation?
- [x] No documentation needed

Nope

Alice falls through the rabbit hole.

### DIFF
--- a/app/services/authentication/providers/facebook.rb
+++ b/app/services/authentication/providers/facebook.rb
@@ -47,17 +47,24 @@ module Authentication
         )
       end
 
-      def image_url
-        token = auth_payload.credentials.token
-        uid = auth_payload.uid
-        request = HTTParty.get("https://graph.facebook.com/#{uid}?fields=picture.type(small)&access_token=#{token}")
-        request["picture"]["data"]["url"]
-      end
-
       protected
 
       def cleanup_payload(auth_payload)
         auth_payload
+      end
+
+      private
+
+      def image_url
+        token = auth_payload.credentials.token
+        response = HTTParty.get(
+          "https://graph.facebook.com/#{auth_payload.uid}?fields=picture.type(small)&access_token=#{token}",
+        )
+        if response.ok?
+          response.dig("picture", "data", "url")
+        else
+          Users::ProfileImageGenerator.call
+        end
       end
     end
   end
--- a/app/services/authentication/providers/facebook.rb
+++ b/app/services/authentication/providers/facebook.rb
@@ -9,7 +9,7 @@ module Authentication
         {
           name: @info.name,
           email: @info.email || "",
-          remote_profile_image_url: @info.image,
+          remote_profile_image_url: image_url,
           facebook_username: user_nickname,
           facebook_created_at: Time.zone.now
         }
@@ -45,6 +45,13 @@ module Authentication
           provider_name,
           **kwargs,
         )
+      end
+
+      def image_url
+        token = auth_payload.credentials.token
+        uid = auth_payload.uid
+        request = HTTParty.get("https://graph.facebook.com/#{uid}?fields=picture.type(small)&access_token=#{token}")
+        request["picture"]["data"]["url"]
       end
 
       protected
--- a/app/services/authentication/providers/facebook.rb
+++ b/app/services/authentication/providers/facebook.rb
@@ -4,6 +4,7 @@ module Authentication
     class Facebook < Provider
       OFFICIAL_NAME = "Facebook".freeze
       SETTINGS_URL = "https://www.facebook.com/settings?tab=applications".freeze
+      PICTURE_URL = "https://graph.facebook.com/%<uid>s?fields=picture.type(small)&access_token=%<token>s".freeze
 
       def new_user_data
         {
@@ -56,10 +57,8 @@ module Authentication
       private
 
       def image_url
-        token = auth_payload.credentials.token
-        response = HTTParty.get(
-          "https://graph.facebook.com/#{auth_payload.uid}?fields=picture.type(small)&access_token=#{token}",
-        )
+        url = format(PICTURE_URL, uid: auth_payload.uid, token: auth_payload.credentials.token)
+        response = HTTParty.get(url)
         if response.ok?
           response.dig("picture", "data", "url")
         else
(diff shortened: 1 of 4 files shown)


app/services/authentication/providers/facebook.rb
```diff
@@ -47,6 +47,13 @@ def self.sign_in_path(**kwargs)
         )
       end
 
+      def image_url
```
rhymes, line 50: this should likely be moved in a `private` section at the bottom

app/services/authentication/providers/facebook.rb
```diff
@@ -47,6 +47,13 @@ def self.sign_in_path(**kwargs)
...
+        token = auth_payload.credentials.token
+        uid = auth_payload.uid
+        request = HTTParty.get("https://graph.facebook.com/#{uid}?fields=picture.type(small)&access_token=#{token}")
+        request["picture"]["data"]["url"]
```
rhymes, line 54: is there any chance this could fail? What if for whatever reason the response is not `response.ok?` - should we have a fallback to a default image? 

Maybe we can do something like: 

```ruby
if response.ok?
  response.dig("picture", "data", "url")
else
  Users::ProfileImageGenerator.call
end
```

this way, worst case scenario, we use a fallback profile image

rhymes (commented):
Cool!

app/services/authentication/providers/facebook.rb
```diff
@@ -47,6 +47,13 @@ def self.sign_in_path(**kwargs)
         )
       end
 
+      def image_url
+        token = auth_payload.credentials.token
+        uid = auth_payload.uid
+        request = HTTParty.get("https://graph.facebook.com/#{uid}?fields=picture.type(small)&access_token=#{token}")
```
Zhao-Andy, line 53: These variables are made because Rubocop said my line was too long :(

rhymes: Ah ah :)

the URL could be moved in another variable, or better yet in a constant template string that gets interpolated by `format/sprintf` https://github.com/rubocop-hq/ruby-style-guide#sprintf

We use it in a few places:

https://github.com/forem/forem/blob/89e0ea79bb0cb6b7cd59b85770f14e70bcb75ddd/app/models/article.rb#L368

https://github.com/forem/forem/blob/89e0ea79bb0cb6b7cd59b85770f14e70bcb75ddd/app/services/slack/messengers/note.rb#L29-L35

Zhao-Andy: Woahhhhh neat! Was wondering if constants with interpolation was a thing!

Zhao-Andy:
Tests are failing, will fix

joshpuetz (approved):

spec/services/authentication/authenticator_spec.rb
```diff
@@ -221,6 +221,8 @@
     # Freeze time since `facebook_created_at` will be based on server time
     before do
       Timecop.freeze
+      allow_any_instance_of(Authentication::Providers::Facebook)
+          .to receive(:image_url).and_return("https://dummyimage.com/400x400.jpg")
```
Zhao-Andy, line 225: So I force pushed this even though Rubocop didn't like it. I'm not sure how to avoid using `allow_any_instance_of` here though. Would I do something like:
```ruby
let!(:facebook_provider) { instance_double(Authentication::Providers::Facebook) }
```

at the top? I think this doesn't work though because each authentication request has a new instance. 🤔

rhymes: What if instead of stubbing `image_url` you stub `HTTParty` and let it return what you need to make the test pass?

joshpuetz: Yeah, that's why `instance_double` doesn't work: each auth request gets a new instance.

I chose to use `allow_any_instance_of` because I was not interested in mocking the HTTP result. If you do go that route: consider using VCR to get an accurate representation of what Facebook is returning.

joshpuetz:
Did you ever ascertain why this suddenly started failing @Zhao-Andy ?

Zhao-Andy:
@joshpuetz I'm not sure still. I only get a 500 error even going through the response, and there's no helpful error message. My best guess is that Facebook changed some permissions. This StackOverflow post helped me, but this is from 8 years ago so it doesn't explain the 500 error. 🤔 

https://stackoverflow.com/questions/11748941/facebook-oauth-no-picture-with-basic-permissions

Zhao-Andy:
@joshpuetz................ I went back to check after seeing your comment and figured it out.................

THE CALLBACK URL IS `HTTP` AND NOT `HTTPS` WHICH IS WHY IT'S BEEN FAILING 😡😡😡😭😭😭

I don't know if Facebook updated their `auth_payload.image` URL manually or something but UGHHHHH

app/services/authentication/providers/facebook.rb
```diff
@@ -9,7 +9,7 @@ def new_user_data
         {
           name: @info.name,
           email: @info.email || "",
-          remote_profile_image_url: @info.image,
+          remote_profile_image_url: @info.image.gsub("http", "https"),
```
fdoxyz, line 12: One question: If Facebook decides to revert back to `https` this would break again, right? Because it would turn `https://fb.com` into `httpss://fb.com`.

Maybe changing this to `@info.image.gsub("http://", "https://")` would still fix the problem and if they switch it back to `https` then it wouldn't find anything to replace and it would stay the same. Does this make sense?

Zhao-Andy: Ahh yes, great point! Will update.

joshpuetz:
> @joshpuetz................ I went back to check after seeing your comment and figured it out.................
> 
> THE CALLBACK URL IS `HTTP` AND NOT `HTTPS` WHICH IS WHY IT'S BEEN FAILING 😡😡😡😭😭😭
> 
> I don't know if Facebook updated their `auth_payload.image` URL manually or something but UGHHHHH

They....WHAT?!!!?

Regardless: great catch!

fdoxyz (approved):
> Regardless: great catch!

x100 ☝️

Issue #354, opened 2018-08-14
@ mention autocompletes in comment box Frontend


benhalpern:
## Feature Request or Task
As a user I want to be able to start typing with @ and have it deliver me a dropdown. This is expected behavior in this context.

This means that the box would have to become a content-editable div and the code should attach to each box. Making that part work could involve some yak shaving.

This should be written in Preact, and if think it makes sense to include a different library, let's discuss. We are critical of each dependency as we want to be efficient in this regard. It doesn't mean you definitely can't use a library, but let's just discuss the implications if you go that route.

The dropdown will use Elasticsearch, should search our user index

### Definition of Done
This is done when all content boxes have a dropdown menu and the behavior works as expected on sites like GitHub and Twitter. Core team is here to help navigate the nuances of the issue.

theoutlander:
I was about to make a request for autocompleting liquid tags and came across this issue. I can take a stab at it (need to get set up first). Following that, if there's consensus, I can look into autocompleting liquid tags.

theoutlander:
Got set up after 3 hours of fighting everything. Need to get acquainted with a few things. I'm going to take a stab at a simpler issue (https://github.com/thepracticaldev/dev.to/issues/245) before I jump on this one.

theoutlander:
@benhalpern is there an endpoint to retrieve user id's or would I need to create that as well?

Assuming I have to create it, there's the endpoint to serve aliases dynamically. I would end up caching the id's on the server-side since this will be a frequently used feature. This cache will also need to be updated as new users are added to the system (or removed). Ideally, we can do this as a micro-service unless you prefer that we stick to the current backend. Otherwise, there's a risk that we overload the backend. If we opt for a micro-service, would NodeJS / Serverless be an option?

As for the front-end, it will listen to a trigger event like '@' and query that endpoint to make automatic suggestions for aliases.

theoutlander:
@benhalpern @pkfrank @jessleenyc @Zhao-Andy @maestromac I am ready to jump on this, but I need feature clarification as detailed in my previous comment.

Also, issues that are assigned should be assigned. This is a big problem at the moment finding issues to work on.

pkfrank:
Hey Nick, thanks for the feedback and nudge.  I think that @benhalpern will clarify the endpoint question tomorrow, and we'll all continue to improve how we handle assignments to stay organized.  We appreciate your help and contributions, and will be in touch soon!

jessleenyc:
[...]

PR #11479, opened 2020-11-18
Feature:Backend Username Search Endpoint for Mention Completion

mstruve:
- [x] Feature

## Description
For a while many have been asking for an autocomplete functionality for mentions. This provides the backend endpoint to get the job done. When given a username or part of a username it will pass that to Elasticsearch, run a search, and then return all matching usernames in an array.

https://github.com/forem/forem/issues/354

## Added tests?
- [x] Yes 💪 

alt_text

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,5 +1,5 @@
 class SearchController < ApplicationController
-  before_action :authenticate_user!, only: %i[tags chat_channels reactions]
+  before_action :authenticate_user!, only: %i[tags chat_channels reactions usernames]
   before_action :format_integer_params
   before_action :sanitize_params, only: %i[listings reactions feed_content]
 
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -82,6 +82,14 @@ class SearchController < ApplicationController
     render json: { result: user_search }
   end
 
+  def usernames
+    usernames = Search::User.search_usernames(params[:username])
+
+    render json: { result: usernames }
+  rescue Search::Errors::Transport::BadRequest
+    render json: { result: [] }
+  end
+
   def feed_content
     feed_docs = if params[:class_name].blank?
                   # If we are in the main feed and not filtering by type return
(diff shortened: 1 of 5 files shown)


app/controllers/search_controller.rb
```diff
@@ -82,6 +82,14 @@ def users
     render json: { result: user_search }
   end
 
+  def usernames
+    usernames = Search::User.search_usernames(params[:username])
+
+    render json: { result: usernames }
+  rescue Search::Errors::Transport::BadRequest
```
vaidehijoshi, line 89: Out of curiosity, what happens if someone hits this endpoint a ton of times? I'm thinking of the frontend implementation of this and how it will need to throttle requests to this endpoint so that we're not searching the entire DB of usernames more times than necessary, but just wondering what kind of error would be raised if that were to happen? 

(I am an ES newbie 😄)

mstruve: That is a great question! Ideally, we don't want someone to spam this endpoint bc it would slow down ES which is why we have throttling on it!
```ruby
module Rack
  class Attack
    throttle("search_throttle", limit: 5, period: 1) do |request|
      if request.path.starts_with?("/search/")
        track_and_return_ip(request.env["HTTP_FASTLY_CLIENT_IP"])
      end
    end
```

atsmith813 (approved):
Good thing we've safe listed all of `/search` so we don't need to update the VCL for `username` 🙏  😅

app/controllers/search_controller.rb
vaidehijoshi: Ah, good to know! I didn't realize we had already accounted for this in `Rack::Attack`, but makes total sense 👍

vaidehijoshi (approved):

app/controllers/search_controller.rb
mstruve: Your question made me check the permissions on this endpoint as well. Given you cannot mention someone without being a user I put it behind our authentication which further decreases the possibility of abuse. This is why questions are so valuable in PRs!!!!!!

vaidehijoshi: AH! I should have remember to check for that myself during review! That seems good + important to do from a security perspective, too.

PR #11485, opened 2020-11-18
Render Profile Fields on user profiles

jacobherrington:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This change allows Forems who have enabled the `profile_admin` feature to create and modify the fields that users display on their profiles.

These changes should be safe to deploy to DEV because we don't have that feature enabled yet, however, we will probably need to make sure our ProfileFields models are configured correctly before making the swap to ensure no data is lost.

One particular place this change ignores is the display of social links on profiles, that's something we haven't totally solved yet and should probably be defined a little better. I think it can be tackled in another PR.

_Large screenshots, so I put them in spoilers!_

**Not DEV**
<details>
  <summary>When feature is enabled:</summary>
  
  image
  
</details>
<details>
  <summary>When feature is disabled:</summary>
  
  image

</details>

**DEV**
<details>
  <summary>When feature is enabled:</summary>

  image
  
</details>
<details>
  <summary>When feature is disabled:</summary>
  
  image

</details>

There shouldn't have been any significant changes from what we have currently!

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

No, but there will need to be some due diligence done before we enable the feature flag.


spec/system/user/user_edits_profile_spec.rb
```diff
@@ -2,6 +2,10 @@
 
 RSpec.describe "User edits their profile", type: :system do
   let(:user) { create(:user) }
+  let!(:profile_field_group) { create(:profile_field_group, name: "Ice Cream") }
+  let!(:profile_field) do
```
citizen428, line 6: When not creating fields through the service objects, you should call `Profile.refresh_attributes!` manually, maybe in the `before` block.

citizen428:
I left some early comments @jacobherrington. Let me also answer the questions you sent me on Slack here, so other can follow along.

> Header fields are the place I'm stuck, we have no idea what they might look like, so how do we design a UI for them?

I don't know either. IIRC I raised this question in the Slack channel for profile generalization once, but there was no answer. Or maybe I'm just imagining that, I'm not sure anymore. I believe this is something @ludwiczakpawel would be more qualified to help with than me. I see basically two options:

1. We allow admins to upload an optional icon for header fields (this could get messy).
2. We try to autodetect profile links like this, maybe even for more sites than the currently supported ones. I think this we also have a user submitted issue for this proposal.

> Should some ProfileFields be protected and unable to be deleted? For example: Name

I've been going back and forth on just keeping name on `User`. Mostly because I think all profile fields should be deletable, though we need sure admins **really** understand what they are doing when removing one.

> Should some ProfileFieldGroups be protected because they will fit a certain design, like "Links" which are fairly unique?

See above. Other communities may not be as link-heavy as DEV.

@jacobherrington Since we're here already: each user can also have up to 5 custom profile fields (`app/models/custom_profile_field.rb`), these also need to go somewhere on the profile.

ludwiczakpawel:
@citizen428 @jacobherrington I'm not sure if I understood Jacob's question correctly but here's like a short overview how this was supposed to work/look like: https://d.pr/v/YXMhZa

TL;DR for header fields:
- this is where and how we wanna display these fields
    Zrzut ekranu 2020-11-19 o 11 26 48
- we should add some constraint like "admin can create up to _n_ header fields" (_n_ TBD)
- another sensible constraint would be "if admin creates a header field, it can only be short form text (`input[text]`)" 
    - in future we could extend that constraint for maybe a `select` type of field, or some `radio` options or `checkbox`es..

---

> We allow admins to upload an optional icon for header fields (this could get messy).
> We try to autodetect profile links like this, maybe even for more sites than the currently supported ones. I think this we also have a user submitted issue for this proposal.

@citizen428, I think we're talking about two separate things here, no? You're referencing [Links section](https://user-images.githubusercontent.com/108287/99654987-6d5af480-2a5b-11eb-97b3-9504c907db71.png), right? And I think this was more about other fields/data we may want to display in header.  Correct me if I'm wrong please.

---

One thing I'd love your thoughts on is: how do we replicate what is currently being known as "Work" field on DEV - it's not one field, it's actually [3 fields combined](https://github.com/forem/forem/blob/master/app/views/users/_metadata.html.erb#L14-L22):
- `employment_title`
- `employer_name`
- `employer_url`

app/models/profile.rb
```diff
@@ -60,4 +60,9 @@ def custom_profile_attributes
   def clear!
     update(data: {})
   end
+
+  def ui_attributes_for(area)
+    names = ProfileField.where(display_area: area).pluck(:attribute_name)
+    attributes["data"].select { |k, v| v.present? && names.include?(k) }
```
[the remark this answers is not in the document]
jacobherrington: I initially wanted to use compact, but I do need to drop the blank values; the intention here is to provide a list of profile fields that **will** be rendered, we shouldn't include the ones that aren't set in this group.

Using slice is definitely a good call, I pushed some changes based on this feedback and moved the method to a decorator. Thanks!

jacobherrington:
@citizen428 Thanks for following up - those answers clarify some things for me. It sounds like our thoughts are somewhat aligned.

@ludwiczakpawel Your video was extremely helpful. I pushed some changes to start using profile fields in the header, but that is where things get dicey. Some of the things we do in the design on DEV break a bit (for example, the "work" column can't be generated dynamically to say "<title> at <employer>" because we can't assume either a job title and employer field exist at all)

image

Similarly, we can't assume that a "looking for work" field exists, so rendering that based on the fact that it is a boolean is a little complicated, potentially we could just render and checkmark icon or something for booleans, but I'm not sure there.

jacobherrington:
**Regarding Links:**

As it stands now, these are "ProfileFields" meaning from the database's perspective these are no different than any other field. At the moment that means they are handled similarly. We could always add an additional `display_area` for links.

Right now, if links are set up like this (as fields in the "header" display area):

image

When I fill them out it will look like this:

image

It seems like either we need to redefine some of the constraints around "ProfileFields" like @ludwiczakpawel has mentioned. Or we need to delineate between things that ARE ProfileFields and things that AREN'T (links, name, etc.)

citizen428:
> One thing I'd love your thoughts on is: how do we replicate what is currently being known as "Work" field on DEV - because it's not one field, it's actually 3 fields combined:

@ludwiczakpawel I have honestly no idea. This seems relatively DEV specific and @benhalpern at one point indicated a willingness to drop things if they are mostly presentational and make generalization harder.

> I think we're talking about two separate things here, no?

@ludwiczakpawel  Yes, probably. I'm at a point where I worked on the backend of this for so long that everything has become a blur.

> Even though this doc is pretty old, I thought we had an agreement that not all fields should be delectable.

@ludwiczakpawel I agree in principle, though I don't seem to find this explicitly in the document (`required` was something else). But maybe my skim reading was sloppy. 

We can definitely make certain fields mandatory and stop admins from deleting/modifying them, which should be relatively easy (famous last words). I'll do this in a follow-up PR as the admin UI is currently disabled anyway.

benhalpern:
I think this is mostly pretty close, and if we have a few less-than-ideal regressions, it will still be nice to make progress and then we can improve over time.

I think colors need to be dropped from the customization (or at least treated as special for now)

Screen Shot 2020-11-25 at 1 12 20 PM

Same with anything we're also displaying in the header.

Screen Shot 2020-11-25 at 1 19 39 PM

IMO we come up with what we can ship, but keep some things hardcoded as exceptions.

Then we can circle back on those exceptions once other stuff has shipped.

If we have some progress made and can focus on certain problems or exceptions one thing at a time I think we'll figure this out vs trying to handle all the questions and exceptions at once before shipping.

ludwiczakpawel:
@jacobherrington let me know where we are with everything and how can I help you / assist with design or frontend work.. since it's been a while and I'm kinda out of the loop right now :D

lib/data/dev_profile_fields.csv
```diff
@@ -1,18 +1,18 @@
 Display email on profile,check_box,,,Basic,settings_only
-Website URL,text_field,https://yoursite.com,,Basic,header
-Summary,text_area,A short bio...,,Basic,header
+Website URL,text_field,https://yoursite.com,,Basic,settings_only
```
jacobherrington, line 2: In this file, I've moved all of the fields that shouldn't be handled by the code rendering "header" and "left_sidebar" fields into the "settings_only" group. I think some of this should be revisited and we might want to add a group such as "links" and another called "protected."

Protected fields should not be deletable or heavily modifiable - in fact, we might not want to render them in the admin UI at all.

app/views/users/_sidebar.html.erb
```diff
@@ -32,52 +32,68 @@
   <% end %>
 
   <% cache "user-profile-sidebar-profile-details-#{@user.id}-#{@user.profile_updated_at}", expires_in: 10.days do %>
-    <% if @user.mostly_work_with.present? %>
-      <div class="crayons-card crayons-card--secondary">
-        <header class="crayons-card__header">
-          <h3 class="crayons-subtitle-3">Skills/languages</h3>
-        </header>
+    <% if FeatureFlag.enabled?(:profile_admin) %>
```
jacobherrington, line 35: This should be safe to deploy to DEV because all of the UI changes are behind feature flags.

app/views/users/_metadata.html.erb
```diff
@@ -1,34 +1,78 @@
...
+          <strong class="crayons-definition__title">
+            <%= sanitized_sidebar title.titleize %>
+          </strong>
+          <p class="crayons-definition__value">
+            <%= sanitized_sidebar value %>
+          </p>
+        </div>
+      <% end %>
+      <%# As we migrate to more generalized profiles, we don't want to break DEV %>
+      <% if SiteConfig.dev_to? %>
```
jacobherrington, line 15: In the case that we are on DEV, we want to preserve the employer and looking for work fields; they need to be hardcoded at the moment.

app/views/users/_metadata.html.erb
```diff
@@ -1,34 +1,78 @@
-<% if @user.employment_title.present? || (@user.looking_for_work_publicly == true && @user.looking_for_work == true) || @user.education.present? %>
-  <div class="profile-header__bottom fs-base">
-    <% if @user.education.present? %>
-      <div class="crayons-definition">
-        <strong class="crayons-definition__title">Education</strong>
-        <p class="crayons-definition__value"><%= @user.education %></p>
-      </div>
-    <% end %>
+<% if FeatureFlag.enabled?(:profile_admin) %>
```
jacobherrington, line 1: This should be safe to deploy to DEV because all of the UI changes are behind feature flags.

app/models/profile.rb
```diff
@@ -16,6 +16,7 @@ class Profile < ApplicationRecord
     brand_color2: :text_color_hex,
     display_email_on_profile: :email_public,
     display_looking_for_work_on_profile: :looking_for_work_publicly,
+    education: :education,
```
jacobherrington, line 19: This allows me to remove education from the hardcoded set of fields for DEV.

app/decorators/profile_decorator.rb
```diff
@@ -0,0 +1,14 @@
+class ProfileDecorator < ApplicationDecorator
+  DEV_HEADER_FIELDS = %w[employment_title employer_name].freeze
+
+  def ui_attributes_for(area:)
+    # Return a Hash of the profile fields that should be rendered for a given
+    # display area, e.g. :left_sidebar
+    names = ProfileField.public_send(area).pluck(:attribute_name)
+    # Temporary workaround: DEV specific header fields are hardcoded in the view
+    if SiteConfig.dev_to?
+      names -= DEV_HEADER_FIELDS
```
jacobherrington, line 10: Eventually, if we build the features to allow DEV's multi-field ProfileField (employer) we can remove this line.

jacobherrington:
@ludwiczakpawel I've updated the screenshots in the initial PR, I think this reasonable, but we might want to discuss how links will work next (I think that will be a separate PR though).

app/decorators/profile_decorator.rb
citizen428: There are currently no plans for multi-field profile fields and unless we have another potential use case for them outside this I'd be very reluctant to change that, as we are bound to get the abstraction wrong.

lib/data/dev_profile_fields.csv
citizen428: > In this file, I've moved all of the fields that shouldn't be handled by the code rendering "header" and "left_sidebar" fields into the "settings_only" group.

We should consider writing a data update script adding a "Links" group and moving links there, then update this accordingly. I think it's a small enough change to be part of this PR, but feel free to do it in a follow-up if you prefer that.

> Protected fields should not be deletable or heavily modifiable

Now that name moved back to `User`, what's an example of a protected field? If it can't be deleted or modified, maybe it shouldn't be a profile field at all.

app/decorators/profile_decorator.rb
```diff
@@ -0,0 +1,14 @@
+class ProfileDecorator < ApplicationDecorator
+  DEV_HEADER_FIELDS = %w[employment_title employer_name].freeze
+
+  def ui_attributes_for(area:)
+    # Return a Hash of the profile fields that should be rendered for a given
```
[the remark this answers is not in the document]
jacobherrington: Good call!

lib/data/dev_profile_fields.csv
jacobherrington: @citizen428 I think you are correct in that some things shouldn't be profile fields at all. For context, I was thinking of `summary`specifically when I wrote this thought down.

app/decorators/profile_decorator.rb
jacobherrington: That's a fair point, maybe the goal should be to inch DEV closer to the vanilla Forem setup than to build a feature to keep this in DEV. Does the way I've written this make sense for now?

benhalpern (approved):
This looks good to me. I went ahead and accepted the changes from the conflict created by @Ridhwana's other PR....

I think this PR is good under the idea that it gets more functionality into the app behind some `if/else` statements, but we'll need to follow up on the next steps.

Closing the loop here is going to be a delicate process. We're going to have to communicate well.

app/decorators/profile_decorator.rb
citizen428: > That's a fair point, maybe the goal should be to inch DEV closer to the vanilla Forem setup than to build a feature to keep this in DEV.

This is generally the direction we're taking right now. 😊

> Does the way I've written this make sense for **now?**

Absolutely 👍

ludwiczakpawel:
> we might want to discuss how links will work next (I think that will be a separate PR though).

I'm happy to. I have a doc (not RFC though 🤷‍♂️ ) specifying behavior I was hoping for.

Ridhwana (approved):
I played around with this PR and it seems good :)

citizen428:
> Closing the loop here is going to be a delicate process. We're going to have to communicate well.

@benhalpern Feel free to use the #profile-generalization channel in Slack to start a discussion regarding the finer points of this, I for one would be happy to hear your thought on this topic.

PR #11495, opened 2020-11-19
Remove under-used admin/growth tab

benhalpern:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This interface has a rich history in our development lifecycle as being associated with the birth of the "single resource admin", but the current interface is specific to an initiative with DEV we don't even use anymore.

"Growth"-related needs can be served by Blazer, etc. better than this. If we ever wanted to circle back on a "growth" view we could, but the current version is not serving that well. I think it's best to remove it for the timebeing.


rhymes (approved):
sayonara 🪓

jacobherrington (approved):
✂️

Issue #11489, opened 2020-11-19
After the comment menu is triggered, the main top-right menu stops working.


benhalpern:
**Describe the bug**

Very odd, but if you trigger the menu dropdown to work, the main menu will all of a sudden _not_ work.
**To Reproduce**

- Go to page with comments
- Open a three-dots menu
- Go to top-right of page and hover... The menu won't work until you refresh.
**Expected behavior**

rhymes:
This has been closed by #11498

PR #11498, opened 2020-11-19
Fix menu dropdown disable issue

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Since the behavior of comment dropdowns strips the "block" class from everything when it is dismissed, this area got messed up because it operates with different functionality. Switching to "inline-block" seems to not affect the style but ensures this problem doesn't happen.

This is still a little fragile but I think the fix is helpful for now.

https://github.com/forem/forem/issues/11489

Follow the instructions in the issue and things should be fixed now.


rhymes (approved):
Tested it manually 🆗

maestromac (approved):
:+1:

PR #11502, opened 2020-11-19
Replace podcast bar helpers with image_tag

maestromac:
- [x] Bug Fix for Forem

## Description
We were generating assets path ourselves instead of using `image_tag` with the asset that's already in our asset folder. This is broken in Forem because we don't use Cloudinary there.

n/a

0. Pull down this PR and boot up the app locally.
1. Make sure you have podcasts locally. You need to have developing caching enabled and running sidekiq. 
2. Navigate to any podcast episodes and press play.
3. You should be able to see the volume and volume mute icon.

image
image

n/a

## Added tests?
- [x] No, and this is why: I would normally include a view spec but this partials' render is very straight forward.

## Added to documentation?
- [x] No documentation needed

n/a

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -62,22 +62,6 @@ module ApplicationHelper
     content_for ? title(title_text) : title_text
   end
 
-  def icon(name, pixels = "20")
-    image_tag(icon_url(name), alt: name, class: "icon-img", height: pixels, width: pixels)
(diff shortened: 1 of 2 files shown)


fdoxyz (approved):
Nice fix 👍

benhalpern (approved):

rhymes (approved):

Issue #1154, opened 2018-11-19
A11y: Exit focused navigation with escape for keyboard users


RyKilleen:
**Is your feature request related to a problem? Please describe.**
When a user's context is focused in navigation, modal, or any other pop up menu, it's useful to refocus outside of that context on an Escape key press.

**Additional context**

Link2Twenty:
Here are the aria-practices pages for modal, menu and menu buttons.
https://www.w3.org/TR/wai-aria-practices/#dialog_modal
https://www.w3.org/TR/wai-aria-practices/#menu
https://www.w3.org/TR/wai-aria-practices/#menubutton

jessleenyc:
Thanks for flagging! This issue is approved for anyone who might want to work on it.

Zhao-Andy:
Not stale, still valid.

maestromac:
Sorry again, not stale.

Link2Twenty:
Not stale

RyKilleen:
I'll attempt to tackle this over break, before the new year!

nickytonline:
@RyKilleen thanks so much for the issue. Since Hacktoberfest starts tomorrow, if you're up for taking this on just say the word.

RyKilleen:
I'm spread thin across a few things atm and don't want to monopolize it!

If anyone else would like to give it a go, I'll defer to them. I'll also send it around

reobin:
#10468 adds a hook for easy global key events. If it is to be merged, I'd be up to use it on this issue

marcysutton:
I can take this one on, since I was about to file an issue for it!

nickytonline:
Sounds great @marcysutton. Assigning this one to you.

cmgorton:
@forem/frontend Another one we may want to consider for internal review. Let me know what you think with this issue? It may be close to finished since I know Marcy was working on it before and has a PR for it that needs some changes.

nickytonline:
I think this is a great candidate @cmgorton

aitchiss:
This ties in to #12297 (Optimise all uses of crayons dropdown for accessibility) as well as the open PR from Marcy - #11509

mstruve:
@aitchiss will this be solved by the Marcy PR you mentioned or will their still be more work to do beyond that?

aitchiss:
[...]

PR #11509, opened 2020-11-19
Header navigation dropdown accessibility

marcysutton:

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [X] Optimization
- [ ] Documentation Update

## Description

To make the header menu more ergonomic and accessible, this PR includes some updates to the markup and JavaScript logic. Including:

- Wrapping the logged-in header menu in an unordered list so the items are counted in a screen reader
- Adding `aria-expanded` to the navigation button so it indicates its state programmatically
- Opening the dropdown when you activate the button with a keyboard "click", as opposed to opening it on focus
- Sending focus to the first item when shown
- Wiring up the Escape key to close and sending focus back to the menu button

Note: the profile focus state could still use a little work, as it's not respecting border radius.

Because the navigation button also opens a logged-out overlay, I specifically did not add `aria-haspopup`, which is meant to indicate a sub menu. I considered adjusting the markup of that overlay to be an unordered list / menu-type-thing too, but it was going to be difficult with all of the wrapping DIVs and non-interactive items like heading and paragraph text. So I opted to leave that part alone.

A follow-up task will be to ensure this logic is repeated on all dropdown instances, starting with Storybook.

Related to https://github.com/forem/forem/issues/1154

Navigate the header with the keyboard, using the Enter or Space key to open it and Escape to close (or tabbing past the last item, which I kept). Notice how you aren't forced through the open menu item anymore when you tab through the page.

Note: because the items are wrapped in list items now, I moved the logic to show the Forem Admin item to the LI instead of the A. This also eliminates an empty list item that was still being rendered when the Admin link was hidden for permissions reasons.

Tested in Voiceover and Safari, iOS Voiceover, Edge and NVDA, Chrome and NVDA, Firefox and NVDA.

## Added tests?

- [ ] Yes
- [X] No, and this is why: We need Cypress for these tests! Alternatively I could rewrite this menu logic into an ES module and test it with Jest.

## Added to documentation?

- [ ] README
- [X] No documentation needed

Multiasking with keyboards and coffee


app/views/layouts/_nav_menu.html.erb
```diff
@@ -1,25 +1,29 @@
...
+        <a href="<%= admin_path %>" class="crayons-link crayons-link--block">Admin</a>
+      </li>
+      <li class="px-1"><a href="<%= dashboard_path %>" class="crayons-link crayons-link--block">Dashboard</a></li>
+      <li class="px-1"><a href="<%= new_path %>" class="crayons-link crayons-link--block">Write a Post</a></li>
+      <li class="px-1"><a href="<%= readinglist_path %>" class="crayons-link crayons-link--block">Reading list</a></li>
+      <li class="border-0 border-b-1 border-solid border-base-20 px-1">
+        <a href="<%= user_settings_path %>" class="crayons-link crayons-link--block">Settings</a>
+      </li>
 
+      <li class="px-1">
```
ludwiczakpawel, line 23:
```suggestion
      <li class="px-1 py-1">
```

Same for spacing.

app/views/layouts/_nav_menu.html.erb
```diff
@@ -1,25 +1,29 @@
...
-    <div class="p-1">
-      <a href="<%= signout_confirm_path %>" class="crayons-link crayons-link--block" id="last-nav-link">Sign Out</a>
-    </div>
+      <li class="px-1 js-header-menu-admin-link data-no-instant hidden">
+        <a href="<%= admin_path %>" class="crayons-link crayons-link--block">Admin</a>
+      </li>
+      <li class="px-1"><a href="<%= dashboard_path %>" class="crayons-link crayons-link--block">Dashboard</a></li>
+      <li class="px-1"><a href="<%= new_path %>" class="crayons-link crayons-link--block">Write a Post</a></li>
+      <li class="px-1"><a href="<%= readinglist_path %>" class="crayons-link crayons-link--block">Reading list</a></li>
+      <li class="border-0 border-b-1 border-solid border-base-20 px-1">
```
ludwiczakpawel, line 19:
```suggestion
      <li class="border-0 border-b-1 border-solid border-base-20 px-1 pb-1">
```

Same for spacing.

app/views/layouts/_nav_menu.html.erb
```diff
@@ -1,25 +1,29 @@
...
-    <div id="user-profile-link-placeholder" class="border-0 border-b-1 border-solid border-base-20 p-1">
-      <a id="first-nav-link" class="crayons-link crayons-link--block">
-        <div>
-          <span class="fw-medium block">...</span>
-          <small class="fs-s color-base-50"></small>
-        </div>
-      </a>
-    </div>
+    <ul class="p-0" id="crayons-header__menu__dropdown__list">
+      <li id="user-profile-link-placeholder" class="border-0 border-b-1 border-solid border-base-20 p-1">
```
ludwiczakpawel, line 4:
```suggestion
      <li id="user-profile-link-placeholder" class="border-0 border-b-1 border-solid border-base-20 p-1 mb-1">
```

This will add a spacing that is currently missing.

app/views/layouts/_nav_menu.html.erb
```diff
@@ -1,25 +1,29 @@
...
-      <a href="<%= admin_path %>" class="crayons-link crayons-link--block js-header-menu-admin-link hidden" data-no-instant>Admin</a>
-      <a href="<%= dashboard_path %>" class="crayons-link crayons-link--block">Dashboard</a>
-      <a href="<%= new_path %>" class="crayons-link crayons-link--block">Write a Post</a>
-      <a href="<%= readinglist_path %>" class="crayons-link crayons-link--block">Reading list</a>
-      <a href="<%= user_settings_path %>" class="crayons-link crayons-link--block">Settings</a>
-    </div>
-    <div class="p-1">
-      <a href="<%= signout_confirm_path %>" class="crayons-link crayons-link--block" id="last-nav-link">Sign Out</a>
-    </div>
+      <li class="px-1 js-header-menu-admin-link data-no-instant hidden">
```
ludwiczakpawel, line 13: `data-no-instant` was an attribute before. Is this turned into a class intentionally?

app/assets/stylesheets/components/header.scss
```diff
@@ -119,6 +119,15 @@
     );
     opacity: var(--dropdown-opacity, 0);
     visibility: var(--dropdown-visibility, hidden);
+
+    ul {
+      list-style: none;
+    }
```
ludwiczakpawel, line 125: This should not be needed. Our CSS reset is taking care of it (hopefully :D).

ludwiczakpawel (changes requested):
Couple little details in the comments, but also something is broken when I click on the avatar and then try to click somewhere else to dismiss the dropdown.

Also, we should figure out what to do with focus style on the avatar because it doesn't look well :D  Actually, maybe we could NOT introduce it in this PR and handle focus state holistically for all buttons but in a separate PR? I mean they do have a focus style but based on your updates here I assume that focus state is not distinguishable enough, right?

app/views/layouts/_nav_menu.html.erb
marcysutton: Ah, no! Good catch. And I'm guessing that should data attribute should probably stay on the link for InstantClick, so I'll replace it. Thanks!

marcysutton:
> Something is broken when I click on the avatar and then try to click somewhere else to dismiss the dropdown.

I see that now, I'll take a look!

> We should figure out what to do with focus style on the avatar because it doesn't look well :D Actually, maybe we could NOT introduce it in this PR and handle focus state holistically for all buttons but in a separate PR? I mean they do have a focus style but based on your updates here I assume that focus state is not distinguishable enough, right?

I totally agree it needed work, and I tried playing with it some more but could not get a style that looked nice so I've removed it for now. The main problem is the menu won't show on focus anymore with this PR, so users won't have any visual indication for keyboard focus on the avatar button without something added. I tried changing `outline` to `border` and that worked better with border radius, but there were layers of borders and the profile image would zoom if I put a border onto the child span or image. A follow-up PR would be cool if it could be done quickly! Otherwise, I can keep tinkering with it.

I've made some of the smaller updates, and will work on it some more. I'd really like to add some tests, too–maybe I can add some with Jest for vanilla JavaScript, since this lives outside of the logged-in experience. 🤔

marcysutton:
I pushed some updates to this PR! 

Clicking outside to close should work now. I had to add a document listener since the menu has focus management in it on open, so the `blur` event on the navigation button was firing too early to be useful.

I also updated the focus style so it shows an outline in the `accent-color`, with some of the padding intact to add a contrasting color. My worry was that without the padding around the avatar image, the blue might not be visible against the user's image depending on its pixel colors. Note: I intentionally made the focus style only show on actual focus of the button, not when the menu is open or the button is hovered. I'm happy to discuss potential adjustments to it!

Button focus style with a blue outline

Dropdown showing style with gray button, blue outline on first focused dropdown item

Left a question for @nickytonline regarding tests, since I wasn't sure if there was any precedent for testing these initializer JS files with plain functions in them.

app/assets/stylesheets/components/header.scss
```diff
@@ -100,11 +100,13 @@
 
   &__trigger {
     all: unset;
-    padding: var(--su-1);
+    padding: var(--su-0);
```
marcysutton, line 103: 4px seemed like a lot of padding inside a blue focus outline, but that was the smallest variable listed. Happy to rename or rethink.

app/assets/stylesheets/components/header.scss
```diff
@@ -120,18 +122,23 @@
...
     .crayons-header__menu__trigger {
+      border: 2px solid var(--header-bg-hover);
       color: var(--header-icons-color-hover);
-      background: var(--header-bg-hover);
+    }
+  }
+  &.showing :focus,
+  :focus {
+    &.crayons-header__menu__trigger {
+      border: 2px solid var(--header-button-focus-color);
```
marcysutton, line 140: One thing I don't love is the flash of the focus style as a mouse user, where the navigation button is focused for a hot second before being moved to the first nav link (for focus management). With `focus-visible` in use, you don't see the focus style being moved to the first nav link as a mouse user. I could try to play with it some more, if anyone feels strongly about it.

nickytonline:
@marcysutton, those JS functions are not in ES modules, so there are some hacky ways to get them into jest for testing, but I think for now, manual testing will do and I will focus on getting us E2E testing in place. Thanks again for the work on this! 🔥

nickytonline:
I restarted one of the build nodes because of flaky tests. Hopefully Travis is happy soon!

nickytonline:
> I pushed some updates to this PR!
> 
> Clicking outside to close should work now. I had to add a document listener since the menu has focus management in it on open, so the `blur` event on the navigation button was firing too early to be useful.
> 
> I also updated the focus style so it shows an outline in the `accent-color`, with some of the padding intact to add a contrasting color. My worry was that without the padding around the avatar image, the blue might not be visible against the user's image depending on its pixel colors. Note: I intentionally made the focus style only show on actual focus of the button, not when the menu is open or the button is hovered. I'm happy to discuss potential adjustments to it!
> 
> Button focus style with a blue outline
> 
> Dropdown showing style with gray button, blue outline on first focused dropdown item
> 

Looks great with the accent color and padding. I'll let @ludwiczakpawel or @lisasy chime in on this too though.

app/assets/stylesheets/components/header.scss
ludwiczakpawel: That extra variables returns 2.4px which looks a bit off (since most screens don't really respect fractions):

image

You can easily notice more padding on the right comparing to the left one. So in this case I'd rather just use one-off like:

```
padding: 2px;
or
padding: calc(var(--su-1) / 2);
```

app/assets/stylesheets/components/header.scss
ludwiczakpawel: I've noticed that too - and seemed a bit jarring. In addition - the dropdown now requires me to explicitly open it when navigating with keyboard - which wasn't a case in previously.. Is that intentional?

ludwiczakpawel (changes requested):
I'd still push to move the trigger style update to a separate PR as suggested earlier:

> maybe we could NOT introduce it in this PR and handle focus state holistically for all buttons but in a separate PR? 

It clearly isn't perfect just yet, so it requires much more polishing and instead of doing it for this one-off I think it would make more sense to do it holistically for all buttons. Also, I wanna be 100% sure the new global variables are worth being added (like the new spacing unit as well as variable for focus color). Lets make these improvements but one at a time :) This PR will already improve accessibility for dropdown without the trigger updates.

app/assets/stylesheets/components/header.scss
marcysutton: >  the dropdown now requires me to explicitly open it when navigating with keyboard - which wasn't a case in previously.. Is that intentional?

Yes, it is intentional as the dropdown forced keyboard users through it before (quite a few extra tab stops if you're trying to go past). The keyboard navigation convention for dropdowns is to open on Enter or Space key on the triggering element, and make the child items only focusable when the dropdown is open.

marcysutton:
Thanks for the review! I have a comment about this, though:
> I'd still push to move the trigger style update to a separate PR as suggested earlier

The dropdown will be pretty unusable to sighted keyboard users if this is merged without a visible focus style, since the dropdown won't open on focus anymore. Here's a screenshot of how it looks on the live site right now, and if the dropdown isn't open you'll have a hard time seeing anything is focused as there is only an extremely faint gray border on the avatar image:

Screenshot of the header dropdown menu open

I've already pushed a fix for the padding, and I'll play around with focus-visible on click. I think it's getting close enough to be viable, and it could still be iterated upon after.

app/assets/stylesheets/components/header.scss
marcysutton: I pushed an update, should be good now with no flash of the focus outline on click.

app/views/layouts/_top_bar.html.erb
```diff
@@ -53,7 +53,7 @@
           <% if user_signed_in? %>
             <span class="crayons-avatar crayons-avatar--l"><img class="crayons-avatar__image" alt="" id="nav-profile-image" width="32" height="32" /></span>
           <% else %>
-            <%= inline_svg_tag("hamburger.svg", aria: true, class: "crayons-icon", title: "Navigation menu") %>
+            <%= inline_svg_tag("hamburger.svg", aria: false, class: "crayons-icon") %>
```
ludwiczakpawel, line 56:
```suggestion
            <%= inline_svg_tag("hamburger.svg", class: "crayons-icon") %>
```

nickytonline: @ludwiczakpawel, I'm guessing you're suggesting to remove it because the [default is false](https://github.com/jamesmartin/inline_svg#accessibility)?

marcysutton: That's how I understood it, thought I made a commit for it but maybe there was more than one in the PR. Thanks!

marcysutton:
I've resolved the merge conflicts on this PR, so it should be good to go.

I am also working on some Cypress tests for the Header dropdown but they need to work behind a login, and will take a bit of time to figure out. I don't think the tests should block this PR, so it doesn't suffer merge conflict fate again!

nickytonline:
> I've resolved the merge conflicts on this PR, so it should be good to go.
> 
> I am also working on some Cypress tests for the Header dropdown but they need to work behind a login, and will take a bit of time to figure out. I don't think the tests should block this PR, so it doesn't suffer merge conflict fate again!

I'm just going to restart the build on Travis. There's rspec errors, but they do not seem to be related to your code changes @marcysutton.

marcysutton:
I merged from master again to see if it makes Travis happy, since the failures looked unrelated to this PR!

nickytonline:
@marcysutton, I tested this locally and it works for focus in the dropdown and pressing escape to get out as expected. One thing I noticed though is it stops working as soon as you change the page in the app via one of the other links or call to action buttons. This is most likely related to [InstantClick](https://docs.forem.com/frontend/instant-click).

Just looking into it.

nickytonline:
> > @marcysutton, I tested this locally and it works for focus in the dropdown and pressing escape to get out as expected. One thing I noticed though is it stops working as soon as you change the page in the app via one of the other links or call to action buttons. This is most likely related to [InstantClick](https://docs.forem.com/frontend/instant-click).
> > Just looking into it.
> 
> Alright, I figured out the issue. The header area does not get reset with new markup when InstantClick is running so the same events keep getting attached because the script runs every time the page changes, but it's InstantClick single page app mode, not a real full refresh of the markup.
> 
> When there are an even number of the same events, it doesn't work, but when there are an odd number of events it works because it basically toggles n times.
> 

Alright I have it fixed and the functionality is now loaded from `app/javascript/packs/base.js`. Need to add some comments and tests still.

app/assets/javascripts/base.js.erb
```diff
@@ -114,8 +114,9 @@ var instantClick
 
   function changePage(title, body, newUrl, scrollY, pop) {
     var pageContentDiv = document.getElementById("page-content");
-    if (document.getElementById("navigation-butt")) {
-      document.getElementById("navigation-butt").classList.remove('showing')
+    var memberMenuButton = document.getElementById("member-menu-button")
```
nickytonline, line 117: The change here is `navigation-butt` is now `member-menu-button`. This new ID is used with other changes in this PR.

app/assets/javascripts/initializePage.js
```diff
@@ -1,7 +1,7 @@
 /*
   global initializeLocalStorageRender, initializeBodyData,
   initializeAllChatButtons, initializeAllTagEditButtons, initializeUserFollowButts,
-  initializeBaseTracking, initializeTouchDevice, initializeCommentsPage,
+  initializeBaseTracking, initializeCommentsPage,
```
nickytonline, line 4: Pulled this out because this logic is now bootstrapped in `app/javascript/packs/base.js`.

app/assets/javascripts/initializePage.js
```diff
@@ -40,7 +40,6 @@ function callInitializers() {
 
   initializeBaseTracking();
   initializePaymentPointers();
-  initializeTouchDevice();
```
nickytonline, line 43: Pulled this out because this logic is now bootstrapped in `app/javascript/packs/base.js`.

app/javascript/packs/base.js
```diff
@@ -2,6 +2,7 @@ import {
   initializeMobileMenu,
   setCurrentPageIconLink,
   getInstantClick,
+  initializeTouchDevice,
```
nickytonline, line 5: This is what was pulled out of base.js.erb.

docs/frontend/plain-js.md
```diff
@@ -33,7 +33,6 @@ that major sections of the websites are bootstrapped, for example:
 
 ```javascript
 initializeBaseTracking();
-initializeTouchDevice();
```
nickytonline, line 36: Removed this line from the docs as it is no longer true.

marcysutton:
Thanks for working on this PR to keep it moving along, @nickytonline! I tested it out locally and it seems to be working well.

app/assets/javascripts/base.js.erb
marcysutton: I didn't even know this code was in there. Thanks for the update!

app/javascript/topNavigation/utilities.js
```diff
@@ -1,3 +1,106 @@
...
+            menuNavButton.focus();
+          }
+        });
+        document
+          .getElementById('last-nav-link')
+          .addEventListener('blur', function (_event) {
+            blurHeaderMenu(
+              memberTopMenu,
+              menuNavButton,
+              document.getElementById('seond-last-nav-link'),
```
marcysutton, line 86: This looks to be a typo with the `seond-last-nav-link` [sic]. Even if corrected, does that Id match any element on the page? I wasn't actually sure.

app/views/layouts/_top_bar.html.erb
```diff
@@ -47,7 +47,7 @@
         </a>
 
         <div class="crayons-header__menu" id="crayons-header__menu">
-          <button type="button" class="crayons-header__menu__trigger navigation-butt" id="navigation-butt" aria-label="Navigation menu">
+          <button type="button" class="crayons-header__menu__trigger" id="member-menu-button" aria-label="Navigation menu">
```
marcysutton, line 50: It makes sense to rename this, but I will add that `navigation-butt` made me chuckle every time I saw it.

app/javascript/topNavigation/utilities.js
nickytonline: I'll take a peek. Thanks @marcysutton!

aitchiss: I've added `second-last-nav-link` to the appropriate menu element and corrected the typo so now we don't close the menu if a user tabs backwards from the last option 👍

aitchiss:
@ludwiczakpawel just catching up on the requested changes in this one so we can move it forward - based on the previous conversations, and the changes to the focus style only being on keyboard navigation, are you happy to re-review?

what it looks like now with keyboard vs mouse click 👇

gif of current styled for keyboard vs mouse focus

db/schema.rb
```diff
@@ -802,6 +801,7 @@
     t.integer "unspent_credits_count", default: 0, null: false
     t.datetime "updated_at", null: false
     t.string "url"
+    t.datetime "latest_article_updated_at"
```
nickytonline: As mentioned in Slack, you can exclude this file from the PR.

nickytonline (changes requested):

app/javascript/topNavigation/__tests__/utilities.test.js
```diff
@@ -134,4 +141,107 @@ describe('top navigation utilitities', () => {
...
+
+  describe('initializeTouchDevice', () => {
+    const initialMenuHTML = `
+      <div class="crayons-header__menu" id="crayons-header__menu" data-testid="menu-dropdown">
+        <button type="button" class="crayons-header__menu__trigger" id="member-menu-button" aria-label="Navigation menu">
+          <span class="crayons-avatar crayons-avatar--l"><img class="crayons-avatar__image" alt="" id="nav-profile-image" src=""></span>
+        </button>
+        <div class="crayons-dropdown left-2 right-2 s:right-4 s:left-auto p-0 crayons-header__menu__dropdown inline-block">
+          <ul class="p-0" id="crayons-header__menu__dropdown__list">
+            <li id="user-profile-link-placeholder" class="border-0 border-b-1 border-solid border-base-20 p-1">
```
ludwiczakpawel, line 153:
```suggestion
            <li id="user-profile-link-placeholder" class="border-0 border-b-1 border-solid border-base-20 p-1 mb-1">
```

app/javascript/topNavigation/__tests__/utilities.test.js
```diff
@@ -134,4 +141,107 @@ describe('top navigation utilitities', () => {
       );
     });
   });
+
+  describe('initializeTouchDevice', () => {
+    const initialMenuHTML = `
```
ludwiczakpawel, line 146: Out of curiosity, since I know very little about JS tests: it looks like we basically duplicate a piece of code here that is also written in frontend itself - is this intentional solution? Is this sustainable?

app/assets/stylesheets/components/header.scss
```diff
@@ -101,18 +103,23 @@
...
 
     .crayons-header__menu__trigger {
+      border: 2px solid var(--header-bg-hover);
       color: var(--header-icons-color-hover);
-      background: var(--header-bg-hover);
+    }
+  }
+  &.showing .focus-visible:focus,
+  .focus-visible:focus {
+    &.crayons-header__menu__trigger {
```
ludwiczakpawel, line 120: I think we can optimize this entire section and make it more readable by relying on CSS variables and `:focus-within`:

```
.crayons-header__menu {
  --trigger-color: var(--header-icons-color);
  --trigger-border-color: transparent;
  ...

  &__trigger {
    ...
    color: var(--trigger-color);
    border: 2px solid var(--trigger-border-color);
  }

  &.showing,
  &.desktop:hover {
    --trigger-border-color: var(--header-bg-hover);
    --trigger-color: var(--header-icons-color-hover);
    ...
  }
  &:focus-within,
  &.showing,
  &.desktop:hover:focus-within {
    --trigger-border-color: var(--header-button-focus-color);
    --trigger-color: var(--header-button-focus-color);
  }
}
```

^ it also requires [removing this piece of code](https://github.com/forem/forem/blob/master/app/assets/stylesheets/tag-edit.scss#L32-L34) - which, to be fair, shouldn't be in that file anyway. and it seems like causing more confusion (especially with specificity) anyway.

Thoughts?

ludwiczakpawel (changes requested):
This is almost good to go - I've left one or two suggestions to make code more readable.

I still think the :focus state should be handled globally for all buttons, not only that one avatar dropdown trigger but we could handle it separately.

app/assets/stylesheets/components/header.scss
ludwiczakpawel: Actually... I've just pushed this code :D Let me know if it makes sense.

aitchiss: The variables definitely making things more readable 👍  Only issue I see though is that removing use of `focus-visible` means that the focus border appears around the profile image on click as well as on keyboard focus. It also seems to be hanging around even if you click the menu and then start tabbing and moving focus away.

I'll tweak what you've added and then hopefully we should be good

app/javascript/topNavigation/__tests__/utilities.test.js
aitchiss: Yes and no 🙂

It's deliberate, but no it's not sustainable/ideal - it's just a stop gap until we can move the tests to E2E tests in Cypress. There's a comment at the top of the test file to explain this (but it was added previously so not obvious in this PR!)

app/assets/stylesheets/components/header.scss
ludwiczakpawel: I thought that's what was originally suggested in this PR, no? Nevertheless, feel free to make any adjustments you want :)

aitchiss: It's been a long thread 😅  The proposal is to show the border around the profile image when it's focused - with the changes it's persisting after a mouse click even though the focus has moved elsewhere 🙃

ludwiczakpawel: > It's been a long thread

Indeed. I no longer know what's good or bad :D As I said - adjust it the way you think makes the most sense 👍

db/schema.rb
aitchiss: reverted this a few commits ago - must have crossed with your review :)

ludwiczakpawel (approved):

nickytonline (approved):
Thanks for finishing up this PR @aitchiss. I'm sure @marcysutton will be happy to see this is done! 🚀

PR #10931, opened 2020-10-19
API: Endpoint to get an organization's details

twinsfan421:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This would provide an api show endpoint `dev.to/api/organizations/{org_username}` to search organizations by their username.  
As discussed in issue 9212, This is the first of 4 related endpoints that will include:
```
dev.to/api/organizations/{org_username}/users
dev.to/api/organizations/{org_username}/articles
dev.to/api/organizations/{org_username}/listings
```
## Related Tickets & Documents
https://github.com/forem/forem/issues/9212

## Added tests?

- [x] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [x] docs.forem.com
- [ ] readme
- [ ] no documentation needed

Issue #9212, opened 2020-07-08
Query organization on dev.to/api


gixxerblade:
## Feature Request

Currently there in no way to query an organization on dev.to/api

**Describe the solution you'd like**

It would be nice to query https://dev.to/api/organization/articles, https://dev.to/api/organization/users, https://dev.to/api/organization/listings where organization is e.g., https://dev.to/api/vetswhocode

The description could read: 
>This endpoint allows the client to retrieve an organization .
>
>"Organizations" is all the data that an organization creates on DEV that typically show up in the feed and a user is connected >to. They can be a blog post, a discussion question, a help thread etc. but is referred to as an organization within the code.
>
>By default it will return featured, published articles ordered by descending date of publication
>
>It supports pagination, each page will contain 30 articles by default.

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

rhymes:
I think the URL structure should be something like:

`dev.to/api/organizations/{org_username}/articles` and so on

twinsfan421:
I can work on this

rhymes:
I've assigned you the issue, thanks @twinsfan421 !

twinsfan421:
@rhymes I am working on it and have couple questions.  
In addition to the articles, users and listings, should I make a show action like dev.to/api/organizations/{org_username} for just the organization's info?
Should there be any authentication for these endpoints?
Thanks

citizen428:
> In addition to the articles, users and listings, should I make a show action like dev.to/api/organizations/{org_username} for just the organization's info?

Yes, that's probably a good idea. It should mirror https://docs.dev.to/api/#operation/getUser.

>  Should there be any authentication for these endpoints?

Rough rule of thumb that the current API seems to follow: if it's public information, no authentication is needed. Best to follow similar API endpoints for now, the API docs are a good reference for this: https://docs.dev.to/api/

cmgorton:
[...]

PR #11514, opened 2020-11-20
API: Endpoint to get an organization's users

twinsfan421:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This would provide an api endpoint `dev.to/api/organizations/{org_username}/users` to get a list of an organization's users.
As discussed in issue 9212, This is the 2nd of 4 related endpoints that will include:
```
dev.to/api/organizations/{org_username}/articles
dev.to/api/organizations/{org_username}/listings
```

#9212  #10931

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [ ] No documentation needed

### DIFF
--- a/app/controllers/api/v0/organizations_controller.rb
+++ b/app/controllers/api/v0/organizations_controller.rb
@@ -9,14 +9,15 @@ module Api
       ].freeze
       private_constant :SHOW_ATTRIBUTES_FOR_SERIALIZATION
 
-      ORG_USERS_FOR_SERIALIZATION = %i[
-        id username name twitter_username github_username profile_image website_url
+      USERS_FOR_SERIALIZATION = %i[
+        id username name twitter_username github_username
+        profile_image website_url location summary created_at
       ].freeze
-      private_constant :ORG_USERS_FOR_SERIALIZATION
+      private_constant :USERS_FOR_SERIALIZATION
 
       def show
         @organization = Organization.select(SHOW_ATTRIBUTES_FOR_SERIALIZATION)
-          .find_by!(username: params[:org_username])
+          .find_by!(username: params[:username])
       end
 
       def users
@@ -24,13 +25,13 @@ module Api
         num = [per_page, 1000].min
         page = params[:page] || 1
 
-        @users = @organization.users.select(ORG_USERS_FOR_SERIALIZATION).page(page).per(num)
+        @users = @organization.users.select(USERS_FOR_SERIALIZATION).page(page).per(num)
       end
 
       private
 
       def find_organization
-        @organization = Organization.find_by!(username: params[:org_username])
+        @organization = Organization.find_by!(username: params[:organization_username])
       end
     end
   end
--- a/app/controllers/api/v0/organizations_controller.rb
+++ b/app/controllers/api/v0/organizations_controller.rb
@@ -1,15 +1,36 @@
 module Api
   module V0
     class OrganizationsController < ApiController
+      before_action :find_organization, only: %i[users]
+
       SHOW_ATTRIBUTES_FOR_SERIALIZATION = %i[
         username name summary twitter_username github_username url
         location created_at profile_image tech_stack tag_line story
       ].freeze
       private_constant :SHOW_ATTRIBUTES_FOR_SERIALIZATION
 
+      ORG_USERS_FOR_SERIALIZATION = %i[
+        id username name twitter_username github_username profile_image website_url
+      ].freeze
+      private_constant :ORG_USERS_FOR_SERIALIZATION
+
       def show
         @organization = Organization.select(SHOW_ATTRIBUTES_FOR_SERIALIZATION)
           .find_by!(username: params[:org_username])
+      end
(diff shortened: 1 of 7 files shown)


mstruve (approved):
Code looks good to me. @rhymes is probably a better person to ensure the docs are good to go. Thanks for taking the time to breakout these API endpoints into their own PRs, it makes our jobs as reviewers a lot easier. We really appreciate it! 

🚀

benhalpern:
This looks good to me in terms of the base premise. (AKA it's all public info and matches good patterns, so totally should exist)..

Also defer to @rhymes for final say on API reviews.

config/routes.rb
```diff
@@ -204,6 +204,7 @@
         resources :organizations, only: [] do
           collection do
             get "/:org_username", to: "organizations#show"
+            get "/:org_username/users", to: "organizations#users"
           end
         end
```
rhymes, line 209: do you think we could use resource routes here?

For example:

```ruby
        resources :organizations, only: [:show], param: :username do
          resources :users, only: [:index], to: "organizations#users"
        end
```

(renamed the parameter to `:username` as it's implicit it's part of an org)

this will also have the effect of generating helpers such as: `api_organization_users_path` and `api_organization_path` that you can use in the tests.

Also, if the `API::OrganizationsController` becomes too big we can easily move the actions to their own resource controller (for example: `API::OrganizationsUsersController` without changing the external URLs)

app/controllers/api/v0/organizations_controller.rb
```diff
@@ -1,16 +1,37 @@
...
     class OrganizationsController < ApiController
+      before_action :find_organization, only: %i[users]
+
       SHOW_ATTRIBUTES_FOR_SERIALIZATION = %i[
         username name summary twitter_username github_username url
         location created_at profile_image tech_stack tag_line story
       ].freeze
       private_constant :SHOW_ATTRIBUTES_FOR_SERIALIZATION
 
+      ORG_USERS_FOR_SERIALIZATION = %i[
```
rhymes, line 12: I wouldn't make a difference between "regular users" and "organization users". As I don't think we plan to tell the API what's their role (admins, non admins) for privacy reasons, I think we can render the same partial we render for `/api/users/:id` in the `Api::V0::UsersController#show` route in a loop in the jbuilder file

Let me know if that makes sense

app/views/api/v0/organizations/users.json.jbuilder
```diff
@@ -0,0 +1,7 @@
+json.array! @users do |user|
+  json.type_of "org_user"
+  json.extract!(user, :name, :username, :twitter_username, :github_username)
+
+  json.website_url user.processed_website_url
+  json.profile_image Images::Profile.call(user.profile_image_url, length: 320)
```
rhymes, line 6: see my point about rendering `api/v0/views/show.json.jbuilder` here (you can use a common partial for the two)

docs/api_v0.yml
```diff
@@ -1147,6 +1147,30 @@ components:
           description: Profile image (640x640)
           format: url
 
+    OrgUser:
```
rhymes, line 1150: if we recycle the `User` entity, this is not needed

docs/api_v0.yml
```diff
@@ -1726,6 +1750,16 @@ components:
         story:
         profile_image: https://res.cloudinary.com/...jpeg
 
+    OrgUser:
```
rhymes, line 1753: same as above

docs/api_v0.yml
```diff
@@ -3023,6 +3057,54 @@ paths:
           source: |
             curl https://dev.to/api/organizations/ecorp
 
+  /organizations/{org_username}/users:
+    get:
+      operationId: getOrgUsers
+      summary: Organization's users
+      description: |
+        This endpoint allows the client to retrieve a list of an organization's users
```
rhymes, line 3065: I would rephrase this as "users belonging to the organization"

rhymes (changes requested):
Hi @twinsfan421, thanks a lot for the PR! I think we need to address a few implementation details but overall it works!

config/routes.rb
twinsfan421: Ya for sure. I just pushed up an update for the routes

app/controllers/api/v0/organizations_controller.rb
twinsfan421: Makes sense. I just pushed up a fix for this

app/views/api/v0/organizations/users.json.jbuilder
twinsfan421: I created a shared partial and pushed it up

twinsfan421:
Thanks @rhymes for reviewing. Let me know if you want me to change anything else

rhymes (approved):
@twinsfan421 it works great! 

Thank you!

PR #11520, opened 2020-11-20
Disable deploy to BenHalpern server

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Have no idea if this will work, but what if Travis log limit is shared among parallel build stages?
Our build stages are exactly ideantical (the destination server is different) so we might be doubling the log size and pushing beyond [the 4MB log limit](https://docs.travis-ci.com/user/common-build-problems/#log-length-exceeded)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,13 +92,13 @@ jobs:
         api_key: '$HEROKU_AUTH_TOKEN'
         app:
           master: practicaldev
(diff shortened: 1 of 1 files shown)


maestromac (approved):
Let's try it.

Issue #11490, opened 2020-11-19
PostCSS Warnings and Errors Flooding logs Causing Travis Build Failures


mstruve:
### Problem
Recently we have had [Travis builds fail](https://travis-ci.com/github/forem/forem/jobs/443192844) bc the log length exceeds the max allowed. This is a big problem bc despite the job showing as failed, the deploy still goes out successfully which sends the wrong message to the developer tracking their deploy. 

Here is an example of some of the logs:
```
yarn run v1.22.5
       $ postcss public/assets/*.css -d public/assets
       
       5:146912	⚠  Custom property ignored: not scoped to the top-level :root element (.site-logo { ... --max-width: ... }), in atrule [postcss-custom-properties]
       5:146969	⚠  Custom property ignored: not scoped to the top-level :root element (.site-logo { ... --max-width: ... }), in atrule [postcss-custom-properties]
       5:146988	⚠  Custom property ignored: not scoped to the top-level :root element (.site-logo { ... --font-size: ... }), in atrule [postcss-custom-properties]
       5:147013	⚠  Custom property ignored: not scoped to the top-level :root element (.site-logo { ... --font-weight: ... }), in atrule [postcss-custom-properties]
       5:147082	⚠  Cust
       
...

            1:104452    ⚠  variable '--su-2' is undefined and used without a fallback [postcss-custom-properties]
       1:106070 ⚠  variable '--body-color' is undefined and used without a fallback [postcss-custom-properties]
       1:106403 ⚠  variable '--body-color' is undefined and used without a fallback [postcss-custom-properties]
       1:107307 ⚠  variable '--link-brand-color' is undefined and used without a fallback [postcss-custom-properties]
       1:107500 ⚠  variable '--card-color-tertiary' is undefined and used without a fallback [postcss-custom-properties]
       1:107878 ⚠  variable '--card-bg' is undefined and used without a fallback [postcss-custom-properties]
       1:108523 ⚠  variable '--link-brand-color' is undefined and used without a fallback [postcss-custom-properties]
       1:108599 ⚠  variable '--card-bg' is undefined and used without a fallback [postcss-custom-properties]
       1:109255 ⚠  variable '--body-color' is undefined and used without a fallback [postcss-custom-properties]
       1:116689 ⚠  variable '--card-secondary-bg' is undefined and used without a fallback [postcss-custom-properties]
```  

[...]

PR #11518, opened 2020-11-20
Attempt at disabling postcss-next warnings

rhymes:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Closes https://github.com/forem/forem/issues/11490

Hopefully this works :D 

## Related Tickets & Documents

- https://github.com/rails/webpacker/blob/master/lib/install/config/postcss.config.js
- https://github.com/csstools/postcss-preset-env#features
- https://stackoverflow.com/a/49629759/4186181
- https://github.com/MoOx/postcss-cssnext/issues/186#issuecomment-123304414

PR #11521, opened 2020-11-20
Suppress Postcss warnings & Re-enable Benhalpern deploy

maestromac:
- [x] Bug Fix

## Description
We've been getting by log flood from `postcss-custom-properties` which is messing up our Travis build. Because that tool no longer provides warning suppression, this PR will suppress it via bash and output it to a file name `err.log` for future debugging.

Resolves https://github.com/forem/forem/issues/11490

Previous attempts
https://github.com/forem/forem/pull/11518
https://github.com/forem/forem/pull/11520

Run `bin/rails assets:precompile` locally and you should not see any warning.

n/a

## Added tests?
- [x] No, and this is why: This is a CI change

## Added to documentation?
- [x] No documentation needed

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "pretest": "npm run lint:frontend",
     "test": "jest app/javascript/ bin/ --coverage",
     "test:watch": "jest app/javascript/ bin/ --watch",
-    "postcss": "postcss public/assets/*.css -d public/assets 2> err.log"
+    "postcss": "postcss public/assets/*.css -d public/assets 2> postcss_error.log"
   },
   "husky": {
     "hooks": {
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "pretest": "npm run lint:frontend",
     "test": "jest app/javascript/ bin/ --coverage",
     "test:watch": "jest app/javascript/ bin/ --watch",
-    "postcss": "postcss public/assets/*.css -d public/assets"
+    "postcss": "postcss public/assets/*.css -d public/assets 2> err.log"
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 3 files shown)


rhymes (approved):
I would name this `postcss_error.log` so that we know it's about that but definitely not blocking

package.json
```diff
@@ -24,7 +24,7 @@
     "pretest": "npm run lint:frontend",
     "test": "jest app/javascript/ bin/ --coverage",
     "test:watch": "jest app/javascript/ bin/ --watch",
-    "postcss": "postcss public/assets/*.css -d public/assets"
+    "postcss": "postcss public/assets/*.css -d public/assets 2> err.log"
```
nickytonline, line 27: question (non-blockin): Is `2>` the same as `&>`? For context, see https://github.com/forem/forem/issues/11490#issuecomment-731256928

nickytonline (approved):
🚀

package.json
rhymes: @nickytonline `2>` only redirects standard error, `&>` redirects both standard output and standard error to the log file, it's a shortcut for `2>&1` I think

nickytonline: Thanks for the explanation @rhymes.

mstruve (approved):
❤️

benhalpern (approved):

benhalpern:
Something about this PR calls out to me

package.json
maestromac: Based on the answer [here](https://stackoverflow.com/questions/24793069/what-does-do-in-bash), `&>file` would log everything into the said file, which is exactly identical to  `1>file 2>file` (1 for standard output, 2 for standard error). I omitted the `1>` because I only want to store the error.

Now that I said that, I think there's a benefit to storing both together. let me update it.

```suggestion
    "postcss": "postcss public/assets/*.css -d public/assets &> precompile-postcss.log"
```

PR #11527, opened 2020-11-20
Update aspect ratio of images on the Admin's config page

maestromac:
- [x] Bug Fix

## Description
The img tags used in the config page doesn't have proper width and height specified. This PR addresses that and change some of the descriptions.

Please feel free to commit to this PR directly.

Resolves https://github.com/forem/InternalProjectPlanning/issues/293

0. Pull down this PR, set add the following role to your own local account `me.role(:single_resource_admin, Config)`
1. Start up the server, navigate to `/admin/config`
2. Click on the 'Images' dropdown.
3. All the images there shouldn't look out of proportion. Nothing should look too big or distorted.

I don't think there should be any concern here as I'm just changing the css.

## Added tests?
- [x] No, and this is why: this is mostly a css change

## Added to documentation?
I'll update the docs as soon as we come to an agreement

- [ ] README
- [ ] No documentation

n/a

### DIFF
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -170,7 +170,7 @@ module Constants
         placeholder: SVG_PLACEHOLDER
       },
       logo_png: {
-        description: "Used as a fallback to the SVG. Recommended minimum of 192x192px for PWA support",
+        description: "Used as a fallback to the SVG. Recommended minimum of 512x512px for PWA support",
         placeholder: IMAGE_PLACEHOLDER
       },
       logo_svg: {
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -170,7 +170,7 @@ module Constants
         placeholder: SVG_PLACEHOLDER
       },
       logo_png: {
-        description: "Minimum 1024px, used for PWA etc.",
+        description: "Minimum 192x192px, used for PWA etc.",
         placeholder: IMAGE_PLACEHOLDER
       },
       logo_svg: {
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -170,7 +170,7 @@ module Constants
         placeholder: SVG_PLACEHOLDER
       },
       logo_png: {
-        description: "Minimum 192x192px, used for PWA etc.",
+        description: "Used as a fallback to the SVG. Recommended minimum of 192x192px for PWA support",
         placeholder: IMAGE_PLACEHOLDER
       },
       logo_svg: {
--- a/app/lib/constants/site_config.rb
+++ b/app/lib/constants/site_config.rb
@@ -178,7 +178,7 @@ module Constants
         placeholder: SVG_PLACEHOLDER
       },
       main_social_image: {
-        description: "Used as the main image in social networks and OpenGraph",
+        description: "Used as the main image in social networks and OpenGraph. Recommended aspect ratio of 16:9 (600x337px,1200x675px)",
         placeholder: IMAGE_PLACEHOLDER
       },
       mailchimp_api_key: {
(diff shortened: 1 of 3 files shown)


app/models/site_config.rb
```diff
@@ -89,7 +89,7 @@ class SiteConfig < RailsSettings::Base
   # Images
   field :main_social_image, type: :string, default: proc { URL.local_image("social-media-cover.png") }
 
-  field :favicon_url, type: :string, default: "favicon.ico"
+  field :favicon_url, type: :string, default: proc { URL.local_image("favicon.ico") }
```
maestromac, line 92: Unless we wanted to purposely avoid providing a default favicon, I updated it to provide the DEV favicon as the default favicon.

app/lib/constants/site_config.rb
```diff
@@ -170,15 +170,15 @@ module SiteConfig
         placeholder: SVG_PLACEHOLDER
       },
       logo_png: {
-        description: "Minimum 1024px, used for PWA etc.",
+        description: "Minimum 192x192px, used for PWA etc.",
```
maestromac, line 173: the 1024px has to be a typo right? I'm not sure if there really is a required size at all but chose 192x192px because it's an ideal size to supports [both ios and android](https://stackoverflow.com/questions/48839338/which-icon-sizes-are-required-for-progressive-web-apps-pwa-as-of-q1-2018). Does that sound alright @benhalpern ?

app/lib/constants/site_config.rb
```diff
@@ -170,15 +170,15 @@ module SiteConfig
...
         placeholder: IMAGE_PLACEHOLDER
       },
       logo_svg: {
         description: "Used as the SVG logo of the community",
         placeholder: SVG_PLACEHOLDER
       },
       main_social_image: {
-        description: "Used as the main image in social networks and OpenGraph",
+        description: "Used as the main image in social networks and OpenGraph. Recommended aspect ratio of 16:9 (600x337px,1200x675px)",
```
maestromac, line 181: I've received mixed answers from my google searches. 16:9 is definitely the ideal ratio but the exact pixel around what I listed there.

app/lib/constants/site_config.rb
```diff
@@ -170,15 +170,15 @@ module SiteConfig
         placeholder: SVG_PLACEHOLDER
       },
       logo_png: {
-        description: "Minimum 1024px, used for PWA etc.",
+        description: "Used as a fallback to the SVG. Recommended minimum of 192x192px for PWA support",
```
maestromac, line 173: the 1024px has to be a typo right? I'm not sure if there really is a required size at all but chose 192x192px because that's the ideal size to support [PWA](https://web.dev/add-manifest/). Does that sound alright @benhalpern ?

benhalpern: I think we need a higher minimum.

Right now 1024px is chosen because we use it with sizes up to 512 and it's always much better to have a bigger logo we can downsize vs the other way around. 512 is probably good enough but I wouldn't go lower than that.

app/lib/constants/site_config.rb
benhalpern: ... It's annoying because [Twitter uses `2:1`](https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image)....

However, I think we might want to not get too clever with this and just say 16:9 and know that Twitter will clip it a little bit. It should be close enough.

app/lib/constants/site_config.rb
maestromac: Sounds good. I'm changing it to 512 x 512px
```suggestion
        description: "Used as a fallback to the SVG. Recommended minimum of 512x512px for PWA support",
```

benhalpern (approved):
This looks good to me

Zhao-Andy (approved):

PR #11559, opened 2020-11-22
Include core assets in serviceworker cache

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Currently when we deploy Forem Cloud we do not retain assets for old versions of the platform, and service workers do not seem to reliably retain the in-memory cached version of assets, so we sometimes get flashes of unstyled content and other frontend issues.

This adds more static assets to the serviceworker cache, and also adds functionality to _periodically_ flush that component of the serviceworker cache to make sure we don't over-cache.

Things should _continue to work as normal_. If you open up the application tab of chrome tools and view Cache Storage you'll see static 1.2 cache including a bunch of assets.

### DIFF
--- a/app/views/service_worker/index.js.erb
+++ b/app/views/service_worker/index.js.erb
@@ -137,6 +137,22 @@
       return;
     }
 
+    // Fetch and/or JS and CSS assets for better persistence than memory cache.
+    if (url.pathname.startsWith("/assets/") || url.pathname.startsWith("/packs/")) {
+      event.respondWith(
+        caches.open(staticCacheName).then(function (cache) {
+          return cache.match(event.request).then(function (response) {
+            return (
+              response ||
+              fetch(event.request).then(function (response) {
+                cache.put(event.request, response.clone());
+                return response;
+              })
+            );
+          });
+        })
+      );
+    }
     if (url.origin === location.origin) {
       if (event.clientId === "" && // Not fetched via AJAX after page load.
         event.request.method == "GET" && // Don't fetch on POST, DELETE, etc.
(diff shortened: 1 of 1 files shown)


atsmith813 (approved):

app/views/service_worker/index.js.erb
```diff
@@ -187,6 +203,22 @@
             "/shell_bottom",
           ]));
       }
+
+      // Periodically delete assets when they pile up
```
nickytonline, line 207: question: The bespoke periodic deletion of assets may not be necessary. I'm not positive, but see the [docs on Eviction](https://chromium.googlesource.com/chromium/src/+/master/content/browser/service_worker/README.md#Eviction) in the Service workers documentation.

benhalpern: I think being explicit here might be safer to rely on any functionality we're not crystal clear on.

We also had some issues early on with eviction with service workers.

nickytonline: Pulling this locally to give it a go.

app/views/service_worker/index.js.erb
```diff
@@ -137,6 +137,22 @@
       return;
     }
 
+    // Fetch and/or JS and CSS assets for better persistence than memory cache.
+    if (url.pathname.startsWith("/assets/") || url.pathname.startsWith("/packs/")) {
```
nickytonline, line 141: This makes sense and will fix the issue. I recommend we do a test drive of this fix with @jdoss   to ensure the fix is solid or did you do that already @benhalpern?

nickytonline (approved):
Seems to behave properly. Any reason we don't change the cache name on deploys, or at least on frontend PRs? That seems like it would help with evicting old assets. From what I can see, the last time we changed the cache name was about a year ago in #5262.

PR #11560, opened 2020-11-22
Use content-visibility: auto for small render performance boost

benhalpern:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

`content-visibility` provides control over when elements are rendered, so rendering can be skipped for elements not yet in the user's viewport. It is only available in Chromium browsers, and shouldn't negatively affect the other browsers which have not launched this feature.

These posts do a great job explaining it...

https://web.dev/content-visibility/
https://dev.to/dailydevtips1/i-made-my-website-28ms-faster-with-content-visibility-466e

This PR only applies to a handful of divs. We could go further with this, but this is some simple low-hanging fruit— Especially with the footer appearing on every single page load.

This functionality uses `contain-intrinsic-size` to tell the browser approximately what size to expect. It doesn't have to be spot on, but it should be approximately correct. That makes it harder to use with, say, comments where they could be really any size variation and predicting a good height to pitch to the browser seems more trouble than it's worth. In this case it takes elements which won't vary greatly in size (even with different window widths).

**SEO:** This should have a positive impact on UX in a very tiny way, but since this is being promoted by Google, I have to think the search crawler should appreciate it.

It's hard to see this working. But if the size was calculated decently the scrollbar shouldn't radically change in sizes.

I believe this would not affect accessibility by implementing this in the standard way, but if anyone has ideas why it would, feel free to chime in.

hidden

### DIFF
--- a/app/assets/stylesheets/article-show.scss
+++ b/app/assets/stylesheets/article-show.scss
@@ -41,11 +41,6 @@ article {
   }
 }
 
-#bottom-content-read-next {
-  content-visibility: auto;
-  contain-intrinsic-size: 450px;
-}
-
 .container {
   width: 880px;
   max-width: 100%;
--- a/app/assets/stylesheets/article-show.scss
+++ b/app/assets/stylesheets/article-show.scss
@@ -41,6 +41,11 @@ article {
   }
 }
 
+#bottom-content-read-next {
+  content-visibility: auto;
+  contain-intrinsic-size: 450px;
+}
+
 .container {
   width: 880px;
   max-width: 100%;
--- a/app/views/articles/_bottom_content.html.erb
+++ b/app/views/articles/_bottom_content.html.erb
@@ -1,5 +1,5 @@
-<% if articles %>
-  <section class="crayons-card crayons-card--secondary text-padding mb-4 print-hidden">
+<% if articles && articles.any? %>
+  <section class="crayons-card crayons-card--secondary text-padding mb-4 print-hidden" id="bottom-content-read-next">
     <h2 class="crayons-subtitle-1">Read next</h2>
     <% articles.each do |article| %>
       <a href="<%= article.path %>" data-preload-image="<%= cloud_cover_url(article.main_image) %>" class="crayons-link mt-6 block">
--- a/app/assets/stylesheets/views/comments.scss
+++ b/app/assets/stylesheets/views/comments.scss
@@ -1,8 +1,6 @@
 @import '../config/import';
 
 .comment-form {
-  content-visibility: auto;
-  contain-intrinsic-size: 78px;
   --preview-display: none;
   --field-display: flex;
 
--- a/app/assets/stylesheets/views/comments.scss
+++ b/app/assets/stylesheets/views/comments.scss
@@ -1,6 +1,8 @@
 @import '../config/import';
 
 .comment-form {
+  content-visibility: auto;
+  contain-intrinsic-size: 78px;
   --preview-display: none;
   --field-display: flex;
 
--- a/app/assets/stylesheets/views/footer.scss
+++ b/app/assets/stylesheets/views/footer.scss
@@ -1,8 +1,14 @@
 @import '../config/import';
 
-.crayons-footer {
+#footer-container {
   content-visibility: auto;
-  contain-intrinsic-size: 320px;
+  contain-intrinsic-size: 300px; // Chromium only. See our discussion here: https://github.com/forem/forem/pull/11560
+  @media (min-width: $breakpoint-l) {
+    contain-intrinsic-size: 260px;
+  }
+}
+
+.crayons-footer {
   --footer-padding: var(--su-5);
   background: var(--footer-bg);
   color: var(--footer-color);
--- a/app/assets/stylesheets/views/footer.scss
+++ b/app/assets/stylesheets/views/footer.scss
@@ -1,6 +1,8 @@
 @import '../config/import';
 
 .crayons-footer {
+  content-visibility: auto;
+  contain-intrinsic-size: 320px;
   --footer-padding: var(--su-5);
   background: var(--footer-bg);
   color: var(--footer-color);


app/views/articles/_bottom_content.html.erb
```diff
@@ -1,5 +1,5 @@
-<% if articles %>
-  <section class="crayons-card crayons-card--secondary text-padding mb-4 print-hidden">
+<% if articles && articles.any? %>
```
benhalpern, line 1: This is a small logic adjustment that should have always been in place

Ridhwana (approved):

app/assets/stylesheets/article-show.scss
```diff
@@ -41,6 +41,11 @@ article {
   }
 }
 
+#bottom-content-read-next {
+  content-visibility: auto;
+  contain-intrinsic-size: 450px;
```
vaidehijoshi, line 46: Interesting, this is a new one for me. How did we decide on/determine on the contain-intrinsic-size pixel value (not just here but across this PR?)

benhalpern: I just eyeballed it and picked a number based on what I think the average space the titles might take up. Anything that includes user-generated content can't be exact, but this value is just designed to be an estimate. I figured this area (four titles) won't have as much variation as, say, a comment section where each comment could be _any_ height.

nickytonline: I imagine we need a media query with a different `contain-intrinsic-size` for smaller screens/devices? Same for all the places below where we use `contain-intrinsic-size`.

nickytonline (commented):
Just a note that this is good for Chrome 85+ (Chromium Edge too), but as far as I know, it's [not implemented in Firefox](https://github.com/mozilla/standards-positions/issues/135) yet. Regardless, a subset of our users (probably majority) will benefit from this.

More here, https://web.dev/content-visibility/#support

app/assets/stylesheets/article-show.scss
vaidehijoshi: I just looked at this `<section>` on DEV and I worry that the hardcoded pixel value doesn't match what the element will actually look like:
Screen Shot 2020-11-23 at 9 10 02 AM

+1 to @nickytonline's point of accounting for the different sizes at various screen sizes here.

benhalpern: My interpretation of how to use `contain-intrinsic-size` is that anything in the ballpark of the ultimate size is fine, so loose estimates are accurate enough. We wouldn't want a _radically_ bad guess (aka comment height could be off by an order of magnitude) but if we're giving the browser a hint it will still be an additive feature.

It's hard to get a 100% obvious answer to this question, but I think that's the case. And if it is the case, the divs I chose happen to be losely similar heights regardless of viewport width.

The word "estimate" is generally used everywhere alongside this feature (though definitely not enough written about this for us to have 100% confidence)

From: https://www.w3.org/TR/css-sizing-4/#cis-scrollbars

> The contain-intrinsic-size property provides an estimate of how large the author expects the content of an element to be, but this estimate is not actual content and does not represent anything that needs to be shown to the user.

Let me know if my interpretation seems offbase.

nickytonline: Based on what you're saying it sounds fine. I've never used this feature so we can always try it out.

nickytonline (approved):
Let's try it out. 🚀

app/assets/stylesheets/article-show.scss
vaidehijoshi: Ok, that's helpful—the additional link you provided makes it seem like it's safe enough to have a loose estimate here. If we notice any weird behavior because of this, we can always change it.

vaidehijoshi (approved):
It would be helpful to include a comment next to the estimated pixel values added in this PR so that people looking at the css files in the future know that we intentionally picked a loose estimate here due to the property's nature (and don't have to look up how the property works in order to get that context).

marcysutton:
Interesting PR and technology! There are some accessibility concerns from my initial testing, but I'll need to seed a local Forem instance to inspect it more (any quick tips for this?). In short, headings and other content will be impacted by this feature which suppresses them from screen readers if considered offscreen. That means content will be hidden on load from the Elements List in NVDA and rotor in Voiceover, impacting the overall page outline and summary read aloud.

I did a little test on one of the [demos from Web.dev](https://codepen.io/marcysutton/pen/ZEpzped) and found `content-visibility` does impact screen readers.

To test this, I moved an `h2` in and out of a DIV with `content-visibility: auto` applied. When the heading is outside of the DIV like they set it up in the demo, it gets rendered. When inside of the region it gets suppressed (as expected):
Codepen demo showing HTML and CSS

Here's a screenshot of NVDA's Elements List with the "Make sure you're stocked up on snacks" h2 suppressed and therefore not listed on load:
Chrome with NVDA's elements list running, omitting a heading

Here it is once more, with the h2 rendered on load:
Chrome with NVDA's elements list running, showing the heading

If `content-visibility` could be applied without headings included it wouldn't impact the page outline as much. But that seems difficult to achieve in my opinion...so for the gains this feature gives you, it also takes some access away.

nickytonline:
> Interesting PR and technology! There are some accessibility concerns from my initial testing, but I'll need to seed a local Forem instance to inspect it more (any quick tips for this?). In short, headings and other content will be impacted by this feature which suppresses them from screen readers if considered offscreen. That means content will be hidden on load from the Elements List in NVDA and rotor in Voiceover, impacting the overall page outline and summary read aloud.
> 
> I did a little test on one of the [demos from Web.dev](https://codepen.io/marcysutton/pen/ZEpzped) and found `content-visibility` does impact screen readers.
> 
> To test this, I moved an `h2` in and out of a DIV with `content-visibility: auto` applied. When the heading is outside of the DIV like they set it up in the demo, it gets rendered. When inside of the region it gets suppressed (as expected):
> Codepen demo showing HTML and CSS
> 
> Here's a screenshot of NVDA's Elements List with the "Make sure you're stocked up on snacks" h2 suppressed and therefore not listed on load:
> Chrome with NVDA's elements list running, omitting a heading
> 
> Here it is once more, with the h2 rendered on load:
> Chrome with NVDA's elements list running, showing the heading
> 
> If `content-visibility` could be applied without headings included it wouldn't impact the page outline as much. But that seems difficult to achieve in my opinion...so for the gains this feature gives you, it also takes some access away.

I'd say if it's affecting accessibility, as Marcy points out, to put this on ice for now.

benhalpern:
Yup, @marcysutton this definitely isn't important enough to sacrifice a11y. Thanks for weighing in.

My only thought here is that maybe we could keep it in `footer`.... Because that is the most straightforward case anyway... It exists on basically every page and is always offscreen, and will generally be consistent with height.... And it includes no headings. So in terms of bang for the buck without sacrifice that would be the only place for it.

How do you feel about stripping everything out, but keeping the footer?

marcysutton:
> How do you feel about stripping everything out, but keeping the footer?

If you applied `content-visibility` to the contents of `footer` and not the element itself, that could keep a landmark element present for screen readers to act as navigation. Seems like a worthy experiment since there aren't currently any headings in the footer (although I'd argue there probably should be).

benhalpern:
@marcysutton I removed the non-footer uses of this and attached it to the inner container instead of the `footer`. How's it looking?

marcysutton:
I've tested it a bit locally, but I could really use more seed data. 

On the live site though, from what I can tell assistive tech users can't easily get to the footer landmark content anyway due to the infinite scroll. So whether this feature suppresses content access is kind of moot on those pages 🤷🏻‍♀️ Infinite scrolling could be a whole other conversation.

The accessibility impact is very low with all the other dynamic content loading going on in the DOM, so I'd say ship it as an experiment and you could always make changes with more information.

Here are some more details I found:

- Chrome 87 isn't exposing the footer ("contentinfo") landmark in Voiceover Rotor to begin with, but Safari is used more often as a combo. Generally, adding `role="contentinfo"` to the footer in HTML would expose it in Chrome regardless of its contents' visibility.
- The Voiceover Rotor shows infinite article tags in the landmark inspector, so you can't ever get to the footer if it follows those elements in the DOM. TIL the rotor updates with DOM changes! Same with NVDA's Elements List.
- I want to test `content-visibility` locally in NVDA on Windows too, but I'm having a login issue on my virtual machine–I'll have to come back to that part.

PR #11581, opened 2020-11-23
[EOY 2020] Replace "ban" with "suspend" in Admin

msarit:

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Renames "ban" to "suspend" throughout the `/admin/` space.
closes https://github.com/forem/InternalProjectPlanning/issues/298

#### NOTE: 👇🏽
I stopped short of changing `banned` and `comment_banned` to `suspended` and `comment_suspended` (respectively) within `app > models > role.rb`, as I am not sure how that move will affect current role assignments in production. I welcome feedback/corrections on my decision. :pray:

1. Navigate to `localhost:xxxx/admin/users` as a `super_admin` role
2. Edit any user and give them either `Suspend` or `Comment Suspend` roles
3. Check that their show page reflects the role changes made

https://www.loom.com/share/6c764ef0ff724b8e8c73433681a2edf1

Changes were made just to certain text, depending on the user's role; there are no UI accessibility concerns introduced, in my humble opinion

## Added tests?

- [X] Yes, tests were modified according to code changes made
- [ ] No, and this is why:

## Added to documentation?

- [X] [Developer Docs](https://docs.forem.com) and/or [Admin Guide](https://forem.gitbook.io/forem-admin-guide/)
https://app.gitbook.com/@forem/s/forem-admin-guide/admin/users/user-roles
- [ ] README
- [ ] No documentation needed

None

Get It Dance

### DIFF
--- a/app/views/admin/users/_activity.html.erb
+++ b/app/views/admin/users/_activity.html.erb
@@ -15,7 +15,7 @@
     <p>(view full history in notes below)</p>
     <ul>
       <% @user.roles.each do |role| %>
-        <% if role.name === "banned" %>
+        <% if role.name == "banned" %>
           <li>suspended</li>
         <% else %>
           <li><%= role.name %> <em><%= role.resource_name.to_s %></em></li>
--- a/app/views/admin/users/_activity.html.erb
+++ b/app/views/admin/users/_activity.html.erb
@@ -15,7 +15,11 @@
     <p>(view full history in notes below)</p>
     <ul>
       <% @user.roles.each do |role| %>
-        <li><%= role.name %> <em><%= role.resource_name.to_s %></em></li>
+        <% if role.name === "banned" %>
+          <li>suspended</li>
+        <% else %>
+          <li><%= role.name %> <em><%= role.resource_name.to_s %></em></li>
+        <% end %>
       <% end %>
     </ul>
   </div>
--- a/app/controllers/chat_channels_controller.rb
+++ b/app/controllers/chat_channels_controller.rb
@@ -103,7 +103,7 @@ class ChatChannelsController < ApplicationController
       else
         render json: {
           status: "error",
-          message: "Ban failed. user with username '#{username}' not found in this channel."
+          message: "Suspend failed. user with username '#{username}' not found in this channel."
         }, status: :bad_request
       end
     when "/unban"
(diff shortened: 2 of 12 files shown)


citizen428:
> I stopped short of changing banned and comment_banned to suspended and comment_suspended (respectively) within app > models > role.rb, as I am not sure how that move will affect current role assignments in production.

What ties users and roles together via `role_id` and `user_id` is the `users_roles` join table, so renaming a role should theoretically work. I just did the following locally:

1. Started a Rails console, added the `banned` role to a user:
    ```ruby
    user.add_role(:banned)
    user.roles_name
    #=> ["banned"]
    ```
1. Change `banned` to `suspended` in `role.rb` (the constant there is only used for validation).
1. Load a Ralls console, update the role name:
    ```ruby
    reload!
    Role.find_by(name: "banned").update(name: "suspended")
    #=> true
1. Reload the user and check the roles:
    ```ruby
    user.reload.role_names
    #=> ["suspended"]

To update this in production the best way would be to write a data update script that renames the two roles, e.g. 

```ruby
Role.find_by(name: "banned").update(name: "banned")
Role.find_by(name: "comment_banned").update(name: "comment_suspended")
```

You would also need to update the following method in `user.rb`:

```ruby
  # methods for Administrate field
  def banned
    has_role? :banned
  end
```

And finally, check for usages of `.banned` or `has_role(:banned)` throughout the codebase.

I would say this can be done in a follow-up PR.

app/views/admin/users/_activity.html.erb
```diff
@@ -15,7 +15,11 @@
     <p>(view full history in notes below)</p>
     <ul>
       <% @user.roles.each do |role| %>
-        <li><%= role.name %> <em><%= role.resource_name.to_s %></em></li>
+        <% if role.name === "banned" %>
```
citizen428, line 18: **suggestion (blocking):** I think a bit of JavaScript snuck in here:

```suggestion
        <% if role.name == "banned" %>
```

While they theoretically both work, `===` in Ruby is used for [case equality](https://blog.arkency.com/the-equals-equals-equals-case-equality-operator-in-ruby/), which is not quite what we want here.

msarit: Thank you so much 🤦‍♀️

msarit:
FYI: Michael's comments [here](https://github.com/forem/forem/pull/11581#issuecomment-732534235) will be addressed in a separate PR 👍

joshpuetz (approved):

benhalpern (approved):

citizen428 (approved):

Issue #11455, opened 2020-11-17
Allow canonical "welcome" thread to be authored by any admin (or mascot)


pkfrank:
**Is your feature request related to a problem? Please describe.**
As a Forem Admin, I'd like more flexibility in setting up the canonical "Welcome" thread (that users get redirected to via `/welcome`).  Currently, that thread must be authored by the "Staff User" set in `Admin > Config > Community Content`.

**Describe the solution you'd like**
I'd like the following user "types" to be eligible authors of the thread:
- Any admin (or superadmin)
- The site mascot

**Additional context**
I _believe_ the relevant code is here:
- [`/app/models/article.rb`](https://github.com/forem/forem/blob/master/app/models/article.rb)
- [`/app/controllers/pages_controller.rb`](https://github.com/forem/forem/blob/master/app/controllers/pages_controller.rb)

I'll ensure that the [Admin Documentation](https://forem.gitbook.io/forem-admin-guide/) is improved as well.

vaidehijoshi:
_Cross-posting from [forem.team](https://forem.team/vaidehijoshi/comment/15c) to here to maintain conversational parity:_

Stepping back for a moment, do we know if other Forem creators use the welcome thread currently? And if they do, how they use it?

I've always felt that the daily threads (like the welcome thread) to be DEV-centric, and I wonder what exactly the problem is that we are solving for. Do we want to establish an idea of all Forems having a "welcome" thread and make that a Forem-wide feature? Or do our Forem creators actually need something else to help them do the welcoming?

I do think we need to move away from the strict usage of the staff user <> welcome thread concept, but I want to raise the larger question of these daily thread concepts since they're coming up naturally in conversation.

benhalpern:
Currently the welcome thread is widely used (it sort of has to be because we always have it in the onboarding).

<img width="671" alt="Screen Shot 2020-11-24 at 10 27 10 AM" src="https://user-images.githubusercontent.com/3102842/100114624-9dc9d680-2e3f-11eb-91bf-97e38082375a.png">

Example: https://the.community.club/community-club/welcome-thread-v2-57oj

mstruve:
@pkfrank it looks like the attached PR will solve this issue. It needs to have the specs fixed but otherwise, it should do what you want. If @benhalpern wants to cleanup the specs and push it through that would be great otherwise I can try to take a whack at it later when I am online.

PR #11597, opened 2020-11-24
Allow any admin to publish welcome thread

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Feedback from Forem creators is that it is unintuitive to restrict the welcome thread to the "staff id" and any admin should be allowed to author a welcome thread.

https://github.com/forem/forem/issues/11455

## Added tests?

- [x] Yes - Adjusted.
      have not been included_

## Added to documentation?

- [ ] README
- [ ] No documentation needed

...Will add docs to admin guide if we merge this.


app/models/article.rb
```diff
@@ -112,7 +112,7 @@ class Article < ApplicationRecord
 
   scope :admin_published_with, lambda { |tag_name|
     published
-      .where(user_id: SiteConfig.staff_user_id)
+      .where(user_id: User.with_role(:super_admin).pluck(:id) + User.with_role(:admin).pluck(:id))
```
benhalpern, line 115: Is there a more efficient query we could use here?

rhymes: This will result in a single query:

```suggestion
      .where(user_id: User.with_role(:super_admin).union(User.with_role(:admin)).select(:id))
```

according to Blazer, with this change, this should take around 210 ms.

*BUT*

with separate queries it should take less than 10 ms (I ran `EXPLAIN ANALYZE` on each separate step and added them together). 

That's because it can use indexes it's not using with the subquery. So I think your option is the most efficient one even if it results in 3 queries

rhymes (approved):
LGTM (left an explanation on why the current query is the best option)

**update** we could probably replace those two queries with `User.with_any_role(:admin, :super_admin).pluck(:id)` which I just found out about :D

app/models/article.rb
```diff
@@ -112,7 +112,7 @@ class Article < ApplicationRecord
 
   scope :admin_published_with, lambda { |tag_name|
     published
-      .where(user_id: SiteConfig.staff_user_id)
+      .where(user_id: User.with_role(:super_admin).union(User.with_role(:admin)).select(:id))
```
mstruve, line 115: Is the mascot guaranteed to be an admin? Want to make sure this won't break any Forems where the mascot might not have admin privileges. @pkfrank you might be able to answer this as well.

mstruve (commented):
Fixed the specs and updated with master. One question to make sure this won't break any Forems.

app/models/article.rb
pkfrank: @mstruve the Mascot does not appear to have admin privileges out of the gate (checking via `/admin/users` on a fresh Forem from yesterday).  Perhaps @vaidehijoshi or @atsmith813 can chime in on whether we'd want to add that to the Mascot?

atsmith813: If we want to add it, it can go here:

https://github.com/forem/forem/blob/4dc1a5375519221f0bf60e210e79762d580f292f/app/services/users/create_mascot_account.rb#L17-L22

with `mascot.add_role(:super_admin)`

mstruve: Whew, thanks for the insight @atsmith813. Rather than add it there and make a DataUpdateScript for all forems, I will update the code to check for that ID as well which actually should make the test changes then unnecessary.

rhymes (approved):
LGTM!

spec/models/article_spec.rb
```diff
@@ -53,6 +53,26 @@ def build_and_validate_article(*args)
 
     it { is_expected.not_to allow_value("foo").for(:main_image_background_hex_color) }
 
+    describe "::admin_published_with" do
+      it "includes mascot published articles" do
+        user = create(:user, id: 1)
```
atsmith813, line 58: The mascot won't always have an ID of 1. I believe I'm misunderstanding what we need here! Should there be logic to check `SiteConfig.mascot_user_id` both here and in the scope?

app/models/article.rb
```diff
@@ -112,9 +112,10 @@ class Article < ApplicationRecord
 
   scope :admin_published_with, lambda { |tag_name|
     published
-      .where(user_id: SiteConfig.staff_user_id)
-      .order(published_at: :desc)
-      .tagged_with(tag_name)
+      .where(user_id: User.with_role(:super_admin)
+                          .union(User.with_role(:admin))
+                          .union(id: SiteConfig.staff_user_id)
```
nickytonline, line 117: Is this line included because the staff user is not necessarily someone with an admin role yet? If they are an admin out of the box, I imagine this line could be omitted?

mstruve: > Is this line included because the staff user is not necessarily someone with an admin role yet?

Correct. It was a much bigger change to make the staff user an admin since we would have to update the code and then we would have to run a data update script to update all Forems so I went with this solution instead.

spec/models/article_spec.rb
benhalpern: Yup, this is correct. I fixed this test and overall logic.

app/models/article.rb
```diff
@@ -112,9 +112,10 @@ class Article < ApplicationRecord
 
   scope :admin_published_with, lambda { |tag_name|
     published
-      .where(user_id: SiteConfig.staff_user_id)
-      .order(published_at: :desc)
-      .tagged_with(tag_name)
+      .where(user_id: User.with_role(:super_admin)
+                          .union(User.with_role(:admin))
```
Zhao-Andy, line 116: (non-blocking) Maybe refactor to:
```ruby
User.with_any_role(:super_admin, :admin)
```

Zhao-Andy (approved):

PR #11599, opened 2020-11-24
Bug Fix:Remove Score Filter From Tag Mod Index

mstruve:
- [x] Bug Fix

## Description
Remove score filter for articles for the moderator's article index so they can view all articles. 

Closes https://github.com/forem/tech-private/issues/420

## Added tests?
- [x] No, they already exist

alt_text


Zhao-Andy (approved):
Ah, not sure why this was there before. Maybe @benhalpern has some context?

rhymes (approved):

PR #11605, opened 2020-11-24
Don't do serviceworker asset caching in development

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We recently added caching of core assets into the serviceworker cache in addition to the general in-memory cache because we have more control over eviction, etc.

However, since locally we do not _fingerprint_ assets, we can't rely on that as a cachebusting mechanism like we can in prod. This makes that that code will not run in development. This is a `.erb` file so we can access Rails where appropriate.

This is kind of a hot fix, but we could re-evaluate for a cleaner approach in the future.

We do not need to adjust other cache eviction logic because if the cache is not accumulating, the eviction logic should be fine if the cache is empty.

### DIFF
--- a/app/views/service_worker/index.js.erb
+++ b/app/views/service_worker/index.js.erb
@@ -137,22 +137,27 @@
       return;
     }
 
-    // Fetch and/or JS and CSS assets for better persistence than memory cache.
-    if (url.pathname.startsWith("/assets/") || url.pathname.startsWith("/packs/")) {
-      event.respondWith(
-        caches.open(staticCacheName).then(function (cache) {
-          return cache.match(event.request).then(function (response) {
-            return (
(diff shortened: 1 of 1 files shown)


nickytonline (approved):
Looks good. Thanks Ben! 🚀

mstruve (approved):
Can we make an Issue outlining basically what is in this PR that asks how we can do this better so development can match production? I want to make sure we keep track of these one-offs so we can work to minimize them. But I am game for this quick fix to make our dev lives easier as well as contributors.

benhalpern:
@mstruve made this issue... https://github.com/forem/forem/issues/11606

PR #11609, opened 2020-11-25
Move HtmlCleaner from app/labor into service

citizen428:

- [X] Refactor

## Description

1. Move `app/labor/html_cleaner.rb` to `app/services/clean_html.rb`
2. Minor code changes/cleanup in `app/services/clean_html.rb`
3. Add `spec/services/clean_html_spec.rb`

## Added tests?

- [X] Yes

## Added to documentation?

- [X] No documentation needed


app/services/clean_html.rb
```diff
@@ -0,0 +1,23 @@
+class CleanHtml
```
rhymes, line 1: could you move this to Feeds::CleanHtml or something like that? I want to eventually refactor `RssReader::Assembler` into something more self explanatory and with less responsibilities and since this is part of that machinery, I wonder if we should start by namespacing this correctly

rhymes (commented):
Nice! Left a comment about namespacing

maestromac (approved):
I agree with what @rhymes mentioned but if the goal is to move off `/labor` then this is LGTM.

app/services/clean_html.rb
citizen428: Yes, that makes sense. My main concern so far was moving things out of `app/labor`, but we might as well properly sort them while we're at it.

citizen428:
@rhymes @maestromac I moved it to the `Feeds::` namespace now.

rhymes (approved):

PR #11615, opened 2020-11-25
Hotfix: Fix Settings/Customization page

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Currently https://dev.to/settings/customization is not working because of nested forms.

This PR removes the nested forms, keeps the floating "Save" button removing the interim ones and restructures "update language settings" that cannot work inside the `Customization` tab as it is but it should be part of the `update` user's action anyway.

1. Load the PR
1. Go to http://localhost:3000/settings/customization
1. Save stuff, especially check if multiple language settings stick after save

## Added tests?

- [x] Yes
      have not been included_

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,9 +1,7 @@
 class UsersController < ApplicationController
   before_action :set_no_cache_header
   before_action :raise_suspended, only: %i[update]
-  before_action :set_user, only: %i[
-    update update_language_settings confirm_destroy request_destroy full_delete remove_identity
-  ]
+  before_action :set_user, only: %i[update confirm_destroy request_destroy full_delete remove_identity]
   after_action :verify_authorized, except: %i[index signout_confirm add_org_admin remove_org_admin remove_from_org]
   before_action :authenticate_user!, only: %i[onboarding_update onboarding_checkbox_update]
   before_action :set_suggested_users, only: %i[index]
@@ -39,7 +37,12 @@ class UsersController < ApplicationController
   def update
     set_current_tab(params["user"]["tab"])
 
-    if @user.update(permitted_attributes(@user))
+    # preferred_languages is handled manually
+    @user.language_settings["preferred_languages"] = Languages::LIST.keys & params[:user][:preferred_languages].to_a
+
+    @user.attributes = permitted_attributes(@user)
+
+    if @user.save
       # NOTE: [@rhymes] this queues a job to fetch the feed each time the profile is updated, regardless if the user
       # explicitly requested "Feed fetch now" or simply updated any other field
       import_articles_from_feed(@user)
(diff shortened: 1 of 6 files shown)


app/controllers/users_controller.rb
```diff
@@ -69,18 +72,6 @@ def update
     end
   end
 
-  def update_language_settings
```
rhymes, line 72: we don't really need to handle this separately, now that in the IA it's in the same page as the rest of the Customization section

app/views/users/_customization.html.erb
```diff
@@ -78,20 +80,15 @@
       </p>
     </header>
 
-    <%= form_for(@user, html: { id: nil, class: "grid gap-4" }) do |f| %>
```
rhymes, line 81: removing nested forms that didn't work

app/views/users/_customization.html.erb
```diff
@@ -105,20 +102,13 @@
...
```
rhymes, line 108: removing nested forms that didn't work

app/views/users/_language_settings.html.erb
```diff
@@ -9,17 +9,18 @@
...
     </h2>
     <p>Select which languages you'd prefer to see in your feed.</p>
-    <p>This setting controls which languages you are more likely to see throughout the site, but you may still see other languages, especially English.</p>
+    <p>
+      This setting controls which languages you are more likely to see throughout the site,
+      but you may still see other languages, especially English.
+    </p>
   </header>
 
-  <%= form_tag users_update_language_settings_path, class: "grid gap-4"  do |f| %>
```
rhymes, line 15: removing nested forms that didn't work

config/routes.rb
```diff
@@ -369,7 +369,6 @@
     get "/async_info/shell_version", controller: "async_info#shell_version", defaults: { format: :json }
 
     # Settings
-    post "users/update_language_settings" => "users#update_language_settings"
```
rhymes, line 372: bye bye

Ridhwana (approved):
This looks good to me @rhymes.

app/views/users/_customization.html.erb
nickytonline: refactor (potentially blocking): I haven't pulled this locally yet, but most likely the nested form will need to be replaced with another element so that the CSS classes `grid gap-4` remain for layout. A [\<fieldset /\>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) should do the trick.

app/views/users/_customization.html.erb
nickytonline: Same comment as above.

app/views/users/_language_settings.html.erb
nickytonline: Same comment as above.

benhalpern (approved):

app/views/users/_customization.html.erb
rhymes: Thanks @nickytonline. I merged it to unblock the functionality. I'll segue with a `<fieldset>`, thanks a lot for the review!

Issue #11529, opened 2020-11-21
Generalize profile creation in the onboarding process


Ridhwana:
We currently have Profile Generalization implemented as a feature flag. In order to bring this feature to completion, we need to implement the profile generalization across multiple areas. This includes the profile creation in the onboarding process which is currently hardcoded.

Ridhwana:
I've started putting together a plan:

The plan:

- Refresh my memory on how Onboarding works.
- Pull the profile field data to the frontend via an api call. The endpoint will display the onboarding scope from the profile field group model.
- Use the new data to render the interface dynamically.
- This will be the MVP. Once I've got this working I'll pull in @pawel to see if we can make optimizations to the UI/UX.
- Update the profile fields through an endpoint once a user hits submit.

PR #11629, opened 2020-11-26
Profile generalization for the onboarding workflow

Ridhwana:
## NOTE

**There's a [pre deployment task](https://github.com/forem/forem/pull/11704) that will need to be run for existing Forems before this PR can be merged.**

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
We currently have Profile Generalization implemented as a feature flag. In order to bring this feature to completion, we need to implement the profile generalization across multiple areas. This includes the profile creation in the onboarding process which is currently hardcoded.

A related Slack thread for anyone with access that's interested in some the decisions: https://dev-internal.slack.com/archives/C0186BE65TN/p1606752839018000

https://github.com/forem/forem/issues/11529

- Enable the profile generalization feature flag. You can follow the instructions [here](https://github.com/forem/forem/pull/11149) or just open up a rails console and run `Flipper.enable(:profile_admin)`
- Go http://localhost:3000/admin/profile_fields
- Configure the Profile fields and tick `show in onboarding` for some of the fields. 
 - The fields that you've ticked should now show up in "Step 2: Build your profile" of the onboarding section http://localhost:3000/onboarding.
- Fill out the fields in this step and make sure they persist by checking the user profile page http://localhost:3000/settings/profile or just looking at the profile information in the console.

If you're not familiar with the feature, you may stumble on replicating this so if anyone needs more verbose instructions, please feel free to ping me and I can make a video or provide more in depth instructions and add it to the PR.

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

## [optional] Are there any **pre deployment** tasks we need to perform?
https://github.com/forem/forem/pull/11704 needs to first be merged and run.

alt_text


app/javascript/onboarding/components/ProfileForm.jsx
```diff
@@ -54,13 +59,125 @@ class ProfileForm extends Component {
...
+          id={field.attribute_name}
+          onChange={this.handleChange}
+        />
+        <label class="crayons-field__label" htmlFor={field.attribute_name}>
+          {field.label}
+        </label>
+        {field.description && (
+          <p class="crayons-field__description">{field.description}</p>
+        )}
+      </div>
```
nickytonline, line 109: refactor (blocking): Use the [FormField](https://storybook.forem.com/?path=/story/3-components-form-components-form-field--radio-variant) design system component instead.

```suggestion
      <FormField variant="checkbox">
        <input
          class="crayons-checkbox"
          type="checkbox"
          name={field.attribute_name}
          id={field.attribute_name}
          onChange={this.handleChange}
        />
        <label class="crayons-field__label" htmlFor={field.attribute_name}>
          {field.label}
        </label>
        {field.description && (
          <p class="crayons-field__description">{field.description}</p>
        )}
      </FormField>
```

app/javascript/onboarding/components/ProfileForm.jsx
```diff
@@ -10,22 +10,24 @@ class ProfileForm extends Component {
...
-      last_onboarding_page: 'v2: personal info form',
+      groups: [],
+      formValues: {},
       canSkip: true,
+      last_onboarding_page: 'v2: personal info form',
     };
   }
 
   componentDidMount() {
+    fetch('/profile_field_groups?onboarding=true')
```
nickytonline, line 25: refactor (blocking): Use the [request](https://github.com/forem/forem/blob/master/app/javascript/utilities/http/request.js#L23) utility function instead via `import { request } from '@utilities/http`. We added the `request` function a little while back after we saw that folks were doing the same things over and over.

With the `request` function, there is no need to pass the headers that are currently specified as they are default headers baked in. As well, there is no need to do a `JSON.stringify` for the body. This is handled within `request`.

app/javascript/onboarding/components/ProfileForm.jsx
```diff
@@ -54,13 +59,125 @@ class ProfileForm extends Component {
...
+          onChange={this.handleChange}
+        />
+        {field.description && (
+          <p class="crayons-field__description">{field.description}</p>
+        )}
+      </div>
+    );
+  }
+
+  colorField(field) {
```
nickytonline, line 153: refactor (blocking): This can be extracted into it's own component. It does not depend on any state of the `<ProfileForm />` component. It does however depend on `this` for `this.handleColorPickerChange`, but this can be passed in as a prop alongside the current `field` parameter, e.g.

`<ColorField field={field} onColorChange={this.handleColorPickerChange} />`


Since this is only used in this file, you can keep the component in the same file, or if you want you can put it in its own separate file.

app/javascript/onboarding/components/ProfileForm.jsx
```diff
@@ -54,13 +59,125 @@ class ProfileForm extends Component {
...
+          onChange={this.handleChange}
+        />
+        {field.description && (
+          <p class="crayons-field__description">{field.description}</p>
+        )}
+      </div>
+    );
+  }
+
+  textAreaField(field) {
```
nickytonline, line 133: refactor (blocking): This can be extracted into it's own component. It does not depend on any state of the `<ProfileForm />` component. `field` can be passed in as a prop, e.g.

`<TextField field={field}  />`

Since this is only used in this file, you can keep the component in the same file, or if you want you can put it in its own separate file.

app/javascript/onboarding/components/ProfileForm.jsx
```diff
@@ -71,7 +188,32 @@ class ProfileForm extends Component {
...
+                  ? this.colorField(field)
+                  : field.input_type === 'text_area'
+                  ? this.textAreaField(field)
+                  : this.textField(field);
+              })}
+            </div>
+          </div>
+        );
+      });
+    }
```
nickytonline, line 216: refactor (non-blocking): The conditional check on groups is not necessary if you destructure `groups` from `this.state` with a default value of `[]`.

```suggestion
    const { canSkip, groups = [] } = this.state;
    const sections = groups.map((group) => {
      return (
        <div class="onboarding-profile-sub-section">
          <h2>{group.name}</h2>
          {group.description && (
              <div class="color-base-60">{group.description})</div>
          )}
          <div>
            {group.profile_fields.map((field) => {
              return field.input_type === 'check_box'
                ? this.checkboxField(field)
                : field.input_type === 'color_field'
                ? this.colorField(field)
                : field.input_type === 'text_area'
                ? this.textAreaField(field)
                : this.textField(field);
            })}
          </div>
        </div>
        );
      });
```

app/javascript/onboarding/components/ProfileForm.jsx
```diff
@@ -71,7 +188,32 @@ class ProfileForm extends Component {
...
+      sections = groups.map((group) => {
+        return (
+          <div class="onboarding-profile-sub-section">
+            <h2>{group.name}</h2>
+            {group.description && (
+              <div class="color-base-60">{group.description})</div>
+            )}
+            <div>
+              {group.profile_fields.map((field) => {
+                return field.input_type === 'check_box'
```
nickytonline, line 204: refactor (maybe blocking): It's a little hard to read the nested ternaries. Consider a switch case in a function and call that function with the `field.input_type`.

nickytonline (changes requested):
Some comments and a few request changes.

Ridhwana:
> Some comments and a few request changes.
Thanks for the verbose feedback Nick! I'll implement these tomorrow morning 😊

Ridhwana:
Oops sorry this was a mistake ☝️

app/controllers/users_controller.rb
```diff
@@ -143,13 +143,24 @@ def remove_identity
   def onboarding_update
     if params[:user]
       sanitize_user_params
-      permitted_params = %i[summary location employment_title employer_name last_onboarding_page]
-      current_user.assign_attributes(params[:user].permit(permitted_params))
+      permitted_user_params = %i[last_onboarding_page]
```
citizen428, line 146: **suggestion (non-blocking):** As always, I'd feel slightly better if we moved this out of the method and into a constant.

app/controllers/users_controller.rb
```diff
@@ -143,13 +143,24 @@ def remove_identity
...
+      current_user.assign_attributes(params[:user].permit(permitted_user_params))
       current_user.profile_updated_at = Time.current
+      if current_user.save
+        success = true
```
citizen428, line 150: I think we don't need this local variable. You can rewrite the next conditional instead and directly use the value of `update_result.success?`:

```ruby
    if current_user.save && params[:profile]
      updated_result = Profiles::Update.call(current_user, { profile: profile_params })
    end
    
    current_user.saw_onboarding = true
    authorize User
    render_update_response(update_result.success?)
```

Note 1: This behaves differently from your original code. In your code a failed call to `Profiles::Update` would not have resulted in `success` being `false`. I think the way it should have looked originally is

```ruby
    if params[:profile] && success == true
      update_result = Profiles::Update.call(current_user, { profile: params[:profile].permit(Profile.attributes!) })

      success = update_result.success?
```

The way the code currently is it would have set `success` to true again (which it already was if we are inside this branch due to the outer conditional), but never to `false`. Please let me know if I misread something.

Note 2: I also added a `profile_params` method call here and would encourage extracting it for readability instead of keeping `params[:profile].permit(Profile.attributes!)` inline.

app/controllers/users_controller.rb
```diff
@@ -162,7 +173,10 @@ def onboarding_checkbox_update
 
     current_user.saw_onboarding = true
     authorize User
-    render_update_response
+    if current_user.save
+      success = true
```
citizen428, line 177: I don't think we gain much from this variable, I'd inline this:

```ruby
     render_update_response(current_user.save)
```

app/controllers/users_controller.rb
```diff
@@ -267,8 +281,8 @@ def determine_follow_suggestions(current_user)
     recent_suggestions.presence || default_suggested_users
   end
 
-  def render_update_response
-    outcome = current_user.save ? "updated successfully" : "update failed"
+  def render_update_response(success)
+    outcome = success == true ? "updated successfully" : "update failed"
```
citizen428, line 285: It's relatively uncommon in Ruby to explicitly check for booleans. Most conditionals rely on truthiness (`nil`/`false` vs. everything else) instead.

app/javascript/onboarding/components/ProfileForm.jsx
```diff
@@ -169,7 +266,6 @@ ProfileForm.propTypes = {
   currentSlideIndex: PropTypes.func.isRequired,
   communityConfig: PropTypes.shape({
     communityName: PropTypes.string.isRequired,
-    communityDescription: PropTypes.string.isRequired,
```
vaidehijoshi, line 172: Curious why this was removed from the required `propTypes` (but not anything else)?

spec/requests/users_onboarding_spec.rb
```diff
@@ -13,17 +13,17 @@
       end
 
       it "updates the attributes on the user" do
```
vaidehijoshi, line 15:
```suggestion
      it "updates the user's last_onboarding_page attribute" do
```
Since this test doesn't actually update any _other_ user attributes anymore (like location) because those are all extracted out of the user and onto the profile fields, we may want to consider changing this.

spec/requests/users_onboarding_spec.rb
```diff
@@ -13,17 +13,17 @@
...
-        params = { user: { location: "Alpaca Town" } }
+        params = { user: { last_onboarding_page: "v2: personal info form" } }
         expect do
           patch "/onboarding_update.json", params: params
-        end.to change(user, :location)
+        end.to change(user, :last_onboarding_page)
       end
 
       it "does not update attributes if params are empty" do
```
vaidehijoshi, line 22:
```suggestion
      it "does not update the user's last_onboarding_page if it is empty" do
```

app/javascript/onboarding/components/ProfileForm.jsx
Ridhwana: ✅

app/javascript/onboarding/components/ProfileForm.jsx
Ridhwana: @nickytonline I'm happy to do this, but I was curious why you'd put it in its own component (instead of just keeping it in a function) since it's not being reused by any other components.

app/javascript/onboarding/components/ProfileForm.jsx
Ridhwana: @vaidehijoshi I assume that communityDescription was at some point being used in the Profile Form, but I noticed that there was no reference to it any longer (we only use communityName), hence I removed it.

app/controllers/users_controller.rb
Ridhwana: ✅

app/controllers/users_controller.rb
Ridhwana: @citizen428 thank you for taking the time to do this write up. Its been really useful and I've refactored accordingly.

app/controllers/users_controller.rb
Ridhwana: Definitely!

app/controllers/users_controller.rb
Ridhwana: ✅

app/javascript/onboarding/components/ProfileForm.jsx
Ridhwana: Thanks @nickytonline I missed a lot since last working on the frontend 😁. I've moved the code over, this is nicer!

app/javascript/onboarding/components/ProfileForm.jsx
Ridhwana: ✅

app/javascript/onboarding/components/ProfileForm.jsx
Ridhwana: ✅

app/javascript/onboarding/components/ProfileForm.jsx
Ridhwana: ✅

app/javascript/onboarding/components/ProfileForm.jsx
Ridhwana: ✅

Ridhwana:
@nickytonline, @citizen428 and @vaidehijoshi thank you all for the reviews, it was very useful feedback.🌸
I've resolved all the comments on this PR.

app/controllers/users_controller.rb
Ridhwana:
```
    if current_user.save && params[:profile]
      update_result = Profiles::Update.call(current_user, { profile: profile_params })
    end
    
    current_user.saw_onboarding = true
    authorize User
    render_update_response(update_result.success?)
```
Just hit a snag and realized that if `current_user.save && params[:profile]` evaluates to false then update_result will not be defined 3 lines down.I didnt remember at the time, but that was the reason I may have defined the success var. I changed this to be  `render_update_response(update_result&.success?)`.

citizen428: Ah, yes, correct @Ridhwana. I didn't try out this code, I just wrote it directly in here, but `&.` seems like an ok solution.

citizen428 (approved):

app/javascript/onboarding/components/ProfileForm.jsx
```diff
@@ -2,33 +2,51 @@ import { h, Component } from 'preact';
...
   componentDidMount() {
+    this.getProfielFieldGroups();
     updateOnboarding('v2: personal info form');
   }
 
+  async getProfielFieldGroups() {
+    try {
+      const response = await request(`/profile_field_groups?onboarding=true`);
+      if (response.ok) {
+        const data = await response.json();
```
rhymes, line 40: ❤️

rhymes (approved):
It seems to be working! I also tried with a checkbox.

I noticed a bug with the onboarding flow but it's present also in production and I'll open a separate issue (filling a field, moving to the next slide and going back results in the field to be empty) :)

vaidehijoshi (approved):
This is great! I also had those codeclimate issues dismissed for you since you can't refactor propTypes, so you should be unblocked on that front! 🎉

app/javascript/onboarding/components/ProfileForm/TextArea.jsx
```diff
@@ -0,0 +1,36 @@
+import { h } from 'preact';
+import PropTypes from 'prop-types';
+import { FormField } from '@crayons';
+
+const TextArea = (props) => {
```
nickytonline, line 5: nitpick (non-blocking) Consider using a function expression instead of an [arrow function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) in the new components created. Arrow functions are only required when you need access to the lexical scope, i.e. dealing with `this`.

app/javascript/onboarding/components/ProfileForm/CheckBox.jsx
```diff
@@ -0,0 +1,35 @@
+import { h } from 'preact';
+import PropTypes from 'prop-types';
+import { FormField } from '@crayons';
+
+const CheckBox = (props) => {
```
nickytonline, line 5: refactor (blocking): A general note about components and JS in general. If it's exported from a file, document it via [JSDoc](https://jsdoc.app) comments.

We weren't doing this in the past, but started doing it a little while back. For an example, see https://github.com/forem/forem/blob/master/app/javascript/utilities/http/request.js

nickytonline (changes requested):
This is looking great @Ridhwana! A few more comments and change requests. 

Looking forward to seeing these improvements in the onboarding!

benhalpern:
There's a little height jump on the fetch that is jarring. I think we should have a minimum height so the box doesn't rapidly become small and then big again while the async function finishes...

But another thought: What if we added these values into the HTML as data attributes like we do with some other config...

```html
<div id="onboarding-container"
  data-community-name="<%= community_qualified_name %>"
  data-community-description="<%= SiteConfig.community_description %>"
  data-community-logo="<%= optimized_image_url(safe_logo_url(SiteConfig.onboarding_logo_image)) %>"
  data-community-background="<%= optimized_image_url(SiteConfig.onboarding_background_image, width: 1680, quality: 75, random_fallback: false) %>">
```

Maybe we want to role with the pattern you've already established, but if we did it like this it would probably simplify the frontend code for anything where the output is not dynamic based on prior slides. (e.g. everyone gets the same fields no matter what happens in previous slides)

app/javascript/onboarding/components/ColorField.jsx
```diff
@@ -0,0 +1,49 @@
...
+            name={attribute_name}
+            id={attribute_name}
+            onChange={onColorChange}
+          />
+        </div>
+        {description && <p class="crayons-field__description">{description}</p>}
+      </FormField>
+    );
+  }
+}
```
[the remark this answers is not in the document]
Ridhwana: ✅

app/javascript/onboarding/components/ProfileForm.jsx
```diff
@@ -71,7 +138,31 @@ class ProfileForm extends Component {
       communityConfig,
     } = this.props;
     const { profile_image_90, username, name } = this.user;
-    const { canSkip } = this.state;
+    const { canSkip, groups = [], error, errorMessage } = this.state;
+
+    if (error) {
+      return (
+        <div class="crayons-notice crayons-notice--danger">
```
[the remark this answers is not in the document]
Ridhwana: Thank you for this reminder @nickytonline

app/javascript/onboarding/components/ProfileForm.jsx
```diff
@@ -109,52 +200,7 @@ class ProfileForm extends Component {
...
-                  name="employer_name"
-                  id="employer_name"
-                  placeholder="e.g. Company name, self-employed, etc."
-                  onChange={this.handleChange}
-                  maxLength="60"
-                  className="onboarding-form-input--last"
-                />
-              </label>
-            </form>
+            <div>{groups && groups.length > 0 && sections}</div>
```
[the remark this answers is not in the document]
Ridhwana: ✅

app/javascript/onboarding/components/ProfileForm.jsx
```diff
@@ -71,7 +138,31 @@ class ProfileForm extends Component {
...
+    const sections = groups.map((group) => {
+      return (
+        <div class="onboarding-profile-sub-section">
+          <h2>{group.name}</h2>
+          {group.description && (
+            <div class="color-base-60">{group.description})</div>
+          )}
+          <div>
+            {group.profile_fields.map((field) => {
+              return this.renderAppropriateFieldType(field);
```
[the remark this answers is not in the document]
Ridhwana: ✅

app/javascript/onboarding/components/ProfileForm/ColorPicker.jsx
```diff
@@ -0,0 +1,50 @@
+import { h, Component } from 'preact';
+import PropTypes from 'prop-types';
+import { FormField } from '@crayons';
+
+export default class ColorPicker extends Component {
```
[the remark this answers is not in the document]
Ridhwana: ✅

app/javascript/onboarding/components/ProfileForm/TextArea.jsx
Ridhwana: ✅

app/javascript/onboarding/components/ProfileForm/CheckBox.jsx
Ridhwana: @nickytonline I couldn't find good practices on the internet to follow, the examples always deviated slightly. Maybe adding this to our documentation will help developers that are not familiar with jsdoc on how we'd like the functions to be documented?

Ridhwana: ✅

Ridhwana:
@nickytonline 

> This is looking great @Ridhwana! A few more comments and change requests.
> 
> Looking forward to seeing these improvements in the onboarding!

@nickytonline all feedback is addressed.

Ridhwana:
> There's a little height jump on the fetch that is jarring. I think we should have a minimum height so the box doesn't rapidly become small and then big again while the async function finishes...

> But another thought: What if we added these values into the HTML as data attributes like we do with some other config...

@benhalpern yeah the height change is definitely jarring. Whilst I like the idea of sending the data through via data attributes, I'd need to figure out the details of how this will work with a dynamic array of objects that we want to pass through. 

Since this PR has been in review for a week now, I'd like to postpone this investigation for a subsequent PR and just set the minimum height for now so that we can have this in production soon and then I'll improve on it.

Ridhwana:
> There's a little height jump on the fetch that is jarring. 

This turns out to be a little more complicated than just adding a max height, so @ludwiczakpawel is working his magic on a more appeasing fix.

Ridhwana:
> > There's a little height jump on the fetch that is jarring.
> 
> This turns out to be a little more complicated than just adding a max height, so @ludwiczakpawel is working his magic on a more pleasing fix.

app/javascript/onboarding/components/FollowTags.jsx
```diff
@@ -90,7 +90,9 @@ class FollowTags extends Component {
     }
 
     const classStyle =
```
Ridhwana, line 92: The changes in this file made by @ludwiczakpawel are not directly related to the profile onboarding but these are small changes for the UI design on the modal to try and keep the modal at the same height so that it doesn't look so jarring when it adjusts the height based on the content.

Ridhwana:
Note: some tests are broken from some UI changes, looking into these this evening.

nickytonline (approved):
Assuming specs get fixed, this LGTM. Thanks @Ridhwana! Looking forward to seeing these improvements in production. 🚀

benhalpern (approved):
Looks great to me pending spec fix!

Ridhwana:
Thank you everyone for the reviews (and @ludwiczakpawel for the UI assistance) on this PR 💚. The review took some time but only because there was loads of useful feedback that made such a big difference to the quality as well the aesthetic of the feature. 

I'll merge this in the my evening (US morning) before/after All Hands so that if anything is weird there is more support online.

PR #11648, opened 2020-11-28
Bug Fix:Remove nil cached_tag_lists when Collecting Tags

mstruve:
- [x] Bug Fix

## Description
There is a chance the `cached_tag_list` for an article is nil, in that case we want to make sure we remove it from our array when we are gathering up tags. 

Fixes: https://app.honeybadger.io/fault/66984/8942c036047ae64eb2372bbcf65e8dc6

## Added tests?
- [x] No, minor implementation change for an edge case I dont think is worth testing

alt_text

### DIFF
--- a/app/workers/follows/update_points_worker.rb
+++ b/app/workers/follows/update_points_worker.rb
@@ -34,7 +34,7 @@ module Follows
       tags = articles.pluck(:cached_tag_list).compact.flat_map { |list| list.split(", ") }
       occurrences = tags.count(tag.name)
       bonus = inverse_popularity_bonus(tag)
-      Math.log(occurrences + bonus + 1) # +1 is purelt to avoid log(0) => -infinity
+      Math.log(occurrences + bonus + 1) # +1 is purely to avoid log(0) => -infinity
     end
 
     def adjust_other_tag_follows_of_user(user_id)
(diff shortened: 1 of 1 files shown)


benhalpern (approved):

app/workers/follows/update_points_worker.rb
```diff
@@ -31,7 +31,7 @@ def calculate_implicit_points(tag, user)
       last_100_long_page_view_article_ids = user.page_views.where(time_tracked_in_seconds: 45..)
         .pluck(:article_id).last(100)
       articles = Article.where(id: last_100_reactable_ids + last_100_long_page_view_article_ids)
-      tags = articles.pluck(:cached_tag_list).flat_map { |list| list.split(", ") }
+      tags = articles.pluck(:cached_tag_list).compact.flat_map { |list| list.split(", ") }
       occurrences = tags.count(tag.name)
       bonus = inverse_popularity_bonus(tag)
       Math.log(occurrences + bonus + 1) # +1 is purelt to avoid log(0) => -infinity
```
citizen428, line 37: I know that's not the point of this PR, but any chance you could also fix the spelling mistake in the comment (`purelt` instead of `purely`), it triggers my OCD :wink:

citizen428, line 37: I know that's not the point of this PR, but any chance we can add this fix for the spelling mistake in the comment, it triggers my OCD :wink:

```suggestion
      Math.log(occurrences + bonus + 1) # +1 is purely to avoid log(0) => -infinity
```

citizen428 (approved):

app/workers/follows/update_points_worker.rb
mstruve: always have time to fix a spelling error :)

PR #11650, opened 2020-11-29
Bugfix: Remove focus from avatars in notifications for better keyboard navigation

### LINKED ISSUE #11403: Keyboard navigation in notifications area results in tabbing over redudant links
**Describe the bug**

When using keyboard navigation or a screen reader, image links in notification cards (avatars of orgs and users) can be tabbed too and are in the accessibilty tree. They are redundant links for a keyboard user/screen reader because there are already textual links for them in a notification card.

**To Reproduce**

1. Go to https://dev.to/notifications
2. Start tabbing with the keyboard. You will land on the avatar of an author or organization.
3. If you press ENTER on either of these avatars it will navigate to the author's or organization's profile (expected).
4. However, if you keep tabbing in the notification card, you will land on the same links in textual form for the author and organization.
5. This all works fine, but it means more tabbing for a keyboard user because of the redundant links and the same thing would occur with a screen reader.

**Expected behavior**

The avatar links in a notifications card are removed from the accessibility tree and they can not be tabbed to via the keyboard.

**Screenshots**

Nov-13-2020 09-26-24

- OS, version: n/a
- Browser, version: n/a

- Device: n/a
- OS, version: n/a
- Browser, version: n/a

**Additional context**

n/a

eevajonnapanula:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR removes notifications' user- and organisation avatars from tab-order and accessibility tree, as the same info can be found textually right after the avatars. 

With the case of multiple followers, I decided to go with [@nickytonline 's suggestion](https://github.com/forem/forem/issues/11403#issuecomment-735402472). This solution can be a bit confusing as sighted keyboard users can see that the first avatar on followers list doesn't get focus, and it can be unclear that the first avatar belongs to the user whose name is mentioned, but on the other hand, it saves a keystroke (which could potentially be really painful for some, so this helps them. Also, one tab less for all who use keyboard for navigation.).  Another option I was considering was to keep all of the avatars in the tab order, so there would have been one extra tab in these cases, but it would have potentially been less confusing. So if there are any ideas for that, I'm happy to refactor 😊 

Closes #11403 

Testing: 
1. Open notifications and have some notifications with user avatar(s)
2. Tab trough the notifications, and notice that the avatars in the notifications are not tabbable. 
3. If testing with a screen reader, pull up the list of links on the notifications page, and see that the avatars (with alt text "*username*'s profile") are not on the list.

Visual demonstration of the tabbing order with this PR

## Added tests?

- [ ] Yes
- [x] No, and this is why: I think I didn't touch any code, that had tests attached.

## Added to documentation?

- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/views/notifications/_aggregated_reactions.html.erb
+++ b/app/views/notifications/_aggregated_reactions.html.erb
@@ -4,14 +4,14 @@
 
 <div class="relative shrink-0 self-start">
   <% if actors.size == 1 %>
-    <a href="<%= actors.first["path"] %>" class="crayons-avatar crayons-avatar--l m:crayons-avatar--xl" tabindex="-1">
+    <a href="<%= actors.first["path"] %>" class="crayons-avatar crayons-avatar--l m:crayons-avatar--xl" aria-hidden="true" tabindex="-1">
       <img src="<%= actors.first["profile_image_90"] %>" class="crayons-avatar__image" alt="link to <%= actors.first["username"] %>'s profile" width="48" height="48">
     </a>
   <% else %>
-    <a href="<%= actors.first["path"] %>" class="crayons-avatar crayons-avatar--l mr-4" tabindex="-1">
+    <a href="<%= actors.first["path"] %>" class="crayons-avatar crayons-avatar--l mr-4" aria-hidden="true" tabindex="-1">
       <img src="<%= actors.first["profile_image_90"] %>" class="crayons-avatar__image" alt="link to <%= actors.first["username"] %>'s profile" width="48" height="48">
     </a>
-    <a href="<%= actors.first["path"] %>" class="crayons-avatar crayons-avatar--l absolute -right-1 -bottom-3 border-solid border-2 border-base-inverted" tabindex="-1">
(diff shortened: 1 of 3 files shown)


app/views/notifications/_aggregated_reactions.html.erb
```diff
@@ -4,14 +4,14 @@
 
 <div class="relative shrink-0 self-start">
   <% if actors.size == 1 %>
-    <a href="<%= actors.first["path"] %>" class="crayons-avatar crayons-avatar--l m:crayons-avatar--xl">
+    <a href="<%= actors.first["path"] %>" class="crayons-avatar crayons-avatar--l m:crayons-avatar--xl" aria-hidden="true" tabindex="-1">
```
nickytonline, line 7: This is looking great! You can probably pull this pull request out of Draft.

ludwiczakpawel:
Since implementation is more of a backend-ish thing, I summon @rhymes  :)

rhymes (approved):
LGTM!

nickytonline (approved):
Thanks so much for tackling this accessibility PR @eevajonnapanula! 🔥

PR #11679, opened 2020-11-30
Move app/labor/error_message_cleaner.rb to service

citizen428:

- [X] Refactor

## Description

This PR moves `ErrorMessageCleaner` from `app/labor` to the `ErrorMessages::Clean` service object.

https://github.com/forem/InternalProjectPlanning/issues/271=

## Added tests?

- [ ] No, and this is why: there were no existing ones and the cost of adding them seems not to be worth the effort. Maybe we should find a better solution for this eventually instead.

## Added to documentation?

- [X] No documentation needed


rhymes (approved):
That's a very specific service 🤣

citizen428:
@rhymes Agreed, which is why I'd like to eliminate it eventually. A very similar block of code executes in two contexts, we probably should catch the original exception which looks like it comes from Psych, and re-raise a custom one with the correct error. But for now this will have to do.

benhalpern (approved):

nickytonline (approved):
Thanks for the cleanup @citizen428! 🚀

PR #11684, opened 2020-11-30
Refactor EdgeCache

atsmith813:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Added tests?

- [ ] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [ ] No documentation needed

alt_text


app/services/edge_cache/bust_comment.rb
```diff
@@ -0,0 +1,32 @@
...
+      commentable.comments.includes(:user).find_each do |comment|
+        bust(comment.path)
+        bust("#{comment.path}?i=i")
+      end
+
+      bust("#{commentable.path}/comments/*")
+    end
+
+    # bust commentable if it's an article
+    def self.bust_article_comment(commentable)
```
citizen428, line 21: Do you want this to be part of the public interface or should we add a `private_class_method :bust_article_comment`?

app/services/edge_cache/bust_comment.rb
```diff
@@ -0,0 +1,32 @@
...
+      bust("/") if Article.published.order(hotness_score: :desc).limit(3).ids.include?(commentable.id)
+
+      if commentable.decorate.cached_tag_list_array.include?("discuss") &&
```
citizen428, line 24: How about moving this check into the decorator? Something like 

```ruby
commentable.decorate.discussion?
```

or even more general

```ruby
commentable.decorate.short_lived?
```

app/services/edge_cache/bust.rb
```diff
@@ -1,49 +1,44 @@
...
         DatadogStatsClient.increment("edgecache_bust.invalid_provider_class",
                                      tags: ["provider_class:#{provider_class}"])
+        false
       end
-
-      self
     end
 
-    attr_reader :provider, :path, :response
+    def self.determine_provider_class
```
rhymes, line 24: should this be private?

app/services/edge_cache/bust.rb
```diff
@@ -1,49 +1,44 @@
...
 
-    def initialize(path)
-      @path = path
-      @provider = determine_provider
-      @response = nil
-    end
-
-    def call
-      return unless provider
+    def self.bust(path)
```
rhymes, line 7: should this be protected?

app/services/edge_cache/bust.rb
```diff
@@ -1,49 +1,44 @@
...
-      if fastly_enabled?
-        "fastly"
-      elsif nginx_enabled?
-        "nginx"
-      end
+      "#{self}::#{provider.capitalize}".constantize
     end
 
-    def fastly_enabled?
+    def self.fastly_enabled?
```
rhymes, line 37: should this be private?

app/services/edge_cache/bust.rb
```diff
@@ -1,49 +1,44 @@
...
       ApplicationConfig["FASTLY_API_KEY"].present? && ApplicationConfig["FASTLY_SERVICE_ID"].present?
     end
 
-    def nginx_enabled?
+    def self.nginx_enabled?
```
rhymes, line 41: should this be private?

app/services/edge_cache/bust_comment.rb
```diff
@@ -0,0 +1,32 @@
+module EdgeCache
+  class BustComment < Bust
```
rhymes, line 2: see my comment about merging `BustComment` and `BustCommentable` into `BustCommentable`: after all we're busting the object the comment is attached to and its comments, not "a comment"

app/services/edge_cache/bust_commentable.rb
```diff
@@ -0,0 +1,11 @@
+module EdgeCache
+  class BustCommentable < Bust
+    def self.call(commentable)
+      return unless commentable
+
+      EdgeCache::BustComment.call(commentable)
+      bust("#{commentable.path}/comments")
+      commentable.index_to_elasticsearch_inline
```
rhymes, line 8: as `BustComment` is never actually called outside `BustCommentable`, what if we merge these two files in `BustCommentable` ?

rhymes (commented):
Cool! I left some comments about hiding stuff inside classes and merging `BustComment` and `BustCommentable`

app/services/edge_cache/bust_commentable.rb
atsmith813: It is called outside `BustCommentable`:

https://github.com/forem/forem/blob/7f066dd2036c91a7a9742b89f26b0c581be5014b/app/services/users/delete_articles.rb#L14

https://github.com/forem/forem/blob/7f066dd2036c91a7a9742b89f26b0c581be5014b/app/services/users/delete_comments.rb#L10

Both of those services call other bust methods under the old `CacheBuster`. I'll refactor them once all of the methods are "moved over".

app/services/edge_cache/bust_comment.rb
atsmith813: Can we hold off on this? I'm trying to make as few changes as possible in each PR so that it's easier to debug if something goes wrong. If I make this change now, I'll have to debug this refactor if something goes wrong vs comparing a copy/paste.

Also, this suggestion means I'd have to update _each_ decorator that qualifies as "commentable" - Article, PodcastEpisode, etc. I'd rather make that a more contained change in its own PR so it stretches the scope of this PR wider.

atsmith813:
@rhymes @citizen428 I was trying to hold off on fully optimizing some of this (i.e. private class methods) so that it would be easier for me to refactor as I go since this is step 1 of what will likely be several PRs. For example, I'm hesitant about making `fastly_enabled?` private. Since we are changing how that class works, we lose the `response` and `provider` attributes on that class:

https://github.com/forem/forem/pull/11684/files#diff-0d8284e88297859e6a2e5c8a3406dd3c23c227cf453033c7e260b56260233121L9-L10

And with this change, we won't have an easy way in code anymore to determine the provider. I was planning to leave `fastly_enabled?` public and see how this refactor shook out - maybe use that method instead of a `provider` attribute. Either way, it's no big deal for me to just make it private right now and change it later if needed, so I went ahead and implemented all of your suggestions to make them private.

app/services/edge_cache/bust_commentable.rb
rhymes: got it! Thanks!

rhymes (approved):
LGTM

app/decorators/article_decorator.rb
```diff
@@ -96,4 +96,10 @@ def co_author_name_and_path
       "<b><a href=\"#{user.path}\">#{user.name}</a></b>"
     end.to_sentence
   end
+
+  # Used in determining when to bust additional routes for an Article's comments
+  def discussion?
+    cached_tag_list_array.include?("discuss") &&
+      featured_number.to_i > 35.hours.ago.to_i
```
vaidehijoshi, line 103: Just curious, how did we come up with 35?

spec/services/edge_cache/bust_spec.rb
```diff
@@ -30,22 +27,12 @@
...
       end
 
-      let(:cache_bust_service) { described_class.new(path) }
-
       it "can bust a fastly cache" do
         allow(fastly_provider_class).to receive(:call)
 
-        cache_bust_service.call
-        expect(cache_bust_service.provider).to eq("fastly")
+        described_class.call(path)
```
vaidehijoshi, line 33: Can we test the expected `determine_provider_class` here as well? Just thinking that if we remove [this line](https://github.com/forem/forem/pull/11684/files#diff-bf52b55fce7912a9ec4a31f89a3f3e0face4025c68f5ce26da00c75c963163f8L39) we don't have any test that asserts that the provider we expect to be returned by the class is actually correct (but now that we've made it `protected`, I think we'd have to use `public_send` here)

spec/services/edge_cache/bust_spec.rb
```diff
@@ -84,8 +68,7 @@
       it "can bust an nginx cache" do
         allow(nginx_provider_class).to receive(:call)
 
-        cache_bust_service.call
-        expect(cache_bust_service.provider).to eq("nginx")
```
vaidehijoshi, line 88: I have the same thought about testing the `determine_provider_class` here

app/decorators/article_decorator.rb
atsmith813: I have absolutely no idea 😅  I just copied it over. Maybe @benhalpern knows?

citizen428 (approved):

spec/services/edge_cache/bust_spec.rb
atsmith813: On second thought, I'm thinking maybe we revisit this in another PR once the refactor is more fleshed out because 1) `determine_provider_class` is now a `private` method and 2) it's tested in a roundabout way [here](https://github.com/forem/forem/pull/11684/files#diff-bf52b55fce7912a9ec4a31f89a3f3e0face4025c68f5ce26da00c75c963163f8R34). If the `determine_provider_class` didn't work as expected, this spec would fail because it wouldn't end up calling the correct class.

How does that sound @vaidehijoshi?

vaidehijoshi: That's fair! I can get down with punting this to another PR with more refactoring done :)

app/decorators/article_decorator.rb
vaidehijoshi: @benhalpern do you have the magic answer about the mystery 35 here? 😛

vaidehijoshi (approved):

benhalpern (approved):

PR #11687, opened 2020-11-30
Adds system spec that reproduces NoMethodError

fdoxyz:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Reproducing error

## Description

Test to reproduce edge case


citizen428:
Just FYI: we closed this after a phone call as it didn't actually demonstrate the issue correctly. However, while talking through it we unearthed the real issue, so this PR served its purpose.

PR #11689, opened 2020-12-01
Improve error message of User Subscriptions

vaidehijoshi:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Currently, you can only use the user subscription feature on an "active" source, aka a published article or a comment that hasn't been deleted. If you try to test out this feature on a source that _isn't_ active, then you'll get an error message about "the Source not being found":
image

While a user _might_ be able to guess what this means, we could be a little more clear in this message I think! I've updated the message to be a bit more specific :) 

Inspired by [this Slack thread](https://dev-internal.slack.com/archives/C7GE84DEJ/p1606784386377500).

To QA this feature, pull down this branch locally, and make sure the user you are logged in as has the `:restricted_liquid_tag` role. Once you have the role, create a new article that includes the user subscription liquid tag in the markdown (see the [relevant docs here](https://app.gitbook.com/@forem/s/team-handbook/tech/user-subscription-liquid-tag)):

```
{% user_subscription I love llamas, click here to subscribe to top-notch llama content!! %}
```

You should see an error message that reads: `Source not found. Please make sure your Article is active!`

Since this just changes the content of an error message, it should be a copy change only and not affect accessibility at all :) 

## Added tests?

- [x] ~Yes~ Updated pre-existing tests! 😉 
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

Nope!

### DIFF
--- a/app/models/user_subscription.rb
+++ b/app/models/user_subscription.rb
@@ -90,6 +90,6 @@ class UserSubscription < ApplicationRecord
 
     return if source_active
 
-    errors.add(:base, "Source not found.")
+    errors.add(:base, "Source not found. Please make sure your #{user_subscription_sourceable_type} is active!")
   end
 end
--- a/spec/models/user_subscription_spec.rb
+++ b/spec/models/user_subscription_spec.rb
@@ -26,7 +26,7 @@ RSpec.describe UserSubscription, type: :model do
                                                                                    published: false)
       user_subscription = described_class.build(source: unpublished_source, subscriber: subscriber)
       expect(user_subscription).not_to be_valid
-      expect(user_subscription.errors[:base]).to include "Source not found."
+      expect(user_subscription.errors[:base]).to include "Source not found. Please make sure your Article is active!"
     end
 
     it "validates the tag is enabled in the source" do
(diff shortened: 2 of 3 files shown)


app/models/user_subscription.rb
```diff
@@ -90,6 +90,6 @@ def active_user_subscription_source
 
     return if source_active
 
-    errors.add(:base, "Source not found.")
+    errors.add(:base, "Source not found. Please make sure your #{user_subscription_sourceable_type} is active!")
```
mstruve, line 93: When you say active, do you mean published? Maybe published would be a clearer term.

mstruve (approved):
Small clarity comment but not blocking. LGTM!

app/models/user_subscription.rb
vaidehijoshi: The reason I didn't go with published is because this error is also raised when the source is a comment—in that case, the comment has to be not deleted rather than not published. I figured that active would be a cleaner way of encompassing published articles and not-deleted comments :)

juliannatetreault (approved):
👍 👍 !

atsmith813 (approved):

benhalpern (approved):

PR #11695, opened 2020-12-01
Fix invalid feed error display in settings

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

When an invalid feed is submitted in the "Extensions" page, one of the various templates rendered there breaks with a 500 error, ths is the fix for that.

1. go to dev.to/settings/extensions and add a URL that's not a feed
1. press save/submit
1. witness the finality of the error
1. do the same with this branch and see the actual error message

## Added tests?

- [x] Yes
      have not been included_

### DIFF
--- a/app/views/users/_publishing_from_rss.html.erb
+++ b/app/views/users/_publishing_from_rss.html.erb
@@ -51,7 +51,7 @@
     </div>
 
     <%= f.hidden_field :tab, value: @tab %>
-    <% button_text = @user.feed_url.present? ? "Save" : "Submit" %>
+    <% button_text = @user.feed_url.present? ? "Save Feed Settings" : "Submit Feed Settings" %>
     <div><button class="crayons-btn" type="submit"><%= button_text %></button></div>
   <% end %>
 
--- a/app/views/users/_response_templates.html.erb
+++ b/app/views/users/_response_templates.html.erb
@@ -33,7 +33,7 @@
 
   <%= form_with model: @response_template do |f| %>
     <section class="flex flex-col gap-3">
-      <% if @response_template.persisted? %>
+      <% if @response_template&.persisted? %>
         <% title = params[:previous_title] || @response_template.title %>
         <% content = params[:previous_content] || @response_template.content %>
         <h3 class="crayons-subtitle-3">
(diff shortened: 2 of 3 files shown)


app/views/users/_response_templates.html.erb
```diff
@@ -33,7 +33,7 @@
 
   <%= form_with model: @response_template do |f| %>
     <section class="flex flex-col gap-3">
-      <% if @response_template.persisted? %>
+      <% if @response_template&.persisted? %>
```
rhymes, line 36: this is the actual fix :D

spec/system/user/user_edits_extensions_spec.rb
```diff
@@ -37,4 +37,20 @@
       expect(page).to have_text("Connected to Stackbit")
     end
   end
+
+  describe "Feed" do
+    before do
+      visit user_settings_path(:extensions)
+    end
+
+    it "fails if the feed URL is invalid" do
```
rhymes, line 46: regression test to make sure it doesn't happen again

app/views/users/_publishing_from_rss.html.erb
```diff
@@ -51,7 +51,7 @@
     </div>
 
     <%= f.hidden_field :tab, value: @tab %>
-    <% button_text = @user.feed_url.present? ? "Save" : "Submit" %>
+    <% button_text = @user.feed_url.present? ? "Save Feed Settings" : "Submit Feed Settings" %>
```
rhymes, line 54: to mimic the other "save" buttons that are self explanatory, also helps in e2e testing

app/views/users/_response_templates.html.erb
juliannatetreault: It's the little things! 😄

juliannatetreault (approved):
While I love the spinning robot, I definitely prefer seeing the error message over the `500` page -- awesome fix! 😄

app/views/users/_response_templates.html.erb
rhymes: ehhehe

Ridhwana (approved):
LGTM

Issue #11194, opened 2020-10-30
Show confirmation modal when there are unsaved changes in Admin/Config


lisasy:
**Is your feature request related to a problem? Please describe.**

Today, when admins make edits in `admin/config`, they must type in a confirmation message and press "Update Site Configuration" to save the changes. However, this confirmation area is very hidden and not intuitive, so I hypothesize that admins are making updates that aren't being saved.

**Describe the solution you'd like**

When an admin takes an action that hasn't been saved and tries to navigate away from `admin/config`, trigger this confirmation modal:

![2020-10-30_Admin-Confirmation-Modal](https://user-images.githubusercontent.com/816402/97748561-e1ccf400-1aaa-11eb-92b7-60e555ef5d38.png)

- If admin selects `Yes, discard changes`, then let them continue navigating away.
- If admin selects `No, keep editing`, then close the modal. 

**Additional context**

There are two ways to implement the modal: 
1. Preact component.
2. Straight HTML markup.

I'd defer to whoever wants to work on this to investigate what would be the best option for us. Thank you!

PR #11696, opened 2020-12-01
[EOY2020 Issue] Default new forems to Email Authentication only

msarit:

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR ensures that new Forems will default to Email Authentication only. The PR also ensures that an Authentication Provider is NOT enabled if either their respective Key or Secret is missing.

closes https://github.com/forem/InternalProjectPlanning/issues/336

None

1. Navigate to `localhost:xxxx/admin/config` as a user with `:super_admin` role
2. Under the `Authentication` section, enable some Auth Providers, but ensure their key and secret are blank. You should see a warning banner about the empty Key and Secret fields. Update Site Configuration. You should find that the Auth Providers you updated were NOT enabled.
3. Repeat #2 but with random strings in the Auth Providers' Key and Secret fields. Update Site Configuration. You should find that the Auth Providers you updated ARE enabled. The warning banner should also be gone.

#### Demo Video:
https://www.loom.com/share/1b9d5a60946b4f469d9781e8f7de344b

None.

## Added tests?

- [X] Yes, all associated tests were updated
      have not been included_

## Added to documentation?

- [ ] README
- [X] No documentation needed

None

Fish Mooney issuing threats


msarit:
@lisasy I'm requesting you on this PR Lisa because I make use of a Warning Banner and want to ensure my usage is kosher 🙏

msarit:
@vaidehijoshi @joshpuetz

In regards to failing specs, I believe a majority of them are due to Authentication Providers not being enabled by default (see `app/models/site_config.rb:33`) :

```ruby
field :authentication_providers, type: :array, default: %w[]
```

**I think I'm good on how to resolve most of them**, but I *may need pairing help* with a few. Just FYI 😄 

List of failing specs (for reference):

```
./spec/requests/github_repos_spec.rb
./spec/requests/user/user_settings_spec.rb

./spec/services/authentication/authenticator_spec.rb
./spec/services/authentication/providers_spec.rb
./spec/services/broadcasts/welcome_notification/generator_spec.rb
./spec/services/users/delete_spec.rb

./spec/system/authentication/redirects_using_referer_spec.rb
./spec/system/authentication/user_logs_in_with_facebook_spec.rb
./spec/system/authentication/user_logs_in_with_github_spec.rb
./spec/system/homepage/user_visits_homepage_spec.rb

./spec/system/user_logs_in_with_password_spec.rb
./spec/system/authentication/user_logs_in_with_twitter_spec.rb

./spec/models/user_spec.rb

```

joshpuetz (commented):
Testing locally matched your expected steps. LGTM! 👍

app/views/admin/configs/_auth_provider_settings.html.erb
```diff
@@ -1,48 +1,53 @@
...
-      data-auth-provider="<%= provider.provider_name %>"
-      data-action="click->config#disableAuthProvider">
-      Undo
-    </button>
-  <% end %>
-</div>
+  <div
+    class="crayons-notice crayons-notice--warning <%= provider_keys_configured?(provider.provider_name) ? "hidden" : "" %>"
+    style="padding: var(--su-2);">
+      <strong>Warning:</strong> Provider will <strong>not</strong> be enabled if Key or Secret missing
```
vaidehijoshi, line 5:
```suggestion
      <strong>Note:</strong>This authentication provider will <strong>not</strong> be enabled if the key or secret are missing.
```

vaidehijoshi: We also should probably add some copy to prompt the user to confirm that their secret and key are correct. If they typo or don't copy it properly, they won't see the warning, but the auth provider still won't be enabled correctly!

Adding some copy to this warning like `Please make sure that you copy your secret and key correctly` might be helpful for our not-as-technical Forem creators.

app/views/admin/configs/_auth_provider_settings.html.erb
```diff
@@ -1,48 +1,53 @@
...
-      class="crayons-btn crayons-btn--secondary"
-      data-auth-provider="<%= provider.provider_name %>"
-      data-action="click->config#disableAuthProvider">
-      Undo
-    </button>
-  <% end %>
-</div>
+  <div
+    class="crayons-notice crayons-notice--warning <%= provider_keys_configured?(provider.provider_name) ? "hidden" : "" %>"
+    style="padding: var(--su-2);">
```
vaidehijoshi, line 4: Let's use a class here rather than an inline style. I would define it in `app/assets/stylesheets/admin.scss`, which is where we define the parent container's class, `.config-authentication-form` as well :)

spec/helpers/authentication_helper_spec.rb
```diff
@@ -38,6 +38,29 @@
     end
   end
 
+  describe "#provider_keys_configured?(provider)" do
+    provider = "facebook"
```
vaidehijoshi, line 42: Let's use rspec's `let` [helper method](https://relishapp.com/rspec/rspec-core/v/2-10/docs/helper-methods/let-and-let) here, which is more idiomatic :)
```suggestion
    let(:provider) { "facebook" }
```

vaidehijoshi (changes requested):
This is very close! Left a few comments to improve our copy + and make our code a bit more idiomatic before merging this :)

lisasy:
Hi @msarit,

Thanks for including me here! I've gotten to review the Loom, and I'd like to push for a different approach that would not only be a better UX, but will fit into the larger strategy for the set of improvements we'll make with Admin in general. 

The goal here is to let the admin know that they need to fill out required form fields for their action to successfully occur. Instead of using the `warning notice` element that appears above empty form fields, the more ideal user experience would be:

1. When user clicks "Enable" to setup new method, don't show any warning notice.
2. They either don't complete the form, or enter information unsuccessfully.
3. They attempt to confirm the update with inputting the confirmation text and clicking "Update configuration"
4. Trigger a modal with messaging akin to what is below:

Screen Shot 2020-12-01 at 8 45 30 PM

---

When it comes to these types of UX, we want to approach them holistically, and this modal approach is apart of a larger, extensible approach ([see here](https://github.com/forem/forem/issues/11194)). If we were to allow a warning notice to communicate to a user what they need to do, why wouldn't we do that for every single other configuration area? You might imagine how messy that could easily become. Let me know what you think about all of this, and I'm happy to work with you and help out.

joshpuetz:
> Hi @msarit,
> 
> Thanks for including me here! I've gotten to review the Loom, and I'd like to push for a different approach that would not only be a better UX, but will fit into the larger strategy for the set of improvements we'll make with Admin in general.
> 
> The goal here is to let the admin know that they need to fill out required form fields for their action to successfully occur. Instead of using the `warning notice` element that appears above empty form fields, the more ideal user experience would be:
> 
> 1. When user clicks "Enable" to setup new method, don't show any warning notice.
> 2. They either don't complete the form, or enter information unsuccessfully.
> 3. They attempt to confirm the update with inputting the confirmation text and clicking "Update configuration"
> 4. Trigger a modal with messaging akin to what is below:
> 
> Screen Shot 2020-12-01 at 8 45 30 PM
> 
> When it comes to these types of UX, we want to approach them holistically, and this modal approach is apart of a larger, extensible approach ([see here](https://github.com/forem/forem/issues/11194)). If we were to allow a warning notice to communicate to a user what they need to do, why wouldn't we do that for every single other configuration area? You might imagine how messy that could easily become. Let me know what you think about all of this, and I'm happy to work with you and help out.

This is indeed a great solution, but seems like a HUGE expansion of the original issue @msarit is trying to solve here (https://github.com/forem/InternalProjectPlanning/issues/336). I'd suggest a new issue that track that UX @lisasy!

msarit:
> This is indeed a great solution, but seems like a HUGE expansion of the original issue @msarit is trying to solve here (forem/InternalProjectPlanning#336). I'd suggest a new issue that track that UX @lisasy!

@lisasy @joshpuetz I agree with Josh. This flow is considerable lift. I will create a new issue that addresses this task, and ask Molly to add to the EOY Priority Board, then I can take it on. Cool?

spec/models/user_spec.rb
```diff
@@ -28,7 +28,10 @@ def provider_username(service_name)
   let(:other_user) { create(:user) }
   let(:org) { create(:organization) }
 
-  before { omniauth_mock_providers_payload }
+  before do
+    omniauth_mock_providers_payload
+    allow(SiteConfig).to receive(:authentication_providers).and_return(%w[facebook github twitter])
```
rhymes, line 33: Instead of hardcoding these you might prefer using `Authentication::Providers.available` - this way when we add more we don't need to go around and add the new one to the list.

It's probably even better to add 

```ruby
allow(SiteConfig).to receive(:authentication_providers).and_return(Authentication::Providers.available)
``` 

to the various functions `omniauth_mock_providers_payload`, `omniauth_mock_facebook_payload`, `omniauth_mock_github_payload`, `omniauth_mock_twitter_payload` so you don't have to explicit this everywhere now that the default value is gone. 

Though I haven't tried, it might not work :) first suggestion still stands

spec/policies/github_repo_policy_spec.rb
```diff
@@ -23,6 +23,7 @@
 
     before do
       omniauth_mock_github_payload
+      allow(SiteConfig).to receive(:authentication_providers).and_return(%w[facebook github twitter])
```
rhymes, line 26: same note

spec/requests/admin/configs_spec.rb
```diff
@@ -75,19 +75,44 @@
       describe "Authentication" do
         it "updates enabled authentication providers" do
           enabled = Authentication::Providers.available.first.to_s
-          post "/admin/config", params: { site_config: { auth_providers_to_enable: enabled },
-                                          confirmation: confirmation_message }
+          post "/admin/config", params: {
```
rhymes, line 78: you can replace it with `admin_config_path`

spec/requests/admin/configs_spec.rb
```diff
@@ -75,19 +75,44 @@
...
+          }
           expect(SiteConfig.authentication_providers).to eq([enabled])
         end
 
         it "strips empty elements" do
           provider = Authentication::Providers.available.first.to_s
           enabled = "#{provider}, '', nil"
-          post "/admin/config", params: { site_config: { auth_providers_to_enable: enabled },
-                                          confirmation: confirmation_message }
+          post "/admin/config", params: {
```
rhymes, line 92: same

spec/requests/admin/configs_spec.rb
```diff
@@ -75,19 +75,44 @@
...
+              auth_providers_to_enable: enabled
+            },
+            confirmation: confirmation_message
+          }
           expect(SiteConfig.authentication_providers).to eq([provider])
         end
 
+        it "does not update enabled authentication providers if any associated key missing" do
+          enabled = Authentication::Providers.available.first.to_s
+          post "/admin/config", params: {
```
rhymes, line 105: same

spec/services/authentication/providers_spec.rb
```diff
@@ -32,10 +34,6 @@
   end
 
   describe ".enabled" do
-    it "lists all available providers" do
-      expect(described_class.available).to eq(described_class.enabled)
-    end
-
```
rhymes, line 38: why did we lose this?

rhymes (commented):
Hi @msarit, thanks for the PR! 

I mainly left some notes about creating a shortcut to avoid the repetition of "available providers" around the code :)

spec/services/authentication/providers_spec.rb
msarit: @rhymes Good question! 😄  In my mind, this spec made sense when we had all providers enabled as default. But that is no longer the case, so comparing `described_class.available` to `described_class.enabled` seems moot? Unless I am misunderstanding... please let me know

spec/models/user_spec.rb
msarit: > this way when we add more we don't need to go around and add the new one to the list

You know @rhymes I thought about this, but my brain was too fried to realize that sing `Authentication::Providers.available` was the solution 😅

spec/services/authentication/providers_spec.rb
rhymes: Oh! you're right! :D Sorry, I got confused hahaha

joshpuetz (approved):

app/views/admin/configs/_auth_provider_settings.html.erb
```diff
@@ -1,48 +1,52 @@
...
-    <button
-      class="crayons-btn crayons-btn--secondary"
-      data-auth-provider="<%= provider.provider_name %>"
-      data-action="click->config#disableAuthProvider">
-      Undo
-    </button>
-  <% end %>
-</div>
+  <div
+    class="crayons-notice crayons-notice--warning warning-padding <%= provider_keys_configured?(provider.provider_name) ? "hidden" : "" %>">
```
vaidehijoshi, line 3:
```suggestion
    class="crayons-notice crayons-notice--warning auth-warning <%= provider_keys_configured?(provider.provider_name) ? "hidden" : "" %>">
```

app/assets/stylesheets/admin.scss
```diff
@@ -147,6 +147,10 @@ label {
   gap: var(--su-4);
   padding-top: var(--su-6);
   padding-bottom: var(--su-2);
+
+  .warning-padding {
+    padding: var(--su-2);
+  }
```
vaidehijoshi, line 153:
```suggestion
  .auth-warning {
    padding: var(--su-2); // Override default crayons notice padding
  }
```

```suggestion
  .auth-warning {
```

I think this is a bit more future-proof rather than include `-padding` in the css class name. When I see the `-padding` here, I almost think it's something from crayons or even a utility class, so I think that could potentially be confusing. But `auth-warning` is more clear that this is a custom class that's overriding the default padding [that comes from crayons](https://github.com/forem/forem/blob/master/app/assets/stylesheets/components/notices.scss#L4) (which is why I'm suggesting the additional comment here 😄)

vaidehijoshi (approved):
💪 this should be good to go pending CI going green 🥗

rhymes:
Great 🔥 ❤️

PR #11698, opened 2020-12-01
Add config post route for admin api

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As a follow up for https://github.com/forem/forem/pull/11358 this PR adds the ability for admins to create/update content within the SiteConfig object.

"Upsert" seemed like the appropriate description for _posting_ to `SiteConfig`.

https://github.com/forem/InternalProjectPlanning/issues/202

No UI

## Added tests?

- [x] Yes
      have not been included_

I didn't add a unit tests, as the request specs do a good job of exercising this new service object at the moment.

## Added to documentation?

- [ ] README
- [ ] No documentation needed

Will add docs to admin guide if merged.

picard


app/controllers/admin/configs_controller.rb
```diff
@@ -181,20 +41,11 @@ def confirmation_text
...
-        MASCOT_PARAMS |
-        IMAGE_PARAMS |
-        ONBOARDING_PARAMS |
-        JOB_PARAMS
 
       has_emails = params.dig(:site_config, :email_addresses).present?
       params[:site_config][:email_addresses][:default] = ApplicationConfig["DEFAULT_EMAIL"] if has_emails
       params&.require(:site_config)&.permit(
-        all_params,
+        SiteConfig.keys.map(&:to_sym),
```
benhalpern, line 48: I think the nature of this object is that it should be entirely readable and writeable by admins and no attribute is off limits per se, so I did away with our allowed list. It seems okay for this particular object.

app/controllers/concerns/config_params.rb
```diff
@@ -0,0 +1,18 @@
+
+module ConfigParams
+  extend ActiveSupport::Concern
+  def config_params
+    special_params_to_remove = %w[authentication_providers email_addresses meta_keywords credit_prices_in_cents]
+    special_params_to_add = %w[auth_providers_to_enable]
```
rhymes, line 6: this can be promoted to constants I guess

app/services/site_configs/upsert.rb
```diff
@@ -0,0 +1,114 @@
+module SiteConfigs
+  module Upsert
+    include AuthenticationHelper
```
rhymes, line 3: it's a bit odd to include a view helper in a service, I wonder if there's a separate way to do this

app/services/site_configs/upsert.rb
```diff
@@ -0,0 +1,114 @@
...
+        mascot_image_url
+        mascot_footer_image_url
+        onboarding_logo_image
+        onboarding_background_image
+        onboarding_taskcard_image
+      ].freeze
+
+    VALID_URL = %r{\A(http|https)://([/|.|\w|\s|-])*.[a-z]{2,5}(:[0-9]{1,5})?(/.*)?\z}.freeze
+
+    def self.call(configs)
```
rhymes, line 21: you could use 

`rails generate service site_configs/upsert config`

which generates a template service, this way you can encapsulate all the methods below in a class with `private` as they all access state, `@configs`, and aren't called anywhere else outside, thus they probably don't need to be class methods

docs/api_v0.yml
```diff
@@ -652,7 +652,11 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/Comment"
-
+    ConfigUpsert:
```
rhymes, line 655: maybe just `SiteConfig` ?

`upsert` is the operation, we're describing the payload

rhymes (commented):
Aside from the various inline notes I think we 100% need to add auditing to this new operation. Everytime someone sends an update to the configuration, this should leave a trace in the audit log. In theory we should also add that to the regular admin config controller. 

You can check 

https://github.com/forem/forem/blob/c5a29826e8a3bf5424ceba57b91525ec2af0a827/app/controllers/admin/response_templates_controller.rb#L4-L6

for an example on how to set that up

app/services/site_configs/upsert.rb
citizen428: I 👍 this suggestion. The reason I added the service generator is to get slightly more consistency in this area, so personally, I'd appreciate it if we start using it a bit more.

app/services/site_configs/upsert.rb
```diff
@@ -0,0 +1,112 @@
...
+      hex.present? && !hex.match?(/\A#(\h{6}|\h{3})\z/)
+    end
+
+    def self.allowed_domains_include_improper_format
+      domains = @configs[:allowed_registration_email_domains]
+      return unless domains
+
+      domains_array = domains.delete(" ").split(",")
+      valid_domains = domains_array
+        .select { |d| d.match?(/^[a-zA-Z0-9][a-zA-Z0-9-]{1,61}[a-zA-Z0-9](?:\.[a-zA-Z]{2,})+$/) }
```
citizen428, line 104: This is not super important, but I'd move this regex out into a constant and give it a name, because that's (a) better for readability  and (b) means we don't need to recompile this regex repeatedly. We can also make it shorter by using the `i` modifier (case-insensitive) instead of repeatedly specifying both character classes.

```ruby
VALID_DOMAIN_REGEX = /^[a-z0-9][a-z0-9-]{1,61}[a-z0-9](?:\.[a-z]{2,})+$/i.freeze

.select { |d| d.match?(VALID_DOMAIN_REGEX)
```

app/controllers/concerns/site_config_params.rb
```diff
@@ -0,0 +1,17 @@
+module SiteConfigParams
+  extend ActiveSupport::Concern
+  SPECIAL_PARAMS_TO_ADD = %w[authentication_providers email_addresses meta_keywords credit_prices_in_cents].freeze
+  SPECIAL_PARAMS_TO_REMOVE = %w[auth_providers_to_enable].freeze
+  def config_params
+    has_emails = params.dig(:site_config, :email_addresses).present?
+    params[:site_config][:email_addresses][:default] = ApplicationConfig["DEFAULT_EMAIL"] if has_emails
+    params&.require(:site_config)&.permit(
```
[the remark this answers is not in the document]
rhymes: Now that you mention it, the second as well:

```ruby
[1] pry(main)> params = ActionController::Parameters.new({})
=> <ActionController::Parameters {} permitted: false>
[2] pry(main)> params.require(:site_config)
ActionController::ParameterMissing: param is missing or the value is empty: site_config
from /Users/rhymes/Development/forem/vendor/bundle/ruby/2.7.0/gems/actionpack-6.0.3.4/lib/action_controller/metal/strong_parameters.rb:462:in `require'
```

there's no chance to get a `nil` here

app/controllers/api/v0/admin/configs_controller.rb
```diff
@@ -2,12 +2,25 @@ module Api
...
         before_action :authenticate_with_api_key_or_current_user!
         before_action :authorize_super_admin
+        after_action :bust_content_change_caches, only: [:create]
 
         def show
           @site_configs = SiteConfig.all
         end
+
+        def create
+          result = SiteConfigs::Upsert.call(config_params)
```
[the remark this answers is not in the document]
benhalpern: 👍

app/services/site_configs/upsert.rb
benhalpern: 👍

app/services/site_configs/upsert.rb
benhalpern: 👍

app/controllers/admin/configs_controller.rb
```diff
@@ -147,30 +21,22 @@ class ConfigsController < Admin::ApplicationController
...
-          update_enabled_auth_providers(value) unless value.class.name != "String"
-        elsif value.is_a?(Array)
-          SiteConfig.public_send("#{key}=", value.reject(&:blank?)) unless value.empty?
-        elsif value.respond_to?(:to_h)
-          SiteConfig.public_send("#{key}=", value.to_h) unless value.empty?
-        else
-          SiteConfig.public_send("#{key}=", value.strip) unless value.nil?
-        end
+      result = SiteConfigs::Upsert.call(site_config_params)
+      if result[:result] == "errors"
```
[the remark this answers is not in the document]
benhalpern: 👍

app/controllers/concerns/site_config_params.rb
```diff
@@ -0,0 +1,16 @@
+module SiteConfigParams
```
citizen428, line 1: **suggestion (non-blocking):** I'm personally not a big of concerns like this one. A service like `SiteConfigs::PrepareParams.call(params)` is a lot more explicit and doesn't require us to `include` a module in the controllers.

app/services/site_configs/upsert.rb
```diff
@@ -0,0 +1,135 @@
...
+      end
+      self
+    end
+
+    def success?
+      @success
+    end
+
+    def clean_up_params
+      %i[sidebar_tags suggested_tags suggested_users].each do |param|
```
citizen428, line 61: **suggestion (non-blocking):** I'd probably move this into a constant if only to be consistent with what we do almost everywhere.

app/services/site_configs/upsert.rb
```diff
@@ -0,0 +1,135 @@
...
+      @errors = []
+      clean_up_params
+      validate_inputs
+      validate_emoji
+      validate_image_urls
+
+      return self if @errors.flatten.any?
+
+      @success = true
+      @configs.each do |key, value|
```
citizen428, line 42: **suggestion (non-blocking):** I'd probably move this out into a private helper like `upsert_configs`, similar to the methods at the beginning of this one.

app/services/site_configs/upsert.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    def validate_image_urls
+      image_params = @configs.slice(*IMAGE_FIELDS).to_h
+      @errors << image_params.filter_map do |field, url|
+        "#{field} must be a valid URL" unless url.blank? || valid_image_url(url)
+      end
+    end
+
+    def update_enabled_auth_providers(value)
+      enabled_providers = []
```
citizen428, line 89: **suggestion (non-blocking):** I think you can combine this line and the next, which would make it more similar to other methods in this service.

```ruby
enabled_providers = value.split(",").filter_map do |entry|
  entry unless invalid_provider_entry(entry)
end
```

app/services/site_configs/upsert.rb
```diff
@@ -0,0 +1,135 @@
...
+      end
+      SiteConfig.public_send("authentication_providers=", enabled_providers) unless
+        email_login_disabled_with_one_or_less_auth_providers(enabled_providers)
+    end
+
+    def email_login_disabled_with_one_or_less_auth_providers(enabled_providers)
+      !SiteConfig.allow_email_password_login && enabled_providers.count <= 1
+    end
+
+    def invalid_provider_entry(entry)
```
citizen428, line 101: **suggestion (non-blocking):** There's a lot of negation here, how about this instead?

```ruby
def valid_provider_entry(entry)
  entry.present? && Authentication::Providers.available.map(&:to_s).include?(entry) &&
    provider_keys_present(entry)
end

def provider_key_present(entry)
  SiteConfig.public_send("#{entry}_key").present? && SiteConfig.public_send("#{entry}_secret").present?
end
```

Then we just need to change the condition from `unless invalid_provider_entry` to `if valid_provider_entry` and everything should be a bit easier to read.

rhymes:
@citizen428 @benhalpern now that `SiteConfig` supports validations, we might want to restructure that part of the code https://github.com/huacnlee/rails-settings-cached#validations

app/controllers/api/v0/admin/configs_controller.rb
```diff
@@ -2,12 +2,26 @@ module Api
...
+        include SiteConfigParams
+
         before_action :authenticate_with_api_key_or_current_user!
         before_action :authorize_super_admin
 
         def show
           @site_configs = SiteConfig.all
         end
+
+        def update
```
rhymes, line 14: this needs

```ruby
skip_before_action :verify_authenticity_token, only: %i[update]
```

at the top.

If you test it manually it won't work because it checks for a CSRF token

rhymes (commented):
I left a comment about the missing `skip_before_action` but I also noticed something else:

both `GET /api/admin/config` and the PUT return value, only list keys that have been set in the past.

On my local brand new database I only get back a handful of keys, not the entire `SiteConfig`:

```shell
✗ curl -H "api-key: API_KEY" localhost:3000/api/admin/config  | jq

{
  "waiting_on_first_user": false,
  "mascot_user_id": 13,
  "authentication_providers": [
    "github"
  ],
  "community_name": "new",
  "admin_action_taken_at": "2020-12-22T10:46:48.077Z"
}
```

I set `authentication_providers` manually in the console, `community_name` with curl and `admin_action_taken_at` is set by `SiteConfigs::Upsert`.

I guess the "show" action/template should return the avaiable keys with their default values, not only those that have been altered.

benhalpern:
> now that SiteConfig supports validations, we might want to restructure that part of the code

This is great! @citizen428 @rhymes

Since this PR already changes a lot of stuff I think it's best to keep this as more of a refactor of the controller, but I'd be excited to take that on next. Better validation will go a long way to reducing creator pain from the bottom up.

app/services/site_configs/upsert.rb
benhalpern: 👍

app/services/site_configs/upsert.rb
benhalpern: 👍

app/services/site_configs/upsert.rb
benhalpern: 👍

benhalpern:
> I guess the "show" action/template should return the available keys with their default values, not only those that have been altered.

@rhymes  Since show action is already in production, why don't we leave that conversation for a separate issue/PR and follow the behavior as it is now?

app/controllers/api/v0/admin/configs_controller.rb
benhalpern: 👍

rhymes:
@benhalpern sure!

rhymes (approved):
Tested the functionality manually and it's working

fdoxyz (approved):
I gave this a try locally and looks good to me

PR #11701, opened 2020-12-02
Logo on /new

ludwiczakpawel:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

There was a bug causing logo to disappear on `/new` view in certain situations. The real reason is actually a combination of weird SVG used as logo with some of our caching/serviceworkers/shell architecture affecting the `/new` view (and other _workarounds_ we've done in the past). Surprisingly, it can be fixed with a bit of CSS which is not ideal but is 100x cheaper than proper solution.

- Closes https://github.com/forem/InternalProjectPlanning/issues/236
- Related https://github.com/forem/forem/pull/11622

Currently, when you use [crappy SVG](https://gist.github.com/ludwiczakpawel/ca34b08fd2d9732413f8260f5a257809) as a logo it will not show up on `/new`. With this bugfix, it should show up. Everything else should work as currently/expected.

## Added tests?

- [ ] Yes
- [x] No, not needed.

## Added to documentation?

- [ ] README
- [x] No documentation needed


app/views/articles/_v2_form.html.erb
```diff
@@ -3,8 +3,29 @@
...
+  }
+
+  .crayons-header .site-logo,
+  .crayons-header .site-logo * {
+    visibility: visible;
+    position: static;
+    pointer-events: auto;
+    opacity: 1;
+    clip: auto;
+    overflow: visible;
```
ludwiczakpawel, line 25: This is probably an overkill, but I wanna be sure it will work :D

ludwiczakpawel: probably worth explaining a bit:

previously we applied `display: none;` to `.crayons-header` - which basically hides the header and everything inside it (including logo). so this piece of code is meant to replace the `display: none;` styling in a way.. BUT without affecting the logo...

ludwiczakpawel:
**Just for reference, here's what actually happened (below I'm pasting the results of investigation @juliannatetreault and I have done):**

## The Problem.

For context: `/new` is a special snowflake in terms of design - it doesn’t have classic header and footer like rest of our views. Because of all our caching systems etc we couldn’t really say “_hey, don’t display header and footer on this view_” - [we had to hide these with CSS](https://github.com/forem/forem/blob/master/app/views/articles/_v2_form.html.erb#L6-L9) - this is very not ideal but it works 🤷‍♂️ .

Why does it matter? Because header (including logo) is still in the DOM but it’s just hidden with CSS. And we then also [“replicate” the very same logo here](https://github.com/forem/forem/blob/master/app/javascript/article-form/components/Header.jsx#L20). So we technically have the same code for the same logo twice in the DOM - one in hidden header and other one in that JSX...

There are many methods and techniques of displaying SVG assets on the site. One of them is by referencing particular elements (also being part of SVG) - basically the way it works is this:

```
<svg>
  <defs>
    <image id="image_id" ...>
  </defs>
  <rect fill=image_id" />
</svg>
```

^it’s a simplified version of what we have… Basically what happens is:

There’s and `image#image_id`  element which contains “source” of the asset.. It has unique ID so it can be referenced elsewhere. But it doesn’t display anything yet.

Then there’s `rect` element which basically is referring that `image_id` - in other words it says “_hey, use whatever is assigned to image_id  and display it inside me_”…

This technique is useful especially when you want to minimize your code so you only keep “source svg” in one place and then reference it elsewhere multiple times (think of [tables with multiple rows, each with some action icons on the right](https://d.pr/i/zQv9MM)).. It's not really useful in our case since we want to display logo only once. But Figma (or whatever other tool) doesn’t know it when exporting SVG and it structures exported code in such a way (also, it’s probably the only way to export PNG in SVG 🤷‍♂️ ).

So we end up having two exactly the same SVG codes. The problem is that both SVGs contain elements with the same ID and ID, by definition, must be unique. So what happens is: our 2nd SVG code (the one from JSX) is trying to reference element with certain ID.. That element unfortunately exists in the both SVG codes (in hidden header and in our JSX) but the one that is being referenced is the 1st one - the one in header (simply because it’s first).. And again - the one in header is hidden and since its hidden in header it will also be hidden where its being referenced :) 

---

## Other solutions we've explored

1. We could technically remove the logo from header with JS but then we will have to bring it back when you leave `/new` 
 (because caching and service workers) page…
2. We could try to parse `siteLogo` prop and make some script that will take all the IDs and try to append a suffix to them like `${id}__cloned` or something..
3. We could find a way to actually load alternative layout on `/new` page, a layout without that header and footer so we wouldn’t have to hide these with CSS.. that solution feels like the most proper one, but the caching/serviceworkers/[enter other magic keywords] is the reason why we did it this way in the first place.
4. We could ALWAYS serve logo as a preact component, similar to what we do with search form..  and then, in certain scenarios (like on the /new view) we could use preact portals to move it elsewhere… this is exactly what Nick described in his post yesterday: https://dev.to/nickytonline/testing-preact-react-portals-with-testing-library-52ja

Solution 3 feels like the most bulletproof and "proper" but it would also be quite expensive in terms of resources and changes that need to made (caching, serviceworkers, layouts, partials...). 

---

## The actual fix (hack)

Hide Header in different, less destructive way than `display: none` and make sure it doesn't affect the logo in header...

spec/system/user_uses_the_editor_spec.rb
```diff
@@ -22,6 +22,16 @@ def fill_markdown_with(content)
     within("#article-form") { fill_in "article_body_markdown", with: content }
   end
 
+  describe "Viewing the editor", js: true do
+    it "renders the logo_svg or Community name as expected" do
+      visit "/new"
+      expect(page).to have_css(".site-logo")
+      within(".truncate-at-2") do
+        expect(page).to have_text("DEV(local)")
```
juliannatetreault, line 30: I observed that the spec renders the `community_name` rather than the logo, so this test ensures that the `.site_logo` CSS is on the page, in addition to the truncated `community_name` that is actually rendered in the test.

juliannatetreault (approved):
The `logo_svg` showed as expected in the header of the editor, so pending a green Travis build, I think this is good-to-go! 🎉

fdoxyz (approved):
I tried this out with `master` and it didn't show the crappy logo, but the fix shows the logo. LGTM

juliannatetreault:
> I tried this out with `master` and it didn't show the crappy logo, but the fix shows the logo. LGTM
> 
> Great job to both of you on the thorough research of the obscure SVG problem :D

Also, extra shoutout to @nickytonline who went down a debugging rabbit hole with me yesterday when I thought it was a Preact-based issue and _not_ a CSS-based issue! 🎉

app/views/articles/_v2_form.html.erb
nickytonline: refactor (non-blocking): This definitely works, but is probably overkill like you said @ludwiczakpawel.

Given that the children of `.crayons-header` are absolutely positioned off the screen and `pointer-events` are set to `none`, you probably only need these properties:

```css
.crayons-header,
  .crayons-header * {
    position: absolute;
    left: -9999em;
    top: -9999em;
    pointer-events: none;
}
```

and then the reset like so.

```
 .crayons-header .site-logo,
 .crayons-header .site-logo * {
  position: static;
}
```

nickytonline (approved):

app/views/articles/_v2_form.html.erb
ludwiczakpawel: We need a bit more here so users can't access it with keyboard navigation... But yea, definitely can be simplified a little. I like to over-engineer things :D

PR #11706, opened 2020-12-02
Split ReCaptcha service and use verbs instead of pronouns

fdoxyz:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Based on feedback received on #10502 ([see this conversation](https://github.com/forem/forem/pull/10502#discussion_r533615758)) this PR splits the `ReCaptcha` service in two.

I originally didn't notice [the service object docs](https://docs.forem.com/backend/service-objects/) that recommend we use verbs instead of nouns.

Also the approach here is a bit cleaner and less complex, i.e. it doesn't initialize an object and return `self` (this is useful sometimes so methods can be chained but not for this simple service object).

- If the `SiteConfig` keys for email registration with reCAPTCHA are enabled you should run into the reCAPTCHA when trying to register.
- Try to report abuse (a comment on a Post) and if the reCAPTCHA SiteConfig keys are not configured you shouldn't see the reCAPTCHA appear in the form
   - Once keys are configured the reCAPTCHA should appear when logged out or the user has a vomit/too recent/banned
   - With keys configured the reCAPTCHA shouldn't appear when trying to report abuse as a Mod/Admin/"clean and non-recent user"

None. This only affects reCAPTCHA logic that currently exists

## Added tests?

- [ ] Yes
- [x] No, and this is why: Existing tests cover this, they were just split in two (just like the service object)

## Added to documentation?

- [ ] README
- [x] No documentation needed

same but different


app/services/re_captcha/check_enabled.rb
```diff
@@ -0,0 +1,37 @@
+# This service encapsulates the logic related to validating if reCAPTCHA is
+# enabled in the current Forem instance. The decision is based on making
+# sure the necessary SiteConfig keys are available and also on the user
+# object passed in.
+#
+# Example use: ReCaptcha::CheckEnabled.call(current_user) => true/false
```
rhymes, line 6: ❤️

app/services/re_captcha/check_enabled.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def initialize(current_user)
+      @current_user = current_user
+    end
+
+    def call
+      # recaptcha will not be enabled if site key and secret key aren't set
+      return false unless keys_configured?
+      # recaptcha will always be enabled when not logged in
+      return true if @current_user.nil?
```
rhymes, line 21: can we invert these? This way we don't risk hitting the DB with `keys_configured?` for anonymous users if we already know it's always enabled for them

app/services/re_captcha/check_enabled.rb
```diff
@@ -0,0 +1,37 @@
...
+      # recaptcha will not be enabled for trusted/admin/tag mod users
+      return false if @current_user.auditable?
+      # recaptcha will be enabled if the user has been banned
+      return true if @current_user.banned
+
+      # recaptcha will be enabled if the user has a vomit or is too recent
+      @current_user.vomitted_on? || @current_user.created_at > 1.month.ago
```
rhymes, line 28: non blocking tip: `@current_user.created_at.after?(1.month.ago)`

rhymes (commented):
Good call @fdoxyz, left a note about checking order

app/services/re_captcha/check_enabled.rb
fdoxyz: Interesting idea but I don't think we can, because if the user is `nil` but the reCAPTCHA keys are not configured we can't show it (it will be broken). The user won't be able to pass the reCAPTCHA and it will render as an error in the form, somewhat like this:

Screen Shot 2020-12-02 at 11 09 32

In other words, if we want to render the reCAPTCHA prompt we need to check the `SiteConfig` keys are configured.

When rendering the reCAPTCHA (in the view) Rails will hit the DB regardless (in order to pass in the site_key config) so they will be cached (AFAICT `SiteConfig` will do that internally) so at least it won't be hit multiple times for the same keys, right?

rhymes: makes sense! thanks for the explanation :)

rhymes (approved):

app/services/re_captcha/check_enabled.rb
fdoxyz: love the activesupport magic ✨

app/services/re_captcha/check_enabled.rb
```diff
@@ -0,0 +1,37 @@
...
+      @current_user = current_user
+    end
+
+    def call
+      # recaptcha will not be enabled if site key and secret key aren't set
+      return false unless keys_configured?
+      # recaptcha will always be enabled when not logged in
+      return true if @current_user.nil?
+      # recaptcha will not be enabled for trusted/admin/tag mod users
+      return false if @current_user.auditable?
```
joshpuetz, line 23: **Suggestion**: explicitly spell out which roles you do or don't mean here. `.auditable?` strikes me as a way of asking "does the current user have admin rights?" or "is the current user just a regular user?"

joshpuetz (approved):
Looks great, one non-blocking suggestion

app/services/re_captcha/check_enabled.rb
fdoxyz: Thanks for the suggestion! I added it because it will definitely make it more clear to understand when reading the code

PR #11710, opened 2020-12-02
Bug Fix: First Published Post Not Showing Properly in the Feed

juliannatetreault:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR resolves a bug that occurs when the first published post on a Forem does _not_ contain a `main_image`, resulting in the post not being "featured." Prior to these changes, if the first, published post on a Forem did not have a cover image, resulting in the post not being considered "featured," the post would not display in the feed until it was updated to contain a `main_image`, was set as `featured: true`, or another post was published _in addition to_ the first published post.

To resolve the bug, a ternary operator was added to `app/javascript/articles/Feed.jsx` to determine where the splicing of the  `feedItems` array (which separates out `featuredStories` from `subStories`) should occur. This proved necessary because when the first published post _does not_ have a `main_image`, it results in an empty array. This then results in a `splice` that removes the _last_ element in the array, ultimately causing no posts to show in the feed because we've removed the posts from the `feedItems` array. 

An alternative approach to resolving this bug would be to add a conditional to lines `29`-`31` in `Feed.jsx` that provides a fallback for when the first published post does _not_ contain a cover image. If my reviewers feel that that approach is more appropriate, I am happy to explore that option!

Closes https://github.com/forem/InternalProjectPlanning/issues/155

**To QA this PR, please do the following**:
- Navigate to `localhost` on your `main` branch and delete all of the posts that you currently have in a rails console:
```
Article.destroy_all
```
- Now that you've deleted all of your posts, create a _new_ post without a `main_image` and publish it:
Screen Shot 2020-12-02 at 12 52 26 PM
Screen Shot 2020-12-02 at 12 52 47 PM

- Observe that the post does _not_ show in the feed:
Screen Shot 2020-12-02 at 12 54 35 PM

- Switch from your `main` branch to my branch and observe that the post now shows in the feed:
Screen Shot 2020-12-02 at 12 55 00 PM

- Try to break things! 🛠️ 

N/A

## Added tests?

- [ ] Yes
      have not been included_
- [x] Question: it looks like the best place to test this would be in `app/javascript/articles/_tests_/Article.test.jsx`, is that correct? If so, I would assume that I might need to add a new test because the existing spec did not catch this bug, but we already test that both an Article and a Featured Article render in the feed... 🤔 

## Added to documentation?

- [ ] README
- [x] No documentation needed

N/A

The Office: Erin and Kelly looking at a computer

### DIFF
--- a/app/javascript/articles/Feed.jsx
+++ b/app/javascript/articles/Feed.jsx
@@ -33,7 +33,8 @@ export const Feed = ({ timeFrame, renderFeed }) => {
 
         // Remove that first one from the array.
         const index = feedItems.indexOf(featuredStory);
-        feedItems.splice(index, 1);
+        const deleteCount = featuredStory ? 1 : 0;
+        feedItems.splice(index, deleteCount);
         const subStories = feedItems;
         const organizedFeedItems = [featuredStory, subStories].flat();
 
--- a/app/javascript/articles/Feed.jsx
+++ b/app/javascript/articles/Feed.jsx
@@ -27,11 +27,15 @@ export const Feed = ({ timeFrame, renderFeed }) => {
         const feedItems = await getFeedItems(timeFrame);
 
         // Ensure first article is one with a main_image
+        // This is important because the featuredStory will
+        // appear at the top of the feed, with a larger
+        // main_image than any of the stories or feed elements.
         const featuredStory = feedItems.find(
           (story) => story.main_image !== null,
         );
 
-        // Remove that first one from the array.
+        // Remove that first story from the array to
+        // prevent it from rendering twice in the feed.
         const index = feedItems.indexOf(featuredStory);
         const deleteCount = featuredStory ? 1 : 0;
         feedItems.splice(index, deleteCount);


app/javascript/articles/Feed.jsx
```diff
@@ -33,7 +33,8 @@ export const Feed = ({ timeFrame, renderFeed }) => {
 
         // Remove that first one from the array.
```
juliannatetreault, line 34: A couple of questions for my reviewers:

- Why do we, just a couple lines up (lines 28-31 in the code show below), "ensure first article is one with a main_image"? Why is it important that the first article contains a cover image?
```
        // Ensure first article is one with a main_image
        const featuredStory = feedItems.find(
          (story) => story.main_image !== null,
        );
```

- Why do we "remove that first one from the array"? Is it to separate the `featuredStory` from the `subStories` or is there another reason?

vaidehijoshi: Hey @juliannatetreault! Based on the fact that we are assigning that story we are assigning on those lines is saved to a const called featuredStory, the code tells me that the featured story actually IS the first article. And, since it is featured at the top of the feed, we want to always ensure that it has an image (the featured story appears at the top of the feed with a larger image than other stories/elements on the feed).

As for your second question, the reason we remove it is because we don’t want the featured story to appear again in the feed—if we didn’t remove it from the list of articles, then it would render twice which would be confusing to users :)

juliannatetreault: Thanks for answering my questions, @vaidehijoshi 😊

vaidehijoshi: Anytime, @juliannatetreault! 😃

rhymes: It'd be AWESOME if this explanation found itself in the inline comments 🔥

rhymes (approved):
Tested! Left a note about adding an explanatory comment about what's going on for future reference

app/javascript/articles/Feed.jsx
juliannatetreault: That's an excellent idea, @rhymes! I'll add an abridged explanation as a comment. 😄

fdoxyz (approved):
Tested locally and seems to be working well from my end! 😃

app/javascript/articles/Feed.jsx
vaidehijoshi: Great call to improve on this comment! I'm sure other folks would have had the same question when looking at this code, so this will be super helpful to our future selves 💯

PR #9870, opened 2020-08-19
Display message from the back end when leaving a channel

AlbertoPdRF:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The messages passed from the back end to the front end when leaving a chat channel are never displayed, this PR addresses that.

## Related Tickets & Documents

I intended to fix this together with #9502, which got fixed by #9804.

## QA Instructions, Screenshots, Recordings

Leave a chat channel and see a snackbar notification appear.

## Added tests?

- [ ] yes
- [x] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

## [optional] Are there any post deployment tasks we need to perform?

Nope.

## [optional] What gif best describes this PR or how it makes you feel?

Adjustment GIF

Issue #9502, opened 2020-07-24
[Connect] User is not redirected after leaving a room, leading to ambiguity and blank error


wafflewitch:
**Describe the bug**

In both full screen (/connect route) and sidecar view, leaving a room does not result in redirecting the user away from that room. This leaves the user in an ambiguous situation where they can still see the room they have left. However, if the user tries to send a message in the left room, a blank error appears.

**To Reproduce**

**Expected behavior**

Upon leaving a room, the user should be redirected (or the page should be refreshed) to officially remove the user from the room and prevent the ambiguous situation described above. 

Also, the error should be improved to display a message. 

Finally, I'm also wondering if the error is supposed to fit the user's chosen color theme... ?)

**Screenshots**

Full screen view:
![dev_connect_full-scr_left-room-empty-err](https://user-images.githubusercontent.com/27379918/88437048-405b4580-cdcb-11ea-8960-e985bf8ebab6.gif)

Sidecar view:
![dev_connect_sidecar_left-room-empty-err](https://user-images.githubusercontent.com/27379918/88437329-de4f1000-cdcb-11ea-9db3-3696635dbb38.gif)

**Desktop (please complete the following information):**

- OS: Mac
- Browser: Chrome
- Version: 83.0.4103.116

juliannatetreault:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @forem/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/forem/forem/blob/master/CONTRIBUTING.md).

AlbertoPdRF:
I'm looking into this one! I may come back with a few questions later on 😛

nickytonline:
Sounds good @AlbertoPdRF. I've assigned it to you.

AlbertoPdRF:
First of all, sorry for the late follow up! So, if I got everything correctly, when a user clicks on the button to leave a channel, and the request succeeds, [this is the code that gets executed](https://github.com/forem/forem/blob/master/app/javascript/chat/ChatChannelSettings/ChatChannelSettings.jsx#L311):

[...]

PR #11711, opened 2020-12-02
Display message from the back end when leaving a channel

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

_Supersedes: https://github.com/forem/forem/pull/9870 (@jacobherrington)_

The messages passed from the back end to the front end when leaving a chat channel are never displayed, this PR addresses that.

I intended to fix this together with #9502, which got fixed by #9804.

Leave a chat channel and see a snackbar notification appear.

## Added tests?

- [ ] yes
- [ ] no, because I need help

## Added to documentation?

- [ ] docs.forem.com
- [ ] readme
- [x] no documentation needed

Nope.

Adjustment GIF

### DIFF
--- a/app/javascript/chat/ChatChannelSettings/ChatChannelSettings.jsx
+++ b/app/javascript/chat/ChatChannelSettings/ChatChannelSettings.jsx
@@ -308,7 +308,12 @@ export default class ChatChannelSettings extends Component {
     const { currentMembership } = this.state;
     if (actionStatus) {
       const response = await leaveChatChannelMembership(currentMembership.id);
+      const { message } = response;
       if (response.success) {
+        this.setState({
+          successMessages: response.message,
+          errorMessages: null,
+        });
         this.props.handleLeavingChannel(currentMembership.id);
       } else {
         this.setState({
@@ -316,6 +321,7 @@ export default class ChatChannelSettings extends Component {
           errorMessages: response.message,
         });
       }
+      addSnackbarItem({ message });
     }
   };
 
--- a/app/javascript/chat/ChatChannelSettings/ChatChannelSettings.jsx
+++ b/app/javascript/chat/ChatChannelSettings/ChatChannelSettings.jsx
@@ -311,7 +311,7 @@ export default class ChatChannelSettings extends Component {
       const { message } = response;
       if (response.success) {
         this.setState({
-          successMessages: response.message,
+          successMessages: message,
           errorMessages: null,
         });
         this.props.handleLeavingChannel(currentMembership.id);


app/javascript/chat/ChatChannelSettings/ChatChannelSettings.jsx
```diff
@@ -308,14 +308,20 @@ export default class ChatChannelSettings extends Component {
     const { currentMembership } = this.state;
     if (actionStatus) {
       const response = await leaveChatChannelMembership(currentMembership.id);
+      const { message } = response;
       if (response.success) {
+        this.setState({
+          successMessages: response.message,
```
citizen428, line 314: You destructure `response` a few lines above (`const { message } = response;`), shouldn't you be using `message` here?

```suggestion
          successMessages: message,
```

jacobherrington: That makes sense!

This isn't actually my code (it's a rebase of a stale PR), so I'm not sure, I'll test that suggestion and commit it if it works as expected.

citizen428 (approved):

app/javascript/chat/ChatChannelSettings/ChatChannelSettings.jsx
```diff
@@ -308,14 +308,20 @@ export default class ChatChannelSettings extends Component {
     const { currentMembership } = this.state;
     if (actionStatus) {
       const response = await leaveChatChannelMembership(currentMembership.id);
+      const { message } = response;
       if (response.success) {
+        this.setState({
+          successMessages: message,
+          errorMessages: null,
```
citizen428, line 315: **nitpick (non-blocking):** Should this maybe be `[]` instead? Personally I'm in favor of always returning the same datatype if possible, Maybe @nickytonline can weigh in on this too.

nickytonline: It appears to be set to `null` whenever it's reset, but I agree it makes sense to keep the same type so that you don't need to make checks to see if it's `null` before doing an array operation.

Having said that, I looked through this whole component, and the `errorMessages` only seem to ever be getting set in component state but never used. Maybe I'm missing something. I noticed the same for the `successMessages`. I'm going to pull this locally.

nickytonline: Alright, I pulled this locally and the type for `errorMessages` is a string, not an array, so `null` probably makes sense, but once again, having looked at the code and running it locally, I see `successMessages` and `errorMessages` get set in component state, but never actually used.

The only usage I see is when the response message gets destructured which has nothing to do with the component state. 🤔

nickytonline: So I did another test where I removed `errorMessages` and `successMessages` from the component's code  and the application still works fine. If there is an an error, the error message is displayed in the Snack bar.

image

So although not the original intention of this PR, but from what I can tell and see, `errorMessages` and `successMessages` are never used in the code aside from being set in component state.

citizen428: Ah, my bad, the name suggested an array and I didn't get around to check out the code.

nickytonline:
@jacobherrington, just following up on this in regards to my previous comments.

msarit:
@jacobherrington Hey Jacob, just pinging you re: Nick's comments on the file `app/javascript/chat/ChatChannelSettings/ChatChannelSettings.jsx:315`

rhymes (approved):

jacobherrington:
I think we should merge this as-is and address removing the unused values from the component's state in another PR; this fixes something immediate and I'd like to get @AlbertoPdRF's code into the application :smile:

Issue #7139, opened 2020-04-07
A11y analysis


tomByrer:
From here: https://github.com/a11y-reviews/a11y.reviews/blob/master/index.md#devto

> 31 January 2019... The overall interface contains many links without accessible names, some of them apparently only existing to enlarge the click area - see the article teaser listing. Said teaser listing also infinite scrolling.

> Regarding keyboard accessibility: outline on focus is not disabled via CSS, but some elements are reachable only on hover (example: the "Follow" control inside a tag list item). Said control, again, is an anchor element with a `href="#"` but should be a button.

benhalpern:
Thanks for pointing this out. WCAG compliance is where we want to be and being on this list is a really bad thing. We'll take this as a note to prioritize closing the loop on some of our accessibility cases like described here with the follow button only accessible via hover. Thank you very much for pointing this out.

tomByrer:
Good news; your site has a screenshot with an AAA rating in an article about accessibility at Google's web.dev:

https://web.dev/a11y-tips-for-web-dev/

marcysutton:
I've been looking to see if any of these items are still outstanding and here's what I found:

- Infinite scrolling is definitely still an issue, but it could be tracked separately in https://github.com/forem/forem/issues/5670.
- Regarding links without accessible names, I remember those as well but I believe most have been fixed. I'm seeing two right now on the DEV homepage that are both coming from a sponsorship widget. I've raised that as an issue internally to start, and will file a separate issue for it if needed.
- I believe the hover-only follow control issue has been fixed, but I'm not 100% sure I was looking at the right thing. @nickytonline or @benhalpern do you know what they meant by "the "Follow" control inside a tag list item"?

I did [create a PR](https://github.com/forem/forem/pull/11721) to address some semantics things I found, like heading order and some missing alt text / form labels. But I think we're near being able to close this issue.

cmgorton:
Thank you @tomByrer for opening this issue. Since we have Marcy working on identifying a11y issues and creating other issues/PRs to monitor those I am going to go ahead and close this issue.

PR #8739, opened 2020-06-17
Upgrade to Preact 10.4.4

nickytonline:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

**THE ONLY REASON THIS IS A DRAFT PR IS SO THAT NO ONE SQUASHES THIS ON MERGE. WE'D LIKE THE TWO COMMITS TO BE ADDED TO THE GIT HISTORY TO REFLECT THE WORK RIDHWANA AND MYSELF DID.**

**DO NOT BE INTIMIDATED BY THE NUMBER OF FILES IN THIS PR. ALL SNAPSHOTS ASIDE FROM CRAYON COMPONENTS HAVE BEEN DELETED**

Preact has been upgraded to the latest version of Preact X, 10.4.4. This gives us hooks, componentDidCatch, fragments etc. See the [official release](https://preactjs.com/guide/v10/whats-new/) for the whole rundown.

Here is the [official migration guide](https://preactjs.com/guide/v10/upgrade-guide) for Preact 8.x to X if you're curious why some changes have been made.

We've done a smoke test of the site on our locals. Onboarding worked fine, search is good and the site seems to behaving normally.

The changes on this PR include:
- preact-render-to-json and preact-render-spy testing tools have been removed due to lack of support for Preact X. 
- [preact-testing-library](https://github.com/testing-library/preact-testing-library) is now the preferred testing library for Preact components. 
- Preact X includes compat by default, so we've removed preact-compat and we already use the ‘prop-types’ package so there was no changes required.
- Snapshot testing has been deprecated for all components except for design system (crayons) components.

The majority of this PR is migrating tests, so we have lots of tests! 😺

### How to Test this PR

Run `yarn test` or `npm test`. This will validate that we have successfully migrated all the tests to `preact-testing-library` if all tests pass.

Although this is a Preact upgrade, the upgrade part was pretty minor. Regardless, the application should be smoked tested.

## Related Tickets & Documents

#5639, #8376, #8224

## Added tests?

- [x] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [x] docs.dev.to
- [ ] readme
- [ ] no documentation needed

## [optional] What gif best describes this PR or how it makes you feel?

Some cool X gif

PR #11721, opened 2020-12-03
Headings and alt text roundup

marcysutton:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [X] Optimization
- [ ] Documentation Update

## Description

In order to address https://github.com/forem/forem/issues/7139, I went through the items mentioned to see if there were any fixes that still hadn't been made. I found some issues with the heading order and some missing alt text, which I did as part of this PR. I'll plan to follow up with another PR to address landmark elements.

TODO: I have one outstanding question about conditional heading markup on the homepage, which I'll add as a comment below.

https://github.com/forem/forem/issues/7139

For heading order, I used View Document Outline in the [Web Developer extension](https://addons.mozilla.org/en-US/firefox/addon/web-developer/) in Firefox and found some issues on multiple pages:

DEV homepage in Web developer Document Outline

Here is the homepage heading order after this PR is applied:

Home page in Web Developer Document Outline with fixes applied

And on the profile page:

DEV profile page in Web Developer Document outline

Here is the order after this PR is applied:

Profile page in Web Developer Document Outline with fixes applied

This PR includes accessibility improvements, namely missing and out-of-order headings. It also adds some missing alt attributes and a textarea label to improve the experience in a screen reader.

## Added tests?

- [ ] Yes
- [X] No, and this is why: There aren't currently automated tests for heading order. That could be cool though.

## Added to documentation?

- [ ] README
- [X] No documentation needed

alt_text


app/views/articles/index.html.erb
```diff
@@ -32,11 +32,12 @@
       data-feed="<%= params[:timeframe] || "base-feed" %>"
       data-articles-since="<%= Timeframer.new(params[:timeframe]).datetime&.iso8601 %>">
 
+    <h1 class="visually-hidden-header"><%= community_name %></h1>
```
marcysutton, line 35: One outstanding question I have is about this h1 that should only live on the homepage. It currently sits outside of `<main>` so it can be rendered before the sidebar, which contains headings that throw off the order as the h1 should come first. Because this heading needs to live inside of a landmark element, I tried conditionally rendering an h1 in `_logo.html.erb` by checking for the homepage and rendering different markup, but it didn't update on the client-side when clicking through pages. I could use some help getting a solution figured out! Maybe `content_for` in Rails to conditionally render in the header on the homepage?

ludwiczakpawel: > I tried conditionally rendering an h1 in _logo.html.erb by checking for the homepage and rendering different markup, but it didn't update on the client-side when clicking through pages.

Yeap, this may not work as you expect since we do lots of weird things around caching and serviceworkers to speed up loading pages... It's pretty tough to overcome TBH, not sure if worth the effort in this case...

Actually... What do you think about removing this element entirely? It's kinda pointless, right? It was pointless earlier when it was saying "Articles" because of its placement (as well as the fact that we do have normal label right below it saying "Posts") and it is a bit pointless now too when it returns community name.. users probably know where they are (it's basically name of the site) and also that element is pretty close to Header where we also have logo and community name so the screen reader would read it anyway, right?

app/views/articles/_sidebar_nav.html.erb
```diff
@@ -50,7 +50,7 @@
     <header class="p-2 pr-0 flex items-center justify-between">
       <h3 class="crayons-subtitle-3">My Tags</h3>
       <a href="/dashboard/following_tags" class="crayons-btn crayons-btn--icon crayons-btn--ghost-dimmed" aria-label="Customize tag priority" title="Customize tag priority">
-        <%= inline_svg_tag("cog.svg", aria: true, class: "crayons-icon") %>
+        <%= inline_svg_tag("cog.svg", aria: false, class: "crayons-icon") %>
```
ludwiczakpawel, line 53: ~~When `aria` attribute is set to `false` is it the same as not having aria attribute at all? Or setting `false` is actually meaningful?~~

I've just checked - when we set `aria` attribute to `false` the output is *not* `aria="false"` in actual markup. It just doesn't show up... In that case, I think better practice would be to remove that attribute entirely instead of changing value to `false`. Unless I'm missing something..?

app/views/articles/tag_index.html.erb
```diff
@@ -8,7 +8,7 @@
     <header class="tag-header crayons-card branded-4" style="border-top-color: <%= HexComparer.new([@tag_model.bg_color_hex || "#0000000", @tag_model.text_color_hex || "#ffffff"]).brightness(0.88) %>">
       <div class="flex gap-4">
         <% if @tag_model.badge_id && @tag_model.badge %>
-            <img src="<%= optimized_image_url(@tag_model.badge.badge_image_url, width: 64) %>" class="" style="transform: rotate(<%= -25 + (@tag_model.id.digits.first * 5) %>deg); width: 64px; height: 64px;" />
+            <img src="<%= optimized_image_url(@tag_model.badge.badge_image_url, width: 64) %>" alt="" class="" style="transform: rotate(<%= -25 + (@tag_model.id.digits.first * 5) %>deg); width: 64px; height: 64px;" />
```
ludwiczakpawel, line 11: Wouldn't it make sense to make it like `alt="<%= @tag %>'s icon"` or something?

app/views/tags/index.html.erb
```diff
@@ -54,7 +54,7 @@
         <% end %>
 
         <% if tag.badge_id && tag.badge %>
-          <img src="<%= optimized_image_url(tag.badge.badge_image_url, width: 180) %>" style="width: 64px; height: 64px; transform: rotate(<%= -25 + (tag.id.digits.first * 5) %>deg);" class="right-1 bottom-1 absolute" />
+          <img src="<%= optimized_image_url(tag.badge.badge_image_url, width: 180) %>" style="width: 64px; height: 64px; transform: rotate(<%= -25 + (tag.id.digits.first * 5) %>deg);" class="right-1 bottom-1 absolute" alt="" />
```
ludwiczakpawel, line 57: Wouldn't it make sense to make it like alt="<%= tag %>'s icon" or something?

ludwiczakpawel (changes requested):
Just few very minor things.

app/views/articles/index.html.erb
marcysutton: That makes sense about the caching, and is consistent with what I saw in the application! Though, this element is not pointless as the community name is only part of the heading structure if we put it in a heading...and that seems important, at least on the homepage. The site logo in the header is a link and won't end up in the heading structure unless changed. 

The easiest way forward, I think, is it put the homepage h1 back into the `<main>` element and let the couple of headings in the sidebar come before it. Surprisingly, this resource from the W3C WAI has an example where the h1 doesn't come first: https://www.w3.org/WAI/tutorials/page-structure/headings/

Therefore, I'll update my PR so the homepage can include a more useful h1.

app/views/articles/_sidebar_nav.html.erb
marcysutton: I can remove it! `aria: true` was adding an `img` role to the SVG but it had no `<title>` or label, so it would fail in most accessibility checkers. The wrapping link also has an `aria-label` on it which supersedes the SVG anyway for its accessible name, so that's why I went this route.

app/views/articles/tag_index.html.erb
marcysutton: The image is decorative since the tag is described in text immediately following this element. So that's why I added an empty alt attribute, since it was likely reading the filename out before.

app/views/tags/index.html.erb
marcysutton: I concluded the image was decorative, since the tag is described in text previously in the DOM. Here's a screenshot where the JS logo follows the tag and follow button:

JavaScript tag with heading, text, button, and logo

ludwiczakpawel (approved):

rhymes (approved):
I merged master in to unblock buildkite and updated the Jest snapshot with `yarn test -u` to fix the failing spec

benhalpern (approved):

nickytonline:
> I merged master in to unblock buildkite and updated the Jest snapshot with `yarn test -u` to fix the failing spec

I wonder why there is a snapshot test. I removed them for all components except the design system when I upgraded us to Preact X and preact-testing-library, https://github.com/forem/forem/pull/8739. 🤔

PR #11728, opened 2020-12-03
Allow admin to add users as podcast owners (MLH Fellowship)

Manasa2850:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This allows admins to add owners to a podcast

https://github.com/MLH-Fellowship/dev.to/issues/81

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/controllers/admin/podcasts_controller.rb
+++ b/app/controllers/admin/podcasts_controller.rb
@@ -2,8 +2,8 @@ module Admin
   class PodcastsController < Admin::ApplicationController
     layout "admin"
 
-    before_action :find_podcast, only: %i[edit update fetch remove_admin add_admin]
-    before_action :find_user, only: %i[remove_admin add_admin]
+    before_action :find_podcast, only: %i[edit update fetch remove_owner add_owner]
+    before_action :find_user, only: %i[remove_owner add_owner]
 
     def index
       @podcasts = Podcast.left_outer_joins(:podcast_episodes)
@@ -34,19 +34,19 @@ module Admin
       redirect_to admin_podcasts_path
     end
 
-    def remove_admin
-      removed_roles = @user.remove_role(:podcast_admin, @podcast)
+    def remove_owner
+      removed_roles = @user.remove_role(:podcast_owner, @podcast)
       if removed_roles.empty?
         redirect_to edit_admin_podcast_path(@podcast), notice: "Error"
       else
-        redirect_to admin_podcasts_path, notice: "Removed admin"
+        redirect_to admin_podcasts_path, notice: "Removed owner"
       end
     end
 
-    def add_admin
-      role = @user.add_role(:podcast_admin, @podcast)
+    def add_owner
+      role = @user.add_role(:podcast_owner, @podcast)
       if role.persisted?
-        redirect_to admin_podcasts_path, notice: "Added admin"
+        redirect_to admin_podcasts_path, notice: "Added owner"
       else
         redirect_to edit_admin_podcast_path(@podcast), notice: "Error"
       end
--- a/app/controllers/admin/podcasts_controller.rb
+++ b/app/controllers/admin/podcasts_controller.rb
@@ -35,20 +35,18 @@ module Admin
     end
 
     def remove_owner
-      removed_roles = @user.remove_role(:podcast_owner, @podcast)
-      if removed_roles.empty?
-        redirect_to edit_admin_podcast_path(@podcast), notice: "Error"
-      else
-        redirect_to admin_podcasts_path, notice: "Removed owner"
-      end
+      @podcast_ownership = PodcastOwnership.find(params[:id])
+      @podcast_ownership.destroy
     end
 
     def add_owner
-      role = @user.add_role(:podcast_owner, @podcast)
-      if role.persisted?
-        redirect_to admin_podcasts_path, notice: "Added owner"
+      @podcast_ownership = PodcastOwnership.new(podcast_ownership_params)
+
+      if @podcast_ownership.save
+        flash[:success] = "Owner created successfully"
       else
-        redirect_to edit_admin_podcast_path(@podcast), notice: "Error"
+        flash[:error] = "User is already an owner"
+        format.html { render :index }
       end
     end
 
--- a/app/controllers/admin/podcasts_controller.rb
+++ b/app/controllers/admin/podcasts_controller.rb
@@ -36,7 +36,7 @@ module Admin
     end
 
     def add_owner
-      @podcast_ownership = PodcastOwnership.new(podcast_ownership_params)
+      @podcast_ownership = PodcastOwnership.new(podcast_id: params["id"], user_id: params["podcast"]["user_id"])
 
       if @podcast_ownership.save
         flash[:success] = "Owner created successfully"
--- a/app/controllers/admin/podcasts_controller.rb
+++ b/app/controllers/admin/podcasts_controller.rb
@@ -2,8 +2,7 @@ module Admin
   class PodcastsController < Admin::ApplicationController
     layout "admin"
 
-    before_action :find_podcast, only: %i[edit update fetch remove_owner add_owner]
-    before_action :find_user, only: %i[remove_owner add_owner]
+    before_action :find_podcast, only: %i[edit update fetch]
 
     def index
       @podcasts = Podcast.left_outer_joins(:podcast_episodes)
@@ -46,7 +45,7 @@ module Admin
         flash[:success] = "Owner created successfully"
       else
         flash[:error] = "User is already an owner"
-        format.html { render :index }
+        render json: { error: "User is already an owner" }, status: :bad_request
       end
     end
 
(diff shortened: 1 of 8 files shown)


Manasa2850:
@rhymes I get this error in Travis CI `Failure/Error: page.check("profile[looking_for_work]")`
But this isn't related to any of the files I've changed.

rhymes (changes requested):
Hi @Manasa2850, I don't think this PR is going in the correct direction.

The first step - https://github.com/forem/forem/pull/11410 - was to add a model `PodcastOwnership` which determines who owns which podcast. Let's use that :) 

Podcast ownership is not a role, it's the link between a podcast and a user determined by the existence of that connection.

Manasa2850:
@rhymes I had a meeting with Ben a couple of days back and he suggested that podcast_owner should be a role. He also suggested that I remove the podcast_admin role and call that podcast_owner

cmgorton:
> @rhymes I had a meeting with Ben a couple of days back and he suggested that `podcast_owner` should be a role. He also suggested that I remove the `podcast_admin` role and call that `podcast_owner`

@benhalpern can you weigh in on this. If I remember correctly you gave Manasa some new guidelines. Can we make sure all of us are on the same page and maybe also update the issue so it is easier for Forem team to review the PRs?

mkcode:
@Manasa2850 - After looking this over, it looks like @benhalpern may have said PodcastOwner should be a "role" in a loose sense, and not in the technical Pundit sense.

Because we have an explicit table to model PodcastOwnership, which is now merged thanks to https://github.com/forem/forem/pull/11410, we should use this to look up who "owns" a podcast.

The thing that is confusing about this, is that we want both a database table and a pundit policy. And this gets to the different purposes of each ...

Inside of our models, we want to represent podcast ownership in terms of the podcast_owners db table, and use this store the truth of 'is this user an owner of the podcast'

Then we also want a Pundit Policy for Podcasts that will use the data from the Podcast Owner database table. This is a similar example in existing codebase: https://github.com/forem/forem/blob/f7a9f0d2b42deb2d38ace3ea3c045f782e6a29fd/app/policies/organization_policy.rb#L24-L28

Notice that, inside of the OrganizationPolicy, it looks looks at the database to see if the user is an admin. Then we use the policy in the controller: https://github.com/forem/forem/blob/cc7d297371f8582e10d61e5f58b98f3cb5acc59d/app/controllers/dashboards_controller.rb#L61

So this gets to the fundamental difference. We use the database as the source of truth to store the data. And we use the pundit policy, inside of a controller, to authorize if a controller action can be completed.

Manasa2850:
I had a meeting with Chris today and we thought of implementing it this way.
@rhymes could you check if the controller is fine? If so, I'll fix the tests accordingly

Manasa2850:
forem

@rhymes I'm getting this spec error. Could you help me out with this? I'm not sure why it's looking for podcasts with `id=5` and `id=6`

Manasa2850:
@cmgorton could you review / ask someone to review this please? 
Thank you!

rhymes (changes requested):
Hi @Manasa2850, this PR looks incomplete and the edit page on the admin is broken. Can you please fix it?

Looking at the code I don't believe it can actually work though in the admin. The admin shouldn't know the ID of the `PodcastOwnership` row, see the `remove_owner`. The UI should send you the ID of the user that's an owner and you should remove it from `podcast.owners` after you found it. 

It should also not break if the ID doesn't exist in `podcast.owners`.

Please test the functionality manually as well.

benhalpern:
Once the changes like this go into production... https://docs.forem.com/backend/data-update-scripts/

mkcode:
From @benhalpern - Because we are not using a feature flag for this, we will need to add a data migration script.

Manasa2850:
@rhymes I would like to just keep the `adding_owner` feature in this PR since my fellowship is ending soon. I will take up the `removing_owner` feature in another PR since there's a bit of work to be done even on the frontend for that.
I've tested this out manually too. 
Thanks!

rhymes:
Hi @Manasa2850, thanks for the change, but it not working yet. I've recorded a session:

add_owner

As you can see it fails when the "add owner" button is pressed. If you closely look at the code you can see why, for example:

```shell
    38: def add_owner
    39:   binding.pry
 => 40:   @podcast_ownership = PodcastOwnership.new(podcast_ownership_params)
    41:
    42:   if @podcast_ownership.save
    43:     flash[:success] = "Owner created successfully"
    44:   else
    45:     flash[:error] = "User is already an owner"
    46:     render json: { error: "User is already an owner" }, status: :bad_request
    47:   end
    48: end

[1] pry(#<Admin::PodcastsController>)> params
=> <ActionController::Parameters {"utf8"=>"✓", "authenticity_token"=>"...", "podcast"=>{"user_id"=>"1"}, "commit"=>"Add Owner", "locale"=>nil, "controller"=>"admin/podcasts", "action"=>"add_owner", "id"=>"4"} permitted: false>
[2] pry(#<Admin::PodcastsController>)> params["podcast"]
=> <ActionController::Parameters {"user_id"=>"1"} permitted: false>
```

The `params` doesn't have a `podcast_ownership` hash, because that's not what the client side is sending to the server, see 
https://github.com/forem/forem/pull/11728/files#diff-508d7a1f98595436d190723629017397255a9d05797108b720001f008c6f3edbR47-R52

Right now the server know the ID of the podcast, from `params["id"]` and knows the ID of the owner from `params["podcast"]["user_id"]`. Given these two piece of information you could already build an ownership object with `PodcastOwnership.new(podcast_id: params["id"], user_id: params["podcast"]["user_id"])`

the rest stays the same.

Let me know if something isn't clear!

Manasa2850:
@rhymes I made the change you suggested. 
forem
Adding an owner using the `Add Owner` button reflects the changes! Please let me know if something else needs to be done.

rhymes:
@Manasa2850 thanks! make sure the tests work and `podcast_ownership_params` isn't needed anymore

Manasa2850:
@rhymes I'm getting a build error `The command "yarn e2e:ci" exited with 1`
I'm not sure if it's related to my PR

rhymes:
@Manasa2850 no unfortunately that's a current issue we're having with Cypress and Travis, I checked the logs, there's this failing test related to this PR: `rspec ./spec/requests/admin/podcasts_spec.rb:33`

Manasa2850:
@rhymes is this fine? I fixed the build error

rhymes (changes requested):
@Manasa2850 thanks for the change, we're almost there.

please test the functionality manually by starting the server and clicking around as I did in the GIF demonstration. If you do you'll notice that when `add_owner` is called there's no path for the user to go to when the server replies. 

If you add a new owner the admin page freezes because the server doesn't redirect anywhere.
If you add an existing owner the admin page renders JSON because that's what the `add_owner` action does, but this is a user facing functionality, so the user ends up in JSON page with no path to escape.

If you check the code for the old `remove_admin` and `add_admin`, see https://github.com/forem/forem/pull/11728/files#diff-584ee5b690e95242eeb1b7833d36db9c120b593d46513cf0729c5834ec9f33a2 you'll noticed those actions redirected the users somewhere.

You can adjust the code to your liking but make sure that 

1 - the user experience works
1- the tests are adjusted to test it, because right now the tests pass but the code is missing some parts

rhymes (changes requested):
Thanks @Manasa2850, I adjusted the code so that the new owner can be seen correctly both in the index and in the edit page.

I extracted it as a patch here https://gist.github.com/rhymes/ad6b00f296e054566daea3863ec4fdd0

If you want to apply it automatically you can download the gist file from https://gist.githubusercontent.com/rhymes/ad6b00f296e054566daea3863ec4fdd0/raw/fb640b3c58714b3c211503add632856717b2f3bc/0001-Finalize-add-owner-page.patch and then issue `git am 0001-Finalize-add-owner-page.patch` from the root directory of the repository project and then `git push`

This way we avoid one separate PR to finalize the last few changes.

rhymes (approved):
LGTM!

Thanks @Manasa2850!

app/controllers/admin/podcasts_controller.rb
```diff
@@ -34,21 +35,12 @@ def fetch
...
-      role = @user.add_role(:podcast_admin, @podcast)
-      if role.persisted?
-        redirect_to admin_podcasts_path, notice: "Added admin"
+    def add_owner
+      @podcast_ownership = @podcast.podcast_ownerships.build(user_id: params["podcast"]["user_id"])
+      if @podcast_ownership.save
+        redirect_to admin_podcasts_path, notice: "New owner added!"
       else
-        redirect_to edit_admin_podcast_path(@podcast), notice: "Error"
+        redirect_to edit_admin_podcast_path(@podcast), notice: @podcast_ownership.errors_as_sentence
```
juliannatetreault, line 43: Great use of `.errors_as_sentence` here! 😄

juliannatetreault (approved):
Thank you so much for tackling this @Manasa2850! I tested your changes locally and everything works great -- well done! 🎉

app/controllers/admin/podcasts_controller.rb
Manasa2850: @rhymes thanks for suggesting that! :smiley:

PR #11732, opened 2020-12-03
Change profile header cache key to purge headers

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We recently shipped a [profile fields overhaul](https://github.com/forem/forem/pull/10707) which resulted in a bug (dumb Ruby and the fact that it treats empty string as `truthy` 🤣 ).

We fixed it in https://github.com/forem/forem/pull/11730 but unfortunately the HTML is cached for 10 days. 

The quickest solution is to change the cache key. This will have a performance hit on DEV and possibly other Forems

#10707
#11730 

## Added tests?

- [ ] Yes
- [x] No, and this is why: caching keys aren't tested

## [optional] Are there any pre deployment tasks we need to perform?

Possibly the capacity of the DEV web server needs to be upped a bit before we deploy this

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -11,7 +11,8 @@
     <%= @user_json_ld.to_json.html_safe %>
   </script>
 <% end %>
-<% cache "main-user-profile-header-area-#{@user.id}-#{@user.profile_updated_at}-#{user_signed_in?}", expires_in: 10.days do %>
+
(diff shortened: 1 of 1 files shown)


maestromac (approved):
:+1: :+1:

fdoxyz (approved):
LGTM

Ridhwana (approved):

PR #11734, opened 2020-12-03
Remove concept of "ongoing" field tests and clarify a/b test instructions/expectations.

benhalpern:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This is a PR to consolidate and clarify how we use field tests and remove the old concept. It also fixes some current issues we have with the latest experiment.

It fixes a current issue where we are not recording the proper results for the new field test delivered recently, but it also adds clarification for using A/B tests for dedicated experiments with a beginning and end.

Ultimately this fixes the state of things right now and adds docs for future uses of `field_test`.

Future evolutions of this functionality includes perhaps opportunities to share results of a/b tests back with us, the maintainers, for different Forems. For now where relevant we can do this informally by coordinating with admins if necessary.

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [ ] No documentation needed

fresh prince


app/workers/field_tests/prune_old_experiments_worker.rb
```diff
@@ -1,13 +0,0 @@
-module FieldTests
-  class PruneOldExperimentsWorker
```
mstruve, line 2: What was the reason we had this before?

mstruve (approved):
Good catch! One small question about why we ever had the prune old events worker otherwise LGTM!

app/workers/field_tests/prune_old_experiments_worker.rb
benhalpern: The idea is that we would take people out of the system after a while and reintroduce them into the pool.

So if I was on feed variation A, and the experiment is meant to be long-lived, I can be reintroduced as a test subject for further evaluation to get more "modern" results. I.e. if lots of things had changed since my results were first accounted for, after a while I should just be reintroduced as a "fresh" subject with a new variation. This is only relevant if we're keeping experiments going for longer— And is really only feasible if our team is set up to administer these ongoing trials.

maestromac (approved):
LGTM

PR #11735, opened 2020-12-03
Improve user subscription UX flow when latency is high

vaidehijoshi:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

We ran into some latency + UX issues today around the user subscription tag. This PR improves on the issues that many users ran into by doing the following:

- [x] Disabling the `Subscribe` button while a request is submitting
- [x] Rendering errors alongside the `Subscribe` button so that users can try re-subscribing
- [x] Changing the text of the subscribe button to `Subscribing...` while the request is submitting

[This DEV post](https://dev.to/devteam/how-to-claim-your-hacktoberfest-completion-badge-3l2h) and related discussions in our support channel.

To QA this feature, pull down this branch locally, and make sure the user you are logged in as has the `:restricted_liquid_tag` role. Once you have the role, create a new article that includes the user subscription liquid tag in the markdown (see the [relevant docs here](https://app.gitbook.com/@forem/s/team-handbook/tech/user-subscription-liquid-tag)):

```
{% user_subscription I love llamas, click here to subscribe to top-notch llama content!! %}
```

To test my changes, you will want to first make some changes to the `UserSubscriptionsController`. You can mock latency by adding a `sleep 10` (or any other number of seconds) to the controller. You can mock errors by just commenting out the code and forcing an error. For example:

Screen Shot 2020-12-03 at 11 02 26 AM

Now, go to the article with the user subscription tag that you just created. You should see a button to subscribe:

Screen Shot 2020-12-03 at 10 57 27 AM

Click the button, and click the "confirm" button in the modal (the modal should automatically close once you confirm that you want to subscribe). Look for the disabled button and the `Submitting...` button text:
Screen Shot 2020-12-03 at 10 57 37 AM

It should take some time for the request to submit, and you should be able to test that you _cannot_ make multiple requests (the button should be disabled, so clicking it many times shouldn't do anything!)

You should see your "mock" error rendered on the user subscription box:
Screen Shot 2020-12-03 at 10 57 49 AM

Now, go back to the `UserSubscriptionsController` and revert those changes. Try re-submitting by clicking the button to try to re-subscribe again. You should see the `Submitting...` text appear again:
Screen Shot 2020-12-03 at 10 58 21 AM

...but this time, the request should be successful, and you should see your _success_ message. The subscribe button should have disappeared now:
Screen Shot 2020-12-03 at 10 58 31 AM

Finally, try refreshing the page entirely. You should still see a slightly different success message telling you that you are already subscribed, and the subscribe button should disappear:
Screen Shot 2020-12-03 at 10 58 57 AM

None that I'm aware of -- the only UI change here is changing the text and adding the `disabled` attribute to a button that should have already had it :) 

## Added tests?

- [x] ~Yes~ Updated pre-existing tests!
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

None!

### DIFF
--- a/app/liquid_tags/user_subscription_tag.rb
+++ b/app/liquid_tags/user_subscription_tag.rb
@@ -13,11 +13,13 @@ class UserSubscriptionTag < LiquidTagBase
     }
 
     // Hiding/showing elements
+    // If clearSubscribeButton is false, we will not clear out the subscription-signed-in area,
+    // which will allow users to re-submit their subscription if they see any error messages.
     // ***************************************
-    function clearSubscriptionArea() {
-      const subscriptionSignedIn = document.getElementById('subscription-signed-in');
-      if (subscriptionSignedIn) {
-        subscriptionSignedIn.classList.add("hidden");
+    function clearSubscriptionArea({ clearSubscribeButton = true } = {}) {
+      if (!clearSubscribeButton) {
+        // Allow users to try submitting again if they see an error.
+        hideSubscriptionSignedIn();
       }
 
       const subscriptionSignedOut = document.getElementById('subscription-signed-out');
@@ -25,10 +27,7 @@ class UserSubscriptionTag < LiquidTagBase
         subscriptionSignedOut.classList.add("hidden");
       }
 
-      const responseMessage = document.getElementById('response-message');
-      if (responseMessage) {
-        responseMessage.classList.add("hidden");
-      }
+      hideResponseMessage();
 
       const subscriberAppleAuth = document.getElementById('subscriber-apple-auth');
       if (subscriberAppleAuth) {
@@ -36,6 +35,15 @@ class UserSubscriptionTag < LiquidTagBase
       }
 
       hideConfirmationModal();
+    }
+
+    // Hides the response message (which displays success/error messages) if it exists.
+    // ***************************************
+    function hideResponseMessage() {
+      const responseMessage = document.getElementById('response-message');
+      if (responseMessage) {
+        responseMessage.classList.add("hidden");
+      }
     }
 
     function showSignedIn() {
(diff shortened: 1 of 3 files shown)


app/liquid_tags/user_subscription_tag.rb
```diff
@@ -251,12 +282,15 @@ class UserSubscriptionTag < LiquidTagBase
     // Handle API responses
     // ***************************************
     function handleSubscription() {
+      hideConfirmationModal(); // Close the modal once the user has confirmed.
```
vaidehijoshi, line 285: This is a change from what we previously did, but I think it makes more sense to do this. With higher latency requests, this modal just stays open, and it's confusing because I think a user would expect that clicking "confirm" would close the modal (rather than them having to click "confirm" and then _also_ click the "X" of the modal to close it).

mstruve (approved):
Tested all the code paths suggested and all worked great!!! LGTM! 🚀

andygeorge (approved):
I'm no javascript expert but the logic makes sense :+1:

app/liquid_tags/user_subscription_tag.rb
```diff
@@ -73,12 +81,20 @@ class UserSubscriptionTag < LiquidTagBase
...
+
       const responseMessage = document.getElementById('response-message');
       if (responseMessage) {
         responseMessage.classList.remove("hidden");
         responseMessage.classList.add(`crayons-notice--${noticeType}`);
         responseMessage.textContent = msg;
+
+        if (noticeType === 'danger') {
+          // Allow users to try resubscribing if they see an error message.
+          const subscribeBtn = document.getElementById('subscribe-btn');
```
nickytonline, line 94: suggestion (non-blocking): Consider getting a reference to the subscribe button only one.

There's four spots where `const subscribeBtn = document.getElementById('subscribe-btn')` is called. I'm assuming it's the same reference, so in that case, you could just define it once. If it's a new reference because DOM is destroyed, this is fine.

nickytonline (approved):
Left a comment, but still good to ship. 🚀

benhalpern (approved):
(Buildkite issue is unrelated)

app/liquid_tags/user_subscription_tag.rb
vaidehijoshi: @nickytonline Great suggestion, as always! I've pulled that element out into a const.

I also added some margin-bottom padding to the logged-in-text so that the error doesn't look as pushed-up against the text whenever it renders:
Screen Shot 2020-12-03 at 12 11 19 PM

nickytonline (approved):
🚢 :shipit:

PR #11738, opened 2020-12-03
Create manual UserSubscriptions' counter-cache update

maestromac:
- [x] Bug Fix

## Description
To ease down a possible future performance issue, this change will reduce the counter cache callbacks and move it into schedule jobs.

shoutout to @mstruve and @rhymes for their suggestions.
n/a

TBD
n/a
## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

n/a


app/models/user_subscription.rb
```diff
@@ -4,12 +4,19 @@
...
+                  column_name: nil,
+                  column_names: {
+                    true => "subscribed_to_user_subscriptions_count"
+                  }
   belongs_to :user_subscription_sourceable, polymorphic: true, optional: true
+  counter_culture :user_subscription_sourceable,
+                  column_name: nil,
+                  column_names: {
+                    true => "user_subscriptions_count"
+                  }
```
maestromac, line 19: This counter cache issue ends up taking way more time than I thought because I went down the route of not using `counter_culture`, but I ran into the issue of not being able to track destroy the item, then @rhymes point out to me a neat way avoid the callback without removing `counter_culture`. 

I am almost there. I just can't figure out why it's not updating the counter properly when a subscription_sourceable is destroy.

app/workers/user_subscriptions/sync_counter_cache.rb
```diff
@@ -0,0 +1,11 @@
+module UserSubscriptions
+  class SyncCounterCache
+    include Sidekiq::Worker
+
+    sidekiq_options queue: :default, retry: 10
+
+    def perform
```
mstruve, line 7: I see this triggered in the specs below but where in the production running code do we trigger this to run?

maestromac: Ah, I left out the `config/schedule.yml` update. Updated!

app/models/user_subscription.rb
mstruve: Neat work around!

config/schedule.yml
```diff
@@ -107,6 +107,9 @@ remove_old_notifications:
 sync_credits_counter_cache:
   cron: "0 16 * * *" # daily at 4 pm UTC
   class: "Credits::SyncCounterCache"
+sync_user_subscription_counter_cache:
+  cron: "*/ * * * *" # every 15 minutes
```
mstruve, line 111: I think you forgot to add the 15 🙂  
```
*/15 * * * *
```

app/models/user_subscription.rb
maestromac: After a bit of messing around, I found out that `UserSubscription.counter_culture_fix_counts` won't be able to properly decrement a polymorphic's count to 0. It can never go below 1 because of [counter_culture's reconciliation strategy](https://github.com/magnusvk/counter_culture/blob/master/lib/counter_culture/reconciler.rb#L48). It works by looking for a distinct `*_type` attributes and it won't be able to when there's no record to work with. This still remains the same if I were to properly use `column_name: :user_subscriptions_count` because what decrement the counter is the AR callback. 

If this is an acceptable side-effect I can adjust the test to make this pass. The other idea I have that would require more work would be to move the whole `create` process into a background job. I'm happy to dive further and am all ears for any suggestion.

maestromac:
We're not in dire need of a solution to this yet. Putting it to sleep for now.

PR #11628, opened 2020-11-26
Update typography in `text-styles`

lisasy:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
Update typography in `text-styles` to improve overall visual rhythm and eliminate unnecessary margins or padding on `:only-child` elements.

#### Before
Screen Shot 2020-11-24 at 2 00 25 PM

Screen Shot 2020-11-24 at 8 45 36 AM

#### This PR
Screen Shot 2020-11-25 at 4 32 48 PM
Screen Shot 2020-11-25 at 4 33 00 PM
Screen Shot 2020-11-25 at 4 33 06 PM
Screen Shot 2020-11-25 at 4 32 31 PM

_Please replace this line with instructions on how to test your changes, as well
as any relevant images for UI changes._

### UI accessibility concerns?

_If your PR includes UI changes, please replace this line with details on how accessibility is impacted and tested. For more info, check out the [Forem Accessibility Docs](https://docs.forem.com/frontend/accessibility)._

## Added tests?

- [ ] Yes
- [ ] No, and this is why: _please replace this line with details on why tests
      have not been included_
- [ ] I need help with writing tests

## Added to documentation?

- [ ] [Developer Docs](https://docs.forem.com) and/or [Admin Guide](https://forem.gitbook.io/forem-admin-guide/)
- [ ] README
- [ ] No documentation needed

## [optional] What gif best describes this PR or how it makes you feel?

alt_text

PR #11743, opened 2020-12-04
Text styles

ludwiczakpawel:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR cleans up our texts styling a bit in the post content area. It is inspired by #11628 - after talking to Lisa we agreed to make some changes and I figured it would make more sense to create a separate PR.

- Closes #11628 

Ideally, you can just click through different posts that you have locally. You can also try to use this [markdown template](https://gist.githubusercontent.com/ludwiczakpawel/ac5c332bfe5a37dfde002949cee65890/raw/907c8cb151efa1f63c120dcdc9ccefb7619a79c7/test.md) - it includes bunch of different combinations of tags.

## Added tests?

- [ ] Yes
- [x] No, not needed.

## Added to documentation?

- [ ] README
- [x] No documentation needed


ludwiczakpawel:
@lisasy friendly remainder :).

lisasy:
There are some use-cases where maintaining a `margin-bottom` via `margin: 0 0 var(--content-rhythm) 0;`. In my original PR, I proposed this as a safeguard:

```
 & > *:only-child {
    margin: 0;
  }
```

Unless you suspect this might have negative consequences, is this something we can bring in here?

Image 2020-12-08 at 1 14 48 PM

lisasy:
This is a comment you left in our 1:1 Slack channel:

> I didn’t feel very strongly about overwriting default color and text-transform of headlines - that was originally included in your PR.
> - we should let users decide what should have uppercase letters and what not. I can imagine situations where not using uppercase letters could be intentional. think of someone thinking about having title like “@BradleyCooper noStringsAttached” (lol :D). if we overwrite original style with uppercase it would be “@BRADLEYCOOPER NOSTRINGSATTACHED”… i mean, lets let users have some creativity freedom :stuck_out_tongue:

That seems like a sensible rationale to me, and my initial approach to us was us having an opinionated typography on Forem so that we can elevate their content via good visual design. One benefit of greater variation of `h` stylings is it also offers people more option to "prettify" or write their content — it could also enhance the reading experience if people can scan through different sections more easily. If in the future we can see some optimizations in typography, I would like for us to be open-minded here :)

app/assets/stylesheets/base/typography.scss
```diff
@@ -96,44 +95,52 @@ body {
...
     padding: 0;
+    font-weight: var(--fw-bold);
+    line-height: var(--lh-tight);
   }
 
   h1 {
-    font-size: 1.85em;
-    line-height: 1.25em;
-    font-weight: 400;
+    font-size: 2em;
```
lisasy, line 105: Something I notice (and I hope it isn't an incorrect observation but let me know if it is) is that when I resize the viewport to the smallest size, the `h` texts stay the same size as when it's on XL viewport. I'm wondering if it might be more ideal for these to scale relative to viewport because something on a phone doesn't need to see a really really big `h1`; all they need to know is that it's the most emphasize title relative to everything else.

I've heard of using `rem`s as one way to approach this and I would love to hear how you're thinking about this.

app/assets/stylesheets/base/typography.scss
```diff
@@ -96,44 +95,52 @@ body {
...
+    font-size: 1em;
   }
 
   hr {
     width: 25%;
     opacity: 0.1;
+    border-radius: var(--radius);
     border: 1px solid var(--body-color);
-    margin: 1.3em auto 1.5em;
+    margin: calc(var(--content-rhythm) * 2) auto;
```
lisasy, line 133: Really nice change!

app/assets/stylesheets/base/typography.scss
```diff
@@ -96,44 +95,52 @@ body {
...
+  h4 {
+    font-size: 1.4em;
+  }
+
+  h5 {
+    font-size: 1.2em;
+  }
+
+  h6 {
+    font-size: 1em;
```
lisasy, line 125: Do you think we can write these lines like this so that we can reference how they fit into our Crayons variables?

As someone who contributes to this code and and wants to maintain consistency, this would be very helpful.

`font-size: 1em // 18px // var(--fs-l);`

lisasy:
In this example here, the font-size of the post title is smaller than the font-size of the `h1` text. Can you help reconcile this so the hierarchy here plays more nicely? 

Image 2020-12-08 at 1 55 04 PM

lisasy:
I was looking at the scale you've defined here, and I see the hierarchy, but when I do the math for some of them, it creates values that aren't referenced in our [design system's typography](https://www.figma.com/file/4FOzCCKLZaLBz6X5TZCVNF/Design-System?node-id=242%3A77).

Given that we've spent good time defining this system and know that it helps things look balanced, what is the trade-off of using `em` values that would result as closely as possible to these variables we've decided on? I thought a benefit is that it allow us to maintain typographic consistency since the text must work with typography elsewhere on the app. (also, this comment is also related to this one about the trade-off of using [rem instead of em](https://github.com/forem/forem/pull/11743#discussion_r538828648))

The tricky thing here: we have different `--content-font-size` for the `base` size and the `medium` size. Here's a possibility based on some quick math....

### What we have right now

```
h1 --> 2em // 36px var(--fs-4xl)
h2 --> 1.8em //  32.5px --> UNEVEN VALUE
h3 --> 1.6em // 28.8px --> UNEVEN VALUE
h4 --> 1.4em // 25.2px --> UNEVEN VALUE
h5 --> 1.2em // 21.6px --> UNEVEN VALUE
h6 --> 1em // 18px var(--fs-l)
```


### Proposed / Base
```
h1 --> 2em // 36px var(--fs-4xl)
h2 --> ??????? 
h3 --> 1.67em // 30px var(--fs-3xl) ?????? 
h4 --> 1.33em // 24px var(--fs-2xl)
h5 --> 1.11em // 20px var(--fs-xl)
h6 --> 1em // 18px var(--fs-l)
```

Those `em` values are not, but I thought that if its endpoints results in aligning more closely to our type scale, that will be better visual design. Taking our literal type system and applying it "breaks" because you have see I don't really know what to put in `h2` :P Let's pair on this!

app/assets/stylesheets/base/typography.scss
ludwiczakpawel: `em` unit is relative to its parent and the parent font-size is dependent on two factors:

1. whether it's article body or article comments
2. breakpoint

So we can't really define what 1em means because we can't tell what's the main context :P. I think adding px values (or referencing variables) may add a bit of confusion in this case TBH.

app/assets/stylesheets/base/typography.scss
ludwiczakpawel: > Something I notice (and I hope it isn't an incorrect observation but let me know if it is) is that when I resize the viewport to the smallest size, the h texts stay the same size as when it's on XL viewport.

I'll check, maybe something is broken.

> I've heard of using rems as one way to approach this

`rem`s are a bit different - these are relative to the font size defined in the `body` element. These are much more useful in scenarios like this for simple layouts and designs, especially when *everything* can be scalable based on font-size (including all kinda of sizes for elements, paddings, margins, etc. etc)...
These won't be very useful in our case (here, where we define article typography) - but we do use `rem`s in other areas, where appropriate.

lisasy: I see -- thanks for that explanation :)

ludwiczakpawel:
> There are some use-cases where maintaining a margin-bottom via margin: 0 0 var(--content-rhythm) 0; (...)

This should be solved now. 

> That seems like a sensible rationale to me, and my initial approach to us was us having an opinionated typography on Forem so that we can elevate their content via good visual design. (...) If in the future we can see some optimizations in typography, I would like for us to be open-minded here :)

Yea, I think I agree and we could revisit that later on since I think this will require more testing against existing posts to make sure we don't break anything for anyone (which I'm also not sure how to tackle in an efficient way tbh). But I can totally see us setting some constraints around designs of various elements. To get this PR out the door, let's just circle back to this at some point.

> In this example here, the font-size of the post title is smaller than the font-size of the h1 text. Can you help reconcile this so the hierarchy here plays more nicely?

Good catch, I've made some adjustments.

> I was looking at the scale you've defined here, and I see the hierarchy, but when I do the math for some of them, it creates values that aren't referenced in our design system's typography.

I think the typography sizes we defined in design system doc, are more for UI elements (buttons, forms, some standalone titles (feed card), and everything else that create the user interfaces).

For content areas (post body, comment body) I think it's ok to not follow the guidelines:
1. It's a dynamic content - not only in terms of WHAT it is (user can generate anything) but also in terms of sizing (we vary size depending on screen resolution and placement - _post body_ vs _comment body_)
2. It doesn't affect sizing of our UI. It should be optimized for reading, not for UI accessibility.
3. You pointed out somewhere that we use uneven values for `h1...h6` - but that's IMO absolutely fine because no matter what fonts will always have subpixels (antialiasing) - otherwise it would be a pixelart :). Here's what I mean: https://d.pr/i/574LPj - when you zoom in any text you will notice that letters are never aligned to full pixels (green lines)

> That's one of the reasons why in my initial proposal, h5 and h6 were the same size, but only with differnet styling

That sounds good to me. Keeping `h5` and `h6` the same size but `h6` with maybe uppercase and lower opacity..? I know it's against what I mentioned earlier but I think that could be a good compromise.

Although, I've also took a look at how GitHub renders markdown (they actually make less common Hs smaller than regular text) and I've decided to combine their approach with what I mentioned above, at least for now - as a start.

```
h1 { font-size: 2em; }
h2 { font-size: 1.5em; }
h3 { font-size: 1.25em; }
h4 { font-size: 1em; }
h5 { font-size: 0.875em; }
h6 {
  font-size: 0.85em;
  color: var(--base-60);
}
```

---

I've also added text underline for links to make them more accessible.

lisasy (approved):
@ludwiczakpawel looks super good to me. thank you so much for helping out on this with me! 💯 😇

lisasy:
> > There are some use-cases where maintaining a margin-bottom via margin: 0 0 var(--content-rhythm) 0; (...)
> 
> This should be solved now.
> 
> > That seems like a sensible rationale to me, and my initial approach to us was us having an opinionated typography on Forem so that we can elevate their content via good visual design. (...) If in the future we can see some optimizations in typography, I would like for us to be open-minded here :)
> 
> Yea, I think I agree and we could revisit that later on since I think this will require more testing against existing posts to make sure we don't break anything for anyone (which I'm also not sure how to tackle in an efficient way tbh). But I can totally see us setting some constraints around designs of various elements. To get this PR out the door, let's just circle back to this at some point.
> 
> > In this example here, the font-size of the post title is smaller than the font-size of the h1 text. Can you help reconcile this so the hierarchy here plays more nicely?
> 
> Good catch, I've made some adjustments.
> 
> > I was looking at the scale you've defined here, and I see the hierarchy, but when I do the math for some of them, it creates values that aren't referenced in our design system's typography.
> 
> I think the typography sizes we defined in design system doc, are more for UI elements (buttons, forms, some standalone titles (feed card), and everything else that create the user interfaces).
> 
> For content areas (post body, comment body) I think it's ok to not follow the guidelines:
> 
>     1. It's a dynamic content - not only in terms of WHAT it is (user can generate anything) but also in terms of sizing (we vary size depending on screen resolution and placement - _post body_ vs _comment body_)
> 
>     2. It doesn't affect sizing of our UI. It should be optimized for reading, not for UI accessibility.
> 
>     3. You pointed out somewhere that we use uneven values for `h1...h6` - but that's IMO absolutely fine because no matter what fonts will always have subpixels (antialiasing) - otherwise it would be a pixelart :). Here's what I mean: https://d.pr/i/574LPj - when you zoom in any text you will notice that letters are never aligned to full pixels (green lines)
> 
> 
> > That's one of the reasons why in my initial proposal, h5 and h6 were the same size, but only with differnet styling
> 
> That sounds good to me. Keeping `h5` and `h6` the same size but `h6` with maybe uppercase and lower opacity..? I know it's against what I mentioned earlier but I think that could be a good compromise.
> 
> Although, I've also took a look at how GitHub renders markdown (they actually make less common Hs smaller than regular text) and I've decided to combine their approach with what I mentioned above, at least for now - as a start.
> 
> ```
> h1 { font-size: 2em; }
> h2 { font-size: 1.5em; }
> h3 { font-size: 1.25em; }
> h4 { font-size: 1em; }
> h5 { font-size: 0.875em; }
> h6 {
>   font-size: 0.85em;
>   color: var(--base-60);
> }
> ```
> 
> I've also added text underline for links to make them more accessible.

Thanks for addressing the feedback and answering the questions/concerns I had above. I'm on board with everything you said, and I've just approved the PR.

👏 Kudos to borrowing from GitHub in how they render markdown — smart, quick approach!
👏 Kudos to adding an underline to links.

jacobherrington (approved):
main-qimg-f310be840b3994e94965c0f0f45a6e7b

app/views/pages/crayons.html.erb
```diff
@@ -76,7 +76,7 @@
 <div class="container">
   <div class="body">
     <h2>Main Typography</h2>
-    <p>Default font is set to 16px (fs-base). It should be standard in UI. Smaller and bigger font sizes should be used carefully with respect to good visual rythm between elements.</p>
+    <p>Default font is set to 16px (fs-base). It should be standard in UI. Smaller and bigger font sizes should be used carefully with respect to good visual rhythm between elements.</p>
```
aitchiss, line 79: just a thought now I'm seeing this, but maybe we should change this to say something like 

```suggestion
    <p>Default font is set to 1rem (fs-base). In standard browser settings this equates to 16px. Smaller and bigger font sizes should be used carefully with respect to good visual rhythm between elements.</p>
```

just thinking it's not totally accurate to say it's set to 16px, even though in most cases that's what 1rem ends up being

aitchiss (approved):
It looks good to me, I've just left a small nitpicky suggestion about the crayons doc page, but feel free to take it or leave it :)

app/views/pages/crayons.html.erb
ludwiczakpawel: I'll actually deprecate this page - it's been there mostly as a playground for myself and I don't think it's very useful to anyone - it's actually the opposite if we want to polish our Storybook and make it the single source of truth.

Nevertheless, that's a good catch!

PR #11748, opened 2020-12-04
Change logic for checking for database on boot.

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This code may run while `ENV["ENV_AVAILABLE"]` is still false, but not only in `assets:precompile` mode.

`ENV["ENV_AVAILABLE"]` is designed to check whether environment variables are present. And I'm thinking that it might be the case that environment variables are not present yet when this code runs, even in the "real" boot.... Which doesn't entirely makes sense to me, but maybe this runs before the `ENV` object is working properly?

This is fairly one-off type of behavior, as described here:

```ruby
  # NOTE: @citizen428 We want to have a current list of profile attributes the
  # moment the application loads. I wish Rails had a hook for code to run after
  # the app started...
  refresh_attributes!
```

At least this shouldn't break anything further I'm pretty sure and might be worth trying?

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -25,12 +25,13 @@ class Profile < ApplicationRecord
 
   # Generates typed accessors for all currently defined profile fields.
   def self.refresh_attributes!
-    return if ENV["ENV_AVAILABLE"] == "false"
     return unless Database.table_exists?("profiles")
 
     ProfileField.find_each do |field|
       store_attribute :data, field.attribute_name.to_sym, field.type
     end
+  rescue PG::ConnectionBad
+    puts "No database connections available. Assumed to be running assets:precompile."
   end
 
   # Returns an array of all currently defined `store_attribute`s on `data`.
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -30,7 +30,7 @@ class Profile < ApplicationRecord
     ProfileField.find_each do |field|
       store_attribute :data, field.attribute_name.to_sym, field.type
     end
-  rescue PG::ConnectionBad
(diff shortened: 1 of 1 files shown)


andygeorge (approved):

fdoxyz:
Travis seems like it's stuck/hung for some reason 🤔

**Question:**

If the problem is that ENV variables aren't set, that means the check `return if ENV["ENV_AVAILABLE"] == "false"` isn't working because `ENV_AVAILABLE` wasn't set before this line was executed.

Would it maybe work to invert it to `return unless ENV["ENV_AVAILABLE"].present?` or `return unless ENV["ENV_AVAILABLE"] == "true"`?

Despite not being fully aware of where this error is being raised in the container build pipeline, would it make sense to add the ENV variable to the [Dockerfile assets:precompile line](https://github.com/fdoxyz/forem/blob/master/Containerfile#L43)? So it's available during the `assets:precompile` execution

```Dockerfile
RUN RAILS_ENV=production NODE_ENV=production bundle exec rake assets:precompile

# vs

RUN ENV_AVAILABLE=false RAILS_ENV=production NODE_ENV=production bundle exec rake assets:precompile
```

fdoxyz:
My comment above doesn't bock or is necessary to be taken into account if it's not inline with the problem at hand by the way, as this seems to be the key part:

> This code may run while ENV["ENV_AVAILABLE"] is still false

Only wanted to share some ideas on it 👍🏼

benhalpern:
I'm going to close this because I don't think this is the root problem.

PR #11750, opened 2020-12-04
[EOY 2020 Issue] Alert admins in modal when enabling Auth Provider without keys

msarit:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR alerts admins via a modal when they attempt to enable Auth Providers without entering their respective key and secret. The PR also refactors the `adminModal.js` function to make it more generalizable for more use-cases.

This PR introduces an improved User Flow, that will replace the Warning Banner in [this PR](https://github.com/forem/forem/pull/11696).

1. As `:super_admin` role, navigate to `localhost:xxxx/admin/config`. Toggle open the Authentication section.
2. Enable one or more Authentication Providers, but leave the keys blank (or you can type random string for Key OR Secret, but NOT both). Update Site Configuration. A modal should popup listing the providers you attempted to enable.
3. Click "Continue editing" button or "X" the modal. You should be taken back to the main Config page AND the Update Site Configuration button should NOT be disabled.
4. Update site configuration again. This time, click "Save anyway". The update action should fire, the page refresh, but the keyless provider(s) you attempted to enable should still be disabled.
5. Re-enable your chosen providers, but this time, enter dummy strings in both key fields. Update site configuration. The page should refresh, and you should see your chosen providers are enabled.

Demo Video of PR:
https://www.loom.com/share/2ffd7af973af4054b21832f59cb9eec9

None anticipated

## Added tests?

- [ ] Yes
- [X] No, and this is why: _current tests cover the changes in this PR_

## Added to documentation?

- [ ] README
- [X] No documentation needed; docs already reflect the need to enter auth provider key and secret

No

Toddler pushing object


app/javascript/admin/controllers/config_controller.js
```diff
@@ -244,4 +244,67 @@ export default class ConfigController extends Controller {
...
+  }
+
+  activateMissingKeysModal(providers) {
+    this.configModalAnchorTarget.innerHTML = adminModal(
+      this.missingKeysModalTitle(),
+      this.missingKeysModalBody(providers),
+      'Continue editing',
+      'closeAdminModal',
+      'Save anyway',
+      '', // this will be the function that Updates Configuration
```
vaidehijoshi, line 297:
```suggestion
      'yourFunctionHere', // this will be the function that Updates Configuration
```

app/javascript/admin/controllers/config_controller.js
```diff
@@ -45,7 +45,7 @@ export default class ConfigController extends Controller {
     }
   }
 
-  closeAdminConfigModal() {
+  closeAdminModal() {
```
vaidehijoshi, line 48:
```suggestion
  yourFunctionHere() {
    // I think you'll want to use fetch() here to make a request to the ConfigsController#create action, which is what happens when you click on the "Update Site Configuration" button
    // Something like this...
    fetch(`/admin/config`, {
      method: 'POST',
      headers: {
        Accept: 'application/json',
        'X-CSRF-Token': document.querySelector("meta[name='csrf-token']")
          .content,
        'Content-Type': 'application/json',
      },
      body: JSON.stringify({
        // I think the request will fail without the confirmation text 😔 
        confirmation: '...'
      }),
      credentials: 'same-origin',
      })
  },

  closeAdminModal() {
```

This will be a little tricky because of all the authorization stuff we have in place when someone tries to update the SiteConfig. I'm not sure that the pseudocode I've thrown in here will work as-is, but that is effectively what you'll need to do if you want to submit a request to the backend from this JS button.

Take a look at [this function](https://github.com/forem/forem/blob/master/app/javascript/admin/controllers/reaction_controller.js#L10-L22), which is kind of similar/along the same lines.

vaidehijoshi (commented):
Hi @msarit! Left a couple comments around my hunch of what you'll need to do to make this work.

It might be a little tricky because the configs controller on the Rails side has a bunch of authorization logic, so you may run into some issues when trying to make the request successfully. You'll also need to nest the params you send from the JS side to the Rails side properly in order for it to work correctly 😉 

Let me know if you want me to record a walkthrough or if you have more questions! I think this might be a little bit involved/tricky because we don't have other JS code that makes a request to the `ConfigsController#create` controller action yet 😅

rhymes (commented):
Hi @msarit - I think the solution is within grasp without having to write backend code.

I'll explain, let's do a step back first.

A common JS pattern of conditionally submitting a form (that's what we are trying to do) can be achieved in very simple terms by something like `window.confirm`, the standard HTML ugly modal, that returns true or false depending if the user has pressed `OK` or `Cancel`.

By checking the return value of this default modal, the programmer knows if the user wants to go ahead or not.

This is usually achieved by something like (bear with me, it's not 100% correct JS code and I haven't tested it):

```js
form.addEventListener('submit', (event) => {
  if (window.confirm("Are you sure you want to do this?" === true) {
   event.target.submit();
  } else {
   event.preventDefault();
  }
});
```

Something like this would result in the form submitting to the server if the user pressed "OK" and to abort if they pressed "Cancel" (in our case they are probably inverted because the blue button is the abort).

This gives us a first clue. The event we have to listen in on is not the click of the button itself but the [submit event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit_event) of the form. In our case the form of the `Authentication` section.

Something like this:

```html
        <%= form_for(SiteConfig.new, url: admin_config_path, html: { data: { action: "submit->config#configUpdatePrecheck" } }) do |f| %>
```

on line `124` of the `configs/show.html.erb` template (the click action in the `_form_submission.html.erb.` file can be removed in my opinion).

So, now we know that we are monitoring the submission event of the form, what's left to do is to put a modal between it and the server so that we can, as the example, call `event.preventDefault()` when the modal says "the user has chosen the option to abort" and call `event.target.submit()` (`submit()` is a [function available to the form DOM element](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit)).

This gives us the second clue. We can get the `<form>` element from the HTML and just activate `.submit()` on it. 

Let me know if anything is not as clear as it should be, great job so far! :)

app/javascript/admin/controllers/config_controller.js
```diff
@@ -47,10 +48,18 @@ export default class ConfigController extends Controller {
...
+      'input[type="submit"]',
+    );
+
     this.configModalAnchorTarget.innerHTML = '';
     document.body.style.height = 'inherit';
     document.body.style.overflowY = 'inherit';
+
+    if (authSectionUpdateConfigBtn.hasAttribute('disabled')) {
+      authSectionUpdateConfigBtn.removeAttribute('disabled');
+    }
```
rhymes, line 62: ❤️

rhymes (approved):
Works great, great job @msarit :) LGTM!

app/javascript/admin/adminModal.js
```diff
@@ -4,28 +4,32 @@
...
-  confirmBtnText,
-  confirmBtnAction,
-  cancelBtnText,
-  cancelBtnAction,
+  firstBtnText,
+  firstBtnAction,
+  secondBtnText,
+  secondBtnAction,
+  firstBtnClasses,
+  secondBtnClasses,
```
vaidehijoshi, line 24: It might be more descriptive to call these `leftButton` and `rightButton` to help make the rendering/usage of these buttons a bit more clear. To me, `firstBtn` isn't as clear as `confirmButton`, but I totally get why you switched the naming here. I think that `leftButton` is a bit clearer because, as a developer, I know where this will _render_; `firstBtn` doesn't necessarily make that as clear (and the documentation we have for it, `The text for the modal's first button.`, also doesn't explain that; a developer would have to look at the markup directly to figure out what the "first button" does)

app/javascript/admin/controllers/config_controller.js
```diff
@@ -246,4 +259,68 @@ export default class ConfigController extends Controller {
...
+    return providersWithMissingKeys;
+  }
+
+  generateProvidersList(providers) {
+    let list = '';
+    for (let i = 0; i < providers.length; i++) {
+      list += `<li>${providers[i]}</li>`;
+    }
+    return list;
+  }
```
vaidehijoshi, line 289: I think we can use `forEach` here (like you do [earlier in the file](https://github.com/forem/forem/pull/11750/files#diff-caedbb7d7984a39e9a441e5df7c79b425ec24ebe6d54b4e50261e0634c183793R267), which will make this easier to read :) 
https://alligator.io/js/foreach-vs-for-loops

app/javascript/admin/controllers/config_controller.js
```diff
@@ -246,4 +259,68 @@ export default class ConfigController extends Controller {
...
+
+  missingKeysModalTitle() {
+    return 'Setup not complete';
+  }
```
vaidehijoshi, line 293: Is there a reason we need to make this a function? Since it is just returning a string, I think we can pass that string along directly to `adminModal`, right?

app/javascript/admin/controllers/config_controller.js
```diff
@@ -246,4 +259,68 @@ export default class ConfigController extends Controller {
...
+      .querySelectorAll('[data-enable-auth="true"]')
+      .forEach((provider) => {
+        const providerName = provider.dataset.authProviderEnable;
+        const officialName = provider.dataset.authProviderOfficial;
+        if (
+          document.getElementById(`site_config_${providerName}_key`).value
+            .length === 0 ||
+          document.getElementById(`site_config_${providerName}_secret`).value
+            .length === 0
+        ) {
```
vaidehijoshi, line 275: I _believe_ that if the element doesn't have a value, it should be an empty string (`""`). Empty strings are falsy in JavaScript, so I think you don't need to do the `.length === 0` check here

app/javascript/admin/adminModal.js
msarit: I agree :)

app/javascript/admin/controllers/config_controller.js
```diff
@@ -246,4 +259,62 @@ export default class ConfigController extends Controller {
...
+      list += `<li>${provider}</li>`;
+    });
+    return list;
+  }
+
+  missingKeysModalBody(providers) {
+    return `<p>You haven't filled out all of the required fields properly to save these settings, specifically for: <ul>${this.generateProvidersList(
+      providers,
+    )}</ul></p>`;
+  }
```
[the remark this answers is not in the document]
vaidehijoshi: Oh! I also just noticed this -- the HTML we are building/rendering here is _not_ semantic, which means that it won't be accessible to screenreaders. We shouldn't nest `<ul>`s inside of a `<p>` tag (you can read more about why [not](https://stackoverflow.com/questions/5681481/should-ol-ul-be-inside-p-or-outside), but the short version is that `<p>` tags only accepts _phrasing_ content inside of them, and `<ul>` is a type of flow content, which makes this rendering illegal html 😅

vaidehijoshi (changes requested):
This is very close!! One blocking issue with inaccessible HTML (will be easy to fix!), and one suggestion around the copy we use to communicate with the Forem admin.

nickytonline:
Just putting some eyes on this @msarit.

app/javascript/admin/adminModal.js
nickytonline: question: Are the button names more generic because for example, the left button isn't always a confirm button? I know it's been changed, but `confirmButton` provides more context than `leftButton`.

app/javascript/admin/adminModal.js
```diff
@@ -4,28 +4,32 @@
...
  * @param {string} customAttrValue The value of the custom attribute "customAttr".
  */
 const adminModal = (
   title,
   body,
-  confirmBtnText,
-  confirmBtnAction,
-  cancelBtnText,
-  cancelBtnAction,
+  leftBtnText,
```
nickytonline, line 19: refactor (blocking): Typically you want 1-3 parameters in a function. More than that and it gets hairy. Having one parameter with properties would work better here, like the props property in React/Preact. With that you get named properties as well which means the order they're defined in won't matter.

```javascript
const adminModal = ({
  title,
  body,
  leftBtnText,
  leftBtnAction,
  rightBtnText,
  rightBtnAction,
  leftBtnClasses,
  rightBtnClasses,
  customAttr = null,
  customAttrValue = null,
}) => `
...
```

nitpick (non-blocking): Prefer a [function expression](https://developer.mozilla.org/en-US/docs/web/JavaScript/Reference/Operators/function) over an [arrow function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) unless you need access to the lexical scope. i.e. access to the `this` property.

app/javascript/admin/controllers/config_controller.js
```diff
@@ -19,6 +19,7 @@ const emailAuthModalBody = `
 export default class ConfigController extends Controller {
   static targets = [
     'authenticationProviders',
+    'authSectionForm',
```
nickytonline, line 22: refactor (blocking): The tests for this will need to be updated.

nickytonline (changes requested):
@msarit, just doing a first path. This is looking good. Just some change requests.

app/javascript/admin/adminModal.js
msarit: Great point, @nickytonline! Will do!

app/javascript/admin/controllers/config_controller.js
nickytonline: Instead of fetch, use our `request` utility function. 

```javascript
fetch('/admin/config', {
  method: 'POST',
  headers: {
    Accept: 'application/json',
    'X-CSRF-Token': document.querySelector("meta[name='csrf-token']").content,
    'Content-Type': 'application/json',
  },
  body: JSON.stringify({
    // I think the request will fail without the confirmation text 😔
    confirmation: '...',
  }),
  credentials: 'same-origin',
});
```

would become this

```javascript
request('/admin/config', {
  method: 'POST',  
  body: { // Note: you do not need to do JSON.stringify. The request function handles that.
    // I think the request will fail without the confirmation text 😔
    confirmation: '...',
  }
})
```

nickytonline: Oops, I think the comment I commented out is no longer relevant. 😅

app/javascript/admin/adminModal.js
msarit: Yes @nickytonline you are right. I made this modal as generic as possible so it can be repurposed as needed.

msarit:
Closed in favor of new PR: https://github.com/forem/forem/pull/12135

PR #11782, opened 2020-12-07
Remove timeframer from app/labor

citizen428:

- [X] Refactor

## Description

This PR removes `app/labor/timeframer.rb`. I ended up moving it to `app/models` for the following reasons:

1. It encapsulates data, not behavior, so it isn't a service.
2. I tried an alternate approach with a `TimeframeHelper`. This ended up feeling clunky since we are also using this in controllers  (where we could get away with using the `helpers` proxy method) and services (where we would have to include the helper module).

I'm open to moving it elsewhere (or go back to the helper approach). I'm still in half a mind to throw it in `app/helpers`, even though it isn't a classical Rails helper. Or maybe it should just go to `app/lib`, we have similar things there, though there's a danger of that directory becoming yet another dumping ground. 

Specs passing should be good enough.

## Added tests?

- [X] Moved existing ones

## Added to documentation?

- [X] No documentation needed


rhymes (approved):
It makes sense, thanks for the change

benhalpern (approved):

atsmith813 (approved):

PR #11788, opened 2020-12-07
Add bust_article service

atsmith813:
- [x] Refactor

## Description

Here are the methods:
- [ ] bust_page
- [ ] bust_tag
- [ ] bust_events
- [ ] bust_podcast
- [ ] bust_organization
- [ ] bust_podcast_episode
- [ ] bust_listings
- [ ] bust_user
- [x] bust_tag_pages
- [x] bust_home_pages
- [x] bust_article
- [x] bust_article_comment
- [x] bust_comment

## Added tests?
- [x] Yes - I'm open to ideas on how to provide better test coverage!

## Added to documentation?
- [x] No documentation needed

ucf_volleyball_spike_setup

### DIFF
--- a/app/services/edge_cache/bust.rb
+++ b/app/services/edge_cache/bust.rb
@@ -1,10 +1,10 @@
 module EdgeCache
   class Bust
     TIMEFRAMES = [
-      [1.week.ago, "week"],
-      [1.month.ago, "month"],
-      [1.year.ago, "year"],
-      [5.years.ago, "infinity"],
+      [-> { 1.week.ago }, "week"],
+      [-> { 1.month.ago }, "month"],
+      [-> { 1.year.ago }, "year"],
+      [-> { 5.years.ago }, "infinity"],
     ].freeze
 
     def self.call(path)
--- a/app/services/edge_cache/bust.rb
+++ b/app/services/edge_cache/bust.rb
@@ -1,5 +1,12 @@
 module EdgeCache
   class Bust
+    TIMEFRAMES = [
+      [1.week.ago, "week"],
+      [1.month.ago, "month"],
+      [1.year.ago, "year"],
+      [5.years.ago, "infinity"],
+    ].freeze
+
     def self.call(path)
       bust(path)
     end
--- a/app/services/edge_cache/bust_article.rb
+++ b/app/services/edge_cache/bust_article.rb
@@ -37,7 +37,7 @@ module EdgeCache
       end
 
       TIMEFRAMES.each do |timestamp, interval|
-        next unless Article.published.where("published_at > ?", timestamp)
+        next unless Article.published.where("published_at > ?", timestamp.call)
           .order(public_reactions_count: :desc).limit(3).ids.include?(article.id)
 
         bust("/top/#{interval}")
@@ -65,7 +65,7 @@ module EdgeCache
         end
 
         TIMEFRAMES.each do |timestamp, interval|
-          next unless Article.published.where("published_at > ?", timestamp).tagged_with(tag)
+          next unless Article.published.where("published_at > ?", timestamp.call).tagged_with(tag)
             .order(public_reactions_count: :desc).limit(3).ids.include?(article.id)
 
           bust("/top/#{interval}")
--- a/app/services/edge_cache/bust_article.rb
+++ b/app/services/edge_cache/bust_article.rb
@@ -1,0 +1,90 @@
+module EdgeCache
+  class BustArticle < Bust
+    def self.call(article)
+      return unless article
+
+      article.purge
+
+      bust(article.path)
+      bust("/#{article.user.username}")
+      bust("#{article.path}/")
+      bust("#{article.path}?i=i")
+      bust("#{article.path}/?i=i")
+      bust("#{article.path}/comments")
+      bust("#{article.path}?preview=#{article.password}")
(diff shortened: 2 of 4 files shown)


spec/services/edge_cache/bust_article_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+RSpec.describe EdgeCache::BustArticle, type: :service do
+  let(:article) { create(:article) }
+
+  before do
+    allow(article).to receive(:purge)
+    allow(described_class).to receive(:bust)
+  end
+
+  it "busts the cache" do
```
atsmith813, line 11: I'm open to ideas on how to expand this test coverage!

vaidehijoshi: Some ideas:
- we could test that `bust` was called _x_ number of times
- we could test that the organization slug is busted if it has an org
- we could test that `bust_home_pages ` and `bust_tag_pages` [were called](https://github.com/forem/forem/pull/11788/files#diff-f9b1966dac6d323e3c03dfd9e957e73a5e586df05546904669c677d14b9b93ccR17-R18)?

i don't think all of that is probably necessary, but it would make it really clear to us if the service was changed down the road because the tests would probably catch it 🤷

app/services/edge_cache/bust.rb
```diff
@@ -1,5 +1,12 @@
 module EdgeCache
   class Bust
+    TIMEFRAMES = [
+      [1.week.ago, "week"],
```
citizen428, line 4: **bug (blocking):** This will not work as expected, as the timeframes will be relative to application start (when the class body gets evaluated), not the current time.

```ruby
TIMEFRAMES = [[1.week.ago, "week"]]
#=> [[Tue, 01 Dec 2020 05:35:36 UTC +00:00, "week"]]
```

I suggest making them lambdas instead and calling them in the code:

```ruby
TIMEFRAMES = [
  [-> { 1.week.ago }, "week"],
  ...
]

# e.g. app/services/edge_cache/bust_article.rb
next unless Article.published.where("published_at > ?", timestamp.call)
```

app/services/edge_cache/bust_article.rb
```diff
@@ -0,0 +1,90 @@
...
+      bust("#{article.path}?preview=#{article.password}")
+      bust("#{article.path}?preview=#{article.password}&i=i")
+      bust("/#{article.organization.slug}") if article.organization.present?
+      bust_home_pages(article)
+      bust_tag_pages(article)
+      bust("/api/articles/#{article.id}")
+
+      return unless article.collection_id
+
+      article.collection&.articles&.find_each do |collection_article|
```
citizen428, line 23: **suggestion (non-blocking):** You probably just copy/pasted this when moving the code, but I think this overuses `&.`: if we got past the guard clause above, we already know we have a collection (referenced by the `collection_id`) and we also know that this collection has at least one article (the one starting this method chain).

citizen428 (dismissed):

spec/services/edge_cache/bust_article_spec.rb
atsmith813: Added all 3 suggestions 💁🏽‍♂️

app/services/edge_cache/bust.rb
atsmith813: Great call, updated!

app/services/edge_cache/bust_article.rb
atsmith813: Updated!

spec/services/edge_cache/bust_spec.rb
```diff
@@ -4,6 +4,10 @@
   let(:user) { create(:user) }
   let(:path) { "/#{user.username}" }
 
+  it "defines TIMEFRAMES" do
+    expect(described_class.const_defined?(:TIMEFRAMES)).to be true
+  end
```
rhymes, line 9: I would use `Timecop` to test that by moving forward or backwards in time those timeframes adjust to the current time inside the Timecop block, just to be on the safe side (I think testing one of them is enough)

rhymes (commented):
Great job @atsmith813, left a note on how to possibly improve the specs for `TIMEFRAMES`

app/services/edge_cache/bust.rb
vaidehijoshi: > the timeframes will be relative to application start (when the class body gets evaluated), not the current time...I suggest making them lambdas instead

Interesting; @citizen428 I wouldn't have thought to do this, TIL!

spec/services/edge_cache/bust_spec.rb
atsmith813: Good idea! I gave it a try, let me know what you think @rhymes. I understand your point, but I'm less sure that I implemented the way you described.

rhymes (approved):
LGTM!

benhalpern (approved):
Looks great

vaidehijoshi (approved):

app/services/edge_cache/bust.rb
citizen428: @vaidehijoshi It's also a common problem in migrations, when people add new date columns and want to set a default: if you don't specify it in the correct way, the default will be the timestamp of the migration run, not evaluated on record creation.

Issue #11791, opened 2020-12-07
Fix 404 on "Recent Emails" via /admin/users


pkfrank:
**Describe the bug**
As a Forem Admin, clicking on a "recent email" via `/admin/users` is leading to a 404.

**To Reproduce**
- Be an admin with sufficient permissions
- Navigate to a user via `/admin/users`
- Click a "recent email" at the bottom of the page
- Hit a 404

**Expected behavior**

Link opens up properly, showing the relevant email

**Additional context**

This is a left-over issue stemming from the `/admin` to `/resource_admin` project.

I think the "fix" could either be:
- Update the link path to `/resource_admin` (though we're trying to keep Forem Admins out of this area in general)
- Move this email history functionality into `/admin` (bigger undertaking, probably some connected Product questions)

fdoxyz:
https://github.com/forem/forem/pull/11792 tackles this by creating a custom view for admins to read these emails, and also removes `resource_admin` altogether.

Are there any specific features that we need from the "recent emails"? @pkfrank 

The links to the recent emails that appear in this area will be "read only", meaning that admins can't edited them, or perform any other action on them (like destroy). Unless there's something we need to explicitly support here I think this issue will be fixed with this PR.

pkfrank:
Looks great to me @fdoxyz!

I don't think we need any extra features here — "read only" is great.

IMO, this issue can be closed once #11792 is merged.

PR #11792, opened 2020-12-07
Remove `resource_admin` dashboards (Administrate)

### LINKED ISSUE #10483: Remove /resource_admin dashboard

As we're moving towards a unified `/admin` dashboard we are discontinuing `/resource_admin` altogether. In order to do this we are closing the gap in the last few capabilities we are still missing. This means that in order to get this done we will need to have the following issues resolved first:

- [x] #10481
- [x] #10482
- [x] #11026

This requires removing the Administrate gem from the Gemfile, the `DashboardManifest` ([here](https://github.com/forem/forem/blob/master/app/dashboards/dashboard_manifest.rb)), existing tests and any other code related to this area.

**Additional context**

This is the end goal of the effort to remove `/resource_admin` in favor of `/admin` dashboard

fdoxyz:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR rips off the bandaid on getting rid of the `/resource_admin` dashboards.

The only missing piece of functionality I found was the ability for Admins to see recent emails sent to a user. Because of this I added a very simple preview for them linked from `/admin/user/:id` at the very bottom (Email Tools).

Closes #10483 and #11791

Also closes https://github.com/forem/InternalProjectPlanning/issues/287 https://github.com/forem/InternalProjectPlanning/issues/8 https://github.com/forem/InternalProjectPlanning/issues/27

**Looking for broken links or functionality**
1. Boot up the app locally
1. Browse through the `/admin` dashboards 
1. Try to look for any lingering `/resource_admin` links or broken functionality

**Email message previews**
1. Go to `/admin/users` locally
1. Click on any seed user
1. Scroll down to "Email Tools"
1. Send them an email
1. Check at the very bottom for the links to "Recent Emails"
1. Clicking on them should show a preview of the email message sent

None at the moment because this PR removes stuff rather than add new features, except for one page to view emails that's pretty small.

## Added tests?

- [ ] Yes
- [x] No, and this is why: I tweaked existing tests to make sure the current coverage remains unchanged.

## Added to documentation?

- [ ] README
- [x] No documentation needed

None

finally


app/views/admin/users/email_message.html.erb
```diff
@@ -0,0 +1,46 @@
...
+        </tr>
+      </tbody>
+    </table>
+  </div>
+
+  <div class="crayons-card p-6">
+    <h2>Email content</h2>
+    <p><em>The content is previewed below without formatting</em></p>
+    <div class="crayons-card my-5 p-5">
+      <%= @email.body_html_content.html_safe %>
```
fdoxyz, line 43: Codeclimate is not happy about rendering the email content as `html_safe`. I don't think we should see any problems with this because this is an admin tool and it's not open for the public to abuse.

That being said, I'm okay with reverting to present this without `html_safe` if anyone feels like that would be better

fdoxyz:
Buildkite check is not passing because a dependency was removed from the Gemfile. Maybe @forem/systems can weigh in on what to do in this case?

andygeorge:
> Buildkite check is not passing because a dependency was removed from the Gemfile. Maybe @forem/systems can weigh in on what to do in this case?

Hey @fdoxyz - the buildkite error is:
```
Run `bundle install` elsewhere and add the updated Gemfile.lock to version control.
```
Can you also remove `administrate` from [`Gemfile.lock`](https://github.com/forem/forem/blob/fdoxyz/remove-administrate-dashboards-10483/Gemfile.lock) and see if that passes the build?

fdoxyz:
I thought `bin/setup` took care of updating Gemfile.lock, turns it doesn't or I missed this. Thanks @andygeorge!

config/routes.rb
```diff
@@ -107,6 +98,8 @@
         resource :moderator, only: %i[create destroy], module: "tags"
       end
       resources :users, only: %i[index show edit update] do
+        get "email_message/:id", to: "users#email_message"
```
citizen428, line 101: Maybe I'm missing something, but how about a normal resource instead? 

```ruby
resources :email_message, only: :show
```

A separate controller seems cleaner than putting more responsibility into `Admin::UsersController` and we'd get a path helper that you could use instead of manually constructing the URL like this:

```ruby
<a href="/admin/users/<%= @user.id %>/email_message/<%= email.id %>"
```

rhymes: That would require adding a controller and not just an action to an existing one. An alternative is to use:

```ruby
#  (message is a bit redundant?)
get "email_message/:id", to: "users#email_message", as: :users_email_message
```

which will also build a useful helper to avoid harcoding the url

rhymes (commented):
Love the change @fdoxyz, I have the same exact feedback about using a helper as @citizen428

The rest seems to be working correctly!

config/routes.rb
citizen428: I'm very much in favor of more controllers with only restful actions instead of adding arbitrary actions to existing controllers.

rhymes: Agreed, it also plays in the wider discussion in coming up with an "internal best practices" kind of document

fdoxyz: Great feedback, thanks @citizen428  and @rhymes! I'll get this restructured soon

rhymes (approved):
Bye bye bye administrate!

Bliss:

Screenshot_2020-12-08 Remove `resource_admin` dashboards (Administrate) by fdoxyz · Pull Request #11792 · forem forem

jacobherrington (approved):
:scissors:

PR #11812, opened 2020-12-08
Create services for busting pages, tags, events, and podcasts

atsmith813:
- [x] Refactor

## Description
_I'm choosing to refactor a handful of methods/services at once in this PR since they're simpler and similar._

Here are the methods:
- [x] bust_page
- [x] bust_tag
- [x] bust_events
- [x] bust_podcast
- [ ] bust_organization
- [ ] bust_podcast_episode
- [ ] bust_listings
- [ ] bust_user
- [x] bust_tag_pages
- [x] bust_home_pages
- [x] bust_article
- [x] bust_article_comment
- [x] bust_comment

## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

lil_waybe_moving_along_gif

### DIFF
--- a/app/services/edge_cache/bust_podcast.rb
+++ b/app/services/edge_cache/bust_podcast.rb
@@ -1,0 +1,9 @@
+module EdgeCache
+  class BustPodcast < Bust
+    def self.call(path)
+      return unless path
+
+      bust("/#{path}")
+    end
+  end
+end
--- a/spec/services/edge_cache/bust_events_spec.rb
+++ b/spec/services/edge_cache/bust_events_spec.rb
@@ -1,0 +1,24 @@
+require "rails_helper"
+
+RSpec.describe EdgeCache::BustEvents, type: :service do
+  let(:paths) do
+    [
+      "/events",
+      "/events?i=i",
+    ]
+  end
+
+  before do
+    paths.each do |path|
+      allow(described_class).to receive(:bust).with(path).once
+    end
+  end
+
+  it "busts the cache" do
+    described_class.call
+
+    paths.each do |path|
+      expect(described_class).to have_received(:bust).with(path).once
+    end
+  end
+end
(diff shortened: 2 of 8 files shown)


app/services/edge_cache/bust_podcast.rb
```diff
@@ -0,0 +1,9 @@
+module EdgeCache
+  class BustPodcast < Bust
+    def self.call(path)
+      return unless path
+
+      bust("/#{path}")
+    end
+  end
+end
```
atsmith813, line 9: Technically, we don't _really_ need this service since it just busts the path itself. I thought it would be nice to leave as a dedicated service though because it makes it much easier to handle Podcast specific caches in the future.

spec/services/edge_cache/bust_events_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  it "busts the cache" do
+    described_class.call
+
+    paths.each do |path|
+      expect(described_class).to have_received(:bust).with(path).once
+    end
+  end
+end
```
atsmith813, line 24: We could probably move some of this logic into helpers/shared examples in the future!

app/services/edge_cache/bust_podcast.rb
benhalpern: This makes sense. I could see this having more functionality in the future, so keeping the interface seems good.

benhalpern (approved):
All in all seems great to me. Nice job with clean code + tests.

app/services/edge_cache/bust_podcast.rb
vaidehijoshi: I like that you pulled it into a service -- even if it's simple for now, I prefer the consistency 👍🏽

vaidehijoshi (approved):
Nice work!! (Also, you're getting so close to finishing this refactor up 👏🏽👏🏽👏🏽 )

citizen428 (approved):
🔥

rhymes (approved):
Awesome!

PR #11825, opened 2020-12-09
Remove app/labor/color_from_image.rb

citizen428:

- [X] Refactor

## Description

This PR removes the `ColorFromImage` class from `app/labor`, as well as the `Articles::UpdateMainImageBackgroundHexWorker` which only existed to call this object. 

Here's some context from @benhalpern:

> [Disabling the Miro gem] was originally the PR which put us over the Heroku threshold for slug size and this gem was pulled in order to make that individual deploy succeed.
>
> However, the Miro gem is pretty old and has a dependency which was renamed and the code is in master, but the latest release doesn't reflect that. So there was a question of whether it was a good dependency for the app to have.
>
> This functionality was put in place to replace an old library which handled this behavior which also had gone unsupported, so it's been a bit of a journey. I like the automatic background color detection, but could see a case for just removing it. We don't do this on other images besides the main image and we clearly have lived with this non-functional endpoint for long enough.
>
>  If we really want this again in the future we can find a cleaner overall solution.

Nothing specific

- [X] No, and this is why: only removing code, not adding any new one

## Added to documentation?

- [X] No documentation needed


rhymes (approved):
🪓

rhymes:
@citizen428 forgot to mention in the review, not blocking obviously, we might want to get rid of this line too: https://github.com/forem/forem/blob/66dc8081b11525826885311c5eebd385f6e1eff0/Gemfile#L60

citizen428:
@rhymes Thanks, done ✅

TODO
```diff
@@ -0,0 +1 @@
+* look into profile updates to remove blank values
```
maestromac, line 1: :eyes: did you accidentally commit this @citizen428 ?

Ridhwana (approved):

benhalpern (approved):

TODO
citizen428: Yes, I did 🤦 It's my private TODO list...

PR #11826, opened 2020-12-09
Move AssignTagModerator from labor to services

citizen428:

- [x] Refactor

## Description

This PR moves `AssignTagModerators` in `app/labor` to 3 services in `app/services`.

1. `TagModerators::AddTrustedRole`: is a separate service because we need this functionality in another context too.
2. `TagModerators::Add`: Adds users tag moderators
3. `TagModerators::Add`: Removes a user as a tag moderator

I also did some minor code reorganization/cleanup in the process, but mostly tried to keep things as they were.

Try to add/remove tag moderator roles via the admin area.

## Added tests?

- [X] Yes and moved existing ones

## Added to documentation?

- [x] No documentation needed


rhymes (changes requested):
I get an error in the https://github.com/forem/forem/blob/1729cf02ad1718df58ea301f2da82fd0062f47d1/app/controllers/admin/tags/moderators_controller.rb#L15 when I tried to add a moderator to a tag

citizen428:
Ah yes, that'd be one of these cases where the code works better when you actually commit all the files 🤦 Will add it tomorrow.

rhymes (commented):
I tried to re-test it but now I get this error, any idea?

Screenshot_2020-12-11 NoMethodError at admin tags 13 moderator

I tried to reimport the fields from the CSV, `ProfileField.count` says 28 and `available_for` is in there. I did notice though there's no data in the profile object. I tried to add it by hand but I still get the same error.

```ruby
=> #<Profile:0x00007fdfc8d5ef08
 id: 11,
 user_id: 11,
 data: {"custom_attributes"=>{"available_for"=>""}},
 created_at: Thu, 10 Dec 2020 18:10:21 UTC +00:00,
 updated_at: Fri, 11 Dec 2020 09:54:28 UTC +00:00>
```

this is the status of my profile object ("available_for" was added manually, quite sure I did not understand the code correctly since that doesn't fix the issue)

citizen428:
@rhymes: `available_for` is not supposed to go into `custom_attributes`, that's solely for the the 5 attributes users can define by themselves. However, whether or not a profile responds to an attribute is independent from the JSONB hash containing this key.

```ruby
p = Profile.new
p.data
p.data
#=> { "custom_attributes" => {} }
p.respond_to?(:available_for)
p.respond_to?(:available_for)
#=> true
```

What's weird is that during the PR for ignoring the columns on user, everything was tested a lot by me, Molly and Fernando. We only saw weird issues like this with caching enabled and now it started re-appearing after I moved the profile initialization code from the profile model to `after_initialize` as you suggested. I have a weird gut feeling that this introduced the problem and here's why:

* My previous solution with calling `refresh_attributes` explicitly in `Profile`'s class body: the moment the class is loaded, it's ready and all accessors are set up.
* The new code using `after_initialize`: when `User` is loaded, `after_initialize` did not run yet. This means the accessors aren't available, which would explain this error. 

I'm very tempted to roll back this change from `after_initialize` to how it was before and see if things work again.

rhymes:
@citizen428 let's do it. Thanks for the explanation!

citizen428:
@rhymes I rebased this against current `master`, please re-review, thanks!

rhymes (approved):
Works great, thanks!

benhalpern (approved):

PR #11832, opened 2020-12-09
Create services for busting organizations, podcast episodes, listings, and users

atsmith813:
- [x] Refactor

## Description
_I'm choosing to refactor a handful of methods/services at once in this PR since they're similar._

Here are the methods:
- [x] bust_page
- [x] bust_tag
- [x] bust_events
- [x] bust_podcast
- [x] bust_organization
- [x] bust_podcast_episode
- [x] bust_listings
- [x] bust_user
- [x] bust_tag_pages
- [x] bust_home_pages
- [x] bust_article
- [x] bust_article_comment
- [x] bust_comment

## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

jersey_shore_moving_on_gif

### DIFF
--- a/app/services/edge_cache/bust_organization.rb
+++ b/app/services/edge_cache/bust_organization.rb
@@ -1,0 +1,17 @@
+module EdgeCache
+  class BustOrganization < Bust
+    def self.call(organization, slug)
+      return unless organization && slug
+
+      bust("/#{slug}")
+
+      begin
+        organization.articles.find_each do |article|
+          bust(article.path)
+        end
+      rescue StandardError => e
+        Rails.logger.error("Tag issue: #{e}")
+      end
+    end
+  end
+end
--- a/app/services/edge_cache/bust_podcast_episode.rb
+++ b/app/services/edge_cache/bust_podcast_episode.rb
@@ -1,0 +1,18 @@
+module EdgeCache
+  class BustPodcastEpisode < Bust
+    def self.call(podcast_episode, path, podcast_slug)
+      return unless podcast_episode && path && podcast_slug
+
+      podcast_episode.purge
+      podcast_episode.purge_all
+
+      begin
+        bust(path)
+        bust("/#{podcast_slug}")
+        bust("/pod")
+      rescue StandardError => e
+        Rails.logger.warn(e)
+      end
+    end
+  end
+end
--- a/app/services/edge_cache/bust_user.rb
+++ b/app/services/edge_cache/bust_user.rb
@@ -1,0 +1,22 @@
+module EdgeCache
+  class BustUser < Bust
+    def self.call(user)
+      return unless user
+
+      username = user.username
+
+      paths = [
+        "/#{username}",
+        "/#{username}?i=i",
+        "/#{username}/comments",
+        "/#{username}/comments?i=i",
+        "/#{username}/comments/?i=i",
+        "/live/#{username}",
+        "/live/#{username}?i=i",
+        "/feed/#{username}",
+      ]
+
+      paths.each { |path| bust(path) }
+    end
+  end
+end
--- a/spec/services/edge_cache/bust_organization_spec.rb
+++ b/spec/services/edge_cache/bust_organization_spec.rb
@@ -16,4 +16,11 @@ RSpec.describe EdgeCache::BustOrganization, type: :service do
     expect(described_class).to have_received(:bust).with("/#{slug}").once
     expect(described_class).to have_received(:bust).with(article.path).once
   end
+
+  it "logs an error" do
+    allow(described_class).to receive(:bust).with("/5").once
+    allow(Rails.logger).to receive(:error)
+    described_class.call(4, 5)
+    expect(Rails.logger).to have_received(:error).once
+  end
 end
--- a/spec/services/edge_cache/bust_organization_spec.rb
+++ b/spec/services/edge_cache/bust_organization_spec.rb
@@ -1,0 +1,19 @@
+require "rails_helper"
+
+RSpec.describe EdgeCache::BustOrganization, type: :service do
+  let(:organization) { create(:organization) }
+  let(:article) { create(:article, organization: organization) }
+  let(:slug) { "slug" }
+
+  before do
+    allow(described_class).to receive(:bust).with("/#{slug}").once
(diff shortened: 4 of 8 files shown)


app/services/edge_cache/bust_user.rb
```diff
@@ -0,0 +1,22 @@
...
+        "/#{username}?i=i",
+        "/#{username}/comments",
+        "/#{username}/comments?i=i",
+        "/#{username}/comments/?i=i",
+        "/live/#{username}",
+        "/live/#{username}?i=i",
+        "/feed/#{username}",
+      ]
+
+      paths.each { |path| bust(path) }
```
atsmith813, line 19: This style was used originally, so I just copied it over!

app/services/edge_cache/bust_podcast_episode.rb
```diff
@@ -0,0 +1,18 @@
...
+
+      podcast_episode.purge
+      podcast_episode.purge_all
+
+      begin
+        bust(path)
+        bust("/#{podcast_slug}")
+        bust("/pod")
+      rescue StandardError => e
+        Rails.logger.warn(e)
```
atsmith813, line 14: I'm not sure why we rescue errors here. I'm guessing it's something with podcast episodes?

app/services/edge_cache/bust_organization.rb
```diff
@@ -0,0 +1,17 @@
...
+      return unless organization && slug
+
+      bust("/#{slug}")
+
+      begin
+        organization.articles.find_each do |article|
+          bust(article.path)
+        end
+      rescue StandardError => e
+        Rails.logger.error("Tag issue: #{e}")
```
atsmith813, line 13: I'm not sure why we rescue errors here or call the errors 'Tag issue' 🤔

app/services/edge_cache/bust_podcast_episode.rb
```diff
@@ -0,0 +1,18 @@
+module EdgeCache
+  class BustPodcastEpisode < Bust
+    def self.call(podcast_episode, path, podcast_slug)
+      return unless podcast_episode && path && podcast_slug
+
+      podcast_episode.purge
+      podcast_episode.purge_all
+
+      begin
```
atsmith813, line 9: If we take a look at the original code...

https://github.com/forem/forem/blob/eee63068dcde7e342a0d204666f3ba98c938ea2c/app/labor/cache_buster.rb#L140-L153

We are making duplicate actions (`bust(path)`, `podcast_episode.purge`, and `podcast_episode.purge_all`). Is this intentional to cover an edge case?

cc: @benhalpern

app/services/edge_cache/bust_organization.rb
rhymes: 🤔🤔🤔 :D

rhymes (approved):
LGTM!

spec/services/edge_cache/bust_podcast_episode_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    expect(podcast_episode).to have_received(:purge)
+    expect(podcast_episode).to have_received(:purge_all)
+  end
+
+  it "logs an error" do
+    allow(Rails.logger).to receive(:warn)
+    allow(described_class).to receive(:bust).and_raise(StandardError)
+    described_class.call(podcast_episode, 12, podcast_slug)
```
vaidehijoshi, line 40: Same question about the `12` here :) The function definition makes me think we should be passing `podcast_path` here instead of `12`...but if not, I am curious why not

spec/services/edge_cache/bust_organization_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    described_class.call(organization, slug)
+
+    expect(described_class).to have_received(:bust).with("/#{slug}").once
+    expect(described_class).to have_received(:bust).with(article.path).once
+  end
+
+  it "logs an error" do
+    allow(described_class).to receive(:bust).with("/5").once
+    allow(Rails.logger).to receive(:error)
+    described_class.call(4, 5)
```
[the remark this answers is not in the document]
atsmith813: I have no idea, I copied it over from the existing spec

https://github.com/forem/forem/blob/c2982ef0fc9dfc75a12ebdbfb8f78fc16110a8ac/spec/labor/cache_buster_spec.rb#L79

My guess is they are "random, invalid arguments".

spec/services/edge_cache/bust_podcast_episode_spec.rb
atsmith813: The same thing here - I have no idea, I copied it over from the existing spec

https://github.com/forem/forem/blob/c2982ef0fc9dfc75a12ebdbfb8f78fc16110a8ac/spec/labor/cache_buster_spec.rb#L92

My guess is `12` is an "invalid argument".

app/services/edge_cache/bust_organization.rb
atsmith813: That makes sense! Maybe @benhalpern has some insight as to where this came from originally.

My plan is to leave it for now. I'm trying to change as little as possible in the logic of the service so it's much easier to debug if something goes wrong when flipping the switch to using the service. By keeping the "copy/pasta" in place, the scope of debugging is much more narrow - I'm basically checking on the calling of the service. Then, once the services are up and running, we can always revisit for specific cleanups

vaidehijoshi (approved):

citizen428 (approved):

app/services/edge_cache/bust_organization.rb
benhalpern: I merged this because it seems fine to maintain the code in the refactor, all else equal.

— But this particular rescue is the sort of thing that dates back to this _solo project rapidly do something to get it working_, so I have no specific idea why this is needed as such. Most likely can be removed if it seems like it can.

app/services/edge_cache/bust_podcast_episode.rb
benhalpern: We used to not have the `purge` and `purge_all` functionality mapping properly to the surrogate keys, so this very well could be completely unneeded duplicate or reminiscent to some case not 100% handled by either version. But I'm not sure which would be the case....

PR #11836, opened 2020-12-09
Don't attempt to call summary on profile during onboarding

jacobherrington:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

At the moment, it is impossible to complete onboarding on master. A NoMethodError is thrown when the `summary_too_long?` method is called. This change makes it possible to complete onboarding. I think we should merge this to fix onboarding, but I think I need to look a little more closely at what is going on here because I might be missing some edgecase where we don't want this validation to be skipped whenever the summary isn't accessible from the profile.

Before this change, the follow steps probably won't work, after they do.

1. Create a new account
2. Go through onboarding
3a. Attempt to "skip for now" on the personal information
3b. Fill out the info and try to continue

## Added tests?

- [ ] Yes
- [x] No, and this is why: The system specs for onboarding seem like a work in progress and I want to get this fix out asap. I'll revisit those tests in another PR (because I think they need more attention) and include a regression test for this bug.

## Added to documentation?

- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/validators/profile_validator.rb
+++ b/app/validators/profile_validator.rb
@@ -32,6 +32,9 @@ class ProfileValidator < ActiveModel::Validator
   private
 
   def summary_too_long?(record)
+    # During onboarding, the attribute is not yet available
+    return unless record.respond_to?(SUMMARY_ATTRIBUTE)
+
     return unless ProfileField.exists?(attribute_name: SUMMARY_ATTRIBUTE)
     return if record.summary.blank?
 


benhalpern (approved):
This check is perfectly harmless so good call. Definitely a deeper thing we'll want to wrap our heads around.

benhalpern:
Not sure what those netlify checks are about. Seems we can ignore.

fdoxyz:
I think it _might_ be related to caching, but I haven't debugged this to know for sure. 

Just wanted to share it as a heads up, because we had to add `responds_to?` checks for the other "non special" fields (summary is special) [when the user was cached](https://github.com/forem/forem/pull/10707#issuecomment-736183882) and this feels very similar... Also a reason why the system specs may pass (we don't test with caching enabled I believe).

fdoxyz:
I'm always paranoid of what happens under the hood when doing rake db:reset (the schema is loaded if available to save time? and maybe other stuff? not sure). This is probably an overkill but this is what I'm running now to test:

```bash
#!/bin/bash

rm db/schema.rb
rake db:reset
bin/setup
bin/startup
```

When going over the onboarding with this check in place I'm getting this (not the summary problem):

Screen Shot 2020-12-09 at 19 17 24

It could totally be that I have something misconfigured (I have a bunch of ENV variables for a custom domain) or that I'm missing something. It will take a minute to run but if anyone else runs into the same result with the script it may help me debug more/better 😅

citizen428 (approved):
Like @fdoxyz said, we do have this check for all attributes, but this one was missed.

app/validators/profile_validator.rb
```diff
@@ -32,6 +32,9 @@ def validate(record)
   private
 
   def summary_too_long?(record)
+    # During onboarding, the attribute is not yet available
```
citizen428, line 35: I would remove this comment, it's misleading. If the attribute isn't available during onboarding, it won't be available after either, no fields get created in the process.

citizen428 (approved):

Ridhwana (approved):

PR #11837, opened 2020-12-09
Comment out ignored columns and temp code causing troubles

benhalpern:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

There seem to be scenarios where the current state of this code is causing problems— Not all the time, but there are too many situations popping up and I am concerned that this will cause issues with any communities which may have already adjusted their profile columns.

I think commenting this out for now while we deploy new things and taking some baby steps here will help.

This is related to this PR as well: https://github.com/forem/forem/pull/11836

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -39,35 +39,35 @@ class User < ApplicationRecord
     youtube_url
   ].freeze
 
-  self.ignored_columns = PROFILE_COLUMNS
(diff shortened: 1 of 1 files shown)


citizen428:
Just to clarify, the problem wasn't with this code, was it?

Issue #11456, opened 2020-11-17
Remove `collective_noun` + `collective_noun_disabled` and use `community_name` instead


vaidehijoshi:
**Is your feature request related to a problem? Please describe.**

After https://github.com/forem/forem/pull/11107 was merged, I [pointed out](https://github.com/forem/forem/pull/11107#issuecomment-717406783) that we probably don't need to add `collective_noun` and `collective_noun_disabled` to the `SiteConfig`, which is _already_ pretty overloaded.

**Describe the solution you'd like**

Instead, it would be ideal to use `community_name` (which already exists and is a required field) in the places where `collective_noun` is used rather than adding another override-able config.

Something to consider is @jacobherrington's [comment](https://github.com/forem/forem/pull/11107#issuecomment-717408490) on the original PR:
> the only issue would be handling instances in which folks didn't want to include the "collective noun" value. For example, when we say "Welcome to DEV" versus "Welcome to the DEV Community"

mstruve:
Seems like a nice little cleanup task. I'm thinking difficulty easy on this one. If you disagree feel free to update the label!

PR #11846, opened 2020-12-10
Replace collective_noun with community_name

juliannatetreault:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR removes `collective_noun` and `collective_noun_disabled` from the SiteConfig and any relevant places in the codebase and replaces `collective_noun` with `community_name`. One concern that was voiced on the issue associated with this work was:
> the only issue would be handling instances in which folks didn't want to include the "collective noun" value. For example, when we say "Welcome to DEV" versus "Welcome to the DEV Community"

I realize what I'm about to say is a bit radical, _but_ to resolve this issue, I have replaced `community_qualified_name` with `community_name` and have updated the copy above the "Community Name" textbox in the SiteConfig to be more explicit about what an Admin could put there, e.g. "DEV" or "DEV Community" or even "DEV Flock." In doing this, I have:

- Cut down on some of our code and complexity in the `application_helper.rb`, which keeps our use of `community_name` a little more consistent (there were places in the codebase where we were using `"<%= community_name %> Community"` and in others, `<%= community_qualified_name %>`)

- Given Admins more flexibility around what their "collective noun" is while also generalizing "community name" a bit more, without adding bloat to the SiteConfig

- Removed an edge case where Admins could potentially set their "Community Name" to something along the lines of "DEV Flock" in the SiteConfig, causing our already appended "Community" to turn their "Community Name" in some areas to "DEV Flock Community" without the Admin realizing

Closes https://github.com/forem/forem/issues/11456
Relates to PR #11973 

To QA this PR, first ensure that you have `admin` and `config` privileges and then please do the following:

- Navigate to `/admin/config` and make sure everything still looks correct now that the `collective_noun` textbox and `collective_noun_disabled` checkbox have been removed from the "Community" section of the config:
Screen Shot 2020-12-10 at 8 04 47 AM

- While testing this out locally, please also look through the CoC, Terms of Use, Contacts page, and any other page where `collective_noun` and/or `community_qualified_name` was replaced to ensure that the pages are still readable:
Screen Shot 2020-12-10 at 8 09 18 AM

- Ensure that all of the checks run and are green! ♻️ 

N/A

## Added tests?

- [ ] Yes
      have not been included_
- [x] I removed and cleaned up existing tests

## Added to documentation?

- [ ] README
- [ ] No documentation needed
- [x] I have removed the "Collective Noun" section from the Admin Guide and have added some further details to the "Community Name" section, but have not yet merged the changes. To view these changes and suggest edits, please take a look at my proposed changes: https://app.gitbook.com/@forem/s/forem-admin-guide/~/drafts/-MOr2aA_SCj61eSjxPLG/admin/config/community-content

I have included a `data_update` script in this PR that will need to run to update the `community_name` for Forems that do not have `collective_noun_disabled` to append their `collective_noun` to their `community_name`.

Happy Bill Nye at a computer


spec/requests/articles/articles_show_spec.rb
```diff
@@ -36,7 +36,7 @@
         "publisher" => {
           "@context" => "http://schema.org",
           "@type" => "Organization",
-          "name" => "#{SiteConfig.community_name} #{SiteConfig.collective_noun}",
+          "name" => SiteConfig.community_name.to_s,
```
juliannatetreault, line 39: This seems to be failing on Travis but is passing locally. 🤔

spec/mailers/devise_mailer_spec.rb
```diff
@@ -11,7 +11,7 @@
     end
 
     it "renders sender" do
-      expected_from = "#{SiteConfig.community_name} Community <#{SiteConfig.email_addresses[:default]}>"
+      expected_from = "#{SiteConfig.community_name} <#{SiteConfig.email_addresses[:default]}>"
       expect(email["from"].value).to eq(expected_from)
     end
```
juliannatetreault, line 16: This is failing because `email["from"].value` returns `DEV (local) Community`, but I changed the value of `expected_from` to _not_ include the appended `Community` and I cannot figure out why `.value` still returns it _with_ the appended `Community` in the test. After looking into the mailer-related code and throwing `binding.pry`s in, I was unable to trace this `.value` back to anything that stood out as the culprit. Am I overthinking this? Should `Community` still remain appended in the spec? 😅 🤔

spec/mailers/notify_mailer_spec.rb
```diff
@@ -16,7 +16,7 @@
 
     it "renders proper sender" do
       expect(email.from).to eq([SiteConfig.email_addresses[:default]])
-      expected_from = "#{SiteConfig.community_name} Community <#{SiteConfig.email_addresses[:default]}>"
+      expected_from = "#{SiteConfig.community_name} <#{SiteConfig.email_addresses[:default]}>"
       expect(email["from"].value).to eq(expected_from)
     end
```
juliannatetreault, line 21: This (and all "community"-related `notify_mailer_spec` tests) is failing because `email["from"].value` returns `DEV (local) Community`, but I changed the value of `expected_from` to _not_ include the appended `Community` and I cannot figure out why `.value` still returns it _with_ the appended `Community` in the test. After looking into the mailer-related code and throwing `binding.pry`s in, I was unable to trace this `.value` back to anything that stood out as the culprit. Am I overthinking this? Should `Community` still remain appended in the spec? 😅 🤔

app/views/mailers/notify_mailer/tag_moderator_confirmation_email.text.erb
```diff
@@ -22,7 +22,7 @@ Tag moderation is something we're continuously iterating on, so you can expect a
 We regularly send out a biweekly Mod Newsletter to keep mods up to date on these changes, so look out for it!
 Of course, don't hesitate to reach out to us at any moment if you have any feedback — feel free to write to <%= SiteConfig.email_addresses[:contact] %> and share your thoughts.
 
-Thanks so much for volunteering your time to benefit the <%= community_qualified_name %>. We sincerely appreciate your help!
+Thanks so much for volunteering your time to benefit <%= community_name %>. We sincerely appreciate your help!
```
juliannatetreault, line 25: Prior to this PR and the previous PR, this read (in a few places):
```
the <%= community_name %> Community
```
Do we have a preference for which way it should be written? I think that the way it is currently written is consistent with the way it is written in other places in the codebase and reads well, but I'm open to suggestions!

app/lib/constants/site_config.rb
```diff
@@ -66,7 +62,7 @@ module SiteConfig
         placeholder: "user"
       },
       community_name: {
-        description: "Primary name... e.g. DEV",
+        description: "Used as the primary name, e.g. DEV, DEV Community, etc.",
```
juliannatetreault, line 65: Is this explicit enough?

spec/mailers/notify_mailer_spec.rb
vaidehijoshi: You're not overthinking this :) I believe your changes aren't showing up here the way we'd expect because there is a spot in the spec configuration/setup where we _explicitly_ override the "from" of an email 😅 

Check out `application_mailer.rb`, specifically these lines:
https://github.com/forem/forem/blob/23235c2276242a0ccd8213bcf3154d09507b541b/app/mailers/application_mailer.rb#L12-L15

Rails [lets you override](https://api.rubyonrails.org/classes/ActionMailer/Base.html) the default hash in the `ApplicationMailer`, which is what we're doing here (not sure _why_, but I am 99% sure it's just a remnant of when we used to hardcode `"Community"` everywhere 😛).

The real culprit here is this bit though, which is invoked in line `13` of the `ApplicationMailer` I linked to above:
https://github.com/forem/forem/blob/23235c2276242a0ccd8213bcf3154d09507b541b/app/mailers/application_mailer.rb#L17-L19

We pass in a `topic`, and then render `topic` between `community_name` and the default `email_address`. And....it looks like in line 13 of the `ApplicationMailer`, we were passing in the string `"Community"` as the topic 😩 

That's why it kept on showing up in those tests, even though you had actually removed it from the code everywhere else.

IMO, you could remove that `topic` parameter from the method signature entirely, so that the `email_from` method looks like this:
```ruby
def email_from
   "#{SiteConfig.community_name} <#{SiteConfig.email_addresses[:default]}>"
end
```

The only other thing I'd add here is that we _may_ want to change `community_name` on the DEV Forem instance to be `DEV Community` (? Maybe check on that with community support/@benhalpern) if we want the "from" in _our DEV emails_ to continue to be `"DEV Community <yo@dev.to>"` on production, rather than `"DEV <yo@dev.to>`, which is what it would change to as a result of my above suggestions :)

spec/requests/articles/articles_show_spec.rb
vaidehijoshi: Hm, is this still failing? I looked on Travis and I don't see any failures for this file (it looks like it was run in [this job](https://travis-ci.com/github/forem/forem/jobs/457485539) but that build is all green). It passes for me locally too ✅ 

Maybe it was flaky?

vaidehijoshi:
I was trying this out locally, and everything works great -- great job!! I noticed one odd behavior where the site logo doesn't seem to update as instantaneously as everywhere _else_ that you have changed the code 🤔 

Here's what I did, in case it's helpful for reproduction purposes!
1. My local Forem instance was called `The LLamaverse Community`. I (successfully) renamed it to `The Hippo Community`:
Screen Shot 2020-12-10 at 11 37 06 AM

2. I clicked the logo in the top left (which updated with the corrrect/new community name) and navigated out of the admin back to the feed. That's when I noticed that the community name was my _old_ community name:
Screen Shot 2020-12-10 at 11 37 16 AM

3. I navigated to the "terms and conditions" page, where I noticed that the text had the correct/new community name, but the logo hadn't updated yet:
Screen Shot 2020-12-10 at 11 37 40 AM

4. Finally, I refreshed the page, and that's when the logo _finally did update_ to match the community name as expected:
Screen Shot 2020-12-10 at 11 37 55 AM

I haven't investigated this, but thought I'd highlight it! 🤔

spec/mailers/notify_mailer_spec.rb
juliannatetreault: Thank you so much for the sanity check! `#{topic}` is really hidden in there. 🔎  

> The only other thing I'd add here is that we may want to change community_name on the DEV Forem instance to be DEV Community (? Maybe check on that with community support/@benhalpern) if we want the "from" in our DEV emails to continue to be "DEV Community <yo@dev.to>" on production, rather than "DEV <yo@dev.to>,

YES! This is something I was thinking about as well--great suggestion! 😄

spec/requests/articles/articles_show_spec.rb
juliannatetreault: I'm rerunning the build now, so hopefully it'll pass this time around! I could have totally been looking at an old build, as I had ~4 Tarvis tabs open. 😅

app/mailers/application_mailer.rb
```diff
@@ -14,8 +14,8 @@ class ApplicationMailer < ActionMailer::Base
     template_path: ->(mailer) { "mailers/#{mailer.class.name.underscore}" },
   )
 
-  def email_from(topic)
-    "#{SiteConfig.community_name} #{topic} <#{SiteConfig.email_addresses[:default]}>"
+  def email_from(_)
+    "#{SiteConfig.community_name} <#{SiteConfig.email_addresses[:default]}>"
```
juliannatetreault, line 18: I removed `#{topic}` per your suggestion, but was unable to remove the `topic` argument without running into some weird failures and this seemed to do the trick! cc @vaidehijoshi

juliannatetreault:
> I was trying this out locally, and everything works great -- great job!! I noticed one odd behavior where the site logo doesn't seem to update as instantaneously as everywhere _else_ that you have changed the code 🤔

It seems to be a caching issue that is resolved with a single, hard refresh!

> 1. I clicked the logo in the top left (which updated with the corrrect/new community name) and navigated out of the admin back to the feed. That's when I noticed that the community name was my _old_ community name:

I was able to reproduce this and immediately resolved it with a hard refresh! I will say that I noticed this happens with our logo, in general, a while back when I was working on a logo-related bug... I'm not really sure how to get around this but will continue to poke around. 

> 1. I navigated to the "terms and conditions" page, where I noticed that the text had the correct/new community name, but the logo hadn't updated yet:

Interestingly, I was not able to reproduce this--my "Terms and Conditions" page was updated immediately both times that I tried this... 🤔 

> I haven't investigated this, but thought I'd highlight it! 🤔

Thank you so much for highlighting this! ✨  I'm going to continue to poke around and see if I can get a better idea for _why_ this is happening in certain areas or not.

app/mailers/application_mailer.rb
vaidehijoshi: Did you remove the “Community” argument from where we invoke the email_from method as well?

https://github.com/forem/forem/blob/8f191d7fa4f1226aa57256b9b5b59b12b627aa8e/app/mailers/application_mailer.rb#L13

I tried it locally and removed it from both places (and didn’t run into any failures)

juliannatetreault: Giant 🤦‍♀️ ...

vaidehijoshi: It happens to the best of us! 🙈

And it's been a heck of a long year 😅

spec/mailers/notify_mailer_spec.rb
juliannatetreault: _Update_: I ended up reverting the changes I made to `application_mailer.rb` and the `notify_mailer_spec.rb` after a bit more digging. It actually appears that the `topic` parameter _is_ necessary and is used in another file and spec, `digest_mailer.rb and `digest_mailer_spec.rb`, for reasons other than appending "Community" to the `community_name`. Below is a snippet from the spec, which shows that the `topic` parameter is used, but not to append "Community" like in the `notify_mailer`:
```
"#{SiteConfig.community_name} Digest <#{SiteConfig.email_addresses[:default]}>"
```
That being said, it looks like the default of "Community" is overridable, so I don't think it'll be harmful keeping it as-is in the `application_mailer.rb` or `notify_mailer_spec.rb`, but I'm more than open to suggestions on how to better generalize this if possible! cc @vaidehijoshi @benhalpern

vaidehijoshi: Interesting, and good digging! I see two possible paths here:

1. We allow a optional parameter to be passed into email_from, and remove "Community", which would allow us to keep the concept of "from [some forem community name] email".
2. We remove the concept of "Digest"—I am not sure what the digest feature looks like for other Forems (maybe @benhalpern has more historical context here). Seems like we could just have all the "from" emails come from one singular "account" for simplicity's sake (it might also just be easier for Forem admins to keep mentally keep track of).

I would be fine either way, and personally think option 2 makes more sense holistically, but I also think that is out of the scope for this PR.

In the meantime, I do think we should remove "Community", but allow an optional param to be passed in (I think that's what you had previously, @juliannatetreault!). The reason I'd advocate to remove "Community" as the default hard-coded value is that if someone names their community "The Llama Community", then we'd have their emails coming from "The Llama Community Community". 😆

rhymes: This discussion is gold! Thanks @juliannatetreault and @vaidehijoshi ❤️

juliannatetreault: Thank _you_, @rhymes! 

**Update for those following this thread**: I was able to get the mailers and mailer-related specs to work by passing a default (`topic = ""`) to the `topic` param in `application_mailer.rb` and refactoring `#email_from` to be a conditional based upon the `topic`, if any, that is passed to the method. By doing this I was able to remove the default "Community" that was appended to the emails in `notify_mailer.rb` and was able to keep the appended "Digest" in `digest_mailer.rb`. Let me know what you all think about the approach! 😄 

cc @vaidehijoshi @benhalpern

app/mailers/application_mailer.rb
```diff
@@ -10,12 +10,16 @@ class ApplicationMailer < ActionMailer::Base
...
   )
 
-  def email_from(topic)
-    "#{SiteConfig.community_name} #{topic} <#{SiteConfig.email_addresses[:default]}>"
+  def email_from(topic = "")
+    if topic.present?
+      "#{SiteConfig.community_name} #{topic} <#{SiteConfig.email_addresses[:default]}>"
+    else
+      "#{SiteConfig.community_name} <#{SiteConfig.email_addresses[:default]}>"
+    end
```
vaidehijoshi, line 22:
```suggestion
  def email_from(topic = "")
    community_name = topic.present? ? "#{SiteConfig.community_name} #{topic}" : SiteConfig.community_name
    
    "#{community_name} <#{SiteConfig.email_addresses[:default]}>"
    end
```
A slightly cleaner way to write this -- since you know that you are always going to have `SiteConfig.community_name`, you could conditionally tack on the `topic` if it is present 😃

app/models/site_config.rb
```diff
@@ -51,8 +51,6 @@ class SiteConfig < RailsSettings::Base
   # Community Content
   field :community_name, type: :string, default: ApplicationConfig["COMMUNITY_NAME"] || "New Forem"
   field :community_emoji, type: :string, default: "🌱"
-  field :collective_noun, type: :string, default: "Community"
-  field :collective_noun_disabled, type: :boolean, default: false
```
vaidehijoshi, line 55: 🔥 this brings me great joy

_thank you_ for improving the siteconfig!!!

vaidehijoshi (approved):
this is _great_!! 👏🏽 really awesome work here. i left a suggestion on how to clean up one method, but i think this is a really great improvement to our siteconfig :)

app/mailers/application_mailer.rb
juliannatetreault: I **la, la, love** this! Thank you for the suggestion--this is great! ✨

lib/data_update_scripts/20201209222407_remove_collective_noun_from_config.rb
```diff
@@ -0,0 +1,7 @@
+module DataUpdateScripts
+  class RemoveCollectiveNounFromConfig
+    def run
+      SiteConfig.find_by(var: "collective_noun")&.or(SiteConfig.find_by(var: "collective_noun_disabled"))&.destroy
```
rhymes, line 4: We should use `.where()` here so that you can use `.or()` correctly and avoid using the `&` operator, if `collective_noun` doesn't exist, the second `OR` is not executed:

```shell
[1] pry(main)> SiteConfig.find_by(var: "collective_noun")&.or(SiteConfig.find_by(var: "collective_noun_disabled"))&.destroy
  User Exists? (52.5ms)  SELECT 1 AS one FROM "users" LIMIT $1  [["LIMIT", 1]]
  SiteConfig Load (7.4ms)  SELECT "site_configs".* FROM "site_configs" WHERE "site_configs"."var" = $1 LIMIT $2  [["var", "collective_noun"], ["LIMIT", 1]]
```

```suggestion
      SiteConfig.where(var: "collective_noun").or(SiteConfig.where(var: "collective_noun_disabled")).destroy_all
```

as you can see, this `OR` searches for both.

You can also use a simpler syntax:

```suggestion
      SiteConfig.where(var: %w[collective_noun collective_noun_disabled]).destroy_all
```

rhymes (changes requested):
Terrific job @juliannatetreault!

I only requested the change because there's an error in the data update script :)

lib/data_update_scripts/20201209222407_remove_collective_noun_from_config.rb
juliannatetreault: Thank you so much for bringing this to my attention and for breaking it all down for me! I really like the simplified syntax and went with that suggestion. 😊

rhymes (approved):
LGTM!

juliannatetreault:
**_Heads Up_**: Pending the final checks pass and the CI build is green, I'll be holding off on merging this until I get @benhalpern's approval! 😊

benhalpern (commented):
Approved, pending my understanding here:

This _will_ change certain situations from saying `DEV Community` to only saying `DEV` now, right?

I think this works in most places, but I wonder if there might be some additional text we use in certain areas, like perhaps the home page?

So `DEV Community` on the home page would now become just `DEV`, but we couldn't necessarily change this without changing _everything_? And on article pages it says `Such and such post about Reactjs - DEV`....

`Snyk Community` would change to just `Snyk`? I wonder if we want to data update script everything so that we don't enforce that change? I think in most place the leaf pages could still just say `— Snyk` but it seems like `Snyk` is not necessarily the correct home page title.

`collective_noun` is a pretty confusing concept and I think it makes sense to move away, but I don't know if we can just do away with a way to modify the home page?

What if we gave people the ability to name a specific title for the home page of their Forem? We could either use DEV or DEV Community, and Snyk could do the same? — And to start we automatically migrate everyone from the name + noun state to using that as the home page title?

@juliannatetreault @vaidehijoshi @rhymes thoughts on all this?

juliannatetreault:
> This _will_ change certain situations from saying `DEV Community` to only saying `DEV` now, right?
Correct!

> I think this works in most places, but I wonder if there might be some additional text we use in certain areas, like perhaps the home page?

Agreed! One thing that this PR does though, is adds additional context around the `community_name` in the SiteConfig so that if an Admin prefers that their community remains "<name> Community" they can set their name to be "DEV Community" or whatever their prefered community name is via the "Community Name" field in the config. I think that this provides enough flexibility and addresses your concerns around needing additional context in certain areas, while making sure that communities don't run into weird edge cases where their community name is automatically appended to "Community" in certain areas, resulting in a community name like "DEV Community Community." If you still do not think that this is sufficient, then I'm happy to explore other implementations!

> So `DEV Community` on the home page would now become just `DEV`, but we couldn't necessarily change this without changing _everything_? And on article pages it says `Such and such post about Reactjs - DEV`....

Correct, but we could update our community name to be "DEV Community" via the "Community Name" field in the SiteConfig, so that it could read "DEV Community" on the homepage and on article pages.

> `Snyk Community` would change to just `Snyk`? I wonder if we want to data update script everything so that we don't enforce that change? I think in most place the leaf pages could still just say `— Snyk` but it seems like `Snyk` is not necessarily the correct home page title.

Again, correct! We would have to have Snyk and any community that prefers to have some sort of "collective noun" appended to their name update their community name via the "Community Name" field in the SiteConfig. In this case, Snyk could update their name to "Snyk Community" via that field and going forward, new communities would just need to set their desired names (with or without a "collective noun") like this upon Forem set up. I've made the "Community Name" field more explicit so that future Admins have a better idea of this going forward. I wonder if a `data_update` script like you suggested would be a better way to handle the transition for current communities, though.

cc @benhalpern @vaidehijoshi @rhymes

vaidehijoshi:
> So DEV Community on the home page would now become just DEV, but we couldn't necessarily change this without changing _everything_?

> but I don't know if we can just do away with a way to modify the home page?

@benhalpern I think I am missing some context here: why do we want to treat the `community_name` on the homepage differently than everywhere else? Shouldn't the `community_name` always be consistent throughout the site?

> What if we gave people the ability to name a specific title for the home page of their Forem?

From what I remember about the [original issue](https://github.com/forem/InternalProjectPlanning/issues/221), the reason we started down the path of not enforcing the `community_name` to always be appended with `"Community"` at the end of it because certain Forems won't really make sense with `"Community"` tacked on at the end. So, I would think that if a Forem _does_ want to include `"Community"` at the end of it, then it would be on the Forem creator/admins to ensure that the `community_name` was updated to be `"Snyk Community"` rather than just `"Skyk"`.

>  I wonder if we want to data update script everything so that we don't enforce that change? 

I think this is a solid idea if you think that would adequately help address your concern! We could write a script that appends `"Community"` to every Forem, and then if a Forem creator _doesn't_ want the word `"Community"` appended to it, then they could just update it and remove it. It's worth noting, however, that this would be annoying for the few Forems that explicitly _don't_ have that right now (like "Community Club" or "Peatseekers Lounge"). Those Forems would have to go in and remove the `"Community"` from the `community_name` in the SiteConfig if we did this in a data update script, which will be a bit of nuisance.

lib/data_update_scripts/20201217184442_append_community_to_community_name.rb
```diff
@@ -0,0 +1,9 @@
+module DataUpdateScripts
+  class AppendCommunityToCommunityName
+    def run
+      SiteConfig.where(var: "community_name").find_each do |community_name|
+        community_name.update!(value: "#{community_name.value} Community")
+      end
+    end
+  end
+end
```
juliannatetreault, line 9: I added this `data_update` script that appends "Community" to `community_name` per your suggestions @vaidehijoshi and @benhalpern . How do we feel about this? I tested it out locally and it worked as expected (see screenshot below for an example of the `community_name` with the appended "Community" after running the `data_update` script locally)!
Screen Shot 2020-12-17 at 1 45 45 PM

juliannatetreault: Question: if we decide to go with this `data_update` script in this PR, will it be okay to keep _both_ `data_update` scripts in this PR or will it be necessary to split the two scripts up between this PR and another PR?

lib/data_update_scripts/20201217184442_append_community_to_community_name.rb
```diff
@@ -0,0 +1,9 @@
+module DataUpdateScripts
+  class AppendCommunityToCommunityName
+    def run
+      SiteConfig.where(var: "community_name").find_each do |community_name|
+        community_name.update!(value: "#{community_name.value} Community")
```
vaidehijoshi, line 5: Thinking about this some more and re-reading the code...

Should we actually be updating the `community_name` to take `collective_noun_disabled` into account?

We used to rely on and render `community_qualified_name` everywhere, which did this:
https://github.com/forem/forem/blob/978b698e7888386f5acf8e907a0ea6b6ffadbb46/app/helpers/application_helper.rb#L176-L180

If the `collective_noun_disabled` was _not_ disabled (aka if it was `false`) we used to build up the community name by interpolating `community_name` and the `SiteConfig.collective_noun`. But, if `collective_noun_disabled` _was_ `true`, then we just used to return `community_name`.

If we just directly use what we have in this data update script:
```ruby
community_name.update!(value: "#{community_name.value} Community")
```

then we are no longer taking that `collective_noun_disabled` into account, which means that any Forem instance that had _not_ disabled the collective noun feature would have `"Community"` appended to it...and that might be an unexpected change that would take those Forem creators by surprise! 😅 

I think could be clever with this data update script and _use_ the `collective_noun_disabled` concept to figure out if we even _need_ to append `"Community"` to the end of the `community_name`. Of course, we'd have to do this _before_ we destroy those fields from the SiteConfig, so order of operations would be essential here.

If we want to reduce the friction and any unexpected community name changes for _all_ Forem instances, I _think_ we want to do this in the data update script:
1) Check the value of `collective_noun_disabled` for the Forem's SiteConfig
2) If the "collective noun" concept has been disabled (`collective_noun_disabled` is `true`)
`collective_noun_disabled`, then we don't need to update anything
3) But if the "collective noun" concept has _not_ been disabled (`collective_noun_disabled` is `false`), then we want to update the `community_name` to be: `"#{community_name} #{SiteConfig.collective_noun}`, which matches [what we do](https://github.com/forem/forem/blob/master/app/helpers/application_helper.rb#L177) in the `community_qualified_name` method (the method that you are eliminating in this PR!)

The other reason I bring this up is that, currently, this data update script is appending the hard-coded string, `"Community"`. But we have this concept of `collective_noun` in the `SiteConfig`, and if there is any Forem out there that overrode the default value of `collective_noun`, we won't be taking that into account with this hardcoded string. 🤔

juliannatetreault: This all makes sense to me! Thank you for the detailed explanation. 😄  I'll make these changes to the `data_update` script and remove the `data_update` script that removes `collective_noun` and `collective_noun_disabled` into a separate PR to ensure proper removal and re-request your review @vaidehijoshi!

juliannatetreault: I've made this change! The newly refined `data_update` script that checks whether `collective_noun_disabled` is true or false and updates the `community_name` based on that value has been updated in this PR, while the `data_update` script that removes `collective_noun` and `collective_noun_disabled` from the SiteConfig has been removed from this PR and opened as a separate PR, which can be found [here](https://github.com/forem/forem/pull/11973).

cc @vaidehijoshi @benhalpern

juliannatetreault:
I believe that I've addressed everyone's concerns and have made all necessary changes to the `data_update` scripts, including removing one from this PR and opening it in a separate PR, which can be found here: #11973. Additionally, I have removed the "Collective Noun" section of the Admin Guide and have updated the "Community Name" section based on all of these changes -- the draft for this work can be found here and all suggestions around wording are welcome: https://app.gitbook.com/@forem/s/forem-admin-guide/~/drafts/-MOr2aA_SCj61eSjxPLG/admin/config/community-content

cc @vaidehijoshi @benhalpern 😄

lib/data_update_scripts/20201217184442_append_community_to_community_name.rb
```diff
@@ -0,0 +1,11 @@
+module DataUpdateScripts
+  class AppendCommunityToCommunityName
+    def run
+      return unless SiteConfig.collective_noun_disabled.false?
+
+      SiteConfig.where(var: "community_name").find_each do |community_name|
+        community_name.update!(value: "#{community_name.value} #{SiteConfig.collective_noun}")
+      end
```
vaidehijoshi, line 8: Is the plan to add another data update script in a separate PR to drop the `collective_noun` and `collective_noun_disabled` fields?

lib/data_update_scripts/20201217184442_append_community_to_community_name.rb
```diff
@@ -0,0 +1,11 @@
+module DataUpdateScripts
+  class AppendCommunityToCommunityName
+    def run
+      return unless SiteConfig.collective_noun_disabled.false?
```
vaidehijoshi, line 4:
```suggestion
      return unless SiteConfig.collective_noun_disabled
```
I think you don't need the `false?` here since `SiteConfig.collective_noun_disabled` returns a boolean and has a default value, so it should _itself_ return `true` or `false`

lib/data_update_scripts/20201217184442_append_community_to_community_name.rb
vaidehijoshi: This is a nice, elegant solution and a great improvement on the previous data update script! 👏🏽 Nice job

vaidehijoshi (approved):
One minor comment around syntax, but otherwise, I think this is a nice improvement on what this PR originally proposed 👍🏽

lib/data_update_scripts/20201217184442_append_community_to_community_name.rb
juliannatetreault: Great ctach! I've updated this

citizen428: Unimportant random comment from the sidelines: the gem we use for `SiteConfig` follows the AR convention of creating predicate methods for boolean options, so this also works and is probably the most idiomatic:

```ruby
return unless SiteConfig.collective_noun_disabled?
```

benhalpern (approved):
Sorry for my delay @juliannatetreault — This looks good pending the conflict fixes.

@vaidehijoshi I think I was overthinking this. We currently don't append the "collective noun" at the end of article pages with the concern that it makes for an extra long title tag... But I'm cool to worry less about this and just simplify how we do things.

lib/data_update_scripts/20201217184442_append_community_to_community_name.rb
juliannatetreault: Updated this @citizen428!

PR #11856, opened 2020-12-11
Make mod button only show up for trusted user

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

There is a current bug which is having the mod action button in the nav bar be _always_ present. There is one missing class we use to toggle this based on stuff we store client side about the user.

Manually change the `config_body_class` in `user_decorator` to verify that this is showing up properly or not. You'll probably need to refresh twice to see the change.

## Tests

We _just_ implemented end-to-end testing with cypress and I think to get this hot fix in we should just ship, but this is the kind of thing that could warrant a regression test in the future.

### DIFF
--- a/app/views/layouts/_top_bar.html.erb
+++ b/app/views/layouts/_top_bar.html.erb
@@ -32,7 +32,7 @@
       <% if user_signed_in? %>
         <a href="<%= new_path %>" class="crayons-btn hidden mr-2 whitespace-nowrap m:block ml-auto">Write a post</a>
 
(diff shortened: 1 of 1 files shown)


ludwiczakpawel:
It seems like pretty simple fix that SHOULD work but I can't really test it due to some issues with my local env. Therefor summoning the cavalry cc @forem/frontend

rhymes (approved):
I tried by adding and removing the trusted role, it works

Ridhwana (approved):

PR #11863, opened 2020-12-11
Disable Google reCAPTCHA in env file sample

rhymes:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

`"Optional"` is not an empty string which leads to unnecessary checks in development and potentially a broken captcha:

https://github.com/forem/forem/blob/bcadebb3efc003f5e46a234d572993ea0cfa6149/app/services/re_captcha/check_enabled.rb#L18-L19

https://github.com/forem/forem/blob/bcadebb3efc003f5e46a234d572993ea0cfa6149/app/services/re_captcha/check_enabled.rb#L33-L35

## Added tests?

- [ ] Yes
- [x] No, and this is why: they are already tested, it's just the template defaulting to a valued environment key

### DIFF
--- a/.env_sample
+++ b/.env_sample
@@ -110,8 +110,8 @@ export PUSHER_BEAMS_KEY=
 
 # Google recaptcha
 # (https://developers.google.com/recaptcha/intro)
-export RECAPTCHA_SECRET="Optional"
-export RECAPTCHA_SITE="Optional"
+export RECAPTCHA_SECRET=
+export RECAPTCHA_SITE=
 
 # Slack for customer alerts
 # (https://api.slack.com/)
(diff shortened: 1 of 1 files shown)


benhalpern (approved):

.env_sample
```diff
@@ -110,8 +110,8 @@ export PUSHER_BEAMS_KEY=
 
 # Google recaptcha
```
jacobherrington, line 111:
```suggestion
# Google recaptcha (Optional)
```

Do you think adding this information preserves the spirit of the original while still fixing this edge-case?

jacobherrington (approved):
LGTM!

I think we could add the comment I suggested, but we can merge it now if you feel like that is unnecessary. 👍

rhymes:
Thanks @jacobherrington!

PR #11870, opened 2020-12-11
Make followers notification overflow-hidden

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In the event that the new follower box overflows, it's better that it be hidden vs escape the box.

Screen Shot 2020-12-11 at 4 16 08 PM

There are alternatives to fixing this, maybe it spans multiple lines, but I think this hot fix is better than leaving the problem as it is.


rhymes (approved):

jacobherrington (approved):
:adhesive_bandage:

PR #11874, opened 2020-12-12
Fix feed navigation shortcuts when first article is not 100% visible

### LINKED ISSUE #11566: Feed navigation does not initialize if the first article is not 100% visible
**Describe the bug**
When entering a page, and pressing `j` or `k`, the list navigation shortcut looks for the first visible article that appears on the screen, and focuses on that.

A problem that occurs is that if the first article in the list is longer than the window's height, it is not considered "visible", so the navigation does not start.

That is because the `isInViewport` function used by the navigation shortcut checks if the element's bounding rectangle is 100% visible in the current window.

_Note_: Perhaps now would be a good time to replace `isInViewport` with the Intersection Observer API, as mentioned by Nick in #10468.
[See comment](https://github.com/forem/forem/pull/10468#discussion_r520201058)

**To Reproduce**
1. Navigate to [/top/infinity](https://dev.to/top/infinity)
2. Make sure your browser's window is not big enough to fit the entire first article
3. Try to initialize the article feed navigation (pressing `j` and `k`)
4. Notice nothing is happening

**Screenshots**
Notice that the first article is longer than the current window.
Screen Shot 2020-11-23 at 08 19 10

**Expected behavior**
When no article is focused on, the first article that is visible should be focused on.

We might want to have a similar behaviour to the current one: we look for the first article that is fully visible, and then fall back to any that is partially visible if nothing is found.

In a case like the following, we would not want the `Announcing Word Wrap Podcast` (partially visible) to be focused on first, but rather the `11 reasons` one:
Screen Shot 2020-11-23 at 08 37 21

- OS, version: macOS 11
- Browser, version: Safari 14

**Additional context**
- Original PR for article feed navigation: #10468
- [List navigation hook](https://github.com/forem/forem/blob/cc4a17992adc898a1490223506b7eb636fb3ce34/app/javascript/shared/components/useListNavigation.js#L45)
- [`isInViewport` function](https://github.com/forem/forem/blob/cc4a17992adc898a1490223506b7eb636fb3ce34/app/javascript/utilities/viewport.js#L13)

gowtham2095:

- [ ] Refactor
- [ ] Feature
- [x]  Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
 The feature was introduced to navigate between articles using the keyboard commands j and k. Previously we were looking for the completely visible first element. There could be cases where we may have partially visible elements, but no completely visible element. 

Closes https://github.com/forem/forem/issues/11566

### To reproduce

* Navigate to /top/infinity
* Make sure your browser's window is not big enough to fit the entire first article
* Try to initialize the article feed navigation (pressing j and k)
* Notice nothing is happening

### Expected behavior

When no article is focused on, the first article that is visible should be focused on.

We might want to have similar behavior to the current one: we look for the first article that is fully visible and then fall back to any that is partially visible if nothing is found. 

It helps improves accessibility by responding to keyboard events to navigate between articles 

## Added tests?

- [ ] Yes
- [ ] No, and this is why:

## Added to documentation?

- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/javascript/shared/components/useListNavigation.js
+++ b/app/javascript/shared/components/useListNavigation.js
@@ -152,7 +152,7 @@ function navigate(
   const nextFocusable = nextContainer?.querySelector(focusableSelector);
   if (nextFocusable) {
     nextFocusable.focus();
-    if (!isInViewport(nextFocusable, 64)) {
+    if (!isInViewport({ element: nextFocusable, offsetTop: 64 })) {
       window.scrollTo({ top: nextContainer.offsetTop - 64 });
     }
   }
@@ -211,7 +211,9 @@ function getPreviousElement(
 }
 
 /**
- * Gets the first visible element that matches a selector
+ * Checks if the first completely visible element is present that matches a selector and returns if it is available
+ * If that isn't visible, it looks for the partially visible element that is present and returns that
+ * If no elements visible(like the banner case which could cover the entire viewport), we select the first element from the list
  *
  * @param {string} selector - The CSS selector
  *
@@ -219,7 +221,19 @@ function getPreviousElement(
  */
 function getFirstVisibleElement(selector) {
   const elements = document.querySelectorAll(selector);
-  return Array.prototype.find.call(elements, (element) =>
-    isInViewport(element),
+  const completelyVisibleFirstElement = Array.prototype.find.call(
+    elements,
+    (element) => isInViewport({ element }),
   );
+  if (completelyVisibleFirstElement) {
+    return completelyVisibleFirstElement;
+  }
+  const partiallyVisibleFirstElement = Array.prototype.find.call(
+    elements,
+    (element) => isInViewport({ element, allowPartialVisibility: true }),
+  );
+  if (partiallyVisibleFirstElement) {
+    return partiallyVisibleFirstElement;
+  }
+  return elements && elements[0];
 }
--- a/app/javascript/shared/components/useListNavigation.js
+++ b/app/javascript/shared/components/useListNavigation.js
@@ -220,20 +220,18 @@ function getPreviousElement(
  * @returns {object} The first visible element
  */
 function getFirstVisibleElement(selector) {
-  const elements = document.querySelectorAll(selector);
-  const completelyVisibleFirstElement = Array.prototype.find.call(
-    elements,
-    (element) => isInViewport({ element }),
+  const elements = [...document.querySelectorAll(selector)];
+  const completelyVisibleFirstElement = elements.find((element) =>
+    isInViewport({ element }),
   );
   if (completelyVisibleFirstElement) {
     return completelyVisibleFirstElement;
   }
-  const partiallyVisibleFirstElement = Array.prototype.find.call(
-    elements,
-    (element) => isInViewport({ element, allowPartialVisibility: true }),
+  const partiallyVisibleFirstElement = elements.find((element) =>
+    isInViewport({ element, allowPartialVisibility: true }),
   );
   if (partiallyVisibleFirstElement) {
     return partiallyVisibleFirstElement;
   }
-  return elements && elements[0];
+  return elements[0];
 }
--- a/app/javascript/utilities/viewport.js
+++ b/app/javascript/utilities/viewport.js
@@ -3,22 +3,45 @@
  *
  * @example
  * const element = document.getElementById('element');
- * isInViewport(element); // true or false
+ * isInViewport({element, allowPartialVisibility = true}); // true or false
  *
  * @param {object} element - The HTML element to check
  * @param {number} [offsetTop=0] - Part of the screen to ignore counting from the top
- *
+ * @param {boolean} [allowPartialVisibility=false] - A boolean to flip the check between partial or completely visible in the viewport
  * @returns {boolean} isInViewport - true if the element is visible in the viewport
  */
-export function isInViewport(element, offsetTop = 0) {
+export function isInViewport({
+  element,
+  offsetTop = 0,
+  allowPartialVisibility = false,
+}) {
   const boundingRect = element.getBoundingClientRect();
   const clientHeight =
     window.innerHeight || document.documentElement.clientHeight;
   const clientWidth = window.innerWidth || document.documentElement.clientWidth;
+  const topIsInViewport =
+    boundingRect.top <= clientHeight && boundingRect.top >= offsetTop;
+  const rightIsInViewport =
+    boundingRect.right >= 0 && boundingRect.right <= clientWidth;
+  const bottomIsInViewport =
+    boundingRect.bottom >= offsetTop && boundingRect.bottom <= clientHeight;
+  const leftIsInViewport =
+    boundingRect.left <= clientWidth && boundingRect.left >= 0;
+  const topIsOutOfViewport = boundingRect.top <= offsetTop;
+  const bottomIsOutOfViewport = boundingRect.bottom >= clientHeight;
+  const elementSpansEntireViewport =
+    topIsOutOfViewport && bottomIsOutOfViewport;
+
+  if (allowPartialVisibility) {
+    return (
+      (topIsInViewport || bottomIsInViewport || elementSpansEntireViewport) &&
+      (leftIsInViewport || rightIsInViewport)
+    );
+  }
   return (
-    boundingRect.top >= offsetTop &&
-    boundingRect.left >= 0 &&
-    boundingRect.bottom <= clientHeight &&
-    boundingRect.right <= clientWidth
+    topIsInViewport &&
+    bottomIsInViewport &&
+    leftIsInViewport &&
+    rightIsInViewport
   );
 }
--- a/app/javascript/utilities/viewport.js
+++ b/app/javascript/utilities/viewport.js
@@ -5,7 +5,7 @@
  * const element = document.getElementById('element');
  * isInViewport({element, allowPartialVisibility = true}); // true or false
  *
- * @param {object} element - The HTML element to check
+ * @param {HTMLElement} element - The HTML element to check
  * @param {number} [offsetTop=0] - Part of the screen to ignore counting from the top
  * @param {boolean} [allowPartialVisibility=false] - A boolean to flip the check between partial or completely visible in the viewport
  * @returns {boolean} isInViewport - true if the element is visible in the viewport


gowtham2095:
https://github.com/forem/forem/issues/11566 Issue Discussion thread

app/javascript/utilities/viewport.js
```diff
@@ -22,3 +22,37 @@ export function isInViewport(element, offsetTop = 0) {
...
+ * @param {number} [offsetTop=0] - Part of the screen to ignore counting from the top
+ *
+ * @returns {boolean} isInViewport - true if the element is visible in the viewport
+ **/
+export function isPartiallyVisibleInViewport(element, offsetTop = 0) {
+  const boundingRect = element.getBoundingClientRect();
+  const clientHeight =
+    window.innerHeight || document.documentElement.clientHeight;
+  const clientWidth = window.innerWidth || document.documentElement.clientWidth;
+  return (
```
reobin, line 44: There is some stuff that can be reused in all those conditions, and it will help make the condition be more readable overall.

I would consider creating variables for each of the bounding rectangle's sides that defines is that particular side is visible in the viewport.

For example:
```js
const topIsInViewport = boundingRect.top <= clientHeight && boundingRect.top >= offsetTop;
const rightIsInViewport = boundingRect.right >= 0 && boundingRect.right <= clientWidth;
const bottomIsInViewport = boundingRect.bottom >= offsetTop && boundingRect.bottom <= clientHeight;
const leftIsInViewport = boundingRect.left <= clientWidth && boundingRect.left >= 0;
```

That way, you can use those variables in any combination in order to get the value you want.

For example, if you want to know if an element is 100% in the viewport, then all of those variables need to be true:
```js
return (
  topIsInViewport &&
  bottomIsInViewport &&
  leftIsInViewport &&
  rightIsInViewport
);
```

And if you want to know if an element is partially in the viewport, it needs to satisfy the following 2 conditions:
- its top OR its bottom is visible
- its left OR its right is visible

That way, you can use the following condition:
```js
return (topIsInViewport || bottomIsInViewport) && (leftIsInViewport || rightIsInViewport);
```

With all of that said, I think we can merge our 2 viewport functions. We can pass it an optional parameter that defines if we want the element to be partially visible or completely visible. We can then use the corresponding condition in our return statement.

gowtham2095: Thank you so much for taking the time to review it and giving valuable suggestions @reobin. I have made the changes as per your suggestion. 

And also there was one more following case apart from the above-mentioned ones,

```
const topIsOutOfViewport = boundingRect.top <= offsetTop;
const bottomIsOutOfViewport = boundingRect.bottom >= clientHeight;
const elementSpansEntireViewport =
    topIsOutOfViewport && bottomIsOutOfViewport;
```

The _elementSpansEntireViewport_ should also be part of the partially visible section check as below

```
return (topIsInViewport || bottomIsInViewport || elementSpansEntireViewport) &&
        (leftIsInViewport || rightIsInViewport)
```

I have added the changes and pushed them. Review it when you find the time and let me know. Cheers!

gowtham2095: Hey @reobin!

I have made the changes you suggested. Can you go over it when you find the time and let me know if it is fine?

juliannatetreault:
Hi, @gowtham2095! 👋  I just wanted to check in to see if you're still working on this PR. Please let us know if you have any questions or if you need any help!

gowtham2095:
> Hi, @gowtham2095! 👋 I just wanted to check in to see if you're still working on this PR. Please let us know if you have any questions or if you need any help!

Hey @juliannatetreault!👋 Hope you are well. I guess it is done. It has to be reviewed by two people. I'm waiting for the green flag from the second reviewer. Not sure if I have missed out on something. If there is something that needs to be done to push it forward, kindly let me know.

juliannatetreault:
> Hey @juliannatetreault!👋 Hope you are well. I guess it is done. It has to be reviewed by two people. I'm waiting for the green flag from the second reviewer. Not sure if I have missed out on something. If there is something that needs to be done to push it forward, kindly let me know.

Hey, @gowtham2095! 

Thank you so much for the update! It looks like my awesome co-worker, @rhymes, added another reviewer to the PR, which should help move the review process along for you. 😊  If you have any further questions, please do not hesitate to reach out to me or the reviewers!

app/javascript/utilities/viewport.js
reobin: Hey @gowtham2095!

Sorry for the delay on the review! I have been very busy with personal life stuff. I'll let the Forem team take care of this. Good work!

gowtham2095: Hi @reobin,

I understand. Not a problem. Take care!

gowtham2095:
> > Hey @juliannatetreault!👋 Hope you are well. I guess it is done. It has to be reviewed by two people. I'm waiting for the green flag from the second reviewer. Not sure if I have missed out on something. If there is something that needs to be done to push it forward, kindly let me know.
> 
> Hey, @gowtham2095!
> 
> Thank you so much for the update! It looks like my awesome team member, @rhymes, added another reviewer to the PR, which should help move the review process along for you. 😊 If you have any further questions, please do not hesitate to reach out to me or the reviewers!

That sounds great. Thank you @rhymes and @juliannatetreault :)

app/javascript/shared/components/useListNavigation.js
```diff
@@ -211,15 +211,29 @@ function getPreviousElement(
...
  *
  * @param {string} selector - The CSS selector
  *
  * @returns {object} The first visible element
  */
 function getFirstVisibleElement(selector) {
   const elements = document.querySelectorAll(selector);
-  return Array.prototype.find.call(elements, (element) =>
-    isInViewport(element),
+  let completelyVisibleFirstElement = Array.prototype.find.call(
```
aitchiss, line 224: we don't change this so we can use `const`

```suggestion
  const completelyVisibleFirstElement = Array.prototype.find.call(
```

app/javascript/shared/components/useListNavigation.js
```diff
@@ -211,15 +211,29 @@ function getPreviousElement(
...
+    elements,
+    (element) => isInViewport(element),
   );
+  if (completelyVisibleFirstElement) {
+    return completelyVisibleFirstElement;
+  }
+  let partiallyVisibleFirstElement = Array.prototype.find.call(
```
aitchiss, line 231: same here re: const

```suggestion
  const partiallyVisibleFirstElement = Array.prototype.find.call(
```

app/javascript/shared/components/useListNavigation.js
```diff
@@ -211,15 +211,29 @@ function getPreviousElement(
 }
 
 /**
- * Gets the first visible element that matches a selector
+ * Checks if the first completely visible element is present that matches a selector and returns if it is available
+ * If that isn't visible, it looks for the partially visible element that is present and returns that
+ * If no elements visible(like the banner case which could cover the entire viewport), we select the first element from the list
```
aitchiss, line 216: Nice! This is a really helpful summary

app/javascript/utilities/viewport.js
```diff
@@ -7,18 +7,35 @@
...
- *
+ * @param {boolean} [isPartiallyVisible=false] - A boolean to flip the check between partial or completely visible in the viewport
  * @returns {boolean} isInViewport - true if the element is visible in the viewport
  */
-export function isInViewport(element, offsetTop = 0) {
+export function isInViewport(
+  element,
+  offsetTop = 0,
+  isPartiallyVisible = false,
+) {
```
aitchiss, line 17: (refactor/blocking) - the updates here seem to work great! Just a small change requested here to make the developer experience a bit easier now that we have three params - let's change it to expect them in an object that we destructure like so:

```suggestion
export function isInViewport({
  element,
  offsetTop = 0,
  acceptPartiallyVisible = false,
}) {
```

So when we call this function it's clearer what the params are, e.g. `isInViewport({element: myElement, acceptPartiallyVisible: true})`

app/javascript/utilities/viewport.js
```diff
@@ -7,18 +7,35 @@
...
+  const bottomIsOutOfViewport = boundingRect.bottom >= clientHeight;
+  const elementSpansEntireViewport =
+    topIsOutOfViewport && bottomIsOutOfViewport;
+  return isPartiallyVisible
+    ? (topIsInViewport || bottomIsInViewport || elementSpansEntireViewport) &&
+        (leftIsInViewport || rightIsInViewport)
+    : topIsInViewport &&
+        bottomIsInViewport &&
+        leftIsInViewport &&
+        rightIsInViewport;
```
aitchiss, line 40: (nitpick / not-blocking) - I think in this case the ternary makes it a little harder to read as we have so many conditionals chained together. I'd suggest splitting out into an `if` to make it more readable:

```js
 if (allowPartialVisibility) {
    return (
      (topIsInViewport || bottomIsInViewport || elementSpansEntireViewport) &&
      (leftIsInViewport || rightIsInViewport)
    );
  }

  return (
    topIsInViewport &&
    bottomIsInViewport &&
    leftIsInViewport &&
    rightIsInViewport
  );
```

aitchiss (changes requested):
Great to see this fix! I've given it a quick spin and it seems to be working great - just a few small suggestions here mostly to make the additions a bit more readable 👍

gowtham2095:
> Great to see this fix! I've given it a quick spin and it seems to be working great - just a few small suggestions here mostly to make the additions a bit more readable 👍

Hey @aitchiss ,

Thanks for taking the time to provide valuable suggestions. I have addressed the review comments as a separate commit and pushed it back. Let me know if it is fine. Will squash them as a single commit and push it back.

app/javascript/shared/components/useListNavigation.js
gowtham2095: Thank you! 😊

app/javascript/shared/components/useListNavigation.js
gowtham2095: Makes sense. Have addressed the same.

app/javascript/shared/components/useListNavigation.js
gowtham2095: Have addressed the same.

app/javascript/utilities/viewport.js
gowtham2095: Yes, plus it gives us flexibility as well. And also a lot more clearer. 👍

app/javascript/utilities/viewport.js
gowtham2095: Thanks! Yes, the ternary demanded more cognitive load. This looks a lot better.

app/javascript/utilities/viewport.js
```diff
@@ -3,22 +3,45 @@
...
- *
+ * @param {boolean} [allowPartialVisibility=false] - A boolean to flip the check between partial or completely visible in the viewport
  * @returns {boolean} isInViewport - true if the element is visible in the viewport
  */
-export function isInViewport(element, offsetTop = 0) {
+export function isInViewport({
+  element,
+  offsetTop = 0,
+  allowPartialVisibility = false,
+} = {}) {
```
aitchiss, line 17: (refactor/blocking) I don't think we need the default empty object here - we don't want the function to be called without at least the `element` property, so we can safely expect an object passed in

```suggestion
}) {
```

aitchiss (changes requested):
Thanks for making those changes so quickly! Just one small thing left.

Don't forget to also sign the CLA linked above 🙂

app/javascript/utilities/viewport.js
gowtham2095: > Thanks for making those changes so quickly! Just one small thing left.
> 
> Don't forget to also sign the CLA linked above 🙂

@aitchiss Hey! 👋  I signed the CLA, but it isn't picking up. Not sure what is missed. Let me try again.

gowtham2095:
> Thanks for making those changes so quickly! Just one small thing left.
> 
> Don't forget to also sign the CLA linked above 🙂

Hey, @aitchiss!  Have included that minor change and also signed CLA. Let me know if anything else needs to be covered 😊. Cheers!

aitchiss (approved):
Looks good to me! 🚀

citizen428:
@nickytonline Pinging you here for a second approval for this PR, thanks!

msarit:
@forem/frontend Hey FrontEnd Team! 👋🏾  This PR needs one more review, thanks! 🙏🏾

app/javascript/utilities/viewport.js
```diff
@@ -3,22 +3,45 @@
...
+  offsetTop = 0,
+  allowPartialVisibility = false,
+}) {
   const boundingRect = element.getBoundingClientRect();
   const clientHeight =
     window.innerHeight || document.documentElement.clientHeight;
   const clientWidth = window.innerWidth || document.documentElement.clientWidth;
+  const topIsInViewport =
+    boundingRect.top <= clientHeight && boundingRect.top >= offsetTop;
+  const rightIsInViewport =
```
nickytonline, line 24: question (potentially blocking): Is it necessary to check if the right and left of the element is in the viewport? Shouldn't top and bottom be enough?

Also, since this is being revisited, we could probably leverage the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)? Or maybe save that for a subsequent PR, cc: @aitchiss.

nickytonline (commented):
One small change request and a question.

app/javascript/utilities/viewport.js
aitchiss: Yeah I agree it would be nice to rework this to use intersection observer - I'd be happy to tackle in a later PR though, since the current implementation fixes the bug 👍

app/javascript/utilities/viewport.js
```diff
@@ -3,22 +3,45 @@
  *
  * @example
  * const element = document.getElementById('element');
- * isInViewport(element); // true or false
+ * isInViewport({element, allowPartialVisibility = true}); // true or false
  *
  * @param {object} element - The HTML element to check
```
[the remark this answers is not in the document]
gowtham2095: @nickytonline Thanks for taking the time to review. I shall add the same and push.

app/javascript/utilities/viewport.js
gowtham2095: Hey, @nickytonline! I have avoided using the Intersection observer since I felt it will be a little overhead. Intersection Observer will be listening to the viewport and callbacks will be triggered when articles come in and leave the viewport. I felt that is not what is required here. What we need is when 'j' and 'k' key events happen, we have to let the user select some element based on what is visible in the viewport. I think the requirement here is to select from what is present in the viewport rather than observing what comes in and leaves the viewport. So felt the Intersection API might not be necessary. Correct me if my understanding is wrong.

gowtham2095: And also regarding the _left_ and _right_ check. Since this "_isInViewport_" function is part of the utilities I had included the left and right check as well to fit its correct definition. If it is specific for only this case I can remove those checks as well.

aitchiss:
@nickytonline - any chance you're able to re-review now changes have been made?

app/javascript/shared/components/useListNavigation.js
```diff
@@ -211,15 +211,29 @@ function getPreviousElement(
...
+    return completelyVisibleFirstElement;
+  }
+  const partiallyVisibleFirstElement = Array.prototype.find.call(
+    elements,
+    (element) => isInViewport({ element, allowPartialVisibility: true }),
+  );
+  if (partiallyVisibleFirstElement) {
+    return partiallyVisibleFirstElement;
+  }
+  return elements && elements[0];
```
nickytonline, line 238: Refactor (blocking): Instead of Array.prototype, use the `find` method on an actual array. This can be accomplished by [spreading](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) the [NodeList](https://developer.mozilla.org/en-US/docs/Web/API/NodeList) into an array.

```suggestion
  const elements = [...document.querySelectorAll(selector)];
  const completelyVisibleFirstElement = elements.find((element) =>
    isInViewport({ element }),
  );

  if (completelyVisibleFirstElement) {
    return completelyVisibleFirstElement;
  }

  const partiallyVisibleFirstElement = elements.find((element) =>
    isInViewport({ element, allowPartialVisibility: true }),
  );

  if (partiallyVisibleFirstElement) {
    return partiallyVisibleFirstElement;
  }

  return elements[0];
}
```

nickytonline (changes requested):
Tested this locally @gowtham2095 and it works great. One small change request. Looking forward to seeing this in production! 🚀

app/javascript/shared/components/useListNavigation.js
gowtham2095: Sure @nickytonline. Thanks for taking the time to go over. Will push the suggested change.

gowtham2095: Have pushed it @nickytonline. Kindly check and let me know when you find time :)

nickytonline (approved):
Thanks so much for the PR @gowtham2095 and thanks for making those changes! Looking forward to seeing this in production. 🚢 😎

Issue #11855, opened 2020-12-11
NoMethodError in Stories#show


0xC000005:
**Describe the bug**

Hi, I'm having a `NoMethodError in Stories#show` error when I trying to access any article ....... 
The error message is ```ActionView::Template::Error: undefined method `brand_color1' for #<Profile:0x00005569dd8f12f8>```, so I try to access articles without logging on an account and this error occur anyways. 

- Error Page: http://yuegaoya.org/jareddvmgrant/a-many-splendoured-thing-excepturi-omnis-58p3
- Honeybadger Bug Report: https://app.honeybadger.io/fault/79424/8a09f8d40889fae03b50383b90e0d469

**To Reproduce**

![](https://raw.githubusercontent.com/0xC000005/image-hosting/master/20201210172930.gif)

**Expected behavior**

**Screenshots**
![](https://raw.githubusercontent.com/0xC000005/image-hosting/master/20201210173106.png)

**Desktop (please complete the following information):**
- OS, version: `Ubuntu 18.04`
- Browser, version: `Microsoft Edge Version 87.0.664.57 (Official build) (64-bit)`

**Smartphone (please complete the following information):**

- Device:
- OS, version:
- Browser, version:

**Additional context**

I'm translating some page text into another language...... maybe I messed up (っ °Д °;)っ 
This is my project: https://github.com/0xC000005/yuegaoya

LE-HU:
Hello,
Did the error appear after your last commit? That is -
**Merge remote-tracking branch 'origin/master' into 0xC00005/dev**
Can you undo the last commit? It shows some major changes not related to translation. Please research the web how to revert the git repository to a previous commit.

citizen428:
Can you please check if you have all current profile fields in your DB? There should be 29 of them.

If you don't have all of them, can you please try the following in a Rails console?

```ruby
csv = Rails.root.join("lib/data/dev_profile_fields.csv")
ProfileFields::ImportFromCsv.call(csv)
```

Then restart your Rails server and see if things work.

LE-HU:
Meanwhile I've set up my own forem instance on local ubuntu following tutorial & using bin/setup method, and encountered the same error as the author.
NoMethodError also spotted at http://localhost:3000/admin_mcadmin 
```ruby
undefined method `display_email_on_profile' for #<Profile:0x00007f01f4b69e80>
```
[...]

PR #11783, opened 2020-12-07
Move profile initialization to after_initialize hook

citizen428:
## What type of PR is this? (check all applicable)

- [X] Refactor

## Description

This PR moves profile initialization from a custom hack to a proper Rails solution using `config.after_initialize`. Thanks to @rhymes for pushing me to give this another try, I have no idea why it didn't work the last time I tried exactly the same 🤷 (well, I have some theories, but it really doesn't matter).

## Related Tickets & Documents

#6994 

## QA Instructions, Screenshots, Recordings

1. Ensure sure all specs pass.
2. Drop your database and make sure you can run `bin/setup` to completion.
3. Ensure profiles are ready to use right after app boot, e.g. by starting a Rails console and verifying that the number of attributes is correct without explicitly calling `Profile.refresh_attribute!`:
    ```ruby
    Profile.attributes.count == ProfileField.count.succ
    ```
## Added tests?

- [X] No, and this is why: plenty of existing tests that should still pass.

## Added to documentation?

- [X] No documentation needed

PR #11881, opened 2020-12-14
Move back from after_initialize to explicit call

citizen428:
## What type of PR is this?

- [X] Refactor

## Description

Context:

1. We somewhat recently [merged a PR](https://github.com/forem/forem/pull/10707) that starts ignoring profile columns on  `User` and delegates them to `Profile` instead. This is supposed to be a temporary workaround so we can take our time to with updating existing app code to not rely on certain profile fields directly. This PR was open for a long time, because it was thoroughly tested by 
2. As part of this PR, @rhymes ]suggested an `after_initialize` hook](https://github.com/forem/forem/pull/10707#discussion_r535016233).
3. I [had my reservations](https://github.com/forem/forem/pull/10707#discussion_r535042360), but then decided to [implement the suggestion anyway](https://github.com/forem/forem/pull/11783).
4. Random bug reports about missing methods started showing up in various places which I found confusing due to the extensive testing mentioned in 1.
5. 

What I believe is happening:

1. The previous solution with an explicit `refresh_attributes!` in `Profile`'s class body made sure all accessors defined via `store_accessor` are available as soon as the class is loaded.
2. I suspect using `after_initialize` for `Profile.refresh_attributes` is too late, since at this point we already loaded `User` which set up delegation to methods that didn't exist yet.
3. I'm not entirely convinced by this explanation, as Ruby's late binding nature usually means things like this work. Here's a small example:
    ```ruby
    class TestProfile; end

   class TestUser
      attr_reader :profile

      def initialize
        @profile = TestProfile.new
      end
     
      delegate :a_method, to: :profile, allow_nil: true
    end

    tu = TestUser.new
    tu.a_method
    #=> NoMethodError: undefined method `a_method' for #<Profile:0x000056071b4fa4c0>

    class TestProfile; def a_method; 42; end ; end
    tu.a_method
    #=> 42
    ```
4. That said, there may be other more subtle effects in play here (classes being cached by Rails, user objects being cached by our application code) that I'm not taking into consideration and I want to eliminate one unknown: the original PR was tested with the explicit `refresh_attributes!` call in `Profile` so I'd prefer moving back to this even if it doesn't solve all our issues as it's also semantically closer to what we want. This isn't a "do this after the app is loaded" scenario, but a "do this while the `Profile` class is loading" one instead.
5. @rhymes' suggestion of using `after_initialize` was absolutely sensible, it was my original comment/explanation that was misleading.

#6994, #11855, https://github.com/forem/InternalProjectPlanning/issues/379

## Added tests?

- [X] No, and this is why:  existing ones should cover this and still be green.

## Added to documentation?

- [X] No documentation needed

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -25,8 +25,6 @@ class Profile < ApplicationRecord
   }.with_indifferent_access.freeze
 
   # Generates typed accessors for all currently defined profile fields.
-  # This is also used in config/application.rb to initialize profiles on app
-  # boot in the after_initialize hook.
   def self.refresh_attributes!
     return if ENV["ENV_AVAILABLE"] == "false"
     return unless Database.table_exists?("profiles")
@@ -35,6 +33,10 @@ class Profile < ApplicationRecord
       store_attribute :data, field.attribute_name.to_sym, field.type
     end
   end
+
+  # Set up all profile attributes when this class loads so all store_attribute
+  # accessors get defined immediately.
+  refresh_attributes!
 
   # Returns an array of all currently defined `store_attribute`s on `data`.
   def self.attributes
--- a/config/application.rb
+++ b/config/application.rb
@@ -85,9 +85,6 @@ module PracticalDeveloper
         top_routes << route
       end
       ReservedWords.all = [ReservedWords::BASE_WORDS + top_routes].flatten.compact.uniq
-
-      # Set up all profile attributes when the app starts
-      Profile.refresh_attributes!
     end
   end
 end


rhymes (approved):
Lgtm

ludwiczakpawel:
Not sure if this is caused by that PR or something else in the past, but my profile looks like this now:

image

^ links were "translated" into classic custom fields on profile.

citizen428:
@ludwiczakpawel I'll look into this, maybe that your local profile is messed up (seems like a lot of test data) or this is not up to date with Jacob's latest changes in `master`.

fdoxyz (approved):
> I suspect using after_initialize for Profile.refresh_attributes is too late, since at this point we already loaded User which set up delegation to methods that didn't exist yet.

> That said, there may be other more subtle effects in play here (classes being cached by Rails, user objects being cached by our application code) ...... This isn't a "do this after the app is loaded" scenario, but a "do this while the Profile class is loading" one instead.

> @rhymes' suggestion of using after_initialize was absolutely sensible, it was my original comment/explanation that was misleading.

I agree with all of the above, there must be some subtle caching/preloading problem with the `after_initialize` hook and that's just "too late" to call `refresh_attributes!`.

Tested this locally and fixed the issues we've been seeing lately 👍

maestromac (approved):

PR #11873, opened 2020-12-12
Switch node-sass to dart-sass

rhymes:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

We explicitly use `node-sass` to compile Storybook, but [node-sass/lib-sass has been deprecated](https://sass-lang.com/blog/libsass-is-deprecated) and has been replaced by [dart-sass](https://github.com/sass/dart-sass).

This PR makes this change and silences some CI logs for StoryBook.

## Related Tickets & Documents

Webpacker is still dependent on node-sass, but it'll soon change when they release version 6:

- https://github.com/rails/webpacker/issues/2673
- https://github.com/rails/webpacker/issues/2673

## QA Instructions, Screenshots, Recordings

Run `yarn storybook-prerequisites` and `yarn build-storybook` locally or `yarn storybook` and see it all works

To check that some logs are suppressed run `CI=true yarn storybook-prerequisites`

## Added tests?

- [ ] Yes
- [x] No, and this is why: just reorganizing CLI commands
- [ ] I need help with writing tests

PR #11895, opened 2020-12-14
Update bust_comment to BustComment

atsmith813:
- [x] Refactor

## Description
This PR updates references for busting comment cache from `CacheBuster` to a new, dedicated service `EdgeCache::BustComment`.

My gameplan is to PR for each method so we can test each piece of cache-busting manually in production after we deploy it. Once all method references are updated, I'll remove the legacy `CacheBuster`. We don't want to remove it preemptively either because there could always be jobs in the queue referencing it. So removing it will be its own PR at the very end when code is no longer referencing it.

Can't really QA this outside of production.

**Yes** - once this PR deploys, we need to test cache busting. We can do this by posting a comment in production and then updating said comment. If the comment updates and displays correctly as expected without having to hard refresh, that means it should be working correctly.

## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

the_office_shocked_gif

### DIFF
--- a/app/services/users/delete_articles.rb
+++ b/app/services/users/delete_articles.rb
@@ -11,7 +11,7 @@ module Users
         article.buffer_updates.delete_all
         article.comments.includes(:user).find_each do |comment|
           comment.reactions.delete_all
-          cache_buster.bust_comment(comment.commentable)
+          EdgeCache::BustComment.call(comment.commentable)
           cache_buster.bust_user(comment.user)
           comment.remove_from_elasticsearch
           comment.delete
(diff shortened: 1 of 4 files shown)


vaidehijoshi (approved):
nice! 🔥

vaidehijoshi:
i think the broken netlify deploys [might be fixed by this PR](https://github.com/forem/forem/pull/11896)?

citizen428 (approved):

rhymes:
@vaidehijoshi good call. It should have been solved by #11873 which is now on master

rhymes (approved):
LGTM

PR #11917, opened 2020-12-15
Fix html variant image prefix logic

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The current naive prefix logic for images here was only accounting for the cloudinary version. This modifies the logic such that it will also work for the imgproxy way.

Currently if an admin edits this file, they'll keep adding prefixes to infinity. This should aolve the issue.

This is a simple hotfix, but added regression test to ensure things are imrpved for the cases we care about.

## Added tests?

- [x] Yes
      have not been included_

### DIFF
--- a/app/models/html_variant.rb
+++ b/app/models/html_variant.rb
@@ -75,6 +75,6 @@ class HtmlVariant < ApplicationRecord
   end
 
   def allowed_image_host?(src)
-    src.start_with?("https://res.cloudinary.com/") || src.start_with?("https://#{SiteConfig.app_domain}/")
+    src.start_with?("https://res.cloudinary.com/") || src.start_with?(URL.url)
   end
 end
--- a/app/models/html_variant.rb
+++ b/app/models/html_variant.rb
@@ -75,6 +75,6 @@ class HtmlVariant < ApplicationRecord
   end
 
   def allowed_image_host?(src)
-    src.start_with?("https://res.cloudinary.com/") || src.start_with?(URL.url)
+    src.start_with?("https://res.cloudinary.com/") || src.start_with?(Images::Optimizer.get_imgproxy_endpoint)
   end
 end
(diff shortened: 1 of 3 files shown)


app/models/html_variant.rb
```diff
@@ -75,6 +75,6 @@ def prefix_all_images
   end
 
   def allowed_image_host?(src)
-    src.start_with?("https://res.cloudinary.com/")
+    src.start_with?("https://res.cloudinary.com/") || src.start_with?("https://#{SiteConfig.app_domain}/")
```
maestromac, line 78:
```suggestion
    src.start_with?("https://res.cloudinary.com/") || src.start_with?(URL.url)
```
Not a huge difference but using our URL helper is a bit cleaner

spec/models/html_variant_spec.rb
```diff
@@ -55,4 +55,16 @@
     cloudinary_string = "/c_limit%2Cf_auto%2Cfl_progressive%2Cq_auto%2Cw_420/https://devimages.com/image.jpg"
     expect(html_variant.html).to include(cloudinary_string)
   end
+
+  it "does not add prefix if cloudinary already start" do
```
juliannatetreault, line 59: _Non-blocking, clarifying, nitpick_: 
```suggestion
  it "does not add prefix if it already starts with cloudinary" do
```

juliannatetreault (commented):
What is the best way to test this out locally?

maestromac (approved):
LGTM!

benhalpern:
@juliannatetreault if you visit `http://localhost:3000/admin/html_variants/new` and create a variant with html that includes an image tag and save it, the proper outcome will that it will not grow in size every time you save it.

... Since this is mostly just adding one scenario to an existing solution, the manual testing isn't so much important vs sanity check the code in this case.

juliannatetreault (approved):
Thanks for the clarifying comment. Everything looks good to me!

PR #11929, opened 2020-12-16
Removed slack notification for admins when user is deleted

lightalloy:
- [x] Refactor

## Description
Removed code related to slack notifications for admins when users are deleted.
Gdpr delete requests were added in #11039 which can replace slack notifications.
Other `forems` may not use slack in the future so we shouldn't rely on it, these particular notifications were not actively used on DEV as well.

#11039
https://github.com/forem/forem/projects/31#card-45606972


rhymes (approved):
Makes sense!

I wonder if Slack configuration is or isn't part of the onboarding experience for new admins (cc @vaidehijoshi @pkfrank) or if anyone is collecting these messages

On the technical side, right now, the Slack client doesn't do anything if a Slack isn't configured. All this code results in no operations done if `ApplicationConfig["SLACK_WEBHOOK_URL"]` is missing

lightalloy:
@rhymes As far as I know, no one is using these particular messages, but I don't know about other slack notifications and onboarding.

jacobherrington (approved):
Makes sense to me! :+1:

PR #11934, opened 2020-12-16
Add Sign in with Apple

fdoxyz:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

**Feature:** Add Sign in with Apple Authentication Provider

Due to changes in the flow I'm trying out a test in a branch rather than a fork PR. The work here started in #7952

Here is a video of the feature working locally (requires HTTPS and Apple Developer Program config): https://share.getcloudapp.com/12uK0kpn

#7952

closes https://github.com/forem/security/issues/337

**Existing OAuth providers**
Since this PR also refactors the OAuth provider buttons to use POST instead of GET (https://github.com/forem/security/issues/337) try to log in and log out with other OAuth service providers

**Apple Auth behind Feature Flag**
1. Boot up locally
2. Go to sign in/sign up path (Apple Auth should not appear)
3. Log in as admin and go to `/admin/config` > Authentication
4. Apple Auth provider **should not appear**
5. Open the rails console and run `Flipper.enable(:apple_auth)`
6. Log in as admin and go to `/admin/config` > Authentication
7. This time Apple Auth provider **should appear**
8. Fill up all four details from Apple Auth to enable it 
   - You can pass in jibberish like "qsdfsdfas" it doesn't matter because you won't be able to test it locally
10. Go to sign in/sign up path and this time Apple Auth **should appear** (it won't work)
11. Go to the rails console and run `Flipper.disable(:apple_auth)`
12. Both the Apple provider from `/admin/config` and the provider option from sign in/sign up paths **should not appear anymore**
13. Yay feature flag! Now lets test in production 😈

Nope, everything was refactored in place. Maybe double check `/enter` to make sure those buttons still work the same?

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [ ] No documentation needed

None, we need to use our Canary Forem instances to fully test out the feature because it's inherently tied Apple Services that require HTTPS and other necessary configurations

the time has come


app/views/admin/configs/_apple_auth_provider_settings.html.erb
```diff
@@ -0,0 +1,77 @@
+<!--
+Apple's config is different compared to the common OAuth provider implemented
+before, this makes the generalized `auth_provider_settings` not work well so it
+requires a custom partial
+-->
```
fdoxyz, line 5: The gist of why we need this partial is because all OAuth providers use `_key` and `_secret` for easier rendering of the original `_auth_provider_settings.html.erb` partial.

Apple uses different keys (`apple_team_id `, `apple_pem`, `apple_key_id ` & `apple_client_id `) therefore needs a different partial

rhymes:
Makes sense! @fdoxyz let me know if I should close the other one or else :)

fdoxyz:
Thanks @rhymes! I do think it's best to wrap this one up in this new PR, just a little less convoluted from the conversation from a few months of working on and off on this feature.

As soon as tests pass I'll just update the PR description with more details and will open for review 😃

app/views/shared/authentication/_providers_registration_form.html.erb
```diff
@@ -1,12 +1,12 @@
...
-      <%= inline_svg_tag("#{provider.provider_name}.svg", aria: true, class: "crayons-icon", title: provider.provider_name) %>
-      <%= params[:state] == "new-user" ? "Sign up" : "Continue" %> with <%= provider.official_name %>
-    </a>
+    <% next if provider.provider_name == :apple && !Flipper.enabled?(:apple_auth) %>
+    <%= form_with url: provider.sign_in_path(state: "navbar_basic"), class: "flex w-100", local: true do |f| %>
+      <%= f.button type: :submit, class: "crayons-btn crayons-btn--l crayons-btn--brand-#{provider.provider_name} crayons-btn--icon-left grow-1 whitespace-nowrap" do %>
+        <%= inline_svg_tag("#{provider.provider_name}.svg", aria: true, class: "crayons-icon", title: provider.provider_name) %>
+        <%= params[:state] == "new-user" ? "Sign up" : "Continue" %> with <%= provider.official_name %>
+      <% end %>
+    <% end %>
```
fdoxyz, line 9: This refactor is necessary to fix https://github.com/forem/security/issues/337. In short we want the OAuth calls that initiate the flow (`/users/auth/:provider`) to be done via `POST` and not `GET` (protect from CSRF). Because of this the PR changes from anchor tags to buttons in Forms.

I thought it was going to help us avoid having to bypass `protect_from_forgery` in `omniauth_callback_controllers` so I started implementing it directly in this PR. Turns out it doesn't help, we still need it but since the fix was in I just went with it and here we are :D

https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -1,6 +1,15 @@
...
   include Devise::Controllers::Rememberable
 
+  # Rails actionpack only allows POST requests that come with an ORIGIN header
+  # that matches `request.base_url`, it raises CSRF exception otherwise.
+  # There is no way to allow specific ORIGIN values in order to securely bypass
+  # trusted origins (i.e. Apple OAuth) so `protect_from_forgery` is skipped
+  # ONLY when it's safe to do so (i.e. ORIGIN == 'https://appleid.apple.com').
+  # The hardcoded CSRF check can be found in the method `valid_request_origin?`:
+  # https://github.com/rails/rails/blob/master/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L453-L463
+  protect_from_forgery unless: -> { safe_apple_callback_request? }
```
fdoxyz, line 11: I'm okay with having this auth callback bypass `protect_from_forgery ` because SIWA uses a signed JWT to pass in the Auth data to the callback:

> the JWT travels through the HTTPS request so I think I'm comfortable with that

[Extracted from conversation about this here](https://dev.to/fdoxyz/comment/14jdl)

I'm also happy about the approach taken to bypass this, because we're only doing so **if it's coming from Apple's trusted path**. An analogy I've used for this before is:

> Instead of leaving this door unlocked for anyone we are only giving 1 trusted path (`https://appleid.apple.com`) the keys to pass through the door

fdoxyz:
Okay we're ready for review again here. Please check out the `QA Instructions` from the PR description, the idea is to have this feature hidden behind a feature flag and test the actual registration/login flow in a Canary instance.

I tried to leave comments throughout the code to explain to give more context (the discussion in #7952 is quite lengthy) but let me know if you have any questions/concerns!

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -1,6 +1,15 @@
 class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   include Devise::Controllers::Rememberable
 
+  # Rails actionpack only allows POST requests that come with an ORIGIN header
+  # that matches `request.base_url`, it raises CSRF exception otherwise.
+  # There is no way to allow specific ORIGIN values in order to securely bypass
+  # trusted origins (i.e. Apple OAuth) so `protect_from_forgery` is skipped
+  # ONLY when it's safe to do so (i.e. ORIGIN == 'https://appleid.apple.com').
+  # The hardcoded CSRF check can be found in the method `valid_request_origin?`:
+  # https://github.com/rails/rails/blob/master/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L453-L463
```
rhymes, line 10: only one non blocking note: can you replace the generic link with the link for Rails 6? Lines are moved around in different releases https://github.com/rails/rails/blob/901f12212c488f6edfcf6f8ad3230bce6b3d5792/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L449-L459

app/models/site_config.rb
```diff
@@ -209,4 +213,15 @@ def self.local?
...
+  # Apple uses different keys than the usual `PROVIDER_NAME_key` or
+  # `PROVIDER_NAME_secret` so these will help the generalized authentication
+  # code to work, i.e. https://github.com/forem/forem/blob/master/app/helpers/authentication_helper.rb#L26-L29
+  def self.apple_key
+    return unless apple_client_id.present? && apple_key_id.present? &&
+      apple_pem.present? && apple_team_id.present?
+
+    "present"
+  end
+  singleton_class.__send__(:alias_method, :apple_secret, :apple_key)
```
rhymes, line 226: damn you Ruby 🤣

app/views/admin/configs/_apple_auth_provider_settings.html.erb
rhymes: damn you Apple 🤣

rhymes (approved):
Let's do this! 7 months after we started working on it :D

app/controllers/omniauth_callbacks_controller.rb
fdoxyz: So tricky to keep those up to date, but absolutely I'll add this update in a sec 💯

joshpuetz:
I appreciate the clear PR baiting that is happening with a RuPaul gif here :)

fdoxyz:
The time has come to ~lip sync~ authenticate with Apple for ~your life~ approval in the App Store 😆

app/services/authentication/providers/apple.rb
```diff
@@ -0,0 +1,88 @@
...
+
+        user_data = {
+          email: info.email,
+          apple_created_at: Time.zone.at(timestamp),
+          apple_username: user_nickname,
+          name: name
+        }
+
+        if Rails.env.test?
+          user_data[:profile_image] = SiteConfig.mascot_image_url
```
joshpuetz, line 23: **Question**: Is there a way to do this without specially adding test code to production?

joshpuetz (approved):

app/services/authentication/providers/apple.rb
fdoxyz: Great question! The reason why this was added is because directly setting `remote_profile_image_url` triggered CarrierWave to fetch the image, which caused the tests to fail for me (the original PR had something similar in place).

I'm not sure if there's a way we can just leave the else clause, i.e. `user_data[:remote_profile_image_url] = Users::ProfileImageGenerator.call`

If anyone can share an idea on how to solve this, like a way to mock CarrierWave somehow (not as trivial as it sounds at least for me) I'd be happy to include it!

This is all behind a feature flag, so we can brainstorm and improve on this when making it generally available (getting rid of the flag) too 👍

joshpuetz: Cool: I think this definitely works for now and I'd hate to hold up the PR because of it!

PR #11945, opened 2020-12-17
Fix Rails/OutputSafety violations

lightalloy:
- [x] Refactor

## Description
Replaced `html_safe` and `raw` with `tag` + `safe_join`

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -283,10 +283,10 @@ module ApplicationHelper
   end
 
   def admin_config_label(method, content = nil)
-    content ||= tag.span method.to_s.humanize
+    content ||= tag.span(method.to_s.humanize)
 
     if method.to_sym.in?(VerifySetupCompleted::MANDATORY_CONFIGS)
-      required = tag.span "Required", class: "crayons-indicator crayons-indicator--critical"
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -282,16 +282,16 @@ def display_estimated_user_count?
     estimated_user_count > LARGE_USERBASE_THRESHOLD
   end
 
-  # rubocop:disable Rails/OutputSafety
   def admin_config_label(method, content = nil)
-    content ||= raw("<span>#{method.to_s.humanize}</span>")
+    content ||= tag.span method.to_s.humanize
```
rhymes, line 286: nitpick non blocking: I noticed there's inconsistency with parenthesis, `tag.span` it's called without it and `tag.a` up in the first file is called with it, and `tag.label` down here is called with them. I vote for parenthesis :D

rhymes (approved):
Yeah!

Ridhwana (approved):
Nice one @lightalloy!

citizen428 (approved):

rhymes:
@lightalloy you can merge master now

app/helpers/application_helper.rb
lightalloy: Nice catch :)

Issue #8940, opened 2020-06-26
Add /feed/latest to RSS feeds


todsac:
**Is your feature request related to a problem? Please describe.**

The only way to ensure you’re getting the latest articles right now is to increment the API

**Describe the solution you'd like**

https://dev.to/feed/latest

**Describe alternatives you've considered**

None

**Additional context**

See email thread from [https://pipedream.com](url) team

rhymes:
Thanks for the issue! We'll take your request into consideration and follow up if we decide to tackle this issue.

To our amazing contributors: issues labeled `type: bug` are always up for grabs, but for feature requests, please wait until we add a `ready for dev` before starting to work on it.

To claim an issue to work on, please leave a comment. If you've claimed the issue and need help, please ping @thepracticaldev/oss and we will follow up within 3 business days.

For full info on how to contribute, please check out our [contributors guide](https://github.com/thepracticaldev/dev.to/blob/master/CONTRIBUTING.md).

ccozkan:
Hey :wave: Is this issue still available? I would like to work on this

citizen428:
Teşekkür ederim for volunteering @ccozkan, it's all yours! 😃

PR #11956, opened 2020-12-17
Add ability to get latest articles on RSS feed

ccozkan:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR adds oppurtinity to receive latest published articles on RSS feed. Query string is used in order to retrieve articles, instead of /feed/latest since it would conflict with username "latest".

Resolves [#8940](https://github.com/forem/forem/issues/8940).

## Added tests?

- [X] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [X] No documentation needed


spec/requests/articles/articles_spec.rb
```diff
@@ -78,6 +78,18 @@
...
 
+    context "when query string is q=latest" do
+      let!(:featured_article) { create(:article, featured: true) }
+      let!(:not_featured_article) { create(:article, featured: false) }
+
+      before { get "/feed?q=latest" }
+
+      it "returns last published articles" do
+        expect(response.body).to include(featured_article.title)
+        expect(response.body).to include(not_featured_article.title)
```
juliannatetreault, line 89: Thank you so much for adding this test! I have a couple of small suggestions for this test, one of which should make this a bit more intuitive for future contributors and another regarding one of the test's expectations.

Right now, line `89` checks that `response.body` includes `not_featured_article.title`, but I think what we actually want to check here is that `response.body` does not include `not_featured_article.title`, as shown below:
```suggestion
        expect(response.body).not_to include(not_featured_article.title)
```

Additionally, I think we could change the names here to make the test a bit more readable for future contributors by renamining `featured_article` and `not_featured_article` to something along the lines of:
```suggestion
        expect(response.body).to include(last_published_article.title)
        expect(response.body).to include(not_last_published_article.title)
```

Let me know if you have any questions about this! All-in-all, awesome job, @ccozkan . 😄

rhymes (commented):
Hi @ccozkan, thanks for the concern, `latest` is not an available username, as it would conflict with https://dev.to/latest which is one of the most important home pages :) 

It's part of the reserved words here https://github.com/forem/forem/blob/master/config/initializers/reserved_words.rb so you can go ahead and switch this back to `/feed/latest` with confidence, right now it points to a 404 page https://dev.to/feed/latest

You could add it to the `routes.rb` file to make it explicit, so the controller doesn't have to wonder if it's a user or an organization

spec/requests/articles/articles_spec.rb
ccozkan: Thanks for the feedback, I will update my PR accordingly :smile: I have a question though :thinking: , why we don't want 'not featured articles' to be a part of the feed?

rhymes (commented):
@ccozkan thanks for the change in URL, if I'm not mistaken though the new URL is exactly like `/feed` that we already have. Shouldn't `latest` return `dev.to/latest` ?

ccozkan:
@rhymes , thank you for the feedback! I see, sorry I misunderstood the issue. So I should call `Articles::Feed::LargeForemExperimental.new.latest_feed` on articles controller, and show those articles on `/feed/latest`

rhymes:
@ccozkan as I understood #8940 right not it's not possible to get a feed with the latest posts. The reverse chronological sequence of published articles is available through https://dev.to/latest for human consumption which calls `Articles::Feeds::LargeForemExperimental#latest_feed` 

I think we want to do something similar for the machine feed.

app/controllers/articles_controller.rb
```diff
@@ -219,6 +221,14 @@ def handle_user_or_organization_feed
     end
   end
 
+  def handle_latest_feed
+    page = params[:page].try(:to_i) || 1
+    ids_of_articles_on_latest = Articles::Feeds::LargeForemExperimental.new(page: page, number_of_articles: 12)
+      .latest_feed.includes([:user])
+      .includes(:taggings).map(&:id)
+    Article.find(ids_of_articles_on_latest)
```
rhymes, line 229: `Articles::Feeds::LargeForemExperimental.new.latest_feed` already returns article objects, so you don't need to re-fetch them with `Article.find`. On line 38 the action is rendering `@articles` so this should assign those articles to `@articles`

Let me know if anything is unclear!

rhymes (commented):
Thanks for the change! Left a note about the logic :)

app/controllers/articles_controller.rb
ccozkan: Hmm yes, I had to this trick since `ActiveModel::MissingAttributeError missing attribute: processed_html` error occurs on `feed.rss.builder` when `@articles` is the returned objects of `Articles::Feeds::LargeForemExperimental.new.latest_feed`

Here is that error line :eyes: 
```
#  app/view/articles/feed.rss.builder line 27
xml.description sanitize(article.processed_html, tags: allowed_tags, attributes: allowed_attributes)
```

rhymes: Ah got it! That's because the RSS feed needs a different set of fields than the `/latest` "human" page needs.

I wonder if we should simply separate the functionality instead of reusing `LargeForemExperimental`.

I looked at the code again, isn't what we need to return simply

https://github.com/forem/forem/blob/90aa6932040f584467c971456b56500e4d333c72/app/controllers/articles_controller.rb#L24

with the addition of `.where("score > ?", MINIMUM_SCORE_LATEST_FEED)` mutuated from `LargeForemExperimental` ?

In plain english: return all the published articles, with the attributes selected by the feed (`Article.feed`), in reverse chronological order of publication, filtering out those that have a score that's too low (so `MINIMUM_SCORE_LATEST_FEED`)

Does that make sense? 

I would add a test to make sure that's the case.

That's probably easier than recycling `LargeForemExperimental`, sorry for the confusion

ccozkan: Yes, sure! I will update my PR accordingly :rocket:

app/controllers/articles_controller.rb
```diff
@@ -26,6 +26,10 @@ def feed
                   handle_user_or_organization_feed
                 elsif params[:tag]
                   handle_tag_feed
+                elsif request.path == latest_feed_path
+                  @articles.where(featured: true)
```
rhymes, line 30: Sorry for not being clearer before but `where(featured: true)` is the condition we're trying to remove, because we want the latest articles, not the feature ones. As I mentioned in the other discussion, we need to replicate the same query we use for `dev.to/latest`, so 

https://github.com/forem/forem/blob/90aa6932040f584467c971456b56500e4d333c72/app/controllers/articles_controller.rb#L24

plus the score filter. That's it. Featured is used in the `else` clause, which is the one of the general feed, not the "latest" feed.

Currently we have these feeds:

- "username feed" with the list of all published articles created by the user (`Article.feed` + `handle_user_or_organization_feed`)
- "organization feed" with the list of all published articles created by the organization (`Article.feed` + `handle_user_or_organization_feed`)
- "tag feed" with the list of all published articles with the given tag (`Article.feed` + `handle_tag_feed`)
- "generic feed" (the `else` clause) with the list of published articles that are featured (`Article.feed` + `where(featured: true)`)

What we want to add is "the list of all published articles above a certain score", so (`Article.feed` + the score filter).

We do not want to filter by `featured` here, otherwise it'd already be the current "generic feed"

Let me know if anything is unclear

rhymes (changes requested):

app/controllers/articles_controller.rb
ccozkan: Oh okay, thanks a lot for the detailed explanation! :smiley: But if I am understanding right, at https://github.com/forem/forem/pull/11956#discussion_r545432001 @juliannatetreault suggested me otherwise on the presence of 'not featured articles' on `feed/latest`

rhymes: Consider that right now the only difference between the two paths is the presence or absence of articles with a lower score, which is not what https://dev.to/latest is

In https://github.com/forem/forem/issues/8940 @todsac mentioned that:

> The only way to ensure you’re getting the latest articles right now is to increment the API

that's because article ids are incremental, so to get the latest you have to come up with a trick to fetch the next article by ID

I might have misunderstood the initial goal, and I think we're in this empasse because the issue itself isn't 100% clear and devoid of a discussion but I don't see the added value to the OP to add a new route which essentially returns the same data.

cc @juliannatetreault for her feedback

Thank you for the patience!

spec/requests/articles/articles_spec.rb
juliannatetreault: Hi, @ccozkan and my sincerest apologies for the delayed response -- I was out of the office! 😊 

Now that I've read through yours and @rhymes's comments, it seems I may have _also_ misunderstood the initial goal, and for that, I am so sorry! The goal of my feedback here was to make these tests a bit more intuitive: since we are adding the ability to access only the "latest" articles and not necessarily the featured articles, I think that it makes more sense to make the names in the spec more in-line with what we're trying to accomplish here. That being said, my thought was that `last_published_article` or even `latest_article` better described what we're looking for in the "GET /feed/latest" test and that `featured_article` wasn't a very descriptive name for what we are trying to test.

Regarding [this](https://github.com/forem/forem/pull/11956#discussion_r549022460) comment: I misunderstood the spec's intention and did not mean to pull you in the wrong direction!

ccozkan: Ok I see now, no problem at all! It's been a good experience for me :relaxed:

juliannatetreault: I'm so happy to hear that! 😊

rhymes (approved):
LGTM! Thanks @ccozkan

juliannatetreault (approved):
This works as expected! Thanks so much for tackling this @ccozkan! 🚀

PR #11969, opened 2020-12-18
Fix the way we check for AndroidBridge global

fdoxyz:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Turns out that when using the `ForemWebView` user agent the `AndroidBridge != undefined` check is raising an exception. It didn't appear before because the user agent check now is shared by both Android and iOS, causing the exception to be raised (because `AndroidBridge` is not defined in the iOS ForemWebView context).

This PR changes the check for this global so it works well on all platforms

Open your local environment and try to click on the ellipses (three dots) on any Article. It should open the modal/pop-up with the options to share the article (meaning that it doesn't break).

If you try out in Dev Tools console run these two lines:

```js
// Will likely raise an exception
AndroidBridge != undefined;

// Should return `false` unless AndroidBridge global is available
typeof AndroidBridge === 'function';
```

If you have an iPhone/iPad running on iOS:
1. Install the DEV app from the App Store
2. Open any Article and try to click on the ellipses (three dots) option to share
3. It fails for me. This PR fixes the problem

None. This will fix a UI feature in iOS

## Added tests?

- [ ] Yes
- [x] No, and this is why: This only happens in clients using `ForemWebView` user agent. There may be a way to test this using Cypress but it feels a little too early for implementing those e2e tests right now. I'm open for ideas to approach testing this and other mobile situations

## Added to documentation?

- [ ] README
- [x] No documentation needed

uuummmm

### DIFF
--- a/app/assets/javascripts/initializers/runtime.js
+++ b/app/assets/javascripts/initializers/runtime.js
@@ -41,7 +41,7 @@ class Runtime {
   static isNativeAndroid(namespace = null) {
     const nativeCheck =
       /DEV-Native-android|ForemWebView/i.test(navigator.userAgent) &&
-      typeof AndroidBridge === 'function';
+      typeof AndroidBridge !== 'undefined';
 
     let namespaceCheck = true;
     if (nativeCheck && namespace) {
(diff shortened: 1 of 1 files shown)


maestromac (approved):

nickytonline (commented):
Good catch! Is `AndroidBridge` a function though or is it an object? If it's the latter, you could change this check to `typeof AndroidBridge !== 'undefined'`

Ridhwana (approved):

fdoxyz:
Thanks everyone for the reviews!

@nickytonline great suggestion, it is indeed a `function` but it's a bit clearer to read and "fool proof" to check for undefined instead 💯

rhymes (approved):

Issue #2147, opened 2019-03-21
Static assets have unusually short caching period


rhymes:
**Describe the bug**

I was testing the homepage with [Google PageSpeed Insights](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fdev.to%2F) and [WebPage Performance Test](https://webpagetest.org/performance_optimization.php?test=190321_8H_be96eb19fc13a0947df3488adf259c64&run=1) (both links lead to the reports) and they both signalled that static assets (JavaScript files, images and CSS) have short caches, ranging from 1 day to 7 days.

Since they are all versioned I was wondering if there was a reason for that otherwise I'd like to suggest to set a longer cache for those, following [Google's best practice](https://developers.google.com/web/tools/lighthouse/audits/cache-policy) and set the cache to one year.

**Expected behavior**

A longer cache period for static assets.

**Screenshots**

Google PageSpeed Insights (same as Lighthouse):

![Screenshot_2019-03-21 PageSpeed Insights](https://user-images.githubusercontent.com/146201/54743950-ec809100-4bc5-11e9-8358-94ab5da90de8.png)

WebPage Performance Test:

![Screenshot_2019-03-21 WebPageTest Optimization Check Results - Dulles dev to - 03 21 19 10 29 24](https://user-images.githubusercontent.com/146201/54743969-fbffda00-4bc5-11e9-9025-8f36aa372314.png)

**Desktop (please complete the following information):**

- OS: macOS
- Browser: Firefox
- Version: 67b3

aspittel:
Hey @rhymes, thanks for checking in on this -- @benhalpern may have some better insights on the cache length!

rhymes:
unstale

mstruve:
@rhymes is this still something that would be worth doing?

rhymes:
Yep, although it's better than March 2019, it's still unclear why we keep short caching periods on fingerprinted static assets, see https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fdev.to%2F ("Diagnostics -> Serve static assets with an efficient cache policy"). It's 30 days, it could be years.

This doesn't happen on forem.dev as it uses nginx for caching IIRC.

Notice the difference (on a fingerprinted CSS file):

**forem.dev** and **the community club** and I guess the others as well:

[...]

PR #11970, opened 2020-12-18
Lengthen local caching max-age

benhalpern:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

From @rhymes comment

> Yep, although it's better than March 2019, it's still unclear why we keep short caching periods on fingerprinted static assets, see https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fdev.to%2F ("Diagnostics -> Serve static assets with an efficient cache policy"). It's 30 days, it could be years.

This changes the `max-age` on instructions for local caching from 30 days to 3000 days, which seems long enough. 😅

I believe that having a more reasonably short `s-max-age` makes sense because this is a directive for the CDN, where we don't need to hold something for so long, but `max-age` will act as the proper directive for browsers to keep stuff around for a long time.

Closes https://github.com/forem/forem/issues/2147

Double check my logic on how this should all work.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -37,7 +37,7 @@ Rails.application.configure do
   # Apache or NGINX already handles this.
   config.public_file_server.enabled = ENV["RAILS_SERVE_STATIC_FILES"].present?
   config.public_file_server.headers = {
-    "Cache-Control" => "public, s-maxage=#{30.days.to_i}, max-age=#{30.days.to_i}"
+    "Cache-Control" => "public, s-maxage=#{30.days.to_i}, max-age=#{3000.days.to_i}"
   }
 
   # Compress JavaScripts and CSS.


config/environments/production.rb
```diff
@@ -37,7 +37,7 @@
   # Apache or NGINX already handles this.
   config.public_file_server.enabled = ENV["RAILS_SERVE_STATIC_FILES"].present?
   config.public_file_server.headers = {
-    "Cache-Control" => "public, s-maxage=#{30.days.to_i}, max-age=#{30.days.to_i}"
+    "Cache-Control" => "public, s-maxage=#{30.days.to_i}, max-age=#{3000.days.to_i}"
```
rhymes, line 40: I think it's fine! I was checking the other Forems and it's set even higher :D

rhymes (approved):

config/environments/production.rb
benhalpern: In 8.2 years we may regret setting it so shot 😅

rhymes: ahahhaah

maestromac (approved):
LGTM!

PR #11972, opened 2020-12-18
Upgrade rubocop* packages to the latest version and fix violations 

rhymes:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

**TLDR;** we were stuck on Rubocop 0.93 because `erb_lint` didn't support it which led to a host of issues tracking the packages dependents from Rubocop and missing bug fixes. 

This PR unlocks `erb_lint` until a new official version will be released (deemed it safe, I checked the code and I'm tracking the issue here https://github.com/Shopify/erb-lint/issues/193) and applies all the changes.

Don't be frigthened by the number of files changed, they're mostly the vendored `erb_lint` gem

https://github.com/Shopify/erb-lint/issues/193

Nope, nothing to QA

## Added tests?

- [ ] Yes
- [x] No, and this is why: Rubocop changes the style of the code, tests need to be passing as they are


.rubocop.yml
```diff
@@ -148,10 +148,18 @@ Lint/DeprecatedOpenSSLConstant:
   Description: "Don't use algorithm constants for `OpenSSL::Cipher` and `OpenSSL::Digest`."
   Enabled: true
 
+Lint/DuplicateBranch:
+  Description: Checks that there are no repeated bodies within `if/unless`, `case-when` and `rescue` constructs.
+  Enabled: false
```
rhymes, line 153: This is enabled automatically. I've disabled it because it's a heavy change and requires a separate PR if we want to enable it. It basically checks for if branches that do the same thing, but it can't be fixed automatically and requires code restructure

.rubocop.yml
```diff
@@ -274,6 +302,11 @@ Naming/PredicateName:
   Exclude:
     - spec/**/*
 
+Naming/VariableNumber:
+  Description: 'Use the configured style when numbering symbols, methods and variables.'
+  StyleGuide: '#snake-case-symbols-methods-vars-with-numbers'
+  Enabled: false
+
```
rhymes, line 309: This is enabled automatically and I'm disabling it. It's a bit picky. You either have to enforce all variables to be in the style of `variable1` or `variable_1` with the underscore. I don't have a preference either way and can live with both styles :D

.rubocop.yml
```diff
@@ -285,6 +318,11 @@ Style/AccessorGrouping:
   Description: 'Checks for grouping of accessors in `class` and `module` bodies.'
   Enabled: true
 
+Style/ArgumentsForwarding:
+  Description: 'Use arguments forwarding.'
+  StyleGuide: '#arguments-forwarding'
+  Enabled: false
+
```
rhymes, line 325: Ruby 2.7 `...` new forwarding syntax which I don't fully understand yet and breaks my editor syntax's checker :D Disabled in the meantime

.rubocop.yml
```diff
@@ -307,6 +345,16 @@ Style/CaseLikeIf:
   StyleGuide: '#case-vs-if-else'
   Enabled: true
 
+Style/ClassAndModuleChildren:
+  Description: 'Checks style of children classes and modules.'
+  StyleGuide: '#namespace-definition'
+  Enabled: true
+  SafeAutoCorrect: true
```
rhymes, line 352: just moved around alphabetically

.rubocop.yml
```diff
@@ -340,12 +391,21 @@ Style/ExplicitBlockArgument:
   StyleGuide: '#block-argument'
   Enabled: true
 
+Style/Encoding:
+  Description: 'Use UTF-8 as the source file encoding.'
+  Enabled: false
+
```
rhymes, line 397: just moved around alphabetically

.rubocop.yml
```diff
@@ -340,12 +391,21 @@ Style/ExplicitBlockArgument:
...
   Description: 'When using exponential notation, favor a mantissa between 1 (inclusive) and 10 (exclusive).'
   StyleGuide: '#exponential-notation'
   Enabled: true
   EnforcedStyle: scientific
 
+Style/FormatStringToken:
+  Description: 'Use a consistent style for format string tokens.'
+  Exclude:
+    - 'config/routes.rb'
+
```
rhymes, line 408: just moved around alphabetically

.rubocop.yml
```diff
@@ -682,12 +750,22 @@ RSpec/DescribeClass:
...
 RSpec/ExampleLength:
   Description: Checks for long examples.
   Enabled: true
   Max: 15
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ExampleLength
 
+RSpec/MessageExpectation:
+  Description: Checks for consistent message expectation style.
+  Enabled: true
+  EnforcedStyle: allow
```
rhymes, line 767: ❤️

Gemfile
```diff
@@ -114,7 +114,7 @@ group :development do
   gem "brakeman", "~> 4.10", require: false # Brakeman detects security vulnerabilities in Ruby on Rails applications via static analysis
   gem "bundler-audit", "~> 0.7" # bundler-audit provides patch-level verification for Bundled apps
   gem "derailed_benchmarks", "~> 1.8", require: false # A series of things you can use to benchmark a Rails or Ruby app
-  gem "erb_lint", "~> 0.0.34", require: false # ERB Linter tool
+  gem "erb_lint", github: "splitwise/erb-lint", branch: "jess/remove-version-pinning", require: false # ERB Linter tool
```
rhymes, line 117: see the PR description here

app/controllers/admin/configs_controller.rb
```diff
@@ -142,7 +142,7 @@ class ConfigsController < Admin::ApplicationController
         onboarding_taskcard_image
       ].freeze
 
-    VALID_URL = %r{\A(http|https)://([/|.|\w|\s|-])*.[a-z]{2,5}(:[0-9]{1,5})?(/.*)?\z}.freeze
+    VALID_URL = %r{\A(http|https)://([/|.\w\s-])*.[a-z]{2,5}(:[0-9]{1,5})?(/.*)?\z}.freeze
```
rhymes, line 145: automatically changed by Rubocop. `|` are redundant here

Guardfile
```diff
@@ -15,7 +15,7 @@
 #
 # and, you'll have to watch "config/Guardfile" instead of "Guardfile"
 
-# Note: The cmd option is now required due to the increasing number of ways
+# NOTE: The cmd option is now required due to the increasing number of ways
```
rhymes, line 18: upper cased automatically by Rubocop

app/controllers/chat_channels_controller.rb
```diff
@@ -181,7 +181,7 @@ def channel_info
 
   def create_channel
     chat_channel_params = params[:chat_channel]
-    chat_channel_name = chat_channel_params[:channel_name].split(" ").join("-")
+    chat_channel_name = chat_channel_params[:channel_name].split.join("-")
```
rhymes, line 184: you'll notice a bunch of these throught the PR. `split(" ")` and `.split` are the same thing. Splitting over space char is the default behavior

vendor/cache/erb-lint-05596ea189ee/.gitignore
```diff
@@ -0,0 +1,12 @@
+# Gem files
```
rhymes, line 1: from here after it's only the vendored erb-lint gem

Gemfile
maestromac: I'm somewhat concerned about pointing to a fork's branch. The past two times we needed to do something like this we would fork it ourselves to ensure that the branch won't be altered or deleted. What do you think?

app/controllers/chat_channels_controller.rb
maestromac: :heart_eyes:

Gemfile
rhymes: Makes perfect sense, i'll fork it and update it while waiting for `erb_lint` to release the new version

rhymes: I forked over `thepracticaldev/erb_lint` because I don't have permission to fork over `forem/*`

.rubocop.yml
citizen428: > which I don't fully understand yet

The best use case would be our service objects:

```ruby
# Before
def self.call(arg1, arg2, kwarg1:)
  new(arg1, arg2, kwarg1: kwargs1).call
end

# After
def self.call(...)
  new(...).call
end
```

Though I'm actually not a fan of it for this specific use-case, the whole reason I want duplication is to make the signature of the main interface more explicit.

app/controllers/admin/configs_controller.rb
citizen428: For anyone wondering, they are redundant because they are used within a character class (`[]`).

citizen428 (approved):

Zhao-Andy (approved):
LGTM!

Gemfile
```diff
@@ -114,7 +114,7 @@ group :development do
   gem "brakeman", "~> 4.10", require: false # Brakeman detects security vulnerabilities in Ruby on Rails applications via static analysis
   gem "bundler-audit", "~> 0.7" # bundler-audit provides patch-level verification for Bundled apps
   gem "derailed_benchmarks", "~> 1.8", require: false # A series of things you can use to benchmark a Rails or Ruby app
-  gem "erb_lint", "~> 0.0.34", require: false # ERB Linter tool
+  gem "erb_lint", github: "thepracticaldev/erb-lint", branch: "jess/remove-version-pinning", require: false # ERB Linter tool
```
nickytonline, line 117: Question: Is this the fix for erb lint? Also, shouldn't `thepracticaldev/erb-lint` be `forem/erb-lint`? Also, I'm not following why it's a branch of Jess'

rhymes: @nickytonline this is in there because `erb_lint` hasn't been updated. There's a working version that was submit as a PR, @maestromac correctly suggested to fork the PR to avoid including temporary code we don't have control on. It was forked over `thepracticaldev` because I don't have the rights to fork on `forem/` and this will be corrected soon, as soon as `erb_lint` releases a new version

.rubocop.yml
rhymes: Exactly, I think the old syntax is clearer. "..." can mean anything

Gemfile
rhymes: Basically a magician trick :D

Issue #382, opened 2018-08-17
Consider Adding End to End (E2E) Testing to the Deployment Pipeline


maestromac:
@nickytonline commented on [Thu Jul 19 2018](https://github.com/thepracticaldev/dev.to_core/issues/602)

# Story - Consider Adding End to End (E2E) Testing to the Deployment Pipeline

@benhalpern @maestromac @Zhao-Andy @jessleenyc et all.

Just wanted to open a discussion here about E2E testing.

There's a lot going on on the site. The backend has lots of unit tests, the new front-end stuff has unit tests and snapshot testing, but there is no E2E testing in place.

When I worked at McAfee, it was essential to have E2E testing in place because we were working on a password browser extension. We had to be certain the UI functioned all the time. In the case of a web site, it's a good idea as well, but a lot of times it's omitted due to lack of resources/time. Not good excuses, but that's the reality I've lived through at other jobs. There's also a burden of maintaining them.

Having said that, they help you spot regressions fairly quickly and it allows you to test full workflows, e.g. on-boarding. In terms or resources, I think we'll have those once the code is fully open sourced. I have a feeling a few people will want to contribute to the code base. 😉

I think it's a win win honestly. The site will be more stable because of the E2E tests and there will be so many tests to do that it will be an easy way for new contributors to get some PRs in (provided we onboard properly).

If we decide to go ahead with this, we'll need to decide on an E2E testing framework. I'm personally leaning towards [Cypress](cypress.io), but let's make this part of the discussion.

---

@benhalpern commented on [Fri Jul 20 2018](https://github.com/thepracticaldev/dev.to_core/issues/602#issuecomment-406684377)

Yeah let's do this. I haven't seen any reason to not go with Cypress. It's definitely been on my list of things to try and haven't seen anything better.

Rails has its own concept of E2E testing which we make some use of but it's been such a pain in the ass I've never had the patience to be good at this part. I'll leave this topic open for anyone to argue that we should stick into this world as opposed to Cypress.

Either way, yeah I'm in agreement with the approach to lead to stability, as long as we don't get too heavy-handed to the point where we need to change 10 files for every change. Let's try to walk the line the right way.

---

@nickytonline commented on [Sun Jul 22 2018](https://github.com/thepracticaldev/dev.to_core/issues/602#issuecomment-406902404)

Also, it currently only supports Chrome, but [FireFox support is on it's way](https://github.com/cypress-io/cypress/issues/1096).

---

[...]

PR #11977, opened 2020-12-18
Reenable End to End Testing... Again 😄

nickytonline:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I've rerun the build 4 times for this PR and it's passed every time. I also created a forked branch PR as I noticed @Ridhwana's PR that ran into e2e issues was from a forked repo. I reran the build 4 times on #11978, and everytime it passed.

This potentially appears to be a rails 6 issue when `RAILS_ENV=test`. See https://github.com/rails/rails/issues/40613.

We have the following set in config/environments/test.rb

```
  # The test environment is used exclusively to run your application's
  # test suite. You never need to work with it otherwise. Remember that
  # your test database is "scratch space" for the test suite and is wiped
  # and recreated between test runs. Don't rely on the data there!
  config.cache_classes = true
```

When `config.cache_classes = true` is set to true, apparently it will implicitly set `config.action_view.cache_template_loading` to true as well. The above mentioned Rails issue talks about that and also references [3.10 Configuring Action View](https://guides.rubyonrails.org/configuring.html#configuring-action-view) from the Rails documentation.

Having said that, the issue references a Rails 4.2 spec for testing this and I came across the [same tests in master](https://github.com/rails/rails/blob/master/railties/test/application/configuration_test.rb#L1251-L1291) for the Rails repository and the similar tests only seem to test for this when in development mode which is why I have explicitly set `config.action_view.cache_template_loading = true` for test.rb.

I don't know if this is the actual fix as I do not have enough Rails experience, but it does appear to fix the issue in the CI/CD pipeline.

- #11978
- #382
- https://github.com/rails/rails/issues/40613

N/A

N/A

## Added tests?

- [ ] Yes
- [x] No, and this is why: This is test infrastructure

## Added to documentation?

- [ ] README
- [x] No documentation needed

Johnny Rose from Shitt's Creek saying "Possibly"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,12 +69,12 @@ script:
   - if [ "$KNAPSACK_PRO_CI_NODE_INDEX" == "1" ]; then bundle exec bundle-audit check --update --ignore CVE-2015-9284; fi
   - if [ "$KNAPSACK_PRO_CI_NODE_INDEX" == "1" ]; then yarn build-storybook; fi
   - if [ "$KNAPSACK_PRO_CI_NODE_INDEX" == "2" ]; then bin/test-console-check; fi
-  # - export FOREM_OWNER_SECRET="secret"
-  # - yarn e2e:ci
-  # # Dropping and recreating the database for Capybara/rspecs
-  # - bundle exec rails db:drop
-  # - bundle exec rails db:create
-  # - bundle exec rails db:schema:load
+  - export FOREM_OWNER_SECRET="secret"
+  - yarn e2e:ci
+  # Dropping and recreating the database for Capybara/rspecs
+  - bundle exec rails db:drop
+  - bundle exec rails db:create
+  - bundle exec rails db:schema:load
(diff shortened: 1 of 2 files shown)


rhymes (approved):
Great find @nickytonline!

I went to check the default state of `test.rb` on a new Rails installation and [this is what they do](https://github.com/rails/rails/blob/901f12212c488f6edfcf6f8ad3230bce6b3d5792/railties/lib/rails/generators/rails/app/templates/config/environments/test.rb.tt#L9-L15):

```ruby
  <%-# Spring executes the reloaders when files change. -%>
  <%- if spring_install? -%>
  config.cache_classes = false
  config.action_view.cache_template_loading = true
  <%- else -%>
  config.cache_classes = true
  <%- end -%>
```

so I think this trick is the right one!

Seems like `cache_template_loading` isn't always `true` as the documentation implies yeah.

LGTM

PR #11982, opened 2020-12-19
Test variants for which posts are fetched on initial "hot articles" home feed query.

benhalpern:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This pull request makes some changes to how we log finishers in field test experiments for better re-usability, but still keeps components near one another knowing that this is inherently code that can be moved around as needed and we don't need to commit to a major architecture.

The experiment itself takes one part of the feed algorithm, the "hot posts" and does more to customize this initial grab. We currently only sort them after they are first fetched, possibly injecting too many posts that are less relevant to the end user.

Besides the refactors of the testing in general, this is a fairly tiny PR. It adds a new a/b test and limits the changes to one place. When the test is done running we will pull it out and limit the total code paths.

This PR also makes some adjustments to "base" removing the "offset" concept altogether, which is not needed based on what we picked from a prior a/b test.

Mostly sanity check that the variants do what they say they do.

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed - When the test is done running, we will write up a retro.


mstruve (commented):
I am noticing that there was not an explicit winner it seems based on the new implicit point system that was added. 
Screen Shot 2020-12-28 at 10 06 06 AM
Screen Shot 2020-12-28 at 10 06 14 AM

Given this test finished pretty inconclusively and the addition of implicit points in the app adds quite a bit of complexity, I think it would be a good idea to remove the implicit point code in addition to removing the testing infrastructure.

benhalpern:
@mstruve I believe there _was_ a winner in that test as "base" had the clear lead for a while that eventually started to erode...

I think the reason it eroded is because the success metrics didn't fit the time the test was running. The things we were measuring became a little too random _over time_ because we were declaring success based on things that were too easy to _eventually_ hit no matter what the case.

Explained more here:

https://github.com/forem/forem/pull/11974

Base was winning by 10-20% over others after about a week. That is why this new test introduces more variations on what we're measuring and eliminates the really "random" ones like _a single reaction_. I think this is a move in the right direction in terms of the feedback loop for more clean and clear results.

mstruve:
Oh shit, I follow you. Thanks for all the context. Should we include some of [this PR description](https://github.com/forem/forem/pull/11974) in the A/B test documentation? like this part maybe? That way in the future we know these tests need to be run quickly and people need to keep an eye on the results immediately.

> About a week into the test, "base" (which is the original new feature) was the clear winner across almost all the tests. It was winning by 10% to 20%.... But as the test has gone on, the results have become closer and seemed random. In inspecting the essence of what we are testing for to see success, this is just not a test that is built to run as long as this one.
Because our success metrics deem success if someone reaches some of these criteria just once results become a little too random over time. If we want to measure success over a 4 week span, we need something that is a little more likely to differentiate than these one-off successes.

app/services/articles/feeds/large_forem_experimental.rb
```diff
@@ -126,26 +126,74 @@ def score_comments(article)
...
+            .where("score >= ? OR featured = ?", SiteConfig.home_feed_minimum_score, true)
+            .order(hotness_score: :desc)
+          # We only want to limit the posts to tagged_with if the participant follows enough tags.
+          articles = articles.tagged_with(followed_tags, any: true) if followed_tags.size > 4
+        when "top_articles_since_last_pageview_3_days_max" # Top articles since last page view (max 3 days)
+          start_time = [(@user.page_views.last&.created_at || Time.current) - 12.hours, 3.days.ago].max
+          articles = Article.published.limited_column_select.includes(top_comments: :user)
+            .where("published_at > ?", start_time)
+            .page(@page).per(@number_of_articles)
+            .order(score: :desc)
```
mstruve, line 171: Why are some of these ordered by `hotness_score` and others by `score`?

app/services/articles/feeds/large_forem_experimental.rb
```diff
@@ -126,26 +126,74 @@ def score_comments(article)
...
-          .where("score >= ? OR featured = ?", SiteConfig.home_feed_minimum_score, true)
-          .order(hotness_score: :desc)
-        featured_story = hot_stories.where.not(main_image: nil).first
         if user_signed_in
-          hot_story_count = hot_stories.count
-          offset = RANDOM_OFFSET_VALUES.select do |i|
-            i < hot_story_count
-          end.sample # random offset, weighted more towards zero
-          hot_stories = hot_stories.offset(offset)
+          hot_stories = experimentalal_hot_story_grab
```
mstruve, line 130:
```suggestion
          hot_stories = experimental_hot_story_grab
```

app/services/articles/feeds/large_forem_experimental.rb
```diff
@@ -126,26 +126,74 @@ def score_comments(article)
...
+          featured_story = hot_stories.where.not(main_image: nil).first
         end
         [featured_story, hot_stories.to_a]
       end
 
       private
 
+      # Disable complexity cop to allow for variant-driven method
+      # rubocop:disable Metrics/CyclomaticComplexity
+      def experimentalal_hot_story_grab
```
mstruve, line 150:
```suggestion
      def experimental_hot_story_grab
```

app/models/reaction.rb
```diff
@@ -36,7 +36,6 @@ class Reaction < ApplicationRecord
   after_create :notify_slack_channel_about_vomit_reaction, if: -> { category == "vomit" }
   before_destroy :bust_reactable_cache_without_delay
   before_destroy :update_reactable_without_delay, unless: :destroyed_by_association
-  after_create_commit :record_field_test_event
```
mstruve, line 39: Bye callback 🎉

mstruve (commented):
Couple of small comments/questions.

One more question I have. Is there an advantage to running many experiment variants(5) at once such as this PR does? 

I ask bc there are a lot of different things going on in this PR and different levers being pulled when it comes to fetching articles. I wonder if we might get better and clearer data if we only ran 1 or 2 experiment variants at a time. I think that would also make reviewing code changes like this a bit easier.

app/services/articles/feeds/large_forem_experimental.rb
benhalpern: "score" is the overall ranking of a post, regardless of when it was created. "hotness" takes into account the age of the post. So if we are not selecting for a time period, "hotness score" works because it has a bias towards recency.

This PR proposes some alternatives to using hotness score which selects _just_ for score, but cuts off how recent based on when the person was most recently active (based on their last page view)... So if you haven't been to the site in a few days, you'll get top posts instead of just current "hot" ones. But if you have been checking back more often you get the more recent top posts. This caters the feed a little more to the browsing habits of the user...

I have no idea if this is going to be more relevant and engaging, but we'll have some results in a few weeks to see. I think new complexity is isolated and will be pulled at the end of the trial.

app/services/articles/feeds/large_forem_experimental.rb
```diff
@@ -126,26 +126,74 @@ def score_comments(article)
...
+            .where("published_at > ?", start_time)
+            .page(@page).per(@number_of_articles)
+            .order(score: :desc)
+        when "top_articles_since_last_pageview_7_days_max" # Top articles since last page view (max 3 days)
+          start_time = [(@user.page_views.last&.created_at || Time.current) - 12.hours, 7.days.ago].max
+          articles = Article.published.limited_column_select.includes(top_comments: :user)
+            .where("published_at > ?", start_time)
+            .page(@page).per(@number_of_articles)
+            .order(score: :desc)
+        when "combination_only_tags_followed_and_top_max_7_days" # Top articles since last page view (max 3 days)
```
fdoxyz, line 178:
```suggestion
        when "combination_only_tags_followed_and_top_max_7_days" # Top articles since last page view (max 7 days)
```

app/services/articles/feeds/large_forem_experimental.rb
```diff
@@ -126,26 +126,74 @@ def score_comments(article)
...
+            .order(hotness_score: :desc)
+          # We only want to limit the posts to tagged_with if the participant follows enough tags.
+          articles = articles.tagged_with(followed_tags, any: true) if followed_tags.size > 4
+        when "top_articles_since_last_pageview_3_days_max" # Top articles since last page view (max 3 days)
+          start_time = [(@user.page_views.last&.created_at || Time.current) - 12.hours, 3.days.ago].max
+          articles = Article.published.limited_column_select.includes(top_comments: :user)
+            .where("published_at > ?", start_time)
+            .page(@page).per(@number_of_articles)
+            .order(score: :desc)
+        when "top_articles_since_last_pageview_7_days_max" # Top articles since last page view (max 3 days)
```
fdoxyz, line 172:
```suggestion
        when "top_articles_since_last_pageview_7_days_max" # Top articles since last page view (max 7 days)
```

app/services/articles/feeds/large_forem_experimental.rb
```diff
@@ -126,26 +126,74 @@ def score_comments(article)
...
+          start_time = [(@user.page_views.last&.created_at || Time.current) - 12.hours, 7.days.ago].max
+          articles = Article.published.limited_column_select.includes(top_comments: :user)
+            .where("published_at > ?", start_time)
+            .page(@page).per(@number_of_articles)
+            .order(score: :desc)
+        when "combination_only_tags_followed_and_top_max_7_days" # Top articles since last page view (max 3 days)
+          start_time = [(@user.page_views.last&.created_at || Time.current) - 12.hours, 7.days.ago].max
```
fdoxyz, line 179: **Question/Suggestion:** I may be way off with this train of thought, but based on this line (that exists in a few of the variants)

```ruby
[(@user.page_views.last&.created_at || Time.current) - 12.hours, 7.days.ago].max
```

These are the two options:
1. If the user has no page views then we're choosing a `start_time` of 12 hours ago (`Time.current - 12.hours` which will always be greater than `7.days.ago`)
2. Otherwise we're choosing a `start_time` of the max of `@user.page_views.last&.created_at - 12.hours` or `7.days.ago`

This means that the newer users will get hot articles from that small window. 

This alternative changes that for option 1 (newer users):

```ruby
[(@user.page_views.last&.created_at || 7.days.ago) - 12.hours, 7.days.ago].max
```

This way the two options now are:
1. If the user has no page views always go with the max range (because the `7.days.ago - 12.hours` will not be chosen over `7.days.ago`)
2. Same as before (max of `@user.page_views.last&.created_at - 12.hours` or `7.days.ago`)

This only affects users that have no page views and I'm not sure if it would have a noticeable impact, so feel free to disregard this in case it doesn't make much sense

fdoxyz (approved):
I gave it a spin locally even though QA request was more for sanity check

It LGTM but I left a few non-blocking questions/suggestions inline

app/services/articles/feeds/large_forem_experimental.rb
benhalpern: Yeah that makes sense!

mstruve:
I think this is close, one question that I posted above in a comment that I would still love to know your thought process on is
> Is there an advantage to running many experiment variants(5) at once such as this PR does?

> I ask bc there are a lot of different things going on in this PR and different levers being pulled when it comes to fetching articles. I wonder if we might get better and clearer data if we only ran 1 or 2 experiment variants at a time. I think that would also make reviewing code changes like this a bit easier.

benhalpern:
@mstruve since there's only one test going on, and we're only modifying a single i/o method, I think it's sensible to test a few related concepts. There are really only a couple ideas here, but it would be a shame if we went with the "3 day" variant, and missed on the "7 day" of a similar test. Now if the 7 day wins we could quickly spin up a subsequent test related to that one with perhaps 6-10 days.

Since we're not doing multivariant testing, and the code that actually varies here is only in one place I think the complexity is pretty isolated to the point where we _could_ run dozens of these if we wanted. But I think the 6 total variants is a decent happy medium as a human-comprehensible case statement.

I think the changes to the test infrastructure in this PR mean more is going here outside of what we're actually testing. Hopefully the simplicity we gain from those changes will mean the next PR will have more code focusing on the thing we're testing and a little lest on supporting code.

mstruve (approved):
Let's give it a shot!

benhalpern:
Awesome @mstruve I decided to remove the `merge by any core` because I feel like being around when this goes live and I'll deploy this a little later today.

PR #7199, opened 2020-04-09
Add Custom Sorting and Scoring For Feed Content from Elasticsearch

mstruve:
## What type of PR is this? (check all applicable)
- [x] Bug Fix
- [x] Optimization

## Description
There are a few things happening in this PR which are going to help us get more relevant results from Elasticsearch. 

- Currently, we weight heavily towards search_score, Elasticsearch will weigh more towards match score but I added [function scoring](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#score-functions) to allow us to use the article score to influence the match score. This allows us to return higher quality match results.  
- In Algolia we use the combination operator AND. In Elasticsearch we will be using OR with combined with a [minimum_should_match](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-minimum-should-match.html) value of 2. The minimum_should_match will allow us to get more precise results but the OR will prevent us from often returning empty results which happens a lot now. One example search I found that currently returns nothing is "Hamburger button in react-bootstrap" and now it will return articles about the Hamburger button in react. 
- I added [boost values](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#query-string-multi-field) to tag names, title, and body text for searching. This means matches in those fields will be boosted by the number after the carrot
- In testing, I was able to get very close to our current search results for common searches such as React, Python, Rails, Docker, and <blank search>.

I have been doing a lot of testing running searches on Algolia vs Elasticsearch using Kibana. Once this is live we can use the below script to run test searches in a console and compare them to Algolia's in the UI. cc @benhalpern 
```ruby
search = "React"
results = Search::FeedContent.search_documents(params: {search_fields: search, class_name: "Article"})
results.map{|h| { title: h["title"], tags: h["tags"].map{|t| t["name"]}, _score: h["_score"]}}
```

## Related Tickets & Documents
https://github.com/thepracticaldev/dev.to/projects/6#card-35511227

## Added tests?
- [x] yes

alt_text

PR #11985, opened 2020-12-21
Spike: Use Postgres full text search

citizen428:

- [X] Spike
- [X] Refactor

## Description

The purpose of this PR is to spike out using Postgres full-text search (FTS) instead of Elasticsearch for our search needs. This is a rough first cut I spent less than 2h on so far, the idea was mainly to illustrate how Postgres FTS works. If we decide to go ahead with this we can work on a proper RFC.

### Notes

* We're using the [pg_search](https://github.com/Casecommons/pg_search) for most of the heavy lifting. I've used this gem extensively in the past, but never in multi-search mode only with search scopes (see [Multi-search vs. search scope](https://github.com/Casecommons/pg_search#multi-search-vs-search-scopes)).
* I've only added search to two models (`Article` and `PodcastEpisode`) for now and only for two fields on each.
* I added a service for index management (`Search::Indices`) which is quite convenient for quickly doing things from a Rails console or similar. Note that the `pg_search` gem also comes with [Rake tasks](https://github.com/Casecommons/pg_search#rebuilding-search-documents-for-a-given-class) for this purpose.
* I enabled two Postgres extensions ([unaccent](https://www.postgresql.org/docs/13/unaccent.html) and [pg_trgm](https://www.postgresql.org/docs/13/pgtrgm.html) because every search I ever built needed them sooner than later.
* I added an example search service (`Search::Multisearch`) and specs to show how things work.

### What's next?

* [X] Adding a GIN index on `pg_search_documents.content`.
* [ ] Add example for `pg_search_scope`
* [ ] Fine-tune search options ("or" searches, negation, prefix searches, trigram config, etc.)
* [ ] Adding one or two more models
* [ ] Adding more attributes to the different models
* [ ] Implement at least one example of custom indexing with `rebuild_pg_search_documents`
* [ ] You tell me 😃 

https://github.com/forem/InternalProjectPlanning/issues/381

1. Check out this branch.
2. Play around with `Search::Indices` from a `rails console`, e.g.
    ```ruby
    Search::Indices.clear!(Article)
    Search::Indices.rebuild!(Article)
    ```
3. Play around with the search function itself, e.g.
    ```ruby
    Search::Multisearch.call("search term") # depending on your seed data
    ```
    Note that the search options aren't fine-tuned yet, so things like partial matches, etc. won't work.

No UI yet.

## Added tests?

- [X] Yes

## Added to documentation?

- [X] No documentation needed, only a spike for now


app/models/podcast_episode.rb
```diff
@@ -3,6 +3,9 @@ class PodcastEpisode < ApplicationRecord
     duration_in_seconds
   ]
 
+  include PgSearch::Model
+  multisearchable against: %i[title body_text] # spike, limited attributes
```
citizen428, line 7: It's a total coincidence that I'm searching against the same attributes here, I'll add some distinct ones in the future.

config/initializers/pg_search.rb
```diff
@@ -0,0 +1,7 @@
+# See:
+# https://github.com/Casecommons/pg_search#trigram
```
rhymes, line 2: this is <https://github.com/Casecommons/pg_search#trigram-trigram-search>

rhymes (approved):
I've played a bit with it both in Italian and in English and it seems to work! 

A couple of questions:

- can multisearchable and scopes co-exist? I'd imagine the latter to be useful for tag autocompletion for example. 
- can it index documents asynchronously? It might not be necessary to have (an `INSERT` should be faster than a network HTTP call) and we can always rebuild the full index in a worker if necessary

Nice!

config/initializers/pg_search.rb
citizen428: It is indeed, somehow that got cut off. Thanks. I fixed it locally now, will push when I update the settings.

citizen428:
> can multisearchable and scopes co-exist?

I see no technical reason why they can't. Multi-search works through one shared table (`pg_search_documents`) and is accessed through `PgSearch.multisearch(term)`, the other goes through the individual model (and potential `tsvector` columns on the same table if we add them) and is used via the scope, e.g. `Article.search_by_title_or_body`. So as long as we call the correct method depending on context, this hopefull just works™.

> can it index documents asynchronously? 

By default, it doesn't. I think this isn't necessarily a problem because as you said, updates are quite simple:

```ruby
INSERT INTO pg_search_documents (...)
SELECT ...
FROM model;
```

BUT: the gem allows us to override `rebuild_pg_search_documents` in any multi-searchable class, so theoretically we should be able to put the indexing code into a worker and just call that from there, something like this:

```ruby
def self.rebuild_pg_search_documents
  Articles::ComplexIndexWorker.perfom_async(...)
end
```

The corresponding worker would look like this:

```ruby
module Articles
  class BustCacheWorker
    include Sidekiq::Worker

    def perform(article_id)
      article = Article.find(article_id)
      
      connection.execute <<~SQL.squish
        INSERT INTO pg_search_documents (searchable_type, searchable_id, content, created_at, updated_at)
        SELECT 'Article' AS searchable_type,
               article.id AS searchable_id,
               (article.title || ' ' || article.body_text) AS content,
               now() AS created_at,
               now() AS updated_at
        FROM articles
      SQL
    end
  end
end

```

rhymes:
Love the flexibility of this gem :)

mstruve:
Nice! 

This cleanly demonstrates that we can get basic functionality out of this gem. I want to take this a step further and spike out what the code looks like that actually runs these searches and how that will sit next to Elasticsearch code. What are your thoughts or ideas about how the two code paths exist together in a clear manner so we can easily turn on one and then the other?

What kind of performance metrics do we get out of complex queries? I think it would be valuable to see a pg search setup that mimics exactly a normal Elasticsearch(searching all of the same fields that we do) we would see from our home page and then look at the comparison.

citizen428:
> I want to take this a step further and spike out what the code looks like that actually runs these searches and how that will sit next to Elasticsearch code.

This is something I was a bit reluctant to do in this spike. The point of the service object was to show what a search controller would eventually call. Having two coexisting code paths at this point in time doesn't seem particularly useful for the following reasons:

* We have no idea at all how the infra side of this will look. Shouldn't we do a spike there as well, that covers optional infra like ES during both setup and later (i.e. create a Forem without ES, then add it)?
* We aren't sure yet if PG and ES search will be mutually exclusive. I'm not sure this needs to be a clear-cut either/or choice, we could, for example, keep basic search always inside PG and leverage ES for advanced queries only (e.g. MLT queries for content recommendations in large communities). This is are the two main directions I wanted to explore during the RFC phase, frontloading them widens the scope of this spike considerably beyond what I mentioned in the "What's next" section above.

LMK what you think and how much time we're willing to invest in this spike, I'm generally very open to exploring more, we just need to make sure we allocate our time wisely between the distinct phases (spike, RFC, implementation) of this project.

> What kind of performance metrics do we get out of complex queries? I think it would be valuable to see a pg search setup that mimics exactly a normal Elasticsearch(searching all of the same fields that we do)

Generally speaking, if you use `tsvector` columns and have the right indexes in place search performance is about the same as with Elasticsearch. I recently helped a friend with Postgres FTS, the DB now does fuzzy string searching (similarity searches) over 1.5M rows in ~30ms. But fleshing this out a bit more was intended to be part of this spike. 

But — and I have to defer to @benhalpern a bit here — performance by itself seems to make up a relatively minor part of this decision. Less moving parts, more straight-forward developer experience, less infra setup, and more efficient hosting of smaller Forems are part of the bigger picture here. I'd wager that for all intents and purposes, performance will be within an acceptable delta of what we have now.

To summarize items that need a decision:

1. Should we do an infra spike in parallel? This would be the first feature where a simple Rails-side feature flag isn't enough, so I'd feel better if we get some more clarity on that end of the stack too.
2. Are we going on for Postgres **or** Elasticsearch or Postgres **and** Elasticearch (progressive enhancement)? Do we want to make this call now, or during the RFC phase?
3. How much time do we spend on performance measurements? Where do we perform them?

db/migrate/20201222070116_add_index_to_search_documents_content.rb
```diff
@@ -0,0 +1,11 @@
+class AddIndexToSearchDocumentsContent < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :pg_search_documents,
+              %[to_tsvector('simple', coalesce("pg_search_documents"."content"::text, ''))],
```
citizen428, line 6: Note: since we support multiple languages (at least on DEV) we can't use a language-specific dictionary here. This means normalization steps like stemming and stop word removal will be skipped:

```sql
PracticalDeveloper_development# SELECT to_tsvector('simple', 'The men ran, the woman runs');
┌──────────────────────────────────────────────┐
│                 to_tsvector                  │
╞══════════════════════════════════════════════╡
│ 'men':2 'ran':3 'runs':6 'the':1,4 'woman':5 │
└──────────────────────────────────────────────┘
(1 row)

PracticalDeveloper_development# SELECT to_tsvector('english', 'The men ran, the woman runs');
┌───────────────────────────────────┐
│            to_tsvector            │
╞═══════════════════════════════════╡
│ 'men':2 'ran':3 'run':6 'woman':5 │
└───────────────────────────────────┘
(1 row)
```

citizen428:
@mstruve I think I just had an idea on how to make this more comparable to what we have right now, without spending too much time on it. 🤞  I'll explore that idea tomorrow.

config/initializers/pg_search.rb
```diff
@@ -1,7 +1,13 @@
 # See:
-# https://github.com/Casecommons/pg_search#trigram
+# https://github.com/Casecommons/pg_search#trigram-trigram-search
 # https://github.com/Casecommons/pg_search#ignoring-accent-marks
 PgSearch.multisearch_options = {
-  using: %i[tsearch trigram],
+  # using: %i[tsearch trigram],
+  using: {
+    tsearch: {
```
citizen428, line 7: I just added two basic options for now because I believe the match what we do with Elasticearch.

db/migrate/20201222070116_add_index_to_search_documents_content.rb
rhymes: Makes sense, we also don't know which languages other Forems use or will use

app/models/tag.rb
```diff
@@ -1,6 +1,9 @@
 require_relative "../lib/acts_as_taggable_on/tag"
 
 class Tag < ActsAsTaggableOn::Tag
+  include PgSearch::Model
+  pg_search_scope :search_by_name, against: :name, using: { tsearch: { prefix: true } }
```
citizen428: This is an example of a simple search scope /cc @rhymes

app/services/search/postgres/tag.rb
```diff
@@ -0,0 +1,13 @@
+module Search
+  module Postgres
+    class Tag
+      def self.search_documents(term)
```
citizen428, line 4: Here we are using the scope from above. Note that we can use the normal AR method chaining, so we can easily return attributes that weren't part of the search. I tucked `as_json` on the end to get a result similar to what the current search controller returns:

```ruby
# current ES code
[
  [0] {
      "hotness_score" => 175,
                 "id" => 7,
    "last_indexed_at" => "2020-12-23T05:22:45.878Z",
               "name" => "javascript",
         "rules_html" => nil,
      "short_summary" => nil,
          "supported" => true
  },
  [1] {
      "hotness_score" => 160,
                 "id" => 6,
    "last_indexed_at" => "2020-12-23T05:23:07.046Z",
               "name" => "java",
         "rules_html" => nil,
      "short_summary" => nil,
          "supported" => true
  }
]

# new PG code
[
  [0] {
    "hotness_score" => 175,
               "id" => 7,
             "name" => "javascript",
       "rules_html" => nil,
    "short_summary" => nil,
        "supported" => true
  },
  [1] {
    "hotness_score" => 160,
               "id" => 6,
             "name" => "java",
       "rules_html" => nil,
    "short_summary" => nil,
        "supported" => true
  }
]
```

The PG code is only missing the `last_indexed_at` attribute, but I suppose we could do something to get that back too. We also didn't need to pass in an explicit wildcard character since I enabled the prefix search option on the scope. 

/cc @rhymes

rhymes: cool! From a brief look it seems like we're using `last_indexed_at` only as stored metadata, added during indexing. We're not using it anywhere during query time. 

Probably solvable with https://github.com/Casecommons/pg_search#more-options, section "Specify additional attributes to be saved on the pg_search_documents table" but it requires new fields and the usage of `multisearchable`, not sure it's worth it for now

db/migrate/20201221033646_pg_search_extensions.rb
```diff
@@ -0,0 +1,11 @@
+class PgSearchExtensions < ActiveRecord::Migration[6.0]
+  def up
+    enable_extension("unaccent") unless "unaccent".in?(extensions)
```
maestromac, line 3: question: Is the `unless` condition here redundant? [The method](https://apidock.com/rails/v6.0.0/ActiveRecord/ConnectionAdapters/PostgreSQLAdapter/enable_extension) use `CREATE EXTENSION IF NOT EXISTS` which I assume would be doing the same thing?

citizen428: Yes, it's redundant, I hadn't looked at the method source code and generally try to make my migrations as idempotent as possible. I will change it, thanks!

app/services/search/postgres/tag.rb
citizen428: Ah, I should have been more specific, I tagged you mostly because you asked about using multisearch and search scopes together. 😃 In this specific case, `last_indexed_at` wouldn't make sense, we don't have a dedicated `tsvector` column on this table (yet) so nothing is really "indexed".  The generated query builds the vector on the fly and uses the [`@@`](https://www.postgresql.org/docs/13/functions-textsearch.html) operator to check if it matches the query. An example query looks like this:

```sql
SELECT "tags".* FROM "tags" 
INNER JOIN 
  (SELECT "tags"."id" AS pg_search_id, tags.hotness_score AS rank 
   FROM "tags" 
   WHERE ((to_tsvector('simple', coalesce("tags"."name"::text, ''))) @@ (to_tsquery('simple', ''' ' || 'java' || ' ''' || ':*')))) AS pg_search_978c2f8941354cf552831b 
ON "tags"."id" = pg_search_978c2f8941354cf552831b.pg_search_id 
ORDER BY pg_search_978c2f8941354cf552831b.rank DESC, "tags"."id" ASC
```

I'm actually starting to doubt if this specific model needs FTS at all, the search scope could probably be replaced with a left-anchored `LIKE `query and an appropriate index.

db/migrate/20201221033646_pg_search_extensions.rb
citizen428: Updated ✅

citizen428:
I looked a bit deeper into allowing us to switch between the two search implementations, but I still think this may be a bit too involved for the scope of this spike. 

The current ES-based search implementation creates a lot of dependencies between different classes and modules: the models themself include a `Searchable` concern. This concern relies on two other classes (a serializer and a search class) to do the heavy lifting. Alas, the way these are referenced is through constants on the model, which aren't easy to override. An annotated example:

```ruby
## app/models/tag.rb

# Include concern
include Searchable

# Configure concern
SEARCH_SERIALIZER = Search::TagSerializer
SEARCH_CLASS = Search::Tag
DATA_SYNC_CLASS = DataSync::Elasticsearch::Tag

after_commit :index_to_elasticsearch, on: %i[create update]
after_commit :sync_related_elasticsearch_docs, on: [:update]
after_commit :remove_from_elasticsearch, on: [:destroy]

## app/models/concerns/searchable.rb

# these methods relied on constants defined in the model
def serialized_search_hash
  self.class::SEARCH_SERIALIZER.new(self).serializable_hash.dig(:data, :attributes)
end

def elasticsearch_doc
  self.class::SEARCH_CLASS.find_document(search_id)
end

def sync_related_elasticsearch_docs
  self.class::DATA_SYNC_CLASS.new(self).call
end
```

This makes it relatively hard to swap out things at runtime because we'd have to unregister callbacks etc. or hackily override them with no-ops because otherwise, we'd still have an implicit dependency on ES even when using Postgres FTS.

Now, a reasonable first step would be to update the `Searchable` concern to be more self-contained (note: all the code here is made-up on the fly and completely untested, it's only meant to illustrate ideas):

1. Eliminate the need for constants defined in the including model. Use methods like `self.search_class` instead which follow a Rails-y convention over configuration approach, e.g.
    ```ruby
    def self.search_klass
      Search.const_get(name)
    end
    ```
    In cases where the search class doesn't match the class name, one would simply override the method directly, which is more straightforward than changing constants.
2. Move callbacks into the concern too. It seems most models would use a reasonable default set of something like this:
    ```ruby
    after_commit :index_to_elasticsearch, on: %i[create update]
    after_commit :sync_related_elasticsearch_docs, on: [:update]
    after_commit :remove_from_elasticsearch, on: [:destroy]
    ```
    If there are exceptions, we could leverage things like the [module builder pattern](https://dev.to/appsignal/configurable-ruby-modules-the-module-builder-pattern-4483) to make them configurable at the site of inclusion. Here's a made-up example:
    ```ruby
    include Searchable.options(sync: %i[create update])
    ```
    The corresponding method in the concern would look like this template:
    ```ruby
    def self.options(index: %i[create update], sync: [:update], remove: [:destroy])
      # configure module based on options and return it
    end
    ```
    If we go down this route we could potentially even move the search class into the options, then we don't need to override anything at all.
3. Method names should probably be more generic, having the search backend included in method names like `index_to_elasticsearch` is hard to reconcile with different strategies.
4. Once the search concern has no more implicit dependencies, it'd be much easier to turn it into something like a strategy pattern where we e.g switch on a feature flag:
    ```ruby
    def self.options(index: %i[create update], sync: [:update], remove: [:destroy])
      strategy = ... # strategy picking logic goes here
      strategy.build_module(index: index: sync: sync, remove: remove)
    end
    ```
5. If the module builder pattern is too obscure/deemed too complex, we can hide it behind something that looks more Rails DSL-like, e.g.
    ```ruby
    has_search class_name: "MySearchClass", sync: %i[create update]
   ```
   This is purely a choice of public API and would not affect the underlying implementation too much.

This would be for the case where we want the two strategies to be completely either/or, but still has some unanswered questions:

* If they are switchable through a feature flag at runtime, we'd need to keep ES around on standby, which negates the purpose of doing this in the first place. This means we'd need a new level of feature flag, that is more tied in with systems, e.g. is a specific piece of infra available? I guess a quick and dirty PoC would simply inject environment variables into the containers.
* Once we have such a flag, we could make the strategy choice only once on app boot: if we have ES, use it, if not fall back to PG. This would simplify some of the things outlined above. We'd still have to look into how far we want to take this, e.g. would we only load one of the two gems and wrap their corresponding strategies in guard clauses like `if defined?(PgSearch)`?

So the longer I think about this, the more I favor the progressive enhancement approach:

app/services/search/postgres/tag.rb
rhymes: Probably the best way to know it is to generate lots of tags and to `EXPLAIN ANALYZE` the two queries. We have 73000 tags in production

citizen428: I mean we could do that, but when you look at the type of search that we perform here (prefix search of a single word against a list of words) it doesn't look like we need any advanced FTS features. But tbh this is probably outside the scope of this spike, optimizing single queries for one search type over the other can definitely wait. 😃

mstruve:
> Are we going for Postgres or Elasticsearch or Postgres and Elasticearch (progressive enhancement)? Do we want to make this call now, or during the RFC phase?

Here is where I have a pretty strong opinion. I think making a change like this and adding this sort of complexity ONLY makes sense if the purpose is to end up with an app that does not need Elasticsearch. The main goal of using Postgres search in the first place is so that we can eliminate the need for Elasticsearch for small Forems and for most dev/local environments to help decrease app size and increase spin up time. 

With that in mind I think the main focus of the spike should be can Postgres full-text search give us everything we need for basic search on a small scale Forem? Secondly, can we side-by-side it with Elasticsearch in our code so switching between the 2 options is seamless? 

If we think we have a path to achieve the second question cleanly in the code base then that is when I think we bring in @forem/systems and do a quick spike to make sure the infra can also do the same. @jdoss feel free to chime in if you have any large concerns that you think make a Postgres/Elasticsearch toggle not feasible. 

> How much time do we spend on performance measurements? Where do we perform them?

I don't think we need to spend any more time on this for the spike. I just wanted to get an idea since I am unfamiliar with Postgres fulltext search if it was inline with Elasticsearch at a small scale and for our query needs.

jdoss:
I am all for removing the need for Elasticsearch for smaller deployments. We can address the "toggle switch" of adding in ES to systems work down the road.

citizen428:
@mstruve I guess it wasn't clear enough that "progressive enhancement" meant "add-on, where Forems that want this pay for the associated infra costs". But anyway, we had this discussion on forem.team, so no need to rehash it here.

From tomorrow on (today I'm mostly playing catch-up after the holidays) I'll focus on two things:

1. Showing that PG can replace ES for basic searches. This is already WIP, but I want to add a couple more use cases/models.
2. A way to keep both code paths around, if that may be temporarily painful since the current code has some issues that make it hard to keep around in parallel without some ugly hacks, so I'd need to do some refactorings in the `Searchable` concern and related places first (see the "Potential solution" section of the forem.team post).

citizen428:
@jdoss 

> I am all for removing the need for Elasticsearch for smaller deployments.

This is definitely the goal for whatever option we decide on.

> We can address the "toggle switch" of adding in ES to systems work down the road.

The more generic problem of "we may need to do infra level changes for plugins/paid add-ons" is probably still something we want to at least start thinking about sooner than later.

citizen428:
Ok, sorry for bringing this up one more time @mstruve and @rhymes, I just want to make :100: sure we're on the same page to so we can agree on what to spend time on (for context, @rhymes volunteered to help out with this project, thanks for that).

---

If we want to keep both ES and PG search in the code and make them switchable, we have to refactor the current ES code quite a bit. I outlined a possible approach in the forem.team post, but that has some issues too:

1. The two searches work in fundamentally different ways. PG FTS doesn't use a dedicated indexing step or index name, so if we want to do something like a strategy pattern, the PG code would have to implement some methods as no-ops, just to follow the same contract. While not necessarily a problem per se, my gut doesn't feel good about this.
2. Duplication: what we store for search (attributes and custom attributes specified via blocks) is defined in the serializer classes. They output a format that is very much tailored for ES (a JSON hash), and not directly useful for PG. We'd have to at least partially duplicate this, which means we may end up having to do changes in two different places if we want to keep both code paths active and interchangeable.

These are some of the original reasons why I preferred a progressive enhancement approach, which apparently I described poorly, so let me do that one more time:

1. All basic search functionality moves to PG. This also gives us a huge incentive to dedicate resources to it and get it right.
2. All current ES code and system-level dependencies get removed.

This achieves our primary goal: local development gets greatly simplified, also for external contributors.

We could then define advanced search features for a (paid) add-on, which would require the following:

* A way for systems to add infra like ES on demand.
* A sort of plugin mechanism that allows us to add code for advanced search features relying on ES to an active Forem.

I think that for the purposes of a spike, having a proper implementation of two runtime swappable implementations is a lot of effort. I propose the following instead:

1. Make all necessary changes to be able to properly evaluate PG FTS. Acceot that temporarily both searches work in parallel.
2, Add a feature flag for toggling between search implementations in the simplest possible way, e.g. by changing the route definition to point at a different search controller.

Everything above is additive, no ES code will be removed.

3. We evaluate PG search on some live Forems, via the feature toggle.
4. If we think it's good enough for our needs, we pause and write a proper RFC. 
5. During the RFC process we have to make the final call on whether we go for 2 swappable implementations, or basic PG for everyone and optional ES as a paid add-on.

The reason I'm proposing this is that I don't want to spend a lot of time during the spike on making two parallel swappable implementations, without a proper design document behind it OR a good idea if this will even be good enough for us. That said, as soon as we have something deployed behind a feature toggle, we could still do a second spike PR to at least explore HOW two parallel search providers could look in practice, even if we don't finish the implementation 100%.

jdoss:
> The more generic problem of "we may need to do infra level changes for plugins/paid add-ons" is probably still something we want to at least start thinking about sooner than later. 

Agreed. It is definitely on our radar.

rhymes:
@citizen428 I understood and I'm onboard with the plan.

ps. I think the initial confusion was that what you call "basic search" misled me to think that the needed both "basic" and "advanced" to fulfill __present__ DEV and Forem needs

mstruve:
@citizen428 THANK YOU for the great write up and for coming up with what I think is a solid way to move forward. I am on board making all necessary changes to be able to properly evaluate PG FTS and accept that temporarily both searches work in parallel. Let's make sure all new code is well documented at a SPIKE or in whatever way we feel will convey to others(mainly contributors) that we are experimenting and not to "add on" to this new code. 

I think one of the hardest things to replicate will be article sorting. It is complicated even by ES standards and took me some time to perfect. I would propose starting with article searching and sorting and seeing if we can get matching functionality and performance there first. 

What I suspect is that we will get it good enough for small Forems but not be able to match DEV ES which is likely why I jumped immediately to how do we do both at once. Thanks for reining me back in :)

mstruve (approved):
Should we put a note in our docs or something high level so people don't see this code and attempt to mess with it while we are doing out spike? Might not even be worth it, idk, just thinking out loud. 

One question about enabling the extensions otherwise LGTM! Lets roll it out when a Heroku admin is online in the event we need a rollback or the db locks up(like it did earlier) for no good reason.

db/migrate/20201221033646_pg_search_extensions.rb
```diff
@@ -0,0 +1,11 @@
+class PgSearchExtensions < ActiveRecord::Migration[6.0]
+  def up
+    enable_extension("unaccent")
+    enable_extension("pg_trgm")
```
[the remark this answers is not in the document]
citizen428: These are both standard PG extensions (docs [here](https://www.postgresql.org/docs/12/unaccent.html) and [here](https://www.postgresql.org/docs/12/pgtrgm.html)), they can just be enabled.

According to [this](https://devcenter.heroku.com/articles/heroku-postgres-extensions-postgis-full-text-search) they are supported on Heroku and [this page](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html) also lists them as supported on Amazon RDS. If we use custom Postgres in our container builds we should make sure that the `postgresql-contrib` package is installed, cc @jdoss.

```sh
$ dnf install postgresql-contrib
```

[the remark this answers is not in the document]
jdoss: I made an issue (https://github.com/forem/infra/issues/165) for us to track this in Systems.

citizen428:
@mstruve 

> I think one of the hardest things to replicate will be article sorting. It is complicated even by ES standards and took me some time to perfect.

Could you maybe elaborate a little bit on the difficulties and interesting things in the current implementation? Is it about `feed_content` being a mapping that includes articles, comments, and podcast episodes, or is there something else that needs our attention?

mstruve:
When a sorting param such as time is not present we default to sorting by the "match score" of the documents. In other words, we sort by how well the documents match the query we are running. To do this we apply an [elasticsearch concept called a function score query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html). This tweaks the document match score based on the score of the actual article. [We do this in our code by applying more weight to really high scoring articles helping to boost them](https://github.com/forem/forem/blob/master/app/services/search/query_builders/feed_content.rb#L171). 

Implementation PR: https://github.com/forem/forem/pull/7199

We also use [a match_phrase query](https://github.com/forem/forem/pull/9562) to help us boost the score of articles when the words we search are adjacent.

rhymes:
I think we should start by evaluating if we can achieve something like "function score query" and "match phrase" with PostgreSQL FTS and the gem or how much different the results are. I'll investigate a bit

citizen428:
Thanks @mstruve, this was super helpful!

@rhymes The way to achieve this with the `pg_search` gem is [ranking](https://github.com/Casecommons/pg_search#ranked_by-choosing-a-ranking-algorithm). Note that ranking algorithms can get quite complex, as they can access columns, call SQL functions, etc. Ranking by "match score" is basically the default, the terminology used is just a bit different:

> By default, pg_search ranks results based on the :tsearch similarity between the searchable text and the query.

Note that the gem also [allows for normalizing](https://github.com/Casecommons/pg_search#normalization) scores, e.g. by taking content length into consideration.

rhymes:
Rather than be bogged down in trying to replicate everything 1:1 I’m going to try to setup tests that search with both and compare the results and we build the functionality from there

rhymes:
I'm looking at integrating `ChatChannelMembership` within Postgre search but I hit a roadstop: the ES index for that model is built on methods that are called at runtime. For example:

https://github.com/forem/forem/blob/7f3859a8b42f2ebacc3546a9d8ffb062d992a9bf/app/models/chat_channel_membership.rb#L38-L41

Given that this is the concatenation of data on the association (it comes from `ChatChannel#channel_name` and `#slug` and `channel_human_names` and if I understood correctly how `pg_search` works, we have two options:

1 - Use https://github.com/Casecommons/pg_search#searching-through-associations to search against those fields
2 - Stick this model in the multi-search `pg_search_documents` table and add the custom attributes to it, see https://github.com/Casecommons/pg_search#more-options

I'm going to try with option 1 and report back

spec/services/search/integration/tag_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "rails_helper"
+
+RSpec.describe "Compare ES search to PG search for Tag", type: :feature do
```
rhymes, line 3: This is the first of a possible bunch of "integration" tests that might help us tracking differences between ES and PG results

app/services/search/query_builders/feed_content.rb
```diff
@@ -87,11 +87,10 @@ def add_sort
         @body[:sort] = { @params[:sort_by] => @params[:sort_direction] }
       end
 
+      # Highlights search results in HTML using `<mark>` returning 2 fragments on 75 chars each ordering them by score
```
citizen428, line 90: `pg_search` supports highlighting and can be configured (what tag to use, etc.). See: https://github.com/Casecommons/pg_search#highlight

app/services/search/query_builders/feed_content.rb
```diff
@@ -148,6 +152,8 @@ def term_keys
         end.compact
       end
 
+      # Search fields based on a range
```
citizen428, line 155: Here we can leverage Postgres itself, as FTS queries can be combined with other conditions.

app/services/search/query_builders/feed_content.rb
```diff
@@ -156,28 +162,36 @@ def range_keys
         end
       end
 
+      # Builds the `simple_query_string` query
+      # This query uses a simple syntax to parse and split the provided query
+      # string into terms based on special operators. The query then analyzes
+      # each term independently before returning matching documents.
+      # https://www.elastic.co/guide/en/elasticsearch/reference/7.10/query-dsl-simple-query-string-query.html
```
citizen428, line 169: Postgres FTS has ` websearch_to_tsquery` which behaves similarly. We'd need to check how to integrate this with the gem. 

https://www.postgresql.org/docs/11/functions-textsearch.html

citizen428:
> I believe this strategy, wrapped into a .search_documents() that adds the various other params and filters, can work. Thoughts?

I think this sounds reasonable @rhymes.

app/serializers/search/postgres/chat_channel_serializer.rb
```diff
@@ -0,0 +1,25 @@
+module Search
+  module Postgres
+    class ChatChannelSerializer < ApplicationSerializer
```
rhymes, line 3: I had to add this because most of the stuff that's returned in the ES "chat channel membership" is serialized at build time but we don't have that, so we need to serialize stuff in output. The format is different though

app/services/search/postgres/chat_channel.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      # NOTE: this is "kinda" working. To test this load the console and compare:
+      # 1 - `Search::ChatChannelMembership.search_documents(params: {channel_text: NAME_OF_ORG})`
+      # => vs `Search::Postgres::ChatChannel.search_documents(NAME_OF_ORG)`
+      # 2 - `Search::ChatChannelMembership.search_documents(params: {channel_text: NAME_OF_USER})`
+      # => vs `Search::Postgres::ChatChannel.search_documents(NAME_OF_USER)`
+      # You will see that the number of returned items is the same for both, the difference is that the
+      # ES based index contains a lot more info which is stored at indexing time related to membership
+      # but we don't have that info inside the DB so we need to use a serializer to get it
+      # Also, the JS layer kinda expects a ChatChannelMembership row, not a ChatChannel
```
rhymes, line 14: This. I've not written any tests for this, just played with it in the console. Still unsure if this is the best strategy as it does quite a few tricks

maestromac (commented):
question @rhymes , I might be misinterpreting it and I know you are currently just testing it out but doesn't it feels off to swap between ChatChannel and ChatChannelMemberhsip depending on which search strategy is being used?

I also don't know if that will cause any problem but certainly, it will raise some confusion.

rhymes:
@maestromac the thing is, it's a bit odd at the present day as we're essentially duplicating info about the channel in each entry in the `membership` index ES, by denormalizing data. That said, I wonder if the right solution here is to understand how we're using the membership index in the first place and see which of the following three is the better strategy:

- move it over to search channels
- keep searching memberships and find a way to paste in channel data
- split the two indexed based on need

For example, I go to dev.to/connect and type your username in the searchbox, I find our direct channel. If I type "Michael" I find all the channels where the word "Michael" appears in, and the chat between me and Michael T. BUT I'm 100% certain he appears in other channels I'm a member of but don't appear in the results. 

I have to investigate how this index is used in Preact/JS because it behaves oddly currently. I'll probably finish the experiment with the other "public" indices used by the search pages and then circle back to this one

app/services/search/postgres/chat_channel.rb
citizen428: I think "best strategy" is outside the scope of the spike. It seems to be a viable strategy though, improving on that would be part of the RFC anyway.

spec/services/search/integration/tag_spec.rb
citizen428: We should always keep in mind that we're willing to accept certain differences as long as the new results are generally sensible. If for example a few documents in the top 10 results aren't listed in the exact same order, I wouldn't necessarily consider that a problem (except for maybe the feed, where we do a lot of extra work to establish a certain order).

rhymes:
I'm currently working on finishing `Search::ReadingList` and we're getting closer to "feature similarity", see https://github.com/forem/forem/blob/bb154883a2aa2efd7cd66c854416aabaf593e392/app/services/search/postgres/reading_list.rb (definitely not cleaned up, as most everything is inline).

I configured highlighting here https://github.com/forem/forem/blob/bb154883a2aa2efd7cd66c854416aabaf593e392/config/initializers/pg_search.rb#L10-L16 - highlighting in PG is slightly different than ES, I'm not returned which fields was used to higlight which I think it's normal in the "multisearch" scenario, as it matches the whole text. That said, we might not need this information.

I'm yet to find out how to replicate this ranking in ES exactly:

https://github.com/forem/forem/blob/1f90c63eacbba0fbb2fe011a1a8f0e54de773dc0/app/services/search/query_builders/feed_content.rb#L6-L10

rhymes:
I hit a bump in the road because `acts-on-taggable-on` is not usable directly from "multisearch":

```ruby
[1] pry(main)> PgSearch.multisearch("Handful").tagged_with
NoMethodError: undefined method `tagged_with' for #<PgSearch::Document::ActiveRecord_Relation:0x00007fac8044e128>
from /forem/vendor/bundle/ruby/2.7.0/gems/activerecord-6.0.3.4/lib/active_record/relation/delegation.rb:109:in `method_missing'
```

I wonder if we should find a way to join the FTS query with the `taggings` and `tags` table or if we should find a find a way to use `pg_search_scope` on reactions (I abandoned this idea pretty early as the FTS search is to be done on articles titles and bodies, not reactions).

I also can't use Ruby to filter on loaded articles because it'd break pagination (unless we load all the results and paginate in memory...)

See https://github.com/forem/forem/pull/11985/files#diff-cc0dbafb7a942d6ea1bdd1e295586513bfbceeb150e5e846621bd885dc5695f6R19 and https://github.com/forem/forem/pull/11985/files#diff-cc0dbafb7a942d6ea1bdd1e295586513bfbceeb150e5e846621bd885dc5695f6R40

Any ideas?

app/services/search/postgres/reading_list.rb
```diff
@@ -3,17 +3,23 @@ module Postgres
...
+        tagged_with_any = tags_mode.to_sym == :any
 
         reading_list_article_ids = user.reactions.readinglist.where(status: status)
           .order(created_at: :desc)
           .pluck(:reactable_id, :id)
 
         article_ids = reading_list_article_ids.map { |article_id, _reaction_id| article_id }
+
         articles = if term.present?
+                     # NOTE: [@rhymes] I'm not sure how to combine `pg_search`'s multisearch
```
citizen428, line 19: Unfortunately, I'm not completely sure either, as I've never used multi-search before, only `pg_search_scope` 😒 Maybe with a manual join pm` searchable_type` and `searchable_id`?

I actually did some digging now, [this issue](https://github.com/Casecommons/pg_search/issues/33) has some pointers including this interesting tidbit:

> It's not well documented, but `PgSearch.multisearch` really just calls through to a `pg_search_scope` called `PgSearch::Document.search`

citizen428:
> This is used in the reading list when you search for a term and also select one ore more tag from the sidebar

Before you spend too much time on it, do we have any idea how commonly a search like this is performed? Does ES provide any statistics in regards to the queries it runs?

app/services/search/postgres/reading_list.rb
rhymes: Interesting! I'm sure we can whip up something with a left join. Need to dig a bit into this

rhymes:
> Before you spend too much time on it, do we have any idea how commonly a search like this is performed? Does ES provide any statistics in regards to the queries it runs?

I explored the results of the [GET /_stats command](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-stats.html) but I couldn't find anything useful. You can enable audit logging but only on paid extensions: https://www.elastic.co/guide/en/elasticsearch/reference/7.10/enable-audit-logging.html

> Also, TIL you can do that on the reading list, maybe now I can actually use it 😆

ahahha

mstruve:
> do we have any idea how commonly a search like this is performed? Does ES provide any statistics in regards to the queries it runs?

Given these queries are all HTTP requests to our server [they will all show up in our logs](https://app.datadoghq.com/logs?from_ts=1610556218302&index=&live=true&query=search&to_ts=1610642618302)

rhymes:
Thanks @mstruve 

I'd say it's used: https://app.datadoghq.com/logs?cols=core_host%2Ccore_service&from_ts=1610556218302&index=&live=true&messageDisplay=inline&query=search%5C%2Freactions%20AND%20tag_names&stream_sort=desc&to_ts=1610642618302

Not a lot, but I wouldn't get rid of that feature if possible :)

rhymes:
Related to the reading list:

I decided to do with the double query for now, as joining 4 tables (`pg_search_documents`, `articles`, `taggings`, `tags`) while using `acts-as-taggable-on` didn't work for me: https://github.com/forem/forem/pull/11985/commits/a167ad3a5e95d80bbcc7ee74eec286f600c8d5ba

This double query only happens within the space of a user's reading list when they both search and select tags. If we find it's too slow in production we can always revisit this idea

app/services/search/postgres/reading_list.rb
```diff
@@ -16,36 +16,93 @@ def self.search_documents(user, term: nil, status: nil, tags: nil, tags_mode: :a
...
+            article.search_highlight = doc.pg_search_highlight
+            article.search_rank = doc.pg_search_rank
+            article
+          end
+        end
+      end
+      private_class_method :full_text_search
+
+      # borrowed from https://github.com/mbleigh/acts-as-taggable-on/blob/47da5036dea61cb971bfaf72de5fa93c85255307/lib/acts_as_taggable_on/taggable/core.rb#L110
+      def self.tagged_with(article_ids, tags, tagged_with_any)
```
citizen428, line 86: Could we potentially do this by changing the first part of this method to use `super`? I think

```ruby
def self.tagged_with(article_ids, tags, tagged_with_any)
  relation = super(tags, { any: tagged_with_any})
  relation.where(id: article_ids)
end
```

rhymes: Can't use `super` as the class doesn't include the `acts-as-taggable-on` module BUT we can just call it directly :D 

```ruby
      def self.tagged_with(article_ids, tags, tagged_with_any)
        relation = Article.tagged_with(tags, any: tagged_with_any)
        relation.where(id: article_ids)
      end
      private_class_method :tagged_with
```

rhymes:
update, cc. @citizen428 @mstruve 

I added `Search::Postgres::Feed` and manually tested it with a some queries and some screenshots (to see if results were similar) and we're almost there, with a few tweaks and a couple of more filters (I haven't implemented "search only my posts" yet). 

Some notes:

- we can search `body_markdown`, though the current search actually searches on `strip_tags(processed_html)`, it's solvable by adding a new column to the table probably but this is one of the earliest notest I wrote down this morning and didn't go back to fix this. The major difference I think is that by searching only markdown we skip the rendered liquid tags
- it's doing A LOT of SQL queries and joins to load data, mostly because serializers are called at runtime in exit (not in input like in ES), but this can also be solved by adding columns to the `pg_search_documents` if we need then and avoiding serializers that need associated models (like `podcast_episode.podcast.creator` to build `podcast_episode.path`, we can just store `path` during the index rebuilding). It can be done, just a note
- we're using `associated_against` a bit but this option doesn't support indexes, which is fine when you search on one table at runtime, but in this case we also need to move that data inside `pg_search_documents`
- I add sorting fields to `pg_search_documents` because I tried a wild combo of `LEFT OUTER JOIN`  which semi-worked, but aside from being super slow, it doesn't also work with sorting for obvious reasons: `ORDER BY articles.published_at` only works for `articles` not `articles` and `comments` :D
- all the various function scores that we have now in ES need to be re-tuned with live results in my opinion, to see what is the best ranking algorithm in PG FTS, too hard to guess now. I'm talking about boosting and function scoring in ES:

```ruby
                "fields": [
                  "tags.keywords_for_search",
                  "tags.name^3",
                  "body_text^2",
                  "title^6",
                  "user.name",
                  "user.username",
                  "organization.name"
                ],
```

```ruby
      "functions": [
        {
          "filter": {
            "range": {
              "score": {
                "gte": 1000
              }
            }
          },
          "weight": 2.5
        },
        {
          "filter": {
            "range": {
              "score": {
                "gte": 500
              }
            }
          },
          "weight": 2
        },
        {
          "filter": {
            "range": {
              "score": {
                "gte": 100
              }
            }
          },
          "weight": 1.5
        }
      ]
```

Anyway, most of these are "solvable" things, either in the spike or the actual implementation. I think we're in a good place, barring performance :)

I'll do some perf testing and then we can decide what do do.

app/controllers/search_controller.rb
```diff
@@ -121,7 +121,18 @@ def reactions
...
+      approved: feed_params[:approved] == "true",
+      class_name: feed_params[:class_name],
+      id: feed_params[:id],
+      sort_by: feed_params[:sort_by],
+      sort_direction: feed_params[:sort_direction],
+      published_at: feed_params[:published_at],
+      page: feed_params[:page],
+      per_page: feed_params[:per_page],
+    )
+    # Search::FeedContent.search_documents(params: feed_params.to_h)
```
rhymes, line 135: tried using postgres locally for searches and the feed, haven't tried the tag search yet

app/models/article.rb
```diff
@@ -6,14 +6,32 @@ class Article < ApplicationRecord
...
+                    tags: %i[name keywords_for_search],
+                    user: %i[username name]
+                  },
+                  additional_attributes: lambda { |article|
+                    {
+                      hotness_score: article.hotness_score,
+                      published_at: article.published_at,
+                      public_reactions_count: article.public_reactions_count
+                    }
+                  },
```
rhymes, line 23: the magic to populate `pg_search_documents` during rebuilding

app/models/article.rb
```diff
@@ -6,14 +6,32 @@ class Article < ApplicationRecord
...
   include UserSubscriptionSourceable
 
+  include PgSearch::Model
+  multisearchable against: %i[body_markdown title],
+                  # do we need `keywords_for_search` ?
+                  associated_against: {
+                    organization: :name,
+                    tags: %i[name keywords_for_search],
+                    user: %i[username name]
+                  },
```
rhymes, line 16: this we should probably move to `against` + `additional_attributes` as associations are the antithesis of FTS indexes AFAIK

app/models/article.rb
```diff
@@ -6,14 +6,32 @@ class Article < ApplicationRecord
...
+                  additional_attributes: lambda { |article|
+                    {
+                      hotness_score: article.hotness_score,
+                      published_at: article.published_at,
+                      public_reactions_count: article.public_reactions_count
+                    }
+                  },
+                  if: :published?,
```
rhymes, line 24: not indexing unpublished articles for now :)

app/services/search/indices.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+
+    def self.rebuild!(klass)
+      PgSearch::Multisearch.rebuild(klass)
+    end
+
+    def self.rebuild_all!
+      MULTISEARCH_CLASSES.each { |klass| PgSearch::Multisearch.rebuild(klass) }
+      nil
+    end
```
rhymes, line 17: Got tired of typing `.rebuild(Article|PodcastEpisode|Comment)` each time I had to rebuild all of them. This would eventually be split in an inline version and an async one I guess

app/services/search/postgres/feed.rb
```diff
@@ -0,0 +1,103 @@
...
+        approved body_text experience_level_rating
+        experience_level_rating_distribution featured featured_number
+        hotness_score language published reactions_count score
+      ].freeze
+
+      def self.search_documents(
+        term: nil, approved: nil, class_name: nil, id: nil, sort_by: nil, sort_direction: nil, published_at: nil,
+        page: 1, per_page: 60
+      )
+        page = (page || 1).to_i + 1 # NOTE: the frontend starts at 0, need to fix this
```
rhymes, line 14: Postgres treats page 0 as if it were page 1, so we can't really just add `+1` to it, but we can always change the frontend start page when we're done and remove this bug

app/services/search/postgres/feed.rb
```diff
@@ -0,0 +1,103 @@
...
+        relation = if term.present?
+                     relation = Search::Multisearch.call(term)
+                       .with_pg_search_highlight
+                       .with_pg_search_rank
+                       .includes(
+                         searchable: [
+                           {  podcast_episode: {  podcast: :creator } },
+                           { articles: :organization },
+                           :tag_taggings, :tags, :comments
+                         ],
```
rhymes, line 26: the local log is a forest of bullet complaining about eager loading and stuff. Hopefully this will disappear when we copy data in `pg_search_documents` during indexing over having to use the serializers to fetch 3 indirections of associations

app/services/search/postgres/feed.rb
```diff
@@ -0,0 +1,103 @@
...
+                     relation = PgSearch::Document
+                     relation = relation.where(searchable_type: class_name) if class_name.present?
+                     select_articles_by_publication_date(relation, published_at, class_name)
+                   end
+
+        relation = sort_documents(relation, sort_by, sort_direction)
+
+        relation.page(page).per(per_page).map do |doc|
+          searchable = doc.searchable
+          serializer = ::Search.const_get("#{searchable.class}Serializer")
```
rhymes, line 42: the 👿

app/services/search/postgres/feed.rb
```diff
@@ -0,0 +1,103 @@
...
+          attributes = serializer.new(searchable).serializable_hash.dig(:data, :attributes)
+          prepare_doc(doc, attributes).as_json
+        end
+      end
+
+      # borrowed from Search::FeedContent.prepare_doc
+      def self.prepare_doc(doc, attributes)
+        attributes[:_score] = doc.respond_to?(:pg_search_rank) ? doc.pg_search_rank : 0
```
rhymes, line 50: this is because search ranking exists only on actual FTS search, not when we fetch `PgSearch::Document` in a projection

app/services/search/postgres/feed.rb
```diff
@@ -0,0 +1,103 @@
...
+        attributes[:id] = attributes[:id].split("_").last.to_i
+        attributes[:tag_list] = attributes[:tags]&.map { |t| t[:name] } || []
+        attributes[:flare_tag] = attributes[:flare_tag_hash]
+        attributes[:user_id] = attributes.dig(:user, :id)
+
+        if doc.respond_to?(:pg_search_highlight)
```
rhymes, line 56: same thing as ranking above

app/services/search/postgres/feed.rb
```diff
@@ -0,0 +1,103 @@
...
+
+        attributes.except(*FIELDS_TO_EXCLUDE_FROM_SERIALIZATION)
+      end
+      private_class_method :prepare_doc
+
+      def self.select_approved_articles(relation, approved, class_name)
+        return relation unless approved || (class_name && class_name != "Article")
+
+        relation
+          .joins("INNER JOIN articles ON articles.id = pg_search_documents.searchable_id")
```
rhymes, line 77: cheating a bit :) SQL FTW

app/services/search/postgres/feed.rb
```diff
@@ -0,0 +1,103 @@
...
+          .where(searchable_type: "Article")
+          .where(articles: { approved: approved })
+      end
+      private_class_method :select_approved_articles
+
+      def self.select_articles_by_publication_date(relation, published_at, class_name)
+        return relation unless published_at.present? || (class_name && class_name != "Article")
+
+        if published_at[:gte].present?
+          relation.where(published_at: published_at[:gte]..)
```
rhymes, line 87: love the new range syntax ❤️

config/initializers/pg_search.rb
```diff
@@ -0,0 +1,20 @@
...
+      any_word: true, # combine search terms with 'or', not 'and'
+      prefix: true,   # search for partial words
+      highlight: {    # https://github.com/Casecommons/pg_search#highlight
+        StartSel: "<mark>",
+        StopSel: "</mark>",
+        MaxFragments: 2,
+        MinWords: 5,
+        MaxWords: 10
+      }
```
rhymes, line 16: these are needed for highlighting

db/migrate/20201221024151_create_pg_search_documents.rb
```diff
@@ -0,0 +1,23 @@
+class CreatePgSearchDocuments < ActiveRecord::Migration[6.0]
+  def up
+    say_with_time("Creating table for pg_search multisearch") do
+      create_table :pg_search_documents do |t|
+        t.text :content
+
+        # columns needed for multisearch sorting (homepage feed and search)
+        t.integer :hotness_score, null: false, index: true
+        t.datetime :published_at, null: false, index: true
+        t.integer :public_reactions_count, null: false, index: true
```
rhymes, line 10: we need these fields because they are used for sorting and filtering

app/services/search/postgres/reading_list.rb
citizen428: Ah, yes, that works, thanks!

citizen428:
Thanks for the update @rhymes, this all seems very promising. My only slight concern is that the feed search may be pushing Postgres FTS to its limits. As I mentioned previously I've primarily used it for single table searches (both with the gem and `pg_search` scope and also without) and it's great for this use case, but it will most likely fall short of ES in complicated multi-model searches. 

The sweet spot is probably somewhere in the intersection of simplifying certain searches while pushing Postgres as far as we find reasonable. I think a deciding factor is how big we expect our average Forem instances to get, otherwise, it'll be hard to judge the tradeoffs we need to make.

rhymes:
@citizen428 I'm a little bit more optimistic. Most of these issues are related to inexperience and "single table search" is not different to "multisearch" (they are literally the same scope in the `pg_search` code), the difference is that the second one builds a common table where to store the data (what ES calls an index). 

The only thing I'm still not sure about is the boosting and function scoring, that might be beyond FTS capability but still, we need to measure if we can get to "good enough"

rhymes:
I did some benchmarking (here's the [script](https://github.com/forem/forem/pull/11985/commits/ee1fcf790b82e4409235058e5b0559a8e3bfe0e6)) and some profiling. I'm confidente that most of the time is spent for the serialization in output, this is most visible in the slowest of operations for PG, the feed query.

<details>
  <summary>Search benchmark results</summary>
  <pre>
Setup: 10000 articles...
Setup: 12000 comments...
Setup: 12 listings...
Setup: 14 tags...
Setup: 4002 users...

                            user     system      total        real
Search Username - ES    0.007555   0.003614   0.011169 (  0.015330)
Search Username - PG    0.004359   0.000735   0.005094 (  0.029955)

                        user     system      total        real
Search User - ES    0.004202   0.002236   0.006438 (  0.012275)
Search User - PG    0.011839   0.001064   0.012903 (  0.103016)

                       user     system      total        real
Search Tag - ES    0.002273   0.001231   0.003504 (  0.006272)
Search Tag - PG    0.004937   0.000335   0.005272 (  0.006755)

                               user     system      total        real
Search ReadingList - ES    0.008820   0.001940   0.010760 (  0.020357)
Search ReadingList - PG    0.066842   0.003069   0.069911 (  0.083829)
                                      user     system      total        real
Filter ReadingList by Tag - ES    0.007715   0.001827   0.009542 (  0.017877)
Filter ReadingList by Tag - PG    0.022054   0.000969   0.023023 (  0.028158)

                           user     system      total        real
Search Listing - ES    0.005956   0.002285   0.008241 (  0.015942)
Search Listing - PG    0.041710   0.001857   0.043567 (  0.053775)
                                       user     system      total        real
Filter Listing by Category - ES    0.003903   0.001583   0.005486 (  0.009473)
Filter Listing by Category - PG    0.027005   0.001054   0.028059 (  0.033770)
                                  user     system      total        real
Filter Listing by Tag - ES    0.003338   0.001435   0.004773 (  0.008053)
Filter Listing by Tag - PG    0.045754   0.002068   0.047822 (  0.054034)

                        user     system      total        real
Search Feed - ES    0.023186   0.004852   0.028038 (  0.081095)
Search Feed - PG    0.013941   0.000543   0.014484 (  0.019846)
                          user     system      total        real
Homepage Feed - ES    0.021767   0.003971   0.025738 (  0.053368)
Homepage Feed - PG    0.750492   0.024377   0.774869 (  0.836370)
                                 user     system      total        real
Homepage Feed - Week - ES    0.021425   0.001878   0.023303 (  0.041276)
Homepage Feed - Week - PG    0.840864   0.015377   0.856241 (  0.914789)
                                   user     system      total        real
Homepage Feed - Latest - ES    0.011536   0.001631   0.013167 (  0.031940)
Homepage Feed - Latest - PG    0.702258   0.016100   0.718358 (  0.778821)

  </pre>
</details>

I've used this script to do some CPU profiling as well:

```ruby
require "ruby-prof"

ActiveRecord::Base.logger = nil

result = RubyProf.profile do
  # Search::Postgres::Feed.search_documents(sort_by: :hotness_score, sort_direction: :desc, class_name: "Article")
  Search::Postgres::User.search_documents(term: "rhymes")
end

printer = RubyProf::GraphPrinter.new(result)
printer.print(STDERR, min_percent: 10)
```

and search for documents for the feed makes it really clear, which confirmed my suspicions and the amount of queries we're doing, that the issue is serialization:

```text
  66.90%   0.02%      0.970      0.000      0.000      0.970              120    *FastJsonapi::ObjectSerializer#serializable_hash 
```

This is an extract from that run: 67% of the time spent in the serializer.

Given this: should I continue to develop it in this spike or wait? Do we want to discuss (a)synchronously what to do? I'm confident we can shave off a lot of time by behaving more like ES (adding the columns we need during the document rebuild step) and minimizing serialization in output. After all, ES would be similarly slow if it had to build objects in output using SQL as well (that's why we serialize in input, not in output).

Let me know what y'all think @mstruve @citizen428

citizen428:
> @citizen428 I'm a little bit more optimistic. Most of these issues are related to inexperience and "single table search" is not different to "multisearch" (they are literally the same scope in the pg_search code), the difference is that the second one builds a common table where to store the data (what ES calls an index).

Let me clarify: I was aware that the search implementation is pretty much the same (though apparently it's exactly the same, which is even better). But for a multi-model search documents table, scoring/boosting seem way more important than for a search performed against a single table. Maybe that assumption is wrong though.

citizen428:
> Given this: should I continue to develop it in this spike or wait? Do we want to discuss (a)synchronously what to do?

I think it's a good insight and I'd be happy to discuss this in any way you see fit. I'll leave the decision as to whether this makes sense to explore as part of this spike to others, but I'm inclined to say yes.

mstruve:
> scoring/boosting seem way more important than for a search performed against a single table.

Serialization being slow is not surprising at all to me. I do think we should flesh out the scoring/boosting first bc I think that will be a harder problem to solve and one where the technology could limit us and if the technology limits us too much then we might have to rethink this all together. I think solving the serialization problem comes with a lot more options. what do you guys think of that?

rhymes:
**Status update**

TLDR; as feared there are quite a few limits on boosting and scoring but we can probably overcome them by unleashing the full capabilities of PostgreSQL

Let me start from the beginning.

We know that the serialization problem is solvable, you can basically call methods during indexing and join data as you wish, see [this commit](https://github.com/forem/forem/pull/11985/commits/459d8c4468e72bea65686d190de7f70874f3b54e).

After having "solved" that, I started going back and forth between the `SearchController`, the live website and the console and started fiddling with data in the console to fully understand how boosting in ES works so I could replicate it.

It's somewhat "easysh" to understand looking at live data: if you search for "Keats" in ES, it will favor articles that have "Keats" in the title, for some reason it boosts more comments whose article has "Keats" in the title over a second article that has the same keyword in the article. But yeah, following this (copied from `Search::QueryBuilders::FeedContent`):

```ruby
        search_fields: [
          "tags.keywords_for_search",
          "tags.name^3", # boost tag names by a factor of 3
          "body_text^2", # boost body text by a factor of 2
          "title^6", # boost title by a factor of 6
          "user.name",
          "user.username",
          "organization.name",
        ]
```

we can see that `title` is top, then tags, then the body and then the rest.

This can be replicated within single search table scopes in PG, by using [weighting](https://github.com/Casecommons/pg_search#weighting): you assign a letter to each field and it uses the order of letters to decide which field is boosted higher. I played with it with chat channels and it worked well.

The first road stop I hit: weights don't work at all for multisearch. If you try to set them in a `multisearchable` the gem breaks, that's because it expects just a list of columns or methods. It technically makes sense, as the point of `against:` in the multisearchable is to load text inside the `content` column and join it for the FTS algorithm to find it later.

The second instrument I'm given from the gem is [ranked_by](https://github.com/Casecommons/pg_search#ranked_by-choosing-a-ranking-algorithm) which does work with the multisearch BUT it requires a numeric weight as well.

It needs to be set in the global `multisearch_options` and it works, I played with this configuration and various alterations:

```ruby
PgSearch.multisearch_options = {
  using: {
    tsearch: {
      any_word: true, # combine search terms with 'or', not 'and'
      prefix: true,   # search for partial words
      highlight: {    # https://github.com/Casecommons/pg_search#highlight
        StartSel: "<mark>",
        StopSel: "</mark>",
        MaxFragments: 2,
        MinWords: 5,
        MaxWords: 10
      }
    },
  },
  # https://github.com/Casecommons/pg_search#ranked_by-choosing-a-ranking-algorithm
  ranked_by: "LENGTH(content)",
  ignoring: :accents
}
```

as you can see at the bottom and in the [ranked_by](https://github.com/Casecommons/pg_search#ranked_by-choosing-a-ranking-algorithm) documentation the only thing you can boost is numeric weights. In this toy example I tried both `LENGTH()`, `-LENGTH()` and I can verify it works. I can rank documents by the size of their `content` field for example.

The problem is: it's totally useless for our purposes :) I can fiddle with knobs and boost [similarity search](https://www.postgresql.org/docs/current/pgtrgm.html#id-1.11.7.40.6) with something like `ranked_by: ":trigram * 3"` which literally translates to `(similarity(unaccent('Keats'), (unaccent(coalesce("pg_search_documents"."content"::text, ''))))) * 3 AS rank` in the SQL. 

As you can see, there's no concept of "if you find `Keats` in the title favor this document". 

The last not fully formed idea I had was the following: we can add a column named `title` with the title itself in the `pg_search_documents` and give it a weight on the fly but we still need a `FUNCTION()` which returns a numeric score for such `title` column, depending on the search keyword. I do not think it's doable by default.

What we actually need is "multicolumn search" instead, `pg_search` is geared towards "single column search".

There's a slight chance it can be achieved, I found a POC in a comment in the gem's repository - https://github.com/Casecommons/pg_search/issues/348#issuecomment-294374642 - it's pretty clever, it uses similarity search on multiple columns and boosts the document with the highest score on the highest weighted column.

I tested the POC on the `pg_search_documents` and it works for what it was written for but as it uses the trigram for similarity, it ignores the pure FTS-builtin tsearch with stemming and so on. The logic it's sound though.

The weigthed algorithm used in the single table search translates to something like `(ts_rank((setweight(to_tsvector('simple', coalesce("chat_channels"."channel_name"::text, '')), 'A') || setweight(to_tsvector('simple', coalesce("chat_channels"."slug"::text, '')), 'B')`, by using the builtin [setweight()](https://www.postgresql.org/docs/current/textsearch-features.html#TEXTSEARCH-MANIPULATE-TSVECTOR) function we can score the documents.

We'd essentially need contribute code to the gem to support `multisearchable` on multiple columns, so that we can use weights on them and then boosting. [Postgres full-text search is Good Enough!](http://rachbelaid.com/postgres-full-text-search-is-good-enough/) has a great explanation of how all of this works (giving different weights to different columns).

Let me know what you think :)

config/initializers/pg_search.rb
```diff
@@ -14,7 +14,9 @@
         MinWords: 5,
         MaxWords: 10
       }
-    }
+    },
   },
+  # https://github.com/Casecommons/pg_search#ranked_by-choosing-a-ranking-algorithm
+  ranked_by: "LENGTH(content)",
```
citizen428, line 20: From what I understand, this is not taking the search scores into account but purely works on the document length. I'd have expected something like 

```ruby
:tsearch * length(content)
```

or similar.

rhymes: yeah, that was just a quick experiment. I also tried using `:tsearch` and the other algorithms. The point is you can't use `pg_search` to rank based on patterns on multiple columns which is what the core of the ES functionality is.

mstruve:
My thoughts on this after reading @rhymes summary is that we cannot duplicate Elasticsearch with pg_search and I think for smaller Forems with less content that might be OK. I think it might push us to improve our filtering options so that people have more explicit knobs such as author, tag, title, published_at, etc rather than having to rely solely on a beefy full-text search to figure out exactly what they want.

I would consider this Spike completed at this stage since we have a clear idea of what we can and cannot do with pg_search. The code here is nicely isolated and follows our existing pattern which would make it ideal for "turning on or off" based on configuration. I view the next step as making a decision about if this is an option that is worth moving forward with in order to simplify the overall base Forem architecture. Time to have some discussions with @jdoss and the product teams.

citizen428:
Agreed with Molly. I wonder if smaller, more specialized search scopes across various models and a revamp of the search interface could be "good enough" for a majority of communities. To be completely honest, for more complex searches on DEV and elsewhere I usually use Google with a site filter anyway, it's surprising how often that works better than any built-in search does.

vaidehijoshi:
> we cannot duplicate Elasticsearch with pg_search and I think for smaller Forems with less content that might be OK. I think it might push us to improve our filtering options so that people have more explicit knobs such as author, tag, title, published_at, etc rather than having to rely solely on a beefy full-text search to figure out exactly what they want.

This feels like the right approach to me. The discussion here gives me deja vu around the discussions of Fastly versus a "simpler, smaller" form of caching; we're having literally the same kind of discussion around [our barebones feed strategy over here](https://forem.team/vaidehijoshi/comment/31e), too.

It seems like we're all slowly circling around the broader philosophical product question of _"What should the basic Forem be able to do out of the box?"_. Forems with smaller content will not _need_ a beefy search, as @mstruve points out

> Agreed with Molly. I wonder if smaller, more specialized search scopes across various models and a revamp of the search interface could be "good enough" for a majority of communities.

I suspect that it would be Good Enough™️ and that the Forems that need the "high traffic" feed would also be the ones that might be interested in a beefier search -- whether we provide that or not is a different story. From where I sit, an ES-backed search would make for a nice add-on down the road, if anything.

citizen428:
@rhymes What's our plan for this PR? Do we keep it around, merge it or close it?

rhymes:
@citizen428 we haven't formulated a plan yet but I think we're going to proceed in steps, not using this PR but introducing feature flagged changes one at a time. I'll confirm later today and act accordingly

PR #11995, opened 2020-12-21
Enable paranoid mode for Devise

citizen428:

- [X] Security

## Description

This PR enables Devise's "paranoid mode" to prevent email enumeration. From the initializer comment:

>\# It will change confirmation, password recovery and other workflows
>\# to behave the same regardless if the e-mail provided was right or wrong.

This means if a user enters an email address in the forgot password form (or similar forms), they will no longer get any indication if the address they entered actually exists in the system.

Closes https://github.com/forem/InternalProjectPlanning/issues/180

Automatic:
* Disable paranoid mode in `config/initialzers/devise.rb` and see the spec fail.

Manual:
* Perform the following specs as outlined in the system spec 
    ```ruby
    visit new_user_session_path
    click_link "I forgot my password"
    fill_in "Email", with: "doesnotexist@example.com"
    click_button "Send me reset password instructions"
    ```
* Do this both with paranoid mode enabled and disabled.

## Added tests?

- [X] Yes
- 
## Added to documentation?

- [X] No documentation needed


nickytonline (approved):
Nice! 🚀

rhymes (approved):
Cool!

fdoxyz (approved):
🙌🏼

citizen428:
Ugh, nevermind the wrong branch name here, I got started on one issue then decided to tackler another first instead...

spec/system/user_logs_in_with_password_spec.rb
```diff
@@ -27,20 +27,6 @@ def submit_login_form(email, password)
       expect(page).to have_text("Invalid Email or password.")
     end
 
-    it "displays a message on the last login attempt" do
```
citizen428, line 30: These two specs got removed because the whole point of Devise's paranoid mode is not exposing information like this.

PR #12015, opened 2020-12-23
Fix profile field production display area functionality and remove SiteConfig.dev_to? hacks for better generalization

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This pull request adds functionality, tests and data update scripts to get us much closer to full profile generalization. It does not close every loop, but if we get this merged, we are in a better position to identify small last mile things without being in such limbo.

The follow-on RFC will be made easier by this, which is mostly clean up, as it does not set any new functionality in stone. I tried to include adequate specs in both setting and displaying profile fields such that we have regression tests in place which will help ongoing development.

### What this PR includes

- Fix fields on existing Forems (which are all currently `display_area: left_sidebar`)
- Removes all `SiteConfig.dev_to?` hacks
- Establishes a few special display type of fields related to `location` and `employer_`...
  - In the future we could further generalize this, but these are both universal types of fields which _if present_ logically could always have a special specification. This seems like a sound way to handle it now.
- Establishes a special way to handle _current_ social link attributes.
  - This seems less longterm great, but this seems like a decent place we can land to set up for future improvements.

### What this PR does not include

- This PR does not include an answer to the question of why `ImportFromCsv` is not working as expected in production, because it seems to work fine in development.
  - Maybe it is fixed in the most recently deployed Forems?
  - This should fix the problem going forward for now as long as long as the data update script is still live.
- This does not include improvements to `admin/profile_fields` which might make the UX _safer_ for rollout. It is still something that we should ensure we have UX which is hard to mess up because this is a feature with high impact on end users.
  - At any point we can make follow-up improvements here or unlock the feature flag for some folks. Improvements to this area seem out of scope.

### What's next?

In short, the outstanding tasks are those additional clean up tasks defined in the issue, along with:

- Defining our true defaults. Currently we default to DEV defaults, which doesn't really make sense
- Improve instructions and UI for admin section and remove the need to hide it behind feature flag
  - In the meantime, our community success team should be able to help admins make safe use of some of this functionality on a case-by-case basis.
 
https://github.com/forem/forem/issues/12161

Ensure the data update script logic makes sense, check that no fields are erroneously showing up on user profiles.

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed


citizen428:
Thanks for jumping in @benhalpern. I'm not sure if you saw my forem.team post, but it will almost certainly take more than one PR to close the loop on this issue. So let's maybe keep them small and focussed so we can tick things off 1 by 1, what do you think?

lib/data_update_scripts/20201223013903_set_proper_display_area_for_profile_fields.rb
```diff
@@ -0,0 +1,10 @@
+module DataUpdateScripts
+  class SetProperDisplayAreaForProfileFields
```
citizen428, line 2: Hm, I thought we had already done something like this in the past and also added display areas to `lib/data/dev_profile_fields.csv`, so if you really want to change it, please change it there too, they should be in sync, at lest for now.

lib/data_update_scripts/20201223013903_set_proper_display_area_for_profile_fields.rb
```diff
@@ -0,0 +1,10 @@
...
+    def run
+      ProfileField.where.not(attribute_name: %w[brand_color1 brand_color2]).update_all(display_area: 0)
```
citizen428, line 4: **suggestion (non-blocking):** Note that Rails enums are smart, so you can use symbols instead of numbers, which will make this DUS easier to read.

```ruby
ProfileField.update_all(display_area: :header)
#  ProfileField Update All (2.8ms)  UPDATE "profile_fields" SET "display_area" = $1  [["display_area", 0]]
# 29
```

lib/data_update_scripts/20201223013903_set_proper_display_area_for_profile_fields.rb
benhalpern: Sounds good. I haven't investigated fully why these weren't set properly but my guess would maybe be enum trying to be set as string somewhere and failed silently?

lib/data_update_scripts/20201223013903_set_proper_display_area_for_profile_fields.rb
benhalpern: 👍  this was sort of my first hack at something smaller, but then realized with just a bit more deep work we could probably get this solved more holistically... So will probably take a different direction.

benhalpern:
Yeah @citizen428 I'll probably leave one or two loops unclosed here. My main goal is to try and get the stuff closest to the user fully out of feature flag mode, including identifying which features can be deemed "special case" in a legitimate way, as opposed to the current state where some stuff is more of an arbitrary hack. And maybe in doing what I plan to do, it will be more than one PR. I'll rename this PR more descriptively once it's at a good place.

I don't plan on touching anything related to dropping columns or last mile DB stuff.

lib/data_update_scripts/20201223013903_set_proper_display_area_for_profile_fields.rb
citizen428: Strings work too:

```ruby
ProfileField.update_all(display_area: 'header')
# ProfileField Update All (4.4ms)  UPDATE "profile_fields" SET "display_area" = $1  [["display_area", 0]]
****

**Comment by citizen428** [comment]
@benhalpern Awesome, this sounds great. Please feel free to ping me anytime something is unclear, you need help, and thanks again for taking on some if this, the extra help is much appreciated. 😃

File: app/models/profile.rb [hunk 3]
```diff
@@ -9,6 +9,34 @@ class Profile < ApplicationRecord
 
   store_attribute :data, :custom_attributes, :json, default: {}
 
+  SPECIAL_FUNCTIONALITY_ATTRIBUTES = %w[
```
citizen428, line 12: Could you maybe elaborate a bit what the use-case for these constants will be? The idea was that once we're done generalizing profiles, new Forems will only get a handful of fields, so most of the ones here won't be available by default.

app/views/users/_metadata.html.erb
```diff
@@ -1,7 +1,8 @@
-<% if FeatureFlag.enabled?(:profile_admin) %>
+<% if true # FeatureFlag.enabled?(:profile_admin) %>
   <% if (header_fields = @profile.ui_attributes_for(area: :header)).present? %>
     <div class="profile-header__bottom fs-base">
       <% header_fields.each do |title, value| %>
+        <% next if Profile.special_attributes.include?(title) %>
```
citizen428, line 5: I think I understand the idea behind this BUT does special for DEV automatically imply special for other Forems?

app/models/profile.rb
benhalpern: My thought here is that these are fields which, if defined, are essentially reserved for special functionality. We don't need to require "email_public", _but_ if it does exist, it will have a certain special function... My thoughts are that we could do further work to make the special fields more general, but I think it's okay if "location" always means "location" and will trigger the icon indicating location. It's pretty global, and the stuff that isn't as global is still special if it exists.

app/views/users/_metadata.html.erb
benhalpern: Yeah, this essentially makes these reserved words for the time being. This doesn't prevent further generalization, but if things like summary, location, etc. are special I think that's okay... Especially brand_color1 etc. That needs to be special.

I think we can clarify this in the interface.

app/models/user.rb
```diff
@@ -292,6 +292,14 @@ def tag_line
     summary
   end
 
+  def twitter_url
+    "https://twitter.com/#{twitter_username}" if twitter_username.present?
+  end
+
+  def github_url
+    "https://twitter.com/#{github_username}" if github_username.present?
```
Ridhwana, line 300:
```suggestion
    "https://github.com/#{github_username}" if github_username.present?
```

app/views/users/_metadata.html.erb
citizen428: I think we're hitting on something interesting here: maybe `brand_color1` and `brand_color2` aren't really profile fields (even though they are set by each user), but some sort of setting instead? Given that we're still working on profile generalization, now would be the perfect time to also introduce other models to move fields to instead of just moving everything from `User` to a `ProfileField`.

benhalpern:
I'll clean this up and get it out of `draft` ASAP to see if I can strip this down and get the useful parts into the codebase.

app/views/users/show.html.erb
```diff
@@ -74,75 +74,10 @@
...
-                    <%= inline_svg_tag("twitch.svg", class: "crayons-icon", aria: true, title: "Twitch logo") %>
-                  </a>
-                <% end %>
-                <% if @user.website_url.present? %>
-                  <a href="<%= @user.website_url %>" target="_blank" rel="noopener nofollow me" class="px-1 align-middle inline-block">
-                    <%= inline_svg_tag("external.svg", class: "crayons-icon", aria: true, title: "External link icon") %>
+              <% Profile.special_social_link_attributes.each do |attribute| %>
+                <% if @user.send(attribute).present? %>
+                  <a href="<%= @user.send(attribute) %>" target="_blank" rel="noopener me" class="px-1 align-middle inline-block">
+                    <%= inline_svg_tag("#{attribute.gsub("_url","")}.svg", class: "crayons-icon", aria: true, title: "Twitter logo") %>
```
aitchiss, line 80: (blocking) - We need to set the correct title for each svg here - this changes them all to be announced to screen reader users as 'Twitter logo' (incidentally, something like 'twitter profile' would be much better naming here)

aitchiss (changes requested):

app/views/users/show.html.erb
benhalpern: Thanks for the catch. Just pushed the fix.

aitchiss (approved):
looks good to me 👍

citizen428 (approved):
I'm approving this now, but I do have some concerns. While this does get rid of `SiteConfig.dev_to?` it's still quite DEV specific. E.g. all profile fields need to have the right names to work as expected.

I don't have a solution for this at this point in time either, but it's something we need to be aware of.

benhalpern:
Yup @citizen428 I think there are outstanding problems to solve. This acts as a bit of a safe rest stop on the way to the top of the mountain.

PR #12027, opened 2020-12-24
Remove swipe gestures and other unnecessary code ✂️

benhalpern:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

We currently have an issue where we're running code on tag pages in a way that is causing a random runtime error. This PR does more than just fix the error, in that we are moving away from this pattern of the sidebars on these kinds of pages in mobile in this way, so I think it's okay to remove the gestures. The buttons are still available by clicking— and I think it's a step in where we're moving.

TLDR: some confusing code causing an error is not really needed anymore. Let's get rid of the functionality.


nickytonline (approved):
The removal of the code doesn't appear to break anything, and the button for the tag sidebar still works.

image

image

Not blocking for this PR, but the only thing I noticed is the button only appears in landscape mode, but it does this in production as well. Don't know if that was by design or not.

ludwiczakpawel (approved):
Works good - I've tested it locally on my phone.
I'm so glad we're dropping it :) Probably for another PR - we could also drop "pull to refresh" feature we have.

PR #12032, opened 2020-12-24
Remove automatic inclusion of Twitter widget.js

benhalpern:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

There are smarter ways we need to decipher when and when not to include a script, but this is a small naive optimization to help us not always include this expensive third party library, as we are currently doing.

### DIFF
--- a/app/views/articles/show.html.erb
+++ b/app/views/articles/show.html.erb
@@ -225,8 +225,8 @@
   <% if has_vid?(@article) %>
     <%= render "articles/fitvids" %>
   <% end %>
-  <% if @article.processed_html.include? 'class="twitter-tweet"' ||
-       @article.processed_html.include? "ltag_twitter_timeline-liquid-tag" %>
+  <% if @article.processed_html.include?('class="twitter-tweet"') ||
+       @article.processed_html.include?("ltag_twitter_timeline-liquid-tag") %>
     <script async src="//platform.twitter.com/widgets.js" charset="utf-8"></script>
   <% end %>
   <% if @article.processed_html.include? "instagram-media" %>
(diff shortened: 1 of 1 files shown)


app/views/articles/show.html.erb
```diff
@@ -225,7 +225,8 @@
   <% if has_vid?(@article) %>
     <%= render "articles/fitvids" %>
   <% end %>
-  <% if @article.processed_html.include? 'class="twitter-tweet"' %>
+  <% if @article.processed_html.include? 'class="twitter-tweet"' ||
```
citizen428, line 228: Hello and thank you for reaching out to the micro-optimization department :wink: 

I was wondering if calling `include?` twice would be relatively slow, but it turned out to be faster than the other things I tried (regex match with regex literal and precompiled regex) over a benchmark with 10000 iterations:

```
Rehearsal --------------------------------------------------------
include?               0.034366   0.000474   0.034840 (  0.035624)
match?                 0.122539   0.001571   0.124110 (  0.131869)
match? (precompiled)   0.104546   0.001282   0.105828 (  0.110414)
----------------------------------------------- total: 0.264778sec

                           user     system      total        real
include?               0.032943   0.000344   0.033287 (  0.034155)
match?                 0.109308   0.001009   0.110317 (  0.113602)
match? (precompiled)   0.103027   0.000669   0.103696 (  0.104777)
```

Not that it matters, but it gave me a reason to read the code for `String#include?` and realizing it's essentially just a wrapper around `String#index`

citizen428 (approved):

rhymes (commented):
LGTM

rhymes (approved):

PR #12034, opened 2020-12-25
Move constant from initializer to app/lib/constants

citizen428:

- [X] Refactor

## Description

`config/initializers/redcarpet.rb` just set up a constant, so I moved it to `app/lib/constants` instead to be more consistent with what we do elsewhere.

n/a

## Added tests?

- [X] No, and this is why: existing tests should cover this

## Added to documentation?

- [X] No documentation needed


jacobherrington (approved):
:+1: :broom:

rhymes (approved):

PR #12035, opened 2020-12-25
Move LanguageDetector to service

citizen428:

- [X] Refactor

## Description

This PR moves the `LanguageDetector` from `app/labor` to the service `Articles::DetectLanguage`. I also updated the specs a bit, as previously each example performed 2 DB updates, one for the title and one for the body markdown. I now use an instance double instead, as the implementation of the article object is irrelevant to the functioning of the service object.

## Added tests?

- [X] Yes, moved/updated existing ones

## Added to documentation?

- [X] No documentation needed


rhymes (approved):
LGTM

citizen428:
I merged `master` back here because I had some weird looking Bootsnap error on Travis when re-running a failed spec.

app/services/articles/detect_language.rb
```diff
@@ -0,0 +1,17 @@
+module Articles
+  class DetectLanguage
+    def self.call(article)
+      text = extract_text(article)
+      response = CLD.detect_language(text)
+      response[:code] if response[:reliable]
+    rescue StandardError
+      nil
```
nickytonline, line 8: Just curious why we do nothing if there is an error. I am probably missing some context.

nickytonline (approved):
LGTM, just left a question.

app/services/articles/detect_language.rb
citizen428: This code just moved, so I'm not 💯 qualified to answer this, but I assume language detection is non-critical, so in case we don't manage to do it, we just move on.

PR #12049, opened 2020-12-28
Update bust_podcast_episode to EdgeCache::BustPodcastEpisode

atsmith813:
- [x] Refactor

## Description
This PR updates references for busting a podcast episode cache from `CacheBuster` to a new, dedicated service `EdgeCache::BustPodcastEpisode`.

My gameplan is to PR for each method so we can test each piece of cache-busting manually in production after we deploy it. Once all method references are updated, I'll remove the legacy `CacheBuster`. We don't want to remove it preemptively either because there could always be jobs in the queue referencing it. So removing it will be its own PR at the very end when code is no longer referencing it.

Can't really QA this outside of production.

**Yes** - once this PR deploys, we need to test cache-busting. We can do this by updating a podcast episode in production. If the podcast episode updates and displays correctly as expected without having to hard refresh, that means it should be working correctly.

## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

goat_slowly_but_surely_gif

### DIFF
--- a/app/workers/podcast_episodes/bust_cache_worker.rb
+++ b/app/workers/podcast_episodes/bust_cache_worker.rb
@@ -1,9 +1,11 @@
 module PodcastEpisodes
   class BustCacheWorker < BustCacheBaseWorker
     def perform(podcast_episode_id, path, podcast_slug)
+      return unless path && podcast_slug
+
       podcast_episode = PodcastEpisode.find_by(id: podcast_episode_id)
 
-      return unless podcast_episode && path && podcast_slug
+      return unless podcast_episode
 
       EdgeCache::BustPodcastEpisode.call(podcast_episode, path, podcast_slug)
     end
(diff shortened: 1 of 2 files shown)


rhymes (approved):

app/workers/podcast_episodes/bust_cache_worker.rb
```diff
@@ -2,9 +2,10 @@ module PodcastEpisodes
   class BustCacheWorker < BustCacheBaseWorker
     def perform(podcast_episode_id, path, podcast_slug)
       podcast_episode = PodcastEpisode.find_by(id: podcast_episode_id)
-      return if podcast_episode.nil?
 
-      CacheBuster.bust_podcast_episode(podcast_episode, path, podcast_slug)
+      return unless podcast_episode && path && podcast_slug
```
vaidehijoshi, line 6: Optimization suggestion: since we don't care about busting a podcast episode unless both `path` and `podcast_slug` are present, could we check for those first (before querying the db for `PodcastEpisode` on [line 4](https://github.com/forem/forem/pull/12049/files#diff-1637a65634c9deae0cf148627c012306f1c54cbc2c37ff48589a3e70da93ebddR4))? Seems like we could avoid an unnecessary query, although admittedly, maybe it's not _that_ much of an optimization 🙃

vaidehijoshi (approved):

app/workers/podcast_episodes/bust_cache_worker.rb
atsmith813: Good idea - updated!

PR #11973, opened 2020-12-18
Remove collective_noun and collective_noun_disabled from the Config

juliannatetreault:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR contains a `data_update` script that removes `collective_noun` and `collective_noun_disabled` from the SiteConfig.

## Related Tickets & Documents
Relates to PR #11846 

## QA Instructions, Screenshots, Recordings
Ensure that all of the checks pass! ♻️ Additionally, you can run the `data_update` script manually to ensure that upon running it, `collective_noun` and `collective_noun_disabled` are deleted.

### UI accessibility concerns?
N/A

## Added tests?

- [ ] Yes
- [x] No, and this is why: my previous PR, #11846 , includes the necessary tests for changes related to the removal of `collective_noun` and `collective_noun_disabled`
- [ ] I need help with writing tests

## Added to documentation?

- [ ] [Developer Docs](https://docs.forem.com) and/or
      [Admin Guide](https://forem.gitbook.io/forem-admin-guide/)
- [ ] README
- [ ] No documentation needed
- [x] I have removed the "Collective Noun" section from the Admin Guide and have added some further details to the "Community Name" section, but have not yet merged the changes. To view these changes and suggest edits, please take a look at my proposed changes: https://app.gitbook.com/@forem/s/forem-admin-guide/~/drafts/-MOr2aA_SCj61eSjxPLG/admin/config/community-content

## [optional] Are there any post-deployment tasks we need to perform?
We will need to ensure that this `data_update` script runs!

## [optional] What gif best describes this PR or how it makes you feel?

Beyonce Single Ladies Waving Bye

PR #12054, opened 2020-12-28
Disable Community Name Data Update Scripts

juliannatetreault:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
When I merged and deployed PR #11846, the `data_upate` script errored out, as shown [here](https://app.honeybadger.io/projects/66984/faults/72660309) on Honeybadger -- it appears that using the predicate method may have caused a `NoMethodError` error: 
```
NoMethodError: undefined method `collective_noun_disabled?' for #<Class:0x00007f5e4e055088>
```
This PR reverts the change from `SiteConfig.collective_noun_disabled?` back to what I initially had, `SiteConfig.collective_noun_disabled`.

Relates to PR #11846 and PR #11973 
Resolves https://app.honeybadger.io/projects/66984/faults/72660309

**To QA this PR, you can do the following:**
- Manually test the `data_update` script
- Ensure all checks pass ♻️ 

N/A

## Added tests?

- [ ] Yes
- [x] No, and this is why: there are tests in PR #11846 around the changes made to `community_name`, `collective_noun`, and `collective_noun_disabled`

## Added to documentation?

- [ ] README
- [x] No documentation needed

This `data_update` script needs to run before the script in PR #11973 (I canceled the build on that PR after the Honeybadger error surfaced and will resume the build once this all passes/gets approved)

Robert Downey Jr. Internal Screamin

### DIFF
--- a/lib/data_update_scripts/20201217184442_append_collective_noun_to_community_name.rb
+++ b/lib/data_update_scripts/20201217184442_append_collective_noun_to_community_name.rb
@@ -1,7 +1,7 @@
 module DataUpdateScripts
   class AppendCommunityToCommunityName
     def run
-      return unless SiteConfig.collective_noun_disabled?
+      return unless SiteConfig.collective_noun_disabled
 
       SiteConfig.where(var: "community_name").find_each do |community_name|
         community_name.update!(value: "#{community_name.value} #{SiteConfig.collective_noun}")
--- a/lib/data_update_scripts/20201217184442_append_collective_noun_to_community_name.rb
+++ b/lib/data_update_scripts/20201217184442_append_collective_noun_to_community_name.rb
@@ -1,5 +1,5 @@
 module DataUpdateScripts
-  class AppendCommunityToCommunityName
+  class AppendCollectiveNounToCommunityName
     def run
       return unless SiteConfig.collective_noun_disabled
 
--- a/lib/data_update_scripts/20201217184442_append_collective_noun_to_community_name.rb
+++ b/lib/data_update_scripts/20201217184442_append_collective_noun_to_community_name.rb
@@ -1,11 +1,13 @@
 module DataUpdateScripts
   class AppendCollectiveNounToCommunityName
     def run
-      return unless SiteConfig.collective_noun_disabled
+      # return unless SiteConfig.collective_noun_disabled
 
-      SiteConfig.where(var: "community_name").find_each do |community_name|
-        community_name.update!(value: "#{community_name.value} #{SiteConfig.collective_noun}")
-      end
+      # SiteConfig.where(var: "community_name").find_each do |community_name|
+      #   community_name.update!(value: "#{community_name.value} #{SiteConfig.collective_noun}")
+      # end
+
+      # We cannot access the column directly, so this has been commented out
     end
   end
 end
--- a/lib/data_update_scripts/20201217184442_append_community_to_community_name.rb
+++ b/lib/data_update_scripts/20201217184442_append_community_to_community_name.rb
@@ -1,5 +1,5 @@
 module DataUpdateScripts
-  class AppendCollectiveNounToCommunityName
+  class AppendCommunityToCommunityName
     def run
       # return unless SiteConfig.collective_noun_disabled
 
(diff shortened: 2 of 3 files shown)


lib/data_update_scripts/20201217184442_append_collective_noun_to_community_name.rb
```diff
@@ -1,7 +1,7 @@
 module DataUpdateScripts
-  class AppendCommunityToCommunityName
+  class AppendCollectiveNounToCommunityName
```
juliannatetreault, line 2: I've updated the name of this script so that it reruns properly.

maestromac (changes requested):
You got the right idea but this isn't going to work. This is from running the command locally on my machine
```
[1] pry(main)> SiteConfig.collective_noun_disabled
  User Exists? (1.6ms)  SELECT 1 AS one FROM "users" LIMIT $1  [["LIMIT", 1]]
NoMethodError: undefined method `collective_noun_disabled' for #<Class:0x0000557eb7ce5dc0>
Did you mean?  collection_cache_versioning
from /home/mac/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/activerecord-6.0.3.4/lib/active_record/dynamic_matchers.rb:22:in `method_missing'
```
There are three things that need to be addressed.

1. You are trying to invoke a method that may or may not exist. You need to check if the method exists first before you call it, or else you will get `NoMethodError`. This involves metaprogramming; see [`respond_to?`](https://apidock.com/ruby/Object/respond_to%3F)
2. This won't rerun in production anymore. The best course of action is for you to comment-out the content within the `#run` method and create another new DataUpdateScript.
3. To make sure that this will indeed work, you should write a spec for it. 
 
You should be using our [`rails generate data_update`](https://docs.forem.com/backend/data-update-scripts/) again and use the spec file that it generated. It can be the same name because the date/time will be different.

vaidehijoshi:
Hm, I can't actually test this out locally because [this line](https://github.com/forem/forem/blob/master/lib/data_update_scripts/20201218173445_remove_collective_noun_from_config.rb#L4) seems to have already run, which has removed the `collective_noun` and `collective_noun` disabled fields already, so SiteConfig is looking for a field that doesn't exist:
```ruby
[6] pry(main)> SiteConfig.collective_noun_disabled
NoMethodError: undefined method `collective_noun_disabled' for #<Class:0x00007f9ac1579280>
```

vaidehijoshi:
> This won't rerun in production anymore. The best course of action is for you to comment-out the content within the #run method and create another new DataUpdateScript.

@maestromac Can you confirm my understanding of how DataUpdateScripts work? Looking at blazer, it looks like [this data update script](https://github.com/forem/forem/blob/master/lib/data_update_scripts/20201218173445_remove_collective_noun_from_config.rb#L4) is on `master`, but hasn't been run yet. We need to ensure that it _doesn't_ ever get run until this is fixed. Does that script need to be commented out to ensure that? (I would assume it does, but want to confirm)

mstruve:
Can confirm that this DataUpdateScript did run and it failed which is why we got the Honeybadger for it. 
Screen Shot 2020-12-28 at 1 40 18 PM

Changing the name of the script will ensure that the new script is run. If the file name was the same then the script would NOT be run. The reason is that we will only put a new script row in our database for a new filename. If the filename is the same it will simply update the existing record but since that script will already have been marked as run it will not run.
```ruby
    def insert_new_scripts
      now = Time.current
      scripts_params = filenames.map do |fn|
        { file_name: fn, created_at: now, updated_at: now }
      end

      DataUpdateScript.insert_all(scripts_params)
    end
```

To ensure our records match our code, what I would do is comment out the body of the old script. 
```ruby
module DataUpdateScripts
  class AppendCommunityToCommunityName
   # Code removed due to bug
  end
end
```
Then make an entirely new data update script. This way when someone compares the code to what is in the database they can see that `20201217184442_append_community_to_community_name` is not some ghost script that they have no idea what it did. 

In the future, we may decide to remove these files but for now we like the idea of having a history and something to trace back to. Let me know if that makes sense!

vaidehijoshi (approved):

maestromac (approved):
LGTM!

PR #12055, opened 2020-12-28
Adds collective_noun Fields Back to SiteConfig Model

juliannatetreault:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR adds the `collective_noun` and `collective_noun_disabled` columns back to the `SiteConfig` model after running into a `data_update` script failure in another PR. This is necessary so that when running the `data_update` script responsible for updating the `community_name`, we have access to the `collective_noun` and `collective_noun_disabled` methods. Additionally, this PR adds a `data_update` script that conditionally updates `SiteConfig.community_name` _if_ `SiteConfig.collective_noun_disabled` returns `false`. If, however, the forem has `collective_noun_disabled` set to `true`, then the `community_name` will remain as-is and untouched.

Relates to PRs #11846 , PR #11973 , and PR #12054 

**To QA these changes, first ensure that all checks pass and the Travis build is green, then**:
- Run the `data_update` script manually and ensure that your Community's name updates accordingly

N/A

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

The `data_update` script included in this PR needs to successfully run!

The Office: Days of Nonsense

### DIFF
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -55,6 +55,8 @@ class SiteConfig < RailsSettings::Base
   # Community Content
   field :community_name, type: :string, default: ApplicationConfig["COMMUNITY_NAME"] || "New Forem"
   field :community_emoji, type: :string, default: "🌱"
+  # collective_noun and collective_noun_disabled have been added back temporarily for
+  # a data_update script, but will be removed in a future PR!
   field :collective_noun, type: :string, default: "Community"
   field :collective_noun_disabled, type: :boolean, default: false
   field :community_description, type: :string
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -55,6 +55,8 @@ class SiteConfig < RailsSettings::Base
   # Community Content
   field :community_name, type: :string, default: ApplicationConfig["COMMUNITY_NAME"] || "New Forem"
   field :community_emoji, type: :string, default: "🌱"
+  field :collective_noun, type: :string, default: "Community"
+  field :collective_noun_disabled, type: :boolean, default: false
   field :community_description, type: :string
   field :community_member_label, type: :string, default: "user"
   field :tagline, type: :string
--- a/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
+++ b/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
@@ -1,0 +1,7 @@
+module DataUpdateScripts
+  class AppendCollectiveNounToCommunityName
+    def run
+      return "#{SiteConfig.community_name} #{SiteConfig.collective_noun}" unless SiteConfig.collective_noun_disabled?
+    end
+  end
+end
--- a/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
+++ b/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
@@ -1,11 +1,9 @@
 module DataUpdateScripts
   class AppendCollectiveNounToCommunityName
     def run
-      if SiteConfig.collective_noun_disabled
-        SiteConfig.community_name
-      else
-        SiteConfig.community_name = "#{SiteConfig.community_name} #{SiteConfig.collective_noun}"
-      end
+      return if SiteConfig.collective_noun_disabled
+
+      SiteConfig.community_name = "#{SiteConfig.community_name} #{SiteConfig.collective_noun}"
     end
   end
 end
--- a/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
+++ b/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
@@ -1,7 +1,11 @@
 module DataUpdateScripts
   class AppendCollectiveNounToCommunityName
     def run
-      return "#{SiteConfig.community_name} #{SiteConfig.collective_noun}" unless SiteConfig.collective_noun_disabled?
+      if SiteConfig.collective_noun_disabled
+        SiteConfig.community_name
+      else
+        SiteConfig.community_name = "#{SiteConfig.community_name} #{SiteConfig.collective_noun}"
+      end
     end
   end
 end
--- a/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
+++ b/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
@@ -1,7 +1,7 @@
 module DataUpdateScripts
   class AppendCollectiveNounToCommunityName
     def run
-      return if SiteConfig.collective_noun_disabled
+      return if SiteConfig.collective_noun_disabled || SiteConfig.collective_noun.blank?
 
       SiteConfig.community_name = "#{SiteConfig.community_name} #{SiteConfig.collective_noun}"
     end
--- a/spec/lib/data_update_scripts/append_collective_noun_to_community_name_spec.rb
+++ b/spec/lib/data_update_scripts/append_collective_noun_to_community_name_spec.rb
@@ -6,7 +6,7 @@ require Rails.root.join(
 describe DataUpdateScripts::AppendCollectiveNounToCommunityName do
   context "when collective_noun_disabled is false" do
     it "appends the collective_noun to the community_name" do
-      SiteConfig.collective_noun = "Club"
+      allow(SiteConfig).to receive(:collective_noun).and_return("Club")
       described_class.new.run
       expect(SiteConfig.collective_noun_disabled).to eq(false)
       expect(SiteConfig.community_name).to eq("DEV(local) Club")
@@ -15,9 +15,8 @@ describe DataUpdateScripts::AppendCollectiveNounToCommunityName do
 
   context "when collective_noun_disabled is true" do
     it "does not append the collective_noun to the community_name" do
-      SiteConfig.collective_noun_disabled = true
+      allow(SiteConfig).to receive(:collective_noun_disabled).and_return(true)
       described_class.new.run
-      expect(SiteConfig.collective_noun_disabled).to eq(true)
       expect(SiteConfig.community_name).to eq("DEV(local)")
     end
   end
--- a/spec/lib/data_update_scripts/append_collective_noun_to_community_name_spec.rb
+++ b/spec/lib/data_update_scripts/append_collective_noun_to_community_name_spec.rb
@@ -1,0 +1,23 @@
+require "rails_helper"
+require Rails.root.join(
+  "lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb",
+)
+
+describe DataUpdateScripts::AppendCollectiveNounToCommunityName do
+  context "when collective_noun_disabled is false" do
+    it "appends the collective_noun to the community_name" do
+      described_class.new.run
+      expect(SiteConfig.collective_noun_disabled).to eq(false)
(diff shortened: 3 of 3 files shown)


spec/lib/data_update_scripts/append_collective_noun_to_community_name_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+require Rails.root.join(
+  "lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb",
+)
+
+describe DataUpdateScripts::AppendCollectiveNounToCommunityName do
+  context "when collective_noun_disabled is false" do
+    it "appends the collective_noun to the community_name" do
+      described_class.new.run
+      expect(SiteConfig.collective_noun_disabled).to eq(false)
+      expect(SiteConfig.community_name).to eq("DEV(local) Community")
```
juliannatetreault, line 11: This continues to return `"DEV(local)"` rather than `"DEV(local) Community"` and after dropping into `pry` and trying a bunch of different things, I still cannot get this to work properly. It almost feels like I need a `reload` here, but I'm unable to use that on a `string` -- thoughts? Am I missing something? 😅 

cc @vaidehijoshi and @maestromac

lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
```diff
@@ -0,0 +1,7 @@
+module DataUpdateScripts
+  class AppendCollectiveNounToCommunityName
+    def run
+      return "#{SiteConfig.community_name} #{SiteConfig.collective_noun}" unless SiteConfig.collective_noun_disabled?
```
vaidehijoshi, line 4: @juliannatetreault This doesn't actually _update_ the SiteConfig at all, it just returns a string. I think that's probably why the test isn't doing what you expect/want it to -- you need actually update the `SiteConfig.community_name` similar to what you [previously did](https://github.com/forem/forem/pull/11846/files#diff-620cde66dafdd0fed658093095e053565f368bee979f03ba83bc2eb9a2fe9dabR7) in your initial PR

spec/lib/data_update_scripts/append_collective_noun_to_community_name_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  "lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb",
+)
+
+describe DataUpdateScripts::AppendCollectiveNounToCommunityName do
+  context "when collective_noun_disabled is false" do
+    it "appends the collective_noun to the community_name" do
+      SiteConfig.collective_noun = "Club"
+      described_class.new.run
+      expect(SiteConfig.collective_noun_disabled).to eq(false)
+      expect(SiteConfig.community_name).to eq("DEV(local) Club")
```
juliannatetreault, line 12: I could also change this to mirror the code snippet below if folks think it's more readable and/or prefer it written this way:
```
      expect(SiteConfig.collective_noun_disabled).to eq(false)
      expect do
        described_class.new.run
      end.to change(SiteConfig, :community_name).from("DEV(local)").to("DEV(local) Club")
```
(The same goes for the test below!)

lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
juliannatetreault: I was totally overthinking this yesterday and looking at it with fresh eyes this morning made me 🤦‍♀️  --  thank you for the feedback!

spec/lib/data_update_scripts/append_collective_noun_to_community_name_spec.rb
vaidehijoshi: I like this a little better, and I think I've mostly seen this as a convention in our codebase!

spec/lib/data_update_scripts/append_collective_noun_to_community_name_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "rails_helper"
+require Rails.root.join(
+  "lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb",
+)
+
+describe DataUpdateScripts::AppendCollectiveNounToCommunityName do
+  context "when collective_noun_disabled is false" do
+    it "appends the collective_noun to the community_name" do
+      SiteConfig.collective_noun = "Club"
```
vaidehijoshi, line 9: Rather than explicitly assigning a value to a field on `SiteConfig` here, I think we should opt to use Rspec's [method stubs](https://relishapp.com/rspec/rspec-mocks/v/2-99/docs/method-stubs) in this test and the one below, too.

By using a stub here, we can be sure that we're not accidentally ever changing the value of `SiteConfig.collective_noun`, which could leak into other tests or cause strange, confusing bugs in tests down the road.

In this test, that would look something like:
```ruby
allow(SiteConfig).to receive(:collective_noun).and_return("Club")
```

juliannatetreault: Ahh, that's great! I could also just lean into the default `collective_noun`, which is "Community"--I set `collective_noun` to "Club" here just to make the test more explicit for folks who look at this code in the future. Thoughts @vaidehijoshi ?

vaidehijoshi: That would work!

But, there's one thing to keep in mind here, which can a bit obtuse until you think about how this test might be interpreted in the future. There is one _benefit_ of being explicit in the test, like the way you currently have it. The benefit is that if the (default) value you are testing against changes down the road (aka someone goes and changes [this line](https://github.com/forem/forem/pull/12055/files#diff-68da686a0f7348bf9a2c59e669043f21019ab8dfb6363cea6daa534c5c8296d7R58)), then this test would start failing. Sometimes, you _want_ your test to fail in that way, and relying on the default makes sense.

But this test isn't really testing the default value of `community_club`; it's testing the behavior of the _data update script_. So, if you were _not_ explicit about the value of this field in this test and if this fails down the road, it could be confusing (at first) to the person making the change and they'd have to figure out that the test was failing because it was _implicitly_ depending upon + testing the default value of the SiteConfig's field. By being explicit here, you are actually avoiding that scenario!

TL;DR, I would vote to be explicit, just as you have it set up already 😉

juliannatetreault: Thank you so much for the thorough explanation! 😄 

I just pushed up changes that incorporated method stubs in the spec before seeing this comment @vaidehijoshi . 😂 I think the stubs sound like a safer way forward, even if they're not as explicit. That being said, would it be helpful if I added a comment to the spec now that I've changed it to give future contributors/us a bit more context?

vaidehijoshi: > I think the stubs sound like a safer way forward, even if they're not as explicit.

@juliannatetreault I think maybe there is a little confusion here: the stubs don't actually have anything to do with the explicitness that I was talking about in my comment. What makes this test explicit is when you declare that `collective_noun` is set to return a value of the string `"Club"`:
```ruby
# explicitly setting the collective_noun by using a method stub
allow(SiteConfig).to receive(:collective_noun).and_return("Club")
```

To be clear, what you had in your initial commit was _also explicit_, even though it didn't use a stub:
```ruby
# explicitly setting the collective_noun by setting the value directly, which is less safe
SiteConfig.collective_noun = "Club"
```
Both of these are explicit because you are explicitly setting the value of `collective_noun` (either with a method stub, or directly) in the body of the test. When someone reads either of those two variations above,  

The _implicitness_ I was talking about would have only applied if you had relied on the default value, like you were describing in your earlier comment:
> Alternatively, I could also just lean into the default value for collective_noun, which is "Community"

So, if you had written your test like this:
```ruby
described_class.new.run
expect(SiteConfig.collective_noun_disabled).to eq(false)
expect(SiteConfig.community_name).to eq("DEV(local) Community")
```

then _that_ would have been implicit, because you would have been implicitly relying on the fact that the default value of `collective_noun` is `"Community"`. 

Hopefully that explanation makes more sense? :) Let me know if any of it is still confusing. Given all of that, I don't think you need to add a comment here because the test is being explicit while using a stub 👍🏽

juliannatetreault: OH! I misread your comment, @vaidehijoshi -- apologies. 😊 I'll leave everything as-is then.

app/models/site_config.rb
```diff
@@ -55,6 +55,8 @@ class SiteConfig < RailsSettings::Base
   # Community Content
   field :community_name, type: :string, default: ApplicationConfig["COMMUNITY_NAME"] || "New Forem"
   field :community_emoji, type: :string, default: "🌱"
+  field :collective_noun, type: :string, default: "Community"
+  field :collective_noun_disabled, type: :boolean, default: false
```
vaidehijoshi, line 59: We might want to add a comment next to these 2 fields explaining that we don't use these across the app elsewhere right now and that they are only here temporarily/will be removed when this issue is fully resolved

vaidehijoshi (approved):
Works great locally, and thank you for adding a test! 🚀

app/models/site_config.rb
juliannatetreault: Great call! Done ✔️

lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
```diff
@@ -0,0 +1,9 @@
+module DataUpdateScripts
+  class AppendCollectiveNounToCommunityName
+    def run
+      return if SiteConfig.collective_noun_disabled
+
+      SiteConfig.community_name = "#{SiteConfig.community_name} #{SiteConfig.collective_noun}"
```
maestromac, line 6: Is there a possibility for collective_noun to be nil?

maestromac (approved):
LGTM!

lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
juliannatetreault: Great question! I just tested this out in the console and it seems like there is a possibility that `collective_noun` could be set to `nil`, but I don't know if this is actually the case for any communities. That being said, I could adjust line 4 to read:
```
return if SiteConfig.collective_noun_disabled || SiteConfig.collective_noun.nil?
```
Or instead, maybe I could add a `nil` guard to line 6 like so:
```
#{SiteConfig&.collective_noun}"
```

Thoughts @maestromac ?

maestromac: That works! you should use `!SiteConfig.collective_noun.present?` instead though since that will account for empty string

juliannatetreault: @maestromac What if I changed it to the code snippet below to account for empty strings, `nil` values, and empty arrays/hashes?
```
SiteConfig.collective_noun.blank?
```

maestromac: great idea!

juliannatetreault: Done! I'll merge/deploy pending a green and passing build. 😄

PR #12069, opened 2020-12-29
Remove Collective Noun-Related Fields from SiteConfig

juliannatetreault:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR does the following:
- Adds a `data_update` script that destroys `collective_noun` and `collective_noun_disabled`
- Adds a spec for the `data_update` script that destroys `collective_noun` and `collective_noun_disabled`

**Heads Up**: This PR is on pause until its previous, related PRs are deployed to the Forem fleet! I will open this PR up for review once all Forems have been updated.

Relates to PR #11846 and PR #12055 

To QA this PR, please ensure that all of the checks pass and that Travis builds green. Additionally, you can run this `data_update` script manually to test that it works as expected.

N/A

## Added tests?

- [x] Yes
- [ ] No, and this is why

## Added to documentation?

- [ ] README
- [ ] No documentation needed
- [x] Removed references to "Collective Noun" and "Collective Noun Disabled" from the Admin guides in a previous PR

The `data_update` script in this PR needs to successfully run! ♻️ 

Kevin Talking With A Few Words

### DIFF
--- a/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
+++ b/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
@@ -1,7 +1,6 @@
 module DataUpdateScripts
   class AppendCollectiveNounToCommunityName
     def run
-      return unless SiteConfig.respond_to?(:collective_noun_disabled)
       return if SiteConfig.collective_noun_disabled || SiteConfig.collective_noun.blank?
 
       SiteConfig.community_name = "#{SiteConfig.community_name} #{SiteConfig.collective_noun}"
--- a/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
+++ b/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
@@ -2,7 +2,7 @@ module DataUpdateScripts
   class AppendCollectiveNounToCommunityName
     def run
       return unless SiteConfig.respond_to?(:collective_noun_disabled)
-      return unless SiteConfig.collective_noun_disabled? || SiteConfig.collective_noun.blank?
+      return if SiteConfig.collective_noun_disabled || SiteConfig.collective_noun.blank?
 
       SiteConfig.community_name = "#{SiteConfig.community_name} #{SiteConfig.collective_noun}"
     end
--- a/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
+++ b/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
@@ -1,7 +1,8 @@
 module DataUpdateScripts
   class AppendCollectiveNounToCommunityName
     def run
-      return if SiteConfig.collective_noun_disabled || SiteConfig.collective_noun.blank?
+      return unless SiteConfig.respond_to?(:collective_noun_disabled)
+      return unless SiteConfig.collective_noun_disabled? || SiteConfig.collective_noun.blank?
 
       SiteConfig.community_name = "#{SiteConfig.community_name} #{SiteConfig.collective_noun}"
     end
--- a/lib/data_update_scripts/20201229230456_remove_collective_noun_from_site_config.rb
+++ b/lib/data_update_scripts/20201229230456_remove_collective_noun_from_site_config.rb
@@ -1,0 +1,7 @@
+module DataUpdateScripts
+  class RemoveCollectiveNounFromSiteConfig
+    def run
+      SiteConfig.where(var: %w[collective_noun collective_noun_disabled]).destroy_all
+    end
+  end
+end
--- a/lib/data_update_scripts/20201229230456_remove_collective_noun_from_site_config.rb
+++ b/lib/data_update_scripts/20201229230456_remove_collective_noun_from_site_config.rb
@@ -1,7 +1,7 @@
 module DataUpdateScripts
   class RemoveCollectiveNounFromSiteConfig
     def run
-      SiteConfig.where(var: %w[collective_noun collective_noun_disabled]).destroy_all
+      SiteConfig.where(var: %w[collective_noun collective_noun_disabled])&.destroy_all
     end
   end
 end
--- a/lib/data_update_scripts/20201229230456_remove_collective_noun_from_site_config.rb
+++ b/lib/data_update_scripts/20201229230456_remove_collective_noun_from_site_config.rb
@@ -1,7 +1,7 @@
 module DataUpdateScripts
   class RemoveCollectiveNounFromSiteConfig
     def run
-      SiteConfig.where(var: %w[collective_noun collective_noun_disabled])&.destroy_all
+      SiteConfig.where(var: %w[collective_noun collective_noun_disabled]).destroy_all
     end
   end
 end
--- a/spec/lib/data_update_scripts/remove_collective_noun_from_site_config_spec.rb
+++ b/spec/lib/data_update_scripts/remove_collective_noun_from_site_config_spec.rb
@@ -5,12 +5,14 @@ require Rails.root.join(
 
 describe DataUpdateScripts::RemoveCollectiveNounFromSiteConfig do
   context "when collective_noun and collective_noun_disabled are removed from the Config" do
-    it "detroys the fields" do
-      allow(SiteConfig).to receive(:collective_noun).and_return("Club")
-      expect(SiteConfig.collective_noun).to eq("Club")
-      described_class.new.run
-      expect(SiteConfig).to include(:community_name)
-      expect(SiteConfig).not_to include(var: %w[collective_noun collective_noun_disabled])
+    it "does not alter the community_name" do
+      allow(SiteConfig).to receive(:community_name).and_return("DEV Club")
+
+      expect do
+        described_class.new.run
+      end.not_to change(SiteConfig, :community_name)
+
+      expect(SiteConfig.community_name).to eq("DEV Club")
     end
   end
 end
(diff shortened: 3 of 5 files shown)


lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
```diff
@@ -1,6 +1,7 @@
 module DataUpdateScripts
   class AppendCollectiveNounToCommunityName
     def run
+      return unless SiteConfig.respond_to?(:collective_noun_disabled)
```
juliannatetreault, line 4: I added this guard clause in the hopes that it would help with the [failing spec](https://travis-ci.com/github/forem/forem/jobs/468925576#L2234) related to this `data_update` script, but it remains red and failing. 😕  At this point, I'm unsure of how to proceed with fixing it. Does anyone have any ideas? Am I missing something or overthinking this? 😅 

cc @vaidehijoshi @maestromac

spec/lib/data_update_scripts/remove_collective_noun_from_site_config_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "rails_helper"
+require Rails.root.join(
+  "lib/data_update_scripts/20201229230456_remove_collective_noun_from_site_config.rb",
+)
+
+describe DataUpdateScripts::RemoveCollectiveNounFromSiteConfig do
+  context "when collective_noun and collective_noun_disabled are removed from the Config" do
+    it "does not alter the community_name" do
```
juliannatetreault, line 8: Since I've removed the `collective_noun` and `collective_noun_disabled` fields, I thought that testing that after the removal, the community name remains untouched might be worthwhile. Thoughts!?

lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
vaidehijoshi: The first thing I want to point out is that data update scripts are like migrations -- you probably shouldn't be editing their content. In rare scenarios, when you really need to, you can comment them out if you don't want the logic of those scripts to be run, but in this case, you are editing [a preexisting script](https://github.com/forem/forem/blob/master/lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb), which is a sign that something is fishy 🎣 

Now, the next question is...why did you end up having to put in the guard clause here in the first place? Well, to answer that, we need to think back to 8 days ago when we first dug into this issue....

We'll recall that `SiteConfig` is a model, and the `field`s are how you set the keys on that model (that's the way that the [`rails-settings-cached`](https://github.com/huacnlee/rails-settings-cached) gem is set up to work using metaprogramming under the hood). The short story (without getting too into the weeds about how this gem works) is that each field on the object will define a method on the object. When you call the method on the object, you get the value from it (much like a PORO, plain old ruby object).

So, when I add this field to the site config:
```ruby
class SiteConfig < RailsSettings::Base
  field :owns_llamas, type: :boolean, default: true
end
```

I am able to do this:
```ruby
SiteConfig.owns_llamas
```
You can see more examples of this in [a portion of the gem's README](https://github.com/huacnlee/rails-settings-cached#get-defined-fields), too.

Now, the reason I can do this is because the gem is defining a method under the hood! (It's pretty cool, [you can check out how it does that here](https://github.com/huacnlee/rails-settings-cached/blob/master/lib/rails-settings/base.rb#L63-L92), but don't worry too much about it if it feels overwhelming).

All of this is to say, **you won't be able to successfully call the `owns_llamas` method on the `SiteConfig` unless you have _defined_ the `owns_llamas` `field` on the `SiteConfig` class.**

This PR's failing test fail with these error messages (note the file where these failures are coming from...they are _not_ coming from a diff in this PR):
```ruby
SiteConfig(id: integer, created_at: datetime, updated_at: datetime, value: text, var: string) does not implement: collective_noun

SiteConfig(id: integer, created_at: datetime, updated_at: datetime, value: text, var: string) does not implement: collective_noun_disabled
```

The error message is trying to tell you that the `SiteConfig` class doesn't implement the method(s) you are trying to call on it. But wait! Where did you actually try to call these methods? 🤔 

The answer is a bit hidden: you call these methods in a totally different test file entirely, which you added in [a previous PR](https://github.com/forem/forem/pull/12055). You are calling them in the form of rspec mocks 😉 

https://github.com/forem/forem/blob/3bf4f33510f4d72978bf71d6f94c3b8e9fb716cf/spec/lib/data_update_scripts/append_collective_noun_to_community_name_spec.rb#L9
https://github.com/forem/forem/blob/3bf4f33510f4d72978bf71d6f94c3b8e9fb716cf/spec/lib/data_update_scripts/append_collective_noun_to_community_name_spec.rb#L18

The spec for the _previous_ data update script is where you call these methods on the `SiteConfig` class. In other words, you are telling Rspec that _it_ should try to call a method called `collective_noun`, but the `SiteConfig` class mock...but the class doesn't have that method defined. The same goes for the `collective_noun_disabled` -- you are trying to call that method when you mock the `SiteConfig`, but it isn't defined. 

All of this is to say:
* The 2 failing specs on this PR are failing for the same reason that [this error popped up on Honeybadger](https://app.honeybadger.io/projects/66984/faults/72660309) the other day :)
* The `append_collective_noun_to_community_name_spec.rb` [spec file](https://github.com/forem/forem/blob/fb3fb80d758142acecb6c077af95e8eb4b949c5b/spec/lib/data_update_scripts/append_collective_noun_to_community_name_spec.rb) is going to fail if you remove the `collective_noun` and `collective_noun_disabled` fields [in this PR](https://github.com/forem/forem/pull/12069/files#diff-68da686a0f7348bf9a2c59e669043f21019ab8dfb6363cea6daa534c5c8296d7L60-L61)
* In fact, I am not even sure that [this data update script](https://github.com/forem/forem/pull/12069/files#diff-2e4943d4028b711400fe9cfedc056fcbf87473d23c476128a3dabbf55e44c841R4) that you added will even work as expected if you remove the `collective_noun` and `collective_noun_disabled` fields in this PR

Going forward, I think you probably want to:
* Remove the guard clause you [added here](https://github.com/forem/forem/pull/12069/files#diff-721ea8309f852633bfe0043f8894c61e1d2096b90d2c2ba7c7241d552bc9dae8R4) and not touch the old data update script
* Keep the the `collective_noun` and `collective_noun_disabled` fields around, and don't remove them in this PR
* Keep the scope of this PR focused only on adding the new data update script + its test.

After that, and once this change has been deployed, we'll need to decide if we want to remove those fields (if we do, we'll need to completely comment out the old data update scripts and delete their specs, which is what is causing the issue).

Tagging @maestromac to see what he thinks about my above ^^ game plan.

vaidehijoshi: Oh, I also meant to add one other thing: if you add the `collective_noun` and `collective_noun_disabled` fields back in, the two failings tests will start passing again! 😄

juliannatetreault: Thank you for the thorough explanation! This all makes sense and I figured this might be where the conversation would lead but figured I'd ask anyway. 😊 

I added the guard clause here after speaking with another engineer yesterday and thought it might do the trick without having to add the fields back to the SiteConfig--it feels weird removing the fields with a `data_update` script without actually removing the fields from the SiteConfig model. That being said, I _am_ curious about @maestromac's thought on adding the guard clause to the script, since I was told that adding the guard clause like this, on a script that will not run again, would be fine for spec-failing purposes and I would like 10000% clarification for future scripts!

Overall, I think it makes sense for me to remove the guard clause and add back the fields (which I already have on my local branch) and make sure the spec passes like you suggested, @vaidehijoshi. 😄

spec/lib/data_update_scripts/remove_collective_noun_from_site_config_spec.rb
juliannatetreault: Ignore the comment above, as this will likely change once I add the fields back in and revert some of the changes I've pushed up in my latest commit!

lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
juliannatetreault: Update: Changes to the script and SiteConfig have been reverted and the `append_collective_noun_to_community_name_spec` is passing as expected!

lib/data_update_scripts/20201229230456_remove_collective_noun_from_site_config.rb
```diff
@@ -0,0 +1,7 @@
+module DataUpdateScripts
+  class RemoveCollectiveNounFromSiteConfig
+    def run
+      SiteConfig.where(var: %w[collective_noun collective_noun_disabled])&.destroy_all
```
rhymes, line 4: a small tip: & in this context is not needed. If no row is returned, then Rails will just not delete anything. There's no scenario where .where() can return nil. It will return an empty collection (precisely an "empty active record relation") if no row is found and thus .destroy_all will be a no-op executing nothing

juliannatetreault: Thank you for sharing, @rhymes ! I've reverted this change. 😊

rhymes (approved):

benhalpern (approved):

vaidehijoshi (approved):

lib/data_update_scripts/20201228194641_append_collective_noun_to_community_name.rb
maestromac: Sound well and fine to me @vaidehijoshi !

maestromac (approved):

PR #12077, opened 2020-12-30
Optimization: Convert tags/_main_feed and widget_list_item partial loads to collections

mstruve:
- [x] Optimization

## Description
Rather than iterating over articles and creating each partial individually, [Rails allows us to load them as a collection](https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-collections) which can [speed up load time](https://medium.com/@coorasse/partial-rendering-performance-in-rails-101fdfb6ffb9). We do this in quite a few places in the code. This just addresses 2 places where I noticed it initially. 

https://github.com/forem/InternalProjectPlanning/issues/273

* Load tag pages and note that you will see the following log instead of 10 separate partial loads
```
Rendered collection of articles/_widget_list_item.html.erb [10 times]
```

## Added tests?
- [x] No, and this is an implementation change, as long as all current tests pass we should be good.

alt_text

### DIFF
--- a/app/views/articles/tags/_main_feed.html.erb
+++ b/app/views/articles/tags/_main_feed.html.erb
@@ -1,6 +1,4 @@
-<% @stories.each_with_index do |story, i| %>
-  <%= render "articles/single_story", story: story, featured: false %>
-<% end %>
+<%= render partial: "articles/single_story", collection: @stories, as: :story, locals: { featured: false } %>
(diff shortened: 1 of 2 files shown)


rhymes (approved):
Great find!

joshpuetz (approved):
Fantastic find! 👍🏻

PR #12081, opened 2020-12-30
Move EmojiConverter to service

atsmith813:
- [x] Refactor

## Description
This PR moves and updates the `EmojiConverter` from `/labor` into `/services`.

## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

dog_driving_gif


maestromac (approved):

app/services/emoji_converter.rb
```diff
@@ -1,15 +1,7 @@
 class EmojiConverter
```
rhymes, line 1: I think we should change the name of the service to an action, see https://docs.dev.to/backend/service-objects/

Something like `Html::ParseEmoji`, this way we can also move `CodeBlockParser` to something lik `Html::ParseCodeBlock` 

What do you think?

rhymes (changes requested):
Left a request change for naming

app/services/emoji_converter.rb
atsmith813: Great idea - updated!

vaidehijoshi (approved):

rhymes (approved):
LGTM!

Thanks for the change @atsmith813

PR #12082, opened 2020-12-30
Bug Fix:Update dev.to check to use DEV Community

mstruve:
Recently we changed our site name on dev.to from DEV to DEV Community. This is used in our code in a couple of places and needs to be updated. From a prod console
```
irb(main):002:0> SiteConfig.community_name == "DEV Community"
=> true
```

### DIFF
--- a/app/services/email_digest.rb
+++ b/app/services/email_digest.rb
@@ -13,7 +13,7 @@ class EmailDigest
         # Temporary
         # @sre:mstruve This is temporary until we have an efficient way to handle this job
         # for our large DEV community. Smaller Forems should be able to handle it no problem
-        if SiteConfig.community_name == "DEV"
+        if SiteConfig.community_name == "DEV Community"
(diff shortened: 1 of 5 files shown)


jdoss (approved):

msarit (approved):

atsmith813 (approved):

maestromac (approved):

vaidehijoshi:
I wonder if we could use the dev_to method here instead of relying on the community name. I had no idea that we were checking against the hardcoded string!

https://github.com/forem/forem/blob/6b0626a5e2ac8748c71f97acf2cb053101aee860/app/models/site_config.rb#L215

maestromac:
Yeah @vaidehijoshi I was just thinking the same thing. I think the domain name is less prone to change

mstruve:
100% game for that. I will make a follow-up PR!

mstruve:
https://github.com/forem/forem/pull/12083

PR #12083, opened 2020-12-30
Refactor:Use dev_to? to Check for DEV

mstruve:
- [x] Refactor

## Description
Rather than checking the community name which we just found out can change easily, lets check the domain like we do in other places since that is far less likely to change. 

https://github.com/forem/forem/pull/12082

alt_text


vaidehijoshi (approved):

maestromac (approved):
:100:

PR #12089, opened 2020-12-31
Add format to image optimizer

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR applies the format of webp to cover images on posts. imgproxy does not have "auto" format, which we use on Cloudinary, but we should be using webp wherever possible. Only about 0.75% of all DEV traffic is from browsers that are not compatible with webp, and this is dropping quickly as more people move to Safari 14. And this number is much lower on authenticated traffic.

I think there is still too much traffic from these browsers to entirely drop support, but I think removing support for old Safari browsers _only for cover images_ is okay. This one image won't render properly for a very small number of users, but the gains are that the rest of the traffic will benefit from the far smaller webp images in their feed.

In a matter of months we may be able to consider applying the format of webp more broadly.

Apple ecosystem tends to have really healthy update cycles and I think we can start stepping into the webp universe.

Screen Shot 2020-12-31 at 6 48 48 PM

**Edit:** It looks like we _can_ auto-detect webp support directly within imgproxy. 

https://docs.imgproxy.net/#/configuration?id=webp-support-detection

We should probably try this and use webp everywhere I'd think.

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/services/images/optimizer.rb
+++ b/app/services/images/optimizer.rb
@@ -34,6 +34,8 @@ module Images
     DEFAULT_IMGPROXY_OPTIONS = {
       height: nil,
       width: nil,
+      format: nil,
+      max_bytes: 500_000, # Keep everything under half of one MB.
       resizing_type: nil
     }.freeze
 
@@ -48,7 +50,7 @@ module Images
       if options[:crop] == "fill"
         options[:resizing_type] = "fill"
       end
-
+      options[:format] = options[:fetch_format] == "auto" ? nil : options[:fetch_format]
       options
     end
 
--- a/app/view_objects/cloud_cover_url.rb
+++ b/app/view_objects/cloud_cover_url.rb
@@ -12,7 +12,7 @@ class CloudCoverUrl
     width = 1000
     img_src = url_without_prefix_nesting(url, width)
 
-    Images::Optimizer.call(img_src, width: width, height: 420, crop: "imagga_scale")
+    Images::Optimizer.call(img_src, width: width, height: 420, crop: "imagga_scale", fetch_format: "webp")
   end
 
   private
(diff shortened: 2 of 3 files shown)


rhymes (commented):
Do we need to change the logic at all? 

Cloudinary will use webp automatically if detected, correct?

> The fetch_format parameter can be set to auto (f_auto in URLs) in order to perform automatic format selection based on the requesting browser. For example, with the automatic format feature, in most cases Chrome users would receive a WebP (.webp), Internet Explorer users would receive a JPEG-XR (.jxr), and Safari users would receive a JPEG-2000 (.jp2). If a browser does not support any of these formats then the image is delivered in the format specified by the file extension. 

see https://cloudinary.com/documentation/image_transformations#automatic_format_selection_f_auto

Imgproxy does use webp if it's enabled (which is disabled right now):

> IMGPROXY_ENABLE_WEBP_DETECTION: enables WebP support detection. When the file extension is omitted in the imgproxy URL and browser supports WebP, imgproxy will use it as the resulting format

from https://docs.imgproxy.net/#/configuration?id=webp-support-detection

and 

> The extension part can be omitted. In this case, imgproxy will use source image format as resulting one. If source image format is not supported as resulting, imgproxy will use jpg. You also can enable WebP support detection to use it as default resulting format when possible.

from https://docs.imgproxy.net/#/generating_the_url_basic?id=extension

Basically, if we enable WebP via the environment variable and don't specify an exstension, it will detect WebP support and serve it

benhalpern:
Yeah, @rhymes this can be moved to infra. I added an issue here: https://github.com/forem/infra/issues/163

PR #12100, opened 2021-01-03
Bug Fix:Update Article Cached Org and Docs When Org is Deleted

mstruve:
- [x] Bug Fix

## Description
We had a couple of bugs showing up here.
1. When an organization is deleted we were not updating the associate article Elasticsearch documents.
2. Upon finding 1, I also discovered we were not updating the `cached_organization` on articles when an organization was removed. 

Why didn't we notice this sooner? 
We didn't notice this sooner bc technically we will not destroy an organization if it has any associated articles according to our organization policy.
```ruby
  def destroyable?
    organization_memberships.count == 1 && articles.count.zero?
  end
```
Regardless, I still think having these in place would be a good idea in the event we remove that requirement one day OR in the event someone manually removes an organization from a console. Ideally, you don't want that to happen but this ensures we are covered if it does. 

Stole this from the [SRE project board](https://github.com/orgs/forem/projects/6)
https://github.com/orgs/forem/projects/18#card-52095568

## Added tests?
- [x] Yes

alt_text

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -11,6 +11,17 @@ class Organization < ApplicationRecord
 
   acts_as_followable
 
+  before_validation :downcase_slug
+  before_validation :check_for_slug_change
+  before_validation :evaluate_markdown
+  before_save :update_articles
+  before_save :remove_at_from_usernames
+  before_save :generate_secret
+  # You have to put before_destroy callback BEFORE the dependent: :nullify
+  # to ensure they execute before the records are updated
+  # https://guides.rubyonrails.org/active_record_callbacks.html#destroying-an-object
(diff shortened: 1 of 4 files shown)


app/models/organization.rb
```diff
@@ -11,6 +11,17 @@ class Organization < ApplicationRecord
 
   acts_as_followable
 
+  before_validation :downcase_slug
```
mstruve, line 14: We have a rubocop that forces us to put callbacks in a certain order. Since I had to move the before_destroy callback before the `dependent: nil` statements the rest of these had to come with it.

rhymes (approved):
LGTM!

benhalpern (approved):

PR #12110, opened 2021-01-04
Add `max_bytes` to imgproxy.

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This feature adds the `max_bytes` default option to help reduce the size of overly big images. This will automatically add optimization only where it is most needed... On the most over-sized images.

This is related to this PR: https://github.com/forem/forem/pull/12089

But as noted, some of that is better implemented with imgproxy ENV vars. But this one seems appropriate.

## Added tests?

- [x] Yes - Adjusted existing test to check for this variable.
      have not been included_

### DIFF
--- a/app/services/images/optimizer.rb
+++ b/app/services/images/optimizer.rb
@@ -34,6 +34,8 @@ module Images
     DEFAULT_IMGPROXY_OPTIONS = {
       height: nil,
       width: nil,
+      format: nil,
+      max_bytes: 500_000, # Keep everything under half of one MB.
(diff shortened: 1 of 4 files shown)


maestromac (approved):
LGTM!

looks like `/spec/views/dashboards/show.html.erb_spec.rb:18` also needs to be updated

rhymes (approved):
LGTM!

PR #12112, opened 2021-01-04
Move CodeBlockParse and MarkdownParser to service

atsmith813:
- [x] Refactor

## Description
This PR refactors `CodeBlockParser`, `MarkdownParser`, and `Html::ParseEmoji`. The previous setup was interesting, markdown and HTML parsing are very tightly connected 🤔 .

- I turned `CodeBlockParser` into `HtmlParser` to look and feel a little more like `MarkdownParser`. The way I view this new setup is `HtmlParser` handles parsing of HTML chunks and `MarkdownParser` focuses more on markdown. I moved methods from `MarkdownParser` to the new `HtmlParser` that were related to HTML to clean that up.
- I decided to scrap `Html::ParseEmoji` as a dedicated service and rolled it into `HtmlParser` as its own method.
- The "bigger change" here is the `HtmlParser` is built in a way to support the chaining of methods allowing us to go from [this](https://github.com/forem/forem/blob/3bf4f33510f4d72978bf71d6f94c3b8e9fb716cf/app/labor/markdown_parser.rb#L35-L46):
```ruby
html = remove_nested_linebreak_in_list(html)
html = prefix_all_images(html)
html = wrap_all_images_in_links(html)
html = add_control_class_to_codeblock(html)
html = add_control_panel_to_codeblock(html)
html = add_fullscreen_button_to_panel(html)
html = wrap_all_tables(html)
html = remove_empty_paragraphs(html)
html = escape_colon_emojis_in_codeblock(html)
html = unescape_raw_tag_in_codeblocks(html)
html = wrap_all_figures_with_tags(html)
wrap_mentions_with_links!(html)
```

to [this](https://github.com/forem/forem/blob/34aba69b76bd816cddeeed41397b5b6ef0dc48ed/app/services/markdown_parser.rb#L129-L143):
```ruby
HtmlParser
  .new(html)
  .remove_nested_linebreak_in_list
  .prefix_all_images
  .wrap_all_images_in_links
  .add_control_class_to_codeblock
  .add_control_panel_to_codeblock
  .add_fullscreen_button_to_panel
  .wrap_all_tables
  .remove_empty_paragraphs
  .escape_colon_emojis_in_codeblock
  .unescape_raw_tag_in_codeblocks
  .wrap_all_figures_with_tags
  .wrap_mentions_with_links
  .html
```
Thanks to @citizen428 for the chaining idea!

In addition to chaining, this setup is still flexible enough for other areas of code to call specific HTML parsing actions as needed.

## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

batman_thinking_gif


app/services/html/prefix_all_images.rb
```diff
@@ -0,0 +1,36 @@
+module Html
+  class PrefixAllImages
+    def self.call(html, width = 880)
+      return unless html
```
citizen428, line 4: **nitpick (non-blocking):** Maybe `unless html.present?`, we don't need to do any work on empty strings.

```suggestion
      return unless html.present?
```

app/services/html/prefix_all_images.rb
```diff
@@ -0,0 +1,36 @@
...
+                       src.gsub("https://media.", "https://i.")
+                     else
+                       img_of_size(src, width)
+                     end
+      end
+
+      doc.to_html
+    end
+
+    def self.allowed_image_host?(src)
```
citizen428, line 27: **nitpick (non-blocking):** Should we use `private_class_method` to remove this from the public interface?

app/services/html/prefix_all_images.rb
```diff
@@ -0,0 +1,36 @@
...
+      # GitHub camo image won't parse but should be safe to host direct
+      src.start_with?("https://camo.githubusercontent.com")
+    end
+
+    def self.img_of_size(source, width = 880)
```
citizen428, line 32: **nitpick (non-blocking):** Should we use `private_class_method` to remove this from the public interface?

app/labor/markdown_parser.rb
```diff
@@ -32,8 +32,8 @@ def finalize(link_attributes: {})
     rescue Liquid::SyntaxError => e
       html = e.message
     end
-    html = remove_nested_linebreak_in_list(html)
-    html = prefix_all_images(html)
+    html = Html::RemoveNestedLinebreakInList.call(html)
```
citizen428, line 35: Are we calling this anywhere else? Right now it seems only `Html::PrefixAllImages` has a separate caller (`app/models/display_ad.rb`), so maybe we could inline this similar to e.g. `wrap_all_images_in_links`? 

Of course, this also depends on what your goals for refactoring the Markdown parser itself are. I could see a use case for for a `HtmlProcessor` object with a fluent interface (for chaining), something like

```ruby
class HtmlProcessor
  attr_reader :html

  def initialize(html)
    @html = html
  end

  def remove_nested_linebreak_in_list
    ...
    self
  end
  
   ....
end
```

This would turn the assignment cascade here into something like

```ruby
html = HtmlProcessor.
  new(html).
  remove_nested_linebreak_in_list.
  prefix_all_images.
  html
```

Anyway, just a random idea, pick what you think is best 😃

app/models/display_ad.rb
```diff
@@ -36,6 +36,6 @@ def process_markdown
     #                                                         attributes: %w[href target src height width style]
     stripped_html = initial_html.html_safe # rubocop:disable Rails/OutputSafety
     html = stripped_html.delete("\n")
-    self.processed_html = MarkdownParser.new(html).prefix_all_images(html, 350)
+    self.processed_html = HtmlParser.new(html).prefix_all_images(350).html
```
citizen428, line 39: Nice, I like how this reads :-) In the interest of keeping our services folder clean, should it maybe be `Html::Parser`?

citizen428 (approved):
Nice, thanks for the major refactor @atsmith813!

app/models/display_ad.rb
rhymes: +1 on this idea, so everything sits under `services/html`

app/services/html_parser.rb
```diff
@@ -1,16 +1,37 @@
-module CodeBlockParser
+class HtmlParser
+  # Each of the instance methods should return self to support chaining of
+  # methods
```
rhymes, line 3: ❤️

app/services/html_parser.rb
```diff
@@ -1,16 +1,37 @@
...
+  #    html
+
   include InlineSvg::ActionView::Helpers
   include ApplicationHelper
 
-  def remove_nested_linebreak_in_list(html)
-    html_doc = Nokogiri::HTML(html)
+  RAW_TAG_DELIMITERS = ["{", "}", "raw", "endraw", "----"].freeze
+
+  attr_reader :html
```
rhymes, line 17: I see this is an attribute reader, is there a specific reason why you opted not to make this an accessor, to you could assign to it? I see down that you assign to `@html` a bunch of times

app/services/html_parser.rb
```diff
@@ -1,16 +1,37 @@
...
+
+  def initialize(html)
+    @html = html
+  end
+
+  def remove_nested_linebreak_in_list
+    html_doc = Nokogiri::HTML(@html)
```
rhymes, line 24: this is just personal preference but by having an attribute reader defined, I'd use that in lieu of the instance variable `@html`

app/services/html_parser.rb
```diff
@@ -71,40 +107,55 @@ def add_fullscreen_button_to_panel(html)
...
-    html_doc = Nokogiri::HTML.fragment(html)
+  def escape_colon_emojis_in_codeblock
+    html_doc = Nokogiri::HTML.fragment(@html)
 
     html_doc.children.each do |el|
       next if el.name == "code"
 
       if el.search("code").empty?
-        el.swap(Html::ParseEmoji.call(el.to_html))
+        parsed_html = HtmlParser.new(el.to_html).parse_emojis.html
```
rhymes, line 139: why aren't we calling the parse emoji method directly instead of instancing a new instance of the same class we are in?

app/services/html_parser.rb
```diff
@@ -71,40 +107,55 @@ def add_fullscreen_button_to_panel(html)
...
+        el.swap(parsed_html)
       else
-        el.children = escape_colon_emojis_in_codeblock(el.children.to_html)
+        el.children = self.class.new(el.children.to_html)
+          .escape_colon_emojis_in_codeblock
```
rhymes, line 143: same thing here, I'm a bit unsure about this. Is it for the chaining mechanism?

rhymes (commented):
One thing that I'd like to explore after this is merged is to see if there's a way for this parser to avoid continuously parsing a fragment, serializing it, then pass the baton to the next method which re-parses, then processes, then serializes, so on and so forth. Maybe we can parse the document once and pass it along the various methods. This is the current implementation and we know it works but I'm making a mental note about looking at it once we're done with the move.

What do you think? Would that make sense?

app/models/display_ad.rb
atsmith813: And change `MarkdownParser` to `Markdown::Parser` under `services/markdown`?

atsmith813:
@rhymes 
> see if there's a way for this parser to avoid continuously parsing a fragment, serializing it, then pass the baton to the next method

Agreed - it would be nice to find a way around it. I thought about it for a bit initially...what makes it a little tricky is that 1) it seems we want to be able to call each of these methods individually _and_ chain them and 2) some methods parse the data slightly different (i.e. `Nokogiri::HTML.fragment`, `Nokogiri::HTML`, no Nokogiri at all, etc.). It should be doable - I had similar thoughts as you that it could be a separate PR to handle it.

app/services/html_parser.rb
atsmith813: That's a good question! Is there a way to do that under this setup?

Originally, this was using recursion and it called the method again. Since we removed the argument (`html`) from all the methods, I didn't see a way to really call the method itself and specify a new `html` argument. If we did that, we would overwrite the current value of `html` and it would return the incorrect data. In other words, we would have to do something like:

```ruby
@html = el.to_html
parsed_html = parse_emojis.html
el.swap(parsed_html)
```

I don't think this would work because it would reassign the value of `@html` to only a subset of the `html` from the loop, not the entire thing reformatted. Does that make sense?

app/services/html_parser.rb
atsmith813: It prevents anyone from reassigning `html` by doing something like:

```ruby
html_parser = HtmlParser.
  new(html).
  some_chained_method.
  html
  
html_parser.html = some_new_html
```

Currently, that re-assignment at the end wouldn't work because it would fail with undefined method `html=`. It would work if we made it an accessor. I'm open to making it an accessor though, I don't expect this would happen often, if at all. Admittedly, I'm probably being more careful than necessary.

@rhymes What do you think, should we change it?

app/models/display_ad.rb
atsmith813: Just kidding, that causes conflicts with `Redcarpet` 😅

app/services/html_parser.rb
rhymes: Ah right! Make sense, maybe the solution is to split it: attr_reader stays public, attr_writer is moved down as private. That might work

app/services/html_parser.rb
rhymes: Ah got it! Maybe this "hack" is an indication that it should be its own separate service?

app/models/display_ad.rb
atsmith813: Updated!

app/services/html_parser.rb
atsmith813: Good idea, updated!

app/services/html_parser.rb
atsmith813: I've had similar thoughts as well. This PR originally started on that path - making these separate services. It feels like the trick here is we sort of use these services/methods inconsistently. Some are called individually outside of the parsing process, some aren't. Breaking only a couple of these methods out into dedicated services feels a little strange to me too since we are supporting chaining and these are mostly used in the parsing funnel of logic.

I do understand what you mean though, there's a slight smell here. I'm definitely open to changing this up. I was leaning towards chalking that up to the fact this is legacy code and this PR still accomplishes the goal (move these to services) while making improvements against the current implementation (now support chaining, more spec coverage, etc.).

app/services/html/parser.rb
```diff
@@ -0,0 +1,284 @@
...
+    end
+
+    def unescape_raw_tag_in_codeblocks
+      return self if @html.blank?
+
+      @html.gsub!(RAW_TAG, "{% raw %}")
+      @html.gsub!(END_RAW_TAG, "{% endraw %}")
+      html_doc = Nokogiri::HTML(@html)
+      html_doc.xpath("//body/div/pre/code").each do |codeblock|
+        next unless codeblock.content.include?(RAW_TAG) || codeblock.content.include?(END_RAW_TAG)
```
atsmith813, line 166: Doing this wasn't my first choice 😅 

When I changed the name of this file to `Html::Parser`, it caused this line to be indented further because we added `module Html` at the top. That caused a rubocop violation for the line then being too long.

Then, I moved the logic to a method, something like:
```ruby
next unless codeblock_includes_raw_tags?(codeblock.content)

def codeblock_includes_raw_tags?(content)
  content.include?("{----% raw %----}") || content.include?("{----% endraw %----}")
end
```

Andddddd then codeclimate complained because the file contained 21 methods which is over the 20 method limit. I try to avoid "hacks" (i.e. disabling Rubocop and/or ignoring codeclimate) so this where I landed to satisfy both 🙈 .

atsmith813:
@citizen428 re-requesting review since I've made more changes 🙃

citizen428 (approved):

rhymes (approved):
LGTM

PR #12114, opened 2021-01-04
Take Apple Authentication out of Beta

fdoxyz:

- [x] Feature
- [x] Documentation Update

## Description

Removing `:apple_auth` feature flag to make SIWA fully available.

https://github.com/forem/internalEngineering/issues/4

1. Boot app locally, sign in with admin user and visit `/admin/customization/config`
2. Within the Authentication section you should be able to add Apple Authentication (without the need for any feature flag)
3. Add placeholder values for all 4 configs to enable the provider
   - Configuring this provider isn't trivial so there's no need to test it works locally unless you're able to
   - You can check the implementation works in DEV or other Forem instances
4. Visit `/enter` in an incognito window and ensure the provider is displayed

None - No changes have been made to the HTML

## Added tests?

- [x] Yes

## Added to documentation?

All authentication changes are slightly more mission critical, in the sense that if it breaks it will avoid users from accessing their accounts. A good post deployment task would be to validate Apple and other authentication providers continue to work as expected in DEV (which gets deployed post merge).

bad apple


app/services/authentication/authenticator.rb
```diff
@@ -86,6 +86,9 @@ def current_user_identity_exists?
     end
 
     def proper_user(identity)
+      Rails.logger.debug "PROPER USER: #{current_user.inspect}"
+      Rails.logger.debug "IDENTITY: #{identity.inspect}"
+      Rails.logger.debug "--------" * 10
```
fdoxyz, line 91: The bug noticed in https://community.benhalpern.com is that when connecting an existing account with the SIWA OAuth process a new account is created (instead of connecting the user currently logged in).

I'm still debugging this part but I suspect we may be running into a problem related to the session not being available (because the Cookies have `SameSite=Lax`)

citizen428: Hi @fdoxyz, just checking in to see how this debugging is going?

fdoxyz: Thanks for bringing this up @citizen428!

The conclusion back then was that we need `SameSite=None` in order to make this work and we were blocked by TLS work (now shipped) in order to enforce `Secure` on our cookies. I'll loop back around this throughout the week since it shouldn't 🤞  be too difficult to sort it out now.

citizen428:
@fdoxyz OSS rotation check-in: what are our plans/next steps for this PR?

fdoxyz:
Hi @citizen428, thank you for checking in. I tried to finish up the last details needed here last week but I was quickly reminded how difficult and cumbersome this feature is to thoroughly test locally.

[This comment](https://github.com/forem/internalEngineering/issues/4#issuecomment-865416028) has an update on the suggested next steps, for now I'll put in some more work on this one this week.

app/views/users/_additional_authentication.html.erb
```diff
@@ -2,10 +2,9 @@
...
       <!--
-        TODO: [@forem/oss] Only until we refactor our cookie configurations
-        we'll be able to connect existing accounts with SIWA. This should be
-        done when Apple Auth Provider is no longer in 'Beta':
-        https://github.com/forem/forem/pull/12114
+        INFO: [@forem/oss] We only support creating new accounts and then
+        signing in with them) using Apple Auth. Connecting Forem existing
+        accounts to Apple Auth isn't supported and is why we skip it here.
       -->
       <% next if provider.provider_name == :apple %>
```
fdoxyz, line 9: The option to connect existing accounts with Apple Accounts hasn't been requested (in the months DEV and other Forems have had SIWA enabled), the technical implications to do so continue to be non-trivial to implement + might result in security vulnerabilities, and a separate project that allows for cross-Forem Authentication capabilities will be worked on in the near future.

Because of the above we won't support this option. I added a spec in `spec/requests/user/user_settings_spec.rb` to ensure this is the case.

app/views/admin/configs/show.html.erb
```diff
@@ -0,0 +1,1306 @@
+<div class="grid gap-6" data-controller="config">
```
maestromac, line 1: Is this file supposed to be here? I think this is a mistake in merge conflict because it was renamed to `app/views/admin/settings/show.html.erb `

maestromac (commented):
I'm not seeing Apple auth in config page. Is there something I missed? should I reset DB?

app/views/admin/configs/show.html.erb
fdoxyz: Great catch @maestromac it was definitely a mistake in a merge conflict resolution, thanks for mentioning this!

fdoxyz (commented):
> I'm not seeing Apple auth in config page. Is there something I missed? should I reset DB?

I don't think that should be necessary now that I pushed a fix for the merge conflict mistake. `bin/setup` is definitely recommended to ensure the DUS runs and the FeatureFlag isn't enabled anymore though.

joshpuetz (approved):

maestromac (approved):
LGTM!

PR #12118, opened 2021-01-05
Simplify onboarding images and centralize config around primary/secondary logos

benhalpern:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR removes specific onboarding and task card images and replaces them with "secondary logo", and instructs the admin to think of this as a playful alternative logo.

I think this is more simple and extensible. I could see us adding more detailed image config back in in the future, but as more thought out progressive enhancement with a refreshed admin UI to complement it. Right now it seems like unimportant additional config we can do without.

I also removed the right and left nav as we no longer use this in core functionality.

Check that the this does not cause issues anywhere now. Eyeball `/onboarding` etc.

This does not modify the UI in this way.

## Added tests?

- [x] Yes - Removed old tests. No new stuff needed.
      have not been included_

## Added to documentation?

      [Admin Guide](https://forem.gitbook.io/forem-admin-guide/) - Admin guide will need to be updated if this goes in.
- [ ] README
- [ ] No documentation needed


atsmith813 (approved):

spec/requests/onboardings_spec.rb
```diff
@@ -23,17 +23,5 @@
...
-
-    it "contains proper data attribute values if the onboarding config is present" do
-      allow(SiteConfig).to receive(:onboarding_logo_image).and_return("onboarding_logo_image.png")
-      allow(SiteConfig).to receive(:onboarding_background_image).and_return("onboarding_background_image.png")
-
-      sign_in user
-      get onboarding_url
-
-      expect(response.body).to include(SiteConfig.community_description)
-      expect(response.body).to include(safe_logo_url(SiteConfig.onboarding_logo_image))
```
atsmith813, line 35: **question (non-blocking):** should we update this spec to check for `SiteConfig.secondary_logo_url` instead of deleting it?

benhalpern: Oh yeah, good call. I'll update this.

benhalpern:
@lisasy @vaidehijoshi I'm having trouble requesting you two as reviewers, but want to put this on your plate.

This is an attempt to simplify our data model in preparation for further project work on admin area...

It reduces the admin's ability to customize a few different places on the site, but in and of itself simplifies the mental model right now.

Let me know your thoughts. If there are any objections I could hold off until RFC mode.

pkfrank (commented):
This approach looks great to me.  It should help really help simplify things for Forem Creators who are getting up and running.

app/lib/constants/site_config.rb
```diff
@@ -278,12 +266,8 @@ module SiteConfig
         description: "Add the secret key for Google reCAPTCHA, which is used for reporting abuse",
         placeholder: "What is the Google reCAPTCHA secret key?"
       },
-      right_navbar_svg_icon: {
-        description: "The SVG icon used to expand the right navbar navigation menu. Should be a max of 24x24px.",
-        placeholder: SVG_PLACEHOLDER
-      },
       secondary_logo_url: {
-        description: "Used as the secondary logo",
+        description: "If you have an alternate logo, this is used in some areas such as onboarding. It's typically the more\"playful\" version of the logo",
```
vaidehijoshi, line 270:
```suggestion
        description: "A place for an alternate logo, if you have one. Used throughout member onboarding and in some sign in forms.",
```
It feels a bit odd for us to specify whether the logo is more playful or not here; I actually don't know if we should be doing that at all, but if that kind of guidance was going to live somewhere, I don't think it should be here (maybe in the creator onboarding guides or something like that?).

app/javascript/article-form/components/ClipboardButton.jsx
```diff
@@ -44,18 +44,13 @@ export const ClipboardButton = ({
...
       icon={CopyIcon}
-    />
-    <span
-      id="image-markdown-copy-link-announcer"
-      role="alert"
-      className={`fs-s ${showCopyMessage ? '' : 'opacity-0'}`}
     >
-      Copied!
-    </span>
+      {showCopyMessage ? 'Copied!' : 'Copy...'}
```
vaidehijoshi, line 52: Why did this change as a part of this PR? (Also, it looks like this...[is already](https://github.com/forem/forem/blob/13ff4c7fbe3a070f6af9dda3c3dfa87e6f2badb7/app/javascript/article-form/components/ClipboardButton.jsx#L52) on master? 🤔)

spec/requests/onboardings_spec.rb
vaidehijoshi: Agree with @atsmith813, we should check that `secondary_logo_url` does in fact render in the onboarding request body 👍🏽

app/javascript/article-form/components/ClipboardButton.jsx
ludwiczakpawel: That's weird... https://github.com/forem/forem/pull/12128

Something went off while merging..?

spec/requests/onboardings_spec.rb
benhalpern: Fixed this.

vaidehijoshi (commented):
Looks like there is [another spec](https://github.com/forem/forem/pull/12118#diff-e3820d144b6c9c1759d5cb081fe847dd1494a037537d55c5ee5bad5ef0d94ccdR28) that needs to be updated in order for the tests [on CI](https://travis-ci.com/github/forem/forem/jobs/470850406#L2271) to pass

vaidehijoshi (approved):

PR #12126, opened 2021-01-05
Feeds::Import: only import feeds that haven't been recently updated

rhymes:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR set at 4 hours the time span to ignore recently updated feeds. By moving the `feed_fetched_at` update to the end of the entire process (fetching from the web, parsing and importing articles) we should be safely confident in ignoring those users whose `feed_fetched_at` has been updated in the past 4 hours.

Thanks @mstruve for suggesting this approach.

1. Populate `feed_url` for at least one user, eg. `u = User.last; u.update(feed_url: "https://medium.com/feed/@vaidehijoshi")`
2. Start Sidekiq in a tab, `bundle exec sidekiq`
3. Run `Feeds::ImportArticlesWorker.perform_async(nil, 4.hours.ago)` manually or enable `:feeds_import` feature flag and run `Articles::RssReaderWorker.perform_async`
4. Verify that articles have been imported
5. Run step 3 again, check Sidekiq logs to see that the user has been ignored
6. Re-run  `Feeds::ImportArticlesWorker.perform_async(nil, 1.second.ago)` and check the user is picked up again (by lowering the time, we make sure the logic works at the end of the interval)

## Added tests?

- [x] Yes
      have not been included_


app/workers/feeds/import_articles_worker.rb
```diff
@@ -4,16 +4,10 @@ class ImportArticlesWorker
 
     sidekiq_options queue: :medium_priority, retry: 10
 
-    def perform(user_ids = [])
-      user_ids = Array.wrap(user_ids)
+    def perform(user_ids = [], earlier_than = nil)
```
mstruve, line 7: In an effort to prevent us from ending up with a never-ending job accidentally, I think we should require an `earlier_than` time here.

app/workers/articles/rss_reader_worker.rb
```diff
@@ -11,7 +11,7 @@ def perform
       return if SiteConfig.dev_to?
```
mstruve, line 11: Is the plan to push this out first, give it a test run, then remove this?

mstruve (commented):
One small suggestion and question otherwise this is 🔥

app/workers/feeds/import_articles_worker.rb
rhymes: Makes sense, though I'm a bit conflicted about changing it because it'd result in "hacks" to select all users, something like `earlier_than: Time.at(0)` or something. Or maybe we make use `perform(nil)`  as "select everything". What do you think?

app/workers/articles/rss_reader_worker.rb
rhymes: Yes!

app/workers/feeds/import_articles_worker.rb
mstruve: We should never be enqueueing one of these with all users based on how long the worker takes so I think it would be fine leaving it in a state where to get all users you would have to do something unconventional or hacky. I would even be game for a comment that warns that using the worker with a lot of users can cause long runtimes even.

rhymes: Okay! I'll rework the parameters to make the time span mandatory

app/controllers/users_controller.rb
```diff
@@ -336,9 +336,11 @@ def destroy_request_in_progress?
   def import_articles_from_feed(user)
     return if user.feed_url.blank?
 
-    worker = FeatureFlag.enabled?(:feeds_import) ? Feeds::ImportArticlesWorker : RssReaderFetchUserWorker
-
-    worker.perform_async(user.id)
+    if FeatureFlag.enabled?(:feeds_import)
```
mstruve, line 339: Do we still need this feature flag? Seems like we have had the`Feeds::ImportArticlesWorker` for a bit and it has been working.

benhalpern (approved):
This looks good to me.

mstruve (approved):
Question about the feature flag still being used but the rest of it looks good to go!

app/controllers/users_controller.rb
rhymes: That's what I want to remove in the next PR.

The next PR will remove `RssReader`, feature flags and everything. Time to rip the bandaid off :D

PR #12133, opened 2021-01-05
Bump Nokogiri to 1.11.1

maestromac:
- [x] Bug Fix

## Description
Upgrading Nokogiri to 1.11. The only breaking feature is  https://github.com/sparklemotion/nokogiri/releases/tag/v1.11.0

> The Node methods add_previous_sibling, previous=, before, add_next_sibling, next=, after, replace, and swap now correctly use their parent as the context node for parsing markup. These methods now also raise a RuntimeError if they are called on a node with no parent.

which was address in this PR.

Resolves https://github.com/forem/InternalProjectPlanning/issues/386
N/A : the change is very minimal so CI is enough

## Added tests?
- [x] No, and this is why: the existing test is sufficient

## Added to documentation?
- [x] No documentation needed

n/a


app/labor/code_block_parser.rb
```diff
@@ -93,7 +93,7 @@ def escape_colon_emojis_in_codeblock(html)
       next if el.name == "code"
 
       if el.search("code").empty?
-        el.swap(Html::ParseEmoji.call(el.to_html))
+        el.swap(Html::ParseEmoji.call(el.to_html)) if el.parent.present?
```
maestromac, line 96: There is already a spec written for this feature which still passes for this change. From what I understand, this code was trying to perform swap on parentless text dom that mainly consists of newline ("\n\n\n\n\n"). It was never meant to work so they added a proper `RuntimeError`

rhymes (approved):
Great job @maestromac!

rhymes:
@maestromac Nokogiri has released a new minor version, you might want to snuck that in while you're at it https://github.com/forem/forem/pull/12145

fdoxyz (approved):
Nice 👍 

non-blocking for me as well @rhymes comment about bumping to `11.1`

PR #12045, opened 2020-12-28
Use Travis builtin caching

rhymes:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Builds are broken, probably for something related to missing packages in `node_modules` - https://travis-ci.com/github/forem/forem/jobs/465368676#L678 - trying to see if by using Travis's builtin caching mechanism - https://docs.travis-ci.com/user/caching/ - things change.

PR #12140, opened 2021-01-05
[Travis] Use directories cache for Bundle and Yarn

maestromac:
- [x] CI

## Description
Essentially reverting 3764cd6. For some reason, this works again but caches are still not showing up on Travis's dashboard. This should be a UI issue as the cache is working as expected.

n/a
How to verify that it is actually working? There are two dependencies command to look for

`bin/ci-bundle` and `yarn install --frozen-lockfile`. 

Each of these should not take more than 5 seconds.

This is a tiny image but it shows that it took only 1.8 second to run yarn install.
image
## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: ruby
 cache:
   bundler: true
-  npm: true
   directories:
+    - node_modules
     - $HOME/.nvm
     - ~/.cache # cache folder with Cypress binary
 rvm:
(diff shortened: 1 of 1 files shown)


nickytonline (approved):
Weird that it's working again the old way after making the change in #12045. Anyways, LGTM.

mstruve (approved):
make up your mind Travis

rhymes:
LOL

PR #12147, opened 2021-01-06
Break StickyArticleCollection into services

citizen428:

- [x] Refactor
- [X] Optimization

## Description

This PR removes `StickyArticleCollection` from `app/labor` and breaks it into two services, `Articles::GetUserStickies` and `Articles::SuggestStickies`. I also refactored the view code, since previously we called both methods even though their results were used in an `elsif`. The refactored code only performs the second DB call if the first one returned an empty collection.

No UI changes.

## Added tests?

- [x] Yes

## Added to documentation?

- [X] No documentation needed


rhymes (approved):

juliannatetreault (approved):
Nice!! 👍 👍

PR #12149, opened 2021-01-06
Add application deployment information to the admin page

Ridhwana:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Forem administrators are now able to to determine from the admin page (http://localhost:3000/admin) when they last deployed and what was the last commit ID that was deployed.

I used the following to determine how to set the deploy and commit variables based on environment:
|                  | Forem Cloud Variables | Heroku Variables          |
|------------------|-----------------------|---------------------------|
| Last deployed    | FOREM_BUILD_DATE      | HEROKU_RELEASE_CREATED_AT |
| Latest Commit ID | FOREM_BUILD_SHA       | HEROKU_SLUG_COMMIT        |

I want to note that I did not use the RELEASE_FOOTPRINT which differs from what the `data-release-footprint` key in the `<body>` tag uses. Instead I used the FOREM_BUILD_DATE and FOREM_BUILD_SHA which seemed more apt. I could use some validation on this, and perhaps we could change the data keys in the body to reflect the same info as the admin page.

https://github.com/forem/InternalProjectPlanning/issues/230
Related slack thread: https://forem-team.slack.com/archives/C0134R988P9/p1607586614240100

Add export HEROKU_RELEASE_CREATED_AT="Optional" to the env file to set the HEROKU_RELEASE_CREATED_AT variable. 

The reason that it says `Optional` under the values is because as per above comment I've set those values in the env file with the string Optional.
Screenshot 2021-01-06 at 12 35 14

Besides changing the values for HEROKU_SLUG_COMMIT and HEROKU_RELEASE_CREATED_AT in the env file, I don't think there's a ton I can do to test this in development. Hence, I'd like to get the opinion of someone who is more knowledgeable than me on #systems and how these export variables are integrated into the app to ensure that this looks like it would work in prod. Based on this discussion I could put it behind a feature flag to test. 

## Added tests?

- [ ] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [ ] No documentation needed

alt_text


.env_sample
```diff
@@ -17,6 +17,7 @@ export RACK_TIMEOUT_WAIT_TIMEOUT=100_000
 export RACK_TIMEOUT_SERVICE_TIMEOUT=100_000
 
 # Heroku release slug used to bust caches on deploys
+export HEROKU_RELEASE_CREATED_AT="Optional"
 export HEROKU_SLUG_COMMIT="Optional"
```
rhymes, line 21: Instead of using "Optional" for these two, can we keep the empty? This way you can test for presence and display "Not available" or something similar in case they aren't

rhymes (commented):
A way to test write a test for this would be to inject environment variables as a setup for the system test, something like: `allow(ENV).to receive(:[]).with("VAULT_TOKEN").and_return("present")` (copied and pasted it from `lib/app_secrets_spec.rb`)

Ridhwana:
> A way to test write a test for this would be to inject environment variables as a setup for the system test, something like: `allow(ENV).to receive(:[]).with("VAULT_TOKEN").and_return("present")` (copied and pasted it from `lib/app_secrets_spec.rb`)

@rhymes thanks! This sounds good - I was just waiting for a review that this is the right path before I worked on the tests.

.env_sample
Ridhwana: Sounds good, I'll change it 👍🏽

app/views/admin/admin_portals/index.html.erb
```diff
@@ -1,3 +1,15 @@
+<div>
+  <div class="tag-card crayons-card p-4 m:p-6 m:pt-4 mb-4 flex flex-row relative">
+    <div class="mr-8">
+      <div class="crayons-field__description">Last deployed</div>
```
jdoss, line 4: `FOREM_BUILD_DATE` is the datetime when the container was created. When we start the container `RELEASE_FOOTPRINT` is the point in time that the container was started or deployed. I would change this to Container Build Date and make another section for Last Deployed that uses `RELEASE_FOOTPRINT` for the last time Forem was deployed.

jdoss (changes requested):

app/views/admin/admin_portals/index.html.erb
```diff
@@ -1,3 +1,15 @@
...
+      <div><%= ApplicationConfig["FOREM_BUILD_DATE"].presence || ENV["HEROKU_RELEASE_CREATED_AT"] %></div>
```
citizen428, line 5: I'm not saying we need this now, BUT I think it could make sense to add a new model (not DB backed) like `ForemInstance` to which we add a bunch of class methods that encapsulate things like this:

```ruby
class ForemInstance
  def self.deployed_at
    ApplicationConfig["FOREM_BUILD_DATE"].presence || ENV["HEROKU_RELEASE_CREATED_AT"
  end

  def self.latest_commit_id
    ApplicationConfig["FOREM_BUILD_SHA"].presence || ENV["HEROKU_SLUG_COMMIT"]
  end
end
```

We probably will have a few more and I'd maybe even move `dev_to?` from `SiteConfig` to that new class since `ForemInstance.dev_to?` makes a lot more sense to me than `SiteConfig.dev_to?`. We could even move the `local?` method for the same reason (`ForemInstance.local?`). 

I found that in a lot of Rails apps the main concept itself often has no representation in code (e.g. blogs have `Author`, `Post` and `Comment` models, but nothing to represent the whole blog), but they often come in handy to collect certain utility functions in a sensible location.

app/views/admin/admin_portals/index.html.erb
Ridhwana: @jdoss thank you so much for this 😊 I'll use RELEASE_FOOTPRINT for Last deploy date. I don't think our users would need to know the Container Build Date so I'll leave that out. Thanks!

app/views/admin/admin_portals/index.html.erb
Ridhwana: @citizen428 this makes sense. I didn't include the changes in this PR, but I'm creating a follow up PR tomorrow. Thanks for the feedback :)

spec/requests/admin/admin_portals_spec.rb
```diff
@@ -22,4 +22,22 @@
       expect(response.body).not_to include("Config: Profile Setup")
     end
   end
+
+  describe "Last deployed and Lastest Commit ID card" do
+    it "shows 'Not Available' if the Last deployed time is missing" do
+      ENV["HEROKU_RELEASE_CREATED_AT"] = ""
```
rhymes, line 28: non blocking: this sets the variable globally for any test

rhymes (approved):
LGTM

jdoss (approved):
LGTM!

spec/requests/admin/admin_portals_spec.rb
Ridhwana: @rhymes I tried to set it so many other like using         
`allow(ENV).to receive(:[]).with("HEROKU_RELEASE_CREATED_AT").and_return("")` but I just couldn't get it to work. I'll look into it again tomorrow.

Ridhwana:
Will look into why this test is failing...

spec/requests/admin/admin_portals_spec.rb
rhymes: Ah got it! No worries, we can keep it like this as well :)

citizen428: @Ridhwana Maybe something like this?

```ruby
stub_const("ENV", ENV.to_h.merge("HEROKU_RELEASE_CREATED_AT" => ""))
```

app/views/admin/admin_portals/index.html.erb
citizen428: No rush, like I said it really doesn't need to be part of this PR, but I feel like this will be a useful addition. Thanks for doing a follow-up PR! 😃

spec/requests/admin/admin_portals_spec.rb
Ridhwana: Thanks @citizen428 this worked!

spec/helpers/application_helper_spec.rb
```diff
@@ -48,6 +48,7 @@
     it "includes SiteConfig.admin_action_taken_at" do
       Timecop.freeze do
         allow(SiteConfig).to receive(:admin_action_taken_at).and_return(5.minutes.ago)
+        allow(ApplicationConfig).to receive(:[]).with("RELEASE_FOOTPRINT").and_return("abc123")
```
Ridhwana, line 51: This test was failing because it seems like ApplicationConfig["RELEASE_FOOTPRINT"] was "", hence it was returning early in the test based on this code 

```
    release_footprint = ApplicationConfig["RELEASE_FOOTPRINT"]
    return path if release_footprint.blank?

    "#{path}-#{params[:locale]}-#{release_footprint}-#{SiteConfig.admin_action_taken_at.rfc3339}"
```

I've reset the value in the test again. I went through my code and I'm not totally sure why this failed in this PR as I'm not resetting the value of RELEASE_FOOTPRINT anywhere in my tests, however it seems very coincidental. 

Whilst this is a fix for the test, I wanted a second pair of 👀's before merging in case I've overlooked something.  

cc @rhymes

rhymes: @Ridhwana I think it was failing because of `config/initializers/set_release_footprint.rb` which is initialized to `HEROKU_SLUG_COMMIT` on startup? That's the only clue I have :D

PR #12169, opened 2021-01-07
Remove RssReader

rhymes:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR rips the bandaid off removing the old `RssReader` by implementing the following things:

- removes programmatically the `:feeds_import` feature flag, via data update script
- removes `RssReader` and all related workers and classes
- renames `RssReader::Assembler` to `Feeds::AssembleArticleMarkdown`
- sets `earlier_than` to default to `nil` for `Feeds::ImportArticlesWorker` moving the default argument inside the worker to support Sidekiq Cron YAML limitations

(finally) closes https://github.com/forem/forem/issues/10996

- https://github.com/forem/forem/pull/12126
- https://github.com/orgs/forem/projects/6#card-44508281
- https://github.com/orgs/forem/projects/14#card-48867659

Start the app, play with `Feeds::ImportArticlesWorker` (optionally wait for the 20th minute after the hour to make sure it's working, or change `schedule.yml` to a closer minute to verify it works :))

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

We need to keep an eye on the `Articles::RssReaderWorker` job which might fail if it's picked up after deploy, no big deal in my opinion because at the following hour `Feeds::ImportArticlesWorker` will start and pick up the new work, we might just want to disable the retry manually of the former job to avoid wasting CPU cycles and resources.

We also need to remove the Heroku schedule job `fetch_all_rss` from rotation, as it's going to fail at the next cycle


app/controllers/users_controller.rb
```diff
@@ -336,11 +336,7 @@ def destroy_request_in_progress?
   def import_articles_from_feed(user)
     return if user.feed_url.blank?
 
-    if FeatureFlag.enabled?(:feeds_import)
-      Feeds::ImportArticlesWorker.perform_async(nil, user.id)
-    else
-      RssReaderFetchUserWorker.perform_async(user.id)
-    end
+    Feeds::ImportArticlesWorker.perform_async(nil, user.id)
```
rhymes, line 339: bye bye feature flag

app/models/user.rb
```diff
@@ -622,11 +622,7 @@ def bust_cache
   def validate_feed_url
     return if feed_url.blank?
 
-    valid = if FeatureFlag.enabled?(:feeds_import)
-              Feeds::ValidateUrl.call(feed_url)
-            else
-              RssReader.new.valid_feed_url?(feed_url)
-            end
+    valid = Feeds::ValidateUrl.call(feed_url)
```
rhymes, line 625: bye bye feature flag

app/services/feature_flag.rb
```diff
@@ -1,6 +1,6 @@
 module FeatureFlag
   class << self
-    delegate :disable, :enable, :enabled?, :exist?, to: Flipper
+    delegate :disable, :enable, :enabled?, :exist?, :remove, to: Flipper
```
rhymes, line 3: need `:remove` for the new data update script. We should also make a habit of adding/removing and controlling feature flags programmatically and not via the web UI if possible, to ensure all Forems are aligned

app/services/feeds/assemble_article_markdown.rb
```diff
@@ -0,0 +1,170 @@
+module Feeds
+  class AssembleArticleMarkdown
```
rhymes, line 2: this is the same as the old `RssReader::Assembler` class, only renamed `.assemble` to `.call`

app/services/feeds/import.rb
```diff
@@ -29,8 +29,8 @@ def call
         # NOTE: doing this sequentially to avoid locking problems with the DB
         # and unnecessary conflicts
         articles = feedjira_objects.flat_map do |user_id, feed|
-          # TODO: replace `feed` with `feed.url` as `RssReader::Assembler`
-          # only actually needs feed.url
+          # TODO: replace `feed` with `feed.url` as `Feeds::AssembleArticleMarkdown`
+          # only actually uses feed.url
```
rhymes, line 33: not going to optimize this right now, will leave for a future downtime as it requires a bit of shuffling that's not worth it right now. We can address this when we optimize `Feeds::AssembleArticleMarkdown`

app/workers/feeds/import_articles_worker.rb
```diff
@@ -4,8 +4,19 @@ class ImportArticlesWorker
 
     sidekiq_options queue: :medium_priority, retry: 10
 
-    def perform(earlier_than, user_ids = [])
-      users = user_ids.present? ? User.where(id: user_ids) : nil
+    # NOTE: [@rhymes] we need to default earlier_than to `nil` because sidekiq-cron,
+    # by using YAML to define jobs arguments does not support datetimes evaluated
+    # at runtime
+    def perform(earlier_than = nil, user_ids = [])
```
rhymes, line 10: I thought about splitting this in two separate jobs (as it was in the past) but: I'm not sure it's worth it, and I didn't want to bother to do another migration (we need to keep the two parameters for a while, exhaust existing jobs and then remove the second parameter).

I'm open to change it or to better ideas

spec/lib/data_update_scripts/remove_feeds_import_feature_flag_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    described_class.new.run
+
+    expect(FeatureFlag.exist?(:feeds_import)).to be(false)
+  end
+
+  it "works if the flag is not available" do
+    described_class.new.run
+
+    expect(FeatureFlag.exist?(:feeds_import)).to be(false)
+  end
```
maestromac, line 19: I was able to manually run `FeatureFlag.remove(:feeds_import)` locally twice to see what happens but then I saw your awesome test :100:

app/workers/feeds/import_articles_worker.rb
maestromac: dumb question, can you elaborate on what you mean by having two separate jobs here and what would we benefit from it?

maestromac (approved):
Oh boy, the final step! LGTM!

citizen428 (approved):
Nice! 🔥

app/workers/feeds/import_articles_worker.rb
rhymes: The second argument right now it's only used when we fetch the feed from the settings page (when the user saves the profile information or explicitly requests a re-fetch), adding a separate job would cleanup the `if`, but nothing more. That's why I'm not super convinced about adding it

rhymes:
Going to keep this as a draft until the US team is online as I'd like SRE to be around when we deploy this, see post deployment actions above

app/workers/feeds/import_articles_worker.rb
maestromac: Thanks for explaining! yeah, I agree it's good as is.

mstruve (approved):

mstruve:
@maestromac can you work with @rhymes to find a good time to deploy this? We will definitely want to keep a close eye on things for a day or so after.

rhymes:
@mstruve we're on it, with both glasses and goggles! :D

PR #12171, opened 2021-01-07
Move Bufferizer to services

atsmith813:
- [x] Refactor

## Description
This PR takes the current `Bufferizer` module in `/labor` and breaks the methods up into dedicated services. I could be convinced to take a different approach (i.e. putting them all in one service).

I chose to break them up because while there is some overlap in basic logic, the difference between arguments (i.e. listing vs article) and the difference between context (i.e. tweets vs posts) felt large enough to justify some isolation. This isolation also gives us more flexibility in the future. I know there has been talking bout finding a replacement for Buffer at some point. There could be a point in the future where we want to post to more platforms. This isolation would make those changes easier to manage.

## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

cleanup_aisle_5_gif


app/services/bufferizer/listings_tweet.rb
```diff
@@ -0,0 +1,22 @@
+module Bufferizer
+  class ListingsTweet
+    def self.call(listing, tweet, _admin_id = nil)
```
citizen428, line 3: I think I'd just remove the last argument instead of ignoring it and update the call site. They are distinct services, they don't necessarily have to have the same public interface.

app/services/bufferizer/facebook_post.rb
```diff
@@ -0,0 +1,37 @@
+module Bufferizer
+  class FacebookPost
+    def self.call(article, post, admin_id = nil)
```
citizen428, line 3: Is `admin_id` really optional? All actual calls in `app/controllers/admin/buffer_updates_controller.rb` seem to pass it in, only the specs don't.

app/services/bufferizer/listings_tweet.rb
```diff
@@ -0,0 +1,22 @@
...
+      BufferUpdate.send_to_buffer(
+        listings_twitter_text(tweet, listing),
+        buffer_listings_id,
+      )
+
+      listing.update(last_buffered: Time.current)
+    end
+
+    def self.listings_twitter_text(tweet, listing)
+      "#{tweet} #{URL.url(listing.path)}" if tweet.size <= 255
```
citizen428, line 17: Should we make 255 a constant with a more intention-revealing name?

app/services/bufferizer/satellite_tweet.rb
```diff
@@ -0,0 +1,36 @@
...
+  class SatelliteTweet
+    def self.call(article, tweet, admin_id = nil)
+      return unless article && tweet
+
+      article.tags.find_each do |tag|
+        next if tag.buffer_profile_id_code.blank?
+
+        text = twitter_buffer_text(tweet, article)
+
+        if text.length < 250 && SiteConfig.twitter_hashtag
```
citizen428, line 11: Same question here, can the 250 be moved into a constant?

app/services/bufferizer/satellite_tweet.rb
```diff
@@ -0,0 +1,36 @@
...
+          tag_id: tag.id,
+          admin_id: admin_id,
+        )
+      end
+
+      article.update(last_buffered: Time.current)
+    end
+
+    def self.twitter_buffer_text(tweet, article)
+      "#{tweet} #{URL.article(article)}" if tweet.size <= 255
```
citizen428, line 31: Constant?

citizen428 (approved):

app/services/bufferizer/listings_tweet.rb
atsmith813: Good catch! Making our linters work overtime, I didn't even do this 😅

app/services/bufferizer/facebook_post.rb
atsmith813: Good catch, updated!

app/services/bufferizer/listings_tweet.rb
atsmith813: Done!

app/services/bufferizer/satellite_tweet.rb
atsmith813: I'm not entirely sure where `250` comes from 🤔  but I gave it a shot!

app/services/bufferizer/satellite_tweet.rb
atsmith813: Done!

vaidehijoshi (approved):

PR #12172, opened 2021-01-07
Feature:Add ability to add a Tag Resource Admin through Admin

mstruve:
- [x] Feature

## Description
I have needed to add a Resource Tag admin twice now and both times I had to do it through a console since its not an option in admin. This changes that :) 

1. Go to `/admin/users`
2. Click on any user and select "Manage User"
3. Add the "Resource Admin: Tag" option and they should be able to work with tags in admin.

## Added tests?
- [x] No, this method and endpoint is test, we are merely adding an additional option

alt_text

### DIFF
--- a/app/lib/constants/role.rb
+++ b/app/lib/constants/role.rb
@@ -19,6 +19,7 @@ module Constants
                      "Resource Admin: Broadcast",
                      "Resource Admin: HtmlVariant",
                      "Resource Admin: DisplayAd",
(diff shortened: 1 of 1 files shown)


Zhao-Andy (approved):
👍

juliannatetreault (approved):
Works as expected! 🚀

vaidehijoshi (approved):

PR #12173, opened 2021-01-08
Remove looking for work columns

citizen428:

- [X] Refactor

## Description

This PR removes the columns `looking_for_work` and `looking_for_work_publicly` from the `users` table. It also removes code using these columns and has a data update script for removing the associated profile fields that were meant to replace them. The columns where previously due to profile generalization work, so we can just go ahead and remove them.

Question: we have a somewhat related column for "Recruiters can contact me about job opportunities", should we remove this too as part of this PR? /cc @vaidehijoshi

https://github.com/forem/internalEngineering/issues/366

Nothing specific, the specs should cover this. You can go to `http://localhost:3000/settings` and verify that the checkbox in the "Work" section is gone after the data update script removed the profile field.

None

## Added tests?

- [X] Yes, updated existing tests
- 
## Added to documentation?

- [X] No documentation needed

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,12 +10,14 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_01_05_183127) do
+ActiveRecord::Schema.define(version: 2021_01_08_031718) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
+  enable_extension "pg_trgm"
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
+  enable_extension "unaccent"
 
   create_table "ahoy_events", force: :cascade do |t|
     t.string "name"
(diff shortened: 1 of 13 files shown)


lib/data_update_scripts/20210108033107_remove_looking_for_work_profile_fields.rb
```diff
@@ -0,0 +1,10 @@
+module DataUpdateScripts
+  class RemoveLookingForWorkProfileFields
+    def run
+      # destroy_by is idempotent by default: if no record can be found an empty
```
citizen428, line 4: Note: I did not add a spec for this script as it would essentially just retest `destroy_by` which didn't seem worthwhile.

vaidehijoshi:
> Question: we have a somewhat related column for "Recruiters can contact me about job opportunities", should we remove this too as part of this PR? /cc @vaidehijoshi

I think we should go ahead and do it! Depending on how many users have this enabled on DEV, we could potentially mention it in a changelog like the looking for work feature, but something tells me not as many users will have this enabled on DEV :)

rhymes (approved):
LGTM

citizen428:
@vaidehijoshi I'll probably just merge this one first, then on Monday look into some stats via Blazer and do a follow-up PR.

db/schema.rb
```diff
@@ -1283,8 +1283,6 @@
     t.string "linkedin_url"
     t.string "location"
     t.datetime "locked_at"
-    t.boolean "looking_for_work", default: false
-    t.boolean "looking_for_work_publicly", default: false
```
vaidehijoshi, line 1287: This is my favorite kind of diff 🔥

vaidehijoshi (approved):
👏🏽 👏🏽

maestromac (commented):
There still is `looking_for_work` and `looking_for_work_publicly` in `User::PROFILE_COLUMNS`. Was that left out on purpose?

citizen428:
@maestromac Yes. To remove a column you first ignore it, remove the code using it, then remove the column from the ignore list. 


`User::PROFILE_COLUMNS` is used as an ignored list

PR #12174, opened 2021-01-08
Verify current password on password change

citizen428:

- [X] Refactor
- [X] Security

## Description

This PR modifies the password changing process to also require the current password as an additional security measure. This prevents account takeovers when a user forgets to lock their computer in a public place or accidentally stays logged in on a shared machine.

Note: as many of you know I'm not a big fan of custom controller actions, but refactoring `UsersController` is a major undertaking in itself and for now I just wanted to get this feature done.

https://github.com/forem/security/issues/351

**Current form**

image

**Updated form**

image

**QA instructions**

Go to `/settings/account` and try to change your password, both by providing an incorrect and correct current password.

None.

## Added tests?

- [X] Yes

## Added to documentation?

- [X] No documentation needed


app/controllers/users_controller.rb
```diff
@@ -252,6 +252,24 @@ def handle_settings_tab
     end
   end
 
+  def update_password
+    set_current_tab("account")
+
+    if @user.update_with_password(password_params)
```
citizen428, line 258: Most of this `if` is a copy/paste from `update`. I'd like to eventually refactor this whole controller, but until then this should do...

app/controllers/users_controller.rb
```diff
@@ -252,6 +252,24 @@ def handle_settings_tab
...
+      redirect_to "/settings/#{@tab}"
```
rhymes, line 259: non blocking: `user_settings_path(@tab)`

app/controllers/users_controller.rb
```diff
@@ -252,6 +252,24 @@ def handle_settings_tab
...
+    else
+      Honeycomb.add_field("error", @user.errors.messages.reject { |_, v| v.empty? })
+      Honeycomb.add_field("errored", true)
+
+      if @tab
+        render :edit, status: :bad_request
+      else
+        flash[:error] = @user.errors.full_messages.join(", ")
```
rhymes, line 267: `.errors_as_sentence` ?

app/controllers/users_controller.rb
```diff
@@ -252,6 +252,24 @@ def handle_settings_tab
...
+        redirect_to "/settings"
```
rhymes, line 268: `user_settings_path`

rhymes (commented):
I tried this locally but, since I'm logged in with GitHub I do not know my password, and since the first password for those who sign in with OAuth is auto generated a bunch of users might not know their passwords.

Should we add a link to the `forgot password` page if we add this requirement?

citizen428:
> Should we add a link to the forgot password page if we add this requirement?

Absolutely, will set this back to draft so it doesn't accidentally get merged over the weekend. Thanks!

citizen428:
@rhymes The problem is that the forgot password page does not work for already signed-in users. 😢 Try to sign in and then go to `/users/password/new`, you'll be redirected to `/?signin=true`. 

I see the following options:

1. Give up on this PR.
2. Sign out the user and make them reset the password (this seems like a horrible UX for user's already signed in through social media).
3. Dig deeper into Devise to see what's possible (I only spent a few minutes on this now).

What do you think?

rhymes:
That looks like the Devise intended behavior @citizen428.

According to https://github.com/heartcombo/devise/blob/45b831c4ea5a35914037bd27fe88b76d7b3683a4/app/controllers/devise/passwords_controller.rb#L3-L4 they require no authentication but if you dig in https://github.com/heartcombo/devise/blob/45b831c4ea5a35914037bd27fe88b76d7b3683a4/app/controllers/devise_controller.rb#L102 they redirect already authenticated users to the homepage because in the default scenario an already authenticated user doesn't need to reset their password. 

I found an old issue - https://github.com/heartcombo/devise/issues/3580#issuecomment-95944905 - where they suggest to apply the same technique we've applied for all the other Devise customization: inherit and customize. In this case I think we just need to disable the "before action"

Also https://twitter.com/account/begin_password_reset works regardless or login status, so we might want to change that to be the same, at least in my mind it makes sense that I should be able to reset my password if I don't remember it, regardless if I'm already signed in, especially in a scenario with multiple ways of signin in as we have.

Devise default behavior might makes sense for a website that only has email/password, not for something that uses OmniAuth.

An alternative: we could add this "existing password" requirement only for users that (in `OR`):

- have changed the password at least once (do we have that info?)
- don't have third party identities and are signed in through email

but that seems like a hack.

Another possible alternative would be to let the field as it is there and change the error message upon inputing the wrong existing password to nudge them towards signing out and clicking forgot password but sounds like a terrible UX experience anyway.

citizen428:
> That looks like the Devise intended behavior,

Yes, I should have been more clear about this: adding the forgot password link is a good idea for users who never set one, but it won't work because of Devise and I should have remembered that.

> don't have third party identities and are signed in through email

I briefly considered this and thought it's pretty inconsistent.

> have changed the password at least once (do we have that info?)

This could actually be somewhat feasible, but `password_changed_at` at is only available if you use the `password_expirable`  module from [Devise Security Extension](https://github.com/phatworx/devise_security_extension).

> inherit and customize

I'll look into this tomorrow, it is the most robust approach and you already did the digging I tried to avoid. Thanks for that!

citizen428:
@rhymes I pushed up something quick, can you please test it a bit?

app/views/users/_account_set_password.html.erb
```diff
@@ -4,10 +4,14 @@
         Set new password
       </h2>
       <p class="color-base-70">
-        As an alternative to signing in via linked social accounts, you may create a password.
+        As an alternative to signing in via linked social accounts, you may create a password. If you signed up with a social media account, please <a href="<%= new_user_password_path %>">reset your password</a> first.
```
rhymes, line 7: we should tell them to reset the password using their primary email, as they might have several :(

rhymes (approved):
I was able to use "forgot password" to set a password I remembered, then I went to account to change it to another one and it worked. There are no visibile notices, but everything works

citizen428:
@rhymes 

> There are no visibile notices, but everything works

I think that's at least partially due to Devise's paranoid mode that we enabled not too long ago.

app/views/users/_account_set_password.html.erb
citizen428: @rhymes I made the message more explicit now, what do you think?

image

atsmith813 (approved):
Tested locally, works great 🚀

atsmith813:
> There are no visibile notices, but everything works

Which notices are missing? I believe Devise uses different errors in its views/layouts so it may be we're calling the wrong one...or not explicitly calling the Devise methods to show the message.

citizen428:
@atsmith813 

> Which notices are missing? I believe Devise uses different errors in its views/layouts so it may be we're calling the wrong one

Again, Devise's paranoid mode swallows a lot of the Devise flash messages to not accidentally leak information about accounts that could be used for enumeration purposes. 

I resolved the merge conflict in this PR now, so once the specs are green again I think this should be ready.

PR #12185, opened 2021-01-08
Add new cache key for article sidebar & bottom_content

Zhao-Andy:
- [x] Feature
- [x] Bug Fix
- [x] Optimization

## Description
This adds a new concept of busting caches for an author's cached content areas, specifically the article's sidebar "More from [author]" and bottom content "Read next" areas. The idea here is to rely more on busting the timestamp `user/organization.latest_article_updated_at`, and therefore can lead to less stale content (broken links, incorrect titles and tags, usernames, etc.). For now, this simply clears the cached areas when a title and tag list is updated. 

I realize this won't cover all cases, but I think this is a good start, and I've also been sitting on this code for too long. Would definitely appreciate feedback here!

Resolves https://github.com/forem/internalEngineering/issues/242

1. Enable caching in development: `rails dev:cache`
2. Mock the articles that will be viewed in:
- `_sticky_nav`: return early in `get_user_stickies.rb, call` method (line 4) with `return author.articles.published.last(3)`
- `_bottom_content`: replace line 210 in `app/views/articles/show.html.erb` with `suggested_articles = (@organization || @user).articles.published.last(3)`
3. Create and publish a new article via `/new`
4. - [ ] See that it shows up in the sidebar and bottom content areas
5. Update the created article's title and tags
6. - [ ] See that the new changes are reflected in the sidebar and bottom content areas

Nope! This doesn't change any front end code.

## Added tests?
- [x] I need help with writing tests. Not sure how to write a test for caching.

## Added to documentation?
- [x] No documentation needed

We will have to monitor our alerting, and maybe turn it off for a bit. The last time we touched `/articles/_sticky_nav.html.erb` the alerting went off a lot, but the site didn't slow down much. cc @maestromac here.

I'm unsure :(


db/migrate/20201208195636_add_latest_article_updated_at_to_users_and_orgs.rb
```diff
@@ -0,0 +1,6 @@
+class AddLatestArticleUpdatedAtToUsersAndOrgs < ActiveRecord::Migration[6.0]
+  def change
+    add_column :users, :latest_article_updated_at, :datetime
+    add_column :organizations, :latest_article_updated_at, :datetime
```
Zhao-Andy, line 4: 1. Not my favorite column name. Taking suggestions :)
2. Not sure if I need to add anything else for this to be a safe migration.

app/models/article.rb
```diff
@@ -664,11 +665,20 @@ def clean_data
     self.canonical_url = nil if canonical_url == ""
   end
 
+  def touch_actor_latest_article_updated_at(destroying: false)
```
Zhao-Andy, line 663: Also not a fan of this method name.

app/javascript/packs/dashboardTagsDisableUnchangedButtons.js
```diff
@@ -1,35 +1,36 @@
-document.getElementById('follows_update_form').addEventListener('submit', checkChanged);
```
citizen428, line 1: Should this file be in here? It mostly seems like formatting changes, no actual code changes? Probably better to do this in a separate cleanup PR.

app/javascript/shared/components/tags.jsx
```diff
@@ -335,7 +335,9 @@ class Tags extends Component {
       // updates searchResults array according to what is being typed by user
       // allows user to choose a tag when they've typed the partial or whole word
       this.setState({
-        searchResults: response.result.filter((t) => !this.selected.includes(t.name)),
```
citizen428, line 338: Same questions as above.

app/views/dashboards/following_tags.html.erb
```diff
@@ -25,7 +25,7 @@
         <%= form_with url: bulk_update_follows_path, method: :patch, local: true, id: "follows_update_form" do |f| %>
           <button type="submit" class="crayons-btn crayons-btn--m mb-3" name="commit">Update Weights</button>
         <% end %>
-
```
citizen428, line 28: This seems like a another change for a separate cleanup PR.

db/migrate/20201208195636_add_latest_article_updated_at_to_users_and_orgs.rb
maestromac: 1. I can't think of a better column name either.
2. I think it should be okay since you don't have a default value
 
Thinking out loud here: `user.latest_article_updated_at` is going to be nil, because there's no default value. That shouldn't cause any issue though since each cache usage includes Article's ID.

maestromac (commented):
I think there's a typo in the direction. Can you elaborate on this step? which file and what line
> _sticky_nav: return early in sticky_article_collection.rb, user_stickies method with return author.articles.published.last(3)

Zhao-Andy:
@maestromac ahh hah the file was deleted when I merged with latest and I didn't realize. Updated the step in the original comment. You can add `return author.articles.published.last(3)` in `get_user_stickies.rb` line 4.

atsmith813 (approved):
I tested this locally and it worked 👍

maestromac (approved):
LGTM!

citizen428 (approved):

PR #12191, opened 2021-01-09
Bug Fix:Fix development.rb to Correctly Detect Rails command

mstruve:
- [x] Bug Fix

## Description
ENV["COMMAND"] no longer seems to be present when a rails command is run. I am not sure when this changed but I have updated our code to use a different method to detect a console or server start in order to prevent it from succeeding if there are data update scripts that need running. 

https://github.com/orgs/forem/projects/6#card-48564710

1. Open a console locally
2. run `DataUpdateScript.last.destroy`
3. close the console
4. Try to reopen the console or start a rails server. Both should fail.
5. run `bundle exec bin/setup`
6. You should be able to open a console and start a server now.

## Added tests?
- [x] No, minor update for local env

alt_text

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -128,7 +128,7 @@ Rails.application.configure do
     Bullet.add_whitelist(type: :n_plus_one_query, class_name: "User", association: :profile)
 
     # Check if there are any data update scripts to run during startup
-    if %w[c console runner s server].include?(ENV["COMMAND"]) && DataUpdateScript.scripts_to_run?
+    if %w[Console Server].any? { |const| Rails.const_defined?(const) } && DataUpdateScript.scripts_to_run?
       message = "Data update scripts need to be run before you can start the application. " \
         "Please run 'rails data_updates:run'"
       raise message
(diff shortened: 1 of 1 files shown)


benhalpern (approved):

config/environments/development.rb
```diff
@@ -128,7 +128,7 @@
     Bullet.add_whitelist(type: :n_plus_one_query, class_name: "User", association: :profile)
 
     # Check if there are any data update scripts to run during startup
-    if %w[c console runner s server].include?(ENV["COMMAND"]) && DataUpdateScript.scripts_to_run?
+    if %w[Console Server].any? { |const| Rails.const_defined?(const) } && DataUpdateScript.scripts_to_run?
```
citizen428, line 131: Nice solution. Maybe we want to also include `DBConsole` here, which drops the user right into a Postgres shell.

citizen428 (approved):

rhymes (approved):

config/environments/development.rb
mstruve: Did not know about that one, I will add it! Thanks!

PR #12202, opened 2021-01-11
Move HexComparer to Hex::Comparer service

atsmith813:
- [x] Refactor

## Description
This PR refactors `HexComparer` out of `/labor` into `Hex::Comparer` in `/services`.

I found myself really liking the current setup using `Hex::Comparer.new` and then calling methods on that. If others feel differently and want to try a different approach (i.e. breaking up methods into individual services and use `.call`, I'm open to that). I found this current setup to be readable and many methods reuse the same private `hex_to_rgb_hash` - making it convenient to leave "as-is". Let me know what you think! 😄 

_I explicitly call out the small changes I made below in PR comments 🙃 ._

## Added tests?
- [x] No, and this is why: we already have tests in place for this service.

## Added to documentation?
- [x] No documentation needed

power_rangers_pose_gif


app/helpers/application_helper.rb
```diff
@@ -2,7 +2,7 @@ module ApplicationHelper
   # rubocop:disable Performance/OpenStruct
   DELETED_USER = OpenStruct.new(
     id: nil,
-    darker_color: HexComparer.new(bg: "#19063A", text: "#dce9f3").brightness,
+    darker_color: Hex::Comparer.new(["#19063A", "#dce9f3"]).brightness,
```
atsmith813, line 5: To be honest...I'm not sure how this was working before 🤔 . I updated the syntax.

app/services/hex/comparer.rb
```diff
@@ -0,0 +1,50 @@
+module Hex
+  class Comparer
+    RGB_REGEX = /^#?(?<r>..)(?<g>..)(?<b>..)$/.freeze
+    ACCENT_MODIFIERS = [1.14, 1.08, 1.06, 0.96, 0.9, 0.8, 0.7, 0.6].freeze
+
+    def initialize(hexes)
+      @hexes = hexes.sort
+    end
```
atsmith813, line 8: I removed the `amount` argument because we aren't using it anywhere in the code. Instead, we've been passing the amount argument to the specific methods themselves. In other words, we do:

```ruby
Hex::Comparer.new([enriched_colors[:bg], enriched_colors[:text]]).brightness(adjustment)
```

Instead of:
```ruby
Hex::Comparer.new([enriched_colors[:bg], enriched_colors[:text]], adjustment).brightness
```

app/services/hex/comparer.rb
```diff
@@ -0,0 +1,50 @@
...
+      smallest
+    end
+
+    # Returns the first valid hex string it finds (# + 6 digits)
+    def accent(amount = 1)
+      ACCENT_MODIFIERS.each do |modifier|
+        with_brightness = brightness(modifier**amount)
+        break with_brightness if with_brightness.size == 7
+      end
+    end
```
atsmith813, line 33: I added the `amount` argument here since I removed it from `initialize`. We don't currently modify this value in the code, but it is referenced in this method. I figured it didn't hurt to leave the flexibility in place since it's already here, even if we aren't using it right now.

app/helpers/application_helper.rb
vaidehijoshi: ....I just re-read `HexComparer` twice to figure out how this would have functioned with named arguments and....I am guessing it just didn't work 😓

vaidehijoshi (approved):

vaidehijoshi:
> _I explicitly call out the small changes I made below in PR comments 🙃 ._

Just wanted to say that I found this strategy really helpful! Thank you :)

app/helpers/application_helper.rb
rhymes: maybe we can move those two colors as a frozen array

app/services/hex/comparer.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    def biggest
+      hexes.last
+    end
+
+    def brightness(amount = 1)
+      rgb = hex_to_rgb_hash(smallest).transform_values do |color|
+        (color * amount).round
+      end
+      format("#%<r>02x%<g>02x%<b>02x", rgb)
```
rhymes, line 22: we could freeze this format string

app/services/hex/comparer.rb
```diff
@@ -0,0 +1,50 @@
...
+    def accent(amount = 1)
+      ACCENT_MODIFIERS.each do |modifier|
+        with_brightness = brightness(modifier**amount)
+        break with_brightness if with_brightness.size == 7
+      end
+    end
+
+    def opacity(value = 0.0)
+      rgba = hex_to_rgb_hash(smallest).merge(a: value)
+      format("rgba(%<r>d, %<g>d, %<b>d, %<a>.2f)", rgba)
```
rhymes, line 37: we could freeze this format string too

rhymes (commented):
Definitely simplified!

One thing: as we should follow the established pattern in naming, instead of `something-er` I think this should be renamed as other services to something more self explanatory: `Color::CompareHex` (we already have the `/color/` folder).

This also follows the guideline of using an action as the name of the service

atsmith813:
@rhymes good idea wrt renaming - updated!

rhymes (approved):
LGTM!

PR #12205, opened 2021-01-11
Move MailchimpBot to Mailchimp::Bot service

atsmith813:
- [x] Refactor

## Description
This PR refactors `MailchimpBot` out of `/labor` into `Mailchimp::Bot` in `/services`. I chose to leave the setup "as-is" with some minor updates. I'm open to other approaches if folks have ideas! My thinking was mostly based on 1) there is a lot of overlap of methods and 2) it's nice to have Mailchimp "contained" in one spot. There have been comments in the past about one day moving to a different email provider. If we ever did that, we could create a new service for the provider and have it live side by side with Mailchimp while we make the transition.

As noted below, there are async workers that call the old/original `MailchimpBot` so we can't delete it in this PR because when we deploy this there could be workers in the queue still referencing it. I'll create a separate PR after this one is deployed to remove it.

## Added tests?
- [x] No, and this is why: I updated the existing specs

## Added to documentation?
- [x] No documentation needed

monkey_reading_newspaper_gif


spec/services/mailchimp/bot_spec.rb
```diff
@@ -1,33 +1,15 @@
...
-  let(:my_gibbon_client) { instance_double(FakeGibbonRequest) }
+  let(:my_gibbon_client) { class_double Gibbon::Request }
   let(:tag) do
-    create(:tag, name: "tagname", bg_color_hex: Faker::Color.hex_color, text_color_hex: Faker::Color.hex_color,
-                 supported: true)
+    create(:tag,
+           name: "tagname",
+           bg_color_hex: Faker::Color.hex_color,
+           text_color_hex: Faker::Color.hex_color,
+           supported: true)
```
atsmith813, line 12: Updated indentation to make this easier to read 🙃

spec/services/mailchimp/bot_spec.rb
```diff
@@ -1,33 +1,15 @@
...
-  end
-  # rubocop:enable Lint/UselessMethodDefinition
-end
-
-RSpec.describe MailchimpBot, type: :labor do
+RSpec.describe Mailchimp::Bot, type: :service do
   let(:user) { create(:user, :with_newsletters, :ignore_mailchimp_subscribe_callback) }
   let(:article) { create(:article, user_id: user.id) }
-  let(:my_gibbon_client) { instance_double(FakeGibbonRequest) }
+  let(:my_gibbon_client) { class_double Gibbon::Request }
```
maestromac, line 6: :star_struck:

maestromac (approved):
LGTM!

app/workers/users/subscribe_to_mailchimp_newsletter_worker.rb
```diff
@@ -6,7 +6,7 @@ class SubscribeToMailchimpNewsletterWorker
     def perform(user_id)
       user = User.find_by(id: user_id)
 
-      MailchimpBot.new(user).upsert if user
+      Mailchimp::Bot.new(user).upsert if user
```
atsmith813, line 9: Since this is called asynchronously, we can't really delete the old/original `MailchimpBot` in this PR because when we deploy this there could be workers in the queue still referencing it. I'll create a separate PR after this one is deployed to remove it.

maestromac: You could keep the file with empty methods hanging around. You could also stub the name somewhere :laughing: 
```
MailchimpBot = Mailchimp::Bot
```
Got the idea from sidekiq https://github.com/mperham/sidekiq/wiki/FAQ#how-do-i-safely-rename-a-worker

vaidehijoshi (approved):
🚀 🎸

rhymes (commented):
Would it make sense to split this by action or it's that going too far?

Something like `Mailchimp::Bot::Upsert.call`, Mailchimp::Bot::ManageCommunityModerator.call` ?

ps. you can just alias the old worker with the new name, see @maestromac comment here https://github.com/forem/forem/pull/12205/files#r555382183

benhalpern (approved):

atsmith813:
@rhymes 

> Would it make sense to split this by action or it's that going too far?

I think it's going too far. I started on that path and ended up going this way in the end. My thinking was mostly based on 1) there is a lot of overlap of methods and 2) it's nice to have Mailchimp "contained" in one spot. There have been comments in the past about one day moving to a different email provider. If we ever did that, we could create a new service for the provider and have it live side by side with Mailchimp while we make the transition. There are also cases where reusing the same instance of `Mailchimp::bot` which sets up the `Gibbon` client could make sense in the future. It's definitely possible to split it out...we'd move the `initialize` logic and some methods to some `base` class that the services would then have to inherit from.

atsmith813:
@rhymes @maestromac 

> ps. you can just alias the old worker with the new name, see @maestromac comment here https://github.com/forem/forem/pull/12205/files#r555382183

That won't work. That would be to rename the worker itself. The trick on this PR is that the worker (`SubscribeToMailchimpNewsletterWorker`) calls the old class `MailchimpBot` inside of it. So if any of these workers are queued on deployment, they'll fail because `MailchimpBot` won't be defined. Even when I did:

```ruby
MailchimpBot = Mailchimp::Bot
```

And then in a console, we still get `uninitialized constant MailchimpBot`.

I think this approach would work if we were trying to change the name of the _worker_ itself and not a call inside of it.

rhymes:
Thanks @atsmith813 for the explanations!

PR #12211, opened 2021-01-11
Encode emails properly before sending to /confirm-email

Zhao-Andy:
- [x] Bug Fix

## Description
Before, some emails were not encoded properly, `andy+hi@forem.com`. This PR ensures that the email is always encoded before redirecting, so that symbols like `+` are not scrubbed accidentally.

Closes https://github.com/forem/internalEngineering/issues/105

1. Visit localhost:3000/confirm-email?email=some+email@test.com
2. Check that it properly displays on the page and form

Nope!

## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

Nope

### DIFF
--- a/spec/system/authentication/user_logs_in_with_email_spec.rb
+++ b/spec/system/authentication/user_logs_in_with_email_spec.rb
@@ -32,6 +32,14 @@ RSpec.describe "Authenticating with Email" do
 
         expect(page).to have_current_path("/confirm-email", ignore_query: true)
       end
+
+      it "displays the properly decoded email" do
+        decoded_email = user.email.sub("@", "+something@")
+        user.email = decoded_email
+        sign_up_user
+
+        expect(page).to have_text(decoded_email)
+      end
     end
 
     context "when trying to register with an already existing email" do
--- a/spec/system/authentication/user_logs_in_with_email_spec.rb
+++ b/spec/system/authentication/user_logs_in_with_email_spec.rb
@@ -13,22 +13,22 @@ RSpec.describe "Authenticating with Email" do
     let(:user) { build(:user, saw_onboarding: false) }
 
     context "when using valid credentials" do
-      it "creates a new user", js: true do
-        expect do
-          visit sign_up_path(state: "new-user")
-          click_link(sign_up_link, match: :first)
-
-          fill_in_user(user)
-          click_button("Sign up", match: :first)
-        end.to change(User, :count).by(1)
-      end
-
-      it "logs in and redirects to email confirmation" do
+      def sign_up_user
         visit sign_up_path(state: "new-user")
         click_link(sign_up_link, match: :first)
 
         fill_in_user(user)
         click_button("Sign up", match: :first)
+      end
+
+      it "creates a new user", js: true do
+        expect do
+          sign_up_user
+        end.to change(User, :count).by(1)
+      end
+
+      it "logs in and redirects to email confirmation" do
+        sign_up_user
 
         expect(page).to have_current_path("/confirm-email", ignore_query: true)
       end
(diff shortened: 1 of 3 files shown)


spec/system/authentication/user_logs_in_with_email_spec.rb
```diff
@@ -13,25 +13,33 @@
...
-
-        fill_in_user(user)
-        click_button("Sign up", match: :first)
+        sign_up_user
 
         expect(page).to have_current_path("/confirm-email", ignore_query: true)
       end
+
+      it "displays the properly decoded email" do
+        decoded_email = user.email.sub("@", "+something@")
```
rhymes, line 37: `CGI.unescape` ?

rhymes (changes requested):
I think the display and the input box value of the email need to encoded too:

Screenshot 2021-01-12 at 10 12 16 AM

Screenshot 2021-01-12 at 10 12 45 AM

spec/system/authentication/user_logs_in_with_email_spec.rb
Zhao-Andy: Hmm sorry? `decoded_email` is to mock an email address that has the `+` in it, and `CGI.unescape` would remove that.

Zhao-Andy:
@rhymes ah that's a good point, but I think we don't need to encode that when the page is visited directly. I know I said to try that in my directions though 😅 Sorry about that. From what I can tell, the only time a user lands on `/confirm-email` is the redirect after registering. I might be wrong though, and if I am then yeah we would need to cover the case you mentioned.

I had trouble trying to do so earlier because by the time you're visiting the page, Rails has already `CGI.unescape`d the email param, and I'm not sure how to get to the email before it's touched. 😕

msarit (approved):
Hey Andy! great work! I experienced the same behavior as detailed by Rhymes, but other than that, things look good to me, so I'll go ahead and approve.

spec/system/authentication/user_logs_in_with_email_spec.rb
rhymes: ah sorry!

rhymes:
@Zhao-Andy not sure I have understood. I started the app, went to `http://localhost:3000/confirm-email?email=andy+hi@forem.com` and when I pressed "Resend confirmation instructions" I got the error, as the email with the space is not valid. As that button is in `app/views/devise/confirmations/new.html.erb`, can't we just unescape the `value` attribute of the input box?

Zhao-Andy:
@rhymes, we could, but the value is already `andy hi@forem.com` when you visit the page with unescaped URL param `?email=andy+hi@forem.com`, so encoding it would only encode the `whitespace`. Since the only place `/confirm-email` is being used is as a redirect registration, I think we only need to encode the URL param in the redirect location of `registrations_controller.rb`.

Sorry, I should update my directions to say:
1. visit `localhost:3000/confirm-email?email=andy%2Bhi%40forem.com`
2. See that the URL is displayed properly in the input and page

@msarit maybe you know since you've worked on this area, but would a user ever visit `/confirm-email` directly? Or is it only ever used as a redirect? Seems like the second from what I could find.

rhymes:
@Zhao-Andy can they land directly there from the confirmation email?

Zhao-Andy:
@rhymes Since we have no mailer view that has that link I assume no.

msarit:
> @msarit maybe you know since you've worked on this area, but would a user ever visit /confirm-email directly? Or is it only ever used as a redirect? Seems like the second from what I could find.

Hey @Zhao-Andy 👋🏾  yes, from what I know, `/confirm-email` is reached via redirect normally.

msarit:
@Zhao-Andy Is it normal that after clicking on "Resend email confirmation", I am redirected to http://localhost:3000/new?signin=true (the New Post page essentially)?

Zhao-Andy:
@msarit I think for whatever reason the resend confirmation is sending the user to the generic "registration complete" route, which can be the homepage or `/new`. Possibly a bug but I'll leave it for a different PR.

rhymes (approved):
LGTM

benhalpern (approved):

PR #12182, opened 2021-01-08
Change HEROKU_SLUG_COMMIT  to HEROKU_RELEASE_CREATED_AT 

Ridhwana:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Change related to https://github.com/forem/forem/pull/12149.

RELEASE_FOOTPRINT is usually set in the Forem Cloud env environment. However, we need to cater for those Forems that are being deployed on Heroku as well. 

RELEASE_FOOTPRINT is the datetime that the deploy is done, however the fallback variable for Heroku was set to HEROKU_SLUG_COMMIT. This causes the Last deployed time to be a commit id instead of a date when RELEASE_FOOTPRINT is not set (like on heroku envs). See the image below: 

 Screenshot 2021-01-08 at 17 35 39

Hence, I've changed the fallback to HEROKU_RELEASE_CREATED_AT. 

## Related Tickets & Documents
https://github.com/forem/forem/pull/12149

## QA Instructions, Screenshots, Recordings
1. In the .env file change the following variables as follows:
```
export HEROKU_RELEASE_CREATED_AT="Some Date"
export HEROKU_SLUG_COMMIT="Some Commit ID"
```

2. Now restart your server and go to http://localhost:3000/admin, you should see the correct fallback variable set.
Screenshot 2021-01-08 at 17 44 12

### UI accessibility concerns?

_If your PR includes UI changes, please replace this line with details on how
accessibility is impacted and tested. For more info, check out the
[Forem Accessibility Docs](https://docs.forem.com/frontend/accessibility)._

## Added tests?

- [ ] Yes
- [x] No, because its a fallback env variable change 
- [ ] I need help with writing tests

## Added to documentation?

- [ ] [Developer Docs](https://docs.forem.com) and/or
      [Admin Guide](https://forem.gitbook.io/forem-admin-guide/)
- [ ] README
- [x] No documentation needed

## [optional] What gif best describes this PR or how it makes you feel?

alt_text

PR #12224, opened 2021-01-12
Refactor deployed_at and latest_commit_id code

Ridhwana:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
I created two PR's https://github.com/forem/forem/pull/12149 and https://github.com/forem/forem/pull/12182 that worked on adding the Deployed at and the Latest commit ID to the admin interface. 

A [suggestion](https://github.com/forem/forem/pull/12149#discussion_r553075918) that came out of https://github.com/forem/forem/pull/12149 was to add a new model (not DB backed) like ForemInstance to which we add a bunch of class methods that encapsulate things that are used for Forem Instances and will change across the different instances. It made sense to me and so I've added that model. 

Aside: As per the comment, we could even add move dev_to? from SiteConfig to the new class since ForemInstance.dev_to? makes a lot more sense than SiteConfig.dev_to?. Same applied to the local? method for the same reason (ForemInstance.local?). This can be refactored and added at a later stage.

**TLDR:**

- This PR focuses on abstracting out the deployed_at and the lastest_commit_id into the ForemInstance model. 
- It also replaces the instances where we use `ENV['RELEASE_FOOTPRINT']` to use ForemInstace.deployed_at. 
- I've removed the set_release_key initializer in favor of the Forem Instance model which explicitly accesses the correct variable directly using ForemInstance Model 
- Finally, in the body we have some data keys that I rename to be more apt and inline with what these variables do

- https://github.com/forem/forem/pull/12149
- https://github.com/forem/forem/pull/12182

It's a bit difficult to test these changes since they swap out an ENV variable for the ForemInstance model method. Some of these changes are also on the caching methods. 

However, you can play aroudn with it by changing the env values, restarting your app and making sure that the data-deployed-at and the data-latest-commit-id update.

None.

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

Keep an eye out for any caching issues.

alt_text


app/controllers/async_info_controller.rb
```diff
@@ -29,7 +29,7 @@ def shell_version
     set_surrogate_key_header "shell-version-endpoint"
     # shell_version will change on every deploy.
     # *Technically* could be only on changes to assets and shell, but this is more fool-proof.
-    shell_version = ApplicationConfig["RELEASE_FOOTPRINT"] + SiteConfig.admin_action_taken_at.to_s
+    shell_version = ForemInstance.deployed_at + SiteConfig.admin_action_taken_at.to_s
```
Ridhwana, line 32: Using the ForemInstance.deployed_at method here as will set ApplicationConfig["RELEASE_FOOTPRINT"] if it exists.

app/helpers/application_helper.rb
```diff
@@ -174,7 +174,7 @@ def community_emoji
   end
 
   def release_adjusted_cache_key(path)
-    release_footprint = ApplicationConfig["RELEASE_FOOTPRINT"]
+    release_footprint = ForemInstance.deployed_at
```
Ridhwana, line 177: Using the newly created ForemInstance.deployed_at method here as will set ApplicationConfig["RELEASE_FOOTPRINT"] if it is present.

app/models/forem_instance.rb
```diff
@@ -0,0 +1,9 @@
+class ForemInstance
```
Ridhwana, line 1: A suggestion that came out of #12149 was to add a model (not DB backed) i.e. ForemInstance to which we add a bunch of class methods that encapsulate things that are used for Forem Instances and will change across the different instances. It made sense to start that model with a deployed_at and the latest_commit_id

app/models/forem_instance.rb
```diff
@@ -0,0 +1,9 @@
...
+  def self.deployed_at
+    ApplicationConfig["RELEASE_FOOTPRINT"].presence || ENV["HEROKU_RELEASE_CREATED_AT"].presence
```
Ridhwana, line 3: For the Forem.cloud environment RELEASE_FOOTPRINT will be set whereas for a Heroku environment HEROKU_RELEASE_CREATED_AT will be set

app/models/forem_instance.rb
```diff
@@ -0,0 +1,9 @@
...
+  end
+
+  def self.latest_commit_id
+    ApplicationConfig["FOREM_BUILD_SHA"].presence || ENV["HEROKU_SLUG_COMMIT"].presence
```
Ridhwana, line 7: For the Forem.cloud environment FOREM_BUILD_SHA will be set whereas for a Heroku environment HEROKU_SLUG_COMMIT will be set

app/views/admin/admin_portals/index.html.erb
```diff
@@ -2,11 +2,11 @@
   <div class="tag-card crayons-card p-4 m:p-6 m:pt-4 mb-4 flex flex-row relative">
     <div class="mr-8">
       <div class="crayons-field__description">Last deployed</div>
-      <div><%= ApplicationConfig["RELEASE_FOOTPRINT"].presence || ENV["HEROKU_RELEASE_CREATED_AT"].presence || "Not Available" %></div>
+      <div><%= ForemInstance.deployed_at || "Not Available" %></div>
     </div>
     <div>
       <div class="crayons-field__description">Latest Commit ID</div>
-      <div><%= ApplicationConfig["FOREM_BUILD_SHA"].presence || ENV["HEROKU_SLUG_COMMIT"].presence || "Not Available" %></div>
+      <div><%= ForemInstance.latest_commit_id || "Not Available" %></div>
```
Ridhwana, line 9: Change the display for the admin portal page to reuse the ForemInstance methos instance of explicitly referencing the variables

app/views/layouts/admin.html.erb
```diff
@@ -26,7 +26,8 @@
 </head>
 <body class="admin default-header"
   data-honeybadger-key="<%= ApplicationConfig["HONEYBADGER_JS_API_KEY"] %>"
-  data-release-footprint="<%= ApplicationConfig["RELEASE_FOOTPRINT"] %>"
```
Ridhwana, line 29: **Confirmation Required** I'm not 100% sure if this key being referenced by an external app, so I need some confirmation here that it is fine to change the data key name. 

It made more sense to name it in accordance to what it references and consistent with the rest of the app. I've also added the latest commit id to the data keys on the body.

config/initializers/set_release_footprint.rb
```diff
@@ -1,2 +0,0 @@
-# We will set RELEASE_FOOTPRINT in our Forem Cloud environment, or use HEROKU_SLUG_COMMIT if set (e.g. Heroku env)
-ENV["RELEASE_FOOTPRINT"] ||= ENV["HEROKU_RELEASE_CREATED_AT"]
```
Ridhwana, line 2: **Confirmation required:** I removed this because we have explicitly set the variables in places where we use them. I could totally be missing an important part of the code, hence it would be helpful to get other developer eyes on this.

Ridhwana:
This PR touches some of the caching stuff by using the new ForemInstance model. It's a pretty simple change and I've outlined some of my rationale for making some changes but would really appreciate a couple of reviews on this PR to ensure that I'm not missing a critical piece of information and possibly breaking some caching and other related stuff.

Ridhwana:
Update: I'm looking into the failing test.

app/models/forem_instance.rb
rhymes: Since both are taken from static values that can never change during the lifetime of the app (you literally need to restart the app to change environment variables), I think the result of the `OR` can be frozen into a constant. We don't need to call a method each time:

`ForemInstance::DEPLOYED_AT` <- this value should never change during the lifetime of the deployed version

app/models/forem_instance.rb
rhymes: same thing about moving this to a frozen constant

rhymes (commented):
Left notes about moving method calls to constants for constant values, I think `dev_to?` and `local?` should be methods instead

app/models/forem_instance.rb
Ridhwana: ✅

app/models/forem_instance.rb
Ridhwana: ✅

config/initializers/set_release_footprint.rb
mstruve: seems like dead code to me.

mstruve (approved):
This is fantastic!!!

app/models/forem_instance.rb
citizen428: Tbh, I still prefer the method, we could always cache it. Frequently reaching into constants defined by other models seems a bit off to me.

```ruby
def self.deployed_at
  @deployed_at ||=  ApplicationConfig["RELEASE_FOOTPRINT"].presence || ENV["HEROKU_RELEASE_CREATED_AT"].presence
end
```

Ridhwana: I spent time considering the pro's and cons of each approach, and I don't think theres a right or wrong answer here (especially using the cached method strategy), but it's rather a matter of preference. 

I had put this PR on hold yesterday evening after struggling with testing the variables. After discussion with Rhymes, we had concluded that it was testing essentially an "or" statement and so it would be relatively safe to remove the tests if we weren't able to stub it out.

With a fresh mind this morning I tackled it again and after reading the comment above I've decided to move it back to a method with caching. This will help us to keep these tests in place as well.

Ridhwana:
Moved this PR from Draft to Ready for Review. Addressed all feedback and fixed the tests. One thing to watch out for when merging and deploying this PR is any cache issues.

rhymes (approved):
LGTM

PR #11332, opened 2020-11-09
🚀 Feature: Ability to report a message in connect.

sarthology:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
The fastest implementation of this feature request that I can think of:
- Now, one can report a message from others.
- And Block them also after reporting.
- Custom Report for admin to see the connect feedbacks.

## QA Instructions, Screenshots, Recordings
Screen Recording 2020-11-09 at 11 35 AM

## Added tests?

- [ ] Yes
- [x] No, and this is why: _Will add soon_
- [ ] I need help with writing tests

## Added to documentation?

- [ ] Docs.forem.com
- [ ] README
- [x] No documentation needed

##  What gif best describes this PR or how it makes you feel?

alt_text

PR #12229, opened 2021-01-12
Report a Message in Connect

nickytonline:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

A user can now report abuse in Connect. Most of this work was done in #11332. I'm just closing the loop on this.

A screenshot of the new report abuse feature in Connect

A screenshot of the new report abuse feature in Connect

One question I have @citizen428 since I don't know what the original specs were for this was. You can report a message, but if you try to report a second message from that user, it says that an abuse submission has already been filed. Is this how the implementation should work? I understand if they block the user as well that maybe it doesn't make sense to report another message, but when just reporting the message, I'm not sure.

This builds off the work in #11332. 

1. Log in as the development mode main admin (email: admin@forem.local, password: password)
2. Make the main admin a tag moderator, for example, for the discuss tag, http://localhost:3000/admin/tags/13/edit
	1. Add the main admin as a tag moderator (their ID is 11)
	2. Click the Add Moderator button to apply the change. (This is required to be able to create new channels)
	3. Go back to Connect, localhost:3000/connect
	4. Create a channel and invite a user to the channel by clicking on the <kbd>+</kbd> icon.

image

	5. The create channel modal will open. Enter a channel name and the username of another user to invite.

image

	6. In the newly created channel, create a message.
1. In another browser (private browser mode or a different browser), log on as the other user. All test users have a password of password. If you need to find the users email address to logon, go into the rails console and search for that user.
2. Once logged on as the other user. Accept the invitation to the new connect channel by clicking on the people icon with a badge notification.

image

3. In the chat channel, hover over the message of the other user. The three dots contextual menu will appear. Hover over it to report abuse.

image

image

4. Select an abuse option. This is required to submit the form. Otherwise the report message button will stay disabled.
5. Click on the report message button. A confirmation window will notify you that the message will be reported and it will also ask you if you’d like to block the user.
	1. If you choose to block the user, the message will be reported and the user will be blocked from interacting with you on the platform.
	2. If you choose not to block the user, only the message will be reported.
6. Note, if you try to report a message from that user again, you will be notified that the report has already been submitted.

image

The three dot contextual menu is not accessible. However, I filed is a separate issue, #12250, as this is more of a global issue in Connect.

image

For this feature, the report abuse panel works fine with keyboard navigation and has focus styles. For the radio buttons we're using the [focus visible polyfill](https://www.npmjs.com/package/focus-visible).

image

The report message button appears to be using the default browser styles. This could be something to improve upon for our button component. cc: @lisasy @ludwiczakpawel.

image

Also screenshare from #11332 

Screen Recording 2020-11-09 at 11 35 AM

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

alt_text


nickytonline:
The code climate issues are not relevant. UI code repeats itself sometimes.

nickytonline:
One question I have @citizen428 since I don't know what the original specs were for this was. You can report a message, but if you try to report a second message from that user, it says that an abuse submission has already been filed. Is this how the implementation should work? I understand if they block the user as well that maybe it doesn't make sense to report another message, but when just reporting the message, I'm not sure.

app/javascript/chat/__tests__/reportAbuse.test.jsx
```diff
@@ -19,24 +19,26 @@ describe('<ReportAbuse />', () => {
   });
 
   it('should render the component', () => {
-    const { queryByText, getByTestId } = render(
```
nickytonline: `getByTestId` was not necessary. We can leverage `getByLabelText` which also encourages a11y.

aitchiss (commented):
I'm not totally sure how the block functionality should work, but when I was testing this, I blocked after reporting (and saw the correct confirmation) but the offender was still able to continue the chat session - is that what we expect?

Also I noticed looking at this PR that everything we display in `sidecar-content` (line 55 of `content.jsx` is wrapped in `aria-hidden="true"`, meaning it's hidden from assistive technology. Some browsers will still announce interactive items as you tab through, but e.g. VoiceOver in Safari will ignore everything we display in there, meaning AT users can't report abuse.

I think refactoring that would fit best with #12250 though, since at the moment a lot of AT users are not going to be able to open that sidecar content until that's fixed anyway. What do you think?

app/javascript/chat/content.jsx
```diff
@@ -80,13 +88,16 @@ export default class Content extends Component {
...
+          resource={resource}
+          closeReportAbuseForm={closeReportAbuseForm}
+        />
       </div>
     );
   }
 }
 
-const Display = ({ resource }) => {
+function Display({ resource, closeReportAbuseForm }) {
```
katiedavis, line 100: Just out of curiosity, did we update this to `function` because it needs to be hoisted? Or is it a style/preference thing?

app/views/admin/feedback_messages/_feedback_message.html.erb
```diff
@@ -52,6 +72,14 @@
               <%= feedback_message.message %>
             <% end %>
           </p>
+          <% else %>
+            <h5 class="fw-bold">
+            Message from Offender:
+          </h5>
+          <div class="reported__message">
+                 <%= raw(feedback_message.message) %>
```
katiedavis, line 80: The formatting on this file is kinda wonky. I'm not super familiar with rubocop/ruby linting but is there an option to lint the `erb` files?

app/views/admin/feedback_messages/_style.html.erb
```diff
@@ -22,4 +22,21 @@
     color: black;
     text-align: center;
   }
+  .badge-connect {
```
katiedavis, line 25: Same here, do our pre-commit hooks format our css?

app/javascript/chat/ReportAbuse/index.jsx
```diff
@@ -0,0 +1,148 @@
...
+      data.user_id,
+    );
+    const { success, message } = response;
+    if (success) {
+      const confirmBlock = window.confirm(
+        `The message will be reported.\n\nWould you like to block this person as well?\n\nThis will:
+      - prevent them from commenting on your posts
+      - block all notifications from them
+      - prevent them from messaging you via chat`,
+      );
```
aitchiss, line 46: during testing I found this part of the flow not very intuitive - because the confirm popup only surfaces OK/cancel buttons it feels a bit like we're asking the user whether they want to go ahead with the report or cancel it (rather than 'OK yes please block' vs 'no thanks'.

Maybe not in design/scope for this ticket, I think it would be cool if this question was part of the main report form so we don't need the popup at all (future optimisation?)

app/javascript/chat/__tests__/reportAbuse.test.jsx
```diff
@@ -0,0 +1,44 @@
+import { h } from 'preact';
+import { render } from '@testing-library/preact';
+import { axe } from 'jest-axe';
+import ReportAbuse from '../ReportAbuse';
+
+describe('<ReportAbuse />', () => {
```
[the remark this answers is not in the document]
katiedavis: I'm really curious about this as well, previously I have handled those kind of actions in jest, but that was honestly due to a lack of an E2E testing framework. But...my gut says this is a great candidate for an E2E test, yeah!

nickytonline:
> I'm not totally sure how the block functionality should work, but when I was testing this, I blocked after reporting (and saw the correct confirmation) but the offender was still able to continue the chat session - is that what we expect?
> 
> 
> 
> Also I noticed looking at this PR that everything we display in `sidecar-content` (line 55 of `content.jsx` is wrapped in `aria-hidden="true"`, meaning it's hidden from assistive technology. Some browsers will still announce interactive items as you tab through, but e.g. VoiceOver in Safari will ignore everything we display in there, meaning AT users can't report abuse.
> 
> 
> 
> I think refactoring that would fit best with #12250 though, since at the moment a lot of AT users are not going to be able to open that sidecar content until that's fixed anyway. What do you think?

@aitchiss, there are lots of accessibility bugs in Connect. I think it makes sense to do this in #12250 or potentially another PR.

app/javascript/chat/content.jsx
nickytonline: Just a preference. We don't have a hat and fa at rule for this. It might make sense to add one to avoid debates and let the robots manage it.

app/views/admin/feedback_messages/_feedback_message.html.erb
nickytonline: Rubocop and I believe erblint handle this. We have a precommit hook that does this.

app/javascript/chat/ReportAbuse/index.jsx
nickytonline: I think that's a great suggestion that we could improve in another PR. I have no specs for this PR, so I think it was what they thought was good.

app/views/admin/feedback_messages/_style.html.erb
rhymes: I don't think it works in CSS blocks inside ERB files. Only in (S)CSS standalone files

nickytonline: What @rhymes said @katiedavis.

app/javascript/chat/message.jsx
```diff
@@ -60,6 +61,22 @@ const Message = ({
       </div>
     </div>
   );
+  const dropdownReport = (
+    <div className="message__actions">
+      <span className="ellipsis__menubutton">
+        <img src={ThreeDotsIcon} alt="dropdown menu icon" />
```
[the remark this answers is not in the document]
nickytonline: Great suggestion @aitchiss.

app/javascript/chat/__tests__/reportAbuse.test.jsx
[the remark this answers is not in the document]
nickytonline: Yeah, once I've wrapped up the Cypress data cleaning, this definitely merits an E2E test. It's a more long term thing, but it'd be good to map out key flows in the app that we want to add end to end tests for. cc: @vaidehijoshi .

app/javascript/chat/ReportAbuse/index.jsx
```diff
@@ -0,0 +1,148 @@
...
+    <Fragment>
+      <section className="mt-7 p-4 grid gap-2 crayons-card mb-4">
+        <h1 className="lh-tight mb-4 mt-0">Report Abuse</h1>
+        <p>
+          Thank you for reporting any abuse that violates our{' '}
+          <a href="/code-of-conduct">code of conduct</a> or{' '}
+          <a href="/terms">terms and conditions</a>. We continue to try to make
+          this environment a great one for everybody.
+        </p>
+        <fieldset className="report__abuse-options p-4 justify-between">
```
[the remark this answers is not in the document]
nickytonline: Yeah, the <legend /> makes more sense given it's a <fieldset />. I'll make the change.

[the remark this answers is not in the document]
nickytonline: Done. I didn't touch any styling for it, as I don't think there was even design for this.

An image of the report abuse form on Forem's Connect

app/javascript/chat/ReportAbuse/index.jsx
nickytonline: Mind creating an issue for this @aitchiss and link it to this PR? Thanks!

nickytonline:
> I'm not totally sure how the block functionality should work, but when I was testing this, I blocked after reporting (and saw the correct confirmation) but the offender was still able to continue the chat session - is that what we expect?
> 

@aitchiss that doesn't sound like it should be expected, but like I mentioned up top, I do not have any specs for this PR that I took over. Deferring to @citizen428.

app/javascript/chat/__tests__/reportAbuse.test.jsx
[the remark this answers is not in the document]
vaidehijoshi: > It's a more long term thing, but it'd be good to map out key flows in the app that we want to add end to end tests for. 

@nickytonline Yeah, this is certainly on my mind; specifically, I've been chatting with @lisasy about creating a full representation of _all_ the flows in our app, just from a _product_ perspective. 

@citizen428 actually gave me the idea of laying out every single screen and state that exists in a Forem, and using that to get a sense of how far-reaching the tentacles of our product really are. The benefit of this being, of course, that those flows literally map to the E2E flows that we'd want to test across the platform.

Suffice it to say, it will be somewhat of an undertaking, and step 1 is just figuring out what the flows are. But it would take us towards this ultimate north start of E2E tests across the key flows of the app, for sure :)

nickytonline:
The code climate issues are not an issue. It's complaining about checkboxes having similar attributes. The only somewhat valid one is code climate complaining that a function is 53 lines instead of 50, but with formatting, this seems like a false positive to me.

app/javascript/chat/actions/requestActions.js
```diff
@@ -92,3 +92,52 @@ export async function updateMembership(membershipId, userAction) {
 
   return response.json();
 }
+
+/**
+ *
+ * @param {string} feedback_message
+ * @param {string}  type_of_feedback
```
citizen428, line 99: **nitpick (non-blocking):** There seems to be an extra space before this param name and the names not being properly aligned massively triggers my OCD 😉 

```suggestion
 * @param {string} type_of_feedback
```

citizen428 (approved):

katiedavis (approved):
It seems like we're tracking the a111y issues elsewhere and this works as expected. LG2M!

PR #12236, opened 2021-01-12
[Travis] Create after_deploy Honeycomb script

maestromac:
- [x] DX

## Description
This will add a marker to our Honeycomb board whenever the app is deployed to Heroku. It provides additional context used for debugging and analytics.

[Honeycomb's documentation](https://docs.honeycomb.io/api/markers/#create-a-marker)

Resolves https://github.com/orgs/forem/projects/6#card-52224759

This is a curl command. I've tested it thoroughly locall.y If it doesn't work, it wouldn't cause any problem and I'll address it immediately.

here's what the feature looks like. It's the dotted line.
image)

## Added tests?

## Added to documentation?
- [x] no documentation needed

n/a

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,6 +83,8 @@ script:
 after_script:
   - ./cc-test-reporter format-coverage -t simplecov -o ./coverage/codeclimate.$KNAPSACK_PRO_CI_NODE_INDEX.json ./coverage/spec/.resultset.json
   - ./cc-test-reporter sum-coverage --output - --parts $COVERAGE_REPORTS_TOTAL coverage/codeclimate.*.json | ./cc-test-reporter upload-coverage --input -
+after_deploy:
+  - bash scripts/after_deploy.sh
 notifications:
   slack:
     if: branch = master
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -110,6 +112,7 @@ jobs:
       if: type != pull_request
       script: skip
       after_script: skip
+      after_deploy: skip
```
maestromac, line 115: Skip after_deploy for BenHalpern Community deploy.

mstruve (approved):
In the future, we should figure out how this could work for all Forems but I am game for at least doing this for now for DEV. 

Given this is not used by Heroku and we are having slug issues, should we add this file to slugignore?

rhymes (approved):
+1 on adding it to slug ignore though it might merit a separate PR as I believe a bunch of stuff in `scripts/` can go in the slug ignore as well

andygeorge (approved):
Seems reasonable so :+1:, although @forem/systems doesn't manage Heroku/Travis so I don't have a super strong opinion on this (until, as Molly referenced, we're using Honeycomb for other Forems).

PR #12237, opened 2021-01-12
API: Endpoint to get an organization's articles

twinsfan421:

- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This would provide an api endpoint dev.to/api/organizations/{org_username}/articles to get a list of an organization's articles.
As discussed in issue 9212, This is the last of the 4 related endpoints.

#9212 #10931 #11514 #11823

## Added tests?

- [X] Yes
      have not been included_

## Added to documentation?

- [X] [Developer Docs](https://docs.forem.com) and/or
- [ ] README
- [ ] No documentation needed


app/controllers/api/v0/organizations_controller.rb
```diff
@@ -21,6 +21,14 @@ class OrganizationsController < ApiController
...
       private_constant :LISTINGS_FOR_SERIALIZATION
 
+      ARTICLES_FOR_SERIALIZATION = %i[
+        id user_id organization_id collection_id
+        title description main_image published_at crossposted_at social_image
+        cached_tag_list slug path canonical_url comments_count
+        public_reactions_count created_at edited_at last_comment_at published
+        updated_at video_thumbnail_url
+      ].freeze
+
```
rhymes, line 31: I wonder if, instead copying `Api::V0::ArticlesController::INDEX_ATTRIBUTES_FOR_SERIALIZATION` here, we can't just refer to it in the controller, this way we make sure that the fields in both endpoints are always aligned.

app/views/api/v0/articles/index.json.jbuilder
```diff
@@ -1,5 +1,5 @@
 json.array! @articles do |article|
-  json.partial! "article", article: article
+  json.partial! "api/v0/shared/article", article: article
```
rhymes, line 2: ❤️

app/views/api/v0/organizations/articles.json.jbuilder
```diff
@@ -0,0 +1,14 @@
+json.array! @articles do |article|
```
rhymes, line 1: can't we just render the `index.json.jbuilder` file from the articles API in the action of this? They are identical except for the if, which in this case it will always be true :-)

rhymes (commented):
Thank you @twinsfan421!

app/controllers/api/v0/organizations_controller.rb
twinsfan421: I updated this

app/views/api/v0/articles/index.json.jbuilder
twinsfan421: I changed it back since I deleted the articles jbuilder file

app/views/api/v0/organizations/articles.json.jbuilder
twinsfan421: Thanks for reviewing. I changed it to render the index jbuilder.

rhymes (approved):
Great job @twinsfan421, thank you!

Ridhwana (approved):
This looks really good. Thank you @twinsfan421!

PR #12241, opened 2021-01-12
Break MarkdownFixer into MarkdownProcessor::Fixer services

atsmith813:
- [x] Refactor

## Description
This PR refactors `MarkdownFixer` in `/labor` into separate services under `MarkdownProcessor::Fixer` in `/services`. I chose this route because each action we call in the fixer is really just calling an order of operations out of a shared group of methods. See PR comments below for more specifics.

_I refactored `MarkdownParser` into `MarkdownProcessor::Parse` in [a separate PR](https://github.com/forem/forem/pull/12245) so we can keep all markdown related services under the `/services/markdown_processor` folder instead of `/services`. We can't use plain `/markdown`/`Markdown` because it conflicts._

None of these services are called in an async job so we shouldn't need to worry about that.

## Added tests?
- [x] Yes

## Added to documentation?
- [x] No documentation needed

the_sink_doesnt_fix_itself_gif


app/services/markdown_processor/fixer/base.rb
```diff
@@ -0,0 +1,95 @@
...
+    #
+    # For example
+    # METHODS = %i[add_quotes_to_tile add_quotes_to_description]
+    class Base
+      FRONT_MATTER_DETECTOR = /-{3}.*?-{3}/m.freeze
+
+      def self.call(markdown)
+        return unless markdown
+
+        self::METHODS.reduce(markdown) { |acc, elem| public_send(elem, acc) }
```
atsmith813, line 16: This "approach" is similar to how to we handle the interface with Elasticsearch under `/search` where each class defines a set of constants and the `base` class handles the logic.

app/services/markdown_processor/fixer/base.rb
```diff
@@ -0,0 +1,95 @@
...
+      def self.add_quotes_to_description(markdown)
+        add_quotes_to_section(markdown, section: "description")
+      end
+
+      # This turns --- into ------- after the first two,
+      # because --- messes with front matter
+      def self.modify_hr_tags(markdown)
+        markdown.gsub(/^---/).with_index { |match, i| i > 1 ? "#{match}-----" : match }
+      end
+
```
atsmith813, line 32: We didn't have any specs for this method and when I went to write them...I realized this method may have actually not been working this entire time 🙈 . Since we don't currently have any specs for it, my understanding of what this is supposed to do is based on the comment here. I updated this logic, please let me know if I misunderstood the purpose of this method!

app/services/markdown_processor/fixer/base.rb
```diff
@@ -0,0 +1,95 @@
+module MarkdownProcessor
```
atsmith813, line 1: We could probably make many of these methods here private...but I like that we explicitly test them. I'm open to suggestions about this!

app/services/markdown_processor/fixer/base.rb
citizen428: I had the same concern about the search classes, IMHO the list of methods should come from a method, even if that just wraps the constant. Methods are much easier to override etc., also while testing.

```suggestion
        fix_methods.reduce(markdown) { |acc, elem| public_send(elem, acc) }
```

```ruby
def fix_methods
  METHODS
end
```

app/services/markdown_processor/fixer/fix_all.rb
```diff
@@ -0,0 +1,17 @@
...
+        add_quotes_to_title
+        add_quotes_to_description
+        lowercase_published
+        modify_hr_tags
+        convert_new_lines
+        split_tags
+        underscores_in_usernames
+      ].freeze
+
+      # #call is implemented in MarkdownProcessor::Fixer::Base
```
rhymes, line 14: I would remove these comments, as it's clear the class inherits from something

rhymes (approved):
Not a superfan of testing private stuff but we all do it sometimes, me incuded, and given how complicated is this logic, better safe than sorry.

app/services/markdown_processor/fixer/base.rb
atsmith813: Good idea, updated!

citizen428 (approved):

PR #12255, opened 2021-01-13
Update I18n default

citizen428:

- [X] Refactor

## Description

When bundling the app I noticed the following output:

> HEADS UP! i18n 1.1 changed fallbacks to exclude default locale.
> But that may break your application.
> 
> If you are upgrading your Rails application from an older version of Rails:
> 
> Please check your Rails app for 'config.i18n.fallbacks = true'.
> If you're using I18n (>= 1.1.0) and Rails (< 5.2.2), this should be
> 'config.i18n.fallbacks = [I18n.default_locale]'.
> If not, fallbacks will be broken in your app by I18n 1.1.x.
> 
> If you are starting a NEW Rails application, you can ignore this notice.

We apparently hadn't updated this setting yet, so this PR takes care of that.

n/a

n/a

n/a

## Added tests?

- [X] No, and this is why: nothing specific to test here

## Added to documentation?

- [X] No documentation needed

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -89,7 +89,7 @@ Rails.application.configure do
 
   # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
   # the I18n.default_locale when a translation cannot be found).
-  config.i18n.fallbacks = true
+  config.i18n.fallbacks = [I18n.default_locale]
 
   # Send deprecation notices to registered listeners.
   config.active_support.deprecation = :notify


rhymes (approved):
LGTM

Ridhwana (approved):
LGTM

citizen428:
I just noticed that we may not actually need this, due to `If you're using I18n (>= 1.1.0) and Rails (< 5.2.2)`. Our Rails version is new enough. What do you think @rhymes and @Ridhwana, should we go ahead with it or not?

Ridhwana:
> I just noticed that we may not actually need this, due to `If you're using I18n (>= 1.1.0) and Rails (< 5.2.2)`. Our Rails version is new enough. What do you think @rhymes and @Ridhwana, should we go ahead with it or not?

My opinion is that we don't merge the change if our versions are more recent those in the warning, since they do explicitly say to ignore the notice in that case.

rhymes:
I would default to what Rails is doing, and Rails master generates this:

https://github.com/rails/rails/blob/291a3d2ef29a3842d1156ada7526f4ee60dd2b59/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt#L83-L85

```ruby
  # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
  # the I18n.default_locale when a translation cannot be found).
  config.i18n.fallbacks = true
```

I think we can skip this then

maestromac:
Seeing that there's a discussion going, i'm going to mark this as draft to avoid accidental merge.

citizen428:
Yup, that's what I thought, should have caught that when originally reading the message. Closing.

PR #12263, opened 2021-01-13
Workaround for Apple auth problems (still in beta)

fdoxyz:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

#11934 added SIWA (Sign in with Apple) in beta mode (feature flag). We found a problem that made it impossible to connect existing accounts with SIWA ([more info here](https://forem.team/fdoxyz/samesite-cookies-and-siwa-39if)).

#12114 is currently blocked due to that problem so we are going to wait until we can sort that out before taking SIWA out of beta.

This PR adds a workaround to avoid showing the option to connect existing accounts with SIWA in order to enable it for DEV, until we are able to finish #12114 

#11934

#12114 

Apple auth workflow remains unchanged (available in https://community.benhalpern.com) but now the button to connect an existing account shouldn't be visible in `/settings`. In order to check this you can try out these steps:

1. Enable Twitter/GitHub/Facebook auth
2. Enable Apple auth (you can add gibberish to the keys that's not a problem)
3. While unauthenticated you should see the OAuth providers (including Apple although it won't work for you locally)
4. Authenticate with `admin@forem.local` and go to `/settings`
5. You should be able to see the "Connect with X" for the non-beta Auth Providers and SIWA should not be visible here
6. If you authenticate with all OAuth providers then no option should be visible here

None

## Added tests?

- [ ] Yes
- [x] No, and this is why: The existing tests around authentication providers remain unchanged (only fixed one that fails until we take Apple auth out from beta)

## Added to documentation?

- [ ] README
- [x] No documentation needed

None

not yet

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -19,7 +19,7 @@ APPLE_OMNIAUTH_SETUP = lambda do |env|
   env["omniauth.strategy"].options[:client_id] = SiteConfig.apple_client_id
   env["omniauth.strategy"].options[:scope] = "email name"
   env["omniauth.strategy"].options[:key_id] = SiteConfig.apple_key_id
-  env["omniauth.strategy"].options[:pem] = SiteConfig.apple_pem
+  env["omniauth.strategy"].options[:pem] = SiteConfig.apple_pem.to_s.gsub("\\n", "\n")
   env["omniauth.strategy"].options[:provider_ignores_state] = true
   env["omniauth.strategy"].options[:team_id] = SiteConfig.apple_team_id
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -531,7 +531,10 @@ class User < ApplicationRecord
   end
 
   def authenticated_with_all_providers?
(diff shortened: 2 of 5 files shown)


config/initializers/devise.rb
```diff
@@ -19,7 +19,7 @@
   env["omniauth.strategy"].options[:client_id] = SiteConfig.apple_client_id
   env["omniauth.strategy"].options[:scope] = "email name"
   env["omniauth.strategy"].options[:key_id] = SiteConfig.apple_key_id
-  env["omniauth.strategy"].options[:pem] = SiteConfig.apple_pem
+  env["omniauth.strategy"].options[:pem] = SiteConfig.apple_pem.to_s.gsub("\\n", "\n")
```
fdoxyz, line 22: The backslashes are escaped when stored in SiteConfig. This fixes that problem and makes it possible to add the private key via the Admin page

benhalpern (approved):
This makes sense to me

rhymes (approved):
😭 LGTM

PR #12266, opened 2021-01-13
Improve Error Messaging for Self-Deletion Confirmation Flow

juliannatetreault:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
**The issue**:
> Occasionally, users are unable to delete their own DEV accounts. Sometimes the link provided to them will lead to a 404. 

**The solution**:

> This PR improves the self-deletion error messaging in `Users::Controller#confirm_destroy` to mitigate confusion when self-deleting an account.

To accomplish this, this PR does the following:
- Reworks the code within `#confirm_destroy` to be more consistent with the other destroy-related (`#request_destroy` and `#full_delete`) methods by mirroring how the error message is set and displayed (`flash[:settings_notice]`).
- Adds `confirm_destroy` to the exception in the `verify_authorized!` `after_action`.
- Adds a conditional to `#confirm_destroy` that sets the `current_user` and does one of two things: authorizes the user or alerts the user that, "You must log in to proceed with account deletion." and redirects the user to the "Sign In" page.
- Adds a conditional that checks whether the `destroy_token` is `blank` or not. If the `destroy_token` is `blank`, then a message is displayed, alerting the user that their token has expired. Otherwise, it raises an error due to a `destroy_token` mismatch.
- Adds tests around all edge cases to ensure that the proper error messages are raised under the proper circumstances.

Relates to issue [374](https://github.com/forem/internalEngineering/issues/374)

⚠️ **Heads-up**: caching is not enabled by default in development. In order to properly QA these changes, you must first enable caching. To enable and disable caching in development, run the following command: `Rails dev:cache`.

🥇 **Pro-tip**: to make QAing these changes easier, add a delimiter to your console to easily search for the email containing the token and/or have `development.log` open so that you can search for the email containing the token by. You can search for the delimiter that you added (I use `------` or `~~~~~~`) or the word `mime` to locate the email containing the token once you have clicked the "Delete Account" button (see QA instructions below for further context!).

**To test the self-deletion flow as a logged-in user with a valid token, please do the following**:
- Log in and navigate to `/settings/account`. Click the "Delete Account" button found at the bottom of the page:
Screen Shot 2021-01-13 at 2 48 46 PM
 
- You should then be redirected back to `settings/account` where you should see a message confirming that you've requested an account deletion:
Screen Shot 2021-01-14 at 8 46 49 AM

- Copy the link containing the `destroy_token` from your logs and paste it into your browser. You should be taken to the "Destroy your account" page as expected. 🛑 **If you choose to destroy your account at this time, you will need to log in as a different user to QA the rest of this PR**:
Screen Shot 2021-01-13 at 2 48 25 PM

**Before QAing the invalid token flow, please read the following heads-up**:

⚠️ **Heads-up**: In order to test the invalid token edge case, please delete the key in a console like so (you will need to know the user's ID in order to delete the key!):
```
[2] pry(main)> Rails.cache.delete("user-destroy-token-11")
Cache delete: user-destroy-token-11
=> 1
```
I have also been setting `destroy_token` within `UsersController#confirm_destroy` equal to an empty string only for testing purposes--this mimics a `.blank` token:
```
  def confirm_destroy
    destroy_token = ""
```

**To test the self-deletion flow as a logged-in user with an invalid token, please do the following**:
- Log in and navigate to `/settings/account`. Click the "Delete Account" button found at the bottom of the page:
Screen Shot 2021-01-13 at 2 48 46 PM

- You should then be redirected back to `/settings/account` where you should see a message confirming that you've requested an account deletion:
Screen Shot 2021-01-14 at 8 46 49 AM

- Copy the link containing the `destroy_token` from your logs and paste it into your browser. Rather than being taken to the "Destroy your account" page, a flash message should appear, stating: "Your token has expired, please request a new one. Tokens only last for 12 hours after account deletion is initiated." You should be redirected back to `/settings/account`:
Screen Shot 2021-01-14 at 10 57 05 AM

⚠️ **Heads-up**: In order to test the invalid token edge case, please delete the key in a console like so (you will need to know the user's ID in order to delete the key!):
```
[2] pry(main)> Rails.cache.delete("user-destroy-token-11")
Cache delete: user-destroy-token-11
=> 1
```
I have also been setting `destroy_token` within `UsersController#confirm_destroy` equal to an empty string only for testing purposes--this mimics a `.blank` token:
```
  def confirm_destroy
    destroy_token = "mismatched token"
```

**To test the self-deletion flow as a logged-in user with a mismatched token, please do the following**:
- Log in and navigate to `/settings/account`. Click the "Delete Account" button found at the bottom of the page:
Screen Shot 2021-01-13 at 2 48 46 PM

- You should then be redirected back to `/settings/account` where you should see a message confirming that you've requested an account deletion:
Screen Shot 2021-01-14 at 8 46 49 AM

- Copy the link containing the `destroy_token` from your logs and paste it into your browser. Rather than being taken to the "Destroy your account" page, you should hit a `404` and an `ActionController::RoutingError, "Not Found"` error should be raised in your console:
Screen Shot 2021-01-14 at 11 06 49 AM
Screen Shot 2021-01-14 at 11 07 23 AM

**To test the self-deletion flow as a logged-out user, please do the following**:
- In an incognito window, copy the link containing the `destroy_token` from your logs and paste it into your browser. Rather than being taken to the "Destroy your account" page, you should be redirected to the "Sign In" page and a flash message should appear, stating: "You must be logged in to proceed with account deletion.":
Screen Shot 2021-01-14 at 10 59 04 AM

N/A

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

N/A

Delete Button


app/controllers/users_controller.rb
```diff
@@ -254,6 +261,17 @@ def handle_settings_tab
...
+  def set_or_sign_in_user
+    @user = current_user
+
+    if @user
+      authorize @user
+    else
+      flash[:alert] = "You must be logged in to proceed with account deletion."
+      redirect_to new_user_session_path
+    end
+  end
```
juliannatetreault, line 273: I have tried rewriting this a handful of times now and have spent a lot of time attempting to move this redirect into `#confirm_destroy` and _out_ of this `before_action` and have been unsuccessful in my attempts. I am open to suggestions if anyone has ideas for how to get this to work within `#confirm_destroy` rather than in the `before_action`, `set_or_sign_in_user`!

vaidehijoshi: What happens when you move this logic into that action? Do you run into errors in your server logs, or does it just not work correctly?

At first glance, my thought is that it might be due to multiple redirects happening within the same controller, but I'm not sure if you're getting an error message or not

juliannatetreault: @vaidehijoshi the most common error I get (I've been playing around with the logic within the action itself a lot 😅 ) is a `500` `Pundit::AuthorizationNotPerformedError `:
```
Completed 500 Internal Server Error in 3ms (ActiveRecord: 0.0ms | Allocations: 2366)

Pundit::AuthorizationNotPerformedError - UsersController:
  app/middlewares/set_cookie_domain.rb:13:in `call'

```
When I trace it back in my logs, I see that I _am_ redirected, but cannot get past the error above. It looks like it's an issue stemming from `authorize @user` and I'm unsure of get past it other than through the code I currently have (using the `before_action`).

app/controllers/users_controller.rb
```diff
@@ -254,6 +261,17 @@ def handle_settings_tab
...
   private
 
+  def set_or_sign_in_user
+    @user = current_user
+
+    if @user
+      authorize @user
+    else
+      flash[:alert] = "You must be logged in to proceed with account deletion."
+      redirect_to new_user_session_path
```
fdoxyz, line 271: I think this should be `sign_up_path` instead, right? At least it's the one I get on `http://localhost:3000/rails/info/routes` to match the `/enter` path.

Also a small heads up: If you're moving this block into the action don't forget the `return` statement after the redirect, because the action won't stop executing after a redirect 🙂

fdoxyz (commented):
I gave this a try and it's working well. Great job in diving into this and thank you for the detailed QA instructions, they were very helpful!

I left a comment that may help you move the the before_action into the action itself though

app/controllers/users_controller.rb
vaidehijoshi: Agree!! This is what I was thinking of when I left my earlier comment about redirects :) if you have more than one, you'll have to return out of them explicitly

app/controllers/users_controller.rb
```diff
@@ -93,9 +94,15 @@ def request_destroy
...
 
-    set_current_tab("account")
+    # rubocop:disable Layout/LineLength
+    if destroy_token.blank?
+      flash[:settings_notice] = "Your token has expired, please request a new one. Tokens only last for 12 hours after account deletion is initiated."
+      redirect_to user_settings_path("account") and return
+    end
+    # rubocop:enable Layout/LineLength
+
+    raise ActionController::RoutingError, "Not Found" unless destroy_token == params[:token]
```
rhymes, line 105: Makes sense, to avoid using the `and return` we might want to restructure the if (haven'te tested) into:

```ruby
if destroy_token.blank?
  redirect
elsif destroy_token != params[:token]
  raise
end

rhymes (commented):
Cool work @juliannatetreault! 

Left a non blocking idea

app/controllers/users_controller.rb
juliannatetreault: When I had tried this previously, Rubocop did _not_ like the conditional written that way, but I just tried this again and it worked, so I've made this change! 😊

app/controllers/users_controller.rb
juliannatetreault: Thank you, @fdoxyz -- works like a charm. 😊

rhymes (approved):

juliannatetreault:
Thanks for pointing me towards a better solution, y'all! I've made the following changes and can confirm that everything still works as expected and would appreciate if you all could give the PR another go: 
- removed the custom `before_action` from the controller
- moved the logic from the custom `before_action` into the `confirm_destroy` action
- added `confirm_destroy` to the `:verify_authorized` `after_action`
- wrapped the `after_action` in `# rubocop:disable Layout/LineLength` (the line got too long!)
- refactored the conditional within `#confirm_destroy`

cc @fdoxyz @rhymes @vaidehijoshi 😊

fdoxyz (approved):
🚀

Issue #12138, opened 2021-01-05
Onboarding Email Opt-In Not Working


jessleenyc:
**Describe the bug**

My opt-in to receive the newsletter + digest is not saved during onboarding. I know this because I didn't receive a welcome email, Mailchimp said I was unsubscribed, and when I visit settings/notifications, it displays that I am not opted-in to either emails. In order to subscribe to both, I need to re-select the boxes from settings/notifications and save from there.

**To Reproduce**

1. visit /onboarding on a forem
2. opt-in to emails
3. go to settings/notifications
4. observe that you are not opted-in to emails

**Expected behavior**

I expect to be subscribed to emails when I opt-in to them.

**Additional context**

This was happening to me on a non-DEV forem. 
I believe it's also happening on other forems: https://creators.forem.com/lee/mailchimp-unsubscribed-via-admin-action-2pp8

msarit:
@jessleenyc Hi Jess! 👋🏾 

So I tried to reproduce the error described in this bug, but I couldn't. Here is a Loom video showing my experience:

https://www.loom.com/share/610c9f72e81d41cfb854d146b72b6bcb

leewynne:
> @jessleenyc Hi Jess! 👋🏾
> 
> So I tried to reproduce the error described in this bug, but I couldn't. Here is a Loom video showing my experience:
> 
> https://www.loom.com/share/610c9f72e81d41cfb854d146b72b6bcb

Hi there, just to confirm - I also get this bug on my forem.

jessleenyc:
@msarit so I just tried your flow on community.codenewbie.org and it is definitely still happening to me. Do you mind trying on that specific Forem? I

benhalpern:
When I go through onboarding veeerrrryyyy slowwwwwwlyyyyyy, it works... But when I go through fast, it does not work as expected...

I'm not sure what the problem is, but I suspect there is something going on with `update_onboarding` which might cause a problem... And potentially a scenario where this response kicks something which undoes the next slide in a sort of worker race condition.

<img width="750" alt="Screen Shot 2021-01-10 at 12 44 45 PM" src="https://user-images.githubusercontent.com/3102842/104130954-a15ae380-5341-11eb-9df0-5645e0057ebf.png">

This definitely needs further exploration.

vaidehijoshi:
Something interesting is definitely going on here 🤔  I tried this on forem.team and can confirm that it does not work when you click through unless you move at a _glacial pace_.

[...]

PR #12267, opened 2021-01-13
Log User Save Error Messages When it Fails

mstruve:
- [x] Debugging

## Description
This is a follow up to #12252 after discovering the following:

If `email_digest_periodic` and `email_newsletter` are both false I have found that those traces do not have update commands. Rails is smart enough to figure out that if nothing on the model is being updated then it doesnt hit the database with an update command.
```
# UPDATE
[14] pry(main)> u.assign_attributes(email_digest_periodic: false)
=> nil
[15] pry(main)> u.save
   (0.3ms)  BEGIN
  BanishedUser Exists? (0.3ms)  SELECT 1 AS one FROM "banished_users" WHERE "banished_users"."username" = $1 LIMIT $2  [["username", "mstruve"], ["LIMIT", 1]]
  User Update (0.6ms)  UPDATE "users" SET "email_digest_periodic" = $1, "updated_at" = $2 WHERE "users"."id" = $3  [["email_digest_periodic", false], ["updated_at", "2021-01-13 21:26:35.351201"], ["id", 13]]
   (0.7ms)  COMMIT
=> true

# NOT UPDATING ANYTHING
[16] pry(main)> u.assign_attributes(email_digest_periodic: false)
=> nil
[17] pry(main)> u.save
   (0.3ms)  BEGIN
  BanishedUser Exists? (0.4ms)  SELECT 1 AS one FROM "banished_users" WHERE "banished_users"."username" = $1 LIMIT $2  [["username", "mstruve"], ["LIMIT", 1]]
   (0.4ms)  COMMIT
=> true
```

That still leaves us with those times when we send `email_digest_periodic`  or `email_newsletter`  true. In the majority of those cases, I see the SQL Update in Honeycomb. BUT I found one where that was not the case bc the update was rolled back. I do not know why we could not save the user but the goal of this is to find out. 

## Related Issue
https://github.com/forem/forem/issues/12138

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -164,7 +164,6 @@ class UsersController < ApplicationController
 
   def onboarding_checkbox_update
     # TODO: mstruve will remove once debugging is done
-    Rails.logger.error("onboarding_checkbox_update_params:#{params}")
     Rails.logger.error("onboarding_checkbox_update_user_params:#{params[:user]}")
 
     if params[:user]
@@ -176,7 +175,15 @@ class UsersController < ApplicationController
 
     current_user.saw_onboarding = true
     authorize User
-    render_update_response(current_user.save)
+
+    # TODO: mstruve will remove once debugging is done
+    result = current_user.save
+    unless result
+      errors = current_user.errors.full_messages.join(", ")
+      Rails.logger.error("onboarding_checkbox_update_errors:#{errors}")
+    end
+
+    render_update_response(result)
   end
 
   def join_org


app/controllers/users_controller.rb
```diff
@@ -164,7 +164,6 @@ def onboarding_update
 
   def onboarding_checkbox_update
     # TODO: mstruve will remove once debugging is done
```
vaidehijoshi, line 166:
```suggestion

```
Since you added a reminder `TODO` on line 179, probably could remove this one?

atsmith813 (approved):
🕵️

vaidehijoshi (approved):

app/controllers/users_controller.rb
mstruve: I figured a TODO for each set of lines to make sure I dont forget any. Better safe than sorry 🤷‍♀️

vaidehijoshi: Whatever works best to remind you works for me! 👍🏽

Issue #11602, opened 2020-11-24
Followup to renaming "banned" as "suspended" in Admin


msarit:
Re: Michael's suggestion [here](https://github.com/forem/forem/pull/11581#issuecomment-732534235), this issue seeks to complete the work started in [this PR](https://github.com/forem/forem/pull/11581).

Tasks to be completed in code:
- Update `banned` to `suspended` in `role.rb`
-  Update all usages of `banned` to the new term `suspended` throughout codebase

In production, would need to update all roles as follows:

```ruby
Role.find_by(name: "banned").update(name: "suspended")
Role.find_by(name: "comment_banned").update(name: "comment_suspended")
```

msarit:
need to update this loop in `app/views/admin/users/_activity.html.erb`:

```ruby
<% @user.roles.each do |role| %>
   <% if role.name == "banned" %>
      <li>suspended</li>
   <% else %>
      <li><%= role.name %> <em><%= role.resource_name.to_s %></em></li>
   <% end %>
<% end %>
```

msarit:
I made several attempts to tackle this issue, but kept getting overwhelmed with testing all touched functionality and ensuring nothing broke 😓

jacobherrington:
I'm going to take a crack at this! Thanks for getting it documented and getting the ball rolling @msarit :+1:

PR #12270, opened 2021-01-14
Rename banned and comment_banned roles

jacobherrington:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Moving from "banned" terminology to "suspended."

This change helps reflect the suggestions in https://github.com/forem/internalEngineering/issues/298 in the code as a followup to https://github.com/forem/forem/pull/11581, which modified the same language in the UI.

Closes https://github.com/forem/forem/issues/11602

## Added tests?

- [ ] Yes
- [x] No, and this is why: no functionality has changed, existing tests should verify that things are working as intended

## Added to documentation?

- [ ] README
- [x] No documentation needed

Attached data update script, otherwise nothing.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -128,7 +128,7 @@ class ApplicationController < ActionController::Base
   end
 
   def raise_suspended
-    raise "SUSPENDED" if current_user&.suspended
+    raise SuspendedError if current_user&.suspended
   end
 
   def internal_navigation?
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -128,7 +128,7 @@ class ApplicationController < ActionController::Base
   end
 
   def raise_suspended
-    raise SuspendedError if current_user&.suspended
+    raise SuspendedError if current_user&.suspended?
   end
 
   def internal_navigation?
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -128,7 +128,7 @@ class ApplicationController < ActionController::Base
   end
 
   def raise_suspended
-    raise "SUSPENDED" if current_user&.banned
+    raise "SUSPENDED" if current_user&.suspended
   end
 
   def internal_navigation?
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -290,6 +290,7 @@ module ApplicationHelper
   end
 
   def role_display_name(role)
+    # TODO: [@jacobherrington] remove the check in this method
     role.name == "banned" ? "Suspended" : role.name.titlecase
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -290,7 +290,8 @@ module ApplicationHelper
   end
 
   def role_display_name(role)
-    # TODO: [@jacobherrington] remove the check in this method
+    # TODO: [@jacobherrington] After all Forems have successfully deployed and the banned role
+    # has been deleted, removed this ternary.
     role.name == "banned" ? "Suspended" : role.name.titlecase
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -374,7 +374,7 @@ class User < ApplicationRecord
   end
 
   def warned
-    has_role? :warned
+    has_role?(:warned)
   end
 
   def admin?
@@ -397,7 +397,7 @@ class User < ApplicationRecord
     return @trusted if defined? @trusted
 
     @trusted = Rails.cache.fetch("user-#{id}/has_trusted_role", expires_in: 200.hours) do
-      has_role? :trusted
+      has_role?(:trusted)
     end
   end
 
@@ -408,7 +408,7 @@ class User < ApplicationRecord
     end
   end
 
-  def comment_suspended
+  def comment_suspended?
     has_role?(:comment_suspended) || has_role?(:comment_banned)
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -370,6 +370,8 @@ class User < ApplicationRecord
   end
 
   def suspended?
+    # TODO: [@jacobherrington] After all of our Forems have been successfully deployed,
+    # and data scripts have successfully removed the banned role, we can remove `has_role?(:banned)`
     has_role?(:suspended) || has_role?(:banned)
   end
 
@@ -409,6 +411,9 @@ class User < ApplicationRecord
   end
 
   def comment_suspended?
+    # TODO: [@jacobherrington] After all of our Forems have been successfully deployed,
+    # and data scripts have successfully removed the comment_banned role,
+    # we can remove `has_role?(:comment_banned)`
     has_role?(:comment_suspended) || has_role?(:comment_banned)
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -371,7 +371,7 @@ class User < ApplicationRecord
 
   # methods for Administrate field
   def suspended
-    has_role? :suspended
+    has_role?(:suspended) || has_role?(:banned)
   end
 
   def warned
@@ -410,7 +410,7 @@ class User < ApplicationRecord
   end
 
   def comment_suspended
-    has_role? :comment_suspended
+    has_role?(:comment_suspended) || has_role?(:comment_banned)
   end
 
   def workshop_eligible?
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -370,8 +370,8 @@ class User < ApplicationRecord
   end
 
   # methods for Administrate field
-  def banned
-    has_role? :banned
+  def suspended
+    has_role? :suspended
   end
 
   def warned
@@ -409,8 +409,8 @@ class User < ApplicationRecord
     end
   end
 
-  def comment_banned
-    has_role? :comment_banned
+  def comment_suspended
+    has_role? :comment_suspended
   end
 
   def workshop_eligible?
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -369,8 +369,7 @@ class User < ApplicationRecord
     end
   end
 
-  # methods for Administrate field
-  def suspended
+  def suspended?
     has_role?(:suspended) || has_role?(:banned)
   end
 
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -71,8 +71,8 @@ class ApplicationPolicy
     user.has_role?(:support_admin)
   end
 
-  def user_is_banned?
-    user.banned
+  def user_is_suspended?
+    user.suspended
   end
 
   def user_is_trusted?
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -71,9 +71,7 @@ class ApplicationPolicy
     user.has_role?(:support_admin)
   end
 
-  def user_is_suspended?
-    user.suspended
-  end
+  delegate :suspended?, to: :user, prefix: true
 
   def user_is_trusted?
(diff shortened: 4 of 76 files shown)


Ridhwana:
Hoping https://github.com/forem/forem/pull/12294 will fix these flakey tests

app/controllers/application_controller.rb
```diff
@@ -114,7 +114,7 @@ def after_accept_path_for(_resource)
   end
 
   def raise_suspended
-    raise "SUSPENDED" if current_user&.banned
+    raise "SUSPENDED" if current_user&.suspended
```
citizen428, line 117: I know you just changed the method name here, but maybe now is a good time to move this from a `RuntimeError` with a message of `"SUSPENDED"` to an actual error (see `app/errors`), like e.g. 

```ruby
# app/errors/suspended_error.rb
class SuspendedError < StandardError; end
```

spec/models/role_spec.rb
```diff
@@ -8,7 +8,7 @@
   describe "::ROLES" do
     it "contains the correct values" do
       expected_roles = %w[
-        admin banned chatroom_beta_tester codeland_admin comment_banned podcast_admin pro restricted_liquid_tag
+        admin suspended chatroom_beta_tester codeland_admin comment_suspended podcast_admin pro restricted_liquid_tag
```
citizen428, line 11: **nitpick (non-blocking):** These were alphabetically sorted before, but changing `banned` to `suspended` changes that.

lib/data_update_scripts/20210119131811_replace_instances_of_banned_and_comment_banned.rb
```diff
@@ -0,0 +1,10 @@
+module DataUpdateScripts
+  class ReplaceInstancesOfBannedAndCommentBanned
+    def run
+      # Update names for the banned and comment_banned roles
+      # Details: https://github.com/forem/forem/pull/11581
+      Role.find_by(name: "banned").update(name: "suspended")
```
citizen428, line 6: I don't have banned users locally, so running this is a Rails console leads to an error:

```ruby
[3] forem(main)> Role.find_by(name: "banned").update(name: "suspended")
  Role Load (0.5ms)  SELECT "roles".* FROM "roles" WHERE "roles"."name" = $1 LIMIT $2  [["name", "banned"], ["LIMIT", 1]]
NoMethodError: undefined method `update' for nil:NilClass
from (pry):3:in `<main>'
```

I then ran the data update script and checked, its status is `failed`.

We should guard against this:

```suggestion
      Role.find_by(name: "banned")&.update(name: "suspended")
```

lib/data_update_scripts/20210119131811_replace_instances_of_banned_and_comment_banned.rb
```diff
@@ -0,0 +1,10 @@
...
+      Role.find_by(name: "comment_banned").update(name: "comment_suspended")
```
citizen428, line 7: Same as above:

```suggestion
      Role.find_by(name: "comment_banned")&.update(name: "comment_suspended")
```

citizen428 (changes requested):
Thanks for the effort, just a minor change required to make the data update script more robust.

msarit:
I'm reviewing this now but, @jacobherrington I just wanna say you are certified #BadAss for taking this on 😅  I'll be learning a lot from this PR 🙏🏾

msarit (commented):
Great work @jacobherrington! 👏🏾 

I just have a few comments but nothing blocking

app/models/user.rb
```diff
@@ -364,8 +364,8 @@ def cached_followed_tag_names
   end
 
   # methods for Administrate field
-  def banned
-    has_role? :banned
+  def suspended
+    has_role? :suspended
```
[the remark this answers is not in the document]
citizen428: @msarit @jacobherrington  In general, this shouldn't be a problem because `has_role?` performs a database lookup and the data update script updates the role name there. BUT: the scripts don't run during deployment, but get scheduled to run in ~5 minutes. During this short window the role would still be called `:banned` but checks would be performed against `:suspended`. So we should indeed keep it like this for the initial deployment but can remove it after.

app/queries/admin/moderators_query.rb
```diff
@@ -4,7 +4,7 @@ class ModeratorsQuery
       state: :trusted
     }.with_indifferent_access.freeze
 
-    VALID_ROLES = %i[banned warned trusted comment_banned].freeze
+    VALID_ROLES = %i[suspended warned trusted comment_suspended].freeze
```
rhymes, line 7: Since `user.suspended` still looks for the `banned` role in the DB during the transition, should this query still keep `banned` and `comment_banned` as valid roles and remove them in a following PR?

app/services/search/query_builders/user.rb
```diff
@@ -21,8 +21,8 @@ def initialize(params:)
 
         @params = params.deep_symbolize_keys
 
-        # default to excluding users who are banned
-        @params[:exclude_roles] = ["banned"]
+        # default to excluding users who are suspended
+        @params[:exclude_roles] = ["suspended"]
```
rhymes, line 25: Same thing. If "legacy" users still have the `banned` role, this will show them in the search. It's probably un unlikely scenario BUT if the data update script fails or if for whatever reason takes too long, we might show banned users in the search.

rhymes (commented):
One general question: is there anything in the various caches that we need to manually purge?

app/services/search/query_builders/user.rb
jacobherrington: We can leave them, then I will follow up with a PR to remove them down the road. Safety first :safety_pin:

jacobherrington:
@rhymes I don't think so, but I'm not confident speaking definitively on that :grimacing:

app/queries/admin/moderators_query.rb
citizen428: @rhymes See https://github.com/forem/forem/pull/12270#discussion_r562338386

citizen428:
@jacobherrington Can we get this unblocked? I guess resolving the merge conflicts would be a first step.

app/models/user.rb
```diff
@@ -364,8 +364,8 @@ def cached_followed_tag_names
   end
 
   # methods for Administrate field
```
citizen428, line 366: Unrelated, but since we're here already, should we maybe get rid of this comment, we no longer use Administrate.

app/models/user.rb
```diff
@@ -364,8 +364,8 @@ def cached_followed_tag_names
...
-  def banned
-    has_role? :banned
+  def suspended
```
citizen428, line 367: Maybe we should name this more idiomatically?

```suggestion
  def suspended?
```

app/models/user.rb
```diff
@@ -407,8 +407,8 @@ def moderator_for_tags
     end
   end
 
-  def comment_banned
-    has_role? :comment_banned
+  def comment_suspended
```
citizen428, line 410: Same suggestion as above:

```suggestion
  def comment_suspended?
```

app/policies/comment_policy.rb
```diff
@@ -65,8 +65,8 @@ def user_is_moderator?
     user.moderator_for_tags.present?
   end
 
-  def user_is_comment_banned?
-    user.has_role? :comment_banned
+  def user_is_comment_suspended?
+    user.has_role? :comment_suspended
```
citizen428, line 69: Let's be a bit more consistent about the style we use:

```suggestion
    user.has_role?(:comment_suspended)
```

app/policies/application_policy.rb
```diff
@@ -71,8 +71,8 @@ def support_admin?
     user.has_role?(:support_admin)
   end
 
-  def user_is_banned?
-    user.banned
+  def user_is_suspended?
```
citizen428, line 74: Nitpick: generally it's a bit unidiomatic to use `is` in predicate methods in Ruby, `user_suspended?` would be the more common form. But that would lead to a lot of extra changes, so is probably not worth it.

```suggestion
  def user_suspended?
```

mstruve:
Hey all, noticed this PR has been open for a bit without much activity. Any blockers I can help with?

mstruve:
Hey @jacobherrington, this has been sitting open for a bit. Do you think you could finish it up and get it merged? If there are any blockers you are hitting please feel free to reach out so I, or someone else on the team, can help unblock you.

app/policies/application_policy.rb
jacobherrington: Unfortunately, we sorta do this everywhere :laughing: I'll fix this instance in this PR and follow up with a 15-minute fix for a few other places that I've seen it!

jacobherrington:
@rhymes Manually testing this with the cache setting enabled on my local machine seems to work as expected.

jacobherrington:
@mstruve Thanks for the ping, I consistently forget that this PR exists. It is ready for review if you'd like to take a look at it :smile:

lib/data_update_scripts/20210119131811_replace_instances_of_banned_and_comment_banned.rb
rhymes: this is likely going to break still

lib/data_update_scripts/20210119131811_replace_instances_of_banned_and_comment_banned.rb
rhymes: same

lib/data_update_scripts/20210119131811_replace_instances_of_banned_and_comment_banned.rb
mstruve, line 6: Is every environment guaranteed to have these roles present? If not we should use a safety operator here and below.

rhymes (commented):
Thanks for picking this up @jacobherrington!

One detail: are we testing that `user.suspended?` in the model is `true` with either role? `banned` or `suspended` ?

app/models/user.rb
```diff
@@ -369,13 +369,12 @@ def cached_followed_tag_names
     end
   end
 
-  # methods for Administrate field
-  def banned
-    has_role? :banned
+  def suspended?
+    has_role?(:suspended) || has_role?(:banned)
```
mstruve, line 373: should we put a TODO here are well to clean up the banned check once this is merged and the data update script has run?

app/helpers/application_helper.rb
```diff
@@ -290,6 +290,7 @@ def admin_config_description(content)
   end
 
   def role_display_name(role)
+    # TODO: [@jacobherrington] remove the check in this method
```
mstruve, line 293: It might be helpful to reference here what will allow this to be removed. For example, once this PR has been merged and deployed. That way if we do accidentally forget about it when someone comes along they have the necessary context to know when and under what circumstances this is safe to remove.

mstruve (commented):
Thanks so much for closing the loop on this PR, I'm excited to get this out the doo. Probably needs someone to go through and give it a good test but I left a couple of small comments.

jacobherrington:
@mstruve I clarified the comments to indicate when we can removed references to banned and comment_banned.

@rhymes it's a little hard to test because the banned role doesn't "exist" anymore - should we force the user model to behave a way it won't in production in a test?

rhymes (approved):
@jacobherrington got it, this PR assumes that the update script has been run successfully. 

In theory you could add `banned` as a role in the test and check the result of `.suspended?` is still `true` but since the role is going to leave at most a few days (non DEV Forems aren't updated automatically), we should be confident enough nobody will change the code in the meantime :) 

Good job!

citizen428 (approved):

PR #12281, opened 2021-01-14
Bug Fix:Increase Default Rate Limit for User Updates to 15

mstruve:
- [x] Bug Fix

## Description
After a few days debugging we finally found the cause of https://github.com/forem/forem/issues/12138, it was rate limiting which became clear thanks to some additional logging. 
Screen Shot 2021-01-14 at 1 04 37 PM

closes: https://github.com/forem/forem/issues/12138

## Added tests?
- [x] Yes

alt_text

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -163,9 +163,6 @@ class UsersController < ApplicationController
   end
 
   def onboarding_checkbox_update
-    # TODO: mstruve will remove once debugging is done
-    Rails.logger.error("onboarding_checkbox_update_user_params:#{params[:user]}")
-
     if params[:user]
       permitted_params = %i[
         checked_code_of_conduct checked_terms_and_conditions email_newsletter email_digest_periodic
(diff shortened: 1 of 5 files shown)


maestromac (approved):
LGTM! :100:

atsmith813:
**suggestion (non-blocking):** For the failing spec, we could update that one and those like it to identify the default values for the `SiteConfig` instead of hardcoding them.

Maybe in that spec or on the `SiteConfig` model we do something like this:

```ruby
def get_default_site_config_value(site_config)
  SiteConfig.get_field(site_config).dig(:default)
 end
 ```
 
 And update the specs, including the [failing one](https://github.com/forem/forem/blob/4c6a7ef234426c33a3427324df8e8be53f3753c5/spec/requests/admin/configs_spec.rb#L711-L716), to:
 ```ruby
 expect do
  default_value = get_default_site_config_value("rate_limit_user_update")
  post "/admin/config", params: { site_config: { rate_limit_user_update: 3 },
end.to change(SiteConfig, :rate_limit_user_update).from(default_value).to(3)
 ```

atsmith813 (approved):

mstruve:
@atsmith813 I love that! We should totally do it for every spec since I noticed all the defaults were hardcoded. I am going to make a follow-up issue with that refactor bc I think it would be worth doing and make those specs more resilient to change.

PR #12282, opened 2021-01-14
Move MarkdownTraverser to MarkdownProcessor::Traverser

atsmith813:
- [x] Refactor

## Description
This PR refactors `MarkdownTraverser` from `/labor` to `MarkdownProcessor::Traverser` in `/services`. I found the current implementation to be pretty solid I left it "as-is".

## Added tests?
- [x] Yes. We previously had no specs specifically for this service so I added some. Given the implementation of this service and how it uses `.each`, I wasn't sure the _best_ way to test it, but I gave it a shot 🙈 .

## Added to documentation?
- [x] No documentation needed

so_close_gif


app/services/markdown_processor/fixer/base.rb
```diff
@@ -53,7 +53,7 @@ def self.split_tags(markdown)
       def self.underscores_in_usernames(markdown)
         return markdown unless markdown.match?(USERNAME_WITH_UNDERSCORE_REGEXP)
 
-        traverser = MarkdownTraverser.new(markdown)
+        traverser = MarkdownProcessor::Traverser.new(markdown)
```
citizen428, line 56: Sorry for coming in so late with this, but to be more consistent with our naming,m I wonder if `Markdow::Fix` and `Markdown::Traverse` would be better names?

atsmith813: @citizen428 No problem - I'm open to that!

My thinking was for services that are written to be initialized, sticking with `er` feels a little better.

```ruby
traverser = MarkdownProcessor::Traverser.new(markdown)
```

I sort of read that in my head as "traverser is a new instance of `Traverser`". And `Traverser` is a noun. That feels better (to me) than saying to myself "traverser is a new instance of `Traverse`" where `Traverse` is a verb.

And on services where we don't use this approach like:
```ruby
fixed_body_markdown = MarkdownProcessor::Fixer::FixForPreview.call(params[:article_body])
```

I read that as "`fixed_mody_markdown` is the result of `call`ing `FixForPreview`" where `FixForPreview` is an action (verb). We are calling the action. And here `Fixer` is also a module name, a verb.

Either way, I'm up for any approach! If we like this or some variation of it, I can also update our docs around service objects. I'm also happy to change it to `Fix` and `Traverse` 🚀 . I definitely can see the value in consistency.

citizen428: Yeah, what you said makes sense and I can definitely live with that.

I tend to not mix these two concepts into `app/services`. If it's a noun, it's a "model", DB-backed or not. If it's a verb, it's a service. In other words, I try to reserve `app/services` for something akin to ["method objects"](https://wiki.c2.com/?MethodObject).

Anyway, personal preferences aside, I think we can agree that the current approach is working just fine, no need for any more changes. Though I did introduce a few non-DB models already, so now I wonder if we should move those to services too...

Zhao-Andy (approved):
LGTM!

rhymes (commented):
LGTM!

rhymes (approved):
forgot to approve :D

PR #12287, opened 2021-01-15
Fix creator flow in /admin/config when :apple_auth feature flag is enabled

fdoxyz:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Our test suite had coverage for OAuth providers user-facing flow, but because of a recent refactor the `/admin/config` flow broke when the `:apple_auth` feature flag is enabled (pretty specific and difficult for to detect manual QA).

This PR fixes the bug and introduces a very simple spec where we check for the available providers in `/admin/config`

In `master` if you enable the feature flag `Flipper.enable(:apple_auth)` and visit `/admin/config` you'll run into a broken page. With this PR the error shouldn't appear and the page should continue to work as expected.

None

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed

flags

### DIFF
--- a/app/views/admin/configs/_apple_auth_provider_settings.html.erb
+++ b/app/views/admin/configs/_apple_auth_provider_settings.html.erb
@@ -5,7 +5,7 @@ requires a custom partial
 -->
 <fieldset class="config-authentication-form">
   <div
-    class="crayons-notice crayons-notice--warning auth-warning <%= provider_keys_configured?(provider.provider_name) ? "hidden" : "" %>">
+    class="crayons-notice crayons-notice--warning auth-warning <%= authentication_enabled_providers.include?(provider.provider_name) ? "hidden" : "" %>">
       <strong>Note:</strong> This authentication provider will <strong>not</strong> be enabled if the key or secret are missing. Please make sure that you enter your key and secret correctly.
   </div>
   <div class="crayons-field">
--- a/spec/system/authentication/creator_config_edit_spec.rb
+++ b/spec/system/authentication/creator_config_edit_spec.rb
@@ -1,0 +1,28 @@
+require "rails_helper"
+
+RSpec.describe "Creator config edit", type: :system, js: true do
+  let(:admin) { create(:user, :super_admin) }
+
+  # Apple auth is in Beta so we need to enable the Feature Flag to test it
+  before { Flipper.enable(:apple_auth) }
+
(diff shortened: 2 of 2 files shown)


app/views/admin/configs/_apple_auth_provider_settings.html.erb
```diff
@@ -5,7 +5,7 @@ requires a custom partial
 -->
 <fieldset class="config-authentication-form">
   <div
-    class="crayons-notice crayons-notice--warning auth-warning <%= provider_keys_configured?(provider.provider_name) ? "hidden" : "" %>">
+    class="crayons-notice crayons-notice--warning auth-warning <%= authentication_enabled_providers.include?(provider.provider_name) ? "hidden" : "" %>">
```
citizen428, line 8: Since we already have a nice `AuthenticationHelper` module, do we maybe want to add a method like `authentication_provider_enabled?(provider_name)` there instead of leaking an implementation detail into the view?

We could change related code in a follow-up PR.

citizen428 (approved):

rhymes (approved):

app/views/admin/configs/_apple_auth_provider_settings.html.erb
fdoxyz: I would be down for sending in a follow up PR with this refactor. `provider_keys_configured?` used to have that role if I'm not mistaken, I can see a few places where `authentication_enabled_providers.include?(provider_name)` is used.

Just to confirm, the implementation of this suggested `authentication_provider_enabled?(provider_name)` would possibly be as simple as wrapping this same statement right? At least this way the views would call a helper method (with test coverage on it) instead of placing these statements in the views, does this make sense or sounds useful?

```ruby
def authentication_provider_enabled?(provider_name)
   authentication_enabled_providers.include?(provider_name)
end
```

citizen428: Yes, that's what I meant. 

I think views should "know" as little as possible, which includes knowing if something will respond to a certain message like `include?` here. As unlikely as it is if we ever change the underlying implementation (e.g. `Authentication::Providers.const_get(klass).enabled?`), we shouldn't have to update views.

Issue #12131, opened 2021-01-05
Saving unpublished blog post after fixing validation error makes it published


DonKarlssonSan:
**Describe the bug**
When saving an _unpublished_ blog post after fixing validation error it becomes _published_

**To Reproduce**

- Create a new blog post
- Add something that is validated, for example: `{% codepen https://codepen.io/DonKarlssonSan/pen/dyMaRJN default-tab=result %}`
- "Save draft"
- Edit the blog post
- Change the content to something that is not valid, for example: `{% codepen https://codepen.io/DonKarlssonSan/pen/dyMaRJN default-tab="result" %}`
- Click "Save draft"
- An error message is displayed: "Whoops, something went wrong: base: Invalid Options"
- Fix the error (remove the "" around result)
- The only option is to click "Save changes" 
- The blog post is _published_

**Expected behavior**

- In the last step: the blog post is saved and kept un-published

**Screenshots**

**Desktop (please complete the following information):**

- OS, version: Windows 10, MacOS 11.1
- Browser, version: 84.0.1 (64 bit)

**Additional context**

nickytonline:
I exhibited this the other day myself. Thanks for reporting @DonKarlssonSan!

vaidehijoshi:
@citizen428 Was this issue fixed by https://github.com/forem/forem/pull/12288, or is there still more work to be done on this front? Just want to confirm before closing it out :)

citizen428:
No, it's done, apparently this whole cycle I forgot add "Closes" in front of the PR numbers and lost automation. 🤦‍♂️

PR #12288, opened 2021-01-15
Fix accidental publishing of draft articles

citizen428:

- [X] Bug Fix

## Description

#12131 described a workflow that led to the accidental publishing of a draft post after fixing validation errors. This PR fixes that.

#12131 

Follow the steps as outlined in the original issue and ensure that you still have the option to save a draft at the last step, not only "Save changes".

- Create a new blog post
- Add something that is validated, for example: `{% codepen https://codepen.io/DonKarlssonSan/pen/dyMaRJN default-tab=result %}`
- "Save draft"
- Edit the blog post
- Change the content to something that is not valid, for example: `{% codepen https://codepen.io/DonKarlssonSan/pen/dyMaRJN default-tab="result" %}`
- Click "Save draft"
- An error message is displayed: "Whoops, something went wrong: base: Invalid Options"
- Fix the error (remove the "" around result)
-  There should be both "Publish" and "Save Draft" buttons

None

## Added tests?

- [X] No, and this is why: verifying this approach first, then figuring out how to write a regression test for this in JS.

## Added to documentation?

- [X] No documentation needed

### DIFF
--- a/app/javascript/article-form/articleForm.jsx
+++ b/app/javascript/article-form/articleForm.jsx
@@ -251,11 +251,12 @@ export default class ArticleForm extends Component {
 
   handleArticleError = (response, publishFailed = false) => {
     window.scrollTo(0, 0);
+    const { published } = this.state;
     this.setState({
       errors: response,
       submitting: false,
       // Even if it's an update that failed, published will still be set to true
-      published: !publishFailed,
+      published: published && !publishFailed,
     });
   };
 
--- a/spec/system/articles/user_fixes_a_draft_spec.rb
+++ b/spec/system/articles/user_fixes_a_draft_spec.rb
@@ -1,0 +1,38 @@
+require "rails_helper"
+
+# Regression test for https://github.com/forem/forem/issues/12131
+RSpec.describe "", type: :system do
+  let(:correct_liquid_tag) do
+    "{% codepen https://codepen.io/user/pen/abcdefg default-tab=result %}"
+  end
+  let(:incorrect_liquid_tag) do
+    '{% codepen https://codepen.io/user/pen/abcdefg default-tab="result" %}'
+  end
+
+  it "let's the user fix a broken draft without publishing", js: true, aggregate_failures: true do
+    # Create a new blog post and add content that validates correctly
+    sign_in create(:user)
+    visit new_path
+    fill_in "article-form-title", with: "Regression spec"
+    fill_in "article_body_markdown", with: correct_liquid_tag
+
+    # Save the draft
+    click_button "Save draft"
+    expect(page).to have_content("Unpublished post")
+
+    # Edit the blog post with content that doesn't pass validations
+    click_link("Click to edit")
+    fill_in "article_body_markdown", with: incorrect_liquid_tag
+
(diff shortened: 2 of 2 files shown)


citizen428:
@nickytonline I was looking for something quick to do towards the end of my working day. I believe the issue here was that `handleArticleError` didn't take the previous article state into account, but I may be off here. That said, applying this change fixes the described issue and other manually tested article interactions also seem to behave as expected. Does this sound plausible to you?

app/javascript/article-form/articleForm.jsx
```diff
@@ -251,11 +251,12 @@ export default class ArticleForm extends Component {
 
   handleArticleError = (response, publishFailed = false) => {
     window.scrollTo(0, 0);
+    const { published } = this.state;
     this.setState({
       errors: response,
       submitting: false,
       // Even if it's an update that failed, published will still be set to true
```
citizen428, line 258: To be honest, I wasn't quite sure about the rationale around this comment. Should this function touch the `published` state at all?

This function is used as error callback (`errorCb`) in `submitArticle` (defined in (`app/javascript/article-form/actions.js`):

```js
// If there is an error and the method is POST, we know they are trying to publish.
errorCb(response, method === 'POST');
```

I do somewhat get the idea behind this, but it (a) doesn't seem to properly take drafts into account and (b) intuitively feels wrong since the article won't in fact be published if it fails validation, even if we set this attribute.

nickytonline: I agree on both counts. That logic should probably be revisited in another PR.

Also looking at the publish logic

https://github.com/forem/forem/blob/c2c1366a4ddc452f9d7e587d1b930391acbe0c06/app/javascript/article-form/articleForm.jsx#L207-L213

`setState` is not immediate and it's bad practice to mutate state like so `state.published` as there are no guarantees. That should probably be reworked as well.

nickytonline (approved):
Tested this locally a few times and all works as expected for the v2 editor. This issue does not affect the v1 editor as the user needs to explicitly write in the markdown properties that the post is published.

nickytonline (changes requested):
Ooops, didn't realize this was in draft still. Removing the approval even though the fix works as this needs some tests to prevent this regression as you mentioned you were validating the fix first.

nickytonline:
I think for testing this, we could do a component integration test for now @citizen428, but long term, this should be an end to end test as part of the suite of flows to test for the editing experience.

app/javascript/article-form/articleForm.jsx
citizen428: @nickytonline If you don't mind, I'll leave further refactoring and fixes up to you, as this is more your domain than mine 😄 Maybe add an issue for it so it finds its way into the next planning meeting.

nickytonline: I created a separate issue, #/12349.

citizen428:
@nickytonline Thanks for creating the issue. Also, as agreed on Slack I added a Rails system test for the regression for now.

nickytonline (approved):
Thanks for adding the test @citizen428! LGTM. 🚢

benhalpern (approved):

Issue #50, opened 2017-06-20
Labels on reaction buttons


bigblind:
I was reading an article on dev.to using VoiceOver on my iPhone, an discovered the new reactions. They look great, but I couldn't make sense of them using VoiceOver. Also, on desktop, when I hover over the icons, I'd love to see a label explaining what the icon means.

If it's useful to you, I can update a video of trying to use these buttons using VO, some time tomorrow.

jessleenyc:
Hey @bigblind, thanks for flagging this. I'll look into and keep you posted.

benhalpern:
@bigblind when you say "label" what exactly do you mean? Like a tooltip or a title attribute?

bigblind:
For me, the title attribute works find, but the W3C [warns against relying on it](https://www.w3.org/TR/html5/single-page.html#the-title-attribute):

>  &#9888; **Warning!** Relying on the title attribute is currently discouraged as many user agents do not expose the attribute in an accessible manner as required by this specification (e.g. requiring a pointing device such as a mouse to cause a tooltip to appear, which excludes keyboard-only users and touch-only users, such as anyone with a modern phone or tablet).

benhalpern:
@bigblind thanks for the heads up.

jessleenyc:
We added alt tags to the images, so closing this for now.

PR #12292, opened 2021-01-15
Navigation Items and Role access for Data Update Scripts

Ridhwana:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
In this PR, the following changes are made to the Data Update Scripts: 
- We keep the route `/admin/data_update_scripts` namespaced under admin and allow access to a single_resource_admin role with resource type DataUpdateScript. 
- Add a nav item under a section called Tech Resources on the /admin and the admin sidebar.
- It is useful to know that both admins and super_admins will also have access to these pages.
- We also write a little script that gives anyone with tech_admin permissions access to this resource by adding the single_resource_admin role with resource type DataUpdateScript role. 

It may be confusing to you why we also give the user a role single_resource_admin DataUpdateScript when a tech_admin role is selected. To read more about it you can read [this comment](https://github.com/forem/internalEngineering/issues/276#issuecomment-771658292) or the full discussion.

Issue: https://github.com/forem/internalEngineering/issues/276

Test that the only single_resource _admins with resource type DataUpdateScript, admins and super _admins have access to data scripts.
- Open a rails console with `rails c`
- Find your user that you're logged in with, we're assuming it's the last user `u = User.last`
- Show all that users roles: `roles = u.roles` 
- Keep note of the roles you have and then destroy them `roles.destroy_all'
- Add a single resource admin role to your user: `u.add_role(:single_resource_admin, DataUpdateScript)` 
- You will now have access to this http://localhost:3000/admin/data_update_scripts.
-
You will also be able to see this resource should you have admin or super_admin roles.

Screenshot 2021-01-15 at 11 16 25
Screenshot 2021-01-15 at 11 33 14

Test that a tech_admin given permissions through Manage Users has access to the page:
 - Go http://localhost:3000/admin/users
 - Click on the user that you want to give a tech admin role too
 - Before continuing it would be best to destroy this users current roles to more effectively test the feature. You can do this 
    - Open a rails console with `rails c`
    - Find find the user and assign the User to u.
    - Show all that users roles: `roles = u.roles` 
    - Keep note of the roles you have and then destroy them `roles.destroy_all'
- Now go back to http://localhost:3000/admin/users, click on the same user, click on `Manage User` 
- Select Tech Admin from the dropdown: 
Screenshot 2021-02-05 at 11 43 10
- You should see two roles pop up for the user:
Screenshot 2021-02-05 at 11 42 33

The user will now have access to http://localhost:3000/admin/data_update_scripts

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [ ] No documentation needed
Not as yet.

alt_text


app/helpers/admin_helper.rb
```diff
@@ -37,6 +37,14 @@ module AdminHelper
 
   PROFILE_ADMIN = { name: "config: profile setup", controller: "profile_fields" }.freeze
 
+  TECH_MENU_ITEMS = [
+    { name: "data_update_scripts", controller: "data_update_scripts" },
+  ].sort_by { |menu_item| menu_item[:name] }
```
Ridhwana, line 42: At some point we'll add more nav items like the Sidekiq page and Feature flags.

README.md
```diff
@@ -67,7 +67,7 @@ within our community. ❤️
 
 - [What is Forem?](#what-is-forem)
 - [Table of Contents](#table-of-contents)
-- [Community](#community) 
+- [Community](#community)
```
Ridhwana, line 70: I saved this file and it just picked up an extra whitespace 🤷🏽‍♀️

Ridhwana (commented):
**NOTE:** This PR is still a work in progress and has changed direction recently, hence it is not ready for review as yet. Waiting on discussion [here](https://github.com/forem/internalEngineering/issues/276#issuecomment-770907584) before continuing to work on it.

lib/data_update_scripts/20210203104631_add_single_resource_admin_role_to_users_with_tech_admin.rb
```diff
@@ -0,0 +1,10 @@
+module DataUpdateScripts
+  class AddSingleResourceAdminRoleToUsersWithTechAdmin
+    def run
```
Ridhwana, line 3: Wanted a second pair of 👀's on this to double check the performance of this query. Maybe there's a way to improve it (especially since we're working with the users).

app/helpers/admin_helper.rb
```diff
@@ -37,6 +37,10 @@ module AdminHelper
 
   PROFILE_ADMIN = { name: "config: profile setup", controller: "profile_fields" }.freeze
 
+  TECH_MENU_ITEMS = [
```
Ridhwana, line 40: This is a list of tech resources that we add. Eventually we may move things like the Sidekiq page to here as well.

lib/data_update_scripts/20210203104631_add_single_resource_admin_role_to_users_with_tech_admin.rb
```diff
@@ -0,0 +1,10 @@
+module DataUpdateScripts
+  class AddSingleResourceAdminRoleToUsersWithTechAdmin
```
Ridhwana, line 2: The reason for this Data Script is because we initially wanted to give permissions to Tech Admins to access this page. We instead decided to give permissions to the SingleResourceAdmins with resource type DataUpdateScript. You can read more about the details on [this thread](https://github.com/forem/internalEngineering/issues/276#issuecomment-771658292).

config/routes.rb
```diff
@@ -66,6 +63,8 @@
                                               destroy], path: "listings/categories"
 
       resources :comments, only: [:index]
+      resources :data_update_scripts, only: %i[index show]
+      post "/data_update_scripts/:id/force_run", to: "data_update_scripts#force_run"
```
rhymes, line 67: this can be Rails-yfied with:

```suggestion
      resources :data_update_scripts, only: %i[index show] do
        member do
          post :force_run
        end
      end
```

this way you also get the proper path helper:

```ruby
[8] pry(main)> show-routes -G force_run
                force_run_admin_data_update_script POST     (/locale/:locale)/admin/data_update_scripts/:id/force_run(.:format)                    admin/data_update_scripts#force_run {:locale=>nil}
```

lib/data_update_scripts/20210203104631_add_single_resource_admin_role_to_users_with_tech_admin.rb
rhymes: I think the script is fine. there's still one thing left: how do we make sure that each time a user is given the `tech_admin` role they are also given the single resource admin? Otherwise this will only work for existing tech admins, not future ones

rhymes (approved):
Cool! Left a note about `routes.rb`

Non blocking :)

rhymes (changes requested):
Sorry, shouldn't have accepted, the second note is blocking: how do we deal with future `tech_admins` promoted in the http://localhost:3000/admin/users pages. Maybe adding `tech_admin` should automatically trigger `single_resource_admin, DataUpdateScript` but that would create more work for a Forem admin I guess

lib/data_update_scripts/20210203104631_add_single_resource_admin_role_to_users_with_tech_admin.rb
Ridhwana: @rhymes what do you think of the ffg:

We can add the single resource admin (DataScript) role to a user when the admin chooses to give a user a tech_admin role under Manage Users. Since this seems to be the only place to assign user roles (from my brief  investigation) it would be a safe solution. This can be done in ManageActivityAndRoles.

However, it seems that tech_admin up until now has not even been assigned through the UI (based on the fact that it's not in the dropdown):
Screenshot 2021-02-04 at 13 40 54.
I'd have to add it to the dropdown first of course.

It does make me a bit uneasy to be adding code along to the tech_admin when it seems like more decisions need to be made about how this role behaves.

Ridhwana:
> Sorry, shouldn't have accepted, the second note is blocking: how do we deal with future `tech_admins` promoted in the http://localhost:3000/admin/users pages. Maybe adding `tech_admin` should automatically trigger `single_resource_admin, DataUpdateScript` but that would create more work for a Forem admin I guess

@rhymes why do you say that it would create more work for a forem admin?

rhymes:
@Ridhwana because in addition to adding `tech_admin` all Forem admins should remember to add `single_resource_admin, DataUpdateScript` as well. If we start adding more that `tech_admin`s should be able to see, then we'd have to tell all Forem admins to keep a list of additional roles they need to manually assign to `tech_admin`, if we instead make it so that by adding `tech_admin` from the Admin - Users page, the system automatically grants the `single_resource_admin, DataUpdateScript` role (and the other ones in the future), then we avoid adding more workload for them.

Unless the point here is not to have _all_ tech admins be able to access that page

Ridhwana:
Ah @rhymes I misunderstood your last comment, I thought that you meant that triggering it automatically in the code will be more work for Forem Admins, which confused me. 😅

Sounds like we came to the same conclusion based on my comment in the code - will implement it and re request review. Thanks!

Ridhwana:
@lisasy you'd given [the go-ahead here](https://github.com/forem/internalEngineering/issues/276#issuecomment-760306837) but just wanted to keep you in the loop as this affects the admin interface and creator experience.

app/services/moderator/manage_activity_and_roles.rb
```diff
@@ -78,6 +78,14 @@ def handle_user_status(role, note)
         check_super_admin
         remove_negative_roles
         user.add_role :super_admin
+      when "Tech Admin"
```
Ridhwana, line 81: When we add a tech admin role,  we push along a :single_resource_admin with DataUpdateScript role too. The rationale for this can be found [here](https://github.com/forem/internalEngineering/issues/276#issuecomment-771658292)

lib/data_update_scripts/20210203104631_add_single_resource_admin_role_to_users_with_tech_admin.rb
Ridhwana: ✅ Done

rhymes (approved):
LGTM!

mstruve (approved):
LGTM 🚀

app/services/moderator/manage_activity_and_roles.rb
```diff
@@ -78,6 +78,14 @@ def handle_user_status(role, note)
...
+        check_super_admin
+        remove_negative_roles
+        user.add_role :tech_admin
```
citizen428, line 84: **nitpick (non-blocking):** Ok, this is a super nitpick, but we're calling `add_role` twice in a row and use different code styles for the parenthesis around arguments. The general consensus is to use them unless it's a DSL method like `belongs_to`, so maybe we can add them and make my OCD happy?

spec/lib/data_update_scripts/add_single_resource_admin_role_to_users_with_tech_admin_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "rails_helper"
+require Rails.root.join(
+  "lib/data_update_scripts/20210203104631_add_single_resource_admin_role_to_users_with_tech_admin.rb",
+)
+
+describe DataUpdateScripts::AddSingleResourceAdminRoleToUsersWithTechAdmin do
+
+  before do
+    users = create_list(:user, 5)
```
citizen428, line 9: You don't need to assign this to a variable, it's never used. In fact, Rubocop should complain about this.

spec/lib/data_update_scripts/add_single_resource_admin_role_to_users_with_tech_admin_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    User.first.add_role(:tech_admin)
```
citizen428, line 10: I think it'd be easier to create three distinct users:

```ruby
tech_admin1 = create(:user, :tech_admin)
tech_admin2 = create(:user, :tech_admin)
admin = create(:user, :admin)
```

then you can assert against things like `tech_admin1.reload.roles`, etc. which seems better to me than going through the AR query interface.

app/services/moderator/manage_activity_and_roles.rb
Ridhwana: > The general consensus is to use them unless it's a DSL method like belongs_to
@citizen428 I was not aware of this, but thanks for the heads up :) 

The reason I did it like this was because glancing at the rest of the methods (and the file) it seemed that the pattern was when [we have one parameter for add_role we do not use parenthesis](https://github.com/forem/forem/blob/1a2dc2ec51d849a28e2ee9f9bdaca8ced440da6c/app/services/moderator/manage_activity_and_roles.rb#L71), but [when we have two parameters then we do](https://github.com/forem/forem/blob/1a2dc2ec51d849a28e2ee9f9bdaca8ced440da6c/app/services/moderator/manage_activity_and_roles.rb#L92). 

It doesn't make sense to just add the parenthesis to this line, if this is the general consensus, we should probably add it to the rest of the file too which I can do if that's the preference? I also think it would beneficial to reference a good style guide, or perhaps Rubocop should have been complaining about this.

spec/lib/data_update_scripts/add_single_resource_admin_role_to_users_with_tech_admin_spec.rb
Ridhwana: It's weird that its not complaining. I also noticed that you fixed some rubocop stuff the other day in a PR related to the DUS, yet when I pushed the commit hook didn't alert. I'll double check if it's some setting on my side.

Ridhwana: Thanks for letting me know though, fixed  ✅

spec/lib/data_update_scripts/add_single_resource_admin_role_to_users_with_tech_admin_spec.rb
Ridhwana: Done ✅

app/services/moderator/manage_activity_and_roles.rb
citizen428: These are some of the related cops. The tl;dr is use parenthesis unless it's an internal or framework DSL or when the method call has no arguments.

https://github.com/rubocop-hq/ruby-style-guide#method-definition-parentheses
https://github.com/rubocop-hq/ruby-style-guide#dsl-method-calls
https://github.com/rubocop-hq/ruby-style-guide#method-calls-with-no-arguments
https://github.com/rubocop-hq/ruby-style-guide#methods-that-are-part-of-an-internal-dsl

spec/lib/data_update_scripts/add_single_resource_admin_role_to_users_with_tech_admin_spec.rb
citizen428: Yes, I wondered about that recent fix PR too. Maybe there's some configuration issue on your side, let me know if you need any help figuring this out.

citizen428 (approved):

Ridhwana:
Just waiting for @lisasy's approval from product and then I'll merge this.

app/services/moderator/manage_activity_and_roles.rb
Ridhwana: Thanks @citizen428, I'll update this file in a subsequent PR 👍

lisasy (approved):
LGTM! 

Since we are updating the admin navigation & IA generally in [RFC #50](https://github.com/forem/rfcs/pull/50), we'll make sure we roll the updates in this PR to *that* one coming up.

PR #12262, opened 2021-01-13
Fix shell version logic in async info controller

atsmith813:
## What type of PR is this? (check all applicable)
- [x] Bug Fix

## Description
This is a quick fix for a failing spec
Screen Shot 2021-01-13 at 11 45 11 AM

## Related Tickets & Documents
https://github.com/forem/forem/pull/12224

## Added tests?
- [x] No, and this is why: specs are already included

## Added to documentation?
- [x] No documentation needed

yellow_ranger_gif

PR #12294, opened 2021-01-15
Attempt to fix random test failures when dealing with the ForemInstance class 

Ridhwana:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Currently on some PR's there are two tests that are failing: 
Screenshot 2021-01-15 at 15 10 29

My debugging process has been as follows:

A test has run prior to this that has set the "HEROKU_RELEASE_CREATED_AT" ENV variable to `A deploy date set on Heroku`. 
Therefore, when we call the method `release_adjusted_cache_key`, it makes a reference to `ForemInstance.deployed_at` which reads as follows: 
```
@deployed_at ||= ApplicationConfig["RELEASE_FOOTPRINT"].presence || ENV["HEROKU_RELEASE_CREATED_AT"].presence
```
Since the value of `HEROKU_RELEASE_CREATED_AT` at a previous point was `A deploy date set on Heroku`, the `deployed_at` gets cached with that value and now for subsequent tests, that value sticks with `A deploy date set on Heroku`. Hence, causing the rest of the tests to fail. 

To solve this I tried to did two things: 
- I made sure that the `HEROKU_RELEASE_CREATED_AT` ENV value was set to and empty string in the before statement 
- I reset the `deployed_at` to nil 

This seemed to have solved the failing test when I tested it on [my PR](https://github.com/forem/forem/pull/12292/commits), so hoping that it works on solving the issue.

I did something similar in https://github.com/forem/forem/pull/12262 which seemed to have worked.

I guess just run application_heper_spec.rb locally, although it was never failing locally so that may be pointless.

## Added tests?

- [ ] Yes
- [x] No, and this is why: This is a flakey spec that I'm trying to fix

## Added to documentation?

- [ ] README
- [x] No documentation needed

alt_text

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -62,10 +62,7 @@ RSpec.describe ApplicationHelper, type: :helper do
   end
 
   describe "#release_adjusted_cache_key" do
-    before do
-      stub_const("ENV", ENV.to_h.merge("HEROKU_RELEASE_CREATED_AT" => ""))
-      ForemInstance.instance_variable_set(:@deployed_at, nil)
-    end
+    after { ForemInstance.instance_variable_set(:@deployed_at, nil) }
 
     it "does nothing when RELEASE_FOOTPRINT is not set" do
       allow(ApplicationConfig).to receive(:[]).with("RELEASE_FOOTPRINT").and_return(nil)
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -63,6 +63,7 @@ RSpec.describe ApplicationHelper, type: :helper do
 
   describe "#release_adjusted_cache_key" do
     before do
+      stub_const("ENV", ENV.to_h.merge("HEROKU_RELEASE_CREATED_AT" => ""))
       ForemInstance.instance_variable_set(:@deployed_at, nil)
     end
 
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -62,6 +62,10 @@ RSpec.describe ApplicationHelper, type: :helper do
   end
 
   describe "#release_adjusted_cache_key" do
+    before do
+      ForemInstance.instance_variable_set(:@deployed_at, nil)
+    end
+
     it "does nothing when RELEASE_FOOTPRINT is not set" do
       allow(ApplicationConfig).to receive(:[]).with("RELEASE_FOOTPRINT").and_return(nil)
       expect(helper.release_adjusted_cache_key("cache-me")).to include("cache-me")
--- a/spec/models/forem_instance_spec.rb
+++ b/spec/models/forem_instance_spec.rb
@@ -2,19 +2,20 @@ require "rails_helper"
 
 RSpec.describe ForemInstance, type: :model do
   describe "deployed_at" do
-    before do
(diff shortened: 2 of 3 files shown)


fdoxyz:
I just restarted a job because of a flaky spec

Ridhwana:
> I just restarted a job because of a flaky spec

There's just flakey spec inceptions all round 🤯

fdoxyz (approved):

spec/helpers/application_helper_spec.rb
```diff
@@ -62,6 +62,11 @@
   end
 
   describe "#release_adjusted_cache_key" do
+    before do
+      stub_const("ENV", ENV.to_h.merge("HEROKU_RELEASE_CREATED_AT" => ""))
```
maestromac, line 66: I don't make sense that you need to to stub `HEROKU_RELEASE_CREATED_AT` because each of the spec in this describe block is already stubbing with `allow(ApplicationConfig).to ...`

I think what you need instead is an afterblock to clear the variable created.
```
after do
   ForemInstance.instance_variable_set(:@deployed_at, nil)
end
```

maestromac: Also, if this is a caching related issue, you should be able to reproduce it consistently locally. And because of that, you should be able to solidly verify if this fix will work or not.

Ridhwana: @maestromac my suspicion is when the first test runs in the describe block then it has the cached value `A deploy date set on Heroku` which was set on `HEROKU_RELEASE_CREATED_AT` in another test. (

Hence I thought that I'd rather be 100% sure that the variable was clear before running the tests, as this will pretty much mimic a real scenario. It's pretty much impossible to test the change locally since this does not replicate locally. 

That being said, just setting the `ForemInstance.instance_variable_set(:@deployed_at, nil)` seemed to have fixed the issue on [another branch](https://github.com/forem/forem/pull/12292/commits), but again I didn't think there was too much harm in making sure that the Heroku Variable was also clear, but if you feel  like just resetting the ForemInstance variable is sufficient, I can undo the HEROKU_RELEASE_CREATED_AT stub.

Ridhwana: In addition, the reason I put `ForemInstance.instance_variable_set(:@deployed_at, nil)` in a before block instead of an after block is incase the cached variable carries over from another test, which has happened before.

It should do the same thing if I move it to after blocks in all the tests where we've added this line.

Ridhwana: Ah sorry i didn't read your updated your comment about the `after blocks`. What you're saying makes sense, and I can move them all to after blocks instead.

Ridhwana: > Also, if this is a caching related issue, you should theoretically be able to reproduce it consistently locally. And because of that, you should be able to solidly verify if this fix will work or not.

I've been grasping at straws, so would love your help if you able to provide more insight. 
I can't definitely tell what the issue is, but I do believe it's a combination of the the order the tests are running in, and the memoization. But again, there has been now way to confirm it locally. However, this fix did seem to sort out the issue in two different PR's; https://github.com/forem/forem/pull/12292/commits and https://github.com/forem/forem/pull/12262

Zhao-Andy (approved):

Issue #11851, opened 2020-12-10
Update dev.to/mods for non-logged in users


michael-tharrington:
**Is your feature request related to a problem? Please describe.**

Yes, `dev.to/mods` has confusing copy for folks who are not already mods (trusted users or tag mods).

**Describe the solution you'd like**

I want the page to read as follows:

```
# DEV Mods

We periodically award some DEV members with heightened privileges to help moderate the community.

Check out our [Code of Conduct](https://dev.to/code-of-conduct) and read through our [Trusted User Guide](https://dev.to/community-moderation) and [Tag Moderation Guide](https://dev.to/tag-moderation). 

If you'd like to assist us as a trusted user or tag mod, please email us at <a href="mailto:yo@dev.to">yo@dev.to</a> and let us know which role you're interested in and why. If it's tag moderation, please tell us what tags you'd like to moderate for.
```

**Describe alternatives you've considered**

None.

**Images**

Here is the current copy for `dev.to/mods` for folks that are logged out.

![Image 2020-12-10 at 5 03 08 PM](https://user-images.githubusercontent.com/16007075/101836307-25376b00-3b0b-11eb-8cd3-0bc205e8cf05.png)

vaidehijoshi:
Note: we should make sure that we are not hardcoding any emails or any DEV-specific content when addressing this (like `yo@dev.to`)

PR #12301, opened 2021-01-15
Update dev.to/mods for non-logged in users

msarit:

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Update the copy on `dev.to/mod`, which is currently confusing for folks who are not already mods (trusted users or tag mods).

closes https://github.com/forem/forem/issues/11851

1. Navigate to `localhost:3000/mod` as a user that is not trusted or a tag moderator.
2. You should see copy such as:
```
We periodically award some DEV members with heightened privileges to help moderate the community.
Check out our Code of Conduct and read through our Trusted User Guide and Tag Moderation Guide.
If you'd like to assist us as a trusted user or tag mod, please email us at yo@dev.to and let us know which role you're interested in and why. If it's tag moderation, please tell us what tags you'd like to moderate for.
```
3. Repeat step 1 on a forem that is NOT DEV. The page copy **should reflect the unique name and email of the forem being viewed**.

_No UI changes_

## Added tests?

- [X] Yes
- [ ] No, and this is why:

## Added to documentation?

- [ ] README
- [X] No documentation needed

None

Cartoon


app/views/moderations/index.html.erb
```diff
@@ -31,12 +31,15 @@
   </main>
 <% else %>
   <div class="container" style="margin-top: 90px;">
-    <h1 style="text-align: center"><%= community_name %> Mods</h1>
+    <h1 class="pt-5" style="text-align: center"><%= community_name %> Mods</h1>
     <div class="body">
       <p>We periodically award some <%= community_name %> members with heightened privileges to help moderate the community.</p>
-      <p>Email <%= email_link %> if you'd like to be considered right away.</p>
+      <% if community_name == "DEV" %>
```
rhymes, line 37: As we've already had an issue about using `community_name` as a condition and since `community_name` on dev.to now is `DEV Community`, I think we should stick with `SiteConfig.dev_to?`

See the recent change here https://github.com/forem/forem/pull/12083

rhymes (changes requested):
Great! I requested as this wouldn't work in production as-is

app/views/moderations/index.html.erb
msarit: Thank you Rhymes! 🙏🏾

rhymes (approved):
Updated!

Should we add a tests as well?

spec/requests/moderations_spec.rb
```diff
@@ -135,4 +135,54 @@
       expect(response.body).to include "circle centered-icon adjustment-icon subtract"
     end
   end
+
+  describe "/mod" do
+    # rubocop:disable Layout/LineLength
+    coc_guides_copy = 'Check out our <a href="https://dev.to/code-of-conduct">Code of Conduct</a> and read through our <a href="https://dev.to/community-moderation">Trusted User Guide</a> and <a href="https://dev.to/tag-moderation">Tag Moderation Guide</a>.'
+    # rubocop:enable Layout/LineLength
+    become_mod_copy = "If you'd like to assist us as a trusted user or tag mod"
+    logged_out_copy = "P.S. You are not currently signed in."
```
vaidehijoshi, line 144: Can we use Rspec's `let` [helper](https://relishapp.com/rspec/rspec-core/v/2-6/docs/helper-methods/let-and-let) here instead of using variables directly?

Same for line 152 and 178, and anywhere else in the file :)

app/views/moderations/index.html.erb
```diff
@@ -31,12 +31,15 @@
   </main>
 <% else %>
   <div class="container" style="margin-top: 90px;">
-    <h1 style="text-align: center"><%= community_name %> Mods</h1>
+    <h1 class="pt-5" style="text-align: center"><%= community_name %> Mods</h1>
     <div class="body">
       <p>We periodically award some <%= community_name %> members with heightened privileges to help moderate the community.</p>
-      <p>Email <%= email_link %> if you'd like to be considered right away.</p>
+      <% if SiteConfig.dev_to? %>
+        <p>Check out our <a href="https://dev.to/code-of-conduct">Code of Conduct</a> and read through our <a href="https://dev.to/community-moderation">Trusted User Guide</a> and <a href="https://dev.to/tag-moderation">Tag Moderation Guide</a>.</p>
```
mstruve, line 38: Is there any way we can generalize this code so we don't have to use `if SiteConfig.dev_to?`?

msarit: Great question @mstruve . I thought about this for some time. The thing is, those links are specific to DEV. I checked if we have any SiteConfig variables that contain custom links like these, but we have none as yet. I think it's a great idea though to add to our Forem generalization goals 😄 

I'm open to any suggestions from you though!

vaidehijoshi: @msarit I think you should be able to link to `/community-moderation` and `/tag-moderation`. It looks like those routes exist on forem.team and forem.dev, and they exist as routes in the `routes.rb` file right now:

https://github.com/forem/forem/blob/27648a938b0168b2b9fc2cd8162b203e44d8100d/config/routes.rb#L418-L421

So, you should be able to link without hardcoding to dev.to (and then you don't need to use `dev_to?` 😄). Ideally, we want to remove the `dev_to?` helper, so we should avoid using that since we know we want to move towards generalizing our codebase entirely :)

msarit: @vaidehijoshi ok this sounds good. my only question is: are all forems going to have these three links at all times? `/community-moderation`, `/tag-moderation` and `/code-of-conduct`?

vaidehijoshi: I suspect we will always have a CoC, but I am not 100% sure about the others. I will say that we have these links throughout our site right now; [see all these CoC occurrences](https://github.com/forem/forem/search?q=code-of-conduct), [tag moderation occurrences](https://github.com/forem/forem/search?q=tag-moderation), and [community-moderation occurrences](https://github.com/forem/forem/search?q=community-moderation). So linking to them directly is following a pattern we already have set in the codebase. If we decide to move away from these links, we're going to have to address all of these occurrences, so I would say that if we _do_ decide not to have these links on every Forem, we'll have to address is holistically.

TL;DR, we have enough links to these 3 pages in the codebase, so I think it's okay to add another one in an effort to not hardcode DEV-specific things :)

msarit: @vaidehijoshi Np! done!

spec/requests/moderations_spec.rb
```diff
@@ -135,4 +136,52 @@
       expect(response.body).to include "circle centered-icon adjustment-icon subtract"
     end
   end
+
+  describe "/mod" do
+    # rubocop:disable Layout/LineLength
+    let(:coc_guides_copy) { 'Check out our <a href="/code-of-conduct">Code of Conduct</a> and read through our <a href="/community-moderation">Trusted User Guide</a> and <a href="/tag-moderation">Tag Moderation Guide</a>.' }
```
vaidehijoshi, line 142:
```suggestion
    let(:coc_guides_copy) { 
        'Check out our <a href="/code-of-conduct">Code of Conduct</a> and read through our <a href="/community-moderation">Trusted User Guide</a> and <a href="/tag-moderation">Tag Moderation Guide</a>.' 
     }
```
By the way, you could make this copy whatever you want since this is a test -- if you shorten it, you can remove the rubocop disable lines 😉

spec/requests/moderations_spec.rb
```diff
@@ -135,4 +136,52 @@
...
+    # rubocop:enable Layout/LineLength
+    let(:become_mod_copy) { "If you'd like to assist us as a trusted user or tag mod" }
+    let(:logged_out_copy) { "P.S. You are not currently signed in." }
+
+    context "when on dev.to" do
```
vaidehijoshi, line 147: Hm, now that we are not using `dev_to?`, do we need 2 different context for these tests? Seems like the behavior should be the same regardless of what Forem you are on, so we can consolidate these I think :)

msarit: @vaidehijoshi Alright I can do that as you say.

I wrote the tests this way because there is a line in the copy that references `community_name` and i thought it a good idea to test this for different communities 🤷🏾 

`app/views/moderations/index.html.erb:39`

```erb
<p>We periodically award some <%= community_name %> members with heightened privileges to help moderate the community.</p>
```
But I'll take it out.

vaidehijoshi: @msarit Thanks for the additional context! 

I wanted to explain a little bit about _why_ I am suggesting this. In the view where you added these changes, you're using the `community_name` helper, which is defined in the `ApplicationHelper`. 

https://github.com/forem/forem/blob/3dd42fe8a0ebc00df619fd2d3a6a44b125cba18d/app/views/moderations/index.html.erb#L36

While the view is using the helper, there isn't any _other_ logic happening here. Previously, when you were checking `SiteConfig.dev_to?`, you _did_ have some conditional rendering logic, but after you updated this PR, you're using the value of that helper directly.

The helper itself is already tested in the `application_helper_spec.rb` (admittedly, there's only one test for this):
https://github.com/forem/forem/blob/8d462801ece4a866a4aaf4a3844759e23d88d866/spec/helpers/application_helper_spec.rb#L39-L44

Looking at the view, the only place where you might encounter different behavior is the line that renders based on whether or not the user is signed in or not; the `community_name` helper's behavior doesn't really have anything to do with the view, it is its own method. So when you test against the two different community names on these lines:
https://github.com/forem/forem/blob/3dd42fe8a0ebc00df619fd2d3a6a44b125cba18d/spec/requests/moderations_spec.rb#L149
https://github.com/forem/forem/blob/3dd42fe8a0ebc00df619fd2d3a6a44b125cba18d/spec/requests/moderations_spec.rb#L174

...you're basically testing the `community_name` helper, somewhat inadvertently (if that makes sense). In other words, the behavior you are testing (setting the `community_name` to `"DEV"` and then to `"Forem"`) isn't behavior that has anything to do with this _view_ -- it is something that the helper handles :)

Turning our attention back to the view, the things that you probably care about testing are that
1) the links render and have the right copy and
2) you have accounted for this conditional logic:
https://github.com/forem/forem/blob/3dd42fe8a0ebc00df619fd2d3a6a44b125cba18d/app/views/moderations/index.html.erb#L39-L41

The good news is that you already have tests for all of this :) 
https://github.com/forem/forem/blob/3dd42fe8a0ebc00df619fd2d3a6a44b125cba18d/spec/requests/moderations_spec.rb#L155-L169

I think from reading the tests, someone might expect the view to _behave_ differently depending on whether it was being rendered on the DEV forem or some other forem instance. But the view won't really behave differently -- the only thing that will change will be the `community_name`, and as we saw, that is already tested 😎 

So all of this is to say, I think you don't need to test against the "dev.to" scenario versus the "other forem" scenario for this, since the behavior of the _view_ doesn't change in any case. 😃 Back when you were checking `SiteConfig.dev_to?`, this would have made perfect sense, but now that we don't have that check anymore, the behavior of the view really is the same across the board.

I hope that this additional context makes more sense. As always, I'm more than happy to give more explanation or context if you ever want it! I never want anyone to feel like they have to make a change "just because I said so"; I care a whole lot more about explaining where I'm coming from and sharing the _why_ so that we can both learn from each other :) If you ever want me to give a more in-depth explanation, I'm 10000% down to take the time to write it out!

msarit: Ok this makes sense then! Thank you! I will complete the changes tomorrow 🙏🏽

vaidehijoshi (approved):
nice work!

Issue #12005, opened 2020-12-22
Generalize stats reporting so that backends other than DataDog can be used, if desired


wyhaines:
Currently, Forem's statistics collection is tied directly to DataDog. I would like to use New Relic instead of DataDog without requiring any customizations to the deployed code, so that a managed Forem service can support both DataDog and NewRelic (and, eventually, other providers) for stats collection.

The `DatadogStatsClient` will be turned into a `ForemStatsClient`. It is the high level interface for Forem to record metrics and events. The default mode of operation will be for the `ForemStatsClient` to use a `DatadogStatsDriver` to invoke the exact same API calls that are invoked currently.

A conditional section will be added to the `Gemfile`. If a New Relic configuration is present, it will include the New Relic gem, but if it is not, it will include the Datadog gem. The `ForemStatsClient` likewise does a conditional require for loading it's driver. Thus, if there is no New Relic configuration, everything will still use DataDog unchanged, just as it does currently. This work will be transparent to all current Forem users.

If there is a New Relic config, though, then Forem will use New Relic for recording metrics and events. The sticky wicket to the plan is really just that while the New Relic platform supports all of the things that the Forem software wants to do with recording of events and metrics, the New Relic Ruby Gem does not. A subset of things cleanly translate. The rest are a dirty translation -- the limitations of the Ruby client mean that we either send the data to New Relic in a way that may be suboptimal for our own experience interacting with the data in the New Relic Dashboard, or we go the hard route and implement code to directly use the HTTP API for New Relic to do the things that the Ruby client won't do, or we lose data. My first pass is to do cleanly what we can do cleanly, and make a best effort at everything else, with NOOP stubs for the things that are hard. The goal is to get something that can be tested and validated and that provides some value quickly, which can then be iterated on to make it better.

The actual `NewRelicStatsDriver` is to be implemented as a separate gem, instead of being bundled into Forem. That way we can make incremental improvements to the integration with New Relic without making code PRs to Forem itself, and both the maintenance and support for the New Relic integration can be cleanly separated from the maintenance and support of the Forem product itself, helping to insulate Forem from increased support burden as a result of this change.

This is being implemented as a couple of discrete PRs. The first creates the `ForemStatsClient` and the `DatadogStatsDriver`.

The second PR adds the conditional section to the `Gemfile` for enabling the use of New Relic in the presence of a New Relic configuration.

In parallel to this, a well tested gem to implement the `NewRelicStatsDriver` is being prepared.

[...]

PR #12304, opened 2021-01-16
This change abstracts the DatadogStatsClient into a ForemStatsClient.

wyhaines:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The purpose of this abstraction is to set the foundation for a subsequent PR that will allow one to use New Relic for recording Forem stats, instead of Datadog, if there is a New Relic configuration found.

This specific change creates an abstraction layer that can be built upon, without changing any actual default behavior. All specs still pass.

https://github.com/forem/forem/issues/12005

This change should result in no _functional_ changes. All stats will be reported to Datadog just as they were before.

None

## Added tests?

- [ ] Yes
- [X] No, and this is why: There were no functional changes, so all of the current specs, with their current use of the stats reporting, are still valid.

## Added to documentation?

- [ ] README
- [X] No documentation needed

alt_text


wyhaines:
Hmmm. I pulled the latest changes from master and rebased.  I didn't rerun the specs after rebase, but all of the specs fail, now, both in my branch and in the updated master.

app/lib/forem_stats_drivers/datadog_driver.rb
```diff
@@ -0,0 +1,37 @@
...
+        c.use :sidekiq
+        c.use :redis
+        c.use :rails
+        c.use :http
+      end
+
+      @driver = Datadog::Statsd.new
+    end
+
+    def increment(*args)
```
citizen428, line 21: **suggestion (non-blocking):** Same comment re delegation as above.  Makes me wonder if we want to share behavior e.g via concern:

```ruby
module ActsAsForemStatsDriver
  extend ActiveSupport::Concern

  class_methods do
    def setup_driver(&setup)
      define_method(:initialize) do
        @driver = setup.call
      end
    end
  end

  included do
    delegate :increment, :time, :gauge, to: :@driver
    alias_method :count, :increment
  end
end
```

This would greatly simplify adding more drivers while still allowing to override or alias methods, etc:

```ruby
module ForemStatsDrivers
  class DatadogDriver
    setup_driver do
      Datadog.configure do |c|
        ...
      end
      Datadog::Statsd.new
    end
 end
end
```

This is not really tested or anything, but the throwaway example I wrote in my Rails console seems to work fine:

```ruby
class ExampleService; def increment(x); x + 1; end; end

class ExampleServiceDriver
  include ActsAsForemStatsDriver
  setup_driver { ExampleService.new }
end

ExampleServiceDriver.new.increment(1)
#=> 2
ExampleServiceDriver.new.public_methods(false)
#=> [:count, :gauge, :increment, :time]
```

app/lib/forem_stats_driver.rb
```diff
@@ -0,0 +1,28 @@
...
+    # Logic will be added here for selecting the correct driver based on
+    # the existence of configuration files for the desired stats recipient.
+    ForemStatsDrivers::DatadogDriver
+  end
+
+  def initialize(*_args)
+    @driver = self.class.select_driver.new
+  end
+
+  def increment(*args)
```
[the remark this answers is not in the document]
wyhaines: Delegation is fine. I made it explicit in the beginning mostly just because it is a beginning, and so I was avoiding all magic.

app/lib/forem_stats_driver.rb
```diff
@@ -0,0 +1,28 @@
+class ForemStatsDriver
+  def self.select_driver
```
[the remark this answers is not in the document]
wyhaines: That's fine. It exists so that when there is more than just one driver implementation (which will be in the next PR), the code can just figure out from the configuration where to send stats. i.e. is there a New Relic API Key? If so, then use the New Relic driver, with the Datadog driver always being the default.

app/lib/forem_stats_drivers/datadog_driver.rb
```diff
@@ -0,0 +1,37 @@
+module ForemStatsDrivers
+  class DatadogDriver
+    attr_reader :driver
```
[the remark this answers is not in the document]
wyhaines: That's a good point.  Thanks.

app/lib/forem_stats_drivers/datadog_driver.rb
wyhaines: :eyes:  Let me do some quick testing on my end, too, as I have the New Relic driver stuff also close-to-ready so that I can deliver it as a separate PR after this foundational one is good. But, my initial impression is that I think that this would be perfectly fine.  I'll follow up soon.

wyhaines:
Insight into the detonation of the specs would be welcome. Running locally, the specs work just fine. Most of the use of the stats reporting are in the requests and workers specs, so, for example, a run of:

`bundle exec rake spec:workers spec:requests`

```

**Comment by wyhaines** [comment]
BTW, as soon as we can get this merged, I want to follow up with the driver implementation for New Relic.  🤞

**Comment by maestromac** [comment]
Hello @wyhaines thank you for this PR! I think it's containing commits that are already part of master. Can you please merge master to clear those up?

**Comment by wyhaines** [comment]
Bless Git.  At an early point, when I realized that the upstream master had diverged a lot from when I forked, I did a pull and rebase, but I must have gone a bit wrong with it, as I ended up with all of those commits in my branch as commits, and my actual change commits were duplicated in the history, once with the commits and once without them. It took a bit of sleuthing to turn everything over and verify it, but now you have my commits without all of the duplicated extras that snuck into there.

File: app/lib/forem_stats_driver.rb [hunk 5]
```diff
@@ -0,0 +1,17 @@
+class ForemStatsDriver
+  def initialize(*_args)
+    @driver = select_driver.new
+  end
+
+  delegate(:increment, :time, :gauge, to: :@driver)
+  alias count increment
```
maestromac, line 7: I don't think `#count` and `#increment` is interchangeable :thinking: . I've looked through the source ([count](https://github.com/DataDog/dogstatsd-ruby/blob/master/lib/datadog/statsd.rb#L161), [increment](https://github.com/DataDog/dogstatsd-ruby/blob/master/lib/datadog/statsd.rb#L134)). `#count` replaces, and requires 2 args while `#increment` increases requires only 1 arg.

wyhaines: Heh.  I just assumed that you knew something that I didn't in your comments, and went with it. :) My error, there.  Hang on, and I will correct this.  Thanks for bearing with me!

wyhaines:
@maestromac Let me know if there is anything else to address here.  Thank you!

app/lib/forem_stats_drivers/datadog_driver.rb
```diff
@@ -0,0 +1,19 @@
+module ForemStatsDrivers
+  class DatadogDriver
+    include ActsAsForemStatsDriver
+    setup_driver do
```
citizen428, line 4: Nice, glad to see my suggestion actually worked, it's really nice how declarative this feels!

citizen428 (approved):

rhymes (approved):
Thanks for the PR @wyhaines!

maestromac (approved):
LGTM!

Issue #10744, opened 2020-10-08
Blocking user doesn't successfully stop user's posts from appearing in feed 


michael-tharrington:
**Describe the bug**

When User A blocks User B, User A can still see User B's posts in their home feed.

**To Reproduce**

1. Go to your home feed on dev.to
2. Click onto an author's name whom you'd like to block
3. Once on the author's profile page, click the '...' icon to the right side of their name card and the block them
4. Return to home feed on dev.to and their post will still be there

**Expected behavior**

When you block someone, their posts should no longer appear in your feed.

**Additional context**

I have not tried to reproduce this behavior, but a user provided us with a video to confirm that this is happening.

Whoever works on this may want to pick up the feature request for mute functionality as well - https://github.com/forem/forem/issues/10497.

rhymes:
Might also be related to https://github.com/forem/forem/issues/5342

vaidehijoshi:
@benhalpern since you have assigned this to yourself, I am assuming you have started work on this already :) If that is the case, can you please add it to the in progress column so we can keep up-to-date with which tasks are currently being worked on? Thank you! 🙏🏽

vaidehijoshi:
Closed by https://github.com/forem/forem/pull/12307 :)

PR #12307, opened 2021-01-16
Fix and refactor hiding content when a user has blocked someone

benhalpern:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR does a few things:

- Fix the client-side check for blocked content.
  - This had broken because we changed an HTML class, but were still using it for client side reference.
- Add functionality so the blocked content is hiddent _on the server_ for the core feed.
  - At the original implementation of blocked content, this was not possible.
  - This lays the groundwork for a better overall approach to safely blocking content.
- Identified and fixed a small inefficiency in the way the home page was loaded
  - We made a useless query + sort in the home feed.

https://github.com/forem/forem/issues/10744

There are no UI changes which should require much user testing— But you can sanity check that this works as expected.

No UI changes

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [x] No documentation needed - Fixes existing expectations

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -168,13 +168,11 @@ class StoriesController < ApplicationController
 
   def handle_base_index
     @home_page = true
-    assign_feed_stories
+    assign_feed_stories unless user_signed_in? # Feed fetched async for signed-in users
     assign_hero_html
     assign_podcasts
     get_latest_campaign_articles if Campaign.current.show_in_sidebar?
     @article_index = true
-    @featured_story = (featured_story || Article.new)&.decorate
-    @stories = ArticleDecorator.decorate_collection(@stories)
     set_surrogate_key_header "main_app_home_page"
     set_cache_control_headers(600,
                               stale_while_revalidate: 30,
@@ -279,6 +277,8 @@ class StoriesController < ApplicationController
       @default_home_feed = true
       @featured_story, @stories = feed.default_home_feed_and_featured_story(user_signed_in: user_signed_in?)
     end
+    @featured_story = (featured_story || Article.new)&.decorate
+    @stories = ArticleDecorator.decorate_collection(@stories)
   end
 
   def assign_article_show_variables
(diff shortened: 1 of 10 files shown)


app/javascript/contentDisplayPolicy/hideBlockedContent.js
```diff
@@ -13,7 +13,7 @@ export default function hideBlockedContent() {
   });
 
   divsToHide.forEach((div) => {
-    if (div.className.includes('single-article')) {
+    if (div.className.includes('crayons-story')) {
```
benhalpern, line 16: This is the client-side patch of existing functionality.

app/models/user.rb
```diff
@@ -435,10 +435,6 @@ def org_admin?(organization)
 
   def block; end
 
-  def all_blocking
```
benhalpern, line 438: I essentially moved this to _one place_ on the `UserBlock` model, to reduce what the `User` cares about.

app/controllers/stories_controller.rb
```diff
@@ -168,13 +168,11 @@ def handle_page_display
 
   def handle_base_index
     @home_page = true
-    assign_feed_stories
+    assign_feed_stories unless user_signed_in? # Feed fetched async for signed-in users
```
benhalpern, line 171: This is only needed when the user is _not_ signed in, and because this ultimately turns into an array and sorts, it will be queried even if we don't use it on the view. This path is mostly cached so it wasn't causing us huge problems, but one less code path traversed on signed-in cold hits. This should reduce app memory consumption by a bit.

I consolidated all the logic related to this.

citizen428 (approved):

app/controllers/stories_controller.rb
rhymes: Would be great to have it summarized in a comment

rhymes (approved):
LGTM

PR #12308, opened 2021-01-16
Declare winner in feed_top_articles_query test

benhalpern:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This PR closes the loop on the most recent feed experiment, ripping out the experimentation code altogether, while refactoring the test code to be more general and ready to consume another a/b test in the future. Functionally this PR mostly just declares the winner (which establishes a slight change in the feed), but it also removes some dead code which is no longer part of the feed algorithm and makes some _slight_ adjustments to logic which was not strictly tested, but represents a likely obvious small improvement based on overall observations.

The experiment `top_articles_since_last_pageview_7_days_max` is running as the winner in most tests of this current a/b test. Will follow on with full evaluation. It seems unlikely that running this experiment any longer will result in a different winner.

This test presents a slightly new approach to arranging the feed: One that arranges content based on the last time you visited the site. So if you visited two days ago, it will mostly show you what you missed in the last few days. Once you have clicked on a couple posts, you'll see mostly content from the last day or so. It structures the site to be more absolute top stuff for occasional visitors and more "top recent" for more active visitors.

The test has now run longer than the intended two weeks, where the data tends to break down because randomness can set in, but at every stage of the way, this variant has either been approximately tied as the winner or clearly the winner by 5-20% depending on the test.

Here are a couple screenshots:

DEV-screenshot

DEV-other

And CodeNewbie has less overall data, and was more random, but a few key tests showed this to be a clear winner, with it being among the best in all tests.

Screen Shot 2021-02-03 at 2 07 38 PM

https://github.com/forem/forem/pull/11982

## Added to documentation?

- [x] Will follow up with evaluation of this test.
- [ ] README
- [ ] No documentation needed


app/services/articles/feeds/large_forem_experimental.rb
```diff
@@ -109,10 +106,6 @@ def score_followed_organization(article)
         user_following_org_ids.include?(article.organization_id) ? 1 : 0
       end
 
-      def score_randomness
```
benhalpern, line 112: This is no longer used based on how we now introduce a shuffle at the end.

app/services/articles/feeds/large_forem_experimental.rb
```diff
@@ -211,12 +159,12 @@ def user_following_users_ids
         @user_following_users_ids ||= (@user&.cached_following_users_ids || [])
       end
 
-      def first_half(array)
-        array[0...(array.length / 2)]
+      def first_quarter(array)
```
benhalpern, line 162: This one change takes the top _quarter_ of results instead of _half_ and presents them in a shuffled order. It's a slight modification that should complement the one we selected for the better.

fdoxyz (approved):
I can notice some of the _slight_ adjustments but nothing stands out as blocking to me, specially since we're halting the conversion tracking.

app/workers/users/record_field_test_event_worker.rb
```diff
@@ -9,20 +9,22 @@ def perform(user_id, goal)
...
-        pageview_goal(24.hours.ago, "DATE_PART('hour', created_at)", 4, "user_views_article_four_hours_in_day")
-        pageview_goal(14.days.ago, "DATE(created_at)", 9, "user_views_article_nine_days_in_two_week")
-        pageview_goal(5.days.ago, "DATE_PART('hour', created_at)", 12, "user_views_article_twelve_hours_in_five_days")
-      when "user_creates_comment" # comments goal. Only page views and comments are currently active.
-        field_test_converted(@experiment, participant: @user, goal: goal) # base single comment goal.
-        comment_goal(7.days.ago, "DATE(created_at)", 4, "user_creates_comment_four_days_in_week")
-      else
-        field_test_converted(@experiment, participant: @user, goal: goal) # base single comment goal.
+      FieldTest.config["experiments"].each_key do |key|
+        @experiment = key.to_sym # Not currently running. Update this to current experiment.
```
benhalpern, line 13:
```suggestion
        @experiment = key.to_sym
```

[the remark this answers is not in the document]
benhalpern: Good call

mstruve (commented):
Let's get those conflicts resolved and then this should be good to go.

mstruve (approved):
🚀

PR #12309, opened 2021-01-16
Improve cumulative layout shift on comment reactions + optimizations

benhalpern:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Cumulative layout shift refers the sudden jump of elements on a page when any change happens after initial render.

Details here: https://web.dev/cls/

This is not only a bad experience for users, Google has warned that this will become a bigger ranking factor:

Details here:
https://www.searchenginejournal.com/cumulative-layout-shift/371946/
https://www.popwebdesign.net/popart_blog/en/2020/09/cumulative-layout-shift-how-to-rank-on-google-in-2021/

I'm not sure it is covered in that post, but I believe Google announced they will step up how much this matters in May 2021.

In general this is _a great thing_ for Forem because we already prioritize user experience and have minimal CLS issues, however, we do have some.

This PR addresses this one:

https://user-images.githubusercontent.com/3102842/104824544-a24bb380-5820-11eb-8362-8e48dd8ed405.mov

Upon loading the page, we asynchronously fetch the number of reactions in the comment trees. This is nice so that we don't have to bust the whole page every time a reaction is made, however as currently implemented, it causes a CLS.

In examining the ways to address this, I believe the best route is what I have chosen to do:

- We include the current public reactions count inline, knowing it may be cached, and use the async fetch as a way to _adjust_ the reactions. This will improve UX, as most of the time there will be no jump at all, and if the reaction count has changed recently it will act as a small _adjustment_, say from 7 to 8 reactions.

As Forems grow (like DEV), they have great potential to become a _knowledge base_. That means that some old threads can stick around to be useful for a long time. However, over time, posts are less likely to have new reactions popping up every few minutes.

This PR takes advantage of this and reduces that async query we make when there has been no new comment activity for a week. That is to say: We do not fetch new _async_ reaction numbers for posts nobody is commenting on anymore. This will reduce one request for _many_ posts.

https://github.com/forem/forem/issues/12310

This should only improve possible issues.

## Added tests?

- [x] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [ ] No documentation needed


app/assets/javascripts/initializers/initializeCommentsPage.js.erb
```diff
@@ -3,74 +3,82 @@ var iconSmallThread = `<svg width="24" height="24" class="crayons-icon" viewBox=
...
-              var publicReactionCounts = responseObj.public_reaction_counts;
-              for (var i = 0; i < reactions.length; i++) {
-                var buttForComment = document.getElementById('button-for-comment-' + reactions[i].reactable_id);
-                if (buttForComment) {
-                  buttForComment.classList.add('reacted');
+
+    // When no logged in user only need to fetch reaction numbers if there has been recent comment activity.
+    // If no recent comments, we do not need the precision of fetching latest counts async. Static value is almost certainly good enough.
+    // Always fetch when user is logged in because we need state of whether the user has reacted.
+    if (hasRecentCommentActivity !== "false" || userStatus === 'logged-out') {
```
benhalpern, line 15: If you don't have `Hide whitespace changes`, this may seem like a bigger change than it is. It's just one line added and a bunch of indentation.

app/helpers/comments_helper.rb
```diff
@@ -31,6 +31,16 @@ def should_be_hidden?(comment, root_comment)
...
 
+  def like_button_text(comment)
+    if comment.public_reactions_count.zero?
+      "Like"
+    elsif comment.public_reactions_count == 1
+      "&nbsp;like"
+    else
+      "&nbsp;likes"
+    end
+  end
```
atsmith813, line 42: **nitpick (non-blocking):** maybe we could use a case statement here if we feel it's easier to read:

```ruby
case comment.public_reactions_count
when 0
  "Like"
when 1
  "&nbsp;like"
else
  "&nbsp;likes"
end
```

Or if we want to import `pluralize` we can do something like:
```ruby
return "Like" if comment.public_reactions_count.zero?

"&nbsp;#{pluralize(comment.public_reactions_count, "like")}"

```

app/helpers/comments_helper.rb
```diff
@@ -31,6 +31,16 @@ def should_be_hidden?(comment, root_comment)
...
   end
 
+  def like_button_text(comment)
+    if comment.public_reactions_count.zero?
+      "Like"
+    elsif comment.public_reactions_count == 1
+      "&nbsp;like"
+    else
+      "&nbsp;likes"
+    end
```
atsmith813, line 41: **nitpick (non-blocking):** maybe we could use a case statement here if we feel it's easier to read:

```ruby
case comment.public_reactions_count
when 0
  "Like"
when 1
  "&nbsp;like"
else
  "&nbsp;likes"
end
```

Or if we want to import `pluralize` we can do something like:
```ruby
return "Like" if comment.public_reactions_count.zero?

"&nbsp;#{pluralize(comment.public_reactions_count, "like")}"

```

docs/technical-overview/architecture.md
```diff
@@ -22,6 +22,22 @@ We also use server-side caching: [Rails caching][rails_caching]. If you see
...
+
+### Examples of this
+
+- We use the `estimated_count` for a more efficient query of registered users on
+the home page. We have deemed that this is probably close enough.
+- On posts comment trees without recent comments, we do not asynchronously fetch
+the absolute latest individual reaction counts for logged-out users because this
+number is likely to be correct without the async call, and if it is off-by-one, we
+can make the choice that it is not important that it be more precise than this. 
+
```
atsmith813, line 39: :heart:

app/views/comments/_comment_footer.html.erb
```diff
@@ -6,8 +6,12 @@
     title="heart">
     <%= inline_svg_tag("small-heart.svg", aria: true, class: "crayons-icon reaction-icon not-reacted", title: "Favorite heart outline button") %>
     <%= inline_svg_tag("small-heart-filled.svg", aria: true, class: "crayons-icon reaction-icon--like reaction-icon reacted", title: "Favorite heart outline button") %>
-    <span class="reactions-count hidden"></span>
-    <span class="reactions-label hidden m:inline-block">Like</span>
+    <span class="reactions-count">
+      <%= comment.public_reactions_count if comment.public_reactions_count.positive?%>
```
juliannatetreault, line 10: **_Nitpick_**: It looks like we could use a space here between `.positive?` and the closing ERB tag:

```suggestion
      <%= comment.public_reactions_count if comment.public_reactions_count.positive? %>
```

app/assets/javascripts/initializers/initializeCommentsPage.js.erb
```diff
@@ -3,74 +3,82 @@ var iconSmallThread = `<svg width="24" height="24" class="crayons-icon" viewBox=
...
-              var allNodes = document.getElementsByClassName('single-comment-node');
-              var publicReactionCounts = responseObj.public_reaction_counts;
-              for (var i = 0; i < reactions.length; i++) {
-                var buttForComment = document.getElementById('button-for-comment-' + reactions[i].reactable_id);
-                if (buttForComment) {
-                  buttForComment.classList.add('reacted');
+
+    // When no logged in user only need to fetch reaction numbers if there has been recent comment activity.
+    // If no recent comments, we do not need the precision of fetching latest counts async. Static value is almost certainly good enough.
+    // Always fetch when user is logged in because we need state of whether the user has reacted.
```
vaidehijoshi, line 14: Since we're in this code and adding some comments...can we use JSDoc here instead? 😃 This function has a lot going on, and definitely could use some documentation. Check out the `initializeBroadcast.js` [file](https://github.com/forem/forem/blob/master/app/assets/javascripts/initializers/initializeBroadcast.js ) which has some helpful examples that you can follow :)

I think our future selves will really appreciate it 🙏🏽

app/helpers/comments_helper.rb
vaidehijoshi: I like either of these, but pluralize seems like a nice solution here :)

docs/technical-overview/architecture.md
```diff
@@ -22,6 +22,22 @@ We also use server-side caching: [Rails caching][rails_caching]. If you see
 `Rails.cache` or `<%= cache ... %>`, this is code affected in production by
 caching.
 
+## Content precision
+
+In some situations we may want more precise content than in others. Often when
+we do not need a precise number, it offers an opportunity to either estimate
+the count or bust the cache less frequently.
```
vaidehijoshi, line 29:
```suggestion
the content or bust the cache less frequently.
```
Or, if you want to keep count here, maybe change line 25 to read "Count precision"? 🤔 Right now, it sounds like this should apply to things beyond just estimating count.

docs/technical-overview/architecture.md
```diff
@@ -22,6 +22,22 @@ We also use server-side caching: [Rails caching][rails_caching]. If you see
...
+
+### Examples of this
```
vaidehijoshi, line 31:
```suggestion
### Examples
```

docs/technical-overview/architecture.md
```diff
@@ -22,6 +22,22 @@ We also use server-side caching: [Rails caching][rails_caching]. If you see
...
+
+- We use the `estimated_count` for a more efficient query of registered users on
+the home page. We have deemed that this is probably close enough.
+- On posts comment trees without recent comments, we do not asynchronously fetch
```
vaidehijoshi, line 35:
```suggestion
- On posts and comment trees without recent comments, we do not asynchronously fetch
```
I'm _guessing_ this is missing?

benhalpern:
> I like either of these, but pluralize seems like a nice solution here :)

@vaidehijoshi

Unless I'm missing something, using `pluralize` itself doesn't have a "without the count" option.... And since this is split up in the HTML and stuff it doesn't cleanly fit in...

The pluralize page comments are actually all about this it seems: https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize

So I could leave as is or add a special helper to our application in general for stuff like this.

app/helpers/comments_helper.rb
benhalpern: I commented below before seeing this specific thread. Pluralize outputs the word _and_ the number so isn't on its own the thing, eh?

atsmith813: @benhalpern you're right, that's my mistake! I gave the wrong example. We'd call pluralize on the string itself if we wanted to go this route:

```ruby
return "Like" if comment.public_reactions_count.zero?

"&nbsp;#{"like".pluralize(comment.public_reactions_count}"
```

app/assets/javascripts/initializers/initializeCommentsPage.js.erb
benhalpern: Thanks for this! I will definitely make much better use of jsdoc in the future.

app/helpers/comments_helper.rb
benhalpern: I added the case statement

vaidehijoshi (approved):

PR #12316, opened 2021-01-18
Update NavigationLink handling

citizen428:

- [x] Refactor
- [X] Creator experience

## Description

When a Forem switches over from the `forem.cloud` subdomain to its live domain, the Navigation Links in the sidebar don't update. To fix this @benhalpern suggested the following plan:

* [X] Allow relative URLs, like `/contact`.
* [X] Automatically strip the protocol and domain part from the URL if it matches the current ones on save.
* [x] Add a data update script for existing Forems to make the Navigation Links relative.

https://github.com/forem/internalEngineering/issues/179

### Screenshots

Updated copy to mention relative URLs explicitly:

image

### QA instructions

1. As admin go to http://localhost:3000/admin/navigation_links
2. Try to enter an invalid URL (e.g. "test"), this should not work.
3. Try to enter a relative URL (e.g. "/test"), this should work.
4. Enter an invalid external URL (e.g. "ftp://example.com"), this should not work. 
5. Enter an external URL (e.g. "https://example.com"), this should work.

None

## Added tests?

- [X] Yes

## Added to documentation?

- [X] No documentation needed. I checked the admin guide, all examples already use relative URLs. Together with the updated copy, this seems sufficient to me.

- [ ] Verify the data update script successfully runs across the fleet.

### DIFF
--- a/app/models/navigation_link.rb
+++ b/app/models/navigation_link.rb
@@ -1,10 +1,33 @@
 class NavigationLink < ApplicationRecord
   SVG_REGEXP = /<svg .*>/i.freeze
 
+  before_validation :allow_relative_url, if: :url?
+  before_save :strip_local_hostname, if: :url?
+
   validates :name, :url, :icon, presence: true
   validates :url, url: { schemes: %w[https http] }, uniqueness: { scope: :name }
   validates :icon, format: SVG_REGEXP
   validates :display_only_when_signed_in, inclusion: { in: [true, false] }
 
   scope :ordered, -> { order(position: :asc, name: :asc) }
+
+  private
+
+  # We want to allow relative URLs (e.g. /contact) for navigation links while
+  # still going through the normal validation process.
+  def allow_relative_url
+    parsed_url = URI.parse(url)
+    return unless parsed_url.relative? && url.starts_with?("/")
+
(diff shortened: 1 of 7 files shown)


app/models/navigation_link.rb
```diff
@@ -1,10 +1,33 @@
 class NavigationLink < ApplicationRecord
   SVG_REGEXP = /<svg .*>/i.freeze
 
+  before_validation :allow_relative_url, if: :url?
```
citizen428, line 4: "Conditional callbacks in code by Michael?" I hear you say?  I'm as surprised as the next person, but the gem we use for URL validation is rather comprehensive and has fewer surprises than `URI.regexp`:

```ruby
URI.regexp(%w[http https]).match?("https://")
#=> true

navigation_link.url = "https://"
#=> "https://"
navigation_link.valid?
#=> false
 navigation_link.errors_as_sentence
#=> "Url is not a valid URL"
```

That's why I decided to just add some pre and post-processing around it instead of writing my own custom validator class (well, technically I did start with the custom validator class, but abandoned that approach quickly).

rhymes (approved):
Cool approach!

benhalpern (approved):

Issue #11936, opened 2020-12-17
Going to account page shows error page, always


ssg:
**Describe the bug**
Error page on account settings page.

**To Reproduce**
1. Login to dev.to and go to <https://dev.to/settings/account>
2. Page shows error and asks me to file a bug report

**Expected behavior**
Account page should be shown.

**Screenshots**
Standard error page with turning robot animation. 

**Desktop (please complete the following information):**
Windows 10 20H2, Edge browser

**Additional context**
My username on Dev.To is "esesci". I have login setup through GitHub integration and I feel like it might be related.

This is consistently reproducible.

LE-HU:
Hello, is it possible to test the issue on different browsers? (disabling add-on's is a good idea)
You can also try to clear browser cache on edge.
Do you have twitter account connected to dev.to? Settings page renders that too.

Zhao-Andy:
Hey @ssg, thanks for the report! Looks like this is an error that the settings page does not properly handle which leads to the 500.

In the settings page we try to show you your emails associated with whatever social provider you've linked (Twitter / GitHub). In this case, your Twitter account for whatever reason does not have the data needed for us to pull, and our chained method of `auth_data_dump.info.email` errors out with `undefined method info on nil 'auth_data_dump'`.

For your case specifically, I've removed your Twitter association from your account. Feel free to add it back in if you'd like via https://dev.to/settings/account.

As for the issue, here is the backtrace of the error: https://app.honeybadger.io/fault/66984/26ac848fae65a336e73fe6ea2b6d55cb

I think here we can update the view `app/views/users/_account_providers_emails.html.erb:15` to:
```
<span><%= identity&.email ? "No email found! Please try relinking your #{identity.provider} account to avoid errors." : identity.email %></span>
```

ssg:
Thank you @Zhao-Andy for looking at this! I'm leaving this open so it can be addressed in code for other users too.

LE-HU:
Hello, I've spent some time investigating in order to find out more about this one.
- thank you @Zhao-Andy for the logs.
- link to [Identity.email function](https://github.com/forem/forem/blob/361d09dd42c4bf510c5dd33f350ba34b068861c1/app/models/identity.rb#L41)
- I was able to replicate the issue locally by connecting a twitter account to local forem, and setting the email function to
```ruby
def email
  nil.info.email
end
```
- I can browse the whole app with the exception of /settings/account. Existing rspec identity_spec.rb catches the change.
[...]

PR #12322, opened 2021-01-18
secure Identity#email in case of missing auth_data_dump hash (#11936)

LE-HU:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Rare occurence of missing auth_data_dump from the oauth provider caused email method to return NoMethodError. 
Fix uses safe navigation operator to ensure presence of auth_data_dump:info:email to retrieve email, prints "no email found, please relink.." string when no email found.

#11936 

Tested locally overwriting auth_data_dump = nil, what replicated the reported issue.
No email found string successfuly fills in the #{provider}.

N/A
## Added tests?

- [ ] Yes
- [X] No, and this is why: existing identity_spec.rb catches happy path. I did not add a test for "no email found" string return. If needed please let me know.
image

## Added to documentation?

- [ ] README
- [x] No documentation needed

### DIFF
--- a/app/models/identity.rb
+++ b/app/models/identity.rb
@@ -39,10 +39,9 @@ class Identity < ApplicationRecord
   end
 
   def email
-    if auth_data_dump&.info&.email
-      auth_data_dump.info.email
-    else
-      "No email found. Please relink your #{provider} account to avoid errors."
-    end
+    auth_data_dump = nil
+    return auth_data_dump.info.email if auth_data_dump&.info&.email
+
+    "No email found. Please relink your #{provider} account to avoid errors."
   end
 end
--- a/app/models/identity.rb
+++ b/app/models/identity.rb
@@ -39,7 +39,6 @@ class Identity < ApplicationRecord
   end
 
   def email
-    auth_data_dump = nil
     return auth_data_dump.info.email if auth_data_dump&.info&.email
 
     "No email found. Please relink your #{provider} account to avoid errors."
--- a/app/models/identity.rb
+++ b/app/models/identity.rb
@@ -39,6 +39,10 @@ class Identity < ApplicationRecord
   end
 
   def email
-    auth_data_dump.info.email
+    if auth_data_dump&.info&.email
+      auth_data_dump.info.email
+    else
+      "No email found. Please relink your #{provider} account to avoid errors."
+    end
   end
 end
--- a/app/models/identity.rb
+++ b/app/models/identity.rb
@@ -42,8 +42,6 @@ class Identity < ApplicationRecord
   end
 
   def email
-    return auth_data_dump.info.email if auth_data_dump&.info&.email
-
-    format(NO_EMAIL_MSG, provider: provider)
+    auth_data_dump&.info&.email || format(NO_EMAIL_MSG, provider: provider)
   end
 end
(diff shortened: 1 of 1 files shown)


app/models/identity.rb
```diff
@@ -39,6 +39,10 @@ def self.build_from_omniauth(provider)
   end
 
   def email
-    auth_data_dump.info.email
+    if auth_data_dump&.info&.email
+      auth_data_dump.info.email
+    else
+      "No email found. Please relink your #{provider} account to avoid errors."
+    end
```
Zhao-Andy, line 46: I think you can refactor this a bit, since using `&.` returns `nil` early on. Maybe turn this into one line, and use a guard statement with an `unless` instead?

Zhao-Andy (changes requested):
Thanks a lot for your PR! I think you're off to a great start, and this `if-else` statement works, but can be refactored a bit.

app/models/identity.rb
LE-HU: Thank you @Zhao-Andy for quick reply with good suggestions.
I've implemented a guard clause.

app/models/identity.rb
```diff
@@ -39,6 +39,8 @@ def self.build_from_omniauth(provider)
   end
 
   def email
-    auth_data_dump.info.email
+    return auth_data_dump.info.email if auth_data_dump&.info&.email
+
+    "No email found. Please relink your #{provider} account to avoid errors."
```
Zhao-Andy, line 44: Another tip here, you can make turn this into a constant to save a bit of memory, like so:
```ruby
# constants are usually defined near the top of the class
class Identity < ApplicationRecord
  NO_EMAIL_MSG = "No email found. Please relink your %<provider> account to avoid errors.".freeze
  
  # ... down below
  
  def email
    return auth_data_dump.info.email if auth_data_dump&.info&.email

    format(MESSAGES[:no_email_found], provider: provider)
  end
end
```

The `format` method will convert the `provider` key value pair just like in normal string interpolation, allowing us to use constants with conditional strings. You can also see `app/models/user.rb` as an example:
https://github.com/forem/forem/blob/0cac61024f63a93da0965a8a202a4cc1952b575d/app/models/user.rb#L76-L82

app/models/identity.rb
```diff
@@ -39,6 +39,8 @@ def self.build_from_omniauth(provider)
   end
 
   def email
-    auth_data_dump.info.email
+    return auth_data_dump.info.email if auth_data_dump&.info&.email
```
Zhao-Andy, line 42: Ah sorry, I misunderstood this earlier. 😅 I think this could be refactored like so:
```
auth_data_dump&.info&.email || "the message"
```

Zhao-Andy (commented):
That definitely works! I misunderstood something earlier about the guard clause -- sorry about that. Left a couple more tips :)

app/models/identity.rb
LE-HU: thanks! will take a look, meanwhile format(..) added - the **%<provider>s** with the s at the end is required to succesfully parse a String, works good tested locally.

LE-HU:
Changes made, thanks for the clever tip to use || as a control flow. Tested locally - works as intended.

fdoxyz (approved):
I like and agree with all the suggestions here. This LGTM

Zhao-Andy (approved):
Awesome job!

PR #12323, opened 2021-01-18
Refactor magic numbers for `/comments` rendering

cyrillefr:
Refactor magic numbers for `/comments` rendering 

- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [X] Optimization
- [ ] Documentation Update

## Description

There are some hard coded values (250 and 8) that determine the way list of comments are displayed in the comment section for a user.
This PR to refactor via removing theses hard coded values and replace them with constants instead.

## Added tests?

- [X] Yes
      have not been included_

## Added to documentation?

- [X] No documentation needed


app/helpers/comments_helper.rb
```diff
@@ -1,4 +1,7 @@
 module CommentsHelper
+  MAX_NBR_OF_COMMENTS_TO_RENDER = 250
+  MIN_NBR_OF_COMMENTS_TO_RENDER = 8
```
fdoxyz, line 3: **non-blocking suggestion**: I think these could just read `MAX_COMMENTS_TO_RENDER` and `MIN_COMMENTS_TO_RENDER`

app/controllers/stories_controller.rb
```diff
@@ -1,4 +1,6 @@
 class StoriesController < ApplicationController
+  include CommentsHelper
```
fdoxyz, line 2: **Suggestion:** Instead of including the comments helper maybe this is better suited to be managed by a `StoriesHelper`?

Another idea could be that if this include is only required for a single call then it could be worth making this a class method instead? This way we can avoid the include and make the call like this:

```ruby
comment_count = CommentsHelper.comments_to_render(params[:view])
```

Totally open to hear other team members opinion on this though

fdoxyz (commented):
Thank you for the contribution @cyrillefr. I'm onboard with making some of these changes to move the logic out from the views and into helpers. left some initial comments but I would like to hear other team members opinion on this as well.

app/helpers/comments_helper.rb
atsmith813: +1 to this suggestion!

app/controllers/stories_controller.rb
atsmith813: +1 to theses ideas!

app/helpers/comments_helper.rb
```diff
@@ -31,6 +34,22 @@ def should_be_hidden?(comment, root_comment)
...
+  end
+
+  def number_of_comments_to_render
+    MAX_NBR_OF_COMMENTS_TO_RENDER
+  end
+
+  def comments_to_render(view)
+    view == "comments" ? MAX_NBR_OF_COMMENTS_TO_RENDER : MIN_NBR_OF_COMMENTS_TO_RENDER
+  end
+
```
atsmith813, line 52: **suggestion (non-blocking):** what if we changed the name of the `comments_to_render` method? When I see that, I expect it to return `Comment` objects, but it actually returns an integer.

I was going to suggest `number_of_comments_to_render` as a name, but I see that is also being used. That said, it looks like that method is returning the constant so we can refer to the constant itself instead of using a method name, or change both method names.

spec/system/user/view_user_index_spec.rb
```diff
@@ -113,4 +117,12 @@ def shows_comments
       end
     end
   end
+
+  def shows_last_comments
+    stub_const("CommentsHelper::MAX_NBR_OF_COMMENTS_TO_RENDER", 1)
+    visit "/#{user.username}"
+    within("#substories .profile-comment-card .pt-3 .fs-base") do
+      expect(page).to have_content("View last 1 Comments")
```
atsmith813, line 125: **suggestion (non-blocking):** maybe we can use the `pluralize` method in the views and then update this spec to read:
```ruby
expect(page).to have_content("View last 1 Comment")
```

app/helpers/comments_helper.rb
cyrillefr: That is fine by me. I will change it.

app/helpers/comments_helper.rb
cyrillefr: Hi @atsmith813 , I see your point. The thing is, there is 2 cases where you display comments(based on the 2 Constants), so I did not find a way to name 2 "number_of_comments".

Maybe I could use for the last(`comments_to_render`) this: `comment_count`. It implies correctly a return of a numeric, and in the controller, I use this very variable as a return for the `comments_to_render` method. Would that be easier to read?

Regarding the Constant, I prefer to use an accessor rather than to have Helper::CONSTANT in the view.

app/controllers/stories_controller.rb
cyrillefr: Yes, seems a bit strange to include a CommentHelper in Stories controller, but it is because we do not get  there via the comment controller. Though, I feel these methods should belong to the comment helper.

But too get rid of this include,  yes, I can change the call with a static method.

spec/system/user/view_user_index_spec.rb
cyrillefr: I do concur. It will be better with pluralize.

app/helpers/comments_helper.rb
atsmith813: Great idea @cyrillefr!

app/controllers/stories_controller.rb
rhymes: I think the best solution here is to use the builtin `helpers` proxy, the call would become `helpers.comments_to_render(params[:view])`

rhymes (commented):
Thanks for taking this on @cyrillefr, I left a note on how to avoid including methods or using class methods

app/controllers/stories_controller.rb
cyrillefr: Did not know that. Settled :)

atsmith813 (approved):

fdoxyz (approved):

Issue #712, opened 2018-09-21
Twitter Embed Doesn't Update Stats


ahmadawais:
## Bug Report

Folks the embedded tweets do not get updated as far as the stats are concerned. If there were 10 likes when you embedded a tweet, it stays that way even when there are more likes. Is it just me or everyone?

### Expected Behavior

Twitter updated stats should be displayed.

### Additional Info (device/browser, helpful links)

Chrome/MacOS

nektro:
This is because the code by Twitter given to embed tweets does not connect to their live updates service. It simply creates an `<iframe>` with the static data from an API call.
https://help.twitter.com/en/using-twitter/how-to-embed-a-tweet

jessleenyc:
Hey sorry for the delay on this. This is approved for anyone who wants to work on it but we want to continue using our liquid tags for the embed and not through the official Twitter embed.

brarnaudovski:
@jessleenyc Can I start with this? 
If I did my research correctly, I'll need some background job, which eventually fetch all tweets (from `Tweet` table, as dev.to stores all the tweets in articles), and will make call through `TwitterBot` to fetch the latest data, and update the counters...

rhymes:
@brarnaudovski I wonder what would be the policy around that. Fetching **all** tweets everytime would probably be too much. We need to think of the right combination of how many tweets and how far back the logic has to go but the problem of "how far back" is that an old tweet might become popular again...

What do you think?

Zhao-Andy:
@rhymes I think _way_ back then we discussed fetching tweets only for articles that have been viewed recently. When we were having that discussion, we didn't have the `PageView` model yet. Maybe in the `PageViewController` we could create a job to update that article's tweets?

rhymes:
@Zhao-Andy That would be an interesting trick! 

I like the idea, the first user gets the "stale count", the next one (or at least the one that see the article when the async job is done) will see the updated count

brarnaudovski:
@Zhao-Andy I beleive it's a great idea. I will try to implement this. As I am still learning the project concept, maybe I'll need help, when the `PageViewController::create\update` is called, so I can test this...

abraham:
Something else to keep in mind is that older tweets will see a lot less activity. For example a tweet less than 24 hours old could probably be updated every hour or so while a tweet that's a couple of months old probably only needs to be updated every couple of months.

brarnaudovski:
[...]

PR #12327, opened 2021-01-18
Fix image assets in tweets and remove like/retweet counts

benhalpern:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This fixes an asset path bug currently on Forem Cloud... We fixed a similar issue on GitHub in the past.

Screen Shot 2021-01-18 at 12 53 10 PM

While I was here I also decided to make a change to remove the like and retweet counts on tweets. This closes a long outstanding issue of our tweets not updating because we serve them as static. Given the immutable nature of tweets, I think this is fine. In fact, having a simple static version of the tweet for display seems much better overall for what should be in a Forem post. Click through for the full live content and all the associated JS on twitter.

I will always be against adding the edit button on Twitter. Tweets were always meant to be immutable. 😄

This adjustment meets all the standards of Twitter's [display guidelines](https://developer.twitter.com/en/developer-terms/display-requirements)

Screen Shot 2021-01-18 at 12 55 46 PM

I believe this _does not_ need an update script to change the counts issue, as I think the content difference here is minimal and we can have the adjustment flush out of the system naturally.

Closes https://github.com/forem/forem/issues/712

## Added tests?

- [x] Yes
      have not been included_


atsmith813 (approved):

fdoxyz (approved):

rhymes (approved):
Let's do this!

PR #7049, opened 2020-04-02
Setup User Elasticsearch Doc Syncing

mstruve:
## What type of PR is this? (check all applicable)
- [x] Refactor
- [x] Bug Fix

## Description
Keep users and ALL their related Elasticsearch documents in sync when they are updated. If people are for this approach then I plan to setup a data sync class like this for each model we index. 

Evntually I would love to move this all out of a callback and have a single flow to update everything but we are not quite there yet. I figured the second best option was to contain this sync logic into its own class. 

## Added tests?
- [x] yes

alt_text

PR #12330, opened 2021-01-19
Reindex comments' Elasticsearch doc when commentable is updated

maestromac:
- [x] Bug Fix

## Description
- Comment's Elasticsearch doc currently has a publish status that defaults to `true`. This updates that to the status of the comment's commendable (ie. articles and podcasts).
- Unpublishing an article now reindexes its comment to reflect the publish status.

n/a

0. Pull down this PR and startup local server _with Sidekiq and Elasticsearch_ running.
1. Create a brand new article, 
1. Add a unique comment to it.
1. Try searching for the comment, you should be able to see the search result.
1. Now unpublished the article you just created.
1. You should no longer be able to search for your old comment.  
## Added tests?
- [x] yes
## Added to documentation?
- [x] no documentation needed

Monitor the status of this DataUpdateScript


app/serializers/search/comment_serializer.rb
```diff
@@ -9,8 +9,8 @@ class CommentSerializer < ApplicationSerializer
       comment.class.name
     end
     attribute :hotness_score, &:score
-    attribute :published do |_comment|
-      true
+    attribute :published do |comment|
```
mstruve, line 12: 🔥

app/models/article.rb
```diff
@@ -100,6 +100,9 @@ class Article < ApplicationRecord
   after_update_commit :update_notifications, if: proc { |article|
                                                    article.notifications.any? && !article.saved_changes.empty?
                                                  }
+  after_update_commit :reindex_comments_to_elasticsearch, if: proc { |article|
```
mstruve, line 103: We actually have a [whole syncing pattern setup for models and their related Elasticsearch docs](https://github.com/forem/forem/pull/7049/files). I would lean towards using that same pattern here so that we stick with convention. This will also allow us to update comments for other possible reasons in the future if we need to.

lib/data_update_scripts/20210118194138_resync_unpublished_articles_comments_elasticsearch_document.rb
```diff
@@ -0,0 +1,9 @@
+module DataUpdateScripts
+  class ResyncUnpublishedArticlesCommentsElasticsearchDocument
+    def run
+      Article.unpublished.where.not(comments_count: 0).each do |article|
```
mstruve, line 4: Only 1230 of these in DEV so this should be good to run in a single go I think. Keep an eye on it in case it fails or gets interrupted by a deploy, we will then want to make sure to finish it by hand.

mstruve (commented):
One suggestion otherwise LOVE this update!!!!

app/models/article.rb
maestromac: Thank you! TIL! That's such a nice pattern. I'm also adding 'title' to the sync reasons too.

lib/data_update_scripts/20210118194138_resync_unpublished_articles_comments_elasticsearch_document.rb
maestromac: Will do

app/services/data_sync/elasticsearch/article.rb
```diff
@@ -0,0 +1,20 @@
...
+      ].freeze
+
+      SHARED_FIELDS = %i[
+        published
+        title
+      ].freeze
+
+      private
+
+      def comments
```
mstruve, line 15: I don't think you need this. I think the sync_related_documents will handle this fine
```ruby
def sync_related_documents
  self.class::RELATED_DOCS.each do |relation_name|
    __send__(relation_name).find_each(&:index_to_elasticsearch)
  end
end
```

maestromac: Just test it, it wouldn't work without it. If I comment it out and run it locally, I would get
```
NoMethodError:
  undefined method `comments' for #<DataSync::Elasticsearch::Article:0x000055ebbbca78b0>
```

If I understand this correctly, things listed in `RELATED_DOCS` is called as a method that I need to provide. Did I miss a step to make this work without that?

mstruve: AH, we used delegate! 
```
delegate :comments, to: :@updated_record
```

maestromac: Updated!

app/services/data_sync/elasticsearch/article.rb
```diff
@@ -0,0 +1,16 @@
+module DataSync
+  module Elasticsearch
+    class Article < Base
+      RELATED_DOCS = %i[
+        comments
+      ].freeze
+
+      SHARED_FIELDS = %i[
+        published
+        title
```
mstruve, line 10: Good call on this!

mstruve (approved):

benhalpern (approved):

rhymes (approved):
LGTM

PR #12331, opened 2021-01-19
Don't try to unsub users when destroying if Mailchimp is not enabled

Zhao-Andy:
- [x] Refactor
- [x] Bug Fix

## Description
This fixes a bug where when destroying a user, we try to unsubscribe them from newsletters that don't exist. This previously had a runtime error with the Mailchimp API where a blank/invalid Mailchimp API key was given when trying to destroy a user.

For the most part, new Forem communities do not have Mailchimp set up. This PR adjusts for that typical situation.

Closes https://github.com/forem/internalCommunity/issues/2

1. Start `rails console --sandbox`
2. Set `SiteConfig.mailchimp_api_key = ""`
3. Attempt to destroy a user: `User.first.destroy`
4. See that it destroys the user successfully

No UI changes

## Added tests?
- [x] No, and this is why: the default test environment already has no Mailchimp API key and skips over this situation. But maybe that means we _should_ write a test for this, since it's still untested?

## Added to documentation?
- [x] No documentation needed

No

### DIFF
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -227,10 +227,6 @@ class SiteConfig < RailsSettings::Base
   end
   singleton_class.__send__(:alias_method, :apple_secret, :apple_key)
 
-  def self.mailchimp_newsletter_enabled?
-    mailchimp_api_key.present? && mailchimp_newsletter_id.present?
-  end
-
   # To get default values
   def self.get_default(field)
     get_field(field)[:default]
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -227,6 +227,10 @@ class SiteConfig < RailsSettings::Base
   end
   singleton_class.__send__(:alias_method, :apple_secret, :apple_key)
 
+  def self.mailchimp_newsletter_enabled?
+    mailchimp_api_key.present? && mailchimp_newsletter_id.present?
+  end
+
   # To get default values
   def self.get_default(field)
     get_field(field)[:default]
--- a/app/models/site_config.rb
+++ b/app/models/site_config.rb
@@ -229,6 +229,6 @@ class SiteConfig < RailsSettings::Base
 
   # To get default values
   def self.get_default(field)
-    get_field(field).dig(:default)
+    get_field(field)[:default]
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -259,7 +259,7 @@ class User < ApplicationRecord
   before_validation :set_username
   before_validation :strip_payment_pointer
   before_create :set_default_language
-  before_destroy :unsubscribe_from_newsletters, prepend: true
+  before_destroy :unsubscribe_from_newsletters, prepend: true, if: -> { SiteConfig.mailchimp_newsletter_enabled? }
   before_destroy :destroy_follows, prepend: true
 
   # NOTE: @citizen428 Temporary while migrating to generalized profiles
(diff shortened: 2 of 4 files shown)


app/models/site_config.rb
```diff
@@ -227,8 +227,12 @@ def self.apple_key
...
   singleton_class.__send__(:alias_method, :apple_secret, :apple_key)
 
+  def self.mailchimp_newsletter_enabled?
+    mailchimp_api_key.present? && mailchimp_newsletter_id.present?
+  end
+
   # To get default values
   def self.get_default(field)
-    get_field(field).dig(:default)
+    get_field(field)[:default]
```
Zhao-Andy, line 236: This is a Rubocop correction.

app/services/mailchimp/bot.rb
```diff
@@ -65,7 +65,7 @@ def resubscribe_to_newsletter
     end
 
     def manage_community_moderator_list
-      return false unless user.has_role?(:trusted)
+      return false unless user.has_role?(:trusted) || SiteConfig.mailchimp_community_moderators_id.blank?
```
Zhao-Andy, line 68: These guard clauses might be overkill, but it seemed to make sense to me -- don't check if we don't have the newsletter enabled.

app/models/user.rb
```diff
@@ -259,7 +259,7 @@ class User < ApplicationRecord
   before_validation :set_username
   before_validation :strip_payment_pointer
   before_create :set_default_language
-  before_destroy :unsubscribe_from_newsletters, prepend: true
+  before_destroy :unsubscribe_from_newsletters, prepend: true, if: -> { SiteConfig.mailchimp_newsletter_enabled? }
```
citizen428, line 262: It seems this is the only place we need this method. I wonder if you want to remove it again from `SiteConfig` and instead update `unsubscribe_from_newsletters` with a guard clause:

```ruby
def unsubscribe_from_newsletters
  return unless SiteConfig.mailchimp_api_key.present? && SiteConfig.mailchimp_newsletter_id.present?
end
```

I'm not actually sure if that's better, but at least it's more self-contained.

app/services/mailchimp/bot.rb
citizen428: `has_role?` accesses the DB, so if you add this extra guard, I think you should add it before the role check, not after.

app/services/mailchimp/bot.rb
```diff
@@ -91,7 +91,7 @@ def manage_community_moderator_list
     end
 
     def manage_tag_moderator_list
-      return false unless user.tag_moderator?
+      return false unless user.tag_moderator? || SiteConfig.mailchimp_tag_moderators_id.blank?
```
citizen428, line 94: Same as above, `tag_moderator?` also accesses the DB.

app/models/user.rb
```diff
@@ -497,7 +497,7 @@ def profile_image_90
   end
 
   def unsubscribe_from_newsletters
-    return if email.blank?
+    return if email.blank? || (SiteConfig.mailchimp_api_key.blank? && SiteConfig.mailchimp_newsletter_id.blank?)
```
citizen428, line 500: Apart from this linme being too long, I think I'd prefer two distinct guard clauses here:

```ruby
return if email.blank?
return if SiteConfig.mailchimp_api_key.blank? && SiteConfig.mailchimp_newsletter_id.blank?
```

citizen428 (approved):
Thanks for making the previously suggested changes @Zhao-Andy! I have one more minor suggestion, but it's definitely non-blocking.

benhalpern (approved):

PR #12335, opened 2021-01-19
Added tag's color in header of tag page selected

### LINKED ISSUE #12286: Updating a tag's colors does not reflect on the tag's page
**Describe the bug**

When I attempt to update the colors for a tag, the colors do not update on the tag's show page.

**To Reproduce**

1. Make sure that you are a Forem admin with _tag resource privileges_.
2. Create a new tag, and go to that tag's edit page (`/admin/tags/1/edit`, for example).
3. Change the colors of the tag at the bottom of the form (`Bg color hex` and `Text color hex`).
4. Save your changes!
5. Go to the tag's show page (`/t/my-new-tag`).
6. Observe that the colors you selected for your new tag are not showing up 😕 

**Note:** If you try _using_ that tag on a _post_, you will notice that the colors actually _are_ correct...so this seems specific to the tag's show page. I suspect that this has to do with how the bg color hex and text color hex are cached on the tag's show page. If I had to guess, I'd say that the cache is not being busted when it should be 😅 

**Expected behavior**

I would expect that when I update a tag, those updates are reflected whenever I look at the tag.

**Screenshots**

Here are some screenshots of me reproducing this on `forem.team` with the `#planning` tag.

1) Here are the colors I chose for my tag:
Screen Shot 2021-01-14 at 12 09 01 PM

2) My chosen colors successfully seem to save and update the tag:
Screen Shot 2021-01-14 at 12 09 11 PM

3) When I go to the tag's show page, those colors are not rendering:
Screen Shot 2021-01-14 at 12 09 20 PM

4) But when I look at _a post_ that uses my tag, the colors do render correctly and as expected:
Screen Shot 2021-01-14 at 12 09 31 PM

joaoGabriel55:

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Design fix at tag's show page for show tag's color at page header.

Closes #12286

### Before
image

### After
image

Any complex UI changes.

## Added tests?

- [X] Yes
      have not been included_

## Added to documentation?

- [ ] README
- [X] No documentation needed

alt_text

### DIFF
--- a/app/views/articles/_single_story.html.erb
+++ b/app/views/articles/_single_story.html.erb
@@ -65,7 +65,15 @@
              tags_to_display = tags_to_display.reject { |tag| tag == flare_tag.name }
            end %>
         <% tags_to_display.each do |tag| %>
-        <a href="/t/<%= tag %>" class="crayons-tag"><span class="crayons-tag__prefix">#</span><%= tag %></a>
+          <% if tag === @tag_model.name %>
+            <a href="/t/<%= tag %>" class="crayons-tag" style="background:<%= @tag_model.bg_color_hex %>;color:<%= @tag_model.text_color_hex %>">
+              <span class="crayons-tag__prefix">#</span><%= tag %>
+            </a>
+          <% else %>
+            <a href="/t/<%= tag %>" class="crayons-tag">
+              <span class="crayons-tag__prefix">#</span><%= tag %>
+            </a>
+          <% end %>
         <% end %>
       </div>
       <div class="crayons-story__bottom">
--- a/app/views/articles/_single_story.html.erb
+++ b/app/views/articles/_single_story.html.erb
@@ -65,7 +65,7 @@
              tags_to_display = tags_to_display.reject { |tag| tag == flare_tag.name }
            end %>
         <% tags_to_display.each do |tag| %>
-          <% if tag === @tag_model.name %>
+          <% if tag == @tag_model.name %>
             <a href="/t/<%= tag %>" class="crayons-tag" style="background:<%= @tag_model.bg_color_hex %>;color:<%= @tag_model.text_color_hex %>">
               <span class="crayons-tag__prefix">#</span><%= tag %>
             </a>
--- a/app/views/articles/tag_index.html.erb
+++ b/app/views/articles/tag_index.html.erb
@@ -13,14 +13,12 @@
         <div class="flex flex-col w-100 justify-center">
           <div class="flex justify-between items-center">
             <h1 class="crayons-title">
-              <span class="crayons-tag crayons-tag--2xl" id="crayons-tag" style="background:<%= @tag_model.bg_color_hex %>;color:<%= @tag_model.text_color_hex %>;">
-                <% if @tag_model && @tag_model.pretty_name.present? %>
-                  <%= @tag_model.pretty_name %>
-                <% else %>
-                  <span class="opacity-10">#</span>
-                  <%= @tag %>
-                <% end %>
-              </span>
+              <% if @tag_model && @tag_model.pretty_name.present? %>
+                <%= @tag_model.pretty_name %>
+              <% else %>
+                <span class="opacity-50">#</span>
+                <%= @tag %>
+              <% end %>
             </h1>
             <% if @tag_model %>
               <button id="user-follow-butt" class="crayons-btn follow-action-button" data-info='{"id":<%= @tag_model.id %>,"className":"Tag"}'>Follow</button>
--- a/app/views/articles/tag_index.html.erb
+++ b/app/views/articles/tag_index.html.erb
@@ -13,12 +13,14 @@
         <div class="flex flex-col w-100 justify-center">
           <div class="flex justify-between items-center">
             <h1 class="crayons-title">
-              <% if @tag_model && @tag_model.pretty_name.present? %>
-                <%= @tag_model.pretty_name %>
-              <% else %>
-                <span class="opacity-50">#</span>
-                <%= @tag %>
-              <% end %>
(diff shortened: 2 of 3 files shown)


ludwiczakpawel (changes requested):
Hi!

Thank you for this PR, definitely a good contribution! This time though I don't think it's exactly what we're looking for in terms of design implementation. I agree we could do much better job in expressing the tag's brand on the tag page, but I'd leave the page title as it is and focus on that colored stripe on top of the header section (I'll talk about it later on).

So usually a tag has two colors, theoretically for _background_ and _text_. And this is pretty much how show certain tags on an article page: we apply a background color and a text color to the tag "cell":

image

But on a tag page itself we intentionally try to be less "loud" in visual representation of a tag. Instead of showing similar cell, we shifted to add colored stripe on top of the header:

image

For some tags it works (looks) better and for some - worse. And I think that's the area where I'd like to focus efforts on fixing - I believe [our code could be a little bit "smarter" when it comes to choosing right color for that stripe](https://github.com/forem/forem/blob/master/app/views/articles/tag_index.html.erb#L8) - I'm not exactly sure how to achieve that but I'm sure this function could be improved - I think it simple picks wrong color sometimes - likely because some of the contrast checking assumptions we're making..

**Overall it's not just because we want it that way. We plan to make this entire section much "richer" in terms of UI and content - that's why we don't want to make specific elements in that area too heavy visually - that would happen if we decided to use the "colored cell" you proposed.**

image

lisasy (changes requested):
I am also echoing what @ludwiczakpawel wrote thoroughly here: https://github.com/forem/forem/pull/12335#pullrequestreview-571580412

joaoGabriel55:
Hello @ludwiczakpawel.

Thanks a lot for your feedback. It helped me to understand better how you define your design guidelines.

vaidehijoshi (changes requested):
@joaoGabriel55 hi there! I noticed in the PR description that you wrote:
> Design fix at tag's show page for show tag's color at page header.

But I don't actually think that this bug requires a design fix -- I suspect it actually has to do with caching.

In the original issue description, I had added this note:
> **Note:** If you try _using_ that tag on a _post_, you will notice that the colors actually _are_ correct...so this seems specific to the tag's show page. I suspect that this has to do with how the bg color hex and text color hex are cached on the tag's show page. If I had to guess, I'd say that the cache is not being busted when it should be 😅 

So I think the approach here actually should be focused on the cache key (and doesn't need to be a design-related change):
https://github.com/forem/forem/blob/8d462801ece4a866a4aaf4a3844759e23d88d866/app/views/articles/tag_index.html.erb#L6

joaoGabriel55:
Ok, I will review my solution. I think now I understand better.

Thanks a lot for your feedback.

joaoGabriel55:
Hello everyone! @vaidehijoshi, before to send the PR. I just want to show the preview
image

Would be this?

vaidehijoshi:
> Hello everyone! @vaidehijoshi, before to send the PR. I just wanted to show a preview

@joaoGabriel55 would you mind pushing up your commits/code changes? It will be much easier to review the code thoroughly rather than try to guess the changes from a screenshot :) Thanks!

joaoGabriel55:
Hello @vaidehijoshi and @ludwiczakpawel this last commit is about second image preview. 

Waiting your feedback.

cmgorton:
Hey @joaoGabriel55  Thanks so much for the work you have done so far. 
Unfortunately, I think we may be going off track on what  issue #12286 was originally intended for. 
If we want to focus on design changes I think that would be a separate issue we can discuss and decide if it should be implemented.

For the [issue](https://github.com/forem/forem/issues/12286) Vaidehi originally posted we were looking for a bug fix in response to caching. It might help to read the original issue again and then decide if this is something you want to work on. 
Please let us know if you have any questions. @fdoxyz and @atsmith813 are on OSS rotation this week so you can ping them for any reviews. 
Thanks!!

joaoGabriel55:
Hello Everyone. I analysed this erb file here:

`app/views/articles/_single_story.html.erb`

And, at 63 line the array used to show post's tags is this:

`story.cached_tag_list_array`

A cached tag list with only tag name, without bg_color_hex and text_color_hex properties. I guess that is the real problem about the tag updated and others tags does not are showing their colors.

Waiting your feedback.

app/views/articles/_single_story.html.erb
```diff
@@ -65,7 +65,15 @@
              tags_to_display = tags_to_display.reject { |tag| tag == flare_tag.name }
            end %>
         <% tags_to_display.each do |tag| %>
-        <a href="/t/<%= tag %>" class="crayons-tag"><span class="crayons-tag__prefix">#</span><%= tag %></a>
+          <% if tag === @tag_model.name %>
```
wenderjean, line 68: I wonder if we really need that triple equals condition here, I mean, It would make sense if the `tag` was a regex and `@tag_mode.name` was a text. I don't think that's the case here.

```ruby
/testing/ === 'Hey, I am just testing...' => true
```

joshpuetz (approved):
Verified this works locally for stories "above the fold". For those below the fold, there's an existing conversation covering tag colors not working: https://github.com/forem/forem/pull/10528#issuecomment-703515421

So: 👍🏻 from me!

atsmith813:
Hey @joaoGabriel55, thank you so much for this insight!

I did a deeper dive into this issue based on what you've done here. I wrote up some discoveries in [a comment on the original issue](https://github.com/forem/forem/issues/12286#issuecomment-767828348).

The short of it is, it turns out the original issue isn't quite an issue - things are working as expected. And the rest is sort of expected behavior - it wasn't designed to be updated right away.

As for the next steps, I'll likely break up the original issue/my comment into smaller, more defined, individual issues sometime this week.

I'll close this PR for now because the issues are going to be redefined with more clarity. Please let me know if you have any questions and feel free to jump in the new issues once we get them up. Thanks again, we may not have discovered these quirks anytime soon without your help!

PR #12340, opened 2021-01-20
Remove OSS team from CODEOWNERS file and add backend team

mstruve:
- [x] Documentation Update

## Description
Since we no longer have a dedicated OSS team and will be working off of our rotation I am removing the OSS team from Github. This means we will no longer use it via CODEOWNERS so I have removed it and replaced it in a couple of places with the Backend team. 

I thought about trying to update the team every week with the new people on the rotation but that seemed like it would be too much manual work and that I would eventually start missing or forget so I decided to remove it instead. Let me know if anyone else has thoughts or suggestions regarding this! 

alt_text

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -10,7 +10,7 @@
 /app/workers/                                       @forem/sre
 /config/                                            @forem/sre
 /db/                                                @forem/sre @forem/backend
-/docs/                                              @forem/oss    @jacobherrington
+/docs/                                              @jacobherrington
 /lib/data_update_scripts/                           @forem/sre
 /lib/sidekiq/                                       @forem/sre
 /spec/rails_helper.rb                               @forem/sre @forem/backend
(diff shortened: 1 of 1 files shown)


atsmith813 (approved):

citizen428 (approved):

citizen428:
> Let me know if anyone else has thoughts or suggestions regarding this!

I think this is fine. I don't feel with our new structure the team is useful enough to spend any time on trying to automate this.

.github/CODEOWNERS
```diff
@@ -9,20 +9,20 @@
 /app/services/search/                               @forem/sre
 /app/workers/                                       @forem/sre
 /config/                                            @forem/sre
-/db/                                                @forem/sre
+/db/                                                @forem/sre @forem/backend
 /docs/                                              @forem/oss    @jacobherrington
```
rhymes, line 13: missed one :)

rhymes (approved):
Agreed! Left one note

.github/CODEOWNERS
mstruve, line 13:
```suggestion
/docs/                                              @jacobherrington
```

mstruve: hiding in plain sight 😂

Issue #11991, opened 2020-12-21
Move validation from Admin::ConfigsController to SiteConfig model


rhymes:
**Is your feature request related to a problem? Please describe.**

Currently the validation logic of various `SiteConfig` fields sits in the [Admin::ConfigsController](https://github.com/forem/forem/blob/master/app/controllers/admin/configs_controller.rb), for example: 

https://github.com/forem/forem/blob/3c297ff05eafc9a18e71c62bd184accfd1b54438/app/controllers/admin/configs_controller.rb#L277-L280

but since version 2.5 the `rails-settings-cached` gem has added Rails validation to fields, see https://github.com/forem/forem/pull/11910 and https://github.com/huacnlee/rails-settings-cached#validations

**Describe the solution you'd like**

Move as many validations over to the model as possible, this will ensure that direct assignment also goes through the required validation and we never leave the data in an invalid state

**Describe alternatives you've considered**

Nope, no alternatives

citizen428:
@rhymes It seems the gem only wraps `validates` (i.e. the default Rails validations) but not `validate` which allows specifying a custom method/block? In that case, we may have to define custom validator classes for some of the validations.

rhymes:
@citizen428 yeah, it doesn't look to be that advanced

citizen428:
Also note that since this issue was originally written, the validations moved from `Admin::ConfigsController` to the `SiteConfigs::Upsert` service.

vaidehijoshi:
Since this was moved to done, are we good to close this issue out? Want to confirm before I do anything in case there was more work to be done in a follow up PR or something :)

citizen428:
@vaidehijoshi Yes, this is done, apparently, I forgot to add "closes" in the PR description so the issue wasn't automatically closed.

PR #12341, opened 2021-01-20
Add validations to SiteConfig model

citizen428:

- [X] Refactor

## Description

Until recently the [`rails_settings_cached` gem](https://github.com/huacnlee/rails-settings-cached) did not support validations, so we performed them explicitly, first in `Admin::ConfigsController` and later in the `SiteConfigs::Upsert` service object. 

This PR moves validations into the model itself. Note that the gem only supports the `validates` option, but not custom validation methods  `validate`, so I had to add a few new validator classes to `app/validators`.

#11991 

Screenshots of the different validation errors:

image
image
image
image

Note: the hex color validation is hard to trigger from the UI as we are using a color input.

For QAing this, go to `/admin/config` and try setting invalid and valid values respectively.

## Added tests?

- [X] Yes
- 
## Added to documentation?

- [X] No documentation needed


app/models/site_config.rb
```diff
@@ -4,14 +4,15 @@
 class SiteConfig < RailsSettings::Base
   self.table_name = "site_configs"
 
-  validates :var, presence: true
```
citizen428, line 7: This has me stumped a bit. While this line works perfectly fine on `master` it breaks on this branch, even though the gem version is the same and no relevant code was changed.

Master:
image

This branch:
image

The column itself already has a non-null constraint, so it's not possible to insert an invalid value. We also never directly interact with `SiteConfig` records but only do so through the gem's interface. 

I disabled this in order to not get stuck in a debugging rabbit hole for a validation that only seems marginally useful to me, but I'm happy to dig deeper if people feel strongly about it.

app/validators/color_contrast_validator.rb
```diff
@@ -0,0 +1,11 @@
+class ColorContrastValidator < ActiveModel::EachValidator
```
rhymes, line 1: ❤️

rhymes (approved):
Wow! So happy they added this and thanks for the great refactoring.

Let's get this merged :D

app/models/site_config.rb
citizen428: For the record, @rhymes and I discussed this on Slack and agreed that this validation can be safely removed.

Ridhwana (approved):
This looks so much better! 😍

